Issue #1324, opened 2017-08-08
Allow negative points in autograding test case


bmcutler:
Use case:

one test case may check for hardcoding, or use of forbidden technique and want to apply a large negative score penalty that effectively zeros out points from other test cases.

scopeInfinity:
Hi @bmcutler, can you please clarify my doubts regarding scoring rules.
Is it allowed to have total score as negative for any case? If yes, which among late submission penalty, negative score testcases or both can contribute to total negative score?

Reference Image
![image](https://user-images.githubusercontent.com/9819066/37541935-59529dc6-2982-11e8-8712-d37cd6a7d366.png)

bmcutler:
Make sure that when penalty autograding is implemented we test & revise the short-circuited submission limit test case implementation from PR #5527 to ensure correct math.

PR #1815, opened 2018-03-16
Reinitialize negative penalty to produce total_pts as zero

scopeInfinity:
Issue: #1324 
Configuration under `config.json`
```json
    "title" : "Forbidden Technique",
    "points" : -100
```
**Expected Output**
If _total_pts_ become negative because of heavy penalties, then reinitialize the negative alloted penalty scores such that _total_pts_ become _zero_. 

**Screenshot**
Result of normal sample submission
forbiddentechnotused

Result of sample submission by using forbidden technique 
forbiddentechused

### DIFF
--- a/grading/main_validator.cpp
+++ b/grading/main_validator.cpp
@@ -16,6 +16,7 @@
 #include <vector>
 #include <string>
 #include <algorithm>
+#include <tuple>
 
 #include "TestCase.h"
 #include "default_config.h"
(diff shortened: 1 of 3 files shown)


scopeInfinity:
Thanks @tcyrus

tcyrus:
The solution is not to change the test

bmcutler:
Thanks for the contribution, but we're going to tackle this in a future refactor.

PR #1820, opened 2018-03-20
Sanitize form input fields

ndixita:
As per the discussion over email, I am adding the fix for XSS bug to prevent injection of malicious scripts. The video attachments for replication of the bugs have been sent over email to @andrewaikens87. The fix involves sanitization of the input fields using FILTER_SANITIZE_STRING switch.


ndixita:
Sure @andrewaikens87. Thank you for your review, modifying it accordingly.

andrewaikens87:
Thank you. Also, for the URL notice that this solution will not handle something with a javascript protocol. Although, since this is the Instructor page this is highly unlikely for an instructor to insert the URL as this. We have code that checks for this and handles it accordingly in the discussion forum. Ideally that code would be utilized anywhere we have a user inserting a url (and if it is a bad url, one that has illegal characters or something of that sort, it would notify the user).  But, the code needs to be refactored to achieve this, and can be added in a future PR.

CouleeApps:
It looks like the correct thing to use here is `htmlentities(... , ENT_QUOTES);`

PR #1823, opened 2018-03-24
This change will make indentation stop at the 7th reply level. 

saxocellphone:
An alert will also be sent out to the author when trying to reply to a post 7 level deep.


andrewaikens87:
Here is something thing that I noticed: 

If you continue to reply to 7 posts it stops the indentation which is good. But, take a look at the screenshot below (I apologize it is small I had to zoom out to show the hierarchy):
bad ordering

Specifically, look at posts "7 -- 1" and "6 -- 1", as you can see they are out of order (take a look at the timestamps). If we get to the point where we are no longer indenting we should be showing posts based on timestamp (as there is no longer a reply to post hierarchy). 

Note: "7 -- 1" is a reply to post with "7" and "6 -- 1" is a reply to post with "6".

If I go further with replying to post "8" it will give me this ordering:
bad order2

saxocellphone:
If we are going to switch to chronological order after 7th level, can we simply just remove the reply button for all the posts after 7

bmcutler:
My proposal was to allow unlimited reply depth, the database has the correct tree structure for unlimited replies, the order of the posts is the same at every reply depth (don't switch to chronological at some point), the only difference is that we set the indentation to zero.

andrewaikens87:
I'm sorry, when I mentioned there was no longer a reply to post hierarchy I was referring to visually with indentation. The hierarchy should still be preserved in the database. Since we aren't visually displaying any hierarchy after this level, it might be confusing having a post being inserted under their parent post without a way to distinguish the parent post (who the user is replying to). I think that using quoting or tagging would be a good approach to solve this and I know that @saxocellphone has mentioned this to me as well. We should discuss more on Wednesday.

MasterOdin:
I'd ask in our current courses, how deep are reply chains getting? 

I do agree with @andrewaikens87 that not indenting past 7 levels but allowing for posts to reply as viewing it is not very intuitive as you're mixing a hierarchal view and a chronological view at the same time and I'm not convinced a non-power user will get what's going on.

Issue #1827, opened 2018-03-25
Even More Forum Requests


bmcutler:
Moved to #2201
~~the size of the reply box for a reply to a non-top level post is currently fixed.  Please make this box auto-expand if the typed reply is lengthy.  Note: the default reply (to the top level post) is re-sizeable.~~

- [x] allow the viewer to collapse/uncollapse a subtree within the thread.  (This will be helpful if there is a lengthy side discussion with many reply-tos.)  If a subtree is collapsed, we should display the count of the # of posts that are hidden/collapsed.

- [x] Add a "reply" button to the top post, to match the other "reply" buttons.  Design proposal:  Pressing this button scrolls/jumps you to the bottom of the page where the reply-to-top-post box already exists.  

- [x] Currently after you finish typing your reply and submitting the reply, the page scrolls back to the top.  Instead, the page should refresh positioned at your posted reply.

- [ ] A "question" post should have a status "open" or "resolved" (design discussion needed...)

@saxocellphone please work with @andrewaikens87 to prioritize and divide up your work on the forum todo list.

PR #1828, opened 2018-03-26
User can now collapse posts.

saxocellphone:
One of the elements of #1827, can possibly add animation, but that's up for discussion.


andrewaikens87:
Overall pretty good, here are some of the initial things I noticed. Take a look at the screenshots below:

Take for example a thread that looks like this:
collapseexample

If we collapse the third reply it works well.
onecollapsed

But notice if we then try to collapse the second post, it breaks, displaying the fourth post again.
second collapse

Also, we probably should not be displaying the `[-]` unless it has functionality. So for a post that doesn't have any replies to it, it probably shouldn't have a `[-]` as it won't have any functionality. Also take a look at [https://fontawesome.com/icons?d=gallery&m=free](https://fontawesome.com/icons?d=gallery&m=free), it might be nice to use this for the +/-. I can take a look at and provide more specific feedback on the code once this is resolved.

saxocellphone:
I tried to recreate the bug, but I can't seem to do it. Here is what I have
capture

MasterOdin:
I think I'd like to add a new CSS class to handle the brackets in the text:
```
.brackets {
    text-indent: 0px;
    white-space: nowrap;
}
.brackets:before {
    content: "[";
}
.brackets:after {
    content: "]";
}
```
to make it easier for us to change this in the future (if we want) or apply this similar setup elsewhere.

andrewaikens87:
I stress tested it and couldn't make the collapsing break so I think we are good as it now has the button. But before we merge did we agree on "hide children" and "[show NUM]" for the button text? It might be more appropriate to do "Hide replies" and possibly "Show replies (NUM)" (the parenthesis can be interchanged with brackets). Also should the button in the instructor view be on the right of the x and edit buttons? What do you think?

positioning

PR #1829, opened 2018-03-26
This change adds a reply button for top-level posts

saxocellphone:
3rd requested feature of #1827


site/app/views/forum/ForumThreadView.php
```diff
@@ -47,7 +47,10 @@ function openFile(directory, file, path ){
 
 			$( document ).ready(function() {
 			    enableTabsInTextArea('post_content');
-			    saveScrollLocationOnRefresh('thread_list');
+				saveScrollLocationOnRefresh('thread_list');
+				var previous_scroll_position = localStorage.getItem("scroll_position");
```
andrewaikens87: We should be modifying saveScrollLocationOnRefresh in order to be able to save multiple scroll bar positions in sessionStorage, take a look at this [https://stackoverflow.com/a/5523174](https://stackoverflow.com/a/5523174), it explains the difference between localStorage and sessionStorage.

Within the function you can do something similar to: `sessionStorage.setItem(id + '_scrollTop', $(element).scrollTop());` to set the unique scroll position. Where `id` is the div id, so the first line should also be changed to use the div id not the class name. Then modify the set line to retrieve the correct value from sessionStorage. As a result of this, `resetScrollPosition()` will have to be modified to clear the saved locations on a redirect, and called in the appropriate location.

site/public/js/server.js
```diff
@@ -943,9 +943,14 @@ function saveScrollLocationOnRefresh(className){
             $(element).scrollTop(sessionStorage.scrollTop);
         }
     });
+    var previous_scroll_position = sessionStorage.getItem("scroll_position");
+    document.getElementById('posts_list').scrollTop = previous_scroll_position;
+    sessionStorage.setItem("scroll_position", 0);
```
andrewaikens87: These three lines are essentially doing what the function is doing. The function should be changed so that you can call  `saveScrollLocationOnRefresh('posts_list')` in addition to `saveScrollLocationOnRefresh('threads_list')` and it will save the location of the scrollbar in sessionStorage for the both of them. The function should take in the value of the id that it wants to save the location of the scroll bar for (this function should work for any id that you give it). I will post a code segment in a comment on the PR.

saxocellphone: Sorry I totally missed that part of your comment. I think my newest commit fixes the problem.

PR #1756, opened 2018-02-22
Write out submitty details to conf folder

MasterOdin:
Closes #1750 (though leaves the files).

Here's the permissions of these new files:
```
root@vagrant:/usr/local/submitty# ls -ld /usr/local/submitty/conf
drwxr-x--x 2 root course_builders 4096 Feb 21 23:37 /usr/local/submitty/conf
root@vagrant:/usr/local/submitty# ls -la /usr/local/submitty/conf
total 20
drwxr-x--x  2 root course_builders 4096 Feb 21 23:37 .
drwxr-x--x 15 root course_builders 4096 Feb 21 23:37 ..
-rw-r-----  1 root root             169 Feb 21 23:37 autograding_workers.json
-r--r-----  1 root hwcron           133 Feb 21 23:37 database_conf.json
-r--r-----  1 root course_builders 1460 Feb 21 23:37 submitty_conf.json
```

I also moved autograding_workers.json into this new folder as it's not currently hooked up to anything and I'd like that file to just be in the proper place to start with.

Made also improvement on checking for existance and read/write of the install/data directories, as well as make it so that we can change these directories by passing in different runtime arguments to CONFIGURE_SUBMITTY.py.

PR #1833, opened 2018-03-27
Timezone Fix

### LINKED ISSUE #1745: Submission timezone not honored
Our server is set to PST and I set the submission deadline to midnight PST (GMT - 8). Yet the submission time shows as three hours later and the students are told that their submissions are late if they are after 9pm.

RylanOC:
Closes #1745 

The timezone is no longer changed in `install_system.sh`, and instead only ever modified in `Vagrantfile`.

PHP doesn't automatically inherit the timezone from the server, so any scripts that deal with time sensitive calculates needed to be updated to read from `/etc/timezone`.

Finally, `config.php` (`site/app/models/Config.php`) was hardcoded as 'America/New_York", and needed to be updated to keep the site from crashing when other timezones are specified in different PHP scripts.


MasterOdin:
This will put a hard requirement of that you HAVE to use Ubuntu/Debian for Submitty, else this will not work and I'm not convinced that's where we want to be going.

Additionally, `site/app/models/Config.php` was not hardcoded to always be `America/New_York`, that's just what the default value was (as it's where we are), but it's overwritten here:
https://github.com/Submitty/Submitty/blob/c859164f49e6d49b482ded534fec843836d15730/site/app/models/Config.php#L223-L228

which will (currently) always catch as the `site/config/master_template.ini` file has this:
https://github.com/Submitty/Submitty/blob/8788391ff982704184181c4bb7c1d64073c9ac50/site/config/master_template.ini#L9

If anything, I'd prefer that we add a question within `.setup/CONFIGURE_SUBMITTY.py` that asks what timezone you want and then writes that to the to the `master.ini` file which PHP then reads and then parses all incoming datetimes into that specific timezone.

site/app/views/NavigationView.php
```diff
@@ -5,6 +5,14 @@
...
 use app\models\Gradeable;
 
+$default_zone = "America/New_York";
+$timezone = file_get_contents("/etc/timezone");
+if (!strlen($timezone)) {
+    $timezone = $default_zone;
+}
+$timezone = trim($timezone);
+date_default_timezone_set($timezone);
+
```
MasterOdin: Unnecessary as we're already setting the `date_default_timezone_set` within `site/public/index.php`.

You also shouldn't introduce 'side-effects' when loading a file that has a class in it per [PSR-1](https://www.php-fig.org/psr/psr-1/#23-side-effects).

MasterOdin:
So I believe the consensus we had was that within CONFIGURE_Submitty.py we write out the timezone to a config file and that we then reference that config file anywhere we're writing or displaying a datetime. That way if someone reconfigures the system (or re-runs install_system.sh), that'll pick up the changes to timezone for them.

@bmcutler, @RylanOC, if that's what we're good on doing, I'll add the necessary field to a file inside the conf file inside of the #1756 PR, as it should just be accomplished by doing:
```
import tzlocal
tzlocal.get_localzone().zone
```
which will give us a string that can be used within PHP or Python.

Docs/student_auto_feed/accounts.php
```diff
@@ -53,17 +53,9 @@
  * -------------------------------------------------------------------------- */
 
 //Univeristy campus's timezone.
-date_default_timezone_set('America/New_York');
+$core = new Core();
+date_default_timezone_set($core->getConfig()->getTimezone()->getName());
```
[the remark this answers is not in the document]
pbailie: _At minimum_, if we don't set the timezone, PHP throws a lengthy warning about reliability.

Docs/student_auto_feed/config.php
```diff
@@ -36,7 +38,9 @@
  * -------------------------------------------------------------------------- */
 
 // Univeristy campus's timezone.  ***THIS NEEDS TO BE SET.
-date_default_timezone_set('America/New_York');
+$core = new Core();
+date_default_timezone_set($core->getConfig()->getTimezone()->getName());
```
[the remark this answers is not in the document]
pbailie: `config.php` runs on a separate server maintained by labstaff and supports the nightly auto feed process.  There is a subtle flaw in PHP, as specifically compiled by Canonical, that makes remote file access to the feed CSV undesirable.

`accounts.php` is run on the Submitty server, but it resides with other sysadmin tools in a place that requires root access -- so not in the same place as the Submitty site code.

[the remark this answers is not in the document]
MasterOdin: Based off this, I'd just leave it setting `date_default_timezone_set('America/New_York');` as setting that whole thing up is documented elsewhere and not try and do anything with `Core` or Submitty at large here.

site/app/models/Config.php
```diff
@@ -95,7 +95,7 @@ class Config extends AbstractModel {
     /** @property @var string */
     protected $authentication;
     /** @property @var string */
-    protected $timezone = "America/New_York";
+    protected $timezone = 'America/New_York'; //Default to EST
```
MasterOdin: Remove comment as it's not really telling anything you can't just glean from looking at the default value. It's also wrong half the year when 'America/New_York' stands for EDT timezone (like right now).

site/app/views/NavigationView.php
```diff
@@ -1,10 +1,10 @@
 <?php
-
 namespace app\views;
-
 use \app\libraries\GradeableType;
+use app\models\Config;
 use app\models\Gradeable;
-
+use app\libraries\Core;
+//date_default_timezone_set($this->core->getConfig()->getTimezone());
```
MasterOdin: remove this commented out line. We'll never need to do this here. 

We can also probably get rid of the
```
use app\models\Config;
use app\libraries\Core;
```
lines.

site/public/index.php
```diff
@@ -106,7 +106,9 @@ function error_handler() {
     $core->getOutput()->addBreadcrumb("", $core->getConfig()->getCourseHomeUrl(),false, true);
 }
 
+echo $core->getConfig()->getTimezone()->getName();
```
MasterOdin: Remove this after done demoing this.

RylanOC:
@MasterOdin @pbailie 
Thanks for your feedback, I think I've made all the fixes that you requested. If you spot any other issues I'd love to know.

Docs/student_auto_feed/config.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use app\libraries\Core;
```
MasterOdin: Remove this

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -133,7 +131,8 @@ def get_input(question, default=""):
             'authentication_method': 1,
             'institution_name' : '',
             'username_change_text' : 'Submitty welcomes individuals of all ages, backgrounds, citizenships, disabilities, sex, education, ethnicities, family statuses, genders, gender identities, geographical locations, languages, military experience, political views, races, religions, sexual orientations, socioeconomic statuses, and work experiences. In an effort to create an inclusive environment, you may specify a preferred name to be used instead of what was provided on the registration roster.',
-            'institution_homepage' : ''}
+            'institution_homepage' : '',
+            'timezone' : tzlocal.get_localzone()}
```
MasterOdin: you probably want to do `tzlocal.get_localzone().zone()` here as `tzlocal.get_localzone()` returns a TzInfo object. Example:
```
>>> import tzlocal
>>> tzlocal.get_localzone()
<DstTzInfo 'America/New_York' LMT-1 day, 19:04:00 STD>
>>> tzlocal.get_localzone().zone
'America/New_York'
```

MasterOdin:
@RylanOC I fixed a typo in that function call that I gave you (zone is a property, not a method, whoops!).
@bmcutler this looks good to merge in my opinion

PR #1837, opened 2018-03-28
Ta grading view redesign

shailpatels:
Updates TA grading on the student's view from a pre-formatted text file. Displays relevant late day information along with published marks. Does not display grader's names if peer grading assignment.

Grading with overall comment/no extra credit
hwview1

Grading with no overall comment/extra credit
hwview2

Updated marks from stars and parentheses to Font Awesome icons.
newchecks


bmcutler:
Small bug fix needed...  
When I look at the grades in Mac Chrome or Mac Safari or Mac Firefox the top boxes for graded by & late day info are placed wonky.  If I make the window very wide it goes away, but if it's medium or small sized the placement is buggy.  See below:

screen shot 2018-04-11 at 4 45 57 pm

shailpatels:
Should work on smaller screen sizes now :)
capture

site/app/views/AutoGradingView.php
```diff
@@ -459,12 +459,12 @@ public function showTAResults(Gradeable $gradeable){
         $lateDayData = $ldu->getGradeable($gradeable->getUser()->getId(), $gradeable->getId());
         $return = <<<HTML
         <div class = "sub">
-            <div class="box half" style="padding: 10px;">
+            <div class="box half" style="padding: 10px; width: 40%;">
```
MasterOdin: is this styling that should be applied to the half class?

shailpatels: I'm not sure if I understand what you mean, I added the width restriction on the right side box to prevent the left box from sitting on top of the left.

bmcutler: Where is the box half class defined?  It's not defined in our source code is it?
Other uses of box half are buggy as well.   It's not important that it get solved elegantly now.  
This is not the final polished layout/design.

shailpatels: Its defined in server.css, its two classes, box , and half. And yea I see what Matt is saying, it seems redundant to have that there since both the float and width are being overwritten by the style tag but yea this is a quick edit that can be added with later updates (since peer grading/new ta grading interface will borrow from this probably)

MasterOdin: I'd like to improve and fix the styling for those two classes as this is a common style usage and the idea behind these was that you should be able to just have two divs with them without requiring additional padding/floats/widths to get the two divs to display next to each other.

PR #1859, opened 2018-04-11
Forum stats page

thatCode-Psy:
Added start of the stat's page

### DIFF
--- a/site/app/views/forum/ForumThreadView.php
+++ b/site/app/views/forum/ForumThreadView.php
@@ -88,6 +88,7 @@ HTML;
 		<div style="margin-left:20px;margin-top:10px; height:50px;  " id="forum_bar">
 
 			<a class="btn btn-primary" style="border:3px solid #E9EFEF" title="Create thread" onclick="resetScrollPosition();" href="{$this->core->buildUrl(array('component' => 'forum', 'page' => 'create_thread'))}"><i class="fa fa-plus-circle"></i> Create Thread</a>
+			<a class="btn btn-primary" style="border:3px solid #E9EFEF" title="Show Stats" onclick="resetScrollPosition();" href="{$this->core->buildUrl(array('component' => 'forum', 'page' => 'show_stats'))}">Stats</a>
 			
 		</div>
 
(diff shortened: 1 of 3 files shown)


site/app/views/forum/ForumThreadView.php
```diff
@@ -612,4 +625,86 @@ public function createThread() {
...
 
+
+	public function statPage($users) {
+
+		$return = <<<HTML
+		
+		<div style="margin-left:20px;margin-top:10px; height:50px;" id="forum_bar">
+
+			<a class="btn btn-primary" style="border:3px solid #E9EFEF" title="Back to threads" href="{$this->core->buildUrl(array('component' => 'forum', 'page' => 'view_thread'))}"><i class="fa fa-arrow-left"></i> Back to Threads</a>
+			<a class="btn btn-primary" style="border:3px solid #E9EFEF" title="Show Stats" onclick="resetScrollPosition();" href="{$this->core->buildUrl(array('component' => 'forum', 'page' => 'show_stats'))}">Stats</a>
```
andrewaikens87: Can we remove this button since we are already on the page that it redirects to (we are already on the stats page)?

site/app/views/forum/ForumThreadView.php
```diff
@@ -612,4 +625,86 @@ public function createThread() {
...
+		
+		</div>
+			<div class="content">
```
andrewaikens87: There is a large gap between the "back to threads"/"Stats" buttons and the content box when viewing the list. Can we make it the same distance as the normal thread view?

site/app/views/forum/ForumThreadView.php
```diff
@@ -612,4 +625,86 @@ public function createThread() {
...
+				$("button").click(function(){
+					var action = $(this).data('action');
+					var posts = $(this).data('posts');
+					var ids = $(this).data('id');
+					var timestamps = $(this).data('timestamps');
+					if(action=="expand"){
+						
+						//console.log(posts);
+						for(var i=0;i<posts.length;i++){
+							$(this).parent().append('<tr id="'+ids[i]+'"><td>'+timestamps[i]+'</td><td colspan = "5" align = "right">'+posts[i]+'</td></tr> ');
```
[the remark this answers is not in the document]
andrewaikens87: Also, if there is a long post, it doesn't preserve the spacing (i.e. if the user inputs a paragraph or something) This should be modified to handle this case (with align left): `align = "left"><pre style="font-family: inherit;white-space: pre-wrap;">'+posts[i]+'</pre>` should do the trick.

thatCode-Psy:
updated with sorting and linking to threads. Also addressed the requested changes.

andrewaikens87:
Legacy posts will display as their encoded value i.e. ` A &#039;quote&#039; is &lt;b&gt;bold&lt;/b&gt `. But I believe that it is something we can talk about after this gets merged as when the instructor clicks on the post it will bring them to the corresponding thread which will have a readable post.

legacypost

andrewaikens87:
I think that it is good enough to merge currently but we should try and get some feedback.

PR #1862, opened 2018-04-12
Warning when about to lose unsaved forum post

### LINKED ISSUE #1849: Warning when about to lose unsaved forum post text
[feature request]

If you type a forum post/reply, but then click on another thread (e.g., to check something else), you will lose your unsaved forum post

It would be nice to have a popup warning that you will lose this unsaved text if you proceed with switching.

scopeInfinity:
Closes #1849 

- Used https://cdnjs.cloudflare.com/ajax/libs/jquery.AreYouSure/1.9.0/jquery.are-you-sure.min.js
  - Source: https://github.com/codedance/jquery.AreYouSure
  - License same as JQuery
- Warning Pop-up for `Create Thread` page, `Forum Threads` page and `Edit Post` form.
- Clears `#edit-user-post` on clicking `Cancel`

###### Screenshot in Chrome
Image1

###### Screenshot in Firefox
Image2


bmcutler:
@scopeInfinity or @andrewaikens87 can you resolve the merge conflict?

scopeInfinity:
Merge conflict is resolved.

Thank you, andrewaikens87 for the fix.
Regarding `editing a post and click cancel it will hide the element regardless if you changed it`,  I assumed a person will click on Cancel button consciously only, so a warning pop-up shouldn't be needed in that case. How does this behavior look to you?

andrewaikens87:
I agree, this should be merged as is and if it we notice it to be a problem we can add in the future.

PR #1874, opened 2018-04-15
Remove Unnecessary References to Grace Period

RylanOC:
`SUBMISSION_GRACE_PERIOD` and ` __SUBMISSION_GRACE_PERIOD_SECONDS__` don't seem to change the submission behavior at all, and can be removed to avoid further confusion. 

When an assignment is submitted shortly after the official due date, a javascript alert warns the student that they will be charged one late day, but no late days are actually charged.
image
image

After removing these lines, the same behavior occurs:
due_date_no_grace
late_day_use_no_grace


RylanOC:
During the meeting on Wednesday, it sounded like there was some disagreement on how the grace period should work. If the consensus is that it should be removed though, I'd be happy to look into that.

PR #1881, opened 2018-04-18
Fixed small mark issues 

LeonMontealegre:
Where clicking from one mark to another didn't save the previous ones. Also fixed visual glitch with Firefox where mark checkbox would wrap upwards. Also fixed visual issue where current points on opened mark wouldn't display correctly.

### DIFF
--- a/site/app/controllers/MiscController.php
+++ b/site/app/controllers/MiscController.php
@@ -50,7 +50,7 @@ class MiscController extends AbstractController {
                  return false;
             }
         }
-        
+
         if (!FileUtils::isValidFileName($path)) {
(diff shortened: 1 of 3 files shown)


bmcutler:
The wrap problem doesn't seem to be fixed.  I'm testing on Mac Firefox.

LeonMontealegre:
Idk what's going on;
whenever I set the min-width to something smaller in the inspector – it stops the wrapping, but when I set it to something smaller in the code – it just keeps wrapping even if I go all the way down to 0 width. What's worse is that it's only happening on Firefox so it's a nightmare to find D:

PR #1883, opened 2018-04-18
Stop Autograding Workers JSON from being Overwritten

emaicus:
A quick fix which prevents the autograding workers json from being deleted when configure submitty is called.

### DIFF
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -8,6 +8,7 @@ import os
 import pwd
 import shutil
 import tzlocal
+import tempfile
 
 def get_uid(user):
     return pwd.getpwnam(user).pw_uid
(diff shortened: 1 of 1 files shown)


MasterOdin:
Would a better solution here be just not removing the config directory on each run?

emaicus:
@MasterOdin Probably. What was our reason for removing it in the first place?

MasterOdin:
So we do this with the `.setup` folder as well, and I believe the reasoning for this was that you'd be able to generate all files from the running of CONFIGURE_SUBMITTY.py (and getting all the answers t to the questions) so we don't leave behind any files that might have been retired and are nice and "clean". 

Perhaps a better solution would be to just load the worker file into memory around [here](https://github.com/Submitty/Submitty/blob/master/.setup/CONFIGURE_SUBMITTY.py#L145-L151) and then just write it out later. We can move up where we define the constants for reference the files [here](https://github.com/Submitty/Submitty/blob/master/.setup/CONFIGURE_SUBMITTY.py#L145-L151) upwards and just go from there, that way it sort of follows the existing design (even if at this time we're not asking any questions about configuring the workers).

MasterOdin:
Well, I guess we'll hold off any potential refactors of CONFIGURE till some later date given this was merged.

PR #1885, opened 2018-04-19
Search, categories, and populating sample course for forum 

andrewaikens87:
This PR includes features for the forum which include: searching, categories, and populating the sample course on vagrant with default values. 

Some of the things still needed
- Multiple selectable categories 
- To have search only show thread title if the match is exclusively with the thread title rather than all of the posts corresponding with that thread. 

screen1
screen2
screen3

update_database.py must be run.

Could someone please review? Thanks.


site/app/controllers/forum/ForumController.php
```diff
@@ -85,23 +94,54 @@ private function checkGoodAttachment($isThread, $thread_id, $file_post){
...
+    }
+
     //CODE WILL BE CONSOLIDATED IN FUTURE
 
     public function publishThread(){
         $title = $_POST["title"];
         $thread_content = str_replace("\r", "", $_POST["thread_content"]);
         $anon = (isset($_POST["Anon"]) && $_POST["Anon"] == "Anon") ? 1 : 0;
         $announcment = (isset($_POST["Announcement"]) && $_POST["Announcement"] == "Announcement" && $this->core->getUser()->getGroup() < 3) ? 1 : 0 ;
+        $category_id = $_POST["cat"];
```
MasterOdin: You should just cast this to an int immediately and not do a `is_numeric()` and `empty()` check. Just cast it and use `hasGoodCategory`.

site/app/controllers/forum/ForumController.php
```diff
@@ -85,23 +94,54 @@ private function checkGoodAttachment($isThread, $thread_id, $file_post){
         } return $imageCheck;
     }
 
+    private function hasGoodCategory($inputCategoryId){
```
MasterOdin: `isValidCategory($category_id)` would make more sense as a name

site/app/controllers/forum/ForumController.php
```diff
@@ -85,23 +94,54 @@ private function checkGoodAttachment($isThread, $thread_id, $file_post){
...
+        if($inputCategoryId < 1)
+            return false;
+        $rows = $this->core->getQueries()->getCategories();
+        foreach($rows as $index => $values){
+            if($values["category_id"] === $inputCategoryId)
+                return true;
+        } return false;
```
MasterOdin: put the return on the next line

site/app/controllers/forum/ForumController.php
```diff
@@ -85,23 +94,54 @@ private function checkGoodAttachment($isThread, $thread_id, $file_post){
         } return $imageCheck;
     }
 
+    private function hasGoodCategory($inputCategoryId){
+        if($inputCategoryId < 1)
+            return false;
```
MasterOdin: Please use `{ }` around all if clauses.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -107,8 +107,19 @@ public function insertSubmittyUser(User $user) {
...
+
+    public function loadThreads($announcements, $category_id){
+    	if($category_id === -1) {
+      		$this->course_db->query("SELECT t.*, e.category_id as category_id, w.category_desc FROM threads t, thread_categories e, categories_list w WHERE deleted = false and pinned = ? and t.id = e.thread_id and e.category_id = w.category_id ORDER BY t.id DESC", array($announcements));
+    	} else {
+
+    		$this->course_db->query("SELECT t.*, e.category_id as category_id, w.category_desc FROM threads t, thread_categories e, categories_list w WHERE deleted = false and pinned = ? and w.category_id = ? and t.id = e.thread_id and e.category_id = w.category_id ORDER BY t.id DESC", array($announcements, $category_id));
+    	}
+        return $this->course_db->rows();
+    }
```
MasterOdin: Why is this one function? Both queries are subtly different (with no obvious way to resolve them together at the moment) and don't both use all parameters. Just have a function `loadAnnouncements` and `loadThreads` or whatever.

saxocellphone:
Another thing I noticed is when I try to load threads without a category (Before this PR came out), it doesn't show up on the thread list. The thread itself can still show up on the search result, but it doesn't have a title. 
capture
I don't think it will be a problem from this point on, but anything posted before might be displayed wrongly.

andrewaikens87:
@saxocellphone Did you run update_database.py?

PR #1898, opened 2018-04-25
fixed sorting and added thread names

thatCode-Psy:
Fixed the sorting and added the thread names to entries


bmcutler:
As discussed in meeting, please shorten date per @andrewaikens87

PR #1916, opened 2018-05-02
Improved display options (Chronological, Tree, Alphabetical)

saxocellphone:
This fixes bugs from the previous PR, but the @ feature takes a lot more work than initially anticipated. Since the parent of each post needs to be calculated each time a user requests the page, my algorithm does not work in alphabetical mode because it assumes that the posts are already in chronological order. I will need to change the algorithm to make the @ feature work, or I can make a new column in the database table which might be more work. So I'm not sure if it is worth it.

Other then that, the rest of the code should work fine. I'm just waiting for the top bar to be implemented so I can add the radiobuttons to that.

### DIFF
--- a/site/app/controllers/forum/ForumController.php
+++ b/site/app/controllers/forum/ForumController.php
@@ -213,15 +213,23 @@ class ForumController extends AbstractController {
         $current_user = $this->core->getUser()->getId();
 
         $posts = null;
+        $option = 'tree';
+        if(isset($_REQUEST["option"])){
+            $option = $_REQUEST["option"];
+        }
         if(isset($_REQUEST["thread_id"])){
-            $posts = $this->core->getQueries()->getPostsForThread($current_user, $_REQUEST["thread_id"]);
+            if($option == "alpha"){
+                $posts = $this->core->getQueries()->getPostsForThread($current_user, $_REQUEST["thread_id"], 'alpha');
+            } else {
+                $posts = $this->core->getQueries()->getPostsForThread($current_user, $_REQUEST["thread_id"], 'tree');
+            }
         } else {
             //We are at the "Home page"
             //Show the first post
             $posts = $this->core->getQueries()->getPostsForThread($current_user, -1);
             
         }
-        $this->core->getOutput()->renderOutput('forum\ForumThread', 'showForumThreads', $user, $posts, $threads);
+        $this->core->getOutput()->renderOutput('forum\ForumThread', 'showForumThreads', $user, $posts, $threads, $option);
     }
 
     public function showCreateThread(){
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -1862,7 +1862,7 @@ AND gc_id IN (
       return $ar;
     }
 
-    public function getPostsForThread($current_user, $thread_id){
+    public function getPostsForThread($current_user, $thread_id, $option='tree'){
 
       if($thread_id == -1) {
         $announcement_id = $this->existsAnnouncements();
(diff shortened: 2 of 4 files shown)


site/app/controllers/forum/ForumController.php
```diff
@@ -213,15 +213,23 @@ public function showThreads(){
         $current_user = $this->core->getUser()->getId();
 
         $posts = null;
+        $option = 'tree';
+        if(isset($_REQUEST["option"])){
+            $option = $_REQUEST["option"];
+        }
```
MasterOdin: You can replace these four lines with:
`$option = $_REQUEST['option'] ?? 'tree';`

site/app/views/forum/ForumThreadView.php
```diff
@@ -251,54 +252,54 @@ function changeName(element, user, visible_username, anon){
...
-							$first= false;
+							if($first){
+								$first= false;
+							}
+							$i++;
 						}
-						$i++;
-					}
-
+
```
MasterOdin: It looks like you just indented everything by one level unnecessarily? Please unindent the block back to match the indentation level of the stuff above it.

site/public/js/server.js
```diff
@@ -955,6 +955,24 @@ function enableTabsInTextArea(id){
...
+            option: option
+        },
+        success: function(data){
+            window.location.replace(buildUrl({'component': 'forum', 'page': 'view_thread', 'thread_id': thread_id, 'option': option}));
+        },
+        error: function(){
+            window.alert("Something went wrong while trying to change views. Please try again.");
+        }
+    });
+}
```
MasterOdin: What is the point of this ajax call? It doesn't look like 'change_display_option' is a real page in the forum component, nor are you really verifying that said call actually did anything.

site/app/controllers/forum/ForumController.php
MasterOdin: You can replace these four lines with:
`$option = $_REQUEST['option'] ?? 'tree';`

site/app/views/forum/ForumThreadView.php
```diff
@@ -182,81 +182,94 @@ function openFile(directory, file, path ){
...
-			    } 
+	    } 
 
-$return .= <<<HTML
+	$return .= <<<HTML
 			</select>
 			</div>
 			<button class="btn btn-primary" style="float:right;position:relative;top:3px;right:5px;display:inline-block;" title="Display search bar" onclick="this.style.display='none'; document.getElementById('search_block').style.display = 'inline-block'; document.getElementById('search_content').focus();"><i class="fa fa-search"></i> Search</button>
+HTML;
+	if($this->core->getUser()->getGroup() <= 2){
```
scopeInfinity: We are checking user authorization in UI but not in the backend therefore even student can add `option` parameter in url directly to change view.
```php
$option = $_REQUEST['option'] ?? 'tree';
```

https://github.com/Submitty/Submitty/pull/1916/files#diff-82b25ad2bdc78a61032e380fe360d299R295

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1912,8 +1911,12 @@ public function getPostsForThread($current_user, $thread_id){
           $thread_id = $announcement_id;
         }
       }
-
-      $this->course_db->query("SELECT * FROM posts WHERE thread_id=? AND deleted = false ORDER BY timestamp ASC", array($thread_id));
+      if($option == 'alpha'){
+        $this->course_db->query("SELECT posts.*, users.user_lastname FROM posts INNER JOIN users ON posts.author_user_id=users.user_id WHERE thread_id=? ORDER BY user_lastname;", array($thread_id));
```
scopeInfinity: Even if `post` is `Anonymous` it is still being sorted according to author's `last_name`.

site/app/views/forum/ForumThreadView.php
andrewaikens87: Good catch, I believe that this can be fixed by simply changing the above expression to a ternary expression that also includes a check for the group being <= 2.

site/app/libraries/database/DatabaseQueries.php
andrewaikens87: Since the visual of the student remains anonymous and only Full Access Graders and Instructors have access to this option I believe that this behavior is fine for now. I am going to make a note of this though and we can possibly change this in the future.

bmcutler:
I got a crash when pressing or submitty a reply in alphabetical mode.  :(
forumcrash

bmcutler:
we don't have to support reply to individual posts while in non tree/ hierarchical mode.  but we should hide the individual post reply buttons as needed and only allow typing in the bottom box?

bmcutler:
Actually it crashes if you make any post in alpha or chrono mode
And, why are the posts in alpha-mode sorted secondarily in reverse chrono order?  It should be alpha, then chrono with early posts at the top.

andrewaikens87:
I just tested in the bottom reply box in non tree/hierarchical modes and it seems to be broken there as well. I am going to investigate.

bmcutler:
@saxocellphone 
please resolve conflicts with the master branch

PR #1917, opened 2018-05-02
Marks bug fixes and refactors

LeonMontealegre:
(Hopefully fixes):
Bug where clicking on marks would "randomly" not save
Bug where adding a new common mark would discard previous edits

Also a lot of nice refactoring that makes the code prettier and more succinct.


site/public/js/ta-grading-mark.js
```diff
@@ -176,6 +226,23 @@ function updateMarksOnPage(num, background, min, max, precision, gradeable_id, u
         var marks = $('[name=mark_'+num+']');
         for (var x = 0; x < marks.length; x++)
             marks[x].remove();
```
MasterOdin: Can we put `{ }` around the for loop body, especially if we're going to then have a separate block scope immediately following this line?

PR #1932, opened 2018-05-08
Separates out Travis into a bunch of build stages

MasterOdin:
This PR breaks apart Travis from one stage/language with a couple of different matrix includes into a bunch of different stages. This allows us to easily test a bunch of different things in a "quicker to fail" fashion, especially as we plan to grow our test suites out. I've also added PHP 7.2 to be tested/tracked and it currently has some build failures (though not sure if it's just how we're testing it or an actual potential issue in the code).

The nice thing about doing this is that we also can expand our efforts more easily into other languages as well to test things (for example, using different clang compilers) and not just PHP throughout.

The downside is that builds take a bit longer to complete as each build, no matter how small has a ~30 second or so cost to run. However, I believe that in the long run, this does make the Travis build a bit easier to maintain as there's a more clear separation of concerns between the different stages, and you don't have to wait as long for say the PHP unit tests anymore.

Image of what this looks like:
Travis Image


MasterOdin:
I've further expanded this to also implement the beginnings of how we'd do a python unit test suite (which could be added to as part of #1933).

PR #1947, opened 2018-05-11
Mark bug fix

### LINKED ISSUE #1931: Must close rubric item to save
I cannot simply click `Move to Next Student` to save a mark - I must actually close the rubric item. As far as I am concerned, this makes Submitty borderline unusable.

Ubuntu 16.04.4, Firefox 59.0.2

LeonMontealegre:
Fixes bug where going to next student without closing the mark wouldn't save it.

Fix #1931

### DIFF
--- a/site/app/views/grading/ElectronicGraderView.php
+++ b/site/app/views/grading/ElectronicGraderView.php
@@ -1410,7 +1410,7 @@ HTML;
             }
 
             $return .= <<<HTML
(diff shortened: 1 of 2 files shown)


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1595,8 +1595,28 @@ function display_files($files, &$count, $indent, &$return, $filename) {
...
+    // Will work for other browsers
+    if (!__unloadSentRequest) {
+        console.log("unload");
+        saveLastOpenedMark('{$gradeable->getId()}' ,'{$user->getAnonId()}', {$gradeable->getActiveVersion()}, '{$your_user_id}', '-1', false, function() {
+            __unloadSentRequest = true;
+        }, function() {
+            // don't error
+        });
+    }
 }
```
MasterOdin: Can you make this a function and you should set `__unloadSentRequest` (though rename this to `__unloadRequestSent`) immediately once entering the if statement just to better avoid any potential race conditions.

```
function unloadPage() {
    if (!__unloadRequestSent) {
        __unloadRequestSent = true;
        ...
    }
}
```

Call the function from `window.onunload` and `window.onbeforeunload`. Please also add a docstring above this stuff to explain why we're doing it with any potential links to resources that explain what's going on just for the future developer who might wonder what's going on here.

Issue #1950, opened 2018-05-12
More forum requests


andrewaikens87:
- [ ] Give instructor ability to mute a user (allows student to still view threads/posts but not post any new content)

- [ ] Instructor ability to merge common threads together (will have to make thread numbering more apparent/accessible, necessary as thread numbers increase)

- [ ] Question resolving system - if teaching staff members are taking turns responding to students questions, they should try and resolve as many questions as possible (this could be included in a "Staff portal" which could encapsulate some of #1677 as well and can be used for instructor/tas to have a settings menu, or could be a generic "staff pinned item"). The student who created the thread would be able to modify the state, whether the question has been resolved or not. 

- [ ] Consider looking into developing a mobile application that could help facilitate notifications (this could be used for the forum, viewing grades, etc).  

- [ ] Refactor forum code to mitigate duplicate code for #1816

bmcutler:
Hi Gagan @scopeInfinity,

Andrew @andrewaikens87 is going to divide up the open tasks on the forum and give you additional items, but I think the first big feature to have you work on is merging common threads.  The 2nd point in the list above.

More description:  When a student makes a new thread that is asking a question that duplicates something asked & answered in another thread, the instructor or full-access or limited-access grader can move that whole thread into the first thread as a subtree (posts within threads are like nodes in a tree -- the first post in a thread has no parent, but when we will change that post to have the other thread as the parent).

Please ask questions here or on Slack.

Thanks!
Barb

andrewaikens87:
Gagan @scopeInfinity, also had interest in working on the instructor ability to mute users. He can work on this once he finishes the other items he is assigned to.

PR #1962, opened 2018-05-16
Merge common threads

scopeInfinity:
Issue #1950 

Current Implementation 
- Assumes child thread must be newer than parent thread
- Assign `$merged_id` as PK of root node of child thread, and set it's parent as root node of parent thread
- Marks child thread as deleted.

**Screenshots**
screenshot from 2018-05-16 14-58-35

After `Thread Y` merged into `Thread Z`
screenshot from 2018-05-16 14-59-03


scopeInfinity:
`Merge Threads` button will pop up list of possible parent threads.

Updated Screenshots

screenshot from 2018-05-17 21-46-40
screenshot from 2018-05-17 21-48-39

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1922,6 +1922,53 @@ public function getPostsForThread($current_user, $thread_id){
...
+                $this->course_db->rollback();
+                return false;
+            }
+            $this->course_db->query("SELECT 1 FROM threads WHERE id = ? and merged_id = -1 and deleted = false", array($parent_thread_id));
+            $result_rows = $this->course_db->rows();
+            if(count($result_rows)<1) {
+                $message = "Can't find parent thread";
+                $this->course_db->rollback();
+                return false;
+            }
```
andrewaikens87: Lines 1929-1942 has duplicate code. It might be worth creating a function that returns a boolean to whether or not a non merged thread exists.

andrewaikens87:
bad allign
The popup button don't seem to stay on the same line

jchulton:
One potential issue is that the attachments of the child thread are not carried over after the merge. Screenshots are below. Similarly, if a thread is an announcement, it will not be after being merge if it is the child thread. I don't remember this coming up during the meetings, so we may want to talk about it during today's meeting. Other than that a potential issue is that you can't have a newer thread as the parent of an older thread, but we discussed this at our meetings.
mergescreenshot1
mergescreenshot2
mergescreenshot3

andrewaikens87:
You are currently storing post ids in `merged_id`, please change the `merged_id` to contain the thread_id that the row merged into.

scopeInfinity:
Thanks, @bmcutler @andrewaikens87 @jchulton for review.

Required changes have been made.
- Refactored code to remove duplicate code
- Fixed button alignment bug
- Fixed attachments bug on merging
- Checkbox confirming to make parent thread as announcement
  - Default selection is true if any among child or parent thread is announcement
- For now, appending child thread title to content of its first post for not losing thread title.
- Added E2E tests  #1975

andrewaikens87:
There seems to be a few warnings with your tests. Take a look at [https://stackoverflow.com/questions/24450395/python-with-selenium-invalid-warning](https://stackoverflow.com/questions/24450395/python-with-selenium-invalid-warning)

`test_basic_operations_thread (e2e.test_forum.TestForumMergeThread) ... /opt/pyenv/versions/3.6.3/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py:788: DeprecationWarning: use driver.switch_to.alert instead
  warnings.warn("use driver.switch_to.alert instead", DeprecationWarning)
ok`

`test_forum_merge_thread (e2e.test_forum.TestForumMergeThread) ... /opt/pyenv/versions/3.6.3/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py:788: DeprecationWarning: use driver.switch_to.alert instead
  warnings.warn("use driver.switch_to.alert instead", DeprecationWarning)`

scopeInfinity:
Thanks @andrewaikens87 
Warnings fixed.

andrewaikens87:
Just found an issue, you can recreate the screenshot error below by having two windows open. Have one merge a thread and the other stay on the child thread page and post something. It will post in the child thread a reply to a post in the parent thread.

view

Merge threads

after merge

Post on child thread

after post

If you take a look in the database notice that the child thread 9 has a post whose parent is a part of thread 5 (which is the parent). This is from a different thread id but has the same result.

db

scopeInfinity:
@andrewaikens87  The bug related to publishing post on a deleted thread.
I tried fixing it by testing thread existence just before publishing post.

andrewaikens87:
I think this is good. The only other thing that I think should be added is the column in the threads table to keep track of the merged thread id (currently we only keep track of the top post id of the child thread). I mentioned above for it to be added in the future but we should include it in this PR as it pertains to the code that is getting merged. @jchulton can you take another look? Thanks.

jchulton:
I'm still having the bug where you loose all attachments on the child branch when you merge: Screenshots below:
Child Thread:
thread merge p1
Parent Thread:
thread merge p2
Merged Thread:
threads merged

jchulton:
Maybe we should change this to a WIP PR and merge it. A lot of progress has been made and this PR seems to be getting a bit long, but I don't think we should call the merging of common threads done until it works with attachments (unless we want to specifically specify that it doesn't).

andrewaikens87:
@jchulton attachments seem to be working for me. I messaged you on slack to discuss more.

bmcutler:
Nice work :)

PR #1963, opened 2018-05-16
add/delete registrations for a course

### LINKED ISSUE #1293: Web UI to add/delete registration sections for a course
Allow instructors to add section at any time.
Can only delete a registration section if there are no users in it.

Note:  Instructions have been added to submitty.org so a sysadmin can directly add registration sections to the database.

tushargr:
Closes #1293 
Add two more options on course setting page 
1. Add registration section - successful only if new registration section isn't already a registration section.
    if successful, add new registration section to sections_registration table.
2. Delete registration section- check if the entered reg section is already a reg section and also no user is enrolled under the reg section. If checks pass, then delete reg section from sections_registration table.


tushargr:
issue1293

bmcutler:
Hi Tushar,

I think instead of placing this option in course settings, we should place it on the "Setup Rotating Sections" page.  Let's rename the page "Setup Sections", and have 2 parts, the lower part being the current information/actions for rotating sections, and then an additional panel(s) above it with information/actions for registration sections.  We can also add a few simple sentences explaining/documenting the difference between registration & rotating sections (a common confusion for new instructors.

Barb

bmcutler:
@pbailie 

Hi Peter, tagging you to ask a question.  For your autoupdate registration script, will there be any problem if we have additional registration sections that aren't on your list?  What if we fully delete a section that has no students?  Will your script crash or error?  Can we make it robust to handle this potential error?  It could happen that an instructor lists a course for cross list, but gets no students in that section.  If the instructor manually deletes the section on Submitty, will your script handle situations of late registration to that section?  Or other errors?

tushargr:
ok I will be shifting it to Setup Sections Page

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -940,6 +940,14 @@ public function insertNewRotatingSection($section) {
         $this->course_db->query("INSERT INTO sections_rotating (sections_rotating_id) VALUES(?)", array($section));
     }
 
+    public function insertNewRegistrationSection($section) {
+        $this->course_db->query("INSERT INTO sections_registration (sections_registration_id) VALUES(?)", array($section));
+    }
+
+    public function deleteRegistrationSection($section) {
+        $this->course_db->query("DELETE FROM sections_registration WHERE sections_registration_id=?", array($section));
```
pbailie: This column is [referenced by another table](https://github.com/Submitty/Submitty/blob/871cc96ea5dc1f952e2cda514cdc619412345f6f/site/data/course_tables.sql#L854-L855).  Although not specifically defined in `course_tables.sql`, a deletion here could trigger a cascade delete on the `grading_registration` table.  Likely, this delete query could entirely fail, if the cascade delete cannot be resolved for some reason.  e.g. `grading_registration` has a multi-column primary key, part of which is referencing `sections_registration`.

tushargr: Hi @pbailie ,
It is ensured that the registration section which is to be deleted do not have any user enrolled in it.
So if we are deleting a registration section (lets say 10) satisfying above condition, then there will be no entry in grading_registration table with 10 since no user will have 10 as registration section. So on deleting 10 in sections_registration table, the grading_registration table same.
Will the above function still cause problem.

bmcutler: There's one more check we should add.  Graders can be assigned to one or more registration sections for grading.  We need to make sure that in addition to no users being registered to the registration section, no users/graders are assigned to grade that registration section.

tushargr: ok got it. For now I have shifted the whole thing to SetUp page

tushargr:
@bmcutler 
Added one more check - now before deleting a registration section , it will also ensure that no grader is assigned for grading to that registration section

tushargr:
@bmcutler 
Changed the UI of setup page as required.

pbailie:
@bmcutler 

Following up with your earlier questions:

>>  Can we make it robust to handle this potential error?
>
> The original version made sure that any needed registration sections were inserted before processing. But in later discussion, it was determined that instructors were primarily responsible for this, so that check was not implemented in the current version. I can add it back in, if desired.

Because the auto feed script is intended to access _only_ submitty's "master" DB, and that the `sections_registration` table exists in a course's DB, I believe the approach to handle this error would be to adjust the [trigger function](https://github.com/Submitty/Submitty/blob/3800ad66ec6f97ef7bf5f549e81552a1e8945fd3/site/data/submitty_db.sql#L186-L222) in `site/app/data/submitty_db.sql` used to copy data from the "master" DB to a course DB.   

The adjustment would be to add an additional query so that any insert or update on a student record (auto feed or otherwise) would trigger through to the course DB and automatically create any registration sections, as needed.

The alternative is that the auto feed script will have to connect to each course's DB to insert any needed registration sections.

pbailie:
@bmcutler 
Following up some more:
>What if we fully delete a section that has no students? Will your script crash or error?

As previously stated, the issue is a FK (foreign key) violation when a registration section does _not_ exist in the course table.  I recreated the error and got the following:

```
PHP Warning:  pg_query(): Query failed: ERROR:  insert or update on table "users" violates foreign key constraint "users_registration_section_fkey"
DETAIL:  Key (registration_section)=(1) is not present in table "sections_registration".
CONTEXT:  Error occurred on dblink connection named "unnamed": could not execute command.
SQL statement "SELECT dblink_exec(db_conn, query_string)"
PL/pgSQL function sync_courses_user() line 18 at PERFORM in {path redacted}/submitty_student_auto_feed.php on line 616
```

However, I need to point out that the error is not caused by the auto feed script in itself.  It is a consequence of a FK violation encountered by the [trigger function](https://github.com/Submitty/Submitty/blob/3800ad66ec6f97ef7bf5f549e81552a1e8945fd3/site/data/submitty_db.sql#L186-L222) (`site/app/data/submitty_db.sql`) I previously mentioned.

> Can we make it robust to handle this potential error?

I certainly can handle the error case more gently in the auto feed script, but the real problem lies in the aforementioned database trigger function.  As such, I am emphasizing one of my previous statements:

> Because the auto feed script is intended to access only submitty's "master" DB, and that the sections_registration table exists in a course's DB, I believe the approach to handle this error would be to adjust the trigger function in site/app/data/submitty_db.sql used to copy data from the "master" DB to a course DB.
>
> The adjustment would be to add an additional query so that any insert or update on a student record (auto feed or otherwise) would trigger through to the course DB and automatically create any registration sections, as needed.

bmcutler:
@tushargr 
Can you please resolve the conflicts for this PR?  Thanks.

PR #1967, opened 2018-05-17
Grading Page Results Loading Rework

### LINKED ISSUE #727: Rewrite diff viewer to only load one diff at a time
Right now, we load all diffs into memory and then go to display them. We should rewrite this such that we only actually load the diff viewer and its output when we go to display it (and then immediately recycle the output).

This allows us to better deal with far larger files than we currently do.

emaicus:
closes #727 

Updates the display of student autograding results to only load in diffs that are currently visible/expanded. This will significantly improve loading time on courses with very large expected output.

Validation is done at loading request time to ensure that users can only load results that belong to themselves or the subset of individuals they are currently allowed to grade. Care was taken to keep hidden testcases hidden and inaccessible.

Created the canIViewThis function in the grading controller, which will hopefully become more widely used to determine whether or not a user has access to view a gradeable for a given student.


site/public/js/server.js
```diff
@@ -16,8 +16,10 @@ window.addEventListener("load", function() {
  * @param {object} parts - Object representing URL parts to append to the URL
  * @returns {string} - Built up URL to use
  */
-function buildUrl(parts) {
-    var url = siteUrl;
+function buildUrl(parts, url="") {
+    if(url===""){
+        url = siteUrl;
+    }
```
MasterOdin: I'm not sure I see the point of this change? What is the use-case that we're worried that we'll need to build a URL to a different site?

MasterOdin: I think if that's the usecase, I'd then prefer just moving script.js to load at the end of the DOM and fire immediately then (when body would be available) than make this hack. Or examine why we're firing events to buildUrls before we've finished loading the initial DOM.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -934,6 +942,33 @@ public function saveSingleComponent() {
...
+        
+
+        //Turns off the header and footer so that it isn't displayed in the testcase output
+        //Don't re-enable. 
+        $this->core->getOutput()->useHeader(false);
+        $this->core->getOutput()->useFooter(false);
+
+        $return = "";
+
+        $popup_css = "{$this->core->getConfig()->getBaseUrl()}css/diff-viewer.css";
```
MasterOdin: are we loading the same CSS for every student output? Is there any reason to not just load it once in the initial controller function that gets fired?

emaicus: This is an artifact of the way Andrew M. wrote the function before. It looks like the openPopUp function in server.js takes in a css file path as input, which it then uses to style the resulting popup window. This css is only loaded if the user chooses to open a popup version of their output.

PR #1968, opened 2018-05-17
TA grade precision fixed for hundredths (if used)

### LINKED ISSUE #1899: TA grade precision: hundreths (if used) not displayed in all places
screen shot 2018-04-25 at 9 03 28 am
screen shot 2018-04-25 at 9 04 15 am

jn7070:
Closes #1899 
If hundredths place option is used for TA grading, it will be displayed in all places.
The spot that was not displayed in was at the checked boxes which should now 
display the correct precision value based on point precision chosen.


site/app/models/GradeableComponent.php
```diff
@@ -191,11 +191,15 @@ public function getGradedTAComments($nl, $show_students, $use_ascii = true) {
         foreach ($this->marks as $mark) {
             $points_string = "    ";
             if ($mark->getPoints() != 0) {
-              $points_string = sprintf("%4.1f",$mark->getPoints());
+              $roundPoints = round($mark->getPoints()/$gradeable->getPointPrecision())*$gradeable->getPointPrecision();
+              if ($gradeable->getPointPrecision() === 0.01)
```
bmcutler: Instead of checking for equality with 0.01, this should be generalized for any precision into the hundredths, e.g., 0.05.
Also, instead of handling only tenths and hundredths, this should handle round number precision (= 1) and thousandths (= 0.001), etc.

site/app/models/GradeableComponent.php
```diff
@@ -214,7 +218,11 @@ public function getGradedTAComments($nl, $show_students, $use_ascii = true) {
             }
             $score_string = "    ";
             if (floatval($this->score) != 0) {
-                $score_string = sprintf("%4.1f",$this->score);
+              $roundScore = round($this->score/$gradeable->getPointPrecision())*$gradeable->getPointPrecision();
+              if ($gradeable->getPointPrecision() === 0.01)
```
bmcutler: same as above

bmcutler:
@jn7070 / @emaicus your recent changes are colliding in a merge conflict here.  Please sort this out.  Thanks.

Issue #1677, opened 2018-01-29
Forum: medium term todo list


bmcutler:
- [ ] ~Instructor/full access ta can specify a 'lock' date for a thread.  After that date, all can view, but only instructor can edit.~  

- [ ] ~~Original author can edit a post if a thread is not locked~~

- [ ] ~~Instructor/full access ta can view/undelete/unhide posts~~

- [x] allow students to 'pin' threads

- [x] thread categories, and viewing/sorting by categories

andrewaikens87:
Hi @scopeInfinity , once you complete the current issue you are working on I think a good next issue could be student pinned threads. This is where a student can self pin a thread that will bump it up to the top of the thread list, below announcements. Students can use this to flag important threads that they are going to return to later. We will also need to have the ability to unpin the threads as well. We can discuss the design through slack.

PR #1973, opened 2018-05-18
Allow students to pin threads

scopeInfinity:
Issue : #1677


jchulton:
One potential issue is that pinned announcements are not ordered higher than non-pinned announcements. I believe this was the desired order that was discussed at Friday's meeting, please respond if it isn't.
pinnedscreenshot

PR #1982, opened 2018-05-19
[WIP] add team export and import feature 

### LINKED ISSUE #1508: teams that exist across multiple gradeables
Long term:  Might want to have teams of students that persist across multiple gradeables.

Implementation ideas:
a) students need to reform teams for every assignment
b) instructor can export teams as a .csv from the first gradeable and upload the .csv of team assignments to the second gradeable
c) teams can be shared across multiple gradeables

Advantage of option c:  Team repos can be shared across gradeables.

tushargr:
Closes #1508 
@bmcutler 

Features-
1. Same team across multiple gradeable.
2. Teams exported and imported should already be an existing team on course.
3. No import if any team exists for gradeable.
4. Empty teams cannot be exported/ imported.
5. New team setting file created for teams on new gradeable on which teams are imported, taking time stamp of importing.
6. If there is team which is across two or more gradeables. If instructor alter that team on one gradeable, it also get altered on other gradeables as it is the same team.

Implementation details-
1. Since now there can same team across multiple gradeables. So the team_id column of gradeable_teams table can't remain primary key. Now (team_id,g_id) of gradeable_teams table is primary key. Also in the teams table team_id can be repeated in multiple rows. 
So created a new table "team_ids" to store just team ids which is a primary key. 
2. gradeable_teams and teams table are now referring (ON DELETE and UPDATE CASCADE) to team_ids table. Those tables which were earlier referring (ON DELETE and UPDATE CASCADE) to gradeable_teams table's team_id column are now referring to team_ids table's team_id  column.
3. If a gradeable is deleted (which is not yet in submitty), team associated with gradeable won't be deleted, i.e., if a row from gradeable_teams table is deleted, then no change on team_ids table.


bmcutler:
I put my comment under the issue  #1508

tushargr:
Last commit features-
When teams are imported , new team_ids will be allotted as if new team created by instructor or student.
These new teams have no connected with gradeable they are obtained from.

KevinMackenzie:
Sorry for making such a mess of this.  All of my changes are reverted, so only the "part 2" stuff will be applied

tushargr:
yes it will be a small change, just in the ElectronicGradeableView.php. 
I will make the changes.

tushargr:
Part 2 is done as required. Made the changes . Export allowed only if >=1 non-empty teams and import allowed only if no teams or just empty teams.

PR #1983, opened 2018-05-19
fix duplicate member on current team bug and team truncation bug

### LINKED ISSUE #1969: Instructor team creation bugs: duplicate member on current team & > 1+ then max team size
When the instructor edits a team to add one or more users, usually if you type the name of a student who is already on another team you get an error (red popup message).

But if you enter the name of a user who is already on the current team, you don't get an error

--------

Also, if you edit a team to make it bigger than the instructor specified max size (which should be allowed, the instructor should be able to make exceptions to the max size), sometimes it truncates the team to the max size, kicking members of the team at the end of the alphabet.

tushargr:
closes #1969 
@bmcutler 
I guess there is no need for max_members for admin team form as admin (instructor) can exploit the max member limit. 
Now whenever admin team form is opened, there will be readonly fields for member already on team plus two extra blank fields (for adding more users to team). Also more blank fields can be added as was before


bmcutler:
Minor tweak:  when I try to add a username that is already on the team I get a "XXX is not a valid username" instead of a "XXX is already on this team" error

PR #1990, opened 2018-05-21
Remove deprecated "individual" field on ElectronicGraderView/Controller

CouleeApps:
Field was being input as a string from $_REQUEST then passed back raw to the page in ElectronicGraderView.php:1219


CouleeApps:
I can't find anywhere this field is used. In the future we may be able to remove it

CouleeApps:
Apparently this field is legacy and should be removed. So I'll do that instead.

PR #1991, opened 2018-05-21
Create/Edit Gradeable warns if grading opens before late days end

### LINKED ISSUE #1989: Create/Edit Gradeable Should Warn User of Inconsistent Dates
It is sometimes desirable to retroactively change the number of allowed late days without adjusting the `Manual Grading Opens` and `Grades Released` dates.  Currently, this is permitted, but Submitty should warn the user if they decide to violate the `must be >= Due Date (+ max allowed late days)` condition to prevent accidental violations.

This may be related to #1431

KevinMackenzie:
Closes #1989 

The instructor should be able to retroactively increase the late day cap on a gradeable, but this should be a very intentional action.  Otherwise, graders may need to redo work if a student resubmits after grading.

The form now prompts the user with a warning when they attempt to save the gradeable if the late day submission cap extends beyond the `Manual Grading Opens` date, but saves it anyway.

### DIFF
--- a/site/app/views/admin/AdminGradeableView.php
+++ b/site/app/views/admin/AdminGradeableView.php
@@ -240,7 +240,7 @@ HTML;
                 <em style='color: orange;'>must be >= Submission Open Date</em>
                 <br />
 
-                How many late days may students use on this assignment? <input style="width: 50px" name="eg_late_days" class="int_val"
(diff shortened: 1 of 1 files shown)


KevinMackenzie:
The Travis CI build failed to complete because the `apt-get install` command failed.  Do we know what causes this?

Issue #1460, opened 2017-09-13
Electronic gradeable with no ta grading should not say "TA grade not available"


bmcutler:
reproduce:
an electronic gradeable, with "no ta grading"
all dates are in the past (the grades have been "released")

The "View Grade" button should never be green (there is no further information beyond autograding for the student to review).

Once you click on view grade (to review the autograding), there should be no panel beneath the autograding for TA grading -- it should not say TA grade not available, that div should not be there!

TA grade not available should only be there if grades have been released, but for some reason the student does not have a TA grade (the TA didn't finish before grades were released).

PR #1994, opened 2018-05-21
Do not show viewed grade status if grade hasen't been released

### LINKED ISSUE #1449: "Viewed Grade" column should not be true/checked if grades are not yet released
* confirm that if the ta has graded the student, but grades are not released, that the student CANNOT see the ta grade report

* confirm that if the ta has graded the student, and the grades are released, that the student CAN see the grade report

* Add a reasonable message, if the ta has not graded the student, but grades are released (those that are finished).  Something like "Your assignment has not been graded, contact XXX for information".  If the student didn't even submit, make sure there's a reasonable message.

* Currently, if the student views the autograding results page after the ta has graded, but before grades are released, the green check box for viewed grade is set, even though they didn't see the ta grade!  THIS NEED TO BE FIXED.  Also confirm that the VIEW GRADE button is correctly green or white.

saxocellphone:
Closes #1449


bmcutler:
Please title your PR as described:
http://submitty.org/developer/how_to_contribute

saxocellphone:
Oops I totally missed that, nice catch. I think I added a reasonable message when the TA grade isn't available but the grades are released. 

On a related note, just like Barb's observation in #1460, students cannot view TA grades in gradeables in sample course, but after a regrade the grade will show up. I believe this can be fixed by changing the way sample course is loaded

PR #1999, opened 2018-05-22
Fix general comment not wrapping

### LINKED ISSUE #1520: TA general comments overflow
The general comment box will overflow in both the grading section (making the "Done" button for saving grades unreachable without a horizontal scroll), and in the student view (also unable to read without a horizontal scroll).

CouleeApps:
Closes #1520

This allows the comment to wrap to new lines on both the student and grader pages. On the grader page, the comment is no longer inside a `<pre>` tag. Also fixes an xss in the general comment.


bmcutler:
The pre tag was in there to allow newlines & blank lines between paragraphs.  Does that display correctly to the student?

Not just for the overall comment, but can the graders also include newlines in marks that are preserved & displayed nicely?  (Do they need to use '\n' or '<br>' or does a simple return work?)

CouleeApps:
Newlines currently don't work, thanks for the heads-up. I'll push a fix for that.

PR #2012, opened 2018-05-23
Improve grading page responsiveness with async and general box 

CouleeApps:
This does a few things:
- Makes the ajax / xhr requests async so your page doesn't freeze when submitting each component
- Makes the component boxes appear in yellow while saving. They still turn orange when saved.
- Makes the general box also get this colored indication
- Makes the general box accessible with keyboard shortcuts

### DIFF
--- a/site/app/views/grading/ElectronicGraderView.php
+++ b/site/app/views/grading/ElectronicGraderView.php
@@ -1425,15 +1425,13 @@ HTML;
             }
 
             //get the grader's id if it exists
-            $grader_id = "";
-            $graded_color = "";
             $displayVerifyUser = false;
             if($question->getGrader() === null || !$show_graded_info) {
                 $grader_id = "Ungraded!";
-                $graded_color = "";
+                $graded_class = "ungraded";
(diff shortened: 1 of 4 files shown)


jchulton:
Looks good except for one small issue. If there are no marks to load, the mark area will constantly display loading. This is a misnomer, since there are no marks to load. I'd recommending quickly changing this to something more clear.
5 24 18 loading screenshot

CouleeApps:
To be fair I think that's more of an issue with the gradeable creator. As far as I'm aware there should never be any question with no marks to load. I'll still attempt to handle it, though.

bmcutler:
There should always be at least one mark, the no credit or full credit.  It should not be editable (either the points or the text).  And it should not be deletable.  Actually that's a problematic bug right now.  In some places it let's you edit or think you're editing the points/mark text.  I'm going to make an issue...

CouleeApps:
This has major merge conflicts with the new ta grading ui. I'm going to remake this in another branch and submit a new PR.

PR #2020, opened 2018-05-24
Remove developer group from Submitty

### LINKED ISSUE #2006: If a Developer edits their sections, they get demoted to Instructor
This is probably not a big problem in practice because (I assume) nobody is a Developer on the live server.

To reproduce:
View graders on a course and edit a Developer. There is no option to keep their group as Developer and they are demoted to Instructor when you submit the form.

MasterOdin:
The group is no longer necessary. The only thing it was controlling is now just shown/not-shown based on whether or not Debug mode is on (or not).

Closes #2006

### DIFF
--- a/.setup/data/users/developer.yml
+++ b/.setup/data/users/developer.yml
@@ -1,6 +1,0 @@
-sudo: true
-user_firstname: Nancy
-user_group: 0
-user_id: developer
-anon_id: developer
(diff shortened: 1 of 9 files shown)


bmcutler:
make sure to remove the developer user from the docs
http://submitty.org/developer/vm_install_using_vagrant

PR #2021, opened 2018-05-24
Twig templates

CouleeApps:
Groundwork for using Twig as our templating engine instead of raw php. Also included is templates for the home page and the users (students, graders, sections) view.


site/app/libraries/Output.php
```diff
@@ -34,6 +37,12 @@ class Output {
     public function __construct(Core $core) {
         $this->core = $core;
         $this->start_time = microtime(true);
+
+        $this->twig_loader = new \Twig_Loader_Filesystem(FileUtils::joinPaths(dirname(__DIR__), 'templates'));
```
MasterOdin: So we could have the options work like the following:
```
$options = [];
if (!$this->core->getConfig->isDebug()) {
    $options['cache'] = FileUtils::joinPaths(dirname(__DIR__), 'cache', 'templates');
}
```

However, @bmcutler, this would require the cache/templates folder being read/write and I'm not sure the permissions on the cached templates would match everything? We could implement/add something like [twig-api](https://github.com/farazdagi/twig-cli) and try and compile all of the templates in the installation process and then we can set the permissions as we like.

site/app/views/HomePageView.php
```diff
@@ -14,167 +14,38 @@ class HomePageView extends AbstractView {
...
+        $ranks = array_filter($ranks, function($rank) {
+            return count($rank["courses"]) > 0;
+        });
+
+        return $this->core->getOutput()->renderTwigTemplate('HomePage.twig', [
+            "user" => $user,
+            "ranks" => $ranks,
+            "changeNameText" => $changeNameText,
+            "showChangePassword" => $this->core->getAuthentication() instanceof DatabaseAuthentication
+        ]);
```
MasterOdin: Do we want to even use these views anymore? It seems like we'd want to move any of this functionality to controllers (or create Services or something) and just try and delete files out of this directory.

bmcutler:
Issues #2052 and #2053 will continue this work.  Agreed to merge as is.

MasterOdin:
I've opened #2052 and #2053 to track the two issues I had with this, rest of it looks fine for now.

PR #2029, opened 2018-05-25
Fix IniParser creating invalid ini files when it encounters newlines

### LINKED ISSUE #2024: IniParser does not correctly escape newlines
If you put a newline in the upload message, it puts the newline into config.ini and breaks the ini parser. This breaks basically the whole website.

To reproduce:
- Go to Course Settings for the sample course
- In the Upload Message box, enter a few newlines and some text
- Watch the whole site melt

This probably applies to the other fields as well, but you can't add newlines to their values _from the frontend webpage_. If someone updated them via a raw url request they would probably break the website in a similar fashion.

CouleeApps:
Closes #2024


site/tests/app/libraries/IniParserTester.php
```diff
@@ -3,13 +3,101 @@
...
 use app\libraries\IniParser;
+use PHPUnit\Runner\Exception;
 
 class IniParserTester extends \PHPUnit\Framework\TestCase {
+    /**
+     * Test that an array writes correctly and is then read back exactly the same
+     * @param $writeData
+     * @param string $message
+     */
+    private static function assertReadWriteEquals($writeData, $message = "") {
```
MasterOdin: Remove the static as there's no real point to having it here. You can also get rid of the second parameter as it's unused?

site/tests/app/libraries/IniParserTester.php
```diff
@@ -3,13 +3,101 @@
...
+    public function testCRLF() {
+        // Pretty much directly from #2024
+        self::assertReadWriteEquals([
+            "course_details" => [
+                "zero_rubric_grades" => false,
+                "upload_message" => "By clicking \"Submit\" you are confirming that you have read, understand, and agree to follow the Academic Integrity Policy.<br />\r\n<br />\r\nwhy",
+                "keep_previous_files" => false
+            ]
+        ]);
+    }
```
MasterOdin: Let's turn all of this into using a [dataProvider](https://phpunit.de/manual/3.7/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers) such that then each test is only testing one specific parse, instead of a function having several. They'll probably run fast enough individually anyway.

site/tests/app/libraries/IniParserTester.php
```diff
@@ -3,13 +3,101 @@
 namespace tests\app\libraries;
 
 use app\libraries\IniParser;
+use PHPUnit\Runner\Exception;
```
MasterOdin: Remove this?

Issue #1979, opened 2018-05-18
Create general hwcron job daemon


MasterOdin:
Create a new daemon that runs as hwcron and can do various jobs (build assignments, generate repos, etc) on the system. It should listen to a queue folder to get job requests.

PR #2031, opened 2018-05-25
[WIP] delete gradeable feature

### LINKED ISSUE #570: Implementation plan for "Delete Gradeable"
(thought we already had an issue, but I don't see it)

proposal:  (for further discussion)

If a gradeable has no submissions associated with it and no TA grade data, then "delete" is allowed (with "are you sure" confirmation) and all trace of the gradeable is cleaned up, and that id can be reused.

If a gradeable has submissions associated with it, we can pop up a message when the instructor user tries to "delete" it, warning that submissions exist.  if the instructor user wants to proceed with delete, the folder will remain in the submissions directory.  Design decision to be made:  We can either allow the gradeable id to be reused, but those submissions are inherited by the new gradeable.  or we can say that id is no longer available (still leaving the submissions for backup/archival purposes).

If a gradeable has TA grading data in the database...  Design decision to be made:  We can either ask for confirmation via super cautious "are you REALLY sure" that all grading data will be deleted and the gradeable id can be reused.  Or maybe we can leave the data behind (juggling foreign key constraints???) and that if the id is reused that data is inherited by the new gradeable (this sounds much messier than the submissions files being inherited).  Or we just say no, you can't delete it.  You can set all dates to infinity so its only visible to instructors?

If a gradeable has both submissions and TA grading...  Design decision?

tushargr:
Closes #570 
Can only delete gradeable if it satisfies following condition-
1. if team assignment - no team (neither empty nor filled)
2. Should be Electronic gradeable
3. No submission for the assignment

issues in deletion-
1. Permission denied for some folder (related to gradeable and need to be deleted). For deleting a file within a directory we need to scandir() which requires x permission for directory and when deleting file we require w permission for directory consisting file. So to delete a file we need rwx all permission for directory. The problem is I guess user:group assigned to such folder (which need to be deleted) is 
hwcron:hwcron . I guess it should hwcron:sample_ta_www as the this group also consists of instructor who is actually trying to delete folder and files

### DIFF
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -11,6 +11,7 @@ use app\models\Gradeable;
 use app\models\GradeableComponent;
 use app\models\GradeableComponentMark;
 use \DateTime;
+use app\libraries\FileUtils;
 
 class AdminGradeableController extends AbstractController {
 	public function run() {
@@ -33,6 +34,9 @@ class AdminGradeableController extends AbstractController {
             case 'quick_link':
                 $this->quickLink();
                 break;
+            case 'delete_gradeable':
+                $this->deleteGradeable();
+                break;    
             default:
                 $this->viewPage();
                 break;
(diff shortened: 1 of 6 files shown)


MasterOdin:
What particular directories are you worried about trying to delete that have the hwcron permissions? We can also utilize #1979 to delete those directories as well (once the daemon is done).

bmcutler:
The folder permissions problem is related to a build gradeable bug.  I'm working on a PR that I'll get in this weekend on Monday.

site/data/course_tables.sql
```diff
@@ -923,7 +923,7 @@ ALTER TABLE ONLY late_days
 --
 
 ALTER TABLE ONLY peer_assign 
-    ADD CONSTRAINT peer_assign_g_id_fkey FOREIGN KEY (g_id) REFERENCES gradeable(g_id) ON UPDATE CASCADE;
+    ADD CONSTRAINT peer_assign_g_id_fkey FOREIGN KEY (g_id) REFERENCES gradeable(g_id) ON UPDATE CASCADE ON DELETE CASCADE;
```
tushargr: database change here

tushargr:
There is database change that I have done 
Just added ON DELETE CASCADE for peer_assign table as if we delete a gradeable then peer assigned for the gradeable should also be deleted

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -564,18 +568,44 @@ private function modifyGradeable($edit_gradeable) {
...
+
+    private function deleteGradeable() {
+        $g_id = $_REQUEST['id'];
+
+        if (!isset($_POST['csrf_token']) || $_POST['csrf_token'] != $this->core->getCsrfToken()) {
+            die("Invalid CSRF Token");
+            $this->returnToNav();
+        }
+        if (!$this->core->getUser()->accessAdmin()) {
+            die("Only admins can edit teams");
```
bmcutler: Error message should be able deleting gradeable, not editing teams

site/app/views/NavigationView.php
```diff
@@ -252,7 +253,33 @@ public function showGradeables($sections_to_list) {
                     $gradeable_title = '<label>'.$g_data->getName().'</label><a class="external" href="'.$g_data->getInstructionsURL().'" target="_blank"><i style="margin-left: 10px;" class="fa fa-external-link"></i></a>';
                 }
                 else{
-                    $gradeable_title = '<label>'.$g_data->getName().'</label>';
+                    if ($g_data->getType() == GradeableType::ELECTRONIC_FILE) {
```
bmcutler: What about the other types of gradeable?  They should be deleteable also, if they have no ta grade data entered for any student.

site/app/views/NavigationView.php
```diff
@@ -252,7 +253,33 @@ public function showGradeables($sections_to_list) {
...
+                        $no_teams_flag=1;
```
bmcutler: why not use true & false?  That might be more clear to future code maintainers.  

In general, this code logic should have comments documenting the checks, and the reason for the checks.

site/app/views/NavigationView.php
```diff
@@ -653,6 +680,28 @@ public function showGradeables($sections_to_list) {
...
+HTML;
+        return $return;
+    }
+
+    public function deleteGradeableForm() {
+        $return = <<<HTML
+    <div class="popup-form" id="delete-gradeable-form" style="width:550px; margin-left:-250px;">
+        <h2>Delete Gradeable</h2> 
+        <p>&emsp;</p>
+        <p>Note: A Gradeable can only be deleted if it do not have any team and submission.
```
bmcutler: Note: A gradeable can only be deleted if it has no student teams and if it has no student submissions.

tushargr:
@bmcutler 
made the changes as were requested

PR #2035, opened 2018-05-26
Add both rotating section and registration section on admin team form

### LINKED ISSUE #1976: Allow Team Rotating Sections to be Modified after Creation Time
At present, there is an option to set a team's registration section after creation time from the instructor team creation dialog on the student index page for a team assignment.

We should add a similar option to modify or add a rotating section for a team on this dialog.

tushargr:
Closes #1976 
Currently on Gradeable graded by registration section do not have rotation section on admin team form, so added it.
Also Gradeable graded by rotating section do not have registration section on admin team form, so added it.
So now independent of whether the gradeable is graded by reg. section or rot. section, the admin- team-form will have both option which can be used to modify registration and rotating section for team.


bmcutler:
please add which issue # this closes

jchulton:
An instructor can move a team to different rotational or registration section. If the grading is by rotational section, anyone who is assigned to the rotational section can grade the assignment. Similarly, if the grading is by registration section, anyone who is assigned to the registration section can grade the assignment. This all works well. Unfortunately, while testing I found that the stats page for both team and single assignments do not work for rotational grading assignments. The stats pages only refer to the assignment's registration section, which makes the stats page misleading for rotational grading.
For example, a student in registration section 5, rotational section 2 submits an assignment. A grader assigned to rotational section 2 and registration sections 3 and 4 would see that no students they were assigned to have submitted. But if they click grading details, they will see that they have been assigned to grade the above student.
I'm only bringing this up in case you noticing the strange stat pages. Since your changes work well, I'd recommend pushing them.

PR #2036, opened 2018-05-27
clangInstall

elizabethdinella:
Files to install clang. Edited so that not all of clang is installed - just what we need


.setup/clangInstall.py
```diff
@@ -61,8 +60,15 @@
...
 os.system(cmd2)
+<<<<<<< HEAD
+'''
+# Just build targets we need
+subprocess.call(["ninja", "install"])
+#subprocess.call(["ninja", "ASTMatcher", "UnionTool"])
+=======
 # Just build targets we need
 subprocess.call(["ninja", "ASTMatcher", "UnionTool"])
+>>>>>>> 449eee90bc75faca0e10cc3b1c6312fd35732a30
```
MasterOdin: It looks like you've got a merge conflict here?

emaicus:
This appears to install well both on primary and worker submitties. I found no errors.

PR #2038, opened 2018-05-28
[WIP] Regrade request system

### LINKED ISSUE #1027: Interface to allow students to request TA regrade through website
This should be configurable per gradeable by the instructor.  If enabled, the student can type a short message detailing the grade complaint.  The TA sees an indicator on their navigation panel and can view this message.

(This interface should be designed to discourage abuse of this feature with frivolous regrade requests.)

shailpatels:
Current status of Regrade Request System
- Students can request a regrade from the homework student view page
- TA/Instructors can respond through the electric grader page
- Student Index page shows any student who requested a regrade
- Added database tables

ToDo:
- [x]  Add option to allow/disallow students from making requests (when creating a gradeable)
- [x]  Allow students to delete request before any response
- [x]  Show regrade requests under stats
- [x] Add support for links (images?)
- [x] Make regrade request box in TA view persistent accross page refreshes 

capture

New Dragable box:
ta

Closes #1027


site/app/views/submission/HomeworkView.php
```diff
@@ -1214,9 +1214,140 @@ function downloadFile(file, path) {
...
+      <div class = "sub">
+        <div style="float: left; width: 50%"><h3>Regrade Discussion</h3></div>
+HTML;
+        $is_disabled = "";
+        $action = "";
+        $url = "";
+        $class = "btn-default";
+        if($gradeable->getRegradeStatus() === 0){
+          $message = "Request Regrade";
+          $action = "showPopUp()";
```
jchulton: I am getting an Uncaught ReferenceError when I try to click the Request Regrade button as below:
regraderequestuncaughtreferenceerror
It looks like the cause is that the ShowPopUp method is never defined. Maybe it was lost in the merging.

shailpatels: Whoops looks like I accidentally deleted that bit of code when I was cleaning it up. Should be good now. 
By design, the regrading interface was supposed to be hidden unless a student requests a regrade, once there is a request in process the yellow toolbar will have an additional button (right now its just a "!") which will open up another box for handling regrade requests. 
The design still needs a bit of finalization on the grader's side.

jchulton:
After a series of merges that have happened in recent days, this will need to be reexamined to work with the current system. Currently a server error occurs after trying to access a student's assignment.

site/app/views/submission/HomeworkView.php
```diff
@@ -1210,14 +1209,198 @@ function downloadFile(file, path) {
...
 
         return $return;
     }
+    public function showRequestForm($gradeable){
+      $thread_id = $this->core->getQueries()->getRegradeRequestID($gradeable->getId(), $gradeable->getUser()->getId());
+      $threads = $this->core->getQueries()->getRegradeDiscussion($thread_id);
+      $existsStaffPost = false;
+      foreach ($threads as $thread) {
+        if($this->core->getQueries()->isStaffPost($thread['user_id'])) $existsStaffPost = true;
+        break;
```
CouleeApps: Why are you immediately breaking from a loop?

site/app/views/submission/HomeworkView.php
```diff
@@ -1210,14 +1209,198 @@ function downloadFile(file, path) {
...
+            </form>
+          </div>
+HTML;
+        }
+        $return .= <<<HTML
+        <div class="modal" id="modal-container">
+          <div class="modal-content" id="regradeBox">
+            <h3>Request Regrade</h3>
+            <hr>
+            <p class = "red-message">Warning: Frivoulous requests may result in a grade deduction, loss of late days, or having to retake data structures!</p>
```
CouleeApps: Any reason this mentions data structures specifically?

site/app/views/submission/HomeworkView.php
```diff
@@ -1210,14 +1209,198 @@ function downloadFile(file, path) {
...
+              $first = false;                                      
+            }                                      
+            $function_date = 'date_format';                                      
+            $$return = <<<HTML
+            <div style="margin-top: 20px ">                                       
+              <div class = '$class' style="padding:20px;">                                       
+                <p>{$content}</p>                                      
+                <hr>                                       
+                <div style="float:right">                                      
+                  <b>{$name}</b> &nbsp
```
CouleeApps: Needs `&nbsp;`

site/public/css/diff-viewer.css
```diff
@@ -82,4 +82,29 @@
 .line_code {
     white-space: pre;
     padding-left: 10px;
+}
+
+.modal{
```
CouleeApps: Why is this in diff-viewer.css? Just use server.css

site/app/views/submission/HomeworkView.php
```diff
@@ -1210,14 +1209,198 @@ function downloadFile(file, path) {
...
+                <div style="float:right">                                      
+                  <b>{$name}</b> &nbsp                                       
+                {$function_date($date,"m/d/Y g:i A")}                                      
+                </div>                                       
+            </div>                                       
+HTML;
+          }
+        $return .= <<<HTML
+        <div style="padding:20px;">
+        <form method="POST" id="replyTextForm" action="{$this->core->buildUrl(array('component' => 'student',
```
CouleeApps: If there are no threads, $return is undefined at this point, throwing a warning.

CouleeApps:
image
Buttons should not be in lower case

CouleeApps:
The panel opening button should be placed with the others in the nav bar. Also, pressing it again should close the panel, removing the need for a dedicated "close panel" button, like how all the other panels work.

site/app/controllers/student/SubmissionController.php
```diff
@@ -58,12 +58,54 @@ public function run() {
...
+        $gradeable_id = (isset($_REQUEST['gradeable_id'])) ? $_REQUEST['gradeable_id'] : null;
+        $student_id = (isset($_REQUEST['student_id'])) ? $_REQUEST['student_id'] : null;
+        if($gradeable_id !== null && $student_id !== null) {
+            if($this->core->getQueries()->insertNewRegradeRequest($gradeable_id, $student_id, $content)){
+                $this->core->redirect($this->core->buildUrl(array('component' => 'student', 'gradeable_id' => $gradeable_id ) ) );
+            }else{
+                //handle error
+            }
+        }else{
+            //handle error
```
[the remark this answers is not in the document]
jchulton: That will be covered in a later pull request

site/app/views/submission/HomeworkView.php
jchulton: For the memes

CouleeApps: 👌👀

shailpatels:
@CouleeApps I'm not able to reproduce the issue your describing, replying to a post from the TA side works fine for me.

site/app/views/submission/HomeworkView.php
shailpatels: Instructors will have the option to write a specific warning message once the feature to allow/disallow requests is added, it looks half-implemented right now?

CouleeApps:
Yep, all the posts load now. I think that for-break thing was the culprit.

jchulton:
new version up, pls take a look

jchulton:
Looks like it the bug was the result of the merge, is resolved now

CouleeApps:
The reply box disappears from the grader's ui after replying. Is that intentional?

CouleeApps:
Adding a comment via the "Request further TA/Instructor action" button adds the same comment twice

CouleeApps:
Also, the [url] and [code] tags don't get rendered when viewing posts:

CouleeApps:
When the button hides because you have no regrade request, the panel should also hide (and disable hotkey if possible)

CouleeApps:
Students can request further action before requesting a regrade at all, which technically requests a regrade but does not seem desirable.

CouleeApps:
Students can delete requests after the TA has regraded them if they don't close/refresh the page. Sidenote: should students be allowed to delete requests at all?

jchulton:
The above push should solve of all the bugs except for the link code buttons. Since this is a WIP, I think we can address that later.

CouleeApps:
One last thing: If regrades are disabled the regrade area still appears but is blank:

CouleeApps:
Bug to deal with later: if you disable regrades while there are requests open, the list will still show "regrade requested"

bmcutler:
see image

screen shot 2018-06-07 at 2 40 35 am

CouleeApps:
You need to run `update_database.py`

bmcutler:
Ah, ok.  I'm going to make this PR test #2 of the migration system :)  So we'll do the update in the new format...
Also, can someone please undo the whitespace changes to the .sql file?  Thx

bmcutler:
@MasterOdin Is going to convert the update_database commands to the new migration format.  Then we'll merge the PR.

MasterOdin:
So, could I get a sparknotes overview of what the two tables are for and kind of data they hold (and how they relate to other tables).  What's the difference between "regrade_id" and "request_id" in regrade_discussion? Can a student make multiple requests?

bmcutler:
A regrade request should be a sequence of posts initiated by the student, generally alternating between student & grader(s) (but not necessarily...  someone can post more than once before the other side replies), with a state open vs. resolved.  (by default, its 'open' if the last post is a student post, and 'resolved' if the last post is a grader post, but anyone can edit the state)

I think one id for the sequence should be sufficient (the thread id in the forum?).  but each post should have its own id (the post id from the forum?).

Hopefully its sharing some of the code, if not using the same tables as the forum system.

MasterOdin:
So we'd want to have a primary key on both ID columns (mostly for convenience for the foreign key constraint between them and other tables) and then a unique constraint on `regrade_requests` on `(gradeable_id, student_user_id)`. `status` could then be changed to be called `is_open` or something and just be a boolean flag (instead of an integer that we treat as a 0/1 boolean?).

I'm still not sure what the difference between `regrade_id` and `thread_id` is.

bmcutler:
Maybe the int was used to differentiate no request & cancelled/deleted request?  (But I vetoed that feature).  A boolean should be sufficient.

Perhaps we should call it regrade_thread_id and regrade_post_id?  (I haven't looked at the code)

MasterOdin:
Reading over the code, it looks like `thread_id` is a link to `regrade_request` and `regrade_id` links to gradeable_id (but isn't really used anywhere). `regrade_request` is a one row per gradeable per user and `regrade_discussion` is then multiple rows per request (with different authors and such), but links back to the single gradeable.

Given that, we can change the following tables to be:
```
CREATE TABLE regrade_requests (
    id serial NOT NULL PRIMARY KEY,
    gradeable_id VARCHAR(255) NOT NULL,
    student_id VARCHAR(255) NOT NULL,
    timestamp TIMESTAMP NOT NULL,
    status INTEGER DEFAULT 0 NOT NULL
);
```
with gradeable_id being a foreign key to gradeable and student_id being a foreign key to users. gradeable_id and student_id must be unique (one regrade request per gradeable per student).

```
CREATE TABLE regrade_discussion (
    id serial NOT NULL PRIMARY KEY,
    regrade_id INTEGER NOT NULL,
    user_id VARCHAR(255) NOT NULL,
    timestamp TIMESTAMP NOT NULL,
    content TEXT,
    deleted BOOLEAN DEFAULT FALSE NOT NULL,
);
```
with regrade_id being a foreign key to regrade_requests, user_id being a foreign key to users, etc. This drops the old (useless) regrade_id and renames thread_id to the more informative regrade_id (or we could call it regrade_request_id). This change should just be a search and replace on the edited lines code in the PR.

site/app/models/Gradeable.php
```diff
@@ -1168,6 +1169,10 @@ public function getComponentsGradedBy($grader_id) {
         return $return;
     }
 
+    public function getRegradeStatus(){
+        return $this->regrade_status;
+    }
+
```
MasterOdin: you don't need this function (though please add it to the top of the file in that big method comment block so that PHPStorm and IntelliJ recognize that it exists). The magic `__call()` function in the AbstractModel maps `getRegradeStatus()` to return `$this->regrade_status` automatically.

jchulton:
So the above pushes should change the regrade_request and regrade_discussion tables as intended, and seem to work on my local machine. However various migrations (like alphanumeric_sections) appear to be failing...

site/app/libraries/ExceptionHandler.php
```diff
@@ -122,18 +122,18 @@ public static function handleException($exception) {
...
-        }
-        else {
+       // }
+       // else {
             return <<<HTML
 An exception was thrown. Please contact an administrator about what you were doing that caused this exception.
 
 HTML;
-        }
+        //}
```
MasterOdin: why?

jchulton:
Before this is merged, the build permissions error we discussed at 6/15's daily meeting needs to be resolved.

PR #2039, opened 2018-05-28
Add late and multiple submission sample

### LINKED ISSUE #1164: "Midsemester" sample course should have late assignments
Splitting this off from #830. Currently all submissions are on time in the sample course.

In "sample", there should be some submissions that cover some or all of the following:

- [ ] Too late (e.g. 5 days late with only 2 late days per semester)
- [ ] Too late (e.g. 5 days late with only 2 late days allowed for the assignment)
- [ ] On time only because extensions
- [ ] On time because of late days (with no late days used)
- [ ] On time where additional late days were required (the "default" starting amount wouldn't be enough)
- [ ] On time because of late days (but late days used previously)
- [ ] Too late because prior usage of late days (e.g. student has 2 days for the semester, but used 1 previously and is trying to use 2 on this submission)
- [ ] On time (no late days used, no extensions used)

EDITED TO ADD:  All of these cases should be exercised in new Selenium tests, including the messages displayed to students at the top of the submission page to show when late day information for the active version, or cancelled version, or no submission, and what the late day status will be if the student submits/re-submits right now.   See also #1767 and #1754.

saxocellphone:
Closes #1164, but not all cases are guaranteed because the late days are generated pseudo-randomly. I played around with the numbers and found a pretty good configuration.


holzbh:
After latest commits, almost all gradeables that used to be present are gone. The sample course for example, now only has "Open homework" and "Open team homework". The tutorial course only has 01 Simple Python. This is a problem.

saxocellphone:
I commented out all the yml for testing, but I forgot to uncomment it, oops

.setup/data/courses/sample.yml
```diff
@@ -255,10 +255,75 @@ gradeables:
   g_grade_start_date: 9997-12-31 23:59:59
   eg_team_assignment: true
   eg_max_team_size: 3
-  eg_max_random_submissions: 0
```
scopeInfinity: Why do we need to update 'open_team_homework' config?

.setup/bin/setup_sample_courses.py
```diff
@@ -753,45 +753,59 @@ def create(self):
...
+                        active_version = random.choice(range(1, versions_to_submit+1))
+                        json_history = {"active_version": active_version, "history": []}
+                        random_days = 1
+                        if random.random() < 0.3:
+                            random_days = random.choice(range(-3,2))
+                        for version in range(1, versions_to_submit+1):
+                            os.system("mkdir -p " + os.path.join(submission_path, str(version)))
+                            submitted = True
+                            submission_count += 1
+                            current_time_string = dateutils.write_submitty_date(gradeable.submission_due_date - timedelta(days=random_days))
```
scopeInfinity:
```bash
root@vagrant:/var/local/submitty/courses/s18/sample/submissions/open_homework/student# cat user_assignment_settings.json 
{
    "active_version": 2,
    "history": [
        {
            "version": 1,
            "type": "upload",
            "who": "student",
            "time": "9996-12-30 23:59:59-0500"
        },
        {
            "version": 2,
            "time": "2018-05-31 14:16:22-0400 America/New_York",
            "who": "student",
            "type": "upload"
        }
    ]
}root@vagrant:/var/local/submitty/courses/s18/sample/submissions/open_homework/student# cat 2/.submit.timestamp 
2018-05-31 14:16:22-0400 America/New_York
```
Timezone is missing. Not sure if it will be an issue.

.setup/data/courses/sample.yml
saxocellphone: This change shouldn't be in this PR, nice catch. I'm currently working on sample submission for team assignments, and I made some changes to the yml file. I'll undo this line from this PR

.setup/bin/setup_sample_courses.py
saxocellphone: I noticed that before, but I couldn't figure out how to add the time zone. However I haven't encountered any issues without the timezone.

bmcutler: The important part of the timezone is the -0500 and -0400.  The location is just for human-readability, its ok that its not there.  I am worried why they don't match. Are we hardcoding the 0500 somewhere?  (I think that's standard time and the 0400 is daylight savings)

holzbh: The timezone should probably be investigated, but I've approved the other changes since the courses are populated and seem consistent, as well as having a good variety of situations.

PR #2040, opened 2018-05-28
Add forum sample that highlights display options, plus attachments

### LINKED ISSUE #1972: expand sample course forum posts to highlight/demo hierarchy vs alpha vs chrono

saxocellphone:
closes #1972. I made a sample thread that can demonstrate the expected behavior of all three display options (chronological, alphabetical, hierarchical). One can now write tests to see if the posts are in the correct order. 

In addition, I made another thread that contains attachments.


site/app/controllers/forum/ForumController.php
```diff
@@ -208,7 +208,8 @@ public function publishPost(){
 
             if($hasGoodAttachment == 1) {
                 $post_dir = FileUtils::joinPaths($thread_dir, $post_id);
-                FileUtils::createDir($post_dir);
+                var_dump(is_dir($post_dir));
```
andrewaikens87: Please remove this line

.setup/bin/setup_sample_courses.py
```diff
@@ -850,7 +850,8 @@ def create(self):
                 conn.execute(forum_thread_cat.insert(), thread_id=thread_id, category_id=threadData[6])
             counter = 1
             for postData in f_data[0]:
-                if(postData[10] != "f"):
+                if(postData[10] != "f" && postData[10] != ""):
```
KevinMackenzie: in python, you need to use `and` instead of `&&`

PR #2047, opened 2018-05-29
Improve grading responsiveness with async saving

CouleeApps:
Rework of #2012: This makes the component boxes save and load their data asynchronously so the page feels much faster. 

## Previews:

After clicking the component before the contents load:

screen shot 2018-05-29 at 15 18 30

After selecting marks and closing the component:

screen shot 2018-05-29 at 15 18 54

### DIFF
--- a/site/public/js/ta-grading-mark.js
+++ b/site/public/js/ta-grading-mark.js
@@ -772,8 +772,11 @@ function saveMark(num, gradeable_id, user_id, active_version, gc_id, your_user_i
     calculatePercentageTotal();
 
     var gradedByElement = $('#graded-by-' + num);
+    var savingElement = $('#graded-saving-' + num);
     var ungraded = gradedByElement.text() === "Ungraded!";
-    gradedByElement.text("Saving...");
+
+    gradedByElement.hide();
+    savingElement.show();
 
     var overwrite = ($('#overwrite-id').is(':checked')) ? ("true") : ("false");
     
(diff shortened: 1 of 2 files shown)


KevinMackenzie:
There are a few different methods where the "Saving..." message will appear and not be replaced by the "Graded by..." message.  The first is when opening and closing a component without changing any marks.  The second is when overwriting a previously saved mark.

1. To reproduce, open the TA grading interface for a student with a graded component.  Open that component and close it without changing the mark.  The "Saving..." message will not go away despite the operation completing.
2. To reproduce, open the TA grading interface for a student with a graded component.  Open that component, change the assigned mark, then close it.  The "Saving..." message will not go away despite the operation completing.

site/public/js/ta-grading-mark.js
```diff
@@ -758,22 +771,24 @@ function saveMark(num, gradeable_id, user_id, active_version, gc_id, your_user_i
...
-                }
+        if (all_false === true) {
+            gradedByElement.text("Ungraded!");
+        } else {
+            if(ungraded || (overwrite === "true")) {
+                gradedByElement.text("Graded by " + your_user_id + "!");
+                var question_points = parseFloat(current_question_num[0].innerHTML);
+                var max_points = parseFloat(current_question_num[0].dataset.max_points);
+                $('#summary-' + num)[0].style.backgroundColor = "#FCFCFC";
+                $('#title-' + num)[0].style.backgroundColor = "#FCFCFC";
```
KevinMackenzie: It appears like these conditions may be incorrect for updating the text.

CouleeApps:
I can't seem to reproduce the issues you're having. Are you getting any errors in the web inspector?

PR #2026, opened 2018-05-25
WIP Grading interface

jchulton:
WIP. closes #1536 closes #1860. Will close #2003.
This is my first draft of the Redesign of the TA Grading Interface. I'll still working on some minor changes, like the font of the marks, but this is my plan for the general layout.
Known bug: After a mark is graded, the Graded By section displays a number, instead of the instructor/TA name. This is fixed after refreshing the page. I'm guessing this is because of changes to Ajax functions? Not really sure.

PR #2056, opened 2018-05-30
Grading interface improvements

### LINKED ISSUE #1543: custom mark text should not be allowed to be empty
We have a bug...  if the custom message is blank, the custom point is not displayed in the rubric box (dunno about the hw report).

longer term: Should not be allowed to just fill in a +/- point value for custom, it must also have a message.

### LINKED ISSUE #2023: "no credit" / "full credit" mark should not be editable
"no credit" / "full credit" is based on whether the component is count up or count down.  
It should look visually similar to the other common marks....  but
You should not be able to edit the points or text.  
(It might look a little different to hint that editing is not possible.)
It should always be the first in the list of marks.

jchulton:
Closes #2023 Closes #1543 Takes care of various issues discussed in the pull request WIP Grading interface #2026. Specifically:
Adding the Add Common Mark Button to the box containing non-common marks
Editing Existing mark no longer makes text box get taller
Cancel/Done text now aligns
Clicking to expand the General Comment box improved.


CouleeApps:
>Editing Existing mark no longer makes text box get taller

Curious why you're doing this. Having vertical scroll in the boxes makes it harder to read them.

site/public/js/ta-grading-mark.js
```diff
@@ -42,18 +42,27 @@ function checkIfSelected(me) {
...
+    if(x==0){
+        color="#e6e6e6";
+        editable="readonly";
+        appearEditable="cursor: not-allowed; border:none; outline:none;";
+    }
+    //onkeyup="autoResizeComment(event) removed from textarea
     return ' \
 <tr id="mark_id-'+num+'-'+x+'" name="mark_'+num+'" class="'+(is_publish ? 'is_publish' : '')+'"'+(is_new ? 'data-newmark="true"' : '')+'> \
-    <td colspan="1" style="'+background+'; width: 90px; text-align: center;"> \
+    <td colspan="1"; width: 90px; text-align: center;"> \
```
CouleeApps: Should just remove background without removing the opening `style=` attribute.

site/public/js/ta-grading-mark.js
```diff
@@ -42,18 +42,27 @@ function checkIfSelected(me) {
 }
 
 function getMarkView(num, x, is_publish, checked, note, pointValue, precision, min, max, background, gradeable_id, user_id, get_active_version, question_id, your_user_id, is_new) {
+    editable="";
+    appearEditable="";
+    color=background;
```
CouleeApps: Should use `var` otherwise these will become global variables

site/public/js/ta-grading-mark.js
```diff
@@ -447,16 +454,32 @@ function calculateMarksPoints(question_num) {
...
+            $('#mark_points_custom-' + question_num)[0].value="0";
+        }
+        if (isNaN(custom_points)) {
+            current_points += 0;
+        } 
+        else {
+            if(custom_points!=0){
+                current_points += custom_points;
+                any_selected = true;
+            }
```
CouleeApps: Selecting only a custom mark with zero points makes it look like you have graded the assignment but shows the score as not graded:

CouleeApps:
Oh I see, https://github.com/Submitty/Submitty/pull/2026#pullrequestreview-124028504 (first bullet point).

Personally I preferred having the boxes expand as it meant you could easily enter multiple lines of text.

jchulton:
Implemented changes above. Please take another look when you can.

PR #2058, opened 2018-05-30
Allow Primary Submitty to Update Worker Machines at Install Submitty Time

### LINKED ISSUE #1832: Coordinate submitty system upgrade/installation of worker machines from primary machine
When INSTALL_SUBMITTY.sh is run on the primary machine, it should notify halt the shipper & worker daemons on all machines, then (download?)/install software on all machines, then restart daemons on all machines.  

Requires communication & correct permissions.  

And the whole thing needs to be properly documented.

emaicus:
Closes #1832

This pull request:
*  Allows the install submitty script to update all reachable worker machines. The machines' daemons are stopped, code from the primary submitty is rsynced to the worker, the worker runs install_submitty, and then the worker daemons are restarted.
*  Adds a script which allows privileged users on primary submitty to interface with the worker and shipper daemons on the worker machine. This allows administrators to easily run a stop or start all workers command.
*  Allows admins of a worker machine to specify the 'Submitty Supervisor,' the user to which hwcron on primary submitty can ssh.
*  Allows administrators to enable/disable individual workers via a new field in the autograding_workers.json
*  Adds tracking of which git version/tag we are on. These variables are stored in submitty.json.
*  Splits out a significant portion of grade_item.py into packer_unpacker.py, fixing an error which was causing workers to fail upon not finding autograding_workers.json.
*  Stops primary submitty from creating shipper threads for workers which cannot be reached.


.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -367,14 +372,15 @@ def write(x=''):
...
+                "enabled" : True
             }
         }
 
         with open(WORKERS_JSON, 'w') as workers_file:
             json.dump(worker_dict, workers_file, indent=4)
-    shutil.chown(WORKERS_JSON, 'root', HWCRON_GROUP)
-    os.chmod(WORKERS_JSON, 0o440)
+    shutil.chown(WORKERS_JSON, HWCRON_USER, 'root')
+    os.chmod(WORKERS_JSON, 0o640)
```
MasterOdin: why are we changing the permissions for the file?

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -204,6 +211,21 @@ if [ "${WORKER}" == 0 ]; then
...
+        echo "" >> /etc/sudoers
+        echo "#grant the submitty user on this worker machine access to install submitty" >> /etc/sudoers
+        echo "%${SUBMITTY_SUPERVISOR} ALL = (root) NOPASSWD: ${SUBMITTY_INSTALL_DIR}/.setup/INSTALL_SUBMITTY.sh" >> /etc/sudoers
+        echo "#grant the submitty user on this worker machine access to the systemctl wrapper" >> /etc/sudoers
+        echo "%${SUBMITTY_SUPERVISOR} ALL = (root) NOPASSWD: ${SUBMITTY_INSTALL_DIR}/sbin/shipper_utils/systemctl_wrapper.py" >> /etc/sudoers
+    fi
+fi
+
+echo "#grant the hwcron user access to the daemon utilities" >> /etc/sudoers
+echo "%${HWCRON_USER} ALL = (root) NOPASSWD: ${SUBMITTY_INSTALL_DIR}/sbin/shipper_utils/systemctl_wrapper.py" >> /etc/sudoers
```
MasterOdin: This probably should be in `install_system.sh` as right now, this will just append these lines each time you run `INSTALL_SUBMITTY.sh` which we probably don't want for `/etc/sudoers`.

.travis.yml
```diff
@@ -238,6 +240,7 @@ jobs:
         - pip3 install -U pip
         - pip3 install PyYAML
         - pip3 install python-dateutil
+        - pip3 install paramiko
```
MasterOdin: We don't need paramiko to test the C++ autograding code?

sbin/shipper_utils/systemctl_wrapper.py
```diff
@@ -0,0 +1,210 @@
...
+      disable_machine(target)
+
+  return correct
+
+if __name__ == "__main__":
+  #parse arguments
+  parser = argparse.ArgumentParser(description='A wrapper for the various systemctl functions. \
+    This script must be run as the submitty supervisor.',
+    epilog='ERROR CODES: 0 = After performing the specified action, the requested daemon is dead, 1 = After performing the \
+    specified action, the requested daemon is alive, 2 = Failure performing operation, 3 = invalid mode, daemon, or target')
```
MasterOdin: For readability, we probably want to put in newlines and do something like: 
```
epilog="""ERROR CODES:
0 = After performing the specified action, the requested daemon is dead
1 = After performing the specified action, the requested daemon is alive
2 = Failure performing operation
3 = invalid mode, daemon, or target"""
```

sbin/shipper_utils/systemctl_wrapper.py
```diff
@@ -0,0 +1,210 @@
...
+  parser.add_argument("--daemon", metavar="DAEMON", type=str, help="Optional. Valid options are worker and shipper. \
+    If unspecified, worker is assumed.")
```
MasterOdin: Add `choices=['worker', 'shipper']` and change type to `type=str.lower`. Then get rid of the if statement in setting it.

It seems you actually probably want to add `str.lower` for the type of all of your arguments.

sbin/submitty_autograding_worker.py
```diff
@@ -27,6 +27,8 @@
 
 JOB_ID = '~WORK~'
 
+ALL_WORKERS_JSON = os.path.join(SUBMITTY_DATA_DIR, "autograding_TODO", "autograding_worker.json")
```
MasterOdin: what is the point of this file when we've got `${SUBMITTY_INSTALL_DIR}/config/autograding_worker.json`?

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -0,0 +1,109 @@
...
+        exit_code = run_systemctl_command(worker, 'stop')
+        if exit_code != 0:
+          print("Could not turn off {0}'s daemon. Please allow rsynching to continue and then attempt another install.".format(worker))
+
+      local_directory = submitty_repository
+      remote_host = '{0}@{1}'.format(user, host)
+      foreign_directory = submitty_repository
+
+
+      # rsynch the file
```
MasterOdin: `rsync`

.setup/bin/track_git_version.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python3
```
MasterOdin: I'm against writing this to `submitty.json`. Given that we're not really using that info really anywhere meaningful, let's just write this to its own file `version.json` or something and that we don't care if it potentially gets blown away.

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -553,6 +578,13 @@ if [[ "$is_active_tmp" == "0" ]]; then
 fi
 # END TO BE DELETED
 
+if [ "${WORKER}" == 0 ]; then
+    # Stop all foreign worker daemons
+    echo -e "\nStopping worker machine daemons"
+    sudo -H -u ${HWCRON_USER} ${SUBMITTY_INSTALL_DIR}/sbin/shipper_utils/systemctl_wrapper.py stop --target perform_on_all_workers
+    echo -e "\n"
+fi
```
MasterOdin: Why is this necessary? Given that we stop the shipper/worker on `INSTALL_SUBMITTY.sh`, why do we need to stop the workers here explicitly instead of just letting the `INSTALL_SUBMITTY.sh` process handle that (and restarting the workers if they're running.

sbin/autograder/packer_unpacker.py
```diff
@@ -0,0 +1,301 @@
...
+
+    submission_path = os.path.join(SUBMITTY_DATA_DIR,"courses",item_name)
+
+    is_batch_job = queue_obj["regrade"]
+    gradingtime=queue_obj["gradingtime"]
+    grade_result=queue_obj["grade_result"]
+
+    print (which_machine,which_untrusted,"unzip",item_name, " in ", int(gradingtime), " seconds")
+
+    grade_items_logging.log_message(job_id,is_batch_job,"unzip",item_name,"grade:",gradingtime,grade_result)
```
MasterOdin: I dislike the fact that a module named `packer_unpacker.py` is when asked to unpack a gradeable is via `unpack_grading_results_zip`:
1. unpacking the gradeable details
1. writing the gradeable data to the database
1. printing a message to the console about grading time
1. logging the details of the grading

which isn't great. This isn't really a thing we need to change for this PR, but we really should try and improve the naming and structure of the autograder code at some point.

sbin/shipper_utils/systemctl_wrapper.py
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/env python3
```
MasterOdin: Should we move all of the relevant starting/stopping of systemctl out of `INSTALL_SUBMITTY_HELPER.sh` to use this module?

.setup/CONFIGURE_SUBMITTY.py
emaicus: hwcron now needs write access, as it is able to update the 'enabled' field, turning off workers as necessary.

.setup/INSTALL_SUBMITTY_HELPER.sh
emaicus: Talked about this in person, moving the lines to `install_system.sh` and wrapping a missing check around it.

.travis.yml
emaicus: Paramiko is now needed to run `INSTALL_SUBMITTY.sh`. SSH is not used, but paramiko is imported into `update_and_install_workers.py`.

sbin/shipper_utils/systemctl_wrapper.py
emaicus: I will do this.

sbin/shipper_utils/systemctl_wrapper.py
emaicus: I didn't know this was a possibility. I will absolutely update this.

emaicus:
@bmcutler @MasterOdin A permissions question. In order to rsync to the worker machine, the `Submitty_Supervisor` user on that machine must have write access to all files in the `GIT_CHECKOUT_Submitty` folder. Right now, we are using `chgrp` at the start of `INSTALL_SUBMITTY.sh` to change `GIT_CHECKOUT_Submitty's` group to be that of the supervisor. However, the write bit is currently unset. 

How would you like me to proceed?

bmcutler:
you want write access to the git checkout machine on the worker machine?  Yes, by all means put a chmod in there.

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -98,6 +97,14 @@ function replace_fillin_variables {
 
 ########################################################################################################################
 ########################################################################################################################
+# if this is not a worker machine, make sure that the submitty checkout belongs to the hwcron group
+if [ "${WORKER}" == 1 ]; then
+    chgrp -R ${SUBMITTY_SUPERVISOR} ${SUBMITTY_REPOSITORY}
+else
+    chgrp -R ${HWCRON_GID} ${SUBMITTY_REPOSITORY}
```
bmcutler: May want to also add a chmod g+r here too (just incase it's 700 or 600)

.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
```diff
@@ -75,3 +76,11 @@ done
 chown -R root:${HWCRON_USER} ${SUBMITTY_INSTALL_DIR}/sbin/autograder
 chmod 750 ${SUBMITTY_INSTALL_DIR}/sbin/autograder
 chmod 550 ${SUBMITTY_INSTALL_DIR}/sbin/autograder/*
+
+if [ "${WORKER}" == 1 ]; then
+    chown -R root:${SUBMITTY_SUPERVISOR} ${SUBMITTY_INSTALL_DIR}/sbin/shipper_utils
+else
+    chown -R root:${HWCRON_USER} ${SUBMITTY_INSTALL_DIR}/sbin/shipper_utils
```
bmcutler: _USER here is incorrect.  it should be _GROUP.

But more important we should be consistent.  In some places we're doing root:hwcron_group in other places we're doing hwcron_user:root.  

If hwcron_group only contains hwcron (and I hope this is true), maybe the group is better (contradicting what I said to do earlier...)  

And let's make sure hwcron only has read in most of these instances, no write (unless of course it needs write)

sbin/autograder/packer_unpacker.py
bmcutler: Can you be more specific about what isn't great?  Is it doing too much?  I think it makes sense.  
(But yes, let's hold off on whatever change you would like to see for another issue/PR)

MasterOdin: I'd say it's doing too much for what the function and module are named. I'd assume a module named packer_unpacker.py would only deal with packing and unpacking zips.

Issue #2048, opened 2018-05-29
Factor component table into a reusable script


CouleeApps:
So create/edit gradeable, and the grading interface can share the same ui.

CouleeApps:
See gradeable-view-js branch for the start of this. First step is mostly complete.

KevinMackenzie:
Making this abstraction is waiting on the electronic grading controller/view to use the new gradeable model since the create/edit gradeable page is using the new model

PR #2061, opened 2018-05-31
Port grading rubric view to Twig/JS

### LINKED ISSUE #2066: Custom mark for each component is always visible in TA grading rubric panel
(isn't expanding/collapsing with the common marks for each component)

CouleeApps:
This replaces the large PHP function for generating the rubric table with lots of Twig, using a build of Twig.js to render it on the browser instead of the server. 

Also included is converting the large amount of dataset attributes to a single page-wide Gradeable object (with handy `getGradeable()` / `getComponent()` / `getMark()` functions) that makes the functions not require 10 arguments. 

I would appreciate it if someone would try to break this. There is quite a bit of functionality in the design and I would like to know if I missed anything. 

Closes #2066


MasterOdin:
We're using twig.js primarily so that you can define the mark template once and then have it be loaded up on demand and then used when adding additional marks?

CouleeApps:
This is the first part of #2048; it moves the grading rubric view to JS so it's no longer tied to php. I used twig.js to render the whole rubric ([here](https://github.com/Submitty/Submitty/pull/2061/files#diff-269a6213d7d26cf7c772a433c78abd9bR6)) and it has the added benefit of letting me create new marks with a template as well.

site/app/models/Gradeable.php
```diff
@@ -1138,4 +1138,64 @@ public function getComponentsGradedBy($grader_id) {
         }
         return $return;
     }
+
+    /**
+     * Get an associative array of all the data needed to render this gradeable and its various components and their marks.
+     * Does not contain submission-specific data like comments or which marks are selected.
+     * @return array
+     */
+    public function getStaticData() {
```
KevinMackenzie: Do we need this function?  I'm pretty sure the `AbstractModel` class has a `ToArray()` method for this purpose.

KevinMackenzie: if we don't mind sharing the static vs grading data, that is.

CouleeApps:
Check if there's an error thrown in the web inspector, but there shouldn't be anything extra needed.

site/app/models/Gradeable.php
MasterOdin: He's filtering out components with a -1 order (not sure when that would happen though).

KevinMackenzie:
403 status code when accessing templates

site/app/models/Gradeable.php
CouleeApps: Was not aware of `toArray` but looking at it now it's a lot more data than we need or would want, especially for peer grading (if that ever happens).

CouleeApps:
Sounds like public/templates is probably not chowned to hwphp. I get that sometimes when I copy files directly as root.

KevinMackenzie:
folder permissions were find and hwphp owned everything, but `.twig` permissions were `-r--r-----` instead of `-r--r--r--`

KevinMackenzie:
image
Should the custom mark still be visible below the component?

site/public/templates/grading/Component.twig
```diff
@@ -0,0 +1,124 @@
...
+    </div>
+    {# /Add new mark #}
+</div>
+
+{# Custom mark #}
+<div class="box" id="mark_custom_id-{{ c_index }}" name="mark_custom_{{ c_index }}">
+    <div class="box-title">
+        <span colspan="1" style="text-align: center; white-space: nowrap;">
+            <span onclick="event.stopPropagation();">
+                <i class="fa {{ component.score == 0 and component.comment == "" ? "fa-square-o" : "fa-square" }} mark fa-lg" name="mark_icon_{{ c_index }}_custom" id="mark_icon_custom-{{ c_index }}" style="visibility: visible; cursor: pointer; position: relative; top: 2px;"></i>&nbsp
```
MasterOdin: `&nbsp;`

CouleeApps:
Yep, that's the current design of the grading page as of #2056

KevinMackenzie:
image
When adding a new common mark, on all components but the first, the component collapses when the `Add New...` button is pressed.  Upon pressing `create` in the popup, the mark doesn't appear to be added to the component.

KevinMackenzie:
image

When adding text to a blank custom mark, the 'points' box doesn't appear to be editible (red cursor) until the UI refreshes or user clicks the box.

bmcutler:
no.... the custom mark should not be outside of the expand/collapse.  Please fix #2066

MasterOdin:
What exactly is bad about tying the grading rubric to PHP? I'm assuming the main goal is that you want to be able to create new components/marks and do it using twig templates? Or is there something more you're looking to do? It's not really clear exactly what the end goal here is, especially given that you're using twig.js to load stuff that's fairly static during page usage (the gradeable container around the components and marks).

MasterOdin:
I'm also concerned about the comment on their [https://github.com/twigjs/twig.js/wiki/Implementation-Notes]wiki implementation notes[/url]:
> currently twig.js does not have the same auto-escaping that Twig does.

CouleeApps:
> I'm assuming the main goal is that you want to be able to create new components/marks and do it using twig templates

Yes, the end goal is to make something that can also be used with the create/edit gradeable page which has dynamic components and marks. While this particular page can be done on php as well, it would split up the templates across server/client and make it more complicated to use.

MasterOdin:
Wouldn't an alternative approach be to at the bottom of the page (or top) to contain a blank component/mark in hidden divs and then whenever you need to add a new one to the form, just copy that into the appropriate place.

What are we gaining by having that process be purely through twig as the empty component/mark isn't really going to have anything dynamically put in? 

I'm also concerned that we could be using two variants of the template engine with potentiallly different behavior in some cases.

CouleeApps:
Post-meeting discussion results: this is the best option we have for now because parts of the ui can dynamically generate and have lots of things that need to be pasted. In the future we should look into rendering the gradeable/components/marks in PHP Twig and having it return HTML that JS can then use.

KevinMackenzie:
Clicking on the disabled `points` numeric input for the custom mark collapses the component

KevinMackenzie:
Clicking on the "people who got" on a mark displays the `Saving...` message similar to when collapsing a component.

CouleeApps:
> Clicking on the "people who got" on a mark displays the Saving... message similar to when collapsing a component.

Because it does in fact save before checking. You just couldn't tell before. Should it?

Issue #1900, opened 2018-04-25
sample course needs more team related gradeables


bmcutler:
* an open gradeable, team lock date has not passed, not all students have formed teams
* an closed gradeable with all teams submitted, but not yet graded

PR #2064, opened 2018-05-31
Team assignment submissions for sample course

### LINKED ISSUE #1287: Add teams and submissions to sample team assignment
Currently there is only one team assignment with no teams or submissions in the Sample course.

setup_sample_courses.py should create randomized teams and submissions for them. We should also have team assignments in the grading stage and with past team lock dates.

saxocellphone:
Most of #1287 and most of #1900. I will work on making some students have no teams, but I think this is a good stopping point for now. 

There are only two team gradeables right now (The two are from #1900), but I can easily add more now that setup_sample_course supports team submissions


saxocellphone:
I fixed the team_history variable missing issue.
For the first issue, I do not think it is due to my changes. I checked the database after setting up sample course and it looks all right. A new issue should be opened to fix that bug

jchulton:
Looks good

PR #2068, opened 2018-06-01
Refactored and Reorganized Create/Edit Gradeables Instructor Interface

KevinMackenzie:
I don't think this closes any issues.

Reorganized the create/edit gradeables interface for the instructor.  This splits the interface into a few different tabs based on different gradable settings.  This also changes the work flow of creating a gradeable.  Before changing most of the settings, the user must choose some base settings and submit the form, which will redirect them to the next page of the form in edit mode, at which point the unchangeable settings, such as type and id are locked in.  

The information and validation should be identical to the previous interface, so please notify me of any omissions / errors.


bmcutler:
When you make a new gradeable, and you press "submit", you are currently kicked back to the navigation page.  Instead, make it stay on the page and allow edits to that gradeable.

I'm editing a gradeable, electronic upload, with to ta grading.  But the "rubric" & "grader assignment"  panels are still visible (but totally empty).  Also, the manual grading open date should be hidden for this gradeable.

Not currently copying # of checkpoints/numerics for template creation of checkpoint/numeric gradeable.  (definitely want this eventually, we use it all the time for prepping the labs each week)

bmcutler:
The order of the questions on the general page should be:

from template?

what is the title?

what is the url with instructions?

what is the unique id?

what type of gradeable? (e/c/nt)

what is the syllabus category?

(we can discuss, but I think this order is more intuitive than the current order)

bmcutler:
The scary  "(Grading has started! Edit Questions At Own Peril!)"  Message should only appear on the Rubric panel.

bmcutler:
"How many late days" should default to the setting from the "Course Settings" page (Default Maximum Late Days Per Assignment)

Can we left justify the "Submit" button at the bottom of the general tab (if you have a narrow window it might be off screen when its right justified)
Same thing for "Save Changes" button on the other tabs.

There should be no due date for the checkpoint or numeric gradeable.  Manual grade date must be >= beta testing date.

An electronic gradeable with no TA grading...  the grades related date should be must be >= due date + late days (not manual grading)

On the numeric/text rubric edit panel, instead of "MAX SCORE" (the all caps one), make it "Total"

The "What are the overall instructors to TA" box should be on checkpoint & numeric, NOT on electronic gradeable.

bmcutler:
The labels for the rubric for the checkpoint gradeable get erased when you go to edit the page (and then they are saved as blank too).

Let's change the "Add" and "Remove" buttons on the checkpoint rubric to be "Add Checkpoint" and "Remove Last Checkpoint".

Since we can't remove the last checkpoint (good feature), we should also always start with 1 checkpoint. (currently we start with 0 checkpoints).

There should be no due date for checkpoint or numeric gradeables.  
The manual grading open date for checkpoint and numeric gradeables should be >= ta beta testing date.

Let's add some text next to the fields on the general tab that cannot be edited once the gradeable is created (id, type, team true/false, git true/false)...  Maybe make it italics and red and say "Cannot edit id after initial creation", "Cannot edit type after initial creation", etc.

KevinMackenzie:
These changes should add all of the improvements you requested.

KevinMackenzie:
As far as the commented out lines, most of them are for features that are kind of in limbo (peer grading) or will be added in a future PR (like inheritable teams, which requires a db change).  In particular, the "this is too powerful" is there because the below line would prevent all irrelevant inputs from being submitted to the controller, but the controller is not setup well to handle missing information.

bmcutler:
numeric/text & checkpoint gradeables should have a grader assignment tab
(they don't have the "will this be graded by a TA?" question because they are always graded by the ta

bmcutler:
If I fill out rotating sections for a gradeable, hit save, they are saved.
But then if I edit a date and press save.
Then go back to the rotating sections for this gradeable, the grading assignments are blank again.

KevinMackenzie:
If an electronic gradeable has TA grading and the rubric is filled out and saved, then the gradeable is edited again and the TA grading 'yes' radio button gets pressed, the 'rubric' tab appears, but marks and other data for the rubric doesn't populate the table.  Pressing save and editing the gradeable shows that the data has not been lost, but it would be scary to see the rubric empty after reenabling this setting.

bmcutler:
checkboxes for rotating sections are missing

numeric/text gradeables is missing the "TA/Instructor Manual Grading Opens Date"

bmcutler:
also checkpoint is missing the ta/instructor manual grading opens date

bmcutler:
Should have the scary popup message about "Grading has begun edit at own peril" on the rubric panel if checkpoint or numeric gradeable and some grades have been entered.
(Make a new issue if you can't quickly solve in this PR)

KevinMackenzie:
as of 1b2dc9d checkboxes should be showing up properly.  Should both checkpoint and numeric have the 'ta/instructor manual grading open date'?

bmcutler:
yep

bmcutler:
If you're viewing one of the later tabs (e.g. dates) and you hit reload on the browser, you're always taken to the 2nd tab.  It would be nice to keep you on the current tab.  (Add to your todo list / new issue)

PR #2069, opened 2018-06-01
Alphanumeric registration section

### LINKED ISSUE #1852: Registered Section Restricted to Numeric Values (Does not support alpha characters.)
Sections at our institution are represented by alpha characters (A, B, C, etc.). It would be nice to be able to use these as section designation in submitty.

The Registered Section is part of the new student screen and on the student view. I am sure there are other places that this information is used.

tushargr:
Closes #1852 
To Be Done - Registration sections are not in proper order.


MasterOdin:
We can fix the ordering to be better by modifying the order by for section_key (registration or rotating) to be something like `"SUBSTRING(u.{$section_key}, '^[0-9]*'), COALESCE(SUBSTRING(u.{$section_key}, '^[0-9]+')::INT, -1)";`

We'll want to create a new function that does this for us (so we can reuse this sort in a bunch of places easily).

Local testing of the query
```
postgres=# SELECT * FROM test;
 test
------
 1
 10
 a
 2
 b
 a1
 a9
 a20
 b1
 b2
(10 rows)

postgres=# SELECT * FROM test ORDER BY SUBSTRING(test, '^[^0-9]*'), COALESCE(SUBSTRING(test, '[0-9]+')::INT, -1);
 test
------
 1
 2
 10
 a
 a1
 a9
 a20
 b
 b1
 b2
(10 rows)
```

tushargr:
@bmcutler 
Corrected the ordering but I noticed that user form (when we edit student or create new student) is not displaying registration section or rotating section. I guess the problem is with twig template for user popup form

bmcutler:
@CouleeApps Can you please debug the issue Tushar raises above?

MasterOdin:
Add str conversions to the following lines:
https://github.com/Submitty/Submitty/blob/master/.setup/bin/setup_sample_courses.py#L650
https://github.com/Submitty/Submitty/blob/master/.setup/bin/setup_sample_courses.py#L682
https://github.com/Submitty/Submitty/blob/master/.setup/bin/setup_sample_courses.py#L699
https://github.com/Submitty/Submitty/blob/master/.setup/bin/setup_sample_courses.py#L793

CouleeApps:
Would be nice if update-database.py edited the tables for this

tushargr:
I guess it will work now, as the section list on user form is corrected.

tushargr:
str conversion is done in setup_sample_courses.py

bmcutler:
@MasterOdin I would like to get your migration system in place before merging this PR, since this might be the first change that we need to undo/roll back to revert the system to use older code.

tushargr:
@bmcutler 
Improved the sorting of registration section (number+letters type)

And the error you are getting when shifting the student to newly created registration section,
not able to replicate it. On my system it is working.

MasterOdin:
We probably want to add to the docs that we support proper sorting of the following section regex strings:
[a-zA-Z]*
[0-9]*
[a-zA-Z]*[0-9]*
[0-9]*[a-zA-Z]*
[a-zA-Z]*[0-9]*[a-zA-Z]*

For any further extendability, we can pull the sorting stuff into its own postgres function and then we can point people to that to modify to fit their bespoke section naming system.

bmcutler:
I was jumping around branches a bunch yesterday when testing, but I believe I started with a 'partial rest' from the current master branch, then checked out this branch, ran update_database.py (the error could be in my patch to update existing courses), then installed this branch and immediately tested.   Please try that.  If it still doesn't work then I will try again.

bmcutler:
I meant partial RESET
http://submitty.org/developer/development_instructions

MasterOdin:
@bmcutler, the queries look right when going through os.system, but when moving to migrator, we'll be executing more directly against the DB so we'd actually want to disable the triggers on a table when changing the foreign key like this, and then once done, reenable the foreign key constraint. Easier than deleting it and re-adding it.

Template for this:
```
ALTER TABLE grading_registration DISABLE TRIGGER ALL;
ALTER TABLE users DISABLE TRIGGER ALL;
ALTER TABLE gradeable_teams DISABLE TRIGGER ALL;
-- alter all tables
ALTER TABLE grading_registration ENABLE TRIGGER ALL;
ALTER TABLE users ENABLE TRIGGER ALL;
ALTER TABLE gradeable_teams ENABLE TRIGGER ALL;
```

bmcutler:
I forgot to add update lines for the master database.  That needs to happen too!  That was perhaps my issue.  @tushargr can you do that?

tushargr:
ok will do that. (master database is also changed for this pr)

tushargr:
@bmcutler 
Modified update_database.py  to include master database changes. Is it fine?

bmcutler:
@MasterOdin Is going to convert the update_database commands to the new migration format.  Then we'll merge the PR.

migration/migrations/course/20180611125006_alphanumeric_sections.py
```diff
@@ -0,0 +1,40 @@
...
+    with conn.cursor() as cursor:
+        # drop foreign key contraints while we typecast the columns
+        cursor.execute("ALTER TABLE ONLY gradeable_teams DROP CONSTRAINT gradeable_teams_registration_section_fkey")
+        cursor.execute("ALTER TABLE ONLY gradeable_teams DROP CONSTRAINT gradeable_teams_rotating_section_fkey")
+
+        # disable foreign key contraints while we typecast the columns
+        cursor.execute("ALTER TABLE ONLY grading_registration DROP CONSTRAINT grading_registration_sections_registration_id_fkey")
+        cursor.execute("ALTER TABLE ONLY users DROP CONSTRAINT users_registration_section_fkey")
+
+        # typecast from integer to character varying
```
bmcutler: @MasterOdin 

Don't you mean from character to integer?

Also...  What will happen if the section isn't an integer?  How will it fail?   We had discussed simply deleting that section and moving all references to that to be null, but that's actually messy & complicated & unnecessary.  Let's just document it really clearly in the PR message.

bmcutler:
@MasterOdin 

Question about this line at the end of update_database.py:

os.system('python3 {}/migration/migrator.py migrate --fake'.format(settings['submitty_repository']))

If we're upgrading a system from 2 months before the migrator to a system 2 months after the migrator, we should:

1) grab the new code

2) run update_database.py to take us from 2 months before to the introduction of the migrator

3) run INSTALL_SUBMITTY.sh to push the database through changes in the 2 months between the migrator introduction to the present and install the new code.

Does this line at the bottom of the update_database actually break this?  Do we miss out on the db changes post introduction of the migrator?

migration/migrations/course/20180611125006_alphanumeric_sections.py
MasterOdin: yeah, I was just copy-pasting the sections from the up method to the down method and then modifying the contents. Whoops.

If the sections are not integers and you try to run down, this will throw an error giving something like `ERROR:  invalid input syntax for integer: "a"`. I'd agree with just documenting that this migration is non-reversible if you change any section to a non-numeric. However, it should be that most people never need to run the down method so I wouldn't spend any time worrying about this case beyond just documenting it.

MasterOdin:
Yes. 

I guess what we want to do is add either a flag `--single` for migrations such that if active we will only run a single migration per environment and then when combined with`--fake` we can just choose to fake as many single migrations as necessary for when we've dropped usage of upgrade_database.py and the migrator (which should just be one single usage of `python3 migrator.py migrate --fake --single`).

bmcutler:
I just think we don't want a call to migrator at the end of update database?
If my system is really old, I have to manually call update_database to get me to the migrator, then run the migrator normally to let it get me to current.

bmcutler:
Once we are past the migrator, then update_database is deprecated and shouldn't be used.

MasterOdin:
The problem here is that update_database gets us past the initial migration that is done by migrator (which we use to bootstrap any new systems). We need to mark that initial migration as skipped somewhere (and there's no way to do that right now). That can be done by a call outside of update_database or by one inside of it, I don't really care, and I'm fine removing it and giving more explicit instructions on just "You have to run update_database, then `migrator.py --fake --single` and then from then on you can call `INSTALL_SUBMITTY.sh` as normal. However, you must call update_database and migrator before you call INSTALL_SUBMITTY or else the installation process will error out midway through.

bmcutler:
ok I understand.  But instead of '--fake --single', can it be '--fake --first' or '--fake --initial' so that it could be re-run without harm?  But yes, we can solve it in another PR.  I'm just waiting for this to pass travis before merging.

PR #1224, opened 2017-07-24
Refactor icons

AndrewMRPI:
I switched the glypicons in the TA grading site and in the create/edit gradeable. I also added a new css class in ta-grading.css. Let me know in anyone finds any place that needs icons. Also should the image in 
site/public/css/images/ui-icons_222222_256x240.png be deleted? I deleted it in this pull request.

PR #2073, opened 2018-06-01
Removed 'down_arrow.png' from styling

KevinMackenzie:
This image does not exist, so it shouldn't be in the style sheet.  Since it doesn't exist, it causes an error on every page that uses server.css (all of the pages).  Now its gone...

I don't think there are any adverse effects of this, but someone else should try this out briefly before merging.

### DIFF
--- a/site/public/css/server.css
+++ b/site/public/css/server.css
@@ -95,7 +95,7 @@ select {
     font-size: 16px;
     border: 1px solid #ccc;
     height: 34px;
-    background: #FFF url('../img/down_arrow.png') no-repeat right 10px center;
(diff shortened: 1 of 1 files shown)


MasterOdin:
So I remembered why this was a thing. On certain older versions of browsers, when we changed the style of the select HTML element, it causes the arrows to disappear to indicate it was a select, so we had to add a custom arrow to the select. It looks like the icon was deleted as part of #1224, but it seems like all browsers now should display alright. We may want to revisit this in the future if we want to have a consistent dropdown design across browsers, but for now, it's fine.

PR #2083, opened 2018-06-04
Forum allow multiple categories for a thread

scopeInfinity:
~Additional Library~
- ~Chosen (JS)~
  - ~https://github.com/harvesthq/chosen~


scopeInfinity:
UI for Selecting Categories on `Create Thread` Page
Screenshot

Category filter on `View Thread` page is expected to filter those thread who have provided category among its categories.

site/app/controllers/forum/ForumController.php
```diff
@@ -147,20 +176,23 @@ public function publishThread(){
...
+        foreach ($_POST["cat"] as $category_id) {
+            $categories_ids[] = (int)$category_id;
+        }
         if(empty($title) || empty($thread_content)){
             $this->core->addErrorMessage("One of the fields was empty or bad. Please re-submit your thread.");
             $this->core->redirect($this->core->buildUrl(array('component' => 'forum', 'page' => 'create_thread')));
-        }else if(!$this->isValidCategory($category_id)){
-            $this->core->addErrorMessage("You must select a valid category. Please re-submit your thread.");
+        }else if(!$this->isValidCategories($categories_ids)){
+            $this->core->addErrorMessage("You must select valid categories. Please re-submit your thread.");
```
andrewaikens87: If the user has JavaScript disabled it defaults to all categories for the thread. Seems it doesn't trigger this error message.

andrewaikens87:
The original design had the category drop down selection be able to filter on multiple categories (so similar to the way you have them selecting in the thread view you could add that into a dropdown). So for example a student could filter on both 'Question' and 'hw1' and see posts that only have both.

andrewaikens87:
When creating a new category it pops up saying that it has been created but the page has to be refreshed to use it with the new toggle buttons. Prior to the toggle buttons the dropdown list would be auto-populated when the new category was created. Can we have the new category auto-populate to a toggle button? We can discuss this more on Slack if it is unclear.

site/app/controllers/forum/ForumController.php
scopeInfinity: Thankyou for pointing out.
Fixed it as discussed on slack.

KevinMackenzie:
When adding a category, the green banner that gets displayed shows the url-encoded string, but it would probably be better to display the unencoded string in quotes instead.
image

The other changes I would request you seem to have already fixed based on the demonstration you gave in the meeting yesterday (and have already been pointed out by @andrewaikens87 )

scopeInfinity:
Thanks @KevinMackenzie
I have pushed the changes for displaying category name in notifications in a better way.
However, the chances related to new category addition/deletion is made on #2135 PR(currently). So, I have pushed them there. I hope thats fine.

KevinMackenzie:
yea that's fine, it will be much less of a nightmare to merge that way.

PR #2085, opened 2018-06-04
Delete deprecated HWReport

### LINKED ISSUE #1987: Delete Homework Reports code
Deprecated now that students see TA grading details directly from database and grade summaries contains more details.   Before deleting confirm that everything has been suitably reproduced in the individual grade summaries report (components, marks, autograding, tagrading, extra credit, hidden, ta overall comments, etc.)

removed from UI with PR #1986

saxocellphone:
Closes #1987. HWReport is only used in ElectornicGraderController and ReportController. I tested both of them out and nothing seems to break.


CouleeApps:
How does this interact with `Gradeable::$grade_file`? If you're deleting the code that generates those files you should clean up the code that tries to load them.

saxocellphone:
@MasterOdin  In the file $grade_file is also used for queueing submissions. I talked to Barb about it and she said those should be named grading_queue_files, and I deleted the portion related to HWReports. I also replaced hasGradeFile with beenTAGraded since they are the same in functionality.

PR #2093, opened 2018-06-05
Share additional repos between host & vagrant

bmcutler:
The developer must have the Submitty/Submitty repository checked out.
We assume this path is  <SOMETHING>/GIT_SUBMITTY
This folder is shared.

If the developer also has Submitty/Lichen (named <SOMETHING>/GIT_LICHEN)
If the developer also has Submitty/AnalysisTools (named <SOMETHING>/GIT_ANALYSIS_TOOLS)
If the developer also has Submitty/Tutorial (named <SOMETHING>/GIT_TUTORIAL)

These additional folders will also be shared.

We can discuss the naming conventions.

### DIFF
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -66,7 +66,7 @@ SUBMITTY_DATA_DIR = args.data_dir
 if not os.path.isdir(SUBMITTY_DATA_DIR) or not os.access(SUBMITTY_DATA_DIR, os.R_OK | os.W_OK):
     raise SystemExit('Data directory {} does not exist or is not accessible'.format(SUBMITTY_DATA_DIR))
 
-SUBMITTY_TUTORIAL_DIR = os.path.join(SUBMITTY_INSTALL_DIR, 'GIT_CHECKOUT_Tutorial')
+SUBMITTY_TUTORIAL_DIR = os.path.join(SUBMITTY_INSTALL_DIR, 'GIT_CHECKOUT','Tutorial')
 
 TAGRADING_LOG_PATH = os.path.join(SUBMITTY_DATA_DIR, 'logs')
 AUTOGRADING_LOG_PATH = os.path.join(SUBMITTY_DATA_DIR, 'logs', 'autograding')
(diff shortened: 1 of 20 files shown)


bmcutler:
Proposal:

We update the docs to suggest VM developers create this file structure on their host machine:
ANYTHING/GIT_CHECKOUT/Submitty  (required)
ANYTHING/GIT_CHECKOUT/Tutorial  (optional -- only if you're developing this repo)
ANYTHING/GIT_CHECKOUT/Lichen  (optional -- only if you're developing this repo)
ANYTHING/GIT_CHECKOUT/AnalysisTools   (optional -- only if you're developing this repo)
etc.

Then they enter into the ANYTHING/GIT_CHECKOUT/Submitty dir and run 'vagrant up'.  And  we share all folders above that exist with these paths
/usr/local/submitty/GIT_CHECKOUT/Submitty 
/usr/local/submitty/GIT_CHECKOUT/Tutorial 
/usr/local/submitty/GIT_CHECKOUT/Lichen
/usr/local/submitty/GIT_CHECKOUT/AnalysisTools
/usr/local/submitty/GIT_CHECKOUT/nlohmann/json   (external C++ json library!)
etc.
Or if they don't exist, we checkout master or specific tagged commit from github.

On repeats of install_system.sh we check commits/tags for all repos and make sure they are sufficiently up-to-date (at or having the specified tag/commit in the linear history) or auto-update to the minimum and crash fail install if merge conflict.

On a live machine, the sysadmin must create 
SYSADMIN_CHOICE/GIT_CHECKOUT/Submitty 
where SYSADMIN_CHOICE is suggested to be /usr/local/submitty, but that ends up being the install-dir config variable
And can optionally manually checkout the other repos
SYSADMIN_CHOICE/GIT_CHECKOUT/Tutorial
etc.
And on install we checkout all repos that don't yet exist

When installing on a primary that has one or more workers, we rsync all dirs/files in the
install_dir/GIT_CHECKOUT/ subtree.  (currently only doing the submitty repo)

TODO: Need to decide how/where the minimum version/tag/commit hash for each repo will be stored.

Comments?

MasterOdin:
So some very brief thoughts, will give more later.

1. What's the use-case of checking out nlohmann/json to the local machine/vagrant? Given it's an external resource and we're presumably not actively developing for it, I'd say we treat it like the AnalysisTools in that we should just download the header file from its https://github.com/nlohmann/json/releases to a known place and then copy/symlink it into place for wherever in the system needs it.

2. We can probably simplify the "host" setup in just saying:
If we detect the following directories:
`../Tutorial`, `../Lichen`, etc. relative to the Submitty repo, they will be shared. I personally like my structure of `~/git/Submitty` to hold projects.

3. I like the idea of condensing all checkout folders into one parent. I would say that "acceptable" versions of each repo that works together should probably just be in this repo.

4. Why does the worker machine require all repos being rsynced across?

PR #2097, opened 2018-06-05
Delete latedays model & added latedays table view

saxocellphone:
Now all late days related functions will use the new method.


saxocellphone:
I did some testing, everything seems to work fine besides the GenerateGradereport. However, the same error occurs on the master branch, so I don't think I had anything to do with it.

saxocellphone:
Also for the SubmissionControllerTester I deleted the latedays related lines but now it's breaking. I'm not quite sure how to approach the problem

jchulton:
Minor errors: Lets say a student has 3 late days and goes to submit 1 day late. Prior to submission, the prompt says that they will be charged one late day and have three remaining. This is incorrect, they would have 2. If the student submits, then goes to resubmit on the same assignment the prompt says they would have 2 late days remaining. This is technically correct, but the # of late days shouldn't change for resubmitting on an assignment which the student already used a late day on.

jchulton:
It looks like the tests are failing on line 77 on Submitty\site/app/controllers/student/SubmissionController.php
It looks like there is a case where gradeable_id is null

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -807,19 +806,23 @@ private function makeTable($user_id, $gradeable, &$status){
             <tbody>
 HTML;
         $total_late_used = 0;
+<<<<<<< HEAD
```
CouleeApps: Nice merge conflict

CouleeApps:
You're failing a test: https://travis-ci.org/Submitty/Submitty/jobs/390876092

jchulton:
Looks good except for one minor banner error: if a student has no late days and goes to submit, if the assignment allows for late days, the banner will read the following, despite the student not having enough late days.
lateday
I don't think this bug should hold up the pull request, since the purpose of this pull request was to remodel how late days are handled, and late days are being calculated properly on the instructor side. I opened issue #2153 to fix the banner.

Also, is Travis now in a working state?

saxocellphone:
Oops forgot to take into account those people who are late instead of good.

bmcutler:
@saxocellphone please resolve these conflicts

CouleeApps:
Considering it's a view I changed in an earlier PR, I should resolve these :)

site/app/controllers/student/LateDaysTableController.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace app\controllers\student;
+
+
+use app\controllers\AbstractController;
+use app\libraries\Core;
+
+class LateDaysTableController extends AbstractController {
+    public function run() {
+        $g_id = isset($_REQUEST["g_id"]) ? $_REQUEST["g_id"] : NULL;
```
MasterOdin: $g_id = $_REQUEST['g_id'] ?? null;

PR #2105, opened 2018-06-06
Assignments can be now rebuild from interface

### LINKED ISSUE #2050: Build (or Rebuild) Assignment Button
It would be nice when working with an assignment if you could rebuild it from the web interface. Simply have a button that ran the build script on the server.

tushargr:
Closes #2050 
@bmcutler


MasterOdin:
Does this queue still work? I thought at one point it was stated that it had been broken and never fixed?

tushargr:
The build logs are stored on server in "build_script_output.txt" in course directory.
When I hit rebuild assignments, I checked  "build_script_output.txt" and it is building all assignments.

tushargr:
@bmcutler 
Made the changes. Now the Rebuild button only for electronic gradeable. And as added new column in table on navigation page, so adjusted e2e test accordingly.

PR #2123, opened 2018-06-07
TA Grading hotkeys window

### LINKED ISSUE #2089: TA Grading hotkeys remap window
It doubles as documentation!

CouleeApps:
Serves two purposes: it tells the user what the hotkeys are, and lets them change the keys if they so desire.

Closes #2089

Pictures:

After modifications:

While remapping:


CouleeApps:
Pressing escape resets keys. I guess I should make that more obvious or include a reset thing.

CouleeApps:
Now with more unset button:

Issue #2094, opened 2018-06-05
Forum: color coded & instructor ordered tags


bmcutler:
It might be nice to give the instructor explicit control over the ordering of the categories (not chrono or alphabetic), and to color code the labels for the categories.

for example:

I could imagine wanting one color for all hws:  (hw1, hw2, hw3, final_project)
another color for tests:  q1, q2, exam1, q3, final_exam
another color for questions: question

And wanting to specify they are ordered:  question, hw1, hw2, hw3, final_project, q1, q2, exam1, q3, final_exam

The ordering & coloring would be used in the thread display sidebar, the pull-down menu for filtering, and in choosing the category for a new thread, or editing the categories of an existing thread.

Let's discuss

scopeInfinity:
TODO
- [ ] Fix order of categories in threads list on view thread page

PR #2126, opened 2018-06-08
Forum Edit Categories (Reordering)

scopeInfinity:
Issue: #2094 (Forum: color coded & instructor ordered tags)

To be merged after PR Forum allow multiple categories for a thread  #2083, need to change Base Branch of PR before merging.

- Categories Reordering
  - Used Sortable JS for UI
  - After successful reordering, refresh categories order on same page without making additional request.


.setup/update_database.py
```diff
@@ -122,6 +122,9 @@
         os.system("PGPASSWORD='{}' psql --host={} --username={} --dbname={} -c 'ALTER TABLE ONLY threads ADD COLUMN merged_thread_id int DEFAULT '-1''".format(*variables))
         os.system("PGPASSWORD='{}' psql --host={} --username={} --dbname={} -c 'ALTER TABLE ONLY threads ADD COLUMN merged_post_id int DEFAULT '-1''".format(*variables))
         
+        # categories ordering for discussion forum
+        os.system("PGPASSWORD='{}' psql --host={} --username={} --dbname={} -c 'ALTER TABLE ONLY categories_list ADD COLUMN rank int'".format(*variables))
+
```
andrewaikens87: Any database change also has to be added into `course_tables.sql`.

site/app/views/forum/ForumThreadView.php
```diff
@@ -847,7 +848,54 @@ public function createThread() {
...
+			$categories = $this->core->getQueries()->getCategories();
+			$return .= <<<HTML
+			<div class="popup-form" id="category-list">
+				<h3>Categories</h3>
+				<pre>(Drag to re-order)</pre><br>
+HTML;
+				if(count($categories) == 0) {
+					$return .= <<<HTML
+					<span id='category-list-no-element' style="margin-left: 1em;" >
+						No category found
```
andrewaikens87: If there is no category found there still displays a draggable bar. Can we please remove that? Also if there exists no categories we should instruct the user to create one. So possibly something like "No categories exist please create one." or something similar. 


draggablebar

site/public/js/server.js
```diff
@@ -1270,6 +1270,101 @@ function addNewCategory(){
...
+
+function reorderCategories(){
+    var data = $('#ui-category-list').sortable('serialize');
+    var url = buildUrl({'component': 'forum', 'page': 'reorder_categories'});
+    $.ajax({
+            url: url,
+            type: "POST",
+            data: data,
+            success: function(data){
+                console.log(data);
```
andrewaikens87: Please remove console log

bmcutler:
There should always be at least one category.  We start with "question" & "comment" (they can be renamed/removed if desired, as long as there is always at least one category).

Only instructor/fullaccess/limitedaccess can create or delete categories.

Only unused categories can be deleted.

You cannot remove the last category.  (a category is required to make a post, and a student cannot make a category)

andrewaikens87:
Yes, there should always be at least one category. The error message should only show up if someone manually deleted the rows in the `categories_list` table. This PR currently has changes for altering the ordering of the categories. Renaming/deleting categories hasn't been implemented yet.

site/app/views/forum/ForumThreadView.php
scopeInfinity: Thanks! Bug fixed.

PR #2127, opened 2018-06-08
Mousefree checkpoint entry

### LINKED ISSUE #767: [UX] Mouse-free checkpoint entry
Hi

Currently the check-point entry workflow is something like this: ctrl-f, type part of students name, move hand to mouse, move cursor to button, click, return hand to keyboard.

If it were possible to streamline this workflow to remove mouse use, it would be much easier to enter checkpoints for large classes. For example ctrl-f, type part of students name, press enter to cycle through multiple matches (this is the default in Chrome), press tab to cycle through checkpoint values for currently selected entry.

By eliminating the need to move the hand to the mouse, and the need to aim the cursor, I think total entry time could be improved to 25-50%. So from 20minutes to enter 200 grades to 10-15minutes. Cumulatively this would save hours per course per semester.

If needed, I would be happy to work on this with some guidance.

ropern123:
Added mousefree checkpoint entry, fixed StatisticsForm.twig to display all statistics instead of just first component, and fixed setupNumericTextCells() and setupCheckboxCells() being called on loadup of every page, not just their intended pages.

closes #767


emaicus:
Just found a small bug. 

If you are using a numeric gradable and you switch from using the arrow keys to using tab, your arrow key context stays put rather than following the cursor.

ropern123:
The most recent version also fixes the bug you mentioned.
Also, as we discussed at the meeting:

- [ ] Add hotkeys menu
- [ ] Highlight cells for numeric

jn7070:
Looks good now.

PR #2129, opened 2018-06-08
Change pdf_check to PyPDF2

### LINKED ISSUE #1958: Replace pdftk with PyPDF2
We're currently only using pdftk in [pdf_check.cgi](https://github.com/Submitty/Submitty/blob/master/site/cgi-bin/pdf_check.cgi) to split apart a PDF into specified chunks. We can just natively use the [PyPDF2](https://pythonhosted.org/PyPDF2/index.html) library to do thus for us using just straight python (instead of messy subprocess calls).

We can implement any necessary capability that pdftk was providing us via this library (just without the CLI interface, but could make one as necessary).

saxocellphone:
Closes #1958.


.setup/Dockerfile
```diff
@@ -7,7 +7,7 @@ hardening-includes p7zip-full patchutils \
 libpq-dev unzip valgrind zip libmagic-ocaml-dev common-lisp-controller libboost-all-dev \
 javascript-common  \
 libfile-mmagic-perl libgnupg-interface-perl libbsd-resource-perl libarchive-zip-perl gcc g++ \
-g++-multilib jq libseccomp-dev libseccomp2 seccomp junit flex bison spim poppler-utils pdftk
+g++-multilib jq libseccomp-dev libseccomp2 seccomp junit flex bison spim poppler-utils PyPDF2
```
MasterOdin: We don't need this in Docker as it's not necessary for grading.

.setup/distro_setup/debian/setup_distro.sh
```diff
@@ -128,7 +128,7 @@ hardening-includes p7zip-full patchutils \
 libpq-dev unzip valgrind zip libmagic-ocaml-dev common-lisp-controller libboost-all-dev \
 javascript-common  \
 libfile-mmagic-perl libgnupg-interface-perl libbsd-resource-perl libarchive-zip-perl gcc g++ \
-g++-multilib jq flex bison spim poppler-utils pdftk
+g++-multilib jq flex bison spim poppler-utils PyPDF2
```
MasterOdin: Move this to `.setup/install_system.sh` and install it via pip3. It's easier for us to keep up-to-date on python packages using pip than apt.

site/cgi-bin/pdf_check.cgi
```diff
@@ -87,23 +86,29 @@ try:
...
+        counter = 0
+        for i in range(0, total_pages, num):
+            cover_writer = PdfFileWriter()
+            cover_writer.addPage(pdfReader.getPage(counter)) 
+            cover_filename = '{}_{}_cover.pdf'.format(filename[:-4], int(i/2))
+            pdf_writer = PdfFileWriter()
+            start = counter
+            for j in range(start, start+num):
+                pdf_writer.addPage(pdfReader.getPage(j)) 
+                output_filename = '{}_{}.pdf'.format(filename[:-4], int(i/2))
```
MasterOdin: Move this string naming outside the for loop, it doesn't change through the loop.

PR #2143, opened 2018-06-12
Fix ordering having system before master for migrations

MasterOdin:
Fix for build break in 9ba562103d7f7df4c09b7eef308d46d0ecfaaed6 where I messed up the ordering of the migrator environments. This is what I get for not waiting for Travis to finish for a "simple fix" that I tested against an already stood-up environment (bug was that --initial for system relies on master already being run for the migrations_system table to be created for us to record the system migrations in).

### DIFF
--- a/migration/migrator.py
+++ b/migration/migrator.py
@@ -45,9 +45,13 @@ def parse_args():
     args = parser.parse_args()
     if args.environments is None:
         args.environments = ENVIRONMENTS
-    # make sure the order is of 'system', 'master', 'course' (which is reverse alphabetical)
-    args.environments.sort()
-    args.environments.reverse()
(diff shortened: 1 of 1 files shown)


migration/migrator.py
```diff
@@ -45,9 +45,13 @@ def parse_args():
...
-    # make sure the order is of 'system', 'master', 'course' (which is reverse alphabetical)
-    args.environments.sort()
-    args.environments.reverse()
+    # make sure the order is of 'master', 'system', 'course' depending on what environments have been selected
+    # system must be run after master initially as system relies on master DB being setup for migration table
+    environments = []
+    for env in ['master', 'system', 'course']:
+        if env in args.environments:
+            environments.append(env)
+    args.envrionments = environments
```
bmcutler: mispelling

PR #2150, opened 2018-06-13
Forum Edit Thread

scopeInfinity:
Issue: #1950

Task(s)
- [x] Shared template for edit post/thread and create new thread
- [x] Edit thread title in the discussion forum(Edit option of first post)
- [x] Edit Anon status of post
- [x] Edit Categories of Thread


andrewaikens87:
I agree with @ropern123, we should be able to edit the thread and the post at the same time. On the note about limiting duplicate code @scopeInfinity we can discuss on slack.

scopeInfinity:
@ropern123 @andrewaikens87 Thanks for the review.
Now we are using common twig template while creating thread and editing post/thead.
And it submits thread and post updation as single request to backend.

PR #2151, opened 2018-06-13
Forum show deleted threads

scopeInfinity:


MasterOdin:
So on my understanding of this, we want to make it so that if a user decides to view deleted threads, if we goes away from the page, and goes back, we again want to show the user the deleted threads?

It would probably make sense to use cookies instead of $_SESSION to maintain this so that it survives closing/opening the browser.

scopeInfinity:
@andrewaikens87 @ropern123 Yah, It got messed up during one of merge, I might have missed it.
It's working now.

PR #2157, opened 2018-06-13
Teamless people can no longer submit & Optimize # versions

### LINKED ISSUE #2110: Sample data: teamless people should not have submissions
Users not on a team are not allowed to submit from the frontend, so they shouldn't have submissions in the sample data.

saxocellphone:
Closes #2110


saxocellphone:
Now there are only around 1700 autograding assignments...

jchulton:
looks good

PR #2161, opened 2018-06-13
Twig autograding / results view

### LINKED ISSUE #2145: Grade badge color consistency
On the results page, badges have different coloring schemes. For consistency's sake, we should only use one of these.

Autograding:
* Green: 100%
* Yellow: 1-99%
* Red: 0%

Manual / TA grading:
* Green: 100%
* Yellow: 50-100%
* Red: 0-50%

CouleeApps:
This includes:

- Autograding results on HomeworkView
- Autograding results on ElectronicGraderView
- TA grading results on HomeworkView

Closes #2145


jchulton:
Looks good. One thing to note is that I should probably change the color scheme of the Manual TA grading rubric to match this color scheme

Issue #2107, opened 2018-06-06
Gradeable model rework


CouleeApps:
Currently the `Gradeable` class contains both assignment details and (sometimes) student submission details. This creates significant data duplication and makes the concept of a "Gradeable" confusing. We should split the class into two parts:
1. The static assignment details, configuration, grading criteria, etc
2. A specific user's submission and results

This new Submission class would hold a reference to the static Gradeable, and the Gradeable would be only general configuration data with no reference to student data.

Here's a quick diagram of the potential structure:

![submissionmodel](https://user-images.githubusercontent.com/711973/41308229-58daa684-6e49-11e8-8819-098231b18b59.png)

CouleeApps:
As discussed in meeting today @KevinMackenzie @MasterOdin @bmcutler

KevinMackenzie:
One particular motivation for these changes is to separate ajax calls for TA grading into those editing a student's assigned marks and those editing the mark itself (order, text, point value).

MasterOdin:
Why do you need to split apart the gradeable to have separate Ajax calls?

MasterOdin:
My suggestion for this would be to make a Gradeable that has the configuration details and then a GradedGradeable that inherits from Gradeable + has graded data. A GradeableComponent and a GradedGradeableComponent that inherits GradeableComponent.

CouleeApps:
Not a bad suggestion, it would make porting considerably easier. One concern is that you still get duplicated data. The upside of splitting it completely is that you only need to get the data for the Gradeable once and can edit it in a single place without worrying about copies.

CouleeApps:
![submissionmodel](https://user-images.githubusercontent.com/711973/41308229-58daa684-6e49-11e8-8819-098231b18b59.png)

KevinMackenzie:
'One thing we need to be careful about is that the previous GradeableComponentMark class used urlencoded strings for its `$notes`. We shouldn't be using that in the future, but it will make backwards compatibility a little confusing.'
-@CouleeApps

KevinMackenzie:
I think I got a good start on the green classes (See branch gradeable-model-refactor-1)

KevinMackenzie:
Why are the mark 'note's url encoded?

MasterOdin:
Probably to avoid XSS bugs within the view layer at the source.  Within this redesign, it should be that within the model is the "pure" data from the DB (XSS exploits and all) and then within the twig display layer is where stuff like escaping is applied.

[...]

PR #2163, opened 2018-06-13
Added new static gradeable model classes and DB queries

KevinMackenzie:
The new blue classes (see #2107) with the required database methods to load them.  These are completely separate classes and are not used anywhere, so to test that they loads properly from the DB, you can insert the following lines: 

```
$test_gradeable = $this->core->getQueries()->getGradeableConfig($admin_gradeable->getGId());
$this->core->getOutput()->renderJson($test_gradeable->toArray());
return;
```

into the `AdminGradeableView` to short circuit it and dump the json representation to the client.

The names I chose are supposed to be more concise than the DB names, but just as descriptive.  Some are slightly reworded so they play better with the `get`, `set`, and `is` prefixes for accessor methods wrangled by the `AbstractModel`.


KevinMackenzie:
Since these classes will be used in a lot of places, we should be picky about style/naming convention now.

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,446 @@
...
+                $errors[$date] = $result;
+            }
+        }
+
+        $ta_view_start_date = $dates['ta_view_start_date'];
+        $grade_start_date = $dates['grade_start_date'];
+        $grade_released_date = $dates['grade_released_date'];
+        $team_lock_date = $dates['team_lock_date'];
+        $submission_open_date = $dates['submission_open_date'];
+        $submission_due_date = $dates['submission_due_date'];
```
CouleeApps: These need to be by reference otherwise you're modifying nothing

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,446 @@
...
+    /** @property @var bool If students can view/download any version of the submitted files, or just the active version */
+    protected $student_any_version = false;
+    /** @property @var bool If the gradeable uses peer grading */
+    protected $peer_grading = false;
+    /** @property @var int The number of peers each student will be graded by */
+    protected $peer_grade_set = 0;
+    /** @property @var \DateTime The date students can start making submissions */
+    protected $submission_open_date = null;
+    /** @property @var \DateTime The date, before which all students must make a submissions (or be marked late) */
+    protected $submission_due_date = null;
```
CouleeApps: Should all the date fields be together? What about team dates? With this many dates it might even be helpful to make a class to store them all in one place

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,446 @@
...
+    /** @property @var bool If the gradeable is using any manual grading */
+    protected $ta_grading = false;
+    /** @property @var bool If students can view submissions */
+    protected $student_view = false;
+    /** @property @var bool If students can make submissions */
+    protected $student_submit = false;
+    /** @property @var bool If students can download submitted files */
+    protected $student_download = false;
+    /** @property @var bool If students can view/download any version of the submitted files, or just the active version */
+    protected $student_any_version = false;
```
CouleeApps: Variable name (while probably matching the db) is not very intuitive without having to read the doc comment. May want to use `$student_download_any_version` to inform people this is about _downloading_

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,446 @@
...
+    /** @property @var \DateTime The date that graders may start grading */
+    protected $grade_start_date = null;
+    /** @property @var \DateTime The date that grades will be released to students */
+    protected $grade_released_date = null;
+    /** @property @var \DateTime The date after which only instructors may change grades (aka when grades are 'due') */
+    protected $grade_locked_date = null;
+    /** @property @var int The minimum user group that can grade this gradeable (1=instructor) */
+    protected $min_grading_group = 1;
+    /** @property @var string The syllabus classification of this gradeable */
+    protected $syllabus_bucket = "Homework";
```
CouleeApps: Enum for this?

site/app/models/gradeable/Mark.php
```diff
@@ -0,0 +1,56 @@
...
+ * @method setPublish($should_publish);
+ */
+class Mark extends AbstractModel
+{
+    /** @property @var int The course-wide unique numeric id of this mark */
+    protected $id = -1;
+    /** @property @var int The number of points this mark will add to the score (negative for deductions) */
+    protected $points = 0;
+    /** @property @var string The description of this mark (aka why a student would lose/gain these points) */
+    protected $note = "";
```
CouleeApps: Gradeables have `title`, Components have `title`, Mark should have `title` as well

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,446 @@
...
+    }
+    private function setSubmissionDueDate($date)
+    {
+        throw new NotImplementedException('Individual date setters are disabled, use "setDates" instead');
+    }
+
+    private function setId($id)
+    {
+        preg_match('/^[a-zA-Z0-9_-]*$/', $id, $matches, PREG_OFFSET_CAPTURE);
+        if (count($matches) === 0) {
```
CouleeApps: If `Gradeable::setId()` is private, should the other classes also have private `setId()`?

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,446 @@
...
+        $this->type = $type;
+    }
+
+    public function setMinGradingGroup($group)
+    {
+        // Disallow the 0 group (this may catch some potential bugs with instructors not being able to edit gradeables)
+        if (is_int($group) && $group > 0) {
+            $this->min_grading_group = $group;
+        } else {
+            throw new \InvalidArgumentException("Grading group must be a natural number");
```
CouleeApps: > natural number

I know some mathematicians who would argue with this... Just say must be an integer greater than zero

site/app/models/gradeable/Component.php
```diff
@@ -0,0 +1,215 @@
...
+        return max($this->upper_clamp, $this->max_value);
+    }
+
+    // Electronic
+    public function isCountUp()
+    {
+        return $this->default === 0;
+    }
+
+    public function isPenalty()
```
CouleeApps: `hasPenalty()`? Components don't have to be only penalty

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,446 @@
...
+    /** @property @var bool If submission after student's max deadline
+     *      (due date + min(late days allowed, late days remaining)) is allowed
+     */
+    protected $late_submission_allowed = true;
+    /** @property @var float The point precision for manual grading */
+    protected $precision = 0.0;
+
+    public function __construct(Core $core, $details, array $components)
+    {
+        parent::__construct($core);
```
CouleeApps: Will it be possible to construct one of these without a details array (like AdminGradeable)? May want to have an optional empty constructor to create a blank Gradeable.

site/app/models/gradeable/Component.php
```diff
@@ -0,0 +1,215 @@
...
+    protected $peer = false;
+    /** @property @var int The order of this component in the gradeable */
+    protected $order = -1;
+    /** @property @var int The pdf page this component will reside in */
+    protected $page = -1;
+
+    /** @property @var Mark[] All possible common marks that can be assigned to this component */
+    protected $marks = array();
+
+
```
CouleeApps: It may be a good idea for components to have a reference back to their Gradeable (so you don't have to pass the entire hierarchy every time). Would be interesting to see if that's doable without making toArray have infinite recursion.

site/app/models/gradeable/Component.php
```diff
@@ -0,0 +1,215 @@
...
+            $max_value = null;
+            $errors['max_value'] = "Value must be a number!";
+        } else {
+            $max_value = floatval($max_value);
+        }
+        if (!is_numeric($upper_clamp)) {
+            $upper_clamp = null;
+            $errors['upper_clamp'] = "Value must be a number!";
+        } else {
+            $upper_clamp = floatval($upper_clamp);
```
CouleeApps: May also want to round these values to the precision

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,446 @@
...
+            }
+            if ($this->ta_grading) {
+                if ($grade_start_date === null) {
+                    $errors['grade_start_date'] = "Value must not be null!";
+                }
+//                if ($grade_locked_date === null) {
+//                    $errors['grade_locked_date'] = "Value must not be null!";
+//                }
+                if (!($submission_due_date === null || $grade_start_date === null) && $submission_due_date > $grade_start_date) {
+                    $errors['g_grade_start_date'] = 'Manual Grading Open Date must be no earlier than Due Date';
```
CouleeApps: Nitpick: single or double quotes for these error messages?

site/app/models/gradeable/Gradeable.php
KevinMackenzie: If you're referring to the assignment on lines 253 through 258, these are just alias's to get named variables into the scope instead of looking up a name in the `dates` array every time to reduce clutter.  After the `foreach`loop, none of the values should be modified.

site/app/models/gradeable/Mark.php
```diff
@@ -0,0 +1,56 @@
...
+ * @method setOrder($order);
+ * @method isPublish();
+ * @method setPublish($should_publish);
+ */
+class Mark extends AbstractModel
+{
+    /** @property @var int The course-wide unique numeric id of this mark */
+    protected $id = -1;
+    /** @property @var int The number of points this mark will add to the score (negative for deductions) */
+    protected $points = 0;
```
CouleeApps: Should be type float

site/app/models/gradeable/Mark.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->order = $details['order'];
+        $this->publish = $details['publish'];
+    }
+
+    /* Overridden setters with validation */
+
+    public function setPoints($points)
+    {
+        if(is_float($points) || is_int($points)) {
+            $this->points = $points;
```
CouleeApps: is_numeric

site/app/models/gradeable/Gradeable.php
CouleeApps: I assumed you wanted to modify the dates array as well because you were modifying `$late_days` and `$grade_start_date` later in the function

KevinMackenzie: Thanks for catching that, should be fixed now

site/app/models/gradeable/Gradeable.php
KevinMackenzie: fixed

site/app/models/gradeable/Mark.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function __construct(Core $core, $details)
+    {
+        parent::__construct($core);
+
+        $this->setId($details['id']);
+        $this->points = $details['points'];
+        $this->title = $details['title'];
+        $this->order = $details['order'];
+        $this->publish = $details['publish'];
```
CouleeApps: Are you going to validate these?

site/app/models/gradeable/Component.php
KevinMackenzie: Done

site/app/models/gradeable/Mark.php
KevinMackenzie: done

site/app/models/gradeable/Mark.php
KevinMackenzie: Yes, should be validating now

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,480 @@
...
+ * @method \DateTime getTeamLockDate();
+ * @method bool isTaGrading();
+ * @method void setTaGrading($use_ta_grading);
+ * @method bool isStudentView();
+ * @method void setStudentView($can_student_view);
+ * @method bool isStudentSubmit();
+ * @method void setStudentSubmit($can_student_submit);
+ * @method bool isStudentDownload();
+ * @method void setStudentDownload($can_student_download);
+ * @method bool isStudentAnyVersion();
```
CouleeApps: This too

site/app/models/gradeable/Gradeable.php
KevinMackenzie: Single quotes used unless double quotes needed

site/app/models/gradeable/Gradeable.php
KevinMackenzie: For now we're going to let it be any string, but in the future we may want to setup a `buckets` table in the database and validate that its one of those.

CouleeApps:
One thing we need to be careful about is that the previous `GradeableComponentMark` class used urlencoded strings for its `$note`s. We shouldn't be using that in the future, but it will make backwards compatibility a little confusing.

site/app/models/gradeable/Component.php
```diff
@@ -0,0 +1,230 @@
...
+        $this->setTaComment($details['ta_comment']);
+        $this->setStudentComment($details['student_comment']);
+        $this->setPoints($details['lower_clamp'], $details['default'], $details['max_value'], $details['upper_clamp']);
+        $this->setText($details['text']);
+        $this->setPeer($details['peer']);
+        $this->setOrder($details['order']);
+        $this->setPage($details['page']);
+    }
+
+    public function getGradeable()
```
CouleeApps: Should be in the class docstring

site/app/models/gradeable/Component.php
```diff
@@ -0,0 +1,230 @@
...
+        }
+        if (!($max_value === null || $upper_clamp === null) && $max_value > $upper_clamp) {
+            $errors['max_value'] = 'Max value can\'t be more than upper clamp!';
+        }
+
+        if (count($errors) === 0)
+            return null;
+        return $errors;
+    }
+    public function setPoints($lower_clamp, $default, $max_value, $upper_clamp)
```
CouleeApps: Would be nice to have docs about what values are valid for this without having to look at the error messages

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,469 @@
...
+
+    /** @property @var string The overall ta instructions for grading (numeric-text/checkpoint only) */
+    protected $ta_instructions = "";
+
+    /* Properties exclusive to electronic gradeables */
+
+    /** @property @var string The location of the autograding configuration file */
+    protected $autograding_config_path = "";
+    /** @property @var string[] The object that contains the autograding config data */
+    private $autograding_config = null;
```
CouleeApps: Any reason this is private?

CouleeApps:
Would be nice to doc the rest of the functions that have constraints

site/app/models/gradeable/Gradeable.php
KevinMackenzie: Originally I wasn't planning on serializing it with the rest of the gradeable, but it probably makes sense to

KevinMackenzie:
I think I got everything you mentioned

site/app/models/gradeable/Component.php
```diff
@@ -0,0 +1,264 @@
...
+ * @method getMaxValue();
+ * @method getUpperClamp();
+ * @method isText();
+ * @method setText($is_text);
+ * @method isPeer();
+ * @method setPeer($is_peer);
+ * @method getOrder();
+ * @method setOrder($order);
+ * @method getPage();
+ * @method setPage($page);
```
CouleeApps: Need doc for getMarks()

CouleeApps:
Would be nice to get `@internal` on all methods that shouldn't be used / are internal so PhpStorm knows not to touch them

KevinMackenzie:
I'm being held up by some things, don't approve this until I fix it.

site/app/models/gradeable/Gradeable.php
```diff
@@ -252,21 +252,21 @@ private function validateDates(array &$dates)
...
-        $submission_open_date = $dates['submission_open_date'];
-        $submission_due_date = $dates['submission_due_date'];
-        $late_days = $dates['late_days'];
+        $dates['late_days'] = intval($dates['late_days'] ?? 0);
+
+        $ta_view_start_date = $dates['ta_view_start_date'] ?? null;
+        $grade_start_date = $dates['grade_start_date'] ?? null;
+        $grade_released_date = $dates['grade_released_date'] ?? null;
+        $team_lock_date = $dates['team_lock_date'] ?? null;
+        $submission_open_date = $dates['submission_open_date'] ?? null;
```
CouleeApps: Doing `thing_that_could_be_null ?? null` is basically a no-op. I'd say either use default values or don't null-coalesce.

site/app/libraries/DateUtils.php
```diff
@@ -77,4 +77,47 @@ public static function validateTimestamp($timestamp) {
...
+    }
+    /**
+     * Converts a provided date reference to a string if it isn't already
+     *
+     * @param $date &\DateTime | &string A reference to the date object to convert
+     */
+    public static function convertDateTimeToString(&$date) {
+        if(gettype($date) === 'string') return;
+        $date = self::dateTimeToString($date);
+    }
```
MasterOdin: Usage of this?

site/app/exceptions/AggregateException.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace app\exceptions;
+
+/**
+ * Class AggregateException
+ * @package app\exceptions
+ *
+ * When multiple separate errors must be communicated through one exception
+ */
+class AggregateException extends BaseException {
```
MasterOdin: Given how you're using this, you should probably just call this "ValidationException" which better reflects that it's being used to point out validation errors.

site/app/models/gradeable/Gradeable.php
KevinMackenzie: Its slightly different for arrays.  It prevents array out-of-bound-ness.  So, for each of those properties, the RHS is equivalent to `isset($dates['property']) ? $dates['property'] : null`.

site/app/libraries/DateUtils.php
KevinMackenzie: Used somewhere that got deleted, will remove

site/app/libraries/DateUtils.php
```diff
@@ -77,4 +77,47 @@ public static function validateTimestamp($timestamp) {
...
+
+    /**
+     * Asserts that the provided date is a \DateTime object and converts it to one
+     *  if its a string, returning any error in parsing.
+     *
+     * @param $date \DateTime|string A reference to the date object to assert.  Set to null if failed.
+     * @param $time_zone \DateTimeZone The timezone to parse with
+     * @return null|string The error message or null
+     */
+    public static function assertDate(&$date, \DateTimeZone $time_zone) {
```
MasterOdin: So couple things:
1. a function prefixed with "assert" I'd expect to either return true/false or throw an "AssertionError" (looking at the assert function in PHP or python) and that's it
2. I dislike the paradigm of having a function that takes in a variable, mutates it, and then returns null (it was successfully mutated) or a string (to indicate an error). To me, a function should either a) mutate the variables passed by reference or b) mutate variable passed by value and return the changed value (and I drastically prefer b). In either case, given this is OOP, throw an exception if we hit an invalid value for that mutation.
3. If your function definition contains an "and' to describe two discrete things that it's doing and that neither feed from the other, this should be two different functions.

site/app/models/gradeable/Component.php
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * Checks the point values to ensure consistency.  See `setPoints` docs for details
+     *
+     * @param $lower_clamp
+     * @param $default
+     * @param $max_value
+     * @param $upper_clamp
+     * @return array|null
+     */
+    private function validatePoints(&$lower_clamp, &$default, &$max_value, &$upper_clamp)
```
MasterOdin: Same as above. A function named "validate", I'd assume it would either return true/false if validation succeeded, not return either an array or null and then mutated the incoming variables. This goes back to above where the description of this function is "validates that parameters are numerics and casts them to floats". Those are two discrete statements and it's weird to see it all done within one function (and hence also why you're doing pass by value and returning things).

Given you probably want to give a status report at the end, I'd do make this function collect errors in an array, and if count($array) > 0, throw ValidationError, else do nothing. You then have roundPointValue which accepts a numeric string or float to handle the floatval for you.

site/app/models/gradeable/Gradeable.php
```diff
@@ -0,0 +1,571 @@
...
+
+        return $details;
+    }
+
+    /**
+     * Validates a set of dates, see docs for `setDates` for the details
+     * @param array $dates A reference to the dates to be checked
+     * @return array|null Any errors in validation
+     */
+    private function validateDates(array &$dates)
```
MasterOdin: same issue as the other validate function.

site/app/models/gradeable/Component.php
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        if (Utils::compareNullableGt($points['lower_clamp'], $points['default'])) {
+            $errors['lower_clamp'] = 'Lower clamp can\'t be more than default!';
+        }
+        if (Utils::compareNullableGt($points['default'], $points['max_value'])) {
+            $errors['max_value'] = 'Max value can\'t be less than default!';
+        }
+        if (Utils::compareNullableGt($points['max_value'], $points['upper_clamp'])) {
+            $errors['max_value'] = 'Max value can\'t be more than upper clamp!';
```
MasterOdin: should this be `upper_clamp` key?

MasterOdin: should this be `upper_clamp` key?

site/app/models/gradeable/Component.php
```diff
@@ -0,0 +1,275 @@
...
+        }
```
MasterOdin: `compareNullableGt` is a weird function in that it validates both values are not null and the `>` relation. I'm not sure there's a good name for such a function as well as in the case one of them is null, it'd be weird to just see error message about things being more/less when something is null.

It feels like doing this might be clearer?
```
$keys = ['lower_clamp', 'default', 'max_value', 'upper_clamp'];
for ($i = 0; $i < count($keys); $i++) {
    if (is_null($points[$keys[$i]])) {
        $errors[$keys[$i]] = str_replace(ucfirst($keys[$i]), "_", " "))." cannot be null";
    }
    elseif (isset($keys[$i-1]) && !is_null($keys[$i-1]) && $keys[$i-1] > $keys[$i]) {
        $errors[$keys[$i-1].'_bound'] = str_replace(ucfirst($keys[$i-1]), "_", " "))." cannot be more than ".str_replace($keys[$i], "_", " ");
    }
}

site/app/models/gradeable/Component.php
KevinMackenzie: It associates the error with `max_value` because `upper_clamp` and `lower_clamp` are supposed to take precedence over `max_value` and `default`

KevinMackenzie:
When this PR gets merged, don't delete the branch.

bmcutler:
The way squash & merge works we always(?) want to delete the branch, just make a new branch with a different name and continue work.

PR #2166, opened 2018-06-14
Fix update_database.py marking all migrations as being done, not just initial

### LINKED ISSUE #2147: migrator option to '--fake --initial' (or similar)
And modify the last line of update_database.py to do the right thing (mark that only the initial migration should not be run)

see discussion #2069

MasterOdin:
closes #2147

### DIFF
--- a/.setup/update_database.py
+++ b/.setup/update_database.py
@@ -147,4 +147,5 @@ for term in os.scandir(os.path.join(settings['submitty_data_dir'],"courses")):
         # add user/database
         print("\n")
 
-os.system('python3 {}/migration/migrator.py migrate --fake'.format(settings['submitty_repository']))
+# On the first run of this, this will mark the first initial migration as fake, and then leave the rest as being un-run.
+os.system('python3 {}/migration/migrator.py migrate --single --fake'.format(settings['submitty_repository']))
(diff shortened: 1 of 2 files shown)


bmcutler:
In issue #2147 I suggested '--fake --initial'

MasterOdin:
Except that `--initial --fake` would be a confusing interaction to only mark the first migration as fake run, and to not run any other migrations. 

`--initial` runs the first migration (if not up) and marks all the rest as up (without running)

`--fake` marks any queued migration as being up without actually running them.

`--fake --initial` implies that it would just mark all migrations as up without running anything (the fake overrides the initial). 

Adding a single flag (run only the first non-up migration) makes sense then as it's potentiallly useful on its own and then combines nicely with fake to fake run the first non-up migration which in the case of update_database.py is the initial migrations. However, this paradigm can be useful for developers so it's nice that it doesn't only apply to faking the initial migration.

bmcutler:
What if we're in a state where some migrations have been run and we update the repo so we have new migrations that haven't been processed.  If we run update_database.py (we don't need to, someone is just following old instructions), what happens?  Does it mark the first un-run migration as fake?  That's what --fake --single implies to me.   

It isn't obvious to me that a command with --fake should do any of the migrations.  
It would make more sense to have 2 calls:
--fake-initial migrator                          
(run normally no options) migrator

MasterOdin:
That's a fair point. I think it's probably best though to just have update_database.py manually insert the row for the initial migration instead of probably spending anymore time on flags and such for this one-off case.

bmcutler:
Yes, much better :)
and migration can be manually run by the sys admin,
or it can wait and be automatically run with the next install_submitty.

PR #2177, opened 2018-06-15
No submit no grade

### LINKED ISSUE #1541: Do not allow grades to be submitted to electronic gradeables with no submission
It appears that you cannot expand the rubric if there is no submission...  but there is a sideways way to get a grade entered.

If you're grading a specific problem, have that rubric component expanded, and click "go to next student" (and that next student has no submission), you can select marks for that rubric item and hit next and they get stored.

jchulton:
Closes #1541. Cannot attempt to grade an assignment if the student has not submitted.


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -947,10 +953,21 @@ public function saveSingleComponent() {
...
-                return;
+
+                $response = array('status' => 'failure');
+                $this->core->getOutput()->renderJson($response);
+                return $response;
             }
         }
+        /*
+        if ($gradeable->getCurrentVersion() === null) {
+            $this->core->addErrorMessage("Cannot grade an assignment with no submission.");
```
CouleeApps: Why did you comment this out?

site/app/templates/grading/electronic/RubricPanel.twig
```diff
@@ -1,54 +1,59 @@
+<script type="application/javascript">
+    var grading_data=null;
+</script>
```
CouleeApps: This should go somewhere else, like RubricPanel.twig

CouleeApps: ** should go further down, in the if-statement on line 40

site/app/controllers/grading/ElectronicGraderController.php
jchulton: This warning would appear on every page, and because the grading rubric now says no submission and does not allow for grading, I didn't think it was needed

jchulton:
Ready for review

PR #2189, opened 2018-06-18
Pdf page settings save properly and behavior more clear

### LINKED ISSUE #2186: Create/Edit Gradeable - pdf pages per component does not save
Related to (but not solved by) edits in PR #2185

KevinMackenzie:
Fixes #2186

I moved the pdf page setting to be within the region that only gets saved when the `save` button is pressed.  It should be clear that it will not be applied unless you press 'save'.

Since the pdf page setting is inferred from the pdf page numbers of the components, changing the setting is meaningless unless you also change the pdf pages of the components.  It was only submitting the changes to the `eg_pdf_page`/`eg_pdf_page_student` setting, thus it wasn't doing anything.  I made a mistake in renaming `pdf_page` to `eg_pdf_page` since the latter name implies the property is from the `electronic_gradeable` table (which it isn't), but I won't update that name until porting to the new gradeable model.


jchulton:
Looks good. The save button now allows for pdf page settings to be enabled.

PR #2190, opened 2018-06-18
Added new per-user gradeable model classes and DB queries

KevinMackenzie:
The new green model classes (see #2107) and database methods to load them.  They are, once again, completely separate classes and not used anywhere else, so to test that they load propertly, you can insert the following lines:

```
$test_gradeable = $this->core->getQueries()->getGradeableConfig($admin_gradeable->getGId());
$gradeable_data = $this->core->getQueries()->getGradeableData($test_gradeable, 'aphacker', null);
$this->core->getOutput()->renderJson($gradeable_data->toArray());
return;
```

into the `AdminGradeableView` to short circuit it and dump the json representation to the client.

### DIFF
--- a/site/app/models/gradeable/GradedGradeable.php
+++ b/site/app/models/gradeable/GradedGradeable.php
@@ -1,0 +1,162 @@
+<?php
+
+namespace app\models\gradeable;
+
+use app\libraries\Core;
+use app\libraries\DateUtils;
+use \app\models\AbstractModel;
+
+/**
+ * Class GradedGradeable
+ * @package app\models\gradeable
+ *
+ * @method string getGradeableId()
+ * @method GradedComponent[] getGradedComponents()
+ * @method Submitter getSubmitter()
+ * @method int getId()
+ * @method string getOverallComment()
+ * @method void setOverallComment($comment)
+ * @method \DateTime getUserViewedDate()
+ */
+class GradedGradeable extends AbstractModel {
+    /** @var Gradeable Reference to gradeable */
+    private $gradeable = null;
+    /** @property @var string Id of the gradeable this grade is attached to */
+    protected $gradeable_id = "";
+
+    /** @property @var int The id of this gradeable data */
+    protected $id = 0;
+    /** @property @var string The grader's overall comment */
+    protected $overall_comment = "";
+    /** @property @var \DateTime The date the user viewed their grade */
+    protected $user_viewed_date = null;
+
+    /** @property @var Submitter The submitter who received this graded gradeable */
+    protected $submitter = null;
+    /** @property @var GradedComponent[] The graded components */
+    protected $graded_components = array();
+
+
+    /**
+     * GradedGradeable constructor.
+     * @param Core $core
+     * @param Gradeable $gradeable The gradeable associated with this grade
+     * @param Submitter $submitter The user or team who submitted for this graded gradeable
+     * @param array $details A property-name-indexed array of values to construct with
+     * @throws \Exception If the 'user_viewed_date' in the $details array is an invalid DateTime/date-string
+     */
+    public function __construct(Core $core, Gradeable $gradeable, Submitter $submitter, array $details) {
+        parent::__construct($core);
+
+        $this->setGradeable($gradeable);
+        $this->setSubmitterInternal($submitter);
+
+        $this->setIdInternal($details['id']);
+        $this->setOverallComment($details['overall_comment']);
+        $this->setUserViewedDate($details['user_viewed_date']);
+    }
+
+    public function toArray() {
+        $details = parent::toArray();
+
+        // When serializing a graded gradeable, put the grader information into
+        //  the graded gradeable instead of each component so if one grader  grades
+        //  multiple components, their information only gets sent once
+        $details['graders'] = [];
+        foreach ($this->graded_components as $graded_component) {
+            if ($graded_component->getGrader() !== null) {
+                $details['graders'][$graded_component->getGrader()->getId()] = $graded_component->getGrader()->toArray();
+            }
+        }
+
+        return $details;
+    }
+
+    /**
+     * Gets the gradeable this grade data is associated with
+     * @return Gradeable the gradeable this grade data is associated with
+     */
+    public function getGradeable() {
+        return $this->gradeable;
+    }
+
+    /* Overridden setters with validation */
+
+    /**
+     * Sets the internal gradeable reference
+     * @param Gradeable $gradeable
+     */
+    private function setGradeable(Gradeable $gradeable) {
+        if ($gradeable === null) {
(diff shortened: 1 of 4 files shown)


site/app/models/gradeable/Submitter.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public function getObject() {
+        return $this->team_or_user;
+    }
+
+    public function getId() {
+        return $this->team_or_user->getId();
+    }
+
```
CouleeApps: I'd also appreciate `getUser()` and `getTeam()` that return a (nullable) User/Team object so we don't have to cast `getObject()` every time we want either of those.

site/app/models/gradeable/GradedGradeable.php
```diff
@@ -0,0 +1,162 @@
...
+     * Sets the array of graded components for this gradeable data
+     * @param array $graded_components
+     */
+    public function setGradedComponents(array $graded_components) {
+        foreach ($graded_components as $graded_component) {
+            if (!($graded_component instanceof GradedComponent)) {
+                throw new \InvalidArgumentException('Graded Component array contained invalid type');
+            }
+        }
+        $this->graded_components = $graded_components;
```
CouleeApps: Did we ever determine how we are handling peer grading with this?

MasterOdin:
You still need to implement that giant DB query to get all the data on the fly in one query unless you get buy in from @bmcutler on pagination on all pages that list all gradeables for some number of students. We can also abuse the fact that you're almost always going to want all parts of the graded gradeable (metadata, component data, mark data) for every page that uses it.

This setup uses x3 queries per graded gradeable per user.

KevinMackenzie:
The data from the `electronic_gradeable_data` and `electronic_gradeable_version` is also missing.

KevinMackenzie:
I added the data for the `electronic_gradeable_data` / `electronic_gradeable_version` tables.  I think that's everything we planned on having in these green classes.

I also created a new set of database queries that retrieves all (or a subset) of `GradedGradeable`s for a given `Gradeable`.  This is done in `O(1)` query count

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -825,5 +829,305 @@ public function updateLateDays($user_id, $timestamp, $days, $csv_option=null) {
...
+                  user_id,
+                  team_id
+                FROM electronic_gradeable_data as in_egd
+                GROUP BY g_id, user_id, team_id
+              ) AS egd ON (egd.team_id=gd.gd_team_id OR egd.user_id=gd.gd_user_id) AND egd.g_id=gd.g_id
+            WHERE gd.g_id=? AND ((gd.gd_user_id IN (?)) OR (gd.gd_team_id IN (?)) OR ?)";
+
+        $this->course_db->query($query, array(
+            $gradeable->getId(),
+            implode(',', $users),
```
MasterOdin: Does this paradigm work in PDO? I'd think that in executing your query, it'd only quote around the whole string (so it'd be doing `gd.gd_user_id IN ('user_1, user_2')` instead of `gd.gd_user_id IN ('user_1', 'user_2')`

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -825,5 +829,305 @@ public function updateLateDays($user_id, $timestamp, $days, $csv_option=null) {
...
+        // Remove null / duplicate values being sneaky in our arrays
+        $filter_null = function($val) { return $val !== null; };
+        $user_ids = array_filter(array_unique($user_ids), $filter_null);
+        $team_ids = array_filter(array_unique($team_ids), $filter_null);
+
+        $users = $this->getUsersById($user_ids);
+        $teams = $this->getTeamsById($team_ids);
+
+        // decode the gradeable data
+        foreach ($db_rows as $db_row) {
```
MasterOdin: It should be noted that the reason the user and team information was being joined was because it allowed us to write one query, which while yes, did get some duplicate information per row, meant that we could use a DatabaseIterator to go through the data row by row when we needed the info. This is drastically more efficient in both time spent and memory used as we largely don't need all of the information at one time, but rather row by row.

KevinMackenzie:
I integrated the submitter/grader information into the main query.  With a very small sample size (5 page loads), there was not a significant difference in page load times.  I think what I did is fine, but someone who knows more about databases than I do might want to look over the query to see if there's anything that could be improved.

site/app/models/gradeable/GradedVersion.php
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * @method int getVersion()
+ * @method float getNonHiddenHonExtraCredit()
+ * @method float getNonHiddenExtraCredit()
+ * @method float getHiddenNonExtraCredit()
+ * @method float getHiddenExtraCredit()
+ * @method \DateTime getSubmissionTime()
+ * @method isAutogradingComplete()
+ */
+class GradedVersion extends AbstractModel {
```
CouleeApps: GradedGradeableVersion? I know the original class was called GradeableVersion

CouleeApps: Actually it seems like this class is mostly autograding results. Maybe it should be named something that indicates that.

site/app/models/gradeable/GradedGradeable.php
CouleeApps: Seems like our best option for peer grading would be some sort of subclass, eg:
```
class PeerGradedComponent extends GradedComponent {
    /** @property @var GradedComponent[] $peer_grades */
    protected $peer_grades;
}
```

site/app/models/gradeable/GradedVersion.php
KevinMackenzie: I renamed it to `AutogradingVersion` since it does only contain information about autograding

KevinMackenzie:
Are there any other major changes that need to be made before merging this?  There are some other changes I'm looking to push and it would make merging less of a headache to get this into master first.

site/app/models/gradeable/GradedGradeable.php
```diff
@@ -0,0 +1,224 @@
...
+    protected $user_viewed_date = null;
+    /** @property @var int The active submission version for electronic gradeables */
+    protected $active_version = 0;
+
+    /** @property @var Submitter The submitter who received this graded gradeable */
+    protected $submitter = null;
+    /** @property @var GradedComponent[][] The an array of arrays of GradedComponents, indexed by component id */
+    protected $graded_components = array();
+    /** @property @var array AutogradingVersion[] The graded versions for electronic gradeables */
+    protected $graded_versions = array();
```
CouleeApps: Need to change the var name

CouleeApps:
No other major changes that I can think of, I'll approve after that one commment

PR #2197, opened 2018-06-18
Fixed twig footer escaping db queries

KevinMackenzie:
The new twig template was escaping the db queries text, which contained html line breaks.  I removed this escaping so the text displays on multiple lines as intended.


CouleeApps:
Nice ninja PR

PR #2205, opened 2018-06-19
Twig/Refactor Navigation Page

### LINKED ISSUE #2152: gradeable url/link & delete gradeable X are not both displayed
If both should be displayed, only the link is displayed.

At the same time, let's convert the red X to a trashcan icon (from our chosen icon library).

CouleeApps:
Massive cleanup of the nav view code and Twig rendering. This is quite a bit of stuff so I'd appreciate if someone could test basically everything on [the spreadsheet](https://camo.githubusercontent.com/f8a9e5c5a33cc95de38e2e1c232f4d21a5ae5d4a/68747470733a2f2f692e696d6775722e636f6d2f54566b397170792e6a7067).

Also note this is _not_ a redesign, it is simply a refactor of the old view. Redesign will come in a later PR.

Closes #2152


saxocellphone:
The delete gradeable button makes a popup, but it doesn't actually delete the gradeable

saxocellphone:
In FUTURE and BETA category, when I click on Open to (whatever), it gives me  Call to undefined method app\models\AbstractModel::updateGradeable()

CouleeApps:
Looks like the latter is a regression as of #2132, I'll fix it for this PR though

saxocellphone:
@bmcutler currently there is no defined behavior for version conflicts in the big spreadsheet. Should we be discussing that?

bmcutler:
What do you mean?  The student finds out about the automatic zero due to version conflict when they view grade.  It doesn't need to appear on the nav page.  Or do you mean what % complete is a TAs grading if there are version conflicts?  Maybe later we'll hook up the regrade request to the after deadline change of active version.  But I don't think version conflict should be a significant visual difference on the nav page.

tests/e2e/test_navigation_page_non_student.py
```diff
@@ -63,7 +63,7 @@ def test_ta(self):
                                 .find_element_by_id('graded_tbody')
                                 .find_elements_by_class_name("gradeable_row")))
 
-        self.assertEqual(7, len(self.driver.find_element_by_class_name(
+        self.assertEqual(4, len(self.driver.find_element_by_class_name(
```
MasterOdin: Why is this being decreased?

site/app/libraries/Button.php
```diff
@@ -0,0 +1,77 @@
+<?php
```
MasterOdin: This feels like it'd be a model to me, albeit one not filled with DB information

CouleeApps: Yeah, I wasn't sure if that counted as model because it's not inheriting from AbstractModel. If PHP supported nesting classes I would have just put it inside NavigationView.

tests/e2e/test_navigation_page_non_student.py
CouleeApps: Graders / TAs were being shown 3 empty columns all the time from the instructor-only buttons. This just cleans that up.

site/app/libraries/Button.php
MasterOdin: Why not have it inherit from AbstractModel? You'd be able to get rid of all of your getter functions and just use the magic `__call` function to handle it?

CouleeApps: That's smart. I'll go change that

PR #2206, opened 2018-06-19
Instructor merge bulk pdf

### LINKED ISSUE #1884: Reduce clicks for merge option on bulk pdf upload & extend to simple upload for student
If an instructor is doing alot of merges on a bulk upload, allow them to check a box once to make merge the default way to handle multiple bulk pdf upload submission for the same student.

Also, allow instructor to upload one or more files by drag & drop with the upload for student option that does a merge with previous version files...  (do something to ensure files with same name don't get clobbered... )

ropern123:
Slight UI change for instructor submit (new validate prompt with new submission, merge, and merge clobber options),
checkbox to make merge default option for instructor submit,
and allows instructors to upload multiple files for a student with a working merge option.

closes #1884

### DIFF
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -922,7 +922,7 @@ class SubmissionController extends AbstractController {
                     }
                 }
                 
-                if(merge_previous) {
+                if($merge_previous) {
                     for($i = 1; $i <= $gradeable->getNumParts(); $i++) {
                         if(isset($uploaded_files[$i])) {
                             $previous_files[$i] = array();
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -409,9 +409,14 @@ class SubmissionController extends AbstractController {
         }
 
         $merge_previous = false;
+        $clobber = false;
         if(isset($_REQUEST['merge'])){
-            if($_REQUEST['merge']  === "true"){
+            if($_REQUEST['merge'] === "true") {
                 $merge_previous = true;
+            }
+
+            if(isset($_REQUEST['clobber']) && $_REQUEST['clobber'] === "true") {
+                $clobber = true;
             }
         }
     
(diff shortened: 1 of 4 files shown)


ropern123:
@bmcutler : You had requested that the previous submission files have \_version_# appended to the end of their names and that the most recent version , and, following this, there are a few cases where name conflicts arise when merging that I'd like to get your input regarding how they should be addressed (if at all):

* An instructor is merging foo.txt to create submission 2, but there is already foo.txt and foo_version_2.txt in submission 1: i.e. old file rename conflicts with another old file. (We would try to change the name of foo.txt in the original submission to foo_version_2.txt, but this conflicts with the previous submission. What name should be used to avoid clobbering and which file should have its name changed? Then, if this new name is already taken, what name should be used, and if that name is taken, what name should be used etc. etc. : i.e. what rule should be used to eventually generate a name that does not conflict?)
* An instructor is merging foo.txt and foo_version_2.txt to create submission 2, but there is already foo.txt in submission 1: i.e. old file rename conflicts with a new file. (We would try to change the name of foo.txt in the original submission to foo_version_2.txt, but this conflicts with an uploaded file. What rule should be used to eventually generate a name that does not conflict, and which file should have its name changed?)
* An instructor is merging foo_version_3.txt to create submission 3, but there is already foo_version_3.txt in submission 2: i.e. the conflicting file already ends with \_version_#. (Does foo_version_3.txt from submission 2 become foo_version_3_version_2.txt or does it get some other name? If those names are taken, what rule should be used to eventually generate a name that does not conflict?)

site/app/controllers/student/SubmissionController.php
```diff
@@ -447,11 +447,18 @@ private function ajaxUploadSplitItem() {
...
         if(isset($_REQUEST['merge'])){
-            if($_REQUEST['merge']  === "true"){
+            if($_REQUEST['merge'] === "true") {
                 $merge_previous = true;
             }
+
+            if(isset($_REQUEST['clobber']) && $_REQUEST['clobber'] === "true") {
+                $clobber = true;
+            }
         }
```
MasterOdin: Condense this all to:
```
$merge_previous = isset($_REQUEST['merge']) && $_REQUEST['merge'] === 'true';
$clobber = isset($_REQUEST['clobber']) && $_REQUEST['clobber'] === 'true';
```
We require `$merge_previous` to be true to use `$clobber`, so we don't need to care about checking the isset of merge for setting.

site/app/controllers/student/SubmissionController.php
```diff
@@ -752,6 +767,19 @@ private function ajaxUploadSubmission() {
...
+        $merge_previous = false;
+        $clobber = false;
+        if(isset($_REQUEST['merge'])) {
+            if($_REQUEST['merge'] === "true") {
+                $merge_previous = true;
+            }
+            if(isset($_REQUEST['clobber']) && $_REQUEST['clobber'] === "true") {
+                $clobber = true;
+            }
+        }
```
MasterOdin: same as above

bmcutler:
We discussed in person. 
Primarily we are concerned with simple merge.
We will assume the student never names their files with suffix "_version_#".

It's ok if this behaves weird/incorrect with those names.  (We don't actually lose data, the previous versions are still there).

Long term, if necessary, we can come up with a different plan.

jchulton:
Looks good to me. Not sure why the line $section !== null is causing failure in Travis builds though.

bmcutler:
@ropern123 
Please resolve conflicts
thanks :)

bmcutler:
@ropern123 
You're failing some travis tests about upload & error conditions.
The error messages don't match.

Can you please look into these tests and figure out if we need to fix the error handling or update the tests.

Thanks

PR #2215, opened 2018-06-20
Extra package installation for Lichen Python & C/C++ tokenization 

bmcutler:


.setup/distro_setup/ubuntu/xenial/setup_distro.sh
```diff
@@ -77,6 +77,9 @@ apt-get install -qqy ninja-build
 echo "installing cmake"
 apt-get install -qqy cmake
 
+# for Lichen (Plagiarism Detection)
+apt-get -qqy install python-clang-3.8
```
MasterOdin: So I'm not sure I follow what all is being installed here and why. What does the clang pypi package get us that python-clang-3.8 doesn't? They're both "Clang python bindings"?

From what I can tell though, the pypi package is also for clang-6 and not 3.8?

PR #2218, opened 2018-06-20
Twig simple grader view

CouleeApps:
Pretty much just a direct port of the php with some minor regrouping of rows


CouleeApps:
Alright, should have resolved those issues.

PR #2220, opened 2018-06-20
Student View Version Conflict Bug

### LINKED ISSUE #2041: Student View of TA grade: Cancelled, Active, Grade Version Bugs
When the active version does not match the graded version, the grade total should be an automatic zero (until the mismatch is corrected).

The student should see an error message.  (currently the message appears if the selected version is not the active version.  But it should instead be comparing to the ta graded version).

Verify that Grade Summaries is also giving an automatic zero if active does not equal ta graded.

### LINKED ISSUE #2211: Version conflict bug
If the TA graded version 2, but student selected version 1 to grade, the student's view TA grade view does not give any errors. However, if the student switches to version 2 in the dropdown menu, the error message appears. This is wrong because the error message should appear if there is a version conflict.

ropern123:
Fixes message displayed to student when changing their active version after grading has started, and grade results only display when active version is the same as the graded version.
Also shows the version that was graded in the results (for clarity).

closes #2041 and closes #2211


saxocellphone:
Something minor. In Grades Released Homework, if the assignment has a version conflict, the student will see "Note: Grading has already begun." I'm not sure if that's the best wording because the grades are released already.

PR #2235, opened 2018-06-22
Database queries for creating/modifying Gradeable/Component/Mark

KevinMackenzie:
Part of (#2107).  The most significant update is a set of database queries for saving and creating gradeables, components, and marks.  To make these methods efficient, I also added support for the `AbstractModel`s `modified` field so unchanged data doesn't need to be queried.

There is also now a guarantee that the components in a gradeable and the marks in a component will be sorted by `order` even if that order is invalid.  Going along with this, the order of components/marks gets rectified when saving to the database.

There are a couple of other misc bugfixes organized by commit

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -11,7 +11,10 @@ use app\libraries\GradeableType;
 use app\models\AdminGradeable;
 use app\models\Gradeable;
 use app\models\gradeable\Component;
+use app\models\gradeable\GradedComponent;
+use app\models\gradeable\GradedGradeable;
 use app\models\gradeable\Mark;
+use app\models\gradeable\Submitter;
 use app\models\GradeableComponent;
 use app\models\GradeableComponentMark;
 use app\models\GradeableVersion;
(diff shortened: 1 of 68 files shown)


KevinMackenzie:
To reduce the number of PR's related to this model, I added some more things to this PR. I added a couple of new properties to the Gradeable model class.  These are `has_manual_grades` which is a value loaded at construct time from the database to tell the user if any entries in the `gradeable_data` table exist for this gradealbe.  The other property is a lookup table to see, given a grader id, which rotating sections they are assigned to.

KevinMackenzie:
Yea, seems fine.

site/app/models/gradeable/Gradeable.php
```diff
@@ -556,6 +599,11 @@ public function setComponents(array $components) {
             }
         }
         $this->components = $components;
+
+        // sort by order
+        usort($this->components, function(Component $a, Component $b) {
+            return $a->getOrder() - $b->getOrder();
+        });
     }
```
CouleeApps: `setComponents` should update `modified` (I assume)

KevinMackenzie: Since changing components doesn't require a write to the `gradeable` or `electronic_gradeable` table, it doesn't.  The components individually get checked for their own `modified` field.

KevinMackenzie:
There are several methods that don't use the `modified` field.  The way that `modified` gets used is to signal whether the database table would need to be`UDPATE`d for that instance.  So, the `gradeable` doesn't have `modified` marked as `true` if the components change.  Similarly, the `component` doesn't have its `modified` marked as `true` if the marks change.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2652,4 +2674,418 @@ public function getTeamsById(array $team_ids) {
...
+                $invalidateComponents = true;
+            } else {
+                $this->updateComponent($component);
+            }
+        }
+
+        // If any components were new, recreate the components array
+        $components = $gradeable->getComponents();
+        if ($invalidateComponents) {
+            $new_components = $this->getGradeableComponentConfigs($gradeable);
```
CouleeApps: Any reason you have to re-query the database here? I'd recommend making `createComponent()` return the id of the added database entry so you don't have to reload to get updated ids.

site/app/models/gradeable/Gradeable.php
CouleeApps: Understandable.

site/app/libraries/database/DatabaseQueries.php
KevinMackenzie: The idea was that components and marks can't have their id's changed during their lifecycle, but I definitely understand that this suboptimal.  Changes to the number of components in a gradeable is rare since only the instructor can do it and they typically only do it a few times.  Adding/deleting marks is more common and results in a similar reload.

CouleeApps: Yes, this won't happen often, but it still seems unnecessary to have to delete and reload all mark and component data, especially if other classes already have references to the old objects.

CouleeApps: Yes, this won't happen often, but it still seems unnecessary to have to delete and reload all mark and component data, especially if other classes already have references to the old objects.

KevinMackenzie:
I'm also writing a `DatabaseRowIterator`-compatable query for static gradeable data.  Don't merge until then. (you can't request changes on your own PR)

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2518,25 +2523,42 @@ public function getGradeableConfig($g_id) {
...
         // Finally, create the gradeable
         $gradeable = new \app\models\gradeable\Gradeable($this->core, $details, []);
 
         // Get the components data
         $gradeable->setComponents($this->getGradeableComponentConfigs($gradeable));
 
+        // Set the rotating sections data
+        $gradeable->setRotatingGraderSections($this->getGradersForAllRotatingSections($g_id));
+
         return $gradeable;
```
CouleeApps: `setRotatingGraderSections` marks the `Gradeable` as modified before it even returns from this function. You should `setModified(false)` it at the end or move those setting functions into its constructor before you unset `modified`.

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -929,6 +931,26 @@ public function getTeamsById(array $team_ids) {
...
     }
 
+    /**
+     * Converts the result of array_to_string(*, ',') into a php
+     *  array using the provided conversion function
+     * @param string|mixed $str string array of values, or single value from database
+     * @param callback $parser A method to parse a string representation of the object (i.e. 'intval')
+     * @return array A php-array representation of the provided array string
+     */
+    private static function parseShallowCommaSeparatedArrayString($str, $parser) {
```
CouleeApps: Since we have Postgres > 9.2, we may want to use this instead: https://stackoverflow.com/a/13670706/214063

KevinMackenzie:
@CouleeApps I added everything that should go in this PR and fixed everything I could find.

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -1277,5 +1278,200 @@ function($elem) { return intval($elem); },
...
+                'order'
+            ];
+            $component_mark_properties = array_map(function ($value) {
+                return 'mark_' . $value;
+            }, $mark_properties);
+
+            // Unpack the component data
+            $unpacked_component_data = [];
+            foreach (array_merge($component_properties, $component_mark_properties) as $property) {
+                $unpacked_component_data[$property] = $this->course_db->fromDatabaseToPHPArray($row['array_' . $property]);
```
CouleeApps: I'd also recommend json_decode for these

KevinMackenzie: I switched to using the json method.  I don't have a precise way to test what the performance difference is between the two though.

KevinMackenzie:
I did some tests and it looks like using the `json_decode` method is faster with a P-val < 0.0001

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2406,250 +2412,460 @@ public function deleteGradeable($g_id) {
...
-        $component_marks = [];
-
-        // Parse the marks per component
-        foreach($marks as $mark) {
-            if (array_key_exists($mark['gc_id'], $component_marks)) {
-                $component_marks[$mark['gc_id']][] = $mark;
-            } else {
-                $component_marks[$mark['gc_id']] = [$mark];
-            }
+    public function getGradeableConfig($id) {
```
CouleeApps: phpdoc with param / return types would be helpful

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2406,250 +2412,465 @@ public function deleteGradeable($g_id) {
...
+              gc_max_value, 
+              gc_upper_clamp,
+              gc_is_text, 
+              gc_order, 
+              gc_is_peer, 
+              gc_page)
+            VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)", $params);
+
+        // Setup the component with its new id
+        $component->setIdFromDatabase($this->course_db->getLastInsertId());
```
CouleeApps: Needs an `intval()` around `$this->course_db->getLastInsertId()` which apparently returns a string

CouleeApps: Same for `createMark`

KevinMackenzie: Ids get parsed as strings in all places inside the model classes now (only some were previously)

CouleeApps:
At this point, `setOrder` is completely useless considering `updateGradeableComponents` changes all the orders to be in sequence. We should either discourage `setOrder` or stop changing the order when updating. I'd vote for the latter because it will make reordering marks easier

KevinMackenzie:
Yea, I screwed this up.  I moved the sorting from the db method to the setter and forgot about this.  With the sort in the db method the calls to `setOrder` will be preserved, but all of the order values will be unique and sequential.

KevinMackenzie:
I'd like to note that a call to `updateGradeable` will not clear the `modified` flag in the `AbstractModel`, but it is very likely that this call will be the last usage of the `Gradeable` instance in its lifecycle, so I don't think its a big deal.

CouleeApps:
Speaking of `modified`, calling `createComponent` / `createMark` doesn't unset `modified`.

KevinMackenzie:
Is there anything keeping this from being merged?

CouleeApps:
Short of that last small thing I mentioned yesterday, nothing I can think of.

KevinMackenzie:
Similar to the gradeable, creating / updating a component / mark, is
typically only done once during its life cycle, so it shouldn't be a big
deal to not reset the modified field.

PR #2238, opened 2018-06-22
Changed student infomation panel to twig

saxocellphone:
~~Since the latedays table is already implemented with Twig, I simply just rendered that instead of building one from scratch. There is only one visual difference: since the twig has an extra content div wrapped around it, each late days table in the TA view will have the same div like this.~~
capture
I happen to really like this design, because it makes the table more "table like". But I can always change it.

No longer does that


CouleeApps:
Seems strange to me that we're including a div.content inside a panel. I'd say factor the table and stuff out and not have a box-within-a-box.

site/app/templates/LateDaysTable.twig
```diff
@@ -1,4 +1,7 @@
+{% if (is_own_page is not defined or is_own_page == true) %}
 <div class="content">
+{% endif %}
```
CouleeApps: I'm really not a fan of {% if %}ing around open/close tags. We should really try to keep all tags matched so there's no chance of losing one if a condition is forgotten.

CouleeApps:
Rendering the late days table adds a breadcrumb entry for each student. That means, for graded team assignments, this happens:

CouleeApps:
Still shows the breadcrumb if you try to grade yourself...

PR #2240, opened 2018-06-22
Twig homework submission view

### LINKED ISSUE #519: Break apart HomeworkView monster function
Right now there's a function that has ~300 lines of code in it which is making it hard to manage and it does simply far too much. This should be broken up into additional view functions which make management of that code easier.

CouleeApps:
Broke this into a bunch of functions and files so it didn't just become one monster twig file. Maybe we can even make batch upload its own page in the future. For now this should look and work the exact same as before. 

Closes #519 

Best bet would be to test all the gradeables in the development repo as it most thoroughly tests this page.


bmcutler:
The naming might be legacy, but we should move away from using "homework" in most instances.
You can submit and grade things that are not homeworks.
Is it gradeable or electronic gradeable?

CouleeApps:
This is only electronic gradeables. Or at least, it's only what `HomeworkView::showGradeable` does.

CouleeApps:
Looks like all the other HomeworkView.php pull requests have been merged, so this shouldn't have anything else to conflict with. Would be nice to get a review.

Issue #2201, opened 2018-06-19
Forum Mid-Summer


andrewaikens87:
Action items for Mid-Summer goal: 

- [x] the size of the reply box for a reply to a non-top level post is currently fixed. Please make this box auto-expand if the typed reply is lengthy. Note: the default reply (to the top level post) is re-sizeable.

- [ ] ~#1971 - discussed over slack~

- [ ] attaching multiple attachments one at a time (& removing an attachment) vs. selecting all at once

- [x] Edit categories: Allow thread publisher/instructor/ta modify the categories for a given thread

PR #2243, opened 2018-06-23
Forum Multiple Attachements

scopeInfinity:
Issue: #2201 

Tasks -

- [x] Multiple Attachments in Create Thread Page
- [x] Multiple Attachments in Reply Posts
- [ ] ~Edit Attachments in Edit Posts~
- [x] Thumbnail for attachmens


jn7070:
Everything looks good, but is there supposed to be multiple images of the same image even after the message pops up asking if you want to replace it? After pressing submit there is only 1 copy of each image so that works.

test

scopeInfinity:
@jn7070 @andrewaikens87 Thankyou for pointing it out.
Fixed it.

PR #2260, opened 2018-06-26
[Initial Work] Create general hwcron job daemon

### LINKED ISSUE #1979: Create general hwcron job daemon

MasterOdin:
Closes #1979


MasterOdin:
@bmcutler, where would you expect this to be installed out of the repo? Or I guess we could run it directly from the repo if we wanted to.

bmcutler:
Still need to hookup jobs to this (and test of course)

PR #2277, opened 2018-06-28
Upload & Display Images of Students

### LINKED ISSUE #1824: Upload & display photos of students
Use case: In a huge class it can be hard for instructors & TAs to learn student names.  We'd like to be able to look up a picture of each student.  

Source of photos:  
1) Instructors may have access to student photos from their university.  Let's assume they can rename the photos to be USERNAME.png or USERNAME.jpg & prepare a zip for upload.
2) We could also allow each student to upload their own image.  However since students may not upload a clear identifiable image, we should store both instructor & student uploaded images.

Display of photos:
1) Instructor & full access graders should have access to view both instructor & student uploaded images.
2) Students can view another student's uploaded image -- but only if the student opts-in to sharing the image.
3) Let's add a page somewhere that let's instructors & full access graders to view a 'yearbook-style' grid of the photos of all students grouped by registration section.

Questions...  Should the photos be used like avatar/profile/icons in other apps? (like gihub/slack?)  Many people select images to express their online personality.  So it's not the same use case.  My primary interest here is to help instructors learn their student's names.

Other comments?

jn7070:
Closes #1824

This is version 1 of the images page. Instructors are the only ones who can upload images by drag and/or drop. Test cases are in sample_files/user_photos which should be stored in /var/local/submitty/courses/s18/sample/uploads/student_images after you click submit. Full-access graders can see all images and limited-access graders can only see the registration sections they are assigned to.

Couple things that need work/not sure how to proceed:
1. Uploading files would cause "ERROR! Please contact administrator that you could not upload files." message to pop up which is disabled for now.
2. Message to print out that the instructor has exceeded file size limit of 10 MB does not work.
3. Window does not automatically reload after uploading so user needs to refresh the page.


site/app/controllers/grading/ImagesController.php
```diff
@@ -0,0 +1,29 @@
...
+use app\libraries\FileUtils;
+
+class ImagesController extends AbstractController {
+    public function run() {
+        switch ($_REQUEST['action']) {
+            case 'view_images_page':
+            		$this->view_images_page();
+                break;
+            default:
+                $this->viewPage();
```
MasterOdin: non-existant function?

MasterOdin: non-existant function?

site/app/controllers/grading/ImagesController.php
```diff
@@ -0,0 +1,29 @@
...
+                break;
+        }
+    }
+
+		public function view_images_page() {
```
MasterOdin: camelCase for function names in OO PHP code.

site/app/views/grading/ImagesView.php
```diff
@@ -0,0 +1,58 @@
...
+                $imgName = $pi['filename'];
+                if ($imgName === "error_image") {
+                    $error_image_data = $expected_img_data;
+                }
+                else {
+                    array_push($images_data_array, $expected_img_data);
+                    array_push($images_names_array, $imgName);
+                }
+            }
+        }
```
MasterOdin: Instead of using glob and then looping over the resulting set, you should just use a [DirectoryIterator](http://de.php.net/manual/en/directoryiterator.construct.php) ([example](https://stackoverflow.com/a/4202257/4616655)) to go through the directory and construct the resulting set at the same time.

site/app/views/NavigationView.php
```diff
@@ -88,6 +88,17 @@ public function showGradeables($sections_to_list) {
         /* @var Button[] $top_buttons */
         $top_buttons = [];
 
+		// ======================================================================================
```
MasterOdin: spaces over tabs

site/app/controllers/student/SubmissionController.php
```diff
@@ -1391,6 +1392,142 @@ private function updateSubmissionVersion() {
...
+                }
+                else {
+                    $extension = strtolower(pathinfo($file, PATHINFO_EXTENSION));
+                    if ($extension === 'png') // just get all png file only
+                        copy($src . '/' . $file,$dst . '/' . $file);
+                }
+            }
+        }
+        closedir($dir);
+    }
```
MasterOdin: Move this function to the `FileUtils` library, since we can then more easily write tests for it, and at that time we can rewrite it to use the DirectoryIterator as well.

site/app/controllers/student/SubmissionController.php
```diff
@@ -1391,6 +1392,142 @@ private function updateSubmissionVersion() {
         return array('error' => false, 'version' => $new_version, 'message' => $msg);
     }
     
+
+	// This copy all files from given dir ($src) and its subdirs into one destination folder, $dst
+	private function recursive_copy($src, $dst) {
+ 		$dir = opendir($src);
```
MasterOdin: spaces over tabs

jn7070:
The zip files in sample_files/user_photos should contain images of most of the students in the sample course of different sections and duplicate students' images for testing purposes.

site/app/controllers/student/SubmissionController.php
```diff
@@ -1390,7 +1391,124 @@ private function updateSubmissionVersion() {
 
         return array('error' => false, 'version' => $new_version, 'message' => $msg);
     }
-    
+
+    private function ajaxUploadImagesFiles() {
+        if (empty($_POST)) {
+           $max_size = ini_get('post_max_size');
+           return $this->uploadResult("Empty POST request. This may mean that the sum size of your files are greater than {$max_size}.", false);
```
ropern123: Add a permissions check to this function so students can't submit image files

MasterOdin:
I think I'd prefer opening an issue to generate the images on the fly via a script (like how GitHub does theirs) since otherwise, we're adding 3+mb of binary files that will break if we ever change how students are generated by setup_sample_courses.py

ropern123:
The images page for a grader with no sections has just the header, background, and footer. It might be nice to have content on the page saying that they have no assigned sections and thus no images to view.

site/app/controllers/student/SubmissionController.php
```diff
@@ -1390,7 +1391,128 @@ private function updateSubmissionVersion() {
 
         return array('error' => false, 'version' => $new_version, 'message' => $msg);
     }
-    
+
+    private function ajaxUploadImagesFiles() {
+        if($this->core->getUser()->getGroup() === 4) {
+			     return $this->uploadResult("You have no permission to access this page", false);
+        }
+
```
ropern123: Also, one needs instructor permissions to view the upload images button, so I think it would make sense to need the same level of permissions to upload images, not just grader level.

PR #2281, opened 2018-06-28
[BUGFIX] fix crash for null value

bmcutler:
09:51:14 06/28/2018 - FATAL ERROR
Error (Code: 0) thrown in /usr/local/submitty/site/app/controllers/grading/ElectronicGraderController.php (Line 879) by:
            $gradeable = $this->core->getQueries()->getGradeable($gradeable_id, $teams_assoc[$who_id]->getLeaderId());

Message:
Call to a member function getLeaderId() on null


CouleeApps:
Do you have a case where the bug happens?

bmcutler:
Yes, on the production machine :)  With a gradeable from S18 Viz.

CouleeApps:
Any chance I could cause something similar to happen on vagrant to test it?

CouleeApps:
Figured a method myself, looks like this works.

PR #2284, opened 2018-06-28
Updated "Upload Config page"(For Evan) and "Create/Edit gradeable page"

### LINKED ISSUE #914: Improve Config Upload Page
@bmcutler said in #911:
> @MasterOdin The upload config page still needs some work. At the minimum, it should say what version was just uploaded (a copy paste line that can be inserted into the add-or-edit-gradeable page). And it would be nice to have it refresh automatically when the build output finishes (requires reload now). And it would be great to see file names and even file contents for previous uploads.

saxocellphone:
New things in "Upload Config Page":
- Moved the view output log to the auto-grading section of create/edit gradeable page.
- Now can delete the unused uploaded configs
- Now can edit the name of all uploaded configs (With warning if any gradable is using them)

New things in "Create/edit gradeable page"
- Added four radio buttons to make selecting the gradeable config easier.
-- Default configs (hardcoded)
-- Uploaded configs (All the configs the user uploaded from the "Upload Config" page)
-- Use private repository (Specified in the course setting, will not show if repo is not used)
-- Provide full path (Just like before, manually enter the path)
- Build output appears at the bottom (the rebuild button turns red if there is an error)

Also closes #914


KevinMackenzie:
is this file supposed to be here? `more_autograding_examples/cpp_cats/config.zip`

site/app/controllers/admin/GradeableController.php
```diff
@@ -74,4 +99,71 @@ public function process_config_upload() {
...
+        if($config_file_path == null){
+            $this->core->addErrorMessage("Unable to find file");
+
+        } else {
+            $new_name = $_POST['new_config_name'] ?? null;
+            $new_dir = FileUtils::joinPaths(dirname($config_file_path, 1), $new_name);
+            if(rename($config_file_path, $new_dir)){
+                $this->core->addSuccessMessage("Successfully renamed file");
+                $this->core->redirect($this->core->buildUrl(array('component' => 'admin', 'page' => 'gradeable',
+                    'action' => 'upload_config')));
```
KevinMackenzie: I think this is redundant because it is called at the end of the function anyway

site/app/controllers/admin/GradeableController.php
```diff
@@ -74,4 +99,71 @@ public function process_config_upload() {
...
+            foreach($this->core->getQueries()->getGradeablesIterator() as $gradeable){
+                if($gradeable->getConfigPath() == $config_path){
+                    $inuse_config[] = $gradeable->getId();
+                }
+            }
+        }
+        return $this->core->getOutput()->renderJson($inuse_config);
+    }
+
+    private function rrmdir($dir) {
```
KevinMackenzie: Can you add php doc to this function?  Its behavior is not obvious from the name (maybe also lengthen name to `recursive_rmdir`)

site/app/templates/admin/UploadConfigForm.twig
```diff
@@ -35,30 +34,47 @@
     <br>
     <b>located in {{ target_dir }}</b>
     <br>&nbsp;<br>
+    <link>
+    <script>
+        function renameConfig(file_path){
+            var url = buildUrl({'component': 'admin', 'page': 'gradeable', 'action': 'check_being_used', 'config': file_path});
+            $('#alert_in_use').html('');
+            $('ol').empty();
```
KevinMackenzie: Maybe give your `ol` an id instead of querying for all `ol`'s  This could become a bug later on if any other `ol`s get used in the page.

site/app/templates/admin/UploadConfigForm.twig
```diff
@@ -35,30 +34,47 @@
...
+            $.ajax({
+                url : url,
+                success(data){
+                    var gradeable_ids = JSON.parse(data);
+                    if(gradeable_ids.length > 0){
+                        $('#alert_in_use').html("Note: Currently these gradeables are using this config, rename at your own risk");
+                        for(var i = 0; i < gradeable_ids.length; i++){
+                            $('ol').append("<li>" + gradeable_ids[i] + "</li>");
```
KevinMackenzie: see above comment about `ol`

site/app/templates/admin/UploadConfigForm.twig
```diff
@@ -35,30 +34,47 @@
...
+        function renameConfig(file_path){
+            var url = buildUrl({'component': 'admin', 'page': 'gradeable', 'action': 'check_being_used', 'config': file_path});
+            $('#alert_in_use').html('');
+            $('ol').empty();
+            $.ajax({
+                url : url,
+                success(data){
+                    var gradeable_ids = JSON.parse(data);
+                    if(gradeable_ids.length > 0){
+                        $('#alert_in_use').html("Note: Currently these gradeables are using this config, rename at your own risk");
```
KevinMackenzie: It seems kind of weird to tell the user about this retroactively.  You could probably just word it differently so its more clear that the changes have already been made.

site/app/views/admin/AdminGradeableView.php
```diff
@@ -84,7 +122,16 @@ public function show_add_gradeable($type_of_action, AdminGradeable $admin_gradea
...
+            "all_graders"    => $graders,
+            //Repository name
+            "repo_name"        => $repo_name,
+            //All the uploaded config paths
+            "all_repository_paths"    => $all_repository_paths,
+            "all_uploaded_paths"      => $all_uploaded_paths,
+            "default_paths"           => $default_paths,
+            "which_option"            => $which_option,
+            //build outputs
+            "cmake_output"            => htmlentities($cmake_output)
```
KevinMackenzie: Either package all of these options (line 126 to 134) into a class or given them better names

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -1,11 +1,53 @@
 <div class="electronic_file">
     <b>Full path to the directory containing the autograding config.json file:</b><br>
-    See samples here: <a target=_blank href="https://github.com/Submitty/Tutorial/tree/master/examples">Submitty GitHub sample assignment configurations</a><br>
-    <kbd>/usr/local/submitty/more_autograding_examples/upload_only/config</kbd>  (an assignment without autograding)<br>
-    <kbd>/var/local/submitty/private_course_repositories/MY_COURSE_NAME/MY_HOMEWORK_NAME/</kbd> (for a custom autograded homework)<br>
-    <kbd>/var/local/submitty/courses/{{ semester }}/{{ course }}/config_upload/#</kbd> (for a web uploaded configuration)<br>
-
-    <input style='width: 83%' type='text' id='config_path' name='eg_config_path' value="{{ admin_gradeable.getEgConfigPath() }}" class="required" placeholder="(Required)" />
+    <fieldset>
+        <input type="radio" id="default_configs" class="eg_config" value="true" {% if which_option == 0 %} checked = "true" {% endif %}/> Default configs
```
KevinMackenzie: Give this class a better name.  Also, the `eg` prefix is unnecessary.  It was my mistake to add that prefix to everything.

site/app/templates/admin/UploadConfigForm.twig
saxocellphone: That ajax call is only for checking if it's being used or not, it's not actually making any changes.

site/app/templates/admin/UploadConfigForm.twig
```diff
@@ -35,30 +34,47 @@
...
+                        for(var i = 0; i < gradeable_ids.length; i++){
+                            $('ol').append("<li>" + gradeable_ids[i] + "</li>");
+                        }
+                    }
+                }
+            })
+            $('.popup-form').css('display', 'block');
```
KevinMackenzie: can you do this by id instead of by class

site/app/templates/admin/UploadConfigForm.twig
```diff
@@ -35,30 +34,47 @@
     <br>
     <b>located in {{ target_dir }}</b>
     <br>&nbsp;<br>
+    <link>
+    <script>
+        function renameConfig(file_path){
```
KevinMackenzie: Since this function doesn't actually rename the config, maybe rename this to `openRenamePopup` or something that indicates that it will present the user with the option to rename.

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -95,5 +179,7 @@
         // Only show the settings if the setting is enabled
         onStudentViewChange();
         $('[name="eg_student_view"]').change(onStudentViewChange);
+        $('.eg_config').change(onConfigChange);
+        $('select').change(onConfigPathChange);
```
KevinMackenzie: I don't think we want to call this event handler on all `select` element changes

site/app/models/Config.php
```diff
@@ -157,6 +157,8 @@ class Config extends AbstractModel {
     protected $vcs_base_url;
     /** @property @var string */
     protected $vcs_type;
+    /** @property @var string */
+    protected $private_config_repository;
```
KevinMackenzie: Add a magic getter method in the class phpdoc for this

site/app/templates/admin/RenameConfigPopup.twig
```diff
@@ -0,0 +1,12 @@
+<div class="popup-form" id="rename_config_popup">
+    <h3>What is the new name?</h3>
+    <p id="alert_in_use"></p>
+    <ol style="margin-left:30px">
+    </ol>
+    <form method="POST" enctype="multipart/form-data" action="{{ core.buildUrl({'component': 'admin', 'page': 'gradeable', 'action': 'rename_config'}) }}">
+        <input name="new_config_name" id="new_config_name" value="">
+        <input name="curr_config_name" type="hidden" value="" id="curr_config_name"><br><br>
+        <input type="submit" value="Submit">
+        <input type="reset" value="Cancel" onclick="$('.popup-form').css('display', 'none');">
```
KevinMackenzie: Can you do this by id instead of by class?

KevinMackenzie:
I can't seem to trigger the message to warn the user about renaming configs in use.
Process:
* Upload config file
* Change gradeable config to be that file path
* Attempt to rename file in `upload config` page

KevinMackenzie:
You should reset the 'select' boxes to the blank option when switching between tabs because if you switch to the `provide full path` and change it, then switch back to another tab, the same select option will be selected and you will need to select something else before selecting the shown choice for the change to take effect.

site/app/controllers/admin/GradeableController.php
```diff
@@ -74,4 +101,76 @@ public function process_config_upload() {
...
+            } else {
+                $this->core->addErrorMessage("Directory already exist, please choose another name.");
+            }
+        }
+        $this->core->redirect($this->core->buildUrl(array('component' => 'admin', 'page' => 'gradeable',
+            'action' => 'upload_config')));
+    }
+
+    public function delete_config() {
+        $config_path = $_GET['config'] ?? null;
```
KevinMackenzie: This method should assert that the path is in the `/var/local/submitty/{semester}/{course}/config` folder AND that the provided path does not escape out of that folder.

site/app/controllers/admin/GradeableController.php
```diff
@@ -74,4 +101,76 @@ public function process_config_upload() {
...
+                $this->core->addSuccessMessage("Successfully renamed file");
+            } else {
+                $this->core->addErrorMessage("Directory already exist, please choose another name.");
+            }
+        }
+        $this->core->redirect($this->core->buildUrl(array('component' => 'admin', 'page' => 'gradeable',
+            'action' => 'upload_config')));
+    }
+
+    public function delete_config() {
```
KevinMackenzie: This got marked as outdated, so I'll put it here again: This method should assert that the path is in the `/var/local/submitty/{semester}/{course}/config` folder and that the provided path does not escape out of that folder.

saxocellphone:
I think this should address all the change requests.

KevinMackenzie:
What is this phantom box character doing? 
image

site/app/controllers/admin/GradeableController.php
KevinMackenzie: This was fixed, I can't delete arbitrary folders anymore

KevinMackenzie:
Yea, looks good to me.  I'll let @emaicus give it a review since it is a feature for him

PR #2286, opened 2018-06-28
Create and use Twig cache

### LINKED ISSUE #2052: Setup twig cache layer
We currently do not cache any generated twig template files which isn't ideally for production. We want to setup folder for that cache. The folder would be `site/cache/twig` (with the last two parts being the new folders). `hwphp` would then to then have permissions to write to the twig directory.

We also need to confirm/test if twig will recreate the generated file when we change the template or if we need to clear out that directory on  installation.

CouleeApps:
Twig seems to be able to handle template creation/updates so all we need to do is point it at a directory and it works. 

Closes #2052


CouleeApps:
Only other question is whether we should clear the cache directory on install. I don't think it's necessary but it doesn't seem like that would cause any problems.

MasterOdin:
I believe we have to clear the cache on install as otherwise twig won't recompile templates?

CouleeApps:
Looks like it checks modify timestamp to see if a file is fresh. Because of that, we should definitely clear cache on install.

PR #2288, opened 2018-06-28
Split Manual/Auto graded gradeable and added Autograding Config model

KevinMackenzie:
The new split graded model is tolerant to either a missing/nonexistant TA grade or a missing/nonexistant autograding grade.  If both are missing, then the database method can't reasonably return any meaningful information.  Also included are some helper functions that calculate grade percentages for Ta and autograding.

The other part of the PR is adding a AutogradingConfig class to the model.  This interprets parts of the autograding config file and loads them into memory in a useful format.  This kind of information is helpful for autograding percent score calculation and will be used in both the navigation view and the homework/submission view.


site/app/models/gradeable/AutogradingConfig.php
```diff
@@ -0,0 +1,248 @@
...
+
+    /** @internal */
+    public function setTotalHiddenNonExtraCredit() {
+        throw new \BadFunctionCallException('Setters disabled for AutogradingConfig');
+    }
+
+    /** @internal */
+    public function setTotalHiddenExtraCredit() {
+        throw new \BadFunctionCallException('Setters disabled for AutogradingConfig');
+    }
```
MasterOdin: why not just set these to private and have an empty body?

KevinMackenzie:
I'm marking this as WIP because there are a few things that I don't understand.  Mainly: I'm not sure what all of the Autograding config options mean, so there are TODO's in the comments for those properties.

site/app/models/gradeable/AutogradingConfig.php
MasterOdin: Stick:
```
if (method_exists($this, $name) && !is_callable($this, $name)) {
    throw new \BadFunctionCallException("{$name} is disabled for {$class_name}");
}
```
or something right after the `$class_name = get_class($this);` and then you can just mark all of these functions as private.

The next step I guess would be looking into metaprogramming the `@method` docs from the top of a class into what the allowed callable functions would be for a given class.

KevinMackenzie: I'll make this an issue and can add it in a dedicated PR since the latter option (which I think is a lot more elegant) means we'll need to verify all of the `@method` php docs for the model classes.

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -977,18 +980,28 @@ public function getTeamsById(array $team_ids) {
...
-     *      otherwise use the $users parameter.  You will not get GradeableData
-     *      for a team submission by passing team member names to $users.
-     * @param \app\models\gradeable\Gradeable $gradeable
+     *  Note: The users' teams will be included in the search
+     * @param \app\models\gradeable\Gradeable[] The gradeables to retrieve data for
      * @param string[]|null $users The ids of the users to get data for
-     * @param string[]null $teams The ids of the teams to get data for
+     * @param string[]|null $teams The ids of the teams to get data for
      * @return DatabaseRowIterator Iterator to access each GradeableData
      * @throws \Exception If any GradedGradeable or GradedComponent fails to construct
```
CouleeApps: `getGradeableConfigs` doesn't throw, yet this does. Should be made consistent

site/app/models/gradeable/AutogradingVersion.php
```diff
@@ -72,6 +75,58 @@ public function toArray() {
...
+    public function getNonHiddenPoints() {
+        return $this->non_hidden_non_extra_credit + $this->non_hidden_extra_credit;
+    }
+
+    /**
+     * Gets the percent of the possible visible points the submitter earned
+     * @param bool $clamp True to clamp the output to 1
+     * @return float percentage (0 to 1), or NAN if no visible percent
+     */
+    public function getVisiblePercent($clamp = false) {
```
CouleeApps: `getNonHiddenPercent` to match `getNonHiddenExtraCredit`?

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -1219,8 +1265,20 @@ public function getGradeableDataAll(\app\models\gradeable\Gradeable $gradeable,
...
 
-            // Create the graded gradeable instance
-            $graded_gradeable = new GradedGradeable($this->core, $gradeable, new Submitter($this->core, $submitter), $row);
+            // Create the graded gradeable instances
+            $graded_gradeable = new GradedGradeable($this->core, $gradeable, new Submitter($this->core, $submitter));
+            $ta_graded_gradeable = null;
+            $auto_graded_gradeable = null;
+            if (isset($row['id'])) {
+                $ta_graded_gradeable = new TaGradedGradeable($this->core, $graded_gradeable, $row);
+                $graded_gradeable->setTaGradedGradeable($ta_graded_gradeable);
```
CouleeApps: This is being created and set, even for Gradebles from `grades_released_homework_onlyauto` which specifically has no ta grading.

CouleeApps: Seems like this is because you SELECT FROM gradeable_data and id / `gd_id` is a primary key, so `$row["id"]` can never be null

CouleeApps: Actually it seems that's just a quirk of the sample course. Deleting the gradeable_data a row seems to load this correctly. Still confusing though

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -1016,16 +1033,42 @@ public function getGradeableDataAll(\app\models\gradeable\Gradeable $gradeable,
...
+        $selector = implode(' AND ', $selector_intersection_list);
 
         $query = "
             SELECT /* Select everything we retrieved */
+            
+              (CASE WHEN gd.g_id IS NULL THEN egd.g_id ELSE gd.g_id END) as g_id,
+            
               /* Gradeable Data */
               gd.gd_id AS id,
               gd.gd_user_id as user_id,
```
CouleeApps: If the `gradeable_data` is null then user_id will be null yet you still have user data from the `users` table `u` below. This means you'll get a user with a name and info but no id.

site/app/models/gradeable/AutogradingConfig.php
```diff
@@ -0,0 +1,247 @@
...
+
+        // defaults to 1 if no set
+        $num_parts = count($details['part_names'] ?? [1]);
+
+        // defaults to 0 if not set
+        $num_textboxes = count($details['textboxes'] ?? []);
+
+        // Get all of the part names
+        for ($i = 1; $i <= $num_parts; $i++) {
+            $j = $i - 1;
```
CouleeApps: What's up with this for-loop

KevinMackenzie: Its copy and pasted code from the previous loading method.  I'm not 100% sure what the schema is for the `build_{id}.json` file that this is loaded from and I think its somewhat in flux / needs to be refactored.  I'm not really sure why some parts are 1 indexed and some are 0 indexed

CouleeApps: Well, it's like that in the original code...

site/app/models/gradeable/Gradeable.php
```diff
@@ -238,32 +244,25 @@ public function getComponent($component_id) {
...
         $course_path = $this->core->getConfig()->getCoursePath();
 
         try {
             $details = FileUtils::readJsonFile(FileUtils::joinPaths($course_path, 'config', 'build',
                 "build_{$this->id}.json"));
+
+            // If the file could not be found, the result will be false, so don't
+            //  create the config if the file can't be found
+            if ($details !== false) {
+                return new AutogradingConfig($this->core, $details);
```
CouleeApps: At least with the original model, autograding configs were cached on a per-gradeable basis to cut down on file reading overhead (see GradeableAutogradingConfig.php:11-15). Would be nice to do similar here.

CouleeApps: Actually I've just realized this is not GradedGradeable... disregard this

site/app/models/gradeable/AutogradingVersion.php
```diff
@@ -72,6 +75,58 @@ public function toArray() {
...
+     * @return float percentage (0 to 1), or NAN if no points possible
+     */
+    public function getTotalPercent($clamp = false) {
+        $config = $this->graded_gradeable->getGradeable()->getAutogradingConfig();
+        $divisor = $config->getTotalNonHiddenNonExtraCredit() + $config->getTotalHiddenNonExtraCredit();
+        $dividend = $this->getNonHiddenNonExtraCredit() + $this->getNonHiddenExtraCredit() +
+            $this->getHiddenNonExtraCredit() + $this->getHiddenExtraCredit();
+
+        // avoid divide-by-zero (== not a typo)
+        if($divisor == 0) {
```
CouleeApps: Or just `(int)` or `intval()` ?

site/app/models/gradeable/AutogradingVersion.php
```diff
@@ -72,6 +75,58 @@ public function toArray() {
...
+            return NAN;
+        }
+        $result = floatval($dividend) / $divisor;
+
+        if ($clamp === true && $result > 1.0) {
+            return 1.0;
+        } else if ($result < 0) {
+            return 0.0;
+        }
```
CouleeApps: This whole block of logic seems to be duplicated multiple times in here and TaGradedGradeable. May want to factor that into its own method.

CouleeApps:
Other than a few minor things it seems to be working at least from my tests. Will approve unless you are going to address the last two things above

KevinMackenzie:
I'll try to address them tomorrow morning

KevinMackenzie:
I lied, i addressed them today

KevinMackenzie:
I'm working on converting some views/controllers and it seems like one of the most common use cases is to get a grade for all users, or all grades for one user, even if no data exist for them yet.  Since the graded gradeable query only returns data if an entry in the `gradeable_data` or the `electronic_gradeable_data` exist, you can't get a complete `DatabaseRowIterator` containing partially blank data.  This is critical for the views that show all students or the navigation view.  I'll make this in a separate PR since nothing actually calls these query methods yet.

PR #2290, opened 2018-06-28
Fix graders getting an error when trying to load the grading page

### LINKED ISSUE #2289: Error on grader trying to grade
TypeError (Code: 0) thrown in /usr/local/submitty/site/app/views/grading/ElectronicGraderView.php (Line 28) by:
    public function statusPage

Message:
Argument 4 passed to app\views\grading\ElectronicGraderView::statusPage() must be an instance of app\models\SimpleStat, array given

CouleeApps:
Closes #2289


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -169,8 +169,8 @@ public function showStatus() {
         $sections = array();
         $total_users = array();
         $component_averages = array();
-        $autograded_average = array();
-        $overall_average = array();
+        $autograded_average = null;
```
jchulton: Could a null autograded_average cause an error (calling a function on a null).
$autograded_average->getCount()

PR #2293, opened 2018-06-29
[WIP] Move repos under git folder in vcs

### LINKED ISSUE #1758: Documentation for git
Git allows you to specify a user/password in the checkout URL by doing '{username}:{password}@{git_checkout_path}' (putting 'http://' in front of the string if it's a URL). Additionally, we can add an ssh key for any user we want to hwcron so that we don't need to specify a full password (especially if using git). 

This information (and all information on VCS) should be documented under http://submitty.org/instructor/

MasterOdin:
Work necessary on main repo to close #1758


bmcutler:
Do we need a migration for this?  (so courses from the spring work?)

PR #2294, opened 2018-06-29
Improvements to the migrator tool

### LINKED ISSUE #2287: Course up/down migrations need semester & course as extra arguments

MasterOdin:
Closes #2287 

This PR does:
- Add semester and course arguments to course migrations
- Add (FAKE) text to the migrator output text to indicate migrations that were fake run
- Fix all existing course migrations to use the new function signature
- Add a system migration to migrate installed course migrations to use the new function signature

Doing a final round of vagrant destroy/up to test.


MasterOdin:
@bmcutler While I think the code should be largely done here, I'm currently doing a vagrant destory/up cycle to ensure that that works and we should definitely wait for Travis to finish on this as well to ensure nothing weird broke due to this change.

bmcutler:
Why not use 18.06.00 as our migrator version ;)

MasterOdin:
Eh, semver is more meaningful for small scripts like this won't see a ton of big changes on the API every week/month/year after we've gone through the initial teething pains of writing migrations within it. If anything, I probably should actually set the version to '0.8.0' or something to reflect its "beta" status and potential for API changes.

Anyway, this PR should be ready to go. Not sure why the PR build is failing on Travis but the push succeeded, but #2296 should fix the PR build once merged.

Issue #2246, opened 2018-06-25
Forum: Original author can edit a post


scopeInfinity:
Moved from #1677

- [ ] Maintain history of edits
- [ ] Allow author and instructor to edit post

Long Term Issue
- [ ] Edit only if the thread is not locked.

PR #2301, opened 2018-06-29
Forum: Allow original author to edit post/thread

scopeInfinity:
Closes: #2246

- [x] Maintain history of edits
- [x] Allow author and instructor to edit the post
- [x]  Markdown support(Code and URL block) in post history box
- [x] Show created timestamp and last edited timestamp for the post.


andrewaikens87:
- I really like the edit history popup, it's easy to read and see who posted the edits. On this popup though, the dates/times need to be formatted similar to how they are on the forum posts. Also it would be nice if we could have the same functionality of as we do on the forum posts in terms of how we display names (rather than just user ids).
- I noticed that if a TA/Instructor edits a post, it changes the author of that post to be the TA/Instructor who edited it. We want to preserve the original author of the post, but it might be nice to also display the person who last edited the post.

scopeInfinity:
Thankyou @andrewaikens87 
PR now preserves the original author and start pushing posts in _forum_posts_history_ only after an edit is made.
And history popup posts are now similar to the actual post in terms of display name and time formatting.

scopeInfinity:
@ropern123 Thank you for pointing out. I have fixed first and third point.
I will be working on second one.

andrewaikens87:
Couple things and it should be good:
- The user info in the view history popup has `Firstname Firstname` in the example below Joe's lastname is supposed to be Student
firstname
- This is optional but in the view history it might be nice to have the gold border around edits by TA/Instructors

scopeInfinity:
Thanks, @andrewaikens87 
I missed that. Fixed it.

Issue #2044, opened 2018-05-29
Non-grader can access marks via custom post requests


CouleeApps:
With a curl POST request to the get_mark_data endpoint they can see mark lists if they can guess component ids and gradeable ids. Many other controller functions have a similar lack of auth checks.

CouleeApps:
Merging into #2337

PR #2303, opened 2018-06-29
[WIP] Test+Fix grading permissions

CouleeApps:
This is planning to try all of the various grading pages / functions as every possible type of user. Also contains a fix for #2044.

Currently WIP because I haven't added all the functions, and also have to figure out how to inform people they should download selenium-requests. Or maybe even switch to a different requests library if that would work better.


MasterOdin:
My thoughts on this is that selenium is the worst possible way to go about testing this as it's slow and outside the purview of what selenium is about (hence the need for a third-party package). We should be just writing unit tests for the controllers (utilizing mock objects as necessary) that allows us to be far more precise on testing these conditions (and others) on failure or whatever for these endpoints.

Also, we're not supporting python 2 anymore, so just make sure it works for python 3.4

CouleeApps:
Fair enough. There are still a few minor changes I made on this branch that I think would be useful. I'll put those in a new PR.

PR #2320, opened 2018-07-02
Add better permissions checks to various grading actions

### LINKED ISSUE #2044: Non-grader can access marks via custom post requests

CouleeApps:
Closes #2044


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1384,6 +1388,45 @@ private function getStats($gradeable, &$sections, $graders=array(), $total_users
...
         }
     }
+
+    /**
+     * Check if the current user is allowed to grade the given gradeable/user pair
+     * @param string $gradeable_id
+     * @param string $user_id
+     * @return bool
+     */
+    public function canIGradeThis(string $gradeable_id, string $user_id): bool {
```
MasterOdin: Have this just accept a gradeable object. All instances of usage of this function, either the query it runs has been called before it or is one the next (or second next) line.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1384,6 +1388,45 @@ private function getStats($gradeable, &$sections, $graders=array(), $total_users
...
+                $sections = $this->core->getUser()->getGradingRegistrationSections();
+                $users_to_grade = $this->core->getQueries()->getUsersByRegistrationSections($sections);
+            } else {
+                $sections = $this->core->getQueries()->getRotatingSectionsForGradeableAndUser($gradeable_id, $this->core->getUser()->getId());
+                $users_to_grade = $this->core->getQueries()->getUsersByRotatingSections($sections);
+            }
+            $user_ids_to_grade = array_map(function(User $user) { return $user->getId(); }, $users_to_grade);
+            if (!in_array($user_id, $user_ids_to_grade)) {
+                return false;
+            }
```
MasterOdin: Just foreach through the array, as using array_map and then in_array has an O(n<sup>2</sup>) runtime if false and we can reduce it to just O(n).

```
foreach ($users_to_grade as $user) {
    if ($user->getId() === $user_id) {
        return true;
    }
}
return false;
```

CouleeApps:
This is getting merged into a bigger refactor as a part of #2337

PR #2321, opened 2018-07-02
[WIP] Added selenium test for late days

saxocellphone:
For setup_sample_course.py I added some special cases for when students have instructor granted extensions.

For the selenium test, I found test cases for most of the late day table status possibilities. (Good, good with an extension, late, late but has an extension, too many used in this assignment, too many used in this term).


saxocellphone:
Currently it's failing all the e2e tests on travis, but it's working fine on my machine. Shouldn't the sample data be the same because it's now deterministic? Or is it different because of different machines?

saxocellphone:
Turns out I messed up when I was trying to gather sample data points, so now all the tests are wrong. I need to gather some new data, but all the testing procedure should be right.

PR #2332, opened 2018-07-03
Finish up initial work for Ubuntu 18.04 (take 2)

MasterOdin:


KevinMackenzie:
I think this pip2 error is because we're ditching python2 for bionic, right?

KevinMackenzie:
I think this pip2 error is because we're ditching python2 for bionic, right?

PR #2341, opened 2018-07-04
Forum Thread Status

### LINKED ISSUE #2323: forum: open vs resolved status for threads
A "question" post should have a status "open" or "resolved" (design discussion needed...)

from #1827

scopeInfinity:
Closes #2323 
For now, PR includes three thread status
- Comment
- Unresolved  (Default Option)
- Resolved

Thread Status is initialized from Create Thread page, and updated from Edit Thread form.
Thread Status can be used for filter thread list in view thread page.

To be modified after #2301 : Allow the original author to edit thread status


andrewaikens87:
Some things I noticed after taking a look:

- As user `student` who created a thread I am unable to change the status of that thread. Seems like this privilege is only available to rank >= 2 (essentially those who can edit posts currently, which you have an open PR for so we might need to merge this after that one)
- I don't think this is related but I noticed that if you open up the edit popup and then close it without modifying anything and try to navigate to a different page it alerts the user that changes that were made may not be saved, but no changes were made. 
- When editing a post that isn't the thread post (meaning one that is a reply), the dropdown is displayed for selecting whether the thread is resolved or not. This should only be available on the thread post.  
- When creating a thread there are three options: `Default, Unresolved, Resolved` as a user I feel like this is might be sort of confusing. Does it make sense to be creating a thread that starts off resolved? Isn't that more of a comment, which could be put under the `Default` category? This is probably something that needs to be discussed.

scopeInfinity:
Thankyou @andrewaikens87 @ropern123 
PR is updated accordingly.

And bug(2) related to are-you-sure on popup form is already fixed in #2301 
Also for allowing the student to edit their own thread status, I'll update this PR after #2301.

bmcutler:
Hi @Gagan, can you please resolve conflicts?

Hi @andrewaikens87, please re-review if you have a chance.  Or open a new issue with requested changes if I merge it before you can review.

andrewaikens87:
Yes I will re-review once the merge conflicts are resolved.

PR #2343, opened 2018-07-04
[WIP] add lichen plagiarism color click events

### LINKED ISSUE #2329: Appropriate action on clicking colored code in plagiarism detector code boxes

tushargr:
Closes #2329


bmcutler:
Can we make the popup box that appears on the click smaller, and centered under the click?  More similar to the drop down box of student names & username that we have in other places on the site when you search by name/username.  So that you don't need to move your mouse very far after clicking to pick the user for the right panel.

Also, after selecting the 2nd person for the right panel, it should scroll the right panel to the matching bit of code.  It looks like it stays at the top of the right panel.

PR #2344, opened 2018-07-05
GradedGradeable query gets data for all teams/users requested

KevinMackenzie:
Now all teams / users requested will get data returned regardless of if any grade data exists for that submitter.  This is important for pages like the nav page, where gradeable instances for all gradeables for a given user are required.  Another important page is the grades list page, where all grades for a gradeable are required for all users, even if grading hasn't been done for them yet.


CouleeApps:
Is getGradedGradeables() with user ids on a team assignment supposed to return the gradeables for the teams those users are in? Because it is. Whether that is good or bad I don't know.

CouleeApps:
If you query user ids on a team assignment you only get one gradedgradeable per team. Seems like this should be either none (because you should query by team) or have multiple copies of the same gradedgradeable to line up with the users requested.

KevinMackenzie:
The intended behavior is for it to get the graded gradeables for the teams the users are on, so something like the nav page would get a grade for the team gradeables without having to get the teams the user is on for all gradeables

CouleeApps:
Opinion on the second comment?

KevinMackenzie:
I think it makes sense to only return one result per team, but I am hesitant to say that we won't want to return one instance per user.  

A bigger issue i'm seeing is that these changes don't return results for team gradeables where the requested user is not on a team.  I'm still thinking about how this case should be handled.

CouleeApps:
If you don't have a team, you can't submit, and you don't have a Submitter. So you can't have a graded gradeable. I think that's acceptable.

CouleeApps:
Yeah. I was trying to come up with the weirdest case to see what would happen. Rest of this looks pretty much good to me.

CouleeApps:
Sounds like you had one other thing you wanted to do, I'll review/approve after that or if you decide not to.

KevinMackenzie:
I think I've resolved/thought out all of the comments/issues I've had.

PR #2347, opened 2018-07-05
Saving GradedGradeables to the database

KevinMackenzie:
Database methods for creating/updating/deleting `TaGradedGradeable` instances, for creating/updating `GradedComponents` (but not deleting them), and for adding/removing assigned marks.

Along with these database methods, there are other functions critical for creating/updating manual grades.  The most complicated is the `TaGradedGradeable`s `getGradedComponent` function.  For a getter, this function has a lot of different behaviors, some of which add new data to the object, but its convenience is indispensable in integrating these models in the existing controllers.


KevinMackenzie:
For reviewing: if you want to test that these changes to the DB are being applied, drop this code in somewhere:
```
$gradeable = $this->core->getQueries()->getGradeableConfig('grading_homework');
$graded_gradeable = $this->core->getQueries()->getGradedGradeable($gradeable, 'aphacker', null);

// The default marks for this gradeable/user
$default = [
	'55' => [129, 130],
	'56' => [133],
	'57' => [137],
	'58' => [139]
];

// All marks for this gradeable/component
$all = [
	'55' => [128, 129, 130],
	'56' => [131, 132, 133, 134],
	'57' => [135, 136, 137, 138],
	'58' => [139, 140, 141]
];

$marks_to_add = [
	'55' => [128],
	'56' => [132, 134],
	'57' => [136],
	'58' => [140, 141]
];
$marks_to_remove = [
	'55' => [129],
	'56' => [133],
	'57' => [137],
	'58' => [139]
];

$ta_grade = $graded_gradeable->getTaGradedGradeable();
$ta_grade->setOverallComment('New Overall Comment');
$ta_grade->setUserViewedDate(new \DateTime());

// Adding a component (if manually deleted / missing)
$ta_grade->getGradedComponent($gradeable->getComponent(55), $this->core->getUser(), true);


// Modifying components
/** @var GradedComponent[] $graded_component */
foreach($ta_grade->getGradedComponents() as $graded_component) {
	foreach($graded_component as $component_grade) {
		$component_grade->setScore(-0.5);
		$component_grade->setComment('Woo');
		$component_grade->setGrader($this->core->getUser());
		$component_grade->setGradeTime(new \DateTime());

		$new_marks = array_merge($marks_to_add[$component_grade->getComponentId()],
					  array_diff($component_grade->getMarkIds(),
						  $marks_to_remove[$component_grade->getComponentId()]));
		$component_grade->setMarkIds($new_marks);

		// To set marks back to defaults per component
		//$component_grade->setMarkIds($default[$component_grade->getComponentId()]);
	}
}
$this->core->getQueries()->saveTaGradedGradeable($ta_grade);
return;
```

CouleeApps:
To be fair we should really make a set of PHPUnit test cases for this instead of copy+pasting code around. I guess that's for a later PR?

KevinMackenzie:
yea, that's probably fair

site/app/models/gradeable/TaGradedGradeable.php
```diff
@@ -91,6 +93,116 @@ public function toArray() {
...
+     *   Component not peer:
+     *     Component has grades:            => return the one TA grade
+     *     Generate false:                  => return null
+     *     Generate true:                   => throw InvalidArgumentException
+     *   Component peer:                    => throw InvalidArgumentException
+     * Grader Not Null:
+     *   Component not peer:
+     *     Component has grades:            => return the one TA grade and sets the grader
+     *     Generate false:                  => return null
+     *     Generate true:                   => return new component with provided user as grader (TA)
```
CouleeApps: All of this creation seems like it would work better in its own function. I wouldn't expect a getter to create and set values. I would assume it returns null and I have to create the GradedComponent myself.

site/app/models/gradeable/GradedGradeable.php
```diff
@@ -69,6 +68,19 @@ public function setTaGradedGradeable(TaGradedGradeable $ta_graded_gradeable) {
...
+    /**
+     * Gets the TaGradedGradeable for this graded gradeable, or generates a blank
+     *  one if none exists and $generate is true
+     * @param bool $generate If a new TaGradedGradeable should be generated if one doesn't exist already
+     * @return TaGradedGradeable
+     */
+    public function getTaGradedGradeable($generate = false) {
+        if ($this->ta_graded_gradeable === null && $generate === true) {
+            $this->ta_graded_gradeable = new TaGradedGradeable($this->core, $this, []);
+        }
```
CouleeApps: Another getter that is doubling as a setter. I would just have a `createTaGradedGradeable` function if you really don't want anyone else touching these classes.

site/app/models/gradeable/GradedComponent.php
```diff
@@ -132,18 +133,33 @@ private function setComponent(Component $component) {
...
+        $this->mark_ids = $actual_ids;
+        $this->marks_modified = true;
+    }
+
+    /**
+     * Called from the db methods to load the mark ids the submitter received
+     * @param int[] $db_mark_ids
+     * @internal
+     */
+    public function setMarkIdsFromDb(array $db_mark_ids) {
```
CouleeApps: `setDbMarkIdsFromDatabase` to be consistent with `setIdFromDatabase`

KevinMackenzie: This sets both, I think adding `db` prefix would imply that it doesn't also set the `mark_ids`

CouleeApps: To be fair the whole db_mark_ids business seems like a messy way of dealing with the lack of a `GradedMark::$modified` type deal for knowing which haven't been saved

CouleeApps: Not sure if I can come up with a better solution than array_diff though

site/app/models/gradeable/TaGradedGradeable.php
KevinMackenzie: The purpose of these kinds of getters is so the caller doesn't always have to be doing the "if exists, then create" check (see 1st PR comment).  In this particular case, when you're changing the grade for the assignment, there will never be a case where you want to not create the grade if it doesn't exist.

CouleeApps: While true, this gets rather confusing. Even in your test code, you call a getter method to create (set) a graded component and it's pretty counterintuitive. The lines of code you save probably aren't worth the confusion of a getter that is also a setter. In my opinion you should really make getters non-modifying and have createOrGet functions if the convenience is really necessary.

KevinMackenzie: I agree.  I'll give it a name change (and the other one below)

site/app/models/gradeable/GradedComponent.php
```diff
@@ -51,35 +60,27 @@ class GradedComponent extends AbstractModel {
      * @param Core $core
      * @param Component $component The component this grade is associated with
      * @param User $grader The user who graded this component
-     * @param int[] $mark_ids The mark ids this graded component received
      * @param array $details any remaining properties
      * @throws \InvalidArgumentException if any of the details are invalid, or the component/grader are null
      */
-    public function __construct(Core $core, Component $component, User $grader, array $mark_ids, array $details) {
+    public function __construct(Core $core, Component $component, User $grader, array $details) {
```
CouleeApps: Would be nice to have TaGradedGradeable in here as well so you don't have to pass around both in the db methods.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2909,4 +2910,226 @@ public function updateGradeableNew(\app\models\gradeable\Gradeable $gradeable) {
...
+        } else {
+            $this->updateTaGradedGradeable($ta_graded_gradeable);
+        }
+    }
+
+    /**
+     * Deletes an entry from the gradeable_data table with the provided data id
+     * @param int $ta_graded_gradeable_id
+     */
+    public function deleteTaGradedGradeable($ta_graded_gradeable_id) {
```
CouleeApps: Any reason this takes an id instead of the object itself?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2909,4 +2910,226 @@ public function updateGradeableNew(\app\models\gradeable\Gradeable $gradeable) {
...
+                        $ta_graded_gradeable,
+                        $component_grade,
+                        $deleted_marks);
+                    $this->createGradedComponentMarks(
+                        $ta_graded_gradeable,
+                        $component_grade,
+                        $new_marks);
+                }
+            }
+        }
```
CouleeApps: UpdateGradedComponents doesn't delete removed components from the database, and there doesn't seem to be a method for that.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2909,4 +2910,226 @@ public function updateGradeableNew(\app\models\gradeable\Gradeable $gradeable) {
...
+            $this->updateTaGradedGradeable($ta_graded_gradeable);
+        }
+    }
+
+    /**
+     * Deletes an entry from the gradeable_data table with the provided data id
+     * @param int $ta_graded_gradeable_id
+     */
+    public function deleteTaGradedGradeable($ta_graded_gradeable_id) {
+        $this->course_db->query("DELETE FROM gradeable_data WHERE gd_id=?", [$ta_graded_gradeable_id]);
```
CouleeApps: Feels like there should be a `DELETE FROM gradeable_component_data` and a `DELETE FROM gradeable_component_mark_data` in here as well

site/app/libraries/database/DatabaseQueries.php
KevinMackenzie: I couldn't think of a use case where you'd want to delete the components without deleting the whole TA grade, but i'll add the feature for completeness, since I'm probably not thinking about everything.

site/app/libraries/database/DatabaseQueries.php
KevinMackenzie: Both of those tables have `ON DELETE CASCADE` on their foreign keys to `gradeable_data`, so I think we're ok

site/app/libraries/database/DatabaseQueries.php
CouleeApps: If you ungrade only one component it removes the GradedComponent (at least visually) but there can still be other GradedComponents

KevinMackenzie: This got marked as outdated, but its still relevant.  I'll add this in.

CouleeApps:
Although there's a merge conflict it's waiting on

Issue #2304, opened 2018-06-29
Add enumeration for user permission levels


KevinMackenzie:
User permission checking is always done using hard-coded numbers whenever the comparison is made, but it would be much more clear if these permission levels were an enumeration.  This could also prevent permission bugs from arising.  As it stands, 

0. Developer (rarely/never used)
1. Instructor
2. TA
3. Grader
4. Student

MasterOdin:
Developer no longer is a class in the system

PR #2349, opened 2018-07-06
Access functions for controllers

### LINKED ISSUE #2337: Access functions refactor
`canIViewThis` / `canIGradeThis` / `checkValidAccess` need to make sure we don't duplicate code

CouleeApps:
Closes #2337 but only for the grading controller-- I'll make an issue for the rest of the stuff (for late July). Also addresses #2304 but need to go through the codebase to clean up the references.


site/app/libraries/Access.php
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        if ($checks & self::CHECK_CSRF) {
+            if ($this->core->checkCsrfToken($_POST['csrf_token'])) {
+                return false;
+            }
+        }
+
+        /* @var Gradeable|null $gradeable */
+        $gradeable = $args["gradeable"] ?? null;
```
MasterOdin: This function is to be only used with gradeable endpoints?

site/app/libraries/Access.php
```diff
@@ -0,0 +1,170 @@
...
+    const ALLOW_STUDENT                 = 1 << 3;
+    const ALLOW_LOGGED_OUT              = 1 << 4;
+    const CHECK_GRADEABLE_MIN_GROUP     = 1 << 5;
+    const CHECK_GRADING_SECTION_GRADER  = 1 << 6;
+    const CHECK_PEER_ASSIGNMENT_STUDENT = 1 << 7;
+    const CHECK_HAS_SUBMISSION          = 1 << 8;
+    const CHECK_CSRF                    = 1 << 9;
+
+    //
+    const ALLOW_MIN_STUDENT    = self::ALLOW_INSTRUCTOR | self::ALLOW_TA | self::ALLOW_GRADER | self::ALLOW_STUDENT;
```
MasterOdin: What's the point of this check?

site/app/libraries/Access.php
CouleeApps: Only gradeable-related checks will need it. The alternative is copying that line into every gradeable-related test, which I'm completely fine with.

site/app/libraries/Access.php
CouleeApps: Mostly consistency with the other access levels so it's obvious what access level is required. Also future-proofing in case we ever decide to add more user groups or do something different.

site/app/libraries/Access.php
MasterOdin: You still have a potential for throwing a null error for the grading tests. But the intention is to expand this beyond just testing of the gradeable object?

CouleeApps:
Should be ready for review.

site/app/libraries/Access.php
```diff
@@ -0,0 +1,248 @@
...
+            if (!$this->checkGradingSection($gradeable)) {
+                return false;
+            }
+        }
+        if ($group === User::GROUP_STUDENT && ($checks & self::CHECK_PEER_ASSIGNMENT_STUDENT)) {
+            /* @var Gradeable|null $gradeable */
+            $gradeable = $this->requireArg($args, "gradeable");
+            if ($gradeable === null) {
+                return false;
+            }
```
MasterOdin: This doesn't really feel like a great paradigm as if we wanted to add a new arg, we're just copying these four lines of boilerplate along and just continuously adding to this one function (which is already almost ~90 lines long). I guess you could do a check of `if ($checks & self:: CHECK_GRADEABLE_MIN_GROUP || $checks & self:: CHECK_HAS_SUBMISSION || ...)` and if so, enter a block that then checks, do we have a gradeable arg, if not, return false. You could do something with a bit flag, to also tell us to go into a "gradeable" specific check function.

site/app/libraries/Access.php
```diff
@@ -0,0 +1,248 @@
...
+        if (!$gradeable->getPeerGrading()) {
+            return false;
+        } else {
+            $user_ids_to_grade = $this->core->getQueries()->getPeerAssignment($gradeable->getId(), $this->core->getUser()->getId());
+            return in_array($gradeable->getUser()->getId(), $user_ids_to_grade);
+        }
+    }
+
+    /**
+     * Check that a a user group has privilege at least equal to minimum
```
MasterOdin: `/a a/a/s`

site/app/libraries/Access.php
```diff
@@ -0,0 +1,248 @@
...
+        $this->permissions["grading.get_gradeable_comment"] = self::ALLOW_MIN_STUDENT | self::CHECK_GRADEABLE_MIN_GROUP | self::CHECK_GRADING_SECTION_GRADER | self::CHECK_PEER_ASSIGNMENT_STUDENT;
+        $this->permissions["grading.add_one_new_mark"] = self::ALLOW_MIN_LIMITED_ACCESS_GRADER | self::CHECK_GRADEABLE_MIN_GROUP | self::CHECK_GRADING_SECTION_GRADER;
+        $this->permissions["grading.delete_one_mark"] = self::ALLOW_MIN_LIMITED_ACCESS_GRADER | self::CHECK_GRADEABLE_MIN_GROUP | self::CHECK_GRADING_SECTION_GRADER;
+        $this->permissions["grading.get_marked_users"] = self::ALLOW_MIN_LIMITED_ACCESS_GRADER | self::CHECK_GRADEABLE_MIN_GROUP;
+        $this->permissions["grading.get_marked_users.full_stats"] = self::ALLOW_MIN_FULL_ACCESS_GRADER;
+        $this->permissions["grading.show_edit_teams"] = self::ALLOW_MIN_INSTRUCTOR;
+        $this->permissions["grading.import_teams"] = self::ALLOW_MIN_INSTRUCTOR | self::CHECK_CSRF;
+        $this->permissions["grading.export_teams"] = self::ALLOW_MIN_INSTRUCTOR;
+        $this->permissions["grading.submit_team_form"] = self::ALLOW_MIN_INSTRUCTOR;
```
MasterOdin: We should condense keys that have the same permission masks unless we absolutely foresee us changing the permission in the very near future (and so why not just do that now?). `save_one_component` and `save_general_comment`, `get_mark_data` and `get_gradeable_comment`, `show_edit_teams` and `export_teams` and `submit_team_form`, `status` and `details`, etc.

Additionally, I foresee a possibility of a time where CSRF token validation would happen automatically which would allow for a bit more consolidation of these.

site/app/libraries/Access.php
CouleeApps: Yeah I was struggling with how to both require arguments for specific checks but then ignore them for others. A generic bit flag sounds like a pretty good way to go.

site/app/libraries/Access.php
CouleeApps: For save/get component vs general, I've added in another check to test for peer graded components, so those aren't combinable anymore. I would combine the team edit permissions except I'm not sure how to handle CSRF.

bmcutler:
@CouleeApps  please resolve conflicts
@MasterOdin  do you approve?

Issue #2317, opened 2018-07-02
New Gradeable Model Integration


KevinMackenzie:
Once #2107 gets finished (gradeable model classes), they will need to be integrated into the rest of the codebase.  The most significant parts are as follows (update as needed):
- [x] Add per-testcase results into `AutoGradedGradeable` (#2379)
- [x] Navigation View (#2407)
- [x] Create/Edit Gradeable (#2356)
- [x] Electronic Grader (#2970)
- [x] Simple Grader (#2406)
- [x] Submission page (first part: #2432)
- [x] Team Controller (#2985)
- [x] Reports Controller (#3174)
- [x] Late Days Calculation (#3043)
- [x] Gradeable Controller (#2986)
- [ ] Delete old model

There will be many other places that will need to be updated, but they shouldn't be as involved.

PR #2356, opened 2018-07-09
Create/Edit Gradeable new gradeable model integration

KevinMackenzie:
(Part of #2317) The create/edit gradeable interface now exclusively uses the new gradeable model.  In this process i also removed the `AdminGradeable` model, which was only used as a vehicle for the parameters passed to the twig templates.  I also removed the `AdminGradeableView` class, which was doing a lot of controller work (fetching and constructing data).  This PR should not change any of the outward behavior of this page.

@saxocellphone Can you check to make sure the the autograding config changes you made are still completely functional.  I made some tweaks to it so it would play more nicely with the autosaving system.


saxocellphone:
When I create a gradeable based on a current gradeable, and exit without making any changes (except for name and id), and try to grade a person, it gives this error

Notice: Undefined offset: 0 in /usr/local/submitty/site/app/views/grading/ElectronicGraderView.php on line 664
--
I took a look at that line, and it seems like it can't find the components. I tried it again on masters and it doesn't happen there.

KevinMackenzie:
This is a semi-known issue where trying to grade a 'no submission' student on a gradeable with student-assigned pdf pages has issues.  The real issue is that I was setting the default pdf page to the wrong value

saxocellphone:
The config uploads stuff works as intended

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -285,7 +285,7 @@ private function newComponent(Gradeable $gradeable) {
             'text' => false,
             'peer' => false,
             'order' => -1,
-            'page' => -1
+            'page' => 0 // 0 means no page, -1 is student-assigned
```
CouleeApps: Use an enum? Or at least constants

PR #2365, opened 2018-07-10
Fixed Configure script to allow switching between worker and primary mode

emaicus:
Fixed a problem which caused CONFIGURE to fail when switching from worker to primary mode.

### DIFF
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -146,7 +146,10 @@ if os.path.isfile(CONFIGURATION_JSON):
         loaded_defaults = json.load(conf_file)
     #no need to authenticate on a worker machine (no website)
     if not args.worker:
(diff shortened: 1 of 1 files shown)


.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -146,7 +146,10 @@ def get_input(question, default=""):
         loaded_defaults = json.load(conf_file)
     #no need to authenticate on a worker machine (no website)
     if not args.worker:
-        loaded_defaults['authentication_method'] = 1 if loaded_defaults['authentication_method'] == 'PamAuthentication' else 2
+        if 'authentication_method' in loaded_defaults:
+            loaded_defaults['authentication_method'] = 1 if loaded_defaults['authentication_method'] == 'PamAuthentication' else 2
+        else:
+            loaded_defaults['authentication_method'] = 2
```
MasterOdin: The else branch shouldn't exist. The "default" value of the fields is handled by the "defaults" dictionary above the outer outer if block and this is just going to be confusing.

PR #2372, opened 2018-07-10
Site theme

### LINKED ISSUE #1922: Allow instructor to specify "wrapper" for Submitty website
Currently in a couple courses we're facilitating nesting submitty inside of an iframe.  The different top level domains is a headache.   CORS I think is the acronym.  the 2 top level domains requires double login to both top domains, and logout must similarly be from both domains and is confusing.  Want to avoid this!

Instead perhaps we allow the instructor to upload a 'wrapper' html that for example pastes a sidebar of links onto the submitty webpage and that side bar hosts the non submitty content for the course.  This allows the instructor to visually match non submitty pages with submitty pages making it look like a single website for the course...

ropern123:
Allow instructors to upload course-specific html to use as a wrapper for the site.

Early stages.

will close #1922


bmcutler:
@ropern123 please add basic documentation with where these files should be placed, and add samples of these files to the sample_files directory.  Thanks

CouleeApps:
I think this is basically done. @bmcutler mind checking out the branch and see if the sample files work for you?

PR #2373, opened 2018-07-10
Sample grader sections

### LINKED ISSUE #2302: Sample data: Grader has no grading sections
Makes it hard to test grading as a grader if the grader has no grading sections and nothing to grade.

CouleeApps:
The sample `grader` user now has two grading sections. Also I've added another grader, `grader2` who has nobody to grade so we can test that as well.

Also added bonus is a recreate_sample_course.py that runs partial_reset and setup_sample_course automatically, for convenience.

Closes #2302


bmcutler:
@holzbh we've got a new user now, not sure if it impacts rainbow grades

PR #2378, opened 2018-07-11
Added sorting to Gradeable and GradedGradeable queries

KevinMackenzie:
It is desirable to sort `GradedGradeables` by different criteria, such as registration section, user id, etc.  Now it is possible to sort by any number of criteria in the `GradedGradeable` query exceptional cases like `registration_section` being alphanumeric instead of numeric are handled internally.

It was trivial to also add this feature to the `Gradeable` query, so that method also can be sorted by any of the gradeable's properties.

Due to the way the `ORDER BY` clause gets generated, it is straight forward to add new exceptions or aliases to sort by so the queries themselves don't need to be modified frequently.

Lastly, I updated both queries to use upper-case for all sql keywords


site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -939,17 +939,67 @@ public function getTeamsById(array $team_ids) {
...
+     */
+    private static function generateOrderByClause(array $sort_keys, array $key_map) {
+        if ($sort_keys !== null) {
+            if (!is_array($sort_keys)) {
+                $sort_keys = [$sort_keys];
+            }
+            // Use simplified expression for empty keymap
+            if (empty($key_map)) {
+                return 'ORDER BY ' . implode(',', $sort_keys);
+            }
```
CouleeApps: I'd also add a check for `empty($sort_keys)`, just in case

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -939,17 +939,67 @@ public function getTeamsById(array $team_ids) {
...
+            return 'ORDER BY ' . implode(',', array_map(function ($key_ext) use ($key_map) {
+                    $split_key = explode(' ', $key_ext);
+                    $key = $split_key[0];
+                    $order = '';
+                    if (count($split_key) > 1) {
+                        $order = $split_key[1];
+                    }
```
CouleeApps: Otherwise specify default order? Or is ORDER BY with no order still allowed?

site/app/libraries/database/PostgresqlDatabaseQueries.php
KevinMackenzie: done

site/app/libraries/database/PostgresqlDatabaseQueries.php
KevinMackenzie: `ORDER BY` defaults to `ASC` if none specified and I think we should preserve that behavior

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -1099,11 +1152,11 @@ public function getGradedGradeables(array $gradeables, $users = null, $teams = n
...
                 LEFT JOIN (
                   SELECT
-                    json_agg(sections_registration_id) as grading_registration_sections,
+                    json_agg(sections_registration_id) AS grading_registration_sections,
                     user_id
                   FROM grading_registration
                   GROUP BY user_id
-                ) as sr ON u.user_id=sr.user_id
+                ) AS sr ON u.user_id=sr.user_id
               ) AS u ON eg IS NULL OR NOT eg.eg_team_assignment
```
CouleeApps: `eg.eg_team_assignment` was renamed to `team_assignment` above

site/app/libraries/database/PostgresqlDatabaseQueries.php
CouleeApps: 👍

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -1118,9 +1174,9 @@ public function getGradedGradeables(array $gradeables, $users = null, $teams = n
...
                       in2_team.team_id,
-                      json_agg(in2_team.user_id) as array_user,
-                      json_agg(in2_team.state) as array_state
-                    FROM teams as in2_team
+                      json_agg(in2_team.user_id) AS array_user,
+                      json_agg(in2_team.state) AS array_state
+                    FROM teams AS in2_team
                     GROUP BY in2_team.team_id
                   ) AS in_team ON in_team.team_id=g_team.team_id
               ) AS team ON eg.eg_team_assignment AND EXISTS (
```
CouleeApps: Also here

PR #2379, opened 2018-07-11
Per-testcase results in GradedGradeable and global GradingQueue

KevinMackenzie:
Added per-testcase results into `AutoGradedVersion` that are lazy-loaded from the result files on the disk.

Additionally, The code for getting queue status has been factored out into a `GradingQueue` class that has an instance in `core` mostly for code organization, but it also can reduce calls to the disk and ensure consistency of position results.


site/app/models/gradeable/AutoGradedVersion.php
```diff
@@ -0,0 +1,439 @@
...
+    /** @property @var float The number of "normal extra credit" points */
+    protected $non_hidden_extra_credit = 0;
+    /** @property @var float The number of "hidden" points */
+    protected $hidden_non_extra_credit = 0;
+    /** @property @var float The number of "hidden extra credit" points */
+    protected $hidden_extra_credit = 0;
+    /** @property @var \Datetime Time the user submitted this version */
+    protected $submission_time = null;
+    /** @property @var bool If the autograding has complete for this version */
+    protected $autograding_complete = false;
```
CouleeApps: If you're going to do `AutoGradedVersion` you should probably do `$auto_grading_complete`

KevinMackenzie: Its kind of a mess, but `Autograding` is long-form.  `AutoGraded` is shorthand for `AutogradingGraded`

CouleeApps: Except "Autograded" is also past tense of "to Autograde" and is spelled with both capses all over the place... AuGhh

PR #2380, opened 2018-07-11
Misc helper methods in new Gradeable model

KevinMackenzie:
This changes are unrelated to any of the other PR's, so I'm putting them in their own, small PR...

I added helper methods in the new `Gradeable` class for getting the number of text/numeric components for `NUMERIC_TEXT` gradeables so the user doesn't need to count them every time.  Additionally, I added helper methods to check if the gradeable is (student) pdf upload and integrated them into the Admin Gradeable controller

Lastly, The Gradeable class lazy loads its teams instead of just its `anyTeams` status, which will be useful in the `ElectronicGraderController`.

### DIFF
--- a/site/app/models/gradeable/Gradeable.php
+++ b/site/app/models/gradeable/Gradeable.php
@@ -785,4 +785,64 @@ class Gradeable extends AbstractModel {
     public function isRotatingGraderSectionsModified() {
         return $this->rotating_grader_sections_modified;
     }
+
+    /**
+     * Gets if this gradeable is pdf-upload
+     * @return bool
+     */
+    public function isPdfUpload() {
+        foreach ($this->components as $component) {
+            if ($component->getPage() !== 0) {
+                return true;
+            }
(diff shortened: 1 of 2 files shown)


site/app/models/gradeable/Gradeable.php
```diff
@@ -92,8 +92,8 @@ class Gradeable extends AbstractModel {
     private $any_manual_grades = null;
     /** @property @var bool If any submissions exist */
     private $any_submissions = null;
-    /** @property @var bool If any teams have been formed */
-    private $any_teams = null;
+    /** @property @var Team[] Any teams that have been formed */
+    private $teams = null;
```
CouleeApps: So _that's_ how you hide a variable from `AbstractModel::__call`

KevinMackenzie: if you don't have the `@property`  tag on it i've had it not work, but I like to be extra sure

Issue #2381, opened 2018-07-11
Create New Docker Containers Per Testcase


emaicus:
To Do
- [X]  Properly copy submission files, etc into test## folders.
- [X]  Properly my_runner into test## folders.
- [X]  Grade using Docker.
- [x] Work to get the solution working with compiled files rather than just python (e.g. C++ code).

PR #2383, opened 2018-07-11
Moves Iteration over Testcases from C++ to Python

emaicus:
Depends on (and contains) PR #2369
Begins work on issue #2381

Moves iteration over autograding testcases out of C++ and into Python. Namely, iteration over testcases has now been moved from main_runner.cpp to grade_items_main_runner.py. This will facilitate creating a new docker container or network per testcase (coming in a future PR). main_runner.cpp now takes in a sixth command line parameter which indicates which testcase it is to execute. If the parameter is not passed, main_runner.cpp executes all testcases.


grading/main_runner.cpp
```diff
@@ -36,31 +41,52 @@ int main(int argc, char *argv[]) {
...
 
+    TestCase my_testcase(config_json,i);
+
+    if (my_testcase.isFileCheck() || my_testcase.isCompilation()){
+      continue;
+    }
+
+    if(test_case_to_run != -1 &&  test_case_to_run != i){
+      continue;
+    }
```
MasterOdin: This continue should be above the instantiation of the my_testcase object.

emaicus: This is a necessary workaround. When test cases are instantiated, their “number” is set via a static global int, so the first one created is 0, then 1, 2, 3, etc. To avoid tearing apart the test case class, we decided to just instantiate unused test cases for now.

sbin/autograder/grade_item_main_runner.py
```diff
@@ -0,0 +1,79 @@
...
+    OPEN_JSON = json.load(open_file)
+SUBMITTY_INSTALL_DIR = OPEN_JSON['submitty_install_dir']
+SUBMITTY_DATA_DIR = OPEN_JSON['submitty_data_dir']
+
+with open(os.path.join(CONFIG_PATH, 'submitty_users.json')) as open_file:
+    OPEN_JSON = json.load(open_file)
+DAEMON_UID = OPEN_JSON['daemon_uid']
+
+
+def executeTestcases(complete_config_obj, tmp_logs, tmp_work, queue_obj, submission_string, item_name, USE_DOCKER, container, which_untrusted):
```
MasterOdin: Talked about this briefly yesterday, but there's no real point in breaking this function out into its own module and should probably just be left in grade_item.py which is a module centered around grading an item (which this function is a part of doing).

PR #2390, opened 2018-07-12
Forum save page state

### LINKED ISSUE #2268: Forum: Page should have state
- [x] state of category selection #1971(Taken from #2201)
- [x] state of chronological/alphabetical selection

andrewaikens87:
Resolves #2268. 

- Saves category selection and display option states
- Includes python script in python_submitty_utils that will insert specified # of threads and posts for a given database (data generated in tables after execution can be dumped to a sql file which might be quicker to source rather than rerunning).


site/app/views/forum/ForumThreadView.php
```diff
@@ -264,7 +264,23 @@ function changeName(element, user, visible_username, anon){
 					<option value="{$categories[$i]['category_id']}" style="color: {$categories[$i]['color']}">{$categories[$i]['category_desc']}</option>
 HTML;
 			}
-
+	$cookieSelectedCategories = '';
+	$category_ids_array = array_column($categories, 'category_id');
+	foreach(explode('|', $_COOKIE['forum_categories']) as $selectedId) {
```
scopeInfinity: Need additional `!empty($_COOKIE['forum_categories'])` in case cookie doesn't exists.

site/app/views/forum/ForumThreadView.php
```diff
@@ -283,6 +299,8 @@ function changeName(element, user, visible_username, anon){
 							{$onChange}
```
scopeInfinity: If after using filter no threads are left, then if we reload page(keeping same filter).
`$onChange` remains empty string. Thus, we are not able to change filter.

```
$onChange = '';
if($thread_count > 0) {
 		$onChange = <<<HTML
 		modifyThreadList({$currentThread}, '{$currentCategoriesIds_string}');
HTML;
}
```

python_submitty_utils/input_forum_data.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
```
MasterOdin: this file should be in .setup/bin.

site/public/js/server.js
```diff
@@ -1240,6 +1241,7 @@ function alterShowDeletedStatus(newStatus) {
 function modifyThreadList(currentThreadId, currentCategoriesId){
     var categories_value = $("#thread_category").val();
     categories_value = (categories_value == null)?"":categories_value.join("|");
+    document.cookie = "forum_categories=" + categories_value + ";";
```
scopeInfinity: Just a note, categories filter are independent of course.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -275,6 +281,10 @@ public function searchThreads($searchQuery){
     	return $this->course_db->rows();
     }
 
+    public function threadExists(){
```
scopeInfinity: I am not sure about this one.
Should we consider deleted thread or not, while checking threadExists.

Rest looks good to me.

Issue #2272, opened 2018-06-27
Move math/logic from views into model


CouleeApps:
A lot of views are doing quite a bit of math and logic that the model should really take care of. Here are a few specific examples:

- [ ] Late days remaining calculations, both overall and per-gradeable
- [ ] Total point values especially with extra credit / hidden points
- [ ] URL locations (should be in controller)
- [x] Page calculation for components, reading from json files
- [x] Grading progress calculation

PR #2397, opened 2018-07-12
Move logic into model

CouleeApps:
Part 1 of probably 2 on #2272

- Add GradingSection model to clean up section checks
- Moving nav page progress bar code into the model (and make it work for teams)
- Add list of `User`s to `Team`s
- Move team history loading into model
- Move page number calculation into model

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -1614,19 +1614,7 @@ WHERE gcm_id=?", $params);
      * @return \app\models\Team
      */
     public function getTeamById($team_id) {
-        $this->course_db->query("
-          SELECT team_id, registration_section, rotating_section
-          FROM gradeable_teams
-          WHERE team_id=?",
-            array($team_id));
-        if (count($this->course_db->rows()) === 0) {
-            return null;
-        }
-        $details = $this->course_db->row();
-
-        $this->course_db->query("SELECT user_id, state FROM teams WHERE team_id=? ORDER BY user_id", array($team_id));
-        $details['users'] = $this->course_db->rows();
-        return new Team($this->core, $details);
+        throw new NotImplementedException();
     }
 
     /**
@@ -1636,22 +1624,7 @@ WHERE gcm_id=?", $params);
      * @return \app\models\Team
      */
     public function getTeamByGradeableAndUser($g_id, $user_id) {
-        $this->course_db->query("
-          SELECT team_id, registration_section, rotating_section
-          FROM gradeable_teams
-          WHERE g_id=? AND team_id IN (
-            SELECT team_id
-            FROM teams
-            WHERE user_id=? AND state=1)",
-            array($g_id, $user_id));
-        if (count($this->course_db->rows()) === 0) {
-            return null;
-        }
-        $details = $this->course_db->row();
-
-        $this->course_db->query("SELECT user_id, state FROM teams WHERE team_id=? ORDER BY user_id", array($details['team_id']));
-        $details['users'] = $this->course_db->rows();
-        return new Team($this->core, $details);
+        throw new NotImplementedException();
     }
 
     /**
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -676,10 +676,24 @@ ORDER BY {$section_key}", $params);
         return $return;
     }
 
+    /**
+     * Get an array of Teams for a Gradeable matching the given registration sections
+     * @param string $g_id
+     * @param array $sections
+     * @param string $orderBy
+     * @return Team[]
+     */
     public function getTeamsByGradeableAndRegistrationSections($g_id, $sections, $orderBy="registration_section") {
         throw new NotImplementedException();
     }
 
+    /**
+     * Get an array of Teams for a Gradeable matching the given rotating sections
+     * @param string $g_id
+     * @param array $sections
+     * @param string $orderBy
+     * @return Team[]
+     */
     public function getTeamsByGradeableAndRotatingSections($g_id, $sections, $orderBy="rotating_section") {
         throw new NotImplementedException();
     }
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -616,17 +616,14 @@ ORDER BY {$section_key}", $params);
 
     public function getTotalSubmittedUserCountByGradingSections($g_id, $sections, $section_key) {
         $return = array();
-        $params = array();
+        $params = array($g_id);
         $where = "";
         if (count($sections) > 0) {
             // Expand out where clause
             $sections_keys = array_values($sections);
-            $where = "WHERE {$section_key} IN (";
-            foreach($sections_keys as $section) {
-                $where .= "?" . ($section != $sections_keys[count($sections_keys)-1] ? "," : "");
-                array_push($params, $section);
-            }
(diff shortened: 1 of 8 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -648,6 +648,48 @@ public function getTotalSubmittedUserCountByGradingSections($g_id, $sections, $s
...
+        $params = array($g_id);
+        $where = "";
+        if (count($sections) > 0) {
+            // Expand out where clause
+            $sections_keys = array_values($sections);
+            $where = "WHERE {$section_key} IN (";
+            foreach($sections_keys as $section) {
+                $where .= "?" . ($section != $sections_keys[count($sections_keys)-1] ? "," : "");
+                array_push($params, $section);
+            }
```
KevinMackenzie: Seems like generating the placeholders is done everywhere else like:
`implode(",", array_fill(0, count($sections), "?"))`
and using `array_merge` on the `params` and `section_keys` will preserve order

site/app/models/Gradeable.php
```diff
@@ -1339,4 +1339,181 @@ public function calculateLateStatus($checkTeam = true, $user_id = null) {
         }
     }
 
+    public function getGradingSectionsForUser(User $user) {
+        if ($this->getPeerGrading() && $user->getGroup() === 4) {
+            $users = $this->core->getQueries()->getPeerAssignment($this->getId(), $user->getId());
+            //TODO: Peer grading team assignments
+            return new GradingSection($this->core, $this, "Peer", [$user], $users, []);
```
KevinMackenzie: This function call also needs a `$teams` parameter

site/app/models/Gradeable.php
```diff
@@ -1339,4 +1339,181 @@ public function calculateLateStatus($checkTeam = true, $user_id = null) {
...
+                $section_names = $this->core->getQueries()->getRotatingSectionsForGradeableAndUser($this->getId(), $user->getId());
+
+                if ($this->isTeamAssignment()) {
+                    foreach ($section_names as $section) {
+                        $teams[$section] = [];
+                    }
+                    $all_teams = $this->core->getQueries()->getTeamsByGradeableAndRotatingSections($this->getId(), $section_names);
+                    foreach ($all_teams as $team) {
+                        /** @var Team $team */
+                        $teams[$team->getRotatingSection()][] = $team;
```
KevinMackenzie: Could avoid this loop if you juts made the return value of `getTeamsByGradeableAndRotatingSections` return the array indexed like this?

site/app/models/Gradeable.php
```diff
@@ -1339,4 +1339,181 @@ public function calculateLateStatus($checkTeam = true, $user_id = null) {
...
+                $section_names = $user->getGradingRegistrationSections();
+
+                if ($this->isTeamAssignment()) {
+                    foreach ($section_names as $section) {
+                        $teams[$section] = [];
+                    }
+                    $all_teams = $this->core->getQueries()->getTeamsByGradeableAndRegistrationSections($this->getId(), $section_names);
+                    foreach ($all_teams as $team) {
+                        /** @var Team $team */
+                        $teams[$team->getRegistrationSection()][] = $team;
```
KevinMackenzie: Could avoid this loop if you juts made the return value of `getTeamsByGradeableAndRegistrationSections` return the array indexed like this?

site/app/models/Gradeable.php
```diff
@@ -1339,4 +1339,181 @@ public function calculateLateStatus($checkTeam = true, $user_id = null) {
...
+            foreach ($section_names as $section_name) {
+                $sections[] = new GradingSection($this->core, $this, $this->isGradeByRegistration(), $section_name, $graders[$section_name], $users[$section_name] ?? null, $teams[$section_name] ?? null);
+            }
+
+            return $sections;
+        }
+    }
+
+    /**
+     * @param Gradeable $gradeable
```
KevinMackenzie: Is this supposed to be here?

site/app/models/GradingSection.php
```diff
@@ -0,0 +1,82 @@
...
+class GradingSection extends AbstractModel {
+    /**
+     * @var Gradeable
+     */
+    protected $gradeable;
+    /**
+     * If this is a registration section (false for rotating)
+     * @var bool
+     */
+    protected $registration;
```
KevinMackenzie: Add a `isRegistration` getter in the `@method`s in the class doc

site/app/models/GradingSection.php
```diff
@@ -0,0 +1,82 @@
...
+ * @method string getName()
+ * @method User[] getGraders()
+ * @method User[] getUsers()
+ * @method Team[] getTeams()
+ */
+class GradingSection extends AbstractModel {
+    /**
+     * @var Gradeable
+     */
+    protected $gradeable;
```
KevinMackenzie: Does this class really need a reference to the `Gradeable` instance?

site/app/models/GradingSection.php
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * If this is a registration section (false for rotating)
+     * @var bool
```
KevinMackenzie: Mark all of the fields accessible by the magic methods with `@property @var`

site/app/libraries/database/DatabaseQueries.php
CouleeApps: Yep, was just copying some old code with this.

site/app/models/GradingSection.php
CouleeApps: No, it really doesn't.

site/app/models/GradingSection.php
CouleeApps: Somehow that makes PhpStorm not understand types... but will do.

site/app/models/Gradeable.php
CouleeApps: The point of those queries was to be a Team counterpart to the User queries, which don't index by section. See 10 lines below this. I could likely convert both but I don't want to make them inconsistent for the sake of moving a for-loop into a different file.

KevinMackenzie: fair enough

KevinMackenzie:
It seems like a lot of these changes aren't used anywhere yet, so they won't break anything.  Will have to be more thorough once all usages of the duplicate code factored out are replaced with these methods.

site/app/models/GradingSection.php
KevinMackenzie: Maybe try putting the description after the type name?

CouleeApps: Correction: it makes PhpStorm not understand array types. `@property @var string` it can get but `@property @var User[]` does not work.

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -1164,21 +1260,17 @@ public function getGradedGradeables(array $gradeables, $users = null, $teams = n
...
               LEFT JOIN (
-                SELECT
-                  g_team.team_id,
-                  registration_section,
-                  rotating_section,
-                  in_team.array_user,
-                  in_team.array_state
-                FROM gradeable_teams g_team
-                  LEFT JOIN (
+                SELECT gt.team_id, gt.registration_section, gt.rotating_section, json_agg(tu) AS array_team_users
```
KevinMackenzie: Try to put each field on its own line

CouleeApps: D:

PR #2406, opened 2018-07-13
New Gradeable model integration into Simple Grader interface

KevinMackenzie:
(see #2317) The simple grader interface and its associated controllers/views (for tests and labs) have been updated to use the new Gradeable model.  There should be no changes in the behavior of the page.


KevinMackenzie:
I still need to update the tests, but the page is working as far as I can tell.

PR #2407, opened 2018-07-13
New Gradeable model integration into Nav page

KevinMackenzie:
(see #2317) The nav page now uses the new Gradeable model.  Since the GradeableList is used in places other than the Nav page, I had to duplicate that class and its unit tests.  The page appears identical for `instructor`, `ta`, `grader`, and `student`.  When removing `student` from its different teams at different stages, the `team` button also appears to update correctly.

A significant number of the line changes are not new, rather they are from duplicating the `GreadeableListTester` for the new `GradeableList` that uses the new model (see above).

Also, the new function in `models/gradeable/Gradeable` for `getGradingProgress` is nearly identical to the one in `models/Gradeable.php`.


site/app/models/gradeable/Gradeable.php
```diff
@@ -739,26 +752,30 @@ public function anySubmissions() {
             // Until we find a submission, assume there are none
             $this->any_submissions = false;
             if ($this->type === GradeableType::ELECTRONIC_FILE) {
-                $semester = $this->core->getConfig()->getSemester();
-                $course = $this->core->getConfig()->getCourse();
-                $submission_path = "/var/local/submitty/courses/" . $semester . "/" . $course . "/" . "submissions/" . $this->getId();
+                $submission_path = FileUtils::joinPaths(
+                    '/var/local/submitty/courses',
```
MasterOdin: Use the config variable to get the `/var/local/submitty`. Don't hardcode this in.

site/tests/app/models/NewGradeableListTester.php
```diff
@@ -0,0 +1,299 @@
+<?php
```
MasterOdin: `site/tests/app/models/gradeable/GradeableListTester.php`

site/tests/app/models/NewGradeableListTester.php
```diff
@@ -0,0 +1,299 @@
...
+
+namespace tests\app\models;
```
MasterOdin: `namespace tests\app\models\gradeable\GradeableList1

PR #2412, opened 2018-07-13
Bump up value of pm.max_children of PHP-FPM conf

MasterOdin:
This now matches the default value of proxy workers we've setup within Apache.

If this fixes things for students, we will have to add to the sysadmin installation instructions details about tuning these values (at least say that they should look at them, and link them to a resource that better explains what it means).

```
sudo cp /usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/php-fpm/pool.d/submitty.conf /etc/php/#.#/fpm/pool.d/
sudo service php#.#-fpm restart
```

where `#.#` is either 7.0 (Ubuntu 16.04) or 7.2 (Ubuntu 18.04)


bmcutler:
seems to have fixed the problem for folks on vagrant, thank :)

Issue #2371, opened 2018-07-10
Add unique key constraint between g_id and team_id in gradeable_data table


KevinMackenzie:
Its all in the title.  There is a unique key constraint between `user_id` and `g_id` in the `gradeable_data` table, but not between `team_id` and `g_id`, which means we can have multiple grades for a team despite only allowing one grade per individual.

bmcutler:
@MasterOdin 
Is this still an open/valid issue?
Is this easy to fix?  Are there any difficulties if old courses do violate uniqueness?

MasterOdin:
This is still an issue. The fix is just doing something like `ALTER TABLE ONLY public.gradeable_data ADD CONSTRAINT UNIQUE (g_id, gd_team_id);`. I believe Submitty does a good enough job here to never violate this constraint in code currently so that we've never run afoul of it.

PR #2432, opened 2018-07-16
New Gradeable model integration into ajax SubmissionController methods

KevinMackenzie:
(see #2317).

The ajax methods in the `SubmissionController` now use the new Gradeable model.  This also means that an entire `GradeableList` does not need to be constructed for these routes.  The `HomeworkView`-rendering route still constructs the `GradeableList` as to not break it.

This conversion is designed to modify as little of the code as possible, so improvements that can be made by having access to the new model will be made in later PR's.


KevinMackenzie:
marked WIP until I get the tests passing

jchulton:
Getting the following error when I try to resubmit
teamsaveerror

if I try to resubmit as a team, I see this on the instructor grading page 
teamsaveerrorp2

KevinMackenzie:
The first error you're seeing should be fixed now, but I couldn't reproduce the second error.

KevinMackenzie:
The second error you're seeing is a reason why we should have #2371

bmcutler:
@KevinMackenzie please fix merge conflicts

PR #2433, opened 2018-07-16
Course materials page

jn7070:
Instructors can post/host files on the Submitty server for everyone to have the ability to view the files by clicking the popup file button if applicable (not docx, zip, or xlsx) and clicking the download button. Instructors can delete the file using the trash can feature. Permissions for the student to view popup files will be worked on later.


jchulton:
Two quick things that I think would be good to include, but won't hold up the pull request I'd recommend adding an error popup when a file or files fails to upload, right now there is no feedback. Also, I think you should mention in the upload materials popup that you can't upload docx, zip or xlsx.

PR #2434, opened 2018-07-16
GradedGradeable query also gets late day exceptions

KevinMackenzie:
Its important to have info for late day exceptions for the submitter.  Since late day exceptions are done per user and not per submitter, there must be a lookup table indexed by user id for the late day exceptions for each team member (or just one entry if not a team assignment).  This also means there needs to be two more `JOIN`s.  One for team late day exceptions, and one for user late day exceptions, one of which will always be null.


KevinMackenzie:
The first error is unrelated to this pr, but is fixed in #2448
The second error is because you're on an old version of the branch.

bmcutler:
@KevinMackenzie please resolve conflict

PR #2447, opened 2018-07-16
Improved popup form CSS

CouleeApps:
Factoring all the forms into a common script and giving them some nicer styles.

Preview:


ropern123:
Only thing I can think of is that now that the forms prevent you from clicking anything outside of them, we might want to make it so that the forms close when you click outside of them as well as having the close button (The statistics form for the simple grader does this).

PR #2462, opened 2018-07-17
Change display of archived courses for instructor

### LINKED ISSUE #2413: Prior Term courses should be separated from current courses on Submitty main page
Right now it's a jumble.  should be:

----------

"Your courses"

Within this section the list should be grouped by role (instructor, full access grader etc)
Then sorted by semester, then by course name.  (it looks like its currently sorted by course name first?)

"Your archived courses"

same sort as above

----------

We had a design for the archive status of a course, and whether it should be hidden from students or not.  Should be stored in the master database.   @pbailie please help as needed with the master database design.

screen shot 2018-07-13 at 12 58 56 pm

ropern123:
Changes how archived courses are displayed by making a new "Your Archived Courses" section under the "Your Courses" section on the homepage.

Also, changed the ordering of courses when retrieving from the database.

closes #2413

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -1978,31 +1978,61 @@ ORDER BY gt.{$section_key}", $params);
     }
 
 	/**
-	 * Retrieves all courses (and details) that are accessible by $user_id
+	 * Retrieves all unarchived courses (and details) that are accessible by $user_id
 	 *
-	 * (u.user_id=? AND u.user_group=1) checks if $user_id is an instructor
-	 * Instructors may access all of their courses
 	 * (u.user_id=? AND c.status=1) checks if a course is active
 	 * An active course may be accessed by all users
-	 * Inactive courses may only be accessed by the instructor
 	 *
 	 * @param string $user_id
 	 * @param string $submitty_path
-	 * @return array - courses (and their details) accessible by $user_id
+	 * @return array - unarchived courses (and their details) accessible by $user_id
 	 */
-    public function getStudentCoursesById($user_id, $submitty_path) {
+    public function getUnarchivedCoursesById($user_id, $submitty_path) {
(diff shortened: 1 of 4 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1978,25 +1978,61 @@ public function insertPeerGradingAssignment($grader, $student, $gradeable_id) {
...
         $this->submitty_db->query("
 SELECT u.semester, u.course
 FROM courses_users u
 INNER JOIN courses c ON u.course=c.course AND u.semester=c.semester
-WHERE (u.user_id=? AND u.user_group=1) OR (u.user_id=? AND c.status=1)
-ORDER BY u.course", array($user_id, $user_id));
+WHERE u.user_id=? AND c.status=1
+ORDER BY u.user_group ASC,
+         SUBSTRING(u.semester, 2, 2)::INT DESC,
+         CASE WHEN SUBSTRING(u.semester, 1, 1) = 's' THEN '1'
```
bmcutler: How does this work for other semester codes? 
(other schools may use different letters or semester labeling schemes)
Does it crash or default to alphabetic (or reverse alphabetic)?

ropern123: No clue at the moment.
Would the desired behavior be that it does this by default, but resorts to alphabetic if this fails?

bmcutler: Yes please.   Non-intuitive order better than crash :)

bmcutler:
Please switch the logic for the checks (sorry I told you wrong and it was undocumented)
1 = active course
2 = archived course
(othernumbers) = reserved for other levels of archiving, possibly hidden from instructors, etc.

So make the check 
Your courses, status==1
Your archived courses, status != 1
(and when we expand the usage to other numbers we will revise this page)

PR #2470, opened 2018-07-18
Course Materials Update

jn7070:
There are now collapsible folders to display files with the view course materials page displaying the first level of the course_materials folder by default with all other nested folders closed. In addition, there is a message that pops up asking if user wants to replace the file if there is already a duplicate file with the same path. The current folder setting textbox has changed to not allow instructors to change the path. They can only add additional folders. If no course materials, don't display the "course materials" button on the navigation page for everyone but instructor. The show to student checkbox is only for instructors to use and determine whether a student can see the file or not.

### DIFF
--- a/site/app/templates/course/UploadCourseMaterialsForm.twig
+++ b/site/app/templates/course/UploadCourseMaterialsForm.twig
@@ -30,14 +30,8 @@
         $(function() {
             $("#submit").click(function(e){ // Submit button
                 var requestedPath = document.getElementById("input_provide_full_path").value;
-                var n = requestedPath.indexOf("..");
-                if (n != -1) {
-                    alert(".. is not supported in the path");
-                }
-                else {
-                    makeSubmission(requestedPath);
-                    e.stopPropagation();
-                }
+                makeSubmission(requestedPath);
+                e.stopPropagation();
             });
         });
     </script>
--- a/site/app/templates/course/UploadCourseMaterialsForm.twig
+++ b/site/app/templates/course/UploadCourseMaterialsForm.twig
@@ -37,8 +37,14 @@
         $(function() {
             $("#submit").click(function(e){ // Submit button
                 var requestedPath = document.getElementById("input_provide_full_path").value;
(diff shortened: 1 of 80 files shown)


jn7070:
I've fixed the bugs and I did intend to leave folders in the server even if all the files/folder are deleted. I also added a checkbox to show to student feature. For the first draft, it will automatically reload/collapse the folders.

site/public/js/server.js
```diff
@@ -2139,6 +2171,10 @@ function escapeHTML(str) {
     return str.replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;');
 }
 
+function changePermission(filename, checked) {
+    // send to server to handle file permission change
+    window.location = buildUrl({'component': 'misc', 'page': 'modify_course_materials_file_permission', 'filename': filename, 'checked': checked});
+}
```
ropern123: It might be better to do use an ajax call here instead of changing window.location so that the page doesn't refresh every time a checkbox is clicked.

ropern123: ignore this, next pr as discussed in meeting

site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -31,8 +30,14 @@
         $(function() {
             $("#submit").click(function(e){ // Submit button
                 var requestedPath = document.getElementById("input_provide_full_path").value;
-                makeSubmission(requestedPath);
-                e.stopPropagation();
+                var n = requestedPath.indexOf("..");
+                if (n != -1) {
+                    alert(".. is not supported in the path");
```
ropern123: This should really be on the php side to make it more secure (interface can't submit paths with ".." but php will accept them with the right request)

ropern123: This should really be on the php side to make it more secure (interface can't submit paths with ".." but php will accept them with the right request)

Issue #2376, opened 2018-07-11
Forum paging for performance: scaling to huge # of posts


bmcutler:
As the # of threads, and total # of posts grows, the forum slows down.
Explore paging and other techniques to still have reasonable load and interaction times.

bmcutler:
@andrewaikens87 has a script to generate lots and lots of posts.  He will check that in and/or share by email.

PR #2473, opened 2018-07-18
Forum Thread paging

scopeInfinity:
Closes: #2376

Tasks
- Refactor DB queries method.
- Load thread list page in chunks(loads more when scroll hits bottom)
- ThreadStatus stored in the cookie
- List of the possible thread for merging is loaded (on clicking "Merge Thread")


scopeInfinity:
Status from an instructor perspective on a local system.

Now, for around 10000 threads
- view_thread page load time: 6.8KB ~300ms
- For loading next 10 threads info(left thread list) : 1.0KB  ~300ms
- Loading threads by clicking "Merge Thread"(~10k threads) : 50KB    200ms(Request) + 300ms(JS) ~500ms

Before this PR, while working on 3000 threads.
- view_thread page load time: ~30 secs

andrewaikens87:
Currently loading 10000 threads to test, I am going to have to let the tool run overnight. I will update this tomorrow

andrewaikens87:
I am getting a couple of undefined variables 
bad

andrewaikens87:
If a user selects a thread that is on a thread page that is greater than the first and refreshes, the thread list doesn't reflect the current thread's active style (the cyan blue color) when trying scrolling down to it again

andrewaikens87:
It seems like if I select a filter I lose the ability to scroll in the thread list and then when I try to select one of the threads that exists in the list I get alerted that there was an issue loading the thread list.

site/app/controllers/forum/ForumController.php
```diff
@@ -594,58 +569,61 @@ private function getSortedThreads($categories_ids, $max_thread, $show_deleted =
     }
 
     public function getThreads(){
-
+        $pageNumber = !empty($_GET["page_number"]) && is_numeric($_GET["page_number"]) ? (int)$_GET["page_number"] : -1;
```
andrewaikens87: We should be validating the lower bound of this page number, i.e. what if the page number is input as -15 (this can be done by simply changing the `next_page` html attribute). This will throw an exception since the offset in the query can't be negative.

scopeInfinity:
Thankyou @andrewaikens87 
Fixed those bugs.

site/app/controllers/forum/ForumController.php
```diff
@@ -569,7 +569,7 @@ private function getSortedThreads($categories_ids, $max_thread, $show_deleted, $
     }
 
     public function getThreads(){
-        $pageNumber = !empty($_GET["page_number"]) && is_numeric($_GET["page_number"]) ? (int)$_GET["page_number"] : -1;
+        $pageNumber = !empty($_GET["page_number"]) && is_numeric($_GET["page_number"]) && $_GET["page_number"] >= 1 ? (int)$_GET["page_number"] : -1;
```
andrewaikens87: Please test, if you input anything that is below 1 it will make the page number -1 (since ternary exp will be false), resulting in an exception when trying to run the offset query. A possible place to put this check might be where we try to load the threads.

PR #2476, opened 2018-07-18
Misc helper methods for new Gradeable Model

KevinMackenzie:
Accumulated helper methods for the new Gradeable model.  Most of these were in the old model and needed to be put somewhere in the new model.  Other helper methods extract code out of existing methods and puts them in better places.  These changes should not affect how any existing functions behave.

### DIFF
--- a/site/app/models/gradeable/AutoGradedVersion.php
+++ b/site/app/models/gradeable/AutoGradedVersion.php
@@ -364,7 +364,7 @@ class AutoGradedVersion extends AbstractModel {
 
     /**
      * Gets the number of days early this version is
-     * @return int reusult clamped to be >= 0
+     * @return int result clamped to be >= 0
      */
     public function getDaysEarly() {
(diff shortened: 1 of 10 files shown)


site/app/models/gradeable/AutoGradedVersion.php
```diff
@@ -332,6 +352,65 @@ public function getTotalPercent($clamp = false) {
...
     }
 
+    /**
+     * Gets the number of days late this version is
+     * @return int result clamped to be >= 0
+     */
+    public function getDaysLate() {
+        return max(0, DateUtils::calculateDayDiff(
+            (clone $this->getGradedGradeable()->getGradeable()->getSubmissionDueDate())
+                ->add(new \DateInterval("PT5M")), $this->submission_time));
```
CouleeApps: Grace period? Should probably be removed.

KevinMackenzie: I agree

PR #2505, opened 2018-07-20
Add view submissions buttons to nav page for ta

bmcutler:
And make sure TA has access to view submissions


bmcutler:
known bug:  full access grader with no assigned registration sections cannot "view all" students for non electronic gradeables.   Opened #2506

PR #2521, opened 2018-07-21
[WIP] PDF annotation

saxocellphone:
This is a very basic implementation of PDF annotation. I'm working on adding new features to it. But if anyone can just test out all the existing features and provide some feedback that would be great. 

Current features:
- Pen
- Selecting strokes (You have to be very precise, or else it doesn't work)
- Zoom (buggy, only zooms in one page. Need to find a non-expensive solution to make zooming better.)
- Saving

Features I'm planning to work on (in order)
- Adding text
- Changing the color and size of pen and text
- Better toolbar
- Better erasing
- Download

Known things to fix:
- There are no permission checks right now, don't worry about that for now...
- The download button is just a placeholder. This feature is going to take a while to implement
- When trying to drag the submission panel in annotation mode, only the top part works. This should be fixed hopefully soon.
- Sometimes when rendering the pdf, the image becomes weirdly messy. This doesn't happen often, so I can't successfully reproduce it. I'm guessing the problem is with the scrolling because the PDFs renders as you scroll, and I'm guessing there is a weird interaction there.


holzbh:
I tested using 8369510a90604b246226d674d42f56162eaf17b0 on a touch screen. In Chrome (Windows) I did not set any experimental flags and the behavior seemed smooth, though I couldn't make points by clicking without moving my cursor.

In Firefox (Windows) I also couldn't make points. My choices in the color/line-size picker were reflected on the toolbar but my line was always the default size and red no matter how many times I picked a new size/color and tried to draw more.

I also noticed I can't move around inside the blue submission window box. My resolution is 3200x1800 with 200% scaling on this computer, so I can't actually see the bottom of the PDF without zooming out more even if I make the blue box take up the entire height of my screen.

When zooming I noticed some odd behavior with lines. See the attached two screenshots which have the same lines but one is zoomed in slightly more (one more click of +). If you look at the word `in` in the title of the document, you can see that a line crosses through the 'n' in one zoom level but not in the other:

exhibita
exhibitb

I haven't tried the most recent commit, but I'm only one behind and it sounds like the most recent one has to do with saving... so I'm not sure if it'll fix any of these observations.

saxocellphone:
The zoom is basically a placeholder right now. It does not work as intended at all. I don't think the annotation library supports single point, so at least two points are needed in order to make a small dot. Also if you disable the stylus enhancement flag in chrome it should make the pen much smoother.

bmcutler:
@saxocellphone 
Please resolve conflicts.
And get the known issues and known next features into a new issue.

PR #2526, opened 2018-07-23
Lab/Test full instructor TA bugs

### LINKED ISSUE #2506: "View Your Sections" / "View All" for non electronic gradable
buggy for a full access grader with no assigned registration sections.

(@jchulton  ask if you don't understand...)

jchulton:
Closes #2506

### DIFF
--- a/.setup/bin/setup_sample_courses.py
+++ b/.setup/bin/setup_sample_courses.py
@@ -59,6 +59,7 @@ with open(os.path.join(SUBMITTY_INSTALL_DIR,"config","database.json")) as databa
     DB_PASS = database_config_json["database_password"]
 
 DB_ONLY = False
+NO_SUBMISSIONS = False
 
 NOW = dateutils.get_current_time()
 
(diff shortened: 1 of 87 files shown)


holzbh:
So I'm not sure what the expected behavior here is supposed to be, but using your branch, if I log in as `instructor` and click "GRADE" for "Grading Lab" (`grading_lab`), on master I see 11 section ("Students Enrolled in Registration Section X", 1-10 + NULL).

Using your branch, and doing the exact same thing, I instead see `No Grading To Be Done! :)`
The `test_header_text` looks for these section headers, and since they don't exist, attempting to extract a `tr` with the class `info persist-header` fails, causing the NoSuchElementException.

If the intended behavior is that the instructor shouldn't see anything in the lab by default, then the test should be updated. (It should also test clicking View All and then seeing that the sections still exist, unless this is tested in a different e2e test.)

If the intended behavior is by default that instructors should see all students, even if not in a section assigned to the instructor, that behavior should be fixed and the test shouldn't be altered.

jchulton:
The intended behavior is that, if the instructor has no sections, the instructor should see all students by default (essentially, should start in "View All" mode, but selecting the "View Sections" button would cause the No Grading to Be Done! :)

jchulton:
The code has been modified to fit the intended behavior above. Note that instructor now shows all students by default (should confirm with Barb this is intended behavior.

bmcutler:
Yes, correct.  If an instructor is assigned no sections, they should view all (including the null section) by default.  Also, if an instructor is assigned no sections, the forward & back buttons in the ta grading interface should visit all students (not including the null section).  If an instructor (or TA) views a null section student the forward & backward should be greyed out.

PR #2530, opened 2018-07-24
Course Materials Part 3

jn7070:
1. Share to student on a date checkbox will show the file(s) to the students on the specified date/time and should be automatically saved. Full-access graders and limited-access graders can see all files except for the show to student checkbox/date and upload materials button.
2. Unzip zip files checkbox should unzip any zip files drag-and-dropped/added into the box when it is checked. It will make not make a new folder with the zip file name containing all the files and folders in it. Instead, dump all files/folders in zip in highest folder level (course_materials or optional folder(s)).
3. Trash can delete should remain on file/folder structure it is on after deleting. Refreshing the page will collapse all folders.
4. Trash can delete for folders


jn7070:
What kind of zip file did you use? Is it bigger than 10 MB?

jn7070:
Ready for a relook thanks!

PR #2531, opened 2018-07-24
install_system.sh tweaks

andrewaikens87:
I was getting the following error when trying to do a vagrant up (caused the vagrant install to crash as the correct user groups weren't being created): 

`grep: ]: No such file or directory
[: missing ]`

These changes allow me to install completely.


jchulton:
These changes allowed me to do a vagrant up as well, I got the same error on master.

PR #2533, opened 2018-07-24
Fixed Create Gradeable 'Build queue entry failed'

KevinMackenzie:
The check that the build queue file and form config file used to determine if the file was available for writing was incomplete.  It would considered the "no file exists" case to be a "can't write to file" case.  Now, the `is_writable` method is used in conjunction with `file_exists` to allow writing in the "no file exists" case.


bmcutler:
This appears to fix the editing / url encoding.

PR #2538, opened 2018-07-24
Full access grader team submit nav page button displays properly if not on team

KevinMackenzie:
If the user was a full access grader and was not on a team, the submit button indicated that they `MUST BE ON A TEAM TO SUBMIT` when they should get the same view as the instructor for `OVERDUE SUBMISSION` since full access graders can submit for students.

### DIFF
--- a/site/app/views/NavigationView.php
+++ b/site/app/views/NavigationView.php
@@ -450,7 +450,7 @@ class NavigationView extends AbstractView {
                 if ($list_section > GradeableList::OPEN) {
                     $class = "btn-danger";
                     if ($this->core->getUser()->accessFullGrading()) {
-                        // team assignment, no team (admin)
+                        // team assignment, no team (full access grader)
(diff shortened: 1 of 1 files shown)


site/app/views/NavigationView.php
```diff
@@ -440,15 +440,17 @@ private function getSubmitButton(Gradeable $gradeable, $graded_gradeable, int $l
...
-                $disabled = true;
+                if (!$this->core->getUser()->accessFullGrading()) {
+                    $title = "MUST BE ON A TEAM TO SUBMIT";
+                    $disabled = true;
+                }
                 if ($list_section > GradeableList::OPEN) {
                     $class = "btn-danger";
-                    if ($this->core->getUser()->accessAdmin()) {
-                        // team assignment, no team (admin)
+                    if ($this->core->getUser()->accessFullGrading()) {
```
KevinMackenzie: This should check against minimum grading group

PR #2539, opened 2018-07-24
[WIP] Homework/Submission page (mostly) using new model

KevinMackenzie:
part of (#2317).

Most of the homework/submission page now uses the new Gradeable model.  The exceptions are the late days banner (waiting on rumored new LateDaysModel class) and the regrade requests (which won't be in the new model until teams are supported).  Both of these page features are easily separable from the rest of the page and making the final conversion should be trivial.

A significant portion of the deletions of this PR are because the `GradeableList` and `GreadeableListTester` were deleted since this view / controller route were the last place that used it.

Marked as WIP because the `SubmissionControllerTester` may not be correctly setup yet, but the page functions should be correct.


CouleeApps:
Also, specifying an invalid version causes some error messages to show up that don't appear on master (less important but I found this while pressing the back button after changing versions). Ex: http://192.168.56.101/index.php?semester=f18&course=sample&component=student&gradeable_id=grades_released_homework_onlyta&gradeable_version=0

CouleeApps:
Trying to use textbox submission throws a server error when submitting:

PR #2543, opened 2018-07-25
Access functions: Simple Grader + new model support

CouleeApps:
Converted the simple grading (lab/test) page to use the new Access functions. This required me to add support for the new model, so I have. Rough changelog:

- Simple grader with new Access
- Access now supports gradeable\Gradeable and GradedGradeable
- Moved some access checks from the view to the controller (electronic grader, simple grader)
- More specific permission path `grading.electronic.xyz` instead of `grading.xyz`
- New model with GradingSection support


KevinMackenzie:
When viewing the lab grading page as instructor / full-access-grader, I see this:
image
But when I log in as the `grader` user, I see this:
image
It might make sense not to show everyone here, but it looks like the `grader` user isn't even supposed to be grading this.
(works correctly on master.... I hate how I feel like I need to suffix an issue with that)

KevinMackenzie:
Found another thing: The grader can view the entire page for `rotating sections lab` of all sections and the student grades, but they can't change them. (insufficient permissions message in console).  On master they get the correct message:  `No grading to be done :-)`.  Note: the minimum grading group for this lab is set to limited access graders on my config.

CouleeApps:
Yep I need to merge #2541

PR #2549, opened 2018-07-25
Gradeable only appears on nav page if it has buttons

KevinMackenzie:
If the gradeable doesn't have any buttons, the user has no action to complete on it, so they should not see it.  This happens when the grader doesn't meet the minimum grading group for an assignment.

Is this a desirable behavior?


CouleeApps:
If they should not be visible at all then we should handle that in the controller. There's a method, `filterCanView`, that checks for that.

PR #2551, opened 2018-07-25
Peer improvements

### LINKED ISSUE #1387: Peer grading code improvement
- [ ] ~~Since in gradeable_component_data the gcd_grader_id is part of the primary key it would be nice to move that to the third column, next to the other columns that make up the primary key~~
- [x]  Break out checking permissions to a seperate function
- [x]  Remove unused code in electronic grader view and controller from when we saved the entire gradeable at once
- [ ]  ~~Update AbstractDatabaseQueries.php~~
- [x]  Should remove total_tagrading_extra_credit and total_peer_grading_extra_credit. Rename total_tagrading_non_extra_credit and total_peer_grading_non_extra_credit to indicate that they are max...there is no longer explicit extra credit or not
- [ ]  ~~Not sure why getgradeablecomponents was rewritten, it's more difficult to read/understand~~
- [ ]  ~~Instead of checking if getGrader() is null to see if a component is graded, we should check if the component has_marks or has_grade.~~

jchulton:
Closes #1387
Only address the final checkbox, everything else has been addressed in the past or is no longer relevant.


site/app/models/Gradeable.php
```diff
@@ -284,11 +284,9 @@ class Gradeable extends AbstractModel {
 
     protected $been_tagraded = false;
 
-    protected $total_tagrading_non_extra_credit = 0;
-    protected $total_tagrading_extra_credit = 0;
+    protected $total_tagrading_points = 0;
```
KevinMackenzie: change to `total_ta_grading_points` to be consistent with `total_peer_grading_points` (and also utilize a magic getter like the `getTotalPeerGradingPoints` method does)

jchulton: the magic getter getTotalTAGradingPoints defaults to return total_t_a_grading_points, so I think we have to stick with a non-magic version if we want to keep the expected naming convention.

site/app/models/Gradeable.php
```diff
@@ -1409,73 +1407,13 @@ public function getGradingSectionsForUser(User $user) {
 
             $sections = [];
             foreach ($section_names as $section_name) {
-                $sections[] = new GradingSection($this->core, $this->isGradeByRegistration(), $section_name, $graders[$section_name] ?? [], $users[$section_name] ?? null, $teams[$section_name] ?? null);
+                $sections[] = new GradingSection($this->core, $this->isGradeByRegistration(), $section_name, $graders[$section_name], $users[$section_name] ?? null, $teams[$section_name] ?? null);
```
KevinMackenzie: Why did you remove the default parameter?  Other than that it looks good

PR #2554, opened 2018-07-26
Removed out-of-place code from new model

KevinMackenzie:
There was a `just_regraded` field added to the `Gradeable` class (both versions).  I do not know what the plans are for this variable, but it doesn't belong in the `gradeable\Gradeable` class.  As described in the class-doc, this class is only for the configuration of a gradeable.

It appears like this variable is being used to notify the user if they have new regrade messages, but this should be done using a notification system shared (at least somewhat) with the forum, et al.  Since notifications need to be per-user, not per submitter, it probably should not be in the GradedGradeable model either.


jchulton:
Glenn's right, the code would cause the View Grade button to appear as green for the student if a regrade request has been closed and the student hadn't viewed their grade since then. I'll work on making the system per-user.

PR #2557, opened 2018-07-26
Grades update when graded by someone different

### LINKED ISSUE #2547: Simple grader marks don't save across graders
If a grader (limited/full access/instructor) grades a simple grading mark, then no other grader is able to change the value of that mark. There is no indication of failure to save the mark until the page is refreshed, at which point the mark is reset to whatever the original grader graded it as.

This is happening for labs, I haven't tested for numeric.

ropern123:
Simple grades now properly update when graded by someone different from their original grader.

closes #2547 

~~Going to add a test case to make sure that this issue does not pop up again in the future.~~ Done


MasterOdin:
The reason I suspect it's like this is I suspect because of peer grading where the grader remains static to a given peer and so the where clause makes sense there. However, it then breaks for normal grading, so it may make sense to have an if statement that switches query depending if peer or not.

Discuss with @jchulton and @bmcutler on the design of this.

MasterOdin:
This same function will probably be used for both normal TA grading and peer grading except for the slight variation I noted above. I'm not sure the status of peer grading implementation however and can't say if this will immediately break things for them. At least put a note on the function linking here discussing implantation of peer grading should be handled as this same bug landed in master last summer due to peer grading work and I'd like to avoid it happening a third time.

PR #2561, opened 2018-07-26
Use json_agg when loading from the model, prevent \ issues

CouleeApps:
Currently, if a mark note ends in a backslash, the entire page breaks. This is because `fromDatabaseToPHPArray` cannot handle backslashes nor the large majority of weird escaped strings. To fix, I swapped array_agg with json_agg because JSON is capable of data (de)serialization without errors.


bmcutler:
@CouleeApps Is this ready to go now?  Matt thinks it is.  Why were we waiting?

PR #2564, opened 2018-07-26
Course settings autosave

### LINKED ISSUE #2480: Course settings page should autosave
A lot of other pages autosave, the course settings page should autosave.

ropern123:
The course settings page autosaves.

closes #2480


site/app/controllers/admin/ConfigurationController.php
```diff
@@ -71,55 +71,49 @@ public function viewConfiguration() {
...
     public function updateConfiguration() {
         if (!$this->core->checkCsrfToken($_POST['csrf_token'])) {
-            $this->core->addErrorMessage("Invalid CSRF token. Try again.");
-            $_SESSION['request'] = $_POST;
-            $this->core->redirect($this->core->buildUrl(array('component' => 'admin',
-                                                              'page' => 'configuration',
-                                                              'action' => 'view')));
+            $response = array('status' => 'fail', 'message' => 'Invalid CSRF token');
+            $this->core->getOutput()->renderJson($response);
+            return $response;
```
KevinMackenzie: replace this with `Output::renderJsonFail`

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -71,55 +71,49 @@ public function viewConfiguration() {
...
-        if (!isset($_POST['course_name']) || $_POST['course_name'] == "") {
-            $this->core->addErrorMessage("Course name can not be blank");
-            $_SESSION['request'] = $_POST;
-            $this->core->redirect($this->core->buildUrl(array('component' => 'admin',
-                                                              'page' => 'configuration',
-                                                              'action' => 'view')));
+        if(!isset($_POST['name'])) {
+            $response = array('status' => 'fail', 'message' => 'Name of config value not provided');
+            $this->core->getOutput()->renderJson($response);
+            return $response;
```
KevinMackenzie: see above

site/app/templates/admin/Configuration.twig
```diff
@@ -148,10 +148,30 @@
...
+                    url: "{{ core.buildUrl({'component': 'admin', 'page': 'configuration', 'action': 'update'})|raw }}",
+                    data: formData,
+                    type: "POST",
+                    processData: false,
+                    contentType: false
+                });
+            });
+        });
+    </script>
+</div>
```
KevinMackenzie: Make a .js file for the ajax call.  Also, look into using `$.getJSON` so you can interpret your json response on the client and tell the user if something went wrong.

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -71,55 +71,49 @@ public function viewConfiguration() {
...
-                'vcs_base_url'              => $_POST['vcs_base_url'],
-                'vcs_type'                  => $_POST['vcs_type'],
-                'forum_enabled'				=> $_POST['forum_enabled'],
-                'regrade_enabled'           => $_POST['regrade_enabled'],
-                'regrade_message'           => $_POST['regrade_message'],
-                'private_repository'        => $_POST['private_repository']
-            )
-        );
+        if(in_array($name, array('default_hw_late_days', 'default_student_late_days'))) {
+            $entry = $entry === "" ? intval($entry) : 0;
```
KevinMackenzie: you should check `ctype_digit`  on these parameters and `Output::renderJsonFail` if they aren't integers.  If you put a string in this field it will save a 0 and you wouldn't know that until you refresh the page.

KevinMackenzie: The boolean expression on 100 actually will only accept `$entry` if it is a blank string, which means this will always save a zero.

KevinMackenzie:
Also, see the `Output::renderJson*` methods for how we're doing json responses going forward

site/app/controllers/admin/ConfigurationController.php
ropern123: whoops

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -71,55 +71,42 @@ public function viewConfiguration() {
 
     public function updateConfiguration() {
         if (!$this->core->checkCsrfToken($_POST['csrf_token'])) {
-            $this->core->addErrorMessage("Invalid CSRF token. Try again.");
-            $_SESSION['request'] = $_POST;
-            $this->core->redirect($this->core->buildUrl(array('component' => 'admin',
-                                                              'page' => 'configuration',
-                                                              'action' => 'view')));
+            $this->core->getOutput()->renderJsonFail('Invalid CSRF token');
```
KevinMackenzie: I think you want to `return` in this case to stop control flow.  On a similar note, the only reason some controller methods return a value is so the PHPUnit tests can get the result of the controller without mocking the view.

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -71,55 +71,42 @@ public function viewConfiguration() {
...
-            )
-        );
+        if(in_array($name, array('default_hw_late_days', 'default_student_late_days'))) {
+            if(!ctype_digit($entry)) {
+                return $this->core->getOutput()->renderJsonFail('Must enter a number for this field');
+            }
+            $entry = intval($entry);
+        }
+
+        if(in_array($name, array('zero_rubric_grades', 'keep_previous_files', 'display_rainbow_grades_summary', 'display_custom_message', 'forum_enabled', 'regrade_enabled'))) {
```
KevinMackenzie: `else if`

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -71,55 +71,42 @@ public function viewConfiguration() {
...
+            $entry = $entry === "true" ? true : false;
+        }
+
+        if($name === 'upload_message') {
+            $entry = nl2br($entry);
+        }
 
-        $this->core->getConfig()->saveCourseIni($save_array);
+        $config_ini = $this->core->getConfig()->readCourseIni();
+        $config_ini['course_details'][$name] = $entry;
```
KevinMackenzie: error check that `$name` is a valid entry for `course_details`

site/public/js/course_settings.js
```diff
@@ -0,0 +1,33 @@
+$(function() {
```
KevinMackenzie: Our naming convention is to hyphenate multi-word .js files.

KevinMackenzie:
you're failing the lint test because there's a missing semicolon in the `PlagiarismController.php`. If you update from master it should be fine.

PR #2580, opened 2018-07-27
Fixed team nav page button

KevinMackenzie:
Full access / instructors should be able to submit for students if they aren't on a team.  The nav page button indicates they can't, but this fixes that.


site/app/views/NavigationView.php
```diff
@@ -449,15 +449,19 @@ private function getSubmitButton(Gradeable $gradeable, $graded_gradeable, int $l
                 $title = "TA GRADE NOT AVAILABLE";
             }
         } else {
-            // This means either the user isn't on a team, or has no submissions yet
+            // This means either the user isn't on a team
             if ($gradeable->isTeamAssignment()) {
+                // access in the view... seems bad
+                $submit_any = $this->core->getAccess()->canI('gradeable.submit_any', ['gradeable' => $gradeable]);
```
CouleeApps: Would highly recommend moving this to the controller and passing in a var eg `$show_submit_no_team`

site/app/libraries/Access.php
```diff
@@ -104,6 +104,8 @@ public function __construct(Core $core) {
         $this->permissions["grading.simple.grade"] = self::ALLOW_MIN_LIMITED_ACCESS_GRADER | self::CHECK_GRADEABLE_MIN_GROUP | self::CHECK_GRADING_SECTION_GRADER;
         $this->permissions["grading.simple.show_all"] = self::ALLOW_MIN_FULL_ACCESS_GRADER;
         $this->permissions["grading.simple.upload_csv"] = self::ALLOW_MIN_FULL_ACCESS_GRADER | self::CHECK_GRADEABLE_MIN_GROUP;
+
+        $this->permissions["gradeable.submit_any"] = self::ALLOW_FULL_ACCESS_GRADER | self::CHECK_GRADEABLE_MIN_GROUP;
```
CouleeApps: Better path would be `gradeable.submit.without_team`

KevinMackenzie:
The case that its relevant for the view is when the user doesn't have a team, but in general this access control configuration is for submitting for another individual

PR #2590, opened 2018-07-28
Moved rebuild button to edit gradeable page

### LINKED ISSUE #2524: Move Rebuild Gradeable button into Edit Gradeable page
The rebuild button is currently a wrench icon on the navigation page, but it would make more sense to put this feature in the Edit gradeable page near the Config setting and build output (Currently `Autograding` tab).  If the build failed, instead of having the regrade button/wrench be red, the pencil icon for `Edit gradeable` becomes red, gets a red background, or something.

ropern123:
Moves the rebuild button from the nav page to the edit gradeable page.

closes #2524

~~Failing travis tests due to less nav page buttons..~~ seem to be working now.


bmcutler:
2 small changes requested:

* Let's make the new rebuild button a blue button, encouraging the user to click it.

* After clicking it, please stay on that page.  (right now it's jumping back to the navigation page).

And then eventually we'll want this page to refresh (every 10-30 seconds?) until the rebuild completes (the rebuild queue file is deleted once it completes) and we can examine the new build log, but we can skip that for now if it's not easy & quick.

Issue #2492, opened 2018-07-19
Student notification of merged thread


bmcutler:
If a student's new thread is merged into another thread it should be obvious where it went.

Options (perhaps we implement all!)

- [x] An toggle option (visible to everyone) to "show merged threads", off by default.  When clicked, we add those threads back into the list of all threads, but those merged threads are just links to the new parent thread.  These merged threads are also colored differently (design needed...)

- [ ] If your thread was merged, you get a notification, it's one of the things that counts in your unviewed forum things badge count.

- [x] When you view the list of all threads, every thread where you have posted at least once is colored or visually marked.  So if your new thread was merged into an old existing thread, you have a 'post' in that old thread and its marked/colored appropriately.

Thoughts anyone?

Issue #2489, opened 2018-07-19
Unified visual & database design for different notifications/badges 


bmcutler:
Need to discuss how we're going to (implement the) display the badges (red rounded square with a number on the upper right corner of the relevant buttons.

Need to discuss/design how each students notifications are stored in the database.

PR #2591, opened 2018-07-28
Notifications

scopeInfinity:
Notifications System #2489
- Notification Page
- Numbers are written on the button(red background) showing unseen notification
- Student Notification on Thread Merge(#2492)
- Notification to all on creation of announcement/upgrade of thread to an announcement
- Notification to the author of the post, on replying.


migration/data/course_tables.sql
```diff
@@ -498,6 +498,25 @@ CREATE TABLE regrade_discussion (
...
 
+--
+-- Name: notifications_type_enum; Type: ENUM; Schema: public; Owner: -
+--
+CREATE TYPE notifications_type AS ENUM ('forum');
+
+--
+-- Name: notifications; Type: TABLE; Schema: public; Owner: -
+--
+CREATE TABLE notifications (
```
KevinMackenzie: We might want to add a `Notification` class in the Model to represent this table

migration/data/course_tables.sql
```diff
@@ -498,6 +498,25 @@ CREATE TABLE regrade_discussion (
...
+    id serial NOT NULL PRIMARY KEY,
+    type notifications_type NOT NULL,
+    metadata TEXT NOT NULL,
+    content TEXT NOT NULL,
+    from_user_id VARCHAR(255) NOT NULL,
```
KevinMackenzie: we may want this to be null so that we can have notifications that don't originate from a user

scopeInfinity:
Thanks, @KevinMackenzie
Tried an attempt with Notification class as Model.

site/app/models/Notification.php
```diff
@@ -0,0 +1,154 @@
...
+
+    /** @property @var string Notification source user (can be null) */
+    protected $notify_source;
+    /** @property @var string Notification target user(s) (null implies all users) */
+    protected $notify_target;
+    /** @property @var string Notification text content */
+    protected $notify_content;
+    /** @property @var string Notification information about redirection link */
+    protected $notify_metadata;
+    /** @property @var string Should $notify_source be ignored from $notify_target */
```
KevinMackenzie: should this be a bool?

site/app/models/Notification.php
```diff
@@ -0,0 +1,154 @@
...
+                break;
+        }
+    }
+
+    /**
+     * Handles notifications related to forum
+     *
+     * @param array $details
+     */
+    private function handleForum($details) {
```
KevinMackenzie: Instead of passing a `details` array here, make this function take explicitly named parameters

site/app/models/Notification.php
```diff
@@ -0,0 +1,154 @@
...
+        switch ($details['type']) {
+            case 'new_announcement':
+                $this->setNotifyMetadata(json_encode(array('component' => 'forum', 'page' => 'view_thread', 'thread_id' => $details['thread_id'])));
```
KevinMackenzie: Why do we need `component` / `page` in the metadata?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2360,6 +2360,94 @@ public function getAllAnonIds() {
...
+     *
+     * @param string $source_user_id     user_id of person who generated notification
+     * @param string $type               Type of notification
+     * @param string $metadata           Information about the notification
+     * @param string $content            Text message of notification
+     * @param string $target_users_query Query selection for list of users
+     * @param string $additional_param   Additional parameters to be appended
+     * @param bool   $ignore_self        Should ignore $source_user_id from target users
+     */
+    public function pushNotification($source_user_id, $type, $metadata, $content, $target_users_query, $additional_param, $ignore_self){
```
KevinMackenzie: We should have one `pushNotification` function in the `DatabaseQueries` class that takes a `Notification` model instance and switches which operation occurs based on the `Notification`s state.

site/app/models/Notification.php
scopeInfinity: The initial idea was to store **url** info in the metadata.
So that using `$this->core->buildUrl` we can resolve metadata into url, easily.

site/app/templates/Notifications.twig
```diff
@@ -0,0 +1,50 @@
...
+                                    </a>
+                                {% endif %}
+                            </td>
+                        </tr>
+                    {% endfor %}
+                </tbody>
+
+            {% if notifications is empty %}
+            <tr>
+                <td colspan="8">No notification found</td>
```
andrewaikens87: It might be nice to have "No new notifications." be displayed for if notifications is empty when looking at the unread notifications

site/app/controllers/NavigationController.php
```diff
@@ -122,4 +125,29 @@ private function filterCanView(Gradeable $gradeable) {
 
         return true;
     }
+
+
+    private function notificationsHandler() {
+        $user_id = $this->core->getUser()->getId();
+        if(!empty($_GET['action'])){
+            if($_GET['action'] == 'open_notification' && is_numeric($_GET['nid']) && $_GET['nid'] >= 1) {
```
andrewaikens87: This will throw an undefined index if `action` is defined but `nid` is not

site/app/models/Notification.php
```diff
@@ -0,0 +1,223 @@
...
+
+    private function actAsUpdatedAnnouncementNotification($thread_id, $thread_title) {
+        $this->setNotifyMetadata(json_encode(array('component' => 'forum', 'page' => 'view_thread', 'thread_id' => $thread_id)));
+        $this->setNotifyContent("Announcement: ".$thread_title);
+        $this->setNotifySource($this->getCurrentUser());
+        $this->setNotifyTarget(null);
+    }
+
+    private function actAsForumReplyNotification($thread_id, $post_content, $target) {
+        // TODO: Redirect to post itself
```
andrewaikens87: In the URL you can simply put #<post_id> at the end, try using the search feature and inspecting element on a resulting post.

site/app/templates/Notifications.twig
```diff
@@ -0,0 +1,50 @@
...
+        {% if show_all %}
+            <a href="{{ core.buildUrl({"component": "navigation", "page": "notifications"}) }}" class="btn btn-primary">Show unread only</a>
+        {% else %}
+            <a href="{{ core.buildUrl({"component": "navigation", "page": "notifications", "show_all": "1"}) }}" class="btn btn-primary">Show All</a>
+        {% endif %}
+    </div>
+    <h2>Notifications</h2>
+    <table class="table table-bordered persist-area">
+                <tr class="info persist-header">
+                    <td colspan="8" style="text-align: center">Course: {{ course  }}</td>
```
andrewaikens87: It might be better to have "New notifications" and "All notifications" for the table since when a user is currently viewing the notifications page it is within a course.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2360,6 +2361,109 @@ public function getAllAnonIds() {
...
+                (case when seen_at is NULL then false else true end) as seen,
+                (extract(epoch from current_timestamp) - extract(epoch from created_at)) as elapsed_time, created_at
+                FROM notifications WHERE to_user_id = ? and {$seen_status_query} ORDER BY created_at DESC", array($user_id));
+        $rows = $this->course_db->rows();
+        $results = array();
+        foreach ($rows as $row) {
+            $results[] = new Notification($this->core, array(
+                    'view_only' => true,
+                    'id' => $row['id'],
+                    'component' => $row['type'],
```
KevinMackenzie: Its confusing that you have a `type` field in the database that gets converted to a `component` field in the model, but that model also takes a `type` parameter that means something different.  If you make `type` consistent with `type` in the database and maybe change `type` as it exists in php now to `action` it would be a lot more clear.

site/app/views/NavigationView.php
```diff
@@ -694,4 +694,13 @@ public function deleteGradeableForm() {
         return $this->core->getOutput()->renderTwigTemplate("navigation/DeleteGradeableForm.twig");
     }
 
+    public function listNotifications($notifications, $show_all) {
```
KevinMackenzie: We could probably put this in a `NotificationsView` class or just call this code in-line where this function gets invoked (`NavigationController`).  It doesn't make much sense to have it be in the Navigation View

Issue #2572, opened 2018-07-27
(minor?) UI improvements to Plagiarism 


bmcutler:
On the "main" page:

- [x] If there is a queue file in the jobs folder -OR- a finished plagiarism run (the ranking.txt file exists) there should be an entry on this page.  (currently it's only looking for ranking.txt?)

- [x] The date should be the timestamp on the ranking.txt file (or N/A if there's only a queue file for a newly created job).  I think the timestamp might be the queue time??  (not sure)

- [x] Next to each entry on the main page, add the # of students in the current term who have submitted this assignment at the time of the last run (# of subfolders in  /var/local/submitty/courses/<SEMSETER>/<COURSE>/lichen/concatenated/<GRADEABLE>/).  And the # of submissions.
(# of subfolders in /var/local/submitty/courses/<SEMSETER>/<COURSE>/lichen/concatenated/<GRADEABLE>/*/).   Format example:  "27 students, 101 submissions"

- [x] Add a boolean checkbox next to each entry on the main page for: "Enable nightly re-run".  By default this option is off.

- [x] (EDITED TO ADD)  If one of the entries has been put in the queue for reprocessing, make the re-build icon and/or date a different color?  (So we know we don't have to press rebuild again).  Maybe we can display the difference between queue file (job not started yet) and the PROCESSING_ file (job started but not completed)

- [x] (EDITED TO ADD) If any of the entries on this page is in the queue to be run, refresh the page every 30 seconds (or 10 seconds or a minute??)  We do this on other pages in the system.

On the "results" page":

- [ ] Make the left & right boxes full height.

- [x] Reduce space between the left & right file boxes.  

- [x] In all places where the user name is displayed (in drop down and selection) use "preferred first name (display name) last name <username>" (standardize to other uses on the website)

- [x] (EDITED TO ADD) Add a 'legend/key' at the top to explain what the grey/yellow/orange/red colors mean.

- [ ] I don't think there should be a "None" choice in the version dropdown menu for the left panel user.

bmcutler:
Sounds good :)

tushargr:
@bmcutler 
Last point in task list-
"None" option for version is required to be displayed when no user is selected in left side. Can we leave this for now.

bmcutler:
I've unchecked the full-height of code option.  If I 'zoom out' (ctrl minus) in my browser, and make my browser tall, I still only have ~30 lines of code visible.  Hoping we can make the code window full height to the current browser size.

screenshot:
[...]

PR #2595, opened 2018-07-29
More minor modifications for lichen

tushargr:
Addresses part of #2572 

Forgot to mention in previous pull request- the main page refreshes in 5 SEC if any lichen job is there.

Also @bmcutler  What color do we choose for showing in queue and in process job on mainpage
Currently they are 
in queue - red
in process - green


bmcutler:
(Sorry I introduced a bug with the missing ; in my edits to the last PR)
However, the code is still not full height (on mac on chrome).

PR #2599, opened 2018-07-29
Forum: Show Merged threads

scopeInfinity:
Issue: #2492

Currently, the merged thread have "fa fa-link" icon in the threads list, it can be changed(suggestions?)


site/public/js/server.js
```diff
@@ -1589,6 +1589,11 @@ function alterShowDeletedStatus(newStatus) {
     location.reload();
 }
 
+function alterShowMergeThreadStatus(newStatus) {
+    document.cookie = "show_merged_thread=" + newStatus + "; path=/;";
```
andrewaikens87: This should be course specific

site/app/controllers/forum/ForumController.php
```diff
@@ -637,18 +643,28 @@ public function showThreads(){
...
+                $this->core->redirect($this->core->buildUrl(array('component' => 'forum', 'page' => 'view_thread', 'thread_id' => $thread['merged_thread_id'])));
+                return;
+            }
             if($option == "alpha"){
                 $posts = $this->core->getQueries()->getPostsForThread($current_user, $thread_id, $show_deleted, 'alpha');
             } else {
                 $posts = $this->core->getQueries()->getPostsForThread($current_user, $thread_id, $show_deleted, 'tree');
             }
+            if(empty($posts)){
+                $this->core->addErrorMessage("No posts found for given threads.");
```
andrewaikens87: Let's make this: "No posts found for selected thread."

andrewaikens87:
For the child thread, when displaying merged threads, the resolve state is shown but there is no way to resolve it. We should either not display the resolve state for the child thread when displaying merged threads or give the ability to the creator of the child thread to resolve their original thread.

test_thread

PR #2612, opened 2018-07-30
Multiple graders for PDF annotations

saxocellphone:
Now multiple graders can grade the same document.


ropern123:
On a note somewhat unrelated to this pr, it seems like the locations of the text comments don't actually change when you move them, save, and refresh.

PR #2560, opened 2018-07-26
Regrade Request Date for every assignment

jchulton:
Regrade requests can now be enabled/disabled for every assignment. Assignments have a regrade request due date: regrade requests must be submitted before this date
Closes #2494

PR #2613, opened 2018-07-30
Added "Grades Due" Date that is between grade start/released date

### LINKED ISSUE #2070: "Grades Due" date that is before the "Grades Released" date
With grade verification happening soon, TAs may need some time after mentors have finished grading to verify their marks. It would be nice to have a due date for grades that is before they are released to students to allow time for this.

KevinMackenzie:
Closes #2070

Added a `g_grade_due_date` to the database for all gradeables,  This will default to the release date for any existing gradeables.  The database constraints needed to be modified to accommodate this new field.  This feature is not in use anywhere, but the necessary code to edit it and validate it are in place.

To test:
1. Run migrator (part of install submitty) and see that no errors occurred in output
2. Run migrator rollback, then migrate again
3. Test Edit Gradeable page to see if it shows an error on invalid date
4. Set grade due date of gradable to be after `now` and then Release grades (nav page)


CouleeApps:
Missing support for the old model? Or are you just assuming that everything will be updated to the new model before this is used?

KevinMackenzie:
We're assuming that everything that wants to use this feature will convert to the new model if it hasn't already.

bmcutler:
Grades due date is correctly not visible for electronic gradeables with no ta grading.  However, we seem to have an invisible dependency.  If I edit the surrounding dates (due date and grades released date) to be before or after the inivisible manual grading due date it will fail to save.

KevinMackenzie:
This is an extension of an existing behavior.  What happens is: if ta grading is disabled, then the grading start date (and now due date) gets set to the grades released date whenever the dates are saved.

KevinMackenzie:
I have the dates greyed out instead of being hidden so the error messages will still show up for debugging.  If you paste `$('.electronic_file_dates input').prop('disabled', false);` Into the web-inspector console, you will be able to edit these dates as you wish (and still have real-time updates).

bmcutler:
"greyed out" looks quite different on the different browsers
screenshot on mac left=firefox middle=safari right=chrome

On chrome this is quite bad/frustrating, those dates are not visibly different, you just can't edit them.  :(  
On safari it's barely visible.

screen shot 2018-08-02 at 8 56 48 pm

KevinMackenzie:
That's weird that the `disabled` styles look so different  I added a style so it should always get a gray background when disabled

KevinMackenzie:
I need to update the migration file before this gets merged to remove the NOT NULL constraint on down and re-add it on up

KevinMackenzie:
Don't merge this before #2560

Issue #2384, opened 2018-07-11
Access function integration


CouleeApps:
After #2337 I need to convert the other controllers to use the Access class.

bmcutler:
status?

CouleeApps:
Low priority at the moment. The rest of the codebase is basic "getGroup() == Instructor" checks that would be pretty easy to add to Access but would just take a while.

bmcutler:
closing, can be re-opened with more specific instructions & files that need refactor, etc.

PR #2628, opened 2018-07-31
Access functions: Checking file permissions

### LINKED ISSUE #2258: Grader can read arbitrary files on the server
[This "temporary hack"](https://github.com/Submitty/Submitty/blob/master/site/app/controllers/MiscController.php#L60-L66) (since 9 months ago) allows instructors/tas/graders to read any file on the server (that php can read). If you have a vagrant running, enjoy this [example](http://192.168.56.101/index.php?semester=s18&course=development&component=misc&page=display_file&dir=results&file=test01_student_file.pdf&path=/etc/passwd).

### LINKED ISSUE #2582: Student Permissions for Course Materials Popup
Student cannot click on the button to open a file in a separate window. It displays a permissions error for the uploads folder.

CouleeApps:
Closes #2582 and closes #2258, part of #2384

This adds checks for files and if a user has access to the gradeable (if given) and directory for various files. Should be decently expandable if we make additional directories in the future.


site/app/libraries/Access.php
```diff
@@ -458,4 +594,121 @@ private function resolveNewGradeable($g) {
...
+                    }
+                    $args["gradeable"] = $this->core->getQueries()->getGradeableConfig($value);
+                    break;
+                case "submitter":
+                    $submitter = $this->core->getQueries()->getSubmitterById($value);
+                    if ($submitter !== null) {
+                        $args["submitter"] = $submitter;
+                        if (array_key_exists("gradeable", $subparts)) {
+                            //If we already have a graded gradeable in the args, make sure this file
+                            // actually belongs to it
```
KevinMackenzie: Does this comment mean that we check that the file is a submission file / results file and we need to check that the student is allowed to access it?  I can't figure out where that might be happening below.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2524,6 +2554,21 @@ public function getGradedGradeable(\app\models\gradeable\Gradeable $gradeable, $
         return null;
     }
 
+    /**
+     * Gets a single GradedGradeable associated with the provided gradeable and
+     *  submitter.  Note: The user's team for this gradeable will be retrived if provided
+     * @param \app\models\gradeable\Gradeable $gradeable
+     * @param $submitter|null The submitter to get data for
```
KevinMackenzie: `@param Submitter|null $submitter`

KevinMackenzie:
You mentioned it in-code in a comment, but I'll restate it here.  As of now, showing the file link or not is our access control to students for course materials.  Clicking a link from a stale page will still grant students access to the files

site/app/controllers/MiscController.php
```diff
@@ -91,7 +91,7 @@ private function checkValidAccess($is_zip, &$error_string, $download_with_any_ro
         // END HACK
 
 
-        $possible_directories = array("config_upload", "uploads", "submissions", "results", "checkout", "forum_attachments", "uploads/course_materials");
+        $possible_directories = array("config_upload", "uploads", "submissions", "results", "checkout", "forum_attachments", "course_materials");
```
KevinMackenzie: Removing the 'uploads' prefix here causes the issues with `checkValidAccess`.  Because of a hack above `//END HACK`, all graders automatically pass this check, so only students would get this far anyway, and they shouldn't have anything other than read access.  Since this function will be killed off, its not a big deal, just worth noting.

site/app/libraries/Access.php
CouleeApps: The thought was that you could pass both a gradeable and a file to `canI`, and this would make sure that file was associated with that gradeable. Not sure how necessary it is but it might prevent some weirdness if you try to access a file across gradeables.

site/app/controllers/MiscController.php
CouleeApps: Removing the `uploads` here is along with removing it from all the other scripts too. Best to not reveal too much about the directory structure (todo eventually: relative paths)

CouleeApps:
Now covers writing/deleting files and checking if students can view/download/any version. Also this means i was able to delete `checkValidAccess`

PR #2631, opened 2018-08-01
PDF can now zoom 

saxocellphone:
It's a bit slow since it has to re-render the pages, otherwise it works great. There is a problem where if you zoom and scroll fast immediately, the rendering will begin to break. To fix it, one just has to zoom in/out again. I don't think that this will be a huge problem, but it might be a bit annoying.

### DIFF
--- a/site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
+++ b/site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
@@ -4,7 +4,6 @@
 {#<script src="https://cdnjs.cloudflare.com/ajax/libs/pdf.js/2.0.550/pdf.js"></script>#}
 {#<script src="https://cdnjs.cloudflare.com/ajax/libs/pdf.js/2.0.550/pdf_viewer.js"></script>#}
 <script src="js/pdf/pdf-annotate.js"></script>
-<script src="js/pdf/initColorPicker.js" type="module"></script>
 {% if student_popup is defined %}
     {#This is currently a hack to get the javascript working. There are
       some duplicate codes in the JS but that's for later#}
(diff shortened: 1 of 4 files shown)


holzbh:
On a touch screen on Firefox, touching is detected and if I drag my mouse around I can see a faint gray trail (no lines are made). If I then move my mouse without clicking, it acts as though I'm holding the mouse down even though I'm no longer touching the screen. Left clicking with the mouse ends this drawing behavior. Any lines added by this weird touch-screen-then-move-cursor-with-touchpad behavior disappear when I zoom. Any lines drawn using the traditional touchpad-only method persist.

In Chrome, the touch screen works as expected. I did not need to play with any browser flags. Touch screen drawings remained even when zooming while in Chrome.

Using a touchpad seems fine still in both browsers. Zooming seems to work correctly.

If my browser window isn't big enough, in Chrome, touch screen moving vertically will scroll instead of drawing. If I try to draw an L shape in this situation, the horizontal part of the L is drawn, but the vertical part does not appear if there was scrolling. This only happens with the touch screen.

In Firefox, with the scrolling situation, touch screen scrolls AND draws though it's a janky sort of sinusoid as the cursor position jumps relative to the document due to the scrolling, so you get giant spikes from the bottom of the screen to the top of the screen. These lines DO remain even after zooming.

saxocellphone:
@holzbh Hmm it seems like that firefox and chrome's behavior differs significantly. I think this is due to the fact that I use completely different event listeners for the two browsers. I made this decision because different event listeners do different things for each browser. It will be hard to test any changes I make since I don't have a touchscreen device. It will be helpful if I can get my hand on the device and do some further testing. 

As for the scrolling situation, I definitely want to see for myself what's going on because I thought I fixed that bug, granted I was testing on the tablet.

holzbh:
After talking with @bmcutler, suggestion is to add a UI element to enable/disable scroll instead of the potential solution of auto-disabling scroll when using drawing tools. That can come in a future PR, so approving this one.

PR #2634, opened 2018-08-01
[Initial Work] New Gradeable model integration into the ElectronicGraderController

KevinMackenzie:
The first steps towards getting the ElectronicGraderController and its views to use the new Gradeable Model.  This includes saving marks and assigned marks, viewing autochecks and diff-viewer output.  Also added some helper methods into the Abstract controller to load model instances and render jsend responses upon error.


site/app/libraries/Access.php
```diff
@@ -72,6 +72,8 @@ class Access {
     public function __construct(Core $core) {
         $this->core = $core;
 
+        $this->permissions["rubric.electronic.save_mark"] = self::ALLOW_MIN_LIMITED_ACCESS_GRADER | self::CHECK_GRADEABLE_MIN_GROUP;
```
CouleeApps: Use grading.electronic.add_one_new_mark

CouleeApps:
That was fast

PR #2475, opened 2018-07-18
Verify graders Take 3

jchulton:
The last Verify graders reverted 3000 lines in its second commit, so I decided just to make a new pull request on a new branch

PR #2645, opened 2018-08-02
Converted last non-view routes in Electronic Gradeable Controller to new model

KevinMackenzie:
All of the non-html-returning routes now use the new model.  This includes all ajax-type requests and all post-requests (form submission with redirection).  The only exceptions to this is grader verification, which will be done after #2475 gets merged, and pdf routes (which are still heavily under development anyway)

A list of things to test:
* Team creation as student
* Team creation through admin team form
* Team import
* Team export
* Electronic rubric grading
* Electronic rubric rubric modifications (mark addition/deletion)
* Electronic rubric "who got mark" popup


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1594,22 +1590,42 @@ public function getAllElectronicGradeablesIds() {
...
-     * @param string $g_id
+     * Generates a new team id from a user's name
      * @param string $user_id
-     * @param integer $registration_section
-     * @param integer $rotating_section
-     * @return string $team_id
+     * @return string
      */
-    public function createTeam($g_id, $user_id, $registration_section, $rotating_section) {
+    public function generateTeamId(string $user_id) {
```
CouleeApps: Sort of minor, but generating team ids and creating them in another function opens you to race conditions... would highly recommend combining the functions

CouleeApps: Or at least don't make this one public

KevinMackenzie: that's true; i'll work on it.  Its part of handling the issue of 'something threw an exception, but the teams were created already.'

CouleeApps:
Mark stats are broken slightly:

CouleeApps:
Adding a new mark and immediately changing the value will cause a conflict popup to show when closing the component. To reproduce:
1. Add new mark, give name etc
2. Before closing component, change mark value
3. Close component

CouleeApps:
Instructor (with no assigned sections) does not default to showing all, instead shows "no grading to be done"

CouleeApps:
Trying to create a team from the index page throws an error:
```Team may not have been properly initialized: SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint "teams_pkey" DETAIL: Key (team_id, user_id)=(00110_baliss, baliss) already exists.```
But still made the team:

CouleeApps:
Moving a team across reg sections throws an error and then a success... but it works

CouleeApps:
Student team forming page works great 👍

KevinMackenzie:
Like we discussed, it will perform the team operation successfully in the database, but when accessing the file system it may have issues and will appear as an error.

CouleeApps:
Team import / export works 👍

CouleeApps:
Attempting to grade a team homework as a limited access grader gives me the following error:

CouleeApps:
Moving a team across rotating sections says it created a new team??

KevinMackenzie:
just fixed that in 9f9a631

KevinMackenzie:
I think the instructor with no assigned section is this: #2651

CouleeApps:
Aahhh yes. Also nice ninja committing

CouleeApps:
Only thing left is that "creating a new team does not set reg or rot section"

PR #2653, opened 2018-08-03
Prevent users from accessing courses

### LINKED ISSUE #2644: Students can still access pages for archived courses
We have only removed it from the main page.
So if they edit the url they can still get to the page.

Also, we should make sure that students in the NULL section of a course cannot access the page if the edit the url (even if the course is non-archived.)

ropern123:
Prevents users from accessing courses if they are not an instructor and either are in the null registration section or if the course is archived.

closes #2644


site/app/libraries/Access.php
```diff
@@ -82,6 +82,8 @@ class Access {
     /** If the current set of flags requires the "gradeable_version" argument */
     const REQUIRE_ARG_VERSION           = 1 << 27;
 
+    /** If the current set of flags requires that the course status be checked */
+    const CHECK_COURSE_STATUS           = 1 << 31;
```
CouleeApps: Move this up to be with the other checks

site/app/libraries/Access.php
```diff
@@ -106,6 +108,8 @@ class Access {
     public function __construct(Core $core) {
         $this->core = $core;
 
+        $this->permissions["view_course"] = self::ALLOW_MIN_STUDENT | self::CHECK_COURSE_STATUS;
```
CouleeApps: Change to `course.view` to be consistent with how the scope stuff works

site/app/libraries/Access.php
```diff
@@ -82,6 +82,8 @@ class Access {
     /** If the current set of flags requires the "gradeable_version" argument */
     const REQUIRE_ARG_VERSION           = 1 << 27;
 
+    /** If the current set of flags requires that the course status be checked */
```
CouleeApps: Should document that this does not apply to instructors

ropern123:
users who are locked out of a course cannot log out on that page

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2088,6 +2088,11 @@ public function getArchivedCoursesById($user_id, $submitty_path) {
         return $return;
     }
 
+    public function getCourseStatus($semester, $course) {
+        $this->submitty_db->query("SELECT c.status FROM courses c WHERE c.semester=? AND c.course=?", array($semester, $course));
```
MasterOdin: No point having aliases if doing a select from a single table.

Issue #2596, opened 2018-07-29
Forum: mailto icon on all posts & thread scroll position


bmcutler:
Please add the mailto icon/email address on all posts (not just the first post in the thread).

And when you click on a new thread, please maintain the current scroll/position of the list of all threads.  It seems to jump back up to the top (or near the top) of the thread list.

PR #2654, opened 2018-08-03
Forum: Remember position of scroll in thread list

scopeInfinity:
Issue: #2596

- [x] Thread list scroll as the bi-directional infinite scroll.
- [ ] Remember the scroll position OR jump to thread page containing current thread.


andrewaikens87:
It seems like I am able to scroll up when I shouldn't be. To recreate, I scrolled down a bit and then scrolled up and continued to do so, it will continue to load the same block of threads over and over. Also my browser is set to 100% zoom (default) and the go to bottom button is way below where it should be. I must set the zoom to 150% to have it in the correct location.

bad

PR #2655, opened 2018-08-03
Vagrant pdf build

saxocellphone:
This builds the pdf-annotate.js code and copies it to the site/js/pdf folder.


.setup/install_system.sh
```diff
@@ -378,6 +378,13 @@ EOF
     chmod 0640 /etc/apache2/suexec/www-data
 
 
+	#############
+	#NPM install
+	#############
+
+	curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
+	sudo apt-get install -y nodejs
```
MasterOdin: you don't need sudo, the script has to be run as root.

Though, please convert the tabs to spaces.

saxocellphone:
It should be working now but someone else should run a vagrant up just to make sure

saxocellphone:
Now that I took a look at how other people do version releases, I think there might be a better approach. Instead of building it when installing submitty, we should just copy the file in the /dist folder in the other repo. When we want to update the code, we can just make a new release in the other repo and update the version number.

MasterOdin:
The reason for the `*.js.map` file is to map the compiled source to where you would look in the uncompiled version for debugging purposes. As such, there's no real reason not to just always use the `*.min.js` version over unminified, so I removed that logic.

I also haven't really tested the PDF functionality, so it'd be worth confirming things still work after my changes.

bmcutler:
@saxocellphone Can you please resolve conflicts?

PR #2663, opened 2018-08-06
CSS Styles cleanup

CouleeApps:
Doing a few things here:
- Fix textboxes being squished
- Fixed grading boxes being 100% height
- Add bootstrap-grid for better 2-column layouts (and allows us more flexibility in the future)
- Replacing the wonky 48% "half" css with 2-column (col-md-6)
- Cleaning up the extra boxes around custom mark
- Some inline css moved into css files, diff viewer, header/footer, rubric


bmcutler:
Can you adjust the width of left (too small) vs right (too big) on the instructor rubric UI 
the publish checkbox & delete icon are wrapping  :(

screen shot 2018-08-07 at 11 31 23 am

bmcutler:
Also, Discussion Forum code snippet boxes are now too tall.

Issue #2666, opened 2018-08-07
late days table shows non-visible assignments


bmcutler:
Late days table should not show future/pre-beta assignments to full/limited access graders
Late days table should not show future or only-TA or cannot view assignments to students

bmcutler:
this issue may already be fixed?  @LeonMontealegre  please test and close or submit a PR :)

bmcutler:
yes, this is closed.

PR #2672, opened 2018-08-07
Hide gradeables you aren't allowed to see in the late days view

CouleeApps:
Addresses #2666
But leaving that open because there's a big todo in this code.


bmcutler:
I'm going to merge this but leave the issue open.

PR #2697, opened 2018-08-08
New UI Design

CouleeApps:
The current site theme is pretty dated and the long lists of buttons make it difficult for people to find and perform actions. Here is the proposed new interface:

Full size images are available here: https://submitty.party/new-ui/


CouleeApps:
Looks like the only failing tests on selenium are forum-related and apparently happen in other branches. The rest of them should work though.

bmcutler:
Discussion Forum & Plagiarism Detection are 2 pages that I think should have the side bar auto collapsed.  

I don't like how the icon size changes when the bar collapses,  It's distracting and I have to move my mouse to re-press collapse/uncollapse.

Just having the first name at the top of the left column looks weird.  Yes it's helpful for development, but it's not what I expect as a user.  The "Hello, Quinn" was ok, or could remove it from the top since it doesn't have a link/action?  The logout button could say "Logout Quinn".

bmcutler:
Requested changes in Issue #2725, address in a (soon!) future PR.

PR #2713, opened 2018-08-10
Deleted big lib files

saxocellphone:
After running uglifyjs we don't need the big pdfjs lib files anymore.


bmcutler:
@saxocellphone Please resolve this conflict

PR #2732, opened 2018-08-16
Add ability for instructor to specify sidebar links

### LINKED ISSUE #2719: Add ability for instructor specified sidebar links
We should add the ability for instructors to create their own sidebar section with their own links. 

For a first draft, we should allow creation of a new json file (`sidebar.json`) that has the following format:
```
[
    {
        "icon": "string for fontawesome icon",
        "title": "Name of Link",
        "link": "http://link_to_resource/"
    },
    ...
}
```

For the icon, we should allow instructors to specify just the name of the icon (ex: `question`) or the full name (ex: `fa-question`). A simple detection of just "does the icon start with `fa-`, and if not, prepend `fa-`", should handle display of the icons. If the icon is left blank, default to some icon (`question`)?

A UI to make it easier to handle creation and writing of this json file will handle after the initial PR implementation.

MasterOdin:
Closes #2719

Instructor can create a file in the config directory in their course (`/var/local/submitty/courses/<semester>/<course>/config`) to generate additional links on the sidebar.

Using the following JSON:
```
[
  {
    "icon": "calendar",
    "link": "http://example.com",
    "title": "Calendar"
  },
  {
    "icon": "fa-github",
    "link": "http://example.com",
    "title": "Github"
  }
]
```
screenshot 2018-08-16 04 01 02

Do we want to have these sidebar links have `target="_blank"` or just have them act like the other links (in that clicking them will take you out of Submitty in your window)?


bmcutler:
not blank, don't force a new window.  I'm going to setup my external site to look very similar with the same sidebar so if you're not paying attention to the URL bar you'll think you're on the same site.

MasterOdin:
I think a further revision would be to allow an instructor to specify a target type (such as "_blank"), but for now, yeah, I'll leave it just as is which will act as a normal link and take your current page to whatever is linked.

PR #2742, opened 2018-08-19
Custom mark fix

### LINKED ISSUE #2678: Making a custom mark is broken/unintutive
Lots of confusion today.  

Clicking on the custom mark checkbox or points or text area was closing the mark instead of allowing use of custom mark.

I don't think it was because they had "Enable editing mode" enabled.

jchulton:
Sorry for delay on this.
closes #2678 
Each part of the custom mark can now be set independently (you can set the score, give a comment and check the mark independently). Trying to save an invalid custom mark (one where the score is non-zero and there is no comment), will cause an alert error.

Note that when saving a mark on page renavigation (refresh or going to next/prev student) an alert error would appear, but the mark would save properly. I've suppressed the alert, but I would apprentice it if someone could take a look at this.


migration/data/course_tables.sql
```diff
@@ -260,6 +260,7 @@ CREATE TABLE gradeable_component_data (
     gd_id integer NOT NULL,
     gcd_score numeric NOT NULL,
     gcd_component_comment character varying NOT NULL,
+    gcd_has_custom boolean NOT NULL,
```
bmcutler: Why is this necessary?

jchulton: To see if the custom mark is checked. Before, if the custom mark had no comment, it was treated as unchecked, and would automatically uncheck if the comment was removed. gcd_custom brings the structure of a custom mark closer to that of a regular mark.

bmcutler:
will be in another PR

PR #2743, opened 2018-08-19
Grader override fix

### LINKED ISSUE #2049: functionality & documentation of "Overwrite Grader" checkbox
This was solved with #903, but it looks like some recent changes to how the assigned marks get saved has caused the "Overwrite" feature to stop working.  No matter what the "Overwrite Grader" Checkbox state is, the data always gets saved to the server.

---------------------------

EDITED by @bmcutler

This checkbox should only be displayed for instructor users.
By default any editing to a component should replace the grader in the database.

Leave the text as "Overwrite Grader".  (Short is good)
It is only visible to instructors.
By default it is checked.  (edits replace the grader).
If unchecked, and previously graded, edits will not change the grader.

To close this issue:
- [ ] rename to "Silent Edit"
- [ ] implement desired behavior
- [ ] write end-to-end test cases to confirm behavior (and avoid re-breaking in the future)
- [ ] document purpose of this checkbox in instructor section of submitty.org

### LINKED ISSUE #2694: Grading an already graded component doesn't override the previous grader
If a component is already grader by user1 and user2 regrades that component, the component will read graded by user1 after reloading the page

jchulton:
Fixes the issue of a grader not overrride an existing grader after regrading a component.
closes #2694 and closes #2049

### DIFF
--- a/site/app/controllers/grading/ElectronicGraderController.php
+++ b/site/app/controllers/grading/ElectronicGraderController.php
@@ -1135,8 +1135,7 @@ class ElectronicGraderController extends GradingController {
         $custom_points = floatval($custom_points);
 
         // Optional Parameters
-        $overwrite = ($_POST['overwrite'] ?? true) === true;
-
+        $overwrite = $_POST['overwrite'] ?? true;
         $grader = $this->core->getUser();
 
         // Get the gradeable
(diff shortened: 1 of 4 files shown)


bmcutler:
Can you also address 2694 and 2049 in this PR?

bmcutler:
@jchulton What is the status of this work?
Please resolve conflicts or close this and open new PR

Issue #2725, opened 2018-08-15
Small tweaks to UI


bmcutler:
- [x] Make left bar icons the same size when collapsed/uncollapsed (so they don't move vertically when toggling this option).  Probably a size in between the current sizes?

- [x] The hover text for "Collapse Sidebar" should be "Un-Collapse Sidebar" or "Expand Sidebar" when it is collapsed.

- [x] Should it be "Manage Students" and "Manage Graders" instead of just "Students" and "Graders" in the sidebar?  (group says yes)

- [x] It could also be "Manage Sections" instead of "Setup Sections".  (group says yes)

- [x] Make it just "Student Photos" (be consistent with "Course Materials").  I originally had requested that both of these are prefixed with "Upload" if the instructor hasn't uploaded anything yet, but that's too complicated/unnecessary.

- [x] Put edit gradeable pencil icon all the way on the left of each gradeable row on the home/navigation page.

- [x] Leave the delete gradeable icon where it is (after the gradeable name) but switch to a black trashcan.

- [x] The create/edit gradeable tabs wrap terribly on a middle/small width screen:
     Make sure to put line breaks in the right places.  And probably reduce the space between the tabs, and reduce the extra space on some of the tabs.

<img width="922" alt="screen shot 2018-08-15 at 1 10 44 pm" src="https://user-images.githubusercontent.com/5871417/44161903-a7730a00-a08c-11e8-8510-667cc635b192.png">
  
- [x] Replace all (in ui text) 'auto-grading' with 'autograding'

- [x] Several of the text boxes on the Course Settings page are too small.  When a sentence length string is expected use a bigger box.

Edited to add: 
- [x] After group discussion:  Cut the "Quinn" from the top of the list.  Instead make it "Logout Quinn" at the bottom.  A little awkward perhaps but then there are no entries in the sidebar without a link&icon, and developers still have a shortcut to remind them who they are logged in as.  We can change it later if someone has a better design.

- [x] If the instructor has specified an external course home page URL, that should be the first link and labeled "Home" on the side bar.  It should be the top link.  Maybe "Course Home" goes to the external URL (if provided, omit this from sidebar if it was not provided).  And "Submitty Main" or "Submitty Navigation" on the sidebar which goes to the navigation/home page.

bmcutler:
Oh, and I think e84848 might be a better warning/red color (more grey, at least to me looks better on the white/lighter background).

bmcutler:
[...]

PR #2747, opened 2018-08-20
Some UI tweaks

bmcutler:
Some items from #2725

### DIFF
--- a/site/app/templates/grading/simple/Display.twig
+++ b/site/app/templates/grading/simple/Display.twig
@@ -15,9 +15,9 @@
         </div>
     {% endif %}
 
-    <h2>
+    <h1>
         {{ gradeable.getTitle() }}
-    </h2>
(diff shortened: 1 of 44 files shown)


site/app/templates/grading/simple/Display.twig
```diff
@@ -15,9 +15,9 @@
         </div>
     {% endif %}
 
-    <h2>
+    <h1>
         {{ gradeable.getTitle() }}
-    </h2>
+    </h1>
     <p>
```
ropern123: This seems to be breaking the travis tests.
Change the "h2" in the click_nav_grade_button call in test_simple_grader (in the run_tests method) to an "h1".

bmcutler: Yup, that was it!  Thanks!

PR #2751, opened 2018-08-20
[WIP] Ta Grading Interface uses new model and Edit Gradeable shares code

### LINKED ISSUE #2048: Factor component table into a reusable script

### LINKED ISSUE #2049: functionality & documentation of "Overwrite Grader" checkbox

### LINKED ISSUE #2670: Delete mark icon should be a trash can not an X
On the instructor UI, and also on the TA UI.

...don't fight over who implements this :)

### LINKED ISSUE #2694: Grading an already graded component doesn't override the previous grader

KevinMackenzie:
Closes #2048 Closes #2670 Closes  #2694 Closes  #2049 (among others, will add later)

The ta grading interface uses the new gradeable model and the electronic file rubric editor in the edit gradeable page uses the same interface.

There are a few things that still need to be changed, but I want to get this reviewed soon, so I will add those in later.

This PR affects the behavior of the entire grading rubric and although it appears very similar and behaves very similar, internally it functions very differently.  This means that every detail of this interface should be tested to make sure it works as it did before (or better).

Instructor edit mode changes:
- [x] Add 'Add' / 'delete' component option to edit gradeable interface
- [x] Support changing 'publish' in edit gradeable
- [x] Automatically switch first mark text from no/full credit
- [ ] Peer checkbox on components
- [x] 'Total Score' box

Grade Mode (or both) changes:
- [x] Team Support
- [x] Verify Graders support (buttons exist, but do nothing)
- [x] Badges colors on components (weird twig.js issue)
- [x] Disable grading if the non-active version is being viewed
- [x] Get a better 'restore mark' fa icon (right now its `fa-shower`)
- [x] Load previously open component from cookie
- [x] Retain 'Silent Edit' checkbox
- [x] Add New Mark text is part of the sortable (and shouldn't be)
- [x] Changes aren't saved on page unload (and should be)
- [ ] Consistent fonts


jchulton:
minor thing. If you create a mark and reorder it then navigate to a different page all without saving, the reording of the mark while not save (the new mark will be at the bottom).

jchulton:
When Creating a new component, if you create multiple marks in the component, then reorder those marks, then refresh not all of the marks are added to the new component.

KevinMackenzie:
This is because the partially asynchronous call chain for saving a component doesn't get finished on page reload.  I'll look into that.

KevinMackenzie:
@bmcutler The `Silent Edit` and `Edit Mode` checkboxes are both off by default (typical for grading).  Do you want the `Silent Edit` state to be loaded from the cookie?  Note that the server only lets instructors silent edit

bmcutler:
Yes, please load silent edit state from a cookie.

KevinMackenzie:
My color choices for the Mark conflict resolution popup are awful, so if someone could come up with some better colors +/ styling, i'd appreciate that
screenshot

jchulton:
I can try to improve the styling. I can also work on verify graders and write an alert when saving an invalid custom mark if you like.

KevinMackenzie:
The custom mark handling I feel is fine since its responsive.  For the mark conflicts, I'm looking mostly just for what colors go better with submitty but are different enough to be distinguishable.

jchulton:
If we want to stick with the interface design style guide, Id argue we should make 'your version' blue (since you usually want to save your own mark), original (since you going back to a previous version, doesn't really carry well), and your server white (since you won't frequently want to discard your changes).

jchulton:
Getting this on accessing a student from grading details. Persists when navigating to a new student until a component is selected. 
strange error

KevinMackenzie:
Here is what the conflict resolution looks like now:
screenshot1
(it doesn't work in 100% of cases yet)

CouleeApps:
This ui is rather unintuitive. Here are a few suggestions:

- Delete the mark id label, it doesn't give any useful information
- Make form title "Mark Conflicts: MARKNAME" instead of having name in the body
- Use radio buttons and have the action be "resolve"
- Primary action button should always be in the corner

Here's a quick mockup:
thispopup

KevinMackenzie:
Thanks, I'll work on that. The pictures I've shown don't show this, but
this pop-up is actually for resolving all mark conflicts for a component
simultaneously.

bmcutler:
Radio buttons + buttons at the bottom require two clicks.  (And a choice of default for the radio button which is not clear in this case).

Please re-read my suggestion above, including the order of options.  It only requires 3 buttons, 2 blue, and the 2 blue are the equally likely good choices.

Submit & close are unnecessary.  Close should not be an option in this case.  You must pick something.  If anything in the future we can make the my edits boxes editable if you want to manually merge.

KevinMackenzie:
How does this look?
screenshot
You are presented conflict resolution options for each conflict one after another.  The changes are made immediately after pressing the button and the popup closes after you resolve all marks

bmcutler:
The order I proposed is chronological, and less confusing I think.  (I showed this to Evan)

Your earlier example with mark delete brings up another complication.  
If a mark has been deleted but I have edited it, my edits should be saved (the mark should be 'undeleted' if possible or a new one created if necessary.)  In any case we don't even need a dialog for this.  My edits should win automatically.  (We can re-delete it later if necessary)

But also, if I'm in grading mode and I use a mark (first time use) but the mark was deleted while I was grading, the mark should be undeleted (if possible if the mark_id can be resurrected) or a new mark is created with that data.  Also in this case it shouldn't be a popup dialog box situation, the mark is just reborn.  (And can be deleted later if necessary, after un-marking that box for this student)

KevinMackenzie:
In the case that the mark is deleted from the server during an edit operation, I feel that, while the user _may_ always choose to 'undelete' the mark, it serves to inform the user that someone _did_ delete it.  In the case of a mark being deleted during a grade operation, it seems weird to be making changes to the rubric when you aren't in 'edit' mode.  Ultimately, I don't think we would benefit from assuming what the user wants to do in these corner cases.

bmcutler:
Add new component should be blue button
default component should have a title
default component title should be "Problem 1", "Problem 2", ...  (can discuss wording...)
triple dot reorder needs a tooltip

When component expanded, put a little more space between the badge &
bold title and the editable component title box.

explicit labels on left side of 3 textboxes "Title:", "Note to graders:", "Note to students and graders:"

labels on the left for the points
"Points:"
"Additional Extra Credit Points:"
"Negative Penalty Points:"

to the right of the points boxes for the extra credit & penalty, initalics 
"Student scores will be clamped from above to Points + Additional Extra Credit Points"
"Student scores will be clamped from below to min(0,Negative Penalty Points)"

"Count Up" -> "Grade by Count Up (from zero)", "Count Down" -> "Grade by Count Down (from maximum Points)"
needs to be a radio button.  It isn't 100% clear which one is selected/unselected.

Add new mark should be a blue button
"No credit"/"full credit" should not have a pubish button (should be false always)

"Publish" -> "Show mark to all students"

horizontal bar between count up/down & no credit/full credit is
currently darker/more visible than the boxes around each component.
It should be the other way around.  The visual separating the
component should be more pronounced.

KevinMackenzie:
I think I addressed everything you asked for.  The only thing I changed from the request is that the wording 'maximum points' is confusing since it refers the full-score point value, and not full-score + extra credit.  My suggestion is to change `maximum points` to `"Points"` since the label of the input box is `Points` (in the code).

The dividers are now the same color as the border, but since they don't extend to the edge of the container, it doesn't appear to be too disruptive.  If they were any thinner, they wouldn't function as dividers as effectively.

bmcutler:
Yay!  Looks alot better.  I hope others find it more intuitive too.  (I've been staring at it a while now so its familiar.  We need more testing by fresh eyes.)

I have not done any testing of the grading interaction / looking for race conditions yet.

KevinMackenzie:
The notable race condition is with two graders simultaneously grading the same component.  It would be possible to make conflict popup for that case, but it also wouldn't be difficult to add in after-the-fact.

PR #2765, opened 2018-08-25
Date parsing for new gradeable accepts years > 9999

KevinMackenzie:
The default php date parsing method does not accept years > 9999, so this new method accepts years with 5 digits, but is stricter.  It accepts the format the database returns and the format from the Edit Gradeable interface.  For good measure, I changed 'infinity' to be towards the beginning of the year 9999 so this is less of a potential issue in the future for code that doesn't use this date parsing method.

The method requires years, months, days, hours, minutes, seconds, and optional sub-second and time zone.

### DIFF
--- a/site/app/libraries/DateUtils.php
+++ b/site/app/libraries/DateUtils.php
@@ -5,7 +5,6 @@ namespace app\libraries;
 use \DateTime;
 use \DateTimeZone;
 use \DateInterval;
-use http\Exception\InvalidArgumentException;
 
 /**
  * Class DateUtils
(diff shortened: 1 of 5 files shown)


MasterOdin:
Hence the comment of completely disallowing that as well, assuming that this is then the quick fix and that's the longer term fix.

KevinMackenzie:
Yea.  Like you mentioned, we should have a db constraint to prevent dates from going that high in addition to php checking this constraint.  Those constraints should have gone into the `clamp_dates` migration

site/app/libraries/DateUtils.php
```diff
@@ -82,21 +82,52 @@ public static function validateTimestamp($timestamp) {
...
+
+        $ms = $matches[7] ?? '';
+        $timezone = $matches[8] ?? '';
+        if ($ms === '') {
+            $ms = 0;
+        }
+
+        $date = new \DateTime('now', ($timezone !== '') ? new DateTimeZone($timezone) : $default_time_zone);
+        $date->setDate($matches[1], $matches[2], $matches[3]);
+        $date->setTime($matches[4], $matches[5], $matches[6], $ms);
```
MasterOdin: This function throws a warning on PHP 7 as the millisecond parameter (fourth argument) was added in 7.1. Given that there's no real point to setting things at the microsecond level as we don't care about anything at that scale, no point worrying about it.

MasterOdin:
It looks like with this, if I change the due date, I have to then rebuild the course for the submission page to show back up and it's not obvious why that is.

To recreate:
1. Go to sample -> Open Homework and go to edit page. 
1. Change due date to "now".
1. Go to submit page
1. page doesn't load with a PHP notice and a message about how assignment hasn't been built yet

PR #2780, opened 2018-08-28
Fix checking out another's repo using grading permissions

MasterOdin:
Fixes allowing instructor to checkout anyone's repos. The authentication process should be tied more into the Access library that Glen wrote at some point for better checking of "is this something you're allowed to see based on having to grade it?", but this at least keeps the functionality equivalent to what it was last semester.

I think the next step for this is to create proper unit testing of all conditions and then go from there for using the access class.


bmcutler:
yay! it works.  @MasterOdin please make new issue(s) as you note above.

PR #2788, opened 2018-08-29
Routerless Mode

### LINKED ISSUE #2750: Enable Networked Programs to Run in 'Routerless" Mode
The Router is a good tool for inserting message delay, loss, and for monitoring student connection types (tcp vs udp). However, if none of these factors are being tested, it represents an unnecessary bottleneck. To this end, we should include an option to turn the router off.

Notes from meeting with Barb:
1. The router should be on by default.
2. Since router output can be turned off or on, it should be explicit about what messages are being intercepted and sent/received, so that it output will be useful for student debugging.

emaicus:
Closes #2750

Adds the ```use_router``` boolean to the instructor configuration file globally and per testcase. If ```use_router``` is false, all containers are added a single network, the router is not injected, and messages are not logged. ```use_router``` is true by default.

### DIFF
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -51,6 +51,12 @@ void AddDockerConfiguration(nlohmann::json &whole_config) {
   }
 
   bool docker_enabled = whole_config["docker_enabled"];
+
+  if (!whole_config["use_router"].is_boolean()){
+    whole_config["use_router"] = true;
+  }
+
+  bool use_router = whole_config["use_router"];
   
   nlohmann::json::iterator tc = whole_config.find("testcases");
   assert (tc != whole_config.end());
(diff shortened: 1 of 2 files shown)


grading/load_config_json.cpp
```diff
@@ -51,6 +51,12 @@ void AddDockerConfiguration(nlohmann::json &whole_config) {
   }
 
   bool docker_enabled = whole_config["docker_enabled"];
+
+  if (!whole_config["use_router"].is_boolean()){
+    whole_config["use_router"] = true;
```
MasterOdin: shouldn't be this be setting it false if use_router isn't a boolean/doesn't exist?

MasterOdin:
Do you mean to say that if `use_router` is true, then then containers will have a router and messages be logged and then then `use_router` is false, the containers are all on one network and don't have a router? The description you gave seems counterintuitive.

PR #2805, opened 2018-09-03
Debug timezone for open/due/etc action buttons & forum

### LINKED ISSUE #2762: Dates on the gradeables page buttons are wrong
I'm seeing "Due Date 03:59", when the date on edit gradeable is 23:59.  It's correctly showing 23:59 on the submission page.

Seems to be ignoring the 4 hour timezone offset.

### LINKED ISSUE #2786: Dates on forum posts look buggy (appears to be UTC rather than timezone adjusted)
Probably related to #2762 
Happening on our production server.
timezone is correctly set to America/New York.

(cannot reproduce this bug on my vagrant box)

I posted this ~ 2pm
screen shot 2018-08-29 at 5 06 34 pm

bmcutler:
Ugly, but it 
Fixes #2762 and fixes #2786 
However I'm not at all sure what it does when the dates for an assignment cross a daylight savings change.

Core problem seems to be that the database(s) have their timezone set to UTC.
See discussion in #2786

### DIFF
--- a/site/app/controllers/forum/ForumController.php
+++ b/site/app/controllers/forum/ForumController.php
@@ -814,7 +814,8 @@ class ForumController extends AbstractController {
             $users[$user]["posts"][] = $content;
             $users[$user]["id"][] = $posts[$i]["id"];
             $date = date_create($posts[$i]["timestamp"]);
-            $users[$user]["timestamps"][] = $function_date($date,"n/j g:i A");
+            $my_timezone = $this->core->getConfig()->getTimezone();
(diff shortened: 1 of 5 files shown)


MasterOdin:
We'd probably want to move the setTimezone commands into the Gradeable Model so that it's stored within the model.

I also don't view the fact the DB stores thing in UTC as that big of an issue, but the whole point of the timezone config value is so that we don't have to care really what timezone any of the components are in. Also, by always explicitly setting the timezone, it will properly handle things within DST vs non-DST as it'll handle setting it to the appropriate offset regardless of what actually got saved in the DB/fie/whatever.

PR #2824, opened 2018-09-06
Null char and smart quotes are now supported

### LINKED ISSUE #2485: Add 'long dash' and 'smart quotes' to list of displayed characters
In the student vs. expected autograde output viewer

saxocellphone:
partial work for #2485. Also made code look better.


bmcutler:
What about long dash?  em dash & en dash?

saxocellphone:
Yes I think that's probably a better idea. I also think it's better if we separate special characters and non-printing characters. I did a google search and non-printing characters are defined as 

- White spaces
- Carriage returns
- Tabs
- Line breaks
- Page breaks
- Null characters

I think I got them all (except page break, I'll add that). I also think we can represent special chars in a different way, because they are already being displayed.

It's also kind of weird is not working for that particular student's case, because I tested null character on my computer and it's being displayed. Can you maybe send a screenshot?

bmcutler:
The screenshot shows the same thing as the original student complaint.
I think I have correctly installed your code and tested it on the original student file.  
(I may have made a mistake, but I don't think so.)
When I said non-printing I meant everything that does not currently display.  That is, all the weird characters too.  So I guess I meant all non-printing and all special characters.

saxocellphone:
I did some testing on grade_released_homework, the special characters (non-ascii) display already, but in escaped form.  Non-printing characters like the ones I listed above don't show up at all. 
capture 
So special characters do show up, but in weird forms. I also think it depends on how character encoding is being handled. I think for this PR we should just focus on non-printing characters. It is weird that it is not showing up for that student, but I'll look into that.

bmcutler:
I think the problem is I cannot view the autograding from the students earlier version
Issue #2846

bmcutler:
ok, now that I can load the prior submissions in grading view, I can confirm that it displays the weird characters.  Unfortunately, only after clicking on the "visualize button".  If a student doesn't know to try this, it will still be very confusing what is wrong.  These weird characters should appear in the original view too.  So we can highlight them in red.

screen shot 2018-09-16 at 10 04 13 pm
screen shot 2018-09-16 at 10 03 57 pm
screen shot 2018-09-16 at 10 04 07 pm

bmcutler:
merging this partial work.

PR #2832, opened 2018-09-08
Fix autograding score for version list

### LINKED ISSUE #2416: Submission page only shows autograding output for active version
The autograding section of the homework view only shows the results for the active version, regardless of which version is selected in the above `select` box.  It should load the autograding results from the selected version.

### LINKED ISSUE #2785: The autograding score for the different submissions versions is often wrong
The numbers in the dropdown are (sometimes) wrong
I don't see a pattern in what's breaking.
screen shot 2018-08-29 at 12 24 57 am
screen shot 2018-08-29 at 12 25 09 am

shailpatels:
Fixes autograding scores in the drop-down menu for versions. Previously the array wasn't sorted correctly and versions would show a score for a different submission.

fixes #2785
fixes #2416

### DIFF
--- a/site/app/views/submission/HomeworkView.php
+++ b/site/app/views/submission/HomeworkView.php
@@ -450,6 +450,9 @@ class HomeworkView extends AbstractView {
             ];
         }, $auto_graded_gradeable->getAutoGradedVersions());
 
+        //sort array by version number after values have been mapped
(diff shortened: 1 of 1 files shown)


andrewaikens87:
This should fix the out of order versions

PR #2833, opened 2018-09-08
Better Electronic Gradeable Presets and Settings

### LINKED ISSUE #2676: File upload:  By student vs Scanned Exam
[ notes from meeting -- comments welcome after draft revision implemented ]

Instead of 3 radio buttons, have 4.  (backend is still 3 core types)

Gradeable Type:
O  Student will upload files
O  TA/Instructor will bulk upload scanned .pdf files
O  Checkpoints (simple data entry)
O  Numeric/Text (simple data entry)

_The difference between the first two types will be stored in a new database variable (I think it needs to be new)._  

On the autograding/submissions tab, a new question for that variable:

hide things below here under an "advanced options expander thing"

Will this gradeable be scanned and uploaded as bulk pdfs by the TA/Instructor?
O Yes  O No   
_(this will be initialized based on choice on first first page, but can be edited which will toggle the choice on the first page if we go back there)_

What can students do?

View the gradeable on the course home page?  O Yes  O No  O Yes, but only when grades are released
Make new submissions?    O Yes  O No
Download submission files?   O Any version  O Active version only  O None

_If "yes" to bulk pdf these default/reset to "When released", "No", "Active only"
If "no" to bulk pdf these default/reset to "Yes", "Yes", "Any version"_

_If/when "No" is selected for View the gradeable, then make submissions and download files are greyed out or(?) hidden (and maybe also set to "no" and "none")._

KevinMackenzie:
Closes #2676 

Added database column for this boolean flag and most of the requested edit gradeable UI changes to accompany this new field.  The change that isn't in this request has to do with the student view "...only when grades are released".  We currently have no way to make this distinction and, among other reasons, it should be added in a separate PR.


KevinMackenzie:
This also fixes the issue you mentioned where the preset value for student 'view' is false instead of true.

KevinMackenzie:
The failing tests appear to be timeouts on the forum tests (which I can't get to pass locally on master either for different reasons)

bmcutler:
@MasterOdin can you help with the travis failures?

KevinMackenzie:
@bmcutler I fixed the tests faililng

bmcutler:
@KevinMackenzie 

The UI change I'd like you to help finish...  Let's have 5 options for gradeable type on the new gradeable page (I've partially coded it up).  3 electronic variants: student upload, bulk instructor pload, or student VCS.  Can you make the "is this vcs" question go away and set it from the gradeable type selection?  (Basically have the popout with extra instructions and the VCS path appear for the gradeable type = VCS).

Also, I think we can abandon the option to toggle student upload & instructor bulk upload on the 2nd tab.  I also added the view "when grades are released" UI option, but it needs to be hooked up.

bmcutler:
@KevinMackenzie 
Please remove the "Electronic File (multiple types)" radio button.  No need to show that internal category to the instructor user.  It's just confusing.  Let's not ever use the term "Electronic File" on the UI, it's an internal thing only.

The "is this a team assignment" is not wrapping correctly, missing a newline?  Put some vertical space between the manual grade question & the team question.

screen shot 2018-09-18 at 11 39 57 am

Cut this  "(Note: The presets on "General" configure these values)" not necessary.

Also...  You should not be able to toggle the VCS type after creating the gradeable.  This isn't how it was before and is introducing a bug.   We do not need to switch between the electronic file types after initial creation (the only one that could be allowed is between student upload & bulk upload, and its not important enough to confuse the UI).  Just make it a choice of 5 types at the same level as I had before.  Locked in after creating the gradeable.

Once the gradeable is created, if you go back to "General", we've lost (or maybe never had) a spot that said whether it was a team or individual assignment (you can't change it, but it would be good to display it (like the unique id, and the gradeable type).

KevinMackenzie:
It looks like most the tests are failing because certain things are timing out.  I've run the suite locally and it sometimes fails, sometimes passes these tests.

bmcutler:
bulk upload, student permissions settings, download submission files should default to "active version"

PR #2834, opened 2018-09-09
Blank Config Repository won't traverse entire course directory

KevinMackenzie:
If the config repository is blank, php would attempt to traverse the entire course file structure since `<course-dir>/<repo-dir>` evaluated to `<course-dir>/`, which may be what is causing out-of-memory errors.  Now, this operation will only occur if there is a non-blank config repo name.


bmcutler:
Yes, this fixes it.  However I realize we aren't actually using/displaying the options in the config repo setting...  so I opened #2836

PR #2842, opened 2018-09-10
Fixes "View All"/"View Your Sections" issue

### LINKED ISSUE #2818: "View All" should also show the NULL section
When looking at the index/list of all students for a checkpoint, numeric, or full rubric grading, if I click on "view all" (only available to instructors & full access graders) it should show the students in the NULL section as well.  This feature has gotten lost more than once in our various rewrites/refactors :(

LeonMontealegre:
Fixed issue where View All and View Your Sections wouldn't work with SimpleGraderViews
Code was copied from the functioning ElectronicGraderView.

fixes part of #2818


LeonMontealegre:
Don't merge this yet - just realized it still isn't showing NULL section; so this is only half fixed

bmcutler:
@LeonMontealegre Make a new PR with the 2nd part :)

PR #2847, opened 2018-09-11
Autograding Results panel Uses New Model

### LINKED ISSUE #2846: TA cannot view autograding from other versions
(I think this was already an issue, but I cannot find it)
When the TA chooses another version (from the black bordered student information panel)
The autograding panel still shows the results from the active assignment (not the selected assignment).
If the TA changes the active version, then it updates correctly.  (but this is not an appropriate workaround)

KevinMackenzie:
Closes #2846 

This PR moves the autograding results panel in the TA grading interface to use the new model (similar to the submission view page).  Since it removes the last dependency on the old template for autograding results, that file is removed.


CouleeApps:
Rename AutoResultsNew.twig to AutoResults.twig ? Seems like the "new" is unnecessary now

KevinMackenzie:
I have a commit that does this, but am having issues pushing it.  Don't merge until I get that pushed.

PR #2877, opened 2018-09-18
Added "Pre Commands" to the assignment configuration.

emaicus:
Pre commands are defined at the testcase level, and are used to prep the testcase environment for a run. 

Pre commands take the form:
```
 "pre_commands" : [
        {
          "command" : "cp",
           "test01",
          "source" : "dir",
          "destination" : "./",
          "option" : "-R",
        }
```

* ```command``` is the action to be taken. At present only copy is allowed.
* ```source``` is the directory or file to copy. The copy is performed as ```submitty_daemon```. Some safeguards have been put in place, but additional security checking would be appreciated.
* ```destination``` is the directory into which the contents of source will be copied. This directory is __always__ assumed to be the directory of the testcase being executed. At the moment, this strips away a layer of directory structure. That is, if source is ```test01/mydir/``` and destination is ```.``` the contents of ```my_dir``` will be copied to ```.``` 
* It is not currently possible to set the source to be a base test directory do to a file permissions issue.
* Option will eventually be used, but is currently just assumed to be ```-R```.
* ```pattern``` will be added in a future PR.


emaicus:
Hold on merging this temporarily.

PR #2902, opened 2018-09-23
Forum block count increase and anonymous notifications

### LINKED ISSUE #2811: notifications might not be preserving anonymous
Reply: Your post "blah... " got a new reply from smithj (even if smithj checked the anonymous box)

andrewaikens87:
This PR closes #2811 and addresses the following:

- Increases forum threads loading block count to 30 from 10
- Anonymous forum notifications when a user selects 'Anonymous (to classmates)'
- Notifications now use same name convention as forum rather than their user_id (DisplayFirstName LastFirstInitial.) 
- Will only work for newly generated notifications after this PR, legacy notifications will not be updated if the user changes anonymous state
anonnotifications


bmcutler:
Why do we need a new field in the notifications table?  This information is redundant and could also be incorrect (if someone changes the anonymous setting after posting, this will not be updated).

andrewaikens87:
Good point, I have removed the column in the db and have implemented it with updating the notification when a user changes anonymous state. I am going to clean it up a bit and then push it tomorrow.

PR #2907, opened 2018-09-24
Vpat

pbailie:
View changes for accessibility that do not (significantly) alter the UI.


pbailie:
I will break this down into smaller pull requests

Issue #2800, opened 2018-09-02
icons & other visuals for forum


bmcutler:
The checkmark in a green circle ('resolved') is hard to see.  Let's just make it a green checkmark (no circle).  Similar to the "student has viewed grade" on index page listing details for all students in TA grading.

Similarly the ! in a yellow circle (unresolved) is hard to see.  Let's make it a question mark (no circle).  Yellow is good.

The thread list page up & page down overlapping the upper left & lower right corners of that thread list aren't well positioned.  They cover up the icons.  I think it would be better to not overlap and instead stretch that button to be wider.  Also, if I'm at the top of the list (or the top of the list has been loaded), that page up button should not appear.  And if I'm at the bottom of the list (or has been loaded), the page down button should not appear.  (or should be greyed out?)

Finally, the forum has a fixed height.  If I ctrl-minus on this page (to try to see more content, there's wasted space at the bottom of the page.

soulspark666:
Does this issue still need any work, or is it was reopened by some mistake??

bmcutler:
The forum height is still problematic.  @andrewaikens87  is working on that (it may not have a specific issue).

PR #2923, opened 2018-09-27
first version of changes for icons in forum

weiji-li:
Just to test whether my small changes can work.

### DIFF
--- a/site/app/views/forum/ForumThreadView.php
+++ b/site/app/views/forum/ForumThreadView.php
@@ -631,7 +631,7 @@ HTML;
 
 							} else {
 								$fa_icon = "fa-question";
-								$fa_color = "#ff8000";
+								$fa_color = "#ffcc00";
 								$fa_margin_right = "5px";
 								$fa_font_size = "1.8em";
 							}


VictorLi718:
I changed the icons, but I don't know which are the best so I just remove the circle around them and adjust the position. Also for the button for scrolling to top, I think change the opacity can help a lot(which is pretty easy to do). And when you hover on it, it will become 100% opacity and appear well.

bmcutler:
Can we use the same version of the green check we use on the viewed grade column (from the instructor page click on grade or regrade for any released assignment. 

screen shot 2018-09-29 at 7 05 37 am

Let's try the icons without a black border?

VictorLi718:
Ok, I will try. And what about the exclamation mark?

bmcutler:
We're trying to use font awesome for all of our icons.  

Maybe the check and x for viewed grade are not font awesome (that should be changed too), so maybe don't use specifically those shapes, but make them look similar (no black outline, and large enough to be clearly visible on the forum).

As I said in #2800, please switch the exclamation point to be a question mark (it is an unresolved question after all)

https://fontawesome.com/icons/question?style=solid
https://fontawesome.com/icons/check?style=solid

VictorLi718:
image
I make a change again, making it a bit larger and without black border. But maybe for question icon, yellow is not the best even if I use the same color as the border of selected answer in discussion forum

bmcutler:
Try an orange color for the question mark.  Maybe ff9933 or ff6600?

VictorLi718:
image
See this. I think it's better?

bmcutler:
Yes, I think that looks good.  But this is not what the code does yet...

VictorLi718:
Sorry, I just make a new push, with orange color. And I just see that maybe the "gold" color, which is used for the starred topic is also good.

site/app/views/forum/ForumThreadView.php
```diff
@@ -624,14 +624,19 @@ public function displayThreadList($threads, $filtering, &$activeThreadAnnounceme
...
+								$fa_icon = "fa-check";
+								$fa_color = "5cb85c";
+								$fa_margin_right = "0px";
+								$fa_font_size = "1.5em";
+
 							} else {
-								$fa_icon = "fa-exclamation-circle";
-								$fa_color = "yellow";
+								$fa_icon = "fa-question";
+								$fa_color = "ff8000";
```
andrewaikens87: Please include a # before the color hex.

site/app/views/forum/ForumThreadView.php
```diff
@@ -624,14 +624,19 @@ public function displayThreadList($threads, $filtering, &$activeThreadAnnounceme
                         }
 						if($thread['status'] !=0) {
 							if($thread['status'] == 1) {
-								$fa_icon = "fa-check-circle";
-								$fa_color = "palegreen";
+								$fa_icon = "fa-check";
+								$fa_color = "5cb85c";
```
andrewaikens87: Please include a # before the color hex.

VictorLi718: Sorry, my bad. I have '#' before color hex value now.

site/app/views/forum/ForumThreadView.php
```diff
@@ -624,14 +624,19 @@ public function displayThreadList($threads, $filtering, &$activeThreadAnnounceme
...
-								$fa_color = "yellow";
+								$fa_icon = "fa-question";
+								$fa_color = "#ff8000";
+								$fa_margin_right = "5px";
+								$fa_font_size = "1.8em";
 							}
 							$return .= <<<HTML
-							<i class="fa ${fa_icon}" style="padding-left:3px;position:relative; float:right; display:inline-block; color:${fa_color}; -webkit-text-stroke-width: 1px;
+							<i class="fa ${fa_icon}" style="margin-right:${fa_margin_right}; padding-left:3px; position:relative; float:right; display:inline-block; color:${fa_color}; font-size:${fa_font_size}; -webkit-text-stroke-width: 0px;
     -webkit-text-stroke-color: black;" aria-hidden="true"></i>
```
andrewaikens87: Please remove the stroke css instead of setting it to zero.

PR #2935, opened 2018-09-28
'Verify' button won't show up before closing component on first grade

KevinMackenzie:
When opening a component for the first time as a full access grader or above, clicking a mark to grade would cause the 'verify' button to appear.  Now, the 'verify' button won't show up unless another grader has already graded the component (as it should).


KevinMackenzie:
@bmcutler I could reproduce the issues you were encountering.  Unfortunately, some logic needs to be duplicated and one of the two spots that was doing some boolean logic had a typo, which was causing the inconsistencies.  Specifically, the message failed to show up on page load, but would show up after  opening / closing components.  c487ebf should fix this.

KevinMackenzie:
This commit was on the wrong branch

PR #2940, opened 2018-09-29
[WIP] Diffviewer monospace problem

### LINKED ISSUE #2863: Switch to a better fixed width/monospace font
Displaying autograding output (actual/expected) works ok with basic ascii, but other printing characters sometimes print with different widths (examples will be shared offline).  Chose another font that works for everything.

saxocellphone:
Closes #2863. This is a weird one. I literally just commented out the font-family css rule and everything seems to be in monospace. All the bang importants are a bit hard to find the root of the problem. Here is a before/after comparison.
Before
capture
After
capture

I tested a bunch of lines from the sample file provided, and they all seem to work. Do we still need to find a new font?

### DIFF
--- a/site/public/css/diff-viewer.css
+++ b/site/public/css/diff-viewer.css
@@ -64,7 +64,8 @@
     vertical-align: top;
     max-height: 20px !important;
     height: 20px !important;
-    font-family: 'Inconsolata', monospace !important;
+    /*Somehow this is causing bunch of problems*/
+    /*font-family: 'Inconsolata', monospace !important;*/
     font-size: 16px;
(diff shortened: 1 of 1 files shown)


MasterOdin:
I'm a little apprehensive of using just "monospace" as the font as that does mean Windows, Mac, and Linux have the possibility of all showing different fonts which may or may not have proper support for the given character outside ASCII.

saxocellphone:
I'm really confused why having any font will cause the spacing problem. I played around with disabling css rules that would interfere with the font, but nothing seems to fix it. The 'monospace' option doesn't seem to do anything

PR #2964, opened 2018-10-04
Prevent lab page crash on sort by last name

### LINKED ISSUE #2959: simple grading page sorting causes crash.
I am a mentor in RCOS and when I tried to grade people in my session, the submitty immediately crushed if I sort the students by their last name. Also when sorting by the first name, it's not working properly.
frogggyyyy

ropern123:
Fixes a bug where the lab page crashes when you try to sort it by last name.

closes #2959, as the sorting by first name appears to be working as intended (sorting by preferred first name if it exists, else the first name ).


ropern123:
This will be fixed by #2963

PR #2971, opened 2018-10-07
Fix Calculation in Late Day Banner Messages

### LINKED ISSUE #2445: More late day banner issues
If a student has an excused extension, the late day banner displays the wrong number of remaining late days.
late days master bug
Note that students are given 3 initial late days, so I don't know why the banner would display 128 remaining late days instead of 126, is it adding the 2 past due date?

Also the double counting bug seems to be back ): Basically the late day banner does not take into account the late days used on the assignment it corresponds with.
The late day banner says I only had one day left and tried to use 2
late day banner bug
However, I actually had 3, I used two on this assignment (Grading Homework), as my late day information shows.
late day banner bug 2
 The late day banner doesn't account for this.

RylanOC:
Fixes #2445, where late days used on previous versions of an assignment are counted against available late days. Also ensures "remaining late days" is correct in banner message when the instructor grants an extension.


RylanOC:
I've done more thorough testing with this, and the behavior seems to be what we're expecting. I've created 4 testing gradables:
1. The due date is in the future
2. The student does not have enough late days
3. The instructor does not allow late days
4. The student has enough late days
   
Here is the late day information for the student before submitting any files:
screenshot 2018-11-04 11 23 49
  
The banner displays the correct message for a late assignment (with enough late days)
screenshot 2018-11-04 11 24 14
  
As well as an assignment which the student does not have enough late days for
screenshot 2018-11-04 11 24 47
  
The banner still shows the correct message for assignment where the instructor does not allow late days
screenshot 2018-11-04 11 42 03
  
After submitting a file, the assignment with enough late days still works
screenshot 2018-11-04 11 25 54
   
As does the assignment where the student does not have enough late days
screenshot 2018-11-04 11 26 36
   
And finally, the assignment where no late days are allowed displays the expected message
screenshot 2018-11-04 11 29 02
  
The "My Late Days" page shows the same information as the banners
screenshot 2018-11-04 11 26 56

bmcutler:
Assigning 2 reviewers.  @ropern123 and @VictorLi718 
Please test this carefully (with every combination you can think of) and also review the code for clarity and maintenance.  Be picky please.  Thanks.

VictorLi718:
Add a new test case. When you don't have enough late days and also the allowed late days is not enough, it only display that "you don't have enough late days". I think it is clear enough.
image
Also works for the "My Late Days" page
image

VictorLi718:
@ropern123 Maybe you can check for corner cases again, I will also test again when I come up with some other conditions

PR #2975, opened 2018-10-10
Forum visual design update 

andrewaikens87:
This PR includes:
- Forum visual design update
- Merge thread searching capability
- Thread ids visible for instructors 
- Code cleanup when generating forum bar. Now uses twig. 
- Displaying only appropriate buttons when there are no threads

Some screenshots:

disc_for

createthread

stats

search

mergesearch


CouleeApps:
Wow, this looks much better! I'm curious about the scrollbar in the first picture... does the whole page scroll as well as the replies? May want to make sure your divs are nested correctly-- some of them will expand to fill the entire page height.

andrewaikens87:
Should be good to go

andrewaikens87:
Should be good now, fixed view on firefox (on linux):
firefox

and chrome:
chrome

Issue #2882, opened 2018-09-19
Autograding total/subtotal results bug


ropern123:
Currently, the autograding results will sometimes not display properly.
In these cases, the autograding total/subtotal will be 0 regardless of the scores from the individual testcases.

I have an idea as to what might be happening:

1) results.json appears in the user's version directory, so the page refreshes
2) AutoGradedVersion gets its results from the database, which has not been updated to reflect the total points yet, as insert_database_version_data.py has yet to finish
3) Each AutoGradedTestcase loads its results from the user's version directory (which is has the point values for the individual test cases) and insert_database_version_data.py finally finishes putting the version data into the database
4) Then, when the user manually refreshes the page, the correct version results are loaded into AutoGradedVersion from the updated database, and the page displays as intended

To fix this, the page must only refresh after the version data has been properly placed into the database.
Also, it is preferable (necessary?) that if one manually refreshes the page after results.json appears and before the database is updated, the resulting page should be displaying the previous active version as it was before the page refreshed.

MasterOdin:
I think we want to move these lines (https://github.com/Submitty/Submitty/blob/master/sbin/autograder/packer_unpacker.py#L294-L303) above (https://github.com/Submitty/Submitty/blob/master/sbin/autograder/packer_unpacker.py#L286-L292) to at least resolve the race condition of the results.json being placed before the insert_database_version_data.py being run.

ropern123:
insert_database_version_data.py uses results.json to calculate the data that will be put into the database, so that will have to be changed as well if results.json is not generated until after it runs.

bmcutler:
We discussed last week that in addition to checking for the presence of the file(s), but we should also check the database to see if it has the autograding scores.  If not, wait and refresh, it will be there next loop.

bmcutler:
please cleanup fixmes in 
site/app/templates/autograding/AutoResults.twig
site/app/views/AutoGradingView.php
from #2978

PR #2978, opened 2018-10-11
Rephrase confusing message about no/incomplete autograding

bmcutler:


bmcutler:
I'm going to merge this now, but I left a few fixme's in the code that should be revisited.  
I think in solving #2882 this code should be cleaned up as needed.

Issue #863, opened 2017-01-24
Allow gradeable not to have a flexible due date


bmcutler:
If an assignment should be submitted "ASAP", it would be good to not display a specific date on the gradeable.  Similarly, late days should not be calcuated for this item and the score should not be automatically set to zero if the item is submitted late.

Examples:  A warmup assignment.  Submission of iclicker id.  Completion of a "once during the term" presentation.

The gradeable should still have internal due date / grading date / release date, so that it can be sorted (need to think about this more) in the overall collection of gradeables.  But grading should be allowed at the beginning (esp for the once a term presentation... ).

KevinMackenzie:
This should have been closed with PR 2993

PR #2993, opened 2018-10-15
[Feedback Requested] Gradeables without a due date

### LINKED ISSUE #1462: "deadline" doesn't make sense for test grading
The due date should automatically be set to the date graders can begin grading?  (but this is weird too, since we might start grading -- and need access to grading, before all items are uploaded)

Workaround is to just reset the dates after all items have been uploaded so no one is marked "late"

KevinMackenzie:
Closes #1462, #863

Adds a `has_due_date` field in the database.  This is automatically set to `false` if `student_submit` is false and, subsequently, if `scanned_exam` is set `true` in the Edit gradeable page.

Student submit false:
image
Student submit true:
image
image

Since there is no due date, the required internal value gets pushed around by the value of other dates to base date validation like any other irrelevant date.

On the submission page, if there is no due date, the warning about late submission is removed (even if the internal `due date` is years in the past).  This is true for both students and non-student submissions.

The late day calculation code is also modified to account for this.  Gradeables without a due date cannot be `late` or `bad`.  Similarly, in the Late Day Summary table, the gradeables without a due date don't show up and don't contribute to the late day usage total.

The navigation page needed to change to account for this too.  For example, gradeables that are student submit, but don't have a due date will always be in the `Open` category and won't show a due date:
image

If the student can't submit, then the gradeable will show up in the `closed for grading` category and will have a different message.  I chose `BULK UPLOAD`, but should it be something else?
image
Similarly, once the `grades released` date is passed, for graders the `Submit` button will still say `BULK UPLOAD`

For the students, it will say `NO SUBMISSION` if it isn't a student-submit and they have no submission.
image
And as before, if they have a submission but no grade it will be:
image

I'm requesting feedback because this affects a lot of different pages and I'm not sure if I missed something.


bmcutler:
Due date still appears in the upper right corner of the submission page.
screen shot 2018-10-25 at 9 46 20 pm

PR #2998, opened 2018-10-17
Mark stats bug with limited access graders

### LINKED ISSUE #2946: "Users who got the mark" for limited access should not show names
They can't grade most of those users so they shouldn't be able to see their names in the "who got the mark" popup. Just general statistics is fine.

LeonMontealegre:
Fixed issue where limited access graders could see all the students with a given mark through the mark statistics popup.

Fix #2946


bmcutler:
The stats confuse me.  (not sure if this is a change or it's always been like this.)

Looking at the same mark..  a full access & instructor see this:
screen shot 2018-10-25 at 9 58 24 pm

but the limited access sees this:
screen shot 2018-10-25 at 9 58 34 pm

Note that I've assigned the full access & limited access to the same sections.

I guess the limited access is only seeing data for the section they are assigned?  Except for the "who got this mark" which is now all.   (that is graded components and total components is just their sections).
But shouldn't the same be true for the full access?  

How about we list both numbers for everyone

who got mark?  5 in your assigned sections (31 across all submitters)
num graded marks?  34 in your assigned sections (194 across all submitters)
...

Or maybe a simple table with column headings so you don't have to keep repeating this.

PR #3005, opened 2018-10-19
Late days not allowed is enforced now

### LINKED ISSUE #2745: Prevent student upload if past deadline and "allow late submission" is false
Finish work started with #2146 and #2620

KevinMackenzie:
Closes #2745 

Implements the control that the `is_late_submission_allowed` field is for.  Submissions are no longer accepted after the due date if the flag is enabled and the submit box does not appear for the user.  The buttons on the nav page for 'closed' gradeables are also updated:

-No submission w/ no late days allowed
-Submission w/ no late days allowed
-Submission w/ late days allowed
image

There is also a box that replaces the submission box to indicate that submission is not allowed (and not broken)
image


bmcutler:
Instructors still need to be able to submit files, so they can upload late materials from students with approved excuses, or to do test submits.  Let's print a message to the instructor that the due date is passed and students cannot submit, but let them have the usual submission access.  (from the main all gradeables page, instructors can look the same as students).

The phrase  "(late submissions may be disabled)"  is going to be confusing.  Either they are disabled or not (let's be specific).

KevinMackenzie:
Should have addressed the issues mentioned

bmcutler:
late day banners are lost :(

PR #3019, opened 2018-10-25
Small UI Tweaks

VictorLi718:
Updated a few different points on UI.


VictorLi718:
Finish three checkpoints. 
1.For the "create/edit gradeable tabs", it looks like this now. 
image
2.For the "text boxes on the Course Settings", I changed to use textarea so that it can have multiple rows, but I am not sure if get value can still work well

VictorLi718:
Closes #2725

PR #3021, opened 2018-10-25
Autograding total/subtotal display bug

### LINKED ISSUE #2882: Autograding total/subtotal results bug

ropern123:
The submission page now checks to make sure that the autograding results are filled in in the database before refreshing the page.
Also, if the user refreshes the page before the autograding results are filled in to the database, it will also wait to display results.

However, I was unable to do extensive testing on whether or not this works, as the issue was hard to replicate on my local machine.

closes #2882

### DIFF
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -1755,16 +1755,28 @@ class SubmissionController extends AbstractController {
 
         // Don't load the graded gradeable, since that may not exist yet
         $submitter_id = $this->core->getUser()->getId();
+        $user_id = $submitter_id;
+        $team_id = null;
         if ($gradeable !== null && $gradeable->isTeamAssignment()) {
             $team = $this->core->getQueries()->getTeamByGradeableAndUser($gradeable_id, $submitter_id);
+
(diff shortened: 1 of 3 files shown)


bmcutler:
@ropern123 can you fix the merge conflicts?

PR #3043, opened 2018-10-27
Late day calculation on the new gradeable model

KevinMackenzie:
part of #2317

Implements the late day calculation logic to use the new model.  Implements the logic in a much more verbose and readable format utilizing a `LateDay` calculation class for each user for all gradeables, and a `LateDayInfo` class that represents each gradeable.


bmcutler:
@KevinMackenzie please fix minor merge conflict

PR #3050, opened 2018-10-28
Team pdf fix

saxocellphone:
Now team assignments can be graded

### DIFF
--- a/site/app/controllers/pdf/PDFController.php
+++ b/site/app/controllers/pdf/PDFController.php
@@ -82,25 +82,20 @@ class PDFController extends AbstractController {
             $active_version = $this->core->getQueries()->getGradeable($annotation_info['gradeable_id'], $user_id)->getActiveVersion();
         }
         $annotation_gradeable_path = FileUtils::joinPaths($course_path, 'annotations', $annotation_info['gradeable_id']);
-        if(!FileUtils::createDir($annotation_gradeable_path) && !is_dir($annotation_gradeable_path)){
-            $this->core->addErrorMessage("Creating annotation gradeable folder failed");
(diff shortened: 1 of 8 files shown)


site/app/controllers/MiscController.php
```diff
@@ -54,35 +54,24 @@ public function run() {
...
-            $this->core->getOutput()->renderJsonError("You do not have access to this file");
-            return false;
-        }
-        $graded_gradeable = $this->core->getQueries()->getGradedGradeableForSubmitter($gradeable, $submitter);
-
-        if ($graded_gradeable === null) {
-            $this->core->getOutput()->renderJsonError("You do not have access to this file");
-            return false;
+        $config = $this->core->getQueries()->getGradeableConfig($gradeable_id);
+        if($config->isTeamAssignment()){
```
KevinMackenzie: this if/else block is unnecessary.  There should be a 'by-submitter-id' version of this function

site/app/controllers/MiscController.php
```diff
@@ -54,35 +54,24 @@ public function run() {
...
-        if ($submitter === null) {
-            $this->core->getOutput()->renderJsonError("You do not have access to this file");
-            return false;
-        }
-        $graded_gradeable = $this->core->getQueries()->getGradedGradeableForSubmitter($gradeable, $submitter);
-
-        if ($graded_gradeable === null) {
-            $this->core->getOutput()->renderJsonError("You do not have access to this file");
-            return false;
+        $config = $this->core->getQueries()->getGradeableConfig($gradeable_id);
```
KevinMackenzie: There's a function in the AbstractController called `tryGetGradeable` or something. Look into using that instead of this

PR #3051, opened 2018-10-28
Enable hotkey remapping for arrow keys in Lab Gradeables Section

### LINKED ISSUE #2813: Enable hotkey remapping for arrow keys
This affects the lab checkpoint grading interface; I'm not sure what the hotkeys in the TA grading interface are yet.

For some reason, while the other various hotkeys can be set to anything, the directional keys are grayed out and cannot be changed. For convenience, I'd like to be able to remap these to vi-keys (hjkl). Can these be made remappable?

VictorLi718:
Closes #2813. Enable the remapping for arrow keys in Lab Gradeables Section only (not for test gradeables as it will disturb the input)


cellocodingryan:
I can set the arrow key to "rightshift + enter" which will then focus on the search bar and make the other arrow keys to stop working. Maybe show the error "enter unique key...." when a user tries to do that.

PR #3056, opened 2018-10-28
[Bugfix] allow students to view files in their own results folder

bmcutler:


CouleeApps:
This allows students to see lots of other autograding files when they download their submission... probably not what we want.

screen shot 2018-10-28 at 01 57 05

PR #3076, opened 2018-10-30
Modify MultiIterator to operate on normal iterators

MasterOdin:
PDO allows for multiple statements to be active at the same time without having to first go through the whole result set of a particular statement first before going into the next statement (you could also do this to interweave the result sets of queries). Additionally, as we're already using DatabaseRowIterator under the hood here, there's little cost to PHP itself to have two open statements.

As such, we can simplify the implementation of the MultiIterator such that we can use it for any type of iterator, not just databases, which eases the implementation and testing of it.

After this change, I saw no affect in:
* Submitting assignment
* Viewing grading summary for individual assignment
* Grading individual assignment
* Viewing grading summary for team assignment
* Grading team assignment

### DIFF
--- a/site/app/libraries/MultiIterator.php
+++ b/site/app/libraries/MultiIterator.php
@@ -2,16 +2,27 @@
 
 namespace app\libraries;
 
+use app\exceptions\NotImplementedException;
+
 /**
  * Class MultiIterator
  * @package app\libraries
  *
- * Allows the chaining of multiple iterators in one iterator object
+ * An iterator that allows for sequentially iterating over each passed
+ * iterator in-order, one at a time. This means that iterator2 will not be
+ * called until iterator1 has been exhausted. This differs from
+ * MultipleIterator as that commits all iterator functions over all attached
+ * iterators in parallel (i.e. calling next() will return an array with the
+ * results of next() on all attached iterators).
+ *
+ * @see \MultipleIterator
  */
 class MultiIterator implements \Iterator {
 
-    /** @var array array of [\Closure, $this] */
-    private $iterator_generators = [];
+    /** @var \Iterator[] */
+    private $iterators = [];
+    /** @var int */
+    private $iterator_key = 0;
     /** @var \Iterator */
     private $curr_it = null;
     /** @var int  */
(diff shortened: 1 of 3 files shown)


site/app/libraries/MultiIterator.php
```diff
@@ -2,27 +2,38 @@
...
+ * An iterator that allows for sequentially iterating over each passed
+ * iterator in-order, one at a time. This means that iterator2 will not be
+ * called until iterator1 has been exhausted. This differs from
+ * MultipleIterator as that commits all iterator functions over all attached
+ * iterators in parallel (i.e. calling next() will return an array with the
+ * results of next() on all attached iterators).
+ *
+ * @see \MultipleIterator
  */
 class MultiIterator implements \Iterator {
```
KevinMackenzie: I think it make sense to rename `MultiIterator` to `CascadingIterator` or `SerialIterator` so its more descriptive and less similar sounding to `MultipleIterator`.

site/app/libraries/MultiIterator.php
```diff
@@ -2,27 +2,38 @@
...
 
-    /** @var array array of [\Closure, $this] */
-    private $iterator_generators = [];
+    /** @var \Iterator[] */
+    private $iterators = [];
+    /** @var int */
+    private $iterator_key = 0;
     /** @var \Iterator */
     private $curr_it = null;
```
KevinMackenzie: We can get rid of the `curr_it` field since we're using `iterator_key` to index the array of iterators

site/app/libraries/MultiIterator.php
```diff
@@ -2,27 +2,38 @@
...
     private $key = 0;
 
     /**
      * MultiIterator constructor.
-     * @param \Closure[] $iterator_generators
+     * @param \Iterator[] $iterator_generators
      */
-    public function __construct(array $iterator_generators) {
-        $this->iterator_generators = $iterator_generators;
+    public function __construct(...$iterator_generators) {
```
KevinMackenzie: rename argument to just `iterators`

MasterOdin:
@KevinMackenzie comments addressed

PR #3094, opened 2018-11-05
Don't archive test_input files

bmcutler:
When test case isolation was introduced, a negative side effect was that we started saving copies of the test_input files.

This PR:

* removes the test_input directory (sibling to the test01, test02, etc) before archiving begins
* removes the test_input files from test01, test02 directories
* automatically archives all files that are an 'actual_file' for a test case validation check (this will reduce the need for customization of "work_to_details")
* a new autograding configuration that has test_input subdirectories and test_output subdirectories (and fixes bugs related to these...)


bmcutler:
A test_input file should never be the "actual_file" from a validation check.  That doesn't make sense.  The actual_files are produced by student code.  I don't want to check the input_files against the files in the work_to_details list, because those could/do have wildcards like *.txt and nothing will actually be deleted/cleaned up.   

The one thing I'm partially worried about is if an input.txt file is named the same as a target output.txt file.  (If the instructor is asking a student to overwrite a test_input file.)  I think that's just a poorly written test.

emaicus:
Ah, alright, I agree. With that in mind, the PR appears to work on Jailed Sandbox, single docker and multi docker gradeables.

Issue #3082, opened 2018-10-31
Generalizing Paragraph Textbox Config


achristoforides:
- [x] Rename "assignment_message" to "gradeable_message" (optional)
- [x] Message should be boxed banner, not green text
- [x] Problem Label should be bold (required)

New Content array (holds content boxes):
- [x] Holds html description, labels (before descriptions), input fields
- [x] Headers for content, newlines, links, italics, bold, colored font, preformatted (fixed width font for code), math symbols, image source=, images included in the assignment (can't see assignment, can't see images/files)

Textbox Modifications:
- [ ] New code box option where instructor selects language

Multiple Choice
- [x] Multiple choice input for questions
- [x] Check-all-that-apply multiple choice input for questions

bmcutler:
@emaicus please review issue & open new issues for remaining items (as appropriate) and then close this issue.

PR #3107, opened 2018-11-09
Refactor of assignment_message and Style Change

achristoforides:
Part of #3082

The assignment_message (found in gradeable json configs) has been refactored to now be called "gradeable_message". This pull request would maintain backwards compatibility with assignment configs which use "assignment_message" instead of the refactored name "gradeable_message". Additionally, rebuilding of old configs is not necessary as the changes proposed check whether or not the old message is used as well. If it is, then it displays the existing message in the new format.

The new style has two different styles which were previously discussed; one with the Submitty blue background and white text and another with the "faded" Submitty blue and black text. I very much favor the first.

Option 1:
screen shot 2018-11-09 at 12 11 41 am

Option 2:
screen shot 2018-11-09 at 12 22 18 am

For option two, I had to calculate the "would be" color of the box because in other places that it exists ("grayed" out buttons), opacity is used to get the color.

Option 1 is eye catching and I think will be hard to miss (which would be preferable if the information there is critical to the assignment and/or has explanations as to how to submit the homework or assignment).

This new box appears below any late day banners/warnings.


bmcutler:
I like the dark blue too.
Going to add Glenn in here if he has an opinion, else I will probably merge & install this weekend.

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -1,6 +1,11 @@
 {% import _self as self %}
 
-<div class="content">
+{# Display the gradeable_message if one exists at the top of the page #}
+{% if has_gradeable_message %}
+    <div class='content' style='background-color: #006398; color: white;'><p>{{ gradeable_message|raw }}</p></div>
```
CouleeApps: Would highly recommend using a css class and putting the style elements in server.css

CouleeApps:
Other than that, I also approve of the dark blue (via meeting)

site/public/css/server.css
```diff
@@ -364,6 +364,11 @@ table tr.table-header {
     }
 }
 
+.gradeable_message {
+    background-color: #006398 !important;
```
CouleeApps: Instead of using `!important` just make the selector more specific eg `.content.gradeable_message`

PR #3115, opened 2018-11-12
Let student know if annotation is available 

### LINKED ISSUE #3025: Make sure student sees pdf annoations
1) If a pdf has annotations, make that clear to the student with a badge and a message "you have feedback from grader" or soemthing.

2) When indicating if a student has viewed the grade, let's add a check to see if they opened the pdf, if the pdf has annotations.

saxocellphone:
closes #3025.
Now the date of when student viewed annotation is stored in the database. Currently, it's displayed nowhere, so that's up for discussion. Also a note will appear in the TAResultbox if an annotation is available for the student.


migration/migrations/course/20181112133026_student_annotation.py
```diff
@@ -0,0 +1,8 @@
+def up(config, conn, semester, course):
+    with conn.cursor() as cursor:
+        cursor.execute("ALTER TABLE gradeable_data ADD COLUMN gd_user_annotation_viewed_date timestamp(6) with time zone DEFAULT NULL;")
+
+
+def down(config, conn, semester, course):
+    with conn.cursor() as cursor:
+            cursor.execute("ALTER TABLE gradeable_data DROP COLUMN eg_regrade_request_date;")
```
KevinMackenzie: You might want to check the column you're dropping : )

site/app/controllers/pdf/PDFController.php
```diff
@@ -71,6 +71,9 @@ private function showStudentPDF(){
                 }
             }
         }
+        $ta_graded_gradeable = $graded_gradeable->getTaGradedGradeable();
```
KevinMackenzie: The result of `getTaGradedGradeable` can be null, so you should check that.  Also noticed that there are a lot of unchecked results above in the function.  Don't need to change those for this PR, but something you might want to more gracefully handle. (i.e. `tryGetGradeable` will return `false` on failure)

bmcutler:
@saxocellphone please resolve conflicts

bmcutler:
@KevinMackenzie  please re-review

bmcutler:
Thanks for the contributions but....
We are closing this since it's quite stale and needs alot of work.  It will be easier to start over.

PR #3124, opened 2018-11-17
fixed can't scroll vertical problem

saxocellphone:
Now it will let you scroll pdf without having to wait for the entire test to render.


saxocellphone:
let me just finish up cleaning up a bit more

saxocellphone:
Actually, this should work for now, but I'll optimize this more in my refactor

PR #3129, opened 2018-11-19
Limit scope of validation container.

emaicus:
Validation container is no longer created before compilation.


emaicus:
Untrusted users now clean up any orphaned containers before autograding.

PR #3136, opened 2018-11-27
Send Email on Instructor Announcement 

kasra-sheik:
This PR adds functionality that allows an Instructor/TA to send an email out when making a discussion forum post. By checking the 'Send Email' check box, an email will be sent to every student enrolled in the class containing the thread title and body (given by the image below). 

email_showcase

The email functionality lives inside of sbin/sendEmail.py and in the case of instructor announcements, is kicked off/handled via the Submitty job queue.   

Possible Future Improvements: 
- Attach uploaded files in the email 
- Attach a link to the forum post in the email 
- Add a batch option to combat any spam filtering done by mail servers due to bulk email sends


bmcutler:
small change request:
Please send the instructor announcement email to all student users who are not in the NULL registration section and all other users (whether they are in the NULL section or a real section).

PR #3140, opened 2018-11-27
Allow instructor to specify number of ports listened to by a container

emaicus:
Added ```ports_per_container``` field both globally and per testcase to the assignment configuration json.

This number defaults to 1, and must range between 1 and 100. It specifies the number of ports allocated to a container in the knownhosts files. Router integration is left for later.

This PR should not require a rebuild of any existing gradeables. All existing gradeables will be assumed to use 1 as the number of ports per container.


bmcutler:
Closing this for now.  We'll re-design the network hosts file to support a range of ports instead.  Don't want to make a change in format mid semester.

PR #3172, opened 2018-12-13
Improvements to submitty_jobs daemon to make it more robust and secure

MasterOdin:
This continues the work done in #3170 in making the Submitty jobs daemon more secure and robust. This PR adds:

* Preventing an exception during job execution from killing one of the processes
* Provide a 'check_required_keys' function to validate that all necessary keys (defined at the class level)
* Provide a 'sanitize_job_details' function to allow jobs to sanitize down their details
* Provide a 'validate_job_details' function for jobs to ensure job_details is set correctly and are valid
* Provide new abstract "CourseJob" which requires semester and course to be set, sanitizes them by taking the basename, and then ensures they point to a valid directory without any directory traversals.
* Provide a new abstract "CourseGradeableJob" which extends "CourseJob" and also requires gradeable, taking the basename of it, and ensuring it's not blank or just a dot.


mattraneri:
Quick question/comment-
I think it would be better to have sanitize_job_details return a boolean on whether the parameters are sanitary, and have the parent class return False by default rather than passing.

My reasoning is that if a future job is being written and someone fails to implement the sanitizer method, their code and implementation for that job will still run- if we make the parents fail by default, it will force future jobs to implement proper checks.

What are your thoughts?

PR #3174, opened 2018-12-13
Reports controller on new gradeable model

KevinMackenzie:
Part of #2317 

Converts the reports controller to use the new Gradeable model.  It should have identical results, but the "status" should be better generated and the 'note' should be more descriptive.  It reuses code when possible in the generation loop.


bmcutler:
@KevinMackenzie please fix travis tests
Thanks :)

KevinMackenzie:
I think the tests were failing because `WIP` was in the title

PR #3175, opened 2018-12-13
Add special case for instructor not on team

### LINKED ISSUE #3168: instructor not on a team, getting error when viewing the submit page
students cannot access this page without being on a team, but the instructor should be able to (so they can do upload as student/bulk upload, etc)

RylanOC:
It looks like the crashes were caused by rendering the late day message when an instructor isn't on a team. I've added a check to avoid this in the special case that the user is an instructor and is not on a team.

fixes #3168

### DIFF
--- a/site/app/models/User.php
+++ b/site/app/models/User.php
@@ -317,6 +317,6 @@ class User extends AbstractModel {
      */
     public function onTeam($gradeable_id) {
         $team = $this->core->getQueries()->getTeamByGradeableAndUser($gradeable_id, $this->id);
-        return $team!==NULL;
+        return $team !== NULL;
     }
 }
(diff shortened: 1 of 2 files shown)


site/app/models/User.php
```diff
@@ -308,4 +308,15 @@ static public function validateUserData($field, $data) {
...
+
+    /**
+     * Checks if the user is on ANY team for the given assignment
+     *
+     * @param string gradable_id
+     * @return bool
+     */
+    public function onTeam($gradeable_id) {
+        $team = $this->core->getQueries()->getTeamByGradeableAndUser($gradeable_id, $this->id);
+        return $team!==NULL;
```
MasterOdin: add a space between the operands and the comparison operator.

PR #3178, opened 2018-12-13
Change "Max. late days allowed" column in late days table

### LINKED ISSUE #3162: "Max # of late days allowed" in the My Late Days/Extensions table
It looks like we are calculating the # of late days the specific student could use on a homework.
Rather it should be the # of late days that are allowed on that assignment (from the create/edit gradeable dates page).

Even if I have 5 late days to spend, if I'm only allowed to use 2 late days on the last homework (because the TAs are going to start grading 2 days after the deadline), I can only use 2 late days.

[please discuss amongst yourself who will fix this]

RylanOC:
Changes "Maximum number of late days allowed for this assignment" column to display the number of late days allowed for _the gradeable_ , instead of the maximum number of late days a _given student_ can spend on the assignment. 

Fixes #3162

### DIFF
--- a/site/app/models/gradeable/LateDayInfo.php
+++ b/site/app/models/gradeable/LateDayInfo.php
@@ -80,11 +80,19 @@ class LateDayInfo extends AbstractModel {
     }
 
     /**
+     * Gets the max number of late days the instructor allows for the gradeable
+     * @return int
+     */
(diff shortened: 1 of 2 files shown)


RylanOC:
Before
screenshot 2018-12-13 14 35 19

After
screenshot 2018-12-13 14 34 10

PR #3185, opened 2018-12-13
Fix networking on vagrant machines

CouleeApps:
There was an apt package that had a bug, but the default vagrant install doesn't have the updated version. So, this fixes that.


MasterOdin:
Is there a link to a bug report or can we include some wordage as to why this is necessary to fix networking within Vagrant.

CouleeApps:
Bug report is here, https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/1768560  
We can remove when netplan.io updates to 0.40+ in Ubuntu stable release and our Vagrant bento box installs it

PR #3189, opened 2018-12-14
Revamped Email Script

kasra-sheik:
This PR is a revamp to how emails are handled within Submitty.  An "emails" table has been created inside the master DB which stores queued emails to be sent. From there, a job scheduler (such as cron) could run "sendMail.py" at some non rate limiting interval which polls for queued emails and sends them. 

This PR includes using this updated mechanism every time an instructor checks the "send mail?" checkbox when making a discussion forum announcement. The forum controller will attempt to insert a row inside emails for each active member of the course. 

Future improvements could include:
- cleanup script that deletes old/sent emails to reduce storage
- storing course/semester for logging purposes
- if an instructor posts an attachment in the forum, include this in the email as well


migration/migrations/master/20181213143003_emails.py
```diff
@@ -0,0 +1,11 @@
+def up(config, conn):
+    with conn.cursor() as cursor:
+        cursor.execute("CREATE TABLE IF NOT EXISTS emails (\
```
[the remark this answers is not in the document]
bmcutler: Yes, that was my one suggestion when we were brainstorming
forgot to pass that on to @kasra-sheik

sbin/sendEmail.py
```diff
@@ -19,6 +20,25 @@
...
+		conn_string = "postgresql://{}:{}@/{}?host={}".format(DB_USER, DB_PASSWORD, db_name, DB_HOST)
+	else:
+		conn_string = "postgresql://{}:{}@{}/{}".format(DB_USER, DB_PASSWORD, DB_HOST, db_name)
+
+	engine = create_engine(conn_string)
+	db = engine.connect()
+	metadata = MetaData(bind=db)
+	return db 
+
+db = setupDB()
```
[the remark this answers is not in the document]
kasra-sheik: The "markSent" function also makes a database call. Should I pass that in from getEmailQueue or leave db global?

bmcutler:
@MasterOdin Can you re-review this PR?  We'd like to get this to master/live Submitty soon.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -3763,4 +3763,30 @@ public function getHasSubmission(gradeable\Gradeable $gradeable, Submitter $subm
...
+        $this->course_db->query('SELECT user_email FROM users WHERE user_group != 4 OR registration_section IS NOT null', $parameters);
+
+        return $this->course_db->rows();
+    }
+    /**
+     * Queues an email to be sent by email job
+     * @param array $email_data
+     * @param string $recipient  
+     */ 
+    public function sendEmail($email_data, $recipient){
```
MasterOdin: let's call this function `createEmail` or `queueEmail` or something to be more accurate what it does.

.setup/INSTALL_SUBMITTY_HELPER_SITE.sh
```diff
@@ -77,6 +77,13 @@ for i in "${array[@]}"; do
     find ${SUBMITTY_INSTALL_DIR}/site/public -type f -name \*.${i} -exec chmod o+r {} \;
 done
 
+#Setup Cron Jobs
+crontab -r -u submitty_daemon
+crontab -l > cron_jobs -u submitty_daemon
```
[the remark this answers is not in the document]
kasra-sheik: Just realized the list command is unnecessary here. Must have carried over from some initial testing.

.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
```diff
@@ -70,7 +70,7 @@ chmod 550 ${SUBMITTY_INSTALL_DIR}/sbin/authentication.py
 chmod 555 ${SUBMITTY_INSTALL_DIR}/sbin/killall.py
 
 # DAEMON_USER only things
-array=( build_config_upload.py run_lichen_plagiarism.py sendEmail.py submitty_autograding_shipper.py submitty_autograding_worker.py submitty_daemon_jobs autograder)
+array=( build_config_upload.py run_lichen_plagiarism.py submitty_autograding_shipper.py submitty_autograding_worker.py submitty_daemon_jobs autograder )
```
MasterOdin: don't we still want to chmod this to proper permissions?

.setup/INSTALL_SUBMITTY_HELPER_SITE.sh
```diff
@@ -77,6 +77,12 @@ for i in "${array[@]}"; do
     find ${SUBMITTY_INSTALL_DIR}/site/public -type f -name \*.${i} -exec chmod o+r {} \;
 done
 
+#Setup Cron Jobs
+crontab -r -u submitty_daemon
```
MasterOdin: Let's not be totally destructive on cron in case we plan to utilize it elsewhere.

kasra-sheik: I was removing the current jobs list just because if not, every time a user runs install, a copy of same job would be appended to the list. 
Is there are better place to put this command? Maybe we can standardize an area for cron jobs in one of the install scripts?

.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
kasra-sheik: Yep. Forgot to re-add that in after the name of the email script changed. Good catch.

.setup/INSTALL_SUBMITTY_HELPER_SITE.sh
MasterOdin:
```
crontab -u submitty_daemon -l > /tmp/cron_jobs
grep "python3 ${Submitty_INSTALL_DIR}/sbin/send_email.py" /tmp/cron_jobs || echo "* * * * * python3 ${SUBMITTY_INSTALL_DIR}/sbin/send_email.py" >> /tmp/cron_jobs
crontab -u submitty_daemon -r
crontab -u submitty_daemon /tmp/cron_jobs
rm -f /tmp/cron_jobs
```

A longer discussion would be do we really need to progromatically manipulate this crontab, or would we be better off just creating a static one that we put in say `.setup/crontabs/` and that we always just install that file. I think I'd really prefer this option as it also makes explicitly clear in the codebase how we might add more cron jobs for a particular user.

bmcutler:
@MasterOdin Can you re-re-review please? :)

PR #3203, opened 2018-12-26
Major changes and improvements to migrator

### LINKED ISSUE #2516: migration rollback should require you to specify -e system/master/course
(It just never makes sense to do the top from each).

Also, rollback should never undo the initial migration (besides the down migration is too scary).

MasterOdin:
I'm opening this PR mainly to start a discussion/early review on this kind of large PR 

This is a large overhaul of the migrator tool that makes the internal code cleaner and writing migrations a bit easier. Additionally, this makes the migrator tool itself DB agnostic, and just leaves the data files/migrations as PostgreSQL specific, making it far easier and less expensive to test.

An overview of the internal changes are:
* Move internal migrator code into a proper python module structure
* Rename runner file, `migrator.py`, to `run_migrator.py` to avoid conflict in name with new module structure
* Move from using raw psycopg2 to using SQLAlchemy
* Add new 'status' command that returns a table of migrations in an environment
* Must specify environment when doing migrate/rollback
* Cannot rollback initial migration

For migration writers, the changes are a bit more subtle, but removes a handful of lines of boilerplate that did exist. The first thing is that we're now passing in a `Database` object (see the `migrator/db.py` file for its definition) that contains the SQLAlchemy connection instead of a raw psycopg2 connection. What this allows though is that instead of having to do:
```
with conn.cursor() as cursor:
    cursor.execute(...)
```
You can now just do:
```
database.execute(...)
```
We still maintain the idea of transactions within a migration (so if a query in the middle of the migration fails, the whole thing is rollbacked).

For checking for if a column exists in a table, instead of doing:
```
cursor.execute("SELECT * FROM table LIMIT 0")
colnames = [desc[0] for desc in cursor.description]
if column not in colnames:
    cursor.execute(...)        
```
You can now just do:
```
if not database.table_has_column(table, column):
    database.execute(...)
```

----------------------

To-Do:
- [x] Test that vagrant up works with new structure of migrator (and that Travis-CI passes fully)
- [x] Add a `base_migration_*.py` file for each environment type with proper docstrings
- [x] Add missing migrations to the status output
- [x] Finish docstring documentation of new module files
- [x] Re-enable flake8 linting on migrator
- [x] Add unittesting for the migrator module (either mocking sqlalchemy or using SQLite setup)
- [x] Test full going up/down all migration environments
- [x] Document what open PRs will need to be adjusted for new migration structure

Closes #2516


MasterOdin:
Affected PRs:
- [ ] #3115
- [ ] #3189

Both of these still require some work, and as such, I believe this can be merged first, and then I'll go in and make adjustments to these to fix their migration script.

MasterOdin:
@bmcutler While I still need to go through all up/down transitions on Vagrant, I believe the code is ready for review/comment.

PR #3206, opened 2019-01-03
[WIP] Add sub-directory to grading and add test for python script

MasterOdin:
This is a spike of how we might go about working on the code under grading as we convert parts of it (like `load_config_json.cpp`) to python while ideally also writing out testcases to validate functionality. To do this easily, it requires moving the code under a "python module" (a folder with a `__init__.py` file) that then in this case unittest works against. For the initial go, I only focus on validating the behavior of the very tiny `json_syntax_checker.py` for more to establish the folder scheme before moving onto the more complex `load_config_json.cpp`.


bmcutler:
Discussed...  Matt will revise the renaming and reopen PR

PR #3210, opened 2019-01-04
Update font awesome, replace some icons

bmcutler:


bmcutler:
@andrewaikens87 @CouleeApps 
This font awesome update has changed the names of some icons.  I fixed the ones I saw on the left sidebar, but I'm afraid other things have changed.

MasterOdin:
I think this may, looking at https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#name-changes, break some more of our usage of FA than just the left sidebar. A quick look at the source and the above link, makes it seem that like `fa-trash` is now `fa-trash-alt`, `fa-pencil` is `fa-pencil-alt`, `fa-window-restore` is now a pro icon (though there is a fas version, but it's slightly different), all of the `-o` icons have lost the `-o` suffix in `TAResults.twig`, etc. 

I think before merging this, we may want to draw up a table of all FA4 icons we currently use, their FA5 counterpart and see if there's any that we may need to replace. We'll also want to replace the plain `fa` at the start of the icons with either `fas` or `fab` I believe (so like the Github icon in the footer should now be `fab fa-github` instead of `fa fa-github`).

Issue #3197, opened 2018-12-21
Formatting text/material around textboxes on submission page


bmcutler:
Basic requirements for markdown/basic html on text blobs for textbox assignments:

```
<tt>...</tt>  (inline fixed width font)

<a href=""> ... </a>  (basic links to offsite references)

<pre>...</pre>  (large code blobs)
```

(or equivalent in markdown)

Issue #3196, opened 2018-12-21
Submission textbox initialized with some starter code/previous answer


bmcutler:
Feature request:  For an exercise where a student is supposed to fix some provided broken code.
The textbox should be initialized with the instructor provided code (specified in the config), so the student can make small edits in the textbox and submit.

If the student has goofed in the edits there should be a button to reset the textbox to the provided code.  (could reload the whole page, but that might be more disruptive.  would reset all textboxes... )

Related request:  We should have a simple method to load the contents of the previous submission into each of the textboxes.  (helps with incremental work/submissions)

bmcutler:
duplicate of #3695 (still open, soon to have PR) & #3696 (closed)

PR #3217, opened 2019-01-14
Paragraph textboxes rework and new input types

achristoforides:
Part of #3082 #3196 #3197 

New content types have been created for new input types, allowing for interaction with the site not only with plaintext textboxes, but also formatted codeboxes, radio buttons, and checkboxes. Customization of these new input types can be done through the traditional method of modifying the config.json for a given gradeable.

Additions/modifications are as follows:

New format scheme for creating informational data on a given gradeable's page:
```
{
    ...
    content: [
        {
            "title" : string, optional
            "exposition" : string, markdown, optional
            "images" : [
                Same structure that is already in place
            ],
            "input" : [
                See below for definition
            ]
        }
    ],
    ...
}
```

New format scheme for creating input fields on a given gradeable's page:
```
"input" : [
    {
        "type" : string, "textbox", "codebox", "multiplechoice", required
        // NOTE: Textbox configuration is identical to what is already in place
        
        // if "codebox" is type, use this field. Fields from "textbox" are also used
        "language" : string, language, required (if type is "codebox")

        // if "multiplechoice" is type, use these fields
        "allow_multiple" : boolean, required
        "choices" : [
            // array of choices, strings, required
        ],

        // same as before
        "filename" : string, required
    }
]
```

At the moment, the ability to create informational content using the "content" object is working partially. The utilization of images within this object hasn't been tested, but the ability to display information in markdown and have it render as html is functional.


MasterOdin:
Can you remove the absolute ton of extra unused files you're bringing in with ACE (and who knows what else) so that the diff isn't so massive with clutter?

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -149,14 +149,74 @@
...
+	    {% for i in c.input %}
+	      {% if i.type == "textbox" %}
+		{{ self.render_textbox(i, image_data, loop.index0) }}
+	      {% elseif i.type == "codebox" %}
+		<style type="text/css" media="screen">
+		  #editor_{{ loop.index0 }} { 
+		  width: 75%;
+		  border: 3px solid #FF00FF;
+		  }
+		</style>
```
MasterOdin: Change this to a class and put it into `server.css`. Then give all divs that class.

MasterOdin:
You can further delete most anything that isn't in the `src-min` folder of `ace-builds` (and then just move all of the files under `ace-builds/src-min/` to just `ace/` for simplicity.

The licence information you'd want to include in https://github.com/Submitty/Submitty/blob/master/LICENSE-THIRD-PARTY.md.

----
However, a larger issue I'd like to discuss is what is the benefit of choosing [ace](https://ace.c9.io/) over [CodeMirror](https://codemirror.net/), especially given that we already use the latter in the code base to do syntax highlighting off code for TA grading.

achristoforides:
Some of the changes you suggested have been made. I am not working on this anymore, so the other questions you raised should be directed towards who ever takes this over (I'm not sure if/when someone will pick this up or who would do it).

emaicus:
## The State of this PR
**Purpose of the PR:** This PR is aiming to expand our existing ```paragraph_textboxes``` to instead be ```AbstractGradeableInput```, an abstract class which is used to polymorphically encapsulate multiple types of subclassed ```GradeableInput```.

### Status of Implementation
**Backend/config.json:** It appears that the work needed to properly integrate this feature into the ```config.json``` is largely finished (e.g. allowing ```content``` to be specified, which may be one of the new input types). This work can be found in ```main_configure```. It is possible that parts of this code could be re-evaluated as the PR progresses to trim down on the number of required fields/to add better defaults.

**New Models:** The ```AbstractGradeableInput```, ```AbstractTextBox```, ```SubmissionMultipleChoice```, ```SubmissionCodeBox```, and ```SubmissionTextBox``` models appear to be in place. These correspond to the new types of GradeableInput introduced by the PR. 

**Controller Changes:** The work in the ```SubmissionController``` should get additional review to ensure that results from the new inputs are properly stored in a format useful to instructors. Based on the current state of the views, this code will likely need to be iterated upon.

**View:** Work has begun in ```SubmitBox.twig``` to display the new input types. It appears that code boxes are currently not implemented, and should be displayed with ```CodeMirror``` if at all possible. A new library has been added to handle the display of instructor markdown content.

emaicus:
The initial version of this PR should be ready for re-review/merge. Images and textboxes now work again, and multiple choice has been updated to work as well. Another revamp will be coming soon in a series of separate PRs which will allow starter code, markdown formatting, re-tool the way that multiple choice is done, and allow code formatting.

PR #3221, opened 2019-01-20
Fix discussion forum thumbtack display

andrewaikens87:
This PR fixes the display icon for pinning a thread.


andrewaikens87:
PR is updated to fix forum filtering issue.

PR #3225, opened 2019-01-22
Fix full height cutting off sidebar

andrewaikens87:
This fix should resolve the issue when the sidebar overflows the height of the display content. This creates an extra scrollbar when viewing content, but this expands the view to maximize the amount of content that is visible (scrolling hides the top breadcrumbs as shown below). We can discuss if this is the right approach to solve this problem. 

Top of scroll:
top

Bottom of scroll:
full page display


andrewaikens87:
Will open new PR per discussion.

Issue #3218, opened 2019-01-15
2 minor forum bugs


bmcutler:
When I click on "alphabetical" or "chronological", the forum switches to the top (most recent/announcement post).  Fortunately, the mode stays and I can return to the thread I was previously on as a workaround.

Also, when someone else deletes a post, the notification is added to my notifications list, but without a link, so I can't clear it (unless I click "mark all as seen")

mattraneri:
We just had a discussion about forum delete behavior:

The desired behavior is the following:

If the user is an instructor and allowed to view deleted forum posts, then the link should go to the forum and "Show deleted threads" should be enabled automatically.
If a student's post is deleted by an instructor or TA, then the student shouldn't get a link and only be able to clear it using the 'marked as read' option on the right hand side of the page

PR #3228, opened 2019-01-24
[Bugfix] Fixed the forum bar, change display, buttons

elihschiff:
The current thread id number was not being passed in when rendering the forum bar which made the 3 change display buttons default to a thread id of "undefined" and the page would then reload to the default top forum post instead of keeping the user on the current forum post.

This fixes half of issue #3218 however I don't have a fix for the other half, so I don't want this to automatically closed that issue. If this gets merged, I can put a comment on that issue to let people know what still needs to be fixed.


elihschiff:
Never mind I see now that this causes an issue when going to the create thread page, as there is no current thread number when you are trying to create a thread. This is probably the reason this code was removed in the first place. I will fix that and make a new pull request with the fix so that it works in both situations.

PR #3232, opened 2019-01-25
C++ Command Line Argument Parsing

emaicus:
**Requires a Rebuild of All Assignments for All Courses**  

*  Updated the ```main_runner```, ```main_validator``` and ```main_compiler``` scripts to use [TCLAP](http://tclap.sourceforge.net/), a C++ command line parser provided under the MIT license. 
*  Added a system migration for TCLAP installation on existing Submitty installs.  
*  Added TCLAP installation to ```install_system.sh```  
*  Updated the way ```grade_item_main_runner.py``` invokes the ```main_``` scripts, including adding a new ```display``` argument to ```main_runner``` to help better detect which display to run windowed programs on.


emaicus:
I had previously indicated in a meeting that it might be able to tweak this PR so that it does not require a rebuild of all gradeables. This would be true if it was not introducing the ```--display``` option, which optionally passes the ```DISPLAY``` environment variable on a system. As it stands, that change will require a rebuild, and would have to be nixed to avoid it.

PR #3234, opened 2019-01-27
Verify grader improvements

### LINKED ISSUE #1959: Verify Graders -- Part 2
Follow up on initial work in Issue #1856 & PR #1966  

- [x] Let's modify the database to add a second optional grader username on every component.

- [x] Then pushing the "verify grader" buttons won't delete/change the original grader name, but will instead add a second name to the component.

- [x] Let's make a course level option (from the instructor course settings page) for "Enable Manual Grading Verification" which has an on / off state.  By default verification is "on".  If it's off, the verify grader button never appears (but we sometimes have 2 graders, see below).

- [x] On the TA grading page, on each component we list "GraderA" if the second grader is null/empty string.  Or "GraderA / GraderB" if the second grader field is not empty.

- [x] From the student view, if the first grader is a limited access grader, we never show that name.  If both first and second grader are full access or instructor, we show both names to the student.

- [x] The second grader is always a full access grader or instructor.  Limited access graders never see the verify grader button, and are never saved as the second grader.

- [x] If a component is already graded and a full access or instructor grader opens it up and edits it (we assume a minor edit) then that person's name is filled in (or replaced) as the second grader.  

- [x] If a limited access grader edits an existing grade, that grader replaces the first grader name and the second grader is set to null/empty string.

Thoughts? Discussion?

NOTE:  Let's write robust code.  If a users status changes after they have graded (if a full access grader is demoted to a limited access grader or a grader is demoted to student only access nothing breaks, no grades are deleted, no authors of grades are lost -- no cascading deletes!  But we stop displaying their names to the students, they lose access to grading interface, etc.

### LINKED ISSUE #2495: Design for enabling Verify Graders Feature
From the course settings page, instructor can select

"Enable Verification of Limited Access Grading"  (additional subtext:  "Use this feature if you would like to require that all grading by limited access grading must be checked and approved by a full access grader or instructor")
or
"Enable Double Grading"  (additional subtext: "Use this feature if you would like to require that ALL grading be checked and approved by a second person")
or
neither.  (Cannot select both)

Note: Changing that selection does not change any of the data already in the system.

If either option is selected, then the verify grader button appears in the ta grading system.

Note: No matter what option is selected...   Editing the marks of another grader always adds that grader as the 2nd grader.  Except when the person editing the marks is a limited access grader in which case that person is set as the first grader and the second grader (if any) is cleared.  Basically a limited access grader editing the mark means that verification or double grading must be re-done.

In the stats page we only show the verified or double graded column stats if one of those options is selected (but not both).  If neither is selected we do not show that data at all.

Thoughts?

shailpatels:
This PR saves the verifier instead of overwriting the grader
vg_1

Full access graders (TA's and Instructors) are able to verify any grader other than themselves. If the original grader is a full access grader they will always be shown, if the grader is limited access and a verifier exists, they will be shown instead

If a component's grade changes the verifier will be set to null if there is one

fixes #2495
fixes #1959


.setup/update_database.py
```diff
@@ -139,4 +139,10 @@
         os.system("""PGPASSWORD='{1}' psql --host={2} --username={3} --dbname={4} -c '{0}'""".format(create_table.format('course'), *variables))
         os.system("""PGPASSWORD='{1}' psql --host={2} --username={3} --dbname={4} -c "{0}" """.format(insert_row.format('course'), *variables))
         # add user/database
+
+        # add verify grader columns
+        os.system("""PGPASSWORD='{}' psql --host={} --username={} --dbname={} -c 'ALTER TABLE gradeable_component_data ADD COLUMN gcd_verifier_id VARCHAR(255)'""".format(*variables))
```
KevinMackenzie: I don't think we want to modify this file. I think this is all pre-migration code.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -224,109 +223,49 @@ private function ajaxVerifyComponent() {
...
-
-        // get the component
-        $component = $this->tryGetComponent($gradeable, $component_id);
-        if ($component === false) {
-            return;
-        }
-
         // Get user id from the anon id
         $submitter_id = $this->tryGetSubmitterIdFromAnonId($anon_id);
-        if ($submitter_id === false) {
```
KevinMackenzie: why did this check move to inside the if block?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -3653,6 +3661,11 @@ private function updateGradedComponents(TaGradedGradeable $ta_graded_gradeable)
                     $deleted_marks = array_diff($component_grade->getDbMarkIds() ?? [], $component_grade->getMarkIds());
                     $this->deleteGradedComponentMarks($component_grade, $deleted_marks);
                     $this->createGradedComponentMarks($component_grade, $new_marks);
+
+                    //change the component to be unverified after changing a mark
+                    $component_grade->setVerifier(null);
+                    $component_grade->setVerifyTime(null);
+                    $this->updateGradedComponent($component_grade);
```
KevinMackenzie: We should avoid this kind of logic in the database saving functions. Maybe put it in `saveGradedComponent`?

site/app/models/gradeable/GradedComponent.php
```diff
@@ -55,6 +55,10 @@ class GradedComponent extends AbstractModel {
     protected $graded_version = 0;
     /** @property @var \DateTime The time which this grade was most recently updated */
     protected $grade_time = null;
+    /** @property @var string The Id of the verifier who verified the grade */
+    protected $verifier_id = "";
```
KevinMackenzie: To keep this parallel with `grader`, I think we want to keep a verifier user object instead of just an id string.

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -677,7 +677,6 @@ public function renderRubricPanel(GradedGradeable $graded_gradeable, int $displa
         $this->core->getOutput()->addInternalJs('ta-grading-rubric-conflict.js');
         $this->core->getOutput()->addInternalJs('ta-grading-rubric.js');
         $this->core->getOutput()->addInternalJs('gradeable.js');
-
```
KevinMackenzie: unnecessary white space change in an otherwise unchanged file

site/public/js/gradeable.js
```diff
@@ -94,7 +94,8 @@ function prepGradedComponent(component, graded_component) {
     if (graded_component.grader_id === '') {
         graded_component.grader_id = undefined;
     }
-    if (graded_component.verifier_id === '') {
+
+    if (graded_component.verifier_id == '') {
```
KevinMackenzie: We should avoid `==`

KevinMackenzie: also, see other note about whitespace for this file too.

PR #3240, opened 2019-01-29
CodeMirror update

andrewaikens87:
This PR updates to the newest version of CodeMirror (Discussion forum and TA grading page both still seem to work with the new version).

### DIFF
--- a/.setup/bin/update_codemirror.sh
+++ b/.setup/bin/update_codemirror.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# Updates the CodeMirror (https://codemirror.net) dependency in the root site/ directory.
+
+set -e
+
+echo "Updating CodeMirror..."
+if [ ! -d /tmp/codemirror ]; then
+	wget https://codemirror.net/codemirror.zip -O /tmp/codemirror.zip
+	unzip /tmp/codemirror -d /tmp/codemirror
+fi
+VERSION=$(ls -1A /tmp/codemirror | grep -E -o "([0-9]+\.[0-9]+\.[0-9]+)")
+
+CODEMIRROR=/tmp/codemirror/codemirror-${VERSION}
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+DEST=${THIS_DIR}/../../site/public/
+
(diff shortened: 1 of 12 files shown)


MasterOdin:
Do we also need to update the parsers for the languages (python, clike, shell) or the themes (eclipse or monokai)?

Here's also a bash script I'd like you to include (at `.setup/bin/update_codemirror.sh`) that we should use over doing this manually:
```
#!/usr/bin/env bash

# Updates the CodeMirror (https://codemirror.net) dependency in the root site/ directory.

set -e

if [ ! -d /tmp/codemirror ]; then
    wget https://codemirror.net/codemirror.zip -O /tmp/codemirror.zip
    unzip /tmp/codemirror -d /tmp/codemirror
fi
VERSION=$(ls -1A /tmp/codemirror | grep -E -o "([0-9]+\.[0-9]+\.[0-9]+)")

CODEMIRROR=/tmp/codemirror/codemirror-${VERSION}
THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
DEST=${THIS_DIR}/../../site/public/

cp -f ${CODEMIRROR}/lib/codemirror.js ${DEST}/js/iframe
echo "// VERSION: ${VERSION}" | cat - ${DEST}/js/iframe/codemirror.js > /tmp/out && mv /tmp/out ${DEST}/js/iframe/codemirror.js
modes=( clike python shell )
for i in "${modes[@]}"; do
    cp -f ${CODEMIRROR}/mode/${i}/${i}.js ${DEST}/js/iframe
done

cp -f ${CODEMIRROR}/lib/codemirror.css ${DEST}/css/iframe
themes=( eclipse monokai )
for i in "${themes[@]}"; do
    cp -f ${CODEMIRROR}/theme/${i}.css ${DEST}/css/iframe
done

rm -rf /tmp/codemirror*
```

andrewaikens87:
This should be good to go now. I have included the bash script and the updated files. I also included the fix for #3269 in this as it is only one line.

bmcutler:
@andrewaikens87 
Remember to write some brief documentation on updating code mirror in the future using your script.  

Probably just put it at the bottom of this page for now:
https://submitty.org/developer/development_instructions
(we should rearrange things later)

PR #3253, opened 2019-01-31
Add warning to user if JavaScript is disabled

bmcutler:
See #3247


bmcutler:
Broke CSS when originally installed

MasterOdin:
Per @bmcutler's comment over email:
> Every page was missing most styling.
I could not reproduce this going from commit 033d14fe74ba81cec86bbb168f8b645308aded59 to a081edd6083eb4335ace606bd50eabe1826ba9ed on master.

I tried all pages on the sidebar (did not go too much further in-depth testing of pages as much of the functionality is broken), and they were all styled appropriately with the new error message. This seems like a case of an installation failure on production, over this code being the direct cause of the break.

PR #3254, opened 2019-01-31
Note Page Extraction Script

holzbh:
This script to get is much cleaner than the solution that we were using before in InstructorTools, which required copying a massive amount of extra information around.

A separate PR will update the README and remove the unnecessary file from the other repo.


bmcutler:
Can you fix the travis tests (looks like Python linting only?)

PR #3258, opened 2019-01-31
Log Autograding Exceptions

emaicus:
Adds logging for various stack traces created by errors caught during autograding.

Should be reviewed, as it was difficult to force good code coverage while testing.

Will likely require tuning as we decide which errors do and do not need to be logged by the system.

### DIFF
--- a/sbin/autograder/grade_item.py
+++ b/sbin/autograder/grade_item.py
@@ -77,6 +77,7 @@ def copy_contents_into(job_id,source,target,tmp_logs):
                 try:
                     shutil.copy(os.path.join(source,item),target)
                 except:
+                    grade_items_logging.log_stack_trace(job_id=job_id,trace=traceback.format_exc())
                     raise RuntimeError("ERROR COPYING FILE: " +  os.path.join(source,item) + " -> " + os.path.join(target,item))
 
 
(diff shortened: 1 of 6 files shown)


emaicus:
Merged Master and removed debug print statements. Ready for review.

sbin/autograder/grade_item.py
```diff
@@ -217,8 +218,11 @@ def grade_from_zip(my_autograding_zip_file,my_submission_zip_file,which_untruste
     # unzip autograding and submission folders
     tmp_autograding = os.path.join(tmp,"TMP_AUTOGRADING")
     tmp_submission = os.path.join(tmp,"TMP_SUBMISSION")
-    unzip_this_file(my_autograding_zip_file,tmp_autograding)
-    unzip_this_file(my_submission_zip_file,tmp_submission)
+    try:
+        unzip_this_file(my_autograding_zip_file,tmp_autograding)
+        unzip_this_file(my_submission_zip_file,tmp_submission)
+    except:
+        raise
```
bmcutler: Is this an incomplete edit?

emaicus: No, that syntax just means re-raise the exception. I added the try/except more as an acknowledgement that ```unzip_this_file``` is a frequent cause of exceptions than anything. I was logging the exception at first, but it is logged at the next level up, and if the copy fails, the function really shouldn't continue.

PR #3261, opened 2019-02-01
Global Container Image Specification

### LINKED ISSUE #3255: Add Global Docker Configuration Fields to Config JSON
Add editable global fields for docker use in config json.

emaicus:
Closes #3255

Added the ```default_container_image``` at the global level of the configuration json. If specified, this image will be used for all testcases. If unspecified, ```ubuntu:custom``` will be used as the default. The default can be overridden at the testcase level.

At the moment, this field is left global. It may be worth discussing whether or not ```default_container_image```, ```use_router```, ```docker_enabled```, and ```single_port_per_container``` should be bundled together into a global ```docker_options``` object. 

This PR does not require any gradeables to be rebuilt.

### DIFF
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -99,7 +99,14 @@ void AddDockerConfiguration(nlohmann::json &whole_config) {
     whole_config["docker_enabled"] = false;
   }
 
+  if(!whole_config["default_container_image"].is_string()){
+    whole_config["default_container_image"] = "ubuntu:custom";
+  }
+
+  std::string default_container_image = whole_config["default_container_image"];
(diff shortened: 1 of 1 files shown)


MasterOdin:
I like adding a top level (and per testcase?) `docker_options` object as we've added quite a few options now and it's possible we'll keep adding more, this adds some nice structure to the object. We probably don't need `docker_enabled` underneath the `docker_options` however, as the existence of the object probably implies that `docker_enabled` is true anyway.

I think we should drop the `default_` prefix on `default_container_image` and just use a top-level `container_image` that falls through onto all testcases unless a testcase specifies its own `container_image`.

bmcutler:
evan is revising and will reopen.

Issue #3223, opened 2019-01-21
Add Tool Tips to Forum Post Badges


KevinMackenzie:
Posts on the forum get badges like this:
![image](https://user-images.githubusercontent.com/3719964/51503066-03ae2880-1da7-11e9-97bd-6454a9117c53.png)
It would be helpful if these icons had some sort of tooltip to describe what each meant since its not labeled anywhere else.

drumilpatel2000:
@MasterOdin @shwehtom89 @KevinMackenzie  can you please briefly explain this issue to me ?

shwehtom89:
@drumilpatel2000 This issue was still open because the pin icon in the forum was missing a tooltip see #3378. The other icon tooltips were added previously in #3280.

Issue #3109, opened 2018-11-09
Keep track of which posts within a thread have been read


bmcutler:
[might be a duplicate]

determining what has been "read" should be pretty lenient.  If it has appeared on the screen, then yes it was read.  Other discussion interfaces seem buggy on this and sometimes you need to push a button "mark all as read" even though you have viewed everything.  Let's err in the other direction.

PR #3268, opened 2019-02-04
Forum Features: filter unread posts, post read status, icon tooltips

shwehtom89:
This PR resolves issues #3223 , #3109 and #3011. 

On any given thread, there are now indicators on which posts are new since the user last opened
the thread. 
newposts

Here read posts are shown with a gray bar and check icon with a tooltip. Unread posts are marked
dark blue and have a corresponding dot with a tooltip. This PR also changes some database queries 
to keep track of when the user last opened a thread rather than deleting timestamps.

This PR also includes a new option for filtering threads by new content.
filter
The option is a checked box and the selection is stored in a cookie.

Lastly there are tooltips for icons on the thread scrollbar.
tooltip
The icons for unresolved thread(?), resolved thread(?), announcement(*) now have tooltips.


andrewaikens87:
Please resolve merge conflicts

site/app/controllers/forum/ForumController.php
```diff
@@ -599,14 +599,15 @@ private function editPost(){
...
+    private function getSortedThreads($categories_ids, $max_thread, $show_deleted, $show_merged_thread, $thread_status, $unread_threads, &$blockNumber, $thread_id = -1){
         $current_user = $this->core->getUser()->getId();
         if(!$this->isValidCategories($categories_ids)) {
             // No filter for category
             $categories_ids = array();
         }
 
-        $thread_block = $this->core->getQueries()->loadThreadBlock($categories_ids, $thread_status, $show_deleted, $show_merged_thread, $current_user, $blockNumber, $thread_id);
+        $thread_block = $this->core->getQueries()->loadThreadBlock($categories_ids, $thread_status, $unread_threads, $show_deleted, $show_merged_thread, $current_user, $blockNumber, $thread_id);
+        // $this->core->getOutput()->renderJson($thread_block);
```
andrewaikens87: Please remove unnecessary comment

shwehtom89: This line was removed

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2587,9 +2623,10 @@ public function getPostsForThread($current_user, $thread_id, $show_deleted = fal
       }
 
       $result_rows = $this->course_db->rows();
-      if(count($result_rows) > 0){
-        $this->course_db->query("INSERT INTO viewed_responses(thread_id,user_id,timestamp) SELECT ?, ?, current_timestamp WHERE NOT EXISTS (SELECT 1 FROM viewed_responses WHERE thread_id=? AND user_id=?)", array($thread_id, $current_user, $thread_id, $current_user));
-      }
+      if(count($result_rows) > 0){ // insert only on first thread visit update otherwise
+        $this->course_db->query("INSERT INTO viewed_responses(thread_id,user_id,timestamp) SELECT ?, ?, current_timestamp WHERE NOT EXISTS(SELECT * FROM viewed_responses WHERE thread_id = ? AND user_id = ?)", array($thread_id, $current_user, $thread_id, $current_user));
```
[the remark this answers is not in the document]
shwehtom89: It should not be necessary. I have changed it back to how it was before

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -183,6 +183,23 @@ public function buildLoadThreadQuery($categories_ids, $thread_status, $show_dele
...
+            "EXISTS(
+                SELECT thread_id 
+                FROM (posts LEFT JOIN forum_posts_history ON posts.id = forum_posts_history.post_id) AS jp
+                WHERE(
+                    jp.thread_id = t.id
+                    AND NOT EXISTS(
+                        SELECT thread_id 
+                        FROM viewed_responses v 
+                        WHERE v.thread_id = jp.thread_id
+                            AND v.user_id = '{$current_user}'
```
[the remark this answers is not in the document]
shwehtom89: Added $current_user to the array $query_parameters. I've noticed are there are other times in buildLoadThreadQuery where php strings are put straight into queries. When should we use the '?' instead of the direct string placement?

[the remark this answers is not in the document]
MasterOdin: Generally never for stuff that're your testing against/setting in the query, especially anything that could be set by an end-user. The only real reason to have PHP variables within a query is because you're dynamically putting together a query (like there's a dynamic number of `?` within an insert), but otherwise use the `?` syntax to avoid potential SQL injections, even if you're totally positive that variable is sanitized and cannot be manipulated as a string by an end user.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2263,7 +2302,7 @@ public function existsAnnouncements($show_deleted = false){
     }
 
     public function viewedThread($user, $thread_id){
-      $this->course_db->query("SELECT * from viewed_responses where thread_id = ? and user_id = ?", array($thread_id, $user));
+      $this->course_db->query("SELECT * FROM viewed_responses v WHERE thread_id = ? AND user_id = ? AND NOT EXISTS(SELECT thread_id FROM (posts LEFT JOIN forum_posts_history ON posts.id = forum_posts_history.post_id) AS jp WHERE jp.thread_id = ? AND (jp.timestamp > v.timestamp OR (jp.edit_timestamp IS NOT NULL AND jp.edit_timestamp > v.timestamp)))" , array($thread_id, $user, $thread_id));
```
andrewaikens87: Do we need to also include edits? Not sure if this was discussed or not. If not, we can simplify this query to just be: `select 1 from viewed_responses v, posts p where v.thread_id = 6 and v.user_id='instructor' and p.timestamp > v.timestamp;`

site/public/js/server.js
```diff
@@ -1783,7 +1785,7 @@ function dynamicScrollContentOnDemand(jElement, urlPattern, currentThreadId, cur
     dynamicScrollLoadIfScrollVisible(jElement);
     $(jElement).scroll(function(){
         var element = $(this)[0];
-        var sensitivity = 3;
+        var sensitivity = 1.5;
```
andrewaikens87: From our discussion, this was modified to fix the vertical cutoff of the thread list? On chrome it is still being cutoff for me.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -183,6 +183,24 @@ public function buildLoadThreadQuery($categories_ids, $thread_status, $show_dele
...
+                WHERE(
+                    jp.thread_id = t.id
+                    AND NOT EXISTS(
+                        SELECT thread_id 
+                        FROM viewed_responses v 
+                        WHERE v.thread_id = jp.thread_id
+                            AND v.user_id = ?
+                            AND (v.timestamp >= jp.timestamp
+                            AND (jp.edit_timestamp IS NULL OR (jp.edit_timestamp IS NOT NULL AND v.timestamp >= jp.edit_timestamp))))))";
+            $query_parameters[] = $current_user;
```
andrewaikens87: As per comment below we might be able to simplify this. I will double check to see if edits should also be included.

site/app/libraries/database/DatabaseQueries.php
andrewaikens87: We should also include edits, please disregard this comment.

site/public/js/server.js
andrewaikens87: Seems like this is happening on master?? Not sure when this began...

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -838,7 +877,7 @@ public function getAverageComponentScores($g_id, $section_key, $is_team) {
         $return = array();
         $this->course_db->query("
 SELECT gc_id, gc_title, gc_max_value, gc_is_peer, gc_order, round(AVG(comp_score),2) AS avg_comp_score, round(stddev_pop(comp_score),2) AS std_dev, COUNT(*) FROM(
-  SELECT gc_id, gc_title, gc_max_value, gc_is_peer, gc_order,
+  SELECT gc_id, gc_title+, gc_max_value, gc_is_peer, gc_order,
```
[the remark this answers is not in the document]
shwehtom89: I don't recall modifying that function. That shouldn't be in there so I will remove it.

andrewaikens87:
The changes I made above will also fix the e2e tests that are currently failing on master (as a result of the new chrome update).

PR #3270, opened 2019-02-06
xlsx and csv user list upload

### LINKED ISSUE #2789: Upload Classlist Needs Better Errors
Uploading a classlist, if you specify a section that doesn't exist, on live will just give a generic frog robot error. The instructions also don't mention this requirement.

On a dev machine you can see the details:
```
Message:
SQLSTATE[23503]: Foreign key violation: 7 ERROR:  insert or update on table "users" violates foreign key constraint "users_registration_section_fkey"
DETAIL:  Key (registration_section)=(2) is not present in table "sections_registration".
CONTEXT:  Error occurred on dblink connection named "unnamed": could not execute command.
SQL statement "SELECT dblink_exec(db_conn, query_string)"
PL/pgSQL function sync_courses_user() line 18 at PERFORM

Strack Trace:
#0 /usr/local/submitty/site/app/libraries/database/PostgresqlDatabaseQueries.php(128): app\libraries\database\AbstractDatabase->query('
INSERT INTO courses_users (semester, course, user_id, user_group, registration_section, manual_registration)
VALUES (?,?,?,?,?,?)', Array)
#1 /usr/local/submitty/site/app/controllers/admin/UsersController.php(750): app\libraries\database\PostgresqlDatabaseQueries->insertCourseUser(app\models\User, 'f18', 'blank')
#2 /usr/local/submitty/site/app/controllers/admin/UsersController.php(43): app\controllers\admin\UsersController->uploadClassList()
#3 /usr/local/submitty/site/app/controllers/AdminController.php(58): app\controllers\admin\UsersController->run()
#4 /usr/local/submitty/site/public/index.php(240): app\controllers\AdminController->run()
```

### LINKED ISSUE #3214: Class list / Grader list upload improvements
As requested to me by email:

*  Detect misuse of list uploads (e.g. uploading a grader list via student management) and provide better error messaging.
*  XLSX file upload is broken.  Please fix.
*  Grader list / classlist upload logic is similar — please refactor into a common function.

This may also correct #2789.

pbailie:
fix #3214
fix #2789 

Fixes user list upload in manage students and manage graders.

- Combined and refactored code.
   - Two functions had mostly (but not all) the same copy/paste code.
   - Closure functions define divergent behavior between 'class list' and 'grader list' processing rather than peppering conditional blocks throughout the parse and process code.
   - Condensed uploaded data validation block to an easier-to-read "list style".
   - Location of `tmp` area to read/write scratch files for cgi scripts is no longer hard coded in relevant controller and cgi code — now is a new property in the `config` class, retrievable by `$this->core->getConfig()->getCgiTmpPath()`.  Property default is `/var/local/submitty/tmp/cgi`.
   - Better code commenting/documentation.
   - Required adjustments to Travis test code (regarding new property in config class)
- Spreadsheet upload is now working.
   - Using `/var/local/submitty/tmp/cgi` to write temp files for converting `xlsx` data to `csv` data.  This works around Linux systemd's "private tmp" policy (that was blocking spreadsheet conversion) without disabling that policy.
- Improved error messaging and exception handling.
   - Error message is shown only briefly.  Reduced message length so user can better read the error.
   - Database exceptions will no longer cause the robot to appear.  Instead, the error message will state that a database exception occurred without interrupting the UI.  Should be much less "scary".
   - ~~Issue #2789 should remain open.  The noted database exception is caught and handled, but the underlying issue is not specifically addressed in this PR.~~  Fix #2789 (DB exception thrown when registration section in uploaded user list didn't actually exist).  Now automatically creates registration sections as needed from user list upload (class list only).
- Bugfixes in data parsing.
   - UTF-8 BOM will no longer cause data validation to fail (MS Excel's primary `csv` save option will write the UTF-8 BOM).
   - Better handling of empty data rows whether it be an actual empty line or a line with value delimiters, but no values (such as a row of all commas in a `csv`).   MS Excel might write rows of all commas when saving a `csv`.
   - Empty rows no longer reorder row numbering (therefore any rows that fail validation won't get renumbered in the error message)

Not fixed:
- Detect type of upload by data present.
   - Not feasible since the data column in question (column 4: `registration_section` or `user_group`) often can have the same values for either class lists or grader lists.


bmcutler:
Hi Peter,
Can you please name the issue with the a more descriptive title?  Thanks
https://submitty.org/developer/how_to_contribute
Barb

mattraneri:
Just added a filter in the code to make sure that the CGI script was only being called from localhost (as the comment header was specifying). I don't think this script was ever called from the client, so this check shouldn't cause issues and will take care of relying solely on configuration.

pbailie:
Please understand that this PR deals primarily with Issue #3214 — as that was specifically requested of me.  I also worked on issue #2789 as it was closely related.

> I tested as much functionality as I could using csv files and found the following issues:
> * Preferred fields (firstname, lastname) could not be updated.
> * Updates do not appear to always occur (perhaps only for preferred fields, as above).

For now, preferred names will not be updated by userlist upload in order to preserve what has been set by individual users.  q.v. Issue #3511 (users.user_updated System Database Field Can Be Erroneously Overwritten To false)
 
> * A "Frog Robot" error was thrown when section was left blank.

Fixed.

> * I'm not certain if this is this PR or another issue which is fixed in master, but a "Password" field is visible when downloading users.
> * The password field is visible when downloading graders.

This is a form UI issue, unrelated to the issues I am working on.  Also, I would prefer more discussion before removing any UI elements, regardless of issue tracking.

> * The results of downloading users cannot be passed to upload classlist.
> * The results of downloading graders can't be passed to upload grader list.

CSV downloads contain _more_ data than is mandated by userlist upload.  Unless the downloaded data set is changed, it could never be used as an upload.  I also would not make changes to the download data set without discussion.

Also, most userlist uploads would occur at the beginning of the semester to populate enrollment in the absence of the student auto feed.  Using a CSV download wouldn't help in that case as it likely would be blank (or nearly so).

Note that it is not necessary to first download student enrollment to build a new or updated list.  e.g. A "late-add" student list that does not contain currently enrolled students is OK.
 
> * Uploading a csv to grader list always appears to result in the maximum number of updates, even if there are no updates to be made.

Fixed.

PR #3275, opened 2019-02-08
Add Global Container Defaults to Assignment Configuration

### LINKED ISSUE #3255: Add Global Docker Configuration Fields to Config JSON

emaicus:
Closes #3255

Added global ```container_options``` object to the autograding configuration json. Within, default values for ```container_image```, ```use_router```, and ```single_port_per_container``` may be specified.  

Changed the global flag ```docker_enabled``` to be a new string ```autograding_method```, which can currently be specified as ```docker``` or ```jailed sandbox```. If ```autograding_method``` is not set, ```jailed sandbox``` is used by default. If an assignment built before this pull request is graded, ```jailed sandbox``` is assumed.

### DIFF
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -108,48 +108,50 @@ void ArchiveValidatedFiles(nlohmann::json &whole_config) {
 
 void AddDockerConfiguration(nlohmann::json &whole_config) {
 
-  if (!whole_config["docker_enabled"].is_boolean()){
-    whole_config["docker_enabled"] = false;
+  //check if docker is globally enabled.
+  if (!whole_config["autograding_method"].is_string()) {
+    whole_config["autograding_method"] = "jailed sandbox";
+  }else{
+    assert(whole_config["autograding_method"] == "docker"
+        || whole_config["autograding_method"]  == "jailed sandbox");
   }
 
-  if(!whole_config["default_container_image"].is_string()){
-    whole_config["default_container_image"] = "ubuntu:custom";
+  //check if there are global container defaults present. If not, add an empty object.
+  if(whole_config["container_options"].is_null()){
+    whole_config["container_options"] = nlohmann::json::object();
   }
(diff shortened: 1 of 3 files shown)


emaicus:
This is failing because the Tutorial Repo hasn't yet been updated to use ```autograding_method```. 

This will be fixed by PR #21 in the Tutorial Repo.

grading/load_config_json.cpp
```diff
@@ -108,41 +108,50 @@ void ArchiveValidatedFiles(nlohmann::json &whole_config) {
 
 void AddDockerConfiguration(nlohmann::json &whole_config) {
 
-  if (!whole_config["docker_enabled"].is_boolean()){
-    whole_config["docker_enabled"] = false;
+  //check if docker is globally enabled.
+  if (!whole_config["autograding_method"].is_string()) {
+    whole_config["autograding_method"] = "jailed sandbox";
```
bmcutler: avoid spaces... switch to jailed_sandbox

grading/load_config_json.cpp
```diff
@@ -211,7 +217,7 @@ void AddDockerConfiguration(nlohmann::json &whole_config) {
 
       if(this_testcase["containers"][container_num]["container_image"].is_null()){
         //TODO: store the default system image somewhere and fill it in here.
-        this_testcase["containers"][container_num]["container_image"] = "ubuntu:custom";
```
bmcutler: where does ubuntu:custom come from now?

grading/load_config_json.cpp
emaicus: Fixed.

grading/load_config_json.cpp
emaicus: It comes from the assignment's default container. If there are no defaults set, the default field is populated as being ubuntu:custom on line 126.

PR #3281, opened 2019-02-11
VCS subdirectory will rebuild gradeable

KevinMackenzie:
Changing the VCS subdirectory for a gradeable will put the gradeable in the build queue (and update the form .json file) similar to how changing the autograding config will.  @bmcutler  is this what you were looking for?


bmcutler:
The bug was that form json was not correctly saving edits to the vcs path/directory
The timestamp on the form json file was changing, but the data in the file was stale
(passing along a secondhand bug report, I did not attempt to recreate)

PR #3285, opened 2019-02-12
Add a new and better datetimepicker

VictorLi718:
Add a new and better datetimepicker (open source). It is implemented through jquery and I include another two files for it. It works both in course material and in gradeable setting. 
Can customize a lot. Link: [flatpickr.js.org](flatpickr.js.org). 
screen shot 2019-02-11 at 19 48 42
screen shot 2019-02-11 at 19 44 21


bmcutler:
Is it missing seconds from the non-textbox data entry portion?  seconds are important.

VictorLi718:
Missed that option before. Should be good now.

site/app/templates/admin/admin_gradeable/AdminGradeableDates.twig
```diff
@@ -1,3 +1,6 @@
+<link rel="stylesheet" href="css/flatpickr.min.css">
+<script type="text/javascript" language="javascript" src="js/flatpickr.js"></script>
```
KevinMackenzie: Instead of using these `script` tags, we use a different method for adding CSS and JS scripts to the web page (See `addInternalJS` and `addInternalCSS` and their uses in `AdminGradeableController.editPage` and `newPage` as a guide)

site/app/templates/course/CourseMaterials.twig
```diff
@@ -1,5 +1,6 @@
 {% import _self as self %}
-<script type="text/javascript" language="javascript" src="js/jquery-ui-timepicker-addon.js"></script>
+<link rel="stylesheet" href="css/flatpickr.min.css">
+<script type="text/javascript" language="javascript" src="js/flatpickr.js"></script>
```
KevinMackenzie: see comment about script/link tags

VictorLi718: I think I have resolved both. Thank you for you advice.

KevinMackenzie:
I'm getting an error now saying that `flatpicker` is not defined when loading the edit gradeable page.  I get this in firefox and chrome.  Can you reproduce this problem?  The picker seems to work fine on the course materials page.

`Uncaught TypeError: $(...).flatpickr is not a function` on the line
`$(".date_picker").flatpickr(...)`

VictorLi718:
Yes I can reproduce it. But even when I add the AddInternal function twice for both newpage and editpage, it still doesn't work. I can't figure out why, but I guess it may be caused by we add the js and css to the AdminGradeableBase and then it calls the AdminGradeableDate and the js wasn't be passed? Could you please help me and see if there is any solution to solve this?

KevinMackenzie:
Yea, I can take a closer look at it

KevinMackenzie:
It looks like flatpickr isn't playing nicely with jQuery.  Using `flatpickr(".date_picker", {...});` instead of `$(".date_picker").flatpickr({...})` seems to fix the problem (this is the no-jquery method from https://flatpickr.js.org/getting-started/).

VictorLi718:
Thank you for your help. I think it works well now.

Issue #1045, opened 2017-06-09
Navigation Page Revisions & Progress Bars


bmcutler:
- [x] Revise the text on the buttons on the navigation page to match the design (screenshot below).
- [x] Revise the colors on each button as needed to match the design.
- [x] Revise the sorting order for items within each category to match the design.
- [x] Implement progress bar option for the buttons.
- [x] Grab the necessary data for the progress bars.
- [x] Improve the logic / organization / code structure of the code on this page.  Break into helper functions / revise variable names, etc.

Note:  for TA grading progress bar, this should be the fraction of your assigned grading that is complete.  If you are an instructor user and not assigned to any students, this should be the fraction of all TA grading that is complete.

Note:  This screenshot might be stale.  The current design is stored in a private Google doc.

<img width="1536" alt="screen shot 2017-06-09 at 3 53 42 pm" src="https://user-images.githubusercontent.com/5871417/26992277-ea1b0a20-4d2b-11e7-8f88-902fb22c619a.png">

Replaces issues #674 and #475

Related to / will also close issue #557....

PR #3301, opened 2019-02-19
Fixes grade button to display grading due date

shwehtom89:
Buster noted that the timestamp under the Grade button for gradeables displayed the grade release date instead of the grading due date. This PR resolves that issue.


holzbh:
The code does address my immediate concern, but creates a new one.

Since the two states that are relevant are "CLOSED" and then "GRADES AVAILABLE", I feel like we should be a little more informative in our box text in this case. Consider the following behavior that now will occur:

If an assignment has grades due today at 18:30 and grades released at 19:00, then at 18:45 the button still says "GRADE, (grades due [today's date] @ 18:30).

I know button text usually involves a bit of a ... conversation, so thoughts @bmcutler ? I would like it to say something like (grades due @ [grade due date]) when the grades aren't yet do, but after that I'd like the button to inform me when grades are supposed to be released and/or have text to indicate as a grader if I haven't finished my job, I'm now overdue.

shwehtom89:
That is a good point. I was thinking about a red/white (depending on how important you want something graded) button for the graders much like how students have a button for an overdue submission. 

The button can be titled something like "Overdue Grade, grades due [due date]@due time" and be labeled with the date that grades were due. Once grades are released, the button could be changed to "REGRADE". 

There is one small case that would probably need some discussion. Suppose a student was never graded and grades were released (maybe by mistake). Should a the graders be notified that this is a late grade or should it return to the default "REGRADE" button? Thoughts?

bmcutler:
This is not on our original button text & color design:
https://submitty.org/developer/interface_design_style_guide
https://github.com/Submitty/Submitty/issues/1045
Because we separated the due date from the release date which should add an extra row...
So I think we really should have "grades due ...." and after that date passes "grades will be released ... "

PR #3305, opened 2019-02-19
VCS gradeable changes

LeonMontealegre:
Changed how VCS is stored in DB and made the UI for creating a VCS gradeable way cleaner.

A VCS gradeable now has an int column that represents its "type":
-1 => not a VCS gradeable
0 => Submitty hosted - this gradeable url
1 => Submitty hosted - custom gradeable url
2 => Public Github
3 => Private Github

For Github, the user and repo information is stored in the .VCS_CHECKOUT file on submission in json.


MasterOdin:
I'm sure a lot of this was discussed within the weekly meeting, so use this as just commentary of an uninformed user. 

* What is the difference between a Github public and private repo? Are we going to allow Gitlab or Bitbucket ever and are we going to then have public/private options for these as well?
* I appreciate that there's been a simplification for "submitty" urls as there really isn't too many variations we've really seen (or could expect to see). Additionally, only showing team_id or user_id variable as allowed depending on team assignment or not is good.
* It's not clear if the middle bit in the "Submitty - Set URL" option is required or not (I expect it is)
* Are we explicitly never allowing students to choose either Submitty or external source for a repo? It must always either be Submitty or Github.
* How would we support Github Classroom (another VCS type option?

bmcutler:
This PR is just getting the UI for getting "github public" to work (I said I'd jump in and hook up the backend for this).

The next step (in a separate PR) will be getting "github private" to work, still need to decide where the submitty-github-user/credentials are stored, and whether its a shared server github-user or could be a per-course github user that the instructor creates and shares credentials with submitty).

And yes, a future step will be to allow any git or any vcs server to work.  But for right now we'll restrict the url pattern to github.

bmcutler:
submitty hosted has 2 options for non-teams.  Either the repo name matches the gradeable name.  Or if the repo is used for multiple gradeables (a project with multiple deadlines throughout the semester), and in that case the repo name can be anything.

MasterOdin:
> still need to decide where the submitty-github-user/credentials are stored, and whether its a shared server github-user or could be a per-course github user that the instructor creates and shares credentials with submitty).

Both options are technically supported through the use of ssh keys (you can specify within the .ssh/config to just try each key till you hit one that works). The only "risk" here is that technically, an instructor could then use this fact to checkout private repos from another course, but I'm not of the belief we need to worry about malicious behavior from instructors / head TAs.

bmcutler:
@LeonMontealegre please fix merge conflicts

LeonMontealegre:
@bmcutler 
Idk how to fix the Travis tests but it appears the master branch is also failing the tests so I assume it's not my code?

bmcutler:
@LeonMontealegre please fix merge conflicts
(sorry, again)

bmcutler:
I'm merging this now, the UI is good (good enough).  
But Submitty-hosted GIT repos are still messed up on the vagrant machine.
#3665 is work-in-progress to address these issues.
Then we still need to fix github hosted repos.

PR #3314, opened 2019-02-21
Add Support for "Server" Containers on a Network

emaicus:
To facilitate database grading, we now allow docker containers to be specified using the new ```server``` field. If a container is a server, it is understood that it will have no commands, and rather persists for the entirety of a testcase. In this way, an instructor can create a container which holds a database, but runs no commands. Other containers then connect to the server with their queries.

Of note: in order to allow postgres to initialize, the container holding a database can't be invoked as the untrusted user. @MasterOdin , what are your thoughts?

### DIFF
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -190,6 +190,7 @@ void AddDockerConfiguration(nlohmann::json &whole_config) {
     }
 
     bool found_router = false;
+    bool found_non_server = false;
 
     for (int container_num = 0; container_num < this_testcase["containers"].size(); container_num++){
       if(this_testcase["containers"][container_num]["commands"].is_string()){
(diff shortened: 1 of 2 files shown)


MasterOdin:
Without going too much into reviewing the code itself, I think it's fine that these "server" containers run as whatever user given that we're not expecting users to run commands directly against them and we're expecting we'll just be using well established server type images (like postgres).

One thing to note is how we're dealing with mounted volumes here. I'm not sure what "mounted_directory" points to here (is it just somewhere in `/var/local/submitty/autograding_TMP`?), but something we might want to consider that a lot of these services allow for initialization scripts on container creation. Postgres and MySQL for example both allow an end-user to place `*.sql`, `*.sql.gz`, and `*.sh` scripts in the `/docker-entrypoint-initdb.d` directory in the container and these are executed on container creation.  Is this something we potentially want to allow for?

PR #3321, opened 2019-02-26
Remove redundant link in sidebar

### LINKED ISSUE #2729: VPAT -- Redundant Link Alert
WAVE reports multiple "redundant Link" alerts with the new UI design (pull request #2697).
Each sidebar menu item has two links pointing to the same URL -- one link for when the sidebar is closed, and one link for when the sidebar is open.  This is detected as redundant links and could cause trouble with keyboard only users and screen readers.

Per WAVE:

> **What It Means**
> Adjacent links go to the same URL.
> **Why It Matters**
> When adjacent links go to the same location (such as a linked product image and an adjacent linked product name that go to the same product page) this results in additional navigation and repetition for keyboard and screen reader users.
> **How to Fix It**
> If possible, combine the redundant links into one link and remove any redundant text or alternative text (for example, if a product image and product name are in the same link, the image can usually be given alt="").
> **The Algorithm... in English**
> Two adjacent links go to the same URL.
> **Standards and Guidelines**
> [2.4.4 Link Purpose (In Context) (Level A)](http://webaim.org/standards/wcag/checklist#sc2.4.4)

drumilpatel2000:
Fixes #2729 
Remove rebundant link in side bar.the <a> showing icon is removed and `button.getTitle()`  is wrapped in span tag which is made  `Display : none`  on pressing sidebar when side bar is opened.

Icon are made large  when side bar is closed. This will increase accessibility of the button

### DIFF
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -128,6 +128,8 @@
                 </span>
             {% else %}
                 <a href="{{ button.getHref() }}"
+                    data-toggle="tooltip"
+                    title="{{ button.getTitle() }}"
                     {% if button.getClass() != "" %}
                         class="{{ button.getClass() }}"
                     {% endif %}
@@ -144,6 +146,7 @@
                         <span class="notification_badge">{{  button.getBadge() }} </span>
                     {% endif %}
                 </a>
+
             {% endif %}
         </li>
     {% endif %}
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -108,22 +108,6 @@
         <li class="short-line"></li>
     {% else %}
         <li>
-            {% if button.getIcon() != null %}
-                <a
-                    class="collapse-icon {{ button.getClass() }}"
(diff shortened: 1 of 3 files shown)


drumilpatel2000:
@pbailie please review this

MasterOdin:
For reference when reviewing...

Current style:
chrome-capture 1

New style:
chrome-capture

MasterOdin:
My view on the changing icon size (and subsequently position) is that I find it kind of jarring, especially given that I'd have to move my mouse to undo the action I just took (which is contrary to how most sites implement this kind of thing I think?).

I think I'd prefer that we make the icon just always a bit larger, no matter if minimized or not, and then when you hit minimized, the icon does not move/change size.

drumilpatel2000:
Thanks @pbailie for reviewing this. Size of fa icon is lowered to normal and some issue  like short-lie not visible properly in collapsed state is removed

drumilpatel2000:
@pbailie Please review . To align fa icons with text chrome demands for 2px padding and firefox demands for 2.5px . As previously chrome was given more preferences over firefox.Therefore I gave 2px padding to collapsed buttons

PR #3330, opened 2019-02-28
Fix date format inconsistency of re-autograding

### LINKED ISSUE #2827: Inconsistent date formats: submission time
On the submission page, after a re-autograde is run, the timestamps are in different formats:

    submission timestamp: 09/07/2018 07:46:13 AM
    days late: 0 (before extensions)
    grading time: 2 seconds
    
    number of re-autogrades: 1
    last re-autograde finished: 2018-09-07 12:15:26-0400

zjxiaohan:
Closes #2827 

The date format of re-autograding is changed to `date('m/d/Y h:i:s A')`, which is consistent with the original submission timestamp.

re-autograde


bmcutler:

PR #3331, opened 2019-03-03
[Bugfix] Fix backward queue order in autograding

### LINKED ISSUE #3318: Queue order is backwards
When you submit you get put at the end of the queue. That is reported as "1 out of 5" ... which means there are 4 others in front of you.

This is backwards from how you'd expect a queue to work, where being first means nobody is in front of you. Just needs a invert and it should be good.

Quarz0:
Closes #3318


MasterOdin:
Thanks for fixing this bug!

Issue #2680, opened 2018-08-08
"Edit rubric" mode additions


bmcutler:
_[collected from below]_
- [x] Make text box when creating/editing mark text bigger (allow multiline marks)
- [x] Add a hotkey to toggle between non edit rubric mode and edit rubric mode
- [x] Toggling edit rubric and non edit rubric mode should not close/collapse the component

bmcutler:
See also #2678 (for new images)

bmcutler:
And let's add a hotkey to toggle between non edit rubric mode and edit rubric mode

bmcutler:
(I think I said this elsewhere, but I can't find it now), 
toggling edit rubric and non edit rubric mode should not close/collapse the component.
Yes, save the current state of the mark checkboxes (the grading).  
But hopefully you can do that without visually collapsing the component.

bmcutler:
Finally Closed

PR #3339, opened 2019-03-06
Switching to/from edit mode won't close component

KevinMackenzie:
Part of #2680 

when toggling the 'edit mode' of the TA grading rubric, the component will not close/open itself.  This changes some pretty important parts of the code, so marked as WIP so it can be more thoroughly tested.

Changes that should be made:
- [x] 'Loading' message should show up when switching modes since it takes a a couple seconds


KevinMackenzie:
My solution prevents the user from rapidly toggling the edit mode button, which is the root cause of many potential problems.

emaicus:
Just tested the PR again. The problem is still present. Better than stopping rapid toggling, can we look into why marks are duplicated to begin with, and the source of the ```Error saving component! Missing mark id in order array and duplicates all new marks.``` message?

KevinMackenzie:
From what I can tell, the event for saving the components is being triggered multiple times before the first one completes.  Multiple identical events are being handled while the asynchronous ajax requests are being completed, so the data is getting into a bad state and the same 'new' marks are being submitted as 'new' every time.  The solution I'm trying to employ is preventing multiple simultaneous events from firing.  I'll find a more robust way to do this.

KevinMackenzie:
I found the source of your mark order error @emaicus.  There was some function composition that wasn't done correctly.  Now you should be able to toggle edit mode as quickly as you want and it shouldn't give you the same errors (or at least not for the same reason).  This should also fix some similar problems when trying to verify a component and edit the rubric in rapid succession.

PR #3341, opened 2019-03-07
VPAT -- TAB key allowed to indent in text area, ESC key navigates to next control.

### LINKED ISSUE #3146: VPAT -- Please Do Not Trap The Tab Key
The Tab key is integral to keyboard navigation, and therefore integral to web accessibility.

Some portions of Submitty's UI intends to trap the tab key into a control element (such as a textbox), preventing the tab key from navigating to another control element.  This _**will**_ severely interrupt a keyboard user's ability to properly navigate through Submitty's UI.

Please correct the UI so that the Tab key is _**always**_ usable to navigate through a page.

Here is a code example of trapping the tab key.  There could be more instances.
https://github.com/Submitty/Submitty/blob/c99b355092dd43cfae0babee2fce28e9c54d744b/site/app/templates/submission/homework/SubmitBox.twig#L334-L358

pbailie:
Fixes #3146 

VPAT requires that all controls can be navigated to and from via keyboard, and `TAB` is the default key to advance through a page.  Trapping the `TAB` key, for text indentation, requires an alternate keyboard control to navigate forward from a textarea.  The `ESC` key has been designated for this use.  Furthermore, `SHIFT+TAB` has been re-enabled to navigate backwards from a textarea.  

- Removed some duplicated code that handled `TAB` keypress events as indentation.
- Also, this PR includes control `<label>` VPAT corrections in _SubmitBox.twig_.


pbailie:
> `esc` implementation is working fine but their are some minute bugs:-
> 
>     1. After pressing `esc` it goes to directly to submit
> 
>     2. After pressing `esc` selected tag is not highlighted

The code specifically builds a list of all interactive controls, finds the one with focus (the textbox) and when `ESC` is pressed, sends focus to the next control in that list.  Depending on the form, that could be submit or that could be something else.  But navigating to the next control (submit or otherwise) would be expected behavior with `tab`, if `tab` wasn't being redefined.

The selected control not being highlighted is likely an issue with CSS on that control.  My testing showed that the next control is getting focus and would be properly interactive even when focus is not visible.  Therefore, what I had set out to do (provide a keyboard shortcut to leave a textbox where `tab` has been redefined) is working.  Fixing the CSS on that control is a separate issue.

Thank you for the review.

pbailie:
Opened Issue #3535 (VPAT — checkbox focus is not visible in the discussion forum form) to address point 2 of previous code review.

drumilpatel2000:
#3538 This may be able to solve #3535 and this PR may be merged

PR #3350, opened 2019-03-07
Fixed upload message spacing

xypnox:
The upload message before this fix was misspaced around the word Submit.

screenshot_2019-03-08 development

This fix corrects the same.

### Notes

- I have also tried using `\"` to escape the quotations but it didn't work. Also, using html tags eg. `<b>` didn't work either. Hence I have removed the quotes altogether.


MasterOdin:
@bmcutler Do we want to make a migration to go through all existing courses and fix this typo via sed?

bmcutler:
No migration necessary.

But I would like to eventually solve the underlying problem.  

Instructors should be able to reliably provide some (restricted) formatting / special characters on strings in various places in the system.

MasterOdin:
I'd probably suggest we move away from ini files then if we want to really let instructors use symbols and things. INI files really aren't the ideal format for moving beyond just simple switches/numbers and into allowing abritrary strings, especially given that we've got to parse and use this file in both Python and PHP.

I'd suggest we either ditch this format for JSON (to be consistent with the rest of the system configuration is done) and it's human readable enough for our needs that I think any instructor/sysadmin would be able to easily grep what they need.

bmcutler:
100% support moving everything to .json.
I never understood why we had any .ini files.

prestoncarman:
I want to second the idea of moving to .json files. It would be nice to store more configuration information in JSON especially for gradeable details.

MasterOdin:
Thanks for the PR. However, closing this in favor of #3375 as that will fix the spacing bug as well as keep the `"` around submit which was what was originally intended.

PR #3352, opened 2019-03-07
TCLAP Fixes and Display Environment Variable Tweaks

emaicus:
* Speeds up tclap install by avoiding installation of examples and tests.
* Windowed assignments no longer run if the display environment variable is not set.
* Should fix lingering issues where the display environment variable is not properly set.

I was unable to reproduce the permissions issues via either migration or fresh install. The directories and files under ```/usr/local/include/tclap``` are all owned by root, but have other read which allows them to execute as expected. @bmcutler , can you comment on the issues and what you had to change the permissions to?


migration/migrator/migrations/system/20190125144723_cpp_argument_parsing.py
```diff
@@ -12,6 +12,7 @@ def up(config):
     os.system("tar -xpzf {0}".format(TCLAP_TAR))
     os.remove(TCLAP_TAR)
     os.chdir(TCLAP_DIR)
+    os.system("sed -i 's/SUBDIRS = include examples docs tests msc config/SUBDIRS = include docs msc config/' Makefile.in")
```
MasterOdin: Are we re-running this migration as right now this line won't have any effect on the live system (which had this migration run today).

We could also just rollback both the python migration and this one and then re-execute this one if we wanted without any danger.

emaicus: That shouldn't be necessary. This change just makes the migration run faster by stopping it from compiling TCLAP's built in tests. Besides a speedup, nothing else should change.

bmcutler:
Maybe the permissions are the same umask issue @MasterOdin and I re-discovered with pip today.
Lots of places in the installation we manually set permissions after installing things.

The problem was that the directory did not have o+rx.  The tclap source files did have o+r, but the daemon or instructor user running BUILD_XXXX.sh hw01 didn't have read access to those files so gradeable build failed.


This is also still missing the automatic install (migration) of tclap on worker machines.  (or maybe that's not necessary?  I'm getting myself confused)

MasterOdin:
Would it make sense to move the umask line in install_system.sh from around the beginning of it to around the end (just before run stuff like setup_sample_courses.py) so that we're not potentially getting bit by umask issues during installation?

bmcutler:
No, it's not just a rearrange lines question.

I thought  (and I still hope) our design was to allow install_system.sh should be re-run on an existing installation without failure and to patch / upgrade anything that might not have been correctly handled by a migration, or fix things that were manually un-installed or broken.

MasterOdin:
Oh yeah, I keep forgetting that's a thing we want to maintain (and should still work?). Never mind me.

PR #3355, opened 2019-03-08
Move to Xenial and fixing e2e tests on Travis

### LINKED ISSUE #2668: Investigate and fix e2e submission tests on Travis
Continued work from the merged PR #2364, which where we've skipped the submission testcases for now on Travis. Following #2660 which would simplify (and make the test less brittle on finding the appropriate buttons), we should re-investigate getting the submission e2e tests to pass on Travis.

### LINKED ISSUE #3257: Test xenial environment for Travis test suite
https://blog.travis-ci.com/2018-11-08-xenial-release

MasterOdin:
Closes #3257
Closes #2668

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -91,6 +91,24 @@ if [ ${WORKER} == 0 ]; then
     python3 ${SUBMITTY_REPOSITORY}/migration/run_migrator.py -e system -e master -e course migrate
 fi
 
+################################################################################################################
+################################################################################################################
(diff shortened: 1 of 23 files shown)


MasterOdin:
The big thing here is that whatever we use to compile TestRunner.java must be the same version we use for running the testcase (which is right now /usr/bin/java) or else it'll throw an error. When not using the full path, Travis-CI was using something out of /opt for compiling and /usr/bin for the grading process causing partial testcase failure.

Do we want to insert a note in execute.cpp to link to INSTALL_SUBMITTY_HELPER.sh mentioning this in case we were to ever update the path from /usr/bin to somewhere else while still leaving a version of Java under /usr/bin?

PR #3358, opened 2019-03-08
Error Handling on Login Page

### LINKED ISSUE #3353: Redirection not handled properly on login page
### Reproduction of Error
Error is thrown when we press login without entering the password.This does not implies wrong password but password = `NULL`
screenshot from 2019-03-08 04-43-26
screenshot from 2019-03-08 04-45-04

### Possible Solution
Front end javascript form validation can be provided .Such that when password or login = `NULL` It throws an error and form is not submitted.

drumilpatel2000:
Fixes #3353 

Fatal error is given when password  is empty provided USER_ID is not empty and points at a real user id. This problem can be easily solved by providing required attribute to both of `Password` and `User_id` and ensuring that they're filled in and non-blank on the front-end as well as make sure backend handles the case where the fields could be blank, returning a failed authentication attempt instead of throwing an error.

### DIFF
--- a/site/app/authentication/PamAuthentication.php
+++ b/site/app/authentication/PamAuthentication.php
@@ -18,7 +18,7 @@ use app\libraries\FileUtils;
  */
 class PamAuthentication extends AbstractAuthentication {
     public function authenticate() {
-        if ($this->user_id === null || $this->password === null ||
+        if ( strlen($this->user_id) === 0 || strlen($this->password) === 0 ||
             $this->core->getQueries()->getSubmittyUser($this->user_id) === null) {
             return false;
         }
--- a/site/app/authentication/PamAuthentication.php
+++ b/site/app/authentication/PamAuthentication.php
@@ -18,7 +18,8 @@ use app\libraries\FileUtils;
  */
 class PamAuthentication extends AbstractAuthentication {
     public function authenticate() {
-        if ( strlen($this->user_id) === 0 || strlen($this->password) === 0 ||
+        // Check for $this->user_id and $this->>password to be non empty
+        if ( empty($this->user_id)  || empty($this->password) ||
             $this->core->getQueries()->getSubmittyUser($this->user_id) === null) {
             return false;
         }
--- a/site/app/templates/Authentication.twig
+++ b/site/app/templates/Authentication.twig
@@ -11,8 +11,8 @@
             {% for key, value in old %}
                 <input type='hidden' name='old_{{ key }}' value='{{ value }}' />
             {% endfor %}
-            <input type="text" name="user_id" placeholder="User ID" autofocus/><br />
-            <input type="password" name="password" placeholder="Password" /><br />
+            <input type="text" name="user_id" placeholder="User ID" required autofocus/><br />
+            <input type="password" name="password" placeholder="Password" required /><br />
             <label for="stay_logged_in">Stay Logged In</label> <input type="checkbox" name="stay_logged_in" id="stay_logged_in" checked /><br />
             <input type="submit" name="login" value="Login" class="btn btn-primary" />
         </form>
--- a/site/cgi-bin/pam_check.cgi
+++ b/site/cgi-bin/pam_check.cgi
@@ -1,9 +1,15 @@
 #!/usr/bin/env python3
 
 """
-Given a filaname, try to open that file, which should contain a JSON object
-containing a username and password, then test that against PAM printing out
-a JSON object that is authenticated and is true or false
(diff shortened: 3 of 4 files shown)


drumilpatel2000:
@MasterOdin and @pbailie  Please review my PR suggest any changes

MasterOdin:
Before merging, I'd like to look at why the fatal error is happening on the backend and see if that's something that needs fixing/can be fixed easily as well.

drumilpatel2000:
@MasterOdin Please review my PR. Issue faced by previous version was that It was assuming `$this->password` to give `NULL` for no value but It is a string with 0 length.So previous backend form validation were not working.
So I changed `$this->password === null` to `strlen($this->password) === 0`

drumilpatel2000:
@MasterOdin I don't know why travis CI is giving error but You can see my changes they cannot produce a running error s i have just changed If statement

MasterOdin:
It's because you marked your PR as a [WIP] in your title. To save on Travis runtime, any PR marked as such auto-fails immediately. Remove it and I can re-run Travis to have it run.

drumilpatel2000:
Thanks for reply @MasterOdin . I have removed [WIP] and my PR is ready to be merged

site/app/authentication/PamAuthentication.php
```diff
@@ -18,7 +18,7 @@
  */
 class PamAuthentication extends AbstractAuthentication {
     public function authenticate() {
-        if ($this->user_id === null || $this->password === null ||
+        if ( strlen($this->user_id) === 0 || strlen($this->password) === 0 ||
```
MasterOdin: Let's change this to `empty($this->user_id) || empty($this->password)` to retain the previous logic of preventing null as well as the additional logic of protecting against empty strings.

drumilpatel2000: Thanks for reviewing my PR @MasterOdin . I have changed to `strlen(x) === 0' to `empty(x)` but I will be taking time for adding testcases because I don't know how to add testcases.Can you please guide me to this particular issue I am facing ?

drumilpatel2000: I probably think they are already present in https://github.com/Submitty/Submitty/blob/master/site/tests/app/authentication/PamAuthenticationTester.php as following name:-
1) `testNoUserId()`
2) `testNoPasswordId()`

MasterOdin: Sure, we use [PHPUnit](https://phpunit.de/) to handle testing the PHP code and that lives under site/tests with the file I linked above (PamAuthenticationTester) handling unit testing of this authentication method. Generally, each Tester class will either extend from the TestCase class of PHPUnit or our custom BaseUnitTest class that contains a number of helper functions for us to create mocks and things that we use in a bunch of tests and would be cumbersome to copy/paste all around the test suite.

For this particular issue, the best thing to base your work off would be:
https://github.com/Submitty/Submitty/blob/fa90f1e56a37b8ae799458726601d185a48e96b0/site/tests/app/authentication/PamAuthenticationTester.php#L66-L72
where you would use that function to create two new ones:
- testEmptyUsername
- testEmptyPassword

where the first test would be just replacing the call to `setUserId` with an empty string and the second test would be adding a new line that would be `$pam->setPassword('');` while having something non-empty passed to `setUserId`.

Line by line explanation of what the test above does is:
1. create a [mock / test-double](https://phpunit.de/manual/6.5/en/test-doubles.html) for our Core class (though it's not really used for this test)
2. Instantiate an instance of the PamAuthentication class which is what we're testing
3. Set the userId for the PamAuthentication instance
4. Assert that the return of calling the `authenticate` function is false.

https://phpunit.de/manual/6.5/en/appendixes.assertions.html is a good place to look through for all of the different types of asserts you can use in your code. For this, assertFalse is fine.

MasterOdin: The existing `testNoUserId` and `testNoPassword` test functions only validate correct behavior if the values are null for user_id and password respectively. I'd like to add testcases for if we get an empty string since this was a new bug and to prevent any sort of regressions going forward.

drumilpatel2000: Thanks for helping me @MasterOdin I will take some time to learn `PHPUnit` and I will commit all left changes soon.

drumilpatel2000: @MasterOdin please review at my PR. I have added two `PHPUnit` testing;

MasterOdin:
Thanks for the PR and bugfix!

PR #3360, opened 2019-03-08
Fix Incorrect Submission Time Displayed during Grading

### LINKED ISSUE #3260: Submission time displayed as current time if waiting to be graded
Not sure if this is a git-based bug but, while an assignment is being graded, the submission timestamp on the submission page is rendered as the current system time. This seems to be just a visual bug as the timestamp reverts once the assignment is autograded.

zjxiaohan:
Closes #3260 

During the grading process, the submission time is incorrectly displayed as current time. This happens because the submission time displayed is loaded from `history.json` generated by the autograder, which is only produced **after** grading. Therefore, the submission time passed to the front-end during the grading process is `null`, and in this case, by default, [twig displays current time](https://twig.symfony.com/doc/2.x/filters/date.html).

To fix this bug, I changed the source of submission time from `history.json` to `$version_instance->getSubmissionTime()`. I tested it and it worked well.

I did do a global search and found that the original `db_submission_time` and `history_submission_time` were not used anywhere else. However, as I am not familiar with the codebase, could you please help me check if I would introduce new bugs by making these changes?


zjxiaohan:
Sorry. I just found that there are some extra unnecessary commits in this PR. I will fix it next time. For now, if this is squash merged to master eventually, please help remove the unrelated commits like e632e81 from other issues. My apologies for that.

MasterOdin:
No problem with the git history here. We squash and merge all PRs when accepted to help keep the history of master clean and since e632e81 has already been merged into master, it has no effect on this PR's changeset when merged into master.

PR #3362, opened 2019-03-09
[Bugfix] http for git push

bmcutler:
Don't know how things were working before.
This allows users to write to a git repo over http.

To patch existing repos, go to /var/local/submitty/vcs/git/<SEMESTER>/ and run:

for i in */*/*; do echo $i; pushd $i; git config --file config http.receivepack true; popd; done


zjxiaohan:
May I have the reason why this PR is closed? I did come across this problem when I tested git submission locally. It seemed that currently git had to be set up over https and I could not find instructions on that.

PR #3363, opened 2019-03-09
Easier way to un-resolve a resolved forum thread

### LINKED ISSUE #3324: Easier way to un-resolve a resolved forum thread
use case:

student asks question

instructor answers question (and believes problem is solved, so presses the "mark as resolved" button)

student (possibly same/possibly different student) wants to ask a followup/extension of the original question and may or may not want to mark the thread as unresolved again should see 2 blue buttons at the bottom of their response:  "Submit Reply to All"   or "Submit Reply to All & Mark as Unresolved" . (Only do this if a thread is in 'resolved' state.  If a thread is in 'comment' or 'unresolved' state just have submit reply to all button'.)

Quarz0:
closes #3324

### DIFF
--- a/site/app/controllers/forum/ForumController.php
+++ b/site/app/controllers/forum/ForumController.php
@@ -41,6 +41,10 @@ class ForumController extends AbstractController {
             case 'publish_post':
                 $this->publishPost();
                 break;
+            case 'publish_post_unresolve':
+                $this->changeThreadStatus(-1);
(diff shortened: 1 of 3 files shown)


holzbh:
This has been stale for a while, and needs to have merge conflicts resolved. The conflicting file has changed quite a bit since the PR was opened. Please re-open if you make updates.

PR #3368, opened 2019-03-10
Fixed position of edit name button

xypnox:
The button to edit the name in the homepage is positioned such that it indicates only that the First name can be edited.
Screenshot_2019-03-10 Submitty

This PR fixes this visual problem by moving the edit button beside the *About You* heading. Indicating that the edit will affect subsequent entries.

Screenshot_2019-03-10 Submitty(2)

### DIFF
--- a/site/app/templates/HomePage.twig
+++ b/site/app/templates/HomePage.twig
@@ -1,7 +1,7 @@
 <div class="content">
     <div class="row">
         <div class="box col-md-6">
-            <h1>About You</h1>
+            <h1>About You <a onclick="userNameChange()" class="edit-name-button"><i class="fas fa-pencil-alt" aria-hidden="true"></i></a> </h1>
             <table>
                 <tbody>
                 <tr>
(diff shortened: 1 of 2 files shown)


MasterOdin:
@bmcutler, I believe if using DB authentication, then this icon is also how you change your password, but I'd have to double check tomorrow.

xypnox:
@bmcutler Do you mean a tooltip like the following (shown while hovering over edit button)?
Peek 2019-03-18 23-22

MasterOdin:
Here's how the system looks when you are using DatabaseAuthentication:

Screenshot 2019-03-20 08 01 17

If we're moving the pencil icon up to the "About You" line, we'd probably want to combine the modals of editing preferred name as well as database password? Thoughts @bmcutler?

bmcutler:
Thanks for this PR, but I think we'll stick with the current design for now.   When we have more information to edit on this page, we may redesign.

PR #3371, opened 2019-03-11
Show git commit hash

### LINKED ISSUE #3309: "Git commit hash" on submit page is null
On the submission page, the git commit hash always displays None. Not sure exactly where the value is being lost.

zjxiaohan:
Closes #3309 

This problem occurs because the `revision` is written to `result.json` instead of `history.json` while the submission box reads from `history.json`

After:
git hash

### DIFF
--- a/sbin/autograder/grade_item.py
+++ b/sbin/autograder/grade_item.py
@@ -698,10 +698,7 @@ def grade_from_zip(my_autograding_zip_file,my_submission_zip_file,which_untruste
         json.dump(queue_obj,outfile,sort_keys=True,indent=4,separators=(',', ': '))
 
     try:
-        with open(os.path.join(tmp_work,"results.json"), 'r') as read_file:
-            results_obj = json.load(read_file)
-        with open(os.path.join(tmp_results,"results.json"), 'w') as outfile:
-            json.dump(results_obj,outfile,sort_keys=True,indent=4,separators=(',', ': '))
+        shutil.move(os.path.join(tmp_work, "results.json"), os.path.join(tmp_results, "results.json"))
     except:
         with open(os.path.join(tmp_logs,"overall.txt"),'a') as f:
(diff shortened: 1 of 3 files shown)


sbin/autograder/grade_item.py
```diff
@@ -699,8 +700,6 @@ def grade_from_zip(my_autograding_zip_file,my_submission_zip_file,which_untruste
     try:
         with open(os.path.join(tmp_work,"results.json"), 'r') as read_file:
             results_obj = json.load(read_file)
-        if 'revision' in queue_obj.keys():
-            results_obj['revision'] = queue_obj['revision']
         with open(os.path.join(tmp_results,"results.json"), 'w') as outfile:
             json.dump(results_obj,outfile,sort_keys=True,indent=4,separators=(',', ': '))
     except:
```
MasterOdin: We can also remove this entire try/except as well since it's no longer doing anything useful.

zjxiaohan: @MasterOdin Thank you for your review. `results.json` seems important for integration test like [here](https://github.com/Submitty/Submitty/blob/348e4f29ced6afeea6c7e6d9fcbdf2f4e8192378/tests/integrationTests/tests/cpp_hidden_tests/__init__.py#L60) if I understand correctly.

Do you mean that this `json_diff` is no longer used, or you mean that it should be removed when it is not under a testing environment?

MasterOdin: Oh, I missed that we were writing the results.json from one location and writing it to another.

Sorry, let's rewrite that read/write operation to just be `shutil.move(os.path.join(tmp_work, "results.json"), os.path.join(tmp_results, "results.json"))`

zjxiaohan: Done. Directly moving saves time and is simpler. Thank you for your suggestion!

PR #3373, opened 2019-03-11
[Bugfix] Make sure points is multiple of precision

### LINKED ISSUE #1055: In the TA grading rubric, the TA can not use the arrow keys to enter in a non whole number.
In the TA grading rubric, the TA can not use the arrow keys to enter in a non whole number. For example, when the maximum point value for a question is 3.5, the TA may not use the arrow keys to advance towards 3.5. However, the TA may manually put in 3.5. However, even though that value may stay in that box, the rubric will not accept it as valid and request the TA to put a number no more than 3 even though 3.5 was specified as the maximum. This seems to be the case for regular, extra, and penalty credit.

J16053:
Closes #1055 

This is my first PR, so any feedback would be greatly appreciated 😄 

When the instructor tries to edit the rubric, the points box will turn red if an invalid point value is entered and the point value is not saved. This should prevent rubrics that lead to the issue above from being created.
image

### DIFF
--- a/site/public/js/ta-grading-rubric.js
+++ b/site/public/js/ta-grading-rubric.js
@@ -2080,22 +2080,18 @@ function onComponentPointsChange(me) {
  * @returns {boolean}
  */
 function dividesEvenly(dividend, divisor) {
-    var mult = Math.max(multiplier(dividend), multiplier(divisor));
-    return ((dividend * mult) % (divisor * mult) === 0);
+    var multiplier = Math.pow(10, Math.max(decimalLength(dividend), decimalLength(divisor)));
+    return ((dividend * multiplier) % (divisor * multiplier) === 0);
 }
 
 /**
- * Returns power of ten to multiply by to make num an integer
+ * Returns number of digits after decimal point
  * @param {number} num
  * @returns {int}
  */
-function multiplier(num) {
-    var parts = num.toString().split('.');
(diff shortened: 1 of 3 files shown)


site/public/js/ta-grading-rubric.js
```diff
@@ -2055,16 +2055,48 @@ function onClickCountDown(me) {
...
+ * Returns power of ten to multiply by to make num an integer
+ * @param {number} num
+ * @returns {int}
+ */
+function multiplier(num) {
+    var parts = num.toString().split('.');
+    if (parts.length < 2) { // not a decimal
+      return 1;
+    }
+    return Math.pow(10, parts[1].length);
```
zjxiaohan: This function could return the length first to save some time for computing exponentiation. (Computing exponentiation is fast on a modern computer indeed, but I think it's always better to write faster codes :smile: )

```
function dividesEvenly(dividend, divisor) {
    var mult = Math.pow(10, Math.max(multiplier(dividend), multiplier(divisor)));
    return (dividend * mult) % (divisor * mult) === 0;
}

function multiplier(num) {
    return (num.toString().split('.')[1] || '').length;
}
```

J16053: Great idea! Will make that change 👍

bmcutler:

PR #3390, opened 2019-03-13
Admin Customize Website Theme File Upload Bug Solution Proposed

duggalrd:
Hey, when playing around with the site, I noticed when the Instructor wanted to customize the Submitty Website Theme, the code did not check which file type was being submitted. For example, before I could submit file type 'python' for Top Bar Html file upload. I fixed this bug and added file upload type checks as seen in the commit.


bmcutler:
Thanks for the contribution, please re-open when you have addressed the comments above

PR #3395, opened 2019-03-15
[Bugfix] Add missing version parameter in downloadZip()

### LINKED ISSUE #3400: Download zip in submission panel gives server error
On my vagrant VM, i got this error after clicking on Download Zip in Submission panel.
Screenshot from 2019-03-15 21-11-23

Sahil333:
Found this bug while trying to download submission zip in TA grading panel. The version number was missing in Request parameter.
Screenshot from 2019-03-15 21-11-23
Closes #3400


Sahil333:
What's the intended purpose of keeping version as an optional parameter in View? Say I allow the version to be optional then by default always the active version will be downloaded. The grader will not be able to download no-active/no-grading versions. This behavior persists even if grader selects a no-grading version in Student Information panel and then click on download zip.
https://github.com/Submitty/Submitty/blob/cd100b2cf14e434ebd99c19b172334e847b5adf7/site/app/controllers/MiscController.php#L321-L328

PR #3397, opened 2019-03-15
[Bugfix] Use cgi_url from config file if non-empty else build from base_url

### LINKED ISSUE #3335: submitty.json cgi_url value is ignored
The **cgi_url** value in the **submitty.json** config file is not used by the code anymore.  Model **Config** builds **cgi_url** from the **base_url** value.  **base_url** is derived from the **submission_url** value in  **submitty.json**

I ran into this when attempting to allow remote machines to access the vagrant install running on my dev machine.   Apache submitty config was modified to listen on all IP addresses.  Vagrant was configured to forward host 11132 to guest 80.

**submitty.json** 
```
  "submission_url": "http://mypublic.rpi.edu:11132",
  "vcs_url": "",
  "cgi_url": "http://127.0.0.1/cgi-bin",

```
resulted in this error.  Note cURL trying to access cgi-bin at the host address.

robot1

I was able to work around this, but the reference to **cgi_url** should be removed from the setup scripts and config files if it cannot be used.

Sahil333:
Closes #3335

### DIFF
--- a/site/app/models/Config.php
+++ b/site/app/models/Config.php
@@ -256,9 +256,14 @@ class Config extends AbstractModel {
         }
 
         $this->timezone = new \DateTimeZone($this->timezone);
-
         $this->base_url = rtrim($this->base_url, "/")."/";
-        $this->cgi_url = $this->base_url."cgi-bin/";
+
+        if (!empty($submitty_json['cgi_url'])){
+            $this->cgi_url = rtrim($submitty_json['cgi_url'], "/")."/";
+        }
+        else {
+            $this->cgi_url = $this->base_url."cgi-bin/";
+        }
 
         if (empty($submitty_json['vcs_url'])) {
             $this->vcs_url = $this->base_url . '{$vcs_type}/';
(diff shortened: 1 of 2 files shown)


site/app/models/Config.php
```diff
@@ -256,9 +256,14 @@ public function loadMasterConfigs($config_path) {
         }
 
         $this->timezone = new \DateTimeZone($this->timezone);
-
         $this->base_url = rtrim($this->base_url, "/")."/";
-        $this->cgi_url = $this->base_url."cgi-bin/";
+
+        if (isset($submitty_json['cgi_url'])){
```
MasterOdin: Let's use `!empty(...)` for this check as by default, I think it might be a blank string.

Sahil333: Yes, right.

MasterOdin:
@bmcutler this is ready to merge, but when you go to install it on our production machine, you'll want to make sure that `submitty.json` doesn't have a value for `cgi_url`.

PR #3403, opened 2019-03-16
[Feature]Enhancing chat with bold,strike,italics and highlight

drumilpatel2000:
## Feature
Bold text by `*bold*`
Strikethrough by `~Strike Through~`
Italics by `_Italics_`
Highlight text by using tilde between text
Screenshot from 2019-03-16 16-04-33
Screenshot from 2019-03-16 16-04-44

### DIFF
--- a/site/app/templates/forum/ThreadPostForm.twig
+++ b/site/app/templates/forum/ThreadPostForm.twig
@@ -29,7 +29,9 @@
         {% set min_height = (post_textarea_large?"40vmin":"100px") %} 
         <textarea name="thread_post_content" id="thread_post_content" class="fill-available post_content_reply" style="resize:none;min-height:{{ min_height }};overflow:hidden;" rows="10" cols="30" placeholder="{{ post_content_placeholder }}" required></textarea>
     </div>
-<br/>
+    <div class="row" style="color: #666666;font-size: .75em ">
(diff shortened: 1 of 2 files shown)


drumilpatel2000:
@MasterOdin  @pbailie @shailpatels Please review this PR and please suggest changes

MasterOdin:
As much as I like markdown, given that we're using [bbcode](https://en.wikipedia.org/wiki/BBCode) for the existing functionality, we should probably stick with that (unless you were to successfully convince @bmcutler and @andrewaikens87 to switch to it). Additionally, we'll need to have established buttons for utilizing these tags in the editor so that users are aware of them (or have some sort of cheatsheet available).

Beyond that, as @zjxiaohan mentioned, we have to be careful that additional BBCodes do not cause any breaks/disruptions within the code blocks. I think it would be more beneficial going forward however to look into using a library to handle BBCode parsing instead of trying to re-invent the wheel on this (using something like http://php.net/manual/en/book.bbcode.php or https://github.com/thunderer/Shortcode. I suspect that before going this route, we'll need to discuss the design a bit more internally first. 

Given this, while we thank you for your contribution to Submitty here, due to the reasons listed above, this is not the proper way to go about this, as well as having some issues with how it's being parsed. Closing this for now as and opening issue #3439.

PR #3404, opened 2019-03-16
Added loading message to testcases

KevinMackenzie:
Added loading message to testcase output since it loads the entire image at once.  This makes the design more responsive, especially for large output files.  Loading messages are similar to the TA grading loading messages.  I just made the text the same style as the old `Details` text, but it would be easy enough to make them look like buttons if we want.

image
image
image

### DIFF
--- a/site/public/js/server.js
+++ b/site/public/js/server.js
@@ -107,19 +107,25 @@ function changeDiffView(div_name, gradeable_id, who_id, version, index, autochec
 }
 
 function loadTestcaseOutput(div_name, gradeable_id, who_id, index, version = ''){
-    orig_div_name = div_name
+    let orig_div_name = div_name;
     div_name = "#" + div_name;
-    var isVisible = $( div_name ).is( " :visible" );
 
-    if(isVisible){
-        toggleDiv(orig_div_name);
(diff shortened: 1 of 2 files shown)


site/public/js/server.js
```diff
@@ -107,19 +107,25 @@ function changeDiffView(div_name, gradeable_id, who_id, version, index, autochec
 }
 
 function loadTestcaseOutput(div_name, gradeable_id, who_id, index, version = ''){
-    orig_div_name = div_name
+    let orig_div_name = div_name;
```
zjxiaohan: It may be irrelevant to this PR, but I wonder if we are supposed to use ES6 keywords.

In `server.js`, I only found 1 occurrence of `const` and 4 occurrences of `let`. Does that mean it is better to avoid using ES6 because of [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#Browser_compatibility), or it is completely fine to use them in this repo (considering that the scope of `let` is often clearer than `var`)?

zjxiaohan: I see. Thank you for your response!

Issue #3414, opened 2019-03-17
Sort People by Lastname instead of User ID


zjxiaohan:
Currently, in the grading page, students are sorted by their user ID.

As user ID can be irregular, and students can have a preferred last name, this may lead to confusion.

For example, in the sample data, there is someone named Alyssa P Hacker, and different from other students, her user ID starts with her first name, making her the first in the list. When a professor grades, she or he may instinctively go to the bottom of the page to find someone whose last name starts with "H", who is actually on the top of the list.

![p1](https://user-images.githubusercontent.com/20266703/54497811-65c56e80-48bc-11e9-8e08-56968f67fe9f.png)

Another case, if someone changes the preferred last name (like Tim Cook changes his name to Tim Apple), he or she does not get relocated. This may be confusing for professors who are used to finding people by last name, too.

![p2](https://user-images.githubusercontent.com/20266703/54497812-6827c880-48bc-11e9-97d8-9c897c476e87.png)

bmcutler:
This is likely a low priority enhancement.  We already have sort by preferred name available on the checkpoint grading which is used in lab where alpha by name are important.  Manual grading of uploaded assignments primarily needs consistent grading.  Or sorting by grading status:  e.g. feature request #3236.

The page for grading of uploaded assignments is getting an overhaul for issue #1314  and PR #3307.  Addition of new sorting should be held until this overhaul is complete.

Kevin-MM:
I believe #4171 may close this issue?

zjxiaohan:
Yeah I believe so I was thinking of the same thing when I saw your PR.

PR #3417, opened 2019-03-18
Fix row sorting in forum statistics

zjxiaohan:
There are two changes.

1. "User  ↓" is changed to "User". Because by default, users are not ordered by their last name (#3414), so it is confusing to have the down arrow here before any sorting is done.

Original:
initial

2. Allow sorting in reverse order for all columns.

After (last name in decreasing order):
after


zjxiaohan:
@shailpatels Thank you very much for your review. I agree that there should be a way to suggest the buttons are clickable. For now, I would highlight the column names in blue to be consistent with the lab grading page.

Screenshot from 2019-03-24 12-35-47

In the long run, probably we need to keep the two places entirely consistent. I would suggest replacing existing icon or arrows with `fa-sort-asc` and `fa-sort-desc`. Also, the grading page needs to be modified so that it does not refresh when it is clicked.

zjxiaohan:
Hi @shailpatels , I wrapped the text with `<a>` tag according to [this StackOverflow answer](https://stackoverflow.com/a/34740294) so that the buttons look consistent with the column names in the grading page.

It is even better because it is accessible by Tab now :)

shailpatels:
Looks good!
@bmcutler works as described, I think this is ready to merge.

PR #3420, opened 2019-03-18
Exclude scanned exams from extension menu

### LINKED ISSUE #3418: Drop down menu for Extensions bug
The drop down menu for excused absence extensions should not include assignments that will be bulk uploaded by ta/instructor.  

Make sure to test this for both individual & team assignments

zjxiaohan:
Closes #3418 

These changes are made so that only electronic gradeables with `eg_scanned_exam=FALSE` will be selected. 

In the new sql command, `g_gradeable_type=0` is removed from `where` because only electronic gradeables remain after `JOIN` operation.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -1881,8 +1881,8 @@ WHERE gcm_id=?", $params);
         $this->course_db->query(
           "SELECT g_id, g_title " .
           "FROM gradeable INNER JOIN electronic_gradeable USING (g_id) " .
-          "WHERE eg_scanned_exam=FALSE " .
-          "ORDER BY g_grade_released_date DESC"
+          "WHERE eg_scanned_exam=FALSE and eg_has_due_date=TRUE " .
+          "ORDER BY eg_submission_due_date ASC"
         );
         return $this->course_db->rows();
     }
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -1875,7 +1875,12 @@ WHERE gcm_id=?", $params);
      * gets ids of all electronic gradeables
      */
     public function getAllElectronicGradeablesIds() {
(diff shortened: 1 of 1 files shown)


bmcutler:
Hi thanks!

I also noticed the order is unclear.  Can you please make sure they're ordered by due date too?

zjxiaohan:
@bmcutler 

Current ordering is by grade released date.
There are grade due date and submission due date, which due date will be more convenient for instructors? (I guess you mean submission due date?)

zjxiaohan:
I changed the code according to what I said above.

Now, only assignments with submission due will appear in the menu, and they are ordered by their due date (ascending) so that the assignment with the earliest submission due date will be on the top.

Thank you very much for the feedback.

bmcutler:
Yup, submission due date is what I'm looking for :)

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1872,10 +1872,18 @@ public function getAllGradeablesIds() {
...
+     * @return array
      */
     public function getAllElectronicGradeablesIds() {
-        $this->course_db->query("SELECT g_id, g_title FROM gradeable WHERE g_gradeable_type=0 ORDER BY g_grade_released_date DESC");
+        $this->course_db->query(
+          "SELECT g_id, g_title " .
+          "FROM gradeable INNER JOIN electronic_gradeable USING (g_id) " .
+          "WHERE eg_scanned_exam=FALSE and eg_has_due_date=TRUE " .
+          "ORDER BY eg_submission_due_date ASC"
+        );
```
MasterOdin: You can just open the `"` and leave it open to have a multiline string instead of doing it a string per line. We don't really care about the added whitespace in the query being sent to the DB, and it makes going back and editing the query a touch easier.

zjxiaohan: Thank you for your feedback! Yes, I agree that only having a multiline string is easier for future modification. I fixed it in the most recent commits.

PR #3421, opened 2019-03-18
[Bugfix] Changed fa-times-circle to a fa-check-circle in success messages

xypnox:
There are several places where the `fa-times-circle` is used incorrectly for success messages.
This PR replaces the incorrectly used `fa-times-circle` to `fa-check-circle`

For example:
The following is the incorrect version of the message:
Screenshot_2019-03-18 DEVELOPMENT

Which after correction results in:
Screenshot_2019-03-18 DEVELOPMENT(1)

Currently due to hard coded `fa-times-circle` in `GlobalHeader.twig` the messages for success do not show `fa-check-circle` for example: successful login.
I am working on adding an if condition based on `message.type` for the same


xypnox:
The hardcoded version for `fa-times-circle` has also been addressed in the latest commit.

Issue #3293, opened 2019-02-13
Improving form validation in `New Gradeable`


drumilpatel2000:
Form validation are only applied when form is submitted. This creates a `bad experience` for the user as they are not able to rectify their mistakes on time.

Fix :- Form validation should be applicable `onchange` not `onsubmit` javascript event. This will make the user to check their mistakes at time of filling the form which creates a `good experience` for the user.

sarthak77:
Can i take up this issue?

MasterOdin:
I would advocate for an approach such that if there is mistakes, the instructor is told the error and also has the form filled in for them with what they put in. That way we're only doing validation in one place (which is hard enough) and not over two.

bmcutler:
current behavior is ok.  would require doing the same check in two places and be harder to maintain.  

Closing issue (please re-open if there are more specific problems)

PR #3422, opened 2019-03-18
Update the new gradeable header as the ID changes

zjxiaohan:
In "New Gradeable" page, if I enter the gradeable title first, and then enter the gradeable id, the title that appears on the top of the page does not change.

This is fixed by binding the function `onTitleChange` to the change event of `g_id` too so that the title on the top changes when `g_id` changes.

g_id

### DIFF
--- a/site/app/templates/admin/admin_gradeable/AdminGradeableBase.twig
+++ b/site/app/templates/admin/admin_gradeable/AdminGradeableBase.twig
@@ -143,6 +143,7 @@
         // update the title bar when the gradeable name changes
         onTitleChange();
         $('#g_title').change(onTitleChange);
+        $('#g_id').change(onTitleChange);
 
 
         // TODO: below is selectively copied from previous ui.  It updates various


zjxiaohan:
> I think you should apply for some type of check before applying changes.
> It will display id even if It is not correct .

Hi Drumil. Thank you for your review!

There are several possible ways that I can think of to implement validation.

1. As soon as the id changes to an existing id, alert as if I clicked "submit".
-> It will make it very user unfriendly to use templates to create new gradeables.

2. If id is duplicated, then do not show things inside the brackets in the title on the top.
-> It may confuse users because there are no text explanations why `[]` shows in the title with nothing in it.

3. Show some friendly texts next to the input area if validation fails.
-> This option looks good to me, but as it is needed to keep the style consistent, much of the code needs to be changed and it may be complicated. I think that discussing this in #3293 is more proper.

This PR only aims at fixing a currently existing problem in the code, which is, while the title depends on two variables, it will only change if one certain variable changes. I would suggest that a redesign of the validation logic may better be discussed separately in #3293 .

MasterOdin:
I agree that we should leave validation changes to a new PR (that addresses it as a whole) and that we can just keep this as a simple bugfix PR.

Thanks for the contribution.

PR #3424, opened 2019-03-19
Add tests for gradeables without due dates

zjxiaohan:
This PR:
- Adds tests for gradeables without submission due dates.
- Refactors the code a little bit by following TODO
- Increases code coverage

### DIFF
--- a/site/tests/app/models/gradeable/GradeableListTester.php
+++ b/site/tests/app/models/gradeable/GradeableListTester.php
@@ -112,7 +112,7 @@ class GradeableListTester extends BaseUnitTest {
         }
     }
 
-    public function testSubmittableAdmin() {
+    public function testSubmittableHasDueAdmin() {
         $gradeables = array();
         $gradeables['01_future_homework_no_ta'] = $this->mockGradeable("01_future_homework_no_ta",
             GradeableType::ELECTRONIC_FILE, '9995-01-01', '9996-01-01', '9997-01-01', '9998-01-01',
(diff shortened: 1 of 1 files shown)


site/tests/app/models/gradeable/GradeableListTester.php
```diff
@@ -199,6 +195,131 @@ public function testSubmittableStudent() {
...
+        $this->assertEquals($expected, array_keys($actual));
+        foreach ($expected as $key) {
+            $this->assertEquals($gradeables[$key], $actual[$key]);
+        }
+
+        $actual = $list->getGradedGradeables();
+        $this->assertCount(0, $actual);
+    }
+
+    public function testSubmittableNoDueGrader() {
```
MasterOdin: It looks like this is a duplicate of the above testcase and doesn't really test anything differently?

zjxiaohan: @MasterOdin No it doesn't test anything different except that one function's core is grader and one is admin.

I may have some misunderstanding here. I was thinking if there were more modifications in the future so that admin's gradeable list is changed, it could be good if the test can tell that happened.

Now I think probably I don't need to consider that. Will remove later.

MasterOdin: Yeah, let's remove it for now and if the code changes in some way for instructors vs graders, they can write up a relevant test case then.

zjxiaohan: @MasterOdin `testSubmittableNoDueAdmin` is now removed because there is no code concerning `access_admin`. So it is find to just test it on graders and students.

Thank you very much for your feedback!

PR #3440, opened 2019-03-21
[Travis] Improve the Travis-CI WIP check for PRs

MasterOdin:
The current implementation will occasionally fail to get the title of a PR which causes the check to not work properly and the PR will attempt to build as normal. This makes it so it's easier to check going forward, as well as adding in an ability to have it try 5 times (over 5 seconds) to get the PR title, which if it fails, will still cause the build to fail. As this gets added to our build, I can work to tweak how many attempts we should make/time waiting between attempts before we just fail the build.


zjxiaohan:
@MasterOdin Instead of checking WIPs, how about encouraging submitting a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) if work is in process.

MasterOdin:
That unfortunately doesn't run any of the Travis test suite. This particular tag was setup internally (back before Draft PRs were a thing) so that we could run the smaller/quicker branch builds for Travis while not having the longer PR build go, principally as during the summer when we've got several students working on it full time, it could cause Travis to take all day to get through all pushes/PR builds.

zjxiaohan:
Oh thank you so much for your reply. I did not consider that point :sweat_smile: 
It looks that [Travis is having an internal discussion on this topic](https://travis-ci.community/t/draft-pull-requests-not-being-built/2434/5). Hopefully it will be easier soon :)

PR #3447, opened 2019-03-22
Forum posts in grading portal

andrewaikens87:
I would like to finish going through the code and making tweaks but thought I would put it up now as the hackathon is tomorrow. Creates a foundation for: "Integration of grading tasks with forum posts and participation" which is a GSOC discussion forum sub-project. 

gradingPortalForum
gradeable


zjxiaohan:
Hello. As you mentioned, this PR "creates a foundation for: 'Integration of grading tasks with forum posts and participation'". Could you please suggest what work is remained to be done?

I could not find an issue regarding the grading integration part, and as far as I understand, this PR has addressed most of the tasks that I could think of, which is, when grading, show certain users' posts in some set threads.

As I am drafting the proposal, I would appreciate it if you could kindly advise some possible further enhancements based on the foundation laid by this PR.

PR #3454, opened 2019-03-23
Enhancing new gradeable

drumilpatel2000:
Enhancing #3422 
With reference to issue #3293
ezgif com-video-to-gif
Automatically corrects `g_id` and uses ajax to check number of g_id with same names and concatenate it with the number

### DIFF
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -57,6 +57,9 @@ class AdminGradeableController extends AbstractController {
             case 'import_components':
                 $this->importComponents();
                 break;
+            case 'number_g_id':
+                $this->numberWithGId();
+                break;
             default:
                 $this->newPage();
                 break;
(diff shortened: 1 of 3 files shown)


drumilpatel2000:
@MasterOdin please review and suggest some changes

drumilpatel2000:
@MasterOdin Any suggestion.

holzbh:
Auto-appending to the end of a g_id is undesirable behavior in my opinion, it's unexpected and almost decidedly not what I intend. I'm fine with being alerted that it's an invalid g_id due to collision. Asked @bmcutler and she agreed that we should not be making auto changes to the gradeable_id.

(Incidentally it looks like the gradeble_id at the top would be inconsistent and use the id that was entered prior to appending, e.g. making a new `c_failure_messages` with title `Test` in the Development course, the bold title will be `New Gradeable: [c_failure_messages1] Test`)

(Also if the gradeable_id already ends in a number this just added 1, for example `c_failure_messages1` became `c_failure_messages11` which is not particularly intuitive behavior. But I do not think the right answer is to try and correct this auto-incrementing.)

PR #3464, opened 2019-03-24
Forensics page

### LINKED ISSUE #3182: bulk upload data history forensics page
Make a new page, visible only by instructors and full access TAs, that we access by clicking on a link button from the entry page of ta grading statistics. The new page should be scraped from these bulk upload data files and have 4 columns:

user name, preferred first & last names
upload timestamp
submit timestamp
filepath --- EDITED TO REVISE: Let's not display or sort by full filepath. Just the filename. The filepath includes the upload time. I just want to sort by "w01_72.pdf" just the filename.
And the user can separately sort these 4 columns, kinda like we can sort the columns of the
forum statistics.

trane2:
Closes #3182


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -210,6 +210,111 @@ public function statusPage(
...
+				        <td style = "cursor:pointer;" width="25%" id="filepath_down">Filepath</td>
+					</tr>
+HTML;
+
+		foreach($users as $user => $details){
+			$first_name = $details["first_name"];
+			$last_name = $details["last_name"];
+            $upload_timestamp = $details["upload_time"];
+            $submit_timestamp = $details["submit_time"];
+            $filepath = $details["file"];
```
andrewaikens87: For the filepath, first_name, and last_name lets escape these values.

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -210,6 +210,111 @@ public function statusPage(
...
     }
 
+    public function statPage($users) {
+
+		$return = <<<HTML
+        
+		<div class="content_upload_content">
+
+HTML;
+        $this->core->getOutput()->addBreadcrumb("Bulk Upload Forensics", $this->core->buildUrl(array('component' => 'submission', 'page' => 'statPage')));
```
andrewaikens87: When clicked it brings the user to a page that says gradeable item doesn't exist (seems like the action just needs to be changed).

andrewaikens87:
Next step would be to convert to twig. This could share the code from the stats page on the forum and the stats page here.

PR #3471, opened 2019-03-24
[URGENT] remove unmerged code

zjxiaohan:
This part of `ConfigTester` is blocking tests.


bmcutler:
This is the wrong edit.  We are using config_json now not config_ini

PR #3478, opened 2019-03-24
Apply trim to frontend and backend in Authentication

### LINKED ISSUE #3477: Authentication failing due to unwanted whitespaces
## Issue
Unwanted spaces are interfering authentication 
ezgif com-video-to-gif(3)
For eg:- when username is " instructor " and password is "instructor" authentication fails but it does not fails when username is "instructor"

drumilpatel2000:
Fixes #3477 
Applying backend and frontend trimming of unnecessary spaces in username


site/app/controllers/AuthenticationController.php
```diff
@@ -118,7 +118,7 @@ public function checkLogin() {
             }
             return false;
         }
-        $this->core->getAuthentication()->setUserId($_POST['user_id']);
+        $this->core->getAuthentication()->setUserId(trim($_POST['user_id'],' '));
```
MasterOdin: Let's make this change in https://github.com/Submitty/Submitty/blob/master/site/app/authentication/AbstractAuthentication.php#L32. You can also just do `trim()` with just one argument and leave the second to the optional default. (so just `trim(strtolower($user_id))`

drumilpatel2000: Thanks @MasterOdin , For reviewing my PR

site/app/templates/Authentication.twig
```diff
@@ -11,7 +11,7 @@
             {% for key, value in old %}
                 <input type='hidden' name='old_{{ key }}' value='{{ value }}' />
             {% endfor %}
-            <input type="text" name="user_id" placeholder="User ID" required autofocus/><br />
+            <input type="text" name="user_id" onchange="$(this).val($(this).val().trim())" placeholder="User ID" required autofocus/><br />
```
xprilion: @bmcutler @drumilpatel2000 This line checks for spaces only on the left and right of the user id input. One would expect instead to disallow typing in spaces at all. Changing the `onchange` to `onkeyup` would effectively implement the prohibition of typing in whitespace in the userid textbox.

PR #3489, opened 2019-03-25
Fix pinning as favorite feature on discussion forum

andrewaikens87:
This resolves the issue that prevented a user from unpinning a thread. ~Haven't tested when viewing in grading portal. Will test that after dinner and update this thread.~


andrewaikens87:
Tested on grading portal and seems to be fine as well.

PR #3491, opened 2019-03-25
Enhancing discussion gradeable portal 

drumilpatel2000:
## Before
Screenshot from 2019-03-26 04-19-52
## After
Initial post of discussion forum is visible for reference of TA and Instructor.So they don't have to go to respective thread  
Screenshot from 2019-03-26 04-17-07


andrewaikens87:
Thank you for the PR, but instructors have not requested this feature.

PR #3493, opened 2019-03-26
Add shortcut(KeyD) for discussion panel in grading portal

drumilpatel2000:
Screenshot from 2019-03-26 10-35-21


drumilpatel2000:
@andrewaikens87 Is this okay?

PR #3502, opened 2019-03-27
Fixed overflow of Overall Comment

### LINKED ISSUE #3501: Extremely long comments overflow in Grading
This is very similar to #335 but apparently, only the special characters part of it was solved when it was closed. 

The Overall comment when very long still overflows after save (during editing it has a vertical scroll bar inside a textarea). 

Screenshot from 2019-03-27 15-26-11

xprilion:
Closes #3501  

Very simple fix : set a maximum height to the comment span, break the long text on overflow and add a scrollbar on overflow to y axis.

https://www.useloom.com/share/bb46f4df0aac4c31a7c866bccd9d0b9a

### DIFF
--- a/site/public/css/ta-grading.css
+++ b/site/public/css/ta-grading.css
@@ -305,6 +305,11 @@ progress::-webkit-progress-value {
 
 #overall-comment-collapsed {
     padding: 0 0 0 20px;
+    max-height: 100px;
+    overflow-x: hidden;
+    overlfow-y: scroll;
+    width: 100%;
+    overflow-wrap: break-word;
 }
 
 .ta-rubric-table {


site/public/css/ta-grading.css
```diff
@@ -305,6 +305,11 @@ progress::-webkit-progress-value {
 
 #overall-comment-collapsed {
     padding: 0 0 0 20px;
+    max-height: 100px;
```
drumilpatel2000: 1. Probably max-height is not working properly for textarea 
   ezgif com-video-to-gif (1)
2. Should use javascript to break a word such that half-word-  and -last-half-word 
ezgif com-video-to-gif (2)

xprilion: Hey @drumilpatel2000 the max-height is working as intended - it keeps expanding till whatever size the edit mode requires. This is good for working on while editing the comment (observe how Facebook handles this). After saving the max height is 100px.

drumilpatel2000: Hey @xprilion Facebook increase height up to some limit and then scrollbar comes. This can be easily achieved by setting max-height.
Considering UI I does not consider this as better implementation as **Manual grading Total** goes out of the frame for long comments

mattraneri:
Max height shouldn't be set- 
Reason being scrollbars inside scrollbars
I am filing a new issue for the odd behavior in the onkeyup function.

PR #3505, opened 2019-03-27
Show due days on excused extension

### LINKED ISSUE #1576: Late Days Calendar Picker for Excused Absense Extensions
When entering the number of late days under the "Excused Absense Extensions" tab, it is not always clear how many late days are needed for the assignment. You must know when the assignment was first due, how many days to now, and how many does the student need. It would be nice to have a calendar to show the original due date and allow you to pick the new submission date. The system could then calculate the number of late days. (Alternatively, even a simple comment about how many late days to the current date would help.)

While the number of late days is a nice concept, we have not been using the feature in the same manner as I expect it was intended. We allow a student to resubmit any one assignment once a week, independent of when the assignment was originally due. Thus an excused absense could be for several weeks.

VictorLi718:
closes #1576. Add a Due Date section showing the due date for each gradeable so that instructors can assign the right number of extension conveniently. 
Screen Shot 2019-03-27 at 3 10 39 PM


VictorLi718:
Fix the old bug and add the date picker option. It will automatically calculate the number of days of extensions needed.
Screen Shot 2019-04-08 at 9 11 50 PM

site/app/templates/admin/Extensions.twig
```diff
@@ -7,18 +7,27 @@
         <div class="option">
             <p> Use this form to grant an extension (e.g., for an excused absence) to a user on a specific assignment.<br><br><br></p>
             <div class="option"><label for="g_id">Select Gradeable:</label><br>
-                <select name="g_id" id="g_id" onchange="loadHomeworkExtensions($(this).val());" style="margin-top: 10px; width: 50%">
+                <select name="g_id" id="g_id" onchange="loadHomeworkExtensions($(this).val(),$(this).find('option:selected').attr('dueDate'));" style="margin-top: 10px; width: 50%">
                     <option disabled selected value> -- select an option -- </option>
-                    {% for index, value in gradeable_ids %}
-                        <option value="{{ value['g_id'] }}">{{ value['g_title'] }}</option>
+                    {% for index, value in gradeables %}
+                        <option value="{{value['g_id']}}" dueDate="{{value['eg_submission_due_date']}}">{{ value['g_title'] }}</option>
```
MasterOdin: Use a [data attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) to contain the due date, not add a custom attribute.

site/public/js/server.js
```diff
@@ -2412,6 +2412,8 @@ function loadHomeworkExtensions(g_id) {
             $('#my_table tr:gt(0)').remove();
             var title = '<div class="option-title" id="title">Current Extensions for ' + json['gradeable_id'] + '</div>';
             $('#title').replaceWith(title);
+            var print_due_date = '<p id ="due_date">' + due_date + '</p>';
+            $('#due_date').replaceWith(print_due_date);
```
MasterOdin: `$('#due_date').text(due_date);`

see: http://api.jquery.com/text/#text2

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1636,12 +1636,13 @@ public function getAllGradeablesIds() {
...
         $this->course_db->query("
-          SELECT g_id, g_title
-          FROM gradeable INNER JOIN electronic_gradeable USING (g_id)
-          WHERE eg_scanned_exam=FALSE and eg_has_due_date=TRUE
-          ORDER BY eg_submission_due_date ASC
+          late_days_for_excused
+            SELECT gradeable.g_id, g_title, eg_submission_due_date
+            FROM gradeable INNER JOIN electronic_gradeable 
+                ON gradeable.g_id = electronic_gradeable.g_id
+            WHERE g_gradeable_type=0 ORDER BY g_grade_released_date DESC
```
MasterOdin: The first line of the query probably shouldn't be there? (`late_days_for_excused`)

It also looks like this will change the result set here to now also include rows where `eg_scanned_exam=TRUE` and `eg_has_due_date=FALSE`?

VictorLi718: I have fixed them now. I think it is good to go now.

PR #3508, opened 2019-03-29
Show current system time on submission page

### LINKED ISSUE #1865: Add "current time on server" dynamic clock to submission page

xprilion:
Partially closes #1865 - it is in discussion over some finer points now. 

Sent a quick PR since it was a long pending issue. 

Screenshot from 2019-03-29 16-30-28


MasterOdin:
@bmcutler, @inlinefan, how often have we gotten drift on the main machine in past year or so? I know we had issues a couple years ago, but it seems like ntp has been stable and preventing any real drift.

xprilion:
Hey @shailpatels I've implemented the following - 

When the page loads, it could be at any second (seconds are universally the same) so a Time fetch from server occurs after the remaining seconds from the next complete minute. 

Henceforth, at every minute, a time fetch from server occurs, effectively eliminating the chance of mismatched time on server and in the client UI, and the overhead of constant polling. Also, the format of time and date display remains consistent. Since the update interval is set irrespective of the response time of the fetch time call, the interval is maintained at every 60 seconds, strictly. 

The time updates only when the call to server is successful, so the UI doesn't flicker nor does the time get cleared from the browser in case of a failure.

inlinefan:
I don't think we have had any more problems since the one time.  I think
that disabling ntp and using polling is a bad idea.

xprilion:
@inlinefan would you suggest simply using JavaScript based time update instead?

MasterOdin:
I would make it so that you fetch time from the server on page load, and then use JS to increment it upwards.

bmcutler:
We had a draft implementation of this issue months ago, but there were bugs in the time/timezone (production installations do not match vagrant in this area) and concerns about the expense of polling and concerns about the visuals being unnecessarily distracting.

This needs lots of testing, and a mini-survey of user input on the final chosen design.

bmcutler:
@xprilion, what is your current status on this PR?

xprilion:
@bmcutler I'm awaiting some conversation over this issue for the design finalization. Shall I draw up some mockups and maybe some of us can vote over them? Or is there any other existing method of doing the design surveys at Submitty, we could definitely use that. Please let me know.

bmcutler:
So this feature is an optional feature (hopefully helpful to students, but not strictly necessary) -- and as such we want to make sure it doesn't cause confusion or mis-information or stress or cause load on the server/network/student machine.

So we want to make sure the display is subtle -- it doesn't update every second, it's not visually distracting.  Maybe it's not even shown if the deadline is more than 24 hours away.  Maybe when the deadline is less than 1 hour away the color changes to red (or some other visual change)?  But no blinking text or anything stressful.  Yes, maybe you can post several options for the visual we can discuss/vote on.  

It should work for all combinations of timezones -- if the student & the server are the same timezone, or if they are different timezones.  We want to ALWAYS display the server timezone (a student doesn't get an extension if they set their local machine to a different timezone!).  This will require lots of testing.  If/when you've done this testing, please explain :)

If the student leaves the page open for a long time without refreshing, or if they put their machine to sleep and wake it back up -- we want to make sure the display doesn't get significantly out of sync (more than a minute or so).   This will require lots of testing.  If/when you've done this testing, please explain.  I assume we need to test on the different OS & browsers as well.  Please ask for help for whatever you don't have access to.

We don't want to poll the server very often, most of this time updating can be done on the client machine exclusively.

bmcutler:
@xprilion  What is the status of your work on this PR?  Are you looking for any guidance on implementation?

xprilion:
@bmcutler the task was a bit complex on the head and after implementing it once, I realized there was a simpler implementation, thus the time taken. But finally, it is done in the following manner - 

[1] At page load, a time out for fetching the server time when the number of seconds on the clock is 0 is set. 
[2] the server time is fetched at clock 0 time, but the latency might cause a few seconds to shift. This is adjusted into the server time. 
[3] a date variable with the server's time offset is set into the browser. This keeps updating on the page every 1 minute. 
[4] After 10 minutes, the server's time is fetched again, synchronizing any time differences that might have crept up.

Worst time difference case: 
[1] Right after time update, the person puts the machine to sleep. The next update is 59 seconds away. 
[2] Person opens up the laptop exactly x minutes later. 
[3] The page will display the wrong time for a maximum of 5 minutes from the time of resuming laptop operation. 

This can be bettered. I shall think of another way, or maybe we can discuss this out! :)

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -418,6 +420,69 @@
...
+            if(nextTimeFetch == 0) {
+                // console.log("Fetch set time remaining: "+(60 - localDate.getSeconds()));
+                nextTimeFetch = 300;
+                setTimeout(function(){
+                    setInterval(fetchServerTime, nextTimeFetch * 1000);
+                    // console.log("Next fetch set");
+                }, (60 - localDate.getSeconds())*1000);
+            }
+
+            $.ajax({
```
shailpatels: Could you remove the console.logs and also the commented out ones and replace them with comments about the logic of the function, I think that would be more helpful

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -418,6 +420,69 @@
...
+                dataType: "json",
+                success: function(result){
+                    localDate = new Date();
+                    serverDate = new Date((result.data.time + (localDate.getSeconds())) * 1000);
+                    console.log("Updated servertime");
+                    if(setTimerOn == 0){
+                        // console.log("Server time remaining: "+(60 - serverDate.getSeconds()));
+                        setTimeout(function(){
+                            // console.log("Setting interval");
+                            serverDate = new Date((result.data.time + (60 - serverDate.getSeconds()))* 1000);
```
shailpatels: Why is there another nest setTimeout function when theres already one above the ajax call. If the ajax response is delayed for whatever reason I don't think this will correct time skew, it might make it worse if the ajax response is really out of sync and just overwrites the setTime function with an outdated time.

xprilion:
@shailpatels thanks for the review. Closing this in favor of a new PR, it has been around quite some time!

Will keep the display as is currently, will change the method of time fetching! :)

PR #3514, opened 2019-03-31
 Fix #3513 Unrendered &quot;

### LINKED ISSUE #3513: Unrendered &quot; in gradable submission page
There's a tiny error here which prevents` &quot;` to render as " as shown below in the highlighted text

Screenshot from 2019-03-31 21-03-42

jhcccc:
Fix #3513 Unrendered &quot; by removing quotation mark.
Technically a way to go, I guess?

### DIFF
--- a/site/config/course_template.json
+++ b/site/config/course_template.json
@@ -8,7 +8,7 @@
         "default_hw_late_days": 0,
         "default_student_late_days": 0,
         "zero_rubric_grades": false,
-        "upload_message": "By clicking \"Submit\" you are confirming that you have read, understand, and agree to follow the Academic Integrity Policy.",
+        "upload_message": "By clicking Submit you are confirming that you have read, understand, and agree to follow the Academic Integrity Policy.",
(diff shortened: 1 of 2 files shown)


xprilion:
Hey @jhcccc I wonder if this 'solves' the issue instead of removing the steps which produced it. If any instructor puts any double quotes in the message for there it will still produce the &quote; thing

Only my thought, this might be a trivial issue so won't really get adamant on this :slightly_smiling_face:

jhcccc:
> Hey @jhcccc I wonder if this 'solves' the issue instead of removing the steps which produced it. If any instructor puts any double quotes in the message for there it will still produce the &quote; thing
> 
> Only my thought, this might be a trivial issue so won't really get u on this 🙂

Thank you for pointing out! I was thinking that instructors won’t bother to customize this message. But you are totally right, instructors may put anything in this message and the problem could reproduce. I’ll try to actually fix quote problem.

PR #3518, opened 2019-04-02
DB schema fix for possible preferred name data corruption

### LINKED ISSUE #3511: users.user_updated System Database Field Can Be Erroneously Overwritten To false.
The system DB schema has a `user_updated` field in the `users` table.  It is supposed to be set `true` when a user updates their own preferred names.  This serves as a "permissions lock" that is to be checked before anyone else (e.g. the instructor, grader, student auto feed, etc.) changes a user's preferred name.

This field is absent in the course DB schema.  Therefore, if a user's details is read from the course DB, `user_updated` is not read, and the Users model will assume the value of `false` by default.

When calling `$this->core->getQueries->getUserById()`, user information is retrieved from the course DB as this retrieves not only information about the user, but also grading registration sections and their forum usage.  As there is no `user_updated` course DB field to read, the User object's `user_updated` property will always be `false`, per default.  Should this same User object be used to update the database via `$this->core->getQueries->updateUser()`, a `true` value in `users.user_updated` (system DB) will be clobbered to `false`.

I can confirm that this is a problem with Student/Grader spreadsheet/csv file uploading, and may exist in other areas.  Therefore, my work on student/grader list file upload (PR #3270) will not permit an upload to change any user's preferred names until this issue is resolved.  (Note: the `users.user_updated` field clobbering can still occur, as this bug is not in the student/grader management page code)

This may be related to issue #3273 (Preferred name is forgotten/removed without any indication why).

pbailie:
Fix #3511 

`user_updated` and `instructor_updated` fields are added to course DB's user table and trigger functions updated accordingly.  Now `user_updated` and `instructor_updated` data can be read from course DB and written back to master DB without corrupting user data (provided that data was originally sync'd in the first place, which it should be).

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -459,6 +459,8 @@ CREATE TABLE users (
     user_group integer NOT NULL,
     registration_section character varying(255),
     rotating_section integer,
+    user_updated boolean NOT NULL DEFAULT false,
+    instructor_updated boolean NOT NULL DEFAULT false,
     manual_registration boolean DEFAULT false,
     last_updated timestamp(6) with time zone,
     CONSTRAINT users_user_group_check CHECK ((user_group >= 1) AND (user_group <= 4))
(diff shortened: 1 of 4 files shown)


bmcutler:
@pbailie please fix travis tests

MasterOdin:
Confirmed this maintains the fields values in the DB as desired. Before running the migrations, when editing a user from the course page as an instructor, it would toggle user_updated to false always and instructor_updated to true. After running the migration, user_updated remains whatever value it was before editing the user.

PR #3520, opened 2019-04-02
[Bugfix] Trim spaces in forum thread titles

### LINKED ISSUE #3519: No Validation for blank input
ezgif com-video-to-gif(9)

drumilpatel2000:
fixes #3519 
ezgif com-video-to-gif(10)


MasterOdin:
While I could agree with trimming the title, I'm not sure there's never a use case an instructor might not use spaces at the beginning of a post for formatting reasons.

drumilpatel2000:
@MasterOdin Please look at this one 
ezgif com-video-to-gif(11)

MasterOdin:
Discussing with @bmcutler and @andrewaikens87, it seems like instructors do rely on that they can just input a single space in a thread body and then use that purely as the body. Both instructors and students have utilized this feature for good so we'd prefer not to take this away, so let's make this feature only affect titles.

site/public/js/server.js
```diff
@@ -2618,6 +2618,11 @@ function toggleSidebar() {
 }
 
 $(document).ready(function() {
+    $("textarea, input[type='text']").change(function () {
+        if($.trim($(this).val()).length == 0){
+            $(this).val($.trim($(this).val()));
+        }
+    })
```
MasterOdin: add a colon here

drumilpatel2000:
@MasterOdin  Is this okay?

MasterOdin:
@drumilpatel2000 This is fine and will be merged asap.

site/app/templates/forum/ThreadPostForm.twig
```diff
@@ -127,4 +127,11 @@
...
 </div>
 <br>
-</div>
\ No newline at end of file
+</div>
+<script>
+    $("input[name='title']").change(function () {
+        if($.trim($(this).val()).length == 0){
+            $(this).val($.trim($(this).val()));
+        }
```
MasterOdin: Let's get rid of the `if` statement and just use the `$(this).val($.trim($(this).val()));` bit to always trim it. We don't ever want the spaces on the start/end of the string.

PR #3521, opened 2019-04-02
Improve testing of migrator tool

MasterOdin:
This finishes up writing a lot of testing of the migrator tool to show that it works. There was some internal functionality changes to make it easier to test, but the external interface and results remain the same. 

The test cases could definitely be improved here, a common base testcase involved, etc, but this does at least add testing for most of the migrator code.


MasterOdin:
@bmcutler: this should be ready to go.

The stuff that you might notice/be concerned about:
1. The deletion of migration/migrator.py is of an unused file that got re-added in a previous "bad" merge conflict where Evan fixed a bug in a PR of his that I put into the new migrator, but we still retained the old file.
2. You can now "fake" a rollback (same as migrate)

The rest of the PR is internal changes to make the tool easier to test (such as making the CLI interface accept a directory for where the config directory is instead of trying to auto-detect it).

PR #3537, opened 2019-04-04
(2) Autograding changes

LeonMontealegre:
Merged all the _doits from other files into a single "dispatch.h" + "dispatch.cpp".

Separated difference.h into "difference.h" and a "change.h"

This is the second PR branching off of DiffBranch1 which was already merged


MasterOdin:
I don't think dispatch is a good name for that file as the file itself isn't doing any dispatching (rather just being run by a function called `dispatch` in another file). validators or graders is probably more appropriate given that's what it contains?

PR #3541, opened 2019-04-05
Fix:  Right aligned text and re-formatted grading details table

xypnox:
Currently, the table for grading view center aligns the text columns such as username, First name and last name. Also, the heading text is not bold and there are borders which do not serve a purpose since rows have different background color to begin with.

Screenshot_2019-04-05 SAMPLE(3)
 

### What is the new behavior?
The new table view has the rows of text as left aligned. Also, the top borders have been removed, the heading is now bold text on white background.

Screenshot_2019-04-05 SAMPLE(2)
 

### Other information?
The implementation is done through a separate class named `table-left` this can then be applied to other tables in the future. This also doesn't affect any other tables.


site/public/css/server.css
```diff
@@ -275,6 +276,21 @@ table tr.table-header {
     vertical-align: middle;
 }
 
+.table-left {
```
drumilpatel2000: Name should be more clear. `table-left` does not signifies that all fonts are to left

drumilpatel2000: According to me, previous configuration was better as first name was easily differentiable between section and first name. Apart from this if we can also change header position still it will create confusion when zoomed in (or for small screen)

xypnox: I agree that the first name was easily differentiable between section and first name. But the increase in readability is more important as several people will glance at the names and find the one they are looking for quickly in a left aligned layout.

As for the name, I think it can be improved to `table-text-left` so It specifies exactly what it is intended to do. I will add this in my next commit.

site/public/css/server.css
```diff
@@ -251,7 +251,8 @@ table tr.table-header {
 }
 
 .table>thead>tr {
```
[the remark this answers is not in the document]
xypnox: Rather than changing `tr` to `th`, It should be applied to both. Since `tr` and `th` are two different elements. 
I am addressing this in my next commit.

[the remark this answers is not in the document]
drumilpatel2000: Sorry but @xypnox  I want to say change `td` to `tr` as it will directly convert normal  text into bold by default. So no need to add CSS for bolding respective text

Issue #3399, opened 2019-03-16
Sidebar does not Automatically Recover from Collapsing


zjxiaohan:
1. Open any page with the sidebar.
2. Open Chrome developer tools.
3. Refresh the page.

Now you will see a collapsed sidebar, which is expected.

![collapsed](https://user-images.githubusercontent.com/20266703/54470456-597acd80-4765-11e9-85fa-332c44ab4dd5.png)

But it will remain like this even after I closed the developer tools.

Another way to reproduce it is to have a small browser window, refresh so that the sidebar is collapsed, and maximize the window. The sidebar does not recover automatically.

I need to either refresh the page or click the "collapse sidebar" icon manually.

Not sure if this is a bug or not. It may be a quite marginal case, but it would be nice if the sidebar collapses when the window size is small and recovers automatically.

PR #3543, opened 2019-04-07
Display error on window resize

### LINKED ISSUE #2164: Display error on window resize
When resizing browser window, content can go outside the central box in an unsightly fashion, and when it does, the navigation bar and Submitty banner no longer fill the entire page.
Examples from the homepage and the grading page:

homepage_display_error

grading_display_error

This doesn't seem to be an issue on the forum...
forum_no_error
...until you refresh and increase the width
forum_error

xprilion:

Fixes #2164 - On resizing the window the Submitty topbar is left out of the screen. Also, the left navigation does not automatically collapse to adjust from larger to smaller screen size and vice versa. 

### What is the new behavior?
The page should display a warning and ask for reload. This would make the necessary changes to the page on fresh load. Preferably reload using cache.

https://www.useloom.com/share/331a6e2a6644492ca200dee4a7295327

### Other information?
Is this a breaking change? - No.
How did you test - Resizing checked on Linux (Elementary OS 5) using Chrome and Firefox.

### DIFF
--- a/site/app/templates/GlobalFooter.twig
+++ b/site/app/templates/GlobalFooter.twig
@@ -38,6 +38,20 @@
         Course Queries: <br /> {{ self.query_list(course_queries) }}
     </div>
 {% endif %}
+
+<script>
+    $(window).bind('resize', function(e)
+    {
+        if (window.RT) clearTimeout(window.RT);
+        window.RT = setTimeout(function()
+        {
+            if(confirm("Window resized. Reload page?")) {
+                this.location.reload(false);
+            }
+        }, 100);
+    });
+</script>
+
 </body>
 </html>
 


drumilpatel2000:
Please uncheck lower two boxes as :- 
1. No test were added 
2. No documentation was changed in this PR

xprilion:
@drumilpatel2000 both the checkboxes for mention "(if possible)" and "if relevant" and in the first case I was unable to find a way to test this while in the second one I was unable to find its relevancy anywhere in the documentation.

xprilion:
@drumilpatel2000 while both your suggestions are good, I did what was mentioned in the issue and solves the problem quickly. 

While your first suggestion is surely doable in a short period, the bar and the content would still take time to get them to adjust to very small screens. The other suggestion will take considerable time and testing, which might not be feasible for a low priority issue for now.

drumilpatel2000:
Sorry @xprilion but according to me this PR neither add some improvement but only increases complexity for user

xprilion:
@drumilpatel2000 the intent of this PR is to provide a quick fix to the issue while in the long run it can be solved by the suggestions you mentioned :) Kindly let me know if it is desirable to have some sort of solution in place for this at the earliest or to leave it pending.

drumilpatel2000:
I probably think this issue should remain pending till a reliable solution is not given by mentors. Merging this PR will cause unnecessary alerts and will harm user interface.
@bmcutler  Please help us with this issue. @ropern123 any design suggestion you came across to solve this issue

MasterOdin:
Agreed with @drumilpatel2000 that this really just a hack around other bugs in the system (which we should address), for example #3399. Realistically, I think we should probably close #2164 in favor of more specific issues with clear actionable items as the issue is somewhat vague on how to fix it (though @drumilpatel2000 is correct that properly handling it is under the purview of making Submitty more mobile friendly).

Sorry for the confusion here, and thanks for contributing to Submitty.

PR #3546, opened 2019-04-07
Show useremail and role on home page

Sahil333:

*  [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)

Screenshot from 2019-04-08 09-29-32
### What is the new behavior?
Screenshot from 2019-04-08 09-26-48
### Other information?

<!-- How did you test >


bmcutler:
Thanks for the PR!

A user can be different roles in different courses.  So please remove that information.  (Also, it's already displayed on the right).  But yes, keep the addition email, that's helpful :)

Sahil333:
@bmcutler Should we allow to change the user's email?

bmcutler:
At this time no, we do not want to allow users to change their email.  At RPI we are sending automatic announcement & notification emails to these addresses.  And if students were to change their address to a non-RPI email address this would impact the ability and/or quantity of emails we could send per minute.  

Eventually we might want to allow users to specify a second email address or to modify the primary address, but that's more complicated than necessary for now.

MasterOdin:
Thanks for the PR!

PR #3547, opened 2019-04-08
Add message for no rotating section

### LINKED ISSUE #2708: Show error when using rotating sections before they are set up
From #2671

> ALSO.... if rotating sections have NOT been configured yet (no sections, everyone in NULL) but the instructor attempts to select rotating in create/edit gradeable, have some scary red message appear saying "Go setup rotating sections first, then return to this page", perhaps with a link to the setup sections page.

VictorLi718:
Add a new message when there is no rotating section. And it will now show the confusing 1 and 0 again when there is no rotating section.
Before:
Screen Shot 2019-04-02 at 6 48 24 PM
Now:
Screen Shot 2019-04-07 at 11 43 25 PM
Closes #2708


bmcutler:
Please change the message to be more helpful:

'Click on "Manage Sections" from the left sidebar to configure Rotating Sections for your course.'

Thanks

VictorLi718:
Should be good now. Thank you for the suggestion.

PR #3557, opened 2019-04-10
VPAT — Improve header heirarchy

### LINKED ISSUE #2309: VPAT - Header 1 / Header Hierarchy
Web Accessibility Evaluation Tool ("WAVE") shows an **alert** that the header hierarchy is incorrect.

WAVE explains:
>Headings facilitate page navigation for users of many assistive technologies. They also provide semantic and visual meaning and structure to the document. A first level heading (\<h1\>) should be present on nearly all pages. It should contain the most important heading on the page (generally the document title).

Analysis
- The global header (first view that is rendered) begins with an \<h2\> tag to display a hello message.
-  Most page views do not have a \<h1\> tag.  This generates an alert: "A page does not have a first level heading".
- Those views that do have an \<h1\> tag have this alert, instead: "A heading level is skipped."
  - Lichen Plagiarism view

Recommendation
- Change the "hello" message from \<h2\> to \<h1\>.  This will make the "hello" message larger.  Therefore, use CSS to restore message to original text size.
- Those page views that do have an \<h1\> tag should be downgraded to a lower level, and then use CSS to restore text size.  Otherwise, the "A heading level is skipped." alert will persist.

pbailie:
Close #2309 

- Breadcrumbs are now wrapped in `<div>` as suggested in issue conversation.
- CSS updated accordingly so breadcrumbs do not visually change.
- e2e testing updated accordingly.

NOTE:
- Discussion forum still throws a header hierarchy warning, but other pages appear to be OK.
- Although probably not affecting production — "Page details" (shows in the footer when debugging is enabled in `config/database.json`) also throws header hierarchy warning.


site/public/css/server.css
```diff
@@ -393,7 +393,17 @@ table tr.table-header {
...
 }
 
-#header-text>h2:nth-child(2) {
+#header-text>div {
+    display: block;
+    font-size: 1.5em;
+    margin-block-start: 0.83em;
+    margin-block-end: 0.83em;
+    margin-inline-start: 0px;
+    margin-inline-end: 0px;
```
MasterOdin: While I do think these tags are neat, they are marked [experimental](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-start) so I think we should look into some other tag for styling

site/public/css/server.css
```diff
@@ -393,7 +393,17 @@ table tr.table-header {
...
+    font-weight: bold;
+}
+
+#header-text>div:nth-child(2) {
```
MasterOdin: I don't think this rule gets triggered anymore and could be deleted.

site/public/css/server.css
pbailie: Those were the tags that Chrome was using to style the text as `<h2>`, per Chrome's developer tools.  But I'll look into other styling.

MasterOdin: I think it looking slightly different is also fine if trying to get the css rules right is a bear.

pbailie: The `margin-block-start` and `margin-block-end` lines appear to be forcing the element to be vertically centered.  We can accomplish the same thing with:
```css
position: absolute;
top: 50%;
transform: translateY(-50%);
```

site/public/css/server.css
pbailie: Done.

mcglow2:
Whats the status of this PR?  Is there still an outstanding change request?  This and PR #3590 are necessary to begin the web accessibility review of Submitty.

bmcutler:
@MasterOdin any final comments or is this ready to approve?

site/public/css/server.css
```diff
@@ -393,8 +393,15 @@ table tr.table-header {
...
-    padding-top: 25px;
+#header-text>div {
+    display: block;
+    font-size: 1.5em;
+    font-weight: bold;
+
+    /* vertical centering */
+    position: absolute;
+    top: 50%;
+    transform: translateY(-50%);
```
MasterOdin: Playing around with this, it looks like the positioning stuff is unnecessary and that removing it would leave the content in the exact same place?

bmcutler: Actually, it looks worse with these lines of code.  If the gradeable title is really long, and overflows a narrow page, with these lines, the full breadcrumb is not visible.  Without these lines it is all visible.  

with:
Screen Shot 2019-04-22 at 10 46 52 PM

without:
Screen Shot 2019-04-22 at 10 57 20 PM

bmcutler: So I've removed it.  It can be added back in (w/ modifications) if necessary.

PR #3567, opened 2019-04-13
replaced occurences of adduser with useradd

### LINKED ISSUE #2392: Replace usage of submitty_utils.glob with Path().glob
We can get rid of this shim by rewriting any usage of it with [Path().glob()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) which natively supports '\*\*' on Python 3.4+ (as opposed to the glob module which only supports '**' on 3.5 upward).

### LINKED ISSUE #2398: Replace usage of adduser with useradd
To make the installation script more portable (and work for #2117), we should be using the `useradd` utility instead of `adduser` as the former is a builtin to Linux and operates the same across distros while the latter is somewhat distro specific (Debian/Ubuntu have their own thing while CentOS/Fedora symlink adduser to useradd).

Fenn-CS:
Signed-off-by: Fenn-CS <fenn25.fn@gmail.com>

Uses version dependent `adduser` instead of more independent `useradd`   in linux scripts Fixes #2398

### What is the new behavior?
Uses `useradd`   in linux scripts


MasterOdin:
While I appreciate the idea, unfortunately it's not just a straight replace of one command with another as useradd does not have the same flags. You'll have to explore how to convert the flags of adduser to useradd to get this to work.

Fenn-CS:
@MasterOdin Thanks, I noticed from the Travis builds... I am correcting currently correcting those.

bmcutler:
@Fenn-CS Please thoroughly test this PR by doing a vagrant destroy and vagrant up and confirm the system is fully functional.  Then you can re-open this PR.  Thanks.

PR #3568, opened 2019-04-15
[FEATURE] NULL user email field Possible

### LINKED ISSUE #3561: NULL user email field
Let's allow a users email field to be NULL.  
This might be a small change to the database, and the various forms that allow us to create/edit this info.
This should be rare, possibly only for admin users.
If a users email is NULL, let's not attempt to send them email for the various notifications.

drumilpatel2000:

Fixes #3561 

### What is the new behavior?
Instructor can change email address in `Manage Student`
User can set to be null and null email address will not get any email


migration/migrator/data/submitty_db.sql
```diff
@@ -211,6 +211,7 @@ ALTER TABLE ONLY sessions
 ALTER TABLE ONLY courses_registration_sections
     ADD CONSTRAINT courses_registration_sections_fkey FOREIGN KEY (semester, course) REFERENCES courses(semester, course) ON UPDATE CASCADE;
 
+ALTER TABLE ONLY users ALTER COLUMN user_email DROP NOT NULL;
```
shwehtom89: No need to alter a column in a table right after you created it. 
Please just delete the NOT NULL constraint from the CREATE TABLE on line 125

migration/migrator/migrations/course/20190415092504_null_email_address.py
```diff
@@ -0,0 +1,7 @@
+def up(config, database, semester, course):
+    database.execute("ALTER TABLE only users ALTER COLUMN user_email DROP NOT NULL")
+    pass
+
+
+def down(config, database, semester, course):
```
shwehtom89: Please add a down function for adding this migrator script to add the constraint back. If this change ever needs to be rolled back on it will make things easier.

bmcutler:
please re-open when issues addressed.

PR #3569, opened 2019-04-15
Updates to Support Custom Python Image Difference

emaicus:
### This PR:
* Updates python custom validators to take their input via a json file (their custom validaton block in the config json) rather than from the command line.
* Fixes a bug which caused python custom validators to improperly set resource limits.
* Allows instructors to specify the name of image difference output files rather than auto-generating them.
* Adds a new package to ```rpi.sh``` for use in image difference.
* Updates relevant more_autograding_examples.


bmcutler:
please fix merge conflicts (Leon's PR rearranged some files)

PR #3572, opened 2019-04-15
Fix user autofill for team gradables  

shailpatels:
Opening a Team gradeable throws a PHP error when trying to populate the autofill data when logged in
as an instructor

### What is the new behavior?
Checks if students are actually in the autofill data array as keys. This prevents PHP from throwing an
error. Student ids may not be loading correctly for team gradeables when populating autofill data.

### Other information?
Steps to reproduce:
Log in as an instructor and press submit/resubmit on a team gradeable.


bmcutler:
@shailpatels 
this code creates alot of php warnings when installed, please fix

Screen Shot 2019-05-02 at 1 55 37 PM

PR #3578, opened 2019-04-16
Display error on blank search input

### LINKED ISSUE #3577: Search bar blank input
**Describe the bug**
The search bar in Discussion Forum takes blank input and returns a no result found page.

https://www.loom.com/share/9289b188795645cb9cef26dce862d9f6

**Expected behavior**
Blank input should not be allowed to the search bar unless a blank query displays all possible results. But the latter will be very inefficient and hence, the search query should not be blank.

**To Reproduce**
1. Open Discussion Forum
2. Click on Search
3. Click on Search again

xprilion:

Closes #3577 

### What is the new behavior?
Warning is displayed when no input is made. Otherwise, on whitespace input the search does not get initiated.

https://www.loom.com/share/9558d2623e4448c78617e6eb50e402ad


drumilpatel2000:
A piece of advice @xprilion use `Fixes` instead of `Closes` when you submit PR which will be used to solve a particular issue as template says `<!-- List issue if it fixes/implements one using the "Fixes #<number>" syntax -->`

drumilpatel2000:
Apart from this I did not see this as problem 
* If I search nothing I shouldn't get anything
* You can see google implementation for blank input when I search whitespace in google I get following output
Screenshot from 2019-04-17 03-33-58

site/app/templates/forum/ForumBar.twig
```diff
@@ -31,8 +31,8 @@
 	{% endif %}
 
 	{% if thread_exists %}
-		<form id="search_block" style="float:right;position:relative;top:3px;right:5px;display:none;" method="post" action="{{ core.buildUrl({'component':'forum', 'page':'search_threads'}) }}">
-			<input type="text" size="35" placeholder="search" name="search_content" id="search_content"/>
+		<form id="search_block" style="float:right;position:relative;top:3px;right:5px;display:none;" method="post" action="{{ core.buildUrl({'component':'forum', 'page':'search_threads'}) }}" onsubmit="if($('#search_content').text().replace(/ /g,'').length == 0){ return false; }">
```
MasterOdin: Would it make more sense to just do:
```
return $('#search_content').text().trim().length !== 0;
```
?

drumilpatel2000: Probably we can use onchange  which will change normal value with trimmed value as at last trim function is somewhat applied in searching in some ways so let us apply before submission and It will also take care of whitespaces as trimming whitespaces will result into blank output

drumilpatel2000: @MasterOdin Is this Issue needed to be fix?

xprilion:
@drumilpatel2000 No the script is not preventing any search other than that which is just whitespace. 

The inline script is expected to strip whitespaces from within anywhere in the string instead of just from the terminals, which is what trim() does, as mentioned here - https://www.w3schools.com/jsref/jsref_trim_string.asp

@MasterOdin your rewrite definitely makes more sense. Also, I'll add support for trimming non-space whitespace characters to the inline script once I get up in the morning (4:06am here 😅)

xprilion:
@drumilpatel2000 thanks for the suggestion about the onchange! I'll implement and push in the morning :)

drumilpatel2000:
@xprilion You can see following implementation
https://github.com/Submitty/Submitty/pull/3478/files#diff-eb1684055642fbb919ca9c2626e8dc14

xprilion:
@drumilpatel2000 Google handles blank search input differently on different devices. Facebook does not permit blank search, as do many other apps/websites including Google on Android. 

Definitely it is intuitive to User Experience that a blank input search should not occur!

xprilion:
Also could you please let me know how you reached that Google blank input screenshot? 😅 My attempts at blank input in Google on Chrome are going vain!!

drumilpatel2000:
Just put a whitespace than submit You will get result as expected

xprilion:
Hi @bmcutler could you please review the update? 

Thanks @drumilpatel2000 for the suggestion! Led me to find a tiny something in the Authentication UI, have added comment!

andrewaikens87:
@xprilion please resolve the merge conflicts.

PR #3580, opened 2019-04-18
added identicon pngs for broken or missing student photos

### LINKED ISSUE #3503: Student Photos displays students without images as broken img tags
On the Student Photos page, currently, when a student does not have any image uploaded for him/her, the list of students is displayed with broken img tags defined as - 

`No Image Available`

However, it would be more intuitive to the design if instead a placeholder image was displayed, or a properly formatted Error message. Using a broken img tag does not feel correct. 

Screenshot from 2019-03-27 18-38-28

Fenn-CS:
Signed-off-by: Fenn-CS <fenn25.fn@gmail.com>

On the student photos page, students with no image uploaded are displayed with broken `img` tags defined as Fixes #3503  

### What is the new behavior?
Now displays an identicon (using https://github.com/stewartlord/identicon.js) for such students.

identicons

### DIFF
--- a/site/public/js/images.js
+++ b/site/public/js/images.js
@@ -2,18 +2,14 @@ $(document).ready(function(){
 
 $('.no-image-available').each(function(i, obj) {
 	var student_name = $(obj).data('student-name');
-	// Identicon requires a harsh of at least 15 chars
+	// Identicon requires a hash of at least 15 chars
 	// So to prevent short names from breaking it, the name is salted.
-	// This would ensure resulting harshes are longer than 15 chars.
+	// This would ensure resulting hashes are longer than 15 chars.
 	student_name = student_name+'0123456789_abcdefghijklmnopqrstuvwxyz';
     // create a base64 encoded PNG
-    var data = new Identicon(uniqueHarsh(student_name), 420).toString();
+    var data = new Identicon(btoa(student_name), 210).toString();
     // update no image tag.
     $(obj).replaceWith('<img width=210 height=210 src="data:image/png;base64,' + data + '">');
 });
-
-function uniqueHarsh(string) {
-    return btoa(string);
-}
 
 });
--- a/site/public/js/images.js
+++ b/site/public/js/images.js
@@ -1,0 +1,20 @@
+$(document).ready(function(){
+
+$('.no-image-available').each(function(i, obj) {
+    // create a base64 encoded PNG
+    var data = new Identicon(randomString(32), 420).toString();
+    // write to a data URI
+    $(obj).replaceWith('<img width=210 height=210 src="data:image/png;base64,' + data + '">');
+});
+
+function randomString(len, charSet) {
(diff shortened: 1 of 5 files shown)


Fenn-CS:
@drumilpatel2000 sounds great... However, this would come at the little expense of not having many random colors :). 

Sending in an update that addresses the review.

site/public/js/images.js
```diff
@@ -1,20 +1,19 @@
 $(document).ready(function(){
 
 $('.no-image-available').each(function(i, obj) {
+	var student_name = $(obj).data('student-name');
+	// Identicon requires a harsh of at least 15 chars
+	// So to prevent short names from breaking it, the name is salted.
+	// This would ensure resulting harshes are longer than 15 chars.
```
MasterOdin: sp: `harshes` -> `hashes`

site/public/js/images.js
```diff
@@ -1,20 +1,19 @@
...
+	student_name = student_name+'0123456789_abcdefghijklmnopqrstuvwxyz';
     // create a base64 encoded PNG
-    var data = new Identicon(randomString(32), 420).toString();
-    // write to a data URI
+    var data = new Identicon(uniqueHarsh(student_name), 420).toString();
```
MasterOdin: remove the one-line `uniqueHarsh` function and just do the `btoa(student_name)` here.

site/public/js/images.js
Fenn-CS: Noted thanks

site/public/js/images.js
Fenn-CS: Alright, thought of this but... doing that right away. Thanks.

MasterOdin:
@bmcutler ready to go assuming you agree with the design

bmcutler:
Please remove the identicons, they do not add any value, they only falsely imply that this image is assigned to the student elsewhere on the site.  A simple text message "no image available" is more appropriate in this situation.

Fenn-CS:
Hello, @bmcutler thanks for the information, however, aren't identical used in the exact scenario where the user has not uploaded an image? otherwise, what is the point? 
And yes, @drumilpatel2000 suggested we used student names (plus a salt) that generates unique hashes in most cases, so it's actually an assignment to the student, if there are other places within the site where this is supposed to appear then we can implement it there as well.

My opinion is, I can't seem to find a modern web development scenario in which text is actually preferred to a default avatar, "place-it" style placeholder or something just like what we've got here.

I would close the PR if there's no further information about this, as _"no image available"_ is actually the current behavior on my browser Firefox (66). That appears without the broken image sign but  that's not the case with @xprilion as shown in the image here: https://github.com/Submitty/Submitty/issues/3503#issue-425949755

bmcutler:
These are not used as avatars, and we don't have plans to do so.  They are just used to instructors and TAs can learn the student names in large courses.

So I'm just going to close this PR and issue.

PR #3586, opened 2019-04-18
Fixing Errors Caused by Special Characters in Emails 

### LINKED ISSUE #3581: Special characters in email announcements break stuff
**Describe the bug**
I tried to put '¯\_(ツ)_/¯' in a discussion forum announcement w/ email.  The announcement posted to the forum, but the emails did not send.

Looking in the email logs we had this error:
Error Sending Email: 'ascii' codec can't encode character '\xaf' in position 873: ordinal not in range(128)

**Expected behavior**
Either escape/encode this character properly, or if that doesn't work for some reason, remove the character.
Make sure it handles all special characters, actually.

kasra-sheik:
fixes #3581 

Tested by recreating error and then encoding the email string and making sure emails containing special characters were sent.

### DIFF
--- a/sbin/send_email.py
+++ b/sbin/send_email.py
@@ -114,7 +114,7 @@ def send_email():
     for email_data in queued_emails:
         email = construct_mail_string(
             email_data["send_to"], email_data["subject"], email_data["body"])
-        mail_client.sendmail(EMAIL_SENDER, email_data["send_to"], email)
+        mail_client.sendmail(EMAIL_SENDER, email_data["send_to"], email.encode('utf8'))
         mark_sent(email_data["id"], db)
 
     LOG_FILE.write("[{}] Sucessfully Emailed {} Users\n".format(
--- a/sbin/send_email.py
+++ b/sbin/send_email.py
@@ -98,8 +98,19 @@ def mark_sent(email_id, db):
 
 def construct_mail_string(send_to, subject, body):
     """Format an email string."""
-    return "Content-Type: text/plain; charset=utf-8\nTO:%s\nFrom: %s\nSubject:  %s \n\n\n %s \n\n" % (
-        send_to, EMAIL_SENDER, subject, body)
+    headers = [
+        ('Content-Type', 'text/plain; charset=utf-8'),
+        ('TO', send_to),
+        ('From', EMAIL_SENDER),
+        ('Subject', subject)
+    ]
+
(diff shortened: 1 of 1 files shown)


MasterOdin:
I guess adding the headers wasn't necessary?

kasra-sheik:
@MasterOdin I was able to get rid of the errors cause by special characters and verify the expected text was being sent upon inspecting nullsmtp logs. However, I'm unsure how an email client would respond to/display the text. I can add the headers if you think it's still necessary.

MasterOdin:
We should definitely include the content-type header to tell email clients that this is UTF-8 content, or else they may (per the email RFC) just assume the content is ASCII and cause it to not display it properly. Whether we choose that to be text/plain or text/html depends on what kind of content we're sending (e.g. how are links and code blocks and things embedded into the emails?)

bmcutler:
For this PR, let's fix the failure to send the emails.  Yes, it would be good to handle links and code blocks better, but that might take a while.  Let's open a separate issue for that.

MasterOdin:
Then I guess let's just go with `test/plain` for now with charset set to utf-8

kasra-sheik:
Just added encoding to the email header. Tested with an email client to be sure expected behavior was being seen. @MasterOdin

sbin/send_email.py
```diff
@@ -98,7 +98,7 @@ def mark_sent(email_id, db):
 
 def construct_mail_string(send_to, subject, body):
     """Format an email string."""
-    return "TO:%s\nFrom: %s\nSubject:  %s \n\n\n %s \n\n" % (
+    return "Content-Type: text/plain; charset=utf-8\nTO:%s\nFrom: %s\nSubject:  %s \n\n\n %s \n\n" % (
         send_to, EMAIL_SENDER, subject, body)
```
MasterOdin: Let's make this a bit more readable and easier to parse:

```
headers = [
    ('Content-Type', 'text/plain; charset=utf-8'),
    ('TO', send_to),
    ('From', EMAIL_SENDER),
    ('Subject', subject)
]
msg = ''
for header in headers:
    msg += "{}: {}\n".format(*header)
msg += "\n\n{}\n\n".format(body)
```

This will also fix the error in Travis on linting this file.

MasterOdin:
I'm good with these changes @bmcutler

PR #3590, opened 2019-04-20
Add Additional Information Links To Footer

### LINKED ISSUE #3332: Customizable page footer
It would be beneficial for site admins to have the ability to add a bit of custom text and HTML to the page footers.

**Specific use case:** Adding links, messages, or disclaimers as required by policy.  Such as RPIs web accessibility policy which requires a link to it on each page.

pbailie:
Fixes #3332 

- A `footer_links.json` file may be added to `submitty/config/` to add additional information links to the footer.  Links are appended after the "Github" icon.  Json data format follows the same standard as `sidebar.json`.  Multiple links may be defined and they will be appended in order.

- This will permit RPI (and other institutions) to add policy-important information links for the public.  i.e. web accessibility policy.

- Links defined by JSON file are individually validated for required attributes (icon, link, and title).  Any Link that fails to provide any of the required attributes will not be shown.  (but there should be no unhandled exception screen)

- Footer links have `aria-label` defined for VPAT compliance.

- Improved visual spacing of all footer elements as footer links can potentially create a lengthy list.

Screenshot (see bottom of screenshot):
Screen Shot 2019-04-19 at 7 59 31 PM


MasterOdin:
I'm not sure we need to make all three things always required. I think it's probably better to just make it required that there's at least either an icon or text, but that it's not required to have both. A link is totally optional.

mcglow2:
I agree that is should allow for either an icon or text or both.  With regard to my accessibility link use case, an icon is not needed.

pbailie:
OK — I'll update the PR to validate either icon (with an aria-label) or text.  In response that a link "is totally optional" — the use case in #3332 specifically describes providing a link to RPI's accessibility policy:
> Specific use case: Adding links, messages, or disclaimers as required by policy. **Such as RPIs web accessibility policy which _requires a link_ to it on each page.**

MasterOdin:
That reads to me that the RPI web accessibility policy has to be a link, but not that every possible thing in the footer must have a link.

site/app/models/Config.php
```diff
@@ -292,6 +296,31 @@ public function loadMasterConfigs($config_path) {
...
+                    default:
+                        //Validation OK.  Include $row.
+                        if (isset($row['icon']) && !Utils::startsWith($row['icon'], "fa-")) {
+                            $row['icon'] = "fa-" . $row['icon'];
+                        }
+                        $this->footer_links[] = $row;
+                    }
+                }
+            }
+        }
```
MasterOdin: I'm not sure this array is something we're going to use reference beyond just the one time you have it so config is probably not the right place for it. I think it might make more sense to just make `getFooterLinks()` be a function off of `app/libraries/Output.php` that just opens the file, reads from it, and returns the resultant array of footer links (and not bother storing it anywhere as we just use it the once. We can refactor for performance if we find we're getting these links more often).

bmcutler:
In our meeting today we decided to change the footer to contain by default 3 things in this order:

COPYRIGHTSYMBOL 2019 Submitty (linked to Submitty) 
GITHUB SYMBOL (linked to submitty github)
An RCOS Project (linked to RCOS)

(separated by vertical bars with a little more space than currently), followed by any optional items...  at RPI we will probably have two additional items:

RPI CS department logo (linked to our rpi cs department pages)
assessibility (linked to the RPI page)

Peter is going to get much of this ready ASAP (today/tomorrow) for an initial merge and anything not finished (items w/o links, icons that aren't fontawesome, additional tweaking) we'll open as a new issue for future PR.

site/app/models/Config.php
pbailie: I put the footer links in the system config class because the issue discussion agreed this was a system-level configuration.

Going right to the Output class would circumvent the GlobalController and GlobalView.  GlobalController and GlobalView are responsible for the footer.  Therefore I've move the Footer Links "configuration" to the GlobalController's `footer()` class.

Issue #3562, opened 2019-04-11
emphasize DO NOT REPLY TO automated emails


bmcutler:
**What problem are you trying to solve with Submitty**
Students hit 'reply' to an automated submitty email/instructor announcement.
Depending on configuration, this might be dumped to an unread email account, or lost.  Most likely does not get seen by the intended recipient (the instructor?).

**Describe the way you'd like to solve this problem**
1. Let's make sure we set any appropriate standard email fields for the 'reply-to'...
Research if there are options to help email clients help users not send bad reply-to emails...

2. And let's append a string to the top and/or bottom of the message 
DO NOT 'REPLY TO' THIS EMAIL.
(let's design discuss this string... )

@holzbh please add comment with suggestions.
@kasra-sheik please implement :)

edited to add:   in the 'footer' let's include the course name again (in case the subject line gets lost/mangled)

bmcutler:
Also let's put the course number & course title (if configured) in the email body text.
To make it clear why you are receiving this email (possibly who sent it if it was an announcement)
Perhaps in the bottom/footer?  (So as not to clutter the top of the email)

MasterOdin:
There isn't a way to help email clients "help users not send bad reply-to emails". Your options are pretty limited in what can be done for no-replay emails:
1. Blackhole any incoming emails and don't bother to look at them (e.g. no-replay@cs.rpi.edu could lead to an inbox that just auto-deletes emails and no one looks at it)
2. Have the address not be a valid one (so that when they send, they get an immediate "this email failed to send response")

However, the client has no way of really recognizing either and will always treat whatever reply-to address you use as a real one that could get a response.

At best, you'd want to put a notice at the top/bottom of the email that's just:
```
********************************************************
This is an automated email.
Any responses will not be looked at or responded to.
********************************************************
```

holzbh:
It would still be nice to actually have a reply-to that was something like no-reply@cs.rpi.edu or no-reply-submitty@cs.rpi.edu so that when someone starting composing a reply it would be a little more apparent that no, this does not look like a good email address to contact.

I agree that notice in the body of the email is the best option, and if users disregard it, well, they shouldn't be surprised when their e-mail is ignored / lost.

PR #3597, opened 2019-04-23
Emphasize 'DO NOT REPLY' in automated emails 

kasra-sheik:
closes https://github.com/Submitty/Submitty/issues/3562

Emails within Submitty lack a DO NOT REPLY indicator which has caused some users to reply to automated emails. 

### What is the new behavior?
In addition to adding a DO NOT REPLY message in the footer of the email body, this PR refactors some of the email code to look almost identical to Notifications by creating an Email model. This way, we can keep our email templates in one place and increase maintainability of email functionality within Submitty. 

Below is an example of a discussion forum announcement email created from the model:

**Subject**
`[Submitty Data Structures] Homework 6 Due Date Pushed to 04/25` 

**Body** 
```
An Instructor/TA made an announcement in the Submitty discussion forum:

Hello Class,

Because we didn't finish covering recursion in class, I have decided to push the due date of Homework #6 to 04/25. 

Professor Foo 


--
NOTE: This is an automated email.
Any responses will not be looked at or responded to. 
``` 

@bmcutler @holzbh  Feel free to suggest any changes to this template. 

### Other information?
Tested all current email types to make sure do-not-replys are being injected and all other expected behavior regarding email sending.


site/app/models/Email.php
```diff
@@ -0,0 +1,78 @@
...
+    private function handleForumAnnouncement($subject, $body) {
+        $formatted_subject = $this->formatSubject($subject);
+        $formatted_body = "An Instructor/TA made an announcement in the Submitty discussion forum:\n\n".$this->formatBody($body);
+
+        $this->setSubject($formatted_subject);
+        $this->setBody($formatted_body);
+    }
+
+    private function handleSeatingAssignment($subject, $body) {
+        $formatted_body = $this->formatBody($body);
```
MasterOdin: Why does this function not use formatSubject? Aren't these emails related to a specific course?

kasra-sheik: So in the case of seating assignments we preview the subject/email to the sender, so I have the [Submitty $course_name] already injected. However, this can be changed/refactored for consistency reasons.

MasterOdin: Yeah, we should refactor this so that we don't have two different methods where that bit gets injected into the title, I don't really have much of a preference on whether that's before we construct the model, or if it's part of a factory method within the model (e.g. `createCourseEmail`). Thinking about this, we can probably move the bit about "Instructor/TA has made an announcement..." into where the the email model is constructed for announcements. I don't think we will re-use that bit beyond for just announcements, whereas as `createCourseEmail` factory method would take in a $core, title, body and then prepend "[Submitty {$course_name}]" to the title, and then `return new Email($core, $title, $body);`.

kasra-sheik:
I also added a 'reply-to' in the email header for extra clarification

PR #3622, opened 2019-04-30
[BUGFIX]Fix unhandled exception when no thread in discussion forum 

### LINKED ISSUE #3621: Unhandled Exception when accessing discussion forum with no posts
If the discussion forum is empty (such as a brand new course that hasn't had any discussions), an unhandled exception is thrown.  Further, PHP gives an undefined variable error.

This can be reproduced in Vagrant by going to the "blank" class, enabling the discussion forum in the options, and then try to access the discussion forum.  I have also recreated this in my development VM in trying to access the discussion forum from a newly created course.

Screenshot from Vagrant:
Screen Shot 2019-04-30 at 4 46 30 PM

drumilpatel2000:

Fixes #3621 
Due to new features implemeted in #3268, queries viewed_response was getting error when no thread id was passed(thread_id = null)  

### What is the new behavior?
Screenshot from 2019-05-01 04-45-27

### Other information?

1) Check on the discussion forum in development(initially not checked)
2) An error was shown as no thread was initially


drumilpatel2000:
@pbailie Can you review this PR?

MasterOdin:
Yeah, this PR isn't really the way we want to fix this. We should modify the `ForumThreadView.php` to not even call this function as well as fix whatever else it might be trying to do with its assumption we'd always have a thread.

PR #3634, opened 2019-05-02
Grade Inquiry Notifications 

kasra-sheik:
Absence of notifications on a grade inquiry 

### What is the new behavior?
On both grade inquiry creation/discussion, relevant users will be notified/emailed. 

### Other information?
Only Full Access Graders are notified. 

Notification/Email settings should be updated in coordination with this PR (once that PR has been merged)


bmcutler:
@kasra-sheik @andrewaikens87 
Can one of you please fix the merge conflicts with this branch and then implement the connection with the notifications from the description.

If the remaining work is non-trivial, but this can be safely merged, then please make a new issue for that work and approve this as a partial work PR.

Thanks

shailpatels:
@bmcutler I think this should be merged as is so we don't keep getting merge conflicts. Its works but the only issue is that notifications show the gradeable id instead of the gradeable title which requires someone to go through and change that which is sort of a nontrival task but would be a good new first issue

Issue #3630, opened 2019-05-02
Add requirements.txt for python packages


MasterOdin:
Right now, we don't have an obvious place for what python packages are required for what component (as well as version). If we want to use automatic check of versions (#3611), we need to add these files somewhere that contain this information.

Python packages:
- [ ] migrator (add under `migration/`)
- [ ] autograder (add under `autograder/`)
- [ ] submitty_utils (add under `python_submitty_utils/`)
- [ ] submitty_scripts (add it somewhere logical under `.setup/`)

wusatosi:
Dependencies in install_systems.sh with latest versions
```
python-pam==1.8.4
PyYAML==3.12
psycopg2-binary==2.8.6
sqlalchemy<1.4.0
pylint==2.8.1
psutil==5.8.0
python-dateutil==2.8.1
watchdog==2.0.3
xlsx2csv==0.7.8
pause==0.3
paramiko==2.7.2
tzlocal==2.1
PyPDF2==1.26.0
distro==1.5.0
distro-info===0.18ubuntu0.18.04.1
jsonschema==3.2.0
jsonref==0.2
docker==5.0.0
```

MasterOdin:
I'm not sure where you're getting `distro-info` from?

PR #3650, opened 2019-05-04
[Refactor] Add crontab framework for submitty_users

### LINKED ISSUE #3256: Standardize Cron Usage in Submitty
As of now, we haven't set a standard for adding/editing cron jobs throughout Submitty. It would be nice to have a centralized place where our jobs live so that future developers can easily manage them. 

@MasterOdin has suggested adding a static file in our .setup directory called `crontabs`. During the install process, we could just read in from that file and assign the job to the appropriate user.

This PR should move the cron job responsible for sending emails (and any other jobs that might exist at the time of reading this) into our static file(s). This way we have example usage for creating/installing cron jobs in Submitty moving forward.

MasterOdin:

Closes #3256 

### What is the new behavior?
This adds a new directory `.setup/submitty_user_scripts` which then has a directory per user to house crontab (and eventually other things like pip packages in the form of requirements.txt files for #1322 and #3630). We can use the expected assortment of `SUBMITTY_*` bash variables as `envsubst` substitutes those in for us as part of the crontab installation process. 

### Other information?

This is not a breaking change and does not change the overall functionality of Submitty.

I tested by deleting the crontab completely and then doing `submitty_install` and then verifying that the crontab was as expected. I also add extraneous stuff to the crontab to verify that whatever is in the file is whatever is going to be in crontab after running submitty install.


MasterOdin:
Closing this as need to have larger discussion on how sysadmins may use/edit cron independently of Submitty and what our policy about it should be.

PR #3655, opened 2019-05-05
[FEATURE]Chronologically Descending Sorting option 

### LINKED ISSUE #3409: Enable users to sort posts from newest to oldest in a thread
Currently, there are three ways to sort posts in a thread: hierarchical, chronological and alphabetical. Among them, chronological mode sorts all posts in a thread from oldest to newest.

However, there may be a scenario where the thread is very long and some users may want to know the latest updates. It would be tiring to scroll all the way down the page. Hence, it may be helpful if, instead of oldest to newest by default, there can be two chronological modes: ascending and descending.

drumilpatel2000:

Fixes #3409 

### What is the new behavior?
A new option to sort from newest to oldest post in particular thread
Screenshot from 2019-05-05 14-59-05


bmcutler:
Please chose a label without the awkward abbreviation.  
Maybe spell it out or even use up and down arrows?

andrewaikens87:
I made some changes to this branch locally but, unfortunately in doing so I had to create a new local branch. I suggest that we should be using font awesome indicators rather than `&uarr;` and `&darr;`. Possibly `fa fa-angle-down` and `fa fa-angle-up` would be appropriate. Also we should add titles to each link to provide a description.

Hierarchical: `Sort posts by reply hierarchy`
Chronological (with up arrow): `Sort posts by ascending chronological order`
Chronological (with down arrow): `Sort posts by descending chronological order`
Alphabetical: `Sort posts by author's last name`

Issue #300, opened 2016-03-07
Investigate use of REST endpoints


MasterOdin:
Should check if we can make available REST endpoints without being logged in to the server (the endpoint should involve the authentication). We should start with the very basic example of "Get All Open Assignments" and return that in an ordered JSON list. If that is possible, than we can expand it as need arises.

MasterOdin:
So the front end of the system can start to support this as we've moved away from the apache authentication and moved into using sessions. We just need to make endpoints and discuss how we want to lay out the code and paths for this.

We most likely want to make an "API" controller, use some method to version it, and and then separate it out as necessary from there.

MasterOdin:
We would want to rewrite the current controllers to have a better divide between "getting information" and putting it into the models. An example of this would be for getting graders we might want to separate the current method of `listGraders()` into `getGraders()` and `displayGraders()` where the first method would incorporate the first line and the second would incorporate the second two lines (a pedantic example). This way the api controller can utilize the former method and the web view can utilize both.

prestoncarman:
I am researching how to interface with Submitty from other applications. A REST API would be a perfect way to make those requests. The comments posted here are from 2016. Do you still feel this approach to adding a REST API is the best way to proceed at this point in time?
- Adding a API controller
- Refactor the current methods into a get data and display functions.
- Building on your example, the endpoint could be a GET request to /api/v1/graders
What are your thoughts about using a library to help with the REST API details (routing, token authentication, etc.)?

MasterOdin:
I'm not sure really what the best method here would be, and I've waffled through a couple different designs in my head over the years. Right now, I think I'd lean more towards having an `api.php` in `site/public/` with Apache setup to route calls to `submitty.example.com/api/<API_STRING>` into calling `api.php` and then having that translate the rest of the route into something Submitty could use. From here, I think modifying the existing controllers to always return something (and be able to set $core->output to a NullOutput so any view calls are silently just swallowed and then turn off redirect() happening). This would also make it increasingly possible to unit test our controllers 
and hopefully improve stability.

With regards to libraries, what would you have in mind? I'm familiar with a couple pieces of symfony, but would be interested to hear what else you might of used/heard good things about.

prestoncarman:
[...]

PR #3663, opened 2019-05-09
Implement simple router for semester/course

### LINKED ISSUE #933: Add in basic routing (semester + course) for URL
We should be able to go to:
https://submitty.cs.rpi.edu/s17/csci1100/index.php
https://submitty.cs.rpi.edu/csci1100/index.php

and be taken to the course page for it (instead of having to do `index.php?semester=s17&course=csci1100`, though this might still be supported?)

More complex style of routing will come later once things are better defined and all functionality is then routed through that central index.php. The more complex style can be tested within a REST API (#300)

MasterOdin:

Closes #933 

### What is the new behavior?
This adds in a basic router that can utilize mod_rewrite within apache to allow us to access pages using both `/index.php?semester={$semester}&course={$course}` or `/{$semester}/{$course}`. 

### Other information?

This is fully backwards compatible with the existing setup such that people can go to either the existing URL scheme, or use the new one. Looking forward, we can use the router to pop off pieces to get `$_REQUEST['component']`, `$_REQUEST['action']`, etc.

This is somewhat necessary before tackling a simple setup for some REST endpoints for #300.

### DIFF
--- a/.setup/vagrant/sites-available/submitty.conf
+++ b/.setup/vagrant/sites-available/submitty.conf
@@ -33,8 +33,12 @@
         RewriteCond %{REQUEST_FILENAME} -f
         RewriteRule ^ - [L]
 
-        # Else rewrite urls to be to some form of /index.php depending of if the
-        # url already contains index.php or not
+        # Else rewrite urls to use index.php, however we have to be aware of two
+        # possible conditions of whether index.php is in the URL already or not.
+        # To be backwards compatible, we want to ensure that having /index.php
+        # is valid and usable. In anycase, we rewrite everything up-to /index.php
+        # to be the "url" parameter of the query, and then append whatever else we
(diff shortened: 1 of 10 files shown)


.setup/vagrant/sites-available/submitty.conf
```diff
@@ -46,6 +46,17 @@
         Require all granted
         Order allow,deny
         Allow from all
+
+        RewriteEngine On
```
MasterOdin: If we were going to implement a capacity of turning on/off "pretty links" (e.g. /s19/sample vs ?semester=s19&course=sample), then we'd want to wrap this in a "<IfModule>" check, but I think we're going to make this a mandatory component of Submitty, so no need, and it'd be better for apache to just fail if the module wasn't enabled.

PR #3668, opened 2019-05-10
Use JWT for session cookies

MasterOdin:

We place the session_id as an encoded JSON string into the cookies. We only check if the session exists for some user and if so, assume it's valid. We do not have a mechanism to protect from a user just randomly changing the session id to see if they can find someone else's session (though it's unlikely they should). Additionally, as we don't validate the user_id in some way, an attacker could just randomly attempt session_ids and end up as any user, which could be another student.

### What is the new behavior?
This uses a JWT for sessions and places the JWT string in the cookie. This adds in a signature to the cookie such that we validate it against a secret stored on the server so that for a malicious actor to change the session_id, they'd have to also figure out the valid signature for that token. We also add validation of the user_id matches the session_id to better protect against accidental session hijacking.

### Other information?

This will require all users to re-log into the website. The library used here can also eventually be used for #300


migration/migrator/migrations/system/20190510173527_jwt_session_secret.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def up(config):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    if 'php_user' in config.submitty_users:
```
MasterOdin: this is to check if the machine is a worker machine or not (if it is, it won't have this user defined)

PR #3679, opened 2019-05-14
[UI Tweak] clickable text 'stay logged in'

D3quien0:
Created the ability to click the text next to the checkbox on the login page, so user does not only have to click inside the box.


bmcutler:
Thanks for the contribution!

PR #3680, opened 2019-05-14
Fixed incorrect time zone writing addressed in #3156

### LINKED ISSUE #3156: CONFIGURE_SUBMITTY.py ignores inputted timezone for json

tnatiuk17piano:

Fixes #3156 
Before, the CONFIGURE_SUBMITTY.py script would prompt the user for a timezone, but only write the given time zone to the INSTALL_SUBMITTY.sh script; it would use the default time zone for submitty_conf.json

### What is the new behavior?

Now, when running CONFIGURE_SUBMITTY.py, entering a custom time zone will correctly implement the time zone into both the INSTALL_SUBMITTY.sh script _and_ submitty_conf.json.

### Other information?

Tested using Vagrant; first ran CONFIGURE_SUBMITTY.py from GIT_CHECKOUT inside the VM, provided a custom time zone, and checked the two files after completion; submitty_conf.json had the incorrect time zone. Ran same tests again after making changes, and submitty_conf.json contained the correct time zone.


bmcutler:
Thanks for the contribution!
It looks like you've accidentally checked in your .vs folder.  Please remove that from this PR.
Thanks!

tnatiuk17piano:
Fixed the problem, there should be a new PR.

~Trevor

PR #3685, opened 2019-05-14
Markdown Descriptions for Textbox Submissions

emaicus:
Enables instructors to write markdown descriptions when using the new "content" (textboxes/multiple choice) options. 

@MasterOdin this required a new composer package. Can you check to make sure I did everything correctly?


site/composer.json
```diff
@@ -12,7 +12,9 @@
   },
   "require": {
     "twig/twig": "^2.9.0",
-    "ramsey/uuid": "^3.8.0"
+    "ramsey/uuid": "^3.8.0",
+    "aptoma/twig-markdown": "~2.0",
```
MasterOdin: Why are we using the 2.x and not 3.x for this plugin? `"aptoma/twig--markdown": "^3.1.0"`, `"michelf/php-markdown": "^1.8.0"`

Let's also prefer to use the `^` over `~` unless there's a set reason to not bring in new minor versions.

MasterOdin:
Given that we're adding markdown here, I'd like to see us remove any legacy usage of instructors typing in HTML (such as for little submit text) and use markdown instead to keep things consist. We can create issues for those separate from this though.

MasterOdin:
I would also like to know how the markdown parser handles potential dangerous HTML, e.g:
```
<script>alert("Hello! I am an alert box!");</script>
```

MasterOdin:
To answer my own question, no, this does not sanitize dangerous HTML tags and will happily display the script as is (causing the script to fire).

I'd also probably suggest we use [erusev/parsedown](https://github.com/erusev/parsedown) over the michelf/php-markdown as it looks to be significantly more popular.

I opened aptoma/twig-markdown#30 to expose options on the ParsedownEngine to set it to not trust the input (and sanitize the html).

site/app/libraries/Output.php
```diff
@@ -4,6 +4,8 @@
 use app\controllers\GlobalController;
 use app\exceptions\OutputException;
 use app\models\Breadcrumb;
+use Aptoma\Twig\Extension\MarkdownExtension;
```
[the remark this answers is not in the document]
emaicus: Fixed by moving to use ```CustomParsedownEngine``` (removing the necessity for a ```use``` statement).

site/app/libraries/CustomParsedownEngine.php
```diff
@@ -0,0 +1,67 @@
...
+use Parsedown;
+
+/**
+ * CustomParsedownEngine.php
+ *
+ * Maps erusev/parsedown to Aptoma\Twig Markdown Extension
+ *
+ * @author Sébastien Lourseau <https://github.com/SebLours>
+ */
+class CustomParsedownEngine implements MarkdownEngineInterface
```
MasterOdin: Probably better to extend off `ParsedownEngine` and then you'd just have the constructor:
```
    public function __construct($instanceName = null) {
        parent::__construct($instanceName);
        $this->engine->setSafeMode(true);
    }
```

We don't really need the ability to configure this as I'm not sure we ever have a case where we would want to allow HTML in the markdown.

emaicus:
As @MasterOdin recommended, I've moved to use ```erusev/parsedown``` rather than ```michelf/php-markdown```.  [aptoma/twig-markdown#30](https://github.com/aptoma/twig-markdown/pull/30) has not been merged as of the time of this comment. As a result, I created a local version of the affected file, ```site/app/libraries/CustomParsedownEngine.php``` which contains the contents of the PR. If the PR is eventually merged, we can update to use the standard ```Aptoma/Twig/Extension/MarkdownEngine/ParsedownEngine``` with minimal effort. I've tested and html appears to now be properly sanitized.

emaicus:
Updated implementation of ```CustomParsedownEngine``` to extend off of ```ParsedownEngine``` per @MasterOdin's comment.

PR #3444, opened 2019-03-22
[FEATURE]TA/Instructor can lock Thread

drumilpatel2000:
Fixes #2269 
## Feature
TA and Instructor can lock thread
Screenshot from 2019-03-22 12-05-36
Locked thread can only be edited Instructor
Screenshot from 2019-03-22 12-03-08
Submit button and textarea will be disabled for student and TAs when thread is locked
Screenshot from 2019-03-22 12-08-46
Screenshot from 2019-03-22 12-08-03
Fa icon are given to those thread which have been locked
Screenshot from 2019-03-22 12-11-57

PR #3703, opened 2019-05-20
[Feature] TA/Instructor can lock Thread

### LINKED ISSUE #2269: Forum: Instructor/full access ta can lock thread
Separated from #1677

Instructor/full access ta can specify a 'lock' date for a thread.
After that date, all can view, but the only instructor can edit.

drumilpatel2000:
Continuation of #3444 

Fixes #2269


drumilpatel2000:
@andrewaikens87 Thanks for reviewing. I solved the first issue but I don't get what is the point of second issue. Can you explain in detail about the problem?

andrewaikens87:
Sure! Please let me know if this is unclear.

Here are two examples of the unexpected behavior (the following screenshots are from a student user):

- When creating a thread as a student, if I post a timestamp with my thread data (refer the the image below), I can set a lock thread date.

studentLocked

Then when the form is submitted: 

ah

- Another example is when the student edits their thread. The same behavior occurs. Please refer to the images below.

test2

Then when the form is submitted: 

studentLockedFromEdit

We shouldn't allow students to modify this field. This ability should be left for privileged users (TA/Instructors) only.

drumilpatel2000:
Screenshot from 2019-05-25 14-43-42
@andrewaikens87  Is this okay?

andrewaikens87:
@drumilpatel2000 I agree that we should not display the edit / delete buttons after we lock a thread. 

Although, we need to make sure that the corresponding functionality doesn't work (i.e. using the `deletePostToggle` JavaScript function I am still able to delete posts from a locked thread).

Also, the explained issue above with students being able to change a thread they've created to being locked still persists. I will post a video in the slack channel demonstrating a simple way of recreating the issue.

drumilpatel2000:
@andrewaikens87 Can you please review it again?
Now user cannot delete using javascript and student cannot set thread lock date

PR #3705, opened 2019-05-21
Updated account information in distro setup script

tnatiuk17piano:
Addresses an inconsistency in the "setup_distro.sh" script, which has information on various accounts and their purposes in the VM environment. (https://submitty.org/developer/vm_install_using_vagrant, step 11)

Some of the listed accounts either did not exist or were categorized incorrectly.

### What is the new behavior?
The script now has information that matches the table in step 11 on the aforementioned page. Clarifies the role and existence of each account.

### Other information?
NOTE: The system user "root" does not have the password "vagrant". Tested by logging in as another user, then back into root. Likewise, the system user "postgres" does not have the password "postgres", tested the same way (can access from root without password, but not from any other account)
NOTE: There is no database user "postgres" despite various documentation of the user (https://submitty.org/sysadmin/database_overview)

Tested using Vagrant; first attempted to access all user accounts listed; then attempted to log into the database with `psql` using each user listed; then logged into the Submitty website using available accounts from  (https://submitty.org/developer/vm_install_using_vagrant, step 11). Updated information as necessary.

### DIFF
--- a/.setup/distro_setup/setup_distro.sh
+++ b/.setup/distro_setup/setup_distro.sh
@@ -88,15 +88,17 @@ if [ ${VAGRANT} == 1 ]; then
 ############################################################
 ${DISTRO_LINE}
 ##                                                        ##
-##  All user accounts have same password unless otherwise ##
-##  noted below. The following user accounts exist:       ##
-##    vagrant/vagrant, root/vagrant, submitty_dbuser,     ##
-##    submitty_php, submitty_cgi, submitty_daemon, ta,    ##
-##    instructor, developer, postgres                     ##
+##  All user accounts have same password as name.         ##
 ##                                                        ##
-##  The following accounts have database accounts         ##
-##  with same password as above:                          ##
-##    submitty_dbuser, postgres, root, vagrant            ##
+##  The following accounts are system users:              ##
+##    vagrant, root, submitty_php, submitty_cgi,          ##
+##    submitty_daemon, ta, instructor, postgres           ##
+##                                                        ##
+##  The following accounts are database accounts:         ##
+##    submitty_dbuser, root, vagrant                      ##
+##                                                        ##
+##  The following users can log into the website:         ##
+##    instructor, ta, ta2, ta3, student                   ##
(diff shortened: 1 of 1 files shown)


.setup/distro_setup/setup_distro.sh
```diff
@@ -88,15 +88,17 @@ if [ ${VAGRANT} == 1 ]; then
...
 ##                                                        ##
-##  The following accounts have database accounts         ##
-##  with same password as above:                          ##
-##    submitty_dbuser, postgres, root, vagrant            ##
+##  The following accounts are system users:              ##
+##    vagrant, root, submitty_php, submitty_cgi,          ##
+##    submitty_daemon, ta, instructor, postgres           ##
+##                                                        ##
+##  The following accounts are database accounts:         ##
+##    submitty_dbuser, root, vagrant                      ##
```
MasterOdin: postgres user has a DB account (and is the account we use to create the `submitty_dbuser` and `vagrant` ones), root does not.

```
root@vagrant:/usr/local/submitty# su postgres
postgres@vagrant:/usr/local/submitty$ psql
psql (10.8 (Ubuntu 10.8-0ubuntu0.18.04.1))
Type "help" for help.

postgres=# \du
                                      List of roles
    Role name    |                         Attributes                         | Member of
-----------------+------------------------------------------------------------+-----------
 postgres        | Superuser, Create role, Create DB, Replication, Bypass RLS | {}
 submitty_dbuser | Superuser, Create role, Create DB                          | {}
 vagrant         | Superuser, Create role, Create DB                          | {}
```

MasterOdin:
Yeah, I suspect that `vagrant/trusty64` had the `root/vagrant` setup (which is what we used when I wrote this thing), but that `bento/ubuntu-18.04` does not. The passwords is also probably somewhat irrelevant knowledge given that we we now boot into Vagrant as root to begin with so I agree with removing that line. Other than the postgres thing, this is definitely an improvement on my long stale documentation, and I'm impressed someone validated this info after all these years! 👍

PR #3706, opened 2019-05-21
[Bugfix & Refactor] Email configuration & log location

### LINKED ISSUE #3430: Move email configuration from database.json to its own email.json
Right now, to utilize the email features, you need to write the necessary keys to database.json. This is a bit of a confusing place to look, and we should be putting these details into its own config json file (email.json). The json file should be readable by submitty_daemon:submitty_daemonphp.

As part of this, we should also add an ability to turn/off email as well.

bmcutler:
Moves the email configuration out of database.json and into email.json.   Closes #3430.
Also fixes the log filenames to have leading zeros on the month & date.

### DIFF
--- a/sbin/send_email.py
+++ b/sbin/send_email.py
@@ -37,7 +37,8 @@ try:
 
     TODAY = datetime.datetime.now()
     LOG_FILE = open(os.path.join(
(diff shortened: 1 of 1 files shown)


bmcutler:
documentation added in submitty/submitty.github.io#83

PR #3710, opened 2019-05-21
Rename textbox to short answer Closes #3694

### LINKED ISSUE #3694: Rename "Textbox" to "Short Answer" for Textbox Grading
Rename ```textbox``` to ```short_answer``` in ```content``` when textbox grading. Also re-evaluate the names of other fields.

edited to add:
Also rename ```content``` to    ```notebook```

Kevin-MM:
Closes #3694

When an instructor wants to use config.json to setup a gradeable they could previously define special fields and formatting in a 'content' block.  (See Development -> Paragraph Textboxes samples).

### What is the new behavior?
'content' has been replaced with 'notebook'.  Instructors should now place special formatting inside a 'notebook' block.  Also 'textbox' is no longer a valid input type.  This option is now called 'short_answer'.

### Other information?
Used Development -> Paragraph Textboxes gradable to test functionality.


MasterOdin:
Why was the array of notebooks rename from `$notebooks` to `$notebook` which is contrary to the other arrays we have in that file (e.g. `testcases`, `part_names`, `inputs`, etc.)

emaicus:
We spoke about this a little bit earlier. The name ```$notebooks``` doesn't really fit correctly, as there is only one notebook, which is composed of multiple smaller components, which we haven't actually decided on a name for yet. I suppose something like ```$notebook_sections``` could fit here. Is their an industry standard term for a segment of, say, a Jupyter notebook?

MasterOdin:
This is probably the time to also have more object models for the configuration where we have a `Notebook` model which has various information, to help establish how we reference it as well.

For an explanation of the structure of Jupyter, see https://jupyter-notebook.readthedocs.io/en/stable/notebook.html. Essentially, a notebook is made up of cells (code, markdown, or raw) and then the type of content/actions available depend on the cell.

PR #3712, opened 2019-05-21
Student Photos Supports More Extensions

### LINKED ISSUE #3704: Photo Scraper Tool Uses Wrong File Extension
**Describe the bug**
The [PhotoScraper](https://github.com/Submitty/InstructorTools/blob/master/PhotoScraper/RPI_SIS_PhotoScraper.py) always writes a .png file regardless of image content.

**Expected behavior**
The extension should be .png if the content is truly a PNG. SIS URLs do not specify the type (they are extensionless) but at present the content is actually JPEG content. The extension should thus be .jpg in these cases.

**To Reproduce**
1. Run the photoscraper script at RPI
2. Download from a course with at least one student still at RPI
3. Files are RCSID.png but contain JPEG content. (Adobe Photoshop uses a filter based on extension as opposed to image header content, and will fail to open these .png files but will open them fine if they are renamed to .jpg)

**Configuration**
 - OS: Any
 - Browser: Any

**Other Notes**
-A PR has been made in the InstructorTools repository to change the format, but this issue cannot be closed until the ImageView supports png/jpg/jpeg. 

~~Also have to figure out what to do if someone uploads multiple formats for the same file - should behavior be undefined or should an error be thrown? An error would have to be thrown when viewing the files since someone could just directly upload into `<semester>/<course>/uploads/student_images` so we can't rely on just the upload interface to catch things like RCSID.png and RCSID.jpeg for the same RCSID.~~

holzbh:

* [N/A] Tests for the changes have been added/updated (if possible)
* [N/A] Documentation has been updated/added if relevant

Closes #3704 .

### What is the new behavior?
Supports png/jpg/jpeg.

### Other information?
Tested using student.png, student.jpeg, instructor.png, and instructor.jpg - started with just the pngs, then with everything uploaded, then deleted the pngs. Behavior was as expected - when there's only one image for a given user it's used, but if there's multiple images, it's whatever order the filesystem chooses to iterate in.

As a side note, since this seemed to be geared towards specifically .png previously, I kept the extensions limited, but I think file extension checking might be pointless since we also extract the content type and look for `image/...` types. I could change that in this PR if we want.

### DIFF
--- a/site/app/views/grading/ImagesView.php
+++ b/site/app/views/grading/ImagesView.php
@@ -46,7 +46,7 @@ class ImagesView extends AbstractView {
                     // Read image path, convert to base64 encoding
                     $expected_img_data = base64_encode(file_get_contents($expected_image));
 
-                    $img_name = $fileinfo->getBasename('.png');
+                    $img_name = pathinfo($fileinfo->getBasename(), PATHINFO_FILENAME);
                     if ($img_name === "error_image") {
                         $error_image_data = $expected_img_data;
                     }
--- a/site/app/views/grading/ImagesView.php
+++ b/site/app/views/grading/ImagesView.php
@@ -25,10 +25,7 @@ class ImagesView extends AbstractView {
         }
 
 
-        //$images_data_array to contain base64_encoded image urls
-        $images_data_array = array();
-        //$images_names_array to contain the names of the images (rcs ids)
-        $images_names_array = array();
+        $image_data = array();
         $error_image_data = '_NONE_';
 
         //Get the expected images path and png files to loop through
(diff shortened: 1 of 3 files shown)


site/app/views/grading/ImagesView.php
```diff
@@ -35,16 +35,18 @@ public function listStudentImages($students, $grader_sections, $instructor_permi
...
+            if (!$fileinfo->isDot() && in_array($fileinfo->getExtension(), $valid_image_extensions)) {
 
                 $expected_image = $fileinfo->getPathname();
                 $content_type = FileUtils::getContentType($expected_image);
                 if (substr($content_type, 0, 5) === "image") {
                     // Read image path, convert to base64 encoding
                     $expected_img_data = base64_encode(file_get_contents($expected_image));
 
-                    $img_name = $fileinfo->getBasename('.png');
+                    $img_name = pathinfo($fileinfo->getBasename(), PATHINFO_FILENAME);
```
MasterOdin: might be easier to do `$fileinfo->getBasename('.' . $fileinfo->getExtension())`? I dunno what best practice here for PHP is.

MasterOdin:
So we could just use `image/`, but that does mean that's there a definite possibility of trying to display HTML as the list of things that use the [image content-type](https://www.iana.org/assignments/media-types/media-types.xhtml#image) is far more than what [img tag supports](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Firefox). I think in an ideal world, we'd rather do filtering on `image/jpeg`, `image/png`, etc. and not do a check at all on file extension.

holzbh:
Wow, I totally missed the .twig having specified PNG. I'm not surprised that works since literally every image we've had from SIS is truly JPEG content despite the extension. It's part of why this issue took so long to get made!

site/app/views/grading/ImagesView.php
holzbh: From my cursory search it seemed like pathinfo() is the suggestion people tend to make, but internet snippets of code doesn't necessarily mean it's best practice. I can change this if you'd like.

MasterOdin:
I suspect it works because browsers ignore the `png` bit and read the actual content-type out of the base64 encoded string (much like how if you pointed it at an image file with the wrong extension, the browser will ignore the extension and read the file to determine the content-type). This is probably why things have worked for so long with the wrong extension.

Looking at this code, there's quite a few structural things wrong here that we should improve upon. If it's alright, I'm going to "take over" here and make the changes (as I had started doing it in my own branch earlier today to fix this issue before you opened this PR).

holzbh:
Sure, go ahead, you can take over. I didn't want to do a restructure in a PR that was supposed to be a quick fix, but the more I look at the code, the more I want to change...

MasterOdin:
Alright, so I made my changes which are:
1. we check that the mime type are image/{jpeg, gif, jpg, png} (and check the mime-type and not just extension like before)
2. properly set the content-type when displaying the image (e.g. `image/jpeg`, or `image/png`, etc.)
3. Don't store image data in a numeric array and user ids in another array that we have to loop through to find the necessary student, instead use an associated array and just do a key look-up (good efficiency pickup)

holzbh:
After updating to the latest version, I noticed the following my console:

>ReferenceError: createArray is not defined[Learn More] index.php:979:9
>    <anonymous> http://192.168.56.111/index.php?semester=s19&course=development&component=grading&page=images&action=view_images_page:979

Futhermore, I can't upload images. I can click "Upload Student Photos" and the modal pops up, but clicking "Drag files or click to open file browser" does nothing. Probably related.

bmcutler:
Still does not handle very large files well.  Opened #3724

PR #3719, opened 2019-05-22
Add image alt text in config

### LINKED ISSUE #3717: Allow for image 'alt' text configuration inside config.json file
Need to allow instructor setting up gradeable config.json file to specify 'alt' text for any images they display.  (Needed for VPAT).

Kevin-MM:
Closes #3717

Instructors have no ability to configure 'alt' text which describes what an image is for use with accessibility software.

### What is the new behavior?
Instructors can now use the 'alt' field when configuring image objects inside the config.json file which will be passed through to image's html 'alt' attribute.


site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -187,15 +187,23 @@
...
                             width="{{ image.width }}"
                         {% endif %}
+
                         {% if image.height > 0 %}
                             height="{{ image.height }}"
                         {% endif %}
+
+                        {% if image.alt is defined %}
+                            alt="{{ image.alt }}"
+                        {% endif %}
```
MasterOdin: I think I'd prefer not having the foreach to apply the alt attribute below and just do:
```
alt="{{ image.alt | default('Instructor Provided Image') }}"
```
here.

bmcutler: I believe this code will be refactored as part of #3728 (we won't have multiple images).
So I'm going to merge this.  But I'll have Kevin look at your feedback.

PR #3721, opened 2019-05-22
Prepopulate boxes

### LINKED ISSUE #3696: Pre-populate Text/Codeboxes with Instructor Data
Allow instructors to specify starter code or text for short_answer boxes and codeboxes.

Kevin-MM:
Closes #3696

Previously, instructors had no ability to prepopulate short_answer boxes.

### What is the new behavior?
Instructors can now specify default text that will be displayed in short_answer boxes when they load.  This is done by including the 'starter_value_string' field inside the input object configuration in the config.json file.  Further, a Reset buttton has been added below each short_answer box which will reset the short_answer box to its original instructor supplied value when clicked.

### Other information?
Tested inside Development -> Paragraph Textboxes gradable.


grading/main_configure.cpp
```diff
@@ -210,26 +210,37 @@ int main(int argc, char *argv[]) {
...
+
+          // starter_value_string, optional
+          // Create a empty string
+          std::string starter_value_string = "";
+
+          // If field inside json was not empty then assign to the new string
+          auto in_starter_value_string = input.find("starter_value_string");
+          if(in_starter_value_string != input.end())
+          {
+            starter_value_string = *in_starter_value_string;
```
emaicus: We should check first here to make certain that ```in_starter_value_string``` is in fact a string, and throw an appropriate, user-readable error if it is not.

bmcutler:
please resolve conflicts

PR #3725, opened 2019-05-23
[Bugfix] Fix upload for student images and efficiency improvements

MasterOdin:

From #3712, the change to not buffer things meant that drag-and-drop.js was not being properly loaded for the page, which caused the upload to break.

### What is the new behavior?
drag-and-drop.js is properly imported. I've also added a new Twig function to allow us to in-place base64_encode image paths so that we don't have to construct a potentially huge array in the View before going to Twig to display the images. Lowers the amount of memory PHP should use per process slightly, though does not appear to fix the underlying problem of potentially to show huge images.


bmcutler:
@holzbh confirmed the bug is fixed.

Issue #3723, opened 2019-05-22
More QR bulk upload improvements


shailpatels:
- [x] Split by QR code has an option page count or page range. And when a split pdf length is not in range, the file is made more obvious so it can be reviewed. For example... I could specify 8-10 pages (8 pages of problems plus a blank page that some students will tear off and we won't scan). So .pdfs that are only 6 pages and pdfs that are 20 pages will appear like a bad username does in red (maybe highlight these in yellow or orange), so it can be fixed if necessary before submitting.

- [x] In addition to saving the student pdf, also save the individual pages of the pdf as .png (all in the submission folder) - so we can use either the pdf or the images in the TA grader UI

- [x] Refactor CGI script to use the submitty job queue

- [x] Handle weird PDFS better   (if the first page does not have a QR code but other pages do -- just make the pages before the first QR code a separate pdf)

- [x] Log status of whenever a bulk upload /var/local/submitty/logs/ handle whenever something goes wrong and also what the user what doing

- [ ] Add CI tests for submission (bulk/normal/version)
      EDIT: Move to new issue #4108

- [x] Assign user Ids as they're processed instead of waiting for the entire file

- [x] Only PING QR process for instructors (previous submission may trigger the polling?) (investigate further - may not actually be happening) EDIT: Unable to reproduce. Confirmed it only loads for instructors.

- [x] Create Json file for normal bulk uploads and always show the page count after they get split to be consistent.

PR #3739, opened 2019-05-26
Highlight page counts that are different than the expected amount

shailpatels:
Instructors can now enter the expected number of pages each split PDF to be after a bulk QR upload,
any page count that is different then the expected (if its given and positive) will be highlighted 

The Submit All button will skip these if the expected pages is given and valid. The expected page count is saved per session.

Interface to add expected count

c3

Examples of highlighting
c1

After changing it to 1

c2

See issue #3723


bmcutler:
Hi Shail,
One small request, can you please only display the box labeled
"Bulk upload processing progress"
when the "bulk update" radio button option at the top of the page is checked?
(I couldn't quickly figure out the syntax for the if statement to make that happen).
Thanks.
(Also double check the UI changes I made -- I think they look a bit cleaner now)

shailpatels:
@

> Hi Shail,
> One small request, can you please only display the box labeled
> "Bulk upload processing progress"
> when the "bulk update" radio button option at the top of the page is checked?
> (I couldn't quickly figure out the syntax for the if statement to make that happen).
> Thanks.
> (Also double check the UI changes I made -- I think they look a bit cleaner now)

@bmcutler fixed, now the progress box only shows when the bulk upload radio button is checked
as for the UI yes I agree it does look cleaner split over multiple lines. Its probably easier to use left-justified too although it kinda feels unbalanced now but that's a discussion for another day/someone else can make it look pretty.

Issue #3740, opened 2019-05-27
Router replacement to cater for the needs of API


zjxiaohan:
**PROGRESS:**

- [x] AdminGradeableController
- [x] ConfigurationController
- [x] EmailRoomSeatingController
- [x] GradeableController
- [x] LateController
- [x] PlagiarismController
- [x] ReportController
- [x] UsersController
- [x] WrapperController
- [x] CourseMaterialsController
- [x] ForumController
- [x] ElectronicGraderController
- [x] ImagesController
- [x] SimpleGraderController
- [x] PDFController
- [x] LateDaysTableController
- [x] RainbowGradesController
- [x] SubmissionController
- [x] TeamController
- [x] AuthenticationController
- [x] HomePageController
- [x] MiscController
- [x] NavigationController
- [x] NotificationController

**What problem are you trying to solve with Submitty**
The router we are using is unable to handle complex requests that API needs. For example, a POST method with an endpoint of `/api/{semester_id}/{course_id}/assignments/{assignment_id}/build` cannot be simply parsed using our current router, which does not support regular expressions.

**Describe the way you'd like to solve this problem**
I plan to use https://github.com/klein/klein.php to replace the router we are using.

**Describe any potential alternatives you'd tried to solve the problem**
Alternatives include:
- Write a router myself. I do not think it is realistic. Although simple requests like `GET /api/users/?id=123` or `POST /api/users` can indeed be processed easily, it is not easy to handle complex requests. Worse, PUT and DELETE can be tricky.
- Use a framework for API such as Lumen. Discussed in #300 but decided not to because of the development resources.
- Use other routers like https://github.com/nikic/FastRoute. I compared several ones, and found Klein.php to be relatively more friendly.

MasterOdin:
Some points:
1. If you're going to use an off the shelf router, we'd probably just go with Lumen or more preferably symfony (as Laravel does too much magic in my opinion) . Additionally, Klein does not look to be supported at all anymore.
1. "which does not support regular expressions" - while true, this isn't as big a concern as you make it to be. The current router would be done by just popping off arguments, done at the various levels of the controller (e.g. index.php pops off the argument for `$_REQUEST['controller']`, the controller then pops off the argument for `$_REQUEST['page']` and so on and so forth, cumbersome, but does allow for backwards compatibility with the current code/web addresses as we move off the current ugly URLs to something a router can do.

[...]

PR #3743, opened 2019-05-28
[Feature] Symfony Routing

zjxiaohan:

Partially addresses #3740 .

### What is the new behavior?
Use symfony/routing for routing.

### Other information?

After looking into the router we already have, I feel that it would be quite messy to use it and "pop" each time for every parameter, especially for long URLs. Besides, API calls do not really fit into the "component"-"page"-"action" fashion. Therefore, I would suggest using Symfony's routing package.

The code in this PR depicts what I think the router would be like. Please give me some suggestions.


zjxiaohan:
@MasterOdin If you feel okay with what the router looks like currently, I would start working on the "main" router, which, compared with the ApiRouter, should have access control like we are doing in `index.php`.

site/public/index.php
```diff
@@ -65,10 +65,14 @@ function error_handler() {
 
 $semester = '';
 $course = '';
+$is_api = False;
 
 if ($core->getRouter()->hasNext()) {
     $first = $core->getRouter()->getNext();
-    if (in_array($first, ['authentication', 'home'])) {
+    if ($first === 'api') {
```
prestoncarman: Can api just be one of the components in the switch statement below?

site/app/controllers/api/TokenController.php
```diff
@@ -0,0 +1,18 @@
...
+
+
+class TokenController extends AbstractApiController {
+
+    /**
+     *
+     * @Route("/token", methods={"GET"})
+     */
+    public function getToken() {
+        $this->core->getOutput()->renderJsonSuccess("test");
```
prestoncarman: I think is just setting up the basic router for the api, but how would you write a test?

site/public/index.php
zjxiaohan: Yes, it can be one of the components. More modifications need to be added though because there are several operations like putting it into $_REQUEST['old']. But yes, we can.

In the meeting today we decided to replace the whole router with Symfony router. Hence it's likely that we are going to have an overhaul of index.php soon.

zjxiaohan:
@MasterOdin 

I am trying to gradually replace the original router with the Symfony one.

What I am doing is to use the new router to deal with things first, and if it fails (throws an exception), we will continue to use the current router.

This enables us to merge the PRs piece by piece, instead of having a huge PR at the end and conflicts with everything. When the new router becomes full-fledged, we could remove duplicate things in the index.php, replacing `buildurl` function, as well as the `run` function in controllers.

For this PR, I have added the following routes:
- /authentication/logout
- /authentication/login
- /home and /home/homepage
- /{semester}/{course} and /{semester}/{course}/navigation

Please give some suggestions.

site/app/libraries/MainRouter.php
```diff
@@ -0,0 +1,141 @@
...
+        $methodName = $this->parameters['_method'];
+
+        if (in_array('semester', $this->parameters) && in_array('course', $this->parameters)) {
+            $semester = $this->parameters['semester'];
+            $course = $this->parameters['course'];
+            $this->core->loadConfig($semester, $course);
+        }
+
+        $controller = new $controllerName($this->core);
+        return $controller->$methodName();
```
MasterOdin: how could we pass in the potential request args for a given path to the function?

e.g. If we have `/submission/{gradeable_id}`, how would we pass that `gradeable_id` to the function: `$controller->$methodName($gradeable_id);`

site/app/libraries/MainRouter.php
```diff
@@ -0,0 +1,141 @@
...
+            }
+        }
+    }
+
+    private function endsWith($string, $ending) {
+        $str_len = strlen($string);
+        $ending_len = strlen($ending);
+        if ($ending_len > $str_len) return false;
+        return substr_compare($string, $ending, $str_len - $ending_len, $ending_len) === 0;
+    }
```
MasterOdin: Can you not just use [Utils::endsWith](https://github.com/Submitty/Submitty/blob/master/site/app/libraries/Utils.php#L127-L129) for this?

site/app/libraries/ApiRouter.php
```diff
@@ -0,0 +1,53 @@
+<?php
```
MasterOdin: let's make a new folder `site/app/libraries/routers` and move all three routers under this folder (rename Router.php to ClassicRouter.php)

site/app/libraries/MainRouter.php
zjxiaohan: Sorry, I did not know there is a `Utils::endsWith`. I have made the changes.

site/app/libraries/MainRouter.php
zjxiaohan: Thank you very much for reminding me of this. I did not notice this because `$course` and `$semester` are quite special that they are processed before calling the navigation controller. I have fixed this by using `call_user_func_array` and removing the parameters starting with `_`

zjxiaohan: This screenshot shows the expected behavior.
Screenshot from 2019-06-04 13-53-06

zjxiaohan:
> Could we add custom properties to the Route annotation method (e.g. to designate a function as front-end only or api only)

Are we going to put API methods into another set of controllers? If so, then it seems not necessary to designate a function as front-end only or API only since they are separately stored. APIs will be in `http://192.168.56.111/api`, marked by a prefix `api`.

If they are on the same file, yes, we can add a `condition` property in the annotation part like `condition: "request.headers.get('User-Agent') matches '%app.allowed_browsers%'"` (which has to be true to access the controller) and change the requests accordingly.

zjxiaohan:
> What does the Symfony\Component\HttpFoundation\Response object look like and what does it do to output HTML vs JSON (e.g. I think we'll probably want to have our own Response object that we can set a view, and also data to pass to that view. If it's front end, call the view and pass the data to it. If it's the API, just output the data as a JSON object (setting appropriate headers and all that).

I am not using `Symfony\Component\HttpFoundation\Response`, I think `app/libraries/Output.php` is sufficient. We could use `renderOutput` to pass the data to the view, and use `renderJsonSuccess`, `renderJsonFail` and `renderJsonError` to return the data.

site/app/libraries/MainRouter.php
zjxiaohan: I changed the route to `/{_semester}/{_course}` in the latest commit considering that `$semester` and `$course` are hardly used in a specific function. (Course-related data are loaded before entering a `controller::method`.)

The most recent commit allows `http://192.168.56.111/s19/sample/student/grading_homework`.

The function looks like:
```
/**
 * @Route("/{_semester}/{_course}/student/{gradeable_id}")
 * @return array
 */
public function showHomeworkPage($gradeable_id = null) 
```

I would use the following code to ensure backward compatibility before the new router is fully set up.

```
if (!isset($gradeable_id)) {
    $gradeable_id = (isset($_REQUEST['gradeable_id'])) ? $_REQUEST['gradeable_id'] : null;
}
```

MasterOdin:
Alright, for the first draft, I'm satisfied with staying with what we've got with output, but so encourage you to look into PSR-7 and give some thought to how we might look to better follow it in the future.

site/app/libraries/routers/MainRouter.php
```diff
@@ -0,0 +1,142 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\RequestContext;
+use Symfony\Component\Routing\Matcher\UrlMatcher;
+use Symfony\Component\Routing\Loader\AnnotationDirectoryLoader;
+use Doctrine\Common\Annotations\AnnotationReader;
+use app\libraries\Utils;
+use app\libraries\Core;
+
+
+class MainRouter {
```
MasterOdin: Let's call this the `WebRouter`

site/app/controllers/HomePageController.php
```diff
@@ -72,10 +69,14 @@ public function changeUserName(){
...
             }
         }
+        $this->core->redirect($this->core->buildNewUrl(['home']));
     }
 
     /**
      * Display the HomePageView to the student.
+     *
+     * @Route("/home")
+     * @Route("/home/home_page")
```
MasterOdin: I think we can probably just use only use `/home` here.

zjxiaohan:
When is this PR planned to be merged?

zjxiaohan:
While as @MasterOdin said we do not need to be perfect at the beginning of the router, I still feel important to merge https://github.com/Submitty/Submitty/pull/3792 here because, when I was writing unit tests, I found several critical mistakes that need to be fixed, including:

- used `in_array` when I actually meant `array_key_exists`.
- `no_access` is for courses, therefore I should redirect users to `$this->parameters['_semester'] . '/' . $this->parameters['_course'] . '/no_access'` instead of simply `navigation/no_access`.

This PR is ready for review.

zjxiaohan:
Several changes since last time:
- Started to gradually kill `ClassicRouter`. Instead of throwing an exception and let ClassicRouter do most of the work, in the most recent code, `WebRouter` is taking full charge of several routes.
- Made `WebRouter` capable of redirecting to the original url after a successful login.
- Modified tests to let it suit `WebRouter`, including removing `access` e2e test and putting the logic in `WebRouterTester`.

zjxiaohan:
If tests are passed like they are in my localhost, then this PR is ready for review. @MasterOdin .

site/tests/BaseUnitTest.php
```diff
@@ -9,8 +9,19 @@
...
+
 
 class BaseUnitTest extends \PHPUnit\Framework\TestCase {
+
+    /**
+     * Loads annotations for routers.
+     */
+    public static function setUpBeforeClass(): void {
+        $loader = require(__DIR__.'/../vendor/autoload.php');
+        AnnotationRegistry::registerLoader([$loader, 'loadClass']);
```
MasterOdin: Is this really needed for all unit tests or is it just needed for the WebRouter tester?

site/app/controllers/student/SubmissionController.php
```diff
@@ -269,8 +270,14 @@ private function changeRequestStatus() {
         }
     }
 
-    private function showHomeworkPage() {
-        $gradeable_id = (isset($_REQUEST['gradeable_id'])) ? $_REQUEST['gradeable_id'] : null;
+    /**
+     * @Route("/{_semester}/{_course}/student/{gradeable_id}")
```
MasterOdin: should this also include `@Route('/{_semester}/{_course}/student/{gradeable_id}/{gradeable_version}` with gradeable_version being optional (and would default to latest if not there).

why is gradeable_id optional here?

site/tests/BaseUnitTest.php
zjxiaohan: It's only needed for WebRouter. Will move it.

site/app/controllers/student/SubmissionController.php
zjxiaohan: Oh. At first I thought it would be backward compatible - Classic Router doesn't pass args to functions and we need a default value. Now it seems useless.

zjxiaohan:
> Scrolling through, there's a handful of places where you've got inconsistent spacing, please fix.

I am not sure what it means.. Is it a space/TAB issue? Could you provide an example?

site/app/views/forum/ForumThreadView.php
```diff
@@ -110,7 +110,10 @@ public function searchResult($threads){
 	*/
 	public function showForumThreads($user, $posts, $unviewed_posts, $threadsHead, $show_deleted, $show_merged_thread, $display_option, $max_thread, $initialPageNumber) {
 		if(!$this->forumAccess()){
-			$this->core->redirect($this->core->buildUrl(array('component' => 'navigation')));
+			$this->core->redirect($this->core->buildNewUrl([
+			    $this->core->getConfig()->getSemester(),
+                $this->core->getConfig()->getCourse()
+            ]));
```
MasterOdin: spacing

site/app/views/forum/ForumThreadView.php
```diff
@@ -930,10 +933,13 @@ public function createPost($thread_id, $post, $unviewed_posts, $function_date, $
...
 
-        if(!$this->forumAccess()){
-            $this->core->redirect($this->core->buildUrl(array('component' => 'navigation')));
-            return;
-        }
+		if(!$this->forumAccess()){
+            $this->core->redirect($this->core->buildNewUrl([
+                $this->core->getConfig()->getSemester(),
+                $this->core->getConfig()->getCourse()
+            ]));
```
MasterOdin: spacing

site/app/views/forum/ForumThreadView.php
```diff
@@ -982,7 +988,10 @@ public function createThread($category_colors){
 	public function statPage($users) {
 
 		if(!$this->forumAccess()){
-			$this->core->redirect($this->core->buildUrl(array('component' => 'navigation')));
+            $this->core->redirect($this->core->buildNewUrl([
+                $this->core->getConfig()->getSemester(),
+                $this->core->getConfig()->getCourse()
+            ]));
```
MasterOdin: spacing

zjxiaohan: I thought there should be an indentation inside the function... it's like 

$this->core->redirect($this->core->buildNewUrl(*some_array*));

How would you do the spacing here?

zjxiaohan: I made them one-line if it is preferred

site/app/controllers/student/SubmissionController.php
```diff
@@ -104,7 +104,11 @@ public function run() {
                 break;
             case 'display':
             default:
-                return $this->showHomeworkPage();
+                
+                return $this->showHomeworkPage(
```
zjxiaohan: Thanks a lot!

PR #3747, opened 2019-05-28
Use Codemirror for Notebook Codebox Inputs

emaicus:
Updates the notebook ```codebox``` input type to be formatted with CodeMirror. Input is properly retrieved and stored.

Know Issues: 
1.  CodeMirror captures ```Tab``` and ```Shift + Tab``` keystrokes, posing an accessibility issue. Further, per the CodeMirror developer board, CodeMirror reports an empty text area to screenreaders.

Also addresses part of #3753


bmcutler:
auto-selecting the radiobutton from the last submission does not seem to be working for me.

PR #3759, opened 2019-05-29
Show server time and update dynamically

### LINKED ISSUE #1865: Add "current time on server" dynamic clock to submission page

xprilion:

Fixes #1865 

### What is the new behavior?
The Current server time is displayed only when 1 day or less is remaining from the due date. It keeps updating every minute and the maximum error for the time is 1 minute. 

### Other information?
Is this a breaking change? - No
How did you test - 
[1] Test Case 1: Laptop in ON state, tab active 
Let the page stay for a while in tab active state, the time kept updating correctly. 

[2] Test Case 2: Laptop in ON state, tab inactive
Let the page stay while on another tab. Returned to find the correct time, and updates too came in correctly.

[3] Test Case 3: Laptop put on sleep
Opened page, checked one correct update. 
Put laptop to sleep for 10 minutes.
Returned to find correct time displayed
Next update at correct time.

(I used [ClockTab](http://www.clocktab.com) to check if updates where happening at 0 seconds on the clock).

### DIFF
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -18,6 +18,7 @@
         {% endif %}
     </div>
 
+    {%  if due_in_a_day %}
     <script>
         $(function(){
 
@@ -33,7 +34,6 @@
             var fetchTimeUrl = window.location.href + "&action=fetch_time";
 
             setTimeout(function(){
-                console.log("Ran");
                 $.ajax({
                     url: fetchTimeUrl,
                     dataType: "json",
@@ -59,6 +59,7 @@
             }
         });
     </script>
+    {% endif %}
 
     {% if show_no_late_submission_warning %}
         <div>
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -10,9 +10,55 @@
     <div class="upperinfo">
         <h1 class="upperinfo-left">New submission for: {{ gradeable_name }}</h1>
         {% if has_due_date %}
-            <h3 class="upperinfo-right">Due: {{ formatted_due_date }}</h3>
+            <h3 class="upperinfo-right">Due: {{ formatted_due_date }} <br>
+                {%  if due_in_a_day %}
+                    <span>Now: <span id="current-server-time">{{ formatted_current_time }}</span></span>
+                {% endif %}
+            </h3>
         {% endif %}
     </div>
+
+    <script>
+        $(function(){
+
+            // Get browser's time, timezone offset and unix timestamp (all in milliseconds)
+            var localDate = new Date();
+            var localTimestamp = localDate.getTime();
+            var localOffset = localDate.getTimezoneOffset() * 60000;
+
+            // Calculate shift to UTC in milliseconds
+            var UTCTime = localTimestamp + localOffset;
+            var serverTimeShift;
+
+            var fetchTimeUrl = window.location.href + "&action=fetch_time";
+
+            setTimeout(function(){
+                console.log("Ran");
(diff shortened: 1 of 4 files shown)


site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -17,10 +17,56 @@
     <div class="upperinfo">
         <h1 class="upperinfo-left">New submission for: {{ gradeable_name }}</h1>
         {% if has_due_date %}
-            <h3 class="upperinfo-right">Due: {{ formatted_due_date }}</h3>
+            <h3 class="upperinfo-right">Due: {{ formatted_due_date }} <br>
+                {%  if due_in_a_day %}
```
zjxiaohan: If the server time is only shown when the gradeable is due in a day, then it seems unnecessary to fetch server time continually when it is not displayed. I would suggest wrapping the script below in an `if` block like this one does.

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -17,10 +17,56 @@
...
+            var localOffset = localDate.getTimezoneOffset() * 60000;
+
+            // Calculate shift to UTC in milliseconds
+            var UTCTime = localTimestamp + localOffset;
+            var serverTimeShift;
+
+            var fetchTimeUrl = window.location.href + "&action=fetch_time";
+
+            setTimeout(function(){
+                console.log("Ran");
```
zjxiaohan: This `console.log` does not make sense to me. How about replacing it with a comment if something is needed here?

site/app/templates/submission/homework/SubmitBox.twig
xprilion: Makes sense, thanks @zjxiaohan! Will make that change!

site/app/templates/submission/homework/SubmitBox.twig
xprilion: Okay, that was there only for me to check if it was running properly at the correct time :D There were other logs too, overlooked this one while removing I guess :p

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -10,10 +10,57 @@
...
+                    dataType: "json",
+                    success: function(result){
+                        // timezone offset returned in seconds from server
+                        var serverOffset = result.data.offset * 1000;
+
+                        // server time shift is shift from UTC and the seconds before the ajax call was made is added to it.
+                        serverTimeShift = UTCTime + serverOffset + ((60 - localDate.getSeconds())*1000);
+
+                        setTime(); // Since setInterval runs after the interval and we need an update right now, run the function once
+                        setInterval(function(){setTime()}, 60 * 1000);
```
shailpatels: What is the point of starting another setInterval loop here? Wouldn't every asynchronous request create its own loop and you're already launching multiple ajax calls since all of that is within its own loop? 

Also this PR looks similar to #3508 where you mentioned theres a better way of doing this? I'm confused on if you think theres a better way to implement this.

xprilion: @shailpatels there are no multiple ajax calls here, it's just a single fetch at the first 00 seconds on the clock after the page loads, henceforth it is always just calculation of server time by shifting the local time to it, which is being taken from the browser. 

This PR is significantly different from the #3508 because this is the better way about which I talked there. The maximum error there was 5 minutes, here it is 1 minute. 

The function inside the setInterval is the only block being called repeatedly, and not in an asynchronous fashion.

xprilion: The setInterval is not being run more than once!

bmcutler:
Closing for now, please re-open the PR when ready for re-review

PR #3761, opened 2019-05-30
Database (master and course) — new column: users.user_numeric_id 

### LINKED ISSUE #3465: Include Student ID number in Student Data
Allow for an optional field to be included in student data that is a numeric-only student ID (e.g. 9-digit RPI RIN)

pbailie:
Close #3465 

* Add a new column `user_numeric_id` to master and course DB schema.
    * New column is `character varying` datatype which will allow delimiter characters (such as a dash) to be included.
    * Index created in course DB for this column as I expect this column may experience a heavy amount of `select [columns] where user_numeric_id=[value]` queries (or part of a larger query) for exam grading.  Especially in large courses like CS1 and DS.
* Adjusted trigger functions so INSERT and UPDATE queries propagate data from master to course DB.
* Adjusted master schema sql files and created new up and  down migrations.


bmcutler:
Thanks Peter!
We should also add the numeric_id field to the New Student/New Grader/Edit User popup GUI.
Is it simple to also add that in this PR?

migration/migrator/migrations/course/20190528164528_numeric_id.py
```diff
@@ -0,0 +1,34 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("ALTER TABLE IF EXISTS users ADD COLUMN user_numeric_id varchar;")
```
zjxiaohan: As [the document on migrations](https://submitty.org/developer/migrations) suggests:

> it is important to ensure that the up migration can be re-run. For example, your up function should not crash on adding the column if the column already exists.

It can be good to change `ADD COLUMN` to `ADD COLUMN IF NOT EXISTS` and `CREATE INDEX` to `CREATE INDEX IF NOT EXISTS`, or adding an `if` to determine if there is already a `user_numeric_id` column.

migration/migrator/migrations/course/20190528164528_numeric_id.py
```diff
@@ -0,0 +1,34 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("DROP INDEX IF EXSISTS users_user_numeric_id_idx;")
+    database.execute("ALTER TABLE IF EXISTS users DROP COLUMN IF EXISTS user_numeric_id;")
```
zjxiaohan: According to the [doc](https://submitty.org/developer/migrations):

> In general, your down migration should only undo the necessary changes to make earlier versions of the Submitty software work. For example, if your migration is adding a column to an existing table in the database, it is probably not necessary or desirable to delete that data in the down function. We don’t want to lose any data since we likely plan to return to this state in the future.

It would be sufficient to `DROP INDEX` only.

migration/migrator/migrations/course/20190528164528_numeric_id.py
pbailie: It's good to bring this up.  However, this also brings up other issues in my mind...
* Presumably, this migration would only crash when the column already exists, but the migration ran anyway.  That implies there is no migration record to disallow another migration, so something else outside of this migration went awry.
* The migration script that is calling these functions probably should be doing exception handling on migration functions.  That would be far more efficient than having every migration function do its own exception handling.  That is not to say that the migrator should ignore bad DB queries.  But the migrator can note a bad query and cleanly post an error message to console and/or log.  And then decide to proceed to the next migration, or not.

But these cases are out of scope for this PR.  I just felt they were worth mentioning.

migration/migrator/migrations/course/20190528164528_numeric_id.py
pbailie: The column is the real essential portion, but the index has its importance too.  It strikes me as pointless to rollback the index, but never the column.

Just as a side note — the kind of "data loss" on rolling back / dropping the column (in this migration) would be non-essential to any given course.

pbailie:
> Thanks Peter!
> We should also add the numeric_id field to the New Student/New Grader/Edit User popup GUI.
> Is it simple to also add that in this PR?

I'll have a look at the twig files.

migration/migrator/migrations/course/20190528164528_numeric_id.py
pbailie: Change made.  Parallel change also made to master DB migration.

migration/migrator/migrations/course/20190528164528_numeric_id.py
pbailie: Change made.  Parallel change also made to master DB migration (or else there would be data sync'ing problems to course DBs).

pbailie:
> > Thanks Peter!
> > We should also add the numeric_id field to the New Student/New Grader/Edit User popup GUI.
> > Is it simple to also add that in this PR?
> 
> I'll have a look at the twig files.

I think that should do it.  I'll revisit this next week if something still isn't right.

New changes involve the New/Edit user form UI (`twig` and `js` files), User model, User Controller, Database Queries, and Travis Test for the User Model.

PR #3766, opened 2019-05-31
Add Timeouts to Worker Machine Interactions

### LINKED ISSUE #3684: Avoid slow connection timeouts when worker is offline
If a worker is offline, the INSTALL_SUBMITTY script and the restart_shippers_and_workers scripts seem to take a long time.  They wait for a timeout?

This is true even if the worker is in disabled mode in the autograding_workers.json file.

Let's do a faster check, and if the the worker is disabled, skip it altogether?

emaicus:
Closes #3684 

Adds reasonable timeouts to all paramiko worker machine interactions made by Submitty.

Not tested locally.

### DIFF
--- a/sbin/shipper_utils/systemctl_wrapper.py
+++ b/sbin/shipper_utils/systemctl_wrapper.py
@@ -86,12 +86,12 @@ def perform_systemctl_command_on_worker(daemon, mode, target):
       ssh = paramiko.SSHClient()
       ssh.get_host_keys()
       ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-      ssh.connect(hostname = host, username = user)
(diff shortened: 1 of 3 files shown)


MasterOdin:
Something to consider is that if a machine is timing out (especially during installation), do we want to automatically disable it until the sysadmin directly intervenes and fixes the problem? We do run the risk of installation timing out, and then if it works during grading, causing potential grading errors as it uses old code.

bmcutler:
If it fails during shipper (worker?) then yes it is auto-disabled.

Yes, all of this is a good idea...  

But I don't think it's worth over-engineering this until we have vagranted the worker and can easily spin up networks of workers.  (& CI tests!)

Issue #3765, opened 2019-05-31
Excused absence drop down menu lists all students twice


bmcutler:
**Describe the bug**

From the excused absence extensions page, where you type the (partial) name/id of a student and get a drop down of all matching students:

All students (both regular & dropped) are listed twice -- as both a regular student, and a student who has dropped and is in the NULL section.  

Students should appear exactly once, and students in the null section should be labeled with the prefix string [NULL section]

Not sure when/how this broke, but it was working rather recently.

PR #3772, opened 2019-06-02
[Extensions] update absence drop down lists to show students only once

Fenn-CS:

As in #3765, All students (both regular & dropped) are listed twice -- as both a regular student and in the NULL section. 

Fixes: #3765
 
### What is the new behavior?
Show students exactly ones in the dropdown.

### Other information?

- I added a `removeStudentWithId()` function which I could use to achieve the same results but did not use it since it's not efficient to add null students in the array before removing them, so students are only added to their respective arrays no need for removal, however, I contemplating if the `removeStudentWithId()` can be useful in the Utils?

### Main change
_Moved  ` $students_full[] = $student_entry; ` into NOT NULL condition_
https://github.com/Submitty/Submitty/blob/263fcda4bc45e6076b9a15505ed83b759dae603d/site/app/libraries/Utils.php#L246-L256

_other changes are just formattting e.g `function(){}` => `function () {}`_

### DIFF
--- a/site/app/libraries/Utils.php
+++ b/site/app/libraries/Utils.php
@@ -240,32 +240,42 @@ class Utils {
      * students_version is an array of user and their highest submitted version
      */
 
-    public static function getAutoFillData($students, $students_version = null){
+    public static function getAutoFillData($students, $students_version = null) {
         $students_full = array();
         $null_section = array();
         foreach ($students as $student) {
             $student_entry = array('value' => $student->getId(),
                     'label' => $student->getDisplayedFirstName() . ' ' . $student->getDisplayedLastName() . ' <' . $student->getId() . '>');
-            $students_full[] = $student_entry;
-            if($students_version != null){
-                if($student->getRegistrationSection() != null && array_key_exists($student->getId(),$students_version)){
+            if($students_version != null) {
+                if($student->getRegistrationSection() != null && array_key_exists($student->getId(),$students_version)) {
                     if ($students_version[$student->getId()] !== 0) {
                         $student_entry['label'] .= ' (' .
                         $students_version[$student->getId()] . ' Prev Submission)';
                     }
                 }
-            }else{
+                $students_full[] = $student_entry;
+            } else {
                 $null_entry = array('value' => $student->getId(),
-                'label' => '[NULL section] ' . $student->getDisplayedFirstName() . ' ' . $student->getDisplayedLastName() . ' <' . $student->getId() . '>'); 
+                'label' => '[NULL section]' . $student->getDisplayedFirstName() . ' ' . $student->getDisplayedLastName() . ' <' . $student->getId() . '>'); 
 
                 $in_null_section = false;
                 foreach ($null_section as $null_student) {
                     if($null_student['value'] === $student->getId()) $in_null_section = true;
                 }
-                if(!$in_null_section) $null_section[] = $null_entry;
+                if(!$in_null_section) $null_section[] = $null_entry; 
+                //$students_full = self::removeStudentWithId($students_full, 'value', $student->getId());
             }
         }
         $students_full = array_unique(array_merge($students_full, $null_section), SORT_REGULAR);
         return json_encode($students_full);
     }
+
+    public static function removeStudentWithId($students, $key, $id) {
+        foreach($students as $subKey => $student) {
+             if($student[$key] === $id) {
+                  unset($students[$subKey]);
+             }
+        }
+        return $students;
+   }
 }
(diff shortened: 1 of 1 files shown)


site/app/libraries/Utils.php
```diff
@@ -240,32 +240,42 @@ public static function getCompareByReference() {
...
-                if($student->getRegistrationSection() != null && array_key_exists($student->getId(),$students_version)){
+            if($students_version != null) {
+                if($student->getRegistrationSection() != null && array_key_exists($student->getId(),$students_version)) {
                     if ($students_version[$student->getId()] !== 0) {
                         $student_entry['label'] .= ' (' .
                         $students_version[$student->getId()] . ' Prev Submission)';
                     }
                 }
-            }else{
+                $students_full[] = $student_entry;
```
zjxiaohan: Moving `$students_full[] = $student_entry;` to here does not solve the problem. It just makes `$students_full` empty. You could test by entering the id of a student with a registration section, and you will find he/she will be categorized as `[NULL section]` under your code.

The root of this problem is that, if `$students_version` is null, then the code jumps to line 258, marking everything as `[NULL section]` without asking if `$student->getRegistrationSection() != null`.

An ugly solution can be adding an `elseif` in line 257:

```
elseif ($student->getRegistrationSection() != null) {
        $students_full[] = $student_entry;
}
```

You could surely come up with a more elegant solution, probably by refactoring the `if-else` logic here.

site/app/libraries/Utils.php
```diff
@@ -240,32 +240,42 @@ public static function getCompareByReference() {
...
-                'label' => '[NULL section] ' . $student->getDisplayedFirstName() . ' ' . $student->getDisplayedLastName() . ' <' . $student->getId() . '>'); 
+                'label' => '[NULL section]' . $student->getDisplayedFirstName() . ' ' . $student->getDisplayedLastName() . ' <' . $student->getId() . '>'); 
 
                 $in_null_section = false;
                 foreach ($null_section as $null_student) {
                     if($null_student['value'] === $student->getId()) $in_null_section = true;
                 }
-                if(!$in_null_section) $null_section[] = $null_entry;
+                if(!$in_null_section) $null_section[] = $null_entry; 
+                //$students_full = self::removeStudentWithId($students_full, 'value', $student->getId());
```
zjxiaohan: Please remove the comments. Comments that are not descriptive are distracting and confusing.

site/app/libraries/Utils.php
```diff
@@ -240,32 +240,42 @@ public static function getCompareByReference() {
...
                         $student_entry['label'] .= ' (' .
                         $students_version[$student->getId()] . ' Prev Submission)';
                     }
                 }
-            }else{
+                $students_full[] = $student_entry;
+            } else {
                 $null_entry = array('value' => $student->getId(),
-                'label' => '[NULL section] ' . $student->getDisplayedFirstName() . ' ' . $student->getDisplayedLastName() . ' <' . $student->getId() . '>'); 
+                'label' => '[NULL section]' . $student->getDisplayedFirstName() . ' ' . $student->getDisplayedLastName() . ' <' . $student->getId() . '>');
```
zjxiaohan: I would say the original one with one space looks better to me. It separates `[NULL section]` and the first name, making it clearer.

site/app/libraries/Utils.php
```diff
@@ -240,32 +240,42 @@ public static function getCompareByReference() {
...
-                if(!$in_null_section) $null_section[] = $null_entry;
+                if(!$in_null_section) $null_section[] = $null_entry; 
+                //$students_full = self::removeStudentWithId($students_full, 'value', $student->getId());
             }
         }
         $students_full = array_unique(array_merge($students_full, $null_section), SORT_REGULAR);
         return json_encode($students_full);
     }
+
+    public static function removeStudentWithId($students, $key, $id) {
```
zjxiaohan: Please do not add code into the codebase if you are not going to use it. For reasons, [this](https://stackoverflow.com/questions/15699995/why-unused-code-should-be-deleted) could be a good reference.

Fenn-CS:
Thanks, @zjxiaohan knowing the significance of being in the NULL section was key. :), updating PR...

site/app/libraries/Utils.php
Fenn-CS: I ended up using the function, since adding the else-if is not regarded as an ideal solution.

Fenn-CS: I ended up using the function, since adding the else-if is not regarded as an ideal solution.

bmcutler:
@Fenn-CS, Before I merge, can you please address @zjxiaohan 's other comments?  Thanks!

site/app/libraries/Utils.php
Fenn-CS: Restored

site/app/libraries/Utils.php
Fenn-CS: Line put into use.

Fenn-CS:
@bmcutler, I did resolve them. I just did not mark them as resolved here, please let me know if there's anyone undone.

PR #3775, opened 2019-06-03
[Feature] Download Zips in Course Material Page

### LINKED ISSUE #3746: Download zip of a subfolder within course_materials
Currently if a student wants to download all of the files in a folder of the course materials, they have to click and download on each individual file.  Let's add a download link on each folder for the course materials that will allow zip & download of all files/folders under that folder with a single click.

zjxiaohan:

* [ ] Documentation has been updated/added if relevant (**Will do if the changes are approved**)

Closes #3746 .

### What is the new behavior?
In the course material page, instructors and students are able to download zips made of files in a directory.

### Other information?

Instructor downloading the `bin` folder:
Screenshot from 2019-06-03 20-44-10

Student downloading the `bin` folder:
Screenshot from 2019-06-03 20-51-12

### DIFF
--- a/site/app/controllers/MiscController.php
+++ b/site/app/controllers/MiscController.php
@@ -517,7 +517,7 @@ class MiscController extends AbstractController {
         $zip = new \ZipArchive();
         $zip->open($zip_name, \ZipArchive::CREATE | \ZipArchive::OVERWRITE);
 
-        if(!$this->core->getUser()->accessAdmin()) {
+        if(!$this->core->getUser()->accessGrading()) {
             // if the user is not the instructor
             // download all accessible files according to course_materials_file_data.json
             $file_data = $this->core->getConfig()->getCoursePath() . '/uploads/course_materials_file_data.json';
--- a/site/app/controllers/MiscController.php
+++ b/site/app/controllers/MiscController.php
@@ -4,6 +4,7 @@ namespace app\controllers;
 
 
 use app\libraries\FileUtils;
+use app\libraries\DateUtils;
 use app\libraries\Utils;
 use app\libraries\Core;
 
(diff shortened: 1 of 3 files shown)


zjxiaohan:
@shailpatels 

Hi. Thank you for your review! Sorry, I forgot to test folder names with space. Thank you very much for doing that.

> I think there is a missing check for individual files for students to download.

Sorry I cannot reproduce this bug. Did you download the files as a student?

> Theres also an issue where if I download a nested file with a space it does not download correctly on firefox.

Fixed by replacing the space with underlines in the recent commits.

shailpatels:
@zjxiaohan 
Yes I think the issue is nested directories. Here's an  example I set up as an instructor:
instructor

So the file "even better.txt" should not be shared with the student and it is not shown to the students view which is good, but if I download the entire folder "nested", "even better.txt" gets downloaded as well.

Student view on submitty:
studeview

However after downloading "nested"
stude

"even better.txt" should gets downloaded anyways

zjxiaohan:
I'll check again. Thank you very much.

zjxiaohan:
@shailpatels 

Hi. I finally realized why this happens.

I only checked if the file exists in the .json, which looks like this:

```
  '/var/local/submitty/courses/s19/sample/uploads/course_materials/nested/blah.txt' => 
    array (size=2)
      'checked' => string '1' (length=1)
      'release_datetime' => string '2019-06-08 03:53:18-0400' (length=24)
  '/var/local/submitty/courses/s19/sample/uploads/course_materials/nested/New folder/even better.txt' => 
    array (size=2)
      'checked' => string '0' (length=1)
      'release_datetime' => string '2019-06-08 03:57:27-0400' (length=24)
```

However, if the prof checked one file and then uncheck that, there will be an entry in the json.

This problem has been fixed.

Thank you so much for testing and coming up with this critical bug!

shailpatels:
@zjxiaohan Happy to review :)

PR #3783, opened 2019-06-04
Add Logging for Bulk Uploads

shailpatels:
### What is the new behavior?
This PR creates the folder /var/local/submitty/logs/bulk_uploads where log files for bulk uploads are created. Both QR and normal are logged 

Also adds try catch around creating the split_uploads folder which should hopefully prevent the race condition from killing the bulk upload process. - This was probably preventing multiple files uploaded at once.

Adds new logger file that locks the file for multiple processes to write to it. Matches format of other logs
The directory for logs is created on the INSTALL_SUBMITTY_HELPER.sh script
logger

see #3723


bmcutler:
Hi Shail,
Does this work with multiple jobs daemon working in parallel, writing to the same log file?  I'm worried about opening the file and holding it open for the period of time it takes to split and process the pdf.  Should the logging write be more isolated?  Perhaps a helper function that opens the file, writes a line, then closes the file?  And perhaps even it should lock the file/directory and wait a second if another job process has got the file locked & open for writing?

shailpatels:
> Hi Shail,
> Does this work with multiple jobs daemon working in parallel, writing to the same log file? I'm worried about opening the file and holding it open for the period of time it takes to split and process the pdf. Should the logging write be more isolated? Perhaps a helper function that opens the file, writes a line, then closes the file? And perhaps even it should lock the file/directory and wait a second if another job process has got the file locked & open for writing?

Every bulk upload job writes its own log file (and every file launches its own job) so they should all be independent of each other. But yes maybe logging should be handled more independently maybe just capturing the output of each job?

bmcutler:
Oh, I didn't even notice the filename.  Does filename include the full directory/course path?  If not it may not be unique, so the logs would clobber themselves.

But let's make it more similar to the other neighboring log directories.  A single logfile per day that all processes write to.  If each write to the logfile is relatively short and fits on a single line, we can prefix the print with the timestamp, the course/gradeable/filename information & perhaps process id, so if multiple bulk jobs are happening at the same time we can just grep the file for one course/gradeable to debug stuff.

shailpatels:
Each file is the timestamp + filename but not the full path so they may clobber each other although I think the file upload system guarantees them to be unique before it reaches this point. But I didn't realize other log files worked like this which I agree is better and would solve the issue of them getting clobbered anyways.
I don't think the entire log can fit on one line but I'd like to talk to you on the format of what gets logged tomorrow :)

PR #3784, opened 2019-06-04
[Bugfix] Only catch exceptions for JWT parsing for cookie session

### LINKED ISSUE #3713: Restrict caught exceptions when logging in
**Describe the bug**
When attempting to access a course without running the latest migration, the site booted me back to the login screen as an exception was thrown when running the `getUserById` function.

**Expected behavior**
Show the regular 'Site has thrown an exception' screen, and not the log-in screen. The exception handler for the authentication code should be restricted for what it'll catch, and not do a broad catch on \Exception.

**To Reproduce**
1. Create a SQL error in PostgresqlDatabaseQueries::getUserById()
2. Log-in to the main page of Submitty
3. Attempt to access course from list on right side
4. See login screen instead of error page

zjxiaohan:

Fixes #3713 .

### What is the new behavior?
1. Let the TokenManager throw AuthenticationException, which is more informative than the original RuntimeException.
2. Let `index.php` only catches AuthenticationException. If the AuthenticationException is caught, which means the token is invalid, clear the cookie. If there are other exceptions, display the default exception page.

### DIFF
--- a/site/app/libraries/TokenManager.php
+++ b/site/app/libraries/TokenManager.php
@@ -2,7 +2,6 @@
 
 namespace app\libraries;
 
-use app\exceptions\AuthenticationException;
 use Lcobucci\JWT\Builder;
 use Lcobucci\JWT\Parser;
 use Lcobucci\JWT\Token;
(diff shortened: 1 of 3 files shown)


MasterOdin:
Let's use an InvalidArgumentException which is probably more appropriate, and is also the exception that `parse()` can throw.

PR #3787, opened 2019-06-04
Testing new, nice colors for the Deleted and New forum threads/posts :)

### LINKED ISSUE #3628: Change color of background of deleted posts
> @bmcutler This looks okay but if you don't mind can you please let me change background to little > bit of light color as test is not readable as color contrast ratio `3.77:1`. If possible make background `C9C9C9`.
> @pbailie what are your views?

_Originally posted by @drumilpatel2000 in https://github.com/Submitty/Submitty/pull/3612#issuecomment-486901993_

Created this issue to discuss if we want to change the color to improve the contrast or some potential other change. We should come to a set conclusion on what we want to do before any PR gets made.

thef1stofhername:
I changed the New thread color to a more pleasant blue. I also changed the Deleted thread color to be less dark and more easily readable. It also stops the thread from darkening when selected. Here is what the old Deleted threads used to look like:
oldDeletedPosts
Here is what the old New threads used to look like:
oldNewPosts
Here is what they look like now:
newColors!

closes #3628

### DIFF
--- a/site/public/css/server.css
+++ b/site/public/css/server.css
@@ -1177,10 +1177,6 @@ fieldset[disabled] .btn-primary.focus {
   background-color: #fff;
 }
 
-.btn-push-down {
-    margin-top: 5px
-}
-
 .panel-heading > .dropdown .dropdown-toggle {
   color: inherit;
 }
@@ -2095,4 +2091,3 @@ end of styles used in the admin gradeable page
     border: 0px;
     padding: 0px;
 }
-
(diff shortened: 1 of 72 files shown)


site/public/css/server.css
```diff
@@ -740,12 +740,16 @@ table tr.table-header {
 
 /* TODO: MAKE THIS CASCADE */
 .deleted {
-    background-color: #A9A9A9;
-    color: #4a4a4a;
+    background-color: #dcdcdc;
+<<<<<<< HEAD
```
drumilpatel2000: Remove merge conflicts

Issue #3125, opened 2018-11-17
Late day banner showing different late days left


shailpatels:
A student submitted an assignment 1 day late and has 6 more late days which is correct, but then banner says if you submit today again you will only have 5 remaining. 

![sharedimage](https://user-images.githubusercontent.com/12129065/48666796-988a6200-ea96-11e8-9a61-abb79067b751.png)

bmcutler:
duplicate/related issues: #3184 #2455

PR #3788, opened 2019-06-04
replaced non explicit message on late day banner.

Fenn-CS:
The late day banner message gives students the impression that late-days are charged based on the number of submissions instead of the number of extra days after the due date. 
Fixes: #3125 

### What is the new behavior?
The new message makes it clear that once a student has been charged for a late day, the student can submit multiples that day without being charged again for that day.

### Other information?

Other minoring changes in wording to make the communication on the banner clearer.


Fenn-CS:
@shailpatels "You won't be charged for another submission today" is compact and better I think it's better to replace the last line with than actually deleting it. As you said it's a minor but necessary nitpick. :)

bmcutler:
As we discussed in the our meeting today, this is not solving the full problem.  Closing for now, please re-open when solution is more complete.

PR #3799, opened 2019-06-05
Fix tab / enter navigation on bulk upload labeling

### LINKED ISSUE #3791: tab / enter navigation on bulk upload labeling
A followup to PR #3789 

Revision request: For a team bulk upload, can we use 'enter' (or 'tab') to move to the next box (the next team member).  That is don't submit.  Except...  on the last box of the team, let's make enter submit for the team.

So, if I've specified the max numbers per team to be 4, there will be 4 boxes.  I type 'al', use downarrow to select "alyssa p hacker", press enter, advancing to the second box, type 'bit', downarrow to select "ben bitdiddle", press enter, advancing to the third box, if that's the whole team I press enter, going to the 4th box, and then (since I'm on the final box for that team) enter again will submit.

VPAT:  The page is mostly ok for keyboard navigation, but sometimes we lose focus on the button.  If it's a resubmit for a student, the popup for rename/overwrite comes up, but I can't press 'enter' to submit.  In fact, tab and enter navigate us on the page under the popup.  There might be other issues as well.  Once you think it's good, let's get @pbailie to look over the finished UI.

We can talk more at the meeting tomorrow if this is unclear or not quite possible

shailpatels:
When the dialog box for previous submissions appears the focus would still be behind the modal object.

Now the modal object radio boxes are focused on. The user can use tab to focus on the next radio buttons, then the close/submit buttons. Escaping the box will return focus to the next submission to assign. Pressing escape will now exit the modal container.

Pressing enter on a team submission will bring it to the next box unless its the very last in the row, then it'll attempt to submit the team gradeable.

fixes #3791

### DIFF
--- a/site/app/templates/submission/homework/BulkUploadBox.twig
+++ b/site/app/templates/submission/homework/BulkUploadBox.twig
@@ -54,13 +54,25 @@
                                     value =""
                                 {% endif %}
 
-                                onkeydown="uploadOnEnter( {{ loop.index }} );"
+                                {# we should only submit on enter for normal bulk uploads or on the last team upload box #}
+                                {% if not team_assignment %} 
+                                    onkeydown="uploadOnEnter( {{ loop.index }} );"
+                                {% else %}
+                                    onkeydown="focusNextOnEnter({{ loop.index }}, 0)"
+                                {% endif %}
                             />
                             {% if team_assignment %}
                                 {# (size - 1) because twig loops are range inclusive #}
                                 {% set outer_loop = loop %}
                                 {% for i in 1..(max_team_size - 1) %}
-                                    <input type="text" id="bulk_user_id_{{ outer_loop.index }}[{{ i }}]" value ="" onkeydown="uploadOnEnter( {{ outer_loop.index }} );" />
+                                    <input type="text" id="bulk_user_id_{{ outer_loop.index }}[{{ i }}]" value ="" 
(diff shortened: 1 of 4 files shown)


bmcutler:
I like the interaction for team uploads, but I don't think the radio buttons or cancel button are highlighted on the popup/modal when you tab between them.  It's unclear where the focus is except when the focus is on the submit button, then it gets the highlight.

pbailie:
I put my accessibility analysis in the issue thread by mistake.  Sorry about that.

https://github.com/Submitty/Submitty/issues/3791#issuecomment-500079572

shailpatels:
@pbailie Fixed WAVE errors and the structured html warnings. There is a contrast warning with the white on green button but this appears to be an issue across submitty. We're working on standardizing colors, I'll bring this up when we create a central location for colors. See #3812

@bmcutler Some of the h3 had to be bumped up to h2, they look a bit bigger but still fit and I don't think overflow. Do you mind looking over it and see if its visually fine (if not we could use CSS to force it to look smaller and still maintain the desired structure)

The font awesome icon to open pdfs can now be focused with the keyboard. I used the recommended method described by Font Awesome https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility

Tested on Debian 9 Firefox and Google Chrome. (I used the WAVE Firefox addon with default settings not sure if we use anything special)

Screenshot from 2019-06-10 14-41-25

PR #3805, opened 2019-06-06
 Reorganized TA grading for better VPAT and UX 

### LINKED ISSUE #3161: Tab interaction on numeric/text and lab pages
Let's make sure tab follows the VPAT accessibility conventions on both of these pages.
Tab should cycle through all of the buttons and input fields in logical order.  
(top of page buttons, each row, left to right within the row)
The arrow keys (hot keys) can probably stay as they are.

Also...  when our focus enters a cell, let's have the text be highlighted, so that typing replaces the text.  Currently it places the cursor in front of the contents of the cell so that if you type '3' in of  a numeric cell with the default value of '0' it becomes '30', which is annoying.

Addresses part of #3155

mktung:
Closes #3161

- Changed DOM order of elements
- Added tabindex to ```Display.twig```
- Slimmed ```simple-grading.js```
- Added options parameter to ```ta-grading-keymap.js```

### DIFF
--- a/site/app/templates/grading/simple/Display.twig
+++ b/site/app/templates/grading/simple/Display.twig
@@ -1,17 +1,6 @@
 {% import _self as self %}
 
 <style>
-    .visuallyhidden {
-        border: 0;
-        clip: rect(0 0 0 0);
-        height: 1px;
-        margin: -1px;
-        overflow: hidden;
-        padding: 0;
-        position: absolute;
-        width: 1px;
-    }
-
     .column-wrapper {
         display:flex; 
         justify-content: space-between;
--- a/site/app/templates/grading/simple/Display.twig
+++ b/site/app/templates/grading/simple/Display.twig
@@ -1,26 +1,6 @@
 {% import _self as self %}
 
-<style>
-    .column-wrapper {
-        display:flex; 
-        justify-content: space-between;
-        align-items: flex-end;
-    }
-
-    .row-wrapper {
-        display:flex; 
-        flex-direction:column;
-    }
-    
-    #data-table td:focus, #data-table input:focus {
-        outline: 3px dashed #337ab7;
-    }
-
-    .fixed-top {
-        position: fixed;
-        top: 0;
-	}
-</style>
+<link rel="stylesheet" type="text/css" href="{{ base_url }}css/simple-grading.css" />
 
 <div class="content">
     <h1>
--- a/site/app/templates/grading/simple/Display.twig
+++ b/site/app/templates/grading/simple/Display.twig
@@ -103,17 +103,17 @@
                 <th width="20"></th>
                 <th width="70">Section</th>
                 <th width="90" style="text-align: left">
-                    <a href="{{ core.buildUrl({'component': 'grading', 'page': 'simple', 'action': action, 'g_id': gradeable.getId(), 'sort': 'id', 'view': view}) }}"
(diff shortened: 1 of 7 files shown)


site/app/templates/grading/simple/Display.twig
```diff
@@ -1,20 +1,28 @@
...
+    
+    #data-table td:focus, #data-table input:focus {
+        outline: 3px dashed #337ab7;
+    }
+
+    .fixed-top {
+        position: fixed;
+        top: 0;
+	}
+</style>
```
KevinMackenzie: We're trying to avoid inline styles going forward

site/app/templates/grading/simple/Display.twig
```diff
@@ -23,60 +31,89 @@
...
+    <table id="data-table" class="table table-striped table-bordered persist-area" style="table-layout: fixed;">
         <caption />
         {# Table header #}
         <thead class="persist-thead">
             <tr>
                 <th width="20"></th>
                 <th width="70">Section</th>
                 <th width="90" style="text-align: left">
-                    <a href="{{ core.buildUrl({'component': 'grading', 'page': 'simple', 'action': action, 'g_id': gradeable.getId(), 'sort': 'id', 'view': view}) }}">
+                    <a href="{{ core.buildUrl({'component': 'grading', 'page': 'simple', 'action': action, 'g_id': gradeable.getId(), 'sort': 'id', 'view': view}) }}"
```
KevinMackenzie: are we supposed to be removing the closing `>` on this tag? (and the three other lines below)

site/app/templates/grading/simple/Display.twig
mktung: We're trying to phase out server.css so I figured keeping it on this page for now would make it easier to move foward. I mostly focused on having classes rather than the inline style attr. Should I make a new css page and link it?

site/app/templates/grading/simple/Display.twig
mktung: Good catch! I've pushed a fix.

KevinMackenzie:
I'd say put it in a css file so the task of breaking up server.css /
removing inline styles doesn't get any bigger.

site/app/templates/grading/simple/Display.twig
mktung: Moved to new css file

thef1stofhername:
I like the changes that were made and everything seems perfectly operable  as much as can be within the parameters of this issue. Once I figured out that tab puts you to scroll after the option (I don't use tab much) it performed as expected. yay!

site/app/templates/grading/simple/Display.twig
```diff
@@ -1,20 +1,8 @@
...
-    #}
-    {% if show_all_sections_button %}
-        <div style="float: right; margin-bottom: 10px">
-            <a class="btn btn-default"
-               href="{{ core.buildUrl({'component': 'grading', 'page': 'simple', 'action': action, 'g_id': gradeable.getId(), 'sort': sort, 'view': view_all ? null : "all"}) }}">
-                {{ view_all ? "View Your Sections" : "View All" }}
-            </a>
-        </div>
-    {% endif %}
+<link rel="stylesheet" type="text/css" href="{{ base_url }}css/simple-grading.css" />
```
KevinMackenzie: I probably should have mentioned this before, but there's a function that we call in PHP called `AddCssInternal` that we use to attach CSS (and another one for JS files) so we can force users to update their CSS/JS cache when the files change.  If you find the place where `Display.twig` is being rendered in PHP you can add a call to `AddCSSInternal` there.

mktung: Fixed

Issue #2656, opened 2018-08-04
Success and error messages hidden behind nav bar


ropern123:
When using the top bar of the site wrapper, error and success messages go behind the admin nav bar and become difficult to read.

MasterOdin:
Need to probably lower the z-index of the nav bar (it's at 9999 right now) and raise the z-index of the alert classes (it's at the default) such that alert > nav.

sarthak77:
Can I take up this issue?

sarthak77:
Is someone working on the issue or i can take it?

PR #3808, opened 2019-06-07
Made pop-up message visible.

EBontemps:
Issue: #2656 Alert messages was hidden behind the nav bar. 
Fix: Changed the z-index for the alert and nav section to make the alert message visible.


bmcutler:
Thanks for the contribution!

PR #3814, opened 2019-06-07
Add option to re-assign team rotating sections

### LINKED ISSUE #2367: Option to re-assign randomly the rotating section assignments of teams
Currently teams are assigned to the rotating section of their initial member.
Add a button on the summary/overview or index page for an electronic gradeable that is set to "grade by rotating section" named "Randomly re-assign teams to rotating sections" that will take all current teams and divide them into (mostly) even groups for each rotating section.  Could do this by randomizing the list of teams and then 'dealing' them out to the rotating sections.

jmcknight98:
Closes #2367 

Pull request includes a new button that randomly re-assigns teams to rotating sections.  (For more information see issue closed).  It does not affect student rotating section for the course, and does not affect students that are not on a team yet.  The button will only appear for instructors and only on gradeables that are team assignments graded by rotating section.  If the gradable is open to grading, a popup will warn instructors and require additional confirmation.
Warning Popup:
popup_warning

Fixed a divide by zero exception caused by the "remove all students from rotating sections" option of Manage sections.  However, this option still causes information on the status page for gradeables that are set to be graded by rotating section to be incorrect/nonsense, so a warning banner was added.

Modified the rotating sections form to display the correct number of rotating sections, even if no students are assigned to them.
Before: 
before
After: 

after

Also fixed a small bug that caused rotating sections to be displayed out of order for team assignments, if no teams were created yet.  
Before:
beforechanges

After:
afterchanges

### DIFF
--- a/site/app/controllers/grading/ElectronicGraderController.php
+++ b/site/app/controllers/grading/ElectronicGraderController.php
@@ -37,6 +37,8 @@ class ElectronicGraderController extends GradingController {
             case 'import_teams':
                 $this->importTeams();
                 break;
+            case 'randomize_team_rotating_sections':
+                $this->randomizeTeamRotatingSections();
             case 'grade':
                 $this->showGrading();
                 break;
@@ -106,8 +108,9 @@ class ElectronicGraderController extends GradingController {
             default:
                 $this->showStatus();
                 break;
+            }
         }
-    }
+
 
     /**
      * Checks that a given diff viewer option is valid using DiffViewer::isValidSpecialCharsOption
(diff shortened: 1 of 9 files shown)


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -37,6 +37,8 @@ public function run() {
             case 'import_teams':
                 $this->importTeams();
                 break;
+            case 'randomize_team_rotating_sections':
+                $this->randomizeTeamRotatingSections();
```
mktung: You should add a `break;` here to speed up the execution time

jmcknight98: Yes, this change has been made.  Thanks!

PR #3817, opened 2019-06-09
Default forum categories and separate Categories management page

### LINKED ISSUE #3298: Create default category for discussion forum
Currently if the discussion forum is enabled the instructor/TA must create a category in order for students to post. We should create a default category to ensure that students can post immediately. It might also be nice to let the instructor know that they should create categories for the discussion forum.

Default categories: "General Questions", "Homework Help", "Quizzes/Tests". 
Course settings page create a link or display popup: "Customize the categories for the Discussion Forum"

xprilion:

Closes #3298

### What is the new behavior?
4 categories are created by default when the Forum is enabled first. The following message is displayed - 
Screenshot from 2019-06-10 02-42-05

The link takes to 'show_categories' - 

Screenshot from 2019-06-10 02-42-13

On Create Thread page the categories can be seen - 

Screenshot from 2019-06-10 02-42-21

### Other information?
Is this a breaking change? - This might get in trouble with the refactor of the ForumController.php
How did you test - Will add e2e tests for this. Tested by using all features change/added one by one.


site/app/templates/admin/Configuration.twig
```diff
@@ -278,6 +281,19 @@
             }
         });
 
+        $(document).on('change', '#forum_enabled', function(){
+            var forum_enabled = $('#forum_enabled').is(":checked");
+
+            console.log(forum_enabled);
```
andrewaikens87: Please remove log

site/app/templates/forum/ShowCategories.twig
```diff
@@ -0,0 +1,119 @@
+<div class="content forum_content">
```
andrewaikens87: This file is an almost direct copy of `site/app/templates/forum/CategoriesForm.twig`. From our conversation I thought we were planning on moving the categories to it's own page exclusively (and get rid of the popup). We should just modify the `CategoriesForm.twig` to be it's own page rather than making this a new file. I can envision this page could become a forum setting page in the future.

xprilion:
@andrewaikens87 I have not yet added the button to the More dropdown consciously. It will be better to do so when the forum view refactor takes place, as we must pass from the ForumThreadView.php the url of the category management page.

site/app/templates/forum/ShowCategories.twig
xprilion: I remembered something along the lines of having the popup where it is, only adding conditions to the rendering of that twig file to use/skip the popup. But on working with the file I realized its construction is such that this was not possible, hence skipped and created a new file instead! Would you suggest to entirely remove the popup? I would only need to make a few changes for that! :)

andrewaikens87: Yeah, let's confirm tomorrow morning (I would personally like to do away with the popup if we're going to have a page). But, if we were to keep both, we could have a twig file that would include everything from line `ShowCategories.twig:15` and down. This would then be able to be included in the popup `CategoriesForm.twig` and the corresponding twig file for the categories page. But let's discuss tomorrow.

site/app/templates/forum/ShowCategories.twig
```diff
@@ -0,0 +1,125 @@
+{% if popup == false %}
```
andrewaikens87: Since we are doing away with the popup, we should remove these if checks that supported both.

site/app/templates/forum/ShowCategories.twig
```diff
@@ -0,0 +1,125 @@
...
+                            {% endfor %}
+                        </select>
+                        &nbsp;
+                        <a class="post_button" title="Delete Category"><i class="fas fa-trash" aria-hidden="true"></i></a>
+                    </div>
+                </li>
+            {% endfor %}
+        </ul>
+
+            {% if popup == false %}
```
andrewaikens87: Refer to above message on this.

site/app/templates/admin/Configuration.twig
```diff
@@ -278,6 +281,17 @@
             }
         });
 
+        $(document).on('change', '#forum_enabled', function(){
+            var forum_enabled = $('#forum_enabled').is(":checked");
```
andrewaikens87: There is a slight issue with the display. If I refresh the page (where #forum_enabled) was not changed the check is off but the text still displays. 

uncheckedStillDisplayed

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -112,6 +113,13 @@ public function updateConfiguration() {
             $entry = nl2br($entry);
         }
 
+        if($name === 'forum_enabled') {
+            if($entry == 1){
+                $ForumController = new ForumController($this->core);
```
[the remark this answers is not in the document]
xprilion: This function makes it safe to insert the category, performing checks before doing so. It might lead to redundant code which has to be changed in two places if ever the method of entry of categories changes.

site/app/templates/forum/ShowCategories.twig
xprilion: Sure!

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -4,8 +4,49 @@
 
 use app\controllers\AbstractController;
 use app\libraries\FileUtils;
+use app\controllers\forum\ForumController;
 
 class ConfigurationController extends AbstractController {
+
+    private function isValidCategories($inputCategoriesIds = -1, $inputCategoriesName = -1){
```
andrewaikens87: This is actually in a different class in the refactor, but can we make this static and put it somewhere as a helper function rather than private here? We can pass $core as a parameter. This would then be able to be called from both here and ForumController.

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -112,6 +153,20 @@ public function updateConfiguration() {
             $entry = nl2br($entry);
         }
 
+        if($name === 'forum_enabled') {
+            if($entry == 1){
+                if($this->core->getAccess()->canI("forum.modify_category")) {
+                    $categories = ["General Questions 2", "Homework Help 2", "Quizzes 2" , "Tests 2"];
```
andrewaikens87: Is there a reason that these have a trailing 2 (`*CategoryName* 2`)? I'm not sure if I missed a conversation but should it just be `General Questions`, `Homework Help`, `Quizzes`, `Tests`?

site/app/controllers/admin/ConfigurationController.php
andrewaikens87: To follow up on our conversation from before, this function already does exist and has been moved as part of the refactor effort. It resides in `site/app/libraries/ForumUtils.php` you should be able to reference this from here instead of copying this function (`$rows` is the categories from the query result). Let me know if this doesn't work.

site/app/templates/forum/ShowCategories.twig
```diff
@@ -0,0 +1,120 @@
...
+        "show_threads" : forumBarData.show_threads,
+        "thread_exists" : forumBarData.thread_exists
+    } %}
+
+    <h3 style="margin-bottom:10px;"> Manage Categories </h3>
+
+    <div id="manage_categories_view" align="center">
+        <div style="max-width: 750px;" align="left">
+
+        <input type="hidden" name="csrf_token" value="{{ core.getCsrfToken() }}" />
```
andrewaikens87: @bmcutler from our conversation today, I'm assuming that this is an issue (core access in twig)? It makes more sense to pass the csrf_token as a param to twig anyway.

site/app/templates/forum/ShowCategories.twig
```diff
@@ -0,0 +1,120 @@
...
+            handle: ".handle",
+            update: function (event, ui) {
+                reorderCategories();
+            }
+        });
+        $("#ui-category-list").find(".fa-trash").click(function() {
+            var item = $(this).parent().parent().parent();
+            var category_id = parseInt(item.attr('id').split("-")[1]);
+            var category_desc = item.find(".categorylistitem-desc span").text().trim();
+            deleteCategory(category_id, category_desc, '{{ core.getCsrfToken() }}');
```
andrewaikens87: Refer to other comment regarding `core` access from twig.

site/app/templates/forum/ShowCategories.twig
```diff
@@ -0,0 +1,120 @@
...
+            item.find(".categorylistitem-desc").show();
+        });
+
+        $("#ui-category-list").find(".fa-check").click(function() {
+            var item = $(this).parent().parent().parent();
+            var category_id = parseInt(item.attr('id').split("-")[1]);
+            var category_desc_original = item.find(".categorylistitem-desc span").text().trim();
+            var category_desc = item.find("input").val().trim();
+            if(category_desc != category_desc_original) {
+                editCategory(category_id, category_desc, null, '{{ core.getCsrfToken() }}');
```
andrewaikens87: Refer to other comment regarding `core` access from twig.

site/app/templates/forum/ShowCategories.twig
```diff
@@ -0,0 +1,120 @@
...
+            item.find(".categorylistitem-editdesc").hide();
+            item.find(".categorylistitem-desc").show();
+        });
+        var refresh_color_select = function(element) {
+            $(element).css("background-color",$(element).val());
+        }
+        $(".category-color-picker").change(function(color) {
+            var category_id = parseInt($(this).parent().parent().attr('id').split("-")[1]);
+            var category_color = $(this).val();
+            editCategory(category_id, null, category_color, '{{ core.getCsrfToken() }}');
```
andrewaikens87: Refer to other comment regarding `core` access from twig.

site/app/templates/forum/ThreadPostForm.twig
```diff
@@ -13,7 +13,7 @@
 
     {% if show_editcat and core.getUser().accessGrading() %}
```
andrewaikens87: Refer to other comment regarding `core` access from twig.

site/app/templates/forum/ShowCategories.twig
```diff
@@ -0,0 +1,120 @@
...
+
+    <h3 style="margin-bottom:10px;"> Manage Categories </h3>
+
+    <div id="manage_categories_view" align="center">
+        <div style="max-width: 750px;" align="left">
+
+        <input type="hidden" name="csrf_token" value="{{ core.getCsrfToken() }}" />
+        <span  style="float: right;">
+                <input id="new_category_text" placeholder="New Category" style="resize:none;" rows="1" type="text" name="new_category" id="new_category" maxlength="50" onkeyup="checkInputMaxLength($(this))" />
+                <button type="button" title="Add new category" onclick="addNewCategory('{{ core.getCsrfToken() }}');" style="margin-left:10px;" class="btn btn-primary btn-sm">
```
andrewaikens87: Refer to other comment regarding `core` access from twig.

site/app/templates/forum/ShowCategories.twig
andrewaikens87: Relating to #3900

site/app/controllers/admin/ConfigurationController.php
xprilion: Must have crept when I was testing the removal of ForumController, will get that fixed!

site/app/templates/forum/ShowCategories.twig
xprilion: If we merge in #3900 after this one, it will get corrected!

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -156,10 +117,10 @@ public function updateConfiguration() {
         if($name === 'forum_enabled') {
             if($entry == 1){
                 if($this->core->getAccess()->canI("forum.modify_category")) {
-                    $categories = ["General Questions 2", "Homework Help 2", "Quizzes 2" , "Tests 2"];
+                    $categories = ["General Questions", "Homework Help", "Quizzes" , "Tests"];
 
                     foreach ($categories as $category) {
-                        if (!$this->isValidCategories(-1, array($category))) {
+                        if (ForumUtils::isValidCategories(-1, array($category))) {
```
andrewaikens87: Be careful with the params here, as I mentioned in my comment the first param that is required is the categories as a result from the query.

xprilion: Fixed

andrewaikens87:
Please resolve merge conflicts.

Issue #2362, opened 2018-07-09
Consistent JSON responses


KevinMackenzie:
As outlined in http://submitty.org/developer/json_responses

We need to go through all of the pages that return json responses (i.e. ajax handler methods) and make sure that they conform to our standard on how json responses should be structured.  As of now, it is up to the developer per controller as to how these responses are structured.  

This issue also would mean removing any non-json text from the response.  For example, calling `addErrorMessage` will contaminate the output with HTML (see `ElectronicGraderController.php`).

Complying with this spec involves changing the controller code and the JS that calls these routes (via ajax)

All places that return JSON violate this spec in some way, but I'll list them anyway:
[ ] AdminGradeableController
[ ] GradeableController
[ ] LateController
[ ] UserController
[ ] Forum
[ ] ElectronicGraderController (*almost* complies, i.e. ~~`failure`~~ -> `fail`)
[ ] SimpleGraderController complies with the spec, but doesn't exclusively return json (sometimes redirects)
[ ] SubmissionController

zjxiaohan:
@MasterOdin 
I tried to look into this issue before, and I found it confusing to distinguish "fail" and "error". Instead of our current standard, now that we are going to implement REST API, what about completely using "success" and "fail" (or "error") plus HTTP status code instead of three types of responses?

PR #3821, opened 2019-06-11
[Refactor] Refactor Submission Controller JSON Responses

zjxiaohan:

Partially addresses #2362 .

### What is the new behavior?
Submission Controller now returns JSend format JSON only.

### Other information?
When reading the code of submission controller, I feel that this controller is way too big and complicated. Might need further refactor besides changing JSON format.

### DIFF
--- a/site/public/js/drag-and-drop.js
+++ b/site/public/js/drag-and-drop.js
@@ -607,7 +607,7 @@ function handleBulk(gradeable_id, num_pages, use_qr_codes = false, qr_prefix = "
             $("#submit").prop("disabled", false);
             try {
                 data = JSON.parse(data);
-                if (data['success']) {
+                if (data['status'] === 'success') {
                     window.location.href = return_url;
                 }
                 else {
--- a/site/public/js/drag-and-drop.js
+++ b/site/public/js/drag-and-drop.js
@@ -369,21 +369,6 @@ function isValidSubmission(){
     return false;
 }
 
-function checkForPreviousSubmissions(csrf_token, gradeable_id, user_id){
-    var formData = new FormData();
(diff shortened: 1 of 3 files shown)


site/public/js/drag-and-drop.js
```diff
@@ -369,21 +369,6 @@ function isValidSubmission(){
     return false;
 }
 
-function checkForPreviousSubmissions(csrf_token, gradeable_id, user_id){
```
MasterOdin: Just to confirm, this function isn't used anywhere and that's why you're deleting it?

zjxiaohan: Yes. I did a full-text search and found no occurrences of `checkForPreviousSubmissions`.

Next time, if in a similar situation, is there a good way to tell reviewers that I delete something because it is not used?

MasterOdin: I think either by leaving an inline comment on the function (such as this), adding a comment about it in the first post of the PR, or make a separate PR that just removes this function.

zjxiaohan: Sure. Thank you for your suggestion. Making a separate PR sounds good as it is easier to review. Deleting something from the codebase really makes me apprehensive considering how big the two files involved are.

MasterOdin:
> When reading the code of submission controller, I feel that this controller is way too big and complicated. 
Yeah, this controller is in dire need of restructuring as has too many different things. When I originally wrote this, it was only supposed to handle functions related to the submission page for a student (submit an assignment, change version, cancel version). I'd like to slim down the controller back to that point and move the other functionality to new/existing controllers (e.g. the submit function for student images should be in the student images controller, not this one).

PR #3823, opened 2019-06-11
[Bugfix] Grade inquiries and grade inquiry discussions with correct time

### LINKED ISSUE #3137: Inconsistent time on regrade requests
submitty broke

A student just submitted a regrade request but the time submitted shown does not match the actual time

### LINKED ISSUE #3594: Add timezone info to regrade_requests.timestamp
**Describe the bug**
Currently, regrade_requests.timestamp relies on the DB's timezone being the same as the rest of the system so that it can assume that the proper time was put into the DB. If the DB uses some other timezone, all dates put into the DB will be off when displayed.

**Expected behavior**
A timezone is attached to any and all timestamps inserted into regrade_requests.

sak6lab:
Grade inquiry posts and requests show time that is in UTC time because PostgreSQL does not retrurn the time back into user's system time. 
closes #3594 
closes #3137 

### What is the new behavior?
Grade inquiry posts and requests will now show correct time that they were made.


migration/migrator/migrations/course/20190608032611_grade_inquiry_with_timezone.py
```diff
@@ -0,0 +1,13 @@
+"""Migration for a given Submitty course database."""
+
+
+def up(config, database, semester, course):
+    database.execute("ALTER TABLE regrade_discussion ALTER COLUMN timestamp TYPE TIMESTAMP WITH TIME ZONE;")
+    database.execute("ALTER TABLE regrade_requests ALTER COLUMN timestamp TYPE TIMESTAMP WITH TIME ZONE;")
+    pass
```
zjxiaohan: I don't think a `pass` is needed here both in `up` and `down` functions - It is mostly used as a placeholder for future code, or just to make things syntactically correct.

> pass is a null operation — when it is executed, nothing happens. It is useful as a placeholder when a statement is required syntactically, but no code needs to be executed

https://docs.python.org/3.2/reference/simple_stmts.html#the-pass-statement

zjxiaohan:
For PostgreSQL, if a column of `timestamp` is altered to be `timestamp with time zone`, then the existing times would be converted to the timestamp which would be exactly like a timestamp made in the local time zone.

I am not sure if my concern is valid or there is any need to worry about this but if some timestamps are made without daylight saving time, and this migration is run under daylight saving time, will there be one hour difference?

sak6lab:
I don't think it is big concern because UTC time is not affected by Daylight Savings Time and because the server converts the timestamp back to the time in the server's timezone everything should be fine as long as the server has correctly changed its time.

bmcutler:
The timestamps on grade inquiries on the production system are quite wrong now (off by UTC - localtime).  So we're not going to worry about a migration to correct those times.  Hopefully this will be be correct for future inquiries though :)

PR #3829, opened 2019-06-11
[Bugfix] Added Migration to Create instructor_solution Directory if Nonexistent

emaicus:
Add a migration to create the ```instructor_solution``` directory in a course directory if it does not exist. The lack of this directory is causing assignment builds to fail on production.

### DIFF
--- a/migration/migrator/migrations/course/20190611034600_add_instructor_solution_directory.py
+++ b/migration/migrator/migrations/course/20190611034600_add_instructor_solution_directory.py
@@ -23,6 +23,8 @@ def up(config, database, semester, course):
         shutil.chown(solution_dir, user=user, group=course_group)
         os.system("chmod u+rwx {0}".format(str(solution_dir)))
         os.system("chmod g+rwx {0}".format(str(solution_dir)))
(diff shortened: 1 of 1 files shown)


migration/migrator/migrations/course/20190611034600_add_instructor_solution_directory.py
```diff
@@ -0,0 +1,30 @@
...
+    course_group_id = stat_info.st_gid
+    uid = stat_info.st_uid
+    course_group = grp.getgrgid(course_group_id)[0]
+    user = pwd.getpwuid(uid)[0]
+
+    if not os.path.exists(solution_dir):
+        os.mkdir(solution_dir)
+        shutil.chown(solution_dir, user=user, group=course_group)
+        os.system("chmod u+rwx {0}".format(str(solution_dir)))
+        os.system("chmod g+rwx {0}".format(str(solution_dir)))
```
bmcutler: just in case of bad umask, lets set other -rwx
os.system("chmod o-rwx {0}".format(str(solution_dir)))

and also set the group sgid bit so subdirectories inherit the group
os.system("chmod g+s {0}".format(str(solution_dir)))

PR #3831, opened 2019-06-12
[Feature] Enable Router to Pass _GET Params to Controllers

zjxiaohan:

Controllers rely on `$_GET` to get query parameters.

### What is the new behavior?
Let the router pass `$_GET` parameters to controllers.

### Other information?
New tests are added to ensure that the user does not have control over the router by through weird queries.

### DIFF
--- a/site/app/libraries/routers/WebRouter.php
+++ b/site/app/libraries/routers/WebRouter.php
@@ -30,6 +30,12 @@ class WebRouter {
     /** @var array */
     public $parameters;
 
+    /** @var string the controller to call */
+    public $controller_name;
+
+    /** @var string the method to call */
+    public $method_name;
+
     public function __construct(Request $request, Core $core, $logged_in) {
         $this->core = $core;
         $this->request = $request;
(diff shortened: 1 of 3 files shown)


site/app/libraries/routers/WebRouter.php
```diff
@@ -30,6 +30,12 @@ class WebRouter {
     /** @var array */
     public $parameters;
 
+    /** @var string the controller to call */
+    public $controller_name;
```
zjxiaohan: I added this variable to make it easier for the tester to check its value.

site/app/libraries/routers/WebRouter.php
```diff
@@ -55,17 +61,22 @@ public function __construct(Request $request, Core $core, $logged_in) {
...
         foreach ($this->parameters as $key => $value) {
             if (Utils::startsWith($key, "_")) {
                 unset($this->parameters[$key]);
             }
         }
 
-        return call_user_func_array(array($controller, $methodName), $this->parameters);
+        // pass $_GET to controllers
+        // the user-specified $_GET should NOT override the controller name and method name matched
+        $this->request->query->remove('url');
```
zjxiaohan: If we go to `http://192.168.56.111`, the `url` would be empty, but no need to worry as the `remove` function calls `unset`, so it will be fine if `url` does not exist in the query.

PR #3832, opened 2019-06-12
[Refactor] Clean Some Code in Auth Controller and Add Tests

zjxiaohan:

### What is the new behavior?
- Removed `isLoggedIn` function. There is no need to call `isLoggedIn` in `loginForm`, as the user will never go to the login page if they are already logged in.
- The `submitty_session_id` cookie is not used elsewhere. I believe it meant to say `submitty_session`.
- Removed `$redirect`. It is no longer used in the new router.
- `NavigationController` is supported by router now as the notification part has been separated.
- Added some tests for `checkLogin`.


zjxiaohan:
@shailpatels 

Hi Shail. Thanks for your review!

For the first one, it is not caused by this PR. I've noticed that and have issue #3843 and PR #3856 to track that. I believe the bug will be fixed after #3856 is merged.

For the second one, I traced back the versions of `index.php`, and do not know when the `reset_cookie` is defined. It was introduced by a commit about JWT (https://github.com/Submitty/Submitty/commit/6c394fb52f8f81574b3d5e105391dbce3a04d92c) by @MasterOdin , but I did not find any definitions of `reset_cookie`. Could @MasterOdin have a look?

zjxiaohan:
I removed `$reset_cookie`. Thanks for the review!

PR #3838, opened 2019-06-12
Allow users to have empty string email address & handle (more) errors in send_email script

### LINKED ISSUE #3561: NULL user email field

bmcutler:
Closes #3561  -- each user is now allowed to have an empty string for their email string.  (if not empty string, it must be a string with an @ sign as before)

When sending a batch of emails, now we don't get stuck on a broken/typoed email address, it logs the error and continues.  Similarly, if the email is empty it logs an error and continues.

ALSO...  the emails table will now store the user id (rather than just the email address)


MasterOdin:
Do we ever envision a point where we might be sending emails to an email address that isn't associated with a particular user? Or should emails only be sent out to valid users of Submitty?

bmcutler:
I think I can guess where you're going...  why does the emails table store the email address?  Shouldn't instead store the user id?  Yes, I agree.  I was irked that I couldn't print a name/username to the logfile.

I will switch it now.  The migration of existing data will be messy/broken, but now's the time to do it, before we add a bunch more cases of types of emails into the system.  I can't think of any reason why we would send email to a non user.

MasterOdin:
Cool, that does make maintaining the email table a bit easier as now if a single user has a bunch of broken emails (due to wrong address), we can just update one centralized place.

The migration shouldn't be too bad in that we should be hopefully able to maintain most of the existing data with a series of queries
```
ALTER TABLE emails ADD COLUMN user_id CHARACTER VARYING;
UPDATE emails SET user_id=subquery.user_id FROM (SELECT user_id, user_email FROM users) AS subquery WHERE emails.recipient = subquery.user_email;
DELETE FROM emails WHERE user_id IS NULL;
ALTER TABLE emails ALTER COLUMN user_id CHARACTER VARYING NOT NULL;
ALTER TABLE emails DROP COLUMN recipient;
ALTER TABLE emails ADD CONSTRAINT emails_user_id_fk FOREIGN KEY (user_id) REFERENCES users (user_id);
```

Though the third query (`DELETE FROM ...`) will lose any data that we cannot tie to a specific user anymore. Maybe we could look through the subject/body to figure it out, but not sure it's worth the hassle.

bmcutler:
There are multiple users with the same email address on production.  Does your query account for that?  (does it pick one of them gracefully or does it break?)

I'm not worried about data loss with this table.  I would even be ok just deleting all old rows.

bmcutler:
OK, I think I've got the migration & data fixed.  But still need to go through the places that add data to the emails table, and fix the send_emails script.  But @MasterOdin if you'd like to check the sql, that would be helpful.

It seems to be ok if there are multiple users with the matching email address (it just picks one).  But I'd like confirmation that its actually valid syntax.

I had to add the check for != '', otherwise it assigned users without email addresses for emails without email addresses (could be completely wrong!)

emaicus:
To deal with the issue of multiple students with the same email address, you can change the update query to use ```DISTINCT ON``` which yields the first value returned (note that this could be any of them, as they are returned in an unordered set by default). That would look like this:

```
UPDATE emails SET user_id=subquery.user_id FROM (SELECT DISTINCT ON(user_email) user_id, user_email FROM users) AS subquery WHERE emails.recipient = subquery.user_email;
```

Here is another query you can run against the database to make sure it does what you want:
```
SELECT user_email FROM emails as e, (SELECT DISTINCT ON (user_email) user_id, user_email FROM users) as subquery WHERE e.recipient != '' and e.recipient = subquery.user_email;
```

MasterOdin:
Yeah, I didn't really account/test for people having the same email addresses when writing the queries, though I think Evan's solution is fine. It might also just be best to truncate the table to just avoid this whole thing.

Another question is what is the difference between a null value and an empty string? When would a user have a null email? When would a user have a blank email?

bmcutler:
null vs empty string...  no difference in meaning.  
It doesn't matter to me which way we represent it.

I'm just stumbling through the implementation (writing both versions of the checks as needed)
If the data source is manually filling out the edit student/grader form it is set to an empty string.  I don't know if it might be set as NULL from another data source (registration feed, ?).

MasterOdin:
It'd probably be better to make the column `NOT NULL DEFAULT ''` and not have to worry about handling both empty string and null, especially given that to me, the only way we'd get a null is on an import operation from registration feed or CSV upload, and it's easy enough to just set the string to blank in those cases.

sbin/send_email.py
```diff
@@ -124,27 +148,55 @@ def send_email():
...
+                email_data["user_id"],
+                email_data["send_to"],
+                str(email_send_error))
+            LOG_FILE.write(e+"\n")
+            print(e)
+
+    e = "[{}] Sucessfully Emailed {} Users".format(
+        str(datetime.datetime.now()), success_count)
+    LOG_FILE.write(e+"\n")
+    print(e)
```
MasterOdin: This shouldn't print anything on a successful run as any output will be sent to the owner of the crontab/value of MAILTO, regardless of its on sysout or syserr.

PR #3856, opened 2019-06-15
Use GET params for redirection

### LINKED ISSUE #3843: Login Through /authentication/login Fails
**Describe the bug**
The user cannot go directly to `/authentication/login` and log in. They will be redirected to `/authentication/check_login/` and be presented with the login form. The login will be successful after the second try.

**Expected behavior**
The user should be able to log in through `/authentication/login`.

**To Reproduce**
1. Go to `/authentication/login`
2. log in

**Configuration**
 - OS: Linux
 - Browser chrome

**Additional context**
The reason why this happens is that the Symfony router does not match `/authentication/check_login/` when `$old` is empty. A quick fix would be adding `/authentication/check_login/` as a new route.

I would do it another way: since #3831 *[Feature] Enable Router to Pass _GET Params to Controllers* has been merged, I would put `$old` in the parameter part like `/authentication/check_login?old=` and set a default value to the root page if `$old=null`.

zjxiaohan:

Fixes #3843 . Our `ClassicRouter` has a feature that, after a successful login, redirects users to the page that they tried to visit but previously failed due to access control. This feature has been implemented in the `WebRouter`. However, a bug exists that you cannot be redirected if you go to `/authentication/login` directly and try to log in.

### What is the new behavior?
Instead of passing the URL to redirect to in the URL like `/authentication/login/{old}`, the old URL to redirect to is passed in GET parameters like `/authentication/login?old={old}`, which looks more natural. When `$old` is empty, go to the homepage as the default page, and the router will take care of it if the user did not successfully log in.

### DIFF
--- a/site/app/libraries/routers/WebRouter.php
+++ b/site/app/libraries/routers/WebRouter.php
@@ -94,10 +94,10 @@ class WebRouter {
             $old_request_url = $this->request->getUriForPath($this->request->getPathInfo());
             $this->request = Request::create(
                 '/authentication/login',
-                'GET'
+                'GET',
+                ['old' => base64_encode($old_request_url)]
             );
             $this->parameters = $this->matcher->matchRequest($this->request);
-            $this->parameters['old'] = base64_encode($old_request_url);
         }
         elseif ($this->core->getUser() === null) {
             $this->core->loadSubmittyUser();
--- a/site/app/libraries/routers/WebRouter.php
+++ b/site/app/libraries/routers/WebRouter.php
@@ -95,7 +95,7 @@ class WebRouter {
             $this->request = Request::create(
                 '/authentication/login',
                 'GET',
-                ['old' => base64_encode($old_request_url)]
+                ['old' => urlencode($old_request_url)]
             );
             $this->parameters = $this->matcher->matchRequest($this->request);
         }
--- a/site/app/controllers/AuthenticationController.php
+++ b/site/app/controllers/AuthenticationController.php
@@ -83,7 +83,7 @@ class AuthenticationController extends AbstractController {
      * @var string $old the url to redirect to after login
      */
     public function loginForm($old = null) {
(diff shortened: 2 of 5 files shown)


site/app/libraries/routers/WebRouter.php
```diff
@@ -94,10 +94,10 @@ private function loginCheck() {
             $old_request_url = $this->request->getUriForPath($this->request->getPathInfo());
             $this->request = Request::create(
                 '/authentication/login',
-                'GET'
+                'GET',
+                ['old' => base64_encode($old_request_url)]
```
shailpatels: I'm not too familiar with this code but will this 'old' url get passed through PHP if so should it be URL-encoded instead of base64.
Since base64 allows the '+', '=', and '/' chars which are reserved in ULRs I wasn't sure if allowing them might break the URL being built here.

zjxiaohan: @shailpatels 

Thanks a lot for mentioning this. I used `urlencode` in the beginning, but it weirdly failed to work (I don't know why, probably because the old way of handling it has some inherent problem). Now I tried it again and it works in my local machine now (please help me check that. Thanks!). Yes, `urlencode` and `urldecode` are more suitable in this scenario and using them is a more common practice.

shailpatels:
@ZishanHuang Tested logging in (with the new URL and old) and everything seems fine. I suspect there was a mismatch between URL encoding and decoding somewherebefore which is always a pain but this seems to work correctly.

zjxiaohan:
@shailpatels Thanks for reviewing, especially at this time (if you are in New York lol)!

shailpatels:
> @shailpatels Thanks for reviewing, especially at this time (if you are in New York lol)!

no prob @zjxiaohan and haha yes its about 3:30am here in NY but as an rpi student my sleep schedule is pretty much non-existent lol.

Issue #3830, opened 2019-06-11
Tabbing on modal (popup) windows can cycles to elements on underlying page


bmcutler:
**Describe the bug**
From the "Manage Graders" page, clicking on "Edit Grader" brings up a modal window to edit the fields of the user/grader.  Pressing tab cycles through the form correctly, but then focus jumps to elements on the background.

We just fixed this on the bulk uploads page in #3799 .  I notice its an issue on this modal window, so I assume it's a problem on most modal windows.  Eventually all will need to be fixed.

@shailpatels , any further comments/suggestions?

shailpatels:
This was annoying to fix and was solved by manually taking control of what the tab key did. 
This issue exists for a lot of other people
https://github.com/valor-software/ngx-bootstrap/issues/1819
And I haven't found a good solution for it. Maybe for our case there can be a general JavaScript function that takes a list of elements and corrects the tab order whenever that modal is open.

jchulton:
I've implemented a general Javascript function that takes in a modal id and prevents tab navigating out of that modal. This is done by grabbing the last and first tab-able elements from the modal and looping back to the first element when tabbing out of the last element. I've edited some of the simple modals created in server.js to use this function. Please check out my pull request #4901 4901. I haven't moved to more complex modals like Edit Grader yet, I wanted to get feedback first. Try Deleting and uploading Course Materials.

PR #3859, opened 2019-06-16
[Bugfix] Sequential Forum Tabs Navigation 

### LINKED ISSUE #3539: Tab navigation not working properly
**Describe the bug**
Some of the focus able element cannot be accessed using tabs

**Expected behavior**
Each focus able element should be accessed using tab index 

**To Reproduce**
1. Open discussion forum in Instructor
2. Using tab navigate to `Create Thread`  
3. You can easily see the error. It jumps from `Create Thread` to `Stats` 
 

**Configuration**
All 

**Screenshots**
ezgif com-video-to-gif(15)

**Additional context**

xprilion:

[EDITED] Completes most of the work for #3539.  Remaining work is in PR #3905.

### What is the new behavior?

It is possible to cycle between buttons using tab. 

https://www.loom.com/share/c48a59e282fa4c3ba36ba4ce809f7390

### Other information?
Is this a breaking change? - None
How did you test - Using all three types of accounts, using tab and shift+tab to forward and backward cycle the tabs.


site/app/templates/forum/ForumBar.twig
```diff
@@ -59,42 +64,44 @@
...
-	              {% endif %}
-	        </ul>
-	    </div>
-	{% endif %}
+		{% if show_more %}
+			<div class="dropdown" style="float:right;top:3px;margin-right:5px;">
+				<a tabindex="{{ buttonCounter }}" title="More option" id="dropdownLabel" role="button" data-toggle="dropdown" class="btn btn-default dropdown-toggle" data-target="#" href="#">
+					More <span class="caret"></span>
+				</a>
+				{% set buttonCounter = buttonCounter + 1 + buttonCounterRight %}
```
zjxiaohan: It works in my browser, but I am not sure what this line means. `buttonCounter` doesn't seem to be used in the code below. Why does it need to be set here, and what does `buttonCounter + 1 + buttonCounterRight` mean?

site/app/templates/forum/ForumBar.twig
```diff
@@ -25,6 +25,8 @@
 	</script>
 {% endif %}
 
+{% set buttonCounter = 1  %}
```
drumilpatel2000: Let tabindex be 0 . If tabindex is greater than 0. It will not work as needed as after all tab is navigated  then we come across `Search`, etc buttons which is not intended

site/app/templates/forum/ForumBar.twig
```diff
@@ -38,19 +40,22 @@
 			<form id="search_block" method="post" action="{{ core.buildUrl({'component':'forum', 'page':'search_threads'}) }}">
 			<div class="row" style="position:relative; top:3px;">
 				<div style="padding:0px; margin-right:5px;" class="col-7">
-					<input style="width:100%; height:34px; " type="text" placeholder="Search here..." name="search_content" id="search_content" onchange="$(this).val($(this).val().trim())" required />
+					<input style="width:100%; height:34px; " type="text" tabindex="{{ buttonCounter }}" placeholder="Search here..." name="search_content" id="search_content" onchange="$(this).val($(this).val().trim())" required />
 				</div>
+				{% set buttonCounter = buttonCounter + 1 %}
```
drumilpatel2000: No need to increment buttonCounter

drumilpatel2000: Filter button is not working as expected

site/app/templates/forum/ForumBar.twig
xprilion: Ah, that's a quick workaround to the fact that on the right, the buttons are in `float: right` CSS, and that makes the button on the left actually being on the right...you know how float-right works! So for all the right floated buttons, count them up separately, and then add that back to the next item, so that the tab indexing goes correctly in the view!

xprilion:
@pbailie thanks for the article! I've little experience with tabindex, which clearly shows in this case :( I shall make a prompt update to this with better solution! :)

site/app/templates/forum/ForumBar.twig
xprilion: Understood! Thanks for pointing out !:)

drumilpatel2000:
Great, all navigation are working fine but when we click on filter we cannot access the popup.ie We cannot navigate through the popup.
Screenshot from 2019-06-20 12-42-53

bmcutler:
I think the buggy popup tabbing is in many places, unfortunately
See #3830

xprilion:
@pbailie It is possible to use Shift+Tab to reach the Forum bar menu, however, it will proceed in reverse fashion. 

I could put an autofocus on the search bar, and it would ensure that both the forum bar are reachable quickly and the left navigation by Shift+Tab. Would you suggest doing so?

bmcutler:
Hi @xprilion 

We don't want to use shift+tab, that's the convention to move backwards through the page (which should work too).

Normal forward tabbing through the page should go through the side bar options, then the search bar, search button, filter, more, create thread, then finally into the threads.

(Whatever that means for the implementation)

xprilion:
Including the fix to focus starting from top of the page into #3905 to avoid merge conflicts and in-page JS in multiple files of the forum!

xprilion:
@pbailie @bmcutler the focus now looks like this - 

https://www.loom.com/share/2c5e12ac5d3345c7911e86e0fbcff727

pbailie:
> @pbailie @bmcutler the focus now looks like this -
> 
> https://www.loom.com/share/2c5e12ac5d3345c7911e86e0fbcff727

What I am (still) experiencing is that when I select the discussion forum, focus is automatically set to the thread list.  The next press of `TAB` sends focus to the 2nd item in the thread list.  This is skipping over the top row of controls.  

Yes, I can `SHIFT+TAB` to get to the back row, but that is not the tab order we want.  What @bmutler is requesting is that once we select the discussion forum, the next press of `TAB` should send focus to the breadcrumbs in the header.  And what I am requesting is that once focus enters the discussion forum view, focus should start at the very top with the search bar.

xprilion:
@pbailie I have included the fix for this in the #3905 to avoid the need of another future PR that requires shifting inline CSS and JS for this tasks's solution! Please let me know if you found that behaviour on the branch associated to that PR!

pbailie:
> @pbailie I have included the fix for this in the #3905 to avoid the need of another future PR that requires shifting inline CSS and JS for this tasks's solution! Please let me know if you found that behaviour on the branch associated to that PR!

I apologize for the misunderstanding.  I'll have a look at when I get a chance.  Thanks for helping!

site/app/templates/forum/ForumBar.twig
```diff
@@ -59,43 +67,49 @@
...
-				<a class="btn btn-primary {{ button.optional_class }}" style="{{ button.style }}" {{ link }} title="{{ button.title }}" {{ onclick }} >
-					{{ button.display_text|raw }}
-				</a>
+				{% if link | length > 0 %}
+					<a role="button" class="btn btn-primary {{ button.optional_class }}" style="{{ button.style }}" {{ link }} title="{{ button.title }}" {{ onclick }} >
+						{{ button.display_text | raw }}
+					</a>
+				{% else %}
+					<button type="button" class="btn btn-primary {{ button.optional_class }}" style="{{ button.style }}" title="{{ button.title }}" {{ onclick }} >
+						{{ button.display_text | raw }}
```
andrewaikens87: Just a note (not relating to this PR) but we should consider doing away with these `| raw` (although not pressing)

bmcutler:
Edited comment to leave Issue #3569 open the other PR is closed (and work is review as a whole).

PR #3860, opened 2019-06-17
[Router] New URL for ImagesController

zjxiaohan:
**The changes are based on PR #3861**

### URL Changes

**Old:** 
- `/index.php?semester=s19&course=development&component=grading&page=images&action=view_images_page`
- `/index.php?semester=s19&course=development&component=grading&page=images`

**New:** `/s19/development/student_photos`

### Tested Entrances
- Sidebar -> Student Photos (As Instructor)
- Direct Access `/s19/development/student_photos` (As Student & Instructor)
- Redirection after Uploading a Student Photo (As Instructor)

### How to Review

I am prettifying all the URLs in Submitty! To make the review process easier, I am breaking the replacement work into several pieces, ideally one PR per controller. Please do not hesitate to review even if you are not familiar with my work. I will generally not modify the logic inside the function but just add new routes. You could suggest changes from the following perspectives.

- Are you satisfied with the naming of the new URL?
- Can you think of other entrances to the page that I may forget to add? (If you see a router exception, then it is very likely that I forgot to change URL somewhere.)
- After checking out this branch, is the router working correctly for the new URL? (I have tested it before submitting a PR, of course :) )

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -244,6 +244,7 @@ mkdir -p ${SUBMITTY_DATA_DIR}/logs
 mkdir -p ${SUBMITTY_DATA_DIR}/logs/autograding
 mkdir -p ${SUBMITTY_DATA_DIR}/logs/emails
 mkdir -p ${SUBMITTY_DATA_DIR}/logs/autograding/stack_traces
+mkdir -p ${SUBMITTY_DATA_DIR}/logs/bulk_uploads
 
 #Make site logging directories if not in worker mode.
 if [ "${WORKER}" == 0 ]; then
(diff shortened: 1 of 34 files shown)


zjxiaohan:
The forced push is a rebase operation.

MasterOdin:
Like you did in another PR, let's consolidate the various image functions into one controller (for example, the actual saving of images is in `student/SubmissionController.php`).

It might also be worth moving this controller under `admin`, or some other folder, and out of grading? Something to consider as we go through these controllers is that we're now less bound by the file hierarchy so that we can somewhat layer things however we want now as it makes sense, with more sub-folders and such.

zjxiaohan:
Close this and to have separate PRs for consolidating functions & replacing URLs. Possibly change file hierarchy at the same time.

PR #3861, opened 2019-06-17
Simplified Function for Building New URL

zjxiaohan:

### What is the new behavior?

As most of the URLs begin with `/semester/course` like `/s19/sample`, it is better if the `buildNewUrl` function can automatically prepend this instead of calling `$this->core->getConfig()->getSemester()` and `$this->core->getConfig()->getCourse()`.

### DIFF
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -1010,8 +1010,6 @@ class AdminGradeableController extends AbstractController {
         $this->core->getQueries()->deleteGradeable($g_id);
 
         $course_path = $this->core->getConfig()->getCoursePath();
-        $semester = $this->core->getConfig()->getSemester();
-        $course = $this->core->getConfig()->getCourse();
 
         $file = FileUtils::joinPaths($course_path, "config", "form", "form_" . $g_id . ".json");
         if ((file_exists($file)) && (!unlink($file))) {
@@ -1166,8 +1164,6 @@ class AdminGradeableController extends AbstractController {
             $this->core->addErrorMessage("Failed to update status of ".$g_id);
         }
 
-        $semester = $this->core->getConfig()->getSemester();
-        $course = $this->core->getConfig()->getCourse();
         $this->core->redirect($this->core->buildNewUrl([], true));
     }
 
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -1019,7 +1019,7 @@ class AdminGradeableController extends AbstractController {
         // this will cleanup the build files
         $this->enqueueBuildFile($g_id);
 
-        $this->core->redirect($this->core->buildNewUrl([], true));
+        $this->core->redirect($this->core->buildNewUrl([]));
     }
 
     private function writeFormConfig(Gradeable $gradeable) {
@@ -1164,7 +1164,7 @@ class AdminGradeableController extends AbstractController {
             $this->core->addErrorMessage("Failed to update status of ".$g_id);
         }
 
-        $this->core->redirect($this->core->buildNewUrl([], true));
+        $this->core->redirect($this->core->buildNewUrl([]));
     }
 
     private function checkRefresh() {
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -1019,7 +1019,7 @@ class AdminGradeableController extends AbstractController {
         // this will cleanup the build files
         $this->enqueueBuildFile($g_id);
 
-        $this->core->redirect($this->core->buildNewUrl([]));
+        $this->core->redirect($this->core->buildNewUrl());
     }
 
     private function writeFormConfig(Gradeable $gradeable) {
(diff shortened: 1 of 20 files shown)


site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1021,7 +1019,7 @@ private function deleteGradeable() {
         // this will cleanup the build files
         $this->enqueueBuildFile($g_id);
 
-        $this->core->redirect($this->core->buildNewUrl([$semester, $course]));
+        $this->core->redirect($this->core->buildNewUrl([]));
```
shailpatels: Since the default parameter for `buildNewUrl` is an empty array, `buildNewUlr($parts = array(),...` you don't have pass in an empty array, you can just call the function without any params.

site/app/libraries/Core.php
```diff
@@ -406,12 +406,18 @@ public function buildUrl($parts=array(), $hash = null) {
     }
 
     /**
+     * Given some URL parameters (parts), build a URL for the site using those parts
```
shailpatels: I think the variable `$with_course_info` should be changed to something more descriptive, without reading the code I'm not sure what this actually does, I recommend something like `$prepend_course_info` or `$add_course_info`

I also think there should be a comment (outside of the @ param line) explicitly stating "this function will add the semester and course to the beginning of the new url by default".
I can see this being an area future developers accidentally misuse and run into getting a URL they're not expecting if the functions behaviour isn't well defined.

site/app/controllers/admin/AdminGradeableController.php
zjxiaohan: Sure! A nice suggestion. Yes, I should do that. Will make the change.

site/app/libraries/Core.php
zjxiaohan: `$prepend_course_info` sounds good to me. Sorry that as I am not a native speaker, I am not sure what is the best way to describe this. I do feel it is important to have sufficient comment on this function. Thank you very much for this advice. Will make the change.

zjxiaohan:
To explain the commit history:

Travis CI - Branch failed in a weird place (in an `echo` command). To fix this, I followed some answer online and amended my last commit with a new time, and force pushed it. It automatically dismissed the approval of @shailpatels , and therefore it looks like a bad practice to restart Travis CI this way.

shailpatels:
> To explain the commit history:
> 
> Travis CI - Branch failed in a weird place (in an `echo` command). To fix this, I followed some answer online and amended my last commit with a new time, and force pushed it. It automatically dismissed the approval of @shailpatels , and therefore it looks like a bad practice to restart Travis CI this way.

In the future you if a job fails you can restart it by pressing the 'details' button of the failing check which will bring you to the Travis Website. Theres a button next to every job to restart somewhere on that page, you just need to log in with your github and you should have permission to restart it.

MasterOdin:
Looking through the codebase, it doesn't really look like anyone uses the final `$hash` parameter to the original `buildUrl` function (the only one I'm aware of is:
https://github.com/Submitty/Submitty/blob/247cac33f40d4c97255078a7827ae96b34b904d7/site/app/controllers/admin/UsersController.php#L106-L107
Given this, I think it's fine to remove the last parameter and let the "end-user" append a hash if they need one (which is what the forums do in a couple places).

On a personal style, I think it might be better to just have this be the signature for the functions:
```
public function buildNewUrl(...$parts);
public function buildNewCourseUrl(...$parts) {
    array_unshift($parts, $this->config->getSemester());
    array_unshift($parts, $this->config->getCourse());
    return $this->buildNewUrl(...$parts);
}
```

as I personally dislike adding "flags" to functions if we can avoid them (and prefer just separate functions to express the difference).

zjxiaohan:
The signatures that @MasterOdin look better to me, but I feel there still should be a `$query` parameter. So how about doing:

```
public function buildNewUrl($parts = array(), $query = array());
public function buildNewCourseUrl($parts = array(), $query = array());
```

zjxiaohan:
`$query` changes will be in https://github.com/Submitty/Submitty/pull/3856

MasterOdin:
Is the `$query` part something that is going to be used often? It strikes me that this is potentially something that's just going to exist maybe once or twice and so there's no real need to add this to the function itself and like a hash can just be appended as needed.

zjxiaohan:
> Is the $query part something that is going to be used often? It strikes me that this is potentially something that's just going to exist maybe once or twice and so there's no real need to add this to the function itself and like a hash can just be appended as needed.

Yes, I would think so. While some URLs can be written as `/something/something/something/something`, I personally think there will be something that is more natural if they appear in the query part. Like `/s19/sample/notifications?action=mark_all_as_read` instead of `/s19/sample/notifications/mark_all_as_read` or `/s19/sample/forum/thread?action=new` instead of `/s19/sample/forum/thread/new`.

That being said, it's really about preference. I do feel weird if some current `action` thing is put between slashes.

MasterOdin:
I strongly disagree and that we should avoid those types of URLs unless absolutely necessary. In both cases, it feels like the appended `?action=` bit just allows you to write a big function that either contains a bunch of logic for different types of actions or a dispatcher on that flag, which is sort of antithetical to the whole point of routes. I'd much prefer seeing `GET /s19/sample/forum/thread/delete/<thread_id>` than the alternative as it more strongly forces separate functions for the action logic. However, the best for that example would be `DELETE /s19/sample/forum/thread/<thread_id>` which the API should use.

zjxiaohan:
@MasterOdin Yes... I realized it goes against the point of routes, and after exploring Github's URLs I found it is not a weird practice to put verbs in URLs now. I would make changes to #3856 late today (NY time).

PR #3862, opened 2019-06-17
[Refactor] Refactor Simple Grader Controller JSON Responses

zjxiaohan:

Partially addresses #2362 .

### What is the new behavior?
Simple Grader Controller now uses our JSend format functions. The controller itself has been returning JSend format JSON all the time. This PR just switches to using `renderJsonFail` and `renderJsonSuccess` for easier future maintenance.

`$action` is removed because it is not used in the code.

### DIFF
--- a/site/tests/app/controllers/grading/SimpleGraderControllerTester.php
+++ b/site/tests/app/controllers/grading/SimpleGraderControllerTester.php
@@ -112,7 +112,7 @@ class SimpleGraderControllerTester extends BaseUnitTest {
         $_POST['scores'] = "123";
         $user = $this->createMockModel(User::class);
         $gradeable = $this->createMockModel(Gradeable::class);
-        $core = $this->createMockCore(['csrf_token' => true], [], ['getGradeableConfig' => $gradeable, 'getUserById' => $user]);
+        $core = $this->createMockCore(['csrf_token' => true], [], ['getGradeableConfig' => $gradeable, 'getUserById' => $user], ['canI' => false]);
         $controller = new SimpleGraderController($core);
         $response = $controller->save();
         $this->assertEquals(
--- a/site/tests/app/controllers/grading/SimpleGraderControllerTester.php
+++ b/site/tests/app/controllers/grading/SimpleGraderControllerTester.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace tests\app\controllers\grading;
+
+use app\controllers\grading\SimpleGraderController;
+use app\models\User;
+use app\models\gradeable\Gradeable;
+use tests\BaseUnitTest;
+
+class SimpleGraderControllerTester extends BaseUnitTest {
+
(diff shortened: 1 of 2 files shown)


site/tests/app/controllers/grading/SimpleGraderControllerTester.php
```diff
@@ -0,0 +1,126 @@
...
+        $response = $controller->save();
+        $this->assertEquals(
+            [
+                'status' => 'fail',
+                'message' => "Didn't submit any scores"
+            ],
+            $response
+        );
+    }
+
```
shailpatels: I think the `testSaveNoAccess()` isn't setting the configuration for a user correctly, you can explicitly set the access levels of the mock user you're creating in `createMockCore`, you can do this by setting the configuration of the second param, `$user_config` or the fourth param, `$access`. I think setting setting `$access` to match a user without permission would make more sense for this test since the actual controller uses the `getAccess()` function in the `core` class.

zjxiaohan:
>  I think setting $access to match a user without permission would make more sense for this test since the actual controller uses the getAccess() function in the core class.

Indeed, it makes more sense if a user is set like one user without any permission.

By the way, I noticed that you are working on tests recently. Could you suggest some approaches if I want to write tests for the remaining part of `save` function? For example, for the `$gradeable->getComponents()` part, do I need to write quite a lot of mock code to have it tested? Or is there a better way to write unit tests?

zjxiaohan:
I marked this as [WIP] because I want to make the tests complete for JSON returning functions. :)

shailpatels:
> > I think setting $access to match a user without permission would make more sense for this test since the actual controller uses the getAccess() function in the core class.
> 
> @shailpatels
> 
> Indeed, it makes more sense if a user is set like one user without any permission.
> 
> By the way, I noticed that you are working on tests recently. Could you suggest some approaches if I want to write tests for the remaining part of `save` function? For example, for the `$gradeable->getComponents()` part, do I need to write quite a lot of mock code to have it tested? Or is there a better way to write unit tests?

yes I've began working on testing (but still new so this is a learning experience for me too!). You will have to mock the component but all you will need are specific pieces of data associated with a gradeable component . I think a way of doing this is building a helper function to create a mock gradeable component in whatever state you'd like to test the save function, and you only define the methods you need in that mock component. Here's an example of defining a method from the submissionControllerTester: 

``` PHP
$graded_gradeable->method('getSubmitter')->willReturn($this->createMockSubmitter('testUser'));
```

Since your not explicitly testing a component in this test all you need is the bare minimum of the gradable component that makes the test(s) work (assume all functions behave as described by the gradable component model).

Since the save function is probably going to test multiple states of a gradeable component I recommend building a helper function that builds a mock component in the exact state you'd like.

Try checking out `submissionControllerTester.php` under tests/app/controllers/submission which is a more fleshed out test class.

PR #3864, opened 2019-06-17
[Bugfix] Fix the vcs_login authentication endpoint

MasterOdin:

Attempting to interact with the internal Submitty git server throws 500 errors.

### What is the new behavior?
Fixes 500 error when trying to interact with internal git server, as the endpoint required semester/course information for the endpoint.

### Other information?

Git clone/git pull/git push of created repo.

### DIFF
--- a/site/app/controllers/AuthenticationController.php
+++ b/site/app/controllers/AuthenticationController.php
@@ -148,7 +148,14 @@ class AuthenticationController extends AbstractController {
     }
 
     /**
-     * @Route("/authentication/vcs_login")
+     * Handle stateless authentication for the VCS endpoints.
+     *
+     * This endpoint is unique from the other authentication methods in
+     * that this requires a specific course so that we can check a user's
+     * status, as well as potentially information about a particular
+     * gradeable in that course.
+     *
+     * @Route("{_semester}/{_course}/authentication/vcs_login")
      */
     public function vcsLogin() {
         if (empty($_POST['user_id']) || empty($_POST['password']) || empty($_POST['gradeable_id'])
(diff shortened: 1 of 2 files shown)


zjxiaohan:
Thanks for making the change. I checked out this branch and git operations & authentication seem to be working.

There are several things I am not sure about... I will put it down here and investigate it tomorrow (like 10 hours later).

1. Is [the document](https://submitty.org/sysadmin/git) outdated? There is a piece of code like this:
```
    <Location "/git">
        Require all granted
    </Location>
```
However, when I go to [the most recent Apache config](https://github.com/Submitty/Submitty/blob/master/.setup/vagrant/sites-available/submitty.conf). I cannot find this. In my local machine, adding this will make git fail.

2. I tried to run `sudo -u www-data /usr/local/submitty/sbin/authentication.py` to test, but it shows
```
True
True
False
True
None
None
None
None
```
instead of
```
    True
    True
    False
    True
    True
    True
    False
    True
```
. I am not sure if I need to do something before running this script.

3. There is a typo in `authentication.py`. `sudo -u www-data /usr/local/submitty/bin/authentication.py` should be `sudo -u www-data /usr/local/submitty/sbin/authentication.py`.

MasterOdin:
1. Hm, yeah that documentation is out of date and needs to be updated (similar with the files under `.setup/apache/`. I'll handle that in a separate PR.

2. Yeah, the four final tests were a legacy feature in that there was a concept that an instructor could make a "course" git repo for a student to use throughout. It was abandoned and a gradeable_id must now be used to generate the repo (though the gradeable_id does not actually have to exist).

3. Thanks for pointing that out. `authentication.py` did live in `bin/` at one point (when there was no `sbin/`).

zjxiaohan:
> Yeah, the four final tests were a legacy feature in that there was a concept that an instructor could make a "course" git repo for a student to use throughout. It was abandoned and a gradeable_id must now be used to generate the repo (though the gradeable_id does not actually have to exist).

I am wondering how my code passed the tests before. Does Travis CI run `authentication.py`? Is there anything broken accidentally?

MasterOdin:
It does not get tested under Travis. Something that should probably be done first is making the Travis apache configuration more closely match the one running under Vagrant and then add some simple "e2e" test, doing something like:
```
git clone <repo_url>
cd repo
touch test.txt
git add .
git commit -m "Test"
git push
touch test2.txt
git add .
git commit -m "Second commit"
git push
git reset --hard HEAD~1
git pull
```

and just verify that none of the commands fail, and the resulting repo has the two empty files.

PR #3868, opened 2019-06-18
[Feature] CSRF Check for POST Methods

### LINKED ISSUE #2399: Use proper checks on GET/POST methods for controllers
Right now, all functions can be activated via just having something in the $_REQUEST field without any further care. We should be limiting it such that methods that must be explicitly called via $_POST require that `$_SERVER['REQUEST_METHOD']` is set to the appropriate type (GET/POST) for a given route. This is definitely necessary for #300 and an extension of the work outlined in #933.

As a side effect of doing this, we can then within our router have it validate the CSRF token automatically for any POST request which also lowers the chance of us forgetting to do that within a controller method and expose a security bug.

zjxiaohan:

Closes #2399 .

### What is the new behavior?
Checks CSRF token for every POST requests except those managed by `AuthenticationController` considering that CSRF token is only present after a session is created.

### DIFF
--- a/site/app/libraries/routers/WebRouter.php
+++ b/site/app/libraries/routers/WebRouter.php
@@ -36,6 +36,16 @@ class WebRouter {
     /** @var string the method to call */
     public $method_name;
 
+    /**
+     * WebRouter constructor.
+     *
+     * The constructor parses the request and obtains raw parameters.
+     *
+     * @param Request $request
+     * @param Core $core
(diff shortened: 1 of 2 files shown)


zjxiaohan:
Close this one and:
- Investigate & report the different types of return value & what happens
- Then we can decide how to move forward

zjxiaohan:
Should we keep the PR for future reference?

Issue #3852, opened 2019-06-14
Error when attempting to delete gradeable


jmcknight98:
When attempting to delete a gradeable an irrecoverable server error occurs.  After manually navigating back to the gradeables page, the gradeable is deleted correctly.  Seems to be a problem for all gradeables, but was mostly reproduced by deleting a gradeable immediately after it was created.

Warning displayed:
![bugReport](https://user-images.githubusercontent.com/34327811/59524649-77f67f00-8ea2-11e9-84e3-7f161ba8b18c.png)

1.  Go to gradeables page
2.  Select delete gradeable 
3.  Confirm Delete on the warning popup

jmcknight98:
The same error occurs on a separate branch that I've been working on when creating a new gradeable.  This has not been able to been reproduced on master, so it may be unrelated (or requires very specific circumstances).  Thought I'd mention it though seeing as the error message was exactly the same.

PR #3872, opened 2019-06-18
New option for grader assignment

### LINKED ISSUE #2683: Grade by Registration / Grade by Rotating / NEW OPTION
Add 3rd radio button option, make vertical list.  Add these lengthier descriptions as inline documentation (discuss/tweak/revise text?)

O Registration Section (fixed assignment specified on Graders page)
O Rotating Section (can cycle assignments per Gradeable)
O All XXXGradersXXX have grading access to All Students (E.g., per component/problem grading assignments, assignments announced/managed external to Submitty)

where XXXGradersXXX is replaced with "Instructors" or "Instructors and Full Access Graders" or "Instructors and Full Access Graders and limited Access Graders".

3rd option Implementation:  It's really like selecting "Grade by Rotating", and clicking "All" for each of the graders in the chosen access level.  Yes, if you mistakenly or out of curiosity click on this option, it will clobber your previously carefully clicked rotating section choices and you'll have to redo them.

jmcknight98:
Closes #2683 

Currently two options for grading method: Rotating and Registration sections.

### What is the new behavior?
Added a third option to assign everyone with permission to grade the assignment to all sections.  Grader assignment method is now stored as an integer to allow for new options in the future.  For more information see the issue referenced above.

### Other small bug fixes

- When setting up rotating section assignments on the grader assignment tab of the edit gradeable page, it would previously not allow you to unselect all the boxes.  When attempting to remove the last check, there would be an "Error Saving Graders".  This has been changed to allow the last box to be unchecked in case an instructor wants to start from scratch.

- On the same page, a bug occurred if rotating sections were previously set up for a gradeable, but the page was loaded in with grade by registration section selected.  When switching back to rotating section, the rotating section assignments appear blank.  However, if the page is refreshed, it will reveal that the previous setup is actually selected.  This has been changed to correctly display the selections without needing to refresh the page.

- An inconsistency occurred if a limited access grader/full access grader was assigned to a rotating section for a gradeable and then the gradeable was changed to no longer give them permission to grade it.  On the Status page, the individual would correctly not be listed as a grader for the sections they were assigned.  On the grading details page, they would still be listed.  This has been changed to no longer list graders without permission on the grading details page.

### Other information?
This is still a WIP because there is an issue creating a new gradeable on this branch.  It seems like it is a redirection issue due to the new router because the error message is identical to the one in #3852.  This issue with create gradeable has not been reproduced on the master branch.  It is certainly possible that I accidentally broke something with my changes, but it could also be very specific circumstances that causes a redirect issue that has yet to be discovered on master.  If somebody that is more experienced with the new router/webpages could take a look at it and let me know their thoughts, I would appreciate it.

Steps to reproduce the behavior (on the new_option branch):
1.  Select New Gradeable on the sidebar
2.  Select any viable options
3.  Click Create New Gradeable


site/app/models/gradeable/Gradeable.php
```diff
@@ -90,8 +90,8 @@ class Gradeable extends AbstractModel {
     protected $instructions_url = "";
     /** @property @var int The type of gradeable */
     protected $type = GradeableType::ELECTRONIC_FILE;
-    /** @property @var bool If the gradeable should be graded per registration section (true) or rotating sections(false) */
-    protected $grade_by_registration = true;
+    /** @property @var int If the gradeable should be graded by all access (2) by registration section (1) or rotating sections (0) */
+    protected $grader_assignment_method = 1;
```
KevinMackenzie: can you make this an enumeration instead of a plain integer (like `Gradeable::GRADE_BY_REGISTRATION`, etc.)?   This way the values will be self-documenting and when they're used scattered throughout the code they won't look like "magic numbers".

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -3257,15 +3257,14 @@ private function updateGradeableComponents(\app\models\gradeable\Gradeable $grad
...
-
         // If the gradeable has been modified, then update its properties
         if ($gradeable->isModified()) {
             $params = [
                 $gradeable->getTitle(),
                 $gradeable->getInstructionsUrl(),
                 $gradeable->getTaInstructions(),
                 $gradeable->getType(),
-                $this->course_db->convertBoolean($gradeable->isGradeByRegistration()),
+                $gradeable->getGraderAssignmentMethod(),
```
KevinMackenzie: There's a similar line to this in `CreateGradeable` that also needs to be changed (3132).  This may be what's causing your crash when creating a gradeable.

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -373,7 +373,8 @@ public function getAverageForGradeable($g_id, $section_key, $is_team) {
 
         return ($this->course_db->getRowCount() > 0) ? new SimpleStat($this->core, $this->course_db->rows()[0]) : null;
     }
-    public function getGradeablesPastAndSection() {
+    public function getGradeablesPastAndSection($gradeable_id) {
```
KevinMackenzie: This feels like like an existing function with straight-forward functionality that a feature got tacked onto it because it was convenient.  It makes more sense to rename the function so it more clearly describes its own behavior, or to make another query for the additional information when needed (looking at the usage of the new features).

jmcknight98: Yes I certainly didn't help what I believe is an already confusing function.  Now I renamed it and described the new behavior, hopefully it is more clear.  As for keeping the old query and creating a new one, I don't believe this is necessary as this function was made for this specific purpose and isn't used anywhere else.  However, I can certainly change this if you believe there is a reason to leave the original.

site/app/libraries/database/DatabaseQueries.php
jmcknight98: Yes, I changed this and it seems to be working.  Thanks for the help!

site/app/models/gradeable/Gradeable.php
jmcknight98: I made this change as suggested.  I did it within the Gradeable file as only the constants were needed instead of in a new file (like GradeableType).  If this is bad practice or if there are any other suggestions please let me know!

KevinMackenzie:
The comments I made have been addressed, but there's a install-related bug in the current version of `libssl` so I can't get a dev box running.  Can someone else test to make sure this works as intended?  Maybe @bmcutler ?

bmcutler:
@jmcknight98 I tested this and the assignment of graders to sections looks ok, but when I go to preview grading, I get the red banner error about rotating sections not being setup, even though they are setup.  Maybe it's because I tested with a gradeable that is not yet open to grading?  We can talk more tomorrow.

PR #3874, opened 2019-06-18
[Bugfix] Fix Broken Vagrant Up

emaicus:
A dependency issue is currently causing vagrant up to fail. This PR addresses the issue by adding a ```dist-upgrade``` to the ```setup_distro.sh``` scripts.

### DIFF
--- a/.setup/distro_setup/ubuntu/bionic/setup_distro.sh
+++ b/.setup/distro_setup/ubuntu/bionic/setup_distro.sh
@@ -16,6 +16,7 @@ fi
 #################
 
 apt-get -qqy update
+yes | apt-get -qq dist-upgrade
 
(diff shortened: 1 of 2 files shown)


.setup/distro_setup/ubuntu/bionic/setup_distro.sh
```diff
@@ -16,6 +16,7 @@ fi
 #################
 
 apt-get -qqy update
+yes | apt-get -qq dist-upgrade
```
MasterOdin: Does `-y` not work for this command?

MasterOdin:
Bento released a new version of the box today. Does that fix things without adding this?

MasterOdin:
Closing this as the `bento/ubuntu-18.04` image was updated and released and `vagrant up` is now functioning as expected.

PR #3879, opened 2019-06-19
[Bugfix] Fix broken redirect on exception on create gradeable

MasterOdin:

Fixes issue that @jmcknight98 commented about in #3872. Any exception caused in the Create Gradeable flow would attempt to be redirected to `index.php?semester={_semester}&course={_course}` instead of `{_semester}/{_course}`.

### What is the new behavior?
Redirect now goes to the new (and working) `{_semester}/{_course}` endpoint.


zjxiaohan:
Close this PR because it is part of #3880.

PR #3881, opened 2019-06-19
[Refactor] Gather Course Material Functions

zjxiaohan:

Some course material related functions are scattered in MiscController and SubmissionController. This makes it hard to add routes to the course material part and make other controllers filled with too many functions.

### What is the new behavior?
- Gathers all functions concerned to CourseMaterialsController.
- Removes unused method (the deletion function in the original CourseMaterialsController) and unused variables.
- Modifies all links concerned.

### Other information?
Tested by manually going through all "actions".

`uploadResultMessage` is shared by `SubmissionController` and `CourseMaterialsController`, so I put it in `Output.php`. It looks like what we would like the API to be, but in the future, we will move to a more implicit way in `index.php` (or the main router).

### DIFF
--- a/site/app/controllers/MiscController.php
+++ b/site/app/controllers/MiscController.php
@@ -28,29 +28,14 @@ class MiscController extends AbstractController {
             case 'download_file_with_any_role':
                 $this->downloadFile(true);
                 break;
-            case 'delete_course_material_file':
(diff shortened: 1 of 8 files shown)


zjxiaohan:
@thef1stofhername Thanks a lot for reviewing so much code. :)

Issue #2171, opened 2018-06-14
Sample data: Note to TA / Student


CouleeApps:
Currently these are always blank. It would be nice if they had a chance of having text.

bmcutler:
Let's add test cases to make sure these messages appear to the TAs (both note to TA & note to student should be displayed), and students (*only* not to student should be displayed -- note to TA might contain secret information that should stay hidden from students)

PR #3885, opened 2019-06-19
[Refactor:SampleData] Added sample TA/instructor notes for components

Fenn-CS:

Currently, Note to TA's and Students in the Rubrics section while creating or editing a gradeable are always blank. 

Fixes: #2171 

### What is the new behavior?
Generates random comments to supply as notes.
### Other information?
There's a file 'TAcomments.txt` that holds some example comments that are used randomly to generate TA comments there's no such file for student comments. Is it better to add such a file with a few comments or use `lipsum` (https://github.com/thanethomson/lipsum), possible for both TA and student comments eliminating the need for both files.

### DIFF
--- a/.setup/bin/setup_sample_courses.py
+++ b/.setup/bin/setup_sample_courses.py
@@ -220,19 +220,26 @@ def main():
     # queue up all of the newly created submissions to grade!
     os.system("/usr/local/submitty/bin/regrade.py --no_input /var/local/submitty/courses/")
 
-def generate_random_user_id(length=15):
-    return ''.join(random.choice(string.ascii_lowercase + string.ascii_uppercase +string.digits) for _ in range(length))
-
-def generate_random_ta_comment():
+def get_random_text_from_file(filename):
     line = ""
-    with open(os.path.join(SETUP_DATA_PATH, 'random', 'TAComment.txt')) as comment:
+    with open(os.path.join(SETUP_DATA_PATH, 'random', filename)) as comment:
         line = next(comment)
         for num, aline in enumerate(comment):
             if random.randrange(num + 2): continue
             line = aline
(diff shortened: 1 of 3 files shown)


Fenn-CS:
random_ta_notes

bmcutler:
As we discussed in the meeting, component_comment is different than ta_comment (or ta_note).  component_comment is the individual feedback from a ta to a specific student about their work.  The ta_comment/ta_note is instructions on how to grade all students for this assignment from the instructor to the ta.  similarly the student_comment is general information from the instructor to the students about how the assignment will be graded.

Please make 3 separate functions.   (The messages for those functions should come from a short list of sample phrases, could be stored in a file, or directly in the function. ) .  I suggest renaming the existing ta comment function to use the word component.  There is also an (optional) 'overall comment' from the grader to the student with specific feedback for that student about their work.

Fenn-CS:
Ok, Barb. I have made the requested changes in the last commit :

ta-student-note

However, I noticed no instructions are given for gradeables that have the "Checkpoints" type. Is this behavior ok?

checkpoint_gradeable

MasterOdin:
Should the randomly selected output have some relation to each other? Right now, it's possible to end up with (for example) a student comment of `Code may make use of classes and methods or just a bunch of functions.` and a TA comment of `Deduct points if student makes use of tabs.`.

Fenn-CS:
@MasterOdin  yeah thought of that. It's weird since some of the student notes actually fall in line with what is noted to instructors as well!
If this is important! I can add some code to ensure that the comments are related. @bmcutler

bmcutler:
It is not necessary to add logic to meaningfully match up the comments.  One step up from lorem ipsum is fine.

PR #3888, opened 2019-06-20
[Feature] Add script to automate hitting generate grade summaries button

### LINKED ISSUE #3711: Nightly run of "Generate Grade Summaries" and Rainbow Grades
Currently, Generate Grade Summaries only happens when the instructor presses the website button.  And Rainbow Grades happens on the instructors local machine (copying files on & off, running C++ code).

We'd like to have this done once per day by a cron job run by submitty_daemon.

submitty_daemon can't press the generate grade summaries button.  And we don't want to port the php code to python.  But this is the right place to do an API call, right?  Maybe?
Generate grade summaries saves files to the server, then the submitty_daemon job takes over easily.

MasterOdin:

Fixes #3711, as currently there was no way to automate "pressing" the grade summaries button.

### What is the new behavior?
A script (that can be hooked up to cron or manually run) that will click the button when passed the appropriate credentials. Script is runnable by the submitty_daemon user.

Runnable by doing:
`/usr/local/submitty/sbin/generate_grade_summaries.py <user_id> <password> <url>`
e.g.:
`/usr/local/submitty/sbin/generate_grade_summaries.py instructor instructor http://192.168.56.111`

### Other information?

A future revision of this is probably to save the user credential information into a config file. Additionally, the base_url should also probably be read from a config file at the same time (I had originally written this as a script in SysadminTools, and moved it here as is).

This will probably be modified somewhat once the API is in place, but the gist of it will be very similar/same.

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
@@ -71,7 +71,7 @@ chmod 550 ${SUBMITTY_INSTALL_DIR}/sbin/authentication.py
 chmod 555 ${SUBMITTY_INSTALL_DIR}/sbin/killall.py
 
 # DAEMON_USER only things
-array=( build_config_upload.py run_lichen_plagiarism.py send_email.py submitty_autograding_shipper.py submitty_autograding_worker.py submitty_daemon_jobs autograder)
+array=( build_config_upload.py run_lichen_plagiarism.py send_email.py generate_grade_summaries.py submitty_autograding_shipper.py submitty_autograding_worker.py submitty_daemon_jobs autograder)
 for i in "${array[@]}"; do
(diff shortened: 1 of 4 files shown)


zjxiaohan:
Hi Matt. Please see https://github.com/Submitty/Submitty/pull/3961 's `Other Information` part. Will it be better?

MasterOdin:
Yeah, it'll be better. We can drop `requests-html` for `requests`, and the sequence will be similar to how you described in your PR (make API call to homepage to get all courses), then make call via API to each course's grade summary generation page.

For now, we can store the admin user's JWT on the server, but we'll need to come up with a good protocol for how we want to handle getting it/setting it up. Doesn't need to be solved in this PR though.

MasterOdin:
Closing in favor of #3977

PR #3896, opened 2019-06-21
[Refactor] Notification refactor

### LINKED ISSUE #3826: Add discussion forum email and notification functionality
Emails regarding discussion forum notifications such as changes to posts, replies, announcements, etc., are not being made. 

**Additional context**
The settings are there but there is no code to implement the emails.
discussion_forum_emails

### LINKED ISSUE #3878: Notifications Refactor
Currently the Notification model and the data source DatabaseQueries has a lot of notification logic that reduces the amount of database queries we have to make but increases the amount of code we have to write and reduces abstraction which makes adding functionality to new notifications difficult as we have to write these special cases.

Notificaiton.php:
```
private function handleForum($details) {
        $this->setType($details['type']);
        switch ($details['type']) {
            case 'new_announcement':
                $this->actAsNewAnnouncementNotification($details['thread_id'], $details['thread_title']);
                break;
            case 'updated_announcement':
                $this->actAsUpdatedAnnouncementNotification($details['thread_id'], $details['thread_title']);
                break;
            case 'new_thread':
                $this->actAsNewForumThread($details['thread_id'], $details['thread_title']);
                break;
            case 'reply':
                $this->actAsForumReplyNotification($details['thread_id'], $details['post_id'], $details['post_content'], $details['reply_to'], $details['child_id'], $details['anonymous']);
                break;
            case 'merge_thread':
                $this->actAsForumMergeThreadNotification($details['parent_thread_id'],  $details['parent_thread_title'], $details['child_thread_title'], $details['child_root_post'], $details['child_thread_author']);
                break;
            case 'edited':
                $this->actAsForumEditedNotification($details['thread_id'], $details['post_id'], $details['post_content'], $details['reply_to']);
                break;
            case 'deleted':
                $this->actAsForumDeletedNotification($details['thread_id'],  $details['post_content'], $details['reply_to']);
                break;
            case 'undeleted':
                $this->actAsForumUndeletedNotification($details['thread_id'], $details['post_id'], $details['post_content'], $details['reply_to']);
                break;
            default:
                return;
        }
    }
```

I would like to refactor the notification system so that the logic is in a new classes where the logic is in two new classes: EventPublisher and EventObserver.
submitty notificaitons design (2)
The EventPublisher is the middleman which sends information about this new event to all observers. This allows us to scale our notifications easily. If we want to add a new kind of notification such as notifications for when a user joins a team we would add it to the respective observer. If we wanted to support say slack notifications we would just add a new observer.
https://codeburst.io/observer-pattern-object-oriented-php-4e669431bcb9

sak6lab:
Notification logic was scattered on the controllers, Notification model, and database queries and some notifications and emails that had settings that controlled whether a user received that notification were not implemented.

### What is the new behavior?
I redesigned the notification system so adding a new kind of notification is as easy as adding a function to NotificationFactory to handle that event.

Submitty Notification Design

A controller will make an event (an array containing essential parameters for Notification and Email) and send as a parameter to the function in NotificationFactory designed specifically to handle that event. The NotificationFactory will usually get the recipients who want this notification/email (in the case of grade inquiries I got the recipients in the SubmissionController because it would have required a query when that information was conveniently already in the graded_gradeable). Finally the NotificationFactory will call a query to insert the notification or an email for all the given recipients (this is one query).

I also implemented the notification and email functionality for the settings that were not implemented.

closes #3878
closes #3826 
  
### Other information?
How I tested was quite tedious and definitely calls for tests. To test every possible situation, I had to test every type by physically mimicking each possibilities (ie. switching users)


MasterOdin:
While Andrew is right, it should be noted that you cannot parameterize a column or table name within a query, and that they have to be dynamically inserted. To make up for this, ideally, there would be a whitelist in the function to ensure that the column name passed to the function is valid and in that whitelist.

sak6lab:
> While Andrew is broadly right, it should be noted that you cannot parameterize a column or table name within a query, and that they have to be dynamically inserted. To make up for this, ideally, there would be a whitelist in the function to ensure that the column name passed to the function is valid and in that whitelist.

By a whitelist do you mean an array of accepted values or like if statements?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2637,83 +2637,144 @@ public function getSubmitterIdFromAnonId(string $anon_id) {
...
+
+    /**
+     * returns all authors who want to be notified if a post has been made in a thread they have posted in
+     * @param $thread_id
+     * @param $column ("reply_in_thread" or "reply_in_thread_email")
+     * @return array
+     */
+    public function getAllThreadAuthors($thread_id, $column) {
+        $params = [$thread_id];
+        $preferences = $this->getAllPreferences();
```
MasterOdin: is this defined anywhere?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2637,83 +2637,144 @@ public function getSubmitterIdFromAnonId(string $anon_id) {
...
+    private function createRecipientArray(int $recipient_count) {
+        return implode(', ', array_fill(0, $recipient_count,'?'));
+    }
+
+    /**
+     * Sends notifications to all recipients
+     * @param $notification
+     * @param $recipients
+     */
+    public function pushNotifications($notification, $recipients){
```
MasterOdin: rename this to be `insertNotifications`

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2637,83 +2637,144 @@ public function getSubmitterIdFromAnonId(string $anon_id) {
...
+
+
+    }
+
+    /**
+     * Queues emails for all given recipients to be sent by email job
+     * @param Email $email
+     * @param string $recipients
+     */
+    public function pushEmails(Email $email, array $recipients){
```
MasterOdin: rename this to `insertEmails`

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2735,7 +2796,7 @@ public function getUserNotifications($user_id, $show_all){
         $rows = $this->course_db->rows();
         $results = array();
         foreach ($rows as $row) {
-            $results[] = new Notification($this->core, array(
+            $results[] = Notification::createViewOnlyNotification($this->core, array(
                     'view_only' => true,
```
MasterOdin: can get rid of this key from the array as it's not used in the factory method.

site/app/models/Email.php
```diff
@@ -25,10 +23,6 @@ class Email extends AbstractModel {
     /** @property @var string user name */
     protected $user_id;
 
-    /** @property @var string Intended receiver of email */
-    // NOTE: THIS IS ESSENTIALLY A DEPRECATED / LEGACY FIELD
-    protected $recipient;
```
MasterOdin: Leave this in for now. Removing this should be accompanied with a migration to delete the column from the DB.

site/app/models/NotificationFactory.php
```diff
@@ -0,0 +1,126 @@
...
+     * @param array $event
+     * @param array $recipients
+     */
+    public function createAndSendEmails(array $event, array $recipients) {
+        if (empty($recipients)) {
+            return;
+        }
+        $event["user_id"] = $this->core->getUser()->getId();
+        $email = new Email($this->core,$event);
+        $this->core->getQueries()->pushEmails($email,$recipients);
```
MasterOdin: The `user_id` of emails is who should receive the email, not who initiated an action that sends emails.

sak6lab: Should the email object resemble a row entry in the emails table? I am using the object a contianer for what the email is and the list of recipients? Can I just rename user_id to author. I think that is definitely clearer.

site/app/libraries/database/DatabaseQueries.php
sak6lab: I could not find a relevant query.

MasterOdin: I'm not sure what that means? It looks like this line refers to a non-existent function in this file and will throw an error when used.

site/app/models/NotificationFactory.php
MasterOdin: For rows in the email table, user_id stands for who we are sending the email to. We don't store the sender in the table. @bmcutler would have to answer if she thinks tracking that information would be useful, though I'm not sure what it would be used for. If it is something we want, probably requires some sort of answer of "what information are we interested in and the potential questions we might ask about it" when looking through the logs.

bmcutler: We didn't briefly consider the possibility of having multiple sender emails (if available) -- but that's not the current design/implementation.  The actual sender is the same for every email.  No need to store that.

In some cases we do have a real user who caused the email to be sent.  E.g. an announcement.  And we would like (in some cases -- perhaps up to the user?) to include their name in the body of the message.  But that's not a separate column of the table.

site/app/models/Email.php
```diff
@@ -22,8 +22,8 @@ class Email extends AbstractModel {
     protected $subject;
     /** @property @var string Body of email */
     protected $body;
-    /** @property @var string user name */
-    protected $user_id;
+    /** @property @var string sender's user id */
+    protected $sender_id;
```
MasterOdin: see discussion in previous comment. `user_id` marks who the email is going to, and recipient is a legacy field that currently holds the value of `user_email` for whoever `user_id` is.

PR #3908, opened 2019-06-24
[SYSADMIN ACTION][API] Authentication API

zjxiaohan:

### What is the new behavior?

Basic code for API, including:
- Generation of JWT (`/api/token` endpoint)
- Verification of JWT
- Redirection & Twig Rendering Prevention

Currently, if the user has a valid JWT, then they can go any endpoints. If the user has an invalid JWT, then they will be defined access. If they have empty `authorization` header, then they can only go to the `/api/token` endpoint.

### Other information?

Screenshot 1: Get Token Success with Empty Token in Header
Screenshot from 2019-06-24 20-13-31

Screenshot 2: Get Token Fail with Empty Token in Header
Screenshot from 2019-06-24 20-14-40

Screenshot 3: Wrong Method and/or Wrong Endpoint with Empty Token in Header
Screenshot from 2019-06-24 20-15-35

Screenshot 4: Invalid Token in Header (Server Refuses to Process)
Screenshot from 2019-06-24 20-16-07

Screenshot 5: Valid Token in Header (Server Responds Correctly)
Screenshot from 2019-06-24 20-16-44


site/app/libraries/Core.php
```diff
@@ -328,6 +338,23 @@ public function getSession(string $session_id, string $expected_user_id): bool {
...
 
+    /**
+     * Given an api_key (which should be coming from a parsed JWT), the database is queried to find
+     * a user id that matches the api key, and returning the user_id.
+     *
+     * @param string $api_key
+     *
+     * @return bool
+     */
+    public function getApiUser(string $api_key): bool {
```
MasterOdin: Let's call this `loadApiUser` as it doesn't return the User model, just a bool flag.

site/app/libraries/Core.php
```diff
@@ -372,6 +399,37 @@ public function authenticate(bool $persistent_cookie = true): bool {
...
+     *
+     * @return string | null
+     *
+     * @throws AuthenticationException
+     */
+    public function authenticateJwt() {
+        $user_id = $this->authentication->getUserId();
+        try {
+            if ($this->authentication->authenticate()) {
+                // Set the cookie to last for 7 days
```
MasterOdin: This comment should be removed as it's not relevant here.

MasterOdin:
@bmcutler Note the change that must be done to the apache configuration when installing this to production.

PR #3918, opened 2019-06-25
[Refactor:UI] Color Stylesheet for universal Submitty colors

### LINKED ISSUE #3812: CSS variables for Submitty standard colors
Let's establish CSS variables for all of the standard colors.

https://developer.mozilla.org/en-US/docs/Web/CSS/:root
(1 example at the top of our server.css)

Also...  please color merged threads the same color (grey) as deleted threads in the forum thread list.   Follow-on to issue #3628

thef1stofhername:
closes #3812


thef1stofhername:
Yeah that's seems valid! I think that's the right course forward. I'm going to go sifting through to find any colors I missed, and then start trying to implement it throughout. Thanks for taking a look!

MasterOdin:
For a point of reference of what should be in this file versus somewhere else, why is their variables for the discussion forum? Are these to be used just in forums.css or are we expecting it to be elsewhere? When should we define a variable in a section specific stylesheet as opposed to the global one?

thef1stofhername:
This file is supposed to help standardize the colors used throughout the Submitty website, so all color variables will be declared here. When they are used in other files, you simply need to call the name of the color you want. This will be used throughout the entire website and contains color variables that are specific to the forum, among others. As the file is just for colors, I don't think it will be touched too much, as people shouldn't be making new colors too often. However, any time a colors is needed ( and the colors in server.css will be standardized to these too), the variable should come from here. I hope this was helpful in clarifying any issues, please feel free to tell me if it doesn't. 
Kevin also does a great job of explaining the point of colors.css here:
https://github.com/Submitty/Submitty/issues/3812#issuecomment-500208073

thef1stofhername:
wow that would make so much more sense... I can fix that real quick

Issue #3828, opened 2019-06-11
IS_EMAIL_CONFIGURED & auto-email on instructor announcement


bmcutler:
**What problem are you trying to solve with Submitty**
Email configuration on a production installation should be optional.  
So if the sysadmin does not provide the necessary information for email configuration (username, sender, password, etc), this new "flag" is false.

From the web GUI, we should disable (not show to users) the options to receive email notifications, and the option to send exam seating assignment emails, and the code shouldn't try to send emails on grade inquiries, etc.

Also, let's combine the "announcement?" and "send email?" options for the instructor discussion forum new thread.  If email is configured, every announcement should send email.  And if not configured, that new thread page should only mention announcement (w/ notification) and not mention email.

PR #3922, opened 2019-06-25
email_enabled field and added configuration questions to CONFIGURE_SUBMITTY

### LINKED ISSUE #3431: Configure details of email in CONFIGURE_SUBMITTY.py
Right now, configuration of email is done manually. We should move this into the CONFIGURE script.

sak6lab:
CONFIGURE_SUBMITTY does not setup email info in email.json.

### What is the new behavior?
CONFIGURE_SUBMITTY asks users for the fields in the email.json and added a field is email enabled so sysadmin can toggle if they want emails to be sent and be shown as an option in the website (in next PR)(issue #3828)

closes #3431 

### Other information?

- run CONFIGURE_SUBMITTY & install_system
- checked /var/local/submitty/logs/emails/mailboxes/
- ran migration migrate and rollback


.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -146,7 +147,14 @@ def get_input(question, default=""):
             'institution_name' : '',
             'username_change_text' : 'Submitty welcomes individuals of all ages, backgrounds, citizenships, disabilities, sex, education, ethnicities, family statuses, genders, gender identities, geographical locations, languages, military experience, political views, races, religions, sexual orientations, socioeconomic statuses, and work experiences. In an effort to create an inclusive environment, you may specify a preferred name to be used instead of what was provided on the registration roster.',
             'institution_homepage' : '',
-            'timezone' : tzlocal.get_localzone().zone}
+            'timezone' : tzlocal.get_localzone().zone,
+            'email_enabled': 'true',
```
MasterOdin: Use true boolean type instead of a string representation of a boolean.

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -239,6 +247,35 @@ def get_input(question, default=""):
 
     CGI_URL = SUBMISSION_URL + '/cgi-bin'
 
+    print("NOTE: Emails can be set up at a later time in the config/email.json file")
+    while True:
+        is_email_enabled = get_input("Will Submitty use email notifications? [y/n]", 'y').lower()
+        if (is_email_enabled == 'y') :
```
MasterOdin: Let's handle yes/no here (`is_email_enabled.lower() in ['yes', 'y']`, `is_email_enabled.lower() in ['no', 'n']`) as that's somewhat standard in the unix world.

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -11,6 +11,7 @@
 import string
 import tzlocal
 import tempfile
+import readline
```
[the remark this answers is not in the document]
sak6lab: I added this because whenever the user tried to delete what they typed they would be stuck with ^H or the respective key.

sbin/send_email.py
```diff
@@ -87,7 +87,7 @@ def construct_mail_client():
         pass
     client.ehlo()
 
-    if EMAIL_USER is not None:
+    if EMAIL_USER != '' and EMAIL_PASSWORD != '':
```
bmcutler: I think we should add a check that EMAIL_ENABLED is true here also?

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -295,6 +331,13 @@ def get_input(question, default=""):
     config['institution_homepage'] = INSTITUTION_HOMEPAGE
     config['debugging_enabled'] = DEBUGGING_ENABLED
 
+    config['email_user'] = EMAIL_USER
```
bmcutler: The email information should only be written to email.json, we don't need this here.

PR #3926, opened 2019-06-26
[Bugfix:Migrator] Fix for destroying migrations for multiple courses

MasterOdin:

When destroying migrations for the course environment, only the first course in the list would actually be affected by the down function, with the rest just removing the migration history. This is due to the fact the migration file is removed after the first course runs it and all future courses then cannot find the missing file.

### What is the new behavior?
All courses are properly affected by the down function of the migration before it is deleted from the filesystem. The down function is also only run if the missing migration was actually run as well before it was deleted.

### Other information?

To test:
1. Have master at commit before 0d68614e5749738d4fc67a4436217b975d18ec93
1. host: `git checkout new_option_for_grading_2683`
1. guest: `submitty_install`
1. host: `git checkout master`
1. guest: `submitty_install`
1. host: `git checkout new_option_for_grading_2683`
1. guest: `submitty_install`

Where without this patch, the last step would throw an error on the second course (the first course would properly migrate). I also tested by inserting a `migrate -e course rollback` step between steps 3 and 4 on a separate run.


bmcutler:
ooh..  sneaky bug!

PR #3929, opened 2019-06-26
[API] Add Response object to handle Web/JSON responses

zjxiaohan:

A controller function either returns a view (generally with the help of twig templates) or returns a JSON.

### What is the new behavior?
An object called `Response` will be returned.

The `Response` object contains a `WebResponse` (could be `null`) and a `JsonResponse` (could be `null`). We can call `response->render()` to render its content in the following fashion:

- If there only exists one kind of non-null response, render it.

- If there exist two kinds of non-null responses, depending on the truth value of `$render` in `Output.php`, render either `WebResponse` or `JsonResponse`.

The `Response` class enables us to have API endpoints that reuse existing code while not breaking internal JSON responses.

#### Why can't we render both JSON and webpages?

Because they are sharing the output buffer. Please see an example in `api_course_example` branch. (https://github.com/Submitty/Submitty/commits/api_course_example)

You could also get the problem in the screenshot below:

Screenshot from 2019-06-26 14-46-30

#### Why can't we simply disable JSON in web access and enable it in API calls?

There are many internal functions that use JSON as a means of communication. Blocking all JSONs in web access will break them. Technically we could add additional parameters or something like flags in individual functions to make them work, but (I believe) it is not elegant enough.

#### Any example of the changes made in this PR?

Let's say the homepage will return two forms of data. If it is a web access, returns the homepage that we know. If we go to `/api/courses`, it returns a JSON containing course information (here, we use `TODO: add real data` as a placeholder).

**Update: we now use new constructors but the gist remains the same.**

The code would be like this:

Screenshot from 2019-06-26 18-56-20

Or more simply:

Screenshot from 2019-06-26 18-57-22

Then, if we go to the homepage, we will see the homepage normally as before. If we go to `/api/courses`, `$core->getOutput()->disableRender()` will be executed (this part of code will be added after #3908 is merged. you could simulate this line by decommenting line 321 in `index.php` of this branch), and we will get JSON!

View:

Screenshot from 2019-06-26 19-03-52

JSON:

Screenshot from 2019-06-26 19-02-09

#### Is this change backward compatible?

Yes.

We assume that in the new router, every call returns a `Response` object (new behaviour) or nothing (old behaviour). Before all responses are refactored, we use an `if` block to make it backward compatible. It looks like:

```
$router = new app\libraries\routers\WebRouter($request, $core, $logged_in);
$response = $router->run();

if ($response instanceof Response) {
    $response->render($core);
}

// else do nothing, because the content has been rendered in $router->run()
```


site/app/libraries/response/JsonResponse.php
```diff
@@ -0,0 +1,92 @@
...
+
+        if ($data !== null) {
+            $response['data'] = $data;
+        }
+        if ($code !== null) {
+            $response['code'] = $code;
+        }
+
+        return $response;
+    }
```
MasterOdin: So the intention here was that the static methods would return a `new JsonResponse` object. An example would be that:
```
class JsonResponse {
    function __construct($type, $data, $message, $code=null) {
        // implementation
    }

    // class implementation

    public static function getSuccessResponse($data) {
        return new JsonResponse('success', $data);
    }
}
```

zjxiaohan:
Finally passed Travis CI - Branch. Turned out to be a missing `$response` in the ClassicRouter.

PR #3931, opened 2019-06-26
This *should* fix the null error for the total score histogram and autograding score histogram

### LINKED ISSUE #3688: Simple plots of grade data for an assignment
The overview page for TA grading of a particular gradeable includes lots of statistics (overall average, per problem average, etc), but only in a table form.  Let's make some simple bar charts.  

Could use d3...  but plotly or chartly are probably simpler and will do the job.

Initially this will just be for the ta & instructors, but we could also allow an option to release this information to students (instructor choice).

note: I thought we already had a basic issue for this, but I can't find it now.

thef1stofhername:
Closes #3688 
Now the  histograms should display properly!
bugFix 2019-06-26 11-24

### DIFF
--- a/site/app/templates/grading/electronic/Status.twig
+++ b/site/app/templates/grading/electronic/Status.twig
@@ -59,10 +59,10 @@
     {% else %}
         <div class = "tab">
             <button class="btn btn-primary" type="button" onclick="openStat(event, 'Numerical Data'), this.blur()" id="defaultOpen">Numerical Data</button>
-            <button class="btn btn-primary" type="button" onclick="openStat(event, 'Total Score Histogram'), this.blur()">Total Histogram</button>
-            <button class="btn btn-primary" type="button" onclick="openStat(event, 'Autograding Score Histogram'), this.blur()">Autograding Histogram</button>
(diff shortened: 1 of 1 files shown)


site/app/templates/grading/electronic/Status.twig
```diff
@@ -316,7 +316,7 @@
                 {% endif %}
             </div>
             {# Graph Setup for Autograding #}
-            <div id="Autograding Histogram" class="tabcontent">
+            <div id="Autograding Score Histogram" class="tabcontent">
```
MasterOdin: These are both invalid ids as per defined by the [HTML 5 spec](https://www.w3.org/TR/html5/dom.html#element-attrdef-global-id), namely that an ID can be any string of characters except for no spaces.

Personally, I think all IDs should be kebab-case.

thef1stofhername: made a note to fix when I revisit with instructor feedback
I already have some tweaks to make. I am new to html and its specs, I will try to improve in future and apologize for the inconvenience.

MasterOdin: Another note to this is that while it's technically allowed to have `#` and `.` in a id identifier, it's really not suggested as you then have to remember to always escape them when referencing them in CSS/JS and that forgetting to escape will make it refer to potentially different elements/scope altogether.

PR #3935, opened 2019-06-26
double click to zoom out and fixed code to be compliant to HTML5 specs

thef1stofhername:
tiny PR to add a permanent message on how to zoom out from the zoom in (the message appears when you reach the page and zoom in for the first time in the top right)
changed variables to reflect html5 standards


site/app/templates/grading/electronic/Status.twig
```diff
@@ -451,24 +460,24 @@
                         {% endif %}
 
                         {% if team_assignment %}
-                            <br />
+                            <br/>
```
zjxiaohan: I am wondering why these `<br />`s are changed to `<br/>`s. They mean the same thing to me. Could I know your considerations?

PR #3936, opened 2019-06-26
[Feature:MobileUI] header/footer, new login/home, menu and sidebar

mktung:
- Reorganized global header and footing HTML structure for better semantics, more modular styling, and mobile compatibility
- Created new stylesheets for different twig files
- Mobile header displays home button, only the latest breadcrumb, and a menu / logout button depending on the page, scrolls on long titles
- Mobile menu, notifications over menu button, scrollable inside menu, clicking outside of menu exits
- Reworked the sidebar to work with the new mobile structure / maintain height on collapse
- New homepageand login page, which use new modular css
- Cleaned css that does not meet our guidelines

To trigger mobile display on the authentication page, I added viewport control in the twig file so it uses mobile styling on phones. This is currently not used on any other page because I want to manually ensure their functionality is maintained. It is bad form to have a meta tag inside the body like this, but I don't want to make mobile truly active on every page until we're sure it works. If someone significantly scales down their browser, the changes will become noticeable - but it shouldn't affect anyone on their phones yet.


mktung:
Finallly.

PR #3941, opened 2019-06-27
[Refactor] Move installation of 3rd party autograder dependencies to INSTALL_SUBMITTY_HELPER

MasterOdin:

The following 3rd party dependencies are installed as part of install_system.sh, requiring a new migration if ever we want to update them:
* JUnit
* Hamcrest
* TCLAPP
* Emma
* JaCoCo
* DrMemory

### What is the new behavior?
Moves their installation inside of INSTALL_SUBMITTY_HELPER.sh so that any changes to their versioning will not require a migration.

This also brings Travis to use the same installation/update mechanisms for these dependencies as production Submitty, ensuring that we use the same version on both.

### Other information?

1. vagrant destroy/vagrant up
1. Editing the version.json into following possible conditions:
- version.json does not exist
- version.json exists, but is empty
- version.json exists, but is missing some keys

After each, I ran `submitty_install test` to run the test suite which verifies all pieces except for TCLAPP/emma.


bmcutler:
I am not in favor of this change.

INSTALL_SUBMITTY.sh is run by some of us developers many times a day.  We should not be making any changes to this file that unnecessarily slows down this process.  When it was originally created the goal of this script (hence the name) was to install the new submitty code to the right areas of the server.  It is not for upgrading the system (less frequently changing 3rd party stuff).  This script should not require internet access (allowing development when offline).

I've always pushed to ensure that install_system.sh can be safely re-run on an existing system.  That should be the thing that upgrades 3rd party packages (that are conveniently handled by a migration).

Most of the packages you've target in here should be moved to docker containers when we finally, eventually move most of autograding to docker containers.

MasterOdin:
> INSTALL_SUBMITTY.sh is run by some of us developers many times a day. We should not be making any changes to this file that unnecessarily slows down this process.

This only slows down things on the first run with these changes (as it pulls down the dependencies, and records their versions) and then all subsequents runs will be no slower than it currently is. I could write a migration to make the first run of this on an existing Submitty system make no time as well if that's truly a big concern.

> This script should not require internet access (allowing development when offline).

Assuming you don't try and change versions of any of these dependencies in versions.sh, no internet access is required once they're installed the first time.

> I've always pushed to ensure that install_system.sh can be safely re-run on an existing system. 

That should be the thing that upgrades 3rd party packages (that are [edit] not conveniently handled by a migration).

This feels like this begs for a more philosphical discussion as to what the different types of dependencies are (to me, we've got apt dependencies, language dependencies, and autograder dependencies). Where should these be installed? Currently, this is split across install_system.sh and SUBMITTY_HELPER.sh with no clear distinction. Personally, I'd like to move everything but apt packages under SUBMITTY_HELPER.sh as I view them as things that are going to have to be potentially upgraded in their lifetime to keep Submitty as a whole actually functioning (see all of the commits by dependabot for example).

Currently, this includes:
1. Submitty Repos
1. PHP packages (via composer)
1. Frontend JS packages (via NPM)
1. nlohmann json (on first run of INSTALL_SYSTEM)

I'd also like to move all of the pip installations in install_system.sh under INSTALL_SUBMITTY as well to match how we're doing PHP/JS, as otherwise, we risk being burned here (I'm surprised we haven't already to be honest) on version updates of packages.

But this of course doesn't gel with your vision of how the scripts are, and we should probably reconcile on one set view. 

I will say, the migration system (to me) should not be used to keep these non-apt dependencies up-to-date when (imo) there's better ways to do it automatically just based on version strings (e.g. through this PR, I can now just edit one string in one file and update DrMemory to v2.1 instead of both writing a migration (copying and pasting `migration/migrator/migrations/system/20180829105219_update_drmemory.py`) and updating the string in the file, even though the migration itself is easy.

> Most of the packages you've target in here should be moved to docker containers when we finally, eventually move most of autograding to docker containers.

I don't disagree with this, and if the consensus is that we're unlikely to want to upgrade JaCoCo or DrMemory before we move to containers, then this change is probably largely irrelevant in the longterm and I'd agree should be closed. However, we do need to probably resolve the above question.

bmcutler:
this will be addressed in other PRs tackling issue #4011

PR #3949, opened 2019-06-27
Made skip nav

### LINKED ISSUE #3912: Skip Navigation -- quick tab/enter option
keyboard/screenreader users often want to quickly skip the sidebar (esp when it contains > 10 menu items) and jump quickly into the interesting items within the content portion of the page.

Let's make the first item on the side bar be 'skip navigation'.
But let's have it be hidden/not visible, EXCEPT when it has focus. 
When it has focus, it should be a visible option, probably overlaid on top of the side bar, towards the top.  

When you first load the page it's not visible (nothing on the page has focus yet).
When you hit tab the first time, the skip navigation now has focus, so it's visible.  
If you hit tab a second time, focus moves to the first thing on the side bar, skip navigation disappears again.
Or instead if you have focus on skip navigation and press enter, focus moves to the first content link within the page.

mktung:
Closes #3912

Made a skip to main content button that becomes visible to non-screen-reader users on focus.

### DIFF
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -20,6 +20,7 @@
 </head>
 <script>var onAjaxInit;</script>
 <body data-site-url="{{ site_url }}" data-base-url="{{ base_url }}" data-course-url="{{ course_url }}" data-csrf-token="{{ csrf_token }}" onload="if (onAjaxInit) { onAjaxInit(); }">
+<a id="skip-nav" href="#nav-positioner">Skip to main content</a>
 <div id="container">
     {% if wrapper_urls['top_bar.html'] != null %}
         {# uploaded homepage redirect can go here? #}
--- a/site/public/css/server.css
+++ b/site/public/css/server.css
@@ -1995,3 +1995,25 @@ end of styles used in the admin gradeable page
     border: 0px;
     padding: 0px;
 }
+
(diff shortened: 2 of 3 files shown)


site/app/templates/GlobalHeader.twig
```diff
@@ -77,6 +77,7 @@
             </div>
             {% endif %}
             <div class="row" id="push">
+                <a id="skip-nav" href="#nav-positioner" tabindex="1">Skip to main content</a>
```
pbailie: Use of `tabindex` greater than zero is nearly always advised against in web accessibility standards.  q.v. https://webaim.org/techniques/keyboard/tabindex

Please remove the `tabindex` property and ensure tab order by logical placement of the link in HTML.

bmcutler:
Yay!  I like this.  But there's a hiccup on the discussion forum page.  I think it's because that page (currently, as of PR #3859) starts with focus on the breadcrumb instead of requiring a tab to get there.

So please, @mktung  @xprilion coordinate who will get skip nav working on the forum page. 

Thanks

site/app/templates/GlobalHeader.twig
mktung: I've made the changes

mktung:
@xprilion I removed a line in the forum js that autofocused the page to the breadcrumb on load. If this was crucial, please let me know.

PR #3955, opened 2019-06-28
[Refactor] Friendly Error Page

### LINKED ISSUE #3927: Make "permission denied" error pages inline to the site
When you go to a page that you don't have access to, you're greeted with a variant of the following page:

Screenshot 2019-06-25 23 46 48

These pages only happen for actions that occur when navigating somewhere within a course.

These pages should be styled "inline" to the site such that it does not break a user's navigation and that they can navigate "back" in a normal fashion. An example of what I'd like to see is the error page you get when you go to a course that you don't have access to:

Screenshot 2019-06-25 23 48 40

wherein a user can still see links on the sidebar, header text, etc.

**Additional context**
The error message from the first screen can/should probably be copied wholesale into what is shown in the second screen for this kind of error with the exception of the listing of courses.

For example:
Screenshot 2019-06-25 23 53 17

zjxiaohan:

Closes #3927 

### What is the new behavior?
Screenshot from 2019-06-28 14-46-54

### Other information?
Besides changing the render part, this PR also:
- Removes `InvalidPage.twig`. It is not used anywhere.
- Puts all error view functions into `ErrorView`.


site/app/templates/error/ErrorPage.twig
```diff
@@ -0,0 +1,6 @@
+<div class="content">
+    You don't have access to this page. <br />
+    Reason: <b>{{ error_message | replace({' ': '&nbsp;'}) | raw | nl2br }}</b><br />
+    Please contact system administrators if you believe this is a mistake.<br />
+    Click <a href = "{{ main_url }}"> here </a> to back to homepage and see your courses list.
+</div>>
```
MasterOdin: Extra `>`

PR #3956, opened 2019-06-28
[Bugfix] Improve router error handling

zjxiaohan:

c
Currently, if `student` is removed from the course `blank`, go to `http://192.168.56.111/index.php?semester=s19&course=blank` and there will be an exception:

```
Symfony\Component\Routing\Exception\NoConfigurationException (Code: 0) thrown in /usr/local/submitty/site/vendor/symfony/routing/Matcher/UrlMatcher.php (Line 93) by:
            throw new NoConfigurationException();
```

If `student` is in the course `sample`, go to `http://192.168.56.111/index.php?semester=s19&course=sample` and there will be a message intended for API.

```
{ "status": "fail", "message": "Endpoint not found." }
```

The **CORRECT** behaviour should be: if there are links that the new router does not recognize, and it goes through the new router, then the user should be redirected to either login or home, depending on if they are logged in or not. (API's case is different. We will return a fail JSON.)

### What is the new behavior?

Going to `http://192.168.56.111/index.php?semester=s19&course=blank` or `http://192.168.56.111/index.php?semester=s19&course=sample` all takes us to the home page.

### Other information?
The root of this problem is that if there is course-semester information in the URL, our old router will load the course. Unfortunately, the new router redirects users depending on if the course is loaded or not. This causes:

1. `index.php`: Hey I see semester and course in the URL and I am loading the course!
2. `WebRouter.php`: It seems everything works good because the course is loaded, meaning the URL is matched to a correct semester and course and components etc. Let's go!
3. `Submitty`: Oh no! Something irrecoverable has happened...


site/app/libraries/routers/WebRouter.php
```diff
@@ -83,14 +85,6 @@ public function __construct(Request $request, Core $core, $logged_in, $is_api =
     }
 
     public function run() {
-        if (is_null($this->parameters)) {
```
zjxiaohan: They are moved to the API part for clarity - These JSON responses are for API only.

site/app/libraries/routers/WebRouter.php
```diff
@@ -116,10 +110,17 @@ private function loadCourses() {
...
             /** @noinspection PhpUnhandledExceptionInspection */
             $this->core->loadConfig($semester, $course);
+            $this->course_loaded = true;
         }
     }
 
     private function loginCheck() {
+        // This is a workaround for backward compatibility
+        // Should be removed after ClassicRouter is killed completely
+        if ($this->core->getConfig()->isCourseLoaded() && !$this->course_loaded) {
```
MasterOdin: Let's add:
```
if ($this->core->isDebug()) {
    throw new \RuntimeException("Attempted to use router for invalid URL. Please report the sequence of pages/actions you took to get to this exception to API developers.");
}
```

just so internally we keep open a valuable way to force developers to tell us about potential various redirects or URL constructions that we might have missed as they'll get an exception page instead of just a silent redirect.

site/app/libraries/routers/WebRouter.php
```diff
@@ -62,11 +62,13 @@ public function __construct(Request $request, Core $core, $logged_in, $is_api =
...
                     !Utils::endsWith($this->parameters['_controller'], 'AuthenticationController')) {
-                    $this->api_authorized = false;
+                    $this->core->getOutput()->renderJsonFail("Unauthorized access. Please log in.");
+                    die($this->core->getOutput()->getOutput());
                 }
             }
             catch (ResourceNotFoundException $e) {
-                $this->parameters = null;
+                $this->core->getOutput()->renderJsonFail("Endpoint not found.");
+                die($this->core->getOutput()->getOutput());
```
MasterOdin: Might not need to do it in this PR (probably better once we kill off the ClassicRouter for good), but we should try to make sure we're not inserting dies into code since it means we cannot really write a test now for this path which is not optimal. (I regret the die that exists in the Output class and look forward to the recent round of changes getting us closer to where we can probably remove it).

zjxiaohan: Thanks for your suggestion. `die` is indeed bad for testing. After we have `Response` probably we can better handle that by having a bad response for both JSON and HTML (currently I move it here because I think it is for API only). I will remove `die`s next time. Let there be `die` in `redirect` only.

MasterOdin: For reference, I'd like to potentially think of adding a `RedirectResponse` class as well so that we can remove the die from the redirect function as well.

Issue #3877, opened 2019-06-18
date picker:  Add quick button to "Now" and "Infinity" (or "never")


bmcutler:
From the calendar/time date picker (use in various places on the site), we'd like to have a button for "Now", that sets the calendar date & time to the current moment.

And we'd also like an "Infinity" or "Never" option that sets the calendar date & time to year a date in 9999  (but don't get too close to rolling over to year 10000 with UTC->local timezone conversions! a previous bug!)

PR #3963, opened 2019-06-29
[Feature:InstructorUI] added shortcut buttons for datepickers

Fenn-CS:

The calendar/time date picker used in various places on the site, does not have "Now"/"Infinity" button that can be easily used to set the time to those respective times.

Fixes: #3877 

### What is the new behavior?

Calendar now includes "Now" and "End of time" buttons.

now_and_end_of_time

_This PR continues from #3901 and all reviews have been incoporated_


site/app/templates/admin/Extensions.twig
```diff
@@ -58,22 +58,47 @@
...
+                    }
+                )],
             allowInput: true,
             enableTime: false,
             enableSeconds: false,
             time_24hr: true,
             dateFormat: "Y-m-d",
-            onChange: function(selectedDates, dateStr, instance) {
-                var due_date_value = document.getElementById("due_date").innerHTML.substr(0,10);
+            onChange: function (selectedDates, dateStr, instance) {
```
zjxiaohan: In the extensions page, when I click `Now` or `End of time` in the `Or Pick a New Date` box, the content of `Give the Days of Extensions` is not automatically calculated. I guess it has something to do with this `onChange` function. Needs to fix because the query to update extension dates uses the number in `Give the Days of Extensions` in effect.

That being said, I doubt if we need `now` or `infinity` in the extension page. Both do not make much sense. @bmcutler  what do you think?

Fenn-CS: Thanks @zjxiaohan lets here what Barb has to say about this, perhaps that could be taken off completely for the extensions page.

bmcutler: Now does make sense for this page.  Often excused absences are entered after-the-fact.  So now could be exactly what we want in those cases.  infinity makes less sense perhaps, but it does allow the instructor to say -- just turn it in before I submit prepare final/semester grades.  Do we have a potential overflow problem since the date is converted into a number of late days allowed?  We might need to clamp a value to prevent overflow?  I'd prefer to keep the UI standard across the site, so leave it.

Can we share code?  I see the same block of code copied & pasted in  4 places here.  Can we make a function/macro?

zjxiaohan: Thanks Barb for introducing how instructors are using it. I am not sure if there will be overflow, but it makes sense to clamp the value because we store the number of days instead of a timestamp.

Also @Fenn-CS I believe auto-calculating `Given the Days of Extensions` is still a problem (the number does not appear after selecting `now` or `infinity`). Could you verify that in your browser?

bmcutler:
@Fenn-CS 
please fix merge conflicts
thanks

site/app/templates/admin/Extensions.twig
Fenn-CS: @zjxiaohan that's true, and I am adding a fix for that.

Fenn-CS:
@zjxiaohan  your concern has been handled please confirm, @bmcutler merge conflicts have been resolved as well.

PR #3966, opened 2019-06-30
[Bugfix] More bulk assign fixes

shailpatels:
The bulkupload job in jobs.py corrects the file permissions of split paths after the entire job is done. If an instructor tries to submit or delete a split pdf before the permissions are corrected it'll throw an error. 

If a file processed 1 split pdf and an instructor submitted or deleted it right away, the controller would check if there were anymore pdfs since the job only processed 1 there would be none and Submitty would wrongly think this is the last pdf being assigned and delete the entire split folder which would cause the job to error and die. I removed this check for now but a better method to delete the split folder when everything is done is still needed.

Since the JSON file containing all the decoded information could be read before a job writes to it we need to check if the php key exists first, this didn't cause any issues but would throw php errors. 

EDIT: Since delete all is asynchronous, the pasge refresh at the end would kill some ajax calls which fail to delete certain split pdfs, now the page removes split pdfs from the user interface without a page reload.

### DIFF
--- a/sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
+++ b/sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
@@ -245,6 +245,14 @@ class BulkUpload(CourseJob):
             if not os.path.isfile(os.path.join(split_path, filename)):
                 shutil.copyfile(os.path.join(bulk_path, filename), os.path.join(split_path, filename))
 
+            # reset permissions just in case, group needs read/write
+            # access so submitty_php can view & delete pdfs when they are
+            # assigned to a student and/or deleted
+            self.add_permissions_recursive(split_path,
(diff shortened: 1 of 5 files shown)


bmcutler:
1.
I'm getting this error when I try to submit a pdf for a student who already has a submission:

Screen Shot 2019-06-30 at 10 29 57 PM

But it looks like the new submission went through.


2.  When I press "Delete All Submissions" it asks if I'm sure I want to delete 3 submissions.  I say yes, but then it only seems to delete the first one.  The other 2 are still there.  But I like how the page refreshes.

shailpatels:
@bmcutler I couldn't reproduce the error for submitting with previous submission (I created a new gradable and submitted multiple times for a student, tried all 3 options too). This PR fixes Fridays issues which are more important I think to get merged first.

I'll look into the issue with multiple submissions, this might be a bug on the PHP side in the controller.

PR #3980, opened 2019-07-02
[Refactor:Router] Remove usage of "die" in WebRouter

zjxiaohan:

To display error messages, sometimes WebRouter has to `die` immediately.

### What is the new behavior?
Now API returns nice error JSON messages as `Response`.

Also added:
- `RedirectResponse`
- CSRF Token Check

### DIFF
--- a/site/app/libraries/routers/WebRouter.php
+++ b/site/app/libraries/routers/WebRouter.php
@@ -3,8 +3,12 @@
 
 namespace app\libraries\routers;
 
+use app\libraries\response\Response;
+use app\libraries\response\JsonResponse;
+use app\exceptions\AuthenticationException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Exception\MethodNotAllowedException;
 use Symfony\Component\Routing\Exception\ResourceNotFoundException;
 use Symfony\Component\Routing\RequestContext;
 use Symfony\Component\Routing\Matcher\UrlMatcher;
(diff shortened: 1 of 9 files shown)


site/app/libraries/routers/WebRouter.php
```diff
@@ -84,7 +82,49 @@ public function __construct(Request $request, Core $core, $logged_in, $is_api =
...
+            !$this->core->checkCsrfToken()
+        ) {
+            $msg = "Invalid CSRF token.";
+            $this->core->addErrorMessage($msg);
+            return new Response(
+                JsonResponse::getFailResponse($msg),
+                null,
+                new RedirectResponse($this->core->buildNewUrl())
+            );
+        }
```
MasterOdin: I'm not sure we do a redirect on any and all CSRF token failures?

site/tests/app/libraries/routers/WebRouterTester.php
```diff
@@ -136,4 +138,34 @@ public function randomUrlProvider() {
...
+            'message' => "Endpoint not found."
+        ], $router->run()->render());
+    }
+
+    public function testApiWrongMethod() {
+        $core = $this->createMockCore(['semester' => 's19', 'course' => 'sample']);
+        $request = Request::create(
+            "/api/token"
+        );
+        $this->expectException(MethodNotAllowedException::class);
```
MasterOdin: What method here throws this exception? I'm not sure I follow the logic for how this gets raised?

zjxiaohan: It is thrown by the matcher of symfony (located in the constructor) when the resource was found but the request method is not allowed.

site/app/libraries/routers/WebRouter.php
zjxiaohan: Sometimes in our current logic the page is redirected to the previous one. However, I assume that it is a very rare case where the website is running correctly and CSRF token check fails. I assume that this only happens when someone tries to attack Submitty. Therefore it looks okay to me to always redirect home.

Not sure if my assumptions make sense.

site/tests/app/libraries/routers/WebRouterTester.php
MasterOdin: But you catch it internally and set a response object instead right?

zjxiaohan: Yes

zjxiaohan: Well it looks that it will make more sense if the test method calls the static method. I'll change it now.

site/app/libraries/routers/WebRouter.php
MasterOdin: Yeah, I don't think this particular failure happens every often in the real world, that it's not really worth worrying about the redirect in and of itself. I'm not sure we really handle more common form failures all that gracefully anyway that there's not a good way to really address what should happen here anyway.

PR #3981, opened 2019-07-02
Changed title scheme

### LINKED ISSUE #2169: Page title
Currently the page title is always _just_ the course name. We may also want to include the current page and/or other information

mktung:
Closes #2169 

Edited title formatting to reflect page and course

### DIFF
--- a/site/app/views/GlobalView.php
+++ b/site/app/views/GlobalView.php
@@ -1,6 +1,7 @@
 <?php
 
 namespace app\views;
+use app\models\Breadcrumb;
 
 class GlobalView extends AbstractView {
     public function header($breadcrumbs, $wrapper_urls, $sidebar_buttons, $notifications_info, $css=array(), $js=array()) {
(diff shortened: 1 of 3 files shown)


mktung:
@bmcutler The reason it's not listing the name of the gradeable is because we don't add those to our breadcrumbs. There are actually a lot of pages without a descriptive breadcrumb; should we add those special cases to just the title, add them to the breadcrumb list, or leave it for another issue?

bmcutler:
We actually don't want to add anything extra to the breadcrumb on the rubric for a single student.  It would be most natural to add the name of the student -- and useful in many cases, but we also want to be able to grade (semi) anonymously, so we chose not to put the name in the breadcrumb.  Anything additional in the breadcrumb would just be unnecessary bulk.  

So I take back my request, let's just leave the title as you've got it for now.

But which other pages are lacking good breadcrumb info?

mktung:
Looks like I was confusing myself by looking at the sidebar rather than the breadcrumb. I think grading is the only place we don't have a full breadcrumb trail.

PR #3983, opened 2019-07-02
Made rubric edit shortcut

mktung:
Refers to a checkbox in #2680 

Added a user editable keyboard shortcut to edit rubrics


bmcutler:
Hmmm.... For hotkeys on the grading page, it's probably ok that the cookies aren't tied to a specific user.  (It's unlikely two TAs are going to use the same machine).  I hope we're not elsewhere storing any significant/sensitive data in a cookie that could get mixed up between users.

PR #3990, opened 2019-07-03
[Router] New URL for TeamController

zjxiaohan:
### URL Changes

**Old:** 
- `/index.php?semester=s19&course=development&component=student&gradeable_id=open_team_homework&page=team`
- `/index.php?semester=s19&course=development&component=student&gradeable_id=open_team_homework&page=team&action=create_new_team`
- `/index.php?semester=s19&course=development&component=student&gradeable_id=open_team_homework&page=team&action=leave_team`
- `/index.php?semester=s19&course=development&component=student&gradeable_id=open_team_homework&page=team&action=invitation`
- `/index.php?semester=s19&course=development&component=student&gradeable_id=open_team_homework&page=team&action=accept`
- `/index.php?semester=s19&course=development&component=student&gradeable_id=open_team_homework&page=team&action=cancel`
- `/index.php?semester=s19&course=development&component=student&gradeable_id=open_team_homework&page=team&action=seek_team`
- `/index.php?semester=s19&course=development&component=student&gradeable_id=open_team_homework&page=team&action=stop_seel_team`

**New:** 
- `/s19/development/open_team_homework/team`
- `/s19/development/open_team_homework/team/new`
- `/s19/development/open_team_homework/team/leave`
- `/s19/development/open_team_homework/team/invitation/new`
- `/s19/development/open_team_homework/team/invitation/accept`
- `/s19/development/open_team_homework/team/invitation/cancel`
- `/s19/development/open_team_homework/team/seek/new`
- `/s19/development/open_team_homework/team/seek/stop`

### Tested Entrances
Click the button on the navigation page to go to the team page. All team operations manually tested.


site/app/controllers/student/TeamController.php
```diff
@@ -3,42 +3,44 @@
...
-use app\libraries\Core;
 use app\libraries\FileUtils;
+use Symfony\Component\Routing\Annotation\Route;
 
 class TeamController extends AbstractController {
     public function run() {
         switch ($_REQUEST['action']) {
             case 'create_new_team':
-                $this->createNewTeam();
+                $this->createNewTeam($_REQUEST['gradeable_id']);
```
shailpatels: Should you be passing in `$_REQUEST['gradeable_id']` into the function calls? What would happen if `$_REQUEST['gradeable']` is null or the gradeable_id passed through the request is invalid?
I think it would be safer to handle this inside of each function and return the appropriate response unless there's an explicit reason why this is set up the way it is.

site/app/templates/submission/Team.twig
```diff
@@ -99,12 +99,12 @@
 
         {# Create new team button #}
         <br />
-        <button class="btn btn-primary" onclick="location.href='{{ core.buildUrl({'component': 'student', 'gradeable_id': gradeable.getId(), 'page': 'team', 'action': 'create_new_team'}) }}'">Create New Team </button>
+        <button class="btn btn-primary" onclick="location.href='{{ create_url }}'">Create New Team </button>
```
shailpatels: Could we change this to `create_team_url`

site/app/controllers/student/TeamController.php
zjxiaohan: This is for backward compatibility to avoid breaking existing testing code if they rely on `run`. `run` is not used in actual routing process. Hmm it's actually safe to remove them completely.

zjxiaohan:
> I was wondering why the run function was completely removed? Are functions no called controlled by the _REQUEST[] switch statement. Not asking for changes just curious why this controller is different than the rest.

The new router calls controller methods directly after a match. It does not go through `run` function as it did before. Therefore, all `run` functions can be safely removed after having new URLs in annotations.

I did not remove `run` before because once we came across a test that relies on `run` to test the controller, which failed due to router changes. I was thinking about removing them all in a "clean up" action after we have all routes covered. We cannot remove single `run`s now because it is defined in the abstract class.

shailpatels:
@zjxiaohan ah ok that makes sense, thanks!

PR #3994, opened 2019-07-03
[Feature:TAGrading] undo/redo for checkpoint gradeables

### LINKED ISSUE #3101: limited scope undo/redo on checkpoint grading page
Use case:  TA/mentor grader using Submitty website on a phone to check off students in lab.   Grader wants to scroll to find the student, but accidentally clicks on a checkpoint, changing the data.  Currently its not clear which box was just changed.  And especially if the page data is stale (it may have been 10 minutes since the grader hit reload, and other graders are also working on this page).

Let's add an undo/redo option!

Only a small amount of history needs to be stored.  Just a few clicks (maybe up to 5).    Maybe we can also highlight those boxes with a "recently edited by you" outline color?  This could just be stored in a cookie?  Store which student, which checkpoint, old value, new value for each action.  And somewhere on the page is an undo button, that just reverts those changes.  And if you've undone, then there can also be a redo?

mktung:
Closes #3101 

Made a cookie that keeps track of checkpoints. The cookie is removed on logout to prevent grading information from escaping and there are different cookies for each gradeable. The undo/redo buttons follow the search bar by sticking to the top of the page on scroll. The undo/redo should be able to go backwards 5 steps.


mktung:
@jmcknight98 We've decided now that we only want to go back 5 steps; the other bug has been fixed.

Issue #3790, opened 2019-06-04
Improve UI for New Student / New Grader


bmcutler:
**What problem are you trying to solve with Submitty**

* The layout of the textboxes for the ids & names & email for new student/new grader popup could be improved/reorganized.   Also the boxes that are optional (numeric id, preferred first & last name) should be labeled as optional.  Also when #3561 is finished, email will be optional too (but strongly recommended for 'real' users so they get announcements & notifications).
 
* If a user is already in the master database, we really don't need to fill out most of the other fields (I think whatever we enter in these boxes is ignored!).  If you're manually adding someone to the course and you know they are already in the master database it is annoying to have to fill out the extra boxes.  It would be great if once we typed in a valid, exists-in-master-database user id, the other boxes on the form (first/last name, email) were pre-populated.   Then we could finish the course specific information (registration section, staff group, etc.)

* Also (edited to add), if you make an error entering the information (e.g., enter an invalid username with spaces or something), you get a red error popup (with a useful message) but the popup closes and you lose all the information you typed in and have to start over.

bmcutler:
@mktung As you are re-working some of the Instructor UI pages, please coordinate with @drumilpatel2000 to not conflict on this page :)

bmcutler:
INSTRUCTOR FEEDBACK:

Let's make sure this UI is more self-documented 
and difference between user id (mandatory)
and numeric id (optional, must be digits only) is clear.

PR #3999, opened 2019-07-04
[Feature:InstructorUI] New edit grading registration sections button

### LINKED ISSUE #3953: Grid of checkboxes for assigning graders to grading registration sections
Currently, assigning graders to grading registration sections can only be done through the 'manage graders' page, by clicking on one grader at a time, and then assigning them to 0,1,or more grading registration sections.  This is time consuming (lots of clicking through many popups) when you have a large number of graders in your course.

Let's add a second UI for of assigning sections, for everyone at once.  Let's add a new button in the upper right of the 'manage graders' page named "Assign Registration Sections for Grading" (yes, rather lengthy title, we can design discuss a shorter label)

When we click that button, we get a popup/modal that shows a grid of checkboxes that looks similar to the grid of checkboxes for assigning rotating sections per gradeable (grader assignment tab under create/edit gradeable).  One column for each registration section (with a shortcut 'all' checkbox that checks all sections for that row).  One row for each grader.  The graders should be grouped by type (instructor/fullaccess/limited access) and then alphabetical.

jmcknight98:
Closes #3953

Currently an instructor has to edit each grader's assigned registration section one by one.
The manage graders page is sorted by User ID only.
aaPrBefore

### What is the new behavior?
The manage graders page is sorted by User Group and User ID, and a button has been added as described in the issue closed.
aaPrAfter

Also fixed a bug that did not allow for a "Preferred First/Last Name" to be removed once one was added in the edit user modal.


bmcutler:
Note: We should be cautious about the edits to the preferred first name, etc.
That extra logic may be important in cases where we are adding an existing user to a course.
Maybe hold off on this work for issue #3790

bmcutler:
@jmcknight98 
Can you please address the flake8 python lint errors?
Thanks!

Issue #3626, opened 2019-05-01
Course materials/student photos upload max size should detect apache limit


bmcutler:
**Describe the bug**
I believe the default apache upload limit is 10M.  But we typically override that limit on our production machine to allow students to upload (somewhat) larger files if the instructor has specified a larger submission limit for an assignment.

However, the course materials upload is (I believe) hard-coded to limit to 10M, even if the apache limit is raised.  Let's make this feature check apache and give an error message only if we exceed the apache limit on the system.

This also might be the case for instructor upload of student photos.

MasterOdin:
I'm not sure we've got apache limiting what we can upload, but rather it's more likely the php.ini value of [post_max_size](https://www.php.net/manual/en/ini.core.php#ini.post-max-size) or [upload_max_filesize](https://www.php.net/manual/en/ini.core.php#ini.upload-max-filesize) which are things we can detect using [ini_get](https://www.php.net/manual/en/function.ini-get.php).

bmcutler:
see also:
https://submitty.org/sysadmin/system_customization#allowing-large-student-file-upload-submissions

PR #4003, opened 2019-07-05
[Router] New URL for ImagesController

zjxiaohan:
**The changes are based on PR #3861**

### URL Changes

**Old:** 
- `/index.php?semester=s19&course=development&component=grading&page=images&action=view_images_page`
- `/index.php?semester=s19&course=development&component=grading&page=images`
- `/index.php?semester=s19&course=development&component=student&page=submission&action=upload_images_files`

**New:**
- `/s19/development/student_photos`
- `/s19/development/student_photos/upload`

### Tested Entrances
- Sidebar -> Student Photos (As Instructor)
- Direct Access `/s19/development/student_photos` (As Student & Instructor)
- Redirection after Uploading a Student Photo (As Instructor)


site/app/controllers/grading/ImagesController.php
```diff
@@ -48,4 +54,139 @@ public function viewImagesPage() {
...
+                    if(FileUtils::isValidFileName($uploaded_files[1]["name"][$j]) === false) {
+                        return $this->core->getOutput()->renderResultMessage("Error: You may not use quotes, backslashes or angle brackets in your file name ".$uploaded_files[1]["name"][$j].".", false);
+                    }
+                    $uploaded_files[1]["is_zip"][$j] = false;
+                    $file_size += $uploaded_files[1]["size"][$j];
+                }
+            }
+        }
+
+        $max_size = 10485760;
```
xprilion: Is it correct to add a file size limit here? This should probably be somewhere in the configuration.

site/app/controllers/grading/ImagesController.php
```diff
@@ -48,4 +54,139 @@ public function viewImagesPage() {
...
+    /**
+     * @Route("/{_semester}/{_course}/student_photos/upload")
+     */
+    public function ajaxUploadImagesFiles() {
+        if(!$this->core->getUser()->accessAdmin()) {
+            return $this->core->getOutput()->renderResultMessage("You have no permission to access this page", false);
+        }
+
+        if (empty($_POST)) {
+            $max_size = ini_get('post_max_size');
```
xprilion: Could this be kept outside and later used instead of hardcoding on line 124?

site/app/controllers/grading/ImagesController.php
```diff
@@ -48,4 +54,139 @@ public function viewImagesPage() {
...
+                    if (isset($uploaded_files[1]["error"][$j])) {
+                        $error_message .= "Error message: ". ErrorMessages::uploadErrors($uploaded_files[1]["error"][$j]). ".";
+                    }
+                    $errors[] = $error_message;
+                }
+            }
+        }
+
+        if (count($errors) > 0) {
+            $error_text = implode("\n", $errors);
```
xprilion: If this is getting rendered on the client side, probably `\n` should be made into `<br>`

site/app/controllers/grading/ImagesController.php
```diff
@@ -48,4 +54,139 @@ public function viewImagesPage() {
...
+
+        $max_size = 10485760;
+        if ($file_size > $max_size) {
+            return $this->core->getOutput()->renderResultMessage("File(s) uploaded too large.  Maximum size is ".($max_size/1024)." kb. Uploaded file(s) was ".($file_size/1024)." kb.", false);
+        }
+
+        // creating uploads/student_images directory
+
+        $upload_img_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "student_images");
+        if (!FileUtils::createDir($upload_img_path)) {
```
xprilion: You might want to specify an explicit permissions string too for security purpose

site/app/controllers/grading/ImagesController.php
```diff
@@ -48,4 +54,139 @@ public function viewImagesPage() {
...
+        }
+
+        if (isset($uploaded_files[1])) {
+            for ($j = 0; $j < $count_item; $j++) {
+                if ($uploaded_files[1]["is_zip"][$j] === true) {
+                    $zip = new \ZipArchive();
+                    $res = $zip->open($uploaded_files[1]["tmp_name"][$j]);
+                    if ($res === true) {
+                        //make tmp folder to store class section images
+                        $upload_img_path_tmp = FileUtils::joinPaths($upload_img_path, "tmp");
```
xprilion: Is this folder accessible from public?

site/app/controllers/grading/ImagesController.php
zjxiaohan: No. They are different.

`post_max_size` specifies the maximum size of things that PHP can handle. If a POST sends things bigger than this size, then PHP will fail to receive them. Instead, the POST field will be empty.

PHP's max size is different from what we want for the max size for one single image file.

site/app/controllers/grading/ImagesController.php
zjxiaohan: Good catch, but this message is also part of JSON. So `\n` is more proper here. I would add an `nl2br` on the twig side. Could you have a look at #4006?

site/app/controllers/grading/ImagesController.php
zjxiaohan: https://github.com/Submitty/Submitty/issues/3626 may be related to this. I am not sure about the current status of that issue though.

site/app/controllers/grading/ImagesController.php
zjxiaohan: Yes they should probably not be here. After looking into it, I found it is a bit more complicated than several lines' changes. For this PR, I would like to keep it for router changes only. Please see #4008 for further discussion.

site/app/controllers/grading/ImagesController.php
zjxiaohan: Don't worry. `umask` is `0027`.

site/app/controllers/grading/ImagesController.php
zjxiaohan: No.

PR #4005, opened 2019-07-05
[Feature] Annotation based access control

### LINKED ISSUE #3965: Easier access control through annotation
Currently we need to manually perform access control before calling sensitive methods. This produces repetitive code and can be potentially dangerous as it can be forgotten sometimes.

Example 1 (`SimpleGraderController::run`):
```
    public function run() {
        if(!$this->core->getAccess()->canI("grading.simple")) {
            $this->core->getOutput()->showError("This account doesn't have access to grading");
        }
        switch ($_REQUEST['action']) {
```

Example 2 (`CourseMaterialsController`):
```
    public function modifyCourseMaterialsFileTimeStamp() {

        if(!$this->core->getUser()->accessAdmin()) {
            $message = "You do not have access to that page. ";
            $this->core->addErrorMessage($message);
            $this->core->redirect($this->core->buildUrl(array('component' => 'grading',
                'page' => 'course_materials',
                'action' => 'view_course_materials_page')));
        }
```

Add an annotation class and it returns an error `Response` if the condition is not satisfied. It would look like:

```
/**
* @AccessControl('admin')
*/
public function modifyCourseMaterialsFileTimeStamp() {
    ....
}
```

Discussion needed.

zjxiaohan:

Fixes #3965.

### What is the new behavior?
We could now handle access control by putting annotations on the top of `class`es or `method`s. This prevents us from adding `if (!$this->core->getUser()->accessAdmin())` everywhere.

Example:

https://github.com/Submitty/Submitty/blob/44dc6fd4d426efbf9d0f79fe53f649a9b41a7ffc/site/app/controllers/admin/ReportController.php#L22-L28

https://github.com/Submitty/Submitty/blob/44dc6fd4d426efbf9d0f79fe53f649a9b41a7ffc/site/app/controllers/course/CourseMaterialsController.php#L202-L210

### Other information?

This only works for new style URLs.


site/app/controllers/admin/ReportController.php
```diff
@@ -21,7 +22,7 @@
 /**
  * Class ReportController
  * @package app\controllers\admin
- *
+ * @AccessControl(role="INSTRUCTOR")
```
MasterOdin: What happens if we put "instructor" here (lowercase string)? Does this cause an exception to be thrown?

zjxiaohan: Yes it does.

https://github.com/Submitty/Submitty/blob/43751a196346fb950101ddbeeed214be89ea0bb2/site/app/libraries/routers/AccessControl.php#L43-L58

zjxiaohan: added `strtoupper` to make lowercase also valid

site/app/libraries/routers/AccessControl.php
```diff
@@ -0,0 +1,82 @@
...
+namespace app\libraries\routers;
+
+use Doctrine\Common\Annotations\Annotation;
+
+/**
+ * Annotation class for @AccessControl().
+ *
+ * @Annotation
+ * @Target({"CLASS", "METHOD"})
+ */
```
MasterOdin: Let's add some documentation/examples of what that might look like, the scope of how you can apply them, do they stack hierarchically (define class level, does defining at method level override it? Can we define a role at class level and permission at method level? We can't and that's fine, but it's good to mention that), etc.

I think we should probably put that documentation here as opposed to `WebRouter::accessCheck` given that IDEs will link here if someone is confused what a particular annotation does. May want to put a comment/link to that particular function though for reference.

zjxiaohan: Several examples have been added. Could you have a look? I will keep it [WIP] until #4030 is good.

site/app/controllers/admin/ReportController.php
```diff
@@ -21,32 +22,20 @@
...
-            case 'summary':
-                $this->generateGradeSummaries();
-                break;
-            case 'reportpage':
-            default:
-                $this->showReportPage();
-                break;
-        }
+        return null;
     }
```
MasterOdin: Is there a reason to leave this function in at all given that it's just returning null?

site/app/libraries/routers/AccessControl.php
```diff
@@ -0,0 +1,134 @@
...
+ *
+ *      The following function is only accessible to full access graders
+ *      and instructors with "grading.simple" permission.
+ *
+ *      @AccessControl(role="FULL_ACCESS_GRADER", permission="grading.simple")
+ *      public function foo() {...}
+ *
+ * Note that @AccessControl() annotation of methods will override the one
+ * of their classes. For clarity, it is recommended not to have class-level
+ * @AccessControl() and method-level @AccessControl() at the same time.
```
MasterOdin: Note that if you use the method level annotation, the class level annotation will not be considered at the same time. For clarity, it is recommended not to have class and method level annotations at the same time.

site/app/libraries/routers/AccessControl.php
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * Example (class & method):
+ *
+ *      The foo() function is NOT accessible to non-instructors with "grading.simple"
+ *      permission. It IS accessible to instructors with OR without "grading.simple"
+ *      permission.
```
MasterOdin: The foo() function will only consider if the user has the "instructor" role, and will NOT check the "grading.simple" permission.

site/app/controllers/admin/ReportController.php
zjxiaohan: AbstractController enforces that. Needs to be removed after old-style routes completely phase out.

MasterOdin: Oh, yeah, that makes sense.

PR #4012, opened 2019-07-05
Fix 'now' and 'end of time' datepicker buttons from not saving on gradeable dates

shailpatels:
Forces the date textareas to update in order to validate and save the dates entered by pressing the buttons on the flatpicker calender in the gradeable edit pages.

(Also changed the end of time value  to be 1998 I think comparing dates with the actual end of time was causing JS math issues and would always say fail to save)


bmcutler:
Thanks Shail!

PR #4015, opened 2019-07-06
[Bugfix:TAGrading] Fix failure to download checkout files

### LINKED ISSUE #3619: download of zip files from checkout browser buggy
**Describe the bug**
If a student has checked in a zip file to their repository, I cannot download that single file from the ta grading panel.

We apparently *still* have lots of places where download of zip file fails, this one is possibly a little different because it is not a zip generated on the fly, but a submitted zip file.  (most submitted zip files are unzipped before they are stored).

zjxiaohan:

Fixes #3619 . Note that the failure happens not only for zip files, but all other files in the `checkout` directory.

### What is the new behavior?
The files can be downloaded correctly.


site/public/js/ta-grading.js
```diff
@@ -746,6 +746,9 @@ function downloadFile(html_file, url_file) {
     else if (url_file.includes("results")) {
         directory = "results";
     }
+    else if (url_file.includes("checkout")) {
```
bmcutler: What is the contents of url_file?  Is it the path to a students file?
If the student names a file in their repository "my_src/my_submissions_results.cpp" for example
will this code break because url_file will be "checkout/my_src/my_submissions_results.cpp"?

Should this be a more specific check so we don't get confused by the file name?

zjxiaohan: True. This is too crude a way to locate a directory. Will fix it.

Issue #3889, opened 2019-06-20
Browser Forensics creates lots of error messages


bmcutler:
When we click on the "Grade" or "Regrade" button for an assignment that is uploaded by students, and then click on "Bulk Upload Forensics", we get a page with many copies of these two errors:

Warning: file_get_contents(/var/local/submitty/courses/s19/sample/submissions/grades_released_homework_onlyauto/thompc/1/bulk_upload_data.json): failed to open stream: No such file or directory in /usr/local/submitty/site/app/controllers/student/SubmissionController.php on line 1977

Warning: file_get_contents(/var/local/submitty/courses/s19/sample/submissions/grades_released_homework_onlyauto/willia/2/bulk_upload_data.json): failed to open stream: No such file or directory in /usr/local/submitty/site/app/controllers/student/SubmissionController.php on line 1978

Seems like no one tested this option for regular submissions.  Likely the bulk upload forensics page only makes sense when we create student submissions by bulk upload.  Perhaps we should hide this option if there are no bulk uploads (but this might be hard to determine).  More importantly, if we do go to this page, we should have some error checking before doing the actions that are causing these errors :)

PR #4017, opened 2019-07-07
[Bugfix:InstructorUI] remove bulk upload forensics from non is_scanned submissions

Fenn-CS:

When we click on the "Grade" or "Regrade" button for an assignment that is uploaded by students, and then click on "Bulk Upload Forensics", multiple errors described in #3889 are gotten.

This is because forensics is only available for bulk uploads (i.e gradeables of type `scanned_exam`), so the functionality fails to work on other gradeables which is not of the mentioned type.

Fixes: #3889 

### What is the new behavior?
Hide "Bulk upload forensics" button if the gradeable does not support bulk uploads.

**_Working gradeable page for gradeable with bulk upload support._**

Screenshot from 2019-07-07 02-21-54

Screenshot from 2019-07-07 02-07-52

**_Otherwise_**

Screenshot from 2019-07-07 02-19-41

### DIFF
--- a/site/app/templates/grading/electronic/Status.twig
+++ b/site/app/templates/grading/electronic/Status.twig
@@ -148,7 +148,7 @@
                     Download Zip of All Students
                 </a>
             {% endif %}
-            {% if core.getUser().accessFullGrading() %}
+            {% if core.getUser().accessFullGrading() and is_scanned_exam %}
                 <a class="btn btn-primary"
                    href="{{ core.buildUrl({'component': 'submission', 'action': 'stat_page', 'gradeable_id': gradeable_id}) }}">
                     Bulk Upload Forensics
--- a/site/app/views/grading/ElectronicGraderView.php
+++ b/site/app/views/grading/ElectronicGraderView.php
@@ -196,6 +196,7 @@ class ElectronicGraderView extends AbstractView {
             "gradeable_title" => $gradeable->getTitle(),
             "team_assignment" => $gradeable->isTeamAssignment(),
             "ta_grades_released" => $gradeable->isTaGradeReleased(),
+            "is_scanned_exam" => $gradeable->isScannedExam(),
             "rotating_sections_error" => (!$gradeable->isGradeByRegistration()) and $no_rotating_sections
                 and $this->core->getUser()->getGroup() == User::GROUP_INSTRUCTOR,
             "autograding_non_extra_credit" => $gradeable->getAutogradingConfig()->getTotalNonExtraCredit(),


Fenn-CS:
@bmcutler I see the python linters are causing this to fail with 

```
"flake8-docstrings" failed during execution due to "module 'pydocstyle' has no attribute 'tokenize_open'"
Run flake8 with greater verbosity to see more details
```

Is this from master?

bmcutler:
I don't think the solution is so simple.  The setting 'is_scanned_exam' is primarily for defaulting a couple parameters for student view.  It does not control whether bulk upload may be used.  

Instructors can (currently) use bulk upload on any gradeable.  

Have you tested if bulk upload forensics are available if bulk upload has been used on a is_scanned_exam=false gradeable?  The stats *should* be available.

Have you tested if bulk upload forensics is error free if is_scanned_exam=true but there are no submissions yet?  Have you tested if bulk upload forensics is error free if is_scanned_exam=true but all current submissions are by 'make submission for a student' rather than bulk upload?

MasterOdin:
To give some context, what needs to be changed is:
https://github.com/Submitty/Submitty/blob/17af50f87429a3697b839d72e8e0d9bdb4dfa1bc/site/app/controllers/student/SubmissionController.php#L1679-L1689

Where it needs to do is a check that `$json_file` exists (if not, continue to next user_id). While here, should rewrite the function so that it:
1. Rename `$json_path` to `$base_path` on line 1670, and get rid of `$path_reset` on line 1671. 
1. Change `$json_path` inside the function to extend off `$base_path` instead of `$json_path`.
1. Do the `$this->core->getQueries()->getUserById($user_id_arr[$i])` once and reuse the returned object for setting firstname and lastname.
1. Use `FileUtils::loadJsonFile($json_path)` to load the file once to a variable, and then use the variable to get `upload_timestamp`, `submit_timestamp`, and `filepath`.
1. Get rid of the set `$json_path = $path_reset;` on line 1688 (as we're now no longer overwriting it)

Given that there can be a forensics page for any upload gradeable, there's really no point hiding the button and I'd think instructors may find it being there/not being there confusing based on the existence of a file (or not).

Fenn-CS:
> Have you tested if bulk upload forensics are available if bulk upload has been used on a is_scanned_exam=false gradeable? The stats should be available.

No. I think it won't work for this case and I would update it to see that it works.

> Have you tested if bulk upload forensics is error free if is_scanned_exam=true but there are no submissions yet? Have you tested if bulk upload forensics is error free if is_scanned_exam=true but all current submissions are by 'make submission for a student' rather than bulk upload?

Yes for these last two cases it works fine, so the only thing to do here is to ensure that this button works properly for all gradeables.

@shailpatels ok thanks for the ideas, I am curious apart from the approach given by Matt was I to check for the path how will I determine the ID for the user who uploaded? It could be any instructor right?.

@Matt thanks for the input.

PR #4019, opened 2019-07-07
[Bugfix:Forum] Maintain Forum thread list scroll state

### LINKED ISSUE #3032: forum thread list scrolling
When you click on a thread from the list, the left thread list scrolls vertically to place the selected thread at the top.  This is confusing.  It would be better to not move the left panel scrolling, leave the clicked item approximately where it was.

xprilion:

Fixes #3032

### What is the new behavior?
Threads are loaded via Ajax, thus maintain the state of scroll 

https://www.loom.com/share/28e73e36cdd749e8921599d35968b2e6

### Other information?
Is this a breaking change?: None yet
How did you test: Untested currently


site/app/templates/forum/displayThreadList.twig
```diff
@@ -1,5 +1,5 @@
 {% for thread in thread_content %}
-    <a href="{{ thread.link }}">
+    <a href="{{ thread.link }}" onclick="loadThread(event, this, {{ thread.thread_id }});">
```
andrewaikens87: Can we have try to avoid inline js here?

site/public/js/forum.js
```diff
@@ -129,6 +129,8 @@ function publishFormWithAttachments(form, test_category, error_message) {
     }
     var submit_url = form.attr('action');
 
+    console.log(submit_url);
```
andrewaikens87: Please remove log.

site/public/js/forum.js
```diff
@@ -137,6 +139,9 @@ function publishFormWithAttachments(form, test_category, error_message) {
         type: 'POST',
         success: function(data){
             try {
+
+                console.log(data);
```
andrewaikens87: Please remove log.

xprilion:
@andrewaikens87 that is a pretty standard practice, and is being used on Quora and LinkedIn, where they display a floating button saying (scroll up for new posts) and when the user scrolls up they get the new posts. Should be an interesting add, but might be slightly unrelated to this issue. So I'll open a new one and we can merge that after this one?

site/app/templates/forum/displayThreadList.twig
xprilion: Definitely, something that can be avoided. I'll change how that works!

MasterOdin:
We could also just put a refresh icon at the top of the thread list which when clicked would refresh the thread list.

site/app/templates/forum/displayThreadList.twig
xprilion: Done!

site/public/js/forum.js
xprilion: Done!

site/public/js/forum.js
xprilion: Done!

xprilion:
@MasterOdin that would probably be the easiest solution on the block! (Which should always be preferred before choosing anything else!)

andrewaikens87:
I agree @MasterOdin that is an easier approach. I am just not sure how often someone would click refresh, although it would lower the number of requests for loading new information. The approach I mentioned above would require some logic to be redefined for the way we load the posts when we scroll upwards and are on the first block. This would require a bit of work, but might be better from a user's perspective. @bmcutler thoughts? Which would you prefer?

bmcutler:
We decided to not add a refresh button on the page.  And to not have a mechanism for automatically refreshing content.  

The user can use the refresh button on their browser to pull new thread & post data.  Or to click to another Submitty page and click back into the forum.  This is sufficient for most use cases at this time.  

(We will probably re-visit this topic -- e.g., using web sockets -- in the future.)

PR #4023, opened 2019-07-08
Fixed css and js link in code.twig

mktung:
Fixed broken Codemirror on TA grading

### DIFF
--- a/site/app/views/MiscView.php
+++ b/site/app/views/MiscView.php
@@ -11,20 +11,25 @@ class MiscView extends AbstractView {
     }
 
     public function displayCode($file_type, $filename, $file_contents) {
-        $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('codemirror', 'codemirror.css'));
-        $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('codemirror', 'theme', 'eclipse.css'));
-        $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('codemirror', 'theme', 'monokai.css'));
-        $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('jquery', 'jquery.min.js'));
-        $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('codemirror', 'codemirror.js'));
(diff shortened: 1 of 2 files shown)


site/app/views/MiscView.php
```diff
@@ -11,20 +11,25 @@ public function displayFile($file_contents) {
...
-        $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('codemirror', 'codemirror.css'));
-        $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('codemirror', 'theme', 'eclipse.css'));
-        $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('codemirror', 'theme', 'monokai.css'));
-        $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('jquery', 'jquery.min.js'));
-        $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('codemirror', 'codemirror.js'));
-        $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('codemirror', 'mode', 'clike', 'clike.js'));
-        $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('codemirror', 'mode', 'python', 'python.js'));
-        $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('codemirror', 'mode', 'shell', 'shell.js'));
+        $code_css = array();
+        $code_css[] = $this->core->getOutput()->timestampResource(FileUtils::joinPaths('codemirror', 'codemirror.css'), 'vendor');
```
zjxiaohan: How about using `addInternalCss`, `addVendorCss`, `addInternalJs` and `addVendorJs`?

zjxiaohan: Sorry I just saw you deleted those lines.

mktung: The `add` functions load stuff into global header, but this is a partial module which is conditionally loaded onto the page, after the header is made. For those scenarios, we use the `timestamp` function to format.

zjxiaohan: Thank you for your explanation. :)

Issue #124, opened 2015-09-05
Rename section_id to section_number in sections table and use section_title in system


MasterOdin:
We treat the section_id as the section_number anyway in the classlist, so might as well have that column name reflect this. We'd also want to get rid of the auto-increment on the table and have users hand put the number. 

Additionally, we want to start using "section_title" in various parts of the system (such as in grading table) as right now, it's not used for anything and kind of a pointless column.

MasterOdin:
Done in f2647d0a412bc31e4746ed65ccd45c5bc1109ed1

PR #4027, opened 2019-07-08
[Example:Autograding] Simplify Python Custom Validation

### LINKED ISSUE #3470: Update Python Custom Validation Example
The python custom validation example in more autograding examples should be made more accessible to users unfamiliar with the system.

emaicus:
Closes #3470
### This PR:
1. Works to make the python custom validation example in ```more_autograding_examples``` easier for new users to understand by restructuring its layout and adding more extensive comments. 
2. Adds the python custom validation example to the development course.


more_autograding_examples/python_custom_validation/config/custom_validation_code/grader.py
```diff
@@ -1,129 +1,250 @@
+"""
```
bmcutler: I think this should all go on submitty.org instead?

emaicus: Moved relevant potion to Submitty.org (PR #124).

emaicus:
Updated to fix the few small items @Kevin-MM  found in his review. 

There is a separate pipeline in place for instructors to write C++ custom validators. Python custom validators are a more recent extension. There are plans to extend the feature to work with compiled languages in the future.

PR #4028, opened 2019-07-09
Detect upload_max_filesize instead of hardcoded file sizes in images and course materials upload.

### LINKED ISSUE #3626: Course materials/student photos upload max size should detect apache limit

### LINKED ISSUE #4008: Remove hard-coded photo size limit
The size limit per student photo or the zip file is hard-coded in the current controller.

https://github.com/Submitty/Submitty/blob/d43bb2643b3681b3095aa114aa707425f86a4bce/site/app/controllers/student/SubmissionController.php#L1686-L1689

We may want to put it into some configuration file.

Questions:
1. Will this config be different from course to course, or there should be a limit in a global config? (Are there courses that require super large student photos?)
2. Now that photos are uploaded manually by instructors, will that be meaningful to set up a limit especially considering PHP itself has a upper bound on the upload size?
3. @shailpatels will this problem still exist after #3724 is done?

shailpatels:
Max files sizes for uploading course materials and student images were hardcoded at 10M 

### What is the new behavior?
CourseMaterialsuploadController and ImagesController now use the `upload_max_filesize` variables from php.ini. "Files must not exceded ..." messages now also check php.ini for max filesize instead of hardcoded values in the uploadCourseMaterials form and uploadImagesForm

### Other information?
Changed php.ini values and uploaded to course materials and student images, see https://submitty.org/sysadmin/system_customization#allowing-large-student-file-upload-submissions

Fixes #4008
Fixes #3626

### DIFF
--- a/site/app/controllers/grading/ImagesController.php
+++ b/site/app/controllers/grading/ImagesController.php
@@ -122,7 +122,7 @@ class ImagesController extends AbstractController {
             }
         }
 
-        $max_size = Utils::return_bytes(ini_get('upload_max_filesize'));
+        $max_size = Utils::returnBytes(ini_get('upload_max_filesize'));
         if ($file_size > $max_size) {
             return $this->core->getOutput()->renderResultMessage("File(s) uploaded too large.  Maximum size is ".($max_size/1024)." kb. Uploaded file(s) was ".($file_size/1024)." kb.", false);
         }
--- a/site/app/controllers/grading/ImagesController.php
+++ b/site/app/controllers/grading/ImagesController.php
@@ -6,6 +6,7 @@ use app\controllers\AbstractController;
 use app\libraries\FileUtils;
 use app\models\user;
 use Symfony\Component\Routing\Annotation\Route;
+use app\libraries\Utils;
 
 class ImagesController extends AbstractController {
     public function run() {
@@ -121,7 +122,7 @@ class ImagesController extends AbstractController {
             }
         }
 
-        $max_size = 10485760;
+        Utils::return_bytes(ini_get('upload_max_filesize'));
         if ($file_size > $max_size) {
             return $this->core->getOutput()->renderResultMessage("File(s) uploaded too large.  Maximum size is ".($max_size/1024)." kb. Uploaded file(s) was ".($file_size/1024)." kb.", false);
         }
--- a/site/app/controllers/grading/ImagesController.php
+++ b/site/app/controllers/grading/ImagesController.php
@@ -122,7 +122,7 @@ class ImagesController extends AbstractController {
             }
         }
 
-        Utils::return_bytes(ini_get('upload_max_filesize'));
+        $max_size = Utils::return_bytes(ini_get('upload_max_filesize'));
         if ($file_size > $max_size) {
             return $this->core->getOutput()->renderResultMessage("File(s) uploaded too large.  Maximum size is ".($max_size/1024)." kb. Uploaded file(s) was ".($file_size/1024)." kb.", false);
         }
--- a/site/app/libraries/Utils.php
+++ b/site/app/libraries/Utils.php
@@ -305,7 +305,7 @@ class Utils {
 
     /**
     * Convert bytes to a specified format thats human readable 
-    * E.g : MB, 10485760 => 10Mb
+    * E.g : MB, 10485760 => 10MB
     * @param string $format
     * @param int $bytes
     * @return string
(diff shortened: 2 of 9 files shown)


site/app/templates/course/CourseMaterials.twig
```diff
@@ -177,5 +177,5 @@
 
 
 {% include('admin/users/UserForm.twig') with {'action': 'update_student'} %}
-{% include('course/UploadCourseMaterialsForm.twig') %}
+{% include('course/UploadCourseMaterialsForm.twig') with {'max_size_string' : max_size_string} %}
```
zjxiaohan: Included templates have access to the variables of the active context. `with` is for additional variables. `with {'max_size_string' : max_size_string}` is safe to be removed.

ref: https://twig.symfony.com/doc/2.x/tags/include.html

site/app/templates/grading/Images.twig
```diff
@@ -49,4 +49,4 @@
     </table>
 </div>
 {% include('admin/users/UserForm.twig') with {'action': 'update_student'} %}
-{% include('grading/UploadImagesForm.twig') %}
+{% include('grading/UploadImagesForm.twig') with {'max_size_string' : max_size_string} %}
```
zjxiaohan: Same as another comment.

site/app/controllers/grading/ImagesController.php
```diff
@@ -121,7 +122,7 @@ public function ajaxUploadImagesFiles() {
             }
         }
 
-        $max_size = 10485760;
+        Utils::return_bytes(ini_get('upload_max_filesize'));
```
zjxiaohan:
```suggestion
        $max_size = Utils::return_bytes(ini_get('upload_max_filesize'));
```

site/app/libraries/Utils.php
```diff
@@ -286,4 +286,33 @@ public static function removeStudentWithId($students, $key, $id) {
...
    }
+
+   /**
+    * Convert the shorthand byte notation in php.ini to bytes.
+    * E.g : php return_bytes(ini_get('post_max_size'))
+    * Src : https://www.php.net/manual/en/function.ini-get.php
+    * @param string $val
+    * @return int 
+    */
+    public static function return_bytes ($size_str){
```
zjxiaohan: Sorry that I failed to make comments all at once, but the function names should be in camelCase like other functions in this file per https://submitty.org/developer/coding_style_guide/php#naming-conventions

site/app/libraries/Utils.php
```diff
@@ -286,4 +286,33 @@ public static function removeStudentWithId($students, $key, $id) {
         }
         return $students;
    }
+
+   /**
+    * Convert the shorthand byte notation in php.ini to bytes.
+    * E.g : php return_bytes(ini_get('post_max_size'))
+    * Src : https://www.php.net/manual/en/function.ini-get.php
+    * @param string $val
```
zjxiaohan:
```suggestion
    * @param string $size_str
```

site/app/libraries/Utils.php
```diff
@@ -286,4 +286,33 @@ public static function removeStudentWithId($students, $key, $id) {
...
+            case 'M': case 'm': return (int)$size_str * 1048576;
+            case 'K': case 'k': return (int)$size_str * 1024;
+            case 'G': case 'g': return (int)$size_str * 1073741824;
+            default: return $size_str;
+        }
+    }
+
+    /**
+    * Convert bytes to a specified format thats human readable 
+    * E.g : MB, 10485760 => 10Mb
```
zjxiaohan: I am not very sure about that but I vaguely remember that `b` denotes `bit` and `B` denotes `byte`. 10485760 bytes is supposed to be 10 megabytes. 

ref: https://www.quora.com/What-is-the-difference-between-Kb-and-KB

zjxiaohan: So maybe not use `ucfirst` below.

shailpatels:
> Thanks for your prompt commits. The PR works as expected and I do fail to upload some file > 10MB. The error message seems unfriendly but it isn't related to this issue.

Yeah I noticed that as well, unfortunately, thats due to the way we're checking for errors on file uploads. PHP decides that if the file size uploaded is larger than the one set in php.ini it will abort which sets the filesize to zero resulting in the wrong error.

Correcting issues like you said is its own problem I plan to do in its own PR
https://github.com/Submitty/Submitty/issues/4029

PR #4010, opened 2019-07-05
[Feature:InstructorUI/RainbowGrades] Initial work on automatic rainbow grades

Kevin-MM:
@holzbh started this feature and most credit for the user interface goes to him

### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [x] Tests for the changes have been added/updated (if possible)
* [x] Documentation has been updated/added if relevant [#123](https://github.com/Submitty/submitty.github.io/pull/123)

### What is the current behavior?
Previously there was no user interface for customization / generation of rainbow grades.

### What is the new behavior
When logged in as a course admin you will now see a 'Web-Based Rainbow Grades Generation' under the 'Grade Reports' tab.  This form allows for some gui based configuration of rainbow grades.  When the 'Save Changes' button is clicked rainbow grades will be automatically generated or if this form has already been used in the past then new changes will be applied.

**How to use**
- First ensure the [jobs daemon is running](https://submitty.org/developer/jobs_daemon) by calling ```sudo systemctl start submitty_daemon_jobs_handler``` prior to running the submitty install script
- Ensure the RainbowGrades repo is cloned in your GIT_CHECKOUT directory
- Log into the 'sample'
- Press the 'Generate Grade Summaries' button inside 'Grade Reports' tab
- Press the 'Web-Based Rainbow Grades Generation' button
- Play with the new customization user interface, saving the form when satisfied
- Check out the results in the 'Rainbow Grades' tab (must be enabled inside the course configuration page)

**Limitations**
- Currently grade summaries must be generated manually before this feature can work
- The 'points distribution' part of gradeables isn't currently implemented
- The first time this form is used, it takes a minute to generate rainbow grades.  The is currently no feedback about when this process has completed (fixed in a future PR)
- The system breaks if you submit the form with a blank messages box (fixed in a future PR)

**Other notes**
- An instructor may place a 'custom_customization.json' file inside their rainbow_grades folder.  This file will always take priority over anything generated by the web user interface and functionality of the web user interface is undefined in this case.
- Output of auto_rainbow_grades.py is captured in 'auto_debug_output.txt' of the course's rainbow_grades directory

### Other information?
- This PR should not break anything, appropriate migrators are included.
- Tested manually

PR #4034, opened 2019-07-09
[Feature:RainbowGrades] Customization UI improvements

Kevin-MM:
This PR builds on #4010 and should be merged after.

Previously the Rainbow Grades Customization UI didn't give any useful feedback about when the server-side processes had completed.

### What is the new behavior?
The UI form now displays a few different status messages indicating the status of rainbow grades generation, specifically when the generation has completed.

See #4010 for instructions on how to use.

A few small bug fixes and visual improvements are included.

### Other information?
Tested manually inside 'sample course'


migration/migrator/migrations/course/20190628100043_create_rainbow_grades_dir.py
```diff
@@ -0,0 +1,28 @@
+import os
+import grp
+from pathlib import Path
```
MasterOdin: These migration files should probably be combined.

bmcutler: this PR will be merged after another PR which introduces these migrations.  
But yes, I agree, the migrations should be combined (and re-numbered to be 201907XX).
And then removed from this PR.

Kevin-MM:
Removed two old migrators that had been merged into a single one.  Also made a few small UI improvements.

bmcutler:
If I get in a state where rainbowgrades errors (don't worry about debugging the rainbowgrade error right now)  

But the page gets stuck with status saving or with "Status: Generating rainbow grades, please wait..."
for a *long* time.  Eventually it says a failure occured.  The auto_debug_output.txt is already updated with the error.  What additional timeout are we waiting for?

root@vagrant:/var/local/submitty/courses/s19/development/rainbow_grades# more auto_debug_output.txt 
rsync: failed to set times on "/var/local/submitty/courses/s19/development/rainbow_grades/raw_data/.": Operation not permitted (1)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1196) [sender=3.1.2]
make: *** [pull_test] Error 23
process_grades.out: /usr/local/submitty/GIT_CHECKOUT/RainbowGrades/output.cpp:602: void start_table_output(bool, const std::vector<Student*>&, int, int, i
nt, int, Student*, Student*, Student*, Student*, Student*): Assertion `GRADEABLES[g].getPercent() < 0.01' failed.
make: *** [overall] Aborted (core dumped)
rsync: link_stat "/var/local/submitty/courses/s19/development/rainbow_grades/individual_summary_html/*" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1196) [sender=3.1.2]
make: *** [push_test] Error 23

bmcutler:
I'm confused what is happening.  There are no jobs in the queue, the new error log output file is produced, but the page is still hanging for 60 seconds.

Kevin-MM:
I believe this problem is occurring because I am waiting for output.html to be produced in rainbow_grades directory to signal when the c++ part of the code has finished running.  In this case the file doesn't get produced so the script times out.  Right now the default timeout is set at 60 seconds, but I think we can probably shorten it.  I'm not sure this bit of code is required as I think I put it in there before I added some other checks so let me remove it and do some testing.

bmcutler:
Now it just always says "successfully generated" even when it failed.
So there's a special case check for the category "Note", which is not supposed to count for a grade.
(We may indeed want to change this, but for now it's a convenient way to check failure).

To reproduce, drag the "Note" category into the assigned buckets, and give it a percentage value > 0.
Then you should get this error:
Assertion `GRADEABLES[g].getPercent() < 0.01'
in auto_debug_output.txt

And output.html is not generated.

And the individual_summary_html is not generated (and wont be copied to ../reports/summary_html)
But if you had an earlier successful run, the old files will still be there.

Regardless, it always says "Rainbow grades successfully generated!" when the job completes, even if there was an error and (new) files were not produced.


Also, it failed travis after the last commit, so I don't want to merge until I see the green check again :) 
It might be the need to merge master apache config thing.  If it passes in the morning (even if we don't have this status this sorted), I'll merge.

.setup/apache/submitty.conf
```diff
@@ -1,144 +0,0 @@
-# If hosting more than just Submitty on Apache, you should set the
```
bmcutler: oops, I don't think your PR should be deleting this file.  That is likely why its failing the travis e2e tests

Kevin-MM: Readded this file.

Kevin-MM:
Updated code so a failure of the rainbow grades c++ code no longer produces "successfully generated" in the ui.

On another note I did notice that setting the Note category to 1% does not produce this error.

PR #4036, opened 2019-07-09
[Bugfix:Autograding] Image Diff Fails Gracefully on Different Sized Images

### LINKED ISSUE #3438: Image diff should fail more gracefully if the student & expected images are of different sizes.
(with an appropriate error message)

emaicus:
Closes #3438

### This PR:
1. Updates the image diff validation type so that it fails gracefully if images are of a different size. 
2. Updates the ```image diff mirror``` gradable in development to fix bugs and make it simpler.

#### Before
before_image_updates

#### After
not_of_same_size

### DIFF
--- a/more_autograding_examples/image_diff_mirror/config/config.json
+++ b/more_autograding_examples/image_diff_mirror/config/config.json
@@ -7,7 +7,6 @@
     "testcases" : [    
         {
             "title" : "Mirror",
-            "command" : "sleep .001",
             "points" : 1,
             "validation" :
             [
@@ -22,7 +21,6 @@
         },
         {
             "title" : "Student 1 Professor 1",
-            "command" : "sleep .001",
             "points" : 1,
             "validation" :
             [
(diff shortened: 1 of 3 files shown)


more_autograding_examples/image_diff_mirror/config/config.json
```diff
@@ -1,23 +1,13 @@
...
-            //Note the multistep compilation which first 
-            //calls cmake and then make. 
-            "command" : ["g++ *.cpp"],
-            "executable_name" : "a.out",
-            "points" : 4
-        },    
         {
             "title" : "Mirror",
-            "command" : "./a.out",
+            "command" : "sleep .001",
```
MasterOdin: Can command be empty if all someone is interested in is the validation step?

emaicus: Good catch. I was under the impression that all testcases required a command (I believe that was a constraint at one point). Commands don't really make sense for this gradeable, as it just takes a user provided image and checks to see if it matches the instructor provided one, so I added the sleep as a workaround. As it turns out, we must have removed the need for a command at some point, so I was able to nix them for this gradeable in my most recent commit.

PR #4040, opened 2019-07-09
[Bugfix:Email] disabled emails from ui if emails is not enabled

### LINKED ISSUE #3828: IS_EMAIL_CONFIGURED & auto-email on instructor announcement

sak6lab:
closes #3828 
If emails are disabled in emails.json, notification settings for emails is still shown, code will still try to send emails on notifications, and the course settings for sending email seating assignments will still allow users to send the seating assignments. Also, when making a thread there is an option to send the thread notification as an email and another for announcement these two options should be combined.

### What is the new behavior?

- Code does not insert into email table if emails are disabled
- Emails settings are not shown and header is changed to just Notification Settings if emails are disabled
no_email_settings
- The course settings page will not allow users to access the email seating assignment page and the email seating assignment page will not allow users to send the seating assignments (if the page is accessed through url hacking) if emails are disabled
course_settings_disabled

- Combined send email and announcement
1) enabled
announcement_email_enabled
2) disabled
announcment_email_disabled


site/app/templates/NotificationSettings.twig
```diff
@@ -115,10 +119,9 @@
                         </label>
                     </div>
                     <div class="option-input col-md-3"><input type="checkbox" name="all_modifications_forum" id="all_modifications_forum" {% if notification_saves['all_modifications_forum'] %}checked="true"{% endif %}></div>
-
-                    <div class="option-input col-md-5"><input type="checkbox" name="all_modifications_forum_email" id="all_modifications_forum_email" {% if notification_saves['all_modifications_forum_email'] %}checked="true"{% endif %}></div>
-
-
+                    {% if email_enabled %}
+                        <div class="option-input col-md-5"><input type="checkbox" name="all_modifications_forum_email" id="all_modifications_forum_email" {% if notification_saves['all_modifications_forum_email'] %}checked="true"{% endif %}></div>
+                    {% endif %}
```
MasterOdin: We should probably update `NotificationController::saveSettings` to not overwrite values that are missing from the page here as it can lead to some unexpectedness if email is enabled for a while, users set their settings turning on email notifications, email is disabled, user goes to the page and disables a setting (all email settings are now turned off), and then email is re-enabled.

sak6lab: The user shouldnt be able to change their email settings if emails are disabled as that column to edit email settings would not be shown to them. However it is probably safe to have that check.

site/app/controllers/NotificationController.php
```diff
@@ -88,6 +88,14 @@ public function viewNotificationSettings() {
     public function changeSettings() {
         //Change settings for the current user...
         $new_settings = $_POST;
+        if ($this->core->getConfig()->isEmailEnabled()) {
+            foreach ($new_settings as $key => $value) {
+                if (!substr_compare($key,'_email',strlen($key)-6)) {
+                    unset($new_settings[$key]);
+                }
+            }
+        }
```
MasterOdin: This doesn't address what I was talking about (and since the input fields aren't even shown on the form, has no effect).

It's rather the next `if` block, namely the:
```
            $values_not_sent = array_diff(self::NOTIFICATION_SELECTIONS, array_keys($new_settings));
            foreach(array_values($values_not_sent) as $value) {
                $new_settings[$value] = 'false';
            }
```

where any setting that wasn't part of the `$_POST` payload is just implicitly set to false, which would include the email settings, regardless of their previous setting. Not sure why this was added, but can probably remove it.

sak6lab: I split the constant array NOTIFICATION_SELECTIONS into one for notification and one for email and decided at run time to add EMAIL_SELECTIONS to the property $selections. That way the changeSettings() function can ignore the email settings. I tried just deleting making the unset settings to false but if all check boxes are false the update part of the query would throw a syntax error as there would be no keys to update.

Issue #2896, opened 2018-09-21
VPAT -- Radio Button grouping


pbailie:
Accessibility readers interpret `<fieldset>` and `<legend>` to identify and describe radio button and checkbox groups.  However, `<fieldset>` and `<legend>` seem to be used for strictly visual grouping that is inconsistent with accessibility tools.  

Minimally, the use of `<fieldset>` and `<legend>` needs to better correspond with accessibility tools to group radio buttons and checkboxes (and perhaps any other pertinently related controls corresponding to the radio buttons / checkboxes).  Optimally, HTML should not be used for visual aesthetics.  That is the purview of CSS.

This issue is particularly prevalent among the Gradeable creation and editing forms.

The following is advice provided by WAVE:

- `<fieldset>` does not exist for a radio button group (WAVE  _alert_):

> **Missing fieldset**
> **What It Means**
> A group of check boxes or radio buttons is not enclosed in a fieldset.
> **Why It Matters**
> A fieldset provides a visual and structural grouping of related form elements. It is typically necessary for groups of check boxes or radio buttons where a higher level description (called a legend) is necessary to understand the function of the check boxes or radio buttons. The description will be identified by a screen reader only if provided in a fieldset legend.
> **How to Fix It**
> Determine whether the grouping of check boxes or radio buttons has or needs text that explains the purpose of the check boxes or radio button grouping. If so, mark up the group within a fieldset and put the group description in a legend element.
> **The Algorithm... in English**
> Two or more checkbox or radio input elements within a form have the same name value, but are not enclosed in a fieldset.
> **Standards and Guidelines**
> [Section 508 (n)](http://webaim.org/standards/508/checklist#standardn)
> [1.1.1 Non-text Content (Level A)](http://webaim.org/standards/wcag/checklist#sc1.1.1)
> [1.3.1 Info and Relationships (Level A)](http://webaim.org/standards/wcag/checklist#sc1.3.1)
> [2.4.6 Headings and Labels (Level AA)](http://webaim.org/standards/wcag/checklist#sc2.4.6)
> [3.3.2 Labels or Instructions (Level A)](http://webaim.org/standards/wcag/checklist#sc3.3.2)

-  `<fieldset>` exists but `<legend>` does not (WAVE _alert_):

> **Fieldset missing legend**
> **What It Means**
> A fieldset does not have a legend.
> **Why It Matters**
[...]

PR #4041, opened 2019-07-09
[VPAT:InstructorUI] html semantics of rotation section form

mktung:
Relates to #2896


site/app/templates/admin/users/RotatingSectionsForm.twig
```diff
@@ -13,31 +13,33 @@
...
-        Courses that are cross-listed in different departments may have multiple course codes/prefixes.<br>
-        <br>
+        Courses that are cross-listed in different departments may have multiple course codes/prefixes.
+    </p>
+    <p>
         Each student in the course is assigned to one registration section.<br>
-        Students who have dropped the course will be assigned to the <em>NULL</em> registration section.<br>
-        <br>
+        Students who have dropped the course will be assigned to the <em>NULL</em> registration section.
+    </p>>
```
jmcknight98: oops, extra >

mktung:
The point of the PR was to reorganize the HTML so that it fit VPAT spec - I just moved the CSS since I thought it was a little thing to change. You're right though that I didn't add a lot of it back, so hopefully I've fixed that in my latest commit.

site/app/templates/admin/users/RotatingSectionsForm.twig
mktung: Fixed in latest commit

PR #4051, opened 2019-07-10
[Bugfix:Flatpickr] Fix onchange not triggering with shortcut buttons, fix extensions math  

shailpatels:
- Changes flatpicker to call onchange trigger upon setting the date using the shortcut buttons. Also changes "end of time" year to be 9998 JS would magically interpret the year 9999 as 9998, 

- remove my hack to force edit gradeable dates to update.

- Fixes the way the number of extension days are calculated when you use flatpickr to select a date.


shailpatels:
> Yup! The reason I moved the math code to `calculateLateDays()` was because `onchange` was not triggering. I could not quickly notice that the `triggerChange` parameter defaults to false. I tried this and it works as expected, also leaving the code outside in the function keeps it cleaner.
> 
> About changing the year from `9999` to `9998` that's not a problem but I would like to know if you have a special reason for doing so?

Haha yeah I used a hack to force updating before I learned how to actually trigger the onchange event and that required going through the source code of the shortcut buttons plugin. 

For the 9998 I did that because it looks like JS is doing some weird conversions or math underneath, if 

```js
<html>
<p id="test">

</p>
<script>
   let t = new Date("9999-01-01");
   document.getElementById("test").innnerHTML = t;
</script>
```

If you run that it'll show 9998 as the year for some reason and if you do any math comparisons with that date object it won't work with 9999. If an instructor put forever for 2 dates the comparison would always fail for some reason so just to be consistent and make sure the displayed date matches what js wants I changed everything to 9998.

MasterOdin:
Passing a full string to the Date constructor has a lot of trapdoors that it's generally not recommended as the route you want to take. If you pass a string datetime, it is parsed as UTC, and then depending on your browser, it'll may be assumed that the object should be translated into your local timezone. The conversion here is why you're seeing 9998:
```
a = new Date('9999-01-01');
> Thu Dec 31 9998 19:00:00 GMT-0500 (Eastern Standard Time)
```

Testing for straight equality of dates won't work since of how JavaScript does objects, but you can do less than/greater than calculations. 
```
a = new Date('9999-01-01');
b = new Date('9999-01-01');
c = new Date('9999-01-02');
a == b
> false
a < c
> true
```

If you want to do equality comparisons, you'd have to do something like:
```
a.getFullYear() === b.getFullYear() && a.getMonth() === b.getMonth() && a.getDate() === b.getDate()
```
(those will all return stuff in the local timezone).

A better method might be to pass in the date numbers individually:
```
a = new Date(9999, 0, 1)
Fri Jan 01 9999 00:00:00 GMT-0500 (Eastern Standard Time)
```
which will cause the browser to assume you wanted that date in your local timezone as is, which is mostly fine.

Having said all that, if we're doing anything really with dates beyond very simple usecases, it's almost always better to use a library like [dayjs](https://github.com/iamkun/dayjs) or [momentjs](https://momentjs.com/).

shailpatels:
@MasterOdin thanks for the explanation, that makes a lot of sense!
and yeah like you said anything complex would be too gross to do with the native js date library but for this it only sets the 'forever' time which I tested with 9998 so i think it should be fine to leave it.

Issue #3975, opened 2019-07-01
User access level in master database


bmcutler:
We need to define access levels for users at the whole submitty level, not just the per-course level.  

Notes from meeting today with proposed design:

Per course we will still have 1=instructor, 2=fullaccessgrader, 3=limitedaccessgrader, 4=student.

But in the submitty master database we will add a new column named "access_level" that will define per 1=superuser, 2=faculty, 3=user. 

Actions allowed per access_level:
(actions here are API actions and/or future web UI actions)
Note this list is initial brainstorm -- still a work-in-progress

1=superuser:  
Can view list of all courses.  
Can add themselves to any course.  
Can have read only access to any course.  (without significant changes to current code, might need to add themselves to the course first?)
Can create new users and give them access level 1 or 2.

2=faculty:
Can create new courses.  (note: linux group for new course must be selected from one of their current courses)
Can view list of all other users on the system(??) 
Can create new users when they add people to their course, but those users will always be access level=3=user.
At RPI this will be restricted to faculty.

3=user:
Can be in any course as any level 1-4.  
At RPI, all students, including TAs and head TAs will have access level 3.  They will not have the ability to create new courses. 

tagging @zjxiaohan for use in API '

-------------

Edited to add note from PR #3987:

I'm going to tack on a second task for this issue to help with testing/setup. Let's modify setup sample courses scripts for our vagrant box to make the Quinn Instructor user a user_access=2(faculty) level user. And let's add a brand new user, maybe Clark Kent? who is user_access=1(superuser). Maybe the superuser is not in any courses initially.

PR #4052, opened 2019-07-10
[Feature:Vagrant] Superuser for Vagrant

pbailie:
Implements superuser for Vagrant and updates needed setup code for user_access_level.  
Related to #3975 

- Update setup_sample_courses.py
    - Added code for user_access_level column.  Default is 3.
    - Fixed bug in user_group assert range check.
    - user_group and user_access_level assert checks now give meaningful error messages.
- New user:  _superuser.yml_, user_access_level is 1, not enrolled in any courses.
- Update user: _instructor.yml_, user_access_level is 2.

### DIFF
--- a/.setup/bin/setup_sample_courses.py
+++ b/.setup/bin/setup_sample_courses.py
@@ -146,6 +146,7 @@ def main():
                               user_lastname=user.lastname,
                               user_preferred_lastname=user.preferred_lastname,
                               user_email=user.email,
+                              user_access_level=user.access_level,
                               last_updated=NOW.strftime("%Y-%m-%d %H:%M:%S%z"))
 
     #Sort alphabetically extra students. Shouldn't affect randomness....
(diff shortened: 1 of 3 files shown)


.setup/bin/setup_sample_courses.py
```diff
@@ -538,7 +541,16 @@ def __init__(self, user):
...
+        try:
+            assert 1 <= self.group <= 4
+        except:
+            raise SystemExit("ASSERT: user {}, user_group is not between 1 - 4. Check YML file.".format(self.id))
+        if 'user_access_level' in user:
+            self.access_level = user['user_access_level']
+        try:
+            assert 1 <= self.access_level <= 3
+        except:
+            raise SystemExit("ASSERT: user {}, user_access_level is not between 1 - 3. Check YML file.".format(self.id))
```
MasterOdin: Let's make both of these try/except blocks just if statements:
```
if self.group < 1 or 4 < self.group:
    raise SystemExit("ASSERT: user {}, user_group is not between 1 - 4. Check YML file.".format(self.id))
```

and 

```
if self.access_level < 1 or 3 < self.access_level:
    raise SystemExit("ASSERT: user {}, user_access_level is not between 1 - 3. Check YML file.".format(self.id))
```

PR #4053, opened 2019-07-11
[Bugfix:InstructorUI] fixed stream not found errors in forensics page

Fenn-CS:

When we click on the "Grade" or "Regrade" button for an assignment that is uploaded by students, and then click on "Bulk Upload Forensics", multiple errors described in #3889 are gotten.

This is because forensics is only available for bulk uploads (i.e gradeables of type `scanned_exam`), so the functionality fails to work on other gradeables which is not of the mentioned type.

Fixes: #3889 

### What is the new behavior?
Shows empty page for gradeables without any bulk upload and hence forensic data.

**_Working gradeable page for gradeable with bulk upload support._**

Screenshot from 2019-07-07 02-21-54

Screenshot from 2019-07-07 02-07-52

**_Otherwise_**

Screenshot from 2019-07-11 03-00-49

### Other information?
Continues : #4017


site/app/controllers/student/SubmissionController.php
```diff
@@ -1667,27 +1667,32 @@ public function checkRefresh() {
...
+            $user = $this->core->getQueries()->getUserById($user_id_arr[$i]);
+            $file_contents = FileUtils::readJsonFile($json_path);
+            $users[$user_id_arr[$i]]["first_name"] = $user->getDisplayedFirstName();
+            $users[$user_id_arr[$i]]["last_name"] = $user->getDisplayedLastName();
+            $users[$user_id_arr[$i]]['upload_time'] = $file_contents['upload_timestamp'];
+            $users[$user_id_arr[$i]]['submit_time'] = $file_contents['submit_timestamp'];
+            $users[$user_id_arr[$i]]['file'] = $file_contents['filepath'];
+        }
+        if(count($users) == 0)
+            return;
```
MasterOdin: What does doing `return` cause to be displayed?

MasterOdin: What gets shown if you pass the empty `$users` array to the renderOutput?

Fenn-CS:
@MasterOdin those lines
```
if(counter($users) == 0)
    return;
````
It is unnecessary with the current restructuring I made. But I did not easily notice because it gave an error without that in the previous code.
https://github.com/Submitty/Submitty/blob/8c9e55b76735787f03df1d42eb762ce647847883/site/app/controllers/student/SubmissionController.php#L1677-L1680

In the previous code (on master) above the users array on line 1677 and 1678 is initialised before the loop and the loop counter is increamented before first iteration `++i`. 

With that arrangment the front page with the knowlege there's something in the users array tries to loop and throws errors like `undefined index firstname not found.` so I had to add a 
```
if(counter($users) == 1)
    return;
````
to account for the 1 user who is initialised before the loop, but I now know this is problematic because in the case where there is actually a single user with forensic information it would return nothing.

Despite all that, after removing the initialisation before the loop, and changing to incrementation after first run, `i++` the users array would be empty if there was no forensic data for any user hence rendering both useless.

https://github.com/Submitty/Submitty/blob/dc211bbe66281b0366669d9a4c7bbc3d31285831/site/app/controllers/student/SubmissionController.php#L1674-L1676

This correction has been made.

File: site/app/controllers/student/SubmissionController.php [hunk 2]
```diff
@@ -1667,27 +1667,31 @@ public function checkRefresh() {
...
-        $user_id_arr = array_slice(scandir($json_path), 2);
-        $user = $user_id_arr[0];
-        $users[$user] = array();
-        for($i = 0; $i < count($user_id_arr); ++$i) {
-            $files = scandir($json_path . $user_id_arr[$i]);
+        $base_path = $course_path . "/submissions/" . $gradeable_id . "/";
+        if(!is_dir($base_path))
+                return;
+        $users = array();      
+        $user_id_arr = array_slice(scandir($base_path), 2);
```
MasterOdin: Can we change these 4 lines above to be:
```
$users = array();
$user_id_arr = is_dir($base_path) ? array_slice(scandir($base_path), 2) : [];
```

so that we we'll just fall through to the bottom of the function and just show the empty tabular page instead of a completely blank page?

MasterOdin:
Thanks for the explanation! Very helpful and was inline with what I thought was happening, and means I can give my final bit of review. Once the above change is made, this should be ready to go I think.

bmcutler:
@Fenn-CS please address review comments.  Thanks!

site/app/controllers/student/SubmissionController.php
Fenn-CS: ok!

Fenn-CS:
@bmcutler done! Also updated with master.
Screenshot from 2019-07-15 19-52-58

site/app/controllers/student/SubmissionController.php
```diff
@@ -1667,27 +1667,29 @@ public function checkRefresh() {
...
-            $users[$user_id_arr[$i]]["first_name"] = $this->core->getQueries()->getUserById($user_id_arr[$i])->getDisplayedFirstName();
-            $users[$user_id_arr[$i]]["last_name"] = $this->core->getQueries()->getUserById($user_id_arr[$i])->getDisplayedLastName();
-            $users[$user_id_arr[$i]]['upload_time'] = json_decode(file_get_contents($json_path), true)['upload_timestamp'];
-            $users[$user_id_arr[$i]]['submit_time'] = json_decode(file_get_contents($json_path), true)['submit_timestamp'];
-            $users[$user_id_arr[$i]]['file'] = json_decode(file_get_contents($json_path), true)['filepath'];
-            $json_path = $path_reset;
+            $json_path = $user_path . "/" . $num_files . "/bulk_upload_data.json";
+            if(!file_exists($json_path))
+                continue;
+            $user = $this->core->getQueries()->getUserById($user_id_arr[$i]);
```
MasterOdin: Add a check of:
```
if ($user === null) {
    continue;
}
```

which will at least prevent the error from @shailpatels comment and he can go ahead and make the issue to track adding this support as I'm not even sure how this would look.

MasterOdin:
@shailpatels can you make an issue about adding support for teams with the various cases of how the directory structure might look and also leave the discussion of what the forensics page should look like in this case since I'm guessing that the bulk upload is being registered as coming from a team?

site/app/controllers/student/SubmissionController.php
Fenn-CS: Thanks Matt

Issue #2273, opened 2018-06-27
Twig cleanup


CouleeApps:
Some of the earlier views that I did have some inefficient / bad practices that I should clean up. Here are a few:
- [ ] Not using `loop.index`
- [ ] `{% for %} {% if %}` can be inlined to `{% for if %}`
- [ ] Weird includes and lack of includes
- [ ] URLs should really be generated in php and fed into Twig
- [ ] Similarly, csrf tokens and basically anything that uses `Core`
- [ ] Code style consistency (currently camel_case)
- [ ] Peer grading made some of the for-component-in-gradeable loops messy
- [ ] `student_full|raw` json data
- [ ] Directory/files layout
- [ ] `|date()` filter instead of using php
- [ ] Potentially factor common sections into one place
- [ ] #780 and #2271 can happen at roughly the same time

bmcutler:
@mktung and @zjxiaohan 
Please discuss on Slack how to divide the remaining tasks on this list.

zjxiaohan:
> Instead of having {{ core.buildUrl() }} in Twig, use $this->core->buildUrl in PHP and pass that into Twig
>
>  Similarly, csrf tokens and basically anything that uses Core should be in PHP. Goal is to have Twig not need {{ core.anything }}

I would fix these two in the process of URL replacement.

> Instead of doing student_full|raw and json_encoding the array in php, pass the array to Twig and use thing|json

I actually think it is good to encode the array as JSON in php because we may want to have APi endpoints that return JSON. That really depends on if we want `raw` or not.

PR #4062, opened 2019-07-12
[Bugfix:Edit User] Fix the edit user/new user page

### LINKED ISSUE #4061: "Edit User" panel (from "Manage Students") does not save
**Describe the bug**
Seems like we've broken the "Edit User" panel.  I press "Submit" but nothing seems to save.
It also does not work to create a "New Student" user.

The related panel for "Edit User" from "Manage Graders" and (edit or "New Grader") does seem to work.

Sadly there is no indication of failure.  No info in site_errors.  No red alert box with an error message.  I just don't get the green alert success message and the data doesn't change.

I do get this message when I try to refresh the page (which is unhelpful)

Screen Shot 2019-07-11 at 11 55 31 PM

jmcknight98:
Closes #4061 

Oops, I broke it (see issue above)... 
Should fix the edit user and New Student page

### DIFF
--- a/site/app/templates/admin/users/UserForm.twig
+++ b/site/app/templates/admin/users/UserForm.twig
@@ -109,7 +109,7 @@
     </script>
 {% endblock %}
 {% block form %}
-    <form method="post" action="{{ return_url_update_grader }}" onchange="userFormChange()">
+    <form method="post" action="{{ core.buildUrl({'component': 'admin', 'page': 'users', 'action': action}) }}" onchange="userFormChange()">
         {{ parent() }}
     </form>
 {% endblock %}
--- a/site/app/templates/admin/users/UserForm.twig
+++ b/site/app/templates/admin/users/UserForm.twig
@@ -109,7 +109,7 @@
     </script>
 {% endblock %}
 {% block form %}
-    <form method="post" action="{{ edit_user_return_url }}" onchange="userFormChange()">
+    <form method="post" action="{{ core.buildUrl({'component': 'admin', 'page': 'users', 'action': action}) }}" onchange="userFormChange()">
         {{ parent() }}
     </form>
 {% endblock %}
--- a/site/app/templates/admin/users/UserForm.twig
+++ b/site/app/templates/admin/users/UserForm.twig
@@ -109,7 +109,7 @@
     </script>
 {% endblock %}
 {% block form %}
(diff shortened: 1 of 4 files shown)


site/app/templates/admin/users/UserForm.twig
```diff
@@ -109,7 +109,7 @@
     </script>
 {% endblock %}
 {% block form %}
-    <form method="post" action="{{ return_url_update_grader }}" onchange="userFormChange()">
+    <form method="post" action="{{ core.buildUrl({'component': 'admin', 'page': 'users', 'action': action}) }}" onchange="userFormChange()">
```
xprilion: I guess the `return_url_update_grader` wasn't being passed to this twig file at all. You might want to shift it in the View file

site/app/views/admin/UsersView.php
```diff
@@ -61,7 +61,6 @@ public function listGraders($graders, $reg_sections, $rot_sections, $use_databas
             "reg_sections" => $reg_sections,
             "rot_sections" => $rot_sections,
             "use_database" => $use_database,
-            "return_url_update_grader" => $this->core->buildUrl(array('component' => 'admin', 'page' => 'users','action' => 'update_grader')),
```
xprilion: Shifting this to `userForm` should help!

xprilion: Might need some modification though

jmcknight98:
Yes it wasn't being passed to the twig file.  I was cleaning up a few of the core.buildUrl calls in twig files (as suggested in #2273) that I happened to notice while working on a separate issue.  But this one was in a file that wasn't included, so I just reverted the change to fix the bug.  It can be cleaned up as part of that issue.

zjxiaohan:
Sorry I just saw your comment after sending the comment.

Yes we could have a quick fix temporarily and I will do a cleanup some time after, but note that this fix is incomplete as it breaks edit grader function and needs to be fixed.

A quick fix would be changing line 55 in `GraderList.twig` to use `with {'action': 'update_grader}`

jmcknight98:
Oh yes I didn't notice they were using the same form when doing the quick fix, they should both work now as intended.

site/app/views/admin/UsersView.php
jmcknight98: I switched it back to using buildUrl in the twig file, its more complicated then I thought so trying to fix it as I passed through the files was a mistake.  This should be changed in the future within #2273, but for now changes are just to make it work as before.

PR #4068, opened 2019-07-12
Fixes error where download for students was null

### LINKED ISSUE #4063: Course Materials download zip buggy
**Describe the bug**
Downloading the zip file of a directory as a student is buggy.  It produces a zero byte file.  

The download interaction is different than as an instructor.  As an instructor I'm asked to select a file name/directory.  As a student it just creates a file with default name in default location.

The zip file as an instructor includes files that are not yet released.

(I'm suspecting a bad merge with @zjxiaohan and @thef1stofhername recent work)

thef1stofhername:
Currently folder downloads will null
closes #4063
now it is full of all the files you want (if they're available for you)
tested by releasing/unreleasing folders/files as an instructor and downloading as student


site/public/js/server.js
```diff
@@ -1750,7 +1750,26 @@ function changePermission(filename, checked) {
 
     $.ajax({
         url: url,
-        success: function(data) {},
+        success: function(data) {csrf_token: csrfToken},
```
zjxiaohan: `fn` is not sent.

thef1stofhername: whoops sorry! I rushed it too much, hopefully this will fix it

thef1stofhername: ok fixed, no errors for me on testing individual files, please let me know if errors persist

Issue #3968, opened 2019-07-01
Sequence Diagram Updates


emaicus:
Update and evaluate the styling for sequence diagrams produced by networked assignments.

PR #4070, opened 2019-07-12
[Feature:Autograding] Add Sequence Diagrams for Networked Gradeables

emaicus:
Should not be merged until knsv/mermaid#847 is addressed.

Adds an initial (rough) version of sequence diagrams for visualization of networked gradeables.
Visualization will be cleaned up per #3968


MasterOdin:
What do the generated diagrams look like?

emaicus:
This is an early example of a diagram produced by this PR. An example planned interaction is the ability to mouse over a message to see its contents in cases where message content is overlong.

sequence_diagram

MasterOdin:
It looks like they've released a new version with a stricter security model which prevents XSS, but at the cost of not having any allowed click action.

PR #4072, opened 2019-07-12
if you are ahead of NY, things should be released according to NY time

### LINKED ISSUE #4065: Course material release time is not time zone sensitive
**Describe the bug**
When clicking "now" in the course material page, the time is set with regard to my time zone, which may be different from the time zone of the server.

**Expected behavior**
No matter where I am, when I, as an instructor, sets a course material to be release now, students should be able to see them immediately.

**To Reproduce**
1. Change the time zone of your computer while Submitty is running in NY.
2. Login as instructor, set one file to be released "now".
3. Login as student. Can't see the file.

**Configuration**
 - OS: Linux
 - Browser Chrome

**Screenshots**

Note 15:26:34 in China === 03:26:34 in NY, and after I set "now" in China, I can't see the course material released as a student.

Instructor Side:
Screenshot from 2019-07-12 15-28-03

Student Side:
Screenshot from 2019-07-12 15-28-29

thef1stofhername:
closes #4065
sets the day back by one if you're ahead of NY so it can be released "now" for everyone

used to be that if you were ahead on utc -4 and chose now, it would release for anyone in your timezone (your now) but now it releases now for everywhere


site/app/templates/course/CourseMaterials.twig
```diff
@@ -334,6 +334,14 @@
...
                                 case 0: 
                                     date = new Date();
+                                    var offset = date.getTimezoneOffset();
+                                    offset = offset/60;
+                                    console.log(offset);
+                                    //if your time is ahead of New York, sets back a day so
+                                    //everything is simultaneous
+                                    if(offset<4){
+                                       date.setUTCDate(date.getDate()-1);
+                                    }
```
sak6lab: I dont think moving the day back is the best way to do this because it is not always the case that the server's timezone is EST in which case your offset would be wrong. I suggest taking the user's time no matter where they are and converting it to UTC and do all you calculation on when to display it with that time. Then when you display it convert it back to the user's time.

Kevin-MM: I agree with this, I am also a firm believer that all timestamps should be stored in UTC using [atom datetime format](https://www.php.net/manual/en/class.datetimeinterface.php#datetime.constants.atom).

MasterOdin: Per @bmcutler, all dates are stored with their respective timezones (and written to json files and log files) to make it easier to debug time-related bugs and issues so that instructors and newbie sysadmins don't have to worry about constantly doing timezone conversions on dates to their local time.

thef1stofhername: ok I'll take saketh's/kevin/matt's suggestions

zjxiaohan: You could have a look at how it stored timestamp before your edit. 

My `course_materials_file_data.json` contains some old data and they look like this:

```
    "/var/local/submitty/courses/s19/sample/uploads/course_materials/TicTacToe-orig/.settings/org.eclipse.jdt.core.prefs": {
        "checked": "1",
        "release_datetime": "2019-06-03 08:10:13-0400"
    },
    "/var/local/submitty/courses/s19/sample/uploads/course_materials/TicTacToe-orig/gen/org/example/tictactoe/Manifest.java": {
        "checked": "1",
        "release_datetime": "2019-06-03 08:10:20-0400"
    },
    "/var/local/submitty/courses/s19/sample/uploads/course_materials/TicTacToe-orig/bin/resources.ap_": {
        "checked": "1",
        "release_datetime": "2019-06-03 08:40:30-0400"
    },
```

but after changing to the new course material interface the json now stores things like this:

```
    "/var/local/submitty/courses/s19/sample/uploads/course_materials/Lab1/Lab01.pdf": {
        "checked": "1",
        "release_datetime": "2019-07-06 12:00:00"
    },
    "/var/local/submitty/courses/s19/sample/uploads/course_materials/Lab1/Lab1.c": {
        "checked": "1",
        "release_datetime": "2019-07-06 12:00:00"
    },
```

PR #4074, opened 2019-07-12
Color individual file date/time box onchange without refresh

### LINKED ISSUE #4064: Date color update without page reload
Remaining issue from PR #3923:

Have the file datestamp highlighting update without reloading the page.

thef1stofhername:
closes #4064 

the colors change when you change the individual file
yay


site/public/css/server.css
```diff
@@ -1369,6 +1369,10 @@ end of styles used in the admin gradeable page
    color: #d9534f;
 }
 
+.grader-invalid{
+   color: #d9534f;
```
zjxiaohan: Looks like there are two `.grader-invalid` in this file.

site/app/templates/course/CourseMaterials.twig
```diff
@@ -297,8 +298,80 @@
     function setNewDateTime(me, path) {
         // pass filename to server to record the new date and time of the file to be released
         var newDateTime = $(me).val();
-
         changeNewDateTime(path, newDateTime);
+        var today = new Date();
+        //to properly compare dates/times they need to match in format
+        //not getting it from the picker leaves out the zero before single digits
+        if(today.getMonth()+1<10){
```
shailpatels: Are you using these formated strings anywhere, if so lines lines 305 to 340 can be replaced with :
```js
var now = new Date();

function pad(str){
  return ('0' + str).slice(-2);
}


var date = now.getFullYear() + "-"+ pad(now.getMonth()) + "-" + pad(now.getDay());
var time = pad(now.getHours()) + ":" + pad(now.getMinutes()) + ":" + pad(now.getSeconds())

var currentDT = date+' '+time;
var neverDT = (now.getFullYear()+10)+'-'+pad(now.getMonth())+'-'+pad(now.getDay())+' '+time;
```

site/app/templates/course/CourseMaterials.twig
```diff
@@ -297,8 +298,80 @@
...
+        var time = hours+":"+mins+":"+secs;
+        var currentDT = date+' '+time;
+        var neverDT = (today.getFullYear()+10)+'-'+month+'-'+day+' '+time;
+
+        //get the value in each file so the color can be assigned
+        //based on the time chosen
+        var fileDT = newDateTime;
+
+        //also custom colors for this page for readability
+        if(fileDT<=currentDT){
```
shailpatels: If all you're doing is using the above calculation to compare dates then you get the string the datepicker outputs into the JS date constructor and compare that

(for equality you'll have to use `getTime()`)
```js
new Date('2019-08-24') < new Date('2019-08-24") ; for <, >
new Date('2019-08-24').getTime() == new Date('2019-08-24').getTime() ; // for equality
```

site/app/templates/course/CourseMaterials.twig
```diff
@@ -297,8 +298,80 @@
...
+            me.style.backgroundColor = "#f76c6c";
+            return "#f76c6c";
+        }
+        else{
+            me.style.backgroundColor = "#ffeb54";
+            return "#ffeb54";
+        }
+    }
+
+    function setChildNewDateTime(path, changeDate,handleData) {
```
shailpatels: I'm confused what this function is supposed to do, why can't you directly call `changeFolderNewDateTime` 

```
changeFolderNewDateTime(path, handleData);
```
which already accepts a function callback to pass its response.

shailpatels:
However you fix the code in courseMaterials.twig could you do the same here

https://github.com/Submitty/Submitty/blob/e51aad995171d7b4257c1acf079c5ce1da335bcf/site/app/templates/course/SetFolderReleaseForm.twig#L38-L63

from #4064

site/app/templates/course/CourseMaterials.twig
zjxiaohan: This may be of help too. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

thef1stofhername: ah I'll get right on that, this sounds dumb, but I literally didn't know js had  a pad option... ;A;

mktung:
I'm not sure if this is part of your issue, but it seems to set the colors based on local timezone. If that's being solved in another PR, then it works perfectly.

thef1stofhername:
I'm about 95% sure that this is part of Kevin's PR, but if not I could always come back and fix it afte I ask him tomorrow

site/app/templates/course/CourseMaterials.twig
```diff
@@ -297,10 +269,40 @@
...
+        //based on the time chosen
+        var fileDT = newDateTime;
+
+        //also custom colors for this page for readability
+        if(new Date(fileDT).getTime()<=new Date(currentDT).getTime()){
+            me.style.backgroundColor = "#87db88";
+
+            return "#87db88";
+        }
+        else if(fileDT>=neverDT){
```
shailpatels: I think you need to change fileDT and neverDT to Js Date Objects before comparison

thef1stofhername: ah I missed thst guy somehow

bmcutler:
Please update the title of this PR 
https://submitty.org/developer/how_to_contribute#how-to-make-a-pull-request-pr-to-submitty
(I'd change it, but I'm not sure what to write)
thanks

Issue #4011, opened 2019-07-05
Refactor Installation Scripts


bmcutler:
New design (from meeting with @MasterOdin today):

* .setup/install_system.sh
     - mostly unchanged, except some functionality (e.g. pip install) moves to
         .setup/install_submitty/install_tools.sh
     - this script is called on a fresh install & calls all of the other scripts
     - it is also safe to call this script on an existing system, and it will update some things
     - note: it calls .setup/configure_submitty.sh and .setup/install_submitty.sh

* .setup/configure_submitty.sh
     - renamed from CONFIGURE_SUBMITTY.sh, but otherwise unchanged

* .setup/install_submitty.sh
     - renamed from INSTALL_SUBMITTY.sh, but much of the details will be pulled out 
          into the other helper scripts.
     - new contents of this script are (approximately)

             run migrator (apparently there is a one-liner to do all 3 env?  please also add 
                         to migrations doc)
             .setup/install_submitty/setup_directories.sh    
             .setup/install_submitty/install_repos.sh    
             .setup/install_submitty/install_tools.sh   
             stop jobs daemon   
             .setup/install_submitty/install_bin.sh   
             start jobs daemon  
             .setup/install_submitty/install_site.sh  
             stop shipper & worker daemons (including all worker daemons on worker machines)  
             .setup/install_submitty/install_autograding.sh  
             start shipper & worker on primary machine  
             .setup/install_submitty/update_workers.sh  
             start workers on worker machines (as updates finish...)  

* .setup/install_submitty/get_globals.sh 
     - a helper script that is called by many of the scripts below

* .setup/install_submitty/setup_directories.sh  
     - I forget what goes in here, things like logs directories?

* .setup/install_submitty/install_repos.sh
     - clones/updates the necessary versions of other submitty repositories

* .setup/install_submitty/install_tools.sh  
     - analysis tools, nlohmann, other vendor things?  (pip/npm/etc?)

* .setup/install_submitty/install_bin.sh  
     - mostly just renamed version of INSTALL_SUBMITTY_HELPER_BIN.sh

* .setup/install_submitty/install_site.sh  
     - mostly just renamed version of INSTALL_SUBMITTY_HELPER_SITE.sh

* .setup/install_submitty/install_autograding.sh  
     - install & compile & fix permissions on C++ code for autograding, 
        and python scripts for autograding

* .setup/install_submitty/update_workers.sh  
     - copy code to worker machines, etc. 

@emaicus

[...]

Issue #3729, opened 2019-05-23
Organize 'your courses' list by semester better


shailpatels:
**What problem are you trying to solve with Submitty**
The course list can get clunky with a lot of courses over multiple semesters, its a bit
hard to see which ones you're currently in versus past courses.
![a](https://user-images.githubusercontent.com/12129065/58289736-4861c480-7d85-11e9-8173-ad27e1464588.PNG)

Here's a potential way to organize it for students but im not sure to handle people who are TAs/Instructors/students at the same time
![b](https://user-images.githubusercontent.com/12129065/58289732-45ff6a80-7d85-11e9-8a08-ca3a8f811bc6.PNG)

MasterOdin:
We might want to take this opportunity to redesign how the underlying model here works. What I mean by that is that our system focuses on things as "semesters", but there are schools that use trimesters or quarters, and that they may name them differently from us, and thus make it hard to order them via reverse lexicographical ordering (which is what we rely on).

I'd propose we add a new table to the `submitty` DB that holds these things and which would have a structure that has (at least) the following structure (with examples from above):

| id | name | start_date | end_date |
|---|-------|------------|------------|
| s19 | Spring 2019 | 01/10/2019 | 05/04/2019 |
| f18 | Fall 2018 | 08/30/2018 | 12/24/2018 |
| u19 | Summer 2019 | 05/20/2019 | 08/17/2019 |
| f19 | Fall 2019 | 08/29/2019 | 12/21/2018 |

Ordering of these can then be done easily through the `start_date` and then for any grouping on which the current date falls outside of the end_date can just be "minimized" as well to save space (with the expectation that students are really not going to be accessing old courses that much).

Other schools that are not RPI would then have an easy time of having their own ids and names for their sessions (whether they be semesters, trimesters, quarters, etc.).

bmcutler:
This is related to #3653 & #1184 archiving courses and converting courses to read-only.   

The design is to put more control in the hands of the instructor, who might choose to allow you to see old course material, or who might remove your access shortly after the course finishes.  It will also be impacted by what period of time the system administrators allow you to have access to the machine.  (If you don't take an RPI CSCI course the next term, you may well lose all access to the server.)

We have been slow to archive the courses from Fall 2018 -- it will probably happen in the next couple weeks.  Spring 2019 courses will probably hang around a bit longer.  Archived courses are currently only visible to instructors, and they are nicely separated & labeled for instructors.  I do agree the students could be displayed better.

PR #4078, opened 2019-07-13
[Feature:InstructorUI] GUI for Course Creation

### LINKED ISSUE #3398: Instructor Web UI for Creating a Course
The process for setting up a class is complicated and uses many steps at the command line. While the documentation explains each step, it would be nice to be able to use one interface for all tasks. I would like to be able to create a course and manage all aspects of the class through the web UI. 

Advantages:
- Lower server administration
- Simplified user experience
- Clean workflow

Features
- Create a new course
- Manage course details

Other featues could be incorporated with the goal of completely managing a course from the web UI.

zjxiaohan:

Closes #3398 .

### What is the new behavior?

1. (ONLY) Faculty members can create new courses via GUI based on existing ones.
Screenshot from 2019-07-13 16-50-25

2. (ONLY) Faculty members can create new courses via API.
Screenshot from 2019-07-13 16-51-25

3. Access level is displayed in homepage
Screenshot from 2019-07-13 16-52-16

### Other information?
This is a minimum working PR for course creation. I opened this PR to prevent it from growing too intimidating for code review.

Future steps to take after this one is merged:
1. Form validation.
2. Return meaningful status messages instead of always giving success response (for API) or nothing (for Web. you need to refresh the page to see the new course).
3. Enhance `@AccessControl()` so that it handles `access_level`.


.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
```diff
@@ -70,6 +70,10 @@ chmod 550 ${SUBMITTY_INSTALL_DIR}/sbin/authentication.py
 # everyone needs to be able to run this script
 chmod 555 ${SUBMITTY_INSTALL_DIR}/sbin/killall.py
 
+# DAEMON_USER should be able to run these scripts as root
+echo "%${DAEMON_GROUP} ALL = (root) NOPASSWD: ${SUBMITTY_INSTALL_DIR}/sbin/create_course.sh" >> /etc/sudoers
```
bmcutler: won't this result in duplicate lines when this script is re-run?

site/app/controllers/GlobalController.php
```diff
@@ -64,6 +64,15 @@ public function header() {
                     "icon" => "fa-star"
                 ]);
             }
+            elseif ($this->core->getUser()->getAccessLevel() === User::LEVEL_FACULTY) {
```
bmcutler: Please allow superusers to also create couses with the interface
(access level should set minimum requirements)

site/app/controllers/HomePageController.php
```diff
@@ -133,6 +134,76 @@ function(Course $course) {
         );
     }
 
+    /**
+     * @Route("/home/new_course", methods={"POST"})
+     * @Route("/api/courses", methods={"POST"})
+     */
+    public function createCourse() {
+        $user = $this->core->getUser();
+        if (is_null($user) || $user->getAccessLevel() !== User::LEVEL_FACULTY) {
```
bmcutler: or superuser.

site/app/controllers/HomePageController.php
```diff
@@ -133,6 +134,76 @@ function(Course $course) {
...
+            return new Response(
+                JsonResponse::getFailResponse("You don't have access to this endpoint."),
+                new WebResponse("Error", "errorPage", "You don't have access to this page.")
+            );
+        }
+
+        $semester = $_POST['semester'];
+        $course_title = $_POST['course_title'];
+        $head_instructor = $_POST['head_instructor'];
```
bmcutler: Is the 'head instructor' copied from the template course (like how we copy the group from the template course)?  Or is the head instructor set to be the person who created the course?  

It should be the person who created the course.  A course might be 'team taught' in the fall with instructors A & B, where A was arbitrarily chosen as the owner for the files.  Then if instructor B creates the course in the spring and teaches the course alone, they should own the files.

site/app/controllers/HomePageController.php
```diff
@@ -133,6 +134,76 @@ function(Course $course) {
...
+            new RedirectResponse($this->core->buildNewUrl(['home']))
+        );
+    }
+
+    /**
+     * @Route("/home/new_course", methods={"GET"})
+     */
+    public function createCoursePage() {
+        $user = $this->core->getUser();
+        if (is_null($user) || $user->getAccessLevel() !== User::LEVEL_FACULTY) {
```
bmcutler: or superuser

site/app/templates/CreateCourseForm.twig
```diff
@@ -0,0 +1,59 @@
...
+                        <option value="{{ course.getSemester }}|{{ course.getTitle() }}">
+                            {{ course.getSemester() }}.{{ course.getTitle() }}
+                        </option>
+                    {% endfor %}
+                </select>
+            </div>
+        </div>
+
+        <input type="hidden" name="csrf_token" value="{{ csrf_token }}" />
+        <input type="hidden" name="head_instructor" value="{{ head_instructor }}" />
```
bmcutler: When a superuser creates a course, they should be able to specify who is the head instructor for the course.  
It must be someone who is in the specified linux group.   Note: We will leave the creation of linux groups and addition of users to linux groups to be linux terminal command line actions.

When a faculty user creates a course, they will be the head instructor.   (It would be good to error check and confirm they are in the linux group.)

site/app/templates/HomePage.twig
```diff
@@ -35,6 +35,7 @@
             </li>
             <li><b>Last Name:</b> {{ user.getDisplayedLastName() }}</li>
             <li><b>Email:</b> {{ user.getEmail() }}</li>
+            <li><b>Access Level: </b> {{ access_level }}</li>
```
bmcutler: Let's only print access level for users who are superuser or faculty

.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
MasterOdin: Yeah, it will. We really shouldn't be trying to modify this file directly but rather add a file under `/etc/sudoers.d/` (much like how we handle cron).

site/tests/app/models/UserTester.php
```diff
@@ -114,6 +114,7 @@ public function testToObject() {
             'user_preferred_lastname' => null,
             'user_email' => "test@example.com",
             'user_group' => 1,
+            'user_access_level' => 3,
```
prestoncarman: Instead of hard coding level three, use the defined access level constant, such as "User::LEVEL_USER". This goes for all places with an access check.

.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
zjxiaohan: Thanks for pointing this out. Is there any example that could be my reference? (I didn't find `sudoers.d` in the code) @MasterOdin

site/app/controllers/HomePageController.php
zjxiaohan: The head instructor is set to be the person who created the course.

There's a problem though... The `head_instructor` shouldn't be `POST`ed because it can be randomly modified by the `POST`er who may be one of faculty. Will change.

site/app/templates/CreateCourseForm.twig
zjxiaohan: Yeah we should make sure that the instructor selected is in the `tas` group.

site/app/controllers/HomePageController.php
zjxiaohan: Do we need to rename the `base_course` to `template_course`? I am not sure about the naming.

.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
MasterOdin: It's not something we're currently doing. Right now there's also two places in install_system.sh that directly modifies `/etc/sudoers`.

I think for now, please just create a file `.setup/submitty_sudoers` with the above lines you were echoing into `/etc/sudoers` and then in `.setup/INSTALL_SUBMITTY.sh`, where we install the crontab file, add:
```
cat ${SUBMITTY_REPOSITORY}/.setup/submitty_sudoers | envsubst > /etc/sudoers.d/submitty
chmod 0440 /etc/sudoers.d/submitty
chown root:root /etc/sudoers.d/submitty
```

zjxiaohan: Looks like the env var is not substituted. I also found it hard-coded in `submitty_crontab`. Is it generated somewhere else?

https://github.com/Submitty/Submitty/blob/e51aad995171d7b4257c1acf079c5ce1da335bcf/.setup/submitty_crontab#L14

zjxiaohan:
So here is the new interface that addresses @bmcutler 's concerns.

This is the interface for a superuser.
Screenshot from 2019-07-15 19-52-40

Head Instructor will be a list of faculty members.
Any time Head Instructor changes, the Base Course list will be automatically updated (via an Ajax call) to a list of courses that the head instructor taught.

For faculty, it's like this:
Screenshot from 2019-07-15 20-00-31

The creator is the head instructor.

For group checking, it is carried out by the shell command. Even if the attacker changes the front-end, it won't work. As a normal user, it is guaranteed that any input displayed in the frontend is valid for course creation.

zjxiaohan:
Do we need to validate the semester though? Should we limit the semester within two semesters? Like if you are in s19, you can only create courses for s19 and u19.

site/app/templates/CreateCourseForm.twig
```diff
@@ -0,0 +1,99 @@
+<div class="content">
+    <h2>New Course -- Work In Progress</h2>
+
+    <form id="course-creation-form" action="{{ course_creation_url }}"
+          method="post" enctype="multipart/form-data">
+        <div class="option row">
+            <div class="option-desc col-md-6">
+                <label for="course_semester">
+                    <div class="option-title">Semester</div>
```
prestoncarman: What about using the "Term" instead of "Semester"? Our school is on the quarter system so we do not have semesters. (summer, autumn, winter, spring) This may be a bigger question, but it would be nice to switch to a more generic term here.

prestoncarman:
> Do we need to validate the semester though? Should we limit the semester within two semesters? Like if you are in s19, you can only create courses for s19 and u19... However, it can vary from school to school.

Remember not all schools are on the semester system. Our school is on the quarter system (summer, autumn, winter, and spring). It would be nice to support other types of school terms. Maybe switching to the using "term" instead of semester would be a good start.

.setup/submitty_sudoers
```diff
@@ -0,0 +1,2 @@
+submitty_daemon ALL = (root) NOPASSWD: /usr/local/submitty/sbin/create_course.sh
+submitty_daemon ALL = (root) NOPASSWD: /usr/local/submitty/sbin/adduser.py
```
MasterOdin: Discussed this with Barb and we're good to go with using this file. I'm fine leaving this as is for now and will replace the hardcoded stuff with variables (in this and the crontab file) in a future PR as part of the larger issue of #4011.

.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
MasterOdin: Yeah, it's because `envsubst` only uses environment variables which in bash means that the variables have `export` in them, and it won't use variables defined in the current script. Don't worry about using the variables here, I'll handle it in a future PR as part of the larger #4011 refactors.

site/app/controllers/HomePageController.php
bmcutler: naming isn't significant right now...  it's fine as it is and we can tweak it later if we change our mind.

bmcutler:
I'm looking for a better label than "Title" (which implies to me "Full course name" which is set by the instructor from the "Course Settings" page, and can have spaces doesn't have to be unique), the field should more strongly be associated with the database column, which is just 'course'.  Something like "Course ID", or "Course Code", and the associate rules should be documented on the page (I forget what we decided, no spaces, only letters numbers and maybe dash or underscore?) .   Other suggestions?

The prior semesters should go back further than 2 terms.  It is rather common for us to teach a class once every other year.  Or perhaps even less frequently.  Or maybe that's not the question you are asking.

And yes I agree with Preston, we should change the word "semester" to "term" everywhere in the system.  Matt has the related issue to add the valid terms to the master database so we can eventually make that a dropdown menu of the valid terms.   Hmm, we should do that ASAP.  Otherwise we're going to have a mess of instructor-created nonstandardized "S19", "s19", "S2019", "Spring2019", ... for that column :(   So that needs to be a dropdown of valid choices before we merge this.

bmcutler:
related #3729 and #4109

zjxiaohan:
> The prior semesters should go back further than 2 terms. It is rather common for us to teach a class once every other year. Or perhaps even less frequently. Or maybe that's not the question you are asking.

No. What I was asking is that do we need to limit faculty members to only create course within a two semester frame, which is if it's fall 2019 currently, how about only allowing selecting from fall 2019 and spring 2020 to create course. If we do not do something like this, the situation would be like what you said, S19, s19, S2019, Spring 2019... etc

If you were saying the drop down menu that contains all the courses the faculty member was involved in for the past two terms, I am not sure if it is necessary (for now). Does average instructors in RPI have taught a LOT of courses that is hard to display them all in one drop down menu?

bmcutler:
For now, let's make the term choice be a drop down containing the currently existing unix directories in the /var/local/submitty/courses/ directory.   We could further restrict the list to the directories that are writeable by the owner (the owner is root) of the directory.  (When we archive old terms we can remove the write bit as a signal to not allow creation of new courses in that term).  This will require the sysadmin to make the first course in a new term from the command line, or to manually make the term directory.  This is ok restriction to start.

Later, when we have implemented the new table in the submitty master database suggested by @MasterOdin indicating the valid terms & start dates and relative order, we can switch this selection to use the database instead of the unix directories.

-----

And for the drop down menu for prior courses, yes, please list all of the courses the faculty member was an instructor for.  Ideally in reverse chronological order (newest course at the top).  But the ordering will be incorrect/a hack until we have the table mentioned above (let's save that for another PR).

zjxiaohan:
> Sometimes after creating a course the success popup appears but the course isn't made (even after waiting/refreshing). I think if the title contains capital letters it causes it to fail, but it seems inconsistent. I know its a work in progress, so maybe just adding a warning if the title is invalid?

Thanks for the bug report. I'll take a look.

zjxiaohan:
@jmcknight98 The reason it happens is that for postgresql unquoted names are always folded to lower case. I am not sure if we want to force all course title (code) to be lowercase? If so, I could convert them all to lowercase. If not, we can have double quotes in postgresql commands. @bmcutler

prestoncarman:
Since we use the course title (code) as part of the url and an database key, I think forcing lowercase is fine. If you go that way, we should force it to be lowercase (even if you just lower the submitted value) and make a note about this restriction on the form page.

zjxiaohan:
Yeah I agree with having lowercase only. That saves us from many potential troubles. Also I noticed that in many places course titles are displayed in uppercase by default. I am not sure if we need to put a note though as we could do `strtolower` in the backend.

bmcutler:
Please fix merge conflicts

bmcutler:
Items I notice:

* When I go to the create new course page, there is no way to go back to the main submitty home page (listing all of the courses & about me).  The breadcrumb should allow us to go back.

* If I just make a directory /usr/local/submitty/courses/u19, with matching permissions, this is not enough to allow me to choose 'u19' as the term for a new course.  Are we requiring a course in this new term in addition to the existence of the directory?

* I made ta2 a faculty access user.  And added ta2 as an instructor of a brand new course named foo.  When I log in as ta2 I see ta2 has grader access to sample and instructor access to foo.  However, ta2 can make a new course for based on both sample and foo.  You should only be able to base a class on a course you are an instructor of.  (and you also need to be in the linux group, but linux group is not enough).

* I made a new course, got a success message asking me to refresh later.  But I don't see the course.  I waited a while, nothing happened.  I checked the jobs daemon, it's running successfully.  Not sure what went wrong.

zjxiaohan:
> I made a new course, got a success message asking me to refresh later. But I don't see the course. I waited a while, nothing happened. I checked the jobs daemon, it's running successfully. Not sure what went wrong.

Could you go to `/var/local/submitty/logs/course_creation` and copy the content of the log?

zjxiaohan:
> If I just make a directory /usr/local/submitty/courses/u19, with matching permissions, this is not enough to allow me to choose 'u19' as the term for a new course. Are we requiring a course in this new term in addition to the existence of the directory?

Yes. The semester list is fetched from database.

zjxiaohan:
> When I go to the create new course page, there is no way to go back to the main submitty home page (listing all of the courses & about me). The breadcrumb should allow us to go back.

> I made ta2 a faculty access user. And added ta2 as an instructor of a brand new course named foo. When I log in as ta2 I see ta2 has grader access to sample and instructor access to foo. However, ta2 can make a new course for based on both sample and foo. You should only be able to base a class on a course you are an instructor of. (and you also need to be in the linux group, but linux group is not enough).

Will do it. Thanks for the testing.

bmcutler:
Yup, I just found the log when I looked at the code.  ta2 was not in the group.  Hopefully we can check for that before allowing them to push the button creating that course.

zjxiaohan:
> Yup, I just found the log when I looked at the code. ta2 was not in the group. Hopefully we can check for that before allowing them to push the button creating that course.

Yeah that's the difficult thing that troubles me all the time. Is there a good way for PHP and shell to communicate? To do this do I need to add a shell script that checks the user group and outputs a json and let PHP create a daemon job to run that and read the output? Is there a better way to do it?

MasterOdin:
You can use the the submitty_cgi interface to drop down into either python or a more unrestricted PHP that can run shell commands. You can use `Core::curlRequest` to hit with a GET/POST request `$this->core->getConfig()->getCgiUrl()."script_name.cgi"` (probably fine to do a GET request here), and then get the response back and keep going all within the same block of PHP execution.

bmcutler:
From the superuser account I only have a single choice of user from the head instructor list, 'instructor'.  All users who are superuser or faculty should be in this drop down menu.

zjxiaohan:
> When I go to the create new course page, there is no way to go back to the main submitty home page (listing all of the courses & about me). The breadcrumb should allow us to go back.

Added.

> If I just make a directory /usr/local/submitty/courses/u19, with matching permissions, this is not enough to allow me to choose 'u19' as the term for a new course. Are we requiring a course in this new term in addition to the existence of the directory?

Do we need to change this? Or let's keep it as it is for now and wait for the semester -> term thing so that we have a better way to get "next" term.

> I made ta2 a faculty access user. And added ta2 as an instructor of a brand new course named foo. When I log in as ta2 I see ta2 has grader access to sample and instructor access to foo. However, ta2 can make a new course for based on both sample and foo. You should only be able to base a class on a course you are an instructor of. (and you also need to be in the linux group, but linux group is not enough).

Fixed.

> I made a new course, got a success message asking me to refresh later. But I don't see the course. I waited a while, nothing happened. I checked the jobs daemon, it's running successfully. Not sure what went wrong.

> Yup, I just found the log when I looked at the code. ta2 was not in the group. Hopefully we can check for that before allowing them to push the button creating that course.

Added a validation. So now after clicking create course there will be an error saying plz contact sysadmin the user group is incorrect.

> From the superuser account I only have a single choice of user from the head instructor list, 'instructor'. All users who are superuser or faculty should be in this drop down menu.

Fixed.

> You can use the the submitty_cgi interface to drop down into either python or a more unrestricted PHP that can run shell commands. You can use Core::curlRequest to hit with a GET/POST request $this->core->getConfig()->getCgiUrl()."script_name.cgi" (probably fine to do a GET request here), and then get the response back and keep going all within the same block of PHP execution.

Thanks a lot for the cgi thing! I struggled with it for a long time today (and it turns out that GET and POST have slightly different commands & I wasn't aware that shell recognizes `&` as `working in background` and wasted a lot of time here) and finally wrote a working script!! It looks powerful!

bmcutler:
just waiting to see if I broke any tests.

PR #4079, opened 2019-07-13
[Router] New URL for ConfigurationController

zjxiaohan:
### URL Changes

**Old:** 
- `/index.php?semester=s19&course=development&component=admin&page=configuration&action=view`
- `/index.php?semester=s19&course=development&component=admin&page=configuration&action=update`

**New:** 
- `/s19/development/config`
- `/s19/development/config/update` (POST)

### Tested Entrances
- As instructor, went to the page from the sidebar and tried to update the settings.

### DIFF
--- a/site/app/controllers/admin/ConfigurationController.php
+++ b/site/app/controllers/admin/ConfigurationController.php
@@ -5,23 +5,30 @@ namespace app\controllers\admin;
 use app\controllers\AbstractController;
 use app\libraries\FileUtils;
 use app\libraries\ForumUtils;
+use app\libraries\response\JsonResponse;
+use app\libraries\routers\AccessControl;
+use app\libraries\response\Response;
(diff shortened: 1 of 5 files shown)


site/app/controllers/admin/ConfigurationController.php
```diff
@@ -46,48 +53,32 @@ public function viewConfiguration()
             'auto_rainbow_grades'            => $this->core->getConfig()->getAutoRainbowGrades()
         );
 
-        if (isset($_SESSION['request'])) {
```
zjxiaohan: This `if` block is removed because I cannot find any usage of `$_SESSION['request']`.

MasterOdin: Yeah, this is was here back when saving the configuration was a set button and not handled via ajax request.

PR #4081, opened 2019-07-13
[Bugfix:InstructorUI] Allow uploading duplicate course material files

### LINKED ISSUE #4069: Unable to upload duplicate course material files
**Describe the bug**
I am unable to upload duplicate course material files.

**Expected behavior**
There should be a confirm window ("Do you want to replace the original file?") or simply a friendly reminder that you have to delete the file first.

**To Reproduce**
1. Upload one file
2. Upload exactly the same file again
3. See error (nothing happens. page doesn't refresh)

**Configuration**
 - OS: Linux
 - Browser chrome

**Additional context**

Screenshot from 2019-07-13 00-20-16

zjxiaohan:

Fixes #4069 .

### What is the new behavior?
Works as before. Pops up a dialog asking if you want to replace the file.

### Other information?
This bug exists because the `confirm` function in `SetFolderReleaseForm.twig` overrides the default JavaScript `confirm` function. Please confirm that my changes are correct @thef1stofhername .

### DIFF
--- a/site/app/templates/course/SetFolderReleaseForm.twig
+++ b/site/app/templates/course/SetFolderReleaseForm.twig
@@ -98,7 +98,7 @@
             });
 
         }
-        function confirm(){
+        function confirmReleaseDate(){
             var directory = document.getElementById("submit_time");
             var dirArr = JSON.parse(directory.dataset.indir);
             var ID = directory.dataset.iden;
@@ -134,7 +134,7 @@
 
 {% block buttons %}
     {{ block('close_button') }}
-    <input name="submit" class="btn btn-primary" id="submit_time" type="submit" value="Submit" data-iden=" " data-inDir=' ' onclick="confirm()"/>
+    <input name="submit" class="btn btn-primary" id="submit_time" type="submit" value="Submit" data-iden=" " data-inDir=' ' onclick="confirmReleaseDate()"/>
 
 {% endblock %}
 
--- a/site/public/js/drag-and-drop.js
+++ b/site/public/js/drag-and-drop.js
@@ -104,15 +104,15 @@ function addFilesFromInput(part, check_duplicate_zip=true){
 // 0 - a file with the same name already selected for this version
 // -1 - does not exist files with the same name
 // Second element: index of the file with the same name (if found)
-function fileExists(file, part){
+function fileExists(filename, part){
     for(var i = 0; i < previous_files[part-1].length; i++){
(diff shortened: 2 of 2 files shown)


thef1stofhername:
I'm on the branch, all the code changes are in, submitty update installed, but when I try to add a duplicate file, it just says it was uploaded successfully. This is the case for folders too. Does it work on your end? If so, have I missed any steps?

zjxiaohan:
@thef1stofhername Hi. Are you using Firefox? It looks like a browser compatibility issue. I've added a fix and tried chrome and firefox. Could you test again? Thanks!

thef1stofhername:
Yes I was using Firefox! I'll give it another  try. Thank you for the update :)

thef1stofhername:
sorry, I'm trying to get it to work but I can't. I'm on your branch, updated, refreshed and this is the result I get
uploadNewFileBug

zjxiaohan:
That's... weird...

Can I have your Firefox version? There shouldn't be a problem with Firefox version though... :(

thef1stofhername:
Ah sorry ;w;  it's Firefox 68.0 for Ubuntu 64 bit
I see your file changes and it *should* work :/

zjxiaohan:
Firefox 68.0 for Ubuntu 64 bit here too... :sweat_smile:

thef1stofhername:
That's super weird... I can keep trying?

zjxiaohan:
@thef1stofhername 

So these are what I've done so far:

1. Noticed that `confirm` was overridden by another function, so changed the name of your `confirm` function.
2. See it failing in Firefox. Turns out that if you use `File` constructor in firefox, there will be colons instead of slashes. Therefore it fails in the string matching, making the upload script believe that the file is not a duplicate. I removed the usage of `File` constructor.
Screenshot from 2019-07-17 23-37-45

I expect it to work like this in Firefox.
Screenshot from 2019-07-17 23-34-49

I would appreciate it if you could find something different between my run and yours... It's a weird case though.

thef1stofhername:
yeah I'll go through the debugger/console and see that's going on

thef1stofhername:
so from what i can see, it is only comparing the name of the file to the full paths and NOT binding the file name to the path of the place you want to put it before comparing it so of course it will accept it every time. I have no idea why this is happening

thef1stofhername:
I just remade my local version of your branch and that seemed to do it!

Issue #3456, opened 2019-03-23
Discussion forum clear filter and display when filtering


andrewaikens87:
We should clear the filter on the discussion forum when we navigate to this page (this would be when the discussion forum on the side bar is clicked, possibly the breadcrumb as well). When filtering we should display to the user that currently there are filters applied.

PR #4082, opened 2019-07-13
[Feature:Forum] Show applied filters in forum bar

xprilion:

Fixes: #3456

### What is the new behavior?
Screenshot from 2019-07-13 23-30-37


xprilion:
@bmcutler @andrewaikens87 looking for design feedback on this! (meanwhile pushing some more commits for clearing filters on Clicking left nav)

thef1stofhername:
So the application of longer filters or multiple wrap into the forum. There is nothing wrong with the placement of the filter, but it should not wrap into the forum posts themselves. Some sites I visit just have a clear filter option to quick access and do not list the filters applied, but I think in the case of a forum it is helpful to have the filter parameter displayed. You could even have Filters: persist without a filter applied and in a little box next to the button.
Screenshot_2019-07-17 Submitty Sample Discussion Forum

Also when you clear the filters and then click one that was selected before it does the same as unapplying the filter and you have to click again to apply it. To me, if I cleared all the filters, I'd want to be able to start from scratch, as if I'd unapplied all the filters. But it this was done intentionally because it made more sense to most users, let me know.
clearFunctionality

site/app/templates/forum/ForumBar.twig
```diff
@@ -61,15 +60,19 @@
 
 						{% endif %}
 					{% endfor %}
+
+					<span id="forum_filters">Filters: <span id="filter_names">None</span> &middot; <a onclick="clearForumFilter()" id="filter_close">Clear </a></span>
```
thef1stofhername: this does not wrap well, especially if you have multiple options enabled (and I have no idea if we will be adding more options in the future

xprilion: @thef1stofhername I am not very good with frontend..hence would need your suggestion or pointer on how to make this better :(

xprilion:
@thef1stofhername It is really tough for me to make those filters appear in a single line, without breaking the current structure. I believe I am not very well versed with Submitty's CSS part. Could you suggest me something which might help me with making that appear on a single line without breaking?

(I even changed the col span of the forumbar just to accomodate the filters text there! :sweat_smile: )

thef1stofhername:
I'll see what I can do! Let me know if you make any progress in the meantime :)

site/app/templates/forum/FilterForm.twig
```diff
@@ -23,32 +23,69 @@
             modifyThreadList({{ current_thread }}, '{{ current_category_ids|join("|") }}', '{{ current_course }}', loadFirstPage, success_callback);
             {% endif %}
         }
+
         $( document ).ready(function() {
```
andrewaikens87: Can we add this into forum.js?

site/app/templates/forum/FilterForm.twig
```diff
@@ -23,32 +23,69 @@
...
+                $('#forum_filters').css('display', 'inline-block');
+            }
+
             $('#unread').prop('checked', {{cookie_selected_unread_value}});
             $("#tree").prop("checked", true);
             {% if display_option in ['tree', 'time', 'alpha'] %}
             $("#{{ display_option }}").prop("checked", true);
             $("#{{display_option }}_label").addClass('active');
             {% endif %}
+
```
andrewaikens87: Please remove white space

site/app/templates/forum/ForumBar.twig
```diff
@@ -88,6 +91,8 @@
 		{% endfor %}
 
 
+
```
andrewaikens87: Please remove white space.

site/public/js/forum.js
```diff
@@ -1105,3 +1106,34 @@ function loadThreadHandler(){
         });
     });
 }
+
+var filters_applied = [];
```
andrewaikens87: Is this supposed to be within loadThreadHandler? Or in a function at all? If my understanding is correct, the intent is not to have this within a function? I'm mostly asking for clarification here since the merge conflict relates to this chunk.

site/app/templates/forum/FilterForm.twig
xprilion: Done!

site/public/js/forum.js
xprilion: Yes, the idea was to have it global to make it accessible to multiple pages!

andrewaikens87:
Just wanted to clarify this from our discussion on Tuesday.

We should remove the popup for filtering. The new row should replace it, where we can toggle categories on or off (similar to how we do so when we create a thread). 

I envision this PR looking something like this (not set on the on/off filter button):

vision2

As I send this screenshot I'm wondering if it would be useful to "deprecate" categories, but this can be a different discussion. @bmcutler, @thef1stofhername has already approved this. Should we merge this as is and then make my suggestions as a v2? Or just change it to included the changes described in the screenshot above?

bmcutler:
Yes, I like the screenshot Andrew posted.  Also need to add 'comment', 'unresolved', and 'resolved'.

For now, I propose that when the filter is enabled, that bar of options is displayed.  When you click the filter button again, the bar goes away and the filter is cleared.  I don't think it's important to hide the bar but leave the filter enabled.

When you say deprecated, do you mean realize that a category is not present in the currently displayed items?  That clicking it will have no impact on what is displayed?  I don't think that extra work is necessary.  If people request it we can revisit.

xprilion:
So now the filtering is radically different as when this PR started. 

https://www.loom.com/share/7c8061df2e3e4644abeeab769da99024

There's this pretty little bar which displays options and then it persists till asked to clear and also the number of threads in any given filtering is displayed! :smile:

bmcutler:
Yay, I like this alot, I'm going to ask for some (hopefully?) small tweaks.

filter_revised

Let's move the clear filter to the main bar, next to the filter button.  If the filter is 'open'/enabled, AND at least one category or modifier is selected, then clear filter appears.  I think it should not be a button, but rather just text with an X.  What do others think?

When you press the category botton, let's make the button background be the category color (like when you make a new thread & choose the category).
 
When you press one of the other modifier buttons  (unread, comment, unresolved, or resolved), those buttons should also fill with a solid color, maybe the Submitty dark blue??  What do others think?  Maybe a medium blue?  I don't really like the yellow highlight color here.

I'm not sure what the comment-unresolved-resolved buttons are doing when pressed in combination right now.  I think you should only be able to press at most 1 at a time.  That is, if "comment" is pressed and you press "resolved", comment should be unselected.

Finally, I think we should teach people what the yellow ? and green check mean.  (Surprising # of people don't know.) . Should we put those symbols on the unresolved and resolved buttons in thread filter?  Symbol plus text?  Or even just symbol (with hover text?) . What do others think?

xprilion:
@bmcutler I would say the multiple selection of the status buttons is useful. It's like negating a single status. 

Like, if I want to see all threads except the ones that are resolved, I'll check both Unresolved and comments

site/public/js/forum.js
```diff
@@ -484,8 +497,21 @@ function alterShowMergeThreadStatus(newStatus, course) {
 }
 
 function modifyThreadList(currentThreadId, currentCategoriesId, course, loadFirstPage, success_callback){
-    var categories_value = $("#thread_category").val();
-    var thread_status_value = $("#thread_status_select").val();
+
+    var categories_value = [];
```
andrewaikens87: This is a duplicate of line 380. Let's consolidate the code here.

site/app/templates/forum/searchResults.twig
```diff
@@ -24,8 +24,8 @@
                     </tr>
                     {% for post in thread.posts %}
                         <tr title="Go to post" class="cursor-pointer" onclick="window.location = '{{ post.post_link }}';" id="search-row-{{ post.count }}" class="hoverable">
-                            <td align="left"><pre class='pre_forum'>
-                                    <p class="post_content pre-wrap">
+                            <td align="left"><pre style="max-width: 67vw">
```
andrewaikens87: Why the change to 67vw?

site/app/templates/forum/searchResults.twig
```diff
@@ -8,7 +8,7 @@
         "show_more" : False
     } %}
 
-    <div id="search_wrapper">
+    <div id="search_wrapper" style="max-width: 67vw">
```
andrewaikens87: Why the change to 67vw?

site/public/js/forum.js
xprilion: Moved to a function!

site/app/templates/forum/searchResults.twig
xprilion: Because 70vw is not working. Since we're probably going to do an Ajax search soon, I'm seeing this as a quick solution for something that will be removed soon!

PR #4083, opened 2019-07-13
[Bugfix:Email] more details for announcements, etc.

sak6lab:
Forum notification email bodies lack detail

### What is the new behavior?
Forum notifications now have detailed bodies.
- Announcements and new threads include the post content
- Posts include parent thread content and thread title
- Deleted and undeleted posts/threads include post content and thread title
- edited posts include new post content and what thread it is in
- edited threads include new post content and new thread title


MasterOdin:
Do any of these notifications include a link to where they are on Submitty?

bmcutler:
Yes, the plan is to include links to appropriate pages from both notifications (many are in place already?) and emails.  But I don't believe any emails have links yet.  Better to wait for the router refactor to finish before we add links.

And in general, no, I am not in favor of shortening the content of the email and forcing people to log in to read the announcement.  However, some data (e.g., grades) should not be included in the message body.  We'll keep sensitive info password-protected on the site.

sak6lab:
I changed notifications' content to be the same as the email subject to avoid having long notification content.

site/app/controllers/forum/ForumController.php
```diff
@@ -388,14 +388,16 @@ public function publishThread(){
                 $metadata = json_encode(array(array('component' => 'forum', 'page' => 'view_thread', 'thread_id' => $thread_id)));
                 // notify on a new announcement
                 if ($announcement) {
-                    $content = "New Announcement: ".$thread_title;
-                    $event = ['component' => 'forum', 'metadata' => $metadata, 'content' => $content, 'subject' => 'New Announcement'];
+                    $subject = "New Announcement: ".Notification::textShortner($thread_title);
+                    $content = "An Instructor/TA made an announcement in the Submitty discussion forum:\n\n".$thread_post_content;
```
thef1stofhername: Per Barb's suggestion here and in the else I think it would be a good idea to include the unshortened title at the beginning of the content in case vital information or the only information is in the title

I think in every other place the title is included in the content somehow but feel free to correct me

site/app/controllers/forum/ForumController.php
```diff
@@ -575,6 +581,7 @@ public function alterPost($modifyType){
                 }
             } else {
                 if($status_edit_thread && $status_edit_post) {
+                    $type = "Thread and Post";
```
andrewaikens87: Thoughts on using an enum here instead (possibly forum alter intent)? I know php doesn't have native support. Possibly introduce an abstract class with const members? Just a thought, would like to get your opinion. I'm personally not the biggest fan of the string comparison down below.

bmcutler:
@sak6lab 
Please re-test all of my wording changes and make sure they look ok.  Then I'll merge this.

site/app/controllers/forum/ForumController.php
sak6lab: I also do not like this string comparison, but I feel like this function in general should be split up into alterThread and alterPost since there a multiple places where there is a logic control block checking if it is a thread or a post.

sak6lab:
Tested all of the event and checked email bodies in database. They are definitely a lot more clearer!

PR #4099, opened 2019-07-15
[Refactor:Router] New URL for LateController

zjxiaohan:
### URL Changes

**Old:** 
- `/index.php?semester=f19&course=sample&component=admin&page=late&action=view_late`
- `/index.php?semester=f19&course=sample&component=admin&page=late&action=view_extension`
- `/index.php?semester=f19&course=sample&component=admin&page=late&action=update_late`
- `/index.php?semester=f19&course=sample&component=admin&page=late&action=delete_late`
- `/index.php?semester=f19&course=sample&component=admin&page=late&action=update_extension`
- `/index.php?semester=f19&course=sample&component=admin&page=late&action=get_extension_details&g_id=open_homework`

**New:** 
- `/f19/development/late_days` GET
- `/f19/development/extensions` GET
- `/f19/development/late_days/update` POST ONLY
- `/f19/development/late_days/delete` POST ONLY
- `/f19/development/extensions/update` POST ONLY
- `/f19/development/extensions/open_homework` GET

### Tested Entrances
- As instructor, went to the page from the sidebar and did all the operations possible. (update single/csv for late days & extensions, delete late days)

### Other Info
Splitted `update` into three functions - `updateLateDays`, `deleteLateDays`, `updateExtension`.

### DIFF
--- a/site/app/controllers/GlobalController.php
+++ b/site/app/controllers/GlobalController.php
@@ -217,14 +217,14 @@ class GlobalController extends AbstractController {
 
             if ($this->core->getUser()->accessAdmin()) {
                 $sidebar_buttons[] = new Button($this->core, [
-                    "href" => $this->core->buildUrl(array('component' => 'admin', 'page' => 'late', 'action' => 'view_late')),
+                    "href" => $this->core->buildNewCourseUrl(['late_days']),
                     "title" => "Late Days Allowed",
                     "class" => "nav-row",
                     "id" => "nav-sidebar-late-days-allowed",
                     "icon" => "fa-calendar-check"
                 ]);
                 $sidebar_buttons[] = new Button($this->core, [
-                    "href" => $this->core->buildUrl(array('component' => 'admin', 'page' => 'late', 'action' => 'view_extension')),
+                    "href" => $this->core->buildNewCourseUrl(['extensions']),
                     "title" => "Excused Absence Extensions",
                     "class" => "nav-row",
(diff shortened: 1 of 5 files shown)


zjxiaohan:
@jmcknight98 Thanks for your review.

> As an instructor, going to /f19/development/extensions/{any_string} (including /update that you mentioned in pr) displays something that seems unintentional (see image below) instead of going to the home page like most invalid urls such as /f19/development/late_days/update.

This endpoint is meant for Ajax instead of human access. It returns extensions of a certain gradeable. While we could `tryGetGradeable` to verify the validity of the gradeable id and deny access if it is an invalid name. Personally I think there is no necessity of adding validity check here because (1) the gradeable id is selected from a predefined set if the instructor doesn't try to maliciously modify the frontend. (2) Even if they do and the gradeable id is invalid, there will be no data returned anyway.

> Also, when a student goes to most of the new urls, it shows "You don't have access to this page". However for the 3rd and 4th new link (/f19/development/late_days/{any_string}) it goes to the home page. This doesn't really matter as students shouldn't be going to these urls, but it would be nice if it was consistent.

It is not the case that student doesn't have access to the 3rd or 4th new link. It's that the router can't find the link because the student is sending a GET request of those links which do not exist. So it equals to going to an invalid URL which should return home page. To see that the student is denied access from the 3rd and 4th new link, you could try to send POST requests to these links as a student.

Issue #1921, opened 2018-05-03
instructor content timed release & user access control


bmcutler:
Use case:
Instructor would like to post/host .pdf or other files on the Submitty server.
And specify which users can access the file(s), and what time they are available.

Examples:
* A lab exercise that should be released/made available at the start time of the student's lab.  Each registration section of students has a different start time.
* A homework assignment that has different versions (perhaps a unique version for each student).

From the navigation page or the page of the assignment, the student is given a url/link, and the file under that link is only accessible if the timestamp & username is valid.

Need to design both the instructor interface for releasing this link.
And make sure its intuitive for how the student downloads it.  (link should be visible but greyed out with a release time?)

- [x] Interface for instructors to release files
- [x] Interface for students to download files
- [x] Allow instructors to release at a specific date
- [x] Don't show to students before that date
- [ ] Enforce date with access control (so students can't guess the urls)

bmcutler:
related to #865

Kevin-MM:
Confirmed students can view unreleased materials provided they have the url.

PR #4101, opened 2019-07-15
Fix file access for limited access graders and greater

Kevin-MM:
Relates to #1921 and #4071

Only instructor/admin level users may view unreleased files.  This had unintended consequences as now limited access graders were unable to view files they needed to grade, or see unreleased course materials.

### What is the new behavior?
Anyone with limited access grader or greater permissions may view submission files, and also view unreleased course materials.

### Other information?
Tested by logging into as 'grader', 'student', and 'instructor' inside vagrant VM to test which users could access which files.

### DIFF
--- a/site/app/controllers/MiscController.php
+++ b/site/app/controllers/MiscController.php
@@ -95,11 +95,15 @@ class MiscController extends AbstractController {
                 return false;
             }
 
-            // If the user attempting to access the file is not at least a grader then ensure the file has been released
-            if(!$this->core->getUser()->accessGrading() AND !CourseMaterial::isMaterialReleased($this->core, $path))
+            // If attempting to obtain course materials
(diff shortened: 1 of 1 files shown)


bmcutler:
Haven't tested, but this looks like it will solve the problem.

But I think it's not quite the right fix.  The isMaterialsReleased function should only be responsible for checking files that are in the course_materials directory:
/var/local/submitty/courses/<SEMESTER>/<COURSE>/uploads/course_materials/

Can you please add a check in that function (or before the call to that function)?   This function should not be validating release dates for files not in that directory.

Otherwise if we later want to change functionality/policy and say that unreleased course materials files are only available to full access grader & above we'll have this bug again.

Thanks

Kevin-MM:
I've updated the code so that the checks only run when attempting to obtain course materials.

PR #4097, opened 2019-07-15
[Bugfix:UI] Fix html for display seating guide email button

MasterOdin:
### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)

### What is the current behavior?

Fixes #4087 where display for email seating configuration button did not look right.
 
No gradeable selected:  
Screen Shot 2019-07-14 at 9 29 15 PM

---
Gradeable selected:
Screen Shot 2019-07-14 at 9 29 28 PM

### What is the new behavior?
No gradeable selected:  
Screen Shot 2019-07-14 at 9 30 12 PM

---
Gradeable selected:
Screen Shot 2019-07-14 at 9 30 26 PM

PR #4104, opened 2019-07-15
Fixed empty button in seating

mktung:
We were accidentally hiding the wrong element


zjxiaohan:
I am a bit confused about this... See: https://github.com/Submitty/Submitty/pull/4097

It looks like we are waiting for a whole page UI refresh?

MasterOdin:
This also retains the button inside a button design which looks (to me) weird:
Screen Shot 2019-07-14 at 9 29 28 PM

mktung:
I fixed the button border, but like Han said I'm not sure if we should be merging this or not? Unclear what the objective is

bmcutler:
Maya,
Please address this issue when you refactor the page for mobile.  
Thanks

PR #4105, opened 2019-07-15
[VPAT:InstructorUI] Made "delete gradeable" button accessible

### LINKED ISSUE #4076: [VPAT:InstructorUI] Device Dependent Event Handler in Gradeables List
The home view that lists available gradeables also has controls to edit (wrench) and delete (trashcan).  The delete control (trashcan) is not keyboard accessible.  WAVE is reporting this as a "Device Dependent Event Handler".  Meaning that this control is only accessible via mouse.
Screen Shot 2019-07-12 at 6 32 34 PM

Looking at the site inspector in Chrome, it appears that the delete function is solely handled by an `onclick()` event handler.  (see lower area, highlighted line).  This will prevent keyboard access to the delete control.
Screen Shot 2019-07-12 at 8 45 14 PM

mktung:
Closes #4076 

Put button inside <button> for semantics and accessibility (should be able to tab and trigger using keyboard)

### DIFF
--- a/sbin/send_email.py
+++ b/sbin/send_email.py
@@ -12,6 +12,43 @@ import json
 import os
 import datetime
 from sqlalchemy import create_engine, MetaData, Table, bindparam
+import sys
+import psutil
+
+
+# ======================================================================
+#
+# Let's make sure we're the only copy of this script running on the
+# server.  Multiple copies might happen if sending emails is slow or
+# hangs and takes longer than 1 minute and the cron job fires again.
+#
+
+# We could just match the program name, but this is problematic if
+# happens to match the filename submitted by a student.
+# my_program_name = sys.argv[0].split('/')[-1]
+
+# So instead let's match the full path used in the cron script
+my_program_name = sys.argv[0]
+
+my_pid = os.getpid()
+
+# loop over all active processes on the server
+for p in psutil.pids():
+    try:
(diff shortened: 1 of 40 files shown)


mktung:
I've made it more noticeable that we can click the button, but I'm not sure if we can use the render button system here. On-click, the delete button calls a JavaScript function to generate a confirmation form. Is there a way we could call that through href? If not, I think it'd be better to leave this as an exception.

shailpatels:
> I've made it more noticeable that we can click the button, but I'm not sure if we can use the render button system here. On-click, the delete button calls a JavaScript function to generate a confirmation form. Is there a way we could call that through href? If not, I think it'd be better to leave this as an exception.

Yes you can execute a JS function through a hyperlink using
 ```html
<a href="javascript:foo('test');">Execute JavaScript</a>
```

You should be able to construct the link since PHP will know all the information at the time and pass it in.

mktung:
The modal keyboard accessibility is another issue. I think everything else should be good in this latest commit

mktung:
My latest commit should fix the delete button. Also my modals aren't that big? Does your CSS look like this image

shailpatels:
image

This is what I'm seeing on firefox, can you try pulling master into your branch, I'm not sure if the issue is from this branch or somewhere else, I'll look more into that later

Issue #4088, opened 2019-07-14
Datepicker not enabled in late days page


zjxiaohan:
**What problem are you trying to solve with Submitty**
The "Single Student Entry" form in "Late Days Allowed" page doesn't have the nice date picker.
![Screenshot from 2019-07-14 14-06-14](https://user-images.githubusercontent.com/20266703/61179960-91751d00-a640-11e9-90c8-48cc4f0b8588.png)

**Describe the way you'd like to solve this problem**
Let's add it just like the extension page.

Issue #3402, opened 2019-03-16
Late Day Datestamp Bug


zjxiaohan:
When using "Single Student Entry" in "Late Days Allowed" page, if "datestamp" is filled with "12/12/97", then the effective date will display as "12/12/1997", while "12/12/30" makes "12/12/2030".

The bug may be related to the handling of unix time, as year 1970 seems to be the dividing line in this scenario.

This issue may be related to #1301 .

![timestamp](https://user-images.githubusercontent.com/20266703/54472009-fac24d80-477e-11e9-9061-4f123ed89493.png)

zjxiaohan:
closed by #4114

PR #4114, opened 2019-07-15
[Bugfix:InstructorUI] Added date picker to late days page

Fenn-CS:
The "Single Student Entry" form in "Late Days Allowed" page doesn't have the nice date picker.
Fixes: #4088 
### What is the new behavior?
Late days page now has a date picker.
Screenshot_2019-07-15 Submitty Sample Late Days Allowed


site/app/templates/admin/LateDays.twig
```diff
@@ -77,4 +77,37 @@
...
+                                case 1: 
+                                    date = new Date("01/01/98");
+                                    break;
+                            }
+                            fp.setDate(date, true);
+                        }
+                    }
+                )],
+            allowInput: true,
+            enableTime: true,
```
zjxiaohan: Let's not `enableTime` or `enableSeconds`. They are not used in the timestamp on this page.

Fenn-CS: ok, cool..

site/app/templates/admin/LateDays.twig
```diff
@@ -77,4 +77,34 @@
...
+                        ],
+                        label: "or",
+                        onClick: (index, fp) => {
+                            let date;
+                            switch (index) {
+                                case 0: 
+                                    date = new Date();
+                                    break;
+                                case 1: 
+                                    date = new Date("01/01/98");
```
zjxiaohan: I noticed that `01/01/98` seems to be interpreted as `1998-01-01` in the date picker. Should be 9998 though.

That being said, consider https://github.com/Submitty/Submitty/issues/3402, it looks the year will be set to 1998 anyways. @bmcutler do we want to use four-digit year in the late days page to address #3402 in another PR?

Fenn-CS: Ah! yeah... Looks like we would have to modify some operations done on the date should we change the format, let's hear what Barb has to say.

shailpatels: Why are we using "01/01/98" here? the rest of the flatpickr "end of time" dates are 9998-01-01 which works fine.

zjxiaohan:
> I haven't looked to closely but I think this is the only calendar input expecting this type of format so we should just standardize it with everything

Yeah I grepped and only found `MM/DD/YY` occurrences in late days.

bmcutler:
We _should_ be using 4 digit years everywhere on the site...

Fenn-CS:
@zjxiaohan I have a added a four-place year format support that would be validated in :

https://github.com/Submitty/Submitty/blob/2a041906b4603417adea01a7fbf5a614cbb56e48/site/app/libraries/DateUtils.php#L48

I think to change the format to anything out of what is validated in this function would mean modifying this `DateUtils` validate function, which is good for a separate issue if necessary also if it's used for any calculation somewhere the calculation may break, so this is the safest fix wrt to this PR.

shailpatels:
@Fenn-CS I think people are focusing too much on the issue of standardizing dates too far for this issue. I meant in my original comment to use the default date that is outputted by flatpickr which by default uses a 4 digit year by default. 

The only issue left on this PR is changing the form to accept the date format outputted by the flatpickr you don't have to worry about dates and validating them across the rest of the system and fix the merge conflict please.

site/app/templates/admin/LateDays.twig
```diff
@@ -77,4 +77,34 @@
...
+                        ],
+                        label: "or",
+                        onClick: (index, fp) => {
+                            let date;
+                            switch (index) {
+                                case 0: 
+                                    date = new Date();
+                                    break;
+                                case 1: 
+                                    date = new Date("01/01/1998");
```
zjxiaohan: Still needs to change to 9998 instead of 1998

site/app/controllers/admin/LateController.php
```diff
@@ -25,13 +25,22 @@ public function run() {
      * @return Response
      */
     public function viewLateDays() {
+<<<<<<< HEAD
```
zjxiaohan: merge conflict isn't resolved yet

Fenn-CS: Looks like I did not commit properly. Please check one more time. :)

site/app/templates/admin/LateDays.twig
```diff
@@ -77,4 +77,34 @@
...
+                                case 1: 
+                                    date = new Date("01/01/9998");
+                                    break;
+                            }
+                            fp.setDate(date, true);
+                        }
+                    }
+                )],
+            allowInput: true,
+            dateFormat: "m-d-Y"
```
shailpatels: Please change this to "Y-m-d" so it matches the rest of the date pickers, or change the rest of the date pickers to match this date format

PR #4115, opened 2019-07-16
[Refactor:Router] New URL for GradeableController

zjxiaohan:
### URL Changes

**Old:** 
- `/index.php?semester=f19&course=sample&component=admin&page=gradeable&action=upload_config`
- `/index.php?semester=f19&course=sample&component=admin&page=gradeable&action=process_upload_config`
- `/index.php?semester=f19&course=sample&component=admin&page=gradeable&action=delete_config&config=something`
- `/index.php?semester=f19&course=sample&component=admin&page=gradeable&action=rename_config`
- `/index.php?semester=f19&course=sample&component=admin&page=gradeable&action=check_being_used?config=something`

**New:** 
- `/f19/sample/gradeable_config` GET
- `/f19/sample/gradeable_config/upload` POST ONLY
- `/f19/sample/gradeable_config/delete` POST ONLY
- `/f19/sample/gradeable_config/rename` POST ONLY
- `/f19/sample/gradeable_config/usage?config_path=somepath` GET

### Tested Entrances
- As instructor, went to the page from the sidebar and did all the operations possible: upload, delete, rename used/unused)

### Other Info
- Changed the name from `GradeableController` to `GradeableConfigController` which I deem more proper.
- `strpos($gradeable->getAutogradingConfigPath(), $path) !== false` is used in several places to check if the config is used, which is problematic because the config starting with same names will be confused. e.g. config `345` is used by `open_homework` => config `3` is used by `open_homework`.
- Added a pop up for deletion because I want deletion to be POSTed.
- Renamed function names to be camelCase.
- Used `FileUtils::recursiveRmdir` instead of `recursive_rmdir`

### DIFF
--- a/site/app/controllers/AdminController.php
+++ b/site/app/controllers/AdminController.php
@@ -3,7 +3,6 @@
 namespace app\controllers;
 
 use app\controllers\admin\ReportController;
-use app\controllers\admin\GradeableController;
 use app\controllers\admin\GradeablesController;
 use app\controllers\admin\AdminGradeableController;
 use app\controllers\admin\UsersController;
@@ -25,9 +24,6 @@ class AdminController extends AbstractController {
         switch ($_REQUEST['page']) {
             case 'users':
                 $controller = new UsersController($this->core);
-                break;
-            case 'gradeable':
-                $controller = new GradeableController($this->core);
                 break;
             case 'late':
                 $controller = new LateController($this->core);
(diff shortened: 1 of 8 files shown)


bmcutler:
Perhaps instead of gradeable_config, we should call this autograding_config?

Stepping back our new thinking [w/ @prestoncarman] for gradeable configuration is in 3 parts, to facilitate re-use in future terms (or even within the same term).  The 3 parts:
1. The autograding configuration
2. The manual grading rubric
3. The 'instance', which includes the various release & due dates
And these pieces can be separately exported (e.g., as json), manually edited if desired, and uploaded/imported to form a new gradeable.

So this UI is for the first part.

Thoughts?

zjxiaohan:
tried `AutogradingConfigurationController` but felt it's too long... Let it just be config...

PR #4116, opened 2019-07-16
[Feature:Developer] twig debug extension on developer environments

shailpatels:
Adds the the Twig DebugExtension if the debug environment is set to true, this will allow the dump function in twig files for development

### DIFF
--- a/site/app/libraries/Output.php
+++ b/site/app/libraries/Output.php
@@ -77,6 +77,11 @@ class Output {
             'cache' => $this->core->getConfig()->isDebug() ? false : $cache_path,
             'debug' => $this->core->getConfig()->isDebug()
         ]);
+
+        if($this->core->getConfig()->isDebug()){
+            $this->twig->addExtension(new \Twig\Extension\DebugExtension());
+        }
+        
         $this->twig->getExtension(\Twig\Extension\CoreExtension::class)
             ->setTimezone($this->core->getConfig()->getTimezone());
         $this->twig->addGlobal("core", $this->core);
--- a/site/app/libraries/Output.php
+++ b/site/app/libraries/Output.php
@@ -79,7 +79,7 @@ class Output {
             'debug' => $debug
         ]);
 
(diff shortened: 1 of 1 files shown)


bmcutler:
Can we add a lint-like travis test to check for dump() in *twig and console.log() in *js?

site/app/libraries/Output.php
```diff
@@ -78,6 +78,12 @@ public function loadTwig($full_load = true) {
             'cache' => $debug ? false : $cache_path,
             'debug' => $debug
         ]);
+
+        if($this->core->getConfig()->isDebug()){
```
MasterOdin: Actually, now that I'm looking at this again, change this to `if ($debug)` to use the variable defined above.

MasterOdin:
Yeah, we can add a linter for JS (like [eslint](https://eslint.org/)) to prevent things like console.log in JS files, but I'm not sure there's a good mature solution for Twig (there's [twigcs](https://github.com/allocine/twigcs), but it's pretty early in development cycle).

I could do something like Symfony itself does which is load a "production" twig environment and then just load all twig environments and if any fail, then fail. Looking into this, it looks like in production (e.g. where `debug = false`), attempting to load a template that uses `{{ dump(...) }}` will throw an exception.

shailpatels:
> Yeah, we can add a linter for JS (like [eslint](https://eslint.org/)) to prevent things like console.log in JS files, but I'm not sure there's a good mature solution for Twig (there's [twigcs](https://github.com/allocine/twigcs), but it's pretty early in development cycle).
> 
> I could do something like Symfony itself does which is load a "production" twig environment and then just load all twig environments and if any fail, then fail. Looking into this, it looks like in production (e.g. where `debug = false`), attempting to load a template that uses `{{ dump(...) }}` will throw an exception.

Yeah if debug is set to false, {{ dump(...) }} will throw a frog robot PHP exception and I think it'll say something along the line of "unknown "dump()" function".

PR #4121, opened 2019-07-16
VPAT fix missing label error in configuration.twig

pbailie:
Fixed missing `<label>` error in configuration view.


bmcutler:
@mktung Please make sure you include this code change in your larger refactor of this page.
(then I think we can probably close this PR.)

PR #4122, opened 2019-07-16
[Bugfix:System] Check for duplicate copy of send_email.py script

bmcutler:
If the send_email script is slow or hangs for more than a minute, it is possible 
it is still running when the cron job kicks off the next instance.   We could end up 
sending duplicate copies of the same email, or worse.

### What is the new behavior?
At the start, the script checks to see if another send_email.py program is running 
and if so, exits immediately.

### DIFF
--- a/sbin/send_email.py
+++ b/sbin/send_email.py
@@ -23,7 +23,13 @@ import psutil
 # hangs and takes longer than 1 minute and the cron job fires again.
 #
 
-my_program_name = sys.argv[0].split('/')[-1]
+# We could just match the program name, but this is problematic if
+# happens to match the filename submitted by a student.
+# my_program_name = sys.argv[0].split('/')[-1]
+
+# So instead let's match the full path used in the cron script
+my_program_name = sys.argv[0]
+
 my_pid = os.getpid()
 
 # loop over all active process on the server
--- a/sbin/send_email.py
+++ b/sbin/send_email.py
@@ -12,6 +12,32 @@ import json
 import os
 import datetime
 from sqlalchemy import create_engine, MetaData, Table, bindparam
+import sys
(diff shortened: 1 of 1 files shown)


MasterOdin:
Would it make sense to have a concept of just having a "pid file" that gets created when this script runs and then deleted when it finishes, and then other instances can just check for the existence of that file before attempting to run? It's a fairly common paradigm and I think would be easier/quicker than mucking about with the process list.

I'd imagine we can make a folder `/var/local/submitty/run` or `/var/local/submitty/pid` or whatever that houses these things, with ownership being `root:submitty_daemon` I think for now and can expand it as necessary.

bmcutler:
Yeah the file would be an option, but I'm worried about the script crashing & not cleaning up the file.
This is dumb simple, and quite fast despite the loop over all pids.   We used to do this for the autograding before shipper/worker refactor and it was robust.  I'd rather not overengineer this minor piece :)

sbin/send_email.py
```diff
@@ -12,6 +12,43 @@
...
+for p in psutil.pids():
+    try:
+        cmdline = psutil.Process(p).cmdline()
+        # if anything on the command line matches the name of the program
+        for i in cmdline:
+            if i.find(my_program_name) != -1:
+                if p != my_pid:
+                    print("ERROR!  Another copy of '" + my_program_name +
+                          "' is already running on the server.  Exiting.")
+                    exit()
```
MasterOdin: you should use either `sys.exit()` or `raise SystemExit("error message...")` here. The `exit` and `quit` functions shouldn't really be used in a program per the [python docs](https://docs.python.org/3/library/constants.html#constants-added-by-the-site-module).

PR #4125, opened 2019-07-17
[Refactor:Router] New URL for NotificationController

zjxiaohan:
### URL Changes

- GET `/f19/sample/notifications` Notification Page
- GET `/f19/sample/notifications?show_all=true` All Notifications Page
- GET `/f19/sample/notifications/1` View Notification nid=1 and Mark as Seen
- GET `/f19/sample/notifications/1?seen=1` View Notification nid=1
- GET `/f19/sample/notifications/1/seen` Mark Notification nid=1 as seen
- GET `/f19/sample/notifications/seen` Mark All as Seen
- GET `/f19/sample/notifications/settings` Notification Settings Page
- POST `/f19/sample/notifications/settings` Notification Settings Update

### Testing
Manually tested all the routes above, including view unseen notifications, view all notifications, mark all as seen, mark single one as seen, view single notification, see settings page and update settings.

### DIFF
--- a/site/app/templates/Notifications.twig
+++ b/site/app/templates/Notifications.twig
@@ -2,13 +2,13 @@
 
 <div class="content">
     <div style="float: right; margin-bottom: 20px;">
-         <a href="{{ core.buildUrl({"component": "notification", "page": "notifications", "action": "mark_all_as_seen"}) }}" class="btn btn-primary">Mark all as seen</a>
+         <a href="{{ mark_all_as_seen_url }}" class="btn btn-primary">Mark all as seen</a>
         {% if show_all %}
-            <a href="{{ core.buildUrl({"component": "notification", "page": "notifications"}) }}" class="btn btn-primary">Show unread only</a>
+            <a href="{{ notifications_url }}" class="btn btn-primary">Show unread only</a>
         {% else %}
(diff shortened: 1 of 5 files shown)


site/app/templates/Notifications.twig
```diff
@@ -2,13 +2,13 @@
 
 <div class="content">
     <div style="float: right; margin-bottom: 20px;">
-         <a href="{{ core.buildUrl({"component": "notification", "page": "notifications", "action": "mark_all_as_seen"}) }}" class="btn btn-primary">Mark all as seen</a>
+         <a href="{{ mark_all_as_seen_url }}" class="btn btn-primary">Mark all as seen</a>
```
sak6lab: Is core still accessible here? I had this question before but could not seem to figure how this twig file had access to core even if it is a global variable.

zjxiaohan: Not sure if I understood it correctly but:

**Is core accessible from twig file?**

Yes.

https://github.com/Submitty/Submitty/blob/5d8bc25d3c0bc38907d89f2291eeca05d31c4234/site/app/libraries/Output.php#L87

**Should we use twig to access core?**

No.

While it looks harmless to do so it is generally not an encouraged practice to mix twig template with php code (that's what twig is for). It is part of a working in progress issue (#2273) to remove that.

PR #4131, opened 2019-07-17
Course materials date picker timezone handling

### LINKED ISSUE #4065: Course material release time is not time zone sensitive

Kevin-MM:
Closes #4065

Address issues outlined in #4065.

**To Test**
- Set your PC timezone to the same time zone as the server. *This is default behavior for vagrant users in New York*
- Navigate to ```Course Materials```
- Upload some files and observe the behavior of the date time picker when using it
- Change your PC timezone to some timezone outside of server timezone
- Refresh the page
- Observe the behavior of the date time picker


bmcutler:
@Kevin-MM 
Can you please add a screenshot of the two cases?  (matching & not matching timezone).
For dev documentation purposes.
thanks!

Kevin-MM:
**In server timezone**
in_time_zone
**Outside of server timezone**
out_of_time_zone

site/app/libraries/DateUtils.php
```diff
@@ -142,4 +142,25 @@ public static function parseDateTimeLong(string $date_time, $default_time_zone =
...
+     * Gets a json which contains the current server time broken up into specific fields
+     * Formatting the data in this manner makes it easier to work with when instantiating javascript Date() objects
+     *
+     * @param $core Application core
+     * @return object
+     */
+    public static function getServerTimeJson($core) {
+        $time = new \DateTime('now', $core->getConfig()->getTimezone());
+
+        $json = (object)[];
```
zjxiaohan: I am not sure why we have to use `$time->format('x')`-like code for multiple times.

If, for PHP code, put `$time->format("Y-m-j G:i:s")` and return it via JSON, you'll get things like `2019-07-18 1:01:02`, and it's good for JavaScript to accept such a string via `new Date('2019-07-18 1:01:02')`.

site/app/templates/course/CourseMaterials.twig
```diff
@@ -316,6 +350,44 @@
         });
     }
 
+    function handleTimeZones() {
+
+        var url = buildUrl({'component': 'misc', 'page': 'get_server_time'});
+
+        $.ajax({
+            type: "POST",
```
zjxiaohan: Note that, although it doesn't hurt to use `POST` in this case, `POST` is generally considered to be creating/updating a resource.

The following content is from https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.3

The GET method means retrieve whatever information (in the form of an entity) is identified by the Request-URI. If the Request-URI refers to a data-producing process, it is the produced data which shall be returned as the entity in the response and not the source text of the process, unless that text happens to be the output of the process.

The POST method is used to request that the origin server accept the entity enclosed in the request as a new subordinate of the resource identified by the Request-URI in the Request-Line. POST is designed to allow a uniform method to cover the following functions:

      - Annotation of existing resources;
      - Posting a message to a bulletin board, newsgroup, mailing list,
        or similar group of articles;
      - Providing a block of data, such as the result of submitting a
        form, to a data-handling process;
      - Extending a database through an append operation.

site/app/libraries/DateUtils.php
Kevin-MM: I believe this is a matter of opinion, but I prefer it this way because the data is more modular.  Having to work between PHP date time formats and javascript date time formats is a nightmare.  Plus if we want to reuse this function in the future maybe we aren't interested in a complete timestamp.  Perhaps on the client side we are only interested in the day of the month and don't want to have to instantiate a javascript Date() object just to get the day back out again.

site/app/controllers/MiscController.php
```diff
@@ -41,9 +42,25 @@ public function run() {
...
+    /**
+     * Get the current server time
+     *
+     * Returns a json string which contains the current server time broken up into year, month, day, hour, minute,
+     * second
+     */
+    private function getServerTime() {
+
+        $json = DateUtils::getServerTimeJson($this->core);
+        $this->core->getOutput()->renderJson($json);
```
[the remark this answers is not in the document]
Kevin-MM: Noted, I will leave as is until your revisit.

site/app/libraries/DateUtils.php
zjxiaohan: Yeah that makes sense. True that we shouldn't be relying too much on the format consistency between PHP and JavaScript. Thanks for your explanation.

site/app/templates/course/CourseMaterials.twig
Kevin-MM: Noted

site/app/libraries/DateUtils.php
MasterOdin: It's also important to note that depending on your browser `new Date("datetime string")` gets rendered in different ways with regards to timezones compared to `new Date(year, month, day)`.

PR #4134, opened 2019-07-18
[Bugfix:System] Clean up permissions on logs directories

bmcutler:
* set correct permissions on logs directories (& refactor/reorganize code)
* replace uses of exit() in send_email.py with sys.exit()

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -286,22 +286,23 @@ fi
 # Set permissions on the top level logs directory
 chown root:${COURSE_BUILDERS_GROUP} ${SUBMITTY_DATA_DIR}/logs
 chmod 751                           ${SUBMITTY_DATA_DIR}/logs/
-# Set permissions for logs directories that exist on both primary & work machines
+# Set owner/group for logs directories that exist on both primary & work machines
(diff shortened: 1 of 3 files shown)


.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -251,18 +251,21 @@ if [ "${WORKER}" == 0 ]; then
...
+# ------------------------------------------------------------------------
+# Create the logs directories that exist on both primary & worker machines
 mkdir -p ${SUBMITTY_DATA_DIR}/logs
 mkdir -p ${SUBMITTY_DATA_DIR}/logs/autograding
-mkdir -p ${SUBMITTY_DATA_DIR}/logs/emails
 mkdir -p ${SUBMITTY_DATA_DIR}/logs/autograding/stack_traces
 mkdir -p ${SUBMITTY_DATA_DIR}/logs/bulk_uploads
-
-#Make site logging directories if not in worker mode.
+mkdir -p ${SUBMITTY_DATA_DIR}/logs/emails
```
MasterOdin: I think both emails and bulk_uploads are only things that happen on the primary machine and should be moved inside the if statement?

bmcutler: thanks! changed

PR #3882, opened 2019-06-19
[Feature:Autograding] Automated output and input generation

drumilpatel2000:
### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?

Presently instructor have to provide both output files and input files for grading.

### What is the new behavior?
After this PR instructor can provide give grading with both random output and random inputs. They have to add programs in instructor_solution and write commands for input and output generation in config.json.This will generate both inputs and output at time of autograding

### Other information?

1. Install submitty
2. run BUILD_development.sh script 
3. Run grades for Cpp simple lab Automated input output

PR #4137, opened 2019-07-18
[WIP]Add generated output

drumilpatel2000:

After PR #3882 is merged Random input and random output is generated at the time of grading but output generation without random input is always the same show it would be not a good method to generate it every time.

### What is the new behavior?
Generated_output will contain all output generated output at the time of building courses. This will only contain output which does not have random input


drumilpatel2000:
Closing this PR as @emaicus  New change in grade item infrastructure  broke the code

PR #4149, opened 2019-07-19
[Feature:Forum] Choice of plaintext vs markdown for post

### LINKED ISSUE #4124: Plaintext vs Markdown for Forum Post
We want to default to plaintext forum posts for simplicity, because we have students users who aren't familiar with markdown and will be confused if their use of basic math symbols turn into italics or bold and the symbols disappear.

When a user makes a new thread or new post in a thread there will be a checkbox at the top "Enable Markdown" that is initially false/unchecked.  

We store the state of the checkbox for the last new thread/post in a cookie (so it remembers whether this user knows how to use markdown).  It's a simple per browser/machine cookie -- not stored in the database per user.

We add a column to the database for every post storing the state of this checkbox at time of thread/post creation.  (The user can change the checkbox if they made an error by editing the post.  Editing a post does not use or change the state of the cookie.)

If the box is unchecked, the contents of the post will be rendered in plaintext.  All symbols will remain as originally typed.  Perhaps the only exception is that we will make a special case check for urls, they will turn into links.  (But just raw links, you cannot put different text for the link in plaintext mode.)

If the box is checked, then the entire post will be rendered as markdown (restricted markdown, no html tags allowed).  If the box is checked, then we have the shortcut buttons for code & url (& maybe some other options later) to help teach markdown syntax.

We have a link with the checkbox (in a ? bubble we use elsewhere on the site) to a tutorial on markdown.  Probably a simple tutorial with the most useful bits of markdown for the submitty forum that we write ourself on submitty.org.  But that tutorial will also link to other 'cheatsheet' pages with the full range of options allowed with our markdown package.

Note: Maybe later we add an option to preview the markdown rendering (similar to github).

Other comments @xprilion @andrewaikens87 @MasterOdin?

xprilion:

Fixes #4124

### What is the new behavior?

Checklist: 

* [x] ~Button to toggle markdown/plaintext~ Viewer cannot toggle, only editor sets
* [x] Option for markdown/plaintext during post/create-thread
* [x] Show buttons for easy markdown post
* [x] Create column in database for saving thread state
* [x] Update cookie during new post (not on edit)
* [ ] Add more helper buttons
* [ ] Link to tutorial
* [ ] Recognize links (no description of links available)
* [ ] Preview post (?)

#### 19 July: 
Button created to toggle plaintext/markdown (Should this be a checkbox as mentioned in issue?)

Plaintext:
Plaintext

Markdown:
Markdown

Creating thread / making reply: 

Plaintext:
Screenshot from 2019-07-20 02-46-44

Markdown:
Screenshot from 2019-07-20 02-46-51

#### 20 July 

Removed the option for markdown toggle on individual posts for the viewer. Only thread maker can set it. 

Added a 'render_markdown' column in 'posts' table for courses, but I am not yet sure if the migration I created for it is perfect. Will put on a vagrant destroy/up sequence tonight. 

Changed the markdown button to - 

Screenshot from 2019-07-21 02-51-58

### Other information?
Currently Untested


bmcutler:
I don't think we want the markdown toggle button as a viewer.  The viewer of a post doesn't need to be able to toggle the rendering.  (Trying to keep the forum clutter-free.  It already has so many buttons and icons!)

I like the editing interface though.  The M button instead of a checkbox is ok.  Should we  use the font awesome markdown icon instead?
https://fontawesome.com/icons?d=gallery&q=markdown
The link to the markdown tutorial can go in there somewhere?

xprilion:
@bmcutler thanks for taking a look here! :D I'll proceed as suggested since I now think of it and it does seem much more intuitive if the viewer isn't toggling around things which the editor has posted :D

xprilion:
@bmcutler I've updated the button for the markdown true! 

@andrewaikens87 @MasterOdin can you please take a look at the migration file I created and let me know if something looks wrong with it! :o

bmcutler:
Pictures look good :)

Migration & .sql edits both look good (I didn't test it).
I don't know if you need the default on the column though.  I assume every time you insert you will provide all values and won't need a default.

xprilion:
I am trying to not have to pass a false a lot of times, since that can be defaulted, and hence a lesser hassle! But that's just my own way of code, open to suggestions!

xprilion:
I'm looking to write tests specific to this feature as soon as possible. Please let me know about the priority for that!

migration/migrator/data/course_tables.sql
```diff
@@ -583,6 +583,7 @@ CREATE TABLE "posts" (
 	"endorsed_by" varchar,
 	"type" int NOT NULL,
   "has_attachment" BOOLEAN NOT NULL,
+  "render_markdown" BOOLEAN NOT NULL DEFAULT 'false',
```
MasterOdin: can you fixing the spacing here (and the line above) to make things to match the surrounding lines?

site/app/controllers/forum/ForumController.php
```diff
@@ -339,6 +339,12 @@ public function publishThread(){
             return $this->core->getOutput()->getOutput();
         }
 
+        $markdown = false;
+
+        if(isset($_POST['markdown_status']) && ($_POST['markdown_status']=="1")){
```
MasterOdin: Assuming `$_POST['markdown_status']` can either be undefined, `"0"`, or `"1"`, we can replace these checks with just `$markdown = !empty($_POST['markdown_status']`.

site/app/controllers/forum/ForumController.php
```diff
@@ -434,6 +442,14 @@ public function publishPost(){
             $this->changeThreadStatus($_POST['thread_status'], $thread_id);
         }
 
+        $markdown = false;
+
+        if(isset($_POST['markdown_status']) && ($_POST['markdown_status']=="1")){
+            $markdown = true;
+        }
```
MasterOdin: see above

site/app/controllers/forum/ForumController.php
```diff
@@ -520,14 +536,21 @@ public function alterPost($modifyType){
         $post_id = $_POST["post_id"] ?? $_POST["edit_post_id"];
         $post = $this->core->getQueries()->getPost($post_id);
         $current_user_id = $this->core->getUser()->getId();
+
+        $markdown = false;
+
+        if(isset($_POST['markdown_status']) && ($_POST['markdown_status']=="1")){
+            $markdown = true;
+        }
```
MasterOdin: see above

site/app/controllers/forum/ForumController.php
```diff
@@ -722,7 +746,14 @@ private function editPost(){
             if(!$this->modifyAnonymous($original_creator)) {
                 $anon = $original_post["anonymous"] ? 1 : 0;
             }
-            return $this->core->getQueries()->editPost($original_creator, $current_user, $post_id, $new_post_content, $anon);
+
+            $markdown = 0;
+
+            if(isset($_POST['markdown_status']) && ($_POST['markdown_status']=="1")){
+                $markdown = 1;
+            }
```
MasterOdin: see above

xprilion:
@andrewaikens87 I guess what you're doing is that you're switching ON/OFF the markdown from the thread reply, and simply changing the thread from the list, without making a reply post. This should probably not cause a change in the markdown cookie, just like making edits on a post does not change the cookie status

migration/migrator/migrations/course/20190720165721_render_markdown.py
```diff
@@ -0,0 +1,11 @@
+"""Migration for a given Submitty course database."""
+
+
+def up(config, database, semester, course):
+    if not database.table_has_column('posts', 'render_markdown'):
+        database.execute('ALTER TABLE posts ADD COLUMN render_markdown BOOLEAN NOT NULL DEFAULT false')
```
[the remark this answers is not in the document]
xprilion: sounds good :D updating!

PR #4150, opened 2019-07-19
[Feature:Installation] latest github release/tag & commit in footer

bmcutler:
At installation, we store the latest git tag & short commit in config/version.json

Then this information can be displayed in the footer.  The commit hash is displayed when the user hovers over the tag.

We assume all tags are valid releases on the master branch of Submitty (this may not be true and then this will result in a broken link).

Note also that the commit may not match the tagged commit (there may be additional commits).  And further, it may not be a publicly available commit. 

tag

### DIFF
--- a/site/app/libraries/Core.php
+++ b/site/app/libraries/Core.php
@@ -108,6 +108,30 @@ class Core {
     }
 
     /**
+     * Load the Submitty software version configuration.
+     * @throws \Exception
+     */
+    public function loadSubmittyVersionConfig() {
+        $this->latest_tag = "xx";
+        $this->latest_commit = "yy";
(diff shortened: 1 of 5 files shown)


site/app/libraries/Core.php
```diff
@@ -107,6 +107,28 @@ public function disableRedirects() {
...
+            throw new Exception("Error parsing version.json file");
+          }
+          $this->latest_tag = $contents['most_recent_git_tag'];
+          $this->latest_commit = $contents['short_installed_commit'];
+        }
+        else{
+          $message = "Unable to access version configuration file " . $version_path;
+          $this->addErrorMessage($message);
+        }
+    }
```
MasterOdin: Was there a reason to not load this file as part of the rest of the Config model?

Issue #2803, opened 2018-09-03
grades_released_homework is missing all common marks


bmcutler:
Not sure when this started...  but this gradeable (and perhaps others) are missing common marks, including a no credit/full credit mark.

This makes editing the rubric from either instructor or ta interfaces (at least with PR #2751, that I'm about to merge) impossible.

Fenn-CS:
@bmcutler I am able to edit this particular homework in it's current state.

MasterOdin:
To give some context here, if you go to [Future (No TAs) Homework](http://192.168.56.111/index.php?semester=f19&course=sample&component=admin&page=admin_gradeable&action=edit_gradeable_page&id=future_no_tas_homework&nav_tab=2) and expand "Readme", you'll see three marks (outlined for emphasis):
![Screenshot 2019-07-09 22 35 18](https://user-images.githubusercontent.com/1845314/60936067-e3612000-a299-11e9-8f91-294a06e5056e.png)

On grades_released_homework, you get no marks starting out (to get things into a proper state, you have to hit "Add New Mark" twice, and then toggle the count up/down setting):
![Screenshot 2019-07-09 22 36 24](https://user-images.githubusercontent.com/1845314/60936118-0390df00-a29a-11e9-89b8-72afe94070a6.png)

which in turn causes some number of UI and other bugs. This is due to the fact that the yml file we're loading does not have any defined marks for that gradeable:
https://github.com/Submitty/Submitty/blob/1d0226d83163f270e6ef02ab93048af7c4b2e106/.setup/data/courses/sample.yml#L641-L659

Compare that to Future (No TAs) homework for example:
https://github.com/Submitty/Submitty/blob/1d0226d83163f270e6ef02ab93048af7c4b2e106/.setup/data/courses/sample.yml#L10-L73

The simple fix is to just define a `marks` object under the grades_released_homework's components, but it should be decided if we also want to make `setup_sample_courses.py` do something as well when it's loading these yml files, such as if it detects a electronic gradeable, and that it has no marks for a component, crash with a meaningful error message, or maybe just output a warning and silently fill in some sort of default mark, or etc.

Fenn-CS:
Thanks for the explanation Matt, going by the last method is better as we discussed on Hangouts, it would be more robust if those non-existent objects are created in `setup_sample_courses.py` with some logging but not termination.

PR #4154, opened 2019-07-20
[Testing:SampleData] added random marks generation for gradeables

Fenn-CS:

When a gradeable lacks the `mark` component defined in `.setup/data/courses/sample.yml` no default values are given hence parts of the interface required to edit sample gradeables disappear.
Fixes: #2803 

### What is the new behavior?
Random common marks are generated and assigned for gradeables without the mark component.

#### Breakdown
`generate_random_marks()` in `.setup/bin/setup_sample_courses.py`
https://github.com/Submitty/Submitty/blob/eeb2aae7fc03e23cd5a81f180e434382abc71f09/.setup/bin/setup_sample_courses.py#L245-L252
returns a list of mark objects(dict) which can be passed to the `Mark` class
mark-pairs

#### Result
Screenshot from 2019-07-20 05-38-14

### Other information?
The following is required for this setup to be complete.

1. How many common marks must a gradeable have or how many should we assign as random defaults to gradeables wher the component is not defined?

_Currently all the mark objects in `.setup/data/random/marks.txt` are attached to each gradeable, I think ideally we should have like 2 or 3?

2. We should make sure random `gradeable_component_mark`, `gcm` that was assigned to a `gradeable_component`, `gc` is not assigned to another `gradeable_component` within thesame gradeable. (Does this even matter? @bmcutler? )


bmcutler:
The instructor must prepare the components, and the total points for each component and whether its up or down grading (which determines whether we start with a full credit or no credit mark).
The instructor can optionally prepare the additional common marks within the component.
But alternatively, the TA can prepare the common marks when they start grading.   (but we always have the full credit or no credit mark)
The instructor or the TA can edit the mark text and points after grading has begun or finished.

For the sample data, it would be nice to have a homework where ta grading has not begun where some or all components have no common marks.
All gradeables that have some ta grading (or completed ta grading) should have at least 2-3 common marks per component.

bmcutler:
@Fenn-CS I'm not sure about point #2. 
I'm going to ask @KevinMackenzie to follow up with you.

MasterOdin:
While I have no idea if @KevinMackenzie has followed up here or not, I will say that I believe that we cannot mix both positive and negative marks within a component. The grading is either strictly count up or down down. Additionally, the 0 point mark ("Full Credit" or "No Credit" depending on counting down or up) must exist and that any other marks are then optional (but as mentioned above should be consistent and in the same direction).

My suggestion here would be to randomly decide if a component is count down or count up. Then, randomly add some marks that are within that particular category. It may be easier to just store the marks in a YAML file that is structured like:
```
count_up:
  - name: Has README
     points: 2
  ...
count_down:
  - name: Lacks README
    points: 2
  ...
```
(replacing the `...` with additional list items).

For @bmcutler's comment:
> For the sample data, it would be nice to have a homework where ta grading has not begun where some or all components have no common marks.

We probably want to just explicitly pick a gradeable in sample to have this property and not rely on random to do this. It'll be vastly simpler to do, and don't have to play around around with the random generator at all to ensure the desired result.

PR #4162, opened 2019-07-21
[Refactor:Router] New URL for ForumController

zjxiaohan:
### URL Changes

- POST `http://192.168.56.111/f19/sample/forum/search`
- POST `http://192.168.56.111/f19/sample/forum/announcements`
- GET `http://192.168.56.111/f19/sample/forum/stats`
- GET `http://192.168.56.111/f19/sample/forum/categories`
- POST `http://192.168.56.111/f19/sample/forum/categories/new`
- POST `http://192.168.56.111/f19/sample/forum/categories/delete`
- POST `http://192.168.56.111/f19/sample/forum/categories/edit`
- POST `http://192.168.56.111/f19/sample/forum/categories/reorder`
- GET or POST `http://192.168.56.111/f19/sample/forum/threads`
- GET or POST `http://192.168.56.111/f19/sample/forum/threads/{thread_id}`
- GET or POST `http://192.168.56.111/f19/sample/forum/threads/new`
- POST `http://192.168.56.111/f19/sample/forum/threads/status`
- POST `http://192.168.56.111/f19/sample/forum/threads/pin`
- POST `http://192.168.56.111/f19/sample/forum/threads/merge`
- POST `http://192.168.56.111/f19/sample/forum/posts/new`
- POST `http://192.168.56.111/f19/sample/forum/posts/history`
- POST `http://192.168.56.111/f19/sample/forum/posts/get`

### Testing
Manually tested the following:
- searching
- make announcements
- see stats
- see categories, adding, deleting, reordering.
- resolve & unresolve threads
- pin & unpin threads
- merge threads
- make & edit posts
- notification

### Other Info
I feel `POST` is overused in ForumController... Due to the extensive nature of this controller and Anu working on this, I didn't modify much inside the function. I don't think many of them should be `POST` though.


site/app/controllers/NotificationController.php
```diff
@@ -76,8 +76,7 @@ public function openNotification($nid, $seen) {
         $user_id = $this->core->getUser()->getId();
         $metadata = $this->core->getQueries()->getNotificationInfoById($user_id, $nid)['metadata'];
         if (!$seen) {
-            $thread_id = Notification::getThreadIdIfExists($metadata);
-            $this->core->getQueries()->markNotificationAsSeen($user_id, $nid, $thread_id);
```
zjxiaohan: I don't understand why `$thread_id` is necessary, so I deleted it.

zjxiaohan:
@sak6lab Could you have a look at the notification part?
I am trying to add a `url` field in the `$metadata` object which is part of notification, and let the url be obtained directly instead of having to build that from `$metadata`. I am not sure about the side effect of this change. Will this affect emails?

sak6lab:
I think you should do that in another PR. Although it will not affect emails as metadata is only for links you will need to change how the notification gets the link when a user clicks on it. https://github.com/Submitty/Submitty/blob/6ce10f2b7f94b2d0058d088e1e9bc8237a2907c5/site/app/models/Notification.php#L114. So it is better to do this in another PR and change all metadata to just be a link. I think storing an url rather than the parts of an url is much cleaner and should be done.

MasterOdin:
> I feel POST is overused in ForumController... Due to the extensive nature of this controller and Anu working on this, I didn't modify much inside the function. I don't think many of them should be POST though.

I agree with this, and would also say that we should be encoding some of the `$_POST` information for some of these calls (e.g. the `thread_id`) into query itself to be better compatible with what the API should look like.

zjxiaohan:
Mark this as WIP to have a notification PR first.

site/app/controllers/NotificationController.php
zjxiaohan: it is necessary

zjxiaohan:
Merge conflicts resolved. Will remove `WIP` after another manual testing.

zjxiaohan:
Tested basic functions. Ready for review.
Also @xprilion could consider modifying some POST methods to GET for (1) semantic reasons, and that (2) since the new router forces POST to use CSRF token, sometimes there is a delay for getting that, potentially resulting in errors.

PR #4167, opened 2019-07-22
[Refactor:API] Endpoints for GET users & graders

zjxiaohan:

### What is the new behavior?
Doc PR: https://github.com/Submitty/api.submitty.org/pull/1
Screenshots:
Screenshot from 2019-07-22 16-56-08
Screenshot from 2019-07-22 16-56-18

### Other information?
Refactored the code so that the two methods concerned return a `Response` object to enable us to reuse the code.

Changes are mainly:
- Removed `renderDownloadForm` function and `include` the form in twig files instead. We cannot render two views in one function if we are going to use `Response`.
- Use `User::SomeGroup` instead of hard coded numbers.
- Moved some JSON logic from view function to controller function so API and Web can use them together.


site/app/controllers/admin/UsersController.php
```diff
@@ -28,36 +31,106 @@ public function run() {
...
+                    break;
+                case User::GROUP_LIMITED_ACCESS_GRADER:
+                    $grp = 'Limited Access Grader (Mentor)';
+                    break;
+                default:
+                    $grp = 'Student';
+                    break;
+            }
+            $first_name = str_replace("'", "&#039;", $student->getDisplayedFirstName());
+            $last_name = str_replace("'", "&#039;", $student->getDisplayedLastName());
```
MasterOdin: Why are we encoding this here?

site/app/controllers/admin/UsersController.php
```diff
@@ -28,36 +31,106 @@ public function run() {
...
+            array_push($download_info, ['first_name' => $first_name, 'last_name' => $last_name, 'user_id' => $student->getId(), 'email' => $student->getEmail(), 'reg_section' => "$reg_sec", 'rot_section' => "$rot_sec", 'group' => "$grp"]);
```
MasterOdin: Get rid of the quotation marks around the last couple elements. We shouldn't be morphing the type of any of these until right before displaying (if at all) and some of these are strings anyway and don't need to be wrapped.

site/app/controllers/admin/UsersController.php
zjxiaohan: Not sure why. Copied these lines from the old code. Tried to remove them and added a user with apostrophe in the name and the csv looked okay.

PR #4170, opened 2019-07-22
[Feature:TAGrading] highlight pos/neg marks w/ count up/down 

### LINKED ISSUE #3413: Configuration of count up/count down for ta grading
This is mostly a design issue.

It should be more intuitive that a component is count up or count down.  When count up most (but not all) marks should be positive.  And vice versa.  From both the instructor rubric creation page and from the ta grading page

Currently only the instructor can switch a component from up to down (it should probably stay that way), but we should make it visually obvious which components are up vs down.  (So an instructor can catch an error in rubric prep more quickly)

thef1stofhername:

Closes #3413 

### What is the new behavior?
Now, when you change the count up/count down grading guideline, any component score that doesn't comply is highlighted. It becomes unhighlighted when you change it to comply.
### Other information?
Testing:
1. Sign in as instructor
2. Go to Gradeables
3. Click on the edit pencil
4. Click on rubric and select a component
5. Play around with possible scores and count up/down configurations
If the score is negative on count up, it should be highlighted and vice versa. Changing the score to comply should remove the highlight.
Instructor highlight 2019-07-23 15-03
6. Sign in as TA and edit rhe grading rubric when you grade a student
TAGrading highlight2019-07-23 11-46


bmcutler:
Please add a couple screenshots to illustrate the changes.

Also we should add some screenshots/documentation on submitty.org :)
[note: this could be a big undertaking, instructor config of ta grading is under-documented & stale]

Thanks!

thef1stofhername:
What component because I tested it in both?

Kevin-MM:
I tried in Future (No TAs) Homework in the Edit Gradeable page and the TA grading interface.  No luck.

thef1stofhername:
ok i will go back and see what is up

PR #4172, opened 2019-07-22
[Feature:GradeInquiry] Migration for grade inquiry per component

sak6lab:
This PR includes the migration to support the grade inquiry per component feature. 
I added:
- A foreign key column in the regrade_discussion table called gc_id that can hold the gradable_component that the grade inquiry post is for
  - This value can be null
- A column in the electronic gradeable table of type boolean that hold the value of whether or not grade inquiry per component is enabled
- A constraint to check that grade inquiry per component is on when grade inquiries are not allowed

Copy of null


migration/migrator/migrations/course/20190722110112_grade_inquiry_per_component.py
```diff
@@ -0,0 +1,21 @@
...
+
+    # add eg_is_gi_per_component column to electronic_gradeable table
+    database.execute("ALTER TABLE electronic_gradeable ADD COLUMN eg_is_gi_per_component boolean DEFAULT false NOT NULL")
+    # add a check constraint to make sure that eg_regrade_allowed is true
+    database.execute("ALTER TABLE electronic_gradeable ADD CONSTRAINT regrade_allowed_true_check CHECK (eg_regrade_allowed is true or eg_is_gi_per_component is false)")
+
+def down(config, database, semester, course):
+    # remove gc_id column from regrade_discusion table
+    database.execute("ALTER TABLE regrade_discussion DROP CONSTRAINT gradeable_component_id_fk")
+    database.execute("ALTER TABLE regrade_discussion DROP COLUMN gc_id")
```
bmcutler: We don't need to drop the data.  
If we roll back on production we don't want to lose information unless absolutely necessary.

You'll then need to modify the up migration to add column *if not exists* (see other migrations for samples)

MasterOdin:
For my understanding, can there be more than one regrade_discussion going at one time for more than one gradeable component?

bmcutler:
Yes, we can have 1 discussion thread for each component.

The plan is on a typical CS1 homework per component grading will be false, a single thread for all questions between student and TA A (we assume a single TA grades entire hw).  gc_id will be NULL in this case.

For a typical CS1 exam, per component grading will be true, and student can start a thread about question 2 with TA A, and a thread about question 5 with TA B.  And perhaps we will still also allow a general thread for discussion (gc_id=NULL) about the test as a whole with everyone who graded any question on the exam (But maybe not.  UI is a draft until we get feedback.)

PR #4173, opened 2019-07-22
[Feature:MobileUI] Mobile Course Settings

### LINKED ISSUE #4087: "Display Room Seating Guide" has an empty button
**Describe the bug**
Screenshot from 2019-07-14 13-19-11

There is an empty thing in the left-bottom corner of the screenshot.

**Expected behavior**
Not sure.

**To Reproduce**
1. Go to `Course Settings` after a default setup.
2. See error

**Configuration**
 - OS: Linux
 - Browser: Chrome

**Additional context**
I am not sure about how this feature works. It looks like there is only documentation for exam seating? Is there anything I can read to understand how the gradeable seating works? @bmcutler

mktung:
Closes #4087 

* Text input fields span the entire width
* Overall larger text areas
* Buttons / dropdowns should fill width when screen shrinks
* CSS and JS moved to separate files
* Removed redudant IDs and HTML elements

Should look something like this:
image


bmcutler:
The email seating assignments button should only be visible if a valid gradeable has been selected for room seating.

PR #4179, opened 2019-07-23
[Refactor:Notification] Let notification metadata store URLs

zjxiaohan:

We currently store URL parts in the notification metadata, which is not compatible with the new URL because we use the old way of assembling them.

### What is the new behavior?
Notification metadata store URL directly.

### Other information?
Tested manually by making announcements & posts.

It is a breaking change - all old notifications will be unavailable. I am not sure if we need a migration, or, gracefully handling this (the code may be a bit dirty as there are database queries).

@bmcutler once said (long time ago) it's okay just let students complain and they will forget. What do you think now?

### DIFF
--- a/site/app/models/Notification.php
+++ b/site/app/models/Notification.php
@@ -107,18 +107,15 @@ class Notification extends AbstractModel {
     /**
      * Returns the corresponding url based on metadata
      *
-     * @param  Core     core
-     * @param  string   metadata
-     * @return string   url
+     * @param  string   $metadata_json
+     * @return string   $url
      */
-    public static function getUrl($core, $metadata_json) {
+    public static function getUrl($metadata_json) {
         $metadata = json_decode($metadata_json);
         if(is_null($metadata)) {
             return null;
         }
-        $parts = $metadata[0];
-        $hash = $metadata[1] ?? null;
-        return $core->buildUrl($parts, $hash);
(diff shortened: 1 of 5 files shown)


bmcutler:
All old notifications will be unavailable?  Or the links in old notifications will be broken?

Broken links is ok.

zjxiaohan:
> All old notifications will be unavailable? Or the links in old notifications will be broken?

Links will be broken.

site/app/models/Notification.php
```diff
@@ -107,26 +107,23 @@ public static function createViewOnlyNotification($core, $details) {
...
 
     public static function getThreadIdIfExists($metadata_json) {
         $metadata = json_decode($metadata_json, true);
         if(is_null($metadata)) {
             return null;
         }
-        $thread_id = array_key_exists('thread_id', $metadata[0]) ? $metadata[0]['thread_id'] : -1;
+        $thread_id = property_exists($metadata, 'thread_id') ? $metadata->thread_id : -1;
         return $thread_id;
     }
```
MasterOdin: Use a consistent group of parameters for `json_decode` for these (and other) functions. We mostly use true elsewhere in the codebase.

You can also simplify the check here with the null coalescing operator:
```
$thread_id = $metadata->thread_id ?? -1;
```

PR #4183, opened 2019-07-23
[Forum: Bugix] Forum read threads marked white

### LINKED ISSUE #4158: Forum threads do not get marked as 'seen' on clicking them
**Describe the bug**
When I click on a thread the post remains blue. When I click to another unread post neither of them get changed to white. If i click on 5 unread threads it looks like I haven't clicked on any which makes it hard to keep track of what i've seen and what i haven't

**Expected behavior**
When I click on a thread it should be marked as read and show as white on the left side. On refresh, things fix themselves so this should be able to be done client side. (onlick = this.color = white) or something to that logic.

I think a post you create should also be marked as read on creation but that might be a separate issue.

**To Reproduce**
1. Click on an unread thread
2. Click on another thread

**Configuration**
 - linux firefox

xprilion:

Fixes #4158

### What is the new behavior?
On clicking a thread its background in the thread list gets changed to white. This helps understand which thread has been read.

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -1085,6 +1085,8 @@ function loadThreadHandler(){
                     return;
                 }
 
+                $(obj).find(".thread_box").css("background", "white");
+
                 $('.thread_box').removeClass('active');
 
                 $(obj).children("div.thread_box").addClass('active');
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -1085,7 +1085,7 @@ function loadThreadHandler(){
                     return;
                 }
 
-                $(obj).find(".thread_box").css("background", "white");
+                $(obj).find('.thread_box').removeClass('new_thread');
 
                 $('.thread_box').removeClass('active');
 


xprilion:
Surely I can! please let me know the expected behaviour! :sweat_smile:

shailpatels:
> Surely I can! please let me know the expected behaviour! sweat_smile

Disregard that comment, (reading it back now more awake it doesn't make sense) I think read/unread posts is working fine.

It would be nice if when I create a thread it will immediately marked read but thats not a major issue when this gets merged.

site/public/js/forum.js
```diff
@@ -1085,6 +1085,8 @@ function loadThreadHandler(){
                     return;
                 }
 
+                $(obj).find(".thread_box").css("background", "white");
```
MasterOdin: Could you do `$(obj).find('.thread_box').removeClass('new_thread');` here instead?

MasterOdin: Could you do `$(obj).find('.thread_box').removeClass('new_thread');` here instead?

xprilion: I get it, did not notice that class there!

Issue #3911, opened 2019-06-24
Add functionality to 'delete user from course' & 'delete user overall' 


bmcutler:
These operations will permanently delete data (not just a bit to mark as deleted).  They should be used very carefully.  Primarily will be used to cleanup automated testing or to immediately fix a data entry typos or erroneously added data.

'delete user from course'
This operation should fail if the student has any submissions or any forum posts or any grades (or anything with a foreign key that we don't want to cascade delete).  We'll add other functionality to delete those bits first if needed.

'delete user overall'
once a user has been deleted from all courses, the user can be scrubbed from the master table as well.  This operation should fail if the user is in any courses.  There is a foreign key to the emails table in the master database, but this is one place where a cascading delete is ok.  Just delete all those rows from the emails table at this time.

related to #3790

PR #4205, opened 2019-07-26
Issue 3911 Delete User from Master DB (database code)

pbailie:
Related to #3911 (possibly closes)

Changed: `submitty_db.sql`
New master migration: `20190722175305_delete_user.py`

- Database code for deleting a user from master DB.  Exception handling implemented to ensure tidy error handling over foreign key constraint violations (especially should the user have referenced data in any course DB).
- User is also automatically scrubbed from master DB `users` table if they are no longer tied to any course in `courses_users`.  
- Email foreign key for `user_id` now permits cascade DELETE.


bmcutler:
Peter and I talked today and he's going to work this into the web UI.

From a specific course, both the manage students and manage grader pages will have a new column all the way on the right for delete user.  A little trashcan icon.

When we click this icon it says "are you sure you want to delete user XXX from your course?" 
If you say yes it will attempt to remove the user, but if the user has any associated data it will fail with a foreign key error and we will stop and print (hopefully helpful) information to the user.  (nothing changes).

It would be nice bonus to have these trashcans only appear on users that can be safely deleted.  Will mean adding more information to the users model/php class.

(Hopefully I've captured this conversation correctly)

pbailie:
Please do not delete branch.

PR #4216, opened 2019-07-29
[Bugfix:InstructorUI] Moved checkboxes to the left on Course Settings

### LINKED ISSUE #4204: Put checkboxes on left of option text on Course Settings Page (& elsewhere)
Will be easier to read/use on wide browser screen
Screen Shot 2019-07-25 at 5 56 29 PM

mktung:
Closes #4204 

Checkboxes, on desktop mode, have been moved to the left side of the screen in course settings, so that it's easier to see what aligns with what on large screens.

Previously:
Screenshot from 2019-07-29 11-04-22

Now:
Screenshot from 2019-07-29 11-00-27


bmcutler:
Why does the checkbox jump to the right side when we get to "mobile" sized window?  Can we just leave it on the left all of the time?

mktung:
My latest commit should change that

PR #4218, opened 2019-07-29
[Bugfix:BulkUploads] Bulk upload focus

### LINKED ISSUE #4185: bulk upload sometimes loses focus on student id box
**Describe the bug**
When matching bulk uploads to usernames, after selecting a user id from the dropdown and pressing enter sometimes the cursor focus correctly goes to the id box of the next pdf to label. 
But sometimes the cursor focus goes elsewhere, sometimes on the pdf of the next pdf to label, sometimes its much farther away (it's inconsistent).

It may take labeling 5-10 or more pdfs to break, but once it breaks, it seems to stay broken (it doesn't go to the id box of the next pdf).

shailpatels:
Fixes bulk uploads losing focus after a bit.
The error messages popping up were caused by duplicate requests being sent and when it would fail
it would try and re-enable the submission box that it thought failed. But because the submission had
actually succeeded, it would error and lose focus. 

Clears messages on page load 

Fixes #4185

### DIFF
--- a/site/app/templates/submission/homework/BulkUploadBox.twig
+++ b/site/app/templates/submission/homework/BulkUploadBox.twig
@@ -98,6 +98,7 @@
     var is_team_assignment = "{{ team_assignment }}" === "1";
     //autofill data
     $(function(){
+        $("#messages").empty();
         if(sessionStorage.getItem("expected_count"))
             document.getElementById("expected_pages_input").value = sessionStorage.getItem("expected_count");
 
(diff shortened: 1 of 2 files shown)


bmcutler:
Looks good, for normal labeling.  (tested a while didn't see the bug)

If I upload a second submission for a student, and I get the popup/modal asking what I want to replace, once I hit return on submit my focus is 'lost' again.  Can we have it go to the next box?

shailpatels:
@bmcutler is this a team assignment I'm not seeing anything on normal bulk uploads for no team

Issue #3904, opened 2019-06-22
Remove emma from Submitty


MasterOdin:
Currently, we install both emma and JaCoCo on Submitty. However, given that emma does not have support for the newer Java versions and features (as it was released when Java 6 came out), it's not really recommended that people should use this tool over JaCoCo.

My recommendation is to remove emma (installation and the grader methods), and delete Submitty/emma and just focus on JaCoCo support.

@bmcutler, thoughts?

Process for removal:
- [x] Remove from installation scripts (#4219)
- [x] Deprecate using emma in grading code
- [x] Remove emma from grading code
- [x] Delete emma repository mirror under Submitty/emma

bmcutler:
closed by PR #6283

PR #4219, opened 2019-07-29
[Refactor:Installation] Remove emma from installation process

MasterOdin:

Emma (a very out of date Java coverage tool) is installed alongside JaCoCo for dealing with Java code coverage. Emma was last updated over 10 years ago and that it does not work on modern versions of Java (Java 6 or 7 was the last truly supported version).

Part of work for #3904.

### What is the new behavior?
Emma is no longer installed in Submitty. The grading code is left as-is to retain backwards compatibility, but that we should discuss appropriate "deprecation" strategy for that functionality and rip it out in a future release.
 
### Other information?

This is not a breaking change, does not affect anyone currently using emma (if any such people exist).

Tested by doing `vagrant up`.


.setup/install_system.sh
```diff
@@ -619,30 +605,29 @@ if [ ${WORKER} == 1 ]; then
...
-    if [ -z "${SUBMISSION_URL}" ]; then
-        SUBMISSION_URL='http://192.168.56.101'
-    fi
-    echo -e "/var/run/postgresql
-    ${DB_USER}
-    ${DATABASE_PASSWORD}
-    America/New_York
-    ${SUBMISSION_URL}
-
+        if [ ${VAGRANT} == 1 ]; then
```
bmcutler: Why did the indentation of this line change?
(Or is this a tabs/spaces thing that github is making look worse?)

Overall the other indentation changes are probably an improvement (but are unrelated to emma & should not affect the functionality of the script, right?)

MasterOdin: Yeah, the indentation shouldn't have any effect on things. The indentation of the contained lines was at the same level as `if [ ${VAGRANT}...` which was a bit confusing and so I indented it, though included this line by accident.

Fixed it so that the indentation is as it should be, while also removing the indentation on all of the echoed lines to better indicate their final values (CONFIGURE_SUBMITTY.py still applies a `.strip()` to any input though).

PR #4225, opened 2019-07-30
[Refactor] Load user from both course db and master db

zjxiaohan:

There are duplicate data in course db and master db and some master db info doesn't exist in course db (access_level).

### What is the new behavior?
Merge the info from course db and master db when loading a user.

### Other information?
May need to remove triggers to sync master db & course db in the future?
Not sure where they are written. Sounds not that easy to remove them safely.


bmcutler:
I don't think we want to remove any existing data (or triggers) from the course database, even if it is duplicated.

The course database should stand on its own.  The idea is that if an instructor wants the data from their course we can simply give them a dump of their entire database (and their courses directory & files) and they could even spin up a Submitty instance based on that data.  (Would have to write a tool to create a new master database, but that should be straightforward.)

The access level bit about the user only needs to live at the master database level.  It doesn't need to be in the course.  Possibly in the future we'll want to add any additional data at the master level only.

site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -25,7 +25,16 @@
...
 
     public function getUserById($user_id) {
-        $this->course_db->query("
+        $details = array();
+
+        $this->submitty_db->query("SELECT * FROM users WHERE user_id=?", array($user_id));
+
+        if (count($this->submitty_db->rows()) > 0) {
+            $details = array_merge($details, $this->submitty_db->row());
+        }
```
MasterOdin: Let's change this to be:
```
if (this->submitty_db->getRowCount() === 0) {
    return null;
}
$details = $this->submitty_db->row();
```

No need to waste time checking the course DB if they don't exist in the master DB

PR #4229, opened 2019-07-30
[Feature:UITweak] Polished mobile gradeables 

mktung:
The gradeables page should now stay on desktop mode for smaller resolutions than before.

Previously:
Mobile: <540px
Tablet: 540-1349px
Desktop: >1350px

Now:
Mobile: <540px
Tablet: 540-949px
Desktop: >949px

Peek 2019-07-30 11-25


site/public/js/server.js
```diff
@@ -1849,7 +1849,7 @@ $.fn.isInViewport = function() {                                        // jQuer
 
 function checkSidebarCollapse() {
     var size = $(document.body).width();
-    if (size < 1000) {
+    if (size < 1150) {
```
bmcutler: Why did this number go up?   Side bar should stick around longer. 
(Yes, I'm asking for the impossible, but many of us have not memorized the symbols)

mktung: At around 1150px, the instructor gradeables page gets too wide to fit on the screen. We have 3 options: allow the table to overflow to the side of the screen (and generate a horizontal scrollabar), switch to tablet mode, or collapse the sidebar. If we collapse the sidebar, we can push the tablet cutoff point from 1150px to 950px, which allows us to maintain desktop mode in half of a 1080p monitor (960px width). I can switch it to something else though.

PR #4231, opened 2019-07-30
[Feature:GradeInquiry] Grade inquiry per component

### LINKED ISSUE #3110: Regrade Request per problem (or per TA)
When the grading of an assignment has been done by multiple people, when a student submits a regrade for specific problem(s) we would like it to notify only the people we graded those problems.  And have the the regrade resolved per problem.

* Maybe list the components like discussion forum categories that the student can check when they start the conversation thread for the regrade.

* The TAs can uncheck the 'categories' (components) as the regrades are resolved.

* The query for deciding if a specific TA should be shown pending regrades can then more correctly direct the TAs to regrade.

* Maybe the stoplight buttons for the grading from the index page can turn red for the problems that have an outstanding regrade request.

[needs design discussion]

sak6lab:
When a grade inquiry is made all graders who may have graded the student's assignment will get notified.

### What is the new behavior?
Now instructors can choose to have their grade inquiries sorted by which component or question the grade inquiry pertains to and visible graders (graders that are allowed to be displayed) who graded the selected component will get notified.

Database Diagram 
database-schema

Student view (similar to instructor)
gipc-student

When switching from grade inquiry per component allowed and not allowed, all of the posts can still be seen however any grade inquiries' status for a component for example, read me, will be not shown. The only grade inquiry status that will be shown will be the grade inquiry that is not assigned to a component.

When switching from grade inquiry per component not allowed and allowed, the grade inquiry that was not assigned to a component will be shown in the all tab along with all of the other posts. Only in this state where a grade inquiry was made before the switch will users be able to interact with this grade inquiry in the all. Once the grade is resolved no further action can be taken it's status will not be shown no more in the all tab.

### Testing
To turn on grade inquiries per component, in the edit gradeables of any gradeable, turn on grade inquries and then turn on grade inquiry per component.
grade_inquiry_per_component_allowed

closes #3110 

Documentation -> Submitty/submitty.github.io#151


sak6lab:
Fixing an issue where the tab selected on reload should be more intuitive. Currently it loads the cookie which expires after 60 seconds or the first component.

It should rather select the component saved in the cookie (which lasts forever which is submission gradeable specific), the first unresolved component, or the first component

sak6lab:
Fixed an issue where if grade inquiry per component was on and grade inquiries were made per component and then the grade inquiry per component is turned off. The status was displayed as resolved when no grade inquiry with out a component was made.

sak6lab:
Hey @andrewaikens87, just curious, what are tweaks you are planning to make?

bmcutler:
@sak6lab  I'll message on slack

PR #4242, opened 2019-07-31
[Refactor:Router] Router cleanup

zjxiaohan:

### What is the new behavior?

1. Removed ClassicRouter and its related code.
2. Removed `buildUrl`. Renamed `buildNewUrl` and `buildNewCourseUrl` to `buildUrl` and `buildCourseUrl` both in PHP and JS. Cleaned all usage of `buildUrl` and fixed several places when looking through them.
3. Removed `run` requirement by the abstract class.
4. Add `@AccessControl` check for API. I don't know why I forgot that.
5. Load course information in the router, after matching URL. That is because, the naive way cannot distinguish semester and course. For example, if we use a naive way to get semester and course, we'll see `/authentication/login` as a course called `login` in `authentication` semester. Previously we hard code those special cases, but we'd better not do that. So first matching semester and course by the router and then loading the info is a better choice.
6. To make `5` happen, we can't load all config so early. So I split `loadConfig` function to two functions, and one of them is for loading course config, which happens later.
7. `5` forces us to check login status in the router, as, when checking, we load users. If the course info is not loaded, then the user info would be insufficient (see #4225 ).
8. Instead of putting login checking in the router, I put it in the `Core`. This is only for easier testing - testing one method in a static method that constructs an instance can be messy.

### Other information?
Please review this after Travis tests are passed. There are many changes (well.. I just didn't realize it's becoming that big) so we'd better be careful about it.


bmcutler:
Please resolve conflicts

zjxiaohan:
@bmcutler conflicts are resolved

site/app/libraries/Core.php
```diff
@@ -139,6 +139,13 @@ public function loadConfig($semester, $course) {
         }
     }
 
+    public function loadMasterConfig() {
+        $conf_path = FileUtils::joinPaths(__DIR__, '..', '..', '..', 'config');
+
+        $this->config = new Config($this, '', '');
```
MasterOdin: Remove the second and third parameters off the class constructor and make it just accept `Core $core`.

site/app/libraries/Core.php
```diff
@@ -118,7 +118,7 @@ public function disableRedirects() {
      * @param $course
      * @throws \Exception
      */
-    public function loadConfig($semester, $course) {
+    public function loadCourseConfig($semester, $course) {
         $conf_path = FileUtils::joinPaths(__DIR__, '..', '..', '..', 'config');
 
         $this->config = new Config($this, $semester, $course);
```
MasterOdin: modify this such that `if ($this->config === null) { throw Exception("Master config has not been loaded") }` at the top and then change `Config::loadCourseJson($course_json_path)` to be `Config::loadCourseJson($semester, $course, $course_json_path)` and set the class properties for those two values in that function.

site/app/libraries/Core.php
```diff
@@ -652,4 +641,79 @@ public function isTesting() {
...
+                    $request->headers->get("authorization"),
+                    $this->getConfig()->getBaseUrl(),
+                    $this->getConfig()->getSecretSession()
+                );
+                $api_key = $token->getClaim('api_key');
+                $logged_in = $this->loadApiUser($api_key);
+            }
+            catch (\InvalidArgumentException $exc) {
+                $this->getOutput()->renderJsonFail("Invalid token.");
+                $this->getOutput()->displayOutput();
```
MasterOdin: Can we just return false in case of exception and have the Router handle showing this fail condition?

MasterOdin:
I'm a bit iffy on points 5-7 (generally moving some of the log-in and config loading into the router), but breaking apart master/course loading makes sense I think.

zjxiaohan:
@MasterOdin 
> I'm a bit iffy on points 5-7

Could you elaborate this?

MasterOdin:
It's more that the Router is increasingly wrapped around Core (and interacts with Core) that it makes it potentially harder to test it (or that we have to spend more time creating expensive mocks), and that the router itself's responsibilities are increasingly expanded. Like I said, it's something I'm iffy on, but not enough to try and argue/figure out why might be a design I'm more comfortable with. It's also entirely possible we're going to refactor things further as more API endpoints are defined.

PR #4250, opened 2019-07-31
[Bugfix:Forum] JSON display thread posts javascript fail

### LINKED ISSUE #4127: [Forum] Post redirect issue
**Describe the bug**
When posts are published, we return a json response in order to send the next location of where the page should refresh. With the introduction of #4019, we have seen that after publishing, the json response is not *always* being consumed. This results in the user landing on a page with only a json object (which is confusing!). This functionality was introduced in GSOC last summer, and we should revisit this design and more importantly whether or not this functionality is required. I need to familiarize myself more with the plan of action for this issue, so I'm assigning myself and @xprilion to this for now.

xprilion:

Fixes #4127 

### What is the new behavior?
The JSON being displayed is when the JS fails to work, and the post goes to Submit because of the HTML submit binding getting into action. The resultant JSON is fair result, so the issue is with the failing of the JS. 

To avoid this, simply adding checks on the URL is enough to stay on the page, with an error message displayed.


site/public/js/forum.js
```diff
@@ -160,12 +160,26 @@ function publishFormWithAttachments(form, test_category, error_message) {
...
 
-function createThread() {
-    return publishFormWithAttachments($(this), true, "Something went wrong while creating thread. Please try again.");
+function createThread(e) {
+    e.preventDefault();
+    try {
+        return publishFormWithAttachments($(this), true, "Something went wrong while creating thread. Please try again.");
+    }
+    catch{
+        alert("There was an irrecoverable fault. Kindly contact administrator.");
```
shailpatels: Please change the error message to "Something went wrong. Please try again"

In this case if creating a thread/post fails an instructor wouldn't be able to do anything anyways and the student can always try again

site/public/js/forum.js
```diff
@@ -160,12 +160,26 @@ function publishFormWithAttachments(form, test_category, error_message) {
...
 
-function publishPost() {
-    return publishFormWithAttachments($(this), false, "Something went wrong while publishing post. Please try again.");
+function publishPost(e) {
+    e.preventDefault();
+    try {
+        return publishFormWithAttachments($(this), false, "Something went wrong while publishing post. Please try again.");
+    }
+    catch{
+        alert("There was an irrecoverable fault. Kindly contact administrator.");
```
shailpatels: change this error msg too

site/public/js/forum.js
xprilion: Done!

site/public/js/forum.js
xprilion: Done!

PR #4256, opened 2019-08-01
[Feature:InstructorUI] Nightly rainbow grades cron and misc

### LINKED ISSUE #3711: Nightly run of "Generate Grade Summaries" and Rainbow Grades

Kevin-MM:
Closes #3711 

This PR makes changes to the new rainbow grades UI and adds the ability to build rainbow grades nightly as part of a cron job.

Changes:
- The submitty-admin API token is now collected as part of the installation script and saved to ```submitty_admin.json```
- If the API token was successfully stored, a new variable in submitty_users.json indicates that the submitty_admin user was verified.
- The course settings page has a new option called "auto rainbow grades generation" which controls if rainbow grades for that course should be rebuilt nightly.  If an API token is not present then this setting cannot be used and appropriate messages are displayed instead.
- Created script ```auto_rainbow_scheduler.py``` which has been added to the ```submitty_crontab``` for a 2AM runtime.  This scripts automatically builds rainbow grades for any course that has "automatic rainbow grades generation" checkbox enabled in their course settings
- The rainbow grades UI can now detect if an API token is not present and displays an appropriate message to the instructor about how to correctly use the tool in this case
- Hid unimplemented options from the UI

rainbow


Kevin-MM:
Resolved conflicts and retested. OK.

PR #4281, opened 2019-08-03
[Refactor:Autograding] ubuntu:custom -> autograding-default

### LINKED ISSUE #3343: Replace internal ubuntu:custom with image in Submitty docker hub
Given that we have a place on Docker Hub to place images, as well as a repository to store Dockerfiles and components that make up them, we should move to storing this image on Docker Hub and in the repo as opposed to it being within the Submitty repo. The suggested name I'd have is "submitty/full" or "submitty/full_course" for the image.

Part of the installation process would then to be just grabbing this image.

MasterOdin:

Closes #3343
The `ubuntu:custom` Docker image was being built as part of `install_system.sh` from a Dockerfile that was contained in the repo.

### What is the new behavior?
The Dockerfile was moved to [Submitty/DockerImages](https://github.com/Submitty/DockerImages) repo and which was used to create the [submitty/ubuntu](https://hub.docker.com/r/submitty/ubuntu). For backwards compatibility, after getting the `submitty/ubuntu` image, we re-tag it as `ubuntu:custom`.

### Other information?

1. `vagrant destroy`
1. `vagrant up`


MasterOdin:
Submitty/ubuntu does have a latest tag (in addition to the date tag). The plain ubuntu:custom cannot have a latest tag as otherwise it'll collide with the official Ubuntu image. Ideally, we'd go through and convert all usages to submitty/ubuntu (or whatever name is come up), but I held off doing that here so as to avoid conflicting with your work.

I'm also fine with changing the name (I don't particularly care for it), but don't really have a good replacement name and I'm open to suggestions.

bmcutler:
let's re-name this:   autograding-default:latest
(on the submitty dockerhub)

MasterOdin:
@bmcutler Renamed it to `submitty/autograding-default` and used the latest tag (but still have a `YYYYMMDD` tag as well for the image on Docker Hub). Let me know what else you'd like changed here.

emaicus:
Added a commit to make it so that ```submitty/autograding-default:latest``` is used by default autograding containers. 

Also, learned that the ```docker``` module in python throws an exception at creation time if the requested image doesn't exist on the system. Before, it would pull the image in automatically. This isn't a problem with this PR, just something that we'll have to sort out; we'll need some infrastructure to help us deploy an image across all worker machines, etc. I added a migration to pull in ```submitty/autograding-default:latest``` to avoid cases where ```autograding-default``` doesn't exist on a primary machine.

bmcutler:
@emaicus or @MasterOdin 
Can one of you fix the merge conflicts?

MasterOdin:
@emaicus I would consider it a good thing that images are not automatically pulled during grading as it prevents an attack vector where if someone could insert a malicious image into the config file and have it pulled in and run without any validation from the sysadmins.

@bmcutler done

PR #4283, opened 2019-08-05
[Refactor:API] More parameter checking for course creation

zjxiaohan:

Course creation endpoint doesn't check if the parameters required exist.

### What is the new behavior?
Add parameter check.


zjxiaohan:
> I couldn't generate the conditions to create an error message,

That's for API as users have fewer restrictions in API and it makes sense to

PR #4290, opened 2019-08-05
[Feature:MobileUI] Excused Absence Extensions

mktung:
- Separated css and js into their own files
- Collapses on mobile
- Date picker for new due date
- Better error handling for inputs
- Notifications done through php and global header instead of js
- Deleted unnecessary CSS class .persist-thead
- Moved js out of server.js into its own
- Moved shared css with latedays.twig to exceptionform.css

Somethin like this:
Peek 2019-08-05 18-00


site/app/templates/admin/Extensions.twig
```diff
@@ -1,113 +1,128 @@
...
+                Student ID:
+                <input name="user_id" id="user_id" class="option-input" type="text">
+            </label>
+            <label for="late_days">
+                Give the Days of Extension:
+                <input name="late_days" id="late-days" class="option-input" type="text" onkeyup="clearDate()">
+            </label>
+            <label for="late_calendar">
+                Or Pick a New Due Date:
+                <input name="late_calendar" id="late-calendar" class="option-input" type="text" onchange="clearLateDays()">
```
jmcknight98: Maybe adding autocomplete="off" here would be good.  Chrome would try to autocomplete the date and it sometimes covers up the date picker.
googlechromeissue

site/app/controllers/admin/LateController.php
```diff
@@ -143,43 +144,76 @@ public function updateExtension() {
...
                 return Response::JsonOnlyResponse(
                     JsonResponse::getFailResponse($error)
                 );
             }
-            if ((!isset($_POST['late_days'])) || $_POST['late_days'] == "" || (!ctype_digit($_POST['late_days']))) {
-                $error = "Late Days must be a nonnegative integer";
+            $late_days = null;
+            if (isset($_POST['late_days']) && $_POST['late_days'] != "") {
+                $late_days = $_POST['late_days'];  
+                if (intval($late_days) < 0 || !ctype_digit($late_days)) {
```
jmcknight98: This should probably only allow positive integers.  Right now if you give an extension of 0 days it gives a success message then does nothing.

site/app/templates/admin/Extensions.twig
mktung: I've added autocomplete="off" with my latest commits

site/app/controllers/admin/LateController.php
mktung: We update with a value of 0 to delete extensions, so rather than remove it as an option, I've added an additional handler if no changes are made. It should look something like this: 
Screenshot from 2019-08-08 14-03-34

mktung:
- I've changed the team popup to be ambiguous if you're making or deleting an extension
- The date picker for extensions should be fixed now
- Latedays should be fixed again

PR #4294, opened 2019-08-06
[Feature:Forum] Add max limit to length of a post

shailpatels:
This PR adds a max content length of 5000 for creating threads/posts and editing them.


site/app/controllers/forum/ForumController.php
```diff
@@ -288,6 +288,14 @@ public function publishThread(){
         $thread_title = trim($_POST["title"]);
         $thread_post_content = str_replace("\r", "", $_POST["thread_post_content"]);
         $anon = (isset($_POST["Anon"]) && $_POST["Anon"] == "Anon") ? 1 : 0;
+
+        if(strlen($thread_post_content) > 5000 ){
+            $this->core->addErrorMessage("Posts cannot be over than 5000 characters long");
+            $result['next_page'] = $this->core->buildNewCourseUrl(['forum', 'threads', 'new']);
+            $this->core->getOutput()->renderJson($result);
```
zjxiaohan: Please use `renderJsonFail` and put the `$result` in the `data` part of it as specified at https://submitty.org/developer/json_responses

A consistent format of JSON responses is helpful for future API development.

shailpatels: @zjxiaohan Is there anything special I have to do with jsonfail other than that, for somereason it just kicks me back to the course list page
This is what I tried doing

```
$result['next_page'] = $this->core->bulldNewCourseURL(['forum', 'threads', 'new']);
return $this->core->getOutput()->renderJsonFail("Posts cannot be over 5000 characters long", $result);
```

zjxiaohan: Oh sorry I didn't notice that the recipient end will only be changed by https://github.com/Submitty/Submitty/pull/4267 which has not been merged yet. This PR is good for now.

shailpatels: Ah ok that makes sense, in that case I'll wait for that PR to get merged and update this to use the expected format :)

bmcutler: merged #4267, please update :)

zjxiaohan:
I'll do the last piece.

andrewaikens87:
I modified this to use a const instead of hardcoding 5000. Didn't test but should work. Please update the corresponding twig files to fix the 5000 limit to use the const value.

PR #4296, opened 2019-08-06
[Bugfix:InstructorUI] robustify course materials

### LINKED ISSUE #4237: Course Materials dates file is missing
When the course_materials_file_data.json is missing and/or the permissions are broken, we can't set the dates for course materials.  The information doesn't save and resets to infinity.  (With no obvious error to the user)

Let's add an error message when we fail to save the date.  So at least the user knows something is wrong with the file/permissions.

Let's write a migration to create this file & set permissions (if it does not exist or permissions are wrong).

thef1stofhername:

Fixes #4237 
Previously, if a course did not have a course_materials.json for whatever reason, saving dates would just fail silently and revert to the original infinity. Also if the permissions were messed up, no error was thrown.

### What is the new behavior?
If no json is created on the first load of the page, one is created for the files. If the permissions are wrong, 

### Other information?
Tested by deleting json and altering permissions. Also made new courses to test if there was some issue with the json creation there with the new fix.


thef1stofhername:
yeah I could only think to throw an error because you can only change permissions through the migration. I'm open to suggestions about how to handle it because I at least wanted people to know that something had gone wrong instead of failing silently

Kevin-MM:
I think you could wrap the code block that is generating the PHP notice inside a conditional that checks that the file is_writable(), then display your own error message if the file is not writable.

https://www.php.net/manual/en/function.is-writable.php

thef1stofhername:
that's sort of what I was doing but php doesn't have an alert option like js as far as I can tell, so this is what I came up with. I'll keep working

MasterOdin:
You should be able to call `$this->core->addErrorMessage(...)` to show an error toast on the page, or else you could just group all error messages into a variable and pass that message to the twig template and render it there in some way.

thef1stofhername:
Thank you very much for the suggestion, I'll try that

bmcutler:
@emaicus or @thef1stofhername 
one of you please resolve merge conflicts.
thanks!

PR #4297, opened 2019-08-06
[Bugfix:Notifications] fixed notifications and emails not sending

sak6lab:
The check for whether the notification is for the current user and whether to send only sends when the current user is the target user of the notification or email and they have self notification or email on.

### What is the new behavior?
The check correctly allows notifications and emails to send. If the target user of the notification or email is the current user or the current user has the preference for self notification then send.


bmcutler:
This fixes some things, but at least one thing is still broken.

Clicking the "send emails to myself" settings box does not save this selection.  (it's unchecked on page reload)

I did a test announcement and did receive that email.  But a test reply to someone else did not send me an email for "a new post".

PR #4302, opened 2019-08-07
[Bugfix:Forum] Prevent img markdown on the forum

andrewaikens87:

Users can post inline images. 

### What is the new behavior?
Regex to convert users inline attachments to links.


MasterOdin:
Looking at this and https://github.com/Submitty/submitty.github.io/pull/146, it's not clear to me (as an outsider in the development of the forum) why this is happening, and I think it's important that gets documented, especially given that it's non-standard markdown feature that anyone would probably expect to work out of the box.

Additionally, some possible cases here that I think will be confusing without knowing of the behavior beforehand:
1. hyperlinked images
`[](https://example.com)` will be rewritten (I believe) to be `[http://underthebridge.co.uk/wp-content/uploads/2014/03/Example-main-image1.jpg](https://example.com)` and then show as a link, but be hyperlinked to somewhere else. This could be confusing.
2. images with titles
Test `Alt Text Test` will be rewritten to `Test http://underthebridge.co.uk/wp-content/uploads/2014/03/Example-main-image1.jpg "Example Title" Test` which feels like it could look odd.

andrewaikens87:
This only takes the url from the image and adds it in as a hyperlink. When a user submits a link in plaintext the engine will convert it to a hyperlink. This regex uses this feature. We prefer users upload images as attachments.

MasterOdin:
Also, from a cursory look again, I think you're going to have to repeatedly call the preg_replace on a post until it does not actually replace anything. For example, currently, an image could be inlined by doing: `)` (repl example: https://repl.it/repls/RunnyPureCategories).

andrewaikens87:
Fixed!

site/app/controllers/forum/ForumController.php
```diff
@@ -786,6 +786,12 @@ public function showThreads($thread_id = null, $option = 'tree'){
         if(!empty($posts)){
             $thread_id = $posts[0]["thread_id"];
         }
+        foreach($posts as &$post) {
+            $count = 1;
+            while($count > 0) {
+                $post['content'] = preg_replace('/(?:!\[(.*?)\]\((.*?)\))/', '$2', $post['content'], -1, $count);
+            }
+        }
```
MasterOdin: Use the little thought of do-while here for a bit more readability in the sequence of what you're doing.
```
do {
    $post['content'] = preg_replace('/(?:!\[(.*?)\]\((.*?)\))/', '$2', $post['content'], -1, $count);
} while ($count > 0);
```

andrewaikens87: fixed

PR #4322, opened 2019-08-09
[Refactor] Split adduser.py into two files

### LINKED ISSUE #4301: Split sbin/adduser.py into two files
The current [sbin/adduser.py](https://github.com/Submitty/Submitty/blob/a030f583116a6f739630512d401dd7e501971ad2/sbin/adduser.py) is used both for creating new users in the Submitty database as well as for adding users to a course. However, when you're just looking to add a user to a course, this is a clunky interface as the system asks you all the questions about user details (name, email, password) which is generally just hitting enter through.

We should split the file into two separate pieces:
1. sbin/adduser.py (just creates a user)
1. sbin/adduser_course.py (just adds a user to a course)

The first one should have the usage of `sbin/adduser.py <user_id>` and the second should have usage of `sbin/adduser_course.py <user_id> <semester> <course>`.

The relevant parts for adding to a course is https://github.com/Submitty/Submitty/blob/a030f583116a6f739630512d401dd7e501971ad2/sbin/adduser.py#L115-L141

Once done, need to modify https://github.com/Submitty/Submitty/blob/e5485b9cae3546b906fd8872f568b1085092c01c/sbin/submitty_daemon_jobs/submitty_jobs/jobs.py#L330
to take advantage of the new script (fixing a bug where that currently won't work if using DatabaseAuthentication). Will also need to update https://submitty.org/sysadmin/course_creation as appropriate and update https://github.com/Submitty/Submitty/blob/7dc7c26f3a2108cac86b876aa710fcfa4f09b59f/.setup/submitty_sudoers.

zjxiaohan:

Closes #4301 

### What is the new behavior?
See #4301

### DIFF
--- a/sbin/adduser_course.py
+++ b/sbin/adduser_course.py
@@ -9,6 +9,7 @@ import argparse
 import json
 from os import path
 import subprocess
+import sys
 from sqlalchemy import create_engine, MetaData, Table, bindparam, and_
 
 CONFIG_PATH = path.join(path.dirname(path.realpath(__file__)), '..', 'config')
@@ -23,7 +24,9 @@ def parse_args():
     parser = argparse.ArgumentParser(description='Adds a user to courses')
 
     parser.add_argument('user_id', help='user_id of the user to create')
-    parser.add_argument('--course', metavar='arg', action='append', nargs=3, help='[SEMESTER] [COURSE] [REGISTRATION_SECTION]')
+    parser.add_argument('semester', help='semester of the course')
+    parser.add_argument('course', help='title of the course')
+    parser.add_argument('registration_section', help='registration section that the user is added into')
 
     return parser.parse_args()
 
@@ -31,6 +34,9 @@ def parse_args():
 def main():
     args = parse_args()
     user_id = args.user_id
+    semester = args.semester
+    course = args.course
+    registration_section = args.registration_section
 
     if path.isdir(DATABASE_HOST):
         engine_str = "postgresql://{}:{}@/submitty?host={}".format(DATABASE_USER, DATABASE_PASS, DATABASE_HOST)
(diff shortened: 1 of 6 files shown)


sbin/adduser_course.py
```diff
@@ -0,0 +1,77 @@
...
+DATABASE_HOST = DATABASE_DETAILS['database_host']
+DATABASE_USER = DATABASE_DETAILS['database_user']
+DATABASE_PASS = DATABASE_DETAILS['database_password']
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Adds a user to courses')
+
+    parser.add_argument('user_id', help='user_id of the user to create')
+    parser.add_argument('--course', metavar='arg', action='append', nargs=3, help='[SEMESTER] [COURSE] [REGISTRATION_SECTION]')
```
MasterOdin: Let's make this three positional arguments instead of the flag as the first two are required to even do anything here, and let's just make it only workable with one course at a time. It's a slight degregation of what you could do, but I'm not sure anyone was actually using this to add a user to multiple courses at once anyway.
```
parser.add_argument('semester', type=str)
parser.add_argument('course', type=str)
parser.add_argument('registration_section', type=int, nargs='?', default=None)
```
(add appropriate help messages)

sbin/adduser_course.py
```diff
@@ -0,0 +1,77 @@
...
+    engine = create_engine(engine_str)
+    connection = engine.connect()
+    metadata = MetaData(bind=engine)
+    users_table = Table('users', metadata, autoload=True)
+    select = users_table.select().where(users_table.c.user_id == bindparam('user_id'))
+    user = connection.execute(select, user_id=user_id).fetchone()
+
+    if 'course' in args and args.course is not None and len(args.course) > 0:
+        courses_table = Table('courses', metadata, autoload=True)
+        for course in args.course:
```
MasterOdin: With above done, can remove the if statement and for loop here and just use one course. Adjust the use of indexes of `course` to just be `args.semester`, `args.course`, `args.registration_section`.

sbin/adduser_course.py
zjxiaohan: Sure! I was also wondering if it was used for adding multiple courses by other users so I didn't change it. Will modify that to one course only.

sbin/adduser_course.py
```diff
@@ -0,0 +1,82 @@
...
+DATABASE_PASS = DATABASE_DETAILS['database_password']
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Adds a user to courses')
+
+    parser.add_argument('user_id', help='user_id of the user to create')
+    parser.add_argument('semester', help='semester of the course')
+    parser.add_argument('course', help='title of the course')
+    parser.add_argument('registration_section', help='registration section that the user is added into')
```
MasterOdin: This was missed from my comment, but let's make this final argument optional with default `None`:
`nargs="?", default=None`

Feels cleaner than requiring typing null or whatever on the command line to set no registration section.

sbin/adduser_course.py
```diff
@@ -74,8 +75,8 @@ def main():
     else:
         query = courses_u_table.update(values={
             courses_u_table.c.registration_section: bindparam('registration_section')
-        }).where(courses_u_table.c.user_id == bindparam('user_id'))
```
zjxiaohan: This didn't work before.

```
sqlalchemy.exc.StatementError: (sqlalchemy.exc.InvalidRequestError) A value is required for bind parameter 'user_id'
[SQL: UPDATE courses_users SET registration_section=%(registration_section)s WHERE courses_users.user_id = %(user_id)s]
[parameters: [{'registration_section': '1'}]]
```

After adding `user_id=user_id`:
```
sqlalchemy.exc.CompileError: bindparam() name 'user_id' is reserved for automatic usage in the VALUES or SET clause of this insert/update statement.   Please use a name other than column name when using bindparam() with insert() or update() (for example, 'b_user_id').
```

zjxiaohan: That's why I am changing this line.

zjxiaohan:
FIXIT: empty string for email

PR #4238, opened 2019-07-31
[Documentation:Forum] add link to forum markdown

bmcutler:
Changed the color to be more clear when markdown is enabled.
The '?' links the user to some basic placeholder documentation on Submitty.org for markdown.

Screen Shot 2019-07-31 at 4 03 01 AM
Screen Shot 2019-07-31 at 4 03 08 AM

(I was trying to get the '?' link to be blue when markdown was enabled...  but CSS)

PR #4326, opened 2019-08-09
[Bugfix:InstructorUI] Help Icons are back

thef1stofhername:

Closes #4238 
The help icons disappeared

### What is the new behavior?
The help icons are back. I greped and found/fixed every instance

### Other information?
Can be tested by visiting each page rendered by the twig and looking for icons


MasterOdin:
Should you also remove https://github.com/Submitty/Submitty/blob/c9688e1ce49e371b2760dad39cfb3c17945e114a/site/public/css/input.css#L73-L79 then as well? Also I guess apply that blue color to all of these instances via one of the standard classes?

PR #4335, opened 2019-08-12
[Feature:Autograding] solution to generate expected output

drumilpatel2000:

Continuation of PR #4137


.flake8
```diff
@@ -32,12 +32,15 @@ exclude=
     autograder/autograder/grade_item.py,
     autograder/autograder/insert_database_version_data.py,
     autograder/autograder/packer_unpacker.py,
+<<<<<<< HEAD
```
emaicus: Fix this

autograder/autograder/autograding_utils.py
```diff
@@ -128,17 +128,15 @@ def log_stack_trace(log_path, job_id="UNKNOWN", is_batch=False, which_untrusted=
 # ==================================================================================
 
 def setup_for_validation(working_directory, complete_config, is_vcs, testcases, job_id, log_path, stack_trace_log_path):
-    """ Prepare a directory for validation by copying in and permissioning the required files. """
```
emaicus: It looks like this was accidentally modified

autograder/autograder/autograding_utils.py
```diff
@@ -128,17 +128,15 @@ def log_stack_trace(log_path, job_id="UNKNOWN", is_batch=False, which_untrusted=
...
-
-    tmp_autograding = os.path.join(working_directory,"TMP_AUTOGRADING")
```
emaicus: I'm unsure why these changed which line they were on.

autograder/autograder/execution_environments/container_network.py
```diff
@@ -467,10 +467,7 @@ def execute_random_input(self, untrusted_user, executable, arguments, logfile, c
         'server' : False,
         'outgoing_connections' : []
     }
-    # Create a container to generate random input inside of.
```
emaicus: It looks like this was mistakenly modified

autograder/autograder/execution_environments/container_network.py
```diff
@@ -467,10 +467,7 @@ def execute_random_input(self, untrusted_user, executable, arguments, logfile, c
...
-    container = Container( container_spec, untrusted_user, self.random_input_directory, False, self.is_test_environment, self.log_message)
```
emaicus: Same as above

site/app/controllers/student/SubmissionController.php
```diff
@@ -697,6 +697,7 @@ public function ajaxUploadSplitItem($gradeable_id, $merge = null, $clobber = nul
             "course" => $this->core->getConfig()->getCourse(),
             "gradeable" => $gradeable->getId(),
             "required_capabilities" => $gradeable->getAutogradingConfig()->getRequiredCapabilities(),
+            "generate_output" => false,
```
bmcutler: Let's make this an optional field.  I would rather not change the queue file for non generated output gradeables.  Only put this field in the queue files for building a gradeable with generated output.

If installed on a server that had a backlog of regular autograding, this would be a breaking change.

Also if this field is required, it will mean we need to change the batch regrade python script.  (Maybe you did that.. 
 haven't finished reviewing).

sbin/submitty_autograding_shipper.py
```diff
@@ -234,8 +234,11 @@ def prepare_job(my_name,which_machine,which_untrusted,next_directory,next_to_gra
 
     # log completion of job preparation
     obj = packer_unpacker.load_queue_file_obj(JOB_ID,next_directory,next_to_grade)
-    partial_path = os.path.join(obj["gradeable"],obj["who"],str(obj["version"]))
-    item_name = os.path.join(obj["semester"],obj["course"],"submissions",partial_path)
+    if not obj["generate_output"]:
```
bmcutler: Make all these if statements if defined (generate output) and generate_output == true 
[in python syntax of course]

grading/TestCase.cpp
```diff
@@ -119,7 +122,9 @@ std::string getPathForOutputFile(const TestCase &tc, std::string &filename, std:
   std::string expectedPath = getOutputContainingFolderPath(tc, filename);
   std::string requiredPath ;
   if (expectedPath.substr(0,11) == "test_output"){
-    requiredPath = expectedPath + id + "/"; 
+    requiredPath = expectedPath + id + "/";
+  } else if (expectedPath.substr(0,13) == "generated_out") {
```
bmcutler: why only match string generated_out?
Future greps for generated_output will miss this instance in the code.

drumilpatel2000:
@bmcutler @emaicus  Any reviews

PR #4348, opened 2019-08-15
[Testing:Travis] Remove pydocstyle dependency declaration

MasterOdin:

`pydocstyle` is specified in the Travis-CI configuration. This was necessary as it was broken after a new version of it was released that broke `flake8-docstrings`. We added the specific version in #4018 and then resolved in #4279.

### What is the new behavior?
Given that we're installing any version of pydocstyle, there's no need to specify it anymore as it's pulled in automatically as part of `flake8-docstrings`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ jobs:
       language: python
       python: 3.6
       install:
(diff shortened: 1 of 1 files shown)


zjxiaohan:
Oh yeah forgot it's that failure made me add it in the first place. Thanks for reverting that.

PR #4352, opened 2019-08-17
[Refactor] Use a Set for JS and CSS files to load

MasterOdin:

Noticed in #4274, we double load some number of JS (and possibly CSS) files in a handful of places around the site.

### What is the new behavior?
Uses a Set class to ensure that we only ever load one copy of a JS/CSS file into the site. Insertion order is preserved in the Set so that printing the elements out to the HTML buffer remains the same as the old array method.

### Other information?

Requires doing `./INSTALL_SUBMITTY.sh restart_web` to have the `Ds\Set` class be available after the apt install (as `php7.2-fpm` needs to be restarted).


migration/migrator/migrations/system/20190816211332_php_ds.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def up(config):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    os.system("apt install -qy php-ds")
```
zjxiaohan: I am not familiar with `pecl`, but, after doing a quick search, it looks possible to have different versions by installing using `apt` and `pecl`. 

https://superuser.com/questions/451832/what-is-the-difference-between-apt-get-install-php-apc-pecl-install-apc

To avoid discrepancy between travis and vagrant that may potentially lead to mysterious bugs (although not likely for this simple case), is it possible to all use pecl?

MasterOdin: In theory, I agree with you that we should use the same, however getting Travis to use apt instead of pecl is a major pain and not really worth the hassle as we transition to BuildBot. On the production server, it's probably better to use apt as it's easier for a sysadmin to manage.

Issue #4350, opened 2019-08-16
Redirect to a simple static 'System Upgrading' page


bmcutler:
**What problem are you trying to solve with Submitty**
When we run INSTALL_SUBMITTY.sh there is generally a 1-2 minute period when some resources related to npm are unavailable (directory permissions?) and most webpages show confusing errors to the users.  If they are patient and hit reload in a minute or two, everything is fine.  During this time lengthy, jumbled messages get logged to site_errors (and drown out any  messages for errors we do need to fix)

**Describe the way you'd like to solve this problem**
During the period when we are updating pieces of submitty that will break the site, can we instead redirect all pages to a simple static page saying something like "Submitty is being updated. Please try again in 2 minutes."  And if we log anything at all, we could just have a 1 liner log that someone tried to load this specific page and we served them the updating page instead??

MasterOdin:
I think the best way to do things is modify the following line to have `index.html` be first:
https://github.com/Submitty/Submitty/blob/72f7b74551794db55883e86211a0d0f5b63de950/.setup/apache/submitty.conf#L22

and then for the first step of `INSTALL_SUBMITTY_HELPER_SITE.sh` to be copy in a `index.html` to `/usr/local/submitty/site/public` which will have the default git permissions of 644 (so it should be served just fine by apache, though we could copy the file to `/tmp` first, set whatever permissions we want, and the move it into place in `site/public`). Then, after the script concludes, we delete the `index.html` file and users should then start being serviced by `index.php` again. We would do no logging (as it's just an html file) and that's probably fine.

PR #4354, opened 2019-08-17
[SYSADMIN ACTION][Feature:Installation] static maintenance webpage

zjxiaohan:
To take advantage of this feature, it is necessary to modify the apache configuration file.
Detailed instructions will be posted here:
https://submitty.org/sysadmin/version_notes/v19.08.03
For upcoming release:
https://github.com/Submitty/Submitty/releases/tag/v19.08.03

While running INSTALL_SUBMITTY.sh to update the submitty source code (~1-2 minutes), there will be a brief interruption in serving all webpages on the site.  To avoid confusing errors, we place a simple static maintenance webpage instructing the user to try again in a couple minutes.

Closes https://github.com/Submitty/Submitty/issues/4350

Implemented as suggested by @MasterOdin

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER_SITE.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER_SITE.sh
@@ -6,6 +6,10 @@
 
 echo -e "Copy the submission website"
 
+echo "Submitty is being updated. Please try again in 2 minutes." > /tmp/index.html
+chmod 644 /tmp/index.html
+cp -p /tmp/index.html ${SUBMITTY_INSTALL_DIR}/site/public
+
 THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 source ${THIS_DIR}/bin/versions.sh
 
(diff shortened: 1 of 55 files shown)


.setup/INSTALL_SUBMITTY_HELPER_SITE.sh
```diff
@@ -149,8 +153,4 @@ chmod 550 ${SUBMITTY_INSTALL_DIR}/site/cgi-bin/git-http-backend
 # cache needs to be writable
 find ${SUBMITTY_INSTALL_DIR}/site/cache -type d -exec chmod u+w {} \;
 
-# return the course index page (only necessary when 'clean' option is used)
```
zjxiaohan: This is removed as a continuing part of https://github.com/Submitty/Submitty/pull/4337

MasterOdin:
@zjxiaohan I made some changes to the branch as there was a couple things I didn't think about:
1. on initial installation, `site/public` does not exist so have to make it
1. have to insert a rewrite rule for when index.html exists in apache so that the new URLs (like `/home`) would serve `index.html` instead of `index.php`.

zjxiaohan:
Rewrite rule is something I didn't consider and now I know that my new URLs were not interrupted during local tests because they were redirected to `index.php`. Thanks for editing this!

MasterOdin:
These two lines however cause a 403 Forbidden error and need to be updated:
https://github.com/Submitty/Submitty/blob/9634054480db5f12e420ab4af23bbde9b619dfd3/.setup/INSTALL_SUBMITTY_HELPER_SITE.sh#L120-L124

I think what we might want to do here is change this to be a find over all files/directories to change the permissions, but to skip `site/public/index.html` (as that needs to retain the other read bit). Additionally, we might want to change it such that we iterate first over all directories giving them the appropriate permissions (551), and then iterate over files giving them the right permissions (440).

MasterOdin:
Checked that the permissions did not change on any files by applying this patch by doing:
```
cd /usr/local/submitty/site
find . -exec stat --format='%n %A %U %G' {} \; | sort
```

and using `diff` on the output from before running this and after.

PR #4359, opened 2019-08-17
Grade Inquiry UI Tweaks

andrewaikens87:
### What is the new behavior?
Adds a scroll to posts when the open regrade inquiry button is clicked. Also hides the button. If there are posts when the page loads the button will not appear and the div will be present.


andrewaikens87:
Thanks for the feedback. I removed the scripts that were in the code segment you shared above. Let me know if I missed anything else.

PR #4370, opened 2019-08-19
[Refactor] Add better error handling for uploaded files

### LINKED ISSUE #4029: Better error handling for uploading files
Currently, many of the checks for uploading a file under course materials, and student images are not working. These checks also appear to be copied over into each function also. There are probably other places that share this copied and pasted code that should be fixed. As a result, a vague error message is sent to the user like "could not upload file(s)". 

Use PHP's upload errors to give a better response to the user, use CONTENT_LENGTH errors to check if a POST request was too large. Create a function (probably under fileutils) to handle validating the status of upload files.
https://www.php.net/manual/en/features.file-upload.errors.php

shailpatels:
This PR adds a function in FileUtils to validate uploaded files.
This properly validates filename, filesize, and any other errors that might have occurred during upload

Todo: there's still a lot of other places that are checking uploads with a copied and pasted way that's not entirely correct that need to be replaced with this function

closes #4029


site/app/libraries/FileUtils.php
```diff
@@ -566,4 +566,77 @@ public static function areWordsInFile(string $file, array $words)
...
+     *
+     * @param array $files - should be in the same format as the $_FILES[] variable
+     * @return array representing the status of each file
+     * e.g. array('name' => 'foo.txt','type' => 'application/octet-stream', 'error' =>      
+     *            'success','size' => 100, 'success' => true)
+     * if $files is null returns failed => no files sent to validate
+     */
+    public static function validateUploadedFiles($files){
+
+        if(!isset($files)){
```
MasterOdin: Let's change this to be `!empty($files)` as the message applies to the empty `[]` as well.

site/app/libraries/FileUtils.php
```diff
@@ -566,4 +566,77 @@ public static function areWordsInFile(string $file, array $words)
...
+                case UPLOAD_ERR_CANT_WRITE:
+                    $err_msg = "Failed to write file to disk";
+                    break; 
+                case UPLOAD_ERR_EXTENSION:
+                    $err_msg = "File upload stopped by extension";
+                    break; 
+                default:
+                    $err_msg = "Unknown error";
+                    $success = false;
+            }
```
MasterOdin: Can you use [ErrorMessages::uploadErrors](https://github.com/Submitty/Submitty/blob/master/site/app/libraries/ErrorMessages.php#L20) function for this switch case. I think it's just missing the `UPLOAD_ERR_INI_SIZE` and `UPLOAD_ERR_FORM_SIZE` cases.

site/tests/app/libraries/FileUtilsTester.php
```diff
@@ -157,4 +157,204 @@ public function testGetAllDirs() {
...
+
+        $_FILES["files{$part}"]['tmp_name'][] = $tmpname;
+        $_FILES["files{$part}"]['error'][] = $err;
+
+    }
+
+    private function fillToSize($fd, $target_size){
+        fseek($fd, $target_size-1,SEEK_CUR); 
+        fwrite($fd,'a'); 
+    }
```
MasterOdin: Let's combine this function with the one above as they are always used together.

site/app/libraries/FileUtils.php
shailpatels: Thanks for pointing that out!

shailpatels:
@MasterOdin do you think it would be worth it to build/modify a similar function to handle validation of zip files, I didn't look to closely at how we're doing it now

site/app/libraries/FileUtils.php
```diff
@@ -3,6 +3,7 @@
 namespace app\libraries;
 use app\exceptions\FileReadException;
 use app\libraries\Utils;
+use app\libraries\ErrorMessages;
```
MasterOdin: Remove these two lines. You don't need to add a `use ...` line for things under the same namespace as the file itself.

site/tests/app/libraries/FileUtilsTester.php
```diff
@@ -157,4 +157,187 @@ public function testGetAllDirs() {
...
         FileUtils::recursiveRmdir($base_dir);
     }
+
+    private function buildFakeFile($fd, $filename, $part = 1, $err = 0, $target_size = 100) {
+        fseek($fd, $target_size-1,SEEK_CUR); 
+        fwrite($fd,'a'); 
+        fclose($fd);
+
+
+
```
MasterOdin: Shrink these three empty lines into just one empty line

MasterOdin:
I can't really comment on it given that I'm not sure how we're handling zips. After unifying the regular file uploads, would probably be worth looking at how the zip uploading works.

shailpatels:
> I can't really comment on it given that I'm not sure how we're handling zips. After unifying the regular file uploads, would probably be worth looking at how the zip uploading works.

Ok sounds good, I'll make the requested changes and look into zips sometime later, I assume we're doing a similar function to uploads that got copied around to different parts but I'll have to look into it more.

PR #4371, opened 2019-08-20
[Bugfix:InstructorUI] web GUI for adding existing student

### LINKED ISSUE #4367: New User UI broken (bad merge?)
**Describe the bug**
Using the instructor web gui to add a user to a course who already exists in the master
database is currently broken.

When I type in the username it does correctly fill in the form with their other data, but I get a popup "A user with that ID already exists".

This may have been broken with a bad merge, or just a general conflict with other refactors.

I'm testing this when on branch PR #4299 (it may also be broken on master).  I did resolve a merge conflict related to users but I don't think this is the cause.

zjxiaohan:

Fixes #4367 

### What is the new behavior?
It works as normal.

### Other information?
This bug is from #4225 . In the past we decide if a user is in a course or not by calling `loadUser` to load it from the **course db**. However, after #4225, the **master db** info is loaded too. That makes `loadUser($user_id) === null` fail to work when the user exists in the **master db** but not in **course db**.

The changes are made so that we don't check if user exists - frontend knows it and prevents regular instructors from doing so. If malicious instructors try to directly send a POST, it will fail when the user is inserted as `user_id` is a primary key.

### DIFF
--- a/site/app/controllers/admin/UsersController.php
+++ b/site/app/controllers/admin/UsersController.php
@@ -273,10 +273,6 @@ class UsersController extends AbstractController {
             }
         }
         else {
-            if ($user->getRegistrationSection() !== null) {
-                $this->core->addErrorMessage("A user with that ID already exists");
-                $this->core->redirect($return_url);
-            }
             $user = $this->core->loadModel(User::class);
             $user->setId(trim($_POST['user_id']));
         }
--- a/site/app/controllers/admin/UsersController.php
+++ b/site/app/controllers/admin/UsersController.php
@@ -273,7 +273,7 @@ class UsersController extends AbstractController {
             }
         }
         else {
-            if ($user !== null) {
+            if ($user->getRegistrationSection() !== null) {
                 $this->core->addErrorMessage("A user with that ID already exists");
                 $this->core->redirect($return_url);
             }


zjxiaohan:
`getUsersById` loads from course db... probably we still need functions to extract user info from course db only? @MasterOdin

site/app/controllers/admin/UsersController.php
```diff
@@ -273,7 +273,7 @@ public function updateUser($type='users') {
             }
         }
         else {
-            if ($user !== null) {
+            if ($user->getRegistrationSection() !== null) {
```
bmcutler: Is this the right check?  The registration section can be NULL when a student drops.
(And I think the value is NULL rather than a string NULL)

bmcutler: I've tested this and the GUI works correctly in all cases I tested.
So I'm going to merge this since it's an urgent bug.
@zjxiaohan and @MasterOdin please follow up if there are remaining concerns.

zjxiaohan: > Is this the right check? The registration section can be NULL when a student drops.
(And I think the value is NULL rather than a string NULL)

No. I removed in the second commit.
There are remaining concerns though. I will open an issue later.

PR #4415, opened 2019-09-02
[SampleData:Numeric Ids] Add numeric ID sample data for users on setup

shailpatels:
Modified setup_sample_courses.py to create numeric ids and populate the database with them for student users
extra users (the randomly generated ones) will have binary strings for their numeric ids

related to #3465


.setup/data/numeric_ids.txt
```diff
@@ -0,0 +1,35 @@
+074746041
```
MasterOdin: where is this file used and do we need it?

bmcutler: these are the 35 ids from the pdf of handwritten RINS

PR #4418, opened 2019-09-02
[Database:Numeric Ids] Add DB query to get user by numeric id

shailpatels:
This PR adds the function to query a user by a numeric id, works like the getUserByID function.

Adds an additional check to when bulk uploading by QR if the user_id is invalid will check against numeric_ids

### DIFF
--- a/site/app/libraries/database/PostgresqlDatabaseQueries.php
+++ b/site/app/libraries/database/PostgresqlDatabaseQueries.php
@@ -24,8 +24,15 @@ use app\models\SimpleStat;
 
 class PostgresqlDatabaseQueries extends DatabaseQueries{
 
-    public function getUserById($user_id) {
-        $this->submitty_db->query("SELECT * FROM users WHERE user_id=?", array($user_id));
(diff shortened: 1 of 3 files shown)


site/app/libraries/database/PostgresqlDatabaseQueries.php
```diff
@@ -67,6 +67,49 @@ public function getUserById($user_id) {
...
+WHERE u.user_numeric_id=?", array($numeric_id));
+
+            if ($this->course_db->getRowCount() > 0) {
+                $user = $this->course_db->row();
+                if (isset($user['grading_registration_sections'])) {
+                    $user['grading_registration_sections'] = $this->course_db->fromDatabaseToPHPArray($user['grading_registration_sections']);
+                }
+                $details = array_merge($details, $user);
+            }
+        }
```
MasterOdin: Move this into a function that both `getUserBy*` call with parameter of just the `user_id` please, no reason to duplicate this across both.

PR #4421, opened 2019-09-03
[Testing:Response] Add class testing for different types of responses

MasterOdin:

The various response classes are tested indirectly by other tests.

### What is the new behavior?
Adds direct tests of the response classes to ensure correct behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ __pycache__/
 *.synctex.gz
 .coverage
 
+/site/cache/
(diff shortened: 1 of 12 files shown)


MasterOdin:
Leaving the gitignore for cache in place but it now shouldn't be generated for any of the existing tets. If one does not set the debug flag for the config to pass to core before using `Output()`, twig will create `site/cache/` in the local directory and place its parsed files there.

PR #4424, opened 2019-09-04
[Version:PDFAnnotator] Update PDF Annotator to latest release version

shailpatels:
changes .setup/bin/versions.sh to use the latest release of submitty/pdf-annotator.js

@MasterOdin I was going to comment about upgrading the version in your PR about changing the pen size but in order to test it I had to manually change it. So I made a PR with the changes

### DIFF
--- a/.setup/bin/versions.sh
+++ b/.setup/bin/versions.sh
@@ -9,7 +9,7 @@ export AnalysisTools_Version=v.18.06.00
 export Lichen_Version=v.18.12.00
 export RainbowGrades_Version=v19.07.00
 export Tutorial_Version=v.19.02.00
-export Pdf_Annotate_Js_Version=v.18.10.00
+export Pdf_Annotate_Js_Version=v.2.1.0
 
 # JAVA
 export JUNIT_VERSION=4.12


bmcutler:
Hmm @MasterOdin 
Why did we jump from v18 to v2?

MasterOdin:
@shailpatels, @bmcutler I'd prefer closing this for #4425 as we don't actually use this variable or the checked out repo anywhere and instead have it defined in the [site/package.json](https://github.com/Submitty/Submitty/blob/master/site/package.json#L15) and have upgrading it handled by dependabot, and so:
a. no point having it hang around causing confusion like this
b. no point wasting some amount of time cloning that repo down and then doing nothing with it anywhere.

@bmcutler Given pdf-annotator.js is a well-defined library with a well-defined (and not fast changing) interface to it, it lends itself well to using semantic versioning where it's easy to tell "should I increment major, minor, or bugfix version" unlike Submitty which is more nebulous (which is better oriented to calendar versioning). This would also make it easier for other people to use the library in their code.

shailpatels:
@MasterOdin Ok thanks that makes sense, I wasn't sure if this was how you wanted to upgrade this dependency its just what I did to test the latest version. 

And yeah NPM should remove the confusion and do this step automatically, I'll close this PR then

PR #4425, opened 2019-09-04
[Refactor:Installation] Remove unused PDFAnnotator version variable and do not grab its repo

MasterOdin:

### Other information?

This variable is unused as we switched to using `npm` to manage the pdf-annotator (and other JS dependencies) some time ago. Removing it so as to not be confusing.

@bmcutler I held off deleting the repo entirely in-case someone was actually using it, but I'm not sure anyone was and so is probably safe to delete if so desired.


bmcutler:
I think it's ok to remove.

PR #4430, opened 2019-09-05
[Testing] Move from coveralls to codecov

MasterOdin:
### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)

### What is the current behavior?

We currently report our test coverage percentage for just PHP to coveralls.io. Had made some prior effort to try and combine reports, but was proving difficult to manage.

### What is the new behavior?

Move to codecov as it's a nicer platform that out of the box supports coverage merging on their end and looks nicer.

### Other information?

Related #4429. However, code coverage is not yet strictly enforced.

Issue #4429, opened 2019-09-05
Add checks on code coverage of new commits


zjxiaohan:
Seeing Matt writing new unit tests these days, I feel we can force new PRs to have at least the same coverage as what we have in the master branch. This would slowly improve the code quality of Submitty over time without too much pain and forces developers to look more closely at their code to avoid some easy bugs (like possibly undefined variables, which happened many times as I can remember).

I am not sure if coveralls can do this. If it can't, codecov is good. I see in other projects codecov bot automatically replies in the PR saying if the coverage is increased/decreased.

shailpatels:
I disagree, forcing new PRs to meet an arbitrary code coverage is going to have developers try to meet to min length as opposed to writing good tests. 

We have a course at RPI that requires hitting a certain coverage as part of the assignment and the majority of student tests try and get the highest coverage possible and as are pretty crappy tests E.g: student hits all the lines through some combination of unit tests but ignores a logical path that causes the program to crash in a valid edge case.

Talking about this subject I'm obligated to mention ["When a measure becomes a target, it ceases to be a good measure."](https://en.wikipedia.org/wiki/Goodhart%27s_law)

However yes we do need to write more tests (this is a major weakpoint of Submitty) but forcing people to write tests or hit a coverage is going to result in crappy code that doesn't help anyone.
I suggest we enforce writing tests through code reviews that ask developers to write tests under certain conditions

- PRs with complex logic e.g late days, timestamps, uploads
- PRs difficult to test manually e.g lots of different states/conditions
- Bugfixes e.g after fixing a bug write a test that specifically prevents it from occuring

There are also areas where we could improve writing tests - if we can reduce the effort and cognitive load to write good tests I'm sure people will write more.

You're points are valid for PHPUnit tests which is pretty well developed and we have some good examples (and we're getting better at this!)
Then there are selenium and integration tests which are less developed and typically harder to set up. They don't factor into code covereage but are still valuable. Then we have large chunks of python scripts doing tons of random back-end functions that probably never get tests or even have any testing infrastructure behind them. So many areas beyond phpUnit tests and coverage we should also consider. 

Sorry for the rant I typically ramble when its late night/morning/whatever time im writing this so I'll go through this tomorow and clear up some points. Glad you're still poking around after gsocs over :)

MasterOdin:
[...]

PR #4434, opened 2019-09-05
[Testing] Replace coveralls with codecov for project-wide code coverage analysis

MasterOdin:

We currently report our test coverage percentage for just PHP to coveralls.io. Had made some prior effort to try and combine reports, but was proving difficult to manage.

### What is the new behavior?

Move to codecov as it's a nicer platform that out of the box supports coverage merging on their end and looks nicer.

### Other information?

Related #4429. However, code coverage is not yet strictly enforced.

Created #4430 so as to hopefully get a working PR message from CodeCov to fully demonstrate it.


zjxiaohan:
Screenshot from 2019-09-06 09-34-31

A dumb question (I don't really know tests except PHP ones..) Are the question mark the expected output?

MasterOdin:
That is a result of me playing around with a couple of different formats for the flags. Originally, I had grouped all python coverage tests together as that was the only way I could get codecov's bash tool to pick them up, but then I looked more into how the tool worked so that I could split it apart into separate pieces. If should be that as we push to master, that matrix loses those rows (I hope).

PR #4435, opened 2019-09-06
[Bugfix:CourseMaterials] incomplete data in course_materials_file_data.json

shailpatels:

* [ ] Tests for the changes have been added/updated (will do this later - courseMaterialsTester needs to be written first so it should be in its own PR 

When an instructor uploads materials using the upload form they are not initially written to the course_materials_file_data.json as a result if an instructor selects the date to 'now' they wont actually be released.

Seen in CV and networking today (not sure if open issue)

### What is the new behavior?
On upload write all files (including zips) to the json file with their upload dates

### Other information?
CourseMaterialsTester doesn't exist yet, I'll add tests for it in a separate PR


bmcutler:
Thanks Shail,
Yes indeed this has been reported as an urgent issue but I didn't know how to reproduce so I haven't made an issue yet.

@emaicus will you please review ASAP so I can install?

PR #4444, opened 2019-09-08
[Testing: Course Materials] Add course materials tests and fix JSON structure

shailpatels:

### What is the new behavior?
adds tests for course materials 

Changes the structure of course_materials_file_data.json in the uploads dir for each course, the top-level 'release_datetime' key has been removed which used to get created on uploading new materials. 

Legacy files will still have the top level release_datetime key


MasterOdin:
Still looking at this and wrapping my head around it, but an initial question I have is what is the usage of the `release_time` top-level value in the `course_materials_file_data.json` file?

shailpatels:
> Still looking at this and wrapping my head around it, but an initial question I have is what is the usage of the `release_time` top-level value in the `course_materials_file_data.json` file?

Honestly I'm not sure, I saw it when I mas making a bugfix to course materials last week but wasn't sure if it was getting used anywhere actually.
My thought was that it was originally there to do some sync between the client time and server time but I don't really know so I didn't touch it.

MasterOdin:
I think it's more to provide a default release time for material that haven't had their date explicitly set. It would be good to clarify with @bmcutler what the intended behavior here is, though removing it or using it again can happen in a different PR.

bmcutler:
There should be no top level release date/time.
That was not the design.  
Every file should have its own release date/time.
If somehow a file is missing from this file (shouldn't happen, but might exist in legacy courses), the release date/time should be end-of-time/infinity.

shailpatels:
> There should be no top level release date/time.
> That was not the design.
> Every file should have its own release date/time.
> If somehow a file is missing from this file (shouldn't happen, but might exist in legacy courses), the release date/time should be end-of-time/infinity.

Ok that makes sense, it looks like right now what its doing is that if there isn't a release datetime passed in on upload it'll use the last used one which is the top level json object in the file. Buy yeah right now on upload a top level release time is created but it should be a small fix.

I can update the course materials file in this PR to switch that format since it'll require updating the tests anyway. marking as wip for now

site/tests/app/controllers/course/CourseMaterialsControllerTester.php
```diff
@@ -0,0 +1,268 @@
...
+        $ret = $controller->ajaxUploadCourseMaterialsFiles();
+        $json = FileUtils::readJsonFile($this->json_path);
+        
+        $expected_ret = array('status' => 'success', 'data' => 'Successfully uploaded!');
+        $this->assertEquals($expected_ret, $ret);
+
+        $files = FileUtils::getAllFiles($this->upload_path, array(), true);
+        $this->assertEquals(4, count($files));
+
+        $f1 = array_key_first($files);
```
MasterOdin: while a useful function, this is only available on PHP >= 7.3 (see header in https://www.php.net/manual/en/function.array-key-first.php). Please add this to `Utils` and use that:
```
public static function getFirstArrayElement($array) {
        foreach($array as $key => $unused) {
            return $key;
        }
        return NULL;
}
```

to match the existing `Utils::getLastArrayElement($array)`

shailpatels: Thanks !

PR #4449, opened 2019-09-09
[Bugfix:Course Materials] Prevent uploading a zip to overwrite other course materials

shailpatels:

Bug introduced in #4435, which allows wriiting to the json file for uploads but for zip files it 
would update every single json entry instead of the new ones

As a result if an instructor had existing materials and uploaded a zip every material would have the release time and status of the new zip file

### What is the new behavior?
Only update the newly extracted zip files instead of everything


site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -406,11 +406,18 @@ public function ajaxUploadCourseMaterialsFiles() {
...
                             $zip->close();
                             $subfiles = FileUtils::getAllFiles($upload_path, array(), true );
                             foreach ($subfiles as $file) {
-                                $json[$file['path']] = array('checked' => '1', 'release_datetime' => $release_time  );
+                                //skip the files that aren't new
+                                if ( in_array($file, $subfiles_before))
+                                    continue;
+
+                                $json[$file['path']] = array('checked' => '1',
+                                'release_datetime' => $release_time  );
```
MasterOdin: If not going to have this all on one line, then have opening/closing braces line up with each array entry on a newline:
```
                                $json[$file['path']] = [
                                    'checked' => '1',
                                    'release_datetime' => $release_time
                                ];
```

shailpatels:
Thanks! I'll make the requested changes in a bit

site/app/controllers/course/CourseMaterialsController.php
shailpatels: Ended up going through the zip file before it was extracted and getting all the filenames from that, it was too messy extracting to a tmp file and then correcting the paths to the real upload_path

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -339,15 +339,26 @@ public function ajaxUploadCourseMaterialsFiles() {
...
+                        $zip = zip_open($uploaded_files[1]["tmp_name"][$j]);
+                        while ($zip_entry = zip_read($zip)) {
+                            $zfiles[] =  zip_entry_name($zip_entry);
+                        }
+                        zip_close($zip);
+
                         $zip = new \ZipArchive();
                         $res = $zip->open($uploaded_files[1]["tmp_name"][$j]);
                         if ($res === true) {
                             $zip->extractTo($upload_path);
```
MasterOdin: Change this to be:
```
                            $entries = [];
                            $disallowed_folders = [".svn", ".git", ".idea", "__macosx"];
                            $disallowed_files = ['.ds_store'];
                            for ($i = 0; $i < $zip->numFiles; $i++) {
                                $entries = $zip->statIndex($i);
                            }
                            $entries = array_filter($entries, function($entry) use ($disallowed_folders, $disallowed_files) {
                                $name = strtolower($entry['name']);
                                foreach ($disallowed_folders as $folder) {
                                    if (Utils::startsWith($folder, $name)) {
                                        return false;
                                    }
                                }
                                if (substr($name, -1) !== '/') {
                                    foreach ($disallowed_files as $file) {
                                        if (basename($name) === $file) {
                                            return false;
                                        }
                                    }
                                }
                                return true;
                            });
                            $zfiles = array_filter($entries, function($entry)) {
                                return substr($entry['name'], -1) !== '/';
                            }
                            $zip->extractTo($upload_path, $entries);
```

This should fix #4475 and not add an entry in `$json` for each folder.

shailpatels:
Requested changes made, thanks @MasterOdin for most of the code !

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -338,15 +338,59 @@ public function ajaxUploadCourseMaterialsFiles() {
...
+
+                        if(!$res){
+                            return $this->core->getOutput()->renderResultMessage("ERROR: Failed to open zip archive", false);
+                        }
+
+                        $entries = [];
+                        $disallowed_folders = [".svn", ".git", ".idea", "__macosx"];
+                        $disallowed_files = ['.ds_store'];
+                        for ($i = 0; $i < $zip->numFiles; $i++) {
+                            $entries[] = $zip->statIndex($i);
```
MasterOdin: Replace `statIndex` with `getNameIndex` and get rid of the indexing on `name` on line 357 and can then also get rid of lines 376-384 as well. I originally was planning on using size and/or crc values on the `$entry` for determining folder or not, but just checking the last character of the name proves reliable enough.

PR #4453, opened 2019-09-10
[Refactor:CourseMaterials] Ceanup Course Materials and bump max uploaded files to 100

### LINKED ISSUE #3208: drag and drop to upload > 20 student photo images might be broken
Is there a limit on # of files?  

I tried to upload 42 images (not using zip, just a collection of .pngs), and only about half of them were uploaded.  Selecting fewer of them at a time they all went through.  

I didn't encounter any error messages.
I don't believe it was a problem with total file size.

shailpatels:

Course Materials still uses a broken method of validating uploads and usually results in wrong/confusing errors. Max limit was set to 20 in the php.ini

### What is the new behavior?
Uses validateUploadedFiles from FileUtils for better error messages.  install submitty has been updated to set max_uploaded_files to 100. Refactors courseMaterialsController to remove a lot of duplicate code as well.

closes #3208

### other information
(note this is going to break one of the test in course materials so I think this should be merged first and then I can go and fix the errors in course materials.

### DIFF
--- a/.setup/install_system.sh
+++ b/.setup/install_system.sh
@@ -491,6 +491,7 @@ EOF
     sed -i -e 's/^post_max_size = 8M/post_max_size = 10M/g' /etc/php/${PHP_VERSION}/fpm/php.ini
     sed -i -e 's/^allow_url_fopen = On/allow_url_fopen = Off/g' /etc/php/${PHP_VERSION}/fpm/php.ini
     sed -i -e 's/^session.cookie_httponly =/session.cookie_httponly = 1/g' /etc/php/${PHP_VERSION}/fpm/php.ini
+    sed -i -e 's/^max_file_uploads = 20/max_file_uploads = 20/g' /etc/php/${PHP_VERSION}/fpm/php.ini
     # This should mimic the list of disabled functions that RPI uses on the HSS machine with the sole difference
     # being that we do not disable phpinfo() on the vagrant machine as it's not a function that could be used for
     # development of some feature, but it is useful for seeing information that could help debug something going wrong
--- a/site/app/controllers/course/CourseMaterialsController.php
+++ b/site/app/controllers/course/CourseMaterialsController.php
@@ -156,10 +156,13 @@ class CourseMaterialsController extends AbstractController {
      * @Route("/{_semester}/{_course}/course_materials/modify_permission")
      * @AccessControl(role="INSTRUCTOR")
      */
-    public function modifyCourseMaterialsFilePermission($filenames, $checked) {
+    public function modifyCourseMaterialsFilePermission($checked) {
         $data=$_POST['fn'];
-        if(count($data)==1){
-            $filename = $filenames;
+        if(is_string($data) || count($data)==1){
+            $data = [$data];
+        }
+    
+        foreach ($data as $filename){
             if (!isset($filename) ||
                 !isset($checked)) {
                 $this->core->redirect($this->core->buildCourseUrl(['course_materials']));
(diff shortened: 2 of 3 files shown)


site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -156,10 +156,13 @@ public function downloadCourseMaterialZip($dir_name, $path) {
...
-        if(count($data)==1){
-            $filename = $filenames;
+        if(is_string($data) || count($data)==1){
+            $data = [$data];
+        }
+    
+        foreach ($data as $filename){
             if (!isset($filename) ||
                 !isset($checked)) {
                 $this->core->redirect($this->core->buildCourseUrl(['course_materials']));
```
emaicus: Github won't let me comment on the exact line because you didn't modify it, but it looks like we are using "now" as the default time here if something goes wrong. Let's use something else to avoid accidental file release (e.g. end of time).

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -233,17 +209,18 @@ public function modifyCourseMaterialsFileTimeStamp($filenames, $newdatatime) {
...
+            $data = [$data];
+        }
+    
+        foreach ($data as $filename){
             if (!isset($filename)) {
                 $this->core->redirect($this->core->buildCourseUrl(['course_materials']));
             }
 
             $file_name = htmlspecialchars($filename);
+            $new_data_time = htmlspecialchars($newdatatime);
```
emaicus: This is declared above as well.

.setup/install_system.sh
```diff
@@ -491,6 +491,7 @@ EOF
     sed -i -e 's/^post_max_size = 8M/post_max_size = 10M/g' /etc/php/${PHP_VERSION}/fpm/php.ini
     sed -i -e 's/^allow_url_fopen = On/allow_url_fopen = Off/g' /etc/php/${PHP_VERSION}/fpm/php.ini
     sed -i -e 's/^session.cookie_httponly =/session.cookie_httponly = 1/g' /etc/php/${PHP_VERSION}/fpm/php.ini
+    sed -i -e 's/^max_file_uploads = 20/max_file_uploads = 20/g' /etc/php/${PHP_VERSION}/fpm/php.ini
```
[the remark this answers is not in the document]
shailpatels: yeah, thanks for the catch!

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -156,10 +156,13 @@ public function downloadCourseMaterialZip($dir_name, $path) {
      * @Route("/{_semester}/{_course}/course_materials/modify_permission")
      * @AccessControl(role="INSTRUCTOR")
      */
-    public function modifyCourseMaterialsFilePermission($filenames, $checked) {
+    public function modifyCourseMaterialsFilePermission($checked) {
         $data=$_POST['fn'];
-        if(count($data)==1){
-            $filename = $filenames;
+        if(is_string($data) || count($data)==1){
```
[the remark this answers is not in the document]
shailpatels: yeah not sure, I added that since it was a check somewhere else but yeah that was the root cause of the optional dir breaking.

site/app/controllers/course/CourseMaterialsController.php
shailpatels: changed, default if no timestamp is given or error is now 'end of time'

MasterOdin:
What is the rationale for bumping this up to 100?

I would (like with apache and postgresql), not have this stuff get configured as part of the main Submitty installation as it really should be something a sysadmin configure specific to their needs and users.  We can certainly offer suggestions in the documentation, or potentially hide this beyond an install option, but making it the default is not the route I'd like to take.

shailpatels:
Its related to this issue https://github.com/Submitty/Submitty/issues/3208

An instructor tried to upload student images manually I think and only the first 20 got uploaded. 
What would the process for making this configurable? Would it be an option in the main install script or do we have something else that instructors can run to edit settings in submitty.

shailpatels:
As discussed today removed increasing max_file_uploads and I'll make a change to documentation on how to instead to increase it by editing the php.ini

MasterOdin:
In a dream scenario, I think the best would be that `install_submitty.sh` makes some suggestions about what the php.ini settings should be and offers to change them to some good default for Submitty. However, it should be possible that a sysadmin can elect to not allow Submitty to change any. Additionally, we should document these suggested values on https://submitty.org/sysadmin/installation (as a new step "Configure PHP") and discuss them. This is important as this is not a "one size fits all" type solution. For example, for my undergrad which had <20 students for most courses, leaving the file upload around 20 (or maybe bumping it to 30) would be ideal. This setting bump would also be for just one form, which already has a good setup for uploading an arbitrary number of files (zip).

With regards to these upload forms, I would suggest we visibly mark what the max is above the drag/drop interface (You can only upload 20 files at a time) and for course uploads and student photos, suggest to use a zip if uploading a large amount of files at once.

shailpatels:
Yeah I think thats a good solution, I believe all the drag and drop interfaces are grabbing the same source code so appending the php.ini info somewhere there is a good idea. I'll make an issue for that

PR #4494, opened 2019-09-15
[Bugfix:InstructorUI] read_file endpoint directory traversal and content-type

MasterOdin:

#4484 removed the usage of the `resolveDirPath` on the Access check, which allowed for arbitrary file traversal using `..` from the site directory through read_file. The check itself relies on the endpoint being run from within a course.

### What is the new behavior?
Re-adds the check and moves the `/read_file` endpoint to `/{_semester}/{_course}/read_file` as it only makes sense when a course is loaded.

### Other information?

It looks like only the wrapper stuff uses this endpoint and the `$wrapper_files` only exist and be non-empty when `Config` loads the course config, so running `buildCourseUrl` should always be safe in this context.


MasterOdin:
Also fixes setting the content-type header so that the CSS file from the `/read_file` endpoint is actually used as as stylesheet:
Screenshot 2019-09-15 08 22 19

PR #4495, opened 2019-09-15
[Bugfix:Installation] Fix Configure Script

emaicus:
Remove bad chowns/chmods.


.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -512,7 +512,8 @@ def write(x=''):
 
 with open(SUBMITTY_USERS_JSON, 'w') as json_file:
     json.dump(config, json_file, indent=2)
-shutil.chown(SUBMITTY_USERS_JSON, 'root', DAEMONPHP_GROUP)
+if not args.worker:
+    shutil.chown(SUBMITTY_USERS_JSON, 'root', DAEMONPHP_GROUP)
```
MasterOdin: This will leave the file chowned to `root:root` and unreadable. Assuming the file is useful, this should probably be:
`shutil.chown(SUBMITTY_USERS_JSON, 'root', DAEMON_GROUP if args.worker else DAEMONPHP_GROUP)`
or replace `DAEMON_GROUP` with whatever group needs read access to this file.

PR #4513, opened 2019-09-19
[Refactor:Discussion Forum] No categories in Discussion Forum improvements

shwehtom89:

Couldn't find an issue, this was requested at the 9/14 Hackathon. 

If an instructor does not create categories, they can still view the page to create new threads which will show the "dummy" category.
 
### What is the new behavior?

 - Dummy category is removed.
 - Cannot create a thread if there are no categories (url redirects). 
 - Notification messages for graders to create categories and for students to contact an instructor to     create categories to use the forum.
- When the forum is enabled in the course settings page, the instructor is notified to create categories if they have not done so.

### DIFF
--- a/site/app/templates/admin/Configuration.twig
+++ b/site/app/templates/admin/Configuration.twig
@@ -31,7 +31,7 @@
                 <div class="option-title">Enable Discussion Forum</div>
                 <div class="option-alt">Choose whether to enable a forum for this course.</div>
                 {% if categories_empty %}
-                    <h4 id="enable-forum-warning" class="red-message">Categories must be created in order to use the Discussion Forum. Click the link below to add categories.</h4>
+                    <h4 id="forum-category-warning" class="red-message">Categories must be created in order to use the Discussion Forum. Click the link below to add categories.</h4>
                 {% endif %}
                 <a id="forum-enabled-message" href="{{ manage_categories_url }}" alt="Customize Categories">Customize categories.</a>
             </label>
@@ -231,19 +231,3 @@
         </div>
     </div>
 </div>
-
-<script type="text/javascript">
-   $(document).ready(function(){
-        $('#enable-forum-warning').hide();
-        var no_categories = "{{categories_empty}}" === "1";
-
-        if (no_categories && $('#forum-enabled').is(':checked')){
-            $('#enable-forum-warning').show();
-        }
-        $('#forum-enabled').change(function(){
-            if (no_categories) {
-                $('#enable-forum-warning').toggle();
-            }
-        });
-   });
-</script>
--- a/site/app/templates/admin/Configuration.twig
+++ b/site/app/templates/admin/Configuration.twig
@@ -30,6 +30,9 @@
             <label for="forum-enabled">
                 <div class="option-title">Enable Discussion Forum</div>
                 <div class="option-alt">Choose whether to enable a forum for this course.</div>
+                {% if categories_empty %}
+                    <h4 id="enable-forum-warning" class="red-message">Categories must be created in order to use the Discussion Forum. Click the link below to add categories.</h4>
+                {% endif %}
                 <a id="forum-enabled-message" href="{{ manage_categories_url }}" alt="Customize Categories">Customize categories.</a>
             </label>
         </div>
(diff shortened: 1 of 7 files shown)


site/public/js/configuration.js
```diff
@@ -3,13 +3,24 @@ $(document).ready(function() {
...
+        updateForumCategoryWarning();
         if ($("#forum-enabled").is(":checked")) {
             $("#forum-enabled-message").show();
         } else {
             $("#forum-enabled-message").hide();
         }
     }
 
+    function updateForumCategoryWarning() {
+        if ($("#forum-category-warning").length) {
```
shailpatels: You don't need to perform a check if the element exists, if JQuery can't find the element show/hide/toggle will have no effect

site/public/js/configuration.js
```diff
@@ -3,13 +3,24 @@ $(document).ready(function() {
...
+            if ($("#forum-enabled").is(":checked")) {
```
shailpatels: Replace the if else logic with $("#forum-enabled").toggle(); instead of using show and hide

site/public/js/configuration.js
```diff
@@ -3,13 +3,24 @@ $(document).ready(function() {
     updateEmailSeatingOption();
 
     function updateForumMessage() {
+        updateForumCategoryWarning();
```
shailpatels: is `updateForumCategoryWarning()` used anywhere else? If not you should just do a
```
 $(#forum-category-warning).toggle();
 $(#forum-enabled-warning).toggle();

File: site/public/js/configuration.js (hunk 1 above)
Comment (shwehtom89) [reply site/public/js/configuration.js]: Fixed!

File: site/public/js/configuration.js (hunk 2 above)
Comment (shwehtom89) [reply site/public/js/configuration.js]: Is there a reason to use `toggle()` over `show()` and `hide()`? Right now we are using `updateForumMessage()` to hide/show the message during document load. I think even if we were to switch to toggle, we would still need to make changes in `Configuration.twig`

File: site/public/js/configuration.js (hunk 3 above)
Comment (shwehtom89) [reply site/public/js/configuration.js]: Removed the function.

File: site/public/js/configuration.js (hunk 2 above)
Comment (shailpatels) [reply site/public/js/configuration.js]: using toggle() is basically the same as having the if else statement with the show and hide but jquery does the work for us. I recommend using it and setting the elements to hidden by default.

You could either do this during the document onload on client side or in the twig create
the elements as hidden.

Either case I think its cleaner to use toggle and do a second client side check to see if we should render the message onload or not

File: site/app/templates/admin/Configuration.twig [hunk 4]
```diff
@@ -30,6 +30,9 @@
             <label for="forum-enabled">
                 <div class="option-title">Enable Discussion Forum</div>
                 <div class="option-alt">Choose whether to enable a forum for this course.</div>
+                {% if categories_empty %}
+                    <h4 id="forum-category-warning" class="red-message">Categories must be created in order to use the Discussion Forum. Click the link below to add categories.</h4>
+                {% endif %}
```
shailpatels: Why isn't the bottom link also part of this twig conditional that way we wouldn't need a client side check to hide it initially

site/app/views/forum/ForumThreadView.php
```diff
@@ -955,6 +957,10 @@ public function createThread($category_colors){
 			$this->core->redirect($this->core->buildCourseUrl());
 			return;
 		}
+        if(empty($this->core->getQueries()->getCategories())){
+            $this->core->redirect($this->core->buildCourseUrl(["forum", "threads"]));
+            return;
+        }
```
MasterOdin: This should be in the controller

MasterOdin:
I would rename this PR to be `[Feature:DiscussionForum] Do not allow thread creation until categories have been created` or something. Refactor should be used when we are just doing internal code change that do not have an effect on the user.

site/app/templates/admin/Configuration.twig
shwehtom89: No reason I can think of besides it being there originally, I will add it to the twig conditional.

site/app/views/forum/ForumThreadView.php
shwehtom89: Will do.

shailpatels:
@shwehtom89  is this ready for review again, if so could you please mark as the requested 
changes as resolved?

site/public/js/configuration.js
shwehtom89: Changed to use toggle. I choose to hide elements in twig.

site/app/views/admin/ConfigurationView.php
```diff
@@ -10,9 +10,11 @@ public function viewConfig($fields, $gradeable_seating_options) {
         $this->core->getOutput()->addInternalJs("configuration.js");
         $this->core->getOutput()->addInternalCss("configuration.css");
         $this->core->getOutput()->addBreadcrumb('Course Settings');
+        $categories = empty($this->core->getQueries()->getCategories());
```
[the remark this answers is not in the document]
shwehtom89: Fixed.

PR #4531, opened 2019-09-21
[Testing] Add tests for all functions in FileUtils

MasterOdin:

Many of the functions in `app/libraries/FileUtils` did not have any tests, or were only indirectly tested.

### What is the new behavior?
All files, with the exception of `recursiveCopy` have explicit tests for them, where that function has been rewritten for #4530 (which has tests for that function).

### Other information?
I fixed/modified/simplified the behavior for:
* createDir - `$recursive` is now the second argument as `$mode` was only used less, and only for testing. All relevant usages have been updated.
* areWordsInFile - will no longer throw PHP warning if passed an invalid or unreadable file, but just return `false` as expected

I moved the following function:
* getGradeablesFromPriorTerm - this function was only used by the `PlagiarismController.php` and its behavior was specific to how that worked that it was not a good candidate for a general utility function.

I removed the following functions:
* getMimeType - This did the same thing as the builtin [mime_content_type](https://www.php.net/manual/en/function.mime-content-type.php) which was (at the time the function was written) marked [mistakenly as deprecated](https://bugs.php.net/bug.php?id=71367). As it wasn't actually deprecated, all usages of this function have been replaced with the builtin.


bmcutler:
@MasterOdin please resolve conflicts

PR #4543, opened 2019-09-25
[Bugfix:Autograding] Properly catch worker/shipper timeouts

emaicus:
```Paramiko``` appears to be able to throw either a ```socket.timeout``` or a python ```TimeoutError```. We now catch both errors, rather than just ```socket.timeout```.


emaicus:
Ready for review.

MasterOdin:
What is causing these timeouts on attempting to connect? Is it RPI network itself being flakey?

After looking at the PR, I have a concern that this does allow an un-connectable machine to completely block the grading queue given enough assignments even if other machines are connectable and could be used.

Issue #4423, opened 2019-09-04
Course Materials restricted to registration section


bmcutler:
**What problem are you trying to solve with Submitty**
Let's add a feature to allow an instructor to release a file or folder to a specific registration section (or sections) and prevent access to those materials from students in any other section.

We'll need to modify the UI slightly.  
But we can store this data in the same .json file used for dates.

By default all course materials file should be available to all students in a valid (non NULL registration section) after the release date.

PR #4568, opened 2019-09-30
[Feature:CourseMaterials] Restrict course materials to registration section

jchulton:
Resolves issue #4423
Allows instructor to assign course materials to only be viewed by certain sections.
Before this is merged, please address why $this->core->getConfig()->getCoursePath(); doesn't actually get the course materials path in UsersController.php. Right now I have to hardcode the course materials.


shailpatels:
`
$this->core->getConfig()->getCoursePath();
`
returns the path of the current course your in, its not actually related to course materials
so if you were to load a view in sample that function would return:
`
'/var/local/submitty/courses/f19/sample'
`

course materials is always located at `COURSE_PATH/uploads/course_materials`
and the json file is located one directory above that under uploads.

most of the system uses this line to get course materials 
`$this->core->getConfig()->getCoursePath() . '/uploads/course_materials_file_data.json';`

you can replace line 415 with that
https://github.com/Submitty/Submitty/pull/4568/files#diff-2eab3c5e45f108d4e1c285da15e55362R415

site/app/controllers/admin/UsersController.php
```diff
@@ -411,7 +411,20 @@ public function updateRegistrationSections() {
             if (User::validateUserData('registration_section', $_POST['delete_reg_section'])) {
                 // DELETE trigger function in master DB will catch integrity violation exceptions (such as FK violations when users/graders are still enrolled in section).
                 // $num_del_sections indicates how many DELETEs were performed.  0 DELETEs means either the section didn't exist or there are users still enrolled.
-                $num_del_sections = $this->core->getQueries()->deleteRegistrationSection($_POST['delete_reg_section']);
+                //why doesn't $this->core->getConfig()->getCoursePath(); work?
```
shailpatels: I don't think it makes sense to not delete a registration section if its used in course materials. A better approach would be to assume a registration section could disappear at any time and if its not valid let course materials ignore it and move on. If there are no valid sections just release to everyone in that case.

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -377,18 +401,41 @@ public function ajaxUploadCourseMaterialsFiles() {
...
                         foreach ($zfiles as $zfile) {
                             $path = FileUtils::joinPaths( $upload_path, $zfile );
-                            $json[$path] = [
-                                'checked' => '1',
-                                'release_datetime' => $release_time
-                            ];
+                            if(!(is_null($sections))){
+                                $sections = @explode(",", $sections);
+                                if($sections == null){
+                                    $sections = [];
```
shailpatels: If `$sections` is null here what is the point of adding an empty array to the json file, I think it would make sense in
the else condition to just not put anything

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -377,18 +401,41 @@ public function ajaxUploadCourseMaterialsFiles() {
...
                     else
                     {
                         if (!@copy($uploaded_files[1]["tmp_name"][$j], $dst)) {
                             return $this->core->getOutput()->renderResultMessage("ERROR: Failed to copy uploaded file {$uploaded_files[1]["name"][$j]} to current location.", false);
                         }else{
-                            $json[$dst] = array('checked' => '1', 'release_datetime' => $release_time  );
+                            if(!(is_null($sections))){
+                                $sections = @explode(",", $sections);
+                                if($sections == null){
+                                    $sections = [];
```
shailpatels: Same comment as above here

site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -15,6 +15,36 @@
             </p>
             <input type="text" id="input-provide-full-path" class="full-width" name="eg_config_path" value="" placeholder="optional/subdirectory"/>
         </label>
+        <label id="upload_sections">
+            <p>
+                <strong>Do you want to restrict this course material to some sections?</strong><br />
```
shailpatels: Get rid of the extra `<br />` here
You could also use <p class="black-message"> instead of a nested `<strong>` since you're just emphasizing all of the 
content under the paragraph tag.

site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -15,6 +15,36 @@
...
+            <label class = "radio">
+                <input id = "all_Sections_Showing" name="show_Some_Section_Selection" type = "radio" onchange="showDiv(0)" checked>
+                No
+            </label>
+            <label class = "radio">
+                <input name="show_Some_Section_Selection" type = "radio" onchange="showDiv(1)">
+                Yes
+            </label>
+            <script type="text/javascript">
+                function showDiv(val){
```
shailpatels: Lets rename this to something more descriptive like `toggleRegistrationSections` or something 
and you can use jquery's '$('#show-Some-Section-Selection`).toggle();` instead of the if else statement

site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -15,6 +15,36 @@
...
+                    } else{
+                        document.getElementById('show_Some_Section_Selection').style.display = "none";
+                    }
+                } 
+            </script>
+            <div id = "show_Some_Section_Selection" style="display: none">
+            {% for section in reg_sections %}
+                <tr>
+                    <td>Section {{ section['sections_registration_id'] }}</td>
+                    <td><input id="section_{{ section['sections_registration_id'] }}" type="checkbox" value="off"/></td>
```
shailpatels: HTML IDs should use hyphens instead of underscores 
https://submitty.org/developer/coding_style_guide/css#styling

site/app/views/course/CourseMaterialsView.php
```diff
@@ -15,14 +15,16 @@ class CourseMaterialsView extends AbstractView {
      * @param User[] $students
```
shailpatels: Update the `@param` comment to relfect the change from `$user_group` to `$user` and get rid of the `return` param, thats outdated

site/app/controllers/course/CourseMaterialsController.php
jchulton: If the json file contains an empty array as sections, the user selected to display only certain sections than didn't choose any sections. If the json file contains nothing as sections, then the user selected to display to all sections. These are two distinct cases that have to be treated differently.

site/app/controllers/admin/UsersController.php
jchulton: You don't think it would be bad to have references to sections that no longer exist in the json file? It seems like that could cause problems later on if we ever want to build off of this...

shailpatels: It would be better if course materials assumes a section could be deleted and handle it accordingly. If an instructor wanted to delet a section it would be more annoying than anything to go into course materials and delete all references.
Instead course materials should be fault tolerant and handle the case if a section does not exist gracefully.

shailpatels: Resolving based on meeting - didn't know this was the expected behavior

site/app/controllers/course/CourseMaterialsController.php
shailpatels: resolving based on meeting - didn't know this was expected behavior

site/app/controllers/course/CourseMaterialsController.php
shailpatels: resolving based on meeting - didn't know this was expected behavior

site/app/templates/course/UploadCourseMaterialsForm.twig
jchulton: Changed to <p class="black-message">

site/app/templates/course/UploadCourseMaterialsForm.twig
jchulton: Changed section_{{ section['sections_registration_id'] to section-{{ section['sections_registration_id']

site/app/templates/course/UploadCourseMaterialsForm.twig
jchulton: Renamed to toggleRegistrationSections and switched to .toggle

site/app/views/course/CourseMaterialsView.php
jchulton: changed as above

jchulton:
Is this all set Shail? I can't see a non-resolved requested change...

jchulton:
Any other changes Shail? I don't see any non-resolved comments. but it still says 1 change requested

shailpatels:
If everything is resolved I'll make sure to begin testing over the weekend :)

site/app/templates/course/CourseMaterials.twig
```diff
@@ -91,6 +92,16 @@
                <a onclick='newDeleteCourseMaterialForm("{{ path | url_encode }}", "{{ dir }}");'> <i class="fas fa-trash" aria-hidden="true" style="font-size: 16px; margin: 5px;"></i></a>
                 <label for="latest-events">Share to student on</label>
                 <input name="release_date" id="date_to_release_{{ id }}" class="date_picker" type="text" size="26"  value="{{ fileReleaseDates[path]}}"/>
+                {% if not(this_file_section is null) %}
+                <label>Sections: 
+                {% for key,value in this_file_section %}
```
shailpatels: You can use the twig `join` command instead of this loop 
https://twig.symfony.com/doc/2.x/filters/join.html

site/app/controllers/admin/UsersController.php
shailpatels: Remove this comment

site/app/controllers/admin/UsersController.php
```diff
@@ -411,7 +411,20 @@ public function updateRegistrationSections() {
...
+                $fp = '/var/local/submitty/courses/f19/sample/uploads/course_materials_file_data.json';
```
shailpatels: You should `$this->core->getConfig()->getCoursePath() . '/uploads/course_materials_file_data.json';` 
since the course path is something that could be configured.

site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -36,24 +62,43 @@
...
+                var reg_sectionArray = JSON.parse('{{ reg_sections|json_encode|raw }}');
+                
+                reg_sectionArray.forEach(function(element){
+                    var currentSection = '#section-' +element['sections_registration_id']
+                    if  ($(currentSection).is(':checked'))  {
+                        sections.push(element['sections_registration_id']);
+                    }
+                });
+                
+                /*for (section in reg_sectionArray){
```
shailpatels: Is this still needed?

jchulton:
The above errors should be fixed.

shailpatels:
Great! I'll rereview over the weekend :)

site/app/controllers/admin/UsersController.php
```diff
@@ -411,7 +411,19 @@ public function updateRegistrationSections() {
             if (User::validateUserData('registration_section', $_POST['delete_reg_section'])) {
                 // DELETE trigger function in master DB will catch integrity violation exceptions (such as FK violations when users/graders are still enrolled in section).
                 // $num_del_sections indicates how many DELETEs were performed.  0 DELETEs means either the section didn't exist or there are users still enrolled.
-                $num_del_sections = $this->core->getQueries()->deleteRegistrationSection($_POST['delete_reg_section']);
+                $this->core->getConfig()->getCoursePath() . '/uploads/course_materials_file_data.json';
```
shailpatels: I think you forgot the $fp = ... 
here, php is complaing that its undefined.

site/app/models/CourseMaterial.php
```diff
@@ -47,6 +47,53 @@ public static function isMaterialReleased(Core $core, string $path_to_file)
             // If current time is greater than release time return true, else return false
             $current_time > $release_time ? $retVal = true : $retVal = false;
 
+            return $retVal;
+        }http://submitty.org/
```
shailpatels: I think you pasted this here accidentally ?

shailpatels:
@jchulton  I think there were a couple of issues when you merging/fixing the last few issues its causing some of the php code to crash. Do you mind fixing that so I can go ahead and review the PR,
and if the changing requested above are finished could hit the resolve button when you've made the changes (it just helps see that you've made the fixes) :)

PR #4585, opened 2019-10-02
[Bugfix] Could not click upper half of items in footer

elihschiff:

When on screens with a width of over 541px (when on a computer/laptop) the top half of the footer would be covered by the "wrapper" div above it. This means the user could not click on the top half of the button/links in the footer. 
image
Note how the "wrapper" div covers up part of the footer and so you cannot click on the footer buttons where it overlaps.

### What is the new behavior?

You can now click on any part of the footer
image

### Other information?

I could not find a reason why this css was there in the first place so I just removed it but there may be a reason I did not see.

### DIFF
--- a/site/public/css/global.css
+++ b/site/public/css/global.css
@@ -137,8 +137,4 @@ footer .footer-separator {
         min-width: 450px;
         transform: translateX(-50%);
     }
-
-    footer {
-        margin-top: -20px; /* 30px content padding and 10px footer padding */
-    }
 }


elihschiff:
It looks like the tests are failing because it is running `git clone --branch footer_height_fix git://github.com/Submitty/Submitty.git .` however my branch is on my fork, not on `Submitty/Submitty`, `git://github.com/elihschiff/Submitty.git`

Is there a way I can fix the tests to pull from my fork instead of it trying to find my branch here and therefore failing?

MasterOdin:
Thanks for this. Yeah, I'm not sure either why the margin exists there, and removing it did not affect usage either for the wider screens or in the cellphone view. 

About the buildbot failures, do not worry about them. Travis is still our primary test infrastructure and requirement and this introduced a case why we cannot quite yet move away from it to our buildbot. I'm currently trying to figure out why it fails to build foreign PRs and am tracking the issue [here](https://github.com/Submitty/submitty-buildbot/issues/15).

PR #4607, opened 2019-10-06
[Testing:API] Create E2E test suite for API

MasterOdin:

### Description

Currently, the API is (lightly) tested via unit tests. While I plan to improve coverage of the test suite for the unit tests, we should, much like the regular site, also create a E2E for the API to ensure that it functions as we expect once all the pieces are put together.

I looked at trying to use [newman](https://github.com/postmanlabs/newman#) or [API Blueprint](https://apiblueprint.org/), but that would involve another language in our testing stack (NodeJS) and that they mostly just validate structure, and I'm not sure how we might test APIs under different users, incomplete parameters, etc. These also seem to fall apart when attempting to test endpoints that expect file uploads, which is something I definitely want to support and test, and that I think it would be easier to do this in a custom format and point to a file under `more_autograding_examples` or wherever under the repo.

Given this, I'd like to just propose a simple in-house API tester of specifying test suites per file using the format:
```
[
  {
    "name": "string",
    "endpoint": "/endpoint",
    "method": "get|post",
    "data": {},
    "response": {}
  }
]
```

There are also a handful (to be documented) replacements that can happen over the data as it shifts in time (e.g. replacing `<current_semester>` with `f19`.

Currently, this includes testing for all endpoints listed on https://api.submitty.org.

To-Do:
- [ ] Mock how testing would work for multiple users
- [ ] Write README in folder on writing tests as well as page for https://submitty.org
- [ ] Improve error message not including test name when test fails
- [ ] Better handle failing the get_token/invalidate_token steps run outside the test suite.

@prestoncarman, @zjxiaohan I am not sure your expertise in testing APIs, and would welcome comment on if you think there's a good tool to use instead of something in-house, etc.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -15,7 +15,7 @@ exclude=
     grading/python/submitty_router.py,
     grading/json_syntax_checker.py
     junit_test_runner,
-    migration/test,
+    migration/tests,
     migration/migrator/migrations,
     more_autograding_examples,
     ParseLogFiles,
--- a/.flake8
+++ b/.flake8
@@ -62,4 +62,6 @@ exclude=
     sbin/untrusted_canary.py,
     site,
     site_internal,
-    tests
+    tests/e2e,
+    tests/integrationTests,
+    tests/rainbowGrades
--- a/.setup/data/users/instructor2.yml
+++ b/.setup/data/users/instructor2.yml
@@ -1,0 +1,8 @@
+user_firstname: Quintin
+user_group: 1
+user_id: instructor2
+anon_id: instructor2
+user_lastname: Instructor2
+user_access_level: 2
+user_numeric_id: 123456789
+courses: []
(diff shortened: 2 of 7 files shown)


MasterOdin:
My initial thought on extending this for multiple users would be to modify the JSON schema to make `response` field a dictionary where each key is a user_id and the value is the expected response. Example:
```
[
  {
    "name": "name",
    "endpoint": "/endpoint",
    "method": "get",
    "response": {
      "student": {
        "status": "fail",
        "message": "You lack proper permissions for this endpoint"
      },
      "instructor": {
        "status": "success",
        "data": null
      }
  }
]
```

The test loader would then be in-charge of loading the tests, and only loading the response appropriate for the given user_id being tested.

zjxiaohan:
[Codeception](https://codeception.com/docs/10-APITesting) can be used for API testing in PHP, but to me the scheme proposed by Matt is good enough and more flexible. (I am not sure how to test multiple user groups using codeception)

PR #4618, opened 2019-10-08
[Documentation: README] : Punctuation

MayankMani00:
Added punctuation where required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,9 @@ The Submitty project is hosted on [GitHub](https://github.com/Submitty).
 # Key Features
 
 * Secure testing of many languages: Python, C/C++, Java, Scheme, Prolog, SPIM, and anything available on GNU / Linux!
-* Customizable automated grading with immediate feedback to students
+* Customizable automated grading with immediate feedback to students.
 * Advanced grading tools: static analysis, JUnit, code coverage, memory debuggers, etc.
-* Student upload by drag-and-drop, zip upload, or version control
+* Student upload by drag-and-drop, zip upload, or version control.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,9 +21,9 @@ The Submitty project is hosted on [GitHub](https://github.com/Submitty).
 # Key Features
 
 * Secure testing of many languages: Python, C/C++, Java, Scheme, Prolog, SPIM, and anything available on GNU / Linux!
-* Customizable automated grading with immediate feedback to students
+* Customizable automated grading with immediate feedback to students.
```
shwehtom89: Periods are not necessary in this case as this is not a complete sentence. I believe part of the confusion comes from ! mark in the first item and etc. being an abbreviation for etcetera. Listed items do not necessarily need periods.

README.md
```diff
@@ -21,9 +21,9 @@ The Submitty project is hosted on [GitHub](https://github.com/Submitty).
...
 * Advanced grading tools: static analysis, JUnit, code coverage, memory debuggers, etc.
-* Student upload by drag-and-drop, zip upload, or version control
+* Student upload by drag-and-drop, zip upload, or version control.
```
shwehtom89: Same as above.

README.md
```diff
@@ -21,9 +21,9 @@ The Submitty project is hosted on [GitHub](https://github.com/Submitty).
...
 * Correct mistakes through multiple submissions, flexible ``late day’’ policy, hidden tests
```
shwehtom89: These quotes do not match. Should be double quotes.

MasterOdin: Fixed in master, separate from this PR.

PR #4621, opened 2019-10-09
[Feature Developer] (experimental) Improved codewatcher

shailpatels:

This is a draft PR to show this as a proof of concept

Enables rsync in the vagrant file to make sure the mounted directory between the guest and host machine is synced in real time. The improved code watcher will then watch that shared directory and update the submission site as changes happen

### Other information?
Needs a lot more testing and also is incomplete
Needs a way to exclude files (twig cache mainly)

Requires the host machine to do an `vagrant rsync-auto` for the code watcher to see 
updates

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -81,7 +81,7 @@ Vagrant.configure(2) do |config|
   owner = 'root'
   group = 'vagrant'
   mount_options = %w(dmode=775 fmode=664)
-  config.vm.synced_folder '.', '/usr/local/submitty/GIT_CHECKOUT/Submitty', create: true, owner: owner, group: group, mount_options: mount_options
+  config.vm.synced_folder '.', '/usr/local/submitty/GIT_CHECKOUT/Submitty', create: true, owner: owner, group: group, mount_options: mount_options, type: "rsync", rsync__exclude: ".git/"
 
   optional_repos = %w(AnalysisTools Lichen RainbowGrades Tutorial CrashCourseCPPSyntax)
   optional_repos.each {|repo|
--- a/.setup/bin/watcher.py
+++ b/.setup/bin/watcher.py
@@ -1,0 +1,54 @@
+import traceback
+from pathlib import Path
+import sys
+try:
+    from watchdog.observers import Observer
+    from watchdog.events import FileSystemEventHandler
+except ImportError as err:
+    print("\nwatchdog python library not installed correctly\n")
+    traceback.print_exc()
+
+
+class EventHandler(FileSystemEventHandler):
+    def on_created(self, event):
+        #self.run_installer(event)
+        print(event.src_path)
+
+    def on_deleted(self, event):
+        #self.run_installer(event)
+        pass
+
+    def on_modified(self, event):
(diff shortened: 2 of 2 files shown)


Vagrantfile
```diff
@@ -81,7 +81,7 @@ Vagrant.configure(2) do |config|
   owner = 'root'
   group = 'vagrant'
   mount_options = %w(dmode=775 fmode=664)
-  config.vm.synced_folder '.', '/usr/local/submitty/GIT_CHECKOUT/Submitty', create: true, owner: owner, group: group, mount_options: mount_options
+  config.vm.synced_folder '.', '/usr/local/submitty/GIT_CHECKOUT/Submitty', create: true, owner: owner, group: group, mount_options: mount_options, type: "rsync", rsync__exclude: ".git/"
```
MasterOdin: Does `rsync` exist on Windows by default, and what is the advantage of setting this over the normal VirtualBox guest additions mounting?

shailpatels: I think `rsync` is on Windows although it might have to be added to windows path for vagrant to see it first. I've been reading this https://www.vagrantup.com/docs/synced-folders/rsync.html for info but haven't tested on windows directly yet.

When I was changing files on my host machine the updates weren't being detected 
by the python watchdog library, my guess is that the mirroring changes doesn't actually 
signal any change so r-sync fixes this but yeah I'll try and investigate a better way that part was kind of hacked together.

When I ssh'd and then edited a file `/usr/local/submitty/GIT_CHECKOUT/Submitty/...` 
it seemed to know the updates but changes from outside the vm weren't seen.

I'll look into this though.

MasterOdin: What is your OS and version of vagrant and VirtualBox?

shailpatels: I'm on Debian 9.9 and vagrant 2.2.4

MasterOdin: For a follow-up (as I think I know what's going on, but just want to double check some stuff first), why are you using the base Observer instead of the commented out PollingObserver?

shailpatels: Originally I was using the polling observer from the old code watcher but I changed to the base observer to see if it would fix that issue I mentioned above of not seeing file changes. 

From my understanding, the base observer picks a platform-specific observer that it uses to watch target directories. Whats the difference between that and polling observer?

MasterOdin: Importing just `Observer` will (at least here) will always get you the `InotifyObserver` which is definitely the "superior" observer when running natively on the system. Unfortunately, it has a limitation/feature that it will not record changes to files that the kernel does not "own", such as VirtualBox shared folders. The PollingObserver does not care about that distinction and so works a bit better out of the box with VirtualBox (as it polls the directory structure instead of using the inotify stuff).

shailpatels: Ah ok, thanks for the information, I'll switch back to polling observer then!

I'll flesh this out over the weekend - i'm away right now but I'll definitely ask for more feedback
once I make some more progress :)

shailpatels:
closing this PR in favor of changes to the installation scripts which should work with the current code watcher in a much faster time.

PR #4624, opened 2019-10-13
[Testing:GradeInquiry] Grade inquiry panel and box e2e tests

sak6lab:
No end to end tests to test the grade inquiry panel and box which have a lot of case dependent elements

### What is the new behavior?
Basic grade inquiry panel and box test that test the panel when there is normal submission, and no submission.

I will add more with future pull requests that fix bugs that I find while writing new tests and outstanding bugs.
### Testing
These test are bypassed for Travis as the Travis instance does not have submissions information
These tests will fail if lakinh or bauchg's submissions or grade inquiries for grades_released_homework in sample course are edited. These tests use the default configuration for users lakinh and bauchg.
see #4626

### DIFF
--- a/tests/e2e/test_grade_inquiry.py
+++ b/tests/e2e/test_grade_inquiry.py
@@ -1,4 +1,5 @@
 from .base_testcase import BaseTestCase
+import os
 class TestGradeInquiry(BaseTestCase):
     def __init__(self, testname):
         super().__init__(testname, log_in=False)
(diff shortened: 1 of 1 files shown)


tests/e2e/test_grade_inquiry.py
```diff
@@ -76,6 +83,9 @@ def test_no_submission_grade_inquiry_panel(self):
 
      # STUDENT SUBMISSION TESTS
     def test_normal_submission_student_grade_inquiry_box(self):
+        # travis should not run this
+        if "TRAVIS_BUILD_DIR" in os.environ and os.environ["TRAVIS_BUILD_DIR"] is not None:
+            return
```
MasterOdin: Mark these tests as properly [skipped](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) by adding annotation `@unittest.skipUnless(os.environ.get('TRAVIS_BUILD_DIR') is None, "cannot run in Travis-CI")` so that we have a visual indicator that is what's going on here.

PR #4632, opened 2019-10-16
[Bugfix:Email] Handle unmatched variables in exam seating email

### LINKED ISSUE #4570: Revamp allowed and handling of replacement variables for seating emails
Currently, when composing the seating email, an instructor is limited to following variables:
https://github.com/Submitty/Submitty/blob/1cc8081ee351038e74fbc6eb925238dd39951878/site/app/controllers/admin/EmailRoomSeatingController.php#L94-L103

It would be useful for personalization effects to have variables for a student (user_id, user_rin, user_display_first_name, user_display_last_name). This expanded list of allowed variables should be added to the UI somewhere (or a link to a page on submitty.org with details of it) so an instructor knows what they can use. 

We may also want to update the opening line for the email to be `Hello {$user_first_name} ({$user_id}),` as well.

Finally, if a value does not exist within the JSON field (from the above currently implemented), the variable is not replaced which can lead to funny looking emails like:
```
Hello,

Listed below is your seating assignment for the upcoming exam quiz01 on Wednesday, September 25th at 6:00 pm - 7:00 pm.

Location: Sage
Exam Room: 3303
Zone: SEE INSTRUCTOR
Row: {$exam_row}
Seat: {$exam_seat}

Please email your instructor with any questions or concerns.

--
NOTE: This is an automated email notification, which is unable to receive replies.
Please refer to the course syllabus for contact information for your teaching staff.
```

All variables should be replaced to something, regardless if we have all of the data (replacing the above with `SEE INSTRUCTOR` or just a blank value).

hdrodz97:

Fixes part of #4570 

In short, when sending out the exam seating assignment email, one can use a limited form of variable substitution to fill in with details of the exam or seating assignment. For example, the string `{$date}` will be replaced with the instructor-provided date for the exam. If a typo is made when typing one of these variables out, they will remain in the email, potentially leading to awkward scenarios where the email text says, for example, "The exam will be on {$day}".

### What is the new behavior?

When performing the variable substitution in the exam seating assignment email, the text is scanned after performing all recognized substitutions for patterns in the form of `{$VARNAME}`, where `VARNAME` is a string composed of upper and lowercase letters, digits, and underscores. These occurrences are most likely typos and are replaced with the text "SEE INSTRUCTOR".


site/app/controllers/admin/EmailRoomSeatingController.php
```diff
@@ -104,11 +104,12 @@ private function replacePlaceholders($message, $data) {
 
         foreach($replaces as $key => $variable) {
             if(isset($data[$key])) {
```
MasterOdin: You'll want to get rid of this if check around the `str_replace` as the `??` is now handling the case for when `!isset($data[$key])`.

PR #4639, opened 2019-10-19
[BugFix:GradeInquiry] Manually Graded Component Averages tab now show

sak6lab:
When there is no grade inquiries that are tied to a component the Manually Graded Component Averages tab does not shows.

### What is the new behavior?
Manually Graded Component Averages tab now shows.


bmcutler:
Thanks for the fix!

Issue #4638, opened 2019-10-17
Simple Grader does not validate backend actually saved something


MasterOdin:
If the payload gets messed up (such that we send no checkmarks) when hitting a checkmark on a lab or changing a numeric score, the frontend will report that the field was saved, with the backend not doing anything:

We should modify the SimpleGraderController such that it returns the actual components that were affected in its JSON response and that the simple-grading.js uses that to double check the clicked on element(s) were saved.

MasterOdin:
The relevant blocks of code are:
https://github.com/Submitty/Submitty/blob/master/site/public/js/simple-grading.js#L212-L235
https://github.com/Submitty/Submitty/blob/master/site/app/controllers/grading/SimpleGraderController.php#L192-L264

MasterOdin:
To be more precise about what needs to happen:

within the SimpleGraderController.php, we should create an array where keys are component_ids and values are new scores, and when we enter the foreach of components, we push onto that array any component whose score we update (i.e. we enter the if `if ($data !== '')` check). The relevant lines are:
https://github.com/Submitty/Submitty/blob/a4f071a7f806404a3bf4abaf2584f3e6c12bfa8b/site/app/controllers/grading/SimpleGraderController.php#L229-L233

and then return that array here:
https://github.com/Submitty/Submitty/blob/a4f071a7f806404a3bf4abaf2584f3e6c12bfa8b/site/app/controllers/grading/SimpleGraderController.php#L261-L263

and then on the success handler in the JS code here:
https://github.com/Submitty/Submitty/blob/master/site/public/js/simple-grading.js#L221-L227
we need to iterate through that array, and confirm that `new_scores` and this returned JSON object match.

PR #4644, opened 2019-10-22
Update simple-grading.js

kshitijandmojo:
Issue #4638   
 Could you please check if this fixes the issue?


MasterOdin:
It will not, as well as breaking different functionality (the border and data-score attributes breaking). I've added additional context in https://github.com/Submitty/Submitty/issues/4638#issuecomment-544936579 to what should happen.

PR #4651, opened 2019-10-25
[Feature:Autograding] Automatic STDOUT Timestamping

emaicus:
This PR introduces a new feature which allows instructors to specify assignments or testcases in which student ```STDOUT``` will be timestamped. Timestamped files are saved with the ```_TIMESTAMP.txt``` ending alongside regular ```STDOUT.txt``` files. 

To turn on timestamped standard output, an instructor can specify ```"timestamped_stdout" : true``` at the global or testcase level in their config.

__Note to reviewers:__
This PR required an amount of juggling of the grading code, so that seccomp filtering was applied after a new forked "timestamp process" was created. This makes it look like more code changed in ```execute.cpp``` than actually did.

__Known Issues:__
The method of timestamping used can only add a timestamp on a buffer flush. This means that programming languages that tend to buffer (e.g. ```Python```) don't play well unless run in unbuffered mode).


emaicus:
Tested locally. Still works.

PR #4652, opened 2019-10-25
[BugFix:BulkUpload] Fixed error handling for large files on Bulk Upload

shailpatels:

Inproper checking of uploaded files would lead to incorrect/vague error messages and
Files sizes larger than the POST or max_filesize_limit would causes a crash and get no response giving a vague error message.

### What is the new behavior?
Better error handling for file uploads by using `validateUploadedFiles` on the server side for large/incorrect file types. 
 The client now checks for this and gives a correct message before it attempts to upload.

### DIFF
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -244,9 +244,7 @@ class SubmissionController extends AbstractController {
             return $this->uploadResult("Empty POST request. This may mean that the sum size of your files are greater than {$max_size}.", false);
         }
 
-        if(isset($_POST['use_qr_codes'])){
-            $is_qr = $_POST['use_qr_codes'] === "true";
-        }
+        $is_qr = isset($_POST['use_qr_codes']) && $_POST['use_qr_codes'] === "true";
 
         if (!isset($_POST['num_pages']) && !$is_qr) {
             $msg = "Did not pass in number of pages or files were too large.";
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -239,7 +239,14 @@ class SubmissionController extends AbstractController {
      * @Route("/{_semester}/{_course}/gradeable/{gradeable_id}/bulk", methods={"POST"})
     */
     public function ajaxBulkUpload($gradeable_id) {
-        $is_qr = $_POST['use_qr_codes'] === "true";
+        if (empty($_POST)) {
+            $max_size =  Utils::returnBytes(ini_get('post_max_size'));
+            return $this->uploadResult("Empty POST request. This may mean that the sum size of your files are greater than {$max_size}.", false);
+        }
+
+        if(isset($_POST['use_qr_codes'])){
+            $is_qr = $_POST['use_qr_codes'] === "true";
+        }
 
         if (!isset($_POST['num_pages']) && !$is_qr) {
             $msg = "Did not pass in number of pages or files were too large.";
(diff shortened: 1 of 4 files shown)


site/app/controllers/student/SubmissionController.php
```diff
@@ -239,7 +239,14 @@ public function ajaxValidGradeable($gradeable_id) {
...
     public function ajaxBulkUpload($gradeable_id) {
-        $is_qr = $_POST['use_qr_codes'] === "true";
+        if (empty($_POST)) {
+            $max_size =  Utils::returnBytes(ini_get('post_max_size'));
+            return $this->uploadResult("Empty POST request. This may mean that the sum size of your files are greater than {$max_size}.", false);
+        }
+
+        if(isset($_POST['use_qr_codes'])){
+            $is_qr = $_POST['use_qr_codes'] === "true";
+        }
```
MasterOdin: If `!isset($_POST['use_qr_codes'])`, this causes `$is_qr` to be undefined for the check on line 351, which is a code smell. You should make sure all variables are defined when using them.

Also, this can be simplified into a one-liner:
`$is_qr = isset($_POST['use_qr_codes']) && $_POST['use_qr_codes'] === "true";`

site/app/views/submission/HomeworkView.php
```diff
@@ -407,7 +410,9 @@ private function renderSubmitBox(Gradeable $gradeable, $graded_gradeable, $versi
             'upload_message' => $this->core->getConfig()->getUploadMessage(),
             "csrf_token" => $this->core->getCsrfToken(),
             'has_overridden_grades' => $graded_gradeable ? $graded_gradeable->hasOverriddenGrades() : false,
-            'days_to_be_charged' => $days_to_be_charged
+            'days_to_be_charged' => $days_to_be_charged,
+            'max_file_size' => $max_file_size,
+            'max_post_size' => $max_post_size
```
MasterOdin: Just do the calls to `Utils` here inside the array definition, no reason to set a variable for this one usage.

site/public/js/drag-and-drop.js
```diff
@@ -715,7 +731,7 @@ function handleBulk(gradeable_id, num_pages, use_qr_codes = false, qr_prefix = "
             catch (e) {
                 alert("Error parsing response from server. Please copy the contents of your Javascript Console and " +
                     "send it to an administrator, as well as what you were doing and what files you were uploading.");
-                console.log(data);
+                console.log(e);
```
MasterOdin: Please revert this change. The exception here (and other places we wrap `JSON.parse`) will not really give any meaningful information, most likely just being `SyntaxError: Unexpected token a in JSON at position 0`, whereas `data` will be the frog robot or other output which will have buried in it some idea of where to start looking.

site/public/js/drag-and-drop.js
```diff
@@ -683,6 +684,21 @@ function handleBulk(gradeable_id, num_pages, use_qr_codes = false, qr_prefix = "
...
+            if (total_size >= max_file_size){
+                alert("ERROR! Uploaded file(s) exceed max file size.\n" + 
+                      "Please visit https://submitty.org/sysadmin/system_customization for configuration instructions.");
+                return;
+            }
+
+             if (total_size >= max_post_size){
+                alert("ERROR! Uploaded file(s) exceed max PHP POST size.\n" + 
+                      "Please visit https://submitty.org/sysadmin/system_customization for configuration instructions.");
+                return;
```
MasterOdin: Go through all of the error conditions in this loop and make sure that they have `$("#submit").prop("disabled", false);` before calling `return` in them so that the user is not stuck with a dead form.

PR #4656, opened 2019-10-26
[Refactor: Queue] Updated ui based on user feedback and added more clear status codes

elihschiff:

The colors and functions of the buttons on many pages don't make sense
Sorting of users that were mass removed from the queue was random
Any removal of a user from the queue would be marked by a '3' as its status code

### What is the new behavior?
Button colors have been updated to follow the other pages on submitty (blue click, white maybe click, red don't click)
Users who were mass removed with the "Empty Queue" button from the queue now sort by the time they joined the queue
Split up the buttons on the instructor/TA/mentor page to now all do exactly one thing
(Close Queue, Generate New Code, Empty Queue)
Student page now has labels to follow accessibility standards
Removing a user from the queue not has 3 new status codes, code 3 for if the user was removed by themselves, code 4 for if a mentor removed them, code 5 for if they were removed as part of a bulk empty of the queue
Added a timer to the mentor page so that mentors can keep track of how long they have been helping a student.

image

image

### Other information?

We plan on working on some larger improvements to the queue but we feel theses changes are important enough to make a pull request now before we start on any larger update with stuff like analytics for example.

Especially the button color changes are important to push as many students and mentors were getting confused and clicking the wrong buttons.


elihschiff:
I agree, adding constants for the status codes is a good idea. I added all of the current status codes to OfficeHoursQueueInstructor.php and added comments to explain what each status code actually means.

elihschiff:
I added a few more small changes to this pull request that Barb asked for.

The columns that instructors see are now in a different order to make it easier to read.
The close queue button is now blue (white before)
image
The time spent in the queue and being helped is now in the format of hour min sec instead of the exact time of day. So it now reads "0h 24m 3s" instead of "5:43 pm" and then you have to do the math to see how long they have been in the queue
The messages for when a student is removed the queue is now also different depending on who removed them

PR #4666, opened 2019-10-26
[Feature:UI] Adding the Submitty Duck Mascot

shailpatels:

Submitty does not have a duck!

### What is the new behavior?
Adds a duck a mascot and a special banner!


migration/migrator/migrations/system/20191026164637_ducky.py
```diff
@@ -0,0 +1,33 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+
+    tgt = str(Path(config.submitty['submitty_install_dir'], 'config', 'submitty.json'))
+
+    with open(tgt, 'r') as infile:
+        data = json.load(infile)
+
+    data['enable_banner'] = False
```
elihschiff: Given this is hidden in a config file for all of submitty I feel like "enable_banner" is not very clear. I think it should be something like "enable_duck_banner" or something that says exactly what banner this is.

Also this is making a change for a migration but I assume there is a default config somewhere that you should also add this to so when somebody makes a new install it is in the default config file.

shailpatels:
@elihschiff Fixed the jumping and tab index.
The banner creating a scrollwheel will be fixed later when we finalized duck special effects for next submitty week.
Updating submitty.org should be a seperate PR .

(Mobile needs to be updated - I need to talk with the person who made the mobile view first fort that)

PR #4726, opened 2019-11-01
[Feature:Submissions] ValidateUploadedFiles now handles zip files

shailpatels:

The method to handle file uploads in PHP only handled normal file uploads, ziparchives
would only have the archive file size read not the filesize of the contents within it.

### What is the new behavior?
ValidateUploadedFiles now checks within zip files for size and names and can be used instead of current method of validating files across submitty

### Other information?
The use of validateUploadedFiles isn't used everywhere and there is still some refactoring needed to get it into every place where files are submitted.
This update will also require refactoring methods of file uploads for zips to get rid of the disgusting triple for loop used for file submission every where else.


site/app/libraries/FileUtils.php
```diff
@@ -480,6 +480,43 @@ public static function areWordsInFile(string $file, array $words) {
...
+                    $err = "Internel Error";
+                    break;
+                case \ZipArchive::ER_INCONS:
+                    $err = "Zip archive inconsistent";
+                    break;
+                case \ZipArchive::ER_CRC:
+                    $err = "Failed to check file integrity (CRC fail)";
+                    break;
+                default:
+                    $err = "Unknown error " . $res;
```
MasterOdin: This kind of switch should live in `ErrorMessages`.

site/app/libraries/FileUtils.php
```diff
@@ -480,6 +480,43 @@ public static function areWordsInFile(string $file, array $words) {
         return $words_detected;
     }
 
+    public static function checkZipFileStatus($file){
+        $zip = new \ZipArchive();
+        //open file with additional checks
+        $res = $zip->open($file, \ZipArchive::CHECKCONS);
```
MasterOdin: let's just return here with the status code, and see below note about splitting the printing of error messages into ErrorMessages

site/app/libraries/FileUtils.php
```diff
@@ -519,14 +568,19 @@ public static function validateUploadedFiles($files) {
                 $err_msg = "Invalid filename";
             }
 
-            $success = $err_msg === "No error.";
+            //if zip file check files inside
+            if($is_zip && !FileUtils::checkFileInZipName($tmp_name) ){
+                $err_msg = "Invalid filename within zip file"; 
+            }
 
+            $success = $err_msg === "No error.";
```
MasterOdin: Let's inline this check into the `$ret[]` definition. No need in creating this temporary variable.

site/app/libraries/FileUtils.php
```diff
@@ -501,13 +538,25 @@ public static function validateUploadedFiles($files) {
...
-            $type = mime_content_type($files['tmp_name'][$i]);
-            $size = $files['size'][$i];
-            $err_msg = "";
-
-            //did anything go wrong?
+            $tmp_name = $files['tmp_name'][$i];
+            $type = mime_content_type($tmp_name);
+            
+            $zip_status = FileUtils::checkZipFileStatus($tmp_name);
             $err_msg = ErrorMessages::uploadErrors($files['error'][$i]);
```
MasterOdin: If `$files['error'][$i]  !== UPLOAD_ERR_OK`, there's no point in bothering with trying to check zip file status.

site/app/libraries/FileUtils.php
```diff
@@ -501,13 +538,25 @@ public static function validateUploadedFiles($files) {
...
             $name = $files['name'][$i];
-            $type = mime_content_type($files['tmp_name'][$i]);
-            $size = $files['size'][$i];
-            $err_msg = "";
-
-            //did anything go wrong?
+            $tmp_name = $files['tmp_name'][$i];
+            $type = mime_content_type($tmp_name);
+            
+            $zip_status = FileUtils::checkZipFileStatus($tmp_name);
```
MasterOdin: can `\ZipArchive` handle anything beyond `application/zip` content-types (which we get from the `$type` variable right above this)? I don't believe so, and so we don't really need to waste time trying to do anything with zips if `$type !== 'application/zip'`.

I'd probably rewrite things as:
```
$is_zip = $type === 'application/zip';
if ($is_zip) {
    $zip_status = FileUtils::getZipFileStatus($tmp_name);
    if ($zip_status !== \ZipArchive::ER_OK) {
        $err_msg = ErrorMessages::getZipErrorMessage($zip_status);
    }
    else {
        $size = FileUtils::getZipSize($tmp_name);
        if (!FileUtils::checkFileInZipName($tmp_name)) {
            $err_msg = "Invalid filename within zip file"; 
        }
    }
}
```

site/app/libraries/FileUtils.php
```diff
@@ -519,14 +568,19 @@ public static function validateUploadedFiles($files) {
...
             $ret[] = [
                 'name' => $name,
                 'type' => $type,
                 'error' => $err_msg,
                 'size' => $size,
-                'success' => $success
+                'success' => $success,
+                'is_zip' => $is_zip
             ];
         }
```
MasterOdin: I should have caught this in my initial review, sorry for not.

This function should be refactored so that we either make it clearer what the final value of `$err_msg` is going to be or collect all possible errors in an array and return them all. Personally, I'd vote for the latter just so the uploader can fix them all in one fell swoop instead of upload, see a new error, upload, see a new error, etc. until they finally after several rounds fix all errors.

site/app/libraries/FileUtils.php
```diff
@@ -480,6 +480,43 @@ public static function areWordsInFile(string $file, array $words) {
         return $words_detected;
     }
 
+    public static function checkZipFileStatus($file){
```
MasterOdin: Let's rename this to `getZipFileStatus`. `check` (to me) implies either the function is going to throw an exception or return true/false.

MasterOdin:
@shailpatels I resolved the merge conflict and style errors that existed in the branch as a result of my merged stuff.

shailpatels:
@MasterOdin Thanks for the merge conflict fixes,
I made the requested changes, the 'error' key is now every error message in a single string instead of just the latest one now. 
The switch statement for zip errors is moved over to the errormessages file and I added your method for checking the zip file.

site/app/libraries/FileUtils.php
```diff
@@ -502,32 +514,62 @@ public static function validateUploadedFiles($files) {
...
             //extract the values from each file
             $name = $files['name'][$i];
-            $type = mime_content_type($files['tmp_name'][$i]);
-            $size = $files['size'][$i];
-            $err_msg = "";
+            $tmp_name = $files['tmp_name'][$i];
+            $type = mime_content_type($tmp_name);
+            
+            $zip_status = FileUtils::getZipFileStatus($tmp_name);
+            $errors = [ErrorMessages::uploadErrors($files['error'][$i])];
```
MasterOdin: Let's make this:
```
$errors = [];
if ($files['error'][$i] !== UPLOAD_ERR_OK) {
    $errors[] = ErrorMessages::uploadErrors($files['error'][$i]);
}
```
Let's only append to `$errors` if there's an actual error message.

site/app/libraries/FileUtils.php
```diff
@@ -502,32 +514,62 @@ public static function validateUploadedFiles($files) {
...
+                $errors[] = "File \"" . $name . "\" too large got (" . Utils::formatBytes("mb", $size) . ")";
             }
 
             //check filename
             if (!FileUtils::isValidFileName($name)) {
-                $err_msg = "Invalid filename";
+                $errors[] = "Invalid filename";
+            }
+
+            $errors = array_unique($errors);
```
MasterOdin: What errors might be duplicated?

site/app/libraries/FileUtils.php
```diff
@@ -502,32 +514,62 @@ public static function validateUploadedFiles($files) {
...
+            $success = true;
+            foreach ($errors as $err) {
+                if ($err !== 'No error.') {
+                    $success = false;
+                }
+            }
```
MasterOdin: Success should be that `$success = count($errors) > 0`. Having `'No error.'` be an entry just begs to be confusing to someone.

site/app/libraries/FileUtils.php
```diff
@@ -502,32 +514,62 @@ public static function validateUploadedFiles($files) {
...
+
+            //something went wrong, remove the 'No error. msg'
+            if (!$success && ($key = array_search('No error.', $errors)) !== false) {
+                unset($errors[$key]);
             }
```
MasterOdin: Same as above, I'd like to see this whole check removed as `$errors` should be empty if there has been no error.

site/app/libraries/FileUtils.php
```diff
@@ -502,32 +514,62 @@ public static function validateUploadedFiles($files) {
...
 
-            $success = $err_msg === "No error.";
 
             $ret[] = [
                 'name' => $name,
                 'type' => $type,
-                'error' => $err_msg,
+                'error' => implode(" ", $errors),
```
MasterOdin: In the case that this were to say `No error.` and `$success === true`, is any place actually uses that string, or can this just be an empty string without affecting anything (beyond the testcases).

shailpatels:
@MasterOdin Alright, sounds good I'll make the requested changes over break . The array unique for error messages is actually the success message which I believe multiple checks return so at one point $errors contained 'no error' message multiple times, changing it so it only contains error messages should fix this though however.

elihschiff:
This page still has issues with large zip files, I can send the the sample zip file I used if you want
http://192.168.56.111/f19/sample/autograding_config

shailpatels:
> This page still has issues with large zip files, I can send the the sample zip file I used if you want
> http://192.168.56.111/f19/sample/autograding_config

This PR this only creates the library function to handle zip files similar to validateUploadedFiles 
and additional PR needs to be done to refactor the controllers which should be done after this PR is merged.

elihschiff:
> > This page still has issues with large zip files, I can send the the sample zip file I used if you want
> > http://192.168.56.111/f19/sample/autograding_config
> 
> This PR this only creates the library function to handle zip files similar to validateUploadedFiles
> an additional PR needs to be done to refactor the controllers to use this function which should be done after this PR is merged.

Oh ok. I will approve it then.

PR #4731, opened 2019-11-02
[Testing:Linter] Fix PHP linter for #4726

shailpatels:

* tried to fix it as best i could , should be merged after #4726

### DIFF
--- a/site/app/libraries/FileUtils.php
+++ b/site/app/libraries/FileUtils.php
@@ -1,6 +1,7 @@
 <?php
 
 namespace app\libraries;
+
 use app\exceptions\FileReadException;
 
 /**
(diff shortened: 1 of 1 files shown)


shailpatels:
i dont know whats happening anymore

MasterOdin:
Use `vendor/bin/phpcbf` to fix lint problems, easier than trying to fix it by hand.

shailpatels:
Does that script put brackets on new lines, for some reason it would change function 
from ```foo(...){``` to
 ```
foo(...)
{
```
which would then be an error

**Comment by MasterOdin** [comment]
You need to use it in the same way you use `phpcs`:
```
vendor/bin/phpcbf --standard=tests/ruleset.xml
```

If you do not pass in the standard, it will default to something that mandates that above style (which then classes with our standard).

shailpatels:
ohh ok thats what i was doing wrong

PR #4733, opened 2019-11-04
[Refactor:PHP] Fix PSR2.Methods.FunctionCallSignature style errors

MasterOdin:

### What is the new behavior?

Fixes `PSR2.Methods.FunctionCallSignature` style errors, which is largely that there should not be a space around the beginning or end of parameters, parameters should not be split between multiple lines unless its one argument per line.

### Other information?

This makes some of the queries in DatabaseQueries (and other Query files) look a bit funny, but I will address fixing them up after more style errors are fixed as else it's going to cause some amount of repeat effort in styling them that I would like to avoid.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -2047,8 +2047,7 @@ VALUES(?, ?, ?, ?, 0, 0, 0, 0, ?)", array($g_id, $user_id, $team_id, $version, $
      */
     public function getUsersSeekingTeamByGradeableId($g_id) {
         $this->course_db->query(
-            "
-          SELECT user_id
+            "SELECT user_id
           FROM seeking_team
           WHERE g_id=?
           ORDER BY user_id",
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -1412,7 +1412,8 @@ ORDER BY rotating_section");
 
     public function getGradersByUserType() {
         $this->course_db->query(
-            "SELECT user_firstname, user_lastname, user_id, user_group FROM users WHERE user_group < 4 ORDER BY user_group, user_id ASC");
+            "SELECT user_firstname, user_lastname, user_id, user_group FROM users WHERE user_group < 4 ORDER BY user_group, user_id ASC"
+        );
         $users = [];
 
         foreach ($this->course_db->rows() as $row) {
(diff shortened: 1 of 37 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1650,14 +1659,12 @@ public function getAllSectionsForGradeable($gradeable) {
                 SELECT * FROM sections_registration
                 ORDER BY SUBSTRING(sections_registration_id, '^[^0-9]*'),
                 COALESCE(SUBSTRING(sections_registration_id, '[0-9]+')::INT, -1),
-                SUBSTRING(sections_registration_id, '[^0-9]*$')"
-            );
+                SUBSTRING(sections_registration_id, '[^0-9]*$')");
```
prestoncarman: It seems this closing parenthesis should be on the next line, but the string starts with a newline. Would it make more sense to take out the new line and let the parenthesis be on a line by itself?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1650,14 +1659,12 @@ public function getAllSectionsForGradeable($gradeable) {
...
         }
         else {
             $this->course_db->query("
                 SELECT * FROM sections_rotating
-                ORDER BY sections_rotating_id"
-            );
+                ORDER BY sections_rotating_id");
```
prestoncarman: Same as above.

prestoncarman:
Side question: Can PHPStorm format PHP files to the standard set up by Submitty? I know in other project I have worked on they provide a format document that is loaded into the IDE and then it will format to the project specification.

site/app/libraries/database/DatabaseQueries.php
MasterOdin: Sort of. I think the best course of action would be to put the query (and all long multiline queries) into a variable above the function call and define it as a heredoc instead of a multiline string. This would I think give the best combination of readability and formatting by phpcs.

MasterOdin:
It can, though I can only point people to https://www.jetbrains.com/help/phpstorm/using-php-code-sniffer.html for help. When I'm back at RPI in a few weeks, I plan to also write up some documentation on setting up automated git pre-commit holds

site/app/libraries/database/DatabaseQueries.php
MasterOdin: For a broader point, We need to codify exactly how we wish to see long queries written into a style document and then have it be something we properly follow in code-reviews, which defines how each block of a SQL query should be indented compared to the previous line. 

I will open an issue to continue that there, and then after we get consensus on agreed style, we can go through the entire file and reformat all queries as appropriate.

PR #4741, opened 2019-11-06
[Refactor:PHP] Fix Squiz.Scope.MethodScope.Missing style errors

MasterOdin:

### What is the new behavior?

Fixes `Squiz.Scope.MethodScope.Missing` style errors, which is that all methods must have the visibility for it defined (instead of just relying on the default public).

### DIFF
--- a/site/tests/ruleset.xml
+++ b/site/tests/ruleset.xml
@@ -48,7 +48,6 @@
         <exclude name="Squiz.PHP.NonExecutableCode.Unreachable" />
         <exclude name="Squiz.PHP.NonExecutableCode.ReturnNotRequired" />
         <exclude name="Squiz.PHP.NonExecutableCode.ReturnNotRequired" />
-        <exclude name="Squiz.Scope.MethodScope.Missing" />
         <exclude name="Squiz.WhiteSpace.ScopeClosingBrace.ContentBefore" />
         <exclude name="Squiz.WhiteSpace.ScopeClosingBrace.Indent" />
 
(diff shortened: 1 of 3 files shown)


site/tests/ruleset.xml
```diff
@@ -48,7 +48,6 @@
         <exclude name="Squiz.PHP.NonExecutableCode.Unreachable" />
         <exclude name="Squiz.PHP.NonExecutableCode.ReturnNotRequired" />
         <exclude name="Squiz.PHP.NonExecutableCode.ReturnNotRequired" />
-        <exclude name="Squiz.Scope.MethodScope.Missing" />
```
zjxiaohan: Why is this style removed from the rule set?

MasterOdin: It's not being removed, it's being set to not be excluded anymore as I've fixed the style errors it was reporting.

zjxiaohan: Oh lol sorry I understood it in a weird way and thought "exclude" was like meaning "preventing `Squiz.Scope.MethodScope.Missing` from happening".

So the style checks in `ruleset.xml` are full of code that we want to get rid of in the future?

MasterOdin: Yeah, and that there is at least one violation of that rule in the codebase. Given the very large number of violations over a large number of rules, I decided to ignore them all, and then go one at a time through them so that:
1. the diffs are focused around a central concept making it easier to review
2. can better understand the impact of a given rule, and decide perhaps if we want to or not follow it
3. better deal with potential of fixes for a style creating a conflict with a PR

zjxiaohan: Thanks a lot for the explanation. :)

PR #4745, opened 2019-11-06
[Refactor:Autograding] Have root shipper process assign autograding jobs to forked shippers

hdrodz97:

The forked shipper processes all lock the `/var/submitty/data/to_be_graded/` directory looking for new jobs to take on. This means only one shipper can check for new jobs at a time and every other shipper must wait for this shipper to take the job before they can check for their own new jobs.

### What is the new behavior?

The forked shipper processes have their own subfolder of `/var/submitty/data/grading/` (a newly-introduced directory) which they monitor as before. This allows them to check for jobs in parallel and without needing to lock resources. The master shipper process continuously monitors `/var/submitty/data/to_be_graded/` for new jobs. If there are any new jobs, and there are any idle workers, then the master shipper process will randomly assign jobs to any idle workers.

### Other information?

The current scheduling behavior is a placeholder and could be improved in future iterations to do more intelligent scheduling.


migration/migrator/migrations/system/20191027013443_separate_shipper_dirs.py
```diff
@@ -0,0 +1,11 @@
...
+import shutil
+
+IN_PROGRESS_PATH = '/var/local/submitty/grading'
+
+
+def up(config):
+    if not os.path.exists(IN_PROGRESS_PATH):
+        os.mkdir(IN_PROGRESS_PATH)
+    shutil.chown(IN_PROGRESS_PATH, 'submitty_daemon', 'submitty_daemonphp')
+    os.chmod(IN_PROGRESS_PATH, 0o770)
```
emaicus: Even if you aren't going to implement a down migration, add one which just passes (does nothing).

sbin/submitty_autograding_shipper.py
```diff
@@ -746,30 +746,31 @@ def shipper_process(my_name,my_data,full_address,which_untrusted,overall_lock):
...
             if not my_job == "":
                 counter=0
-                grade_queue_file(my_name,which_machine,which_untrusted,os.path.join(INTERACTIVE_QUEUE,my_job))
+                grade_queue_file(my_name,which_machine,which_untrusted,os.path.join(my_folder, my_job))
                 continue
             else:
                 if counter == 0 or counter >= 10:
-                    print ("{0} {1}: no available job".format(my_name, which_untrusted))
+                    autograding_utils.log_message(AUTOGRADING_LOG_PATH, JOB_ID,
+                        message="{0} {1}: no available job".format(my_name, which_untrusted))
```
emaicus: This may clutter the logs on off hours. @bmcutler , what are your thoughts?

emaicus: Missed the line above. This would print to the log once every ten iterations.

andrewaikens87: Are we going to continue to use `string.format`? Or use `f-Strings` for python string formatting moving forward: https://realpython.com/python-string-formatting/#3-string-interpolation-f-strings-python-36

emaicus: Yes, we should be moving to ```f-strings```. Our current policy is to replace instances of ```.format()``` as we encounter them. @hdrodz97 , feel free to clean this one up.

emaicus:
Re-reviewed. Needs to have merge conflicts resolved, but grading appears to work as expected. @bmcutler pointed out that this needs to be integrated with the ```grading_done``` script.

autograder/autograder/autograding_utils.py
```diff
@@ -94,6 +94,8 @@ def log_stack_trace(log_path, job_id="UNKNOWN", is_batch=False, which_untrusted=
 
     now = dateutils.get_current_time()
     datefile = "{0}.txt".format(datetime.strftime(now, "%Y%m%d"))
+    if not os.path.exists(log_path):
+        os.mkdir(log_path)
```
MasterOdin: `os.mkdir(log_path, exist_ok=True)`

bin/grading_done.py
```diff
@@ -175,6 +180,27 @@ def print_status(epoch_time,num_shippers,num_workers,
...
 
+class QueueItem:
+    def __init__(self, json_file, epoch_time, is_grading=False):
+        # If this is for a queue item currently grading; then we ensure the 
+        # provided JSON file begins with 'GRADING_', and we get rid of the
+        # 'GRADING_' prefix for the regular queue file, while also reading
+        # the 'GRADING_' file.
+        if is_grading:
+            base, tail = os.path.split(json_file)
+            assert(tail[:8] == 'GRADING_')
```
MasterOdin: `assert(tail.startswith('GRADING_'))`

bin/grading_done.py
```diff
@@ -175,6 +180,27 @@ def print_status(epoch_time,num_shippers,num_workers,
...
+            grading_json_file = json_file
+            json_file = os.path.join(base, tail[8:])
+
+            with open(grading_json_file, 'r') as infile:
+                self.grading_queue_obj = json.load(infile)
```
MasterOdin: Let's re-arrange this. No point making a temp variable.

```
with open(json_file) as infile:
    self.grading_queue_obj = json.load(infile)
json_file = os.path.join(base, tail[8:])
```

migration/migrator/migrations/system/20191027013443_separate_shipper_dirs.py
```diff
@@ -0,0 +1,15 @@
+import os
+import shutil
+
+IN_PROGRESS_PATH = '/var/local/submitty/grading'
+
+
+def up(config):
+    if not os.path.exists(IN_PROGRESS_PATH):
+        os.mkdir(IN_PROGRESS_PATH)
```
MasterOdin: ```os.mkdir(IN_PROGRESS_PATH, exist_ok=True)```

migration/migrator/migrations/system/20191027013443_separate_shipper_dirs.py
```diff
@@ -0,0 +1,15 @@
+import os
+import shutil
+
+IN_PROGRESS_PATH = '/var/local/submitty/grading'
```
MasterOdin: This should be constructed in the `up` method and using the `config.submitty['submitty_data_dir']` variable

sbin/submitty_autograding_shipper.py
```diff
@@ -843,6 +835,17 @@ def launch_shippers(worker_status_map):
     total_num_workers = 0
     processes = list()
     for name, machine in autograding_workers.items():
+        thread_count = machine["num_autograding_workers"]
+        
+        # Cleanup previous in-progress submissions
+        worker_folders = [worker_folder(f'{name}_{i}') for i in range(thread_count)]
+        for folder in worker_folders:
+            if not os.path.exists(folder):
+                os.mkdir(folder)
```
MasterOdin: `os.mkdir(folder, exist_ok=True)`

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -293,6 +294,8 @@ if [ "${WORKER}" == 0 ]; then
     chmod  770                                      $SUBMITTY_DATA_DIR/to_be_graded_queue
     chown  ${DAEMON_USER}:${DAEMONPHP_GROUP}        $SUBMITTY_DATA_DIR/daemon_job_queue
     chmod  770                                      $SUBMITTY_DATA_DIR/daemon_job_queue
+    chown  ${DAEMON_USER}:${DAEMONPHP_GROUP}        $SUBMITTY_DATA_DIR/grading
+    chmod  770                                      $SUBMITTY_DATA_DIR/grading
```
MasterOdin, line 298: If `submitty_php` is only reading out of this directory, and never writing anything into it, let's make the permission `750`.

migration/migrator/migrations/system/20191027013443_separate_shipper_dirs.py
```diff
@@ -0,0 +1,14 @@
+import os
+import shutil
+
+
+def up(config):
+    IN_PROGRESS_PATH = os.path.join(config.submitty['submitty_data_dir'], 'grading')
+
+    os.makedirs(IN_PROGRESS_PATH, exist_ok=True)
+    shutil.chown(IN_PROGRESS_PATH, 'submitty_daemon', 'submitty_daemonphp')
+    os.chmod(IN_PROGRESS_PATH, 0o770)
```
MasterOdin, line 10: This directory is made as part of the regular `SUBMITTY_INSTALL_HELPER.sh` script, so we do not need a migration file for it.

site/tests/app/libraries/GradingQueueTester.php
```diff
@@ -16,6 +16,7 @@
 class GradingQueueTester extends BaseUnitTest {
```
MasterOdin, line 16: either expand the existing tests or add some new ones to verify the changed behavior in `GradingQueue` model.

PR #4762, opened 2019-11-11
[Refactor:GradeSummary] Greatly reduce number of run queries for csv generation

### LINKED ISSUE #4755: Generate grade summaries is very inefficient
Running the generate grade summary functions is very inefficient as it runs a couple queries once per user of a course instead of doing any sort of batching. For example, for the `sample` course, with 25 gradeables and 139 users, generating a 11kb CSV runs 7093 queries. Running the code for CS1 with >700 students is not clear how long it'll take (easily well over 30 minutes according to CS1 instructor).

MasterOdin:

Fixes #4755 

Generating the CSV grade report is very inefficiently, currently running over 7,000 queries for the sample course in Vagrant. For real courses at RPI, this count is most likely going over 30,000 queries, which causes the page to take a very long time and generally just time out (with its 700+ students).

### What is the new behavior?

Fixes the slowdown, reducing the above query count to a constant 6 (or so) queries for the basic use-case. This is achieved by getting the following information in one batch instead of one at a time per user per gradeable.
1. Get all late day updates for all students
2. Get if a grade override exists for all gradeables for all students

This saved information is then used throughout as needed without requiring any subsequent queries for more information. The downside is that this does require additional memory throughout the entire process, however the upside is that this only requires a set number of queries used to run this page.

### Other information?

This is marked as a draft pull request until I am able to complete my assessment of memory overread through this method, as well as perform additional testing around grade overrides with team assignments.

### DIFF
--- a/site/app/controllers/admin/ReportController.php
+++ b/site/app/controllers/admin/ReportController.php
@@ -3,12 +3,8 @@
 namespace app\controllers\admin;
 
 use app\controllers\AbstractController;
-use app\exceptions\FileReadException;
-use app\libraries\Core;
-use app\libraries\DateUtils;
 use app\libraries\FileUtils;
 use app\libraries\GradeableType;
-use app\libraries\Output;
 use app\libraries\routers\AccessControl;
 use app\models\gradeable\AutoGradedGradeable;
 use app\models\gradeable\Gradeable;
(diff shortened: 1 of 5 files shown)


MasterOdin:
Each new entry of late day table adds about ~1300 bytes used, which for 700 student class adds about ~1MB for all students having one row.

Looks like grade overrides is very similar in size used.

While not necessarily great, it's not exactly breaking the bank here so I can live with this.

PR #4781, opened 2019-11-14
[Bugfix: Autograding] Move Stack Trace Logs

emaicus:
1. Adds a migration which moves the stack trace logs from ```var/local/submitty/logs/autograding/stack_traces``` to ```var/local/submitty/logs/stack_traces```.
2. Appropriate down migration also added.
3. New stack trace log names are no longer prepended with ```stack_trace_```. 
4.  ~~On failure to prepare a job, the autograding shipper now waits 5 seconds before trying again instead of instantly retrying. This should help to slow down log build-up in cases of failed grading. Future revisions should be made to further fix this error.~~  ___Moving to another PR___

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -213,7 +213,7 @@ fi
 # Create the logs directories that exist on both primary & worker machines
 mkdir -p ${SUBMITTY_DATA_DIR}/logs
 mkdir -p ${SUBMITTY_DATA_DIR}/logs/autograding
-mkdir -p ${SUBMITTY_DATA_DIR}/logs/autograding/stack_traces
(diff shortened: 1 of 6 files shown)


MasterOdin:
If we're going to move the autograding stack traces folder to be a top level log folder, but still only be for autograding stack traces, can we prepend `autograding_` to that then? Just naming it `stack_traces` and only for autograding does not seem like it'll be obvious to newcomers.

MasterOdin:
Also, even if it's just a one line change in the same file/subsystem, we should always be making separate PRs for completely unrelated bugfixes/features, so that we can actually roll things back in a more granular way.

emaicus:
Changed ```stack_traces``` to ```autograding_stack_traces``` per @MasterOdin's request. Also moving the repeated logs fix to another PR.

Issue #4736, opened 2019-11-05
Create code style for long SQL queries in Queries.php files


MasterOdin:
From discussion in #4733 

The Queries.php files have several different style for long queries. Unfortunately, I do think this is something that PHPCS can help us, and must be (for now) something we enshrine and self-maintain. We should determine what our preferred style is, write that on submitty.org and start refactoring Queries.php to match (and then maintain over code reviews of PRs).

To start, here are some examples I saw while browsing through the files:
1. 
```
            $query = "
                UPDATE gradeable_data SET
                  gd_overall_comment=?,
                  gd_user_viewed_date=?
                WHERE gd_id=?";
            $this->course_db->query($query, $params);
```

2. 
```
                $this->course_db->query("
                    UPDATE electronic_gradeable SET
                      eg_submission_open_date=?,
                      eg_submission_due_date=?,
                      eg_is_repository=?,
                      eg_subdirectory=?,
                      eg_vcs_host_type=?,
                      eg_team_assignment=?,
                      eg_max_team_size=?,
                      eg_team_lock_date=?,
                      eg_use_ta_grading=?,
                      eg_scanned_exam=?,
                      eg_student_view=?,
                      eg_student_view_after_grades=?,
                      eg_student_submit=?,
                      eg_has_due_date=?,
                      eg_config_path=?,
                      eg_late_days=?,
                      eg_allow_late_submission=?,
                      eg_precision=?,
                      eg_peer_grading=?,
                      eg_peer_grade_set=?,
                      eg_regrade_request_date=?,
                      eg_regrade_allowed=?,
                      eg_grade_inquiry_per_component_allowed=?,
                      eg_thread_ids=?,
                      eg_has_discussion=?
                    WHERE g_id=?", $params);
```

3. 
```
        $this->course_db->query("SELECT COUNT(*) as cnt
FROM gradeable_component_data as gcd
WHERE gcd.gcd_grader_id IN {$grader_list}
AND gc_id IN (
  SELECT gc_id
  FROM gradeable_component
  WHERE gc_is_peer='t' AND g_id=?
)", $params);
```

4.
```
        $this->course_db->query("
        SELECT count(u.*), u.registration_section
        FROM users as u
        INNER JOIN(
            SELECT gd.* FROM gradeable_data as gd
            LEFT JOIN(
                gradeable_component_data as gcd
                LEFT JOIN gradeable_component as gc
[...]

PR #4786, opened 2019-11-15
Update DatabaseQueries.php

kshitijandmojo:
@Referring to discussion on issue #4736 
@MasterOdin  @bmcutler  @prestoncarman  @emaicus 
I have tried creating a coding style of the Queries.php files 
please review them if they are correct/suitable. In case of any Mistakes please let me know so that I can make the changes elsewhere in the files.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -500,7 +500,25 @@ class DatabaseQueries {
 
         try {
             //insert data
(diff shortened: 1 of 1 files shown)


MasterOdin:
While we appreciate the effort here, we are not yet at a stage for opening PRs to update this file to confirm to a particular coding style as we have not yet event agreed to one.

Please feel free to voice your opinion on how you think queries should be formatted in #4736, giving some rationale for certain decisions, like why 10 characters for indentation in some places, when/why you're splitting a query in half, etc.

PR #4793, opened 2019-11-19
[Bugfix:Autograding] Added site_log_path to Workers

emaicus:
The ```site_log_path```, which could more accurately be called the ```submitty_log_path```, was missing from worker machines, causing an error when autograding. This PR adds the ```site_log_path``` to workers.


emaicus:
For affected systems, this PR will require either a manual edit of the ```submitty.json``` or a re-run of ```CONFIGURE_SUBMITTY.py```.

PR #4819, opened 2019-11-28
[Bugfix:Forum] Corrected invalid typeof check for the forum

elihschiff:

a typeof is checked to see if it equals "int"

### What is the new behavior?

typeof will never return the value "int" in javascript, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof for the list of types that can be returned. I changed "int" to "number" as that is the value that would be returned.

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -1205,7 +1205,7 @@ if (!Array.prototype.toggleElement) {
     Object.defineProperty(Array.prototype, 'toggleElement', {
         value: function(element, comparer) {
             var index = this.inArray(comparer);
-            if ((typeof(index) == "boolean" && !index) || (typeof(index) == "int" && index === 0)) {
+            if ((typeof(index) == "boolean" && !index) || (typeof(index) == "number" && index === 0)) {
                 this.push(element);
             }
             else {


shailpatels:
Approving, looks like the forum only ever passes a string to this function and thats why this issue has never been hit.

shailpatels:
I removed the second condition and tested with and without your changes and the behavior stayed the same, my only guess is that this function is trying to handle the case if the elements exists at position 0 and JS interprets this as false from the above function 'inArray'.

'inArray' should return -1 if its not found if thats the case. But Jquery has its own inArray function we should be able to use so not sure why this is even here?

MasterOdin:
Looks like this fixes a bug in the forum filters where if you have "Unread Only" and one category filter applied, if you un-click the category filter, "Unread Only" will also de-select.

I do agree though that this function is confusing and it might be a good idea to refactor it out of existence in the three spots it's currently used.

elihschiff:
To be honest, I am not 100% sure how I could go about refactoring this function as you were suggesting above. I dont really know enough about what it is doing/how I would fix it. Maybe I should make an issue for this (and reference this merge request in it) or if you want I could let one of you try to fix this yourself.

PR #4834, opened 2019-12-03
[Bugfix:Submission] Inconsistent behaviour while uploading zip file

### LINKED ISSUE #4833: Dialog box for uploading zip inconsistent with expected behavior
**Describe the bug**
When I drag a zip into submission box, I get a dialog box that says all files in the bucket will be deleted if I upload the zip, and do I want to continue. Hitting cancel on that dialogue adds the zip to the bucket and leaves all of my files in-place. My expectation is that hitting cancel would neither remove any files from the bucket, nor add the zip to the bucket.

**Expected behavior**
Either the dialog should be made more clear on what will happen if you hit cancel. I'd expect either one of the two to happen:
1. Hitting cancel will not delete any previously uploaded files in the bucket, but will also add the zip (as it currently does)
1. Hitting cancel will cause the zip to not be uploaded.

where if it's the first case, the dialog message should be made more clear, and if it's the second case, we should fix the behavior (and maybe reword the dialog box slightly?).

**To Reproduce**
1. Put files into bucket
2. Drag zip file into bucket
3. See Dialogue
1. Hit cancel

**Configuration**
 - Browser [e.g. chrome, safari] Chrome

**Screenshots**
Screenshot 2019-12-03 06 16 41
Screenshot 2019-12-03 06 16 47

**Additional context**
Before any PR gets opened, it should be determined in this issue what exactly the bug is.

The zip does not contain any files that would match the other files in the bucket.

mukul-kmr-jha:

Closes #4833 

### What is the new behavior?
Bug is now fixed and respective feature is working as per the expectation in the mentioned [issue](https://github.com/Submitty/Submitty/issues/4833)

### DIFF
--- a/site/public/js/drag-and-drop.js
+++ b/site/public/js/drag-and-drop.js
@@ -173,17 +173,15 @@ function isFolder(file){
 function addFile(file, part, check_duplicate_zip=true){
     var i = fileExists(file.name, part);
     if( i[0] == -1 ){    // file does not exist
-        // empty bucket if file is a zip and bucket is not empty
+        // When uploading a zip, we confirm with the user to empty the bucket and then only add the zip
         if(check_duplicate_zip && file.name.substring(file.name.length - 4, file.name.length) == ".zip" && file_array[part-1].length + previous_files[part-1].length > 0 ){
             if(confirm("Note: All files currently in the bucket will be deleted if you try to upload a zip: " + file.name + ". Do you want to continue?")){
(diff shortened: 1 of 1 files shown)


mukulkrjha:
@MasterOdin Hey PTAL!

MasterOdin:
@mukulkrjhaThanks for the PR. I needed to confirm with @bmcutler that this was the desired behavior before merging.

PR #4838, opened 2019-12-05
[VPAT:Gradeable] Poor Contrast Ratio In New Gradeables View

### LINKED ISSUE #4836: VPAT — Poor Contrast Ratio In New Gradeables View
Gradeables_view(2)

Grey text (disabled tab labels near top) that is circled in blue is colored `#888888` and has a contrast ratio of 3.54:1.  Contrast ratio should be at least 4.5:1.

mukul-kmr-jha:

Closes #4836

Grey text (disabled tab labels near top) that is circled in blue is colored #888888 and has a contrast ratio of 3.54:1.

### What is the new behavior?
Contrast ratio is now 5.50: 1 .

### DIFF
--- a/site/public/css/admin-gradeable.css
+++ b/site/public/css/admin-gradeable.css
@@ -75,7 +75,7 @@ a.nav-bar.normal-btn:hover {
 }
 
 a.nav-bar.disabled-btn {
-    color: var(--standard-medium-gray) ;
+    color: var(--standard-medium-dark-gray) ;
     cursor: default;
     border: none !important;
     pointer-events: none;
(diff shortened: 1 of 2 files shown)


MasterOdin:
Thanks for this. I'd rather not change the color if we don't have to as it's used for borders and such in a number of places.

I'd prefer let's just change this line:
https://github.com/Submitty/Submitty/blob/43d90ebbeef11980ea62d23c6cdb66f98660f06a/site/public/css/admin-gradeable.css#L78

to use `--standard-medium-dark-gray` instead of `--standard-medium-gray`

MasterOdin:
Thanks for fixing this!

PR #4846, opened 2019-12-09
[Bugfix:Submission] Confetti lags on multiple clicks

### LINKED ISSUE #4841: Confetti lags when you toggle it a lot
**Describe the bug**
My friend showed me that if you get a perfect score on the autograding subtotal, you can click on it (the perfect score) to make some confetti appear. The confetti scrolls down the screen slowly as if it is falling. Clicking on the score again will toggle the confetti off. However, if you click the perfect score many times in a short period of time, the confetti appears to lag more every time it reappears.

**Expected behavior**
I expect that clicking on the perfect score should just toggle the confetti on/off without affecting how the confetti appears on the screen (it should not lag).

**To Reproduce**
1. Get a perfect score on an assignment with autograding.
2. Click the score on the submission page to turn on the confetti.
3. Continue clicking on the score to keep toggling the confetti off and on.
4. See error (the confetti will appear laggy).

**Configuration**
 - Chrome (I haven't tested other browsers)
 - Window

**Screenshots**
image
This doesn't really show that it lags, but here's proof that the confetti exists.

**Additional context**
Have a happy holidays!

mukul-kmr-jha:

### What is the current behaviour?

Closes #4841  
1. Clicking on perfect/score badge makes confetti toggle on and off , but if we click it multiple times it starts to lag heavily.
2. Confetti makes the screen disturbed and adds a lot of empty space in right side.
IMG_20191209_220849

### What is the new behaviour?
Both the issues are resolved and confetti is now optimized.


mukul-kmr-jha:
@MasterOdin @shailpatels PTAL!

mukul-kmr-jha:
Hii @shailpatels ,Thanks for review and suggestion!
I agree with the function syntax thing, I'll revert that but I also think that the refactoring I did is required, I have made the code more modular  (i.e I've broken the whole confetti functions into smaller sub-functions so that code becomes clean, optimized, easier to read and debug) .
Just for example Like on line no. 95 we are using extra if check  which can easily be handled by a `default` case inside the `switch` block. 
On line 8, each time `addConfetti` is called, a new event listener is being created and kept attached  on `canvas` reducing the performance.
On line 170 and 119 each time a `piece` is added we call `getRandomColor()` and create color array to just get a color from probably same set of colors (as this set of color depends on the month value, which does not change so often). (Note that this happens *2500* times). 
And making the pieces array global saves us from creating new set of pieces each time we call `addConfetti`.

mukul-kmr-jha:
I agree that it will be difficult to review but I believe that the changes that I made are required. But if you insist I can revert these changes and just make the bug fix in this PR like as you suggested.

mukul-kmr-jha:
@MasterOdin  what are your thoughts on this, a second opinion would be helpful.

shailpatels:
@mukul-kmr-jha After talking this over with bmcutler, we recommend you implement this using the suggested changes, since you only need the additional if statements to fix this. The core problem is that the draw and update functions silently keep running when the user clicks to end confetti early. This can be done with the check if the canvas display is set to none and return if so.

A refactor is not required, the click handler doesn't actually add a new one on every click just once and moving the if statement into to switch doesn't justify a refactor. 

Overall confetti is more of an Easter egg and this is not a high priority issue. We'd like to close the issue  
And pr and since the bugfix is so small we could just create a simple two line fix pr otherwise.

mukul-kmr-jha:
@shailpatels , I have reverted the code refactor and implemented the requested changes for the bugfix as per your suggestions. Please take another look . Thank you!

site/public/js/confetti.js
```diff
@@ -90,12 +91,12 @@ function addConfetti(){
...
+      //make sure we have a default if parsing goes wrong
+      default:
+        colors = ['#8FD7FF', '#316498', '#34CA34', '#FFFF40', '#FF2929', '#9c84a4'];
+
+    }
 
-		//make sure we have a default if parsing goes wrong
-		if(colors.length === 0)
-			colors = ['#8FD7FF', '#316498', '#34CA34', '#FFFF40', '#FF2929', '#9c84a4'];
-	    return colors[Math.floor(Math.random() * colors.length)];
```
MasterOdin: We still want this return, right?

mukul-kmr-jha: Yes we do, Just missed it out.

MasterOdin:
> the click handler doesn't actually add a new one on every click just once and moving the if statement into to switch doesn't justify a refactor.

This is false. A new handler is appended each time as you're using anonymous functions for the click handler, and which JS will never consider two anonymous functions equivalent. Switching to named functions would have it only attach a single listener. This would probably be a good thing to fix.

> @MasterOdin what are your thoughts on this, a second opinion would be helpful.

I agree with @shailpatels that a full-on refactor of this easter egg feature is probably not necessary, and I'd like to keep all variables and functions as self-contained as possible to prevent global scope creeping into other places (there's too much of that already and it's causing us headaches). Having said that, I do agree it would be good to do some of the easy optimizations. 


For example, I would probably get rid of the random color function and do the following:
```
colors = {
  0: ['#406bc9','#ffffff','#809bce','#9ac8de','#b6c7be'],
  1: ['#df3b57','#ee4b6a','#7d2335','#86cec5','#b2e6f1'],
  2: ['#8db62f','#7b9233','#034121','#022607','#ffcc00'],
  ...
  11: ['#d7cdcc','#f7b11d','#1f5e00','#de1a1a','#ffffff']
}

let d = new Date();
let month = d.getMonth();

// if we parsed the submission due date, use that instead
if(submission_date.length >= 1) {
  month = parseInt(submission_date[0], 10) - 1;
  // default to 8 if bad parse, as it has submitty blue and shail green, the best colors
  month = isNaN(month) ? 8 : Math.max(0, Math.min(11, month));
}

function Piece (x, y) {
  this.x = x;
  this.y = y;
  this.color = colors[month][Math.floor(Math.random() * colors.length)];
}
```
It saves on calling that function 2500 times and all that, but like most of everything here, this is in principle a micro-optimization in that it most likely does not actually affect performance in the real-world on people's machines, and so not much point in focusing on it.

MasterOdin:
Thanks for the PR fixing this.

PR #4818, opened 2019-11-26
[Refactor:Configuration] Refactor and cleanup configuration controller

MasterOdin:
### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [x] Tests for the changes have been added/updated (if possible)

### What is the current behavior?

Configuration Controller was not tested, and did several questionable operations in its functioning.

### What is the new behavior?

Cleans up the configuration controller, as well as how the submitty_admin user is handled in the config model. The config model no longer runs a query to determine if the user is an instructor in the course (the controller does that), and the verified submitty admin user field is read in on the initial read of the submitty_users.json file, removing a second opening and reading of it.

PR #4869, opened 2019-12-13
[Bugfix:Instructor UI] Exam Seating Dropdown Dropping An Option

holzbh:
If a gradeable is set up to be used for exam seating (i.e. the `reports/seating/<g_id>` directory exists and `g_id` is a valid gradeable), but the gradeable TITLE is alphabetically first among all gradeable titles in the course, then it did not display. This was due to the "None" option always occupying position 0 in the resulting array.

This has been rectified by making the "None" option occupy position -1 which should never be occupied by the query/filtering of valid gradeable IDs.


MasterOdin:
Test case added for this in #4818.

PR #4871, opened 2019-12-13
[Bugfix:TAGrading] popup error when writing a custom message

### LINKED ISSUE #4729: Ask for confirmation when cancelling a mark when writing a custom message to a student
Complaint from a grader, after writing a custom message to the student and accidentally hitting 
cancel the entire marked is closed and the message is deleted. 

Either a warning should appear when the custom text area box is not empty or we should not clear the textarea and only wipe it on save if not confirmed.

mukul-kmr-jha:

Closes #4729 
If we make any changes made in custom message for some mark or overall message and hit *cancel* there is prompt for the TA which lead to accidental loss of that message.  
### What is the new behavior?
Now if there is any changes made in the custom message of any mark or the overall component and you press *cancel* submitty will ask you for a confirmation.  

confirm

### DIFF
--- a/site/public/js/ta-grading-rubric.js
+++ b/site/public/js/ta-grading-rubric.js
@@ -883,10 +883,8 @@ function setRubricDOMElements(elements) {
  */
 function getComponentIdFromDOMElement(me) {
     if ($(me).hasClass('component')) {
-        console.log('element itself has id for comp', parseInt($(me).attr('data-component_id')));
         return parseInt($(me).attr('data-component_id'));
     }
-    console.log('element parent contain id for comp', parseInt($(me).parents('.component').attr('data-component_id')));
     return parseInt($(me).parents('.component').attr('data-component_id'));
 }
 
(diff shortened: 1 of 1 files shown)


mukul-kmr-jha:
@MasterOdin @shailpatels  please take a look !

shailpatels:
> @MasterOdin @shailpatels please take a look !

We get an email whenever a Pr is made/updated no need to tag us :)

site/public/js/ta-grading-rubric.js
```diff
@@ -1867,11 +1886,27 @@ function onComponentOrderChange() {
  * @param me DOM element of the cancel button
  */
 function onCancelOverallComment(me) {
-    toggleOverallComment(false)
-        .catch(function (err) {
+  // if the overall Comment is changed then prompt the TA for loss in comment
+  ajaxGetOverallComment(getGradeableId(), getAnonId()).then((lastSavedOverallComment) => {
```
shailpatels: This function is almost identical to the above, lets have a helper function somewhere that accepts the getGradedComponent promise and live custom comment and compares them, if they're different give a prompt otherwise do nothing.

That way both these methods can call that function and in the future we need to change/add functionality it'll be easier.

site/public/js/ta-grading-rubric.js
```diff
@@ -1832,11 +1832,30 @@ function onClickComponent(me) {
...
 function onCancelComponent(me) {
-    toggleComponent(getComponentIdFromDOMElement(me), false)
-        .catch(function (err) {
+  const component_id = getComponentIdFromDOMElement(me);
+  const gradeable_id = getGradeableId();
+  const anon_id = getAnonId();
+  ajaxGetGradedComponent(gradeable_id, component_id, anon_id).then((component)=>{
+    // If there is any changes made in comment of a component , prompt the TA
+    if ( component.comment !== $('#component-' + component_id).find('.mark-note-custom').val().trim()) {
+      if(confirm("Note: Changes made to the Student Message will be lost! Do you want to continue ? ")){
```
shailpatels: This warning message should be more direct,
something like "Are you sure you want to discard all changes to the student message?"

site/public/js/ta-grading-rubric.js
```diff
@@ -1832,11 +1832,30 @@ function onClickComponent(me) {
...
  */
 function onCancelComponent(me) {
-    toggleComponent(getComponentIdFromDOMElement(me), false)
-        .catch(function (err) {
+  const component_id = getComponentIdFromDOMElement(me);
+  const gradeable_id = getGradeableId();
+  const anon_id = getAnonId();
+  ajaxGetGradedComponent(gradeable_id, component_id, anon_id).then((component)=>{
+    // If there is any changes made in comment of a component , prompt the TA
+    if ( component.comment !== $('#component-' + component_id).find('.mark-note-custom').val().trim()) {
```
shailpatels: I don't think trimming the message is required in this case since adding/removing whitespace could be a valid change the user would like to discard.

mukul-kmr-jha: done!

site/public/js/ta-grading-rubric.js
mukul-kmr-jha: done!

site/public/js/ta-grading-rubric.js
mukul-kmr-jha: @shailpatels the two methods just look identical but they use different DOM components, so I guess writing a helper function for just checking the equality in comments (which is currently done inside if condition) will be too much. What do you think?

shailpatels: Oh ok I see now, I guess its not ideal but if theres not a good way to consolidate code then this is good, I'll approve now.

shailpatels:
Looks good, works as described

site/public/js/ta-grading-rubric.js
mukul-kmr-jha: Thanks!

PR #4883, opened 2019-12-15
[VPAT:Forum] ctrl+enter for direct submission of post

### LINKED ISSUE #3570: "CTRL + ENTER" To direct submission of messages in discussion forum
Using `CTRL + ENTER` for direct submission of post in discussion forum
This feature is used in following organisation:-
1. Github
2. Gmail
Many organisation uses `CTRL + ENTER` to submit posts.

elihschiff:

There is no way for a user with just a keyboard to submit a post/comment on the discussion forum

### What is the new behavior?

In all the textboxes on the discussion forum pressing ctrl+enter will submit their post.
closes #3570

### Other information?

I tested on new posts, comments, and editing.

### DIFF
--- a/site/app/templates/forum/ThreadPostForm.twig
+++ b/site/app/templates/forum/ThreadPostForm.twig
@@ -81,7 +81,16 @@
     </div>
     <div class="form-group row">
         {% set min_height = (post_textarea_large?"40vmin":"100px") %}
-        <textarea name="thread_post_content" id="thread_post_content" class="fill-available post_content_reply" style="resize:none;min-height:{{ min_height }};max-height:300px" rows="10" cols="30" placeholder="{{ post_content_placeholder }}" required aria-label="Thread Post Textarea" maxlength="{{ post_content_limit }}" ></textarea>
+        <textarea name="thread_post_content" id="thread_post_content" class="fill-available post_content_reply" style="resize:none;min-height:{{ min_height }};max-height:300px" rows="10" cols="30" placeholder="{{ post_content_placeholder }}" required aria-label="Thread Post Textarea" maxlength="{{ post_content_limit }}"
+        onkeydown="
+            if ((event.keyCode == 10 || event.keyCode == 13) && event.ctrlKey) {//ctrl+enter keys pressed
+                var theForm = $(this).closest('form');
+                if(theForm.hasClass('dirty')){
+                    theForm.trigger('reinitialize.areYouSure');
+                    theForm.submit();
+                }
+            }
+        "></textarea>
     </div>
 <br/>
 {% endif %}
--- a/site/app/templates/forum/ThreadPostForm.twig
+++ b/site/app/templates/forum/ThreadPostForm.twig
@@ -193,16 +193,15 @@
 </div>
 <br>
 </div>
-<script>
+<script type="text/javascript">
     $(function(){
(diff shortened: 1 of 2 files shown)


site/app/templates/forum/ThreadPostForm.twig
```diff
@@ -81,7 +81,16 @@
     </div>
     <div class="form-group row">
         {% set min_height = (post_textarea_large?"40vmin":"100px") %}
-        <textarea name="thread_post_content" id="thread_post_content" class="fill-available post_content_reply" style="resize:none;min-height:{{ min_height }};max-height:300px" rows="10" cols="30" placeholder="{{ post_content_placeholder }}" required aria-label="Thread Post Textarea" maxlength="{{ post_content_limit }}" ></textarea>
+        <textarea name="thread_post_content" id="thread_post_content" class="fill-available post_content_reply" style="resize:none;min-height:{{ min_height }};max-height:300px" rows="10" cols="30" placeholder="{{ post_content_placeholder }}" required aria-label="Thread Post Textarea" maxlength="{{ post_content_limit }}"
+        onkeydown="
```
shailpatels: Could you please move the JS into its own function in the script tag at the bottom of the file and have the onkeydown even call that instead of the embedded JS inside of the HTML, it makes the code harder to read.

site/app/templates/forum/ThreadPostForm.twig
```diff
@@ -199,3 +200,14 @@
         });
     });
 </script>
+<script type="text/javascript">
```
shailpatels: You can use the existing script tag defined above, no need to create another.

site/app/templates/forum/ThreadPostForm.twig
```diff
@@ -199,3 +200,14 @@
...
+    function submitNearestForm(e, textarea){
+        if ((e.keyCode == 10 || e.keyCode == 13) && e.ctrlKey) {//ctrl+enter keys pressed
```
shailpatels: I have a feeling this isn't going to work on Mac, e.ctrlkey is windows hardware-specific you're probably going to need to do `e.metaKey || e.ctrlKey`

elihschiff:
I tried to add `&#10; &#13;` and it did not work for me. Is there any reason \\n is bad to use as it seems to be working fine.

Also I am unable to test on mac, I have only tested on linux, so if somebody on mac can review this pr and make sure it works that would be great.

shailpatels:
> I tried to add `&#10; &#13;` and it did not work for me. Is there any reason \n is bad to use as it seems to be working fine.
> 
> Also I am unable to test on mac, I have only tested on linux, so if somebody on mac can review this pr and make sure it works that would be great.

Oh thats annoying, looks like the CRLF is a firefox specific thing, but it looks like a \n will
work on firefox and chrome, I was testing by inspect elementing the page so I guess preparing the \n in php makes it work.

shailpatels:
Also is this a VPAT issue? I think [Feature:Forum] is a better fit but I don't remember if this issue was made in the context of VPAT stuff.

elihschiff:
I labeled it vpat because I was under the impression that it is for people who only use the keyboard which I believe is mostly a vpat thing. Do you think I should switch it to be labeled as forum? I don't know how much it matters either way though

PR #4893, opened 2019-12-19
[Bugfix] PHP code incorrectly rounds valid numbers

### LINKED ISSUE #4890: PHP code incorrectly rounds valid numbers
**Describe the bug**
Using a precision of 0.05 and a max value of 6 for points for a component, the PHP code will incorrectly set it to 6.05 always.

**Expected behavior**
The component should save as 6.

**To Reproduce**
1. Edit a gradeable's rubric
2. Set the precision to 0.05
3. Set the max value for component to 6
4. Hit save

**Additional context**
The problem exists in this function in the Gradeable model:
https://github.com/Submitty/Submitty/blob/c617b754c99a279997bcef35bc2cde2a0d6249e7/site/app/models/gradeable/Gradeable.php#L1157-L1179

where given a `$this->precision = 0.05` and `$points = 6`, it returns 6.05.

mukul-kmr-jha:

Currently there is a floating point bug inside the points calculation, this behavior can be seen by setting the precision of points as 0.05 and total point value as 6, as soon as we focuses out of the points input it gets converted to  6.05 which shouldn't be the case as 6 can be the point value using 0.05 precision. 

### What is the new behavior?
The bug is removed and it closes #4890. It improves the algorithm such that it should now better handle a number of different precisions and values. Given a small enough precision, may still have some (slight) rounding bugs, but it's not clear if we're going to hit those cases in practical usage.

### Other Information
Currently working on developing test cases to check and verify the intended behavior.

### DIFF
--- a/site/app/libraries/NumberUtils.php
+++ b/site/app/libraries/NumberUtils.php
@@ -11,33 +11,42 @@ namespace app\libraries;
 class NumberUtils {
 
     /**
-     * @param $x float The number to round
-     * @param $y float The number $x to be rounded with respect to this variable $y
-     * @return float The rounded result to the nearest multiple of $y
+     * NumberUtils constructor.
+     * @param Core $core
      */
-    public static function roundPointValue($x, $y) {
-
-        // No $y, no rounding
-        if ($y === 0.0) {
-            return $x;
-        }
-
-        $x = floatval($x);
-        $q = (int) ($x / $y);
-        $r = self::fmod_round($x, $y);
-
-        // If the remainder is more than half the $y away from zero, then add one
-        //  times the direction from zero to the quotient.  Multiply by $y
-        return ($q + (abs($r) > $y / 2 ? ($r > 0 ? 1 : -1) : 0)) * $y;
+    public function __construct(Core $core) {
+        parent::__construct($core);
     }
 
     /**
-     * @param $x float
-     * @param $y float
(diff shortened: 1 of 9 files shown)


site/app/libraries/NumberUtils.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+class NumberUtils {
+
+    /**
+     * @param $x float The number to round
+     * @param $y float The number $x to be rounded with respect to this variable $y
+     * @return float The rounded result to the nearest multiple of $y
+     */
+    public static function roundPointValue($x, $y) {
```
MasterOdin: Let's use `$value` and `$precision` as the parameter names.

We can also add the float as direct type hints and not just in phpdoc string.

bmcutler:
Please revise the top level post for this PR to actually describe the problem and fix.
It is not acceptable to just link to the issue.  These PR notes are part of the documentation for our release notes and should not require the reader to follow links to find the problem description.

We'll need to update the submitty.org documentation and github PR template to make these expectations clear.

site/app/libraries/NumberUtils.php
mukul-kmr-jha: @MasterOdin okay

MasterOdin:
Thanks for fixing this! :+1:

PR #4900, opened 2019-12-21
[Feature:Submission] Randomize order of answers in Notebook

### LINKED ISSUE #4420: Randomize order of answers in Notebook Gradeable
Let's add an option to the multiple choice notebook cell type:
"randomize_order" : true

(if omitted this field defaults to false)

When true, the choice in the the question should be psuedo-randomized based on a hash of the student name and the gradeable id.  This way if the same student returns to the page they have the same order of answers, but 2 different students will probably have different orderings.

Note that the answers stored in the file will not be randomized, since each answer has a fixed "value".

mukul-kmr-jha:

Currently the options in multiple choice notebook appears as they are in the `choices` array without and randomization.

### What is the new behavior?
Closes #4420 
Now an option `randomized_order` can be used to set the psuedo-randomization in order of choices.

*Psuedo-randomization* This randomization is based on a hash of the student name and the gradeable_id, which makes the choices appearing in the random-fashion for different students (most likely), but also makes sure that it is not different each time a user revisits page. 
Also this doesn't affect the choices stored in file.


mukul-kmr-jha:
@MasterOdin @bmcutler please take a look!

mukul-kmr-jha:
Thanks @Kevin-MM for reviewing this PR,  sorry for no instructions on how to test these changes. 
A clean installation is required to test the feature changes.
PR involves changes in grading directory,  so theoretically running these commands should have been enough `sudo /usr/local/submitty/.setup/INSTALL_SUBMITTY.sh clean`  and then `/var/local/submitty/courses/s20/<course>/BUILD_<course>.sh` 
But for some unknown reasons to me these commands don't serve the purpose, and I tested and implemented it using  the former approach of clean installation. 
Thanks!

Kevin-MM:
Could you please resolve the merge conflicts and I will re-review.

mukul-kmr-jha:
@Kevin-MM PTAL, Thanks!

mukul-kmr-jha:
@Kevin-MM Added a check per your suggestion. btw this error doesn't show up on re-installation but adding a check ofcourse makes it more robust to implement. Thanks!

mukul-kmr-jha:
Yeah,  let me update the docs.
Thanks!

PR #4901, opened 2019-12-23
Only tab navigate within a modal

jchulton:
This aims to fix issue #3830 

The bug this aims to fix is, when a modal is created a user could tab navigate to elements outside of that modal. The fix is a general Javascript function located in server.js which grabs the last and first tab-able elements from the modal and loops back to the first element when tabbing out of the last element. 
This fix has been implemented to some simple modals in server.js but not all modals yet (wanted to get feedback first).

This can be tested in Upload Course Materials and Delete Course Materials in Course Materials

### DIFF
--- a/site/public/js/server.js
+++ b/site/public/js/server.js
@@ -158,6 +158,7 @@ function newDeleteGradeableForm(form_action, gradeable_name) {
     $('[name="delete-gradeable-message"]', form).append('<b>'+gradeable_name+'</b>');
     $('[name="delete-confirmation"]', form).attr('action', form_action);
     form.css("display", "block");
+    tabModalFix("delete-gradeable-form");
 }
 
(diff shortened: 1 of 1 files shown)


jchulton:
I've implemented shift+tab, I'm using the focusable filter, I've changed the function name to captureTabInModal and I've added a helper function reelaseTabFromModal to toggle the tab redirecting using off. The helper function will be automatically be killed when the modal is hidden/closed. This should address the given feedback

jchulton:
The visible filter or the input list itself isn't the problem because default functionality only changes if the current element is the first or last in the list: the script doesn't loop through each element in the input list. 

The upload box is also being skipped in master. I think this should be addressed in another pull request. I'm working on a separate pull request for elements in the forum being skipped, I can add another one for elements in the course materials or add it on to the forum one

PR #4904, opened 2019-12-29
[Feature:TAGrading] "Rollback Available" TA grading

### LINKED ISSUE #4891: "Rollback Available" TA grading
When a TA is grading and encounters a late submission (that will get an automatic zero unless an extension is granted), we'd like to visually indicate whether that student has a different version that is *not* late, and that the assignment can be "rolled back" prior to grading.  (Ideally the student would have set the version correctly, but it happens frequently that they do not.

Let's modify the banner message:

Screen Shot 2019-12-18 at 3 16 20 PM

To say either:
"Late Submission  --  No on time submission available"
Or:
"Late Submission --  Rollback to on time submission available"

We can revisit the background colors we use to indicate this and other different situations (no submission, cancelled submission, version conflict)

mukul-kmr-jha:

Currently, if a student submits multiple times (i.e multiple versions) then by default the most recent version is set to be graded (student can, however, decides this and selects which version is to be graded by the grader). But Sometimes the student uploads the assignment in a late period even when s/he already has a good submission, this may lead to evaluation as zero. 

### What is the new behavior?
Closes #4891 
Provides a visual clue to the instructor that the student has one of their assignment submitted on time, if s/he has more than one good submissions then submitty suggests the most recent good submission to `rollback`. This helps to avoid any accidental misevaluation as the instructor can `"roll-back"` to that suggested version for grading.
### Other information?
Rollback is Available
Screenshot 2020-02-21 at 1 56 53 PM

No Good Submission Found.
Screenshot 2020-02-21 at 1 57 31 PM

### DIFF
--- a/site/app/controllers/grading/ElectronicGraderController.php
+++ b/site/app/controllers/grading/ElectronicGraderController.php
@@ -1042,6 +1042,29 @@ class ElectronicGraderController extends AbstractController {
         else {
             $late_status = $ldi->getStatus();
         }
+        $rollbackSubmission = -1;
+        $previousVersion = $display_version-1;
+        // check for rollback submission only if the Active version is greater than 1 and that too is late.
+        if ($previousVersion && $late_status !== LateDayInfo::STATUS_GOOD) {
+            while($previousVersion) {
+                $graded_gradeable->getAutoGradedGradeable()->setActiveVersion($previousVersion);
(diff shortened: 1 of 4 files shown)


bmcutler:
@shailpatels 
will fix last piece in #5092

shailpatels:
Approved based on : 
> @shailpatels
> will fix last piece in #5092

PR #4907, opened 2020-01-01
[Refactor:InstructorUI] Course Materials no longer uses 'check'

### LINKED ISSUE #4538: Refactor Course Materials to no longer use 'check'
Before course material used release datetimes, instructors had an option of a binary checkmark which releases the materials immediately. This was replaced with an 
option to select a date on which the materials would be released.

However, the code base still has a lot of leftover code and specifically the variable
'checked' is still being passed from the client to the server which is then being written to
`course_materials_data_file.json` even though the value of 'check' for a file is never being used anymore.

Refactor course materials controller to remove using the value of check and update the tests accordingly

mukul-kmr-jha:

Earlier Instructors had an option to release course material immediately. This option (type : checkbox) was replaced with an option to select a date on which the materials would be released. Maybe to provide backward compatibility,  variables and methods were not removed completely from the `course-material` module, leading to left-over code. 

### What is the new behavior?
Closes #4538 
The `course-material` module along with its test cases are refactored .


mukul-kmr-jha:
@shailpatels @MasterOdin Please take a look.
Thanks!

bmcutler:
@jchulton please review ASAP

PR #4910, opened 2020-01-04
Logging in after attempting to view page with query string does not work

### LINKED ISSUE #4556: Logging in after attempting to view page with query string does not work
**Describe the bug**
When going to a page with `$_GET` parameters when logged out, you will be taken to the login screen as expected, but after logging in, you will be redirected to a page without the `$_GET` parameters.

**Expected behavior**
The `$_GET` parameters to be part of the redirected URL.

**To Reproduce**
1. Go to http://192.168.56.111/f19/sample/gradeable/open_homework/grading/grade?who_id=aphacker&sort=id&direction=ASC
2. Delete all cookies
3. Hit Refresh to get login screen and login in

mukul-kmr-jha:

Currently, if we get logged-out on a page with some query parameters we are asked to log in (`intended`) and then after successfully logging-in you will get redirected to the page but without query parameters. So query-parameter is not working within log-out events.

### What is the new behavior?
Closes #4556 
Now, If you get logged-out at any page which is having query-params, you will be redirected to the same old URL you are were initially at with correct params.

### DIFF
--- a/site/tests/app/libraries/routers/WebRouterTester.php
+++ b/site/tests/app/libraries/routers/WebRouterTester.php
@@ -41,7 +41,7 @@ class WebRouterTester extends BaseUnitTest {
         );
         $response = WebRouter::getWebResponse($request, $core);
         $this->assertEquals(
-            $core->buildUrl(['authentication', 'login']) . '?old=' . urlencode($request->getUriForPath($request->getPathInfo())),
+            $core->buildUrl(['authentication', 'login']) . '?old=' . urlencode($request->getUriForPath('/s19/sample')),
             $response->redirect_response->url
         );
     }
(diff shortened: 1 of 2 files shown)


mukul-kmr-jha:
@MasterOdin PTAL.
Thanks!

site/tests/app/libraries/routers/WebRouterTester.php
```diff
@@ -39,9 +39,18 @@ public function testRedirectToLoginFromCourse() {
...
+        $filtered_query_obj = array_filter($query_obj, function ($k) {
+            return $k !== "url";
+        }, ARRAY_FILTER_USE_KEY);
+
+        $query_string = empty($filtered_query_obj) ? null : '?' . http_build_query($filtered_query_obj);
+
         $response = WebRouter::getWebResponse($request, $core);
         $this->assertEquals(
-            $core->buildUrl(['authentication', 'login']) . '?old=' . urlencode($request->getUriForPath($request->getPathInfo())),
+            $core->buildUrl(['authentication', 'login']) . '?old=' . urlencode($old_request_url . $query_string),
```
MasterOdin: Welp, this was missed in my initial review of this. For both tests, can we replace the usage of a variable created off the `$request` variable to just a set hard-coded string. This will make it easier to reason about what it's doing and catch if behavior does subtle change going forward.

mukul-kmr-jha: Yes, I will keep this in mind for future PRs

mukul-kmr-jha:
@MasterOdin Done! Please take another look. Thanks.

PR #4912, opened 2020-01-07
[Grading:Feature] Incremental Peer Grading Update

### LINKED ISSUE #4873: Re-enable Peer Grading
"Turn On" Peer grading. Re-enable it from the gradeable config menu and fix any bugs.

emaicus:
Closes #4873

This is an incremental PR which begins to re-evaluate and integrate the existing, non-used peer grading code. 

### This PR
1. Adds an ```Add Peer Component``` button to the instructor rubric creation view -- currently with a warning not to click it.
2. Enables/re-enables the ability for multiple users to assign different scores for the same peer component.
3. Fixes a variety of bugs with peer components.

### Changes to Peer Grading
The initial draft of the peer grading system imagined that whether a gradeable was a peer gradeable or not would be defined at gradeable creation time much like team grading. Instead, this PR moves towards a system where an instructor can introduce both peer a non-peer components to a gradeable. A peer gradeable, then, is a gradeable which has one or more peer components, and a non-peer gradeable is a gradeable which has no peer components.

### For Closer Review
This PR modifies an existing API Route (Component Creation) to add an ```is_peer``` boolean. Please comment if this is done improperly or if there is a preferred way to modify Routes.


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1412,7 +1412,8 @@ public function ajaxSaveComponent($gradeable_id) {
         $default = $_POST['default'] ?? null;
         $max_value = $_POST['max_value'] ?? null;
         $upper_clamp = $_POST['upper_clamp'] ?? null;
-        $peer = $_POST['peer'] ?? 'false';
+        // You cannot change the "peer-ness" of a component.
+        //$peer = $_POST['peer'] ?? 'false';
```
MasterOdin: let's remove the commented out code here. I'm not necessarily against the comment, but it should probably be moved into the comment for the function, rather than in the body of the function.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1455,7 +1456,7 @@ public function ajaxSaveComponent($gradeable_id) {
         if (strval(intval($page)) !== $page) {
             $this->core->getOutput()->renderJsonFail('Invalid page parameter');
         }
-        $peer = $peer === 'true';
+        //$peer = $peer === 'true';
```
MasterOdin: remove

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1487,7 +1488,7 @@ public function ajaxSaveComponent($gradeable_id) {
                 'upper_clamp' => $upper_clamp
             ]);
             $component->setPage($page);
-            $component->setPeer($peer);
+            //$component->setPeer($peer);
```
MasterOdin: remove

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1623,15 +1624,17 @@ public function saveComponentsPage(Gradeable $gradeable, int $page) {
 
     /**
      * Route for adding a new component to a gradeable
-     * @Route("/{_semester}/{_course}/gradeable/{gradeable_id}/components/new", methods={"POST"})
+     * @Route("/{_semester}/{_course}/gradeable/{gradeable_id}/components/{peer}/new", methods={"POST"})
```
MasterOdin: This is improper usage of the route. Setting of the `$peer` variable should be done through a `$_POST` variable, not as a part of the route.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1651,7 +1654,7 @@ public function ajaxAddComponent($gradeable_id) {
                 0,
                 0,
                 false,
-                false,
+                $peer,
```
MasterOdin: given we're not using `$peer` except here, let's delete the change on line 1636 and make this line read `$peer === 'true'` (or more preferably, `$_POST['peer'] === 'true'`)

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -50,7 +50,7 @@ public function statusPage(
     ) {
 
         $peer = false;
-        if ($gradeable->isPeerGrading() && $this->core->getUser()->getGroup() == User::GROUP_STUDENT) {
+        if ($gradeable->isPeerGrading()) {//&& $this->core->getUser()->getGroup() == User::GROUP_STUDENT) {
             $peer = true;
         }
```
MasterOdin: This can just be simplified to `$peer = $gradeable->isPeerGrading()` instead of having the if structure.

PR #4914, opened 2020-01-07
[Feature:InstructorUI] common location for student images

bmcutler:
Student images were previously only stored per course, and the instructor was expected to upload them.  e.g.  /var/local/submitty/courses/s20/csci1000/uploads/student_images/student.jpeg

Now in addition to searching for a match in the previous location (prioritized), we will also search in a common location:  /var/local/submitty/student_images/s20/student.jpeg

This PR does not provide a web interface for uploading images to this common location (should be done by the sysadmin).


MasterOdin:
I think it would probably make sense to also include a common location that does not include term information?

PR #4918, opened 2020-01-10
[Grading:Feature] Correctly Display Peer Grading Marks

emaicus:
This is a second incremental PR in a sequence meant to integrate peer grading into Submitty.

This PR is based off of the branch ```peer_grading_part_one``` which is used as the base for PR #4912 
All differences between this PR and #4912 are contained in ```DatabaseQueries.php``` Other changes should be reviewed in #4912.

### This PR:
Updates the ```getGradedGradeablesUserOrTeam``` query to include data about which user a mark is associated with. This information is vital to the computation of peer grading scores and to the display of peer components in the ta grading interface.


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5615,10 +5615,11 @@ private function getGradedGradeablesUserOrTeam(array $gradeables, $users, $teams
...
                       json_agg(gcm_id) AS string_mark_id,
                       gc_id,
-                      gd_id
+                      gd_id,
+                      gcd_grader_id
                     FROM gradeable_component_mark_data
-                    GROUP BY gc_id, gd_id
-                  ) AS gcmd ON gcmd.gc_id=in_gcd.gc_id AND gcmd.gd_id=in_gcd.gd_id
+                    GROUP BY gc_id, gd_id, gcd_grader_id
+                  ) AS gcmd ON gcmd.gc_id=in_gcd.gc_id AND gcmd.gd_id=in_gcd.gd_id AND gcmd.gcd_grader_id=in_gcd.gcd_grader_id
```
MasterOdin: So that I understand this, for peer grading, each peer will submit their own gradeable_component_data row with its own gradeable_component_mark_data?

emaicus:
The concern that was brought up at our previous meeting -- that this change could break non-peer grading when multiple graders touch the same component -- seems to be a non-issue. When marks are saved to the database for a non-peer component, ownership of all of them is claimed by the grader. I.e. the most recent grader has ownership. @bmcutler  indicated that this is the intended design.

PR #4919, opened 2020-01-10
[Bugfix: InstructorUI] Fix CSV Upload "not a string" exception

holzbh:
Uploading a CSV to either Manage Students or Manage Graders page causes an exception due to `$uploaded_data` being used instead of `$row`, leading to things like the new user ID being an array of information instead of the ID

### What is the new behavior?
Uploading a CSV correctly reflects the level/registration section and works for new users and updated users. No exceptions are thrown.

### Other information?
Tested in a VM with master, both with newly created users and then users with sections changed. Tested for both students and graders.

### DIFF
--- a/site/app/controllers/admin/UsersController.php
+++ b/site/app/controllers/admin/UsersController.php
@@ -823,13 +823,13 @@ class UsersController extends AbstractController {
          *
          * @param string $action "insert" or "update"
          */
-        $insert_or_update_user_function = function ($action, $user) use (&$semester, &$course, &$uploaded_data, &$return_url) {
+        $insert_or_update_user_function = function ($action, $user) use (&$semester, &$course, &$row, &$return_url) {
             try {
                 switch ($action) {
                     case 'insert':
                         //User must first exist in Submitty before being enrolled to a course.
-                        //$uploaded_data[0] = authentication ID.
-                        if (is_null($this->core->getQueries()->getSubmittyUser($uploaded_data[0]))) {
+                        //$row[0] = authentication ID.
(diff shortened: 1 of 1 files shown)


site/app/controllers/admin/UsersController.php
```diff
@@ -823,13 +823,13 @@ public function uploadUserList($list_type = "classlist") {
...
-        $insert_or_update_user_function = function ($action, $user) use (&$semester, &$course, &$uploaded_data, &$return_url) {
+        $insert_or_update_user_function = function ($action, $user) use (&$semester, &$course, &$row, &$return_url) {
             try {
                 switch ($action) {
                     case 'insert':
                         //User must first exist in Submitty before being enrolled to a course.
-                        //$uploaded_data[0] = authentication ID.
-                        if (is_null($this->core->getQueries()->getSubmittyUser($uploaded_data[0]))) {
+                        //$row[0] = authentication ID.
+                        if (is_null($this->core->getQueries()->getSubmittyUser($row[0]))) {
```
MasterOdin: Is there a case where `$row[0] !== $user->getId()`?

MasterOdin: So looking at the usages of this closure and the binded `$row` variable, it's only used for this snippet of code:
https://github.com/Submitty/Submitty/blob/7e62e32d4bcefde838cccc7ea7bb25a99af7bfe2/site/app/controllers/admin/UsersController.php#L947-L964

which looks like the value of `$row[0] === $user->getId()` so let's rip out the `$row` variable altogether from the closure and make this a bit simpler and less error prone.

PR #4927, opened 2020-01-15
Upload peer graders

jchulton:
Allows an instructor to upload a csv file containing peer grader pairs for a peer assignment:
1. This option only appears for a peer assignment
2. The csv file is expected to contain a header row. The csv file must contain a student row and a grader row. These rows do not have to be in a particular order. The file can contain more rows, they will be ignored.
3. The page will display the current peer grader pairs, this will be updated when the user uploads a new file and refreshes the page (TODO: see if this can be updated without refreshing the page).

image


site/app/models/gradeable/Gradeable.php
```diff
@@ -133,6 +134,9 @@ class Gradeable extends AbstractModel {
      * that contains filename, file path, and the file size.
      */
     private $split_pdf_files = null;
+    
+    /* idk why, but pylint doesn't like me commenting the type ):*/
```
MasterOdin: it's phpstan, and it's because `@property` is a phpDoc construct that we're not using properly (see https://manual.phpdoc.org/HTMLSmartyConverter/PHP/phpDocumentor/tutorial_tags.property.pkg.html). You can use the new `@prop` instead to get the same functionality:
`/** @prop @var array */`

PR #4843, opened 2019-12-08
[Feature:Submission] keep previous files uploaded

mukul-kmr-jha:
### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)

### What is the current behavior?
Closes #4837 
### What is the new behavior?
*Keep Previous Files uploaded* option is now removed and the previous files are now kept in bucket by default.

PR #4954, opened 2020-01-22
[Feature:PHP] Add support for gating features behind feature flags

### LINKED ISSUE #4923: Create policy on feature flags for in-progress work
With #4912 and #4918, work is going again into fleshing out, fixing, and generally finishing peer grading. However, during this work, it's expected that the options will exist, but that instructors are heavily instructed to not use them (as they may break, change, or whatever). It would be good to come up with a policy of using [feature flags](https://martinfowler.com/articles/feature-toggles.html) such that it might be possible to turn a feature on for a specific course, only under debug mode, etc. so that we do not have to worry about instructors breaking their courses in production.

This would also apply to the stuff for the automatic rainbow grade generation as well.

MasterOdin:

Closes #4923. 

New features that are under-development, but not ready for production are still shown to instructors with a warning to not use them. This is not ideal as instructors could still use them and cause things to break in unexpected ways.

### What is the new behavior?
Can gate features behind feature flags where in the PHP code:
```php
if ($this->core->getConfig()->checkFeatureFlagEnabled('flag_name')) { }
```
and in twig:
```twig
{% if feature_flag_enabled('flag_name') %}
```

Feature flags are on by default if debug mode is enabled (such as within vagrant) or if in the course config file, a top-level key "feature_flags" object is added where the value is a map where each key is a feature flag and the value is boolean true/false indicating if it's enabled. The feature flag is only considered enabled if they key exists and it is true, ensuring full backwards compatibility with existing courses. It is fine for the course config file to also have old flags no longer used, assuming that we always use a unique name for each flag to prevent accidental collisions.

### Other information?

Not a breaking change, and does not require any migrations to take advantage of it.

### DIFF
--- a/site/app/libraries/Output.php
+++ b/site/app/libraries/Output.php
@@ -117,9 +117,12 @@ HTML;
             if ($this->core->getConfig()->wrapperEnabled()) {
                 $this->twig_loader->addPath(
                     FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), 'site'),
-                    $namespace = 'site_uploads'
+                    'site_uploads'
                 );
             }
+            $this->twig->addFunction(new \Twig\TwigFunction("feature_flag_enabled", function(string $flag): bool {
+                return $this->core->getConfig()->checkFeatureFlagEnabled($flag);
+            }));
         }
         $engine = new ParsedownEngine();
         $engine->setSafeMode(true);
--- a/site/app/libraries/Output.php
+++ b/site/app/libraries/Output.php
@@ -120,7 +120,7 @@ HTML;
                     'site_uploads'
                 );
             }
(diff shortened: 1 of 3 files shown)


MasterOdin:
@bmcutler This is ready to go and merge in my opinion, and have put the plagiarism tool behind it. To make it available on a course, you would (per the documentation here), add:
```
{
  "feature_flags": {
    "plagiarism": true
  },
  other settings
}
```

to a course config.json file at that top level. Please let me know your thoughts on the documentation, or assign someone you think should review this before merging.

MasterOdin:
@emaicus What is the glitch when upgrading?

emaicus:
Here is a screenshot of the error:

robot

MasterOdin:
Alright, that error would be a result of going from master to this branch without having merged master into it first, as that setting was removed in #4843.

PR #4973, opened 2020-01-28
[Bugfix:Autograding] Properly Acquire Lock in Get Job

emaicus:
Updates the ```get_job``` function to use a ```with``` statement for lock acquisition rather than the ```lock.acquire``` and ```lock.release``` functions. 

Previously, if acquiring a job failed with an exception, the lock would never be released, causing all of autograding to hang until restarted. The with statement will cause the lock to be released in such cases.


MasterOdin:
I think the PR title would more appropriately be "[Bugfix:Autograding] Properly release job lock on exceptions".

PR #4977, opened 2020-01-28
[Feature:TAGrading] Add Peer Panel to TA Grading Interface

### LINKED ISSUE #4876: Create a "Peer Grading" panel on the TA Grading Page
Create a new panel on the peer grading page which displays the peer results of a submission to Instructors and Full Access Graders. This panel should not be present on non-peer gradeables. This design will be revisited after peer grading is fully implemented.

emaicus:
Closes #4876 

This PR adds a peer grading panel to the TA grading interface. This panel can be accessed by instructors and full access graders to view which marks have been selected by which peers for a given student.


emaicus:
peer_panel

Peer Panel on right.

jchulton:
This interacts very oddly with limited access graders. When a limited access grader changes an instructor's graded peer marks, those changes do not show up for the instructor. The changes do show up for limited access graders, so there are now two graded versions. The student seems to see the instructor grades. I also got a component Id not found error when trying to use the verify all button to verify non-peer marks if there are also peer marks in the gradeable

site/app/templates/grading/electronic/NavigationBar.twig
```diff
@@ -24,6 +24,9 @@
         {% if not peer %}
             <button title="Show/Hide Student Information (Press S)" class="invisible-btn" onclick="toggleInfo(); updateCookies();"> <i class="fas fa-user icon-header" ></i> </button>
         {% endif %}
+        {% if peer %}
```
MasterOdin: `{% else %}`

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -50,7 +50,7 @@ public function statusPage(
     ) {
 
         $peer = false;
-        if ($gradeable->isPeerGrading() && $this->core->getUser()->getGroup() == User::GROUP_STUDENT) {
+        if ($gradeable->isPeerGrading()) {//&& $this->core->getUser()->getGroup() == User::GROUP_STUDENT) {
```
MasterOdin: Delete the commented out part here, or add a comment as to why you're leaving it commented out.

site/public/js/ta-grading-rubric.js
```diff
@@ -194,14 +194,15 @@ function ajaxGetComponentRubric(gradeable_id, component_id) {
...
  * @return {Promise} Rejects except when the response returns status 'success'
  */
-function ajaxGetGradedGradeable(gradeable_id, anon_id) {
+function ajaxGetGradedGradeable(gradeable_id, anon_id, all_peers) {
     return new Promise(function (resolve, reject) {
         $.getJSON({
             type: "GET",
             async: AJAX_USE_ASYNC,
-            url: buildCourseUrl(['gradeable', gradeable_id, 'grading', 'graded_gradeable']) + `?anon_id=${anon_id}`,
+            url: buildCourseUrl(['gradeable', gradeable_id, 'grading', 'graded_gradeable']) + `?anon_id=${anon_id}&all_peers=${all_peers}`,
```
MasterOdin: `all_peers.toString()` to make it the action here explicit as to what should happen with the boolean, to avoid any potential surprises on implicit casting.

site/public/templates/grading/EditGradeable.twig
```diff
@@ -17,7 +17,8 @@
 {% endblock %}
 
 {% block footer_content %}
-    <input type="button" class="btn btn-primary" value="Add New Component" onclick="onAddComponent()" />
+    <input type="button" class="btn btn-primary" value="Add New Component" onclick="onAddComponent(false)" />
+    <input type="button" class="btn btn-danger" value="(DO NOT USE) Add New Peer Component" onclick="onAddComponent(true)"  />
```
MasterOdin: I think we should wait to merge this until after #4954 is merged and then wrap this with a feature flag:
```
{% if feature_flag_enabled('peer_grading') %}
    <input type="button" class="btn btn-danger" value="Add New Peer Component" onclick="onAddComponent(true)"  />
{% endif %}
```

as I have to imagine somebody is going to try and use this even with the warning otherwise.

emaicus:
@jchulton, Working through your feedback. As for the verify all bug, that appears to be on master, not just my branch.

emaicus:
Just made an issue.

emaicus:
@jchulton, I cannot seem to reproduce the behavior you are experiencing in regards to limited access graders. Could you post a set of steps you took to encounter the problem?

jchulton:
@emaicus Realizing the issue is different than I described, sorry for the confusion. The steps and a better description of the issue are below:
1.  Create a new gradable with some regular components and peer components that can be graded by a limited access grader.
2. As an instructor, push the gradable to Grading in progress and submit for a student.
2. As a limited access grader, grade every component for that student, including the peer components. For example, give a peer component 2/7
3. As an instructor, grade a peer component graded by the limited access grader for the same student in the grading rubric. For example, give the peer component 0/7
4. Now there are three different scores being displayed: the limited access grader sees 2/7, the instructor sees 0/7 in the grading rubric and 1/7 in the peer grading. The 1/7 is the final grade shown to the student.

He's the instructor's view:
image


My understanding was that the instructor should override the limited access grader, so the final grade should be 0/7 instead of 1/7. It looks like the average is being taken. The 2/7 and 0/7 make sense as those amount to different peer grades. The instructor override may be something we work on in a future pull request though.

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1024,6 +1035,46 @@ public function renderRubricPanel(GradedGradeable $graded_gradeable, int $displa
...
+     * Render the Grading Rubric panel
+     * @param GradedGradeable $graded_gradeable
+     * @return string
+     */
+    public function renderPeerPanel(GradedGradeable $graded_gradeable, int $display_version) {
+        $return = "";
+        $gradeable = $graded_gradeable->getGradeable();
+
+        // Disable grading if the requested version isn't the active one
+        // $grading_disabled = $graded_gradeable->getAutoGradedGradeable()->getActiveVersion() == 0
```
jchulton: Can the commented-out code be removed?

emaicus:
@jchulton, That is the intended behavior. Each grader sees the world from their own perspective, and the peer panel shows it as it actually is (that is, peer 1 see's that they gave the student 1 point, peer 2 sees that they gave the student 2 points, and the peer panel shows that the student has received the average, 1.5 points).

site/app/views/grading/ElectronicGraderView.php
emaicus: Fixed

PR #4978, opened 2020-01-29
[Feature:Forum] Adds Markdown Templates

matgar1997:
When selecting the markdown option it allows you to add code blocks and URLs to a discussion forum post.

### What is the new behavior?
In addition to the previous options, it now includes bold and italics

### Other information?
This should not break anything since its text in the discussion forum text box. Worst case if they mess with the spacing the markdown may not work as expected but by default, there are spaces inserted between so it would not be an issue unless they delete any of the preformatted inserted text

### DIFF
--- a/site/app/templates/forum/ThreadPostForm.twig
+++ b/site/app/templates/forum/ThreadPostForm.twig
@@ -27,6 +27,8 @@
         <span style="display: {{ render_markdown?"block":"none" }}" id="markdown_buttons_{{ post_box_id }}">
             <button id="link_btn" type="button" title="Insert a link" onclick="addMarkdownCode(1, $(this).closest('.thread-post-form').find('[name=thread_post_content]'))" style="margin-right:10px;" class="btn btn-default">Link <i class="fas fa-link fa-1x"></i></button>
             <button id="code_btn" title="Insert a code segment" type="button" onclick="addMarkdownCode(0, $(this).closest('.thread-post-form').find('[name=thread_post_content]'))" class="btn btn-default">Code <i class="fas fa-code fa-1x"></i></button>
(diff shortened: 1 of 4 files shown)


matgar1997:
Screen Shot 2020-01-29 at 2 29 28 PM

site/app/templates/forum/ThreadPostForm.twig
```diff
@@ -26,7 +26,9 @@
...
-            <button id="code_btn" title="Insert a code segment" type="button" onclick="addMarkdownCode(0, $(this).closest('.thread-post-form').find('[name=thread_post_content]'))" class="btn btn-default">Code <i class="fas fa-code fa-1x"></i></button>
+            <button id="code_btn" title="Insert a code segment" type="button" onclick="addMarkdownCode(0, $(this).closest('.thread-post-form').find('[name=thread_post_content]'))" style="margin-right:10px;" class="btn btn-default">Code <i class="fas fa-code fa-1x"></i></button>
+            <button id="bold_btn" title="Insert bold text" type="button" onclick="addMarkdownCode(2, $(this).closest('.thread-post-form').find('[name=thread_post_content]'))" style="margin-right:10px;" class="btn btn-default">Bold <i class="fas fa-bold fa-1x"></i></button>
+            <button id="italic_btn" title="Insert italic text" type="button" onclick="addMarkdownCode(3, $(this).closest('.thread-post-form').find('[name=thread_post_content]'))" style="margin-right:10px;" class="btn btn-default">Italics <i class="fas fa-italic fa-1x"></i></button>
```
MasterOdin: Let's make a class `markdown_btn` or `md_btn` that has the property `margin-right: 10px` and give it to all buttons. Stick it in forums.css.

matgar1997: Not sure what I'm missing, where is the file forums.css. I only see it under addInternalCss

PR #4980, opened 2020-01-29
[Bugfix:UI] Fixes buttons not following css styling on ios

elihschiff:

ios buttons don't follow the css style and so they are extra small and look different. I believe this only happens for buttons inside forms.

image

### What is the new behavior?

This should make them ignore the built in css and therefore look the same as they do on computers and android phones

### Other information?

I did not test this. I dont have an iphone and so I have no way to test this. I found this fix here https://stackoverflow.com/questions/5449412/styling-input-buttons-for-ipad-and-iphone
and was hoping somebody with an iphone can test this for me.

I have been told this can also be tested on the ios simulator on an apple computer.

### DIFF
--- a/site/public/css/global.css
+++ b/site/public/css/global.css
@@ -154,3 +154,7 @@ footer .footer-separator {
 #loading-bar-percentage{
     margin-left: 10px;
 }
+
+input{
+  -webkit-appearance: none;
+}
--- a/site/public/css/global.css
+++ b/site/public/css/global.css
@@ -154,7 +154,3 @@ footer .footer-separator {
 #loading-bar-percentage{
     margin-left: 10px;
 }
-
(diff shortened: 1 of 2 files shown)


matgar1997:
On the iOS simulator, it almost matches the version on a computer except that finish helping is green. Other than that it does seem to override the built-in CSS.
 
Simulator Screen Shot - iPhone 11 - 2020-02-02 at 19 40 42

elihschiff:
@matgar1997 the green color is actually a feature and unrelated to this PR, although I admit the reason for the colors are not that clear. I have changed that up a bit in #5001 

As for this PR, because the button looks correct and not like the old ios buttons, I think it should be ok to merge as it fixes the issue.

PR #4981, opened 2020-01-29
[Bugfix:Forum] Fix auto population of default categories on first enable

### LINKED ISSUE #4932: Fix buggy implementation of default categories for the forum
**Describe the bug**
When enabling the forum, its handling of default categories does not work as expected, not creating the categories if they're missing, and causing an exception to be thrown if they do.

**Expected behavior**
When the forums are enabled, and there are no existing categories, which only happens on the first time it is enabled, then the default categories should be created.

**To Reproduce**
To reproduce the exception being thrown, you can do the following sequence:
1. Enable forum
2. Add category "Tests"
3. Disable and then attempt to re-enable forum

**Additional context**
The offending bit of code is the following:
https://github.com/Submitty/Submitty/blob/ff7c4b2af7a41ba52d29da4c5c5ab8941bf91169/site/app/controllers/admin/ConfigurationController.php#L188-L201

Where the line `ForumUtils::isValidCategories($rows, -1, array($category))` returns true if the category exists, otherwise false, which is the root cause of the exception, but broadly misses the above point of not being in-effect on subsequent toggles of the forum.

That code should be modified such that it follows this code (the permission check from above is dropped as it's trivially true for anyone who can get to the configuration page):
```
if ($name === 'forum_enabled' && $entry == 1) {
    if (empty($this->core->getQueries()->getCategories())) {
        foreach ([["General Questions", "Homework Help", "Quizzes" , "Tests"] as $category) {
            $this->core->getQueries()->addNewCategory($category);
        }
    }
}
```

As part of this issue, two test cases must be added to the ConfigurationControllerTester, one for enabling the forum where `getCategories()` returns `[]` and one where it does not and that then the `addNewCategory` function is called appropriately (tested using mocks).

shwehtom89:

Closes #4932

### What is the new behavior?
Default categories are added if there exist no categories currently in the Forum.
If there exists categories and the forum is re-enabled, default categories are not added.
Messages prompting instructors to create default categories were removed.

### Other information?
PHP unit tests were added for the CourseConfigurationController.
Tests were added for empty categories and existing categories.


site/tests/app/controllers/admin/ConfigurationControllerTester.php
```diff
@@ -379,4 +379,72 @@ public function testUpdateConfigNonWriteable() {
...
+        $queries
+            ->expects($this->exactly(0))
+            ->method('addNewCategory');
+        $core->setQueries($queries);
+        $response = $controller->updateConfiguration();
+        $this->assertNotNull($response->json_response);
+        $_POST['name'] = 'default_hw_late_days';
+        $_POST['entry'] = '2';
+        $controller = new ConfigurationController($core);
+        $response = $controller->updateConfiguration();
```
MasterOdin: Let's delete this.

site/tests/app/controllers/admin/ConfigurationControllerTester.php
```diff
@@ -379,4 +379,72 @@ public function testUpdateConfigNonWriteable() {
...
+                [$this->equalTo('General Questions')],
+                [$this->equalTo('Homework Help')],
+                [$this->equalTo('Quizzes')],
+                [$this->equalTo('Tests')]
+            )
+            ->will($this->onConsecutiveCalls(0, 1, 2, 3));
+
+        $core->setQueries($queries);
+        $response = $controller->updateConfiguration();
+        $this->assertNotNull($response->json_response);
```
MasterOdin: Let's add:
```
        $expected = [
            'status' => 'success',
            'data' => null
        ];
        $this->assertEquals($expected, $response->json_response->json);
```

to this test.

site/tests/app/controllers/admin/ConfigurationControllerTester.php
```diff
@@ -379,4 +379,72 @@ public function testUpdateConfigNonWriteable() {
...
+        $config->loadCourseJson('f19', 'sample', $this->course_config);
+        $core->setConfig($config);
+        $_POST['name'] = 'forum_enabled';
+        $_POST['entry'] = 'true';
+        $queries = $this->createMock(DatabaseQueries::class);
+        $queries
+            ->expects($this->once())
+            ->method('getCategories')
+            ->with()
+            ->willReturn(['General Questions', 'Homework Help', 'Quizzes', 'Tests']);
```
MasterOdin: Let's have this return just `['Category']` to show that it does not make any calls even if none of the default categories exist.

site/app/controllers/admin/ConfigurationController.php
```diff
@@ -184,16 +184,15 @@ public function updateConfiguration(): Response {
...
-                            $this->core->getQueries()->addNewCategory($category);
-                        }
+        if ($name === 'forum_enabled' && $entry == 1) {
+            $categories = ["General Questions", "Homework Help", "Quizzes" , "Tests"];
+            $rows = $this->core->getQueries()->getCategories();
+
+            // Only create default categories when the forum is first enabled
+            if (empty($rows)) {
+                foreach ($categories as $category) {
+                    if (!ForumUtils::isValidCategories($rows, -1, array($category))) {
```
MasterOdin: We can get rid of this check, as the function will always trivially return false with `$rows` being empty.

site/app/templates/forum/ShowForumThreads.twig
```diff
@@ -26,21 +26,8 @@
...
-                </h4>
-                <a class= "text-center" id="forum-enabled-message" href="{{ manage_categories_url }}" alt="Customize Categories">Customize categories.</a>
-
-            {% else %}
-                <h4 class="text-center red-message">No categories have been created yet. Please contact your instructor.</h4>
-
-            {% endif %}
-        {% else %}
-            {%  include "forum/ForumBar.twig" with {
+    {% if thread_exists == false %}
```
andrewaikens87: Could use `if not thread_exists`

jchulton:
Tested and working well.

bmcutler:
@shwehtom89 
Please address TravisCI failures

PR #4996, opened 2020-01-31
[Bugfix:InstructorUI] Add user modal does not clear on close

### LINKED ISSUE #4991: Hitting close on add user modal does not clear modal
**Describe the bug**
When opening the modal to add a user, use auto-complete the find a user and fill out all fields, hit close, and then hit the button to open the modal, the modal is still filled totally in with the previous user's information.

**Expected behavior**
The second time I open the modal, all fields are blank.

**To Reproduce**
1. Open new student modal
2. Type in "pearsr" and then tab to auto-complete all fields
3. Hit close on modal
3. Hit button to re-open modal

**Configuration**
 - Browser [e.g. chrome, safari] Chrome, Safari

**Additional context**
If we are afraid of losing data by accident when hitting the close button, we should add a mechanism to warn a user then that hitting close will lose the data, and not have it be that they can just re-open the modal and keep working on that student.

mukul-kmr-jha:

The form modal opened by the `New Student` button keeps the data filled by the auto-completion feature even after hitting `Close` button 
Similarly, the `New Grader` form modal is storing the previous user's information.

### What is the new behavior?
The state of form is being reset on close event. Closes #4991

### DIFF
--- a/site/public/js/userform.js
+++ b/site/public/js/userform.js
@@ -259,18 +259,36 @@ function completeUserFormInformation(user) {
     }
 }
 
+function isUserFormEmpty() {
+  let form = document.querySelectorAll('form')[0];
+  let isFormEmpty = true;
+  for (const formElement of form.elements) {
+    const {name, value, type, checked} = formElement;
+    // skipping all the hidden fields, user-group and manual-registration
+    if (type === 'hidden' || name === 'user_group' || name === 'manual_registration') {
+      continue;
+    }
+    else if ((type === "text" && value.length) || (type === "select-one" && value !=='null') || checked){
+      isFormEmpty = false;
+      break;
+    }
+  }
+  return isFormEmpty;
+}
+
 function clearUserFormInformation() {
-    var form = $("#edit-user-form");
-    $('[name="user_numeric_id"]', form).val("");
-    $('[name="user_firstname"]', form).val("");
-    $('[name="user_preferred_firstname"]', form).val("");
-    $('[name="user_lastname"]', form).val("");
-    $('[name="user_preferred_lastname"]', form).val("");
-    $('[name="user_email"]', form).val("");
-    $('[name="registered_section"] option[value="null"]', form).prop('selected', true);
-    $('[name="rotating_section"] option[value="null"]', form).prop('selected', true);
-    $('[name="manual_registration"]', form).prop('checked', true);
-    $("[name='grading_registration_section[]']").prop('checked', false);
+    let form = document.querySelectorAll('form')[0];
+    for (const formElement of form.elements) {
+      const {name, value, type, checked} = formElement;
+      if (type === "text") {
+        formElement.value = "";
+      } else if (type === "select-one") {
+        formElement.value = "null";
+      } else if (type === "checkbox") {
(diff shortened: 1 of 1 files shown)


site/public/js/userform.js
```diff
@@ -204,9 +204,11 @@ function autoCompleteOnUserId(user_information) {
 }
 
 function completeUserFormInformation(user) {
-    var form = $("#edit-user-form");
+  var form = $("#edit-user-form");
+
+  $('[name="edit_user"]', form).val("true");
 
-    $('[name="user_numeric_id"]', form).val(user['user_numeric_id']);
+  $('[name="user_numeric_id"]', form).val(user['user_numeric_id']);
```
elihschiff: It looks like you removed some white-space here and that messes with the indentation. I think it looks better with the original white-space where everything lines up.

site/public/js/userform.js
```diff
@@ -204,9 +204,11 @@ function autoCompleteOnUserId(user_information) {
 }
 
 function completeUserFormInformation(user) {
-    var form = $("#edit-user-form");
+  var form = $("#edit-user-form");
```
elihschiff: It looks like you removed some white-space here and that messes with the indentation. I think it looks better with the original white-space where everything lines up.
This also will require fixing the new line 209 to be at the same indentation level.

site/public/js/userform.js
```diff
@@ -51,7 +51,6 @@ function newUserForm() {
     $("#edit-student-modal-title").css('display','none');
     $("#edit-grader-modal-title").css('display','none');
     $("#user-form-already-exists-error-message").css('display','none');
-    $('[name="edit_user"]', form).val("false");
```
elihschiff: I do not think you want to remove this line. With your code, if you edit a user, press cancel, then add a new user. The edit_user value will be true. 
Lines 208/209 might also relate to this issue.

site/public/js/userform.js
```diff
@@ -282,11 +305,12 @@ function clearValidityWarnings() {
 }
 
 function closeButton() {
-    $('#edit-user-form').css('display', 'none');
-    if($('[name="edit_user"]').val() == 'true') {
+    // If form contain some data, prompt the user for data loss
+    if(!isUserFormEmpty() && confirm('Changes to the Form will be lost!')){
```
MasterOdin: `Form` should lowercased to `form`

site/public/js/userform.js
```diff
@@ -282,11 +305,12 @@ function clearValidityWarnings() {
...
         $('[name="user_id"]', $("#edit-user-form")).val("");
         clearUserFormInformation();
     }
+    $('#edit-user-form').css('display', 'none');
```
MasterOdin: I think we want that this should only happen if we confirm. Right now, if we hit cancel after we get the confirm box, the modal will still close, but the data will remain the next time we open the modal.

MasterOdin:
Close to merging this (thanks for the patience on this), just one last thing.

site/public/js/userform.js
mukul-kmr-jha: Added changes and rebased.
PTAL, Thanks!

MasterOdin:
Thanks for your work on this!

PR #5001, opened 2020-02-01
[Feature:InstructorUI] office hours queue improvements

elihschiff:

There are a number of smaller bugs that needed to be fixed. As well as the entire queue page gets very slow when the history grows really large.

### What is the new behavior?
- The queue history is now limited to 25 entries with the ability to see all of them if you click a button. (See image below)
- The timers that keep track of how long you have been helping somebody only run on visible rows in the queue (I think this will improve performance)
- There is now a restore button that moves somebody from the queue history back to the queue, just in case there was a mistake.
image
- Queues now have a name and a code. So you can have one office hours queue but the ability to switch out the code so that people have to be in the room to sign up.
image
- Students that removed themselves from the queue no longer show their user id under the removed by column, it is now the clickable eye button
image
- Every button on the page can now be interacted with by the tab key to help with accessibility
- The green button now means that you were the one who started to help a user
image
- A star next to the help button means that a user has not been in that queue yet today, this can be used in really busy office hours to help prioritize students.
image
- The queue now auto refreshes every 30 seconds for ta/mentors, if there is a change.
- Fixed a bug where the numbers on the student queue were were wrong
- Fixed a bug where the mentor numbers were wrong when they turned on a filter
- Fixed a bug where the queue empty error message was always visible even when the queue was not empty

### Other information?

~~I made a lot of changes so I have a feeling this wont be merged by Wednesday the 5th for the next set of DS and CS1 labs. If this is in fact the case can you let me know with enough time to put in a much much smaller PR with just the change that makes the queue history limited to 25 (with no ability to see more) as that is really needed so that mentor's phones don't crash for the next lab day. That PR would only change a few lines and should be much easier to approve.~~
Edit: This has been done and merged


site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -32,11 +32,21 @@ public function showQueue() {
...
 
+        $queries = array();
+        parse_str($_SERVER['QUERY_STRING'], $queries);
+
+        $full_history = false;
+        if (array_key_exists('full_history', $queries) && $queries['full_history'] === "true") {
+            $full_history = true;
+        }
+
+        $this->core->getOutput()->addInternalCss('officeHoursQueue.css');
```
zjxiaohan: This line for adding CSS needs to be put in `OfficeHoursQueueViewer` to conform with MVC pattern.

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -32,11 +32,21 @@ public function showQueue() {
             );
         }
 
+        $queries = array();
+        parse_str($_SERVER['QUERY_STRING'], $queries);
```
zjxiaohan: You don't need to parse the query string yourself. Adding a `full_history` argument in the function signature is fine if all you need is to parse `/office_hours_queue?full_history=true`. Please have a look at https://submitty.org/developer/router#route-with-parameters

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -289,4 +342,62 @@ public function deleteQueue($queue_code) {
             new RedirectResponse($this->core->buildCourseUrl(['office_hours_queue']))
         );
     }
+
+    /**
+    * @Route("/{_semester}/{_course}/office_hours_queue/checkUpdates", methods={"GET"})
```
zjxiaohan: `checkUpdates` is inconsistent with `change_token` and `current_queue` below.

```suggestion
    * @Route("/{_semester}/{_course}/office_hours_queue/check_updates", methods={"GET"})
```

site/app/controllers/OfficeHoursQueueController.php
elihschiff: I put it in the OfficeHoursQueueViewer constructor, is that the expected place I should have put it?

zjxiaohan: Oh sorry I misunderstood the code. It is supposed to be in `OfficeHourQueueView`.

zjxiaohan: https://github.com/Submitty/Submitty/blob/90437c056fb7fa8d6b5de2081df52d4a376d3b64/site/app/views/OfficeHoursQueueView.php#L11

here

elihschiff: Wait, I think you may have meant to say it should be put in `OfficeHoursQueueView`, am I correct?

`OfficeHoursQueueView` is the view for the office hour queue.
`OfficeHoursQueueViewer` is a model that stores some information for the person viewing the office hour queue, making them a "viewer".

Maybe it would make sense for me to rename `OfficeHoursQueueViewer` to `OfficeHoursQueueModel` or `OfficeHoursQueueUser`?

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -32,11 +32,21 @@ public function showQueue() {
...
+        if (array_key_exists('full_history', $queries) && $queries['full_history'] === "true") {
+            $full_history = true;
+        }
+
+        $this->core->getOutput()->addInternalCss('officeHoursQueue.css');
+
         return Response::WebOnlyResponse(
             new WebResponse(
                 'OfficeHoursQueue',                      //Goes to this file OfficeHoursQueueView.php
                 'showTheQueue',                          //Runs this functin showTheQueue()
```
zjxiaohan:
```suggestion
                'showTheQueue',                          //Runs this function showTheQueue()
```

zjxiaohan: Actually I feel the comments are not necessary here as the meaning of `new WebResponse(somearg)` is consistent throughout Submitty. There is nothing special about the usage here.

zjxiaohan:
I am happy with your changes regarding to my review, but I cannot approve this or review further because 1) I don't have Submitty environment installed in my machine so I can't test the features myself, 2) I am not familiar with the office hour queue thing.

You could request other people's review for this PR. :)

Issue #4537, opened 2019-09-23
Refactor Views to only have access to Output, not Core


MasterOdin:
The `View` class currently has access to the `Core` which allows people to mistakenly use it to do stuff like run queries or the like, which should be happening within the `Controller`. It is additionally used a container for getting at the current user and config, leading to writing tests for view functions a bit more difficult. To help make it easier to follow the proper structure, the `AbstractView` should be modified to only accept an instance of `Output`, and better enforce getting data to fill the view passed in from the Controller, not accessible through the `Core` super-global.

We will probably need to add some compatibility for the `buildUrl` and `buildCourseUrl` functions into `Output`, though I think it may make sense to just whole-sale move them from core, into `Output`.

MasterOdin:
For reference, here's a grep of Views that currently run a DB query in their logic:
```
grep -R "\$this\->core\->getQueries" app/views
app/views/NavigationView.php:            $gradeable_ids_and_titles = $this->core->getQueries()->getAllGradeablesIdsAndTitles();
app/views/NavigationView.php:            $teams = $this->core->getQueries()->getTeamsByGradeableId($gradeable->getId());
app/views/NavigationView.php:                    $this->core->getQueries()->getTeamViewedTime($graded_gradeable->getSubmitter()->getId(),$this->core->getUser()->getId()) === null) {
app/views/grading/ElectronicGraderView.php:        $team_gradeable_view_history = $gradeable->isTeamAssignment() ? $this->core->getQueries()->getAllTeamViewedTimesForGradeable($gradeable) : array();
app/views/grading/ElectronicGraderView.php:        $students = $this->core->getQueries()->getAllUsers();
app/views/grading/ElectronicGraderView.php:            $posts = $this->core->getQueries()->getPostsForThread($this->core->getUser()->getId(), $threadId, false, 'time', $submitter_id);
app/views/admin/LateDayView.php:        $students = $this->core->getQueries()->getAllUsers();
app/views/admin/ExtensionsView.php:        $students = $this->core->getQueries()->getAllUsers();
app/views/admin/ExtensionsView.php:        $users = $this->core->getQueries()->getUsersWithExtensions($current_gid);
app/views/admin/GradeOverrideView.php:        $students = $this->core->getQueries()->getAllUsers();
app/views/admin/PlagiarismView.php:               $title = $this->core->getQueries()->getGradeableConfig($saved_config['gradeable'])->getTitle();
[...]

PR #5002, opened 2020-02-02
[Feature:Forum] Instructor/TA/Mentor posts now are yellow in thread list

elihschiff:

Threads posted by Instructor/TA/Mentor show up exactly the same in the thread list as a normal student post (unless they are an announcement in which case they have a star)
image

### What is the new behavior?

Threads in the thread list now have the yellow border if they were made by an Instructor/TA/Mentor which matches how they look when you actually open the thread.
image

### Other information?

This was talked about at a DS mentor meeting last semester because it is not always clear when an important thread is posted. Anyone who uses the forum knows that they yellow border on a reply means it was made by course staff so it seems to me that it makes sense to have the sidebar follow this same format.

### DIFF
--- a/site/app/views/forum/ForumThreadView.php
+++ b/site/app/views/forum/ForumThreadView.php
@@ -619,6 +619,11 @@ class ForumThreadView extends AbstractView {
             if ($thread["deleted"]) {
                 $class .= " deleted";
             }
+
+            if($this->core->getQueries()->getUserById($thread['created_by'])->getGroup() <= 3){
+              $class .= " important";
+            }
+
             //fix legacy code
             $titleDisplay = $thread['title'];
 
--- a/site/app/views/forum/ForumThreadView.php
+++ b/site/app/views/forum/ForumThreadView.php
@@ -620,7 +620,7 @@ class ForumThreadView extends AbstractView {
                 $class .= " deleted";
             }
 
-            if ($this->core->getQueries()->getUserById($thread['created_by'])->getGroup() <= 3) {
+            if ($this->core->getQueries()->getUserById($thread['created_by'])->accessGrading()) {
                 $class .= " important";
             }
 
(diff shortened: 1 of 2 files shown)


site/app/views/forum/ForumThreadView.php
```diff
@@ -619,6 +619,11 @@ public function displayThreadList($threads, $filtering, &$activeThreadAnnounceme
             if ($thread["deleted"]) {
                 $class .= " deleted";
             }
+
+            if ($this->core->getQueries()->getUserById($thread['created_by'])->getGroup() <= 3) {
```
zjxiaohan: It is preferred to use constants instead of random numbers for readability.

```suggestion
            if ($this->core->getQueries()->getUserById($thread['created_by'])->getGroup() <= User::GROUP_LIMITED_ACCESS_GRADER) {
```

zjxiaohan: Also to conform with MVC architecture, it is recommended not to get queries in your view function. Instead, it is better to put the queries in controllers and pass them to the view.

It is an ongoing task to forbid people from doing queries in views.

https://github.com/Submitty/Submitty/issues/4537

shailpatels: Thanks for catching this @zjxiaohan ! It looks like there are a lot of other queries also being made in this file as well there.

 It might be worth merging this PR for now and making a separate PR to move all of them to the forum controller at once but I'm not too familiar with this area @andrewaikens87 @elihschiff  so not sure whats easier/makes more sense.

zjxiaohan: @shailpatels Thanks for the reply. I didn't notice that there are a lot of queries in this file... Doing it in separate PR(s) would be better as far as I am concerned.

elihschiff: I actually think I found an even better solution. I switched it to use the accessGrading() function which does exactly what I was doing before but is a member function of user so there are no random numbers.

zjxiaohan: @elihschiff Yeah true that's the best solution. I haven't committing code for a long time and forgot that we have `accessGrading`. :)

PR #5009, opened 2020-02-04
Revert "[Bugfix:UI] buttons not following css styling on ios (#4980)"

bmcutler:
This reverts commit 383baba1fc55790cf17312a1238e6c870b2b0e1b.

Revert "[Bugfix:UI] buttons not following css styling on ios (#4980)"

radio buttons were not displaying on mac Chrome, mac Safari or Linux Chrome (didn't test anything else)

Screen Shot 2020-02-04 at 12 19 13 AM

### DIFF
--- a/site/public/css/server.css
+++ b/site/public/css/server.css
@@ -182,10 +182,6 @@ input.disabled {
     background-color: var(--standard-light-gray);
 }
 
-input{
-  -webkit-appearance: none;
-}
-
 /****************************
  * TABLE STYLES
  ***************************/


elihschiff:
:fearful: sorry about that! Looks like I never encountered a radio button while testing so I did not catch this. I thought webkit css only impacted safari but I was wrong.
I believe this could be fixed going forward by this, but I will clearly need to do a lot more testing.
```css
input[type="button"]{
  -webkit-appearance: none;
}
```

There is another other option I only very recently figured out. The root cause of the issue is using <input type="submit" class="btn"...> however <button type="submit" class="btn"> does the exact same thing and looks the same even on ios

Issue #4779, opened 2019-11-13
Bugs with current loading implementation of annotation files


MasterOdin:
**Describe the bug**
The current system uses just the filename + _ + user_id for the name of the annotation files. This fails to work if we have underscores in the user_id (as it assumes the last split will always be full user_id) and also if we have the same filename at multiple points in a folder hierarchy (through git). If you have say `filename` in `folder1/` and `folder2/` and make annotations in one PDF, then when you open the other PDF, it will attempt to load the annotations for it.

**Expected behavior**
Annotation file should load properly and with the right annotations.

**Proposed Solution**

 I'd rather us just create a md5 hash of the filepath + grader_id and use that to store the annotations. The filepath and grader_id should be stored within the PDF for convenience when looking at the file by a human. For example, assume the following directory tree and annotations being done by `user_1`:

```
filename.pdf
filename2.pdf
folder/
  filename.pdf
```

I would expect the annotation folder to have:

```
md5("filename.pdf" + "user_1")
md5("filename2.pdf" + "user_1")
md5("folder/filename.pdf" + "user_1")
```

all of these are probably going to be unique. There's a chance of collisions, but it's probably not significant enough to worry about in the real world usage.

MasterOdin:
The original post here failed to account for nested folders and files, which causes an issue in that we do not want a tree of:
```
folder/
  folder_2/
    filename_2/
      grader_2.json
  filename_1/
    grader_1.json
```

as that can get messy to parse out and deal with. Instead. I'd like to just do an md5hash of (file_name + '_' + grader_id) and use that as the filename. The JSON structure would then get the file_name added to it for convenience of someone manually looking at these files, but I do not think anyone is.

PR #5016, opened 2020-02-05
[Bugfix:TAGrading] Fix error when loading pdfs from non-root submissions folders

### LINKED ISSUE #2622: PDF Annotation supports subfolders within submission folder
Currently, the annotator only supports PDF submissions with no subfolder. If, however, a gradeable has multiple sections, the pdf annotation cannot save and read annotations.

### LINKED ISSUE #5023: Clicking the +/- zoom buttons when grading a pdf does not work
**Describe the bug**
When you click +/- buttons to zoom in on a pdf it gives an error
image
It does work if you refresh the page after though

**Expected behavior**

No error it should just zoom in with no page refresh

**To Reproduce**
1. Open an exam with a pdf
2. click the +/- button

**Configuration**
Arch Linux Firefox

**Screenshots**
image

MasterOdin:

Closes #2622
Closes #5023

Fixes error when attempting to load PDFs for annotation when the PDFs were submitted via VCS or when within a sub-folder for regular submission. This was due to the code hard-coding the directory to top-level `submissions` folder when attempting to load the PDFs.

### What is the new behavior?

PDFs load as expected and can be annotated. As a side-effect of wiring the new changes up to the function signature, it looks like I fixed the zoom/rotation bug that existed before this PR.

### Other information?

Tested opening PDFs and annotating them when PDFs were submitted by both VCS as well as regular file upload.


bmcutler:
This doesn't seem to have fixed it. 
I tested on both MacChrome & MacFirefox.
If it matters the pdf is in a subdirectory of the checkout folder:  checkout/answers/foo.pdf

Screen Shot 2020-02-06 at 12 49 11 AM

MasterOdin:
Yeah, subfolders would break loading PDFs whether they were uploaded via file upload or VCS. Back to the proverbial drawing board once more to do even more rewriting of how this works.

jchulton:
Whelp I did some testing: not too much to add except this doesn't seem browser specific (breaks on chrome and firefox). The error seems to happen somewhere in either encodePDF in MiscController.php or in PDFAnnotateEmbeded.js, but that's just the pdf rendering chain so that's unsurprising

MasterOdin:
So fixed the errors above with subfolders. There remains a bug where if you have a PDF in a subfolder with the same name as a PDF elsewhere in that submission, the annotations will collide. Fixing that (and #4779) will be done in a different PR as it requires overhauling some other parts of it, as well as writing a migration, which I'd rather have this in sooner so at least Konstantin can grade.

PR #5018, opened 2020-02-06
[VPAT] Moved meta viewport tag inside <head> tag

### LINKED ISSUE #5000: VPAT — <meta> Tag Errors
As noted in an internal accessibility audit, there are errors with a `<meta>` tag.  This particular tag should not have a `name` attribute, but should have either an `itemprop` or `property` attribute.

Note that `itemprop` requires the element to be a property of an item.  q.v. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/itemprop

(I recommend using `property` for now)

Screen Shot 2020-01-31 at 8 47 06 PM

elihschiff:

Our meta tags dont adhere to the w3 standards because they are not in the \<head> part of the page

closes #5000

### What is the new behavior?

This fixes the issue by moving the problematic line to the global header template and then adding the option for pages to say they support the mobile viewport when rendering using `$this->core->getOutput()->useMobileViewport();`


site/app/controllers/GlobalController.php
```diff
@@ -364,7 +364,7 @@ public function header() {
         //else if(...){}
         //more Holidays go here!
 
-        return $this->core->getOutput()->renderTemplate('Global', 'header', $breadcrumbs, $wrapper_urls, $sidebar_buttons, $unread_notifications_count, $css->toArray(), $js->toArray(), $duck_img);
+        return $this->core->getOutput()->renderTemplate('Global', 'header', $breadcrumbs, $wrapper_urls, $sidebar_buttons, $unread_notifications_count, $css->toArray(), $js->toArray(), $duck_img, $use_mobile_viewport);
```
MasterOdin: Let's simplify this by removing passing `$use_mobile_viewport` from Output -> Controller -> View and just have it be the view calls `$this->output->useMobileViewport()` (see below for comment about function changes).

site/app/views/GlobalView.php
```diff
@@ -52,7 +52,8 @@ public function header($breadcrumbs, $wrapper_urls, $sidebar_buttons, $notificat
             "system_message" => $this->core->getConfig()->getSystemMessage(),
             "csrf_token" => $this->core->getCsrfToken(),
             "enable_banner" => $config_data['duck_special_effects'],
-            "duck_img" => $duck_img
+            "duck_img" => $duck_img,
+            "use_mobile_viewport" => $use_mobile_viewport
```
MasterOdin: To follow-up from above, get rid of the `$use_mobile_viewport` from above, and have this just be `"use_mobile_viewport" => $this->output->useMobileViewport()`

bmcutler:
please resolve conflicts with master

elihschiff:
@bmcutler it should be able to merge now

site/app/controllers/GlobalController.php
```diff
@@ -9,7 +9,7 @@
 
 class GlobalController extends AbstractController {
 
-    public function header() {
+    public function header($use_mobile_viewport) {
```
MasterOdin: We can remove this variable now.

site/app/libraries/Output.php
```diff
@@ -385,7 +386,7 @@ public function getOutput() {
 
     protected function renderHeader() {
         if ($this->use_header) {
-            return $this->controller->header();
+            return $this->controller->header($this->use_mobile_viewport);
```
MasterOdin: We can remove this variable now.

PR #5031, opened 2020-02-12
[Feature:Forum] Added split post functionality

### LINKED ISSUE #4853: Unmerge / Split thread
2 use cases:

1.  Instructor has incorrectly merged two threads (can happen if the search is used incorrectly).

2.  Student has asked a new, good question inside of another thread, that really should have been a new thread.

Proposed interface..   When viewing all of the posts within a thread, an instructor (or full access grader?) can select one of the posts in the thread and press a button "Split into new thread", which will take that post, and any direct replies (that subtree) and create a new top level thread.

Note:  Need to test a bit and confirm that the thread ordering can handle this a new thread (with higher thread #) for a post that may have an earlier initial creation date that some threads with lower thread #.

DefinitiveAbove:

Currently you cannot split posts into new threads.

### What is the new behavior?
Closes #4853
You can now split posts into new top level threads, changing the title and categories as needed.

If it was a previously merged thread, it will reuse the old information from the database.
image
image

### Other information?
Currently has an issue where posts with identical timestamps may produce unpredictable behavior (due to difference in ordering in the database after editing post information).

Not a direct issue with this PR (existed before this PR), but will be fixed in a future PR.


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -452,6 +452,55 @@ public function getCategoriesIdForThread($thread_id) {
...
+                    $this->course_db->query("UPDATE threads set merged_thread_id=-1, merged_post_id=-1 where id=?", array($thread_id));
+                    $this->course_db->query("DELETE FROM thread_categories where thread_id=?", array($thread_id));
+                }
+                else {
+                    $this->course_db->query("INSERT INTO threads (title, created_by, is_visible, lock_thread_date) VALUES (?, ?, ?, ?)", array($title, $post["author_user_id"], true, null));
+                    $this->course_db->query("SELECT MAX(id) as max_id from threads where title=? and created_by=?", array($title, $post["author_user_id"]));
+                    $thread_id = $this->course_db->rows()[0]["max_id"];
+                }
+                foreach ($categories_ids as $id) {
+                    $this->course_db->query("INSERT INTO thread_categories (thread_id, category_id) VALUES (?, ?)", array($thread_id, $id));
```
andrewaikens87: You should be able to limit this to one insert via: 
```
 INSERT INTO table (column1, column2, …) VALUES (value1, value2, …),   (value1, value2, …) ,...;
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -452,6 +452,55 @@ public function getCategoriesIdForThread($thread_id) {
...
+                }
+                while (count($posts) > 0) {
+                    $check_posts = array();
+                    $str = "";
+                    for ($i = 0; $i < count($posts); $i++) {
```
andrewaikens87: nit: It does not look like `$i` is being used anywhere other than accessing `$posts`. Let's use a foreach here.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -452,6 +452,55 @@ public function getCategoriesIdForThread($thread_id) {
...
+                $this->course_db->query("SELECT id from threads where merged_post_id = ?", array($post_id));
+                $thread_id = $this->course_db->rows();
+                if (count($thread_id) > 0) {
+                    $thread_id = $thread_id[0]["id"];
+                    $this->course_db->query("UPDATE threads set merged_thread_id=-1, merged_post_id=-1 where id=?", array($thread_id));
+                    $this->course_db->query("DELETE FROM thread_categories where thread_id=?", array($thread_id));
+                }
+                else {
+                    $this->course_db->query("INSERT INTO threads (title, created_by, is_visible, lock_thread_date) VALUES (?, ?, ?, ?)", array($title, $post["author_user_id"], true, null));
+                    $this->course_db->query("SELECT MAX(id) as max_id from threads where title=? and created_by=?", array($title, $post["author_user_id"]));
```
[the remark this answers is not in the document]
DefinitiveAbove: Changing the code in AbstractDatabase.php to always run the code under Utils::startsWith($lower, 'select') will get the `returning` syntax to work, but I won't be pushing that change since this is used by a lot of things (so if it was changed it may be better to make a new commit) and I don't know enough about how the database works to understand if there would be any issues that result from this.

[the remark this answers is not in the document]
andrewaikens87: It may be better to use a comment like: 
```
// TODO: Update AbstractDatabase.php to work with returning syntax 
```
rather than commented out code.

site/app/templates/forum/CreatePost.twig
```diff
@@ -48,6 +48,15 @@
 
         {% if userGroup <= 3 %}
             <a class="btn btn-default btn-sm post_button_color text-decoration-none" tabindex = "0" onClick="showHistory({{ post.id }})">Show History</a>
+            {% if (isThreadLocked != 1 or accessFullGrading) and first != 1 %}
+                <a class="btn btn-default btn-sm post_button_color text-decoration-none" tabindex = "0" onClick="showSplit({{ post.id }})">
+                {% if thread_previously_merged == true %}
```
andrewaikens87: this can be simplified to `if thread_previously_merged`

site/app/views/forum/ForumThreadView.php
```diff
@@ -867,6 +870,13 @@ public function createPost($thread_id, $post, $unviewed_posts, $function_date, $
                     "ud_button_icon" => $ud_button_icon
                 ];
 
+                if ($this->core->getUser()->getGroup() <= 3) {
```
andrewaikens87: Let's use the limited access grading function check rather than using a raw `3`

Issue #5032, opened 2020-02-12
Limited Access Graders Cannot View PDFs


jchulton:
**Describe the bug**
A Limited Access Grader cannot view PDFs when grading students in their section, even when gradeable is set to allow limited access graders.

**Expected behavior**
The Limited Access Grader should be able to open and view the PDF as normal

**To Reproduce**
1.  Create a gradeable and minimum grader to limited access grader
2. Submit for a student
3. Attempt to grade as a limited access grader

![image](https://user-images.githubusercontent.com/11095297/74321443-3f1fea00-4d50-11ea-8eb4-380bd7a3c394.png)

**Screenshots**

elihschiff:
@jchulton it looks like #5032 solved this issue. Is there a reason why this issue is still open or did you just forget to close it?

PR #5033, opened 2020-02-12
[Bugfix:TAGrading] Limited Access Grader cannot view PDFs

jchulton:
As described in issue #5032, a limited access grader cannot view a PDF when grading students. They can access the grading page, but get a permissions error:

image

### What is the new behavior?
The limited access grader can view the PDF normally.
### Other information?
Solves issue #5032


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1167,6 +1167,11 @@ public function ajaxGetGradedGradeable($gradeable_id, $anon_id = '', $all_peers
         if ($submitter_id === false) {
             return;
         }
+        
+        $section = $this->core->getQueries()->getUserById($submitter_id)->getRotatingSection();
+        if ($gradeable->isGradeByRegistration()) {
+            $section = $this->core->getQueries()->getUserById($submitter_id)->getRegistrationSection();
+        }
```
MasterOdin: Use an if/else here, so that we're not wasting a potential query.

jchulton: Added the if else

PR #5043, opened 2020-02-14
[Bugfix:Forum] Fixed edit categories page not loading and remove eye icon when there is nothing to show 

elihschiff:

There are 2 bugs with the forum. The first one is that users cannot view the page to modify categories because a normal get request does not contain a csrf token. This fix makes it so that to load the page you dont need the token, but to actually modify the categories you still need the token.
This is a bug introduced in a previous pr that attempted to limit students from viewing the page, but in the process limited anyone trying to view the page if they did not have a csrf token.

The second issue is that the eye button is visible in the history for a post even though when you click it there is nothing to reveal. This happens for limited access graders who cannot unhide somebody's real name.
image

### What is the new behavior?

The modify categories page now loads (only if you have access to modify the categories)

The eye button when you view the history of a post no longer shows up for limited access graders
image

### Other information?

These are 2 small fixes code wise so I thought it might be better to put them in one pr. If it is better to split them let me know.

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -675,6 +675,9 @@ function showHistory(post_id) {
                 var visible_user_json = JSON.stringify(visible_username);
                 info_name = JSON.stringify(info_name);
                 var user_button_code = "<a style='margin-right:2px;display:inline-block; color:black;' onClick='changeName(this.parentNode, " + info_name + ", " + visible_user_json + ", false)' title='Show full user information'><i class='fas fa-eye' aria-hidden='true'></i></a>&nbsp;";
+                if(!author_user_id){
+                  user_button_code = ""
+                }
                 box.find("h7").html("<strong>"+visible_username+"</strong> "+post['post_time']);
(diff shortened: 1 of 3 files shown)


site/public/js/forum.js
```diff
@@ -675,6 +675,9 @@ function showHistory(post_id) {
                 var visible_user_json = JSON.stringify(visible_username);
                 info_name = JSON.stringify(info_name);
                 var user_button_code = "<a style='margin-right:2px;display:inline-block; color:black;' onClick='changeName(this.parentNode, " + info_name + ", " + visible_user_json + ", false)' title='Show full user information'><i class='fas fa-eye' aria-hidden='true'></i></a>&nbsp;";
+                if(!author_user_id){
```
shwehtom89: Nit: this condition is unclear. Perhaps define in a variable that an empty `author_user_id` string means that a limited access grader is viewing the post history.

bmcutler:
I'm fine with fixing 2 small bugs in one PR, in this case.

PR #5060, opened 2020-02-20
[VPAT] Added default font color to html body

### LINKED ISSUE #4460: VPAT — Default Font Color (contrast ratio issue)
As noted in https://www.w3.org/TR/WCAG20-TECHS/F24.html, some users with vision impairment may set their own default text and background colors, within their browser, to help them read sites that do not define these attributes.  Therefore,  both `color` and `background-color` CSS properties should be defined in a way that operate throughout all views.

Submitty does define background colors that operate in all views, but not a text color.  Text colors are defined in CSS, but only for certain situations.  In many places, the reason text is black is because the browser's default font color is black.

Perhaps a vision impaired user set his browser to default of a black background with white text.  Submitty will render white and off-white background colors as intended, but now some text will be white against Submitty's white and off-white backgrounds, therefore becoming unreadable (see attached screenshot of gradeables list for an example).

Screen Shot 2019-09-11 at 6 51 44 PM

The goal is to ensure that all text will have a defined color with appropriate contrast ratio.  One suggestion is to define `color` property as black within `body` in CSS.  Through inheritance, all text that did not have a specifically defined color will now be black, even when the browser's default is white.

Following up with this consideration — a challenge is that there are multiple CSS files in use with some redefining properties within the same element.  Notably, `global.css` and `reset.css` attempt to define some of the same properties within `body`.  This can further confuse inheritance which may present difficulties in correcting this issue.  Although my observation is that most views seem to favor inheritance from `global.css`

I also observed that the discussion forum appears to favor inheritance from `reset.css` rather than `global.css`.  In addition, setting `color` to black within `body` has made some text black that may have been intended for another color (such as white or gray).  See screenshot, below (circled areas).

Screen Shot 2019-09-11 at 7 18 25 PM

elihschiff:

closes #4460

### What is the new behavior?
I added `color: var(--text-black);` to body in global.css
I also took a look at multiple css files and switched a lot of colors that were based on `--default-black` to use `--text-black` as well as the same for the white version. While both are black, by making the colors based on the css variable used for text, it should be easier for a user to just switch that one variable and switch the color of text on the website while not messing with other colors on the website.

### DIFF
--- a/site/public/css/colors.css
+++ b/site/public/css/colors.css
@@ -129,8 +129,8 @@
     --focus-dropdown: #f5f5f5;
 
     /*text and fonts*/
-    --text-black: #000000; /*use only on white/background-blue*/
-    --text-white: #ffffff; /*use on all colors*/
+    --text-black: blue; /*use only on white/background-blue*/
+    --text-white: red; /*use on all colors*/
 
     /*discussion forum*/
     --viewed-discussion-post-blue: #e8f0f7;
--- a/site/public/css/colors.css
+++ b/site/public/css/colors.css
@@ -129,8 +129,8 @@
     --focus-dropdown: #f5f5f5;
 
     /*text and fonts*/
-    --text-black: blue; /*use only on white/background-blue*/
-    --text-white: red; /*use on all colors*/
+    --text-black: #000000; /*use only on white/background-blue*/
+    --text-white: #ffffff; /*use on all colors*/
 
     /*discussion forum*/
     --viewed-discussion-post-blue: #e8f0f7;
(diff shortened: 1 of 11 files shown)


site/public/css/colors.css
```diff
@@ -129,8 +129,8 @@
     --focus-dropdown: #f5f5f5;
 
     /*text and fonts*/
-    --text-black: #000000; /*use only on white/background-blue*/
-    --text-white: #ffffff; /*use on all colors*/
+    --text-black: blue; /*use only on white/background-blue*/
+    --text-white: red; /*use on all colors*/
```
MasterOdin: This is not correct.

Issue #4491, opened 2019-09-15
WCAG, JAWS, and HTML5 accessibility


mktung:
Following John Fisher's accessibility overview, there are a number of things we need to fix. I've pushed solutinos to some of them in #4489 and these are the rest:

- [x] Script between head and body
- [ ] JS mimicking link
- [ ] Link text replaced by aria-label attribute
- [x] Element meta missing itemprop
- [x] Input descendants of label must have matching id to for
- [ ] Each \<a> element must contain text or an img with an alt

Forum Specific:
- [x] Hidden h1
- [x] Empty headings
- [x] Duplicate ID
- [x] Onclick should also have keyboard option
- [x] Detached labels
- [x] Input cannot be descendant of \<a>
- [ ] \<input type=text> with aria label

mktung:
One thing I'm concerned about is:
- Element meta missing itemprop

He says this is an HTML5 spec but I can't find a source for it. I'll keep it in the issue but I don't have plans to fix this unless it's explained to me. From what I've read on MDN, W3C, and SO, it seems to imply that either you use the name or itemprop tag but not both (some suggest having 2 meta tags to capture both options). This is contradictory to the issue that we should add it.

bmcutler:
Tagging @pbailie @mcglow2

elihschiff:
Do you have more details on what the problems I listed below mean, or where you got them from (ex: WAVE, validator.w3.org, etc.) so I can understand the problems and how to fix them:

"JS mimicking link"
"Link text replaced by aria-label attribute"
"Detached labels"
"\<input type=text> with aria label"

pbailie:
According to the report:
* "JS mimicking link" is in regards to WCAG 2.0 A F42.
* "Link text replaced by aria-label attribute" is an issue related to JAWS 2018 assistive technology, and maybe others.
* "Detached labels" is in regards to WCAG 2.0 A 1.4.2.  Maybe there are form labels that are not properly associated with form control IDs.
* "<input type=text> with aria label" is an issue with the NVDA 2018 assistive technology, and possibly others.

As an aside:  John Fisher told me that Rensselaer requires WCAG AA conformance, although AAA conformance is encouraged.

elihschiff:
I looked and cannot find any examples of "JS mimicking link", is there a way to use a tool to detect this? I was searching with regex based off what this said the issue is https://www.w3.org/TR/WCAG20-TECHS/F42.html

[...]

PR #5062, opened 2020-02-21
[Feature:UI] Made all labels toggle checkbox/radio buttons when clicked

### LINKED ISSUE #3592: Checkbox and radio labels should be clickable
Clicking on the text next to a checkbox does not change the stated of the checkbox.

Clicking on the text next to a checkbox should change the state of the checkbox.

Require users to click on a small widget when they could click on a larger area.

**Additional context**
[Checkbox and radio labels should be clickable](https://stackoverflow.com/questions/2257606/making-checkbox-and-radio-labels-clickable)

elihschiff:

When you click a label next to a checkbox it should toggle that checkbox 
Closes #3592

### What is the new behavior?
Every label for a checkbox or radio box should toggle that box. 

### Other information?

Some pages dont have a label for a specific checkbox, for example the notifications settings page. So I did not do anything to that page. It does have labels but they dont correspond 1 to 1 to a specific checkbox so I think it should stay as is.

I used regex to find every checkbox and radio box and then I tested them all by hand to see if they work so I am pretty sure I tested every one on the entire website, meaning it is safe to fully close the issue.

### DIFF
--- a/site/app/templates/course/UploadCourseMaterialsForm.twig
+++ b/site/app/templates/course/UploadCourseMaterialsForm.twig
@@ -15,7 +15,7 @@
             </p>
             <input type="text" id="input-provide-full-path" class="full-width" name="eg_config_path" value="" placeholder="optional/subdirectory"/>
         </label>
-        <label>
+        <div>
             <p class="black-message">
                 Do you want to restrict this course material to some sections?
             </p>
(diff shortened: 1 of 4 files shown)


elihschiff:
This also solves "Input descendants of label must have matching id to for" in #4491

PR #5071, opened 2020-02-22
[VPAT] Removed redundant link in breadcrumbs

### LINKED ISSUE #4075: [VPAT] Redundant Link in Breadcrumbs
I am seeing WAVE report a redundant link in the breadcrumbs.  See screenshot:
Screen Shot 2019-07-12 at 6 32 16 PM

What this is saying is WAVE has detected multiple links associated with the breadcrumbs (presumably attached specifically to "Submitty") that point to the same URL.  This may cause some screen readers to unnecessarily stop on the "Submitty" breadcrumb multiple times.

elihschiff:

There is a redundant link in the breadcrumbs
Closes #4075

### What is the new behavior?
This uses js to remove and add back the redundant link in the breadcrumbs. The reason for the redundant link was because there is the mobile and desktop version of the button. By removing from the page dom, whichever one was not being show in the first place, it solves the issue.

### DIFF
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -96,26 +96,28 @@
             <div id="menu-overlay"></div>
             <nav class="flex-row shadow">
                 <div id="nav-links">
-                    <a id="home-button" class="black-btn" href="{{ base_url }}" aria-label="Go to Submitty Home">
+                    <div id="mobile_home_link_parent">
+                      <a id="home-button" class="black-btn" href="{{ base_url }}" aria-label="Go to Submitty Home">
                         <img id="submitty-mascot-home-btn"
                          src="{{ base_url }}/img/{{ duck_img }}"
                          alt="Submitty's duck mascot!"
                          height = "50px"
                          />
-                     </a>
+                      </a>
+                    </div>
                     <div id="breadcrumbs">
                         {% for b in breadcrumbs %}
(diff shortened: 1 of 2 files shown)


elihschiff:
@pbailie my bad. I think it is now fixed

MasterOdin:
Before this gets merged, I would like some clarification here on the visibility of the elements. `#home-button` is marked as `display: none` when the mobile layout is not active. Once you shrink the page to be small enough for mobile, the breadcrumb elements (except for the last one) is marked as `display: none`. @mktung raised similar concerns in https://github.com/Submitty/Submitty/issues/4075#issuecomment-511431148. Given that webaim's own documentation on [hiding links](https://webaim.org/techniques/css/invisiblecontent/#techniques) has this to say:

> ### display:none; or visibility: hidden;
> These styles will hide text from all users. The text is removed from the visual flow of the page and is ignored by screen readers. Do not use this CSS if you want the content to be read by a screen reader. But DO use it for content you don't want read by screen readers.

How is this not in compliance to begin with and is this just a bug in WAVE that we're blindly following because of "compliance"?

MasterOdin:
To add to this, hitting tab on the shown page with the WAVE error is (number corresponds to tab presses):

1. skip to main content
1. link to submitty homepage
1. link to http://submitty.org/developer/rcos_moorthy
1. http://submitty.org/

So it's not exactly clear what is adjacent here.

Issue #4619, opened 2019-10-08
link back from uplload config page to gradeable -- current gradeable


bmcutler:
When an instructor is editing a gradeable, and uses the upload config option to uploads the config.json & etc through the webpage, we'd like a link back to the edit gradeable page we came from.

(further design discussion may be needed)

PR #5078, opened 2020-02-26
[Feature:InstructorUI] link back from upload config page

mukul-kmr-jha:

When an instructor is editing a gradeable, and uses the upload config option to upload the config.json & etc through the webpage, there is no way to go back to the edit gradeable page where he came from.

### What is the new behavior?
A link is shown on the config upload screen letting the instructor go back to the gradeable editing page on the exact same location from where he came from.

Closes issue #4619

### DIFF
--- a/site/app/controllers/admin/AutogradingConfigController.php
+++ b/site/app/controllers/admin/AutogradingConfigController.php
@@ -22,7 +22,7 @@ class AutogradingConfigController extends AbstractController {
      * @param string $g_id gradeable Id
      * @return Response
      */
-    public function showConfig($g_id='') {
+    public function showConfig($g_id = '') {
         $target_dir = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "config_upload");
         $all_files = FileUtils::getAllFiles($target_dir);
         $all_paths = array();
--- a/site/app/controllers/admin/AutogradingConfigController.php
+++ b/site/app/controllers/admin/AutogradingConfigController.php
@@ -19,9 +19,10 @@ use Symfony\Component\Routing\Annotation\Route;
 class AutogradingConfigController extends AbstractController {
     /**
      * @Route("/{_semester}/{_course}/autograding_config", methods={"GET"})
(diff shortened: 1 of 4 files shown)


site/app/controllers/admin/AutogradingConfigController.php
```diff
@@ -42,6 +43,7 @@ public function showConfig() {
                 'uploadConfigForm',
                 $target_dir,
                 $all_files,
+                $gradeable_id = $g_id,
```
MasterOdin: Assignment here is not necessary, please remove it.

site/app/templates/admin/UploadConfigForm.twig
```diff
@@ -26,6 +26,8 @@
         Upload Config: <input type="file" name="config_upload" /><br />
         <input type="submit" value="Upload" />
     </form>
+    <br>
+    <a class="btn btn-primary" style="margin-top: 6px" href={{ back_url }}>Go Back To AutoGrading Page</a>
```
MasterOdin: Capitalization should be `Autograding` (lower case g)

site/app/views/admin/GradeableView.php
```diff
@@ -18,6 +18,9 @@ public function uploadConfigForm($target_dir, $all_files, $inuse_config) {
             "rename_url" => $this->core->buildCourseUrl(['autograding_config', 'rename']),
             "delete_url" => $this->core->buildCourseUrl(['autograding_config', 'delete']),
             "display_url" => $this->core->buildCourseUrl(['display_file']),
+            "back_url" => $gradeable_id !== ''
```
MasterOdin: Let's set `back_url` to `''` if `$gradeable_id === ''`

site/app/templates/admin/UploadConfigForm.twig
MasterOdin: Let's also hide the button if `{{ back_url }}` is blank (with change below) as it does not make sense.

PR #5101, opened 2020-03-01
[Feature:InstructorUI] Fixed ios buttons not following css

elihschiff:

\<input type="button"> on iphones do not look the same as on all other browsers

### What is the new behavior?

This fixes the buttons on iphone by switching it to \<button type="button"> instead of \<input type="button">


MasterOdin:
It would be good to know if applying this CSS:
```
button,
[type="button"],
[type="reset"],
[type="submit"] {
  -webkit-appearance: button;
}
```

fixes the issue.

Issue #5086, opened 2020-02-29
VPAT and HTML 5


pbailie:
In order to best ensure that assistive technologies interpret content accurately, The W3C recommends that markup languages (e.g. HTML 5) fully conform to their specification as a measure to meet WCAG success criterion 4.1.1.  A markup validator is also recommended to help validate markup code.

q.v. https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G192

Submitty has many HTML 5 errors according to W3C's HTML 5 validator at https://validator.w3.org/.

I recommend that individual errors have their own issues opened and maybe reference this issue.

PR #5106, opened 2020-03-01
[VPAT] w3 errors on course materials page

elihschiff:

see #5086
There are a number of errors found by w3 validator on the course materials page https://validator.w3.org/nu/#file
image

### What is the new behavior?

Everything checked in the image above has been resolved


jchulton:
Looking through the code, the only changes to formatting, so the functionality of course materials still works. One thing I noticed though is that the new line between the Yes No radio buttons for Section and the Section checkbox has been removed, as below:

Now:

image


Before:

image

If it isn't too big a hassel, I think it looks nicer with the new line, but this isn't a big concern

elihschiff:
@jchulton 
Should be fixed now. Thanks for looking over my PR

elihschiff:
Sorry, I realize there were some wave errors I did not fix in this pr.
(I thought I fixed them in a different pr but that is not the case)
Could you also review the changes made in the last commit (https://github.com/Submitty/Submitty/pull/5106/commits/162a09a5f798f320e1867f02631401875818b355)

PR #5134, opened 2020-03-05
[Bugfix: PDFAnnotator] TA grading impacts student view

mukul-kmr-jha:
### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
See detailed description in this [issue](https://github.com/Submitty/Submitty/issues/5122#issue-574416623)

### What is the new behavior?
Closes #5122 
Now in the student view of the PDF , cookie values are ignored and its just shows the plain PDF without any rotation and zoom.

PR #5137, opened 2020-03-07
[Bugfix: PDFAnnotator] Student View of Unannotated PDF 

### LINKED ISSUE #5073: Student view of unannotated pdf buggy
note from PR #5063

"During testing, did notice the "Original" button for the PDF shows the annotated version. I am not sure when this last worked properly, given that it goes to the same URL as the annotated URL, can probably make a secondary issue for that."

### LINKED ISSUE #5122: pdf rotation cookie from TA grading impacts student view
**Describe the bug**
If I'm a grader in course A, and I use the TA grading interface to view and grade and annotate PDF files.  If the last time I used the TA grading interface I rotated a pdf, that setting is saved in a cookie.

If I'm just a student in course B, and I view the annotations on a pdf I've submitted for that course, the cookie setting for TA grading seems to carry over into the viewer.  from the student side I cannot control this rotation setting.

**Expected behavior**
Either introduce the rotation option on the student view of annotated PDFs, or just ignore the cookie setting for rotation and don't let it rotate the student view

mukul-kmr-jha:

See detailed description here #5073 

### What is the new behavior?
Closes #5073 
Unannotated view of pdf files are fixed.

### DIFF
--- a/site/app/controllers/MiscController.php
+++ b/site/app/controllers/MiscController.php
@@ -115,11 +115,6 @@ class MiscController extends AbstractController {
                     return false;
                 }
             }
-
-            if (!$this->core->getUser()->accessGrading() && !CourseMaterial::isSectionAllowed($this->core, $path, $this->core->getUser())) {
-                $this->core->getOutput()->showError("Your section may not access this file.");
-                return false;
-            }
         }
         $file_name = basename(rawurldecode(htmlspecialchars_decode($path)));
         $corrected_name = pathinfo($path, PATHINFO_DIRNAME) . "/" .  $file_name;
--- a/site/app/templates/autograding/TAResults.twig
+++ b/site/app/templates/autograding/TAResults.twig
@@ -70,7 +70,7 @@ $( document ).ready(function() {
                                     <button class = 'btn btn-default key_to_click' onclick='downloadFile("{{ file.path|url_encode }}", "submissions")' aria-label="Download the file"> Download
                                         <i class="fas fa-download" title="Download the file"></i></button>
                                 {% endif %}
-                                <a class="btn btn-primary key_to_click" tabindex="0" onclick="openUrl('{{ student_pdf_view_url }}?filename={{ file.relative_name }}&path={{ file.path|url_encode }}')">View Popup <i class="fas fa-window-restore"></i></a>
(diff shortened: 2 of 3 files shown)


site/app/controllers/MiscController.php
```diff
@@ -115,11 +115,6 @@ public function displayFile($dir, $path, $gradeable_id = null, $user_id = null,
                     return false;
                 }
             }
-
-            if (!$this->core->getUser()->accessGrading() && !CourseMaterial::isSectionAllowed($this->core, $path, $this->core->getUser())) {
```
shailpatels: Why is check removed?

mukul-kmr-jha: This check is meant for the courseMaterial files but currently it is being used for submission file also, which break the submitty. Also this check is already in place for the courseMaterial file (see above `if` conditions Line no. 107 and 113).

mukul-kmr-jha: > I still don't understand why the check for section-specific course materials is being removed.
> I'm not sure what you're talking about with the if condition on 107, 113, thats the condition you're removing in this PR.

@shailpatels `isSectionAllowed` check is only meant for the courseMaterial files and currently, we are testing submission files also against this check which is not correct (please correct me if I'm wrong?)  
Also, this check is causing submitty to crash when you try to view the submission files (i.e using `display_file` function). 
On line 107 and 113 I was quoting that this condition( `isSectionAllowed`) is already present for the CoursMaterial files i.e ` if ($dir == 'course_materials') `. In line 113 you'll find that this is the exact same code (or to say,  `isSectionAllowed` condition) which is good and it should be as it is. So, that's the reason behind removing the check (buggy).

MasterOdin:
Thanks for the contribution @mukul-kmr-jha. Would you be able to split this PR into two parts, one addressing #5122 and one addressing #5073. I ask this as I think the former is going to have to undergo some amount of review process while the changes for #5073 can pretty much be merged as is (with variable rename only).

mukul-kmr-jha:
Done, [here](https://github.com/Submitty/Submitty/pull/5134) is the PR for issue #5122 
Please take another look. Thanks!

site/app/controllers/MiscController.php
jchulton: Testing course materials, removing this check does not cause any new problems.

PR #5143, opened 2020-03-10
 [Feature:Install] Always use new dev site installer when debug enabled

MasterOdin:

In #4790, a new developmental site installer was added, which used the results of the rsync command to run a selective amount of chmods, chowns, and package installers. This was gated behind the `submitty_install_site_dev` command in Vagrant or running it directly. I've been using the script since it was added, but I am unsure how many other people were using it.

### What is the new behavior?

This makes the new script the installer used for running the regular submitty install script (`bash .setup/INSTALL_SUBMITTY_HELPER.sh`) if the debug flag is set to true. This primarily will affect Vagrant developers, who will now all be using the new script. Production systems will continue to use the existing (legacy) system. Documentation has been added to the top of both scripts to explain their purpose, and a plan of phasing out the legacy script.

### Other information?

Tested running the install script (`bash .setup/INSTALL_SUBMITTY_HELPER_SITE.sh`) toggling the debug flag (`config/database.json:debugging_enabled`) from true to false and back again several times, along with jumping around the commit tree.

### Tests:

- [x] Fresh install (site does not exist)
- [x] Regular file change affecting PHP files
- [x] Regular file change affecting files in `site/public/`
- [x] File change on composer.json
- [x] File change on package.json
- [x] File change with new folders


MasterOdin:
To test this, I have been using the following python3 script:
```
#!/usr/bin/env python3

import os

for root, dirs, files in os.walk('/usr/local/submitty/site'):
    for name in dirs:
        print(f"{oct(os.stat(os.path.join(root, name)).st_mode)[-3:]} {os.path.join(root, name)}")

    for name in files:
        print(f"{oct(os.stat(os.path.join(root, name)).st_mode)[-3:]} {os.path.join(root, name)}")
```

and then running it with `python3 perms.py > perms.log` after running the installer with the new or old script and then using diff to compare the results.

PR #5145, opened 2020-03-12
[Feature:InstructorUI] Option to collect contact information for office hours queue

elihschiff:

RPI has shut down the campus and we have no way to run office hours

### What is the new behavior?

I added a setting to the course settings page to enable the ability for students to add contact info to the office hours queue.
image
Once checked, students will not be required to add their contact info to join the office hours queue for that class.
image
Contact info is just a required textbox, so you can send an email, a url, or anything else.
For instructors/mentors/tas they will see this contact info under the student in the queue
image

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -1237,7 +1237,8 @@ CREATE TABLE IF NOT EXISTS queue(
   time_out TIMESTAMP,
   added_by TEXT NOT NULL REFERENCES users(user_id),
   help_started_by TEXT REFERENCES users(user_id),
-  removed_by TEXT REFERENCES users(user_id)
+  removed_by TEXT REFERENCES users(user_id),
+  contact_info TEXT
 );
 CREATE TABLE IF NOT EXISTS queue_settings(
   id serial PRIMARY KEY,
--- a/migration/migrator/migrations/course/20200311135223_oh_queue_contact_info.py
+++ b/migration/migrator/migrations/course/20200311135223_oh_queue_contact_info.py
@@ -1,5 +1,6 @@
 """Migration for a given Submitty course database."""
-
+import json
+from pathlib import Path
 
 def up(config, database, semester, course):
     """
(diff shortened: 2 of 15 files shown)


matgar1997:
I think it would be helpful for the student view to also see the instructions for contacting on their page incase they put in a type. Or perhaps a more dynamic approach would allow the instructor to ask for clarification and the student could edit on their page without needing to leave and re-enter the queue. Sort of like a chat system but only keep track of one editable string for simplicity.

I also probably wouldn't specifically mention pandemic under the instructor settings. Leave it for any case which would need a remote queue. 

Other then the suggestions the functionality looks good.

elihschiff:
I feel like if the problem can be solved by a text question they can just use the discussion forum. The only reason to join the office hours queue is if you need to screen share or voice chat. So I cannot think of a time when you would need more clarification that you cannot hear in person once you connect to them.

Also I do not think it is smart to put instructor contact info on the student page. I as a mentor do not want students contacting me, I cannot see how that will end well, I want to contact them when it is their turn.

The use of "pandemic" is very important :frowning:, I vote to keep it

shailpatels:
Is this an actual issue that was discussed in person? The mentions of pandemic make this feel like a joke PR but it looks like a lot of work has gone into it.
I don't see the point of implementing a method to contact students through the queue virtually since the queue was originally designed for managing people in a physical space.

If a school decides to run a course online while using Submitty I think the existing infrastructure they are/will using be able to handle students who want help. I image jumping between Slack/WebEx/Zoom and then to Submitty office hours Q and then back to whatever isn't very intuitive...

elihschiff:
Ok the use of pandemic I just added. I can remove it 😢

But the rest of this pr was discussed over slack. Because rpi is going fully online for classes we need a way for students to let mentors know how to contact them when it is their turn for help. Students would put in their webex, skype, discord, etc. Account and a mentor would then contact them.

bmcutler:
Thanks @elihschiff!

PR #5146, opened 2020-03-12
[UI/UX:InstructorUI] Fixed button styling on Course Settings

areeshatariq:
Signed-off-by: Areesha Tariq areeshatariq02@gmail.com

Button styling was inconsistent with the buttons on other pages:

Screenshot 2020-03-27 14 07 29

### What is the new behavior?
Button styling is now consistent with the other buttons:

Screenshot 2020-03-27 14 08 54

### DIFF
--- a/site/app/templates/admin/Configuration.twig
+++ b/site/app/templates/admin/Configuration.twig
@@ -2,7 +2,7 @@
     <header>
         <h1>Course Settings</h1>
         <div class="btn-wrapper">
-            <a href="{{ theme_url }}" class="btn btn-default">Customize Website Theme</a>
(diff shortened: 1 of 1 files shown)


elihschiff:
I actually think the color that was there does follow the Submitty interface style guide. As you can see here https://submitty.org/developer/interface_design_style_guide
"white/grey buttons: are infrequently-used optional actions. Examples: Cancel an assignment submission. Request a TA regrade."
I feel like changing the website theme is not something most professors would do and so while it remains an option, keeping it white over blue makes mores sense

bmcutler:
Correct, this button is white by design.

PR #5150, opened 2020-03-14
[Feature:InstructorUI] customizable welcome message for office hours queue

elihschiff:

There is no way for instructors to give a custom message to students before they join the queue

### What is the new behavior?

Added a custom course wide message that an instructor can set which shows up at the top of the office hours queue page
image
image

### DIFF
--- a/migration/migrator/migrations/course/20200313213344_queue_message.py
+++ b/migration/migrator/migrations/course/20200313213344_queue_message.py
@@ -1,0 +1,45 @@
+"""Migration for a given Submitty course database."""
+import json
+from pathlib import Path
+
+def up(config, database, semester, course):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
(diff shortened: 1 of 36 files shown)


migration/migrator/migrations/course/20200313213344_queue_message.py
```diff
@@ -0,0 +1,45 @@
...
+
+        if 'queue_message' not in j['course_details']:
+            j['course_details']['queue_message'] = 'Welcome to the office hours/lab queue. Choose a queue name and enter the code secret code you have been provided.'
+
+        with open(config_file, 'w') as out_file:
+            json.dump(j, out_file, indent=4)
+
+
+def down(config, database, semester, course):
+    """
```
shwehtom89: Do the course configuration unit tests still work if you roll back to a previous version without this entry added? If they fail then you should delete the entry `j['course_details']['queue_message']` on a `down` migration.

site/config/course_template.json
```diff
@@ -23,6 +23,7 @@
         "room_seating_gradeable_id": "",
         "auto_rainbow_grades": false,
         "queue_enabled": false,
-        "queue_contact_info": false
+        "queue_contact_info": false,
+        "queue_message": "Welcome to the office hours/lab queue. Choose a queue name and enter the code secret code you have been provided."
```
shwehtom89: Do we want to have this default message for all Office Hours queues or do you want this to just be a power user feature?

elihschiff: Now that I think about it I agree. This message adds nothing so I will make it a power user feature

PR #5159, opened 2020-03-16
[Bugfix:Forum] Improve forum markdown

### LINKED ISSUE #4135: Multiple Forum Markdown issues
Apparently, the forum has markdown now?

This allows inline images, is that something we want?

Markdown doesn't render on the thread list on the left 
Lists are very buggy
They don't render for replies 
Markdown lists cut off to the left of each post
\* a
\* b
\* b
\* c
\* d

Results in 

image

\1. Item 
\2. Item  
--\* Mixed        <- pretend there are 2 spaces before * Mixed 
--\* Mixed   
\3. Item

Results in :

Should be : 
1. Item 
2. Item  
    * Mixed    
    * Mixed   
3. Item

\* a
\* b
\* c

\- a
\- b
\- c

results in : 
In addition to the cutting off, mixing the two types of list options causes weird spacing making
the bullets not line up with the line.
image

Should be :

* a 
* b
* c

- a 
- b
- c

Code highlighting doesn't do much I prefered the old highlight, right now it looks like all it does 
it convert it to a monospace font?

"````
int foo(void)
{
return 0 ;
}
results in :

image

selecting a language doesn't do anything either
````c
int main(void)
{
	return 0;
}
````
github's syntax highlight on c : 
image

Long code blocks don't wrap or scroll they just cut off 
image

On github it adds a scroll whell
```c
char * longy = aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

I don't know which flavor of markdown we're using but quotes and tables don't render
> a
> b
> c
> * a
> * b
> * c

I can't tell if anything is happening (or if its supposed to)

image

results in : 

image

Github renders it like 

|A|B|C|
|-|-|-|
|A|B|C|

DefinitiveAbove:

Some markdown does not format correctly on the discussion forum.

### What is the new behavior?
Fixes #4135 
Relevant markdown features now format correctly (bullets, tables).

### Other information?
Tested by making posts with different content in the markdown fields.

### DIFF
--- a/site/app/templates/forum/CreatePost.twig
+++ b/site/app/templates/forum/CreatePost.twig
@@ -32,7 +32,7 @@
 
     {% endif %}
 
-    <pre class='pre_forum'><span class="post_content pre-wrap {% if render_markdown == 1 %}forum_markdown{% endif %}" >{% include "forum/RenderPost.twig" with { "post_content": post_content, "render_markdown": render_markdown } only %}</span></pre>
+    <pre class='pre_forum'><span class="post_content pre-wrap" >{% include "forum/RenderPost.twig" with { "post_content": post_content, "render_markdown": render_markdown } only %}</span></pre>
 
     <hr style="margin-bottom: 3px;">
 
--- a/site/app/templates/forum/CreatePost.twig
+++ b/site/app/templates/forum/CreatePost.twig
@@ -32,7 +32,7 @@
 
     {% endif %}
 
-    <pre class='pre_forum'><span class="post_content pre-wrap" >{% include "forum/RenderPost.twig" with { "post_content": post_content, "render_markdown": render_markdown } only %}</span></pre>
+    <pre class='pre_forum'><span class="post_content pre-wrap {% if render_markdown == 1 %}forum_markdown{% endif %}" >{% include "forum/RenderPost.twig" with { "post_content": post_content, "render_markdown": render_markdown } only %}</span></pre>
 
     <hr style="margin-bottom: 3px;">
 
--- a/site/public/css/forum.css
+++ b/site/public/css/forum.css
@@ -547,19 +547,29 @@ body {min-width: 925px;}
     font-size: 1.8em;
 }
 
-.forum_markdown li {
+.post_content li {
     white-space: normal;
 }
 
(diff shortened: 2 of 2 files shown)


site/public/css/forum.css
```diff
@@ -546,3 +546,20 @@ body {min-width: 925px;}
...
+.forum_markdown ol, .forum_markdown ul {
+    white-space: normal;
+    padding-left: 15px;
+}
+
+.forum_markdown table {
+    border-collapse: collapse;
+}
+
+.forum_markdown table, .forum_markdown th, .forum_markdown td {
```
MasterOdin: Put these on new lines please

site/public/css/forum.css
```diff
@@ -546,3 +546,20 @@ body {min-width: 925px;}
     margin-right: 5px;
     font-size: 1.8em;
 }
+
+.forum_markdown li {
+    white-space: normal;
+}
+
+.forum_markdown ol, .forum_markdown ul {
```
MasterOdin: Put these on new lines please

site/app/templates/forum/CreatePost.twig
```diff
@@ -32,7 +32,7 @@
 
     {% endif %}
 
-    <pre class='pre_forum'><span class="post_content pre-wrap" >{% include "forum/RenderPost.twig" with { "post_content": post_content, "render_markdown": render_markdown } only %}</span></pre>
+    <pre class='pre_forum'><span class="post_content pre-wrap {% if render_markdown == 1 %}forum_markdown{% endif %}" >{% include "forum/RenderPost.twig" with { "post_content": post_content, "render_markdown": render_markdown } only %}</span></pre>
```
MasterOdin: Can we make all of the list styling based off the `.post_content` class? Not sure we would ever be in the case where we would want different styling of lists in posts based on whether markdown is used or not, though it is not possible to create lists right now otherwise.

PR #5162, opened 2020-03-18
[Bugfix:InstructorUI] Add id back to config selection dropdown

emaicus:
In a recent PR, the ```id``` attribute was removed from the configuration selection dropdown on the create/edit gradeable form. This created an error which stopped gradeables from successfully building when selected via the dropdown. This PR re-introdcues an ```id``` to fix the error.


emaicus:
@elihschiff, could you review this and comment on why the id was removed? Thanks!

elihschiff:
I think I might have fixed the bug and solved the id issue I listed above.
I did not want to push to this branch so I made a new branch `fix_config_selection_eli ` with my fix if you want to check it out just to make sure it works.

If it looks good, the easiest thing would most likely be just merge that branch with this one and then this pr can be merged with master.

emaicus:
Thanks @elihschiff, that fixed the bug. I've merged your changes into this PR, and recommend that it be merged @bmcutler

PR #5172, opened 2020-03-21
[Feature:Submission] Warn user when trying to upload more than the max allowed files

### LINKED ISSUE #4456: Display maximum number of files that can be uploaded to user
Right now across submitty's file upload drag and drop areas users can add as many files as they want even the the max_file_uploads is restricted to 20 by php (by default).

On every interface to upload files display the max allowed to users and then do a client side check to prevent users from adding 20+. Since php already handles checking for a number of file.

marwanatef2:
Fixes #4456 

Users were allowed to add any no. of files they want however, the maximum allowable no. for submission in php is restricted by default to 20.
Now users are not allowed to add more than 20 files per submission where if they tried they get alerted that they have exceeded the limit.

The new behavior can be seen in this gif
addvalid

Please let me know if any messages or something need to be changed.


marwanatef2:
> the best place to actually check for the file limit is the `handleSubmission` & `handlebulk` functions

my point is i don't want user to submit 20+ files then alert some message that submission file limit is exceeded so he has to remove some files, i want to prevent him from adding more than 20 files, stopping him at the limit, so it is better handled with addition not submission.

MasterOdin:
I agree on showing the message on adding, it's a friendlier user experience.

bmcutler:
Instead of hardcoding "20" into our source code, can we read the current system setting?
For example, if a production installation has modified the php.ini setting to be 50 files
https://www.codexworld.com/how-to/increase-files-upload-limit-in-php/

Also, can we add a line to the warning/error message that says
"Maximum allowed number of files to be uploaded is 20.
Please package your files in a single .zip file if it is necessary for you to submit more than this limit."

shailpatels:
@marwanatef2 are you still working on this PR?

marwanatef2:
> @marwanatef2 are you still working on this PR?

Yes, I am. I'll get it done I was just working on my proposal.

marwanatef2:
> Instead of hardcoding "20" into our source code, can we read the current system setting?

this part is done.

> Also, can we add a line to the warning/error message that says
> "Maximum allowed number of files to be uploaded is 20.
> Please package your files in a single .zip file if it is necessary for you to submit more than this limit."

this part can be shown in this screenshot
Capture

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -2,7 +2,10 @@
 
 {# Display the gradeable_message if one exists at the top of the page #}
 {% if has_gradeable_message %}
-    <div class='content gradeable_message'>{{ gradeable_message | markdown }}</div>
+    <div class='content gradeable_message'>
+        {{ gradeable_message | markdown }}
+        Maximum allowed number of files to be uploaded is {{ max_file_uploads }}.
```
MasterOdin, line 7:
```
        {% if part_names|length > 1 %}
            <div id="upload-boxes">
                {# Submit boxes #}
                {% for part in part_names %}
                    {# uploadIndex and input-file + Index required for drag-and-drop.js #}
                    <div tabindex="0"
                        id="upload{{ loop.index }}"
                        class="upload-box"
                        onkeypress="clicked_on_box(event)"
                        role="text"
                        aria-label="Press enter to upload your {{ part }} file"
                    >
                        <h2 class="label" id="label{{ loop.index }}">
                            {% if part_names|length > 1 %}
                                Drag your {{ part }} file(s) here or click to open file browser
                            {% else %}
                                Drag your file(s) here or click to open file browser
                            {% endif %}
                        </h2>
                        <input type="file" name="files" id="input-file{{ loop.index }}" class="hide" onchange="addFilesFromInput({{ loop.index }})" multiple aria-label="Select Files to upload" />
                        <table class="file-upload-table" id="file-upload-table-{{ loop.index }}">
                        </table>
                    </div>
                {% endfor %}
                <p>Maximum allowed number of files to be uploaded is {{ max_file_uploads }}.</p>
            </div>
        {% endif %}
```

(I'm adding a surrounding `{% if %}` here as well as with your change.)

site/public/js/drag-and-drop.js
```diff
@@ -17,6 +17,9 @@ var label_array = [];
 var use_previous = false;
 var changed = false;        // if files from previous submission changed
 
+var total_files_added = 0;
+var MAX_NO_OF_FILES;
```
MasterOdin, line 21: Use `MAX_NUM_OF_FILES` as the name.

Use `let` keyword please for new variables.

site/public/js/drag-and-drop.js
```diff
@@ -219,6 +246,7 @@ function deleteFiles(part) {
     var labels = dropzone.getElementsByClassName("file-label");
     while(labels[0]){
         dropzone.removeChild(labels[0]);
+        total_files_added --;
```
MasterOdin, line 249: Get rid of the space here between the variable and the operator

site/public/js/drag-and-drop.js
```diff
@@ -243,6 +271,7 @@ function deleteSingleFile(filename, part, previous) {
             if (file_array[part-1][j].name == filename) {
                 file_array[part-1].splice(j, 1);
                 label_array[part-1].splice(j, 1);
+                total_files_added--;
```
MasterOdin, line 274: Get rid of the space here between the variable and the operator

marwanatef2:
> your message is causing the submission area to be pushed to the left side

it looks like I added the message outside the "upload box" div, modified now.
 
> On chrome after I submit past the max number of files, and get the alert, and I try to submit more than the max number again I don't get the alert, nothing seems to happen. This appears to be on chrome only, firefox does not have this issue on Mac.

I am working on chrome already and I had the same problem, It's fixed now as shown in the gif below.

readding

bmcutler:
@shailpatels @MasterOdin 
Can one or both of you please re-review & approve

PR #5183, opened 2020-03-24
[Feature:Autograding] Ease of Use Updates to the Python Custom Validator

emaicus:
This PR makes the following updates to the python custom validator specification:
1. The Instructor may write log output to ```validation_logfile.txt```.
2. The Instructor may dump score output to ```validation_results.json```
3. Text written to stdout will be treated as a ```validation_results.json``` to avoid breaking old validators. However, this feature is now considered deprecated and will be removed in a future release. If both stdout and a ```validation_results.json``` exist, stdout is disregarded/lost.
4. Any output written to stderr is now captured and stored in ```validation_stderr```
5. All files listed above are renamed and stored in the details folder for a submission. The renaming is as follows: ```<filename>_<testcase number>_<autocheck number>.<extension>```. So testcase 1, autocheck 1 validation logfile is renamed ```validation_logfile_0_0.txt```.

This PR updates the Python Custom Validation gradeable in the Development course to use both the old and new validation scheme.

stderr


bmcutler:
The location of the validation files shouldn't be at the 'top' level of the results folder.
This is alot of clutter.
They don't really all belong in the details or in the logs subfolders.
So maybe we should just make a new subfolder named "validation" or "custom_validation".
I'm leaning towards just validation

PR #5189, opened 2020-03-26
[Feature:Autograding] Shipper rejects stale jobs

emaicus:
This PR updates the autograding shipper to include a 64 character random string identifier in the queue file of each autograding job. This string id is checked when the shipper retrieves a finished job from a worker machine to make certain that the returned job is not stale. In this way, we ensure that if the shipper is restarted, it will not accept outstanding "stale" jobs from its workers.


sbin/submitty_autograding_shipper.py
```diff
@@ -355,24 +358,37 @@ def unpack_job(which_machine,which_untrusted,next_directory,next_to_grade):
...
         os.remove(local_done_queue_file)
 
-    msg = "Unpacked job from " + which_machine if success else "ERROR: failure returned from worker machine"
     print(msg)
     autograding_utils.log_message(AUTOGRADING_LOG_PATH, JOB_ID, jobname=item_name, which_untrusted=which_untrusted, is_batch=is_batch, message=msg)
     return True
 
+def generate_random_string(length):
+  if length < 1:
+    return ''
```
hdrodz97, line 390: This is redundant

emaicus: Removed

emaicus:
@hdrodz97, ready for re-review.

sbin/submitty_autograding_shipper.py
```diff
@@ -355,24 +358,35 @@ def unpack_job(which_machine,which_untrusted,next_directory,next_to_grade):
...
     with contextlib.suppress(FileNotFoundError):
         os.remove(local_done_queue_file)
 
-    msg = "Unpacked job from " + which_machine if success else "ERROR: failure returned from worker machine"
     print(msg)
     autograding_utils.log_message(AUTOGRADING_LOG_PATH, JOB_ID, jobname=item_name, which_untrusted=which_untrusted, is_batch=is_batch, message=msg)
     return True
 
+def generate_random_string(length):
+  return ''.join(random.choice(string.ascii_uppercase) for i in range(length))
```
MasterOdin, line 389: Let's add this to the submitty_utils python module as we use this same function more or less in three different places.

sbin/submitty_autograding_shipper.py
```diff
@@ -355,20 +358,29 @@ def unpack_job(which_machine,which_untrusted,next_directory,next_to_grade):
...
+        msg = "ERROR: failure returned from worker machine"
         success = False
 
     with contextlib.suppress(FileNotFoundError):
         os.remove(local_done_queue_file)
 
-    msg = "Unpacked job from " + which_machine if success else "ERROR: failure returned from worker machine"
     print(msg)
     autograding_utils.log_message(AUTOGRADING_LOG_PATH, JOB_ID, jobname=item_name, which_untrusted=which_untrusted, is_batch=is_batch, message=msg)
     return True
```
MasterOdin, line 386: shouldn't this be `return success`?

python_submitty_utils/submitty_utils/string_utils.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def generate_random_string(length):
+    """
+    Return an uppercase string of n letters.
+
+    :return:
+    :rtype: string
+    """
+    return ''.join(random.choice(string.ascii_uppercase) for i in range(length))
```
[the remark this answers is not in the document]
emaicus: Fixed and added tests.

sbin/submitty_autograding_shipper.py
emaicus: Good catch, updated.

sbin/submitty_autograding_shipper.py
emaicus: Updated

python_submitty_utils/tests/test_string_utils.py
```diff
@@ -0,0 +1,26 @@
+import unittest
+
+from submitty_utils import string_utils
+
+
+class TestUser(unittest.TestCase):
+    def testNegativeLength(self):
+        for i in range(-100, 0):
+            self.assertEqual(string_utils.generate_random_string(i), '')
```
MasterOdin, line 9:
```suggestion
    def testNegativeLength(self):
        self.assertEqual(string_utils.generate_random_string(-1), '')
```

python_submitty_utils/tests/test_string_utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def testZeroLength(self):
+        self.assertEqual(string_utils.generate_random_string(0), '')
+
+    def testPositiveLengths(self):
+        for i in range(1, 100):
+            self.assertEqual(len(string_utils.generate_random_string(i)), i)
```
MasterOdin, line 16:
```suggestion
    def testPositiveLength(self):
        self.assertEqual(len(string_utils.generate_random_string(1)), 1)
```

PR #5201, opened 2020-03-28
[Bugfix : Submission] Max number of files which a user can upload

### LINKED ISSUE #4456: Display maximum number of files that can be uploaded to user

vorabrijesh:
Fixes #4456

Follows PR:  #5172
Users were allowed to add any no. of files they want, however, the maximum allowable no. for submission in PHP is restricted by default to 20 from ini_get().
Now users are not allowed to add more than 20 files per submission where if they tried they get alerted that they have exceeded the limit.

This PR tries to solve [Comment](https://github.com/Submitty/Submitty/pull/5172#pullrequestreview-378959732) from @shailpatels. 

Below screenshot shows the behaviour. 

Screenshot (17)

### DIFF
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -2,7 +2,7 @@
 
 {# Display the gradeable_message if one exists at the top of the page #}
 {% if has_gradeable_message %}
-    <div class='content gradeable_message'>{{ gradeable_message | markdown }}</div>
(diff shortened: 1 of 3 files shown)


vorabrijesh:
Can anyone review this?

shailpatels:
Hi @vorabrijesh Thanks again for your PR, after reviewing #5172 again it looks like that one is closer to being completed than this PR, and since #5172 was originally made first we'd like to see that Github user get credit for work they did.

PR #5206, opened 2020-03-29
[Refactor:PHP] Simplify Response classes

MasterOdin:

* [x] Documentation has been updated/added if relevant: Submitty/submitty.github.io#213

Currently, it is expected that all classes return a `Response` instance. Inside this overarching class, we can then have `WebResponse`, `JsonResponse`, `RedirectResponse`. When originally planning it, it was expected that we would always have multiple types.

### What is the new behavior?

in practice, this has proven not to be true. Instead, this makes it clearer that a function can return a single response type (e.g. `JsonResponse`) and have it work perfectly fine. This will make testing functions a bit shorter for those that only return one type, as well as allow us to specify narrower return types for functions to make phpstan work a bit better at its job.

### Other information?

While this is breaking, through a combination of phpstan and phpcs (and manual clicking around), I am pretty confident that nothing has broken.


shailpatels:
@MasterOdin Is there a reason `WebOnlyResponse` has been changed to `webOnlyResponse` I guess it doesn't really matter but I was just wondering since the other two responses are capitalized?

PR #5208, opened 2020-03-29
[Bugfix:InstructorUI] fix bugs in dark theme

elihschiff:

If an instructor uses custom css the dark mode may have broken some of their css.
The submitty logo may have become very large 
image
as well as if they use a light background, the sidebar buttons might be very hard to see when in dark mode
image

This also fixes an issue where when a user's device defaults them to dark theme (usually on phones with a system wide dark theme) the checkbox to choose the theme is inverted until they click it and refresh the page.

### What is the new behavior?

This fixes the issues listed above.

### DIFF
--- a/site/app/templates/misc/SourceSettings.twig
+++ b/site/app/templates/misc/SourceSettings.twig
@@ -13,7 +13,7 @@
     else {
         editor{{ number }}.setSize("100%", "auto");
     }
-    var theme = localStorage.getItem("codeDisplayStyle");
+    var theme = localStorage.getItem("theme");
     editor{{ number }}.setOption("theme", (theme == null || theme === "light") ? "eclipse" : "monokai");
     editor{{ number }}.setOption("mode", "{{ type }}");
     $("#myTab").find("a").click(function (e) {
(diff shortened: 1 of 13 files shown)


bmcutler:
Sort of related question...
Do you think we can get rid of the light vs dark toggle selection in TA grading for code/text files now?  Is there a reason to keep independent control?  (I'd like to make the UI simpler and just have 1 control)

bmcutler:
Also...  the hover color (dark blue) for links is terrible in dark mode.
(e.g. links back to documentation from the instructor's course settings page)

elihschiff:
I removed the button, it now just follows the theme you have set.
The hover link color is now also much easier to see.

site/app/templates/misc/SourceSettings.twig
```diff
@@ -13,7 +13,7 @@
     else {
         editor{{ number }}.setSize("100%", "auto");
     }
-    var theme = localStorage.getItem("codeDisplayStyle");
+    var theme = localStorage.getItem("theme");
```
MasterOdin: Will this properly account for `window.matchMedia("(prefers-color-scheme: dark)").matches` as well?

elihschiff:
@clino47 and @MasterOdin both your suggestions should now be fixed. Along with a few other minor color issues that I have been informed about.

PR #5214, opened 2020-03-30
[Feature:Developer] convert bools & dates to correct sql

### LINKED ISSUE #5211: Have AbstractDatabase::query automatically convert certain parameter types
Currently, when writing to the database, the developer has to know that certain types (like booleans and DateTimes) have to be converted to a string or number before they can be used. For booleans, we have historically had a `AbstractDatabase::convertBoolean` method, but again, it requires the developer to know they have to call this method.

It would be nicer / cleaner if the `AbstractDatabase::query` method automatically handled casting these types for us, so that the developer can just use booleans / datetimes as is, without worrying about the ins and outs of the conversion.

elihschiff:

When you want to add a bool or a date to a sql query, you have to convert it to the correct format first before adding it to the parameter array.

### What is the new behavior?

The code now auto converts bools and dates, when added to the parameter array, to the correct format
fixes #5211

### Other information?

I only made it auto convert bools and dates however, if there are other types that need to be converted that I do not know about let me know and I can add them.

### DIFF
--- a/site/app/libraries/database/AbstractDatabase.php
+++ b/site/app/libraries/database/AbstractDatabase.php
@@ -143,6 +143,16 @@ abstract class AbstractDatabase {
     public function query($query, $parameters = array()) {
         try {
             $this->query_count++;
+            foreach ($parameters as &$parameter) {
+                if (gettype($parameter) === "boolean") {
+                    $parameter = $this->convertBoolean($parameter);
+                }
+                elseif (gettype($parameter) === "object") {
+                    if (get_class($parameter) === "DateTime") {
+                        $parameter = $parameter->format("Y-m-d H:i:sO");
+                    }
+                }
+            }
             $this->all_queries[] = array($query, $parameters);
             $statement = $this->link->prepare($query);
             $result = $statement->execute($parameters);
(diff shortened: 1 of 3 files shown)


site/app/libraries/database/AbstractDatabase.php
```diff
@@ -143,6 +143,16 @@ public function isConnected() {
     public function query($query, $parameters = array()) {
         try {
             $this->query_count++;
+            foreach ($parameters as &$parameter) {
+                if (gettype($parameter) === "boolean") {
+                    $parameter = $this->convertBoolean($parameter);
+                }
+                elseif (gettype($parameter) === "object") {
+                    if (get_class($parameter) === "DateTime") {
+                        $parameter = $parameter->format("Y-m-d H:i:sO") . " " . $parameter->getTimezone()->getName();
```
MasterOdin, line 152: Looking at this, I believe we can just rid of the concatenation (so just have `$parameter => $parameter->format("Y-m-d H:i:sO");`) as the `O` provides us the UTC offset and I think the DB is just stripping off the concatenation anyway.

PR #5222, opened 2020-04-01
[Feature:TAGrading] Multiple submitted files

### LINKED ISSUE #4659: Popup grade view for non handwritten solutions
To provide a quick way for an instructor to see that a student submitted extra files
beyond the original scanned test. This way a student doesn't have a question mistakenly
marked with a 0.

Have the grading view detect that more files have been uploaded for this student
and if they have show a popup when the grader selects the next student

Have the students write "laptop" in the box on the test but often students either forget
or only put it on the cover sheet

matgar1997:

It does not display anything extra if multiple files are in the submissions directory

Closes #4659

### What is the new behavior?
A banner appears if 2 or more files were submitted, mainly used for tests with students who use laptops but forget to mention that on a blank test

### Other information?

Confirmed with students who had no submissions, exactly 1 file, more than 1 file submitted


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -789,7 +788,24 @@ public function hwGradingPage(Gradeable $gradeable, GradedGradeable $graded_grad
                 "message" => "Late Submission (No on time submission available)"
             ]);
         }
+        elseif ($graded_gradeable->getAutoGradedGradeable()->hasSubmission() && count($display_version_instance->getFiles()["submissions"]) > 1 && $graded_gradeable->getGradeable()->isScannedExam()) {
```
shailpatels, line 791: One more issue, this appears to be flagging the other files that get uploaded with bulk uploads as extra files when they shouldn't be.
you should ignore the following files:
 ```
.submit.timestamp
bulk_upload_data.json
`

PR #5254, opened 2020-04-14
[Refactor:System] Removes hardcoded instances of Submitty install_dir

shailpatels:

The Submitty install directory is hardcoded accross submitty to be the default `/usr/local/submitty/` and is not read from the `config/submitty.json` file.

### What is the new behavior?
Modifies the `config` model to include functionality to read from the config file and set a variable for the submitty install dir path. Adds a method to get this and removes the hardcoded usage accross submitty.

### Other information?
Python scripts in sbin should also dynamically read the location of the install dir.

Hopefully not

Tested bulk uploads, sites that were affected (gloabl view and creating a gradeable), and updated tests


sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
```diff
@@ -257,7 +257,7 @@ def run_job(self):
         logger.write_to_log(log_file_path, log_msg)
         #create paths
         try:
-            with open("/usr/local/submitty/config/submitty.json", encoding='utf-8') as data_file:
+            with open(os.path.join(INSTALL_DIR, "config", "submitty.json"), encoding='utf-8') as data_file:
                 CONFIG = json.loads(data_file.read())
```
MasterOdin, line 261: Remove this and just use the already available `INSTALL_DIR` or `DATA_DIR` variables for whatever path you want in this job.

site/app/models/Config.php
```diff
@@ -224,6 +226,10 @@ class Config extends AbstractModel {
     protected $queue_contact_info;
     /** @prop @var string */
     protected $queue_message;
+    /** @prop @var string */
+    protected $submitty_install_dir;
```
MasterOdin, line 230: should be `$submitty_install_path` to match the other path properties (e.g. `submitty_path`, `course_path`)

site/app/models/Config.php
```diff
@@ -274,6 +280,8 @@ public function loadMasterConfigs($config_path) {
 
         $this->base_url = $submitty_json['submission_url'];
         $this->submitty_path = $submitty_json['submitty_data_dir'];
+        $this->submitty_install_dir = $submitty_json['submitty_install_dir'];
```
MasterOdin, line 283: need to add this to the list of paths that are validated and then `rtrim` below on lines 332 - 337. Feel free to re-arrange the lines that set paths to be closer to that block to make this connection more clear.

shailpatels: thanks for pointing this out!

PR #5276, opened 2020-04-21
[Bugfix:TAgrading] Peer grading team interface fixes

jchulton:

Peer team rotating section assignments will produce a robot when a student tries to grade them.

Students can see all peer assignments in the navigation menu if they are assigned to grade one peer assignment

### What is the new behavior?
Fixes these errors: peer team rotatin section assignments can be graded properly, students can only see gradeables they are assigned to grade in the navigation menu.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -1744,6 +1744,11 @@ ORDER BY g.sections_rotating_id, g.user_id",
     public function getUsersByRotatingSections($sections, $orderBy = "rotating_section") {
         $return = array();
         if (count($sections) > 0) {
+            $orderBy = str_replace(
+                "Peer",
+                1,
+                $orderBy
+            );
             $placeholders = $this->createParamaterList(count($sections));
             $this->course_db->query("SELECT * FROM users AS u WHERE rotating_section IN {$placeholders} ORDER BY {$orderBy}", $sections);
             foreach ($this->course_db->rows() as $row) {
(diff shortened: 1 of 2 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1744,6 +1744,11 @@ public function getRotatingSectionsForGradeableAndUser($g_id, $user_id = null) {
     public function getUsersByRotatingSections($sections, $orderBy = "rotating_section") {
         $return = array();
         if (count($sections) > 0) {
+            $orderBy = str_replace(
+                "Peer",
+                1,
+                $orderBy
+            );
```
MasterOdin, line 1751: Why was this necessary? Where this function being called that the `$orderBy` parameter has the word "Peer" in it?

jchulton: On line 526 of ElectronicGraderController.php. For team peer graders, we named the rotating section Peer so that it could be differentiated from non-peer rotating sections. There are other ways we could differentiate that I am working on

MasterOdin: Is that change in some future PR that you and @emaicus are working on as line 526 of ElectronicGraderController.php is just this:
https://github.com/Submitty/Submitty/blob/c9873ecbe67a9129f62450ed134034e60dd72902/site/app/controllers/grading/ElectronicGraderController.php#L526

where there is no second parameter to the function call...

PR #5298, opened 2020-04-24
[Bugfix:Autograding] Ensure results json is utf8 encoded

emaicus:
This PR makes certain that the results json created by ```main_validator``` is utf-8 encoded by reading and re-writing the file during archival. Non utf-8 characters are stripped when reading the file.

### DIFF
--- a/autograder/autograder/autograding_utils.py
+++ b/autograder/autograder/autograding_utils.py
@@ -16,6 +16,7 @@ import zipfile
 import stat
 import subprocess
 import shutil
+import codecs
 import glob
 
 def just_write_grade_history(json_file,assignment_deadline,submission_time,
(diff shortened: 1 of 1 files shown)


autograder/autograder/autograding_utils.py
```diff
@@ -389,7 +390,16 @@ def archive_autograding_results(working_directory, job_id, which_untrusted, is_b
...
+
+            # Make certain results.json is utf-8 encoded.
+            results_json_path = os.path.join(tmp_work, 'results.json')
+            with codecs.open(results_json_path, 'r', encoding='utf-8', errors='ignore') as infile:
+                results_str = "".join(line.rstrip() for line in infile)
+                results_obj = json.loads(results_str)
+            with open(results_json_path, 'w') as outfile:
+                json.dump(results_obj, outfile, indent=4)
+
+            shutil.move(results_json_path, os.path.join(tmp_results, "results.json"))
```
MasterOdin, line 402: Would something like the following work?
```
with codecs.open(results_json_path, 'r', encoding='utf-8', errors='ignore') as infile, \
        open(os.path.join(tmp_results, 'results.json'), 'w') as outfile:
    outfile.write(infile.read())
```

PR #5307, opened 2020-04-27
[Bugfix:Autograding] Fix random input/output

emaicus:
After a recent change a number of new files were being copied in while autograding (e.g. ```user_assignment_access.json```).  However, autograding crashed when these files did not exist, as is the case when generating random input/and expected output files. 

This PR adds logic so that a crash does not occur when these files do not exist.


emaicus:
Encountered the issue described in #5308 while fixing this.

Issue #5238, opened 2020-04-07
Autosave on Notebook Gradeable


bmcutler:
**What problem are you trying to solve with Submitty**
If a student is taking a lengthy exam (e.g. 2 hours) that is coded up as a notebook gradeable (typing lengthy answers in form boxes), it would be nice to have some form of autosave, to ensure the student doesn't lose their work.  

Yes, the student can currently manually press "save" (which makes a full submission & new version).  But this requires conscious effort, and the student might not know if this is allowed/encouraged (the instructor might limit the number of penalty free submissions).

Also, ideally this would work even if the student has spotty internet, or closes the laptop lid, etc.

Let's discuss possible implementation solutions.

This might be tangentially related to #1635.  The instructor might specify a frequency and/or specific time(s) for autosave/autosubmit (e.g. at the gradeable deadline)

MasterOdin:
The easiest (and best?) would be to just use the builtin [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) API with an attached, writing to it every handful of seconds as they type, doing something like in JS pseudo-code:
```javascript
let notebook_data = JSON.parse(localStorage.getItem('notebook_data'));
let gradeable_data = {
  id: gradeable_id
};
for (const idx in notebook_data) {
  if (notebook_data[idx].id === 'b') {
    gradeable_data = notebook_data[idx];
    notebook_data.splice(idx, 1);
  }
}

notebook_data.unshift(gradeable_data);

if (notebook_data.length > 5) {
  notebook_data.pop();
}

let timeout;
notebook_textarea.addEventListener('change', () => {
  if (!timeout) {
    timeout = setTimeout(() => {
      notebook_data[gradeable_id][notebook_textarea.dataset.id] = notebook_textarea.value;
      localStorage.setItem('notebook_data', JSON.stringify(notebook_data));
    }, 5000);
  }
}
```

This is designed that we maintain a couple of gradeables (in-case they are doing stuff concurrently for whatever reason), but that we cannot let localStorage grow unbounded as we are limited to 5MB on mobile.

There are optimizations here in that within the timeout function, we probably want to just gather all form elements at once, and so can just use one global timeout object throughout.

bmcutler:
Maybe there are 3 parts to this:

1) A warning if the user leaves the page with unsaved / unsubmitted work (Noted by @MasterOdin noted in an email)

2) An auto save of the content boxes to local storage.  (Described in previous comment)

[...]

PR #5323, opened 2020-05-01
[Feature:Submissions] Notebook Gradeable Autosave + Leave Without Submission Warning

hdrodz97:

See issue #5238

Long story short, for notebook gradeables, if one navigates away from the page without submitting, then all of their work is lost. This in particular becomes an issue with unintentional navigation: some examples include accidentally pressing Ctrl+R on the keyboard (which reloads the page) instead of typing a capital R or accidentally tapping the "back" button on mobile.

### What is the new behavior?

Two new behaviors are introduced to minimize inconvenience:

First, notebook gradeables now autosave to localStorage on every input change. When the notebook gradeable is loaded, if there is any local autosaved data present, then that data will be loaded into the gradeable so that a user may leave the page without losing all of their work.

Secondly, when any change is made to the notebook gradeable, then the page will display an "are you sure you would like to leave?" message. This notifies the user that the changes they have made are not yet submitted to the server.

### Other information?

This has been tested thus far on Google Chrome, Mozilla Firefox, and Microsoft Edge (the non-Chromium version) with mostly everything working as intended. No testing has been done on Safari yet as I lack any modern Mac or iOS devices.

There is one minor quirk in Chrome and Edge where the leave confirmation dialog box will not display if no user-initiated interaction (such as clicking and typing, *but not scrolling*) has occurred on the page yet. In the case of Chrome (and presumably any Chromium-based browser), this is intentionally disallowed behavior that is shown in the developer console. Edge does not display any message in the console, but the behavior seems to follow the same cues. To see this in action, in the "Notebook Basic" gradeable under the "DEVELOPMENT" course on the Vagrant machine, change any answer, leave the page, open the page again, and immediately leave again. In Firefox, this will properly display the warning box; however, Chrome and Edge will not show anything.

Because of this behavior as well as the fact that the "are you sure you would like to leave?" message is not customizable in modern browsers, I'm thinking we might want to additionally have a banner warning users in no uncertain terms that their changes have not been submitted yet. This is the only major thing keeping me from not tagging this as WIP.


site/public/js/gradeable-notebook.js
```diff
@@ -99,39 +188,47 @@ $(document).ready(function () {
...
-          $(recent_button_id).attr("disabled", false);
+            $(recent_button_id).attr("disabled", false);
+            $("#submit").attr("disabled", false);
+            window.onbeforeunload = () => true;
         }
-    });
+    }));
+
+    // Saving to LocalStorage on every keypress isn't fantastic, but some 
+    // quick-and-dirty profiling seems to indicate it's fine...
```
elihschiff, line 230: If you end up using what I mention above, I dont think you need this comment anymore

hdrodz97:
> You mention that you dont think we will run out of space in local storage however what happens if the same student uses lots of notebook gradeables over time, is there a chance they could run out of space? If that is a worry, maybe there needs to be a system where, maybe on page load, it checks all notebook gradeables saved in local storage and removes anything over 1 month (or any time-frame) old. This would mean adding a date of when it was saved to the json in local storage

So on this note, this shouldn't be an issue as the autosave data is cleared when the submit button is clicked. Of course, if the user doesn't ever click the submit button and does this for several gradeables then this would definitely be a worry. That being said, having a fail-safe is probably a good idea.

site/public/js/gradeable-notebook.js
```diff
@@ -65,6 +65,89 @@ function setCodeBox(codebox_id, state)
...
+ */
+function saveToLocal() {
+    if (autosaveEnabled) {
+        localStorage.setItem(AUTOSAVE_KEY, JSON.stringify({
+            multiple_choice: gatherInputAnswersByType("multiple_choice"),
+            short_answer: gatherInputAnswersByType("short_answer"),
+            codebox: gatherInputAnswersByType("codebox")
+        }));
+    }
+}
```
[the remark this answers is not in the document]
hdrodz97: I think this is a pretty good solution to this issue; I'll do this instead for the text boxes.

hdrodz97:
> Works well on firefox on ubuntu. The following errors occurred using chrome on ubuntu:
> After installing and navigating to the Notebook Basic, the following error occurred.
> 
> ```
> Uncaught SyntaxError: Unexpected end of JSON input
>     at JSON.parse (<anonymous>)
>     at cleanupAutosaveHistory (gradeable-notebook.js?v=1588463395:112)
>     at HTMLDocument.<anonymous> (gradeable-notebook.js?v=1588463395:405)
>     at e (jquery.min.js?v=1588463562:2)
>     at t (jquery.min.js?v=1588463562:2)
> ```

Looking through the code, I noticed we're using localStorage in several other places, so this is just an artifact of some of the things we're storing not being JSON objects. I'll put some extra checks to make sure we're not clearing out anything unrelated.

> I also noticed that the leave prompt on chrome says "leave site", even if you are still navigating within Submitty.

This is an unfortunate limitation of leave prompts. Customizing the text on them has been deprecated for a while, and all major browsers offer no support for it anymore (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Browser_compatibility)). Judging from [this StackOverflow answer](https://stackoverflow.com/a/821227) as well as from the little bit I've searched on the topic, this event seems to be the only tool we have to stop the user from accidentally leaving the page.

bmcutler:
Still needs to be tested on safari

site/public/js/gradeable-notebook.js
```diff
@@ -65,6 +65,161 @@ function setCodeBox(codebox_id, state)
     }
 }
 
+const AUTOSAVE_KEY = `${window.location.pathname}-autosave`;
```
MasterOdin, line 68:
```suggestion
const AUTOSAVE_KEY = `${window.location.pathname}-notebook-autosave`;
```

site/public/js/gradeable-notebook.js
```diff
@@ -65,6 +65,161 @@ function setCodeBox(codebox_id, state)
...
+                // JSON.parse() will throw a SyntaxError. We don't care about
+                // these entries, so skip over them.
+                if (e instanceof SyntaxError) {
+                    continue;
+                } else {
+                    throw e;
+                }
+            }
+        }
+        toDelete.forEach(localStorage.removeItem, localStorage);
```
MasterOdin, line 135: Let's use the [endswith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) to limit the keys we care about (just need unique suffix) instead of parsing all keys. This has the added benefit of that if any value in localStorage related to this feature gets corrupted for whatever reason, we can properly just dispose of it instead of it hanging around infinitely.
```suggestion
        for (let i = 0; i < localStorage.length; i++) {
            const key = localStorage.key(i);
            if (!key.endswith('-notebook-autosave')) {
                continue
            }
            try {
                const { timestamp } = JSON.parse(localStorage.getItem(key));

                if (msToDays(Date.now() - timestamp) > 30) {
                    toDelete.push(key);
                }
            } catch (e) {
                // item has gotten messed up, just delete it as it's unusable
                toDelete.push(key);
            }
        }
        toDelete.forEach(localStorage.removeItem, localStorage);
```

MasterOdin:
Can fix that if someone ever actually notices, as fixing it would require a decent overall of how the things get displayed, and also who knows how it'll pay with randomized notebooks. Maybe make an issue, but probably fine to just ignore it for now.

bmcutler:
@elihschiff 
This will actually likely happen when the instructor is incrementally creating & testing the gradeable and could be very confusing.
Yes its not as critical as a student seeing it during an exam, but is worth fixing.
We can make an issue for future if its not trivial to fix now.

MasterOdin:
I think in that case, let's just do a simple check of if the length of fields != the number saved in autosave, do not use autosave? That probably meets with instructor's expectations in testing the output while building up the notebook?

bmcutler:
Sure....  but I can also see us swapping the order of some questions, so the # of questions/fields is the same.

Again, if its not trivial to fix, that's ok.  I'd rather get this PR in.

PR #5353, opened 2020-05-09
[Feature:System] System update UI

### LINKED ISSUE #5350: Notify system superuser when update is available
Currently, there is no way for admins to know when an update is available beyond looking on GitHub

Provide a message in a nonintrusive way to the user with the most privileges within the system

This was not an issue we are currently addressing beyond GitHub releases

matgar1997:
* [x] The PR title and message follows our [guidelines]

Closes #5350

No way to check easily for updates within submitty

### What is the new behavior?
superusers can check for recent updates as well as see what changes
It compares the current version with the most recent version available

### Other information?

1 Login with account `superuser` `superuser`
2 view "your courses" page

Screenshot of new top level page
image

*Needs discussion on how to display raw markdown release notes to user*


site/app/templates/admin/SystemUpdate.twig
```diff
@@ -0,0 +1,43 @@
...
+
+        } else {
+            document.getElementById('text').innerHTML = "A newer version of Submitty is available";
+            document.getElementById('tag').innerHTML = "Current Version Installed: " + latest_tag + "\n" + "Most Recent Version: " + myArr[0]["tag_name"];
+        }
+
+        document.getElementById('release_notes').innerHTML = parseReleaseNotes(myArr[0]["body"]);
+        document.getElementById('release_notes_container').style.display = "block"
+        document.getElementById('update_status').style.display = "block"
+    }
```
MasterOdin, line 37:
```suggestion
    async function getReleases(latest_tag) {
        try {
            const req = await fetch('https://api.github.com/repos/Submitty/Submitty/releases');
            const release = await req.json()[0];
            if (release["tag_name"] === latest_tag) {
                document.getElementById('text').innerHTML = "Submitty is up to date!";
            }
            else {
                document.getElementById('text').innerHTML = "A newer version of Submitty is available";
                document.getElementById('tag').innerHTML = "Current Version Installed: " + latest_tag + "\n" + "Most Recent Version: " + release["tag_name"];
            }
        }
        catch (exc) {
            document.getElementById('text').innerHTML = 'Could not fetch version information';
        }

        document.getElementById('release_notes').innerHTML = parseReleaseNotes(release["body"]);
        document.getElementById('release_notes_container').style.display = "block"
        document.getElementById('update_status').style.display = "block"
    }
```

site/app/templates/admin/SystemUpdate.twig
```diff
@@ -0,0 +1,43 @@
...
+
+    function parseReleaseNotes(raw_str) {
+        str = raw_str.replace(/\n/g, "<br />").replace(/\r/g, "").replace("*", "").replace(RegExp("\[(.*?)\]"), "");
+        return str;
```
MasterOdin, line 41:
```suggestion
        return raw_str.replace(/\n/g, "<br />").replace(/\r/g, "").replace("*", "").replace(RegExp("\[(.*?)\]"), "");
```

site/app/templates/admin/SystemUpdate.twig
```diff
@@ -0,0 +1,43 @@
...
+        <h4 id="tag"></label> 
+    </div>
+
+    <div class="content" id="release_notes_container" hidden>
+        <label id="release_notes"></label>
+    </div>
+</div>
+
+
+<script>
```
MasterOdin, line 17: This should not be inline. Move into an existing js file or its own.

site/app/templates/admin/SystemUpdate.twig
MasterOdin: Let's also trigger this function on page load, and not require instructor to press a button. If they're coming to this page, chances are they want to know version information and if update is available.

MasterOdin: So for some words on the above:
1. The [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) api is awesome and we should use that instead of jQuery or old school `XMLHttpRequest` as it's a sane, promise based interface
1. We should handle the case that we cannot fetch the version information for whatever reason (no internet, github is down, etc.)
1. Use `const` / `let` over `var` and avoid wacky scoping rules

site/app/controllers/GlobalController.php
```diff
@@ -321,6 +321,16 @@ public function header() {
                 ]);
             }
 
+            if($this->core->getUser()->getAccessLevel() == 1) {
```
MasterOdin, line 324: There's a constant in `User` class for superuser level

site/app/templates/admin/SystemUpdate.twig
matgar1997: After moving it would I need $this->core->getOutput()->addInternalJs in homePageController, since there isnt a php file for the new page at the moment

PR #5354, opened 2020-05-09
[UI/UX] Improved a number of pages when in dark mode

elihschiff:

There are a number of issues with dark mode.

### What is the new behavior?

Based on user feedback I switched the check-boxes to enable dark mode to a dropdown.
The dropdown has a number of options including allowing Submitty to follow the device theme. (This mostly shows up on phones where they are light theme for half the day and dark theme at night)
image

I also removed a number of shadows across the the website (only in dark mode) which looks a lot better overall.
image
This is what it used to look like
image

White buttons are no longer white and are instead a less bright color
image

Text at the top of the office hours queue is now white on blue instead of black on blue which was hard to read
image

### DIFF
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -65,6 +65,8 @@
 
         //determines if the user has a set theme
         function detectColorScheme(){
+            document.documentElement.setAttribute("data-theme", "");
+            document.documentElement.setAttribute("data-black_mode", "");
             let theme="light";    //default to light
             //local storage is used to override OS theme settings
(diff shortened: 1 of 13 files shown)


elihschiff:
I fixed most of the errors you listed. The one thing I am having trouble with is the red text on a dark background. The issue is that I cannot switch to background away from dark because that is the entire point, and I cannot switch the red away from red because the red color signifies something (it is not just for looks)
I switched the red to this more pink color which I think is easier to read. WAVE says the contrast is good so I think it should be fine to use. Here is an example:
image

Issue #4961, opened 2020-01-24
Rainbow Grades Instructor UI # of items in category


bmcutler:
Early in the semester, before all gradeables for a category have been created, the configuration might look like this:
<img width="305" alt="Screen Shot 2020-01-24 at 12 04 41 AM" src="https://user-images.githubusercontent.com/5871417/73045746-21d8b980-3e3d-11ea-8008-e028977c0a06.png">

We will have 10 homeworks, but so far we only have created 2 of them.  If homeworks are 50% of the grade, we have only graded 2/10 homeworks, so only 10% of the overall semester grade is available.

It seems to work initially, but when I go back to the UI it resets the first box (items) to 2.  Then the math gets off for rainbow grades.

Kevin-MM:
Will investigate where they category information is being pulled from and adjust as needed.

technostarbtd5:
I've been looking over this issue and I don't see any components of it that haven't been successfully addressed by a pull request. Are we aware of any lingering problems with this issue preventing it from being closed?

PR #5366, opened 2020-05-11
Add per-gradeable curve inputs to the rainbow grades web UI

Kevin-MM:
Partially handles #4961

No way to adjust the per-gradeable curve from the rainbow grades web UI.

### What is the new behavior?
Input boxes are now available for each gradeable in the web UI to allow instructors to enter per-gradeable curves from the web browser.

per_gradeable_curve


bmcutler:
@Kevin-MM 
Can you please add screenshots to your top post (for initial review comments & longer term documentation for release notes -- PRs are linked) thanks!

holzbh:
As a comment I'm also not sure on the curve "icon". I don't have an alternate suggestion off the top of my head, but intuitively to me, when I clicked the button I expected it to show me the grade distribution histograms.

Kevin-MM:
- I've updated the validation rule to now check to make sure that all input boxes are less than or equal to the max score, and also less than or equal to the previous input box.

- I'm unable to replicate log error so I'm unsure about that right now.

- I'll wait for further guidance on which icon to use.

- There are a few assertions in the rainbow grades code that seem to imply 0 as the bottom floor for the per-gradeable curves

https://github.com/Submitty/RainbowGrades/blob/fd4cab8f62e81b93a27a70f1f15570ad430ea0f7/main.cpp#L696-L699

and 

https://github.com/Submitty/RainbowGrades/blob/fd4cab8f62e81b93a27a70f1f15570ad430ea0f7/output.cpp#L151-L167

holzbh:
If the Rainbow Grades code doesn't allow scores less than zero than your form shouldn't either - I was able to save a cutoff of `-300` without any issues. I think it's fine to assume anything less than zero is a problem unless @bmcutler says otherwise, in which case hitting Save should cause an error if I use a negative cutoff.

Kevin-MM:
Ready to review again.  I did more work to make the user interface more intuitive.  Curve input boxes are now only displayed when the corresponding box is check in the 'display benchmarks' section.

holzbh:
Regarding the log error - I've recommended @Kevin-MM do a destroy&up, since the issue I'm observing is that during setup the `/var/submitty/logs/rainbow_grades` subdirectory isn't being created, which is causing the auto_rainbow_grades.py script to fail when it tries to open the log file for the day. For the sake of going forward with testing I manually created the subdirectory in my VM and was able to proceed.

The checkboxes seem to work correctly (until I pick one of the `lowest...` benchmarks, the curve icon isn't there). I tried this for all single benchmarks and then some random combinations. So that's good. The customization.json is also correctly generated.

However, as an example I tried just picking `perfect` and `lowest c-` as benchmarks, only added in the Tests category in `sample`, and gave one of the tests a `3` (it correctly only showed me a box for lowest C-). The JSON generated looked correct. On refreshing the page though, the box was empty instead of having a `3`, so there's an issue with loading values back in. I believe what's happening is the first value is being assumed to be the "lowest A-" instead of being assigned to the first visible text box for that gradeable. In fact, if you then click "lowest A-" so it's checked, the text box for lowest A- will have the first value. So you need to actually combine knowledge from the global `display_benchmark` and per-gradeable `curve` fields to know which text fields to populate.

After messing with values for a while, and only having an A- and C- benchmark (no B- benchmark) I hit save again, and got
```
process_grades.out: /usr/local/submitty/GIT_CHECKOUT/RainbowGrades/main.cpp:699: void preprocesscustomizationfile(const string&, std::vector&): Assertion `p_score >= a_score && a_score >= b_score && b_score >= c_score && c_score >= d_score' failed.
```

Looks like Rainbow Grades doesn't allow for gaps in selected benchmarks (so just C- is fine, just A- is fine, {A-, B-, C-} is fine, but {A-, C-} is not OK because the missing B- benchmark confuses Rainbow Grades). I guess the web form should try to enforce that rule as well.

Kevin-MM:
@holzbh I did a fresh destroy and up, but I am still unable to replicate the logging issue.  When you referred to the logging directory did you mean `/var/local/submitty/logs/rainbow_grades`?  What is your typical workflow when installing a new branch?  For me I will checkout the branch then run `/usr/local/submitty/.setup/INSTALL_SUBMITTY.sh`.  I can see the log directory being generated here:

https://github.com/Submitty/Submitty/blob/383b5a6f2c64c7307265a8ba7902ea533b04fb9e/.setup/INSTALL_SUBMITTY_HELPER.sh#L213-L229

I can also see that `install_system.sh` is lacking lines for rainbow_grades logging here:

https://github.com/Submitty/Submitty/blob/383b5a6f2c64c7307265a8ba7902ea533b04fb9e/.setup/install_system.sh#L714-L774

I'm unsure of the exact interplay between these two files, what are your ideas here?

Kevin-MM:
@holzbh The per-gradeable curve values should now land in the correct boxes even if there are gaps between which display benchmarks are selected.

I wasn't able to replicate the error you described with have {A-, C-} selected.

holzbh:
My workflow for a new branch is

```
vagrant destroy
vagrant up
```

You should not have to run any .sh files after the `vagrant up` install process completes - if you need to, that's a problem. I can confirm that IF I run INSTALL_SUBMITTY.sh manually after vagrant up, the directory exists.

You're better off asking someone like @MasterOdin about the organization of those install scripts - I don't typically play with them, though my understanding was install_system.sh should invoke the generated INSTALL_SUBMITTY.sh during the setup process.

I'll take a look at the rest of your comments in a separate reply.

holzbh:
Still seeing the other error. To reproduce, from my fresh VM (after creating the log file)

1. Go to Sample
2. Go to Grade Reports
3. Go to Web-Based Rainbow Grades Generation
4. Click the checkboxes for `lowest a-` and `lowest c-`
5. Scroll down to the categories and add `Test (5 items)`
6. Click the curve box for `future_no_tas_test`
7. Put `3` in the `A-` box and `1` in the `C-` box.
8. Scroll to the top of the page and click Save.
9. After a while you should get a failure message and be able to click `Log` to see the assertion failure mentioned before.

My `customization.json` at this point looks like:

```
{
    "section": {
        "1": "1",
        "2": "2",
        "3": "3",
        "4": "4",
        "5": "5",
        "6": "6",
        "7": "7",
        "8": "8",
        "9": "9",
        "10": "10"
    },
    "display_benchmark": [
        "lowest_a-",
        "lowest_c-"
    ],
    "messages": [],
    "display": [
        "grade_summary",
        "grade_details"
    ],
    "benchmark_percent": {
        "lowest_a-": 0.9,
        "lowest_c-": 0.7
    },
    "gradeables": [
        {
            "type": "test",
            "count": 5,
            "percent": 0,
            "ids": [
                {
                    "max": 5,
                    "release_date": "9998-12-31 23:59:59-0500",
                    "id": "future_no_tas_test",
                    "curve": [
                        3,
                        1
                    ]
                },
                {
                    "max": 5,
                    "release_date": "9998-12-31 23:59:59-0500",
                    "id": "future_tas_test"
                },
                {
                    "max": 5,
                    "release_date": "1972-01-01 23:59:59-0500",
                    "id": "grades_released_test"
                },
                {
                    "max": 5,
                    "release_date": "9998-12-31 23:59:59-0500",
                    "id": "grading_test"
                },
                {
                    "max": 5,
                    "release_date": "9998-12-31 23:59:59-0500",
                    "id": "grading_test_rotating"
                }
            ]
        }
    ]
}
```

MasterOdin:
The problem arises that `INSTALL_SUBMITTY.sh` creates the folder, and then `install_system.sh` (in the block shown above) deletes all log folders and then sets up symlink folders one at a time so that the logs are all viewable outside of Vagrant on the host machine. Once INSTALL_SUBMITTY.sh is run again, the folder is re-created and all is well. It's a bug, but one that only affects Vagrant, and only until you re-run INSTALL_SUBMITTY.sh as you both discovered above.

I've also created #5391 to track making this system a bit more robust and simpler. A quick fix here would be to add an entry into the lines under `install_system.sh` that create the symlink for the folder in the same vein as the rest of `ln -s` lines.

PR #5375, opened 2020-05-12
 [Bugfix:InstructorUI] Stops submitty from being used in iframe and stops external domain redirects on login

### LINKED ISSUE #5264: [Security Bug] UI redressing/Clickjacking Attack Possible at Login Authentication Page
This is @humblelad fellow Gsoc aspirant. 
While doing security assessment, I found that main login page is vulnerable to clickjacking attacks as the page  can be loaded within an iframe. 

**To Reproduce**
`<html>

	<style>
		iframe { 
		width: 800px; 
		height: 500px; 
		position: absolute; 
		top: 0; left: 0; 
		filter: alpha(opacity=50); 
		opacity: 0.5; 
		}  
	</style>
	<iframe src="http://localhost:1501/authentication/login">
</html>`

save this as click.html and open in a new window. We see that the page gets loaded and the creds can be easily stolen by attackers. 

**Configuration**
 - OS: Mac Catalina
 - Browser Firefox Developer Edition

**Screenshots**
sdf

The mitigation may involve adding proper X Frame Origin Headers as either "deny" or "sameorigin" or have an CSP Policy.

### LINKED ISSUE #5265: [Security Bug] Open Redirection Vulnerability at Login Page
**Describe the bug**
It is possible to redirect a user to an attacker owned domain and trick the user. I am investigating on the potential chaining of this bug to perform other attacks.

**Expected behavior**
Redirection to external urls must not be allowed.

**To Reproduce**
This occurs at the  main login page when the user enters an invalid username/password.
The url gets changed to 
`http://localhost:1501/authentication/login?old=http%3A%2F%2Flocalhost%3A1501%2Fhome`
An attacker can change the url(to any attacker owned domain which may mimic submitty interface) to something like 
`localhost:1501/authentication/login?old=http%3A%2F%2Flgoogle.com` 
and trick redirection to google.com  users after they enter their creds.

**Additional context**
This can be prevented either by-
1. Adding regex checks on url and whitelisting the url.
2. Completely removing GET request based url redirection. 

I am trying to fix this but thought to create this issue for others to contribute the fix if interested.

elihschiff:

Submitty has the chance of redirecting users to other domains when they log in.
Submitty could be put into an iframe to trick users into giving their passwords to another website

### What is the new behavior?

Redirects now always check that the url they are redirecting to on login comes from submitty's domain
The Submitty login page will no longer work in an iframe
fixes #5265
fixes #5264

### Other information?

To test:
Log out of submitty 
make an html file with the content of this issue #5264 
the page should not work
then go to this url http://localhost:1501/authentication/login?old=https://duck.com and log in
it should not redirect to duck.com and just go to the main submitty homepage

### DIFF
--- a/site/app/controllers/AuthenticationController.php
+++ b/site/app/controllers/AuthenticationController.php
@@ -75,7 +75,6 @@ class AuthenticationController extends AbstractController {
         if (!is_null($old) && !Utils::startsWith(urldecode($old), $this->core->getConfig()->getBaseUrl())) {
             $old = null;
         }
-        header('X-Frame-Options: DENY'); // Stops clickjacking on all pages
         return MultiResponse::webOnlyResponse(
             new WebResponse('Authentication', 'loginForm', $old)
         );
--- a/site/app/controllers/AuthenticationController.php
+++ b/site/app/controllers/AuthenticationController.php
@@ -72,6 +72,10 @@ class AuthenticationController extends AbstractController {
      * @return MultiResponse
      */
     public function loginForm($old = null) {
+        if (!is_null($old) && !Utils::startsWith($old, $this->core->getConfig()->getBaseUrl())) {
+            $old = null;
+        }
+        header('X-Frame-Options: DENY'); // Stops clickjacking on all pages
         return MultiResponse::webOnlyResponse(
             new WebResponse('Authentication', 'loginForm', $old)
         );
@@ -89,6 +93,9 @@ class AuthenticationController extends AbstractController {
      * @return MultiResponse
      */
     public function checkLogin($old = null) {
+        if (!is_null($old) && !Utils::startsWith($old, $this->core->getConfig()->getBaseUrl())) {
+            $old = null;
+        }
         if (isset($old)) {
             $old = urldecode($old);
         }
--- a/site/app/controllers/AuthenticationController.php
+++ b/site/app/controllers/AuthenticationController.php
@@ -72,7 +72,7 @@ class AuthenticationController extends AbstractController {
      * @return MultiResponse
      */
     public function loginForm($old = null) {
-        if (!is_null($old) && !Utils::startsWith($old, $this->core->getConfig()->getBaseUrl())) {
+        if (!is_null($old) && !Utils::startsWith(urldecode($old), $this->core->getConfig()->getBaseUrl())) {
             $old = null;
         }
         header('X-Frame-Options: DENY'); // Stops clickjacking on all pages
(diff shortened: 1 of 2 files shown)


site/app/controllers/AuthenticationController.php
```diff
@@ -72,6 +72,10 @@ public function logout() {
      * @return MultiResponse
      */
     public function loginForm($old = null) {
+        if (!is_null($old) && !Utils::startsWith(urldecode($old), $this->core->getConfig()->getBaseUrl())) {
+            $old = null;
+        }
+        header('X-Frame-Options: DENY'); // Stops clickjacking on all pages
```
MasterOdin, line 78: This should be set in `site/public/index.php`, and let's use `SAMEORIGIN` to continue to allow for Barb's original use-case of embedding Submitty in an iframe on her course website. Ideally this would also be a separate PR, as it's a different security thing as the above, regardless of how small a change it is currently, but we can leave it in the same PR here since it's already started.

site/app/controllers/AuthenticationController.php
```diff
@@ -72,6 +72,10 @@ public function logout() {
      * @return MultiResponse
      */
     public function loginForm($old = null) {
+        if (!is_null($old) && !Utils::startsWith(urldecode($old), $this->core->getConfig()->getBaseUrl())) {
```
shailpatels, line 75: Setting the header is enough to fix this problem, moving it to `index.php` like @MasterOdin suggested should fix the issue of using iframes to use cross-origin hacks on Submitty.

I don't think there's any point of this check once that's in place, the `SAME-ORIGIN` policy will perform the same check and most likely will be more effective than this.

MasterOdin: `SAME-ORIGIN` policy will not stop redirects to other domains.

site/app/controllers/AuthenticationController.php
bmcutler: I don't use iframes anymore -- haven't ever since we moved all the menu options to the left sidebar.  I changed my website to mimc the submitty styling (but no page is a mix of domains).  I don't think we should continue to support external iframes.

elihschiff: I think another issue with SAMEORIGIN is that a member of RPI (or a hacked account) could use their personal homepage to make a fake Submitty on the same origin

site/app/controllers/AuthenticationController.php
elihschiff: This confusion is my bad, I should have split this up into 2 prs. The header line is to stop iframes from working with submitty. All the other lines are to stop the login redirect from sending you to another website after you log in. They are semi unrelated issues and require different fixes.

MasterOdin: It should be only if they run their personal site at the same origin as Submitty, which is that `https://submitty.cs.rpi.edu` has a different origin than `https://rpi.edu/~pevelm`. But given that no one is using iframes, fine to just deny for now and revisit if we have the need.

bmcutler:
@shailpatels @MasterOdin please re-review
Thanks :)

PR #5378, opened 2020-05-13
[Bugfix:TAGrading] Fixing inability to switch to drawing sometimes

jchulton:

Sometimes, when a grader attempts to annotate a pdf and clicks the pencil icon, the pencil will be highlighted, but when the uses tries to draw an annotation, the pencil icon is unhighlighted and the user cannot draw an annotation. This seems to occur when the user tries to interact with the pdf annotation bar before everything is fully loaded (the panel gives no feedback as to when everything is loaded, so it's easy to interact with the page before it's loaded).

I've had trouble reproducing this bug realiable. 

### What is the new behavior?

The grader will be able to select the pencil and annotate as normal all the time.

### DIFF
--- a/site/public/js/pdf/PDFInitToolbar.js
+++ b/site/public/js/pdf/PDFInitToolbar.js
@@ -2,7 +2,6 @@ if (PDFAnnotate.default) {
   PDFAnnotate = PDFAnnotate.default;
 }
 
-var loaded = sessionStorage.getItem('toolbar_loaded');
 window.onbeforeunload = function() {
     sessionStorage.removeItem('toolbar_loaded');
 };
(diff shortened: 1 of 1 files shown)


site/public/js/pdf/PDFInitToolbar.js
```diff
@@ -176,9 +175,7 @@ window.onbeforeunload = function() {
     function handleToolbarClick(e){
         setActiveToolbarItem(e.target.getAttribute('value'));
     }
-    if(!loaded){
```
jchulton, line 179: I do not understand the logic behind loaded here. It seem like the toolbar is only enabled when loaded = false. If loaded = true, it seems like you get the inability to switch to drawing mode. Is there logic I'm missing?

MasterOdin: It's because the implementation for loading multiple PDFs is janky, in that some elements are destroyed, some are not. The toolbar is one of those things and so the this variable is there to prevent adding multiple event listeners to the toolbar elements. For example, without this, each time you load a new PDF (with reloading the page), hitting the zoom button will rotate the PDF by an extra 90 degrees.

Honestly, we're probably at a point where it would be better to just rewrite this whole thing to be cleaner implementation than just trying to fix minor bugs with it.

PR #5380, opened 2020-05-14
[Refactor:Notebooks] Refactor Notebook gradeables to use models

shailpatels:

Notebook gradeables have been programmed in a procedural style and have been mixed into a number of other models and views they shouldn't be. This was done by parsing files and configs and passing around php arrays which has made it harder to debug and add new features. 

### What is the new behavior?
Notebook files have been reorganized and the majority of relevant notebook code has been placed into their own models. A new twig macro has been created from the existing code inside of `submitBox.twig` to - this will allow later feature of viewing what a notebook gradeable

Userspecific notebook inherits from Notebook is the new setup, I'm not sure if this the best design since the notebooks still need to use some functionality from autograding_config which gets generated before we create notebooks from itempools so I'm not sure what the best solution to this problem is.

### Other information?
Testcases are note being parsed correctly on itempool notebooks (autograding isn't fully set up so this isn't a big problem for them)

Creating this now to get some feedback for new organization. 

I didn't rewrite the actual code itself but instead mainly moved it to different locations and wrote a bit of glue to tie the new pieces together. Happy to accept feedback for the design and code so Notebook gradeables work better :)


site/app/models/notebook/AbstractNotebookInput.php
```diff
@@ -1,21 +1,21 @@
...
+namespace app\models\notebook;
 
 use app\libraries\Core;
 use app\models\AbstractModel;
 
 /**
- * Class AbstractGradeableInput
- * @package app\models\gradeable
+ * Class AbstractNotebookInput
+ * @package app\models\notebook
```
MasterOdin, line 10: Please just delete all of these `@package` annotations. We're not generating phpdocs, and they just replicate the information that the namespace already provides.

site/app/models/gradeable/AutogradingConfig.php
```diff
@@ -351,44 +162,42 @@ public function getTestcases() {
...
     }
 
-    /**
-     * Gets the abstract inputs for this configuration
-     * @return AbstractGradeableInput[]
-     */
-    public function getInputs() {
-        return $this->inputs;
+
+    public function getNotebook() {
```
MasterOdin, line 166: `getNotebookConfig`?

shailpatels:
An idea is to have `notebook.php` be a class for `autogradingConfig.php` and have a notebook object as a property of `autogradingConfig.php` that way the notebook models will have access to testcases and can update them when the notebook config has been re-parsed and the item sections have been replaced with notebook components.

Kevin-MM:
Notes about notebook.php
- The ```use app\models\notebook\AbstractGradeableInput;``` is unresolved
- What is the purpose of ```Notebook->buildUserSpecificNotebook()```?  Why not just let the calling code instantiate a user specific notebook directly?  Maybe a notebook factory method or factory class would be appropriate here and would enable use to simplify calling code.
- Need imports for ```NotImplementedException```, ```AuthorizationException```, ```FileNotFoundException```, ```IOException```

Notes about UserSpecificNotebook.php
- There are a few typos in the comments

General Notes:
- I like where you're going with having a different model for each notebook cell type.
- I feel a little weird about ```AbstractTextBox``` having information about whether or not it is a code box.  I don't think the inheritance structure is working correctly when the parent needs to store information about the child (child in this case ```SubmissionCodeBox``` or ```SubmissionTextBox```).
- I think maybe we should have models for the markdown and image type cells also, basically any thing that could be a cell in a notebook page should be modeled, should make it easier to create new cell types in the future by pretty much just adding a model for it.
- I think the addition of notebook.twig was a good move

shailpatels:
Thanks for the feedback @Kevin-MM !

And yeah the AbstractTextBox, AbstractSubmissionBox and all those files were created a while back and were originally part of the gradeable folder, I've just moved them under the new notebook folder but yeah I think I'm going to start rewriting them to be a bit better.

Here's my thought in moving things around, a Notebook cell can either be an input or a static cell. Static cells would be markdown/images/ stuff thats defined by the instructor and we just display

Inputs are things that students will interact with it text boxes, radio inputs & whatever else. For now I think it doesn't make sense to model every time of cell, like images and markdown there probably isn't enough functionality in each to merit their own models. Of course, this could change in the future.

I think for now we can have a more general inheritance structure since notebooks are newer and theres still room to change things and in the future we can make things more strict with a more explicit object separation. I'm a bit afraid of going too deep into an object-oriented style and we just end up solving problems caused by our design in the future instead of actual problems.

image

shailpatels:
So I think I can update the testcases in AutogradingConfig.php without having notebooks extend it. 
The rest of this requires implementing the changes `AbstractTextBox`, `AbstractNotebookInput`, SubmissionMultipleChoice` & `SubmissionTextBook`

This will require combining them since a lot of these are small files without much functionality the second step actually requires rewriting some of the notebook code to use these classes since a lot of the internal code is still procedural and not taking advantage of these classes.

shailpatels:
This PR should be ready to get merged so more work on Notebooks can be done by other developers, the rest of the refactor will focus on cleaning up some of the internal code in the notebook models since this really just re-organized things

PR #5381, opened 2020-05-14
[UI/UX/API:InstructorUI] Peer Matrix Tab

### LINKED ISSUE #5369: Split "Grader Assignment" tab into "Peer Matrix" tab
Currently, peer grading information is on the "Grader Assignment" tab for "Edit Gradeable".
Let's split that functionality onto a new tab.

If an assignment has at least one peer component, let's add a new tab to that menu "Peer Matrix".  In that new panel we will have the place to upload the peer matrix csv and the display of the current peer assignments.

If possible, let's also try to get rid of the need to reload the page after adding your first peer component, to get the page to recognize that this gradeable is now a peer gradeable.
[@emaicus fill in more details here]

listeph:

* [ ] Documentation has been updated/added if relevant (https://github.com/Submitty/submitty.github.io/pull/226)

Currently, after peer components are added to the rubric, the way peers are assigned are by uploading a CSV to the bottom of a prompt from the Grader Assignment tab. After adding your first peer component, the page has to be explicitly reloaded to see the prompt to upload a CSV.

### What is the new behavior?
Closes #5369 
The new behavior is that if there is at least one peer component in the Rubric page, then a Peer Matrix tab will pop up, and that is where the prompt to upload a CSV will be. When you click onto the tab after adding your first peer component, you don't need to reload.

### Other information?
I tested by creating a new gradeable and checking Yes for "Will any or all of this assignment be manually graded (e.g., by the TAs or the instructor)?" Then, I go to the rubric tab and add / delete  normal / peer components and see if the Peer Matrix tab appears / disappears when you expect it to.


site/app/templates/admin/admin_gradeable/AdminGradeableBase.twig
```diff
@@ -53,10 +55,10 @@
 <script>
 
     let adminGradeableNavTab = 0;
-    let navItemsStatus = new Map([['0',true],['1',true],['2',true],['3',true],['4',true]]);
+    var navItemsStatus = new Map([['0',true],['1',true],['2',true],['3',true],['4',true],['5',false]]);
```
shailpatels, line 58: Is there a reason this being changed from `let` to `var`. We typically want to use `let` since JS will scope `var` to the most immediate function body, `let` will be scoped to the most immediate set of brackets.

JS scoping can sometimes get weird causing unexpected errors so using `const` or `let` instead is usually better - here it might not matter as much since they're globally defined in that script

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -0,0 +1,24 @@
+{% if peer %}
```
shailpatels, line 1: Instead of having a server-side check for peer grading what you can do is always load this within a div and then hide it by default (display:none) and then perfom a javascript check to see if there are any peer components and if there are you can unhide.

site/app/templates/admin/admin_gradeable/AdminGradeableBase.twig
shailpatels, line 58: Is there a reason this being changed from 'var' to 'let', the scoping for 'var' works a bit weirder and can lead to some unexpected bugs because JS considered the scope to be the most immediate function instead of the local block you created that variable in so you might get a variable defined where you don't expect to be.

In this case it looks like it doesn't matter too much but we should try and use `let` and `const` because its a stronger signifier

site/app/templates/admin/admin_gradeable/rubric/ElectronicFileRubric.twig
```diff
@@ -30,7 +30,17 @@
 <script>
     $(document).ready(function() {
         loadTemplates().then(function() {
-            reloadInstructorEditRubric($('#g_id').val());
+            reloadInstructorEditRubric($('#g_id').val()).then( function() {
+                $('body').on('DOMSubtreeModified', '.component-container', function() {
+                    let yesPeer = ($('.peer-component-container').length == 0) ? false : true;
+                    if (yesPeer) {
+                        $('#page_5_nav').show();
+                        $('#page_5_nav').addClass("normal-btn");
```
shailpatels, line 38: Instead of adding the "normal-btn" css class lets just include it as part of the html under `AdminGradeableBase.twig`

site/app/templates/admin/admin_gradeable/AdminGradeableBase.twig
shailpatels: I thought github deleted my comment here but i guess it posted it twice - my bad

PR #5386, opened 2020-05-15
[Feature:InstructorUI] Add ability to add links to course materials

### LINKED ISSUE #5384: URL to external resource can be a "course material"
Currently, course materials are files hosted on Submitty.  Instructors would also like to direct students to files/resources stored on external servers.  For example, posting a URL to a lecture video hosted on YouTube.

Allow the instructor to specific a "link" instead of a file that contains 2 fields: "url" and "title".  The link can be released on a specific date just like files.

elihschiff:

If a professor wants to link to an external location for example a youtube/box video lecture, there is no way to do that on the course materials. This often leads to professors making entire discussion forum posts which can clog up the forum.
Fixes #5384

### What is the new behavior?

The course materials page can now also have links to external locations
instructor view
image
student view
image
how you add a link to the course materials
image

light mode screen shots:
image
image
image

### Other information?

In order to do this the best way I though of was to store a json file that contains the url and what the instructor wants to call the link.
This means that all the functionality such as the delete button, or the ability to choose which sections can see it, etc. all still work.
Each of theses link files is a json with a url and a name field. The way the website knows if it should display that file as an uploaded json or to actually display it as a link is that in the course_materials_file_data.json file I add a `"external_link":true` line to that file.
```json
{
    "/var/local/submitty/courses/s20/sample/uploads/course_materials/homework 1": {
        "release_datetime": "2020-05-14 20:06:58",
        "hide_from_students": "off",
        "external_link": false
    },
    "/var/local/submitty/courses/s20/sample/uploads/course_materials/external-link-2020-05-14-20-06-42.json"{
        "release_datetime": "2020-05-14 20:06:52",
        "hide_from_students": "off",
        "external_link": true
    },
    "/var/local/submitty/courses/s20/sample/uploads/course_materials/sample folder/external-link-2020-05-14-20-07-46.json"{
        "release_datetime": "2020-05-14 20:07:59",
        "hide_from_students": "off",
        "external_link": true
    }
}
```

One other thing to note, if you download the course materials folder, the json files will be included. The reason for this is that a student may still want to see what websites were linked to even if they downloaded the folder. If we instead do not want them to ever see theses json files I can do that as well, it is one if statement.


elihschiff:
I added 2 radio buttons so that that you can choose if you want to upload a file or add a url. By default it will open the modal on file upload.

image
image

jchulton:
I'm getting the following error when I try to edit any course material (link or file): 
image

elihschiff:
@jchulton I am having trouble reproducing the error on my end. What browser were you using? as well as do you have any more detailed instructions on how to reproduce the bug? Are your items in folder, what exactly did you click once you clicked the edit button, anything else you can think of that might help.

jchulton:
I was using chrome. I tried but items in and not in the folder and got the error. I just clicked Upload Course Materials and didn't change any settings. I simply chose URL/file, filled in the title and URL for the exeternal link and uploaded a file for the file, then tried to click on the pencil icon to edit and got an error. Are there steps to take to get this branch working besides just install Submitty? Sorry this isn't very helpful but I don't have anything else to add

jchulton:
Sorry, I hadn't updated my master for a while. The edit bug is on master as well ):

elihschiff:
@jchulton I think I might have fixed it, if you dont mind can you take another look. I removed a function call that did not actually do anything because the function was never implemented anywhere and I am not sure what it was supposed to do. Everything seems to work fine without it.

jchulton:
Yep that fixed it. Not sure why you would change permissions when changing the date though. One interesting thing to note is that uploading an external link with the same URL and name does not override an existing external link: it makes a duplicate of it. This isn't a concern because an instructor can just delete the extra links though. This should  be set to merge after the minor comments I made above.

elihschiff:
After thinking about the possibility of having more than one course material link with the same name/url, I think that might actually be a good thing. Yes there is the fact that you have to delete the old one which is annoying. However in some situations you may want to release the same link to different people at different times. For example maybe you are linking to a lecture but your class has 2 sections (rpi opsys this summer has 2 sections at different times for example). You may want to let section 1 see the link at noon and lecture 2 see it at 2pm. If I removed the ability to have duplicate links of the same name/url that would not be possible. The solution would be giving them a different name, but I actually think the current way is better.

Ideally we should add a system where the same course material or link can be released at different times for different sections. But right now that is not a feature without making duplicates.

bmcutler:
@elihschiff please resolve conflict

bmcutler:
@elihschiff 
please resolve conflicts

bmcutler:
ach!  again???  🙄 

@elihschiff
please resolve conflicts

PR #5396, opened 2020-05-21
[Feature:MobileUI] Add Collapsible gradeables

mukul-kmr-jha:

### What is the new behavior?
On smaller screen devices gradeables are shown in the stacked form and makes it hard to search for gradeable of certain type. This PR adds collapsible feature (on the basis of type gradeable list i.e FUTURE, OPEN, CLOSED...) in the gradeables page to make navigating between gradeables easier on these devices. 
Also, gradeables that are OPEN and those for which grades are released are kept expanded by default.


site/public/css/navigation.css
```diff
@@ -62,21 +71,82 @@
...
+/*@media (min-width: 780px) {*/
+/*    .gradeable-row {*/
+/*        border: none;*/
+/*    }*/
+
+/*    .btn-nav {*/
+/*        min-width: 95%;*/
+/*    }*/
+/*}*/
+
```
elihschiff, line 149: You should just remove this extra commented out css, unless there is a good reason for it to stay here that I don't know about.

site/public/css/navigation.css
```diff
@@ -62,21 +71,82 @@
...
+    .course-main .course-title {
+        font-size: 12px;
+    }
+    .course-button .btn, .subtitle {
+        font-size: 11px;
+    }
+}
 
-    .gradeable-row {
-        border: none;
```
elihschiff, line 72: Personally I think the borders on desktop are a bit weird looking. This css is the line that used to hide them. I think they look fine on mobile if you want to leave them. My issue with them is that it makes it look like each item is its own section instead of part of another section (ex:  BETA,  OPEN, CLOSED, etc.). This might be because it stretches the full width even though the lines for each section stop with a bit of space on the left and right side.

This is what your code currently looks like 
image
I think it should look like this (which is how it used to look)
image

The mobile version I think looks fine with or without the bottom border line.

site/public/css/navigation.css
mukul-kmr-jha: It's not needed, removed.

site/public/css/navigation.css
mukul-kmr-jha: removed border from the gradeables when shown in row-form.

bmcutler:
I'll merge this now, but note that we now have inconsistent styling / coloring of the expand/collapse for the gradeables page and for the TA grading index page.

We should get feedback and decide which we prefer.

PR #5404, opened 2020-05-23
[Feature:InstructorUI] Add new term when creating a course

### LINKED ISSUE #4109: Change "Semester" to "Term"
Submitty uses the working for semester, but not all schools use semesters. Some other schedules include block or quarter. 

Submitty should pick a word to support these different schedule types instead of semester. Such as "Term".

Other words might be appropriate: "academic term".

**Additional context**
The naming for this "Term" should be open to what ever the user defines. Don't limit the user to only "A" and "U". They could use "f", "A", "Fall", or "AUTUMN".

marwanatef2:
Closes #4109

When an instructor is creating a new course it is assumed that they choose one of the existing semester without the ability to add a nicer label if they're using _quarters_ instead of _semesters_ for example.

### What is the new behavior?
An instructor can now add a new **"term"** with the required name, start and end dates which is more convenient then create a course in this term, the word _"term"_ is the more generic way of expressing _"semesters"_

The new behavior can be seen in the following gif

addterm

### Other information
I mocked the behavior of the change_prefered_name pop-up form for consistency purposes, if there is any better way let me know and i'll fix accordingly.


site/app/controllers/HomePageController.php
```diff
@@ -290,10 +290,40 @@ public function createCoursePage() {
...
+            $term_id = $_POST['term_id'];
+            $term_name = $_POST['term_name'];
+            $start_date = $_POST['start_date'];
+            $end_date = $_POST['end_date'];
+
+            $terms = $this->core->getQueries()->getAllTerms();
+            if (in_array($term_id, $terms)) {
+                $this->core->addErrorMessage("Term id already exists.");
+            }
+            else if ($end_date < $start_date) {
```
elihschiff, line 316: I think this is using normal string comparison instead of comparing date objects. String comparison might work I am not 100% sure so I thought I would bring it up.

site/app/controllers/HomePageController.php
```diff
@@ -290,10 +290,40 @@ public function createCoursePage() {
                 'showCourseCreationPage',
                 $faculty ?? null,
                 $this->core->getUser()->getId(),
-                $this->core->getQueries()->getAllUnarchivedSemester(),
+                //$this->core->getQueries()->getAllUnarchivedSemester(),
```
elihschiff, line 293: is there a reason why you are keeping this commented out? You probably just want to remove it

site/app/controllers/HomePageController.php
```diff
@@ -290,10 +290,40 @@ public function createCoursePage() {
...
+            }
+            else if ($end_date < $start_date) {
+                $this->core->addErrorMessage("End date should be after Start date.");
+            }
+            else {
+                $this->core->getQueries()->createNewTerm($term_id, $term_name, $start_date, $end_date);
+                $this->core->addSuccessMessage("Term added successfully.");
+            }
+
+            return Response::RedirectOnlyResponse(
```
elihschiff, line 324: I would move this outside of the if statement on line 306 and then add an else with an error message to that if statement as well.

site/app/templates/AddTerm.twig
```diff
@@ -0,0 +1,43 @@
...
+    <form method="post" action="{{ add_term_url }}">
+        {{ parent() }}
+    </form>
+{% endblock %}
+{% block body %}
+    <p class="std-margin">
+        A message is to be displayed here...
+    </p>
+    <p class="std-margin">
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Etiam tempor sed velit nec porttitor. Maecenas eu dolor est. Sed sit amet lacus sed elit accumsan molestie ullamcorper ut est. Curabitur molestie quam sit amet ex sodales venenatis. Proin porttitor erat vitae justo pharetra, id hendrerit felis sagittis.
```
elihschiff, line 14: Before this pr is merged, this should be switched to real text or just removed

site/app/templates/AddTerm.twig
```diff
@@ -0,0 +1,43 @@
...
+{% block popup_id %}add-new-term{% endblock %}
+{% block title %}Add new Term{% endblock %}
+{% block form %}
+    <form method="post" action="{{ add_term_url }}">
+        {{ parent() }}
+    </form>
+{% endblock %}
+{% block body %}
+    <p class="std-margin">
+        A message is to be displayed here...
```
elihschiff, line 11: Before this pr is merged, this should be switched to real text or just removed

site/app/templates/CreateCourseForm.twig
```diff
@@ -92,7 +93,18 @@
...
 
+{% include 'AddTerm.twig' %}
+
 <script>
+    function addNewTerm() {
+        $('.popup-form').css('display', 'none');
+        var form = $("#add-new-term");
+        form.css("display", "block");
+        form.find('.form-body').scrollTop(0);
+        // $('[name="term-id"]', form).val("");
```
elihschiff, line 104: is this comment still needed?

site/app/controllers/HomePageController.php
marwanatef2: i think it doesn't matter if they are string compared as long as they have the same format, they are submitted to the database as date formats as well.

site/app/controllers/HomePageController.php
marwanatef2: yeah, will remove

site/app/templates/AddTerm.twig
marwanatef2: it should be replaced with real text but i don't know what exactly should be written, any suggestions?

site/app/templates/CreateCourseForm.twig
marwanatef2: no, will remove

site/app/controllers/HomePageController.php
marwanatef2: this if statement will always evaluate to true since all inputs will be set as they're all required, so i guess there is no need for an else statement

marwanatef2:
@elihschiff can you show me when these inputs are unaligned so i can fix them, as they are always aligned with me as shown in the gif.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -256,6 +256,33 @@ public function getAllUnarchivedSemester() {
...
+    /**
+     * @param string $term_id
+     * @param string $term_name
+     * @param date $start_date
+     * @param date $end_date
+     */
+    public function createNewTerm($term_id, $term_name, $start_date, $end_date) {
+        $this->submitty_db->query(
+            "INSERT INTO terms (term_id, name, start_date, end_date) VALUES (?, ?, ?, ?)",
+            array($term_id, $term_name, $start_date, $end_date)
```
andrewaikens87, line 282: Please switch `array()` syntax to use `[]` syntax for the instantiation of arrays

elihschiff:
> @elihschiff can you show me when these inputs are unaligned so i can fix them, as they are always aligned with me as shown in the gif.

It looks like the issue only shows up for me on Firefox, not Chrome for some reason. It should be able to be fixed it with some css to make the width of the textbox slightly smaller.

site/app/libraries/database/DatabaseQueries.php
marwanatef2: okay, I'll. But can i understand the reason behind that?

andrewaikens87: As of PHP 5.4 we can use the short syntax for arrays and it is preferred.

Example (taken from https://php.net/manual/en/language.types.array.php):
```
<?php
$array = array(
    "foo" => "bar",
    "bar" => "foo",
);

// as of PHP 5.4
$array = [
    "foo" => "bar",
    "bar" => "foo",
];
?>
```

marwanatef2:
> It looks like the issue only shows up for me on Firefox

I installed Firefox and the appearance was the same as Chrome, i modified the styling a little bit anyway, hope it looks better by now.

site/app/templates/AddTerm.twig
```diff
@@ -0,0 +1,45 @@
...
+            <label class="col-md-6" for="term-id">
+                <div class="row">
+                    <span class="col-md-5">Term id:</span>
+                    <input type="text" name="term_id" placeholder="s20" id="term-id" class="col-md-6" required/>
+                </div>
+            </label>
+            <label class="col-md-6" for="start-date">
+                <div class="row">
+                    <span class="col-md-5">Start Date:</span>
+                    <input type="date" name="start_date" id="start-date" class="col-md-6" required/>
```
marwanatef2, line 22: this line fails the e2e tests for the following reasons.
> "The 'date' input type is not supported in all browsers. Please be sure to test, and consider using a polyfill."

it's more of a warning rather than an error, we can use this polyfill to fix this issue https://github.com/liorwohl/html5-simple-date-input-polyfill

MasterOdin: Or we can just ignore this warning as all of our targeted browsers support date and I refuse to have to cater to old dead browser versions.

Rewrite:

https://github.com/Submitty/Submitty/blob/44d39248f698f15d89f54b5c803705a0943f903f/tests/e2e/test_accessibility.py#L96-L101

to be:

```python
                for error in response.json()['messages']:
                    # For some reason the test fails to detect the doctype even though when you actually look at the rendered
                    # pages this error is not there. So therefore the test is set to just ignore this error.
                    skip_messages = [
                        "Start tag seen without seeding a doctype first",
                        "Possible misuse of “aria-label”",
                        "The 'date' input type is not supported in all browsers. Please be sure to test, and consider using a polyfill.",
                    ]
                    if error['message'] in skip_messages:
                        continue
```

bmcutler:
@marwanatef2 
please fix merge conflicts :)

bmcutler:
@marwanatef2 
Please follow up on failing end-to-end tests

Issue #5318, opened 2020-04-30
More Student Info:  Local Time Zone & Student Photo


bmcutler:
**What problem are you trying to solve with Submitty**

1) For online classes, it can be helpful to know the student's current local time zone (so TA can know what contact hours are convenient for office hours, or instructor can schedule synchronous exams, etc.)

2) Instructors (or sysadmins) can upload photos of every user.  Currently we do this from freshman student IDs.  These photos can be quite out of date for senior students.  And the photos are not available for all students.  It would be nice to allow students to upload their own photo.  (Making clear this should be a recognizable passport style photo of themselves, not an avatar, etc.)

**Describe the way you'd like to solve this problem**

- [x] Let's add timezone to the users table, default to server timezone.

- [x] Display and allow users to edit their timezone on the "About Me" page.

- [x] Allow instructors to view the students timezone from the Manage Students page (either in a column, or just in the modal popup?

- [x] From the users "About Me" page, show the currently uploaded student photo, and allow users upload their own photo to be used instead.  Store that photo somewhere (don't overwrite the sysadmin/instructor uploaded version).  

- [x] When displaying the "View Student Photos" page, use the student uploaded one if available, then the sysadmin/instructor uploaded version.

- [x] Allow instructor/sysadmin to indicate that the student uploaded version is not sufficient -- and the instructor/sysadmin version should be used instead. [design discussion needed]

Note: We could eventually get fancy and display due dates & other dates on the gradeable screen in the users specified time zone.

bmcutler:
tangentially related to issue #1865

bmcutler:
And also issue #5170

MasterOdin:
From slack, as we want to move towards displaying things in student's specified timezone, it should be accomplished by first using Twig to display all dates (instead of formatting them in PHP) and then we can just set `$twig->getExtension(\Twig\Extension\CoreExtension::class)->setTimezone($user->getTimezone());` and Twig will handle converting timezones for us on displaying them.

PR #5407, opened 2020-05-25
[Feature:InstructorUI] preferred timezone implementation

Kevin-MM:
Partially #5318 

- Course staff have no ability to view users' time zone information.
- Important dates for gradeables are not converted to the user's time zone.

### What is the new behavior?
- Course staff may now view users' time zone information on the `Manage Students` tab.  Additionally, the `Download CSV` button on the same page also now includes time zone information.
- Buttons on the main `Gradeables` navigation page now display important dates and times in the user's time zone (or server time zone if the user has not set a time zone)
- Dates and times on the gradeable submission page are now converted for the user's time zone (or server time zone if the user has not set a time zone)
- Dates and times appearing on student information tab of the TA grading interface will be shown in the user's time zone.
- Dates and times on the Late Days Table will now be displayed in the user's time zone
- Dates and times on the Lab Grading page will now be shown in the user's time zone (only visible when "Show Dates Graded" is checked).

Also included are a view visual upgrades requested by @bmcutler.  These changes made to help the `Manage Students` page feel less cluttered.
- The manual registration column has been removed.  If a student has been manually registered a hammer icon now appears next to the number in the `Registration Section` column
- The button to view a student's rainbow grades entry has been replaced with a clickable icon.


MasterOdin:
I continue to think we need to show the GMT offset instead of or in addition to just the timezone identifier. As is, I imagine that most instructors will have to be googling what Africa/Conakry actually means to them.

Kevin-MM:
Very well, I will add in the UTC offset today.

Kevin-MM:
I'm looking for feedback on what would be the best information to display on the `Manage Students` page in regards to to time zone information.  Please see the below ideas

**Manage Students Page**
1. Only the time zone
2. Only the UTC offset
3. Both the time zone and UTC offset (might become cluttered)
4. Only the UTC offset, but hovering the mouse over the offset would popup the user's time zone

**Download CSV**
1. I think its fine to include both time zone and UTC offset in separate columns.

bmcutler:
@Kevin-MM 
Minor suggestion:  Can we combine the registration section & manual registration columns.
For a manual registration=false student, just put the registration section # "1".  But for a manual registration=true student, put the hammer icon after the section#, so it would be "1<hammer>" in that column.  Maybe you can make a hover text on the hammer that says "manual registration" but the column label is just "registration section"

bmcutler:
@MasterOdin 
Can you please take a careful look at the database changes and the migration?
Thanks!

Kevin-MM:
I am running into a problem working on a future branch.  It appears the install sequence runs the master migrators before the course migrators, but I the course migrator was created first, and thus should run first.  I'm looking more into it now.

migration/migrator/data/course_tables.sql
```diff
@@ -497,6 +497,8 @@ CREATE TABLE users (
     instructor_updated boolean NOT NULL DEFAULT false,
     manual_registration boolean DEFAULT false,
     last_updated timestamp(6) with time zone,
+    time_zone VARCHAR NOT NULL DEFAULT 'NOT_SET/NOT_SET',
+    display_image_state VARCHAR NOT NULL DEFAULT 'system',
```
MasterOdin, line 501: Why is this column being added here?

site/public/js/server.js
```diff
@@ -575,7 +575,7 @@ function downloadCSV(code) {
     var temp_element = $('<a id="downloadlink"></a>');
     var address = "data:text/csv;charset=utf-8," + encodeURIComponent(csv_data);
     temp_element.attr('href', address);
-    temp_element.attr('download', 'submitty_user_emails.csv');
+    temp_element.attr('download', 'course_users_data.csv');
```
MasterOdin, line 578: can we make the name here include semester / course details?

site/app/models/User.php
```diff
@@ -215,6 +215,24 @@ public function updateTimeZone(string $time_zone): bool {
...
+        return $this->time_zone === 'NOT_SET/NOT_SET' ? 'NOT SET' : $this->time_zone;
+    }
+
+    /**
+     * Get the UTC offset for this user's time zone.
+     * @return string The offset in hours and minutes, for example '+9:30' or '-4:00'
+     */
+    public function getUTCOffset() {
+        return DateUtils::getUTCOffset($this->time_zone);
+    }
```
MasterOdin, line 234: add tests for these functions

site/app/models/Button.php
```diff
@@ -38,6 +38,9 @@ class Button extends AbstractModel {
     protected $title;
     /** @prop @var string|null $subtitle */
     protected $subtitle;
+    /** @prop @var DateTime|null A DateTime object representing the time for this Button.  For example if this was a
+    * grade button then $date may represent the date and time grading begins or ends */
+    protected $date;
```
MasterOdin, line 43: add `@method getDate(): DateTime|null` above. Also add test for function.

migration/migrator/data/course_tables.sql
Kevin-MM: I was trying to sneak it in because its part of the same issue, but you're right it has nothing to do with the PR so I removed all references to it.

site/public/js/server.js
Kevin-MM: Yes, added this functionality.

site/app/models/User.php
Kevin-MM: Tests added.

site/app/models/Button.php
Kevin-MM: Done.

site/app/models/User.php
```diff
@@ -2,6 +2,7 @@
```
MasterOdin, line 2: Set:
https://github.com/Submitty/Submitty/blob/5cf9cdc1a7f31db964a250ec737af4db3c886e92/site/app/models/User.php#L194-L196

to be $this->time_zone = $details['time_zone'] ?? 'NOT_SET/NOT_SET';

site/app/models/User.php
```diff
@@ -215,6 +218,35 @@ public function updateTimeZone(string $time_zone): bool {
         return false;
     }
 
+    /**
+     * Get the user's time zone, in 'nice' format.  This simply returns a cleaner 'NOT SET' string when the
+     * user has not set their time zone.
+     *
+     * @throws NullException Found null user time zone, but this should never happen
```
MasterOdin, line 225: Why do we need this null check if it should never happen? What are the cases that are causing this to be null, and can we better protect against those upstream, by making the change above, or just throwing an exception where `$this->time_zone` is set to null?

site/app/views/NavigationView.php
```diff
@@ -215,7 +214,8 @@ public function showGradeables($sections_to_list, $graded_gradeables, array $sub
             "display_room_seating" => $display_room_seating,
             "seating_only_for_instructor" => $this->core->getConfig()->isSeatingOnlyForInstructor(),
             "gradeable_title" => $gradeable_title,
-            "seating_config" => $seating_config
+            "seating_config" => $seating_config,
+            "DATE_TIME_FORMAT" => DateUtils::DATE_TIME_FORMAT
```
MasterOdin, line 218: Can we use https://twig.symfony.com/doc/3.x/functions/constant.html instead?

site/app/models/User.php
Kevin-MM: You make a good point here.  I got rid of the null exception, and changed updateTimeZone to be setTimeZone, so now the only ways to set it are in the constructor and with the setter.  It should never end up null unless the user isn't loaded.

site/app/models/User.php
Kevin-MM: Changed.

site/app/views/NavigationView.php
Kevin-MM: I changed all these in my next PR so don't worry about this, it will be changed.

PR #5409, opened 2020-05-25
[UI:Mobile] Add mobile screen view for the grade-report page 

mukul-kmr-jha:

Currenlty the page layout breaks on smaller screen devices...
Screenshot 2020-05-25 at 7 51 32 PM

### What is the new behavior?
Added responsiveness and now the page looks like this
image

### Other information?
Involves a little code refactoring also as I scrapped out the table layout plus inline styling in the report page and created a different CSS file for styling.


mukul-kmr-jha:
> A bigger issue is that we should be writing mobile-first CSS this means if we're adding new CSS the display page should be first programmed to render on mobile screens and media queries should be used after to target desktop and tablet views.

Done for this PR.
I'm planning to use this suggestion for future PRs also as I move on to the next pages ( which will result in refactoring) because currently, we have targeted the desktop screen instead of using mobile-screen first strategy.
Thanks!

PR #5421, opened 2020-05-29
[VPAT] Stop tabbing from escaping modals

### LINKED ISSUE #3830: Tabbing on modal (popup) windows can cycles to elements on underlying page

elihschiff:

On many modals when you use tab to move among elements, you can escape the modal and tab on elements behind the modal on the main page

### What is the new behavior?

fixes #3830
when you press tab from the last element in a modal instead of cycling to the page behind the modal, it will move the focus back to the first item

I also made it so that the captureTabInModal() function runs automatically as long as a modal is generated with Popup.twig (which as far as I can tell they all are)
This will mean going forward no extra work should need to be done when making a new modal.

There is still one situation captureTabInModal() will need to be run manually. If a modal disables the submit button for any reason, the function must be run every time the state of the submit button is changed. (so both when it is disabled as well as when it is no longer disabled)

### Other information?

I went around the website and tried to find and test every modal but there is a chance I missed one.

### DIFF
--- a/site/public/js/server.js
+++ b/site/public/js/server.js
@@ -194,6 +194,7 @@ function newDeleteGradeableForm(form_action, gradeable_name) {
     $('[id="delete-gradeable-message"]', form).append('<b>'+gradeable_name+'</b>');
     $('[name="delete-confirmation"]', form).attr('action', form_action);
     form.css("display", "block");
+    captureTabInModal("delete-gradeable-form");
 }
 
 function displayCloseSubmissionsWarning(form_action,gradeable_name) {
@@ -203,6 +204,7 @@ function displayCloseSubmissionsWarning(form_action,gradeable_name) {
     $('[id="close-submissions-message"]', form).append('<b>'+gradeable_name+'</b>');
     $('[name="close-submissions-confirmation"]', form).attr('action', form_action);
     form.css("display", "block");
+    captureTabInModal("close-submissions-form");
     form.find('.form-body').scrollTop(0);
 }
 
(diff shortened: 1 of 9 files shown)


site/public/js/server.js
```diff
@@ -304,17 +299,20 @@ function newEditCourseMaterialsForm(path, this_file_section, this_hide_from_stud
...
+    var form = $("#".concat(formName));
 
     /*get all the elements to tab through*/
     var inputs = form.find(':focusable').filter(':visible');
     var firstInput = inputs.first();
     var lastInput = inputs.last();
 
     /*set focus on first element*/
-    firstInput.focus();
+    if(resetFocus){
```
shailpatels, line 313: When would you not want the focus to land on the first element in the modal?

elihschiff: In some modals, the submit button is disabled at the start. This means you cannot tab to it. However when you enable it, you must rerun this function so that tabbing to the submit button works. However in that situation you want to keep the focus on the current item because that is already in the modal. The goal is just to make the submit button tabable again. An example of this is the new student button on the manage students page.

elihschiff:
@shailpatels I reverted the change that makes it run automatically because I did not think about what you said. As for the issue of where your tab lands after you exit the modal. Instead of hard coding the correct element and passing it into captureTabInModal, I think it would be better to have the function auto figure out what element you were on using `document.activeElement` and then put you back there when you exit the modal. I am pretty sure this should be doable.

PR #5424, opened 2020-05-30
[Feature:Submission] Increase sequence diagram message length

emaicus:
After recent updates to `mermaid.js`, actors in sequence diagrams now automatically space themselves and messages no longer overlap. This PR increases the number of characters displayed of a message before truncation occurs from 72 to 240.

72 characters displayed (old display)
old-sequence-diagram

240 characters (new display)
updated-sequence-diagram


emaicus:
Note the differences in aspect ratio in the images provided. The spacing between actors is actually equivalent.

PR #5426, opened 2020-05-30
[Feature: Forum] Attach image from clipboard to forum post

### LINKED ISSUE #5379: Attach image from clipboard to forum post
Currently image attachments on the forum must be a file on your computer.
It would be nice to be able to paste an image from the clipboard -- without first explicitly saving it to a file.
From windows you can do: Win-Shift-S to quickly take a screen shot of an area.

marwanatef2:
Closes  #5379 
When saving image/screenshot to clipboard (like that taken with PrtScr button) and pasting it to the post _textarea_ nothing happens.

### What is the new behavior?
When an image/screenshot is pasted from the clipboard to the post _textarea_ it is added along with the files attached from input.

### New behavior gif
pastaimgclipboard


marwanatef2:
> the image name appears to always be "image1.png" if I try and paste a second image even if its different it will also be named image1 and the only option is to overwrite the original image, I think changing the naming scheme to image1.png, image2.png

I thought of this solution but the problem it brings is that if the user pasted the same image multiple times it will be added with different names, so i thought we can limit pasting to one image for now.

shailpatels:
I think that's fine @marwanatef2 , there's already a limit to attachments so we don't have to worry about spamming, adding multiple images by pasting is a valid use case so we should support it.

marwanatef2:
alright, will work on it.

PR #5433, opened 2020-06-01
[Bugfix:Router] Handle POST requests larger than post_max_size

shailpatels:

Currently, a POST request larger than post_max_size is not handled by the router, this causes issues since data is normally lost when this occurs by PHP causing the superglobal variable to be empty or malformed.

### What is the new behavior?
The router will now check for POST requests being too large and send a JSON fail response. This was done since a large POST request typically occurs during file upload controls or custom made POST requests through the API.

This will make sure the controllers no longer need to check for this and would often handle it incorrectly leading to confusing or wrong error messages or nothing happening at all.

### Other information?

Default is 10MB, can test by going to any file submission site and uploading a large file


site/public/js/drag-and-drop.js
```diff
@@ -1145,12 +1131,11 @@ function handleUploadCourseMaterials(csrf_token, expand_zip, hide_from_students,
...
 
-                if (jsondata['status'] === 'success') {
-                    window.location.href = return_url;
-                }
-                else {
+                if (jsondata['status'] !== 'success') {
                     alert(jsondata['message']);
                 }
+                
+                window.location.href = return_url;
```
MasterOdin, line 1138: why were these all changed to always redirect on failure? There's no mention in the PR as to why you're making changes to this file.

shailpatels: Previously an error would be shown through and alert and the server would also display an error by $this->core>addError(); function but that would only display on page change.

I changed it to redirect which most cases would just be a page refresh to the same page so show the error once from the server, this would also prevent showing duplicate errors when the user moves changes to another page.

MasterOdin: We should try to not refresh / redirect under normal usage (like hitting max POST size). Stuff like CSRF token violation or uploading a file for someone that is not you is fine to refresh on since those are abnormal errors that should not really happen under normal usage, and if they do, at least in the case of the first error, a refresh is the only way to fix it.

site/public/index.php
```diff
@@ -116,6 +118,26 @@ function error_handler() {
...
+            JsonResponse::getFailResponse($msg)
+        );
+
+        if ($response instanceof ResponseInterface) {
+            $response->render($core);
+        }
+
+        return $core->getOutput()->displayOutput();
+    }
+}
```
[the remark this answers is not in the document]
shailpatels: would you say this should be moved to `webRouter.php` and be called from `getWebResponse` and `getApiResponse`?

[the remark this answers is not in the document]
MasterOdin: yes

Issue #2501, opened 2018-07-19
Close popup form when you click outside of them


bmcutler:
per review from #2447

CouleeApps:
The main reason I did not add this is because some forms don't have a "cancel" option (eg the teammate extension form is only "Yes" or "No")... I can add it for the ones that have cancel though.

PR #5435, opened 2020-06-01
[VPAT] Removed device dependent event handlers from popups

elihschiff:

Every popup made using the popup.twig file would cause 2 WAVE `Device dependent event handler` errors. 

### What is the new behavior?

I do not believe either of theses onclick attributes even has a use anything anymore, they are just there from old code and currently do nothing. So I just removed them. For whoever tests this, if you are able to find a change in page functionality by removing them please let me know but I dont think there is any.

### Other information?

image

### DIFF
--- a/site/app/templates/generic/Popup.twig
+++ b/site/app/templates/generic/Popup.twig
@@ -2,9 +2,9 @@
 <div class="popup-form" id="{% block popup_id %}#popup-form{% endblock %}" style="display: none;">
     {% block form %}
         {# This lets us center the window #}
-        <div class="popup-box"  onclick="event.stopPropagation(); closePopup();">
+        <div class="popup-box">
             {# Actual displayed window #}
-            <div class="popup-window" onclick="event.stopPropagation();">
+            <div class="popup-window">
                 {% block title_panel %}
                     <div class="form-title">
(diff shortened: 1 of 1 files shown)


MasterOdin:
Do all of the points in https://github.com/Submitty/Submitty/pull/2536 still remain fixed / working following this, as that's where the `closePopup()` function was added.

elihschiff:
After looking into it a bit more, sometime after the pr you linked and today, #4139 made it so that you can no longer click on the darkened background outside the modal to close the modal. I am not sure if this was intentional or not. As of right now, this pr is consistent with what currently happens on Submitty (clicking on the background does nothing). I could modify this pr to add that feature back in though if you think that is something we want.

MasterOdin:
I'll let @bmcutler answer as she originally opened the linked issue #2501 requesting that to be done.

bmcutler:
I opened that issue, but it came out of a group discussion.  I think it was appropriate when the popup/modal was visually small...  and more of a reminder/warning/annoyance, rather than a control/action/input that the user requested.  (Most of?) Our modals now seem more complicated, and the user *should* interact with them actively, not have a shortcut to close them.  Let's leave it as is, that clicking the background does nothing.

PR #5442, opened 2020-06-02
[Bugfix:TAGrading] peer grading total on index page

josharsh:
Score under TA Grading column in Grading index for is wrong, for both instructor and student views

### What is the new behavior?
In Instructor View, instructor can now see the total manual score obtained and in the student view the student can see the peer score only.


emaicus:
I misunderstood the cause of the problem before. Perform the following steps to reproduce:
1. Log in as instructor
2. Create a new gradeable
3. Modify the default TA component to be worth 10 points
4. Add a peer component worth 10 points
5. Perform peer assignment. student `A` grades `B`
6. Make a submission for student `B`
7. Open the assignment to grading
8. Log in as student `A`
9. Grade student `B`
10. Navigate to the grading index to see `10/20` points instead of `10/10`.

bugged_peer_points

PR #5453, opened 2020-06-03
[Refactor:System] eslint config for linting JS

MasterOdin:

* [x] Documentation has been updated/added if relevant: Submitty/submitty.github.io#278

We do not currently lint JS files.

### What is the new behavior?

Adds a [eslint](https://eslint.org/) configuration to be used for linting the JavaScript files of Submitty. It uses the recommended rules plus a handful of ones I like. See their [rules page](https://eslint.org/docs/rules/) for the full list of rules.

Enabled rules:
- [arrow-spacing](https://eslint.org/docs/rules/arrow-spacing)
- [block-spacing](https://eslint.org/docs/rules/block-spacing)
- [for-direction](https://eslint.org/docs/rules/for-direction)
- [getter-return](https://eslint.org/docs/rules/getter-return)
- [no-async-promise-executor](https://eslint.org/docs/rules/no-async-promise-executor)
- [no-compare-neg-zero](https://eslint.org/docs/rules/no-compare-neg-zero)
- [no-cond-assign](https://eslint.org/docs/rules/no-cond-assign)
- [no-constant-condition](https://eslint.org/docs/rules/no-constant-condition)
- [no-control-regex](https://eslint.org/docs/rules/no-control-regex)
- [no-debugger](https://eslint.org/docs/rules/no-debugger)
- [no-dupe-args](https://eslint.org/docs/rules/no-dupe-args)
- [no-dupe-else-if](https://eslint.org/docs/rules/no-dupe-else-if)
- [no-dupe-keys](https://eslint.org/docs/rules/no-dupe-keys)
- [no-duplicate-case](https://eslint.org/docs/rules/no-duplicate-case)
- [no-empty](https://eslint.org/docs/rules/no-empty)
- [no-empty-character-class](https://eslint.org/docs/rules/no-empty-character-class)
- [no-ex-assign](https://eslint.org/docs/rules/no-ex-assign)
- [no-extra-boolean-cast](https://eslint.org/docs/rules/no-extra-boolean-cast)
- [no-extra-semi](https://eslint.org/docs/rules/no-extra-semi)
- [no-func-assign](https://eslint.org/docs/rules/no-func-assign)
- [no-import-assign](https://eslint.org/docs/rules/no-import-assign)
- [no-inner-declarations](https://eslint.org/docs/rules/no-inner-declarations)
- [no-invalid-regexp](https://eslint.org/docs/rules/no-invalid-regexp)
- [no-irregular-whitespace](https://eslint.org/docs/rules/no-irregular-whitespace)
- [no-misleading-character-class](https://eslint.org/docs/rules/no-misleading-character-class)
- [no-obj-calls](https://eslint.org/docs/rules/no-obj-calls)
- [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins)
- [no-regex-spaces](https://eslint.org/docs/rules/no-regex-spaces)
- [no-setter-return](https://eslint.org/docs/rules/no-setter-return)
- [no-sparse-arrays](https://eslint.org/docs/rules/no-sparse-arrays)
- [no-unexpected-multiline](https://eslint.org/docs/rules/no-unexpected-multiline)
- [no-unreachable](https://eslint.org/docs/rules/no-unreachable)
- [no-unsafe-finally](https://eslint.org/docs/rules/no-unsafe-finally)
- [no-unsafe-negation](https://eslint.org/docs/rules/no-unsafe-negation)
- [use-isnan](https://eslint.org/docs/rules/use-isnan)
- [valid-typeof](https://eslint.org/docs/rules/valid-typeof)
- [no-case-declarations](https://eslint.org/docs/rules/no-case-declarations)
- [no-empty-pattern](https://eslint.org/docs/rules/no-empty-pattern)
- [no-fallthrough](https://eslint.org/docs/rules/no-fallthrough)
- [no-global-assign](https://eslint.org/docs/rules/no-global-assign)
- [no-octal](https://eslint.org/docs/rules/no-octal)
- [no-redeclare](https://eslint.org/docs/rules/no-redeclare)
- [no-self-assign](https://eslint.org/docs/rules/no-self-assign)
- [no-unused-labels](https://eslint.org/docs/rules/no-unused-labels)
- [no-useless-catch](https://eslint.org/docs/rules/no-useless-catch)
- [no-useless-escape](https://eslint.org/docs/rules/no-useless-escape)
- [no-with](https://eslint.org/docs/rules/no-with)
- [no-delete-var](https://eslint.org/docs/rules/no-delete-var)
- [no-shadow-restricted-names](https://eslint.org/docs/rules/no-shadow-restricted-names)
- [no-undef](https://eslint.org/docs/rules/no-undef)
- [no-unused-vars](https://eslint.org/docs/rules/no-unused-vars)
- [no-mixed-spaces-and-tabs](https://eslint.org/docs/rules/no-mixed-spaces-and-tabs)
- [constructor-super](https://eslint.org/docs/rules/constructor-super)
- [no-class-assign](https://eslint.org/docs/rules/no-class-assign)
- [no-const-assign](https://eslint.org/docs/rules/no-const-assign)
- [no-dupe-class-members](https://eslint.org/docs/rules/no-dupe-class-members)
- [no-new-symbol](https://eslint.org/docs/rules/no-new-symbol)
- [no-this-before-super](https://eslint.org/docs/rules/no-this-before-super)
- [require-yield](https://eslint.org/docs/rules/require-yield)
- [brace-style](https://eslint.org/docs/rules/brace-style) (stroustrup style, like PHP)
- [comma-dangle](https://eslint.org/docs/rules/comma-dangle)
- [default-param-last](https://eslint.org/docs/rules/default-param-last)
- [eol-last](https://eslint.org/docs/rules/eol-last)
- [indent](https://eslint.org/docs/rules/indent) (2 spaces for indentation)
- [keyword-spacing](https://eslint.org/docs/rules/keyword-spacing)
- [linebreak-style](https://eslint.org/docs/rules/linebreak-style) (use `\n` to end lines)
- [no-trailing-spaces](https://eslint.org/docs/rules/no-trailing-spaces)
- [no-var](https://eslint.org/docs/rules/no-var)
- [prefer-arrow-callback](https://eslint.org/docs/rules/prefer-arrow-callback)
- [prefer-const](https://eslint.org/docs/rules/prefer-const)
- [prefer-template](https://eslint.org/docs/rules/prefer-template)
- [quotes](https://eslint.org/docs/rules/quotes) (prefer single quotes)
- [semi](https://eslint.org/docs/rules/semi) (always have semi-colons)
- [semi-style](https://eslint.org/docs/rules/semi-style) (semi-colons are at end of lines)
- [template-curly-spacing](https://eslint.org/docs/rules/template-curly-spacing) (enforces no spaces between curly braces and variable in template literals)

As a test run of things, I've updated the websocket.js file (and it is not ignored).

### Other Information

Similar to Python and flake, I've added an ignore file that includes all existing JS files. I will over an extended period be going through each file and cleaning up errors one-by-one, hopefully avoiding any files that someone is currently working in and has an open PR against it.


site/.eslintrc.json
```diff
@@ -0,0 +1,33 @@
+{
+    "env": {
+        "browser": true,
+        "es6": true,
```
shailpatels, line 4: Should this be es5? Or would es5 and 6 have the same linting rules.

shailpatels:
This is probably for the best, here are my thoughts on it : 
* How will this work with Twig templates adding populating variables serverside in twig files like ` let x = "{{ filename }}";`
* How will this work with a plugin like jquery?
* I may not be reading the docs right but was is the executor keyword do in the context `function executor(resolve, reject) {` for the no-async-promise-executor rule, is there a way we can enforce the (resolve, reject) => standard when creating promises instead?
* Why is no-inner-declarations considered bad?
I don't see how a nested function declaration is better than a nested function expression
if the linter doesn't like this
```js
function a(){
    function b(){}
}
```
why is this allowed?
```js
function a(){
    let b = function(){};
    b();
}
```
* no-irregular-whitespace rule doesn't appear to have a comment saying it can be autofixed by the `--fix` CMD option, is it possible to get os-specific issues that can't be fixed locally from this?
* Similar twig issue with the no-undef rule, if some of a JS file is being constructed by twig and that same code/declarations is being expected somewhere else how can we fix that
* default-param-last how would this handle multiple default params 
* no-var seems kind of extreme rule to implement, can every instance of `var` be replaced with `let` and `const` automaticlly or is this something we would have to spend time refactoring ?
* I don't understand what this rule does or enforces: prefer-arrow-callback 
* personally I don't think it matters if we use double and single quotes in JS (since we do this in php right?) but if it can be auto fixed away I don't have strong feelings on it
* prefer-template rule, would this interfere with twig templates ? 

I'm guessing the twig issues could be solved by only targetting JS but at the same time we have a lot of inline js in <script> tags in twig files.

site/package.json
```diff
@@ -26,5 +26,10 @@
     "shortcut-buttons-flatpickr": "0.3.0",
     "twig": "1.15.1"
   },
-  "devDependencies": {}
+  "devDependencies": {
+    "eslint": "^7.1.0"
+  },
+  "scripts": {
+    "lint": "eslint public/js/*.js public/js/pdf/*.js"
```
shailpatels, line 33: Can we also have a script for auto fixing using eslint
and is there a way to pass in a specific file as an argument instead of hitting every file?
something like:
```
eslint --fix --fix-type problem,layout <file>
```

MasterOdin: I was going to add a `"lint:fix": "eslint public/js/*.js public/js/pdf/*.js"` run-script. You can also activate eslint directly by doing `node_modules/.bin/eslint`. Yes, I will add documentation once we're agreed on the rules.

MasterOdin:
> How will this work with Twig templates adding populating variables serverside in twig files like let x = "{{ filename }}";

Twig templates are probably not going to be linted and we should be working to not inline as much JS as we can. There is https://github.com/BenoitZugmeyer/eslint-plugin-html which in-theory supports twig, but not if you're mingling twig templating within the `<script>` tags like we are.

> How will this work with a plugin like jquery?

eslint supports detecting that we're using jQuery (and so understands that `$` is referencing jQuery), but we do not have any rules as to what in jQuery you can or cannot do.

> I may not be reading the docs right but was is the executor keyword do in the context function executor(resolve, reject) { for the no-async-promise-executor rule, is there a way we can enforce the (resolve, reject) => standard when creating promises instead?

`executor` was just the name of the function they were using. It does not enforce that you have to call the function executor (or anything), just that you cannot have it be marked as `async` (for the reasons listed). The `prefer-arrow-callback` rule enforces that we get `(resolve, reject) => ` over `function (resolve, reject) {`.

> Why is no-inner-declarations considered bad?

It's mainly to avoid potential inconsistencies with different browsers and how they "hoist" things. For example, the statement:
```
if (false) {
  function foo() {}
}
```
has the function foo defined in some browsers as functions are hoisted to the top of their function block, which is not intuitive (safari did this I think?). As such, we want to avoid this since it can lead to incorrect / surprising behavior, and there's almost always a better way to accomplish what you want. Both of your examples though are accepted by this rule.

> no-irregular-whitespace rule doesn't appear to have a comment saying it can be autofixed by the --fix CMD option, is it possible to get os-specific issues that can't be fixed locally from this?

We have no violations of this rule so it's a bit of a moot point. No idea why it's not "auto-fixable", but it's not really something that should come up often.

> Similar twig issue with the no-undef rule, if some of a JS file is being constructed by twig and that same code/declarations is being expected somewhere else how can we fix that

Specify variables that come from outside the script in a `/* global <var_name> */` comment at the top of the file. This improves our documentation of where variables are coming from, though we should strive to not have globals if possible.

> default-param-last how would this handle multiple default params

It works with any number of parameters with defaults, so long as they come after any without.

> no-var seems kind of extreme rule to implement, can every instance of var be replaced with let and const automatically or is this something we would have to spend time refactoring ?

It is something we would have to spend time to do. This rule may not be initially activated, but it is one that should be worked for and activated as it's JS best practices to use let/const over var in all cases (as it avoids hoisting and other weird scope stuff.

> I don't understand what this rule does or enforces: prefer-arrow-callback

Use `(bar) => {}` over `function (bar) {}`. I fixed the name and reference to the eslint rule as there was a typo.

> personally I don't think it matters if we use double and single quotes in JS (since we do this in php right?) but if it can be auto fixed away I don't have strong feelings on it

This is a more personal style rule than anything. I'm fine removing it, but it's just what I roll with within my own code which is where I'm getting this eslint from.

> prefer-template rule, would this interfere with twig templates ?

See above, we would not be parsing twig templates (as parsing them is kind of tricky). We should aim to not inline as much JS as we can so that we can let eslint do its work.

> personally I prefer 4 spaces for indents on JS can we get other peoples input on this?

This is a personal preference and am willing to go with 4 if that's what the majority wants. For some consideration, it seems like 2 spaces is somewhat the standard in the wider JS community from my experience.

site/package.json
MasterOdin: added.

MasterOdin:
> > personally I prefer 4 spaces for indents on JS can we get other peoples input on this?
>
> This is a personal preference and am willing to go with 4 if that's what the majority wants. For some consideration, it seems like 2 spaces is somewhat the standard in the wider JS community from my experience.

More of the codebase uses 4 spaces for indent, including a number of recently added files. Therefore, while I have my personal preference, I will go with the majority files here and to reduce adoption friction, have gone with four spaces.

PR #5455, opened 2020-06-03
[Feature:TAGrading]Peer graders anonymization of files

jchulton:

A peer grader can inspect element and see the student id of the student they are grading because the id is used in our file structure

### What is the new behavior?
While our file structure hasn't changed, inspect element will show anonymized paths, in which the userid is replaced with the anon_id


jchulton:
Sorry about the error, that should be fixed now

jchulton:
The student's username should no longer be exposed in getNonAnonPath

emaicus:
I'm very confused. I just re-reviewed this, and it looks like many of the places where the student's id was previously anonymized are now exposed. I am grading `aufded` as `aphacker`. Also of note, the discussion panel is not anonymized.

jchulton:
The fix I pusehd should remove most places where the id was anonymized. I am still unsure of how to remove it from the call to getNonAnonPath. I think it would have to be encoded/decoded, is there a preferred way to do this?

PR #5469, opened 2020-06-05
[Bugfix:Travis] use pyenv to set python to 3.6

MasterOdin:

Travis uses Python 3.5 by default if no version of python is specified. This causes issues with Submitty as we potentially move to using more python 3.6 features (like f-strings), such as in #5465. 

### What is the new behavior?

Explicitly set that we are using python 3.6 in tests were we do not specify the python version, but utilize python scripts (e2e, integration tests). Adds a few minutes to these cases, but unfortunately it's not really avoidable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -196,6 +196,9 @@ jobs:
       before_install:
         - source ${TRAVIS_BUILD_DIR}/.setup/travis/before_install.sh
         - phpenv config-rm xdebug.ini
+        - pyenv install 3.6.3
+        - pyenv global 3.6.3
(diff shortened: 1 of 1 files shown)


MasterOdin:
Misunderstood what was happening. It's not an issue with pyenv (that's actually setting it to use python 3.7), but rather that libapache-mod-wsgi-py3 (which we use to trigger authentication.py) uses the python3 out of apt-get which for xenial is python 3.5. We need to move to using bionic for the Travis builds to resolve this particular problem.

PR #5509, opened 2020-06-10
[VPAT:Forum] Improved tab navigation on the forum

elihschiff:

It is very hard to get around the forum using only the tab key

### What is the new behavior?

I added a bunch of skip buttons to the forum that show up as you tab. They can be used to jump around the forum more quickly
If this gif is a bit hard to see, I would open it in another tab so it is larger
image


listeph:
The skips are good! One thing I noticed though is when the user does want to scroll through the threads, it jumps back to the top. Here is how it looked like to scroll through before your changes:

normal_scroll

Here is after:

buggy_scroll

site/app/templates/forum/GeneratePostList.twig
```diff
@@ -57,7 +58,7 @@
         <input type="hidden" name="display_option" value="{{ display_option }}" />
         {% include "forum/ThreadPostForm.twig" with {
             "show_post" : true,
-            "post_content_placeholder" : "Enter your reply to all here...\nPress enter for newline or ctrl+enter to post",
+            "post_content_placeholder" : "Enter your reply to all here...\nPress enter for newline or ctrl+enter to post\nor esc to exit textbox",
```
shailpatels, line 61: Placeholders are visual cues only and will be ignored by most screen readers. 
Some sort of label should be used here to convey this information, especially because the escape to move to the next element is nonstandard navigation.

I think this should say "Press escape to exit textbox" if its going to be on its own line, the "or" doesn't make grammatical sense to me

site/app/templates/forum/GeneratePostList.twig
```diff
@@ -1,6 +1,7 @@
 {% set show_unresolve = false %}
 <data id="current-thread" value="{{ post_data[0].thread_id }}"></data>
-
+<a class="skip-btn" href="#thread_box_link_{{thread_id != "" ? thread_id : first_post_id}}">Back to thread list</a>
+<a class="skip-btn key_to_click" href="#" onclick="$('#reply_box_{{post_box_id}}').focus(); return false;">Skip to reply</a>
```
shailpatels, line 4: You should not use `href="#"` in this context, having an empty element indicates to the browser to jump to the top of the page.

If an anchor tag is not being used to point to a valid URL or element it should not use the href tag, however this will mean you will have to set tabindex manually and potentially set some of the styling

I recommend passing in the id of the "post to thread" box (the one at the very bottom) through twig and setting the href to point at that element like the rest of the navigation skip links.

site/public/css/server.css
```diff
@@ -1749,6 +1748,10 @@ end of styles used in the admin gradeable page
     z-index: 999;
 }
 
+#skip-nav {
+    margin: 90px 10px 10px; /* 90px for header */
```
shailpatels, line 1752: This comment isn't very descriptive, something like "make skip to main content box" larger
would be more helpful if you're going to add a comment here.

elihschiff:
I have resolved the comment shail left as well as fixed the scrolling issue that steph pointed out. It now keeps the current post always in the center of the screen which makes it very easy to follow when tabbing through the page, as well as it makes sure that nothing is hidden when tabbing.
Peek 2020-06-22 03-59

shailpatels:
@elihschiff Did you add a label for the forum textarea input, currently, a screenreader will not read the information "press esc" to exit instead of normal tabbing out.

> Placeholders are visual cues only and will be ignored by most screen readers.
Some sort of label should be used here to convey this information, especially because the escape to move to the next element is nonstandard navigation.

elihschiff:
@shailpatels sorry I misread what you said originally. I added an aria-label with the placeholder information so that somebody with a screen reader can hear that info. I dont think anyone not using a screen reader needs to see that info twice so I think an aria-label makes more sense than a regular label and they both work for screen readers

@listeph I added a way to get back to the thread list after you are at the reply box
image

PR #5527, opened 2020-06-14
[Feature:Autograding] Short-circuit non-autograde submissions

### LINKED ISSUE #3683: Short circuit shipper-worker for assignments with no autograding
If there are no test cases...  shipper can prepare the necessary results.json, etc. files and not pack & send to worker.

hdrodz97:

Closes #3683.

There are some gradeables that don't require autograding (e.g. uploading a PDF of an essay). Despite this, they still have to go through the whole process of shipping from the main Submitty machine to a worker machine (or process), which then needs to pick up the autograding job, run the autograder, and ship the results back, even though the only thing being done is a submission count check which could be done on the main shipper.

### What is the new behavior?

This PR introduces the idea of "trivially gradeable" autograding jobs. Currently, a job is trivially gradeable if its only autograding test case is a submission limit check.

If a job is trivially gradeable, then the "grading" process is executed on the shipper process instead of on a (potentially-remote) worker.

### Other information?

Singular submissions have been tested on the "Upload Only" gradeable in the "Development" course.

Bulk upload was also tested, and that seems to be working as well.


sbin/submitty_autograding_shipper.py
```diff
@@ -789,6 +805,317 @@ def shipper_process(my_name,my_data,full_address,which_untrusted):
...
+    ###########################################################################
+    #
+    # NOTE: Editing this file? Make sure that this function stays in-sync with
+    #       grading/main_validator.cpp::validateTestCase()
+    #
+    ###########################################################################
+
+    results = {'testcases': testcase_outputs}
+    with open(results_json, 'w') as fd:
+        json.dump(results, fd)
```
emaicus, line 1072: We usually write json files with `indent=4` for ease of debugging (probably not best practice, though).

sbin/submitty_autograding_shipper.py
```diff
@@ -789,6 +805,317 @@ def shipper_process(my_name,my_data,full_address,which_untrusted):
...
+
+    excessive = max(subnum - max_submissions, 0)
+    points = floor(excessive * penalty)
+    points = max(points, possible_points)
+    view_testcase = points != 0
+    
+    return {
+        'test_name': f"Test 1 {testcase['title']}",
+        'view_testcase': view_testcase,
+        'points_awarded': points
```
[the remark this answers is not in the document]
hdrodz97: It's a deduction. Code is mostly parallel to https://github.com/Submitty/Submitty/blob/master/grading/main_validator.cpp#L323-L334.

On the earlier note on how excessive submissions don't actually result in a penalty, I found this strange too. The original code as-written seems to use the `possible_points` field as a cap for how many points you can lose as a result of excessive submissions. As the "Upload Only" gradeable has `possible_points` set to 0, this means that there is no penalty on that gradeable for making too many submissions. Whether this was the original intended behavior, I don't know.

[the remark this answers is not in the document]
bmcutler: Initial design for autograding was that each test was non negative and integer points.

Longstanding open issue to allow negative points (#1324), and probably also allow floating point (with controllable precision).  But I don't see an issue for that yet.

So it's not relevant to your PR now.  But when that other issue gets resolved it must be tested.  I've made a note in that issue.

sbin/submitty_autograding_shipper.py
bmcutler: @hdrodz97 are you making this quick change?

hdrodz97: Just put it in

PR #5529, opened 2020-06-15
[Feature: Forum] Use websockets in Discussion Forum

marwanatef2:
### Please check if the PR fulfills these requirements:

* [ ] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
Closes #4328 
The discussion forum fetches new data from server only upon refresh while changes happen without the user noticing.
Closes #5517 

### What is the new behavior?
Using web sockets in the discussion forum will introduce real-time updates without having the users to refresh the page every now and then to get the latest updates, this also reduces the load on the server as changes occur at the client window (browser).

### Other information?
This PR is based on the sockets code introduced in PR #5205 , you have to go by this PR in order to enable websockets in the forum.
This PR also grabs some functionality in PR #4329 which contains initial code for using websockets in the forum.

PR #5514, opened 2020-06-11
[Feature:TAGrading]Peer_feedback_backend_implementation

jchulton:
### Please check if the PR fulfills these requirements:

* [ ] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
There is a peer feedback interface, but it doesn't actually save user input to the backend: it is just a UI
### What is the new behavior?
A student can give feedback via the UI, and a peer grader can view that feedback via the TA_Grading page.

Steps to use:
1. View grdaes of a student who has been peer graded
2. Use the Switch to Feedback Display button to switch to the feedback UI
3. Select a peer tab to give feedback to
4. Select the feedback to give in the peer feedback bar
5. Switch to the peer who have just given feedback to
6. Go to the TA_Grading page in which you graded that student who gave feedback. You should see the feedback at the bottom of the grading rubric

image

PR #5443, opened 2020-06-02
New stats page

cellocodingryan:
### Please check if the PR fulfills these requirements:

* [ ] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
None, this is a new page

### What is the new behavior?
New page for students to see current and past load on submitty for the current semester (fall or spring)

### Other information?
This pull request should not affect other parts of submitty

PR #5425, opened 2020-05-30
Randomization for peer graded assignments.

josharsh:
### What is the current behavior?
Peer grading is done by uploading a CSV file. 

### What is the expected new behavior?
Peer grading is assigned randomly among students in registration sections with a button click.

PR #5534, opened 2020-06-15
[Feature:PHP] Move course endpoints under top-level /courses/ URL path

### LINKED ISSUE #5388: Move /{_semester}/{_course}/ endpoints under /courses top-level
The current resource list for resolving a URL request looks something like:

* /authentication/login
* /current_user/change_password
* /api/courses
* /home/courses
* /{_semester}/{_course}
* /{_semester}/{_course}/office_hours_queue
* etc.

To support the first group of things that are not under a course, we need a couple of things to happen right now:
* must have a name that starts before `MiscController.php`
* cannot be in a sub-folder
* must be added as a special case in the WebRouter for authentication stuff

While the design made sense at first (as we only had one top-level page), as we expand to more things (like #5337 and #5353), dealing with the above is increasingly going to be a pain, regardless of documentation.

I'd like to propose to move all existing `/{_semester}/{_course}` endpoints under a top-level `/course` value. This would resolve any issues of the above, and make it easy to do the right thing with regards to routing.

The con would be that most URLs in the system would be a bit longer and break any previous links. I'm not sure how much people are using past course links after a semester ends, but should aim to do this between spring and fall semesters.

MasterOdin:

* [x] Documentation has been updated/added if relevant: Submitty/submitty.github.io#243

Closes #5388

Urls to the course pages are all under `/{_semester}/{_course}` right now. The decision to use this scheme was from when we thought we were only going to have a couple of top-level endpoints (`/api`, `/authentication`, `/home`). As we've continued to add new endpoints, this distinction has become messy, and a source of error as developer has to ensure their controller gets loaded before all other controllers by symfony or else their route is un-reachable.

### What is the new behavior?

This moves all of the course pages to be under a top-level `/courses` path (`/courses/{_semester}/{_course}`) which makes collisions at the top level impossible (i.e. `{_semester}` can be named anything now like `api` if so wanted), as well as removes the issue with controller loading as discussed in linked issue.

### Other information?

To test, I clicked around the site, made a gradeable, made a forum thread and posts, managed users, and nothing seemed to break.

Open PRs that will need their routes modified:
- [ ] #5529
- [ ] #5514
- [ ] #5443
- [ ] #5425
- [ ] #5416
- [ ] #5227


bmcutler:
@MasterOdin please resolve conflicts

MasterOdin:
@bmcutler done

PR #5539, opened 2020-06-16
[Feature:System] Preferred User Photos

### LINKED ISSUE #5318: More Student Info:  Local Time Zone & Student Photo

### LINKED ISSUE #5494: Submitty needs access to a PHP image library
Working on the image portion of #5318 I need the ability to resize images.  Currently I can't find any good way using the native functionality of PHP.

I'd like to have PHP configured to include one of the two libraries:
- https://www.php.net/manual/en/book.imagick.php
- https://www.php.net/manual/en/book.image.php

This problem doesn't appear to be as simple as pulling in a library with composer.  I believe it requires special configuration of PHP.

Kevin-MM:
Closes #5494
Closes #5318

Student images are saved on a per-course basis.  There are several different directories student images might be uploaded into and there is just as much duplicated code which scans these directories.

#### New Functionality
- Users may now upload a preferred image of themselves by clicking the pencil on the homepage (this is the same modal that allows you to change your preferred name).
- Also on the homepage users will now see their user image.  If they have not uploaded one they will see the one supplied by the system admin or instructor.  If no system image exists then they won't be shown an image.
- Instructors have the ability to flag student preferred images as inappropriate by clicking the flag icon next to the student user id on the student photos page.  Once a student has had their image flagged they lose the ability to upload new preferred images.

#### New Directory Structure
- A new directory structure has been introduced so that images are now saved on a per-user basis instead of a per-course basis.
- The `Student Photos` page and upload feature has been updated to retrieve/save images using this new directory structure
- User data will be stored at `/var/local/submitty/user_data`
- Images uploaded by system admin or instructor will be placed in `/var/local/submitty/user_data/<user_id>/system_images/<image_file>`
- Images uploaded by users will be saved to `/var/local/submitty/user_data/<user_id>/user_images/<image_file>`


site/app/controllers/grading/ImagesController.php
```diff
@@ -4,51 +4,26 @@
...
 use Symfony\Component\Routing\Annotation\Route;
 use app\libraries\Utils;
+use app\libraries\routers\AccessControl;
 
 class ImagesController extends AbstractController {
+
     /**
      * @Route("/courses/{_semester}/{_course}/student_photos")
      */
     public function viewImagesPage() {
```
shailpatels, line 18: Instead of the PHP user group check you can use the `@AccessControl(role="STUDENT")` route annotation and avoid the extra code

site/app/controllers/grading/ImagesController.php
```diff
@@ -181,4 +177,21 @@ public function ajaxUploadImagesFiles() {
...
+    public function flagUserImage() {
+        $result = $this->core->getQueries()->updateUserDisplayImageState($_POST['user_id'], 'flagged');
+
+        if ($result) {
+            $this->core->addSuccessMessage($_POST['user_id'] . '\'s image was successfully flagged.');
+            $this->core->redirect($this->core->buildCourseUrl(['student_photos']));
+        }
+        else {
+            $this->core->addErrorMessage('Some error occurred flagging ' . $_POST['user_id'] . '\'s image.');
+            $this->core->redirect($this->core->buildCourseUrl(['student_photos']));
```
shailpatels, line 194: You can get rid of each redirect call here and instead return a response object after the else statement so it looks like

```php
return MultiResponse::RedirectOnlyResponse(
    new RedirectResponse($this->core->buildUrl(['student_photos']))
);
```

site/app/models/DisplayImage.php
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        $sub_dir = null;
+        if ($display_image_state === 'system' || $display_image_state === 'flagged') {
+            $sub_dir = 'system_images';
+        }
+        elseif ($display_image_state === 'preferred') {
+            $sub_dir = 'user_images';
+        }
+
```
shailpatels, line 77: There are only 3 image states right - can this be re-written as 
```php
sub_dir = $display_image_sate === 'preferred' ? 'user_images' : 'system_images';
```

site/public/js/server.js
```diff
@@ -1615,3 +1615,26 @@ function enableKeyToClick(){
     });
   }
 }
+
```
shailpatels, line 1618: Is there a reason why this is specifically creating a form through JS ? 

It seems like this function should instead either use ajax or fetch?
I'm not sure if we agreed on transitioning fro ajax to the fetch webApi yet but you can instead try something like:

```js
fetch(buildCourseUrl(['flag_user_image']), {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
  },
  body: JSON.stringify(  
     { 'user_id' : user_id, 'csrf_token' : csrf_token }
  ),
})
.then(response => response.json());
```
or just use plain ajax

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -437,6 +440,7 @@ cp  ${SUBMITTY_REPOSITORY}/.setup/bin/reupload_old_assignments.py   ${SUBMITTY_I
 cp  ${SUBMITTY_REPOSITORY}/.setup/bin/reupload_generate_csv.py   ${SUBMITTY_INSTALL_DIR}/.setup/bin/
 cp  ${SUBMITTY_REPOSITORY}/.setup/bin/track_git_version.py   ${SUBMITTY_INSTALL_DIR}/.setup/bin/
 cp  ${SUBMITTY_REPOSITORY}/.setup/bin/init_auto_rainbow.py   ${SUBMITTY_INSTALL_DIR}/.setup/bin/
+cp  ${SUBMITTY_REPOSITORY}/.setup/bin/setup_sample_user_data.py   ${SUBMITTY_INSTALL_DIR}/.setup/bin/
```
MasterOdin, line 443: This should only be copied if there's an expectation this will be run in production

migration/migrator/migrations/system/20200518135745_create_user_data_dir.py
```diff
@@ -0,0 +1,39 @@
...
+    script_path = os.path.join(config.submitty['submitty_install_dir'], '.setup', 'bin', 'setup_sample_user_data.py')
+
+    # Generate user_data directory
+    if not os.path.isdir(user_data_path):
+        os.system('mkdir ' + user_data_path)
+        os.system('chmod 770 ' + user_data_path)
+        shutil.chown(user_data_path, user, user)
+
+    # Execute script to populate user_data directory with sample images
+    os.system('python3 ' + script_path)
```
MasterOdin, line 26: What is the purpose to running this in production?

migration/migrator/migrations/system/20200518135745_create_user_data_dir.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    user_data_path = os.path.join(config.submitty['submitty_data_dir'], 'user_data')
+    user = config.submitty_users['php_user']
+    script_path = os.path.join(config.submitty['submitty_install_dir'], '.setup', 'bin', 'setup_sample_user_data.py')
+
+    # Generate user_data directory
+    if not os.path.isdir(user_data_path):
+        os.system('mkdir ' + user_data_path)
+        os.system('chmod 770 ' + user_data_path)
```
MasterOdin, line 22: Use the `os` functions of `mkdir` and `chmod` instead of `os.system`.

site/app/controllers/GlobalController.php
```diff
@@ -211,43 +211,14 @@ public function header() {
...
-                        $sidebar_buttons[] = new Button($this->core, [
-                            "href" => $this->core->buildCourseUrl(['student_photos']),
-                            "title" => "Student Photos",
-                            "class" => "nav-row",
-                            "id" => "nav-sidebar-photos",
-                            "icon" => "fa-id-card"
-                        ]);
-                    }
-                }
+                $at_least_one_grader_link = true;
```
MasterOdin, line 214: this (and all usages of it) can be removed. There's never a time where `$this->core->getUser()->accessGrading()` will be `true` and `$at_least_one_grader_link` will be `false`.

site/app/controllers/HomePageController.php
```diff
@@ -95,6 +96,20 @@ public function changeUserName() {
...
             }
+
+            // If we received an image file attempt to save it
+            if ($_FILES['user_image']['tmp_name'] !== '') {
+                $meta = explode('.', $_FILES['user_image']['name']);
+                $file_name = $meta[0];
+                $extension = $meta[1];
+
+                // Save image for user
+                $result = $user->setDisplayImage($file_name, $extension, $_FILES['user_image']['tmp_name']);
```
MasterOdin, line 107: Please move the DB and file system logic out of the model. The models should be just containers for information that the controller uses to do whatever. Adding the functionality into the model makes testing them that much more difficult.

site/app/controllers/grading/ImagesController.php
```diff
@@ -126,7 +101,10 @@ public function ajaxUploadImagesFiles() {
         }
 
         if (isset($uploaded_files[1])) {
+            $users = $this->core->getQueries()->getListOfCourseUsers();
+            // For each item that was uploaded
```
MasterOdin, line 105: Can remove this comment and the others added in this function. The code itself is readable enough on what it's doing, and these comments do not provide any further meaningful context.

site/app/libraries/Output.php
```diff
@@ -100,17 +100,11 @@ public function loadTwig($full_load = true) {
...
             return call_user_func_array('self::renderTemplate', $args);
         }, ["is_safe" => ["html"]]));
-        $this->twig->addFunction(new \Twig\TwigFunction('base64_image', function (string $path, string $title): string {
-            $valid_image_subtypes = ['png', 'jpg', 'jpeg', 'gif'];
-            [$mime_type, $mime_subtype] = explode('/', mime_content_type($path), 2);
-            if ($mime_type === "image" && in_array($mime_subtype, $valid_image_subtypes)) {
-                // Read image path, convert to base64 encoding
-                $image_data = base64_encode(file_get_contents($path));
+        $this->twig->addFunction(new \Twig\TwigFunction('base64_image', function (string $base64_data, string $mime_type, string $title): string {
+                // Return html image tag with image embedded as a base64 character string
```
MasterOdin, line 104: can remove

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -206,6 +206,32 @@ public function getAllUsers($section_key = "registration_section") {
...
+     * Update master and course database user's display_image_state to a new state
+     *
+     * @param string $user_id
+     * @param string $state
+     * @return bool
+     */
+    public function updateUserDisplayImageState(string $user_id, string $state): bool {
+        $sql = 'update users set display_image_state = ? where user_id = ?';
+        $this->submitty_db->query($sql, [$state, $user_id]);
+        return ($this->submitty_db->getRowCount() === 1);
```
MasterOdin, line 232:
```suggestion
        return $this->submitty_db->getRowCount() === 1;
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -206,6 +206,32 @@ public function getAllUsers($section_key = "registration_section") {
         return $return;
     }
 
+    /**
+     * Gets an indexed array of all the user_ids for all users who are members of the current course.
+     *
+     * @return array An array of all the users
+     */
+    public function getListOfCourseUsers(): array {
+        $sql = 'select user_id from users';
```
MasterOdin, line 215:
```suggestion
        $sql = 'SELECT user_id FROM users';
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -206,6 +206,32 @@ public function getAllUsers($section_key = "registration_section") {
...
+
+    /**
+     * Update master and course database user's display_image_state to a new state
+     *
+     * @param string $user_id
+     * @param string $state
+     * @return bool
+     */
+    public function updateUserDisplayImageState(string $user_id, string $state): bool {
+        $sql = 'update users set display_image_state = ? where user_id = ?';
```
MasterOdin, line 230:
```suggestion
        $sql = 'UPDATE users SET display_image_state = ? WHERE user_id = ?';
```

site/app/models/DisplayImage.php
```diff
@@ -0,0 +1,212 @@
...
+     *
+     * You may optionally resize the image.
+     *
+     * You may pass both parameters to have fine control over both image dimensions.
+     *
+     * You may pass only a single parameter and leave the other one null.
+     * Doing so will resize that dimension to the specified length, while maintaining the form factor of the image.
+     *
+     * @param int|null $cols Length in pixels to resize this dimension to
+     * @param int|null $rows Length in pixels to resize this dimension to
```
MasterOdin, line 115: I get that imagemagick uses this terminology, but I would argue we should use `height` and `width` like most other image software.

site/app/models/DisplayImage.php
```diff
@@ -0,0 +1,212 @@
...
+     * @return string The base64 encoding of the image
+     * @throws \ImagickException
+     */
+    public function getImageBase64(int $cols = null, int $rows = null): string {
```
MasterOdin, line 119:
```suggestion
    public function getImageBase64(?int $cols = null, ?int $rows = null): string {
```

https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types

site/app/models/DisplayImage.php
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        // Decrease image size while maintaining form factor
+        // If bigger image dimension is already smaller then IMG_MAX_DIMENSION don't do any resizing.
+        $imagick = new \Imagick($tmp_file_path);
+        self::resizeMaxDimension($imagick, self::IMG_MAX_DIMENSION);
+
+        // Save file (will overwrite any image with same name)
+        $imagick->writeImage(FileUtils::joinPaths($folder_path, $new_image_name . '.' . $image_extension));
+    }
```
MasterOdin, line 211: Neither of these functions make any usage of the model code other than a class constant. Please move them into a `ImageUtil` class under `app/libraries` similar to the other static functions we have in this vein.

site/app/models/User.php
```diff
@@ -188,6 +191,15 @@ public function __construct(Core $core, $details = []) {
             $this->setGradingRegistrationSections($details['grading_registration_sections']);
         }
 
+        if (isset($details['display_image_state'])) {
+            try {
+                $this->display_image = new DisplayImage($core, $details['user_id'], $details['display_image_state']);
+            }
+            catch (\Exception $exception) {
+                $this->display_image = null;
+            }
```
MasterOdin, line 200: Can we just store `$details['display_image_state']` to `$this->display_image` and then make a concrete `getDisplayImage()` that does the `DisplayImage` construction. If we find we're calling `getDisplayImage` multiple times on the same User model, then we can cache the result in that function, but not sure we are.

We should not waste the cycles on constructing the model unless we actually are going to call this function, which I do not think we do in a large number of cases.

.travis.yml
```diff
@@ -98,11 +98,20 @@ jobs:
     - &02-php-unit
       stage: 02 unit test
       php: 7.2
+      addons:
+        apt:
+          sources:
+            - sourceline: "deb http://archive.ubuntu.com/ubuntu bionic multiverse"
+            - sourceline: "deb http://archive.ubuntu.com/ubuntu bionic-updates multiverse"
```
MasterOdin, line 105: Just confirming that adding these are necessary to install `imagemagick` and `php-imagick`?

.travis.yml
```diff
@@ -98,11 +98,20 @@ jobs:
...
+          packages:
+            - imagemagick
+            - php-imagick
```
MasterOdin, line 108: how is this different than the `pecl install imagick`?

.setup/distro_setup/ubuntu/bionic/setup_distro.sh
```diff
@@ -89,6 +89,7 @@ update-java-alternatives --set java-1.8.0-openjdk-amd64
 
 # Install Image Magick for image comparison, etc.
 apt-get install -qqy imagemagick
+apt-get install -qqy php-imagick
```
MasterOdin, line 92: move this up with the other `php-*` installs so that they're all consolidated in one place.

bmcutler:
Review of functionality (haven't look at code yet)

looks nice :)

---

* I thought were were going to name the files with the date.  But it looks like we are preserving the upload filename.  Are we relying on the timestamp of the files to determine the most recent?  This is fragile (if the files are copied/restored from backup without preserving timestamps we will lose important information).

---

The implementation of the flagging is incomplete, so perhaps should be commented out for this first version.

* After "flagging" a student uploaded image is inappropriate, both the student & their instructors should see an indication that the student image is skipped.  

* There should be an option for the instructor to "undo" this action.

* The student should be able to upload a new image if their previous image was flagged as inappropriate.  (I got an unhelpful error on attempted upload.)

---

* existing system bug: Images on the student photos page are displayed with fixed aspect ratio.  It should display the image with the aspect ratio of the original photo (like on the user homepage).

migration/migrator/migrations/system/20200518135745_create_user_data_dir.py
Kevin-MM: It shouldn't have been there.  I removed it.

.setup/INSTALL_SUBMITTY_HELPER.sh
Kevin-MM: Removed

site/app/models/DisplayImage.php
Kevin-MM: Changed

site/app/controllers/grading/ImagesController.php
Kevin-MM: Changed

site/app/controllers/grading/ImagesController.php
Kevin-MM: Changed

site/app/controllers/grading/ImagesController.php
Kevin-MM: I worked with this code quite a bit and I didn't find it readable, thus the comments.  I stand behind keeping them.

site/app/models/DisplayImage.php
Kevin-MM: I would argue that we aren't writing image software, we are consumers of it.  So I think it makes sense to use the same conventions as the the software we are using.

migration/migrator/migrations/system/20200518135745_create_user_data_dir.py
Kevin-MM: Done

site/app/controllers/GlobalController.php
Kevin-MM: Done

site/app/libraries/Output.php
Kevin-MM: Removed

.travis.yml
Kevin-MM: It looks like pecl install is all thats needed.  I removed the other lines.

.travis.yml
Kevin-MM: It looks like pecl install is all thats needed.  I removed the other lines.

site/app/controllers/grading/ImagesController.php
MasterOdin: If the code is hard to understand / read that these short one line comments are necessary, that's a code smell and that this should be cleaned / modified in some way to be clearer. However, I'm not convinced (yet) and would like to perform a dive on why you think the comment is necessary so can think on making this block of code more readable.

Let's start with this comment: `// Item was a zip file`

The line of code it comments: `if ($uploaded_files[1]["is_zip"][$j] === true) {`.

What does this comment add to understanding this line is testing that something is a zip?

site/public/js/server.js
Kevin-MM: I'm going to shelve this until a later PR.  See #5587.

site/app/controllers/HomePageController.php
Kevin-MM: I'm going to shelve this until a later PR.  See #5587.

site/app/models/DisplayImage.php
Kevin-MM: I'm going to shelve this until a later PR.  See #5587.

site/app/models/User.php
Kevin-MM: I'm going to shelve this until a later PR.  See #5587.

bmcutler:
The flag feature works well if a student has a system uploaded photo.  (the instructor can undo and the student sees the message)

But if a user has no photo, uploads a preferred photo, then that photo is flagged...  the photo is just empty again.  Neither the undo option nor the student message appear.

Kevin-MM:
@bmcutler The newest code will resolve the issue you experienced.

site/app/controllers/grading/ImagesController.php
bmcutler: I don't have a problem with this comment, let's not make a fuss over it

bmcutler:
@Kevin-MM Let me know when you're done with the  bugfixes/tweaks for this PR.

PR #5542, opened 2020-06-16
[Bugfix:UI] Fix validation of email failing valid addresses

### LINKED ISSUE #5489: Regex-checks for user_email too strict
**Describe the bug**
Some valid emails are rejected by the regex-checks.
Affected mails have a '-' in the domain-part other than the first label.
For example `student@foo-bar.university.edu` is accepted, but `student@faculty.university-of-xyz.edu` is rejected.

**Expected behavior**
Emails like `student@faculty.university-of-xyz.edu` should be accepted since they are valid.

**To Reproduce**
1. Log in to a course
2. Manage students
3. Click "New Student"
4. Enter `student@faculty.university-of-xyz.edu` as Email
5. Click in another field so that the Email gets checked
6. See Email marked as invalid

**Configuration**
 - Server-OS: Linux
 - Client-OS: Windows
 - Browser: Firefox

**Additional context**

/site/app/models/User.php line 344:
`^[^(),:;<>@\\\"\[\]]+@(?!\-)[a-zA-Z0-9\-]+(?<!\-)(\.[a-zA-Z0-9]+)+$`
'-' is not allowed in domain-parts other than the first one.
Suggestion:
`^[^(),:;<>@\\\"\[\]]+@(?!\-)[a-zA-Z0-9\-]+(?<!\-)(?:\.(?!\-)[a-zA-Z0-9\-]+(?<!\-))+$`
Basically it adds dash to later labels, including the negative-checks to prevent labels from starting or ending with dash. Also use non-capturing group since there is no point in capturing.

/site/public/js/userform.js line 170:
`^[^(),:;\<\>@\\"\[\]]+@(?!\-)[a-zA-Z0-9\-]+(?!\-)(\.[a-zA-Z0-9]+)+$`
Here the negative-lookbehind from the previous file was changed to a negative-lookahead. That makes no sense because the following group starts with a '.', so it can never be '-'.
Lookbehinds are only available since 2018 in JS, so maybe it should not be used.
Suggestion:
`^[^(),:;\<\>@\\"\[\]]+@(?!\-)[a-zA-Z0-9\-]*[a-zA-Z0-9](?:\.(?!\-)[a-zA-Z0-9\-]*[a-zA-Z0-9])+$`
Work around lookbehind limitation by requiring one non-dash character at the end of each label, plus optionally other characters before that, but prevent starting with dash using negative-lookahead; allow dash in later labels; use non-capturing group

MasterOdin:

Fixes #5489

The current regex used for email validation rejects valid email addresses.

### What is the new behavior?

This changes it so that on the PHP side of things we use the [egulias/email-validator](https://packagist.org/packages/egulias/email-validator) composer package to validate email addresses (which is what symfony itself uses). For the JS side, while I propose in the issue to use the builtin browser validation of `type="email"`, unless we switch all fields over to using it, it does not appear consistent. As a stop-gap, I've changed the regex to one that more closely resembles what the HTML5 spec calls for doing for email inputs.

Both PHP and JS now allow the provided email address in the issue (`person@faculty.university-of-xyz.edu`).


bmcutler:
please fix merge conflict
(How does composer.lock get conflicted?)

MasterOdin:
The file contains a hash of the contents of the composer.json file, to ensure that the lockfile lines up with the json file. So anytime a PR gets merged with composer changes, any open PRs with composer changes also need to be updated to reset the hash. Dependabot does this action automatically for us, so usually it should be that we hold-off merging any PHP dependabot PRs until we merge any open (from us) PRs with new dependencies to make it easier to avoid the conflict.

Issue #4194, opened 2019-07-25
No scrollbar on TA grading interface


Kevin-MM:
**Describe the bug**
No scroll bar on the TA grading interface (see image below).

**Expected behavior**
If the navigation bar is greater than the height of the window, you should be able to scroll down to elements on the bottom of the navigation bar

**To Reproduce**
1. Log in as instructor
2. Enter the sample course
3. Click Gradeables
4. Grade the 'Grading Homework'
5. Click 'Index of Grading Details...'
6. Click any of the 'Grade' buttons
7. If the height of the navigation bar is not greater than the height of the window, zoom your browser until it is.
8. Observe that there is no scroll bar, and you can no longer access the bottom part of the navigation bar.

**Configuration**
 - OS: Ubuntu 18.04
 - Browser Chrome

**Screenshots**
![lost_scroll_bar](https://user-images.githubusercontent.com/7605020/61874727-21c02700-aeb7-11e9-97a3-b99dcc59c055.png)

bmcutler:
Something has changed, but it's not that we want to add a scrollbar to the whole screen including the panels.  

The side bar should be able to move up & down (and the breadcrumb can moves off the top of the screen).

The forum has a similar problem.  Please see how it works right now (it hasn't changed).

Also tagging @andrewaikens87 because maybe he can explain what the ta grading page used to do when the side bar is taller than the page.

mukul-kmr-jha:
I think It can be closed, thanks!

PR #5543, opened 2020-06-17
New TA grading interface

mukul-kmr-jha:

### What is the new behavior?
Adds a new TA grading interface (beta). 

How to test and review this new interface.
There is a new button added in the TA navigation bar from where you can access this interface.
image

What's new in this interface?
- Fixes the sidebar menu scroll [issue](https://github.com/Submitty/Submitty/issues/4194) as earlier it was not possible to pick the menu items which were offscreen. 
- Scraps out the earlier floating panel layout as they were hard to manage and less acessible. 
- A fixed structure is added (aka single-panel-mode) where a single panel is shown at a time taking all the available space.
- Each panel can be toggled with the help of buttons given inside the new panel-header-box which corresponds to one of these panels OR by just pressing the corresponding hotkey which can be configured on-site.
- Introduces a two-panel-mode for power-users of this interface, where two panels can be opened side-by-side. A drag-bar is provided in between these two panels which can be used to resize and adjust the width of panels as per the user's choice.
- Introduces full-left-column-mode where left column of the two-panel-mode takes the complete height of its container and the navigation-bar + panels-header-box is shifted to the right side of the layout. And this is again resizeable just as the normal two-panel-mode.
- Also there is a button provided in the navigation bar to exchange the positions of currently opened panels i.e panel opened in left goes to right and right one goes to left. 
- Adds another power-user feature i.e the full-screen-mode or focus-mode in which the grading interface goes full screen and there is no side-bar or submitty top navigation bar, and this works for both single and two-panel-mode.
 
### Screenshots

image

### Other information?
~This interface doesnot know about all panels that are present for a certain gradeable, and might be missing some key panels (or displaying extra panel) like peer-grading panel or discussion panel, and I'm working on that.~ DONE


elihschiff:
When you first load up a page the open "Autograding Testcases" button is not filled in red even though it is open. If you click on the button twice (once to close, once to reopen) the button is then filled in.
When page first loads
image
After I click on the button twice:
image

Also, if you open a layout and then switch to the next student it always resets to just "Autograding Testcases" for me. It would be nice if you layout stayed between students. It would be nice if it remembered what layout and which panels were open and keep them open between students.

mukul-kmr-jha:
Yes, I have mentioned this in ta-grading-v2.js and I also think it will be nice to have such type of persistence...Thanks!

mukul-kmr-jha:
@emaicus  implemented the suggested changes. Please take another look, Thanks!

site/app/templates/grading/electronicV2/DiscussionForumPanel.twig
```diff
@@ -0,0 +1,19 @@
+{% import _self as self %}
+<div id="discussion_browser" class="draggable rubric_panel" style="display:block;">
```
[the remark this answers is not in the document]
mukul-kmr-jha: Nope, I missed this.

mukul-kmr-jha:
Hii Kevin, the first two comments of your review are addressed on this [PR 5609](https://github.com/Submitty/Submitty/pull/5609) which is just the mobile view for TA-grading interface.
And for the `full-left-col` button let's discuss about where should it go and which icon to use as @emaicus also requested changes for that. Thanks!

site/public/css/electronic.css
```diff
@@ -0,0 +1,1236 @@
...
+}
+
+/* Animations */
+@keyframes scaleUp {
+    from {
+        position: fixed;
+        height: 40vh;
+        top: 30%;
+        left: 20%;
+    }
```
emaicus, line 1227: I don't know much about css animation, but can we make the panels start from their current position rather than specifying a starting position?

mukul-kmr-jha:
Hii @emaicus, I'm working on all those suggested points, and I think I can address the following issues in a follow-up PR (if we are fine with that) 

- [ ] The background of the panels does not respect the background color of the page.
- [ ] Proper fix for the information regarding the student status (Late, canceled, no submission)
- [ ] There is no way to leave the new interface and return to the old one (We don't need to work on this, as per @bmcutler )
- [ ] Top panel buttons (e.g. forward/backward) do not highlight on tab navigation.

site/public/css/electronic.css
mukul-kmr-jha: added a little change in animation

PR #5563, opened 2020-06-19
[UI/UX:InstructorUI] Notebook Builder 1.0

Kevin-MM:
### What is the new behavior?
Started implementation of the new notebook builder UI.  To access the UI first install this branch then visit http://localhost:1501/courses/s20/sample/notebook_builder .

- This is only a rough implementation of the UI.  It is not hooked up to any gradeable.
- The `Show JSON` button can be used to view a mock up of the notebook portion of an autograding configuration
- The `Show JSON` button is only for testing and debugging, it won't be included in its current form in future implementations


bmcutler:
This looks prettty good!  nice start!

Cosmetic/functional change requests (don't have to be completed before merging this PR):

* Please make the box for the markdown cell type start taller (3 lines?)

* If I resized the markdown textbox too wide:

Screen Shot 2020-06-23 at 9 40 51 AM

* Let's a on page documentation for markdown -- a simple link to basic markdown documentation.  Placeholder link to start:
https://submitty.org/student/discussion_forum#formatting-a-post-using-markdown

* The textbox for the "value" of multiple choice should be visually smaller than the "description" box. 

* The description box for multiple choice should allow newlines (full markdown).  Probably make this a 2 line tall box initially (and make it resizeable).

* It wasn't clear that you have to press "Add" to finalize the addition of a choice for multiple choice.  It sort of looks like you can just fill it out and it will be part of the form.

* How do you edit or reorder previously added choices for multiple choice?

* How do you specify the height of a short answer? (make it an input when 'large' is selected?)

* Make the "selector" box (to add a new cell) visually different (different color?) than the notebook cells.  Also maybe label it "Add New Notebook Cell" instead of just "Selector"?

site/public/js/notebook_builder/notebook-builder.js
```diff
@@ -0,0 +1,106 @@
...
+        };
+    }
+
+    /**
+     * Add a widget to the notebook builder form
+     *
+     * @param widget
+     */
+    widgetAdd(widget) {
+        this.widgets.splice(this.widgets.length - 1, 0, widget);
```
MasterOdin, line 56: Is this not just `this.widgets.push(widget)`?

site/public/js/notebook_builder/notebook-builder.js
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Move a widget down one position in the notebook builder form
+     *
+     * @param widget
+     */
+    widgetDown(widget) {
+        const index = this.widgets.indexOf(widget);
+
+        // If widget is already at the end of the form then do nothing
+        if (index === this.widgets.length - 2) {
```
MasterOdin, line 98: `this.widgets.length - 1`?

site/public/js/notebook_builder/notebook-builder.js
Kevin-MM: No the widget array also contains the 'selector' widget which should always be last.  Any new widgets should be inserted next to last.

site/public/js/notebook_builder/notebook-builder.js
Kevin-MM: No the selector widget should always be last so we don't want the down button to allow widgets to move below it.

site/public/js/notebook_builder/notebook-builder.js
Kevin-MM: I think I may end up refactoring this when it's needed to support a top, mid, and bottom widget arrays.  Bottom will hold the selector.  Top will hold configuration options that are not part of the notebook, for example assignment message.  Then mid will hold notebook widgets and will be the only part that allows reordering.  This will simplify some of the logic for rearrangement of widgets.

MasterOdin: Can we not do this now with this bottom widget? Unless there's a design that we're going to be swapping the bottom out with something else, making it part of this list is just confusing and makes it too easy to introduce sorting bugs for the unwary.

PR #5560, opened 2020-06-19
[Refactor:Autograding] Lint testcase.py

emaicus:
Removes  `autograder/autograder/testcase.py`, `autograder/autograder/__init__.py`, and `autograder/autograder/execution_environments/__init__.py` from the `Flake8` exclusion list and updates `testcase.py` to conform to our style guide.

PR #5570, opened 2020-06-22
[Refactor:Autograding] Lint autograding shipper/worker

### LINKED ISSUE #5558: Lint autograding_shipper/worker
- [ ] sbin/submitty_autograding_shipper.py,
- [ ] sbin/submitty_autograding_worker.py,

hdrodz97:

Closes #5558

`sbin/submitty_autograding_shipper.py` and `sbin/submitty_autograding_worker.py` were excluded from linting with flake8.

Additionally, thanks to linting now being enabled for these files, two typos were found in `submitty_autograding_shipper.py`. 

The first typo was found in the `establish_ssh_connection` function; the SSH connect retry delay was attempted to be set to at most 10 seconds, but the delay was not actually propagated as the result was assigned to `retry_relay` instead of `retry_delay` (the former of which got flagged by the linter as an unused variable).

The second typo was found in the `grading_cleanup` function; if an error was encountered removing either the grading or queue files, then an `AttributeError` would be thrown as the calls to the logging functions make reference to a `my_name` variable which does not exist in the function's scope.

### What is the new behavior?

`sbin/submitty_autograding_shipper.py` and `sbin/submitty_autograding_worker.py` are no longer excluded from linting with flake8 and have been properly linted.

The SSH connect retry delay is properly limited to be at most 10 seconds.

A `my_name` parameter was added to the `grading_cleanup` function so we can propagate this value into the function.

### Other information?

While linting this file, I found three different ways of formatting strings: f-strings, the `.format()` function, and concatenation. I left these as-is as the linter doesn't complain about this; however, I do think as we go through these files we should standardize one method, preferably f-strings.


sbin/submitty_autograding_shipper.py
```diff
@@ -115,60 +118,73 @@ def update_worker_json(name, entry):
         host = autograding_worker_to_ship[name]['address']
     except Exception as e:
         print("ERROR: autograding_workers.json entry for {0} is malformatted. {1}".format(e, name))
-        autograding_utils.log_message(AUTOGRADING_LOG_PATH, JOB_ID, message="ERROR: autograding_workers.json entry for {0} is malformed. {1}".format(e, name))
-        autograding_utils.log_stack_trace(AUTOGRADING_STACKTRACE_PATH, job_id=JOB_ID, trace=traceback.format_exc())
+        autograding_utils.log_message(AUTOGRADING_LOG_PATH, JOB_ID,
```
emaicus, line 121: A copied comment from @MasterOdin from his review of pr #5560 which I agree with: 
> I find for Python (and other languages), that if the list of arguments to a function or whatever extends past one line, to just have the starting element on its own newline with one level of indentation. It's I think a bit nicer looking, easier to add to the list a longer argument (as you're not starting a line at character 78 or whatever) and that indentation is a regular multiple instead of starting at some non-standard number.

sbin/submitty_autograding_shipper.py
```diff
@@ -115,60 +118,73 @@ def update_worker_json(name, entry):
...
-            ssh = establish_ssh_connection(None, user, host, only_try_once = True)
+            ssh = establish_ssh_connection(None, user, host, only_try_once=True)
         except Exception as e:
-            autograding_utils.log_stack_trace(AUTOGRADING_STACKTRACE_PATH, job_id=JOB_ID, trace=traceback.format_exc())
-            autograding_utils.log_message(AUTOGRADING_LOG_PATH, JOB_ID, message="ERROR: could not ssh to {0}@{1} due to following error: {2}".format(user, host,str(e)))
-            print("ERROR: could not ssh to {0}@{1} due to following error: {2}".format(user, host,str(e)))
+            autograding_utils.log_stack_trace(AUTOGRADING_STACKTRACE_PATH, job_id=JOB_ID,
+                                              trace=traceback.format_exc())
+            autograding_utils.log_message(AUTOGRADING_LOG_PATH, JOB_ID,
+                                          message="ERROR: could not ssh to {0}@{1} due to following error: {2}".format(user, host, str(e)))  # noqa: E501
```
emaicus, line 158: We are trying to move away from `.format`, so if a `.format` makes a line to long, absolutely convert it to a `fstring`.

sbin/submitty_autograding_shipper.py
```diff
@@ -407,67 +453,76 @@ def grade_queue_file(my_name, which_machine,which_untrusted,queue_file):
...
+    my_dir, my_file = os.path.split(queue_file)
     directory = os.path.dirname(os.path.realpath(queue_file))
     name = os.path.basename(os.path.realpath(queue_file))
     grading_file = os.path.join(directory, "GRADING_" + name)
 
     # Try to short-circuit this job. If it's possible, then great! Clean
     # everything up and return.
     if try_short_circuit(queue_file):
-        grading_cleanup(queue_file, grading_file)
+        grading_cleanup(my_name, queue_file, grading_file)
```
emaicus, line 464: What's going on here with the new parameter?

hdrodz97: (Can't really link to the direct line since Github collapses the diff of the file anyway, so bear with me)

In the `grading_cleanup` function, there are two exception handlers; one handling errors on removing queue files, and the other handling errors on removing grading files. These print a message both to stdout and the log, which reference a `my_name` variable that didn't exist in the function -- presumably due to the parameter never actually being added to the function. I added a `my_name` param to the function signature to rectify this, which is why there's a few `grading_cleanup` diffs in this style.

sbin/submitty_autograding_shipper.py
```diff
@@ -1206,20 +1428,29 @@ def launch_shippers(worker_status_map):
 
             num_workers_on_machine = machine["num_autograding_workers"]
             if num_workers_on_machine < 0:
-                raise SystemExit("ERROR: num_workers_on_machine for '{0}' must be non-negative.".format(machine))
+                raise SystemExit("ERROR: num_workers_on_machine for '{0}' must be non-negative.".format(machine))  # noqa: E501
```
emaicus, line 1431:
```suggestion
                raise SystemExit(f"ERROR: num_workers_on_machine for '{machine}' must be non-negative.")
```

sbin/submitty_autograding_worker.py
```diff
@@ -134,33 +174,43 @@ def launch_workers(my_name, my_stats):
     # launch the worker threads
     address = my_stats['address']
     if address != 'localhost':
-        which_machine="{0}@{1}".format(my_stats['username'], address)
+        which_machine = "{0}@{1}".format(my_stats['username'], address)
```
emaicus, line 177:
```suggestion
        which_machine = f"{my_stats['username']}@{address}"
```

sbin/submitty_autograding_worker.py
```diff
@@ -38,51 +38,85 @@
...
+                    autograding_utils.log_message(
+                        AUTOGRADING_LOG_PATH, JOB_ID,
+                        message=f"{e} {results_zip}"
+                    )
 
                 os.remove(results_zip_tmp)
                 with open(todo_queue_file, 'r') as infile:
                     queue_obj = json.load(infile)
-                    queue_obj["done_time"]=dateutils.write_submitty_date(milliseconds=True)
+                    queue_obj["done_time"] = dateutils.write_submitty_date(microseconds=True)
```
MasterOdin, line 98:
```suggestion
                    queue_obj["done_time"] = dateutils.write_submitty_date(milliseconds=True)
```

sbin/submitty_autograding_worker.py
```diff
@@ -92,10 +126,13 @@ def worker_process(which_machine,address,which_untrusted,my_server):
...
                 shutil.rmtree(tmp_dir)
-                done_queue_file = os.path.join(SUBMITTY_DATA_DIR,"autograding_DONE",servername_workername+"_"+which_untrusted+"_queue.json")
+                done_queue_file = os.path.join(
+                    SUBMITTY_DATA_DIR, "autograding_DONE",
+                    f"{servername_workername}_{which_untrusted}_queue.json"
+                )
                 with open(todo_queue_file, 'r') as infile:
                     queue_obj = json.load(infile)
-                    queue_obj["done_time"]=dateutils.write_submitty_date(milliseconds=True)
+                    queue_obj["done_time"] = dateutils.write_submitty_date(microseconds=True)
```
MasterOdin, line 135:
```suggestion
                    queue_obj["done_time"] = dateutils.write_submitty_date(milliseconds=True)
```

MasterOdin:
@emaicus The stack trace error is a result of the parameter rename from #5545. I've added suggestions on how to fix them.

hdrodz97:
> 1. It looks like the comment `# noqa: E501` was added throughout the linted files. Please remove it. I started making "suggested changes," but it looks like there are enough of them to warrant just doing a find and replace locally.

Regarding this, `#noqa: E501` disables the line length parameter, and I've only limited it to lines which display user-facing text. [This is done in the Linux kernel as well],(https://www.kernel.org/doc/html/v4.10/process/coding-style.html#breaking-long-lines-and-strings) argument being that breaking these strings over multiple lines can break the ability to search for them when trying to find the source of a message. I'm still open to breaking these strings up, but I figured this was a point worth considering.

emaicus:
I removed the `# noqa: E501` lines before reading @hdrodz97's comment above. After this is merged, let's have a conversation about when/if we want to allow devs to circumnavigate our style guide.

PR #5576, opened 2020-06-23
[Refactor:Router] Mark MultiResponse::only* deprecated

MasterOdin:

The methods are used, and without any indication within the codebase they should be ignored (though https://submitty.org/developer/router#what-is-the-responseinterface-object-is-it-required discusses what to return). #5206 discusses the rationale for this.

### What is the new behavior?

Adds a `@deprecated` tag to these functions, with an expectation that they will be removed from the codebase in follow-up PRs.

### Other information?

@shailpatels Since there's the push to have two pages on the subject, please update https://submitty.org/developer/development_instructions/router-response#the-multiresponse-object and following sections to fit this defined behavior

### DIFF
--- a/site/app/libraries/response/MultiResponse.php
+++ b/site/app/libraries/response/MultiResponse.php
@@ -35,6 +35,8 @@ class MultiResponse implements ResponseInterface {
     }
 
     /**
+     * @deprecated should not be used, just return WebResponse directly
+     *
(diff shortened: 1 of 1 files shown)


shailpatels:
https://github.com/Submitty/submitty.github.io/pull/257 Updates docs to match these changes

PR #5594, opened 2020-06-25
[Feature: Forum] Use websockets to update posts in Discussion Forum

marwanatef2:
The discussion forum use websockets only to update threads (adding, deleting, merging, etc.)

### What is the new behavior?
The discussion forum should fully work with websockets updating both threads and posts on real-time basis.

### Other information?
This PR should not be merged unless PR #5529  is merged since it is based on it.


site/app/controllers/forum/ForumController.php
```diff
@@ -428,11 +428,31 @@ public function publishPost() {
...
+     */
+    public function getSinglePost() {
+        $post_id = $_POST['post_id'];
+        $reply_level = $_POST['reply_level'];
+        $post = $this->core->getQueries()->getPost($post_id);
+        if (($_POST['edit']) && !empty($this->core->getQueries()->getPostHistory($post_id))) {
+            $post['edit_timestamp'] = $this->core->getQueries()->getPostHistory($post_id)[0]['edit_timestamp'];
+        }
+        $thread_id = $post['thread_id'];
+        $GLOBALS['totalAttachments'] = 0;
```
Kevin-MM, line 449: We would like to remove all usages of the $GLOBALS array.  If this isn't to much work could you include it in this PR?  If it is more trouble than its worth right now feel free to create an issue and someone can work on it later.

marwanatef2:
> Sorry for the slow turn around time on this review.

No problem for sure.

> If I am viewing a thread a blue border shows up around that thread in the thread list to denote which thread I have selected.  If someone edits the title or tags for that thread I get the appropriate updates, however the blue border disappears from the thread list, so its no longer clear which thread I have selected.

I fixed this issue, i was removing the "active" class which causes the blue border on new threads since upon their first creation they won't be active for any user but the author, so i removed this when it's only a new thread not a modified one.

> When using 'Reply to all' I would click the button once, but occasionally the post would be duplicated 2 or 3 times. 

This problem is a forum related not sockets related, i think if you tried on your master branch making a reply (new post) after multiple tries you would be able to see this, so please try this out and tell me so i can know this bug is related to which part of the code.

site/app/controllers/forum/ForumController.php
marwanatef2: in the previous version the _create_post_ function which takes the _totalAttachments_ variable passed as reference was only called in the _ForumThreadView.php_ file https://github.com/Submitty/Submitty/blob/67fedcf561cbb8bc9f9f98b30d65d41ecd1ad1ec/site/app/views/forum/ForumThreadView.php#L790
Now I'm calling this function in the _ForumController.php_ file so i couldn't pass this variable by reference from one file to another so i used global variables instead.
So we can open a new issue for this and use the global variable in the meantime.

MasterOdin: Let's make this a broad issue to clean up all of the usages of `$GLOBALS` in the ForumController and ForumView.

Kevin-MM:
As far as I know @MasterOdin had also mentioned implementing a socket routing scheme inside `Server.php` so I'd like to get his feedback before reviewing.

MasterOdin:
This PR has the basics of routing in it. It's fine for the immediate need, and we can revise it in follow-up work separate from this PR once this is merged.

Kevin-MM:
How dire is the immediate need?  I think we need to make a best effort to make sure the socket routing is as complete as it can be here.  We change it later, any features built on top of it will need modifications.  We can save time and effort by getting it set now.

marwanatef2:
I don't think changes to socket routing will affect the functionality built on it since those changes will supposedly be in _Server.php_ and _websocket.js_ files, other files just use the websockets.
The two socket configuration files i mentioned deal with the sockets **opening** and what should be done when a client tries to open a new socket connection; check authentication, etc. While other files just open the socket connections and deal with **sending & receiving** messages.

marwanatef2:
I fixed the duplicate post issue, when testing please post with attachments to make sure nothing breaks.

marwanatef2:
I changed some database query without changing all its function calls, now fixed.

PR #5604, opened 2020-06-30
[Feature:InstructorUI] office hour queue announcements & misc

elihschiff:

Right now there is only a way for an instructor to make an announcement on the office hours queue. However often times mentors want to write a note that students can also see.

Queue names and codes cannot have spaces in them. All spaces were converted to `_`

### What is the new behavior?

There is now an announcement box for mentors to add text to if they wish. The box also supports markdown
image
image

The filter settings was also moved into a modal to keep it more consistent with the design of the rest of the website. However the functionality should be exactly the same. 
image

Also, queue names and codes can have spaces and underscores in them. Do note that all queues currently open will default to having underscores in them. There is now way to know if a professor was telling their students to use spaces or underscores so there is a chance this causes confusion, but I do not know how I would get around this issue.


bmcutler:
@elihschiff please fix merge conflict

site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -134,7 +139,9 @@
             $("#current_queue").replaceWith(newQueue);
             filterQueues();
             if(notifications_enabled && queueEmpty && $( ".help_btn:visible" ).length > 0){
```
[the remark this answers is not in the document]
elihschiff: This was by design. The reason most people want notifications is that nobody is in office hours so they are doing something else like working on their own homework. When somebody joins the queue they will want a notification so they can pause their homework to help the student.

If office hours, all of a sudden, become busy then there is no need to send them a notification as they will see that more people need help as soon as they mark the current student as done being helped.

site/app/templates/officeHoursQueue/QueueHeader.twig
```diff
@@ -3,18 +3,38 @@
...
+      <a target="_blank" href="https://submitty.org/grader/queue">
+        Managing the Office Hours Queue
+        <i style="font-style:normal;" class="fa-question-circle"></i>
+      </a>
+    {% endif %}
     <br>
+    <br>
+  </p>
+  {% if viewer.getQueueAnnouncementMessage() | length > 0 %}
+    <h3>Todays Announcements:</h3>
```
marwanatef2, line 28: I think the announcement header should rather be **h1** to match "Today's History" for consistency purposes, the header along with the paragraph containing the announcement should rather be in a div with some css as a background indicating that this is the announcement as this is how it looks in a student page, it is not separated from what follows.

image

elihschiff: I added a background. I actually made the "today's announcement" a h2 not a h1 as well as switched the "today's history" also to h2. I am pretty sure a page is only supposed to have 1 h1 (in this case I have "Office Hours Queue"), so the "today's history" being a h1 is my bad to begin with.
image

migration/migrator/migrations/course/20200527001041_queue_whiteboard_message.py
```diff
@@ -0,0 +1,63 @@
...
+            j = json.load(in_file)
+
+        if 'queue_announcement_message' in j['course_details']:
+            del j['course_details']['queue_announcement_message']
+
+        with open(config_file, 'w') as out_file:
+            json.dump(j, out_file, indent=4)
+
+
+    database.execute("ALTER TABLE queue DROP COLUMN IF EXISTS paused;");
```
bmcutler, line 63: please review
https://submitty.org/developer/migrations#writing-new-migrations

we don't want to drop columns / lose data if we rollback on a production machine -- unless necessary

bmcutler:
@MasterOdin @elihschiff 

I don't understand the edits to the courses_table.sql file.

I used Matt's new sql dump command and it removed those edits.
Perhaps I performed and incorrect resolution to the merge.  
Or perhaps my vagrant database is invalid somehow.

The types of those queue fields don't match. 
And I don't know why the courses_table.sql file would have an "if exists".

migration/migrator/migrations/course/20200527001041_queue_whiteboard_message.py
```diff
@@ -0,0 +1,60 @@
...
+
+    course_dir = Path(config.submitty['submitty_data_dir'], 'courses', semester, course)
+    # add boolean to course config
+    config_file = Path(course_dir, 'config', 'config.json')
+    if config_file.is_file():
+        with open(config_file, 'r') as in_file:
+            j = json.load(in_file)
+
+        if 'queue_announcement_message' in j['course_details']:
+            del j['course_details']['queue_announcement_message']
```
bmcutler, line 57: Migration is unnecessarily destructive

PR #5617, opened 2020-07-02
[Feature:Ducks!] seasonal ducks svg

hertacalvo:
Added seasonal ducks in svg format (January-July) at /site/public/img
File names are ##-monthname.svg format.

This PR changes no code and changes no existing files so testing is not necessary.


hertacalvo:
Below are screenshots of each duck design (GitHub doesn't support svg here)

January
image

February
image

 March 
image

April 
image

May
image

June
image

July
image

mukul-kmr-jha:
Wow!

site/app/controllers/GlobalController.php
```diff
@@ -343,20 +343,75 @@ public function header() {
         }
 
         $now = getDate(date_timestamp_get($this->core->getDateTimeNow()));
-        $month = $now['mon'];
-        $day = $now['mday'];
+        $duck_img = $this->getDuckImage($now);
```
MasterOdin, line 346: Can we just pass `$this->core->getDateTimeNow()` to `$this->getDuckImage()` and then use `$day = (int) $now->format('j')` and `$month = (int) $now->format('n`)` inside of the function.

bmcutler: done

PR #5618, opened 2020-07-02
[Testing:Autograding] Properly validate notebook gradeables

emaicus:
This PR:
1. Updates the `complete_config_schema` to properly validate notebook gradeables and gradeables with item pools.
2. Adds schema integration tests for all existing notebook gradeables.
3. Updates `main_configure.cpp` to properly validate notebook objects within an `item_pool`.
4. Updates `load_config_json.cpp` to properly autocomplete/fill in defaults for testcases within an `item_pool`.
5. Updates `submitty_schema_validator.py` to print better error messages for errors which occur in a notebook or item pool.
6. Updates all schema integration tests to print out better errors on schema failure.

The above required a refactor of `load_config_json.cpp` to validate arbitrary arrays of testcases (previously, the script was hardcoded to validate only the global `testcases` array in a config.json). Similarly, it required a refactor of `main_configure.cpp` to validate arbitrary notebooks rather than just the global `notebook`.


tests/integrationTests/tests/c_failure_messages/__init__.py
```diff
@@ -37,8 +38,12 @@ def cleanup(test):
 def schema_validation(test):
     cleanup(test)
     config_path = os.path.join(test.testcase_path, 'assignment_config', 'complete_config.json')
-    test.validate_complete_config(config_path)
-   
+    try:
+        test.validate_complete_config(config_path)
+    except Exception:
+        traceback.print_exc()
+        raise
```
MasterOdin, line 45: I think we should modify this block:
https://github.com/Submitty/Submitty/blob/master/tests/integrationTests/lib.py#L118-L128

to print out the full exception stack trace (instead of just the error message) instead of having this pattern in every schema validation testcase.

python_submitty_utils/submitty_utils/submitty_schema_validator.py
```diff
@@ -13,12 +13,14 @@ class SubmittySchemaException(Exception):
 
     def __init__(self, config_json, schema, message, title, schema_error):
         """Use to aid in printing helpful information about schema errors."""
-        super(SubmittySchemaException, self).__init__(schema_error.message)
+        error_message = schema_error.message if schema_error is not None else ''
+
+        super(SubmittySchemaException, self).__init__(error_message)
```
MasterOdin, line 18: In python 3, you can do `super().__init__(error_message)`.

python_submitty_utils/submitty_utils/submitty_schema_validator.py
```diff
@@ -83,16 +70,32 @@ def complete_config_validator(j_, s_, warn=True):
...
+            for c in t['solution_containers']:
+                validate_schema(c, c_schema, prefix=t_name, warn=warn)
+        else:
+            raise SubmittySchemaException(
+                t,
+                None,
+                'Testcase is missing "solution_containers" field',
+                'Testcase is missing "solution_containers" field',
+                None
+            )
```
MasterOdin, line 96:
```suggestion
        for key in ['containers', 'solution_container']:
            if key in t:
                for c in t[key]:
                    validate_schema(c, c_schema, prefix=t_name, warn=warn)
            else:
                raise SubmittySchemaException(
                    t,
                    None,
                    f'Testcase is missing "{key}" field',
                    f'Testcase is missing "{key}" field',
                    None
                )
```

tests/integrationTests/tests/c_failure_messages/__init__.py
emaicus: Resolved per conversation at today's meeting.

python_submitty_utils/submitty_utils/submitty_schema_validator.py
emaicus: Updated in recent commit.

python_submitty_utils/submitty_utils/submitty_schema_validator.py
emaicus: Updated in recent commit.

emaicus:
In my most recent commit I: 
1. Resolved Issues identified in the above review.
2. Removed notebook gradeables from the `test_submitty_schema_validator.py` skip list and updated relevant complete configs.
3. Identified and fixed a bug which was causing `STDOUT.txt` and `STDERR.txt` not to be automatically archived for compilation testcases.

bmcutler:
@emaicus 
please resolve conflicts

grading/load_config_json.cpp
```diff
@@ -145,160 +197,125 @@ void ArchiveValidatedFiles(nlohmann::json &whole_config) {
...
+void AddDockerConfiguration(nlohmann::json &testcases, nlohmann::json &whole_config) {
 
   int testcase_num = 0;
-  for (typename nlohmann::json::iterator itr = tc->begin(); itr != tc->end(); itr++,testcase_num++)
-  {
-    std::string title = whole_config["testcases"][testcase_num].value("title","BAD_TITLE");
+  for (typename nlohmann::json::iterator itr = testcases.begin(); itr != testcases.end(); itr++,testcase_num++) {
+    std::string title = (*itr).value("title","BAD_TITLE");
 
-    nlohmann::json this_testcase = whole_config["testcases"][testcase_num];
```
bmcutler, line 188: why did you remove this local variable?  (I think it helps with code readability -- *itr is less clear in this lengthy function)

Do we need to switch it to a reference?

PR #5622, opened 2020-07-04
[UI/UX:InstructorUI] Notebook Builder 2.0

Kevin-MM:
### New Functionality
This PR adds the ability to save a `config.json` constructed by the notebook builder to the server.  A button has been added to the autograding configuration tab of the `Edit Gradeable` page.  The notebook builder should now be accessed via this button.

#### Currently when the `Save` button is pressed you can expect
1. The current state of the notebook builder is compiled into a valid `config.json`
1. The `config.json` is uploaded to the server and saved in a **new** configuration directory
1. The gradeable will be automatically updated to use the new configuration and a rebuild of the gradeable is triggered
1. A link to the submission page will be shown so that the person using the notebook builder can inspect their work

#### Things to be addressed in the *next* version of notebook builder
- The ability to load and update existing `config.json` will be added (instead of strictly create new ones like this PR has)
- Handling of default file names for input fields which ensures they are unique and don't collide with previously set ones


site/public/js/notebook_builder/form-options-widget.js
```diff
@@ -0,0 +1,72 @@
...
+        const file = new File([JSON.stringify(notebook_builder.getJSON(), null, 2)], 'config.json', {type: "text/plain"});
+        const g_id = window.location.pathname.split('/').pop();
+        const url = buildCourseUrl(['notebook_builder', 'save']);
+
+        const form_data = new FormData();
+        form_data.append('config_upload', file, 'config.json');
+        form_data.append('csrf_token', csrfToken);
+        form_data.append('g_id', g_id);
+
+        const makeRequest = async () => {
```
shailpatels, line 52: is the `makeRequest` function used more than once? If not probably don't need to put the request in a function.

site/public/js/notebook_builder/form-options-widget.js
```diff
@@ -0,0 +1,72 @@
...
+            const status_div = document.querySelector('.form-options-widget .status');
+            status_div.innerHTML = 'Saving...';
+
+            const response = await fetch(url, {method: 'POST', body: form_data});
```
shailpatels, line 56: Instead of wrapping the `await` key `fetch` is async by default like ajax so you can use promises to handle this type of logic. I think it'll look something like this
```js
fetch(url, {
  method: 'POST',
  body: form_data,
})
.then(
   response => {
        JSON.parse(response)
   }
)
.then(data => {
  if(data.status === 'success){ 
     //foo
  }else{

  }
})
.catch((error) => {
  console.error('Error:', error);
});
```

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -223,6 +223,7 @@ private function editPage(Gradeable $gradeable, $semester, $course, $nav_tab = 0
         $this->core->getOutput()->addInternalJs('admin-gradeable-updates.js');
         $this->core->getOutput()->addInternalCss('admin-gradeable.css');
         $this->core->getOutput()->renderTwigOutput('admin/admin_gradeable/AdminGradeableBase.twig', [
+            'core' => $this->core,
```
shailpatels, line 226: Is there a reason why 'core' is getting passed to this twig file? 
Core typically contains a lot of information and we want to try restricting its usage to controller classes and propagate only relevant information to views so we don't risk exposing more information to the user than intended

site/public/js/notebook_builder/form-options-widget.js
Kevin-MM: `await` only works inside of `async` functions.

site/public/js/notebook_builder/form-options-widget.js
Kevin-MM: `await`/`async` is syntactic sugar over promises.  Any `async` function will always return a promise.  I think using `await`/`async` makes the code much more readable than using all then chaining `.then()`.

site/app/controllers/admin/AdminGradeableController.php
Kevin-MM: I wanted to use the `buildCourseUrl()` function in the twig file.  I don't agree that we shouldn't pass core to the twig file.  I've seen many cases where the view is passing 20 parameters (all information collected from core) to the twig where it could just be passing the core a single time.

MasterOdin: What out of core do you want or need in twig files? Are the 20 parameters just different urls? I agree we should figure out how to slim down some of the huge parameter lists, but do not want to pass in core unless there's a tremendously good reason as I have a strong feeling it's going to lead to abuses of things like running queries in the templates like we've got queries inside the views.

Kevin-MM: I propose we create a white list of functions that would be useful inside the twig file, for example `buildCourseUrl()`, and maybe keep them in a `json` or `const` multi-dimensional array or something so they stay organized.  Then `Output::loadTwig()` can loop over the functions and register them as Twig functions the same way a few other functions are already being registered in there.

MasterOdin: Yes, but before bothering with that, I would appreciate if you could take a stab at answering:

> What out of core do you want or need in twig files? Are the 20 parameters just different urls?

Off the top of my head, I could see providing a twig function for `buildUrl` and `buildCourseUrl` and then a twig variable for csrf token. I am loathe to allow `getUser` or `getConfig` as I firmly believe we should pass models in to templates to minimize coupling.

Kevin-MM: At this time I can only think of `buildUrl` and `buildCourseUrl`.  Not sure about the csrf token as I think its getting loaded in somewhere.  It seems to always been available inside the browser even when I haven't explicitly defined it.

Kevin-MM: I removed core from the twig so we can move ahead with this PR.

site/public/js/notebook_builder/form-options-widget.js
shailpatels: Ok sounds good, I didn't realize `async` will always return a promise in that case I agree its more readable, maybe we can add this to the [coding style guide for JS](https://submitty.org/developer/coding_style_guide/javascript) @MasterOdin so we're consitent about using this.

personally I think we should start moving towards using `fetch` wherever we would use ajax in the past so having a consistent way to call it with promises would be nice.

Kevin-MM:
@bmcutler I've updated the multiple choice option widget as discussed.  I also added a link pointing to the edit gradeable page that gets shown after pressing the save button.  The browsers back button also works here.

@shailpatels Sorry about that, I changed the operator to a ternary which has much better browser support.

PR #5623, opened 2020-07-04
[Feature:Submission] Rework gradeable access log

### LINKED ISSUE #5282: Revise gradeable page load/access log. (2 notes)
The bug:
The date in the access log is incorrectly formatted, mm-dd-yyyy instead of the submitty standard yyyy-mm-dd.  But let's not fix that right now (end of term -- people have temporary scripts using this data) we'll update this early summer.

Requested revision for improved performance:
On each gradeable page load access for each student we:
1. append an entry to the global log file (this is ok, we'll move this to faster disk)
but we also:
2. append an entry to a personal log file in the students submission directory.  we believe this is a bottleneck, and can't easily be moved to faster disk.

So let's instead store all the load/accesses in a new database table for the course and write the complete access log file at each submission time.  So we'll still have all of the access information, but skip writing it to file until we are saving an entire submission.

MasterOdin:

Closes #5282

The original implementation to record access by a user to a gradeable page was done using a log file that was read completely into memory on each page load and then written back out with a new entry. The file was then copied around during the autograding process, and potentially used. The format of the timestamp in the logfile was not in the right format, using `MM-DD-YYYY` instead of `YYYY-MM-DD`. Additionally, during a period of high usage, reading and writing the file was causing an unfortunate system bottleneck. This file was saved one level above the submission version folder, and so required being copied in during the grading process.

Old format:
```json
{
    "page_load_history": [
        {
            "time": "07-04-2020 14:30:33-0400",
            "who": "instructor"
        },
        {
            "time": "07-04-2020 14:34:33-0400",
            "who": "instructor"
        }
    ]
}
```

### What is the new behavior?

We now write to the DB each access, and this is then written out to a file on submission. The timestamp of each entry is also now in the proper format. This file is saved now within the submission folder as a dotfile (`.user_assignment_access.json`), which removes the need for copying it in during the grading process.

The new format:
```
[
    {
        "id": 4,
        "g_id": "open_homework",
        "user_id": "instructor",
        "team_id": null,
        "accessor_id": "instructor",
        "timestamp": "2020-07-04 12:35:10-0400"
    },
    {
        "id": 5,
        "g_id": "open_homework",
        "user_id": "instructor",
        "team_id": null,
        "accessor_id": "instructor",
        "timestamp": "2020-07-04 12:35:28-0400"
    }
]
```

### Other information?

Currently I am writing out the entire gradeable_access table row to the file. I expect to get feedback on trimming it down, as well as naming of fields within the review, which will help revise things.

To Do:
- [ ] Decide if to include or not backwards compatibility with old format
- [x] Commit DB changes to course_schema.sql
- [ ] Update submission tests for checking new generated file


MasterOdin:
As somewhat hinted at in the OP, this patch as initially committed completely breaks backward compatibility, as it's not clear how much we want to retain it from the linked issue. It would be good to get feedback on how much we might want that, as well as if it would be better to hold off on this PR until semesters change or some other preferred option. There's also a discussion on how much backwards compatibility we want in the file structure itself as well.

Thoughts @bmcutler, @emaicus.

bmcutler:
@MasterOdin 
I've looked over the code and overall approve it.
I've tested this, and debugged the autograding for time limit notebooks.
So I'm ok to merge, but we are failing travis tests.

Can you please debug travis?  Thanks

emaicus:
Someone should review the migration for this and run the e2e tests locally.

PR #5631, opened 2020-07-06
[Bugfix:TAGrading] Late Submission Check

### LINKED ISSUE #4588: TA grading index page for bad="too many days late" assignment
**Describe the bug**

Currently, viewing an assignment submitted too late displays an all red background.  That's good. 

But from the main index page listing all students, it's still a blue "Grade" button.  It should be a red "Too Many Days Late" or bad status button.

Also, the data/stats page currently counts this against the % of grading done.
TAs should not have to grade things that are "too late".  (They can grade it and we record and show the student the autograding & manual grading scores, but rainbow grades will show a zero.)

Things that are too late, should not be included in the grading done.  The button for grading should not be red if all on time (or within late policy) submissions are graded.

jportorreal00:

Fixes #4588- When a user submits an assignment that is too late, the TA should have a red button that says "Too Many Late Days" instead of a blue "Grade" button. There should also an option to remove all late submissions from the "grading to be done" stat page.

### What is the new behavior?
These changes mean that a grader can know if an assignment was submitted after the deadline and/or too many late days were used. The grander can now see the stats for the number of graded assignments excluding late assignments. This option is saved with a cookie.

### Other information?
This was tested by making new gradeables and looking at late submissions. It could be toggled by changing the grading version if there is a late upload available. I also checked the option of excluding some data by grading late assignments.


josharsh:
@jportorreal00  Looks like lint test still fails and this time it can be automatically solved using *phpcbf* 
Please refer: 
https://submitty.org/developer/coding_style_guide/php

jportorreal00:
@josharsh I’ve ran phpcbf and phpcs on the file that is giving me trouble and it says “no fixable errors were found” every time. I’m not sure what’s going on

josharsh:
@jportorreal00 Did you run phpcbf with submitty's standard ruleset? Before running it did you perform 
~~~bash 
composer install
~~~
from site directory?

Please read this for reference: https://submitty.org/developer/coding_style_guide/php

jportorreal00:
@josharsh Yes, I followed this. I don’t get why it does this because the variable it is complaining about already existed before I started making any changes. I’ve compared it to the master branch and

josharsh:
@jportorreal00  i am in doubt too, looks strange, try merging master to it and hope it works. 
if not, we might need some help from  @emaicus and/or @jchulton :)

MasterOdin:
It's because this is not a phpcs error and so running it is not going to give you anything. It's a phpstan error, see https://submitty.org/developer/testing/linting_static_analysis#php-static-analysis for discussion on that.

For this particular issue, whenever you see an error of the format:
```
Ignored error pattern #/some patter/# in path                                                    
         file_path was not matched in reported errors.
```

is because you've fixed some error in our phpstan baseline file (https://github.com/Submitty/Submitty/blob/master/site/phpstan-baseline.neon) and can remove the related lines from it. For example, it looks like the particular error you might have fixed would be https://github.com/Submitty/Submitty/blob/master/site/phpstan-baseline.neon#L128-L131 and so you would remove those lines and re-run phpstan.

jportorreal00:
@MasterOdin that solved it! Once I removed the lines it runs with no errors

jchulton:
Got this when I tried to grade the Grading Homework gradeable in sample courses as an instructor
image

jportorreal00:
@jchulton is this error a time out?

site/app/models/gradeable/GradedGradeable.php
```diff
@@ -77,6 +77,22 @@ public function getGradeable() {
...
+     * @return bool
+     */
+    public function isOnTimeSubmission(User $user) {
+        if ($this->gradeable->isTeamAssignment()) {
+            $user = $this->getSubmitter()->getTeam()->getMemberUsersSorted()[0];
+        }
+        else {
+            $user = $this->getSubmitter()->getUser();
+        }
+        $late_day = new LateDays($this->gradeable->core, $user, [$this]);
```
shailpatels, line 91: Probably don't want to create a new LateDay here,
`$late_days = LateDays::fromUser($this->core, $this->core->getUser());`

would be a little better but still not idea since it needs to make a database query per student

shailpatels (changes requested):
The functions
`getBadGradedComponents`
`getBadSubmissionsByGradingSection`

should be refactored to use new database queries that can ideally select all users with bad late submissions at once or at least break part of that into smaller users.

Right now these functions are calling the new Late days constructor multiple times but this is causing multiple db queries per student which probably won't scale to a class with like 300 students

PR #5632, opened 2020-07-06
[Feature:System] Use the schema validator on gradeable build

emaicus:
This PR adds a new script `bin/config_syntax_check.py`, which compares a built `complete_config.json` with the `complete_config_schema.json`. This script now called during assignment build, and reports any schema found after assignment build to the instructor.

To test: build the tutorial assignment `15_GLFW`, which is currently failing schema validation (to be fixed in a future PR), and then build the tutorial assignment 01_simple_python, which is currently passing schema validation.

Issues to fix before this moves out of draft:
1. Modify the permissions on `bin/json_scemas/complete_config_schema.json` so that it is readable both by instructors and users relevant to the web assignment build process.

Open Issues after this PR is complete:
1. Fix schema validation errors across the handful of assignments where they exist (e.g. `15_GLFW`)
2. Modify the schema to validate instructor written (non-complete) configuration files.


bin/config_syntax_check.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
```
MasterOdin, line 1: What do people view as the long-term prospects of this script, by which I mean is it envisioned this code will be used by instructors at some point in the future, or just to be used within the build homework compilation process?

bin/build_homework_function.sh
```diff
@@ -176,6 +175,9 @@ function build_homework {
     # quit (don't continue on to build other homeworks) if there is a compile error
     make -j 8  > $hw_build_path/log_make_output.txt 2>&1
 
+    # run the schema validator on the complete_config generated by the build process
+    $SUBMITTY_INSTALL_DIR/bin/config_syntax_check.py $hw_source $assignment $semester $course
```
MasterOdin, line 179: On resolving the merge conflict, this line should (like the `make_generated_output.py` one) be placed underneath the if block that checks the make exit code?

PR #5639, opened 2020-07-07
[Feature:InstructorUI] Student Photos Upgrades

### LINKED ISSUE #5587: Student Photos Improvements
There is a bit of a checklist left over from #5539.  Some additional improvements we can make to preferred student photos include:

- Update the flag / unflag operations on the student photos page so that they use ajax or fetch, instead of the current form submission and full page reload.

- Allow full access graders to flag images. 

- Full access graders and above should have an option to select **view all students** or **view only their assigned registration sections**.  By default show only assigned sections, unless they have no sections assigned, then show all sections.

Kevin-MM:
Closes #5587 

### What is the new behavior?
- Full access graders and above now have the ability to flag and unflag a user's preferred photo.
- The flag / unflag operations now happens using a fetch call instead of a full page reload.
- Full access graders and above will see a button on the student photos page allowing them to select between viewing only their assigned registration sections, or all students. (Only visible when grader has at least one section assigned)


site/app/controllers/grading/ImagesController.php
```diff
@@ -177,22 +182,37 @@ public function ajaxUploadImagesFiles() {
...
+                'image_mime_type' => $display_image->getMimeType(),
+                'icon_html' => $icon_html,
+                'href' => $href
+            ]);
         }
 
-        $result = $this->core->getQueries()->updateUserDisplayImageState($_POST['user_id'], $new_state);
-        $result ? $this->core->addSuccessMessage($success_msg) : $this->core->addErrorMessage($fail_msg);
-        return new RedirectResponse($this->core->buildCourseUrl(['student_photos']));
+        return JsonResponse::getErrorResponse('There was an error flagging or unflagging the user\'s image');
```
mukul-kmr-jha, line 216: can we keep using $fail_msg to show the exact msg whether it was flagging or unflagging time error

site/app/templates/grading/Images.twig
```diff
@@ -1,6 +1,12 @@
 <div class="content">
-    {% if hasInstructorPermission %}
+    {% if has_full_access %}
         <div style="float: right; margin-bottom: 20px;">
+            {% set url = core.buildCourseUrl(['student_photos']) %}
```
mukul-kmr-jha, line 4: I think we should not use core in twig files and for this one, we can just use template values

site/app/controllers/grading/ImagesController.php
```diff
@@ -177,22 +182,37 @@ public function ajaxUploadImagesFiles() {
...
+        $result = $this->core->getQueries()->updateUserDisplayImageState($user_id, $new_state);
+        $user = $this->core->getQueries()->getSubmittyUser($user_id);
+
+        if ($result) {
+            $display_image = $user->getDisplayImage();
+
+            return JsonResponse::getSuccessResponse([
+                'first_last_username' => $user->getDisplayedFirstName() . ' ' . $user->getDisplayedLastName(),
+                'image_data' => $display_image->getImageBase64MaxDimension(ImagesView::IMG_MAX_DIMENSION),
+                'image_mime_type' => $display_image->getMimeType(),
```
mukul-kmr-jha, line 210: Add a conditional whether $display_image is null because when user doesn't have an image or when its flagged getDisplayImage returns null and its throwing PHP error (internally... I mean you dont see a normal error page with all the details as in other normal cases because of the JsonResponse type) when accessing methods on this null value for assigning values to $image_mime_type and $image_data assignments (L209-L210)

site/app/controllers/grading/ImagesController.php
Kevin-MM: Updated error message to indicate if the photo was being flagged vs unflagged.

site/app/templates/grading/Images.twig
Kevin-MM: Removed core.

site/app/controllers/grading/ImagesController.php
Kevin-MM: Added `$display_image` null check.

site/app/controllers/grading/ImagesController.php
```diff
@@ -177,22 +182,36 @@ public function ajaxUploadImagesFiles() {
...
+        $user = $this->core->getQueries()->getSubmittyUser($user_id);
+        $display_image = $user->getDisplayImage();
+
+        if ($result && $display_image) {
+            return JsonResponse::getSuccessResponse([
+                'first_last_username' => $user->getDisplayedFirstName() . ' ' . $user->getDisplayedLastName(),
+                'image_data' => $display_image->getImageBase64MaxDimension(ImagesView::IMG_MAX_DIMENSION),
+                'image_mime_type' => $display_image->getMimeType(),
+                'icon_html' => $icon_html,
+                'href' => $href
```
mukul-kmr-jha, line 211:
```suggestion
        if ($result) {
            return JsonResponse::getSuccessResponse([
                'first_last_username' => $user->getDisplayedFirstName() . ' ' . $user->getDisplayedLastName(),
                'image_data' => !is_null($display_image) ? $display_image->getImageBase64MaxDimension(ImagesView::IMG_MAX_DIMENSION) : "",
                'image_mime_type' => !is_null($display_image) ? $display_image->getMimeType() : "",
                'icon_html' => $icon_html,
                'href' => $href
```

site/public/js/server.js
```diff
@@ -1625,46 +1625,66 @@ function enableKeyToClick(){
...
+                const image_container = document.querySelector(`.${user_id}-image-container`);
+
+                // Change image
+                const old_image = image_container.querySelector('img');
+
+                const new_image = document.createElement('img');
+                new_image.setAttribute('alt', data.first_last_username);
+                new_image.setAttribute('src', `data:${data.image_mime_type};base64,${data.image_data}`);
+
+                image_container.removeChild(old_image);
```
mukul-kmr-jha, line 1673: Say you flag someone's photo and after that, you hit reload. Now you're given with the button to unflag that photo but when you do that you don't see changes in effect right away and need to refresh the page. because this user doesn't have an associated image and this line breaks the code.
But when the user has an image (it's not flagged) you can play with the report button and it shows the changes without reloading.

site/app/controllers/grading/ImagesController.php
Kevin-MM: This shouldn't be needed.  If `$display_image` is `null` then it will be coerced to `false` when used in the conditional, there by canceling the success response and returning an error response.

mukul-kmr-jha: Actually, cancelling the success response and returning an error response whenever you try to flag a photo requires you to hit page refresh to reflect those changes...

Kevin-MM: Changed

site/public/js/server.js
Kevin-MM: Changed

PR #5650, opened 2020-07-08
removed inline css and script

Saurabh200598:

The file include inline CSS and inline scripts.

### What is the new behavior?
The inline CSS and inline scripts have been removed from the file.


MasterOdin:
We do not want to just remove the code, but rather move in into its own JS and CSS file. In this particular page's case, I would probably leave the width's inline (since I see no point making a class with a single percentage width declaration), and move the inline JS to `public/js/stat_page.js` adding a `$this->core->getOutput()->addInternalJs('stat_page.js');` below this line https://github.com/Submitty/Submitty/blob/master/site/app/views/forum/ForumThreadView.php#L1138.

PR #5669, opened 2020-07-11
[Bugfix:Autograding] Clean up shipper/worker communication on autograding failure

emaicus:
Previously, when the worker failed to successfully autograde an assignment, it would return a results zip file which contained only the file `failure.txt`. Unfortunately, the shipper would then report an error, but then continue to wait forever for a good response.

### What is the new behavior?
This PR greatly cleans up communication between the shipper and worker. `failure.txt` is removed. In it's place, a new field, `autograding_status` is added to the "done" queue file. This field includes a `status` (`fail` or `success`), and a string `message`. In the case of a failure, the worker machine includes the associated stack trace in the `message`.

On the shipper side, the `unpack_job` function has been simplified and restructured to remove redundant code and to separate variable declarations from key logic. Previously, the `unpack_job` function returned a `success` boolean. Instead, the function now returns a member of an enumeration, either `success`, `failure`, or `waiting`. Upon receiving a response from the worker, the shipper now opens the done queue file to determine whether the job finished successfully or unsuccessfully and returns `success` or `failure` accordingly. If the shipper looks for a finished job but none exists, it returns `waiting`.

### TODO
At present, on worker failure, the associated results directory is abandoned on the worker machine. Furthermore, if it was properly zipped and sent to the worker, it would still likely not be properly archived. This is because `packer_unpacker.unpack_grading_results_zip` is not robust to missing files within the results directory. As future work, we should update the worker to rescue partial results, and `packer_unpacker` to properly store them.


sbin/submitty_autograding_worker.py
```diff
@@ -54,53 +53,47 @@ def worker_process(which_machine, address, which_untrusted, my_server):
...
+
+    # Set up key autograding_TODO directories
+    todo_dir = os.path.join(SUBMITTY_DATA_DIR, "autograding_TODO")
+    autograding_zip = os.path.join(todo_dir, f"{worker_name}_autograding.zip")
+    submission_zip = os.path.join(todo_dir, f"{worker_name}_submission.zip")
+    todo_queue_file = os.path.join(todo_dir, f"{worker_name}_queue.json")
+
+    # Establish the the directory in which we will do our work
+    working_directory = os.path.join(
+        "/var/local/submitty/autograding_tmp/",
```
MasterOdin, line 72:
```suggestion
        os.path.join(SUBMITTY_DATA_DIR, 'autograding_tmp')
```

sbin/submitty_autograding_shipper.py
```diff
@@ -488,51 +517,110 @@ def unpack_job(which_machine, which_untrusted, next_directory, next_to_grade, ra
...
+            return status
 
     try:
         with open(local_done_queue_file, 'r') as infile:
             local_done_queue_obj = json.load(infile)
+
         # Check to make certain that the job we received was the correct job
         if random_identifier != local_done_queue_obj['identifier']:
-            success = False
+            status = GradingStatus.WAITING
```
MasterOdin, line 534: I think this line can be wholly removed, since it looks like `GradingStatus.WAITING` is being explicitly returned on line 551.

sbin/submitty_autograding_shipper.py
```diff
@@ -488,51 +517,110 @@ def unpack_job(which_machine, which_untrusted, next_directory, next_to_grade, ra
...
             msg = f"{which_machine} returned a stale job (ids don't match). Discarding."
+            # Even though this job was not the one we are waiting for, report errors.
+            if local_done_queue_obj['autograding_status']['status'] == 'fail':
+                msg += ' discarded job failed. Check the stack traces log for details.'
+                autograding_utils.log_stack_trace(
+                    AUTOGRADING_STACKTRACE_PATH, job_id=JOB_ID,
+                    trace=f"ERROR: {worker_name} returned the following error for a stale job:\n"
+                          f"{local_done_queue_obj['autograding_status']['message']}"
+                )
+            print(msg)
```
MasterOdin, line 544: stray print?

sbin/submitty_autograding_shipper.py
```diff
@@ -411,25 +420,38 @@ def unpack_job(which_machine, which_untrusted, next_directory, next_to_grade, ra
...
         SUBMITTY_DATA_DIR, "autograding_DONE",
-        f"{servername_workername}_{which_untrusted}_queue.json"
+        f"{worker_name}_queue.json"
     )
 
+    # status will be set to a GradingStatus.
+    status = None
+
+    # If the worker we are retrieving from is local
     if which_machine == "localhost":
```
MasterOdin, line 451: This comment as is is not very helpful to someone new to this code, as the if statement already covers that we're retrieving a localhost worker (or not). It would be more helpful to comment here instead why the differentiation is made between local vs remote for the steps that follow (if the code itself does not make that clear enough).

sbin/submitty_autograding_shipper.py
emaicus: Good catch. Removed.

sbin/submitty_autograding_shipper.py
emaicus: Not enough prints, actually. We want the shipper to be a little bit chatty on stdout when it receives a completed job, in case it is being run manually (sometimes done for debugging). This one survived my refactor, but a few others were taken out. I added them back in a recent commit.

sbin/submitty_autograding_shipper.py
emaicus: Updated the comment.

Issue #2808, opened 2018-09-04
Followup on timezone issue


bmcutler:
Followup to #2805 

move the setTimezone commands into the Gradeable Model so that it's stored within the model.

Also, reproduce the timezone problem on the vagrant (if the setTimezone is removed) and write test cases so we'll immediately know if this breaks again.

Also, investigate assignment dates that span a daylight savings time change.

bmcutler:
This situation happened again.

We copied the databases to a new server, but did not copy the configuration file (postgresql.conf).  The new linux VM had timezone UTC, the new database server had timezone UTC.  The columns in the database that were missing explicit timezone (office hours queue) were thus interpreted inconsistently (since the main linux server and submitty timezone in /usr/local/submitty/config/submitty.json was America/New_York) and caused student/TA confusion.

Let's intentionally set the vagrant linux time to be America/New_York, and intentionally set the database timezone to be  Chatham Islands, New Zealand. UTC +13:45 (or something similarlly different) -- so we catch missing timezones and installation timezone mismatch in development.

PR #5673, opened 2020-07-13
[Bugfix:InstructorUI] Fixed timezone issues on office hours queue

### LINKED ISSUE #4702: Store timezones with timestamps for office hours queue
I missed this in my review of the initial implementation, sorry!

The table structure for queues does not contain the timezone information. This introduces a couple potential setups for bugs which include:

1. This means that the queue ordering will become wonky on the DST change such that it might mean people joining the queue at the end of the rollover will be sorted after people who join right after the rollover when clocks roll backwards, or that any statistics we might run like "time between queues" might have a missing or additional hour if running around the DST change.
1. If the DB and server have different timezones, it may introduce unexpected offsets as dates go into or out of the DB and into the server.

While I do not think our implementation and RPI usecase will necessary run afoul of this, I would still like to avoid the potential.

I would not worry about any potential wrong offsets that might occur on historic data (such as it is) when doing the migrate from `timestamp` to `timestamp with timezone`. I do not think anything further will have to happen, but should at least give a once over that everything looks like it's working appropriately.

elihschiff:

The office hours queue does not store timezones with the times
The times for the office hours queue comes from the sql database time

### What is the new behavior?

All office hours queue times now store a timezone as well
The times in the office hours queue not correctly get times based on the submitty config file so that the sql database time should not matter anymore 

Fixes office hours queue parts of #2808
Closes  #4702

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -168,7 +168,7 @@ class DatabaseQueries {
      */
     public function getSimilarNumericIdMatches(string $id_string): array {
         $this->course_db->query("
-            SELECT user_numeric_id from users where 
+            SELECT user_numeric_id from users where
             cast(user_numeric_id as text)
             like ?
         ", [$id_string]);
@@ -2535,7 +2535,7 @@ VALUES(?, ?, ?, ?, 0, 0, 0, 0, ?)",
     public function updateTeamRegistrationSection($team_id, $section) {
         $this->course_db->query("UPDATE gradeable_teams SET registration_section=? WHERE team_id=?", [$section, $team_id]);
     }
-    
+
     /**
      * Set team $team_id's anon_id
      *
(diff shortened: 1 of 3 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -168,7 +168,7 @@ public function getUserByIdOrNumericId($id) {
      */
```
MasterOdin, line 168: Do we need to run the formats in this file or can we just live with passing in the `\DateTime` object directly?

The places you use format:
1. `getPastQueue`
2. `getNumberAheadInQueueThisWeek`
3. `getNumberAheadInQueueToday`

elihschiff: The reason I use format sometimes in this file is because that is equivalent to `CURRENT_DATE` while passing in a date is equivalent to `current_timestamp`

`CURRENT_DATE` does not have any hours/min so it rounds to midnight which is what I want. I use that for cutoff to know if somebody has been in the queue today or not. If I used the current time it would include the hour/min which would mean that instead of checking if somebody has been in the queue today, it would instead check if they have been in the queue in the last 24 hours which is not what we want.

MasterOdin: Ah, alright. You'll probably want to do then `->format('Y-m-d 00:00:00O')` so that we're explicit here on the timezone of the server, and not letting postgres put an arbitrary value for it.

PR #5675, opened 2020-07-13
[UI/UX:InstructorUI] Notebook Builder 3.0

Kevin-MM:
#### Open New / Edit Existing
- Notebook builder can now be opened in one of two different modes.
- When opened in 'new' mode notebook builder will generate a new configuration, set the current gradeable to use that configuration, and rebuild automatically.
- When opened in 'edit' mode notebook builder will load the notebook section of the gradeable's config.json into the form for easy editing and resaving.  Resaving also fires off an automatic rebuild of the gradeable.  Care has been taken here to not modify any sections that are not part of the notebook.  For example if the config.json has already setup testcases, then that information **will** persist after editing, despite not being visible in notebook builder.
- 'Start New' and 'Edit Existing' buttons have been placed on the 'Edit Gradeable' submission tab to facilitate opening notebook builder in the correct state.  The 'Edit Existing' button is hidden when the gradeable is using any of the default configurations to prevent accidental editing of global level config.json files.  Checks are also in place to prevent editing a global level config.json by entering invalid paths in the url.
- This PR lays the groundwork for form validation when the save button is used.  Right now only filenames are being validated to check their uniqueness, but it will be fairly straightforward to add additional validation later.

#### Caveats
- Notebook builder currently only supports markdown, short_answer, or multiple_choice cells.  Editing and saving configurations which contain other types of cells will result in those other cells being lost.
- Notebook builder currently cannot load config.json where any `markdown_string` is formatted onto multiple lines.  An example of this would be entering (backslash + enter) when manually editing a config.json.
- Notebook builder **does** support loading new lines as long as they were entered like (backslash + n).  When using the html textarea found inside the markdown and multiple choice widgets, new lines don't require any special handling, just hit enter.
- Tab character usage is undefined.
- Notebook builder will remove any c-style comments from a config.json.  For editing/resaving a config.json where c-style comments are present this means those comments will be lost.  C-style comments inside of a `markdown_string` or multiple choice description will cause notebook builder to fail.


site/app/controllers/admin/AutogradingConfigController.php
```diff
@@ -232,14 +233,47 @@ public function configUsedBy($config_path = null) {
     }
 
     /**
-     * @Route("/courses/{_semester}/{_course}/notebook_builder/{g_id}", methods={"GET"})
+     * @Route("/courses/{_semester}/{_course}/notebook_builder/{g_id}/{mode}", methods={"GET"})
```
MasterOdin, line 236: We can add a requirement on the final parameter to enforce it has to be a specific value:

```suggestion
     * @Route("/courses/{_semester}/{_course}/notebook_builder/{g_id}/{mode<new|edit>}", methods={"GET"})
```

See https://symfony.com/doc/current/routing.html#parameters-validation for documentation on this.

site/app/controllers/admin/AutogradingConfigController.php
```diff
@@ -232,14 +233,47 @@ public function configUsedBy($config_path = null) {
...
      */
-    public function notebookBuilder(string $g_id) {
-        $gradeable = $this->core->getQueries()->getGradeableConfig($g_id);
+    public function notebookBuilder(string $g_id, string $mode) {
+        $gradeable = null;
+        try {
+            $gradeable = $this->core->getQueries()->getGradeableConfig($g_id);
+        }
+        catch (\Exception $exception) {
+            // Pass
```
MasterOdin, line 248: We should not be passing here, rather just explicitly make a redirect back to `$this->core->buildCourseUrl()` or something, no point going forward if we don't have a gradeable given it's required below.

site/app/controllers/admin/AutogradingConfigController.php
```diff
@@ -232,14 +233,47 @@ public function configUsedBy($config_path = null) {
...
+        }
+
+        // Redirect home if attempting to start notebook builder with invalid state
+        // Mostly only possible if attempting to access by manually setting invalid url parameters
+        $valid_mode = in_array($mode, ['new', 'edit']);
+        $invalid_state = $gradeable && $gradeable->isUsingDefaultConfig() && $mode === 'edit';
+        if (!$valid_mode || $invalid_state) {
+            $this->core->addErrorMessage('Invalid settings used when attempting to start Notebook Builder.');
+            return new RedirectResponse($this->core->buildUrl());
```
MasterOdin, line 257: Assuming the change on the check for gradeable above, and so we have a valid gradeable at this point, can redirect back to the gradeable edit page here with a more precise error message about the default configs.

site/app/controllers/admin/AutogradingConfigController.php
Kevin-MM: Committed suggestion.

site/app/controllers/admin/AutogradingConfigController.php
Kevin-MM: Changed

site/app/controllers/admin/AutogradingConfigController.php
Kevin-MM: Changed

PR #5681, opened 2020-07-15
[Feature:TAGrading] display notebooks in TAGrading panel

shailpatels:

Currently, instructors can see the submission of a student notebook by going through the files generated, for notebooks with lots of multiple choices and short answer, this results in a lot of files with little content.

### What is the new behavior?
The new electronic grading page now has a button for displaying a notebook panel which will show what a student's notebook gradeable looks like on their end. (If there is a submission or not)

### Other information?
This new panel only displays on the new grader interface

Differences between the grader view and student view: 

Instructor announcements panel is not displayed
Notebook warning are not displayed (if theres the possibility a student can receive the same question twice)
Files uploaded to dropzones are not displayed

<s>**Current bug** displaying the panel only works in full-screen mode - not split (i'll fix this tomorrow)</s>

image

image

image


site/public/js/drag-and-drop.js
```diff
@@ -1223,3 +1223,13 @@ function handleEditCourseMaterials(csrf_token, hide_from_students, requested_pat
...
+
+
+function initializeDropZone(id){
+    var dropzone = document.getElementById(id);
+    dropzone.addEventListener("click", clicked_on_box, false);
+    dropzone.addEventListener("dragenter", draghandle, false);
+    dropzone.addEventListener("dragover", draghandle, false);
+    dropzone.addEventListener("dragleave", draghandle, false);
+    dropzone.addEventListener("drop", drop, false);
+}
```
Kevin-MM, line 1235: Please add a new line to the end of the file.

shailpatels:
@Kevin-MM yeah I still have to fix the split view setting forgot about that - my bad

As for the images yeah the images are rendered the same way on notebooks so most likely reloaded until the browser caches them. 

Not sure what the alternative would be?

Kevin-MM:
I wasn't able to test navigating between students because I only had one submission.  I think it all really depends.  If we can count on the images being cached by the browser then I think we don't really need to do anything in this situation.

elihschiff:
Sorry for the second comment. 2 more things:

First off it seems that the instructor is able to click and edit the answers from the student (the save button is missing so they cannot save their answers), I think it would make more sense to disable all input as it is confusing how it is right now.

Also I was not able to view the student's answers in the new panel. I tried a bunch of things but nothing seemed to show their answers in the new notebook panel, everything was left blank/the default value.

Peek 2020-07-29 01-06

shailpatels:
@bmcutler @elihschiff Yeah there was an issue where it would load the data from the current user logged in instead of the student assigned to that notebook, this has been fixed so now it'll load item pools and the student answers correctly. Fixed a bug where files submitted were not appearing as well. 

All inputs for a notebook are disabled now

There is a current bug about the notebook icon on the gradeable panel disappearing I messaged @mukul-kmr-jha about to see if he can quickly understand what's going on since I couldn't understand why it's buggy.

The codeboxes are also not rendering correctly but I think this is a separate issue with rendering them on the grading side. As a work arround you can click on the codeboxes to see submitted code but the text is messed up and highlighted incorrectly. This will take a bit longer to debug i think for now this PR should probably get merged first.

Kevin-MM:
> The codeboxes are also not rendering correctly but I think this is a separate issue with rendering them on the grading side. As a work arround you can click on the codeboxes to see submitted code but the text is messed up and highlighted incorrectly. This will take a bit longer to debug i think for now this PR should probably get merged first.

@shailpatels There is a lot of additional js/css files that need to be loaded for the codeboxes to work correctly.  I'm currently working on a centralized code mirror utilities class that will make it a lot more streamlined to load and use the code mirror boxes.  My vote would be, don't worry about the codeboxes for right now. After I get my initial PR in there will be a few spots I need to go back to and rollover other codemirror code to use the improved loading utility.  This would be one of them.

shailpatels:
@Kevin-MM ah ok that makes sense,  yeah I'll hold off on getting them to render right I'll see if the refresh trick can at least make them visible a little later today.

bmcutler:
@shailpatels Status?  Looking forward to merging this :)

shailpatels:
@bmcutler I have to implement Kevin's codemirror utils, to fix code box rendering, should be ready soon

shailpatels:
@mukul-kmr-jha my bad, when you were talking about randomized order based on a hash I thought you meant randomized item pool notebooks which was fixed.

The random order question boxes were based on the current user's id, not the submitter id which I just pushed a fix for they should match now

PR #5685, opened 2020-07-16
[Refactor:Autograding] Add config class to autograding shipper and worker

emaicus:
The autograding shipper and worker scripts, as well as many of the scripts under `autograder/autograder`, read in submitty configuration files at the global scope, making it impossible to import them if they are in any environment other than their installed location. This makes writing tests for the scripts very difficult.

### What is the new behavior?
This PR add a `Config` class to the `autograder` module. This class has two constructors, one of which takes the path to the submitty config directory, and the other of which takes dictionaries holding the expected contents of relevant config files. Furthermore, most of the functions in the autograder package now take a `Config` as their first argument, allowing most of the global variables to be removed.


bmcutler:
@emaicus please resolve conflicts..   & test :)

hdrodz97:
Managed to run this after #5686 was merged, graded the sample courses against it with both local and remote workers, and everything completed without a hitch.

PR #5695, opened 2020-07-20
[Bugfix:TAGrading] remove redundant peer database fields

jchulton:

Extraneous database variables eg_peer_grade_set and eg_peer_grading in the electronic_gradeable table may be causing problems 

### What is the new behavior?
The variables eg_peer_grade_set and eg_peer_grading have been removed.


jchulton:
Fixed the frog robot and altered the migration such that the eg_is_peer_grading column will be properly set on migration down. I can't figure out how to set the eg_peer_grade_set column. @josharsh is it possible to determine what the eg_peer_grade_set column would be using other values in the database?

bmcutler:
@jchulton please resolve conflict

josharsh:
Tested the functionality:
1 - Existing gradeables on the courses seem to work fine (Tested with Non Peer, Peer and Peer Assignments)
2 - New Gradeable Creation works fine (Both Peer and Non Peer)
3 - Grader Set (When Peer grading is used) works fine for both student assignments and calculating average from peer and instructor score. 
4 - Creation of new Team Gradeable and peer assignments seems to work fine.
(However I am not sure on eg_peer_grade_set, given that it is an integer, I might possibly have the number of peer graders [But I am not sure if we have ever used it in code] )
@jchulton

josharsh:
As for setting eg_peer_grade_set, @jchulton does it look okay to set it to 0 (Which is default) on migration down?

PR #5704, opened 2020-07-21
[Bugfix:TAGrading] precision of student display of marks/components

jchulton:

If a student is given a mark with a decimal point (3.5) the mark is displayed with a point value of 4 when the student views their grade. The component score is still calculated with the decimal point mark

### What is the new behavior?
If a student is given a mark wit ha decimal point, the mark is correctly displayed with the decimal point, as below

image


bmcutler:
After offline discussion, let's switch the specification of precision to be a radio button or drop down (rather than an un-constrained input text box).

The 9 hardcoded choices:  10, 5, 1, 0.5 (the default), 0.33, 0.25, 0.1, 0.05, 0.01

And thus, all marks & components will be displayed with with a consistent number of digits beyond the decimal:
hardcoded values corresponding to the above:  0,0,0,1,2,2,1,2,2

PR #5706, opened 2020-07-22
[Bugfix:System] grading_done.py patch -- required_capabilities is optional

bmcutler:
grading_done.py patch -- required_capabilities is optional

### DIFF
--- a/bin/grading_done.py
+++ b/bin/grading_done.py
@@ -266,7 +266,9 @@ def update_queue_counts(queue_or_grading_file,is_grading,epoch_time,queue_counts
         else:
             queue_counts["interactive"] += 1
 
-    capability = entry.queue_obj["required_capabilities"]
+    capability = "default"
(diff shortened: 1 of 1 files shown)


bin/grading_done.py
```diff
@@ -266,7 +266,9 @@ def update_queue_counts(queue_or_grading_file,is_grading,epoch_time,queue_counts
         else:
             queue_counts["interactive"] += 1
 
-    capability = entry.queue_obj["required_capabilities"]
+    capability = "default"
+    if "required_capabilities" in entry.queue_obj:
+        capability = entry.queue_obj["required_capabilities"]
```
MasterOdin, line 271:
```suggestion
    capability = entry.queue_obj.get('required_capabilities', 'default')
```

bmcutler: thanks, I'll keep thank in mind for future.

PR #5709, opened 2020-07-23
[UI/UX:InstructorUI] Notebook Builder 4.0

Kevin-MM:
### Testing information
- Create a new gradeable or edit an existing gradeable
- Navigate to the second tab which is the 'Submissions/Autograding' tab
- Launch notebook builder using the included buttons

### Image Widget
- A new image widget has been added which allows instructors to upload/include images in their notebook gradeable

### Improved Widget Reordering
- Widgets are now 'rendered' a single time.  When reordering occurs the DOM element is preserved in memory and just reattached to the tree at a different point.  This makes it easier to preserve the state of the widget during re-ordering.

### New Notebook Builder Controller
- All the notebook builder functions which used to reside in the `AutogradingConfigController` have been moved to their own `NotebookBuilderController`.


bmcutler:
Thanks @listeph for your feedback, good suggestions!
@Kevin-MM will be incorporating many of those revisions in future PRs.

PR #5713, opened 2020-07-23
[Bugfix:VCS] Fix auto repo generation for arbitrary name

### LINKED ISSUE #5702: VCS gradeable with instructor specified repository name does not work
A gradeable configured like this:

Screen Shot 2020-07-21 at 3 50 50 PM

Gives an error when we try to clone:

> git clone http://localhost:1501/git/f20/blank/THING/student TMP_THING
Cloning into 'TMP_THING'...
Username for 'http://localhost:1501': student	
Password for 'http://student@localhost:1501': 
fatal: repository 'http://localhost:1501/git/f20/blank/THING/student/' not found

But when we use the gradeable id instead of THING it works.  The instructor specification is being ignored when repos are created.

MasterOdin:

Fixes #5702.

Previously, the code would only work for automatically generating gradeables if the name was the gradeable_id. Even if an instructor were to specify their own name for the folder, it would still just create folders using the gradeable_id. Additionally, repos were always being created, even if the gradeable specified it was using GitHub repos.

### What is the new behavior?

This fixes it so that it will now properly use the gradeable_id for the first option, and whatever the instructor types in for the second option. It will also no longer create repos if the GitHub options are selected.

### Other information?

Tested both options, and that they created the repositories as expected.

### DIFF
--- a/bin/generate_repos.py
+++ b/bin/generate_repos.py
@@ -43,6 +43,7 @@ def create_folder(folder):
 
 
 parser = argparse.ArgumentParser(description="Generate git repositories for a specific course and homework")
+parser.add_argument("--non-interactive", action='store_true', default=False)
 parser.add_argument("semester", help="semester")
 parser.add_argument("course", help="course code")
 parser.add_argument("repo_name", help="repository name")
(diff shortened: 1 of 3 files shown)


bmcutler:
Do you mean to link issue #5702?

MasterOdin:
Yes, thanks for pointing that out.

PR #5719, opened 2020-07-24
[Feature:Notebook] item_pool map student to problem

### LINKED ISSUE #5647: Option to specify assignments of notebook items from item_pool by username
Currently, the assignment of items from the item_pool is done by hash(username + other stuff) % number of items.
It might be desirable for the instructor to explicitly specify the assignment of items of some or all students.

Let's add an optional field...  something like:

"mapping" : { "aphacker" : 2, "bitdiddle" : 0, "student" : 1, "instructor" : 0 }

If the username is not included in the mapping, we do the hash.
The value specified must be an integer in the range zero to number of items - 1.

This will also allow simplify instructor testing of the notebook gradeable & associated autograding, allowing them to explicitly select the item from a pool for themself or other users.

tagging @Kevin-MM and @holzbh

mukul-kmr-jha:

Currently, if there are multiple items that are available in `from_pool`, hashing is done to pick one of these questions.
Also, there is an unhandled `undefined key error` case ( for 'type' ) which is popping out on some gradeable.

### What is the new behavior?
Closes #5647 
Fixes undefined index error. 
Adds support for optional field `user_item_map`.  This object contains a mapping of usernames to the index of items present in the `from_pool` array. 
```
          "user_item_map": {
                "<username_0> : <index_0>,
                "<username_1> : <index_1>....
              },
```
Users in this mapping will only be shown item corresponding to the mentioned index next to their username and no hash-based randomization will be done for them.
These ensure instructors can proofread all the question before exam... see #5647 for more info

### Other information?
Added this key in item_pool_random gradeable config.json for testing :)

### DIFF
--- a/grading/main_configure.cpp
+++ b/grading/main_configure.cpp
@@ -224,7 +224,7 @@ nlohmann::json validate_notebook(const nlohmann::json& notebook, const nlohmann:
         // Check if the indices mapped to users are in 'from_pool' array range
         for (auto i=user_item_map.begin(); i!=user_item_map.end(); i++) {
            int item_index = i.value();
-           if(item_index < 0 || item_index > in_notebook_cell_from_pool.size()) {
+           if(item_index < 0 || item_index >= in_notebook_cell_from_pool.size()) {
                std::cout << "ERROR: user (" << i.key() <<") mapped with index \"" << item_index;
                std::cout << "\" which is out of 'from_pool' array range: 0 to " << user_item_map.size() << std::endl;
                throw -1;
--- a/grading/main_configure.cpp
+++ b/grading/main_configure.cpp
@@ -221,6 +221,15 @@ nlohmann::json validate_notebook(const nlohmann::json& notebook, const nlohmann:
                       << (item_label.empty() ? "[no label]" : item_label) << std::endl;
           }*/
         }
+        // Check if the indices mapped to users are in 'from_pool' array range
+        for (auto i=user_item_map.begin(); i!=user_item_map.end(); i++) {
+           int item_index = i.value();
+           if(item_index < 0 || item_index > in_notebook_cell_from_pool.size()) {
+               std::cout << "ERROR: user (" << i.key() <<") mapped with index \"" << item_index;
+               std::cout << "\" which is out of 'from_pool' array range: 0 to " << user_item_map.size() << std::endl;
+               throw -1;
+           }
+        }
 
         // Write the empty string if no label provided, otherwise pass forward item_label
         out_notebook_cell["item_label"] = item_label;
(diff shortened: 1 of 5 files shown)


grading/main_configure.cpp
```diff
@@ -220,9 +221,22 @@ nlohmann::json validate_notebook(const nlohmann::json& notebook, const nlohmann:
                       << (item_label.empty() ? "[no label]" : item_label) << std::endl;
           }*/
         }
+        // Check if the indices mapped to users are in 'from_pool' array range
+        for (auto i=user_item_map.begin(); i!=user_item_map.end(); i++) {
+           int item_index = i.value();
+           if(item_index < 0 || item_index > in_notebook_cell_from_pool.size()) {
```
shailpatels, line 227: Is notebook indexing based on 0 indexing, I think `item_index` should be >= `in_notebook_cell_from_pool.size()` in this case since I believe the rest of system assumes 0 is the first index in a notebook pool

site/app/models/notebook/UserSpecificNotebook.php
```diff
@@ -109,7 +109,19 @@ private function replaceNotebookItemsWithQuestions(array $raw_notebook): array {
      */
     private function getItemFromPool(array $item): string {
         $item_label = $item['item_label'];
-        $selected = $this->getNotebookHash($item_label, count($item['from_pool']));
+        $selected = null;
+        // Check if user-mapping is available or not
+        if (
```
shailpatels, line 114: Instead of this check you can do
` $selected = $item['user_item_map'][$this->user_id] ?? null`;

Since a similar check is performed in the C++ side and if the value does somehow become invalud/out of bounds you can fall back to using the hash by setting it null

grading/main_configure.cpp
mukul-kmr-jha: Yeah, I missed `=` in there, thanks!

site/app/models/notebook/UserSpecificNotebook.php
mukul-kmr-jha: are you suggesting to remove this conditional checking whether the user index is set and value lies in the valid range?

shailpatels: yeah since the null coalescing operater will check if the value user_item_map and this->user_id is set and not null so you can avoid the check, if the value is out of range and note there it will default to null and then fallback to the normal hashing method. I think its a cleaner way than the large if statement

mukul-kmr-jha: Yeah, make sense since I'm already performing the rane validation check in CPP code, thanks:)

Issue #5710, opened 2020-07-23
Handle broken gradeables more gracefully


emaicus:
If gradeable creation fails, it currently makes it impossible to access the gradeables index (first page within a course). Ideally, we would handle this more gracefully. 

Here is an example of such an error:
```
Server Error
Oh no! Something irrecoverable has happened...

app\exceptions\DatabaseException (Code: 0) thrown in /usr/local/submitty/site/app/libraries/database/DatabaseQueries.php (Line 4183) by:
                throw new DatabaseException("Electronic gradeable didn't have an entry in the electronic_gradeable table!");


Message:
Electronic gradeable didn't have an entry in the electronic_gradeable table!

Stack Trace:
#0 unknown file(unknown line): app\libraries\database\DatabaseQueries->app\libraries\database\{closure}(Array)
#1 /usr/local/submitty/site/app/libraries/database/DatabaseRowIterator.php(56): call_user_func(Closure, Array)
#2 /usr/local/submitty/site/app/models/gradeable/GradeableList.php(72): app\libraries\database\DatabaseRowIterator->next()
#3 /usr/local/submitty/site/app/controllers/NavigationController.php(29): app\models\gradeable\GradeableList->__construct(app\libraries\Core)
#4 unknown file(unknown line): app\controllers\NavigationController->navigationPage()
#5 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(217): call_user_func_array(Array, Array)
#6 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(195): app\libraries\routers\WebRouter->run()
#7 /usr/local/submitty/site/public/index.php(127): app\libraries\routers\WebRouter::getWebResponse(Symfony\Component\HttpFoundation\Request, app\libraries\Core)
                  ,--.    ,--.
                 (( O))--(( O))
               ,'_`--'____`--'_`.
              _:  ____________  :_      _____
             | | ||::::::::::|| | |        \ \
             | | ||::::::::::|| | |         \
             | | ||::::::::::|| | |
             |_| |/__________\| |_|
               |________________|
            __..-'            `-..__
         .-| : .----------------. : |-.
       ,\ || | |\______________/| | || /.
      /`.\:| | ||  __  __  __  || | |;/,'\
     :`-._\;.| || '--''--''--' || |,:/_.-':
     |    :  | || .----------. || |  :    |
     |    |  | || '----SSt---' || |  |    |
     |    |  | ||   _   _   _  || |  |    |
[...]

PR #5721, opened 2020-07-25
[Bugfix:Submission] Better handle broken gradeable

Kevin-MM:
Partially #5710

The information needed to instantiate certain gradeable objects is stored across many database tables.  In the event certain information is missing, a single broken gradeable will throw an exception on the 'Gradeables' tab and result in a condition where users may not easily navigate to other parts of submitty that are not broken.

### What is the new behavior?
If a broken gradeable is detected, instead of crashing the entire 'Gradeables' page the main content area now shows an error message, and navigation via the left hand menu remains intact.

broken_gradeable


site/app/templates/error/genericError.twig
```diff
@@ -0,0 +1,5 @@
+<div class="content">
+    {% for msg in error_messages %}
```
MasterOdin, line 2: I think we should just make this page show a singular error message until such a time as we actually need to show multiple error messages, as I'm not sure we'll ever hit that point.

site/app/templates/error/genericError.twig
```diff
@@ -0,0 +1,5 @@
+<div class="content">
```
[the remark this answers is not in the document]
Kevin-MM: Changed.

site/app/templates/error/genericError.twig
Kevin-MM: I'm trying to promote flexibility and re-usability.  You'd prefer I change to a less flexible, less re-usable design?  You don't have to pass discrete error messages, one could instead pass multiple sentences they want formatted onto separate lines instead of one single paragraph.

MasterOdin: And my argument is that I'm not necessarily sure we'll ever use that feature, so why have it. Designing things for maximum flexibility and reusability "just in case" is as much an anti-pattern as not having any flexibility / reusability, if that "just in case" never happens.

But this is not a hill I will die on.

site/app/controllers/NavigationController.php
```diff
@@ -26,7 +27,13 @@ public function noAccess() {
      * @Route("/courses/{_semester}/{_course}", requirements={"_semester": "^(?!api)[^\/]+", "_course": "[^\/]+"})
      */
     public function navigationPage() {
-        $gradeables_list = new GradeableList($this->core);
+        try {
+            $gradeables_list = new GradeableList($this->core);
+        }
+        catch (DatabaseException $e) {
```
MasterOdin, line 33: Please add a call of `ExceptionHandler::handleException($e);` here so that sysadmins have a meaningful log output to act upon.

Kevin-MM: Changed.

PR #5723, opened 2020-07-26
[BugFix: TAGrading] Revise custom Peer marks

### LINKED ISSUE #5328: Revise marks & custom marks
A number of grading situations suggest we should redesign custom & common marks for the TA grading rubric.

* In some cases, it is desirable to prevent use of custom mark.  In a large class, to facilitate adjustment of component and mark point values after grading, it is essential that all scoring be global.   Design proposal:  add a new setting in the rubric configuration "allow use of custom mark". 

* It would be useful to view all custom marks previously used for grading that component (a grader might want to reuse a mark for another student that seemed like a one-time use situation the first time it was written). See issue #1998.  Design proposal: all custom marks are just regular marks, but we 'hide' some marks from the default TA view of that component.  Maybe we have "common" marks and "infrequent" marks.  All current custom marks are converted to "infrequent".  And there's a button on the TA interface to "view all infrequent marks" that is false by default.  Bonus:  A student can receive multiple infrequent marks.

* With peer grading, peers should not be allowed to write custom text in the component nor assign custom points.  Peers should not be able to add or edit common marks.  The full rubric should be set in advance.  Peers can leave an overall message for the gradeable, but not write custom text message per component.  This is not implemented yet!

* For random selection of problems from a notebook itempool, if the itempool problem is manually graded by TAs, there will usually be different rubric marks that apply to the different problem selections.  Design Proposal:  We discussed and decided we should probably have a single component for the item pool, and have a subset of marks for each possible problem.  When the TA grades, we automatically hide the marks that belong to the problems that were not selected for that user.

@emaicus @holzbh @shailpatels @jchulton

josharsh:
closes #5328 

With peer grading, peers can write custom text in the component nor assign custom points. Peers should not be able to add or edit common marks.
### What is the new behavior?
With peer grading, peers should not be allowed to write custom text in the component nor assign custom points. Peers should not be able to add or edit common marks.

### DIFF
--- a/site/public/templates/grading/GradingComponent.twig
+++ b/site/public/templates/grading/GradingComponent.twig
@@ -56,7 +56,7 @@ Required inputs:
 {% block extra_mark_rows %}
     {# Custom mark (only display in non-edit mode #}
     {% set f_mark = component.marks|first %}
-    {% if not edit_marks_enabled %}
+    {% if not edit_marks_enabled core.getUser().getGroup() < 4%}
         <div class="custom-mark-container mark-container row"
              data-mark_id="0">
             <span class="mark-selector-container">
--- a/site/public/templates/grading/GradingComponent.twig
+++ b/site/public/templates/grading/GradingComponent.twig
@@ -56,7 +56,7 @@ Required inputs:
 {% block extra_mark_rows %}
     {# Custom mark (only display in non-edit mode #}
     {% set f_mark = component.marks|first %}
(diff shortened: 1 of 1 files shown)


jchulton:
When a peer tries to grade a component, they get the following alert. This should not appear:
image

Also this doesn't fully close #5328. #5328 Includes other features, including the ability to restrict custom marks from the rubric tab on edit/create gradeable.

PR #5727, opened 2020-07-27
[Bugfix:Autograding] Add grading_parameter props to schema

bmcutler:
update the complete config schema validation


MasterOdin (commented):
Searching the codebase for a usage of a penalty points field showed the following two:

https://github.com/Submitty/Submitty/blob/4ec89c8a84bce09a2c0978beefc92baba0301206/grading/main_validator.cpp#L459

https://github.com/Submitty/Submitty/blob/3cbbb24c335550900b1c0375969b87041f36a51b/more_autograding_examples/cpp_hidden_tests/config/config.json#L2-L6

It does not look like this field is in our schema, and so is this a new field that's for a planned feature or aiming to cover the above?

PR #5736, opened 2020-07-28
[Refactor:UI/UX] refactor sidebar construction

bmcutler:
helper function & reduce code duplication


tests/e2e/base_testcase.py
```diff
@@ -13,9 +13,14 @@
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.common.by import By
 
+import sys
+# explicitly add this import path, so we can run it on a local host
+sys.path.append('../python_submitty_utils/')
```
bmcutler, line 18: @MasterOdin 

I added this path explicitly so that the end-to-end tests would run on my host

Is my host incorrectly configured, or is there a better way of solving this?
Will this break Travis?  (we'll see soon)

MasterOdin: go into `python_submitty_utils` and do `python3 -m pip install -e .` would be my suggested mechanism of doing this.

bmcutler: yuck, really??
This should be documented in our instructions then:
https://submitty.org/developer/testing/site_end_to_end_tests

(I don't believe I had to do that when I last ran these tests successfully...)

MasterOdin: I would ideally like to see the `python_submitty_utils` split off into its own repo and then made installable via package manager similar to how pdf-annotate.js, submitty-php-codesniffer, etc. are. It would simplify (to me) a number of these sorts of inter-dependencies of test suites for other things on that being installed (or inserting this hacky `sys.path.append` line).

PR #5737, opened 2020-07-28
[Bugfix:TAGrading] consistent precision of marks/components

jchulton:

The instructor manually sets the precision of a gradeable on the rubric tab. There are several bugs regarding precision
1. No matter what the precision is, the lowest possible precision for the total score of a component is 0.1. That means if the instructor enters a precision of 0.01, the gradeable max will round to the nearest 0.1
2. Sometimes, a mark score will be dislpayed incorrectly (1.910000001 instead of 1.91)
image

3. The student displays decimals with trailing zeroes (0.10) while they are not displayed in instructor and grader views (0.1)
### What is the new behavior?
The instructor will choose a precision from a dropdown list. The bugs above have been fixed


jchulton:
I've fixed the rubric cancel button and removed 0.33. When editting the point value, the number is compressed to fit the box (so if you enter something large like 1000000.01 you have to scroll left to right within the box). You should be able to see the entire value if you scroll.

jchulton:
I've added more space to the mark points decimal area for both the edit rubric and grading rubric. I've also fixed the overflowing decimals

jchulton:
Added the widening of the decimal space as above

PR #5742, opened 2020-07-29
[Bugfix:System] no course directory on initial install

bmcutler:
PR #5735 exists the migrator on error.
However, the first time we install the system (e.g., a fresh vagrant up)
we don't yet have the /var/local/submitty/courses directory....

@MasterOdin, perhaps you would like to propose an alternate patch.


MasterOdin:
See #5743 for an alternative patch which I would prefer, as otherwise we should just remove the print altogether as it's a meaningless warning to me.

PR #5752, opened 2020-07-30
[Bugfix:TAGrading] disable peer custom marks

josharsh:
works on #5328 

Peer Grades are allowed to use edit rubric and custom marks.
### What is the new behavior?
Peer Grades are restricted from using custom marks while peer grading.


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -756,6 +756,14 @@ function () use ($csv_file) {
         $this->core->redirect($return_url);
     }
 
+    /**
+     * Returns Custom marks access boolean
+     * @Route("/courses/{_semester}/{_course}/gradeable/{gradeable_id}/getCustomMarksAccess", methods={"GET"})
+     */
+    public function getCustomMarksAccess($gradeable_id) {
```
jchulton, line 763: Currently silent_edit permissions is defined as instructor only in Access.php. This makes sense, because only instructors should be able to edit a rubric. However, full and limited access graders should be able to leave custom marks. I believe this should be changed to say if the current user's group < GROUP::STUDENT, return true, else return false

josharsh: @jchulton Made the changes for it.

jchulton:
Getting the following error when I try to open a component for grading logged in as an instructor:
image

josharsh:
@jchulton It was due a route returning true than JsonResponse::getSuccessResponse(true);
Sorry, it skipped my mind. I checked, it seems to work fine now.

jchulton:
Minor thing: the grading rubric custom mark contains the word 'true'  
image

site/public/js/gradeable.js
```diff
@@ -196,15 +196,33 @@ function renderPeerGradeable(grader_id, gradeable, graded_gradeable, grading_dis
...
+            context: document.body,
+            global: false,
+            async:false,
+            success: function(res) {
+                console.log(res);
+                return res.data;
+            }
+        }).responseText;
+    var custom_mark_obj=JSON.parse(custom_mark_req);
+    var custom_mark_enabled = custom_mark_obj.data;
```
emaicus, line 214: Is there a reason you are running this with `async:false` instead of using promises?

site/public/js/ta-grading-rubric.js
```diff
@@ -72,7 +72,21 @@ AJAX_USE_ASYNC = true;
...
+                type: 'GET',       
+                url: buildCourseUrl(['gradeable',gradeable_id, 'getCustomMarksAccess']),
+                dataType: 'html',
+                context: document.body,
+                global: false,
+                async:false,
+                success: function(res) {
+                    console.log(res);
+                    return res.data;
+                }
```
emaicus, line 86: Same

site/public/js/ta-grading-rubric.js
```diff
@@ -72,7 +72,21 @@ AJAX_USE_ASYNC = true;
...
+let gradeable_id = getGradeableId();
+var custom_mark_req = $.ajax({	        
+                type: 'GET',       
+                url: buildCourseUrl(['gradeable',gradeable_id, 'getCustomMarksAccess']),
+                dataType: 'html',
+                context: document.body,
+                global: false,
+                async:false,
+                success: function(res) {
+                    console.log(res);
```
emaicus, line 84: stray print?

site/public/js/gradeable.js
josharsh: Nothing particular, just don't have much of experience with promises. Should I update it to use promises instead? @emaicus

site/public/js/ta-grading-rubric.js
josharsh: Removed!

josharsh:
image
@emaicus I am sorry, I don't understand "access control on the saving of custom comments". Custom marks are not accessible to peer graders. Does access control on custom components serve any purpose. Am I misinterpreting anything?

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1945,6 +1945,12 @@ public function ajaxSaveMark($gradeable_id) {
             return;
         }
 
+        // get the user
+        if ($this->core->getUser()->getGroup() === User::GROUP_STUDENT){
+            $this->core->getOutput()->renderJsonFail('Insufficient permissions for custom marks');
+            return;
+        }
```
emaicus: I think this will stop peers from being able to grade at all. What we actually want is for them to not be able to modify/add/remove custom marks.

The check that you want to add should be in `saveGradedComponent`, where `setComment` and `setPoints` are called for a graded component. Confusingly, I think that these calls represent the changing of a custom mark.

jchulton: This is in the ajaxSaveMark method, which is used when a limited access grader or above edits a mark (changing title, amount of points the mark is worth, etc.). This already contians a check to prevent students from editting the mark (line 1961). While this change doens't stop peers from being able to grade, Evan is correct that the check should be made in saveGradedComponent

jchulton:
Tested and works. Looks good

bmcutler:
closing this -- replaced by #6551

PR #5753, opened 2020-07-30
[Bugfix:System] run migrator only on courses in db

MasterOdin:

Another side-effect of the stricter error checking from #5735 is that older courses may have their databases removed, but may remain on the filesystem, to be used for plagiarism detection or other purposes. Because we're using the filesystem to determine the list of courses to migrate, there was no practical way to detect a course that was in this state versus a course that is active.

### What is the new behavior?

The migrator now consults the master database to determine the list of courses to migrate, only affecting those that have a status of 1 or 2 (see https://submitty.org/sysadmin/course_archiving for meaning). As such, courses that exist only on the filesystem or have a status not in (1, 2) are skipped. This prevents the crash especially in the case of courses that exist only on the filesystem, and not within the database, nor have their own database, while retaining the stricter error checking of #5735 that we do not error on a database that should exist, but does not (e.g. for active courses).


bmcutler:
I've already started this in another branch because it broke on production.
Another bug you introduced when you merged #5735 without review.

MasterOdin:
I did not merge #5735...
Screen Shot 2020-07-30 at 9 38 24 AM

bmcutler:
@MasterOdin 

Yeah, this again crashed when I tried to run it on our production system.

-------------------------------------------------------------------------------------

The error:
Migrator Error:  Could not find directory for s21 <REDACTED>_PRACTICE

Yes, it's a practice course, but it does indeed have a corresponding directory, looks like a real course directory to me on quick glance.  I'm not sure what's different.  (Yes, probably something is different, I don't know the history of how this course was created many months ago, maybe there was some funny business. )

I made this change below to limp along:
Of course this isn't what you want in the PR.

-------------------------------------------------------------------------------------

diff --git a/migration/migrator/main.py b/migration/migrator/main.py
index 4a2958c92..d2ad3ef35 100644
--- a/migration/migrator/main.py
+++ b/migration/migrator/main.py
@@ -248,9 +248,11 @@ def handle_migration(args):
             for semester in courses:
                 for course in courses[semester]:
                     if not Path(course_dir, semester, course).exists():
-                        raise SystemExit(
-                            f'Migrator Error:  Could not find directory for {semester} {course}'
-                        )
+                        #raise SystemExit(
+                        #    f'Migrator Error:  Could not find directory for {semester} {course}'
+                        #)
+                        print (f'Migrator Error:  Could not find directory for {semester} {course}')
+                        continue
                     loop_args = deepcopy(args)
                     cond1 = loop_args.choose_course is not None
                     cond2 = [semester, course] != loop_args.choose_course

bmcutler:
Following on to my previous comment.  

I don't think we want to force the database & the file system for old terms to be in lockstep.

We have been moving whole directories from old terms to slower disk, maybe eventually we will remove them from the production system entirely.  Disk space, and fast (SSD) disk space is limited resource.

Similarly, at some point we will probably want to untangle the databases for some old courses / some old terms and remove them from the db server -- this will be a larger challenge since the main database will need to be edited or rebuilt.

bmcutler:
I fixed up our database (goof in course creation from 2 years ago) and the migration ran.
If/when something similar happens in the future it will break on the next installation, but maybe that's an appropriate failure mode.

PR #5762, opened 2020-07-31
[UI/UX:InstructorUI] Notebook Builder 4.1

Kevin-MM:
### Installation Notes
- You should run `INSTALL_SUBMITTY.sh clean` to ensure that some additional codemirror dependencies are copied into place.

#### New CodeMirror Utilities
- A new PHP class called `CodeMirrorUtils` has been added which consolidates codemirror related code into one place.
- `CodeMirrorUtils` defines a set of default dependencies and makes it much easier for PHP to register these dependencies with the `Output` object.
- A javascript `code-mirror-utils` file has also been added which provides useful codemirror functions javascript developers will need.
- Currently only the notebook builder and notebook submission page have been rolled over to use these new utilites.  Any other codemirror boxes being used on submitty will be rolled over in a later PR.

#### Improved Short Answer Widget
- A new drop down has been added which allows the instructor to specify 'Default' for a text answer, or a programming language if they would like to collect an answer that is written in a specific language.
- Instructors may also specify a height of the input box.  When using a programming language the instructor will see real time changes to the codemirror box so they can get a good idea of how it will appear to students on the submission page.  If they have 'Default' selected they will see a text area.  Its worth noting that depending on how they set the height the input, on the submission page it may appear as either a `<input>` or `<textarea>`.  This is strange right now but as we move towards using codemirror for collecting all notebook inputs this bug will go away.


shailpatels:
PR works great! just a couple of small changes, here is a screenshot of the short answer widget set to a codebox

image

Kevin-MM:
@shailpatels I got rid of the question mark and added a message that is always visible.  For the record, I don't like it because the widget is starting to feel very cluttered.  At some point in the future I plan to write some proper documentation and then maybe I will just add a question mark link to every widget linking to their corresponding documentation.

I also added the requested accessibility message about using TAB and ESC.

As for the filename validation we will need to better define what constitutes a valid filename from a submitty perspective before I am able to move forward on that.  From a linux perspective the only illegal values appear to be [the null byte and forward slash](https://unix.stackexchange.com/questions/230291/what-characters-are-valid-to-use-in-filenames).

PR #5766, opened 2020-08-03
[Testing:TAGrading] Next/Prev e2e tests

jchulton:

We have no testing for the next/prev student buttons in TAGrading and they are regularly being broken

### What is the new behavior?
There are a series of testcases for the next/prev/next-ungraded/prev-ungraded that can be run only locally to confirm the buttons still work


Kevin-MM:
@jchulton I'm a little bit confused.  Github doesn't show any changes to any files?

Kevin-MM:
@jchulton Hi Jack, Can you please resolve the e2e test failures, or add needed exclusions.

Kevin-MM (approved):
This is a good start.  

Some notes:
- Right now these are only testing the instructor and ta user.  These users by default don't have any sections assigned.  For a later PR, we will need more thorough tests.  For example user ta3 has sections 1 and 3 assigned.  We will need to test that the buttons work when skipping over section 2.
- These tests run on the current TA grading interface, as we transition to the new TA grading interface these may need changes.
- These tests didn't work for me by default.  I had to recreate sample courses because as I've been developing new features I've been making changes to the sample course.  I stand firmly with what I've said before being that we need to figure out a way to get these tests running in travis.  I think its going to be a common thing where developers don't run them and they become stale.  In the event they do run them they may not recreate sample courses and the tests may fail, leaving them wondering what is broken?  Is it the tests, is it my development environment? In the event they do recreate sample courses they are wasting a huge amount of their time they could be putting to better use.

PR #5778, opened 2020-08-05
[VPAT:System] esc key to close popups

### LINKED ISSUE #5775: "Escape" key on a modal should close/cancel the modal
A standard on many websites is that pressing "esc" key when focused on a modal escapes/closes/cancels that modal.
Let's do this!  It will be good usability / accessibility change.  Is it simple, just a tweak to our base modal class?

assigning both @jportorreal00 and @elihschiff.  Please coordinate who will work on this.

elihschiff:

The only way to close an open popup is to click the close button (or tab until you are focused on it and press enter)

### What is the new behavior?
partial work for #5775
pressing the escape key when a popup is open will now also close the popup

### DIFF
--- a/site/app/templates/admin/users/MoreExtensions.twig
+++ b/site/app/templates/admin/users/MoreExtensions.twig
@@ -10,7 +10,10 @@
     {% endfor %}
     </ul>
 {% endblock %}
+{% block close_button_id %}id="team-extension-cancel"{% endblock %}
+{% block close_click_action %}javascript:confirmExtension(0){% endblock %}
+{% block close_button_text %}No{% endblock %}
 {% block buttons %}
-    <a id="team-extension-cancel" class="btn btn-default" href="javascript:confirmExtension(0)">No</a>
+    {{ block('close_button') }}
     <a class="btn btn-primary" href="javascript:confirmExtension(1)">Yes</a>
 {% endblock %}
--- a/site/app/templates/admin/users/UserForm.twig
+++ b/site/app/templates/admin/users/UserForm.twig
@@ -102,7 +102,9 @@
         {{ parent() }}
     </form>
 {% endblock %}
+
+{% block close_click_action %}closeButton(){% endblock %}
 {% block buttons %}
-    <a onclick="closeButton()" class="btn btn-default close-button key_to_click" tabindex="0">Close</a>
+    {{ block('close_button') }}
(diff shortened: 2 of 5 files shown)


jportorreal00:
@elihschiff When I make changes to one of the fields in the popup and press escape, it closes nicely. But if I try to leave the page after I am left with a popup asking me if I want to save my changes even though I can no longer see the popup. I'm not sure if we ant to keep that. I think a good idea would be reverting any changes after we press escape. What do you think?

elihschiff:
It ended up being a bit more complicated to fix that then I thought. The problem was some popups implemented their own custom close buttons which did not work with my generic close on esc press I had just implemented.

To fix this I made the default close buttons more generic by adding optional arguments using twig blocks.
You can see my main changes in Popup.twig on about lines 24-27. I then just went to all the custom buttons and made them follow my more generic format.

jportorreal00:
For some reason this doesn't work for editing a forum post. The escape key does nothing :(

elihschiff:
> It'll probably be easier to just add a listener for the escape key under generic popout and trigger the close button when it happens then defining twig blocks .

I do have this. (line 32-41 as of this comment)
image
Your version does not use jquery, which might be better and I can switch it to do that.

The reason I use twig blocks is because not all modals use the popup.twig `closePopup()` function to close. Some modals have a custom function they want to run. So by default the twig block calls `closePopup()` but in the case that a modal prefers to use its own custom close function I added the twig block to allow that. The other option is to remove the twig block and copy the entire keypressed detection code next to every modal that does not use the default `closePopup()` function. But that imo is a worse solution because it requires a ton of copied code that will be really hard to track down if we need to change it for any reason.

shailpatels:
@elihschiff I should have separated the requested changes and the comment, the instructor edit post button is also broken for me for some reason something is weird about that modal you could use the event listener specifically for that modal to fix it but once I tried it out I was getting issues with it always complain changes made would be lost even though I didn't do anything.

Not sure what the best way around that is for now, maybe just leave it at always giving the popup or commenting out the lines that are trying to check if you made changes or not in the edit post form like I suggested

shailpatels:
@bmcutler I think we should merge this PR as it is now since it works fine with every modal on the site, other than a couple on the forum which behave weirdly from old code and make a new issue specifically for the forum.

PR #5660, opened 2020-07-09
[SYSADMIN ACTION][Feature:InstructorUI] websockets in office hours queue

marwanatef2:
### What is the current behavior?
The office hours queue makes a request every 30 seconds to check if there is an update in the queue, and if so it renders the parts of the page that needs to get updated without refreshing the whole page.

### What is the new behavior?
Introducing websockets in the office hourse queue will only update the page on new socket messages without the need to ask (request) every 30 seconds whether there is an update or not.

### Other information
To enable websockets in order to test this PR please see [Docs](https://submitty.org/sysadmin/version_notes/v20.06.00)

[ draft docs PR : https://github.com/Submitty/submitty.github.io/pull/272 ]

PR #5779, opened 2020-08-05
[Feature:instructorui] Add pause button to office hours queue

elihschiff:

When the office hours queue is long a student may be waiting for a long period of time to get help. In the unlucky situation where say they need to go to the bathroom right as they are next in the queue, they may get called on and then lose their spot in line because they were unable to respond 

### What is the new behavior?
This pr adds the ability for students to let mentors know they will be away for a few min through a pause button.

Student view:
image

Grader view:
image
If you try and help that student, it will warn you making sure your really want to help the student even though they are paused
image

### Other information?

There is a chance that a student could abuse this system and stay paused until they have a question, currently the way we talked about dealing with that is to just remove students from the queue that are paused for longer than a reasonable amount of time

### DIFF
--- a/site/app/controllers/OfficeHoursQueueController.php
+++ b/site/app/controllers/OfficeHoursQueueController.php
@@ -193,6 +193,25 @@ class OfficeHoursQueueController extends AbstractController {
     }
 
     /**
+     * @Route("/courses/{_semester}/{_course}/office_hours_queue/togglePause", methods={"POST"})
+     * @return MultiResponse
+     */
+    public function setQueuePauseState() {
+        if (empty($_POST['pause_state'])) {
+            $this->core->addErrorMessage("Missing pause state");
+            return MultiResponse::RedirectOnlyResponse(
+                new RedirectResponse($this->core->buildCourseUrl(['office_hours_queue']))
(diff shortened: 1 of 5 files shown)


elihschiff:
Thanks for reviewing my pr!

I think that isn't too much of an issue as once web sockets are enabled, the mentor page will automatically update with the fact a student is paused. This update happens almost instantly after the student pauses.

elihschiff:
I didn't disable it because I don't want to enforce not helping paused students. There might be a legitimate reason to help them. I think adding a warning making sure they actually wanted to help the paused student is enough to make sure nobody accidentally clicks help before realizing that student is paused

bmcutler:
Looks great!  However, it does not update dynamically.
Was this previously working or did it break with the websockets update I just merged?
@elihschiff @marwanatef2

marwanatef2:
This functionality wasn't in #5660 , I think as this is a new feature we should update it using websockets either in this PR or a new one but as we don't use auto-refresh anymore, the refresh is on socket messages.

elihschiff:
This pr should now work with websockets, it was just a small 1 line change

PR #5789, opened 2020-08-07
[Refactor:System] Replacing Hardcoded Anon Ids

jchulton:

When sample courses are created, several users are created with hardcoded anon_ids, instead of having randomly generated anon_ids, as is normal

### What is the new behavior?
When sample courses are created, all users' anon_ids are randomly generated


josharsh:
@jchulton After you've rolled back the changes, and introduced hard coded anon ids which are no more simply the username should the PR description be updated before it is merged?

PR #5765, opened 2020-08-02
[Feature:TAGrading] websockets in grading inquiry

marwanatef2:
### What is the current behavior?
Students/Graders should refresh the page to get new posts in the grade inquiry discussion which may be boring if the discussion goes a long way or may be they will forget and the inquiry will not be resolved in the same time of discussion.

### What is the new behavior?
The grade inquiry discussion is updated on the go with websockets whether new posts are added of the inquiry is resolved users will get the updates immediately.

### Other information?
In order to test you'll need to have websockets enabled, so please check the [docs](https://submitty.org/sysadmin/version_notes/v20.06.00)

PR #5794, opened 2020-08-10
[Refactor:System] Specify socket client page manually

marwanatef2:

Socket clients' pages are detected from the Breadcrumb (course and page), this behavior was working just as fine in discussion forum and office hours queue because their breadcrumb title was the same for all, however it won't work on grading inquiry where the page breadcrumb for graders is different from that of students (containing "Grading" at the end).

image
image
image

Edit to add:  This PR also fixes a bug where for installations that have specified the (optional) institution name, which appears as the first item in the breadcrumb.

### What is the new behavior?
The socket page is passed as an argument to the socket initialization function.

### Other information
This PR needs to get merged before #5765


Kevin-MM:
Hi @marwanatef2 , If I have one forum tab open, and one office hours tab open, are these two separate socket connections?

Kevin-MM:
Also could you please update the function documentation for `Server::broadcast()`?  It is no longer accurate and also I was confused about what the third parameter did initially.

marwanatef2:
> are these two separate socket connections?

yeah, the socket client is per user's window, so any new window/tab opens a new socket client.

But this point reminds me to refactor socket clients storing with users ids to be able to store multiple clients per user_id, thanks for that @Kevin-MM

PR #5812, opened 2020-08-13
[Testing:TAGrading] additional PDF e2e tests

### LINKED ISSUE #5641: access to pdf file while grading buggy
**Describe the bug**

Peer graders can view the popout of the PDF,  but not the inline version of the PDF.

TA graders can view the inline version of the PDF, but not the popout version of the PDF.

---

We need regression tests for this.  This has broken way too many times.
There are many combinations that need to be tested -- both success & failure to access files based on access.

instructor vs full access vs limited access vs peer
rotating vs registration
explicitly assigned vs has access event though not assigned (instructor & full access)
individual vs team
popout vs inline

jchulton:

Closes #5641
PR #5700 adds a basic test suite for opening PDFs while grading. But misses some cases.

### What is the new behavior?
This covers the testcases listed in #5641 that #5700 does not cover. This includes tests for rotating sections, all grades all, and testing the popup pdf view.
### Other information?
This adds new gradeables to the sample course: recreate_sample_courses must be run. This should be merged after #5700


tests/e2e/base_testcase.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, testname, user_id=None, user_password=None, user_name=None, l
         """ :type driver: webdriver.Chrome """
         self.options = webdriver.ChromeOptions()
         self.options.add_argument('--no-sandbox')
-        self.options.add_argument('--headless')
+        #self.options.add_argument('--headless')
```
shailpatels, line 53: Should add `--headless` back in before merging

shailpatels (changes requested):

shailpatels (approved):
New tests appear to work correctly in CI

GitHub is complaining about not seeing codecoverage but should be good to merge

PR #5829, opened 2020-08-16
[Feature:TAGrading] websockets for checkpoint & numeric

marwanatef2:

Simple grading (checkpoints / numeric) is accessed by multiple graders simultaneously but doesn't refresh upon changes so graders can't see the latest page updates.

### What is the new behavior?
Integrating websockets in simple grading will allow changes to any checkpoint/numeric-input to take place concurrently on all open graders pages (windows) so they can see the latest updates in real time.

### DIFF
--- a/site/app/controllers/grading/SimpleGraderController.php
+++ b/site/app/controllers/grading/SimpleGraderController.php
@@ -172,8 +172,6 @@ class SimpleGraderController extends AbstractController {
             $graders[$section->getName()] = $section->getGraders();
         }
 
-        $grader = $this->core->getUser();
-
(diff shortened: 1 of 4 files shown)


site/app/controllers/grading/SimpleGraderController.php
```diff
@@ -172,6 +172,8 @@ public function gradePage($gradeable_id, $view = null, $sort = null) {
             $graders[$section->getName()] = $section->getGraders();
         }
 
+        $grader = $this->core->getUser();
```
jchulton, line 175: Can you not call this in SimpleGraderView.php instead of calling this here and passing the variable through?

marwanatef2: I'm sorry can you elaborate more?

PR #5836, opened 2020-08-17
[Bugfix:System] Fix display_image_state sync on new installations

Kevin-MM:
When a system admin creates a brand new submitty installation (for example during a fresh vagrant up) the database `sync_user` trigger function doesn't include any functionality for propagating `display_image_state` in the `user` table from the master database to course level databases.  Because of this, much of the student photos preferred photos and flagging is broken on new installations.

### What is the new behavior?
This issue has been resolved by correcting the `sync_user()` trigger function.

### Other information?
This bug only manifest itself on new installations.


shailpatels:
How should this be tested, a vagrant destroy down and then check the database ? Can a migration be written to avoid a destroy?

Kevin-MM:
No the only way to test would be to destroy and up.

Kevin-MM:
Further, existing installations can be repaired by rolling back and running the most recent master migration.

Kevin-MM:
I've added another migration which will bring current installations up to speed.

PR #5837, opened 2020-08-17
[Bugfix:System] detect websocket server is down

marwanatef2:

If the socket server is down for any reason, some pages as office hours queue produce an unwanted behavior that users should handle manually.

image

### What is the new behavior?
Users' actions are completed successfully as they should but they get a notice message that the server socket is down and the page won't dynamically load.

socketserverdown

### Other Infromation
In order to test this you have to stop the websocket server so it doesn't accept connections running `systemctl stop submitty_websocket_server` on your vagrant machine.

### DIFF
--- a/site/app/controllers/OfficeHoursQueueController.php
+++ b/site/app/controllers/OfficeHoursQueueController.php
@@ -510,6 +510,10 @@ class OfficeHoursQueueController extends AbstractController {
         );
     }
 
+    /**
+     * this function opens a WebSocket client and sends a message with the corresponding update
+     * @param array $msg_array
+     */
     private function sendSocketMessage($msg_array) {
         $msg_array['user_id'] = $this->core->getUser()->getId();
         $msg_array['page'] = $this->core->getConfig()->getCourse() . "-office_hours_queue";
@@ -518,7 +522,7 @@ class OfficeHoursQueueController extends AbstractController {
             $client->send($msg_array);
         }
         catch (WebSocket\ConnectionException $e) {
-            $this->core->addNoticeMessage("WebSocket Server is down, Page won't load dynamically.");
+            $this->core->addNoticeMessage("WebSocket Server is down, page won't load dynamically.");
         }
     }
 }
--- a/site/app/controllers/OfficeHoursQueueController.php
+++ b/site/app/controllers/OfficeHoursQueueController.php
@@ -512,7 +512,12 @@ class OfficeHoursQueueController extends AbstractController {
     private function sendSocketMessage($msg_array) {
         $msg_array['user_id'] = $this->core->getUser()->getId();
         $msg_array['page'] = $this->core->getConfig()->getCourse() . "-office_hours_queue";
-        $client = new Client($this->core);
-        $client->send($msg_array);
+        try {
+            $client = new Client($this->core);
+            $client->send($msg_array);
+        }
+        catch (\Websocket\ConnectionException $e) {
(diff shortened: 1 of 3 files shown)


bmcutler:
@marwanatef2 
please fix php stan / lint errors

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -512,7 +513,12 @@ public function showNewAnnouncement() {
...
-        $client = new Client($this->core);
-        $client->send($msg_array);
+        try {
+            $client = new Client($this->core);
+            $client->send($msg_array);
+        }
+        catch (WebSocket\ConnectionException $e) {
+            $this->core->addNoticeMessage("WebSocket Server is down, Page won't load dynamically.");
+        }
     }
```
Kevin-MM, line 523: Can you please add documentations above this function and also type hints.

Additionally can you change the error message to be "WebSocket Server is down, page won't load dynamically."  The `p` shouldn't be capitalized.

site/app/libraries/socket/Server.php
```diff
@@ -213,7 +213,7 @@ public function onMessage(ConnectionInterface $from, $msgString) {
 
         $msg = json_decode($msgString, true);
 
-        if ($msg["type"] === "new_connection") {
+        if ($msg["type"] === "new_connection" && isset($msg['page'])) {
             if (!array_key_exists($msg['page'], $this->clients)) {
```
Kevin-MM, line 217: If `$msg['page']` is not set, we will end up in the `else` block and call `$this->broadcast()`.  Is that desirable in this case?

marwanatef2: No it's not, as broadcast needs the page the client is on which is **not set**

site/app/controllers/OfficeHoursQueueController.php
marwanatef2: Is the documentation added here clear enough?

site/app/libraries/socket/Server.php
```diff
@@ -214,15 +214,20 @@ public function onMessage(ConnectionInterface $from, $msgString) {
...
-                $course_page = explode('-', $this->getSocketClientPage($from));
-                echo "New connection --> user_id: '" . $this->getSocketUserID($from) . "' - course: '" . $course_page[0] . "' - page: '" . $course_page[1] . "'\n";
+                if ($this->core->getConfig()->isDebug()) {
+                    $course_page = explode('-', $this->getSocketClientPage($from));
+                    echo "New connection --> user_id: '" . $this->getSocketUserID($from) . "' - course: '" . $course_page[0] . "' - page: '" . $course_page[1] . "'\n";
+                }
+            }
+            else {
+                $from->close();
             }
```
Kevin-MM, line 231: Small recommendation here would be to log a message using the `Logger` class in `Logger.php` when this block is reached.  We shouldn't really end up here but if we do it would be nice to have something logged to help debugging by sys admins.

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -509,10 +510,19 @@ public function showNewAnnouncement() {
         );
     }
 
+    /**
+     * this function opens a WebSocket client and sends a message with the corresponding update
+     * @param array $msg_array
+     */
     private function sendSocketMessage($msg_array) {
```
Kevin-MM, line 517:
```suggestion
    private function sendSocketMessage(array $msg_array): void {
```

PR #5847, opened 2020-08-21
[Feature:InstructorUI] Communicate schema warnings to webUI

### LINKED ISSUE #5733: Add "Build Completed with Warning" status to the WebUI for building gradeables
When building a gradeable from the web, the UI says "Build Status: Gradeable build complete" even if the schema validator gave warnings. Update the UI to instead say "Build Status: Gradeable built with warnings" in orange in this case.

It may be necessary to add a special string to the schema validator's output to grep for to tell when warnings have occurred.

Kevin-MM:
Closes #5733 

On the 'Submissions/Autograding' tab of the edit gradeables page a small box indicates the build status of the gradeable.  If a gradeable is successfully build it will show `Build Status: Gradeable build complete` even if the schema validator gave warnings.

### What is the new behavior?
If the schema validator generated warnings during the gradeable build process the message will now show `Build Status: Gradeable built with warnings`.

### DIFF
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -1231,7 +1231,7 @@ class AdminGradeableController extends AbstractController {
     /**
      * @Route("/courses/{_semester}/{_course}/gradeable/{gradeable_id}/build_log", methods={"GET"})
      */
-    public function ajaxGetBuildLogs($gradeable_id) {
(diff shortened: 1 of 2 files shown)


shailpatels (approved):
Whoops not sure how I missed this PR - It feels like I've reviewed it before.

But it correctly propagates info about warnings in the build and expanding the log shows what gets logged 
Example output from a config with error

image

PR #5878, opened 2020-09-09
[Bugfix:Submission] Clean invite rcsid on team invite

jshom:

On team invite, the RCSID must be all lower case and must not have spaces around the input.

Results in errors like this:
Screen Shot 2020-09-08 at 8 25 02 PM

### What is the new behavior?
Cleans the input so now the issue in the screenshot shouldn't occur.
Would lookup `shomsj` that exists; instead of `Shomsj` that doesn't.

### Other information?
_Is this a breaking change?_
Nope

_How did you test_
`./site/tests/app/controllers/student/TeamControllerTester.php`

### DIFF
--- a/site/app/controllers/student/TeamController.php
+++ b/site/app/controllers/student/TeamController.php
@@ -524,7 +524,7 @@ class TeamController extends AbstractController {
      * @param string $invite_id
      * @return string
      */
-    private static function cleanInviteId(string $invite_id) {
+    private static function cleanInviteId(string $invite_id): string {
         return trim(strtolower($invite_id));
     }
 }
--- a/site/app/controllers/student/TeamController.php
+++ b/site/app/controllers/student/TeamController.php
@@ -168,13 +168,13 @@ class TeamController extends AbstractController {
         $invite_id = self::cleanInviteId($_POST['invite_id']);
         $invited_user = $this->core->getQueries()->getUserById($invite_id);
 
-        if (empty($invited_user)) {
+        if ($invited_user === null) {
             // If a student with this id does not exist in the course...
             $this->core->addErrorMessage("User {$invite_id} does not exist");
             $this->core->redirect($return_url);
         }
 
-        if (empty($invited_user->getRegistrationSection())) {
+        if ($invited_user->getRegistrationSection() === null) {
             // If a student with this id is in the null section...
             // (make this look the same as a non-existant student so as not to
             // reveal information about dropped students)
@@ -520,7 +520,7 @@ class TeamController extends AbstractController {
     }
 
     /**
-     * Get cleaned invite RCSID
+     * Get clean user_id invite
      * @param string $invite_id
      * @return string
      */
--- a/site/app/controllers/student/TeamController.php
+++ b/site/app/controllers/student/TeamController.php
@@ -165,14 +165,16 @@ class TeamController extends AbstractController {
             $this->core->redirect($return_url);
         }
 
-        $invite_id = $_POST['invite_id'];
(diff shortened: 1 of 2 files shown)


site/app/controllers/student/TeamController.php
```diff
@@ -165,14 +165,16 @@ public function sendInvitation($gradeable_id) {
             $this->core->redirect($return_url);
         }
 
-        $invite_id = $_POST['invite_id'];
-        if ($this->core->getQueries()->getUserByID($invite_id) === null) {
+        $invite_id = self::cleanInviteId($_POST['invite_id']);
+        $invited_user = $this->core->getQueries()->getUserById($invite_id);
+
+        if (empty($invited_user)) {
```
shailpatels, line 171: You shouldn't use `empty` on php strings because of its weird type casting rules e.g the string "0" is considered an empty string when its technically a valid user_id

a stronger comparison would be 
```php
if ( $string === '') {
//handle empty string
```

site/app/controllers/student/TeamController.php
```diff
@@ -165,14 +165,16 @@ public function sendInvitation($gradeable_id) {
...
             // If a student with this id does not exist in the course...
             $this->core->addErrorMessage("User {$invite_id} does not exist");
             $this->core->redirect($return_url);
         }
 
-        if ($this->core->getQueries()->getUserByID($invite_id)->getRegistrationSection() === null) {
+        if (empty($invited_user->getRegistrationSection())) {
```
shailpatels, line 177: This change should be reverted back as a comparison to `null` since a student's registration section could be null if they've dropped the class or haven't been assigned to one yet.

site/app/controllers/student/TeamController.php
```diff
@@ -516,4 +518,13 @@ public function showPage($gradeable_id) {
         $this->core->getOutput()->addBreadcrumb("Manage Team For: {$gradeable->getTitle()}");
         $this->core->getOutput()->renderOutput(['submission', 'Team'], 'showTeamPage', $gradeable, $team, $members, $seekers, $invites_received, $seeking_partner, $lock);
     }
+
+    /**
+     * Get cleaned invite RCSID
```
shailpatels, line 523: Should change to "get clean user_id invite", RCS ID is an RPI specific term

shailpatels (changes requested):

Just a couple of small changes requested should be ready to merge soon

site/app/controllers/student/TeamController.php
jshom: Happy to change to null comparison (I assume you were referring to the `empty($invited_user)`), taking a second look before I do:
```php
// site/app/libraries/database/DatabaseQueries.php
143     /**
144      * Gets a user from the database given a user_id.
145      */
146     public function getUserById(string $user_id): ?User {
147         return $this->getUser($user_id);
148     }
```
Since `getUserById` is returning a nullable User, it should either be User or null that should allow us to use `empty($invited_user)` right?

shailpatels: I would still compare the object to null php would convert the object into an array before performing the empty check and I don't trust php to perform any implicit cat

It's a bit more clear too, you can use `isset` as well for $user I think that's used more often though submitty

jshom: 👍 will both to `isset`

shailpatels (approved):
Looks good, and works as described

jshom:
> Looks good, and works as described

Sweet! Don’t have merge privs, so merge whenever you’re ready. Thanks!

PR #5893, opened 2020-09-17
[Bugfix:Developer] Remove debug print statement which was triggered on every keystroke

emaicus:
A debug print statement was introduced to master recently which triggers on every keystroke on many pages of the system. This PR removes that statement.


shailpatels (approved):
Thanks for this! I missed debug statement in my review for the modal refactor

PR #5894, opened 2020-09-17
[Bugfix:Notifications] Add keepalive messages/heartbeats to websockets

emaicus:
By default, `nginx` drops inactive connections after sixty seconds. As a result, our websocket connections are frequently dropped and subsequently reconnected after an interval. To combat this issue, this PR adds a "ping" heartbeat message to each websocket, which is sent once every fifty seconds.


shailpatels:
I think this should work,
one note is that JavaScript intervals on tabs that are inactive will be considered 'low priority' and the timer may either be delayed to stopped altogether but if that's not a problem this should be fine then.

Alternative would be to set [proxy_read_timeout](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout) to some big number (10m, 1h, 1d etc) I searched a little bit and it appears you can't just disable this however.

shailpatels:
I would say maybe extend the timeout to 5m or something else long and we can have an interval on each page that uses sockets that checks the websocket `readyState` every x interval and trys to reconnect if closed

this would give extra protection against people who may tab out to do something else which might be a usecase with something like office hours q ?

MasterOdin:
It's almost certainly better to be dealing with keep alive frames from the server to the client, as it also allows the server to kill off dead connections from its memory for connections that may have been improperly closed client side. Ratchet provides this functionality, it just needs to be enabled through a function call. The benefit of this method is that this would then follow the spec for websockets and ping/pong frames: https://tools.ietf.org/html/rfc6455#section-5.5.2.

shailpatels:
Make sure the websockets service is restarted after installing this code
`systemctl restart submitty_websocket_server.service`

shailpatels (approved):
Approved, best code i've ever seen

PR #5906, opened 2020-09-23
[Bugfix:installation] Fix initial vagrant up & vagrant provision

shailpatels:

A fresh vagrant up appears to be broken, doing a `vagrant reload --reload` breaks and leaves a broken vm.

### What is the new behavior?
Attempts to fix this
Main issue with a fresh vagrant up,
postgresql would fail to perform peer authentication, not sure why but set it to trust any localhost account

vagrant provision would crash with
setting the motd file permissions
trying to install haskell stack if it already exists
removing files that don't exist
creating a symlink thats already there

### Other information?
need to test if this actually fixes vagrant provision, will do that later but should fix the vagrant up issue


shailpatels:
Vagrant up is reported to be working fine by others,
will test vagrant provision later

PR #5909, opened 2020-09-23
[Feature:Submission] Show due dates in user/server timezone on submission page

### LINKED ISSUE #5888: Add due date in server timezone to submission page
Now that we have implemented timezone handling site wide, some students have become confused about whether due dates are formatted for their local time or server time.

For this reason, if a student has their timezone set outside of the server timezone, they will be shown 2 timestamps on the submission page.  One will be clearly formatted for their timezone, the other will be clearly formatted for the server timezone.

Kevin-MM:
Closes #5888 

When a user navigates to the submission page for a gradeable they will see the due date formatted for their local time zone.  Despite including a time zone identifier this was still confusing for some users.

### What is the new behavior?
On the submission page, if a user has their timezone set differently than the server timezone, they will see the due date formatted in both their local time and the servers time.

If a user has their timezone set to the same timezone as the server, they will see no changes on this page.

image

### DIFF
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -20,7 +20,20 @@
             {% endif %}
         </h1>
         {% if has_due_date %}
-            <h2>Due: {{ due_date|date(date_time_format) }}</h2>
+            {% set diff_time_zone = server_time_zone_string != user_time_zone_string %}
+
+            <div id="gradeable-info-due-dates">
+                <h2>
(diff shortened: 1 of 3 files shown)


shailpatels (approved):

Is there a point to also showing the server time? It wouldn't effect the student so I guess its not super critical if we do show it or not

Kevin-MM:
Requested by @bmcutler .  Some students were still confused about what time zone the time stamp was formatted for despite it including a time zone identifier.  This will (hopefully) remove all confusion.

PR #5920, opened 2020-09-30
[TAGrading: Bugfix]Cancelled Submissions Inconsistently Counted in Stats

jchulton:

If a submission is graded, then the submission is canceled by the student, the submission will count as graded in the numerator of the TA Grading by Section calculations, but not the denominator. This can lead to a numerator greater than the denominator.

### What is the new behavior?
The gradeable is counted in both the numerator and denominator. This prevents percentages greater than 100%


emaicus:
@jchulton, let's approach this differently. Instead of counting canceled submissions, let's discount them all together.

bmcutler (changes requested):
please see Evan's comment
the cancelled should not be in the numerator or the denominator.

jchulton:
Cancelled submissions are no longer in the numerator or denominator, as above

emaicus (changes requested):
This works well for standard (TA graded) gradeables.

At a glance, peer totals are incorrect. For the "Peer PDF Homework" in the sample course, I see `Students who have submitted: 180 / 101 (178.2%)`

Also, how should we be counting version conflicts? At the moment, they are counted towards completed grading. Perhaps we should list them separately?

emaicus (approved):
After a group conversation, it was agreed that peer stats are a separate issue and that version conflicts are being tallied correctly.

I have opened #6039, and approve this PR.

PR #5921, opened 2020-09-30
[BugFix:Forum] Clear autosave data for empty forum reply boxes

hdrodz97:

In the forum, if you type some text into a reply box, and the contents of the box are autosaved, and later you decide to clear out the text of that box, the old contents of the box will be shown the next time the page loads as the autosave data for that text box is not overwritten (due to a check to avoid polluting `localStorage` with a bunch of empty text box data).

### What is the new behavior?

When a forum reply box's data is about to be saved, if the reply box is empty, then its autosave data will be cleared from `localStorage` and a refresh will result in the reply box staying empty as intended.

### Other information?

Tested on Chrome.


shailpatels (approved):
Looks good, tested on firefox + chrome on Linux

PR #5927, opened 2020-10-02
[Feature:Autograding] Enable Autograding for Itempools

emaicus:
This PR aims to enable autograding for gradeables which utilize random item pools.

This work requires the following:
- [X] An update of the python autograding code (the `autograder` module) to read in `.submit.notebook` and run the associated testcases
- [X] An update to the `C++` grading code (`/grading`) to utilize string testcase id's rather than numeric indices
- [X] An update to the `main_validator` script to read in the `.submit.notebook` file and validate the associated testcases
- [x] An update to the submission page to display the results of itempool testcases
- [x] An update to the `AutogradingConfig` model to return student specific arrays of `AutogradedTestcases`
- [x] An update to the stats page to better display the statistics regarding itempool testcases
- [x] An update to the TA grading page to display the results of itempool testcases
- [x] A potential update to `Update Grade Summaries` to ensure compatability with itempool results


grading/load_config_json.cpp
```diff
@@ -1276,13 +1292,15 @@ void AddSubmissionLimitTestCase(nlohmann::json &config_json) {
...
+  assert (testcases != config_json.end());
+  for (nlohmann::json::iterator tc = testcases->begin(); tc != testcases->end(); tc++) {
     //This input to testcase is only necessary if the testcase needs to retrieve its 'commands'
     std::string container_name = "";
-    TestCase my_testcase(config_json,i,container_name);
-    int points = (*tc)[i].value("points",0);
+    std::string testcase_id = tc->value("testcase_id", "");
+    assert(testcase_id != "");
+    TestCase my_testcase(*tc, testcase_id, container_name);
+    int points = tc->value("points", 0);
```
bmcutler, line 1303: Is this counting up the points on the itempools too?
It should be.
But not double counting testcases in the global array + itempool testcases.  Right?

grading/load_config_json.cpp
```diff
@@ -1297,6 +1315,7 @@ void AddSubmissionLimitTestCase(nlohmann::json &config_json) {
     limit_test["type"] = "FileCheck";
     limit_test["title"] = "Submission Limit";
     limit_test["max_submissions"] = MAX_NUM_SUBMISSIONS;
+    limit_test["testcase_id"] = "submission_limit_test";
```
bmcutler, line 1318: Are we making sure somewhere that no one manually named their testcase "submission_limit_test"
(or if they did, it was actually overriding the submisison test and not some other test)

grading/load_config_json.cpp
```diff
@@ -1404,3 +1419,28 @@ void VerifyGraderDeductions(nlohmann::json &json_graders) {
     std::cout << "ERROR! DEDUCTION SUM < 1.0: " << sum << std::endl;
   }
 }
+
+std::vector<std::string> gatherAllTestcaseIds(const nlohmann::json& complete_config){
```
bmcutler, line 1423: similar logic to function "find_testcase_by_id"
Just wondering if they should use the same helper function?

grading/load_config_json.cpp
```diff
@@ -1404,3 +1419,28 @@ void VerifyGraderDeductions(nlohmann::json &json_graders) {
...
+      nlohmann::json item_testcases = item_itr->value("testcases", nlohmann::json::array());
+      for(nlohmann::json::iterator it_itr = item_testcases.begin(); it_itr != item_testcases.end(); it_itr++) {
+        std::string testcase_id = it_itr->value("testcase_id", "");
+        assert(testcase_id != "");
+        testcase_names.push_back(testcase_id);
+      }
+    }
+  }
+
+  return testcase_names;
```
bmcutler, line 1445: Might be good to throw an assertion in here that this return value vector has no duplicate values.

grading/load_config_json.h
```diff
@@ -24,11 +24,11 @@ void formatPreActions(nlohmann::json &testcases, nlohmann::json &whole_config);
 
 void RewriteDeprecatedMyersDiff(nlohmann::json &testcases, nlohmann::json &whole_config);
 
-void InflateTestcases(nlohmann::json &testcases, nlohmann::json &whole_config);
+void InflateTestcases(nlohmann::json &testcases, nlohmann::json &whole_config, int& testcase_id);
```
bmcutler, line 27: again these are index-or-count-or-counter not id
(int vs string)

grading/main_compile.cpp
```diff
@@ -24,14 +24,14 @@ int main(int argc, char *argv[]) {
...
-  int testcase_to_compile = -1;
+  std::string testcase_to_compile = "";
 
   TCLAP::CmdLine cmd("Submitty's main compilation program.", ' ', "0.9");
   TCLAP::UnlabeledValueArg<std::string> homework_id_argument("homework_id", "The unique id for this gradeable", true, "", "string" , cmd);
   TCLAP::UnlabeledValueArg<std::string> student_id_argument("student_id", "The unique id for this student", true, "", "string" , cmd);
   TCLAP::UnlabeledValueArg<int> submission_number_argument("submission_number", "The numeric value for this assignment attempt", true, -1, "integer" , cmd);
   TCLAP::UnlabeledValueArg<std::string> submission_time_argument("submission_time", "The time at which this submissionw as made", true, "", "string" , cmd);
-  TCLAP::ValueArg<int> testcase_to_compile_argument("t", "testcase", "The testcase to compile. Pass -1 to compile all testcases.", false, -1, "int", cmd);
+  TCLAP::UnlabeledValueArg<std::string> testcase_to_compile_argument("testcase", "The testcase to compile. Pass -1 to compile all testcases.", true, "", "string", cmd);
```
bmcutler, line 34: why did it change false to true?

grading/main_compile.cpp
```diff
@@ -72,108 +72,102 @@ int main(int argc, char *argv[]) {
   std::cout << "Compiling User Code..." << std::endl;
 
   system("find . -type f -exec ls -sh {} +");
+  assert (config_json.find("testcases") != config_json.end());
 
-  // Run each COMPILATION TEST
-  nlohmann::json::iterator tc = config_json.find("testcases");
-  assert (tc != config_json.end());
-  for (unsigned int i = 1; i <= tc->size(); i++) {
```
bmcutler, line 79: Why does this loop go away?
We can still have multiple compilation test cases.
We still need a loop.

bmcutler: Is this because main_compile will never do "all" since we got rid of the -1 option?

grading/main_compile.cpp
```diff
@@ -72,108 +72,102 @@ int main(int argc, char *argv[]) {
...
+  nlohmann::json testcase_config;
+  try {
+    testcase_config = find_testcase_by_id(config_json, testcase_to_compile);
+  } catch(char* c) {
+    std::cout << c << std::endl;
```
bmcutler, line 81: This should never happen?
Maybe be more clear that this is an error with that print statement?
cerr?  And write "ERROR" or something too?

grading/main_configure.cpp
```diff
@@ -319,11 +320,13 @@ int main(int argc, char *argv[]) {
...
-    if (tc.isSubmissionLimit()) {
-      max_submissions = tc.getMaxSubmissions();
+    nlohmann::json my_testcase_json = *itr;
+    TestCase my_testcase(my_testcase_json, test_id, container_name);
+    if (my_testcase.isSubmissionLimit()) {
+      max_submissions = my_testcase.getMaxSubmissions();
     }
-    all.push_back(printTestCase(tc));
+
+    all.push_back(printTestCase(my_testcase));
```
bmcutler, line 329: this is going away with your refactor to merge build & complete configs right???
because i think we're missing itempool testcases...

grading/main_runner.cpp
```diff
@@ -99,22 +101,21 @@ int main(int argc, char *argv[]) {
...
   std::string generation_type = "";
 
   TCLAP::CmdLine cmd("Submitty's main runner program.", ' ', "0.9");
   TCLAP::UnlabeledValueArg<std::string> homework_id_argument("homework_id", "The unique id for this gradeable", true, "", "string" , cmd);
   TCLAP::UnlabeledValueArg<std::string> student_id_argument("student_id", "The unique id for this student", true, "", "string" , cmd);
   TCLAP::UnlabeledValueArg<int> submission_number_argument("submission_number", "The numeric value for this assignment attempt", true, -1, "integer" , cmd);
-  TCLAP::UnlabeledValueArg<std::string> submission_time_argument("submission_time", "The time at which this submissionw as made", true, "", "string" , cmd);
-  TCLAP::ValueArg<int> testcase_to_run_argument("t", "testcase", "The testcase to run. Pass -1 to run all testcases.", false, -1, "int", cmd);
+  TCLAP::UnlabeledValueArg<std::string> submission_time_argument("submission_time", "The time at which this submission as made", true, "", "string" , cmd);
+  TCLAP::UnlabeledValueArg<std::string> testcase_to_run_argument("testcase", "The id of the testcase to run.", true, "", "string", cmd);
```
bmcutler, line 113: again, switch from false to true

grading/main_runner.cpp
```diff
@@ -159,46 +160,37 @@ int main(int argc, char *argv[]) {
...
-    std::vector<nlohmann::json> dispatcher_actions;
-    
-    if (my_testcase.isFileCheck() || my_testcase.isCompilation()){
-      continue;
-    }
+  nlohmann::json testcase_config;
+  try {
+    testcase_config = find_testcase_by_id(config_json, testcase_to_run);
+  } catch(char* c) {
+    std::cout << c << std::endl;
```
bmcutler, line 170: should never happen?
put to cerr and add an "ERROR " to the printed string?

grading/main_validator.cpp
```diff
@@ -397,13 +398,106 @@ void ValidateATestCase(nlohmann::json config_json, int which_testcase,
     testcase_message,
     testcase_pts,
     all_testcases);
+    WriteToGradefile(my_testcase,gradefile,testcase_pts);
+}
+
+/**
+* Get all testcase objects associated with an assignment.
+*/
+std::vector<TestCase> getAllTestcases(nlohmann::json config_json) {
```
bmcutler, line 407: wait this is a duplicated function?
gather all test cases was somewhere above?

grading/main_validator.cpp
```diff
@@ -397,13 +398,106 @@ void ValidateATestCase(nlohmann::json config_json, int which_testcase,
...
+        if((*item)["item_name"] == item_name) {
+          return *item;
+        }
+    }
+    throw "ERROR: Could not find an item with item_name " + item_name;
+}
+
+
+/**
+* Get all global testcase objects for an assignment + all testcases specified in .submit.notebook
```
bmcutler, line 449: Seems to overlap alot in functionality with getalltestcases
(and other functions in other files?)

library / standardize naming, 
use same function with additional args to do different things?

grading/main_validator.cpp
```diff
@@ -435,12 +529,17 @@ int validateTestCases(const std::string &hw_id, const std::string &rcsid, int su
...
 
   // =======================================
-  // LOOP OVER ALL TEST CASES
-  nlohmann::json::iterator tc = config_json.find("testcases");
-  assert (tc != config_json.end());
-  for (unsigned int i = 0; i < tc->size(); i++) {
+  // Loop over all global testcases
+  std::vector<TestCase> testcases;
+  if (generate_all_output) {
+    testcases = getAllTestcases(config_json);
```
bmcutler, line 535: oh, now I understand need for all vs for_submission.
but that should be in the comment at least, if not in the function naming
again, also think about merging into 1 function?
and pass in the generate output flag?

bmcutler (changes requested):
Haven't looked at grade item yet

autograder/autograder/grade_item.py
```diff
@@ -9,6 +9,294 @@
...
+    queue_obj,
+    working_directory,
+    which_untrusted,
+    item_name
+):
+    testcase_objs = []
+    testcase_specs = complete_config_obj['testcases']
+
+    for item in complete_config_obj['item_pool']:
+        testcase_specs += item_obj['testcases']
```
bmcutler, line 31: If there are multiple test cases in the item, does it add an array of multiple test cases or does it append them as individual test cases?  (the latter I hope/assume)

autograder/autograder/grade_item.py
```diff
@@ -9,6 +9,294 @@
...
+def get_all_testcases(
+    complete_config_obj,
+    config,
+    queue_obj,
+    working_directory,
+    which_untrusted,
+    item_name
+):
+    testcase_objs = []
+    testcase_specs = complete_config_obj['testcases']
```
bmcutler, line 28: this should be build config not complete config?
(will be same thing)

autograder/autograder/grade_item.py
```diff
@@ -9,6 +9,294 @@
...
+            '',
+            config.log_path,
+            config.error_path,
+            is_test_environment=False
+        )
+        testcase_objs.append(tmp_test)
+
+    return testcase_objs
+
+def get_testcases_for_user(
```
bmcutler, line 56: duplicate code...
should be merged with the above with an extra argument?

autograder/autograder/grade_item.py
```diff
@@ -9,6 +9,294 @@
...
+):
+    # ARCHIVE STUDENT RESULTS
+    print(f"{seperator}ARCHIVING STARTS", file=log_file)
+    log_file.flush()
+    for tc in testcases:
+        # Removes test input files, makes details directory for the testcase.
+        tc.setup_for_archival(log_file)
+
+    try:
+        # Perform archival.
```
bmcutler, line 282: unnecessary comment?

autograder/autograder/grade_item.py
```diff
@@ -59,121 +347,48 @@ def grade_from_zip(
...
-                working_directory,
-                testcases,
-                '',
-                config.log_path,
-                config.error_path,
-                is_test_environment=False
-            )
-
+        for tmp_test in get_all_testcases(
+            complete_config_obj,
```
bmcutler, line 366: calling the variable it complete vs build...  hmm

bmcutler (changes requested):
grade_item comments

bmcutler:
What are our naming restrictions on the testcase_id string?
* must be unique across the global array and all itempool objects
* no spaces or '/'
* underscores allowed
* restrict other special characters?
* ?

How we will use this string when we render/display each testcase?
Currently they render as "Test 1", "Test 2", etc. and are in directories named "test01", "test02", etc.

I think we could get in trouble if the testcase id name clashes with one of our other autograding related folders.
I think we should require that the testcase id begin with "test".

And then we can display by chopping of the "test" replacing it with "Test ", and also chopping off the leading zeros on the remaining part of the id, if any.  And perhaps also replacing all underscores with spaces.  And perhaps also replacing multiple spaces in a row with a single space.

So for display:

"test01" -> "Test 1"
"test001" -> "Test 1"
"test1.2" -> "Test 1.2"
"test_alpha" -> Test alpha"

autograder/autograder/grade_item.py
emaicus: Yes, this returns the union of the two lists.

grading/load_config_json.cpp
```diff
@@ -816,9 +814,13 @@ bool validShowValue(const nlohmann::json& v) {
            v == "on_success"));
 }
 
-void InflateTestcase(nlohmann::json &single_testcase, nlohmann::json &whole_config) {
+void InflateTestcase(nlohmann::json &single_testcase, nlohmann::json &whole_config, int& testcase_id) {
```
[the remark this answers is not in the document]
emaicus: This `testcase_id` is only here to allow us to name testcases with the same numbered scheme that we had previously, so as to preserve backwards compatibility. A better name might just be `testcase_count`.

grading/main_compile.cpp
emaicus: The cpp scripts used to be able to be called so that the loop was run internally, rather than handled by the python code. There is no place left in the system that still wants the cpp code to run all testcases, and it was only helpful when the script was being manually run by a dev.

grading/main_validator.cpp
emaicus: I'll type up an issue about making an  AutogradingConfig class in the cpp code.

tests/integrationTests/build_main_configure.sh
```diff
@@ -0,0 +1,53 @@
...
+    exit 1
+fi
+
+# Remove the intermediate config
+rm ${config_directory}/complete_config.json
+
+# Copy the build config into the config directory
+cp ${which_test_dir}/data/build_config.json ${config_directory}
+cp ${which_test_dir}/data/build_config.json ${config_directory}/complete_config.json
+cat ${config_directory}/complete_config.json
```
bmcutler, line 49: is this line necessary?  Is this extra output for debugging or is captured/checked by something?

tests/integrationTests/build_main_configure.sh
```diff
@@ -0,0 +1,53 @@
...
+
+ls -la .
```
bmcutler, line 51: debugging?

tests/integrationTests/tests/c_failure_messages/validation/grade.txt_alternate
```diff
@@ -1,8 +1,8 @@
...
-Testcase  3: BEHIND THE SCENES:  raw output                       3 /  5  
-Testcase  4: BEHIND THE SCENES:  checking the sorted output       5 /  5  
-Testcase  5: BEHIND THE SCENES:  checking the second and third words   5 /  5  
-Testcase  6: Submission Limit                                             
+test01: Compilation                                          5 /  5  
+test02: NORMAL GRADING:  (hiding the details)               10 / 10  
+test03: BEHIND THE SCENES:  raw output                       3 /  5  
+test04: BEHIND THE SCENES:  checking the sorted output       5 /  5  
+test05: BEHIND THE SCENES:  checking the second and third words   5 /  5  
+SubmissionLimit: Submission Limit
```
bmcutler, line 6: do we somewhere check/prevent using this id for an instructor-written testcase?
(would be better not to have this limitation at all, but I don't have a better suggestion right now)

tests/integrationTests/build_main_configure.sh
```diff
@@ -0,0 +1,53 @@
...
+cpp_res=$?
+if (( $cpp_res != 0 )); then
+    echo -e "\nFailed to run cpp preprocessor"
+    exit 1
+fi
+
+# Remove cpp markers and run the complete config json through a python json syntax checker.
+python3 ${GRADINGCODE}/json_syntax_checker.py ${config_directory}/complete_config.json
+py_res=$?
+if (( $cpp_res != 0 )); then
```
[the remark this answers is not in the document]
emaicus: Good catch. Updated here and in `bin/build_homework_function.sh`

tests/integrationTests/build_main_configure.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# Use the C Pre-Processor to strip the C & C++ comments from config.json
+cpp ${config_directory}/config.json ${config_directory}/complete_config.json
+cpp_res=$?
+if (( $cpp_res != 0 )); then
+    echo -e "\nFailed to run cpp preprocessor"
+    exit 1
+fi
+
+# Remove cpp markers and run the complete config json through a python json syntax checker.
```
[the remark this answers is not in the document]
emaicus: This and running the cpp pre-processor used to be a compound step. Updated.

tests/integrationTests/build_main_configure.sh
```diff
@@ -0,0 +1,53 @@
...
+./configure.out ${config_directory}/complete_config.json ${which_test_dir}/data/build_config.json test_assignment
+configure_res=$?
+
+if (( $configure_res != 0 )); then
+    echo -e "\nFailed to create a complete_config.json"
+    exit 1
+fi
+
+# Remove the intermediate config
+rm ${config_directory}/complete_config.json
```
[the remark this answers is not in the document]
emaicus: This file is actually an intermediate config. The true complete config is found at `${which_test_dir}/data/build_config.json`, and is copied to `${config_directory}/complete_config.json` below.

tests/integrationTests/build_main_configure.sh
emaicus: Good catch, yes, this was for debugging.

tests/integrationTests/tests/c_failure_messages/validation/grade.txt_alternate
emaicus: At the moment, instructors can't provide there own testcase ids. There is a comment in the code where this value is set (requested in your previous review) which lists adding a check for duplicate ids as a todo.

tests/integrationTests/build_main_configure.sh
emaicus: This was a debug line. I've removed it.

Issue #5316, opened 2020-04-30
Order of items in rainbow grades "bucket"


bmcutler:
**Describe the bug**
In the rainbow grades UI, gradeables are placed in each category correctly, but the default order of the items is not as expected.

I think we decided to order items by grades released date (or due date?)
But I think the order might be affected by the other gradeable settings.  Items in my "project" category are a mix of student upload, text numeric gradable, bulk upload, and some peer assignments, and this is the initial order everytime I go to the page:

<img width="387" alt="Screen Shot 2020-04-30 at 11 58 57 AM" src="https://user-images.githubusercontent.com/5871417/80732219-00d9d180-8ada-11ea-8e31-319349887bed.png">

Which doesn't make sense based on either due date or release date.  I manually drag them into this order:

<img width="388" alt="Screen Shot 2020-04-30 at 11 59 39 AM" src="https://user-images.githubusercontent.com/5871417/80732293-1b13af80-8ada-11ea-8259-6492192a26dc.png">

And it saves to customization.json ok, but when I return to the page the next time it's back to the wrong order again.

bmcutler:
Talked with @holzbh 

Request this ordering:

* gather all items that belong in the bucket.

* first, the items (if any) from that were previously saved in the customization.json (if it exists), in that order.
* second, the items newly assigned to the bucket (not in the customization) that have a due date (will this assignment have a due date == yes) , ordered by due date. 
* third, the items newly assigned that do not have a due date, ordered by release date.

shailpatels:
Changes look good and works as described

Saving buckets works and the order is preserved when manually edited.
Grades are ordered by due date first, with the gradeables that are due the oldest towards the top and the ones due in the future towards the bottom
Grades without any due date are then ordered after the future most due date gradeables

@bmcutler I believe this matches the description in issue #5316 , does this sound right to you?

PR #5929, opened 2020-10-03
[Bugfix:RainbowGrades] Rainbow grades customization gradeable order

technostarbtd5:
Fixes https://github.com/Submitty/Submitty/issues/5316

Current behavior: In Rainbow Grades instructor UI, gradeables present in a bucket will only be listed in the order they were added to the database. Users can reorder gradeables and the new order will be saved to the rainbow grades customization JSON, but reloading the page will cause gradeables to revert to their old order.

### What is the new behavior?
If a rainbow grades customization JSON is present, gradeables will be ordered based on the JSON ordering. If a rainbow grades customization JSON is not present, gradeables will be ordered by grade release date. 

### Other information?

Tested using both classes containing rainbow grades customization JSON and not containing rainbow grades customization JSON. Also tested with reordering gradeables and adding new gradeables. 

These changes should not change which gradeables are sent to the client. It should only mutate the ordering.

### DIFF
--- a/site/app/models/RainbowCustomization.php
+++ b/site/app/models/RainbowCustomization.php
@@ -124,90 +124,55 @@ class RainbowCustomization extends AbstractModel {
         }
 
         // Reorder buckets
+        $this->reorderBuckets();
+
+        //XXX: Assuming that the contents of these buckets will be lowercase
+        $this->available_buckets = array_diff(self::syllabus_buckets, $this->used_buckets);
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


shailpatels (approved):

Saving buckets works and the order is preserved when manually edited.
Grades are ordered by due date first, with the gradeables that are due the oldest towards the top and the ones due in the future towards the bottom
Grades without any due date are then ordered after the future most due date gradeables

@bmcutler I believe this matches the description in issue #5316 , does this sound right to you?

PR #5930, opened 2020-10-03
[Feature: InstructorUI] Update some polling pages to be GET request instead of POST requests

BenKelly99:
Currently, the following pages use POST requests to transfer the relevant poll ID to each page:
viewPoll
viewResults
editPoll

### What is the new behavior?
Now, these pages use a GET request instead.


shailpatels (changes requested):
Its kind of strange that the URL's are getting changed to `GET`s but the URL isn't being updated.
The poll id should be passed through the URL e.g:

```php
/**
 * @Route("/courses/{_semester}/{_course}/notifications/{poll_id}")
 */
public function showPoll($poll_id) {
```

That way you can also copy and paste the URL directly into the browser 
as  `http://localhost:1501/courses/f20/sample/polls/viewPoll/1` instead of `http://localhost:1501/courses/f20/sample/polls/viewPoll?poll_id=1`

If you know the poll id is always a number you can also have the router enforce that with a requirement regex like `[1-9]*` or `\d*`

```php
/**
 * @Route("/courses/{_semester}/{_course}/notifications/{poll_id}", requirements={"poll_id": "\d*"})
 */
public function showPoll($poll_id) {
```

Is there a reason why you're using forms to send the request? Since each of these buttons is going to a new page an `< a >` tag would make more sense (you can keep the button styling on it) since you have the base url and the poll id anyways in the twig file

shailpatels (approved):

bmcutler:
@vikrsri @shailpatels 
please resolve conflicts

PR #5945, opened 2020-10-07
[Feature:InstructorUI] Add additional notebook builder error checking

Kevin-MM:
There is currently an issue with notebook builder not being able to correctly save/edit files in our production environment.

### What is the new behavior?
This PR adds additional error checking any time a notebook builder file is saved / edited / uploaded.  Error messages are now more verbose which will hopefully allow us to perform a more thorough debugging.

### Other information?
Tested by changing file permissions/ownership.  I used a PHP debugger to pause script execution, allowing me to manually change settings after file generation but before error checks ran.

### DIFF
--- a/site/app/libraries/FileUtils.php
+++ b/site/app/libraries/FileUtils.php
@@ -625,7 +625,7 @@ class FileUtils {
      * @param string $path Absolute path to file or directory
      * @param string|null $expected_owner Expected owner name of the file, or null to omit this check
      * @param string|null $expected_group Expected group name owner of the file, or null to omit this check
-     * @return array Empty array if no errors were detected or an error containing one or more error message strings if
+     * @return array Empty array if no errors were detected or an array containing one or more error message strings if
      *               errors were detected.
      */
     public static function checkForPermissionErrors(string $path, ?string $expected_owner, ?string $expected_group): array {
--- a/site/app/libraries/FileUtils.php
+++ b/site/app/libraries/FileUtils.php
@@ -670,4 +670,30 @@ class FileUtils {
 
         return $results;
     }
+
+    /**
+     * Recursively traverse a directory structure starting at $dir.  The passed in $results array will be populated
+     * with the absolute path to each file or sub-directory.
+     *
+     * NOTE: Sub-directories are treated like files, and thus will be included as their own entries in the array.
+     *
+     * Credit:
+     * https://stackoverflow.com/questions/24783862/list-all-the-files-and-folders-in-a-directory-with-php-recursive-function
+     *
+     * @param string $dir The starting directory (not included in results)
+     * @param array $results An array passed by reference which will be populated
+     */
+    public static function getDirContents(string $dir, &$results = array()): void {
+        $files = scandir($dir);
+
+        foreach ($files as $key => $value) {
+            $path = realpath($dir . DIRECTORY_SEPARATOR . $value);
+            if (!is_dir($path)) {
+                $results[] = $path;
+            } else if ($value != "." && $value != "..") {
+                self::getDirContents($path, $results);
+                $results[] = $path;
+            }
+        }
+    }
 }
--- a/site/app/libraries/FileUtils.php
+++ b/site/app/libraries/FileUtils.php
@@ -610,4 +610,64 @@ class FileUtils {
 
         return $ret;
     }
+
+    /**
+     * Given an absolute path check a file or directory for permission problems.  If any problems are detected return
+     * an array of error strings describing the problems.
+     *
+     * Use this function to check that:
+     * - File or directory at $path exists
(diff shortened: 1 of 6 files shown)


shailpatels (changes requested):
Here is an example of one of the newer error messages:
image


Note I had to rebuild a gradable otherwise, I would keep getting this error message, this might just me not doing that in a while though
Screenshot from 2020-10-08 01-56-30

I do not get a good error message when going to edit an existing file that has broken permissions, I think this is a potential case we should check for 
Here's how I tested
 went to `/var/local/submitty/courses/f20/development/config_upload/7`
I then change the permissions of config.json with `chown root:root config.json`
then I hit edit existing on the submissions/autograding page but all I get is a generic 
" Failure attempting to load the current configuration." error message.

The second test I hit the new button and then once in the notebook builder I edit the permissions of the config file to be owned by root but when I click "save" it happily reports back that my gradeable is being installed. But this is a pretty rare condition I think since if we made it to the builder most of the file paths should be fine, however, checking permissions doesn't seem expensive so we could also add the check here as well

shailpatels:
The code looks good and hopefully this gives additional info on why the notebook builder is failing on production
 @Kevin-MM I consistently see the notebook error "Notebook builder may only edit ... " on this branch any idea why doing a gradeable rebuild might fix it ? I don't see anything different with the config paths its looking for

Kevin-MM:
I've updated the code so that when notebook builder is initializing all files and sub-directories in the autograding directory for the current gradeable are checked for permission errors.  Now we should hopefully be reporting errors anytime a new directory or file is generated, and additionally anytime notebook builder is started.

site/app/libraries/FileUtils.php
```diff
@@ -670,4 +670,30 @@ public static function checkForPermissionErrors(string $path, ?string $expected_
...
+     *
+     * NOTE: Sub-directories are treated like files, and thus will be included as their own entries in the array.
+     *
+     * Credit:
+     * https://stackoverflow.com/questions/24783862/list-all-the-files-and-folders-in-a-directory-with-php-recursive-function
+     *
+     * @param string $dir The starting directory (not included in results)
+     * @param array $results An array passed by reference which will be populated
+     */
+    public static function getDirContents(string $dir, &$results = array()): void {
```
shailpatels: What's the difference between this function and `getAllFiles` which is declared at the top of this file?
They look like they do the same thing

shailpatels (changes requested):

site/app/libraries/FileUtils.php
```diff
@@ -610,4 +610,91 @@ public static function validateUploadedFiles($files) {
...
+     *
+     * NOTE: Sub-directories are treated like files, and thus will be included as their own entries in the array.
+     *
+     * Credit:
+     * https://stackoverflow.com/questions/24783862/list-all-the-files-and-folders-in-a-directory-with-php-recursive-function
+     *
+     * @param string $dir The starting directory (not included in results)
+     * @param array $results An array passed by reference which will be populated
+     */
+    public static function getDirContents(string $dir, &$results = []): void {
```
shailpatels, line 686: Whats the difference between this function and `getAllFiles` defined in fileUtils ? 
https://github.com/Submitty/Submitty/blob/2e1447add11d77a2d703fd6824a0c8d7bd497bf0/site/app/libraries/FileUtils.php#L31

shailpatels (changes requested):

site/app/libraries/FileUtils.php
shailpatels: Ignore the other requested changes, i was looking at an older version

Kevin-MM: The result from `getAllFiles()` doesn't include the paths to sub-directories, only to files in those sub-directories.  It also includes a lot of other meta-data I don't need.  I considered using this function but I would have written just as much, if not more code in the caller to parse the result for the useful information I needed.

shailpatels (approved):
Additional checks work as described

site/app/libraries/FileUtils.php
```diff
@@ -610,4 +610,91 @@ public static function validateUploadedFiles($files) {
...
+    }
+
+    /**
+     * Recursively traverse a directory structure starting at $dir.  The passed in $results array will be populated
+     * with the absolute path to each file or sub-directory.
+     *
+     * NOTE: Sub-directories are treated like files, and thus will be included as their own entries in the array.
+     *
+     * Credit:
+     * https://stackoverflow.com/questions/24783862/list-all-the-files-and-folders-in-a-directory-with-php-recursive-function
```
MasterOdin, line 681: You should use the second answer to this question, not the first: https://stackoverflow.com/a/24784020/4616655

bmcutler:
Hi @Kevin-MM 
What is the status on this PR?

Issue #5924, opened 2020-10-01
Specify Files to Hide From Peer Graders


holzbh:
**What problem are you trying to solve with Submitty**
In the current peer grading implementation, a peer has access to all files in the submission they are currently grading. Instructors may want to require students submit files that are not anonymous or contain information intended for the grading staff, but not peer graders. As it stands, this cannot be done - either the peer graders would see these "confidential" files, or the instructor would need to make a separate gradeable to collect the files - inconvenient for both instructor and students.

**Describe the way you'd like to solve this problem**
The instructor should be able to specify a list of one or more filenames (wildcards are great, but exact matching is sufficient and will probably be faster to implement?) that are not shown to peer graders. In my use case, there is a "README_instructor.pdf" that I would like to be able to read, but I don't want peer graders to see it.

**Describe any potential alternatives you'd tried to solve the problem**
The current "low tech" solution I've used is a gradeable where students submit everything, then as the instructor I run a script I wrote to create .zip files for each student that contains all files EXCEPT the confidential file (if it exists). I then manually make a new gradeable for peer grading, and manually submit each student's .zip on behalf of the student. While I haven't used Submitty's Peer Grading in the past, this 'low tech' process would still work, but it's needlessly tedious and doesn't scale well.

PR #5952, opened 2020-10-12
[TAGrading: Feature] Allow instructors to exempt files from peer view

jchulton:

A peer grader can see all of the files a student submitted

### What is the new behavior?
On the peer matrix tab while creating/editting a gradeable, an instructor can enter wildcards. Filenames that match the attached wildcards will not be shown to the student (but will be shown instructors, Full and limited access TAs) 
image

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -198,6 +198,7 @@ CREATE TABLE public.electronic_gradeable (
     eg_thread_ids json DEFAULT '{}'::json NOT NULL,
     eg_has_discussion boolean DEFAULT false NOT NULL,
     eg_grade_inquiry_start_date timestamp(6) with time zone NOT NULL,
+    eg_hidden_files character varying(1024),
     CONSTRAINT eg_grade_inquiry_due_date_max CHECK ((eg_grade_inquiry_due_date <= '9999-03-01 00:00:00-05'::timestamp with time zone)),
     CONSTRAINT eg_grade_inquiry_start_date_max CHECK ((eg_grade_inquiry_start_date <= '9999-03-01 00:00:00-05'::timestamp with time zone)),
     CONSTRAINT eg_regrade_allowed_true CHECK (((eg_regrade_allowed IS TRUE) OR (eg_grade_inquiry_per_component_allowed IS FALSE))),
@@ -1050,12 +1051,6 @@ CREATE TABLE public.viewed_responses (
     thread_id integer NOT NULL,
     user_id character varying NOT NULL,
     "timestamp" timestamp with time zone NOT NULL
-);
-
-CREATE TABLE public.electronic_gradeable_hidden_files (
-    g_id character varying(255) NOT NULL,
-    file_wildcard character varying(255) NOT NULL,
-    lowest_access_group INTEGER NOT NULL DEFAULT 3
 );
 
 
@@ -2199,14 +2194,6 @@ ALTER TABLE ONLY public.gradeable_access
 --
 ALTER TABLE ONLY public.gradeable_access
     ADD CONSTRAINT gradeable_access_fk3 FOREIGN KEY (accessor_id) REFERENCES public.users(user_id) ON DELETE CASCADE;
---
--- Name: gradeable_access gradeable_access_fk2; Type: FK CONSTRAINT; Schema: public; Owner: -
---
-ALTER TABLE ONLY public.electronic_gradeable_hidden_files
-    ADD CONSTRAINT electronic_gradeable_hidden_files_pk PRIMARY KEY (g_id, file_wildcard);
-
-ALTER TABLE ONLY public.electronic_gradeable_hidden_files
-    ADD CONSTRAINT electronic_gradeable_hidden_files_fk FOREIGN KEY (g_id) REFERENCES public.electronic_gradeable(g_id) ON DELETE CASCADE;
 --
 -- PostgreSQL database dump complete
 --
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -1055,7 +1055,7 @@ CREATE TABLE public.viewed_responses (
 CREATE TABLE public.electronic_gradeable_hidden_files (
     g_id character varying(255) NOT NULL,
     file_wildcard character varying(255) NOT NULL,
-    lowest_access_group INTEGER NOT NULL DEFAULT 3,
+    lowest_access_group INTEGER NOT NULL DEFAULT 3
 );
 
 
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -1052,6 +1052,12 @@ CREATE TABLE public.viewed_responses (
     "timestamp" timestamp with time zone NOT NULL
 );
 
+CREATE TABLE public.electronic_gradeable_hidden_files (
+    g_id character varying(255) NOT NULL,
+    file_wildcard character varying(255) NOT NULL,
+    lowest_access_group INTEGER NOT NULL DEFAULT 3,
+);
+
 
 --
 -- Name: categories_list category_id; Type: DEFAULT; Schema: public; Owner: -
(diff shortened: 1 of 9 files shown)


site/app/models/gradeable/Gradeable.php
```diff
@@ -2005,4 +2005,10 @@ public function isUsingUploadedConfig(): bool {
 
         return !(strpos($this->getAutogradingConfigPath(), $config_upload_path) === false);
     }
+    public function setFilesToOmit() {
```
shailpatels, line 2008: Are theses empty for a reason?

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -33,6 +33,13 @@
                 </div>
             </div>
         </div> 
+        <div class="row">
+            <div class="card col-md-12" style="margin:0px">
+                <div class="content"style="margin:0px">
+                    Enter any file wildcards that should not be viewable by students. Seperate wildcards with commas <br>
+                    <input type="text" id="files_to_omit" name = "files_to_omit" class="auto_save" placeholder="" value = {{hidden_files}}>
```
shailpatels, line 40: use dashes for id's, names', and classes instead of underscores, e.g: `files-to-omit`

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1673,6 +1673,47 @@ public function getAverageForGradeable($g_id, $section_key, $is_team) {
         }
         return new SimpleStat($this->core, $this->course_db->rows()[0]);
     }
+
```
shailpatels, line 1676: Can you add PHP type hints to all of these functions, it will make static analysis perform better wherever they are used.

shailpatels (changes requested):

site/app/models/gradeable/Gradeable.php
jchulton: How we handle gradeable properties means I need setter and getter functions, otherwise an error is thrown. They are never actually used

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
jchulton: Changing to files-to-omit means the name of the setters and getters would have to be Files-To-Omit setFiles-to-omit()/ getFiles-to-omit(). I think it would be better to just use files_to_omit.

site/app/libraries/database/DatabaseQueries.php
jchulton: Added the PHP type hints

site/app/models/gradeable/Gradeable.php
emaicus: Jumping in a little late, but is there a reason that the value for `filesToOmit` isn't stored as a class variable here?

jchulton: In the database, I added a table that contains each wildcard, the gradable, and the level of permission that cannot see the file (only 4 rn). I think this a more versatile implementation that allows for easy changing down the line. However, it does mean that we have to handle filesToOmit differently, so storing it as a class variable wouldn't add anything (I think)

emaicus (changes requested):
It looks like the migration is failing at the moment per Travis. Please fix that and the issues that I mention in this review, then we should be good to merge.

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1180,23 +1180,34 @@ public function setAnonPath($file_path) {
...
                 foreach ($new_files as $file) {
-                    if ($start_dir_name == "submissions") {
-                        $file["path"] = $this->setAnonPath($file["path"]);
+                    $skipping = false;
+                    if ($this->core->getUser()->getGroup() == User::GROUP_STUDENT) {
+                        foreach (explode(',', $hidden_files) as $hidden_file) {
+                            if (fnmatch($hidden_file, $file["name"])) {
+                                $skipping = true;
+                            }
+                        }
```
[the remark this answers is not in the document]
jchulton: Correct, made the change

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1180,23 +1180,34 @@ public function setAnonPath($file_path) {
...
                     }
-                    $path = explode('/', $file['relative_name']);
-                    array_pop($path);
-                    $working_dir = &$files[$start_dir_name];
-                    foreach ($path as $dir) {
-                        if (!isset($working_dir[$dir])) {
-                            $working_dir[$dir] = [];
+                    if (!$skipping) {
```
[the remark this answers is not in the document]
jchulton: They could, I've changed Access.php to prevent this

migration/migrator/data/course_tables.sql
```diff
@@ -1052,6 +1052,12 @@ CREATE TABLE public.viewed_responses (
     "timestamp" timestamp with time zone NOT NULL
 );
 
+CREATE TABLE public.electronic_gradeable_hidden_files (
+    g_id character varying(255) NOT NULL,
+    file_wildcard character varying(255) NOT NULL,
+    lowest_access_group INTEGER NOT NULL DEFAULT 3,
```
emaicus, line 1058:
```suggestion
    lowest_access_group INTEGER NOT NULL DEFAULT 3
```

jchulton: Made the above change

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
shailpatels: CSS ids should use dashes see the style guide : https://submitty.org/developer/coding_style_guide/css,
its fine to leave for now if there's JS logic depending on the ID using underscores but in the future new code should stick to using dashes instead

shailpatels (dismissed):
I think this PR is ready to merge.

I tested the following file wild cards : *, ?, [ ] and they worked as intended correctly blocking student files. Exact matches against specific filenames work too, matches are capitalization sensitive. Input box seems to handle putting spaces or no spaces between comma separated entries

Students do not see any file that has a name that matches a wildcard as expected, they cannot URL hack either. There is no identifying information in the file structure. Assigned graders are able to see all files uploaded by a student.

shailpatels:
This PR refers to issue  #5924

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1180,23 +1180,34 @@ public function setAnonPath($file_path) {
      * @return string by reference
      */
     public function renderSubmissionPanel(GradedGradeable $graded_gradeable, int $display_version, bool $showNewInterface) {
-        $add_files = function (&$files, $new_files, $start_dir_name) {
+        $add_files = function (&$files, $new_files, $start_dir_name, $graded_gradeable) {
             $files[$start_dir_name] = [];
+            $hidden_files = $this->core->getQueries()->getOmmitedFiles($graded_gradeable->getGradeableId());
```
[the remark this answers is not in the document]
jchulton: Hidden_files is now an auto-updating property of Gradeable, and its value can be got by using an associated get method (getHiddenFiles) on Gradeable.php

shailpatels (changes requested):
This PR works correctly for the wildcards `* ? [ ]` 
Students cannot URL hack and there is no existing information about the student in the file submission browser. Exact matches work as well

However adding a space between the comma delimiter and the wildcard match do not save even though the page says it sucessfully saved.

To reproduce:
Enter a string like `match_1, match_2` 
Make sure there is a space between the wildcards
Hit enter to save and refresh the page, the second wildcard is not there.

When I first tested this I thought the PR was broken because I entetered many wildcards and they weren't working, when I went back to the peer grading matrix page I saw they were all gone with no indication.

This should be handled better since it may lead to an instructor thinking they've hidden files when really they didn't

jchulton:
I've added some input processing to remove all leading and trailing spaces between commas. Upon testing, this fixes the above error.

shailpatels (changes requested):
A similar issue still persists:
Enter a string of multiple wildcards with spaces in between: e.g: `test1,   test2,   test3`
Hit enter and confirm the page says "all changes saved"
Refresh the page
The third wildcard has disappeared without any information on why

jchulton:
Fixed the conflicting files. I can fix the bug pointed out by @shailpatels in a future PR. For Buster's purposes, this should be good to merge.

shailpatels (approved):
Approved to get this merged.
For instructors using this, avoid using spaces for now but having a single exact match against a filename works fine

emaicus (approved):
Tested by omitting `README.txt`. The file was properly omitted for peers, and was properly _not_ omitted for full access graders or instructors. Also tested the down migration.

Everything looks good!

PR #5959, opened 2020-10-15
[Bugfix:OfficeHours] Fix updating queue dynamically 

shailpatels:

PR https://github.com/Submitty/Submitty/pull/5950 introduced a bug where the office hours queue would say "request type not supported" when the queue would dynamically update if a student entered/left.

This was because pr 5950 forced ajax requests to receive JSON responses back, but the office hours queue was using ajax to grab the new html and updating the dom via javascript, preventing a full page refresh. There were no checks of the response type so whatever was sent back was displayed to the user.

### What is the new behavior?
Since developers might be interested in receiving raw HTML from a controller in scenarios like this the router will
will now by default send back whatever the controller that was called returned. If the client wants to enforce a JSON request they can set the "accepts" headers to "text/json". I think this is a good middle ground

### DIFF
--- a/site/public/index.php
+++ b/site/public/index.php
@@ -126,11 +126,12 @@ if ($is_api) {
 }
 else {
     $response = WebRouter::getWebResponse($request, $core);
-}
+    $request_json = isset($_SERVER['HTTP_ACCEPT']) && $_SERVER['HTTP_ACCEPT'] === "text/json";
 
-if ($request->isXmlHttpRequest() && ($response instanceof MultiResponse)) {
-    //If its an Ajax based request, we can only send a JSON payload back
-    $response = $response->convertToJsonResponse();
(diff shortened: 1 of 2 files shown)


site/public/index.php
```diff
@@ -126,11 +126,12 @@ function error_handler() {
 }
 else {
     $response = WebRouter::getWebResponse($request, $core);
-}
+    $request_json = isset($_SERVER['HTTP_ACCEPT']) && $_SERVER['HTTP_ACCEPT'] === "text/json";
```
MasterOdin, line 129: `application/json` is the official mime-type for JSON per [RFC 4627](https://www.ietf.org/rfc/rfc4627.txt).

vikrsri (approved):
The changes correctly fix the described bug.

PR #5963, opened 2020-10-18
[Bugfix:InstructorUI] Prevent access to course materials before they are released

Kevin-MM:
Students may access course materials prior to the course materials release date, provided they obtain a url to the file.

### What is the new behavior?
This behavior has been corrected and students may no longer access course materials that haven't yet been released.

### Other information?
This bug was the result of PHP being unable to correctly parse the release time string into a DateTime object.  When this condition occurs instead of throwing an exception the DateTime library will simply return `false`.  The `false` was then used in a comparison causing `true` to always be returned.

The parsing bug has been resolved.  Additionally the return type is now checked so that if a parsing error occurs an exception is now thrown.

### DIFF
--- a/site/app/models/CourseMaterial.php
+++ b/site/app/models/CourseMaterial.php
@@ -3,6 +3,7 @@
 namespace app\models;
 
 use app\exceptions\FileNotFoundException;
+use app\exceptions\MalformedDataException;
 use app\libraries\Core;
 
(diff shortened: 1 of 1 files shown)


shailpatels (approved):

I was able to reproduce the issue on master, this PR fixes it properly. URL hacking now brings the student to the "you do not have access" page. Tested with single files and zips.

PR #5965, opened 2020-10-20
[UI/UX:InstructorUI] Office Hours Queue UI Changes

vikrsri:

The modal to edit queues is currently visually straining. It is difficult to tell whether a specific queue is open or closed at a glance: 
image

### What is the new behavior?
Changes to the modal: the colors have been removed, the checkbox to hide/show a queue has been moved to the main page, and the open/close button for opening/closing queues have been changed to checkboxes so that the user can more quickly determine if a certain queue is open/closed:
image

Changes to the main queue page: The colors of the queues have been changed to a lower saturation to be easier on the eyes, and buttons to show/hide queues have been made.

This is what it looks like when all the queues are shown:
image

This is what it looks like when Queue 6 is hidden. The Queue 6 button becomes colorless to indicate that it is hidden:
image

image


site/app/templates/officeHoursQueue/QueueFilter.twig
```diff
@@ -29,45 +29,24 @@
...
-              <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
-              <input type="hidden" name="queue_code" value="{{queue['code']}}"/>
-              <input type="hidden" name="queue_state" value="{{queue['open']}}"/>
-              {% if queue['open'] == 0 %}
-                <button type="submit" class="btn btn-primary filter_settings_btn open_queue_btn">Open</button>
-              {% else %}
-                <button type="submit" class="btn btn-primary filter_settings_btn close_queue_btn">Close</button>
-              {% endif %}
-            </form>
+              <input type="checkbox" id="toggle_queue_{{loop.index}}" onchange="toggleQueue({{loop.index}}, '{{queue['code']}}')" {% if queue['open'] %} checked {% endif %} />
```
shailpatels, line 49: CSS Ids and classes should use dashes instead of underscores,
see our style guide here: https://submitty.org/developer/coding_style_guide/css

If there is JS logic attached to this id and its too much work to refactor in this PR we can leave it for now

site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -228,5 +234,18 @@
       });
   }
 
+  function toggleQueue(index, code){
+      console.log($(`#toggle_queue_${index}`)[0].checked)
```
shailpatels, line 238: is this console log required?

site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -228,5 +234,18 @@
...
+      $.post( `{{base_url}}/${code}/toggle`, {
+          csrf_token: '{{ csrf_token }}',
+          queue_code: code,
+          queue_state: $(`#toggle_queue_${index}`)[0].checked ? 0:1
+      });
+  }
+
+  function refreshPage(){
```
shailpatels, line 246: I don't think we need to wrap this in a function, since its only being used in that one place.
You can just do a `onclick="window.location.reload()"`

site/app/templates/officeHoursQueue/QueueHeader.twig
```diff
@@ -20,6 +20,43 @@
         Managing the Office Hours Queue
         <i style="font-style:normal;" class="fa-question-circle"></i>
       </a>
+        <br>
+        <br>
+        Open Queues:
+        {% for queue in viewer.getAllQueues() %}
+            {% if queue['open'] != 0 %}
+
+                <label tabindex="0" class="btn filter-buttons"  style="word-wrap: break-word; border-color: {{ viewer.getColorFromCode(queue['code']) }}; background: {{ viewer.getColorFromCode(queue['code']) }}; color: inherit; {# color: {{ viewer.getColorFromCode(queue['code']) }}; #}">{{queue['code']}}
```
shailpatels, line 29: is the commented out code still needed?

shailpatels (changes requested):

shailpatels:
It looks like the Office hours e2e tests are actually failing due to some of the queue names getting changed on the sidebar and q interface

shailpatels (changes requested):
The newer changes look good, if there are many queues or qs with very long names they stack vertically, I think this is fine.

A couple of accessibility issues, modals should be able to be closed when hitting the "escape" button, the easist way to do this is probably add an event listener for the close key down event and do a `.click()` on the close button.

You can also see what the generic popup template is doing to solve this
https://github.com/Submitty/Submitty/blob/master/site/app/templates/generic/Popup.twig

Also each checkbox should have a label associated with it sine its an input. This is for screen readers. See https://submitty.org/developer/interface_design_style_guide/web_accessibility and
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

shailpatels (changes requested):
There is no limit to the length of the queue name, I created a queue with a name that had a length of 58310 characters and could no longer close the queue or delete it. I kept seeing errors every time I did something with it.

It looks like the delete q is sending the information as a url parameter, either this should be moved into the post body or there should be a `maxlength` attribute on the textareas, creating a client side restriction of under 8000 characters for apache - but we should restrict it under 2048 if the data is getting added to a url.

shailpatels (approved):

PR #5966, opened 2020-10-20
[Bugfix:Autograding] Stop subsequent testcase failure on docker errors

emaicus:
At present if an error occurs as the result of a failed docker operation (e.g, container creation), any hanging containers are not cleaned up. This means that subsequent testcases fail due to docker name collisions.

This PR adds container cleanup before testcases are run, intermediately between each testcase, and after execution has finished.

### DIFF
--- a/autograder/autograder/execution_environments/container_network.py
+++ b/autograder/autograder/execution_environments/container_network.py
@@ -792,8 +792,8 @@ class ContainerNetwork(secure_execution_environment.SecureExecutionEnvironment):
             self.log_stack_trace(traceback.format_exc())
             self.log_message("ERROR: Could not verify execution mode status.")
             return
-
-        autograding_utils.cleanup_stale_containers(self.untrusted_user)
+        finally:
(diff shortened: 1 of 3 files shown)


autograder/autograder/execution_environments/container_network.py
```diff
@@ -793,6 +793,8 @@ def execute_helper(self, containers, script, arguments, logfile):
             self.log_message("ERROR: Could not verify execution mode status.")
             return
 
+        autograding_utils.cleanup_stale_containers(self.untrusted_user)
```
hdrodz97, line 796: Is there any reason this isn't inside a `finally` block with the previous `try`...`except`?

hdrodz97 (commented):
Overal seems to work well. Only have one minor comment and then I'll feel good about approving this!

bmcutler:
@emaicus 
Status on this PR?

autograder/autograder/execution_environments/container_network.py
emaicus: Moved.

PR #5968, opened 2020-10-22
[Bugfix:InstructorUI] Course materials fix hide from students option

Kevin-MM:
When working with course materials as an instructor, there is a file option to 'Hide from students'.  In testing this option didn't appear to do anything.

### What is the new behavior?
This behavior has been corrected so that when this option is selected the file will not be visible to students (but will still be accessible via a link, as described in the options documentation).

### DIFF
--- a/site/app/models/CourseMaterial.php
+++ b/site/app/models/CourseMaterial.php
@@ -3,6 +3,7 @@
 namespace app\models;
 
 use app\exceptions\FileNotFoundException;
+use app\exceptions\MalformedDataException;
 use app\libraries\Core;
 
 class CourseMaterial extends AbstractModel {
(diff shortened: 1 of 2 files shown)


shailpatels (changes requested):
Looks better, before files would always appear 
Now files are not displayed but download links for files and folders work correctly when copied into to console or browser

However, the popup url does not seem to work.
I uploaded a file and check the hide but allow URL access, I right-clicked and copied the href URL for the popup button for the file but kept getting the you do not have access to this file.

Even pressing the button does not work, maybe the url is broken?

Kevin-MM:
@shailpatels I'm having trouble replicating the behavior you are describing.  In my testing everything works correctly in the scenario you are describing.  We just recently fixed a bug related to file release time.  Are you sure you have set the release time to 'now'?  Also ensure your submitty timezone and PC timezone are the same before attempting to set a release time.

shailpatels (approved):
@Kevin-MM Approving now, after testing again on a clean vagrant this PR works as expected. 
Whatever was happening before might of been some weird state with timezones or user error like you mentioned

PR #5994, opened 2020-11-01
[Bugfix:System] Fix install_system for initial vagrant up

shailpatels:

CGI scripts are not able to run after an initial vagrant up. This will prevent users from logging in to the sample system since it's based on pam authentication. The existing suexec config file is not updated if it already exists, this prevents apache from executing any cgi script since it thinks the docroot is located somewhere else.

Apache was most likely creating default files before the install_system script could copy the correct versions over.

### What is the new behavior?
Removes the `-n` flags so the install_system will clobber the submitty.conf and suexec/www-data with whatever is on the host system.
This should make sure those files are correct on initial up.

### Other information?
If you have this issue you might be able to skip a vagrant destroy/up with the following commands:
```bash
cp ${SUBMITTY_REPOSITORY}/.setup/apache/www-data  /etc/apache2/suexec/www-data
chmod 0640 /etc/apache2/suexec/www-data
service apache2 restart
```

### DIFF
--- a/.setup/install_system.sh
+++ b/.setup/install_system.sh
@@ -484,7 +484,7 @@ EOF
         fi
     fi
 
-    cp ${SUBMITTY_REPOSITORY}/.setup/php-fpm/pool.d/submitty.conf /etc/php/${PHP_VERSION}/fpm/pool.d/submitty.conf
+    cp -n ${SUBMITTY_REPOSITORY}/.setup/php-fpm/pool.d/submitty.conf /etc/php/${PHP_VERSION}/fpm/pool.d/submitty.conf
     cp ${SUBMITTY_REPOSITORY}/.setup/apache/www-data /etc/apache2/suexec/www-data
     chmod 0640 /etc/apache2/suexec/www-data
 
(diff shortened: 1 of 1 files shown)


emaicus:
Hi @shailpatels, as I recall, I added the `-n` option to save users from clobbering local edits to those config files on production machines. Is this only an issue in vagrant? If so can we add an if/else to avoid clobbering files on production?

shailpatels:
@emaicus yeah this most likely a vagrant only issue. Since the config for the our sample submitty was not getting copied over

shailpatels:
@emaicus I thought about this issue a bit more and added back the `-n` for submitty.conf but I think the `www-data` file should always be copied over from whatever is in source control.

The only two lines in that file that do anything are:
```
/usr/local/submitty/site/
cgi-bin
```
which is the location of the root of the document tree for the website and the location of the CGI files. The first value probably doesnt matter but the second has to be "cgi-bin" otherwise apache won't execute any cgi scripts. I can't see a situation where someone's configuration would need to change this without basically changing the source code of Submitty.

emaicus (approved):
Approved.

Issue #5912, opened 2020-09-24
Polling: Existing Feature Improvements


BenKelly99:
Here are a list of existing features that can be improved for polling:

- [ ] Overall UI improvements. The UI is messy and can be improved, especially for the new/edit polls, and for student view for answering a poll.

- [x] Currently, an instructor can only open a poll once. That is, once it is opened for students to see, it cannot be closed. Instructors should have the ability to set a poll to any of three states at any time. Closed: cannot be viewed or answered by students. Open: Can be viewed and answered. Ended: Can be viewed, but no longer answered.

- [x] Currently, the viewPoll, editPoll, and viewResults page all are POST requests whose payloads contain the poll ID to show. This should be GET requests instead. 

- [x] Currently, when an instructors submits edits or a new poll and there is a part that is an invalid setup, it will simply kick the instructor back to the all polls page, and discard their changes. This is very frustrating so instead if there is an invalid part it should kick them back to the page they were on, retaining their changes, but not updating the polls in the database at all.

Note: if fixing one of these issues, do not use the "Closes"  tag unless the PR resolves all remaining issues. I will put check marks next to the issues as they are fixed.

Harshalzzzzzzz:
@BenKelly99  Dear Mentor ,
I would like to work on this issue, can you please guide me further on what needs to be done and how I can approach this issue and contribute to the project?
I have already completed the setup.

Thank you

Harshalzzzzzzz:
@BenKelly99 Dear Mentor, 

Although both the issues interest me, I would like to work on this and then move on to the other if possible. Can you please guide me further? I'd like to get started on this.

BenKelly99:
I would choose one of the bullet points on one of the issues, and start on that. Which one looks good to you? In general, most of the polling work is done in PollController.php, PollView.php, PollModel.php, the twig templates in site/app/templates/polls, and the database stuff is in DatabaseQueries.php

sbelsk:
@BenKelly99 The only bullet point left on the list is "Overall UI improvements". What, in your opinion, makes the student poll view and instructor new/edit poll UI's messy that can be improved? Perhaps it's been so long that this was handled and is no longer relevant... :)

BenKelly99:
[...]

PR #6003, opened 2020-11-07
[Feature:Submission] Three States Poll Setting

listeph:

Works on #5912 2nd bullet -- "Currently, an instructor can only open a poll once. That is, once it is opened for students to see, it cannot be closed. Instructors should have the ability to set a poll to any of three states at any time. Closed: cannot be viewed or answered by students. Open: Can be viewed and answered. Ended: Can be viewed, but no longer answered."
Instructor Interface:
old_instructor
Corresponding student interface:
student

### What is the new behavior?
Instead of a toggle button to move polls between open / closed / ended states, now there are two checkbox columns (visible, accepting answers) to toggle the polls between the states. These checkboxes have more intuitive meaning to the user than the previous toggle button states.
New Instructor Interface (corresponding to the same student interface above):
instructor

### Other UI Changes
- Polls no longer jump between future / past / today's polls categories by being opened / ended
     - Future polls are not visible or accepting answers by default (but their release dates can be changed to today in "Edit Polls" if necessary)
- Old Polls and Future Polls now display a release date column
- Polls are no longer ordered by id
    - Today's Polls are ordered alphabetically by poll name
    - Future / Past polls are ordered by most recent date first, then alphabetically by poll name
- Dropdown states are now saved in cookies. If the instructor opens (or closes) a dropdown, and leaves the page then returns or refreshes, the dropdown is still opened (or closed). If there's no saved cookie state, the today polls are open by default, and past and future polls are closed by default.

### DIFF
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -27,13 +27,23 @@ class PollController extends AbstractController {
      */
     public function showPollsPage() {
         if ($this->core->getUser()->accessAdmin()) {
+            // Check if we have a saved cookie session with the dropdown states for each of the instructor polls sections
+            $dropdown_states = ['today' => true, 'old' => false, 'future' => false];
+            foreach ($dropdown_states as $key => $val) {
+                $cookie_key = $key . "_polls_dropdown";
+                if (array_key_exists($cookie_key, $_COOKIE)) {
+                    $dropdown_states[$key] = $_COOKIE[$cookie_key] === 'true';
+                }
+            }
+
             return MultiResponse::webOnlyResponse(
                 new WebResponse(
                     'Poll',
                     'showPollsInstructor',
                     $this->core->getQueries()->getTodaysPolls(),
                     $this->core->getQueries()->getOlderPolls(),
-                    $this->core->getQueries()->getFuturePolls()
+                    $this->core->getQueries()->getFuturePolls(),
+                    $dropdown_states
                 )
             );
         }
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -196,9 +196,7 @@ class PollController extends AbstractController {
     public function closePoll() {
         if (!isset($_POST["poll_id"])) {
             $this->core->addErrorMessage("Invalid Poll ID");
-            return MultiResponse::RedirectOnlyResponse(
-                new RedirectResponse($this->core->buildCourseUrl(['polls']))
-            );
+            return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
         $this->core->getQueries()->closePoll($_POST["poll_id"]);
 
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -205,10 +205,10 @@ class PollController extends AbstractController {
     }
 
     /**
-    * @Route("/courses/{_semester}/{_course}/polls/updateDropdownStates", methods={"POST"})
+     * @Route("/courses/{_semester}/{_course}/polls/updateDropdownStates", methods={"POST"})
      * @AccessControl(role="INSTRUCTOR")
      * @return JsonResponse
-    */
+     */
     public function updateDropdownStates() {
         $user_id = $this->core->getUser()->getId();
         $expire_time = time() + (7 * 24 * 60 * 60); // 7 days from now
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -10,6 +10,7 @@ use app\libraries\response\RedirectResponse;
 use Symfony\Component\Routing\Annotation\Route;
 use app\libraries\routers\AccessControl;
 use app\libraries\DateUtils;
+use app\libraries\Utils;
 use app\libraries\routers\FeatureFlag;
 
 /**
@@ -201,6 +202,18 @@ class PollController extends AbstractController {
         $this->core->getQueries()->closePoll($_POST["poll_id"]);
 
         return new RedirectResponse($this->core->buildCourseUrl(['polls']));
+    }
+
+    /**
+    * @Route("/courses/{_semester}/{_course}/polls/updateDropdownStates", methods={"POST"})
+     * @AccessControl(role="INSTRUCTOR")
+     * @return JsonResponse
+    */
+    public function updateDropdownStates() {
+        $user_id = $this->core->getUser()->getId();
+        $expire_time = time() + (7 * 24 * 60 * 60); // 7 days from now
+        Utils::setCookie($_POST["cookie_key"], $_POST["new_state"], $expire_time);
+        return JsonResponse::getSuccessResponse($_COOKIE[$_POST["cookie_key"]]);
     }
 
     /**
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -189,6 +189,25 @@ class PollController extends AbstractController {
     }
 
     /**
+     * @Route("/courses/{_semester}/{_course}/polls/setClosed", methods={"POST"})
+     * @AccessControl(role="INSTRUCTOR")
+     * @return MultiResponse
+     */
+    public function closePoll() {
+        if (!isset($_POST["poll_id"])) {
+            $this->core->addErrorMessage("Invalid Poll ID");
+            return MultiResponse::RedirectOnlyResponse(
+                new RedirectResponse($this->core->buildCourseUrl(['polls']))
+            );
+        }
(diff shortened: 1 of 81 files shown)


site/app/controllers/PollController.php
```diff
@@ -188,6 +188,25 @@ public function endPoll() {
...
+    /**
+     * @Route("/courses/{_semester}/{_course}/polls/setClosed", methods={"POST"})
+     * @AccessControl(role="INSTRUCTOR")
+     * @return MultiResponse
+     */
+    public function closePoll() {
+        if (!isset($_POST["poll_id"])) {
+            $this->core->addErrorMessage("Invalid Poll ID");
+            return MultiResponse::RedirectOnlyResponse(
+                new RedirectResponse($this->core->buildCourseUrl(['polls']))
```
shailpatels, line 200: If you're sending back only one type of response, you can just do 
```php
return new RedirectResponse( ...
```
Remember to edit the phpdocstring `@return` or remove it, otherwise phpstan will complain

shailpatels (changes requested):

site/app/controllers/PollController.php
listeph: @BenKelly99  did that for most functions in PollController.php so I'm scared to change them all in case it was intended to be like that

shailpatels: That's fine, we can refactor the older code later as long as the new stuff we're adding matches our up to date functions

BenKelly99: Yeah when I made this it was right when a lot of changes were made to the responses. I can refactor it later

site/app/libraries/Core.php
```diff
@@ -730,6 +730,23 @@ public function isWebLoggedIn(): bool {
...
+     * @return array list of three booleans for dropdown states
+     */
+    public function getPollsDropdownStates(): array {
+        $ret = [true, false, false]; // corresponding to default dropdown states for (today, old, future)
+        $cookie_keys = ['today_polls_dropdown', 'old_polls_dropdown', 'future_polls_dropdown'];
+        for ($i = 0; $i < count($cookie_keys); $i++) {
+            if (array_key_exists($cookie_keys[$i], $_COOKIE)) {
+                $ret[$i] = $_COOKIE[$cookie_keys[$i]] === 'true';
+            }
+        }
```
listeph, line 746: Maybe I should put this code directly in PollView.php? I just saw some other cookie functions defined in Core.php so I put it here

site/app/libraries/Core.php
```diff
@@ -730,6 +730,23 @@ public function isWebLoggedIn(): bool {
         return $logged_in;
     }
 
+    /**
+     * Check if we have a saved cookie session with the dropdown states for each of the instructor
+     * polls sections (today, old, future), otherwise sets it to the default dropdown states (true, false, false).
+     *
+     * @return array list of three booleans for dropdown states
+     */
+    public function getPollsDropdownStates(): array {
```
listeph, line 739: Maybe I should put this code directly in PollView.php .. I put it in Core.php because I saw other cookie functions here , but maybe it's unnecessary.

shailpatels: Yeah I would add this in the PollController and pass the information onto the view.
You can also have a dictionary which is easier to understand than an array of booleans:
```php
$ret = [
   'today' => true
   'old' => false
   'future' => false
];
```

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -29,31 +34,12 @@
...
-                            <td>
-                                <form method="post" action="{{ base_url }}/setOpen">
-                                    <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
-                                    <input type="hidden" name="poll_id" value="{{ poll.getID() }}"/>
-                                    <button type="submit" class="btn btn-primary">Reopen Poll</button>
-                                </form>
-                            </td>
-                        {% endif %}
+                        <td>
+                            <input class="key_to_click" tabindex="0" type="checkbox" id="poll_{{ poll.getID() }}_visible" aria-label="label" onclick='updatePollVisible("{{ poll.getID() }}")' {{ (poll.isOpen() or poll.isEnded) ? 'checked' : '' }}>
```
shailpatels, line 38: You can remove the `key_to_click` class here, HTML5 checkboxes will allow users to select them by tabbing and toggling their checkboxes with the enter key by default

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -29,31 +34,12 @@
...
+                        </td>
+                        <td>
+                            <input class="key_to_click" tabindex="1" type="checkbox" id="poll_{{ poll.getID() }}_view_results" aria-label="label" onclick='updatePollAcceptingAnswers("{{ poll.getID() }}")' {{ poll.isOpen ? 'checked' : ''}}>
```
shailpatels, line 41: Same comment as above and remove the tabindex, we should default to the natural input order in this case since the inputs are nested within a table.
A tabindex of 0 would give the element a tab order based on its location in the DOM, a tab index of 1 would hard code it as the first element to be selected when hitting tab. This causes the user to jump to the "Accepting Answers" input even though they should start at the input before.

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -111,25 +101,13 @@
...
-                                    <button type="submit" class="btn btn-primary">Open Poll</button>
-                                </form>
-                            </td>
-                        {% endif %}
+                        <td> {{ poll.getReleaseDate() }} </td>
+                        <td>
+                            <input class="key_to_click" tabindex="0" type="checkbox" id="poll_{{ poll.getID() }}_visible" aria-label="label" onclick='updatePollVisible("{{ poll.getID() }}")' {{ (poll.isOpen() or poll.isEnded) ? 'checked' : '' }}>
+                        </td>
+                        <td>
+                            <input class="key_to_click" tabindex="1" type="checkbox" id="poll_{{ poll.getID() }}_view_results" aria-label="label" onclick='updatePollAcceptingAnswers("{{ poll.getID() }}")' {{ poll.isOpen ? 'checked' : ''}}>
```
shailpatels, line 109: same comments as above

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -220,3 +192,85 @@
...
+    }
+
+    function updateDropdownStates(curr_state, cookie_key) {
+        let csrf_token = $("#csrf_token").val();
+        let url = $("#base_url").val() + "/updateDropdownStates";
+        let fd = new FormData();
+        fd.append('csrf_token', csrf_token);
+        fd.append('new_state', !curr_state);
+        fd.append('cookie_key', cookie_key)
+        console.log(cookie_key);
```
shailpatels, line 261: is this console log required?

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -1,25 +1,30 @@
 <div class="content">
+    <input type="hidden" id="csrf_token" value="{{ csrf_token }}"/>
```
shailpatels, line 2: There should be a globally defined `csrfToken` JS variable you can use instead of passing it in again here. You can do a console.log to confirm

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -1,25 +1,30 @@
...
+    <input type="hidden" id="base_url" value="{{ base_url }}"/>
```
shailpatels, line 3: is this needed? If all of your JS in also within twig files then you can use the same `"{{ base_url }}"` to inject the value you need into the JS code, remember the quotes if you do do this to make sure the twig replacement expands to valid JS code.

shailpatels (changes requested):

site/app/templates/polls/AllPollsPageInstructor.twig
shailpatels: Oh sorry I meant you can remove both tabindexes and the browsers default ordering will work here.

I meant to say tabindex 1 makes it so when the user presses tab they will be brought to that element which is not what we want since there are other element's on the page that should be visited first (like the skip to main content section)

listeph: Okay, thanks for the knowledge! I removed both tab indexes then.

vikrsri (approved):
Changes word as proposed - tested adding/deleting polls, switching from visible to not visible and accepting answers to not accepting answers. Also tested changing dates and ordering of polls.

shailpatels (approved):

PR #6008, opened 2020-11-09
[Feature:Testing] Initial work on migrating from TravisCI to Github Actions

shailpatels:

Travis CI doesn't have a free plan for open source projects anymore

### What is the new behavior?
Creates a new github Actions workflow which performs the following:
- JS linting/static analysis (eslint)
- php linting by (phpcs)
- php static analysis (phstan)
- php unit tests (phpunit)
- python3 linting (flake8)
- python3 unit tests (unittest)
    - note autograding unit tests are broken right now

This PR also rolls back phstan to an earlier version, the newer one has added new checks that are now failing
unsure why it originally passed travis - might of been before we fixed it 

### Other information?
Will leave these steps in travis for now until everything is moved over

submitty github workflow

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,7 @@ jobs:
             - run: npm install
             - name: Run eslint
               run: npm run lint
-    php:
+    php-static:
         runs-on: ubuntu-latest
         defaults:
             run:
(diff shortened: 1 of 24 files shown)


MasterOdin:
You will probably want to use the [`actions/setup-node`](https://github.com/actions/setup-node), [`actions/setup-python`](https://github.com/actions/setup-python), and [`shivammathur/setup-php`](https://github.com/shivammathur/setup-php) actions for setting the specific versions of a language for the individual tests. This helps to avoid the runners breaking if GH decides to modify what's in a runner, and the lattermost action features some useful bits for adding in extensions (like imagemagick).

Additionally, given your "TODO" message, unfortunately, github does not feature a way to define blocks for reuse like Travis did as GH only utilizes a subset of the full yaml spec. Given the speed of running the linters, it's probably easier to just combine all of the language stuff into a singular build stage, and then if you really wanted to replicate Travis, use conditionals on language version to trigger or ignore specific run commands.

shailpatels:
@MasterOdin Thanks for the info, yeah it makes sense we would want to explicity control the language versions in the runners. And ok for somereason I thought actions could define blocks that could be called from jobs but looks like thats not a feature.

I think for now it would be good to see how the testing suite works split up into seperate jobs even though it does repeat some of the same setup steps but since GH actions allows more parallel jobs than travis it might be worth it.

Issue #5884, opened 2020-09-14
standardize markdown css for messages, etc. throughout submitty


bmcutler:
We have some css settings (applied after reset.css sets everything to zero) to improve the appearance of rendered markdown in various places, but we are missing some of the basic formatting in numerous places.

Let's make a standard markdown.css (or similar) and use it everywhere.

places we render markdown (probably incomplete):

* gradeable assignment message (blue box)

* queue welcome message (blue box)
<img width="164" alt="Screen Shot 2020-09-14 at 10 46 58 AM" src="https://user-images.githubusercontent.com/5871417/93100851-a059c400-f677-11ea-938b-a43b67feca0a.png">

* queue announcements (grey box)
<img width="95" alt="Screen Shot 2020-09-14 at 10 47 37 AM" src="https://user-images.githubusercontent.com/5871417/93100947-b8314800-f677-11ea-913e-1e7c5c33854c.png">

* forum posts

* notebook

* ?

places we should render markdown:

* Submission upload message (url wasn't handled in standard markdown way)

* ?

Things that are often overlooked/misformatted:

* bulleted/numbered list is rendered with insufficient left indentation

* insufficient vertical space between paragraphs

* insufficient vertical space after the last item of a bulleted list

* when we are on the dark blue background, the urls shouldn't be rendered with blue text

* no planning for or poor contrast when we switch to dark mode.

sidpaswan:
I would like to work on this issue. Enlist me this issue and as i am new to this repository please guide me.
Thank You

bmcutler:
@Kevin-MM 
Another place to wrap nicely formatted markdown...  around the display of the "Grade Inquiry Warning Message".
Currently it does not render as markdown.

bmcutler:
When we display the notebook for a specific student in the new TA grading interface, there is no padding/margin to the left of the content.

Probably other small things to tweak too... 

<img width="101" alt="Screen Shot 2020-10-24 at 12 49 28 PM" src="https://user-images.githubusercontent.com/5871417/97087348-60cbb380-15f7-11eb-9b91-135d13d823c4.png">

PR #6011, opened 2020-11-10
[Refactor:UI] Standardize markdown for gradeable assignment/load message

Kevin-MM:
Partially #5884 

Gradeable assignment messages and gradeable load messages do not inherit markdown styling rules from the new standard set of markdown styling rules.

### What is the new behavior?
Markdown rendered for these two spots in submitty now will be styled according to the recently implemented markdown.css.

Assignment message screenshot:
image


vikrsri (approved):
Changes look good and work as proposed.

PR #6015, opened 2020-11-11
[Feature:InstructorUI] Restrict access to course materials by user

Kevin-MM:
There is no ability to restrict access to course materials on a user-by-user basis.

### What is the new behavior?
This PR begins to implement this behavior.  A user interface for this feature will be added in a later PR.

To test simply add a 'user_allow_list' array to a desired file in the `course_materials_file_data.json` as illustrated below.

image

**Note:** Using the instructor facing website interface of course materials to edit any files will likely clobber custom made edits to `course_materials_file_data.json`. 

Adding a 'user_allow_list' to a file will restrict access and visibility of that file to *only* those users in the allow list, or users that have a higher access level than students (for example instructors).

### DIFF
--- a/site/app/controllers/MiscController.php
+++ b/site/app/controllers/MiscController.php
@@ -14,6 +14,8 @@ use app\models\User;
 
 class MiscController extends AbstractController {
 
+    const GENERIC_NO_ACCESS_MSG = 'You do not have access to this file';
+
     /**
      * Get the current server time
      *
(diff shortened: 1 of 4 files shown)


jchulton:
This PR works as described above (tested and works for URL hacking). Once the user list is implemented with the interface, this feature should be ready to be used.

jchulton (approved):

PR #6022, opened 2020-11-18
[Feature:TAGrading] Keep notebook scroll position in TA grading page

### LINKED ISSUE #5976: TA grading -- auto scroll to same position in notebook...
If I'm TA grading the same notebook problem for every student, it would be nice as I move from student to student to have the notebook panel to remember and scroll to approximately the same position.   Currently it resets to the top every time.

Let's make sure this works even when using itempool gradeables (so students have different problems).

DefinitiveAbove:

When grading inside of the new TA grading interface, when a grader goes to a different student, the scroll position in the notebook resets.

### What is the new behavior?
Fixes #5976 
When grading inside of the new TA grading interface, when a grader goes to a different student, the scroll position in the notebook goes to the same content section from the previous one (if it is a random item pool item, it will go to the top of that item as a whole).

### Other information?
Tested by going through each section and going between students to make sure it went to the correct content section in the notebook.

### DIFF
--- a/site/app/models/notebook/UserSpecificNotebook.php
+++ b/site/app/models/notebook/UserSpecificNotebook.php
@@ -73,6 +73,8 @@ class UserSpecificNotebook extends Notebook {
         $new_notebook = [];
         $seen_items = [];
         $tests = [];
+        
+        $item_ref = 0;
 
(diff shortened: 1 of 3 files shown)


vikrsri (approved):
Tested functionality, changes seem to work as proposed.

PR #6027, opened 2020-11-22
[Feature:Duck] Add thanksgiving moorthy-duck for November

shailpatels:
There's no special moorthy-duck for November!

### What is the new behavior?
Adds a duck for thanksgiving week, credit to @hertacalvo !
image


emaicus (changes requested):
I do want to fast track this, but I don't think thanksgiving is always week 48, so this would have to be updated year on year. Do we already have a November duck? If not, let's just display this for all of November.

shailpatels:
> I do want to fast track this, but I don't think thanksgiving is always week 48, so this would have to be updated year on year. Do we already have a November duck? If not, let's just display this for all of November.

Whoops, yeah you're right thanksgiving would change to a different week each year, I updated the comparison to handle the yearly change - should work as intended now for the week of Thanksgiving

emaicus (approved):
Looks good!

PR #6029, opened 2020-11-24
[Bugfix:InstructorUI] Fix due date display

emaicus:
Add a missing ```</div>``` to fix the display of due dates


Kevin-MM (approved):
Changes look good.

Poked around in the tabs on the edit gradeable page for both a peer-grading gradeable and non peer-grading gradeable and all tabs appear to work correctly now.

PR #6040, opened 2020-12-02
[Bugfix: Submission] Correct Powerpoint Submission Extraction

### LINKED ISSUE #5995: .pptx submission gets extracted as if it were a zip
**Describe the bug**
When uploading a .pptx, Submitty appears to extract the file as if it were a zip (which I guess it technically is). 

**Expected behavior**
Submitty should really not be extracting the pptx as even though it's technically a zip, no one wants it to be extracted. 

**To Reproduce**
1. Visit gradeable submission
2. Click and drag .pptx file in 
3. Click submit
4. See a massive list of submitted files that were extracted out of the .pptx file

**Configuration**
 - OS: Windows 10
 - Browser: Chrome

**Screenshots**
image
image

**Additional context**
This specific assignment has no autograding so I don't think that's part of the issue. 

Also, if I zip the pptx, then it unzips the zip and leaves the pptx alone as expected.

myamamoto99:
Fixed #5995 

Originally Submitty was extracting a .pptx file submission as if it were a zip and thus having a long list of submitted files (fonts, pictures, etc.) when there should have only been one .pptx file. Now, Submitty will extract it as only a .pptx file and not have a long list of files. Submitty only shows one file, the .pptx file, in the list of submitted files.

### DIFF
--- a/autograder/autograder/autograding_utils.py
+++ b/autograder/autograder/autograding_utils.py
@@ -602,8 +602,13 @@ def zip_my_directory(path,zipfilename):
 def unzip_this_file(zipfilename,path):
     if not os.path.exists(zipfilename):
         raise RuntimeError("ERROR: zip file does not exist '", zipfilename, "'")
+    _, file_ext = os.path.splitext(zipfilename)
     zip_ref = zipfile.ZipFile(zipfilename,'r')
-    zip_ref.extractall(path)
+    if file_ext is ".pptx":
+        zip_ref.extract(path)
+    else:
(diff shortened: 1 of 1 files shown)


hdrodz97:
Hey! Thanks for your contribution!

Unfortunately, the file you edited does not handle zip files directly submitted by users via the Submission page. Instead, it handles zip files created in the backend during the autograding process, so this doesn't quite fix the bug.

Feel free to make another PR if you do figure this one out, though!

PR #6046, opened 2020-12-04
[Bugfix:TAGrading] Correct Stats for Peer Grading

### LINKED ISSUE #6039: Statistics are incorrect for peer gradeables on the stats page
On the stats page, grading statistics for peer gradeables are being tallied incorrectly.

For the "Peer PDF Homework" in the sample course, it can be seen that `Students who have submitted: 180 / 101 (178.2%)`

myamamoto99:
Fixes #6039 

Instead of counting the number of submissions to calculate the total who submitted, we count the users who submitted. This is because students can submit more than once, or cancel their submission, so counting all the submissions is not the right way. Counting the number of users who have submitted will give the correct percentage and number for the total who submitted.
Example: Before the "Peer PDF Homework" in the sample course, showed "Students who have submitted: 180 / 101 (178.2%)".
Now it shows "Students who have submitted: 75 / 101 (74.3%)".


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -517,10 +517,11 @@ public function showStatus($gradeable_id) {
         $total_who_submitted = 0;
         $peers_to_grade = 0;
         $peer_graded_components = 0;
+        $total_users_who_submitted = 0;
```
MasterOdin, line 520:
```suggestion
        $total_users_who_submitted = [];
```

The variable shape here must be an array with key/value pairs to foreach over, otherwise line 632 would error if we hit it without re-assigning `$total_users_who_submitted`. As such we should either give a default shape that will work (`[]`) or no value at all with the expectation that it'll always be re-assigned if we need it. I don't know enough about the code to say if we will never hit 632 without re-assigning the variable, so go instead to just give a default that'll function fine there.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -628,6 +629,12 @@ public function showStatus($gradeable_id) {
                 }
                 $total_submissions += $value;
             }
+            foreach ($total_users_who_submitted as $key => $value) {
+                if ($key == 'NULL') {
```
MasterOdin, line 633: Always better to use the `===` operator in PHP (and JS) instead of `==` as the latter is a weak form of equivalence vs `===` which is strong equivalence (ensures things like types are equivalent and does not do implicit type coercion).

```suggestion
                if ($key === 'NULL') {
```

MasterOdin (changes requested):

DefinitiveAbove (approved):
Looks good to me.

bmcutler:
@myamamoto99 
Thanks for your contribution!  Sorry it took so long to review and merge :)

Issue #6053, opened 2020-12-07
Late days allowed should not store effective dates with hours/minutes/seconds 


emaicus:
At present, the "Late days allowed" page stores effective dates with precision down to the second. Ideally, this should instead be stored with month/day/year precision. This change will require a migration to fix old entries.

PR #6054, opened 2020-12-07
[Bugfix:InstructorUI] Allow late days allowed entries to be deleted

emaicus:
Late days added on the "late days allowed" page are stored with hours/minutes/seconds precision. However, when an entry is deleted, it is deleted with month/day/year precision, resulting in failed deletions.

### What is the new behavior?
Deletion now allows "seconds" precision.

### Other information?
Per a conversation with @bmcutler, the intended behavior for this interface is actually for all portions to use "days" precision. However, at present, that would require a migration/more complicated refactor. This PR restores full functionality to the late days allowed page, while leaving future work per issue #6053


shailpatels (commented):

shailpatels (approved):

PR #6058, opened 2020-12-08
[Refactor:InstructorUI] Course Materials Refactors

Kevin-MM:
There is a lot of duplicated code related to access checking for course materials files.

### What is the new behavior?
Some code related to course materials access checking has been refactored.  The results of this refactor is as follows:
1. The `CourseMaterial` model has been removed.  All of this code has been moved into the `CourseMaterialsUtils` file.  All these functions are static, and following Submitty conventions, it made a lot more sense to have these functions in a Utils file instead of a Model file.
1. Each check previously opened/read/closed the course materials metadata json file.  This has been refactored so the file is read one time and its contents are stored in memory to be passed to each checking function.
1. Much code was duplicated in two separate spots to accommodate the two different cases of a file being downloaded vs viewed in the browser.  Most of the duplicated code has now been moved into a single function call.


shailpatels (approved):
Changes looks good and didn't notice any differences to functionality.

PR #6061, opened 2020-12-09
[Bugfix:Autograding] Log worker crashes

hdrodz97:

When a worker process crashes, there is no obvious indication that it has crashed. There is a monitoring loop in place, but it incorrectly always thinks that all processes are alive due to a typo.

### What is the new behavior?

When a worker process throws an exception, the exception is logged to the stack traces directory, and a message is logged to the autograding logs directory.

On the monitoring side, the loop now properly detects how many worker processes are still alive and will print a warning to the log when any worker has crashed. Furthermore, the polling frequency has been decreased from once a second to once a minute — this should be frequent enough to be noticeable when inspecting the logs but not frequent enough to cause log sizes to blow up dramatically.

### Other information?

To test the monitoring thread, I added the following to `submitty_autograding_worker.py` on a local clone:

At the top of the file:
```python
import random
```

Around line 27, in the beginning of the `worker_process` function:
```python
if random.choice([True, False]):
    raise Exception("unlucky")
```

This will cause each launched worker to have a 50% chance of crashing on startup.


emaicus (approved):
Looks good to me.

PR #6073, opened 2020-12-13
[Feature:TAGrading] Graders can leave notes that won't be seen by students

### LINKED ISSUE #4788: TA only notes on individual rubric marks
It would be helpful to have an option to specify TA only notes for every mark on the rubric.  These would only be displayed to the TAs during grading, but not to the students when they view their grade.

At the same time, it will become even more important to improve the display of extra long rubric text.  Currently, the boxes (at least on the instructor UI) are limited in height to 1 row of text.

rsdrew:
Closes #4788 

When grading a gradeable, TA's can only leave comments that the students can see, either through a custom message for a certain component, or an overall comment.

Before view:
image

### What is the new behavior?

There was a feature request to allow TAs to leave notes for each mark that only other graders can see.  Having a separate textbox for each mark is redundant, so one textbox has been provided for each component. Now, TAs can leave internal notes that only other graders can see. These notes are not visible to students when viewing their grades.

After view:
image

### Testing

Notes can be left when leaving a grade:
image

Notes can be left without leaving a grade:
image

Instructors can leave notes (Notice the "Graded by Instructor"):
image

And TAs can read them (Notice the "Logout Jill" and "Graded by Instructor"):
image

TAs can leave notes:
image

And Instructors can read them:
image

Other TAs can also read them (Notice the "Logout Jack"):
image

And students cannot see these notes when viewing their grade:
image

### DIFF
--- a/.setup/bin/setup_sample_courses.py
+++ b/.setup/bin/setup_sample_courses.py
@@ -261,6 +261,9 @@ def generate_random_ta_note():
 def generate_random_student_note():
     return get_random_text_from_file('StudentNote.txt')
 
+def generate_random_internal_grader_note():
+    return get_random_text_from_file('InternalGraderNote.txt')
+
 
 def generate_random_marks(default_value, max_value):
     with open(os.path.join(SETUP_DATA_PATH, 'random', 'marks.yml')) as f:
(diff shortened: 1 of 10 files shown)


shailpatels:
@bmcutler Was this feature implemented when the new TA grading interface was made default:
The solution/TA nodes seems like it does the same things:

image

bmcutler:
Thanks for sharing your ideas, but we have implemented an alternate method for sharing TA notes.

Issue #6080, opened 2020-12-18
notebook rendering of version rollback is wrong and/or confusing


bmcutler:
**Describe the bug**
If a student submits to a notebook gradeable multiple times, 
and rolls back to a previous version the display of the "currently active" 
answers is confusing.

This may be tied into attempts of the browser page to locally store/display 
recent unsaved/unsubmitted edits to the page.

Student opens multiple tabs / different browsers / different machines.
Could have conflicting locally stored edits on multiple machines.

for each box we could be displaying:
default initial value (might be empty or nonempty depending on config)
Most recent submitted-to-server value
active submitted value (active might not be most recent)
selected submitted value (current selected might not be active nor most recent)
what is in the local browser storage for this assignment on this machine.

also...  pulling from the local storage is buggy if we have logged out & log 
back in as a different user.  It is remember answers or edits done as the other user.

NOTE: I've assigned a bunch of people.  We need to discuss the UI, the code needs to be
very carefully reviewed, and we need to test on every os/browser we can think of that might
have different behavior.  And also write automated tests.

PR #6091, opened 2020-12-21
[Bugfix:Submission] Separate autosave by user; fix multiple choice mangling

hdrodz97:

This addresses a sub-issue in #6080, wherein when multiple users are sharing the same browser, notebook autosave data does not distinguish between said users.

Furthermore, this also includes a bugfix for autosave where multiple choice data for a notebook gradeable with more than 10 multiple choice questions would get mangled.

### What is the new behavior?

* Each user's notebook autosave data is now stored independently of one another.

* Restoring autosave data of notebooks with more than 10 multiple choice questions now works as intended.

### Other information?

Tested on Windows 10 + Chrome, macOS 11.1 + Safari, macOS 11.1 + Firefox.

As there currently are no notebook-specific end-to-end tests, this PR does *not* include any tests. Rather, I believe these should come in as a separate PR to avoid excessive scope creep.


shailpatels (approved):

Tested this by entering notebook questions across multiple browsers
Also did not run into the issue of a notebook with 10+ questions getting local storage questions loaded wrong

PR #6094, opened 2020-12-22
[Feature:InstructorUI] Add import and export for polling

BenKelly99:
### What is the new behavior?
Adds the ability for instructors to export the current semester's poll as a JSON so they can import them into another class easily later.

The Export button downloads the polls as a JSON file, and the import buttons adds new polls based on the selected JSON file.

If one of the polls has an error (missing a required field), the other polls will still be added and a message will report the number of successfully added polls, as well as the number of errors.


site/app/controllers/PollController.php
```diff
@@ -402,4 +402,75 @@ public function viewResults($poll_id) {
...
+        }
+        return JsonResponse::getSuccessResponse($data);
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/polls/importPolls", methods={"POST"})
+     * @AccessControl(role="INSTRUCTOR")
+     * @return JsonResponse
+     */
+    public function importPollsFromJSON() {
```
shailpatels, line 435: same comment as above

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -280,4 +282,69 @@
...
+                document.body.removeChild(element);
+            }
+        });
+    }
+
+    function importPolls() {
+        let input = document.createElement('input');
+        input.type = 'file';
+        
+        input.onchange = e => {
```
shailpatels, line 315: you should put a paren here so its `input.onchage = (e) => { ...`

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -280,4 +282,69 @@
...
+                element.setAttribute('download', "{{semester}}-{{course}}-polls.json");
+                element.style.display = 'none';
+                document.body.appendChild(element);
+                element.click();
+                document.body.removeChild(element);
+            }
+        });
+    }
+
+    function importPolls() {
```
shailpatels, line 311: I'm confused at what this importPolls function is doing with ajax it looks like you're refreshing the page at the end of this anyways and only just uploading a file so it would be possible to just do this with a plain HTML form.

```html
<form action="{{base_url}}/importPolls">
  <input type="file" id="import-polls-area" name="polls-file">
  <input type="submit">
</form>
```

shailpatels (changes requested):

site/app/templates/polls/AllPollsPageInstructor.twig
BenKelly99: The thing I don't like about this solution is that it makes you have to click a button for uploading the file, then click submit. The way I have it set up now makes it all one "Import" button instead of two separate buttons. I agree about the Ajax it was a kind of work around for something a little messy. I'll find a cleaner solution.

MasterOdin: You could add a `onChange` event on the file input that then fires a `submit()` event on the overall form, bypassing the need for hitting the submit button, but retain the overall "it's a form" simplicity @shailpatels is suggesting.

BenKelly99:
Combining with related PR which needed code used here.

PR #6096, opened 2020-12-22
[Feature:TAGrading] Cached late days for more stats

jportorreal00:

All late days are currently calculated with context. This means that we generate the late day information every single time per user. We calculate it by running through every single graded gradable a user is being graded for and subtracting any late days if we see a late submission. This works well, but it means we can't show cool stats on the TAGrading page because its to expensive to do for every student at once :(
 
### What is the new behavior?
When there is a late day recalculation trigger, the late day values are calculated and then cached within the database in order for easy access. This means we can customize the TA status and grading index page a lot more.

### Other information?

All "triggers" I accounted for:
-submit a gradeable
-gradeable version change
-drop submission
-late day addition (late days allowed page)
-late day change (gradeable admin page)
-gradeable due date change
-extension

This PR is in order to finish https://github.com/Submitty/Submitty/pull/5631. This PR will use the caching system in order to add more stats and visual assistance to the TAs

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -293,6 +293,41 @@ CREATE TABLE public.gradeable (
 
 
 --
+-- Name: gradeable_access; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.gradeable_access (
+    id integer NOT NULL,
+    g_id character varying(255) NOT NULL,
+    user_id character varying(255),
+    team_id character varying(255),
+    accessor_id character varying(255),
+    "timestamp" timestamp with time zone NOT NULL,
+    CONSTRAINT access_team_id_check CHECK (((user_id IS NOT NULL) OR (team_id IS NOT NULL)))
+);
+
+
+--
+-- Name: gradeable_access_id_seq; Type: SEQUENCE; Schema: public; Owner: -
+--
+
+CREATE SEQUENCE public.gradeable_access_id_seq
+    AS integer
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
+
+
+--
+-- Name: gradeable_access_id_seq; Type: SEQUENCE OWNED BY; Schema: public; Owner: -
+--
+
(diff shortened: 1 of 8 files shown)


jportorreal00:
@emaicus Hey Evan. This PR is for the caching of late days and the status for each gradeable for a given user. Its supposed to be used for PR #5631 so we can show some stats for the TAGrading pages.

This is still a draft. It works as expected, when Late days need to be recalculated, they are saved onto the database and are accessible easily within each graded gradeable.

Concerns:
-Efficient enough?
-Takes a long time for each recalculation (Its one late day per user and each late day uses EVERY gradeable to calculate the late day information for each of them)
-The most expensive operations are editing a gradeable. If we change the due date or the number of late days we have to recalculate late days for every student in the course (it takes 30+ seconds). '
-Is there a way to run recalculations on another thread?

Since layout is here maybe there could be a way to completely move the late day recalculation to the database with triggers as mentioned before in another meeting. Just a thought

Sorry for the long comment!

MasterOdin:
Looking briefly at the code, one concern on the efficiency would be in the number of queries it runs on the admin page. On changing a gradeable, it looks like it runs `((3 + NUM_GRADEABLES) * NUM_USERS)` queries, which for a large class like CS1 with ~840 students and ~12 gradeables is going to tank performance on that page at ~30,000 queries. The solution here would be to try and cache as much as possible where we can, and to then do bulk selects + updates where we can. For example, some quick ideas:

* in `LateDays::cacheLateDayInfoForUser`, you are passing in a `$user_id`, and then doing a DB query to get the user info. Can we just mandate passing in a user object from the caller? What callers would that require changing to have that info?
* for each user we construct a `LateDays` object for with `LateDays::fromUser`, we are doing a select of all gradeable configs, then filtering down that list to the ones applicable to that user. We should cache the the gradeable configs and re-use the same base object on each filter
* You can consolidate the cache update to do a "bulk update" so that we only send on query for all gradeables. To do this, you would use the `UPDATE ... FROM` syntax of postgres ([doc page](https://www.postgresql.org/docs/9.6/sql-update.html)). The bottom bit of this [SO post](https://stackoverflow.com/a/20224370) gives an idea of how that might look.

These are the easiest gains, and would hopefully take us down to a much more manageable query count, and giving a nice speed bump.

Let me know if this makes sense or if you've got any questions about the above.

PR #6097, opened 2020-12-22
[Bugfix:Submission] Don't extract Microsoft Office files as zip

### LINKED ISSUE #5995: .pptx submission gets extracted as if it were a zip

hdrodz97:

Fixes #5995

When uploading certain Microsoft Office files, the submission controller identifies these as zip files, which, while *technically* true, is not intended behavior. This happens because the submission controller exclusively uses the MIME type of the file to determine whether it is a ZIP file or not, 

### What is the new behavior?

The submission controller now checks if the file extension is zip *in addition to* inspecting the MIME type of the file. This way, a user can be confident in knowing that their upload will only be expanded if it is a valid zip file with the zip extension.

### Other information?

Exporting a docx from Docs, a pptx from Slides, or an xlsx from Sheets will reliably trigger this bug.

This is an artifact of an [upstream bug](https://bugs.astron.com/view.php?id=46) in `libmagic`, part of the `file` utility, which was fixed in October 2018. The version of `file` included in Ubuntu 18.04 has the same bug: running `file --mime-type` on a Microsoft Office file exported from GSuite will display `application/zip`.

### DIFF
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -1167,7 +1167,7 @@ class SubmissionController extends AbstractController {
                 if (isset($uploaded_files[$i])) {
                     $uploaded_files[$i]["is_zip"] = [];
                     for ($j = 0; $j < $count[$i]; $j++) {
-                        if (mime_content_type($uploaded_files[$i]["tmp_name"][$j]) == "application/zip") {
+                        if (mime_content_type($uploaded_files[$i]["tmp_name"][$j]) === "application/zip" && pathinfo($uploaded_files[$i]["name"][$j], PATHINFO_EXTENSION) === "zip") {
                             if (FileUtils::checkFileInZipName($uploaded_files[$i]["tmp_name"][$j]) === false) {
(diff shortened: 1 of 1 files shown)


shailpatels (approved):
Looks good, appears to stop PowerPoint presentations from Microsoft and ones generated from google drive getting unzipped, the long term solution would be to add an option to unzip a folder as discussed earlier but this should stop the immediate issue mentioned

PR #6099, opened 2020-12-23
[Testing:System] Validate PR Title

bmcutler:
Add GitHub Action that runs regex on the PR title, ensuring the format matches:
https://submitty.org/developer/how_to_contribute

Uses:
https://github.com/naveenk1223/action-pr-title

Note:  This workflow validates against the current list of valid "TYPE"s & "MODULE"s.
We will be revising these lists soon (in both documentation & this github action.)


.github/workflows/pr_title.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: naveenk1223/action-pr-title@master
+        with:
+          #
+          # pull request titles format rules here:
+          # https://submitty.org/developer/how_to_contribute#how-to-make-a-pull-request-pr-to-submitty
+          #
+          # [<TYPE>:<MODULE>] <SUBJECT>
+          #
+          regex: '^\[(Bugfix|Feature|Refactor|Testing|Documentation|VPAT|UI/UX|Dependency):(Submission|Autograding|Forum|Notifications|TAGrading|InstructorUI|RainbowGrades|System|Developer|API)\] .{2,40}$'
```
MasterOdin, line 22: I believe we need to start the regex with `(?:\[SYSADMIN ACTION\] )?`, otherwise looks good. May also want to add `DevDependency` as another type and `PHP` and `JS` as subtypes.

I can talk to @shailpatels on setting up an action for auto renaming dependabot PRs to fit this scheme so that those PRs won't start showing up as failed after this is merged.

MasterOdin (approved):

.github/workflows/pr_title.yml
shailpatels: Sounds good, 
Looks like we can also configure dependabot to change it's commit message/pr title by having a config yaml for it

https://github.com/dependabot/dependabot-core/pull/1227

MasterOdin: Unfortunately, I don't think dependabot gives us quite enough granularity to specify the appropriate sub-type (PHP / JS), and so with the `prefix` and `prefix_development` config variables we would only be able to get `[Dependency]` or `[DevDependency]`.

bmcutler: If I understand correctly the limitations...
I've modified the regex to not require (or allow) the ":module" for dependency or devdependency

MasterOdin: Yeah, I think losing the subtype is probably not that big a deal for `[Dependency]` or `[DevDependency]` as I do not think anyone was really making use of that information. If someone clamors for it, we can look to re-add it then.

PR #6105, opened 2020-12-25
[Testing:System] Fix regex for PR titles for dependabot

MasterOdin:

The regex for PR titles added in #6099 is a bit too strict for dependabot which regularly generates titles that are between 50-70 characters long. This issue is shown in the failing on titles for recent dependabot PRs, like #6103.

### What is the new behavior?

Adjusts the regex so that the `[Dependency]` and `[DevDependency]` tag allow for up-to 70 characters in the title (to accommodate for our longest dependency title `@fortawesome/fontawesome-free`.

### Other Information

It does not look like dependabot surfaces ability to generate a shortened message (dropping say the `in /site` bit for us). We could maybe look into an action to deal with automatic PR title rewriting, but my feeling from discussion in #6099 is we would like to avoid that except as a last resort.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -31,3 +31,6 @@ exclude=
     sbin/run_lichen_plagiarism.py,
     sbin/untrusted_canary.py,
     tests
+
+per-file-ignores =
+    .github/workflows/pr_title_tester.py:E501,W605
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -19,4 +19,4 @@ jobs:
           #
           # [<TYPE>:<MODULE>] <SUBJECT>
           #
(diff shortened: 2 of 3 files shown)


bmcutler (approved):
rearranged a couple parens
and added a file I was using to (manually) test the regex -- not sure if it will break github if I dump it in this folder

MasterOdin:
@bmcutler I added an ignore on the regex and line too long errors for linting on that particular file since getting the errors fixed seems more bother than worth, and the script is only being used internally to self-test the GH workflow, and not used in or to test Submitty itself.

bmcutler:
@MasterOdin 
thanks for the adding the exceptions

PR #6122, opened 2021-01-01
[Feature:InstructorUI] Office hours queue audible alerts

### LINKED ISSUE #6088: alert sound when student enters the office hour queue
Office hours are quiet today, and the TA has the office hours queue window open, but they are doing other work.  TA might like to configure an audible bell when the student enters the queue so they can help them promptly.

note: I think someone had instructions for how to customize the browser to get some form of alerts (even before we setup websockets).  Is this documented?  Do we need to / should we add a configuration option to the page to for an audible alert?

Kevin-MM:
Closes #6088

Teaching staff using the office hours queue are limited in their ability to be alerted when changes occur in the office hour queue.

### What is the new behavior?
A new button has been added which allows teaching staff to hear an audible chime when a student joins or leaves the queue.

**Notes**
1. Most browser policies forbid playing audio without first interacting with the page.  For this reason it is required to explicitly enable this feature (by clicking the bell) on first page load, or after any full page reload.
1. The chime sound was sourced from the [material design toolkit](https://material.io/design/sound/sound-resources.html).  It is released under the creative commons `Attribution 4.0 International` license.  I am unsure of compatibility of this license and any submitty open source licenses.
1. The exact sound is question is titled `alert_simple.ogg` or `alert_simple.wav`.

image


MasterOdin:
> The chime sound was sourced from the material design toolkit. It is released under the creative commons Attribution 4.0 International license. I am unsure of compatibility of this license and any submitty open source licenses.

From my understanding of CC for non-commercial products, it's expected that you make a best effort to provide attribution and not claim we made something. Accomplishing this can probably be done by just including a link to the source material and their license in the place we use it, probably both in the code itself and maybe also in [LICENSE-THIRD-PARTY](https://github.com/Submitty/Submitty/blob/master/LICENSE-THIRD-PARTY.md).

site/public/js/notification-sound.js
```diff
@@ -0,0 +1,19 @@
...
+        }
```
MasterOdin, line 10: Why include both, and not just wav? Comparing support for [wav](https://caniuse.com/?search=wav%20audio%20format) vs [ogg](https://caniuse.com/?search=ogg%20audio%20format), it looks like wav supports all the same platforms as ogg with the addition of Safari.

site/public/js/notification-sound.js
```diff
@@ -0,0 +1,19 @@
...
+        else {
+            this.sound = null;
+        }
+    }
+
+    play() {
+        this.sound ? this.sound.play() : alert(`A queue change occured, but your browser doesn't support playing audio on this page.`);
```
MasterOdin, line 17: Looking at this, if a user activates sound on a computer that does not support and then wanders away for a while and a bunch of queue stuff happens, when they come back, are they then going to have to click through a bunch of alert windows all saying this error message?

site/public/js/notification-sound.js
Kevin-MM: Good catch, I was looking at an outdated table when I was trying to decide what audio types to include.  I will rework this to only check and include wav.

site/public/js/notification-sound.js
Kevin-MM: Good point, I will rework this.

MasterOdin: My view would be to do this check on attempting to turn on the icon, and perhaps disable / gray out the icon altogether on page load if they cannot play sound with a mouse over text of "Your browser does not support step sound" or something?

site/public/js/notification-sound.js
MasterOdin: Looking at the docs page for [canPlayType](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType), we might want the conditional here to also be just `this.sound.canPlayType('audio/wav') !== ''` as this function can return `'probably'`, `'maybe'`, and `''`, even though in practice `'maybe'` and `''` are the only ones returned.

Kevin-MM: Reworked this

site/public/js/notification-sound.js
Kevin-MM: I reworked this so that the button is now only displayed if we are reasonably sure the browser supports playing the audio.

site/public/js/notification-sound.js
```diff
@@ -0,0 +1,15 @@
+class NotificationSound {
+    constructor() {
+        this.sound = document.createElement('audio');
+
+        if (this.sound.canPlayType('audio/wav') === '') {
```
MasterOdin, line 5: it occurs to me that we probably want to convert this sound file into MP3, which would shrink the raw audio file from 168kb to 9kb (and giving us a similar reduction in this file). It looks like MP3 is well supported on all browsers we support, and that I doubt anyone will notice the very slight degradation of quality.

MasterOdin (approved):
Minor comment on just converting the file content, however, this looks and works fine once I jury-rigged past my websocket issues. I'm fine merging this as-is (though please make an issue to track converting this to mp3) or after you convert the file to mp3.

bmcutler:
worked for me!  awesome!
Classic UI problem tho...  I was confused about whether the sound effect was enabled or not enabled.  
(button symbol is confusing.  hover message is good.  might also want to grey it out and make it look disabled?)
But the TAs will figure it out via trial & error :)

PR #6129, opened 2021-01-07
[Feature:InstructorUI] Polls data dump

BenKelly99:
Currently, there is no way to access poll data without sysadmin access.

### What is the new behavior?
Now, any instructor can get access to a JSON summary of all the polls, so they can use it to assign late days, give extra points, etc... The JSON is generated in the .../reports/ directory when "Generate Grade Summaries" is run. The file is only generated if polling is enabled.

### DIFF
--- a/migration/migrator/migrations/course/20210111163413_add_poll_data_folder.py
+++ b/migration/migrator/migrations/course/20210111163413_add_poll_data_folder.py
@@ -7,20 +7,16 @@ def up(config, database, semester, course):
     polls_dir = Path(course_dir, 'polls')
 
     #create directory
-    os.mkdir(polls_dir)
+    polls_dir.mkdir(mode=0o750, exist_ok=True)
 
     php_user = config.submitty_users['php_user']
 
     # get course group
-    stat_info = os.stat(str(course_dir))
-    course_group_id = stat_info.st_gid
+    course_group_id = course_dir.stat().st_gid
     course_group = grp.getgrgid(course_group_id)[0]
 
     # set the owner/group/permissions
     os.system("chown -R "+php_user+":"+course_group+" "+str(polls_dir))
-    os.system("chmod -R u+rwx "+str(polls_dir))
-    os.system("chmod -R g+rxs "+str(polls_dir))
-    os.system("chmod -R o-rwx "+str(polls_dir))
 
 
 def down(config, database, semester, course):
--- a/migration/migrator/migrations/course/20210111163413_add_poll_data_folder.py
+++ b/migration/migrator/migrations/course/20210111163413_add_poll_data_folder.py
@@ -1,0 +1,27 @@
+import os
+import grp
+from pathlib import Path
+
+def up(config, database, semester, course):
+    course_dir = Path(config.submitty['submitty_data_dir'], 'courses', semester, course, 'reports')
+    polls_dir = Path(course_dir, 'polls')
+
+    #create directory
+    os.mkdir(polls_dir)
+
+    php_user = config.submitty_users['php_user']
+
+    # get course group
+    stat_info = os.stat(str(course_dir))
+    course_group_id = stat_info.st_gid
+    course_group = grp.getgrgid(course_group_id)[0]
+
+    # set the owner/group/permissions
+    os.system("chown -R "+php_user+":"+course_group+" "+str(polls_dir))
+    os.system("chmod -R u+rwx "+str(polls_dir))
+    os.system("chmod -R g+rxs "+str(polls_dir))
+    os.system("chmod -R o-rwx "+str(polls_dir))
+
+
+def down(config, database, semester, course):
+    pass
--- a/migration/migrator/migrations/course/20210111163413_add_poll_data_folder.py
+++ b/migration/migrator/migrations/course/20210111163413_add_poll_data_folder.py
@@ -1,5 +1,4 @@
-import os
-import grp
+import shutil
 from pathlib import Path
 
 def up(config, database, semester, course):
@@ -13,10 +12,9 @@ def up(config, database, semester, course):
 
     # get course group
     course_group_id = course_dir.stat().st_gid
-    course_group = grp.getgrgid(course_group_id)[0]
 
     # set the owner/group/permissions
-    os.system("chown -R "+php_user+":"+course_group+" "+str(polls_dir))
+    shutil.chown(polls_dir, php_user, course_group_id)
 
 
 def down(config, database, semester, course):
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -417,6 +417,7 @@ class PollController extends AbstractController {
      */
     public function importPollsFromJSON(): RedirectResponse {
         $filename = $_FILES["polls_file"]["tmp_name"];
+        $polls = FileUtils::readJsonFile($filename);
         if ($polls === false) {
             $this->core->addErrorMessage("Failed to read file. Make sure the file is the right format");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -10,6 +10,7 @@ use app\libraries\response\RedirectResponse;
 use Symfony\Component\Routing\Annotation\Route;
 use app\libraries\routers\AccessControl;
 use app\libraries\DateUtils;
+use app\libraries\FileUtils;
 use app\libraries\Utils;
 use app\libraries\routers\FeatureFlag;
 use app\libraries\PollUtils;
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -417,10 +417,7 @@ class PollController extends AbstractController {
      */
     public function importPollsFromJSON(): RedirectResponse {
         $filename = $_FILES["polls_file"]["tmp_name"];
-        $file = fopen($filename, "r");
-        $contents = fread($file, filesize($filename));
-        $polls = json_decode($contents, true);
-        if (json_last_error() != JSON_ERROR_NONE) {
+        if ($polls === false) {
             $this->core->addErrorMessage("Failed to read file. Make sure the file is the right format");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
@@ -428,8 +425,10 @@ class PollController extends AbstractController {
         $num_errors = 0;
         foreach ($polls as $poll) {
             if (
-                !array_key_exists("name", $poll) || !array_key_exists("question", $poll)
-                || !array_key_exists("responses", $poll) || !array_key_exists("correct_responses", $poll)
+                !array_key_exists("name", $poll)
+                || !array_key_exists("question", $poll)
+                || !array_key_exists("responses", $poll)
+                || !array_key_exists("correct_responses", $poll)
                 || !array_key_exists("release_date", $poll)
             ) {
                 $num_errors = $num_errors + 1;
@@ -451,7 +450,7 @@ class PollController extends AbstractController {
             $this->core->getQueries()->addNewPoll($name, $question, $responses, $answers, $release_date, $orders);
             $num_imported = $num_imported + 1;
         }
-        if ($num_errors == 0) {
+        if ($num_errors === 0) {
             $this->core->addSuccessMessage("Successfully imported " . $num_imported . " polls");
         }
         else {
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -404,11 +404,20 @@ class PollController extends AbstractController {
     /**
      * @Route("/courses/{_semester}/{_course}/polls/export", methods={"GET"})
      * @AccessControl(role="INSTRUCTOR")
-     * @return JsonResponse
      */
-    public function getPollExportData(): JsonResponse {
-        $polls = $this->core->getQueries()->getPolls();
-        return JsonResponse::getSuccessResponse(PollUtils::getPollExportData($polls));
+    public function getPollExportData() {
+        $polls = PollUtils::getPollExportData($this->core->getQueries()->getPolls());
+        $file_name = date("Y-m-d") . "-" . $this->core->getConfig()->getSemester() . "-" . $this->core->getConfig()->getCourse() . "-" . "poll-data" . ".json";
+        $data = FileUtils::encodeJson($polls);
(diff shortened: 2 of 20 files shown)


MasterOdin:
For a course like CS1 or CS2 (or any course with a lot of students), how big is this json file going to get?

bmcutler:
For Data Structures: 28 lectures x 6 multiple choice questions per lecture x # of students x 700 students. 
For Fall 2020 (with ~300 students) a dump (different format) of this data is ~700kb.

BenKelly99:
@MasterOdin @bmcutler I added 28 * 6 = 168 polls to the database, each with a title of ~30 characters and then added 700 student responses to each of them. The file ended up being 3.3M.

MasterOdin:
@bmcutler Thanks for providing that information

@BenKelly99 Thanks for running those numbers to see the load. I am somewhat cautious on this part of the codebase on eating up too much memory, but I think we can spare 3.3M overhead in this singular function and can leave this function simpler above complicating it with streaming out the JSON file piece by piece as we do elsewhere in the report generator.

site/app/controllers/admin/ReportController.php
```diff
@@ -81,6 +81,18 @@ public function generateGradeSummaries() {
             $this->saveUserToFile($base_path, $a, $b, $c, $d);
             return null;
         });
+
+
+        if ($this->core->getConfig()->isPollsEnabled()) {
+            $poll_base_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), 'reports');
+            // Check that the directory is writable, fail if not
+            if (!is_writable($base_path)) {
```
MasterOdin, line 89:
```suggestion
            if (!is_writable($poll_base_path)) {
```

BenKelly99:
So leave as is for now? Sounds good.

site/app/controllers/admin/ReportController.php
```diff
@@ -81,6 +81,18 @@ public function generateGradeSummaries() {
             $this->saveUserToFile($base_path, $a, $b, $c, $d);
             return null;
         });
+
+
+        if ($this->core->getConfig()->isPollsEnabled()) {
```
MasterOdin, line 86: I would be in favor of moving this before we generate the grade summaries themselves. That way, if we have `is_writable` failures, we hit those first before attempting to write out any data. My general preference on error handling is to cover as much up-front (especially these sort of static checks) so that we don't have to give return messages to the user like "Generated grade summaries, but failed to generate poll summaries" or whatever.

site/app/controllers/admin/ReportController.php
```diff
@@ -367,6 +379,26 @@ private function saveUserToFile(string $base_path, User $user, array $ggs, LateD
...
+        $polls_data = [];
+        foreach ($polls as $poll) {
+            $poll_data = [];
+            $poll_data["name"] = $poll->getName();
+            $student_response_data = [];
+            foreach ($poll->getUserResponses() as $student => $response) {
+                $student_response_data[$student] = $poll->isCorrect($response);
+            }
+            $poll_data["responses"] = $student_response_data;
+            $polls_data[] = $poll_data;
```
MasterOdin, line 397: Very slight micro optimization, but we can save on the memory steps of creating a new variable and assigning data to it and instead just push the entire thing to `$polls_data` once we're ready:

```suggestion
            $student_response_data = [];
            foreach ($poll->getUserResponses() as $student => $response) {
                $student_response_data[$student] = $poll->isCorrect($response);
            }
            $poll_data["responses"] = $student_response_data;
            $polls_data[] = [
                'name' => $poll->getName(),
                'responses' => $student_response_data,
            ];
```

site/app/controllers/admin/ReportController.php
```diff
@@ -367,6 +379,26 @@ private function saveUserToFile(string $base_path, User $user, array $ggs, LateD
         file_put_contents(FileUtils::joinPaths($base_path, $user->getId() . '_summary.json'), FileUtils::encodeJson($user_data));
     }
 
+    /**
+     * Generates a summary of all polls over a semester if polling is enabled
+     * @param string $base_path the base path to store the report
+     */
+    private function generatePollSummaryInternal($base_path) {
```
MasterOdin, line 386:
```suggestion
    private function generatePollSummaryInternal(string $base_path): void {
```

MasterOdin (approved):
Slight comment, but otherwise the code looks and works fine.

For reference, this looks to add 3 queries per poll to the generation page. Not exactly a huge deal here (I think) with ~90 new queries, but something that can probably be revised in the future to take advantage of joins and such to reduce that number. Same with looking at lowering the memory. Could be lowered, but not important I think at this stage as <3.5M is not that much memory in the scheme of things for a page that is not that often run.

site/app/controllers/admin/ReportController.php
```diff
@@ -367,6 +380,27 @@ private function saveUserToFile(string $base_path, User $user, array $ggs, LateD
...
+     */
+    private function generatePollSummaryInternal(string $base_path): void {
+        $polls = $this->core->getQueries()->getPolls();
+        $polls_data = [];
+        foreach ($polls as $poll) {
+            $student_response_data = [];
+            foreach ($poll->getUserResponses() as $student => $response) {
+                $student_response_data[$student] = $poll->isCorrect($response);
+            }
+            $poll_data["responses"] = $student_response_data;
```
MasterOdin, line 395:
```suggestion
````

File: site/app/controllers/admin/ReportController.php [hunk 6]
```diff
@@ -67,6 +67,14 @@ public function generateGradeSummaries() {
             $this->core->redirect($this->core->buildCourseUrl(['reports']));
         }
 
+        $poll_base_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), 'reports');
+
+        // Check that the directory is writable, fail if not
+        if (!is_writable($poll_base_path)) {
```
MasterOdin, line 73: I was suggestion moving the entire if block here, and not just the error check, but this is fine. I do think we still want to run this error check only if polls are enabled or else in an incorrectly configured setup, someone might hit this error with polls disabled, and then complain about this as a bug that polls are doing something they shouldn't be.

```suggestion
        if ($this->core->getConfig()->isPollsEnabled() && !is_writable($poll_base_path)) {
```

BenKelly99:
Combined this PR with the import/export PR since they became closely related after the changes Barb and I discussed.

site/app/controllers/PollController.php
```diff
@@ -398,4 +399,64 @@ public function viewResults($poll_id) {
...
+            $this->core->addErrorMessage("Failed to read file. Make sure the file is the right format");
+            return new RedirectResponse($this->core->buildCourseUrl(['polls']));
+        }
+        $num_imported = 0;
+        $num_errors = 0;
+        foreach ($polls as $poll) {
+            if (
+                !array_key_exists("name", $poll) || !array_key_exists("question", $poll)
+                || !array_key_exists("responses", $poll) || !array_key_exists("correct_responses", $poll)
+                || !array_key_exists("release_date", $poll)
```
MasterOdin, line 433: Put all elements on their own lines. This will (eventually) be a lint rule:

```suggestion
                !array_key_exists("name", $poll)
                || !array_key_exists("question", $poll)
                || !array_key_exists("responses", $poll)
                || !array_key_exists("correct_responses", $poll)
                || !array_key_exists("release_date", $poll)
```

site/app/controllers/PollController.php
```diff
@@ -398,4 +399,64 @@ public function viewResults($poll_id) {
...
+                $responses[$response_id] = $response;
+                $orders[$response_id] = $i;
+                $i = $i + 1;
+            }
+            $answers = $poll["correct_responses"];
+            $release_date = $poll["release_date"];
+            $this->core->getQueries()->addNewPoll($name, $question, $responses, $answers, $release_date, $orders);
+            $num_imported = $num_imported + 1;
+        }
+        if ($num_errors == 0) {
```
MasterOdin, line 454:
```suggestion
        if ($num_errors === 0) {
```

site/app/controllers/admin/ReportController.php
```diff
@@ -367,6 +381,23 @@ private function saveUserToFile(string $base_path, User $user, array $ggs, LateD
...
+        $polls = $this->core->getQueries()->getPolls();
+        $polls_data = [];
+        foreach ($polls as $poll) {
+            $polls_data[] = [
+                "id" => $poll->getID(),
+                "responses" => $poll->getUserResponses()
+            ];
+        }
+        file_put_contents(FileUtils::joinPaths($base_path, "poll_responses_summary.json"), FileUtils::encodeJson($polls_data));
+        file_put_contents(FileUtils::joinPaths($base_path, "poll_data_summary.json"), FileUtils::encodeJson(PollUtils::getPollExportData($polls)));
```
MasterOdin, line 398: Use [`FileUtils::writeJsonFile`](https://github.com/Submitty/Submitty/blob/a9e09405eaf6b370fdd8978805a144a71042306e/site/app/libraries/FileUtils.php#L287):

```suggestion
        FileUtils::writeJsonFile(FileUtils::joinPaths($base_path, "poll_responses_summary.json"), $polls_data);
        FileUtils::writeJsonFile(FileUtils::joinPaths($base_path, "poll_data_summary.json"), PollUtils::getPollExportData($polls));
```

site/app/libraries/PollUtils.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace app\libraries;
+
+use app\libraries\Core;
+use app\libraries\FileUtils;
+use app\libraries\response\JsonResponse;
```
MasterOdin, line 7: unused `use` statement:

```suggestion
```

site/app/libraries/PollUtils.php
```diff
@@ -0,0 +1,36 @@
...
+ * Class FileUtils
+ *
+ * Contains various useful functions for the polling system.
+ */
+class PollUtils {
+    /**
+     * Generates a PHP array containing the all the poll data used in the export procedure
+     * @param array $polls the array of polls to be summarized
+     */
+    public static function getPollExportData(array $polls): array {
```
MasterOdin, line 19: can you add a test case or two (e.g. test for empty `$polls` array, test for array with few elements) for this method?

MasterOdin (changes requested):

site/app/controllers/PollController.php
```diff
@@ -398,4 +399,64 @@ public function viewResults($poll_id) {
             )
         );
     }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/polls/getPollExport", methods={"GET"})
+     * @AccessControl(role="INSTRUCTOR")
+     * @return JsonResponse
```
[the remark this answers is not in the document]
MasterOdin: This isn't asking expressly to change anything, just for clarification for myself to understand this code / purpose.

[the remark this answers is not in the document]
BenKelly99: It should just be for the ajax download. If there's a better way to make this function to reflect that, let me know and I can change it.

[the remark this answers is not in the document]
MasterOdin: You could mimic some of the functions in [MiscController](https://github.com/Submitty/Submitty/blob/master/site/app/controllers/MiscController.php) (like displayFile) where you turn off header/footer, manually set some headers to register it as a download, and then use `$this->core->getOutput()->renderString` to output the JSON. The JS could then be removed and replaced with a regular `<a>` tag. If you want to do it, that's fine, can also hold off and I think first put in place a generic `DownloadResponse` that makes doing this easier around the system.

Let me know which you'd prefer to do here.

migration/migrator/migrations/course/20210111163413_add_poll_data_folder.py
```diff
@@ -0,0 +1,23 @@
...
+    polls_dir.mkdir(mode=0o750, exist_ok=True)
+
+    php_user = config.submitty_users['php_user']
+
+    # get course group
+    course_group_id = course_dir.stat().st_gid
+    course_group = grp.getgrgid(course_group_id)[0]
+
+    # set the owner/group/permissions
+    os.system("chown -R "+php_user+":"+course_group+" "+str(polls_dir))
```
MasterOdin, line 19: We want to use built-in python stuff as much as possible so that we get error handling and such if things blow up. By using `os.system`, we would shallow any possible errors that might happen there. So for chown, we can use [`shutil.chown`](https://docs.python.org/3/library/shutil.html#shutil.chown). As this then relies on `os.chown`, we can also use the `Path` directly without having to first convert it to a string.

```suggestion
    shutil.chown(polls_dir, php_user, course_group_id)
```

MasterOdin (approved):

site/app/controllers/PollController.php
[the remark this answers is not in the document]
BenKelly99: I'll work on making a DownloadResponse. Seems like it will be useful.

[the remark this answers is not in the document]
BenKelly99: For the DownloadResponse, would it make sense for it to just take in a path to a file that will be downloaded? Then it can grab the mime info from the file and download it that way.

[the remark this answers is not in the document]
MasterOdin: Let's not do the DownloadResponse within this PR, and instead do it separately (after this is merged in the next day or so) as then that separate PR can address all of the places such a new class would benefit, and the class can be built to cater to all of them from the get-go. For doing that, can you spin this conversation into a new issue and then also include all of the functions that would benefit from this new DownloadResponse? This should then give us a good idea of where we would expect to utilize this to users. I expect that some of the methods would utilize a file (and so would use the php function `readfile` as well) while others are just going to output some string buffer (e.g. this method).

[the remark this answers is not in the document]
BenKelly99: Roger that. I've updated this PR to use the same method as MiscController::displayFile which should be easy to change over once the new Response type is made.

bmcutler:
@MasterOdin @BenKelly99 
Is this in a state to be merged & installed?

Also what is the timing to remove the feature flag gating on polling?
Instructors would like to use this for Spring 2021 term :)

BenKelly99:
@bmcutler I think it should be ready to go unless Matt has comments on the new HTTP header version of the download. I can remove the feature flag in this PR or open a separate PR for that purpose. Let me know which you would prefer.

bmcutler:
Please open another PR to remote the feature flag

PR #6134, opened 2021-01-11
[Bugfix:Autograding] config schema errors

MasterOdin:

The schema fails the strict validation of the [`ajv`](https://ajv.js.org/) tool. Most of the errors were either missing the type field or using `minLength` instead of `minItems` for an array field. The former error is something that `ajv` and `python-jsonschema` can paper over and infer if need be, while the latter was just ignored, but did allow empty lists in. The `acceptable_threshold` field was then placed outside the `properties` block and so was completely ignored for validation purposes.

### What is the new behavior?

Fixes the error so that our schema passes the strict validation mode of `ajv`.

### Other information?

We may wish to hook up the schema to be self-validated in some way by something like `ajv`, including in our test suite:

```
npm install -g ajv-cli
ajv compile -s bin/json_schemas/complete_config_schema.json
```


bin/json_schemas/complete_config_schema.json
```diff
@@ -676,9 +683,10 @@
             {
               "type" : "array",
               "items" : {
-                "type" : "string"
+                "type" : "string",
+                "minLength": 1
```
MasterOdin, line 687: The `minLength` is included here as the field above this (which is part of the containing `anyOf` is a string with `minLength: 1`.

emaicus (approved):
Looks good to me

PR #6135, opened 2021-01-12
[Bugfix:System] Fix building clang for static analysis

MasterOdin:

Running a fresh `vagrant up` was failing as the `http://lvm.org/git/*` endpoints are no longer available. These endpoints were a mirror of the sub-directories of the main llvm-project which were meant to only be available as long as it took for the transition from SVN to git (started in 2018).

### What is the new behavior?

Transitions to the clone to use the current home of the project, https://github.com/llvm/llvm-project. I also pinned the clone to use the latest minor release (7.1.0) of the project for when it was originally added, which was when 7.0.0rc1 was released, as I am not certain the tools are being tested against newer versions of clang, and so might of broken on 8+.

### DIFF
--- a/.setup/install_system.sh
+++ b/.setup/install_system.sh
@@ -609,10 +609,12 @@ if [ ! -d "${clangsrc}" ]; then
 
     mkdir -p ${clangsrc}
 
-    # checkout the clang sources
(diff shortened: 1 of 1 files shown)


vikrsri (approved):
Changes work as proposed, running a fresh `vagrant up` now works properly.

PR #6141, opened 2021-01-17
[Feature:Autograding] worker access via ssh with ProxyJump

bmcutler:
Previously the use of paramiko did not allow connections to machines that had been configured to go through an intermediary.  

### What is the new behavior?
Now, the code will read the ssh config file and correctly perform system updates/restarts and send jobs to a worker machine that is accessed indirectly (using proxyjump).

Documentation on submitty.org has been updated with configuration of the worker machine for SSH ProxyJump:

https://submitty.org/sysadmin/worker_installation
https://submitty.org/sysadmin/worker_installation#configuring-ssh-access
https://submitty.org/sysadmin/worker_installation#ssh-access-via-proxyjump

### DIFF
--- a/autograder/submitty_autograding_shipper.py
+++ b/autograder/submitty_autograding_shipper.py
@@ -32,16 +32,16 @@ from autograder import packer_unpacker
 from autograder import config as submitty_config
 
 
-import sys
-sys.path.append('/usr/local/submitty/sbin/shipper_utils/')
-import ssh_jump_proxy
-
-
 INTERACTIVE_QUEUE = ''
 IN_PROGRESS_PATH = ''
 JOB_ID = '~SHIP~'
 
 
+"""
+This function is copied, would rather use the version in this file instead.
+import sys
+sys.path.append('/usr/local/submitty/sbin/shipper_utils/')
+import ssh_jump_proxy
 """
 def ssh_connection_allowing_jump_proxy(target_user, target_host):
 
@@ -55,7 +55,7 @@ def ssh_connection_allowing_jump_proxy(target_user, target_host):
         conf.parse(open(config_file))
         o = conf.lookup(target_host)
         if 'proxyjump' in o:
-            (intermediate_user,intermediate_host) = str(o['proxyjump']).split('@')
+            (intermediate_user, intermediate_host) = str(o['proxyjump']).split('@')
         if 'identityfile' in o:
             my_key_filename = o['identityfile'][0]
 
@@ -67,9 +67,9 @@ def ssh_connection_allowing_jump_proxy(target_user, target_host):
             intermediate_connection = paramiko.SSHClient()
             intermediate_connection.get_host_keys()
             intermediate_connection.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-            intermediate_connection.connect(hostname=intermediate_host, 
-                                            username = intermediate_user,
-                                            timeout=60, 
+            intermediate_connection.connect(hostname=intermediate_host,
+                                            username=intermediate_user,
+                                            timeout=60,
                                             key_filename=my_key_filename)
             my_sock = intermediate_connection.get_transport().open_channel(
                 'direct-tcpip', (target_host, 22), ('', 0)
@@ -78,18 +78,17 @@ def ssh_connection_allowing_jump_proxy(target_user, target_host):
         target_connection = paramiko.SSHClient()
         target_connection.get_host_keys()
         target_connection.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-        target_connection.connect(hostname = target_host, 
-                                  username = target_user, 
-                                  timeout=60, 
-                                  sock=my_sock, 
+        target_connection.connect(hostname=target_host,
+                                  username=target_user,
+                                  timeout=60,
+                                  sock=my_sock,
                                   key_filename=my_key_filename)
         # return both connections (so they can be cleaned up)
-        return (target_connection,intermediate_connection)
+        return (target_connection, intermediate_connection)
 
     except Exception as e:
         print("ERROR: could not open a connection to the target host: "+str(e))
         raise e
-"""
 
 
 def instantiate_global_variables(config):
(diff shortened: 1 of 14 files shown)


autograder/submitty_autograding_shipper.py
```diff
@@ -37,6 +37,58 @@
 JOB_ID = '~SHIP~'
 
 
+# This function is copied, would rather use the version in this file instead.
```
bmcutler, line 40: @MasterOdin
If you tell me how I should properly refer to the python function in the other directory, I will update this code to remove repeated code.  What I was trying wasn't working.  Thanks.

autograder/submitty_autograding_shipper.py
```diff
@@ -37,6 +37,58 @@
...
+def ssh_connection_allowing_jump_proxy(target_user, target_host):
+
+    # load and parse the ssh config
+    my_key_filename = None
+    intermediate_host = None
+    intermediate_user = None
+    config_file = os.path.expanduser("~")+"/.ssh/config"
+    if (os.path.exists(config_file)):
+        conf = paramiko.SSHConfig()
+        conf.parse(open(config_file))
```
emaicus, line 53: A with statement would be preferable here to ensure the file is closed.

autograder/submitty_autograding_shipper.py
```diff
@@ -37,6 +37,58 @@
...
+        o = conf.lookup(target_host)
```
emaicus, line 54: What is `o`, exactly?

autograder/submitty_autograding_shipper.py
```diff
@@ -37,6 +37,58 @@
...
+        if 'proxyjump' in o:
+            (intermediate_user, intermediate_host) = str(o['proxyjump']).split('@')
+        if 'identityfile' in o:
+            my_key_filename = o['identityfile'][0]
```
emaicus, line 58: Is it possible for `o['proxyjump']` or `o['identityfile']` to be formatted in an unexpected way? (specifically, missing an `@` or not having index `0`? If so, does this need to go inside of a try/except, or do we expect that calls to this function should always be wrapped in a try/except?

sbin/shipper_utils/ssh_jump_proxy.py
```diff
@@ -0,0 +1,50 @@
+import os
```
emaicus, line 1: Same comments as above.

emaicus (changes requested):
Is there ever a case when we are using `intermediate_connection` for anything besides acquiring `target_connection`? If not, do we actually need to return it, or could we just close it inside of `ssh_connection_allowing_jump_proxy` and return`target_connection`?

autograder/submitty_autograding_shipper.py
```diff
@@ -37,6 +37,58 @@
...
+# import sys
+# sys.path.append('/usr/local/submitty/sbin/shipper_utils/')
+# import ssh_jump_proxy
+def ssh_connection_allowing_jump_proxy(target_user, target_host):
+
+    # load and parse the ssh config
+    my_key_filename = None
+    intermediate_host = None
+    intermediate_user = None
+    config_file = os.path.expanduser("~")+"/.ssh/config"
```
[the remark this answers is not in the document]
bmcutler: actually sometimes its root user, sometimes submitty_daemon.

autograder/submitty_autograding_shipper.py
bmcutler: I'm returning intermediate_connection because I am closing both connections once I'm done.
Just trying to ensure that resources aren't leaked.  If it will close automatically (or needs to be rewritten to close automatically) I'll update.

autograder/submitty_autograding_shipper.py
bmcutler: Yes, I should add error checking around these items...

autograder/submitty_autograding_shipper.py
MasterOdin: I would recommend moving this to `submitty_utils` as the point of this library was to make it easy to share code between various discrete components.

autograder/submitty_autograding_shipper.py
```diff
@@ -37,6 +37,58 @@
 JOB_ID = '~SHIP~'
 
 
+# This function is copied, would rather use the version in this file instead.
+# import sys
+# sys.path.append('/usr/local/submitty/sbin/shipper_utils/')
+# import ssh_jump_proxy
```
hdrodz97, line 43: If you do go this route, you're probably gonna want to replace
```python
import ssh_jump_proxy
```
With:
```python
from shipper_utils import ssh_jump_proxy
```

bmcutler:
@hdrodz97 @emaicus 

I've updated the documentation here:
https://submitty.org/sysadmin/worker_installation
https://submitty.org/sysadmin/worker_installation#configuring-ssh-access
https://submitty.org/sysadmin/worker_installation#ssh-access-via-proxyjump

Standard(?) stuff -- mostly taking notes for myself for next time I have to set this up...

Let me know if you have comments on the documentation.

autograder/submitty_autograding_shipper.py
bmcutler: Thanks.  I believe I've got it as you suggest and there is no more code duplication - yay

autograder/submitty_autograding_shipper.py
bmcutler: changed!

autograder/submitty_autograding_shipper.py
```diff
@@ -37,6 +37,58 @@
...
+def ssh_connection_allowing_jump_proxy(target_user, target_host):
+
+    # load and parse the ssh config
+    my_key_filename = None
+    intermediate_host = None
+    intermediate_user = None
+    config_file = os.path.expanduser("~")+"/.ssh/config"
+    if (os.path.exists(config_file)):
```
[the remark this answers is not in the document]
bmcutler: changed!

autograder/submitty_autograding_shipper.py
bmcutler: changed!

autograder/submitty_autograding_shipper.py
bmcutler: renamed variable

bmcutler:
@emaicus @hdrodz97 @MasterOdin 
Thanks!  I think I've address all feedback.

PR #6145, opened 2021-01-19
[Feature:InstructorUI] Add registration subsection

vikrsri:

There is currently no way to put students into subsections, allowing further sorting of students.
Addresses #6064.

### What is the new behavior?
This PR adds a registration subsection that can be set by an instructor. Further PR's will allow sorting and grading by students' registration subsections.

The subsection can be set when an instructor either edits or adds a new student in the Manage Students page. If a subsection has been set for a student, it will be appended to the student's registration section in the Manage Students page. Otherwise, there will be no change.


migration/migrator/migrations/course/20210113174909_registration_subsection.py
```diff
@@ -0,0 +1,7 @@
+
+def up(config, database, semester, course):
+    database.execute('ALTER TABLE users ADD COLUMN registration_subsection character varying(255)')
+
+
+def down(config, database, semester, course):
+    database.execute('ALTER TABLE users DROP IF EXISTS registration_subsection')
```
bmcutler, line 7: I don't think it should be necessary to delete this column in the down migration (we don't want to lose data if we temporarily revert this PR).

migration/migrator/migrations/course/20210113174909_registration_subsection.py
```diff
@@ -0,0 +1,7 @@
+
+def up(config, database, semester, course):
+    database.execute('ALTER TABLE users ADD COLUMN registration_subsection character varying(255)')
```
bmcutler, line 3: Make this an 'add if does not exist' 
(see other migration files for exact syntax)

bmcutler (changes requested):

bmcutler (approved):
Quick code review and quick test look ok!  Thanks!
This is what we agreed for the step 1 PR.

@MasterOdin -- @vikrsri and I discussed the unexpected additional edits to the .sql file (changes were made using the postgres dump instructions... ).  Can you please review and advise?

site/app/models/User.php
```diff
@@ -388,6 +394,10 @@ public function updateUserNotificationSettings($key, $value) {
         $this->notification_settings[$key] = $value;
     }
 
+    public function getRegistrationSubsection() {
+        return $this->registration_subsection;
+    }
```
MasterOdin, line 399: you can get rid of this function, it's not needed (as using `@prop` annotation automatically makes available this getter field.

site/app/models/User.php
```diff
@@ -418,6 +428,10 @@ public function setRegistrationSection($section) {
         $this->registration_section = ($section !== null) ? $section : null;
     }
 
+    public function setRegistrationSubsection($section) {
+        $this->registration_subsection = $section;
+    }
+
```
MasterOdin, line 434: you can get rid of this function, it's not needed (as using `@prop` annotation automatically makes available this setter field.

MasterOdin (approved):

MasterOdin:
The SQL file changes are fine. The `gradeable_access` table was added by hand in #5623 and then no one has run the dump DB command since then. The reason for that it looks like `gradeable_access`'s schema changed is that `SERIAL` is a pseudo-type in postgresql that translates under the hood to creating a sequence, assigning ownership, etc. It's a shorthand that the pg_dump command does not use, as it is more inline with how the data structures are actually stored in the DB.

PR #6150, opened 2021-01-20
[Feature:InstructorUI] Revise JSON file formats of Subminis

### LINKED ISSUE #6149: Revise json file formats of the poll/subminis
Now that I'm working with actual course data, I have feedback for the exported formats
(let's try to get these changes in quickly)

* exported file names are non-intuitive.  let's revisit.  how about poll_questions.json and poll_responses.json?  

* the "data" file is missing "status" ("open", "closed", "ended" ... )

* why is top level an associative array in the "data" file, but a simple array with "id" field in "responses" file.  probably just make both an array.

* id is a string in the data/questions file, but an integer in poll_responses.  let's be consistent.

* order the items starting with the smallest id (currently it appears to be starting with the largest)

BenKelly99:
Closes #6149

### What is the new behavior?
Now, the file names that are generated when generate grade summaries are run are poll_questions.json and poll_responses.json
Poll statuses are now exported
Both files use simple arrays with an id field
IDs are now all ints instead of sometimes being strings
Reorders polls to be older (created first) polls are at the top. This also reorders them on the main poll page.


MasterOdin (approved):
Code looks fine. Adding @bmcutler as reviewer for approval on formatting and filename details.

bmcutler (approved):
Looks great!  Thanks!

PR #6153, opened 2021-01-22
[Feature:RainbowGrades] version update for Submini polls

bmcutler:
Update rainbow grades to support polls.
NOTE:  The local directory structure for running rainbow grades for a specific course has changed slightly to place the student grading files in a subdirectory, copying more of the directory structure from the server.


BenKelly99:
I tried running rainbow grades after installing the new version and got the following error:
[error.txt](https://github.com/Submitty/Submitty/files/5854345/error.txt)

PR #6162, opened 2021-01-26
[Feature:Autograding] student git repos use main branch

bmcutler:
Autograding will now expect student work to be in a main branch.
If that branch does not exist it will default to the now deprecated master branch.

note: this is work-in-progress, more testing and error messages will be added before it is finalized & merged.


emaicus (approved):
I verified that:
1. When there is a master branch and no main branch, the master branch is checked out
2. When there is a main branch, it is checked out

I wondered what would happen if cases where neither branch exists, but it is suitably difficult to delete a master branch that I doubt a student will accidentally end up in that state.

PR #6175, opened 2021-02-01
[UI/UX:Submission] Show graded version worker information

hdrodz97:

Currently, no information is given by Submitty as to which worker machine has graded a gradeable. This can make debugging worker setups with many different worker hardware configurations difficult, as sometimes bugs may manifest on one worker but not another (e.g. a graphics assignment being graded differently on different workers due to worker hardware).

### What is the new behavior?

A gradeable's version information box now displays the name of the worker that graded that specific version alongside other information such as queue wait time, first access timestamp, etc. There are three different output cases:

* Output in most cases:
image

* Output on a graded version which does not have associated worker machine information (e.g., it was graded before this information was added to the stored queue file):
image

* Output on a graded version that has been short-circuited by the autograder:
image

### DIFF
--- a/site/app/templates/submission/homework/CurrentVersionResults.twig
+++ b/site/app/templates/submission/homework/CurrentVersionResults.twig
@@ -53,17 +53,14 @@
                     <span>Grading time:</span>
                     <span>{{ grade_time }} seconds</span>
                 </span>
-                {% if autograder_machine is not null %}
-                    <span class="flex-row">
-                        <span>Graded on machine:</span>
+                <span class="flex-row">
+                    <span>Graded on machine:</span>
+                    {% if autograder_machine is not null %}
                         <span>{{ autograder_machine }}</span>
-                    </span>
-                {% else %}
-                    <span class="flex-row">
-                        <span>Graded on machine:</span>
+                    {% else %}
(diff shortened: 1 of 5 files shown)


site/app/templates/submission/homework/CurrentVersionResults.twig
```diff
@@ -53,6 +53,17 @@
...
+                    <span class="flex-row">
+                        <span>Graded on machine:</span>
+                        <span>{{ autograder_machine }}</span>
+                    </span>
+                {% else %}
+                    <span class="flex-row">
+                        <span>Graded on machine:</span>
+                        <span>(unknown)</span>
+                    </span>
+                {% endif %}
```
shailpatels, line 66: You can remove the duplication in html with the twig ternary statement
```html
<span class="flex-row">
   <span> Graded on machine: </span>
   <span> {{ (autograder_machine is not null) ? autograder_machine : '(unknown)' </span>
</span>
```

Alternatively, you can move the if statement inside the inner `<span>` after the string "Graded on machine"

shailpatels (changes requested):

bmcutler (approved):

PR #6180, opened 2021-02-04
[Feature:SubminiPolls] Add Images to Submini Polls

BenKelly99:
### What is the new behavior?
Instructors now have the ability to add images to a Submini poll. This image is displayed to students when they answer the polls. The images are stored in a new folder in the course `uploads` folder called `polls`. The names of the files are based on both poll ID and the name of the file the instructor uploaded. Specifically, the format is `poll_image_{poll_id}_{name_of_uploaded_file}`.

### Other small changes
Updates the sidebar to read "Submini Polls" instead of "Polls"
Update HTML in poll page to remove unneeded tag


site/app/controllers/GlobalController.php
```diff
@@ -166,7 +166,7 @@ public function prep_course_sidebar(&$sidebar_buttons, $unread_notifications_cou
         if ($this->core->getConfig()->isPollsEnabled()) {
             $sidebar_buttons[] = new Button($this->core, [
                 "href" => $this->core->buildCourseUrl(['polls']),
-                "title" => "Polls",
+                "title" => "Submini Polls",
```
MasterOdin, line 169: Can we pull this into its own PR? This seems unrelated to the rest of the changeset, and so in-case we have to revert the rest of this PR, we wouldn't want to revert this change as well.

site/app/models/PollModel.php
```diff
@@ -15,8 +15,9 @@ class PollModel extends AbstractModel {
     protected $user_responses;
     protected $release_date;
     protected $status;
+    protected $image_path;
```
MasterOdin, line 18: You should mark this as a `/** @prop-read */` and can then delete `getImagePath()` function and take advantage of the automatically provided getters from `AbstractModel` as described [here](https://submitty.org/developer/developing_the_php_site/models).

site/app/views/PollView.php
```diff
@@ -58,11 +58,18 @@ public function showPollStudent($poll) {
         $this->core->getOutput()->addBreadcrumb("View Poll");
         $this->core->getOutput()->addInternalCss('polls.css');
         $this->core->getOutput()->enableMobileViewport();
+        $image_path = $poll->getImagePath();
+        $file_data = null;
+        if ($image_path != null) {
```
MasterOdin, line 63:
```suggestion
        if ($image_path !== null) {
```

site/app/controllers/PollController.php
```diff
@@ -363,6 +378,10 @@ public function deletePoll() {
             $this->core->addErrorMessage("Invalid Poll ID");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
+        $image_path = $this->core->getQueries()->getPoll($_POST["poll_id"]);
+        if ($image_path != null) {
```
MasterOdin, line 382:
```suggestion
        if ($image_path !== null) {
```

site/app/controllers/PollController.php
```diff
@@ -363,6 +378,10 @@ public function deletePoll() {
             $this->core->addErrorMessage("Invalid Poll ID");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
+        $image_path = $this->core->getQueries()->getPoll($_POST["poll_id"]);
```
MasterOdin, line 381:
```suggestion
        $image_path = $this->core->getQueries()->getPoll($_POST["poll_id"])->getImagePath();
```

site/app/controllers/PollController.php
```diff
@@ -329,6 +338,12 @@ public function submitEdits() {
                 new RedirectResponse($this->core->buildCourseUrl(['polls']))
             );
         }
+        $file_path = null;
+        if (array_key_exists("image_file", $_FILES) && $_FILES["image_file"]["name"] != "") {
```
MasterOdin, line 342: same as above.

site/app/controllers/PollController.php
```diff
@@ -152,7 +154,14 @@ public function addNewPoll() {
                 $answers[] = $_POST["option_id_" . $i];
             }
         }
-        $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
+        $poll_id = $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
+        $file_path = null;
+        if (array_key_exists("image_file", $_FILES) && $_FILES["image_file"]["name"] != "") {
+            $file = $_FILES["image_file"];
+            $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $poll_id . "_" . $_FILES["image_file"]["name"]);
```
MasterOdin, line 161: Why is maintaining the name of the file important on the filesystem?

site/app/controllers/PollController.php
```diff
@@ -152,7 +154,14 @@ public function addNewPoll() {
                 $answers[] = $_POST["option_id_" . $i];
             }
         }
-        $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
+        $poll_id = $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
+        $file_path = null;
+        if (array_key_exists("image_file", $_FILES) && $_FILES["image_file"]["name"] != "") {
```
MasterOdin, line 159: Probably safer to use `isset` here as if `$_FILES['image_file] === null`, then this if statement would pass. `isset` passes only if the key exists and its value is not null, plus it's less characters.

```suggestion
        if (isset($_FILES['image_file']) && $_FILES["image_file"]["name"] !== "") {
```

MasterOdin (changes requested):

site/app/controllers/PollController.php
bmcutler: This was my suggestion.  Eventually when we export the images associated with polls, it will be helpful to have the original names as part of the filenames, they might be meaningful to the instructor.

MasterOdin: That's fair. I would greatly prefer that we utilize the new `image_path` DB field to store the name that an instructor uses and that on the filesystem itself we just store it as `FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $poll_id)`. On export, we can provide the original name from the DB, while on the filesystem itself, we do not have to worry about the instructor uses god knows what characters.

BenKelly99: Sounds good to me

site/app/controllers/GlobalController.php
BenKelly99: I'll put this in another PR

site/app/controllers/PollController.php
BenKelly99: @MasterOdin I talked to Barb about this and she suggested we keep the current format but remove any special characters from the provided name first (just allow letters, numbers, some characters like _, -, etc). Do you think this makes sense?

site/app/templates/polls/PollPageStudent.twig
```diff
@@ -1,6 +1,9 @@
 <div class="content">
     <h1> {{ poll.getName() }} </h1>
-    <p> {{ poll.getQuestion() | markdown}} </p>
+    {{ poll.getQuestion() | markdown}}
+    {% if not file_data is same as(null) %}
```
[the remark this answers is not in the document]
BenKelly99: For some reason I had a memory the same as() check was necessary. I'll change it

site/app/controllers/PollController.php
MasterOdin: So the philosophy I would cater towards is that it's always better to store a filename within a database and then use a simple, unique identifier for the file on-disk (or in this case `"poll_image_" . $poll_id` which we let get overwritten on each new upload). It simplifies certain elements of handling files (e.g. on upload, we can remove the explicit `unlink` step for the old file, and just have the filesystem overwrite on copy) and in a more generic world where an upload folder is more flat, upload name collision attacks.

This isn't a hill I'm going to die on though, and if you or @bmcutler feel like you want to have the name on disk, so be it.

site/app/controllers/PollController.php
```diff
@@ -329,6 +338,12 @@ public function submitEdits() {
                 new RedirectResponse($this->core->buildCourseUrl(['polls']))
             );
         }
+        $file_path = null;
+        if (isset($_FILES['image_file']) && $_FILES["image_file"]["name"] !== "") {
+            $file = $_FILES["image_file"];
+            copy($file["tmp_name"], $file_path);
+            $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $_POST["poll_id"] . "_" . $_FILES["name"]);
+        }
```
MasterOdin, line 346: This method should probably be modified to remove the old poll image, unless we want to leave these files on the filesystem for some reason (though not reachable beyond ssh)?

site/app/libraries/PollUtils.php
```diff
@@ -23,13 +23,14 @@ public static function getPollExportData(array $polls): array {
...
-                "id" => $poll->getID(),
+                "id" => $poll->getId(),
                 "name" => $poll->getName(),
                 "question" => $poll->getQuestion(),
                 "responses" => $responses,
                 "correct_responses" => $poll->getAnswers(),
                 "release_date" => $poll->getReleaseDate(),
-                "status" => $poll->getStatus()
+                "status" => $poll->getStatus(),
+                "image_file" => $poll->getImagePath()
```
MasterOdin, line 33: Would it make sense to make this be the base64 image string for the export, so that it would make it possible to import the poll and get the same image?

MasterOdin (changes requested):

site/app/libraries/PollUtils.php
BenKelly99: Yeah the original plan was to leave the import/export part for images to another time but I think this works pretty well. I'll confirm with barb tomorrow.

BenKelly99: Talked with Barb -- she wants the export file to remain as human readable as possible so she'd rather ignore the import/export for now and then go back later and have the download be a zip file containing the data and the images themselves.

site/app/controllers/PollController.php
```diff
@@ -333,6 +344,16 @@ public function submitEdits() {
                 new RedirectResponse($this->core->buildCourseUrl(['polls']))
             );
         }
+        $file_path = null;
+        if (isset($_FILES['image_file']) && $_FILES["image_file"]["name"] !== "") {
+            $file = $_FILES["image_file"];
+            $current_file_path = $this->core->getQueries()->getPoll($_POST["poll_id"])->getImagePath();
+            if ($current_file_path != null) {
```
MasterOdin, line 351:
```suggestion
            if ($current_file_path !== null) {
```

I really need to activate this lint rule. 🤔

site/app/controllers/PollController.php
```diff
@@ -152,11 +154,20 @@ public function addNewPoll() {
...
             new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
-        $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
+
+        $poll_id = $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
+        $file_path = null;
+        if (isset($_FILES['image_file']) && $_FILES["image_file"]["name"] !== "") {
+            $file = $_FILES["image_file"];
+            $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $poll_id . "_" . $_FILES["image_file"]["name"]);
+            copy($file["tmp_name"], $file_path);
```
MasterOdin, line 168: Better to use [`move_uploaded_file`](https://www.php.net/manual/en/function.move-uploaded-file.php) when dealing with uploading files that need to get moved into the filesystem

```suggestion
            move_uploaded_file($file["tmp_name"], $file_path);
```

site/app/controllers/PollController.php
```diff
@@ -333,6 +344,16 @@ public function submitEdits() {
...
         }
+        $file_path = null;
+        if (isset($_FILES['image_file']) && $_FILES["image_file"]["name"] !== "") {
+            $file = $_FILES["image_file"];
+            $current_file_path = $this->core->getQueries()->getPoll($_POST["poll_id"])->getImagePath();
+            if ($current_file_path != null) {
+                unlink($current_file_path);
+            }
+            $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $_POST["poll_id"] . "_" . $_FILES["image_file"]["name"]);
+            copy($file["tmp_name"], $file_path);
```
MasterOdin, line 355:
```suggestion
            move_uploaded_file($file["tmp_name"], $file_path);
```

site/app/libraries/PollUtils.php
MasterOdin: Alright, I think it might make sense to rename this output to `image_path` to better represent what this is, but a small nitpick, and I do not really care.

site/app/controllers/PollController.php
```diff
@@ -152,11 +154,20 @@ public function addNewPoll() {
...
-        $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
+
+        $poll_id = $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
+        $file_path = null;
+        if (isset($_FILES['image_file']) && $_FILES["image_file"]["name"] !== "") {
+            $file = $_FILES["image_file"];
+            $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $poll_id . "_" . $_FILES["image_file"]["name"]);
+            copy($file["tmp_name"], $file_path);
+        }
+        $this->core->getQueries()->setPollImage($poll_id, $file_path);
```
MasterOdin, line 170: No need to run this query if we don't have an image. Can delete the `$file_path = null`; above the if statement as well.

```suggestion
            $this->core->getQueries()->setPollImage($poll_id, $file_path);
        }
```

MasterOdin (approved):
small nitpicks, otherwise LGTM.

PR #6184, opened 2021-02-08
[Refactor:TAGrading] Subsection Default Value Change

vikrsri:

Currently, the default value of the registration subsection is null.

### What is the new behavior?
The default value is now an empty string, and it cannot be null.


migration/migrator/migrations/course/20210204140110_modify_subsection.py
```diff
@@ -0,0 +1,3 @@
+
+def up(config, database, semester, course):
+    database.execute('ALTER TABLE users ALTER COLUMN registration_subsection SET DEFAULT '' NOT NULL')
```
MasterOdin, line 3: You probably need to do the following sequence to deal with any existing null values in the DB:
```suggestion
    database.execute("UPDATE users SET registration_subsection = '' WHERE registration_subsection IS NULL")
    database.execute('ALTER TABLE users ALTER COLUMN registration_subsection SET DEFAULT '' NOT NULL')
```

MasterOdin: Should include a down migration as well to reverse the column default / not null change. Don't need to include an update though to reverse `''` into nulls as they're interchangeable in the system.

MasterOdin (changes requested):

migration/migrator/data/course_tables.sql
```diff
@@ -1062,7 +1062,7 @@ CREATE TABLE public.users (
     last_updated timestamp(6) with time zone,
     time_zone character varying DEFAULT 'NOT_SET/NOT_SET'::character varying NOT NULL,
     display_image_state character varying DEFAULT 'system'::character varying NOT NULL,
-    registration_subsection character varying(255),
+    registration_subsection character varying(255) DEFAULT (NOT NULL::boolean),
```
vikrsri, line 1065: For some reason when I ran the course dump, it set the default to NOT NULL::boolean. Do you know why this would be happening @MasterOdin ?

MasterOdin: It's because your alter table statement isn't right. With postgres, you can only change one aspect of a column at a time, and if you want to change multiple things, you need multiple alters. By doing `SET DEFAULT '' NOT NULL`, postgres is doing a weird transformation of `'' NOT NULL` into a singular default value which ends up being `(NOT NULL::boolean)`.

I've updated my suggestion below with what I think the command should be, though did not test against actual DB.

MasterOdin (approved):

migration/migrator/data/course_tables.sql
MasterOdin: With your latest change, my guess would be that you were getting a weird syntax quirk of using the single quotes to define your python string literal as well as default value inside that string. I made a small commit that seems to have fixed this and I am getting the expected result on dumping the migrated schema.

PR #6191, opened 2021-02-10
[Refactor:Autograding] Move scheduler into own class

hdrodz:

Currently, the autograding shipper's scheduling behavior all happens inline in the `monitoring_loop` function in `submitty_autograding_shipper.py`. This makes scheduling behavior harder to test in isolation.

### What is the new behavior?

This PR refactors the scheduling logic of the shipper into its own set of classes: `BaseScheduler`, which encapsulates core functionality for all scheduling algorithms (e.g. getting the current list of jobs available), and `FCFSScheduler`, which implements the basic first come, first serve scheduling behavior. This will also allow for extending the scheduler with different user-configurable scheduling behaviors.

Furthermore, unit tests now exist for the shipper scheduler.


hdrodz97:
Opening this up for review since it seems to be working as expected now

autograder/autograder/scheduler.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    def __init__(self, path: str):
+        self.path = path
+        self._queue_obj = None
+
+    @property
+    def queue_obj(self) -> dict:
+        if self._queue_obj is None:
+            with open(self.path) as f:
+                self._queue_obj = json.load(f)
```
emaicus, line 34: Should we handle errors here, or expect calls to this to be wrapped in a try/except?

emaicus (changes requested):
A small, cosmetic tweak: : it looks like the shipper process must be writing a json file to `stdout`, as I can see it when I run a `systemctl status` on the shipper process.

I initially spotted another error, but it was specific to my system rather than this PR.

emaicus (changes requested):
I have found two additional bugs while testing this PR, which may or may not be directly related. 
1. If grading fails at the python level (e.g., `grade_item.py` returns an error) the shipper re-trys to grade the gradeable. In particular, if an assignment is run which requires a docker container which is not installed on the local system, grading will re-try ad infinitum, causing infinite output in the error log.
2. I suspect that this one is directly linked to this PR. After grading for a time, my shipper process began to run out of control. Looking at the output of `top`, it was maxing out my system's CPU when grading ordinary assignments, causing system freeze and vagrant failure.

hdrodz97:
@emaicus Would you happen to have any of the shipper log output around the time the shipper went haywire? Hopefully something was being printed which could help narrow down where things went wrong.

autograder/autograder/scheduler.py
hdrodz97: Fixed

williamjallen (approved):
I have reviewed the code and run the tests, as well as some manual testing.  I am not particularly familiar with this part of the codebase so I don't know all of the edge cases which may occur and thus, my testing is probably not completely thorough.  I was unable to generate any unexpected behavior and the code looks clean so I'll approve.

PR #6204, opened 2021-02-15
[Testing:Developer] Move e2e tests to Github actions

shailpatels:

### What is the new behavior?
This PR moves e2e & integration tests from travis to github actions.
Travis stages that have already been moved over (phpunit, linting, etc) have been removed
Integration tests have been moved to their own workflow (CI-3) which wasn't working anyways.

Currently, the only thing running on Travis is the git tests which need to be debugged further when moving over to gh actions


.travis.yml
```diff
@@ -292,13 +189,13 @@ jobs:
...
       script:
-        - echo "Authentication Method => $(sudo jq -r ".authentication_method" /usr/local/submitty/config/database.json)"
-        - TEST_URL="http://localhost" python3 -m unittest discover -v --start-directory tests
-        - sudo sed -ie "s/Pam/Database/g" /usr/local/submitty/config/database.json
-        - echo "Authentication Method => $(sudo jq -r ".authentication_method" /usr/local/submitty/config/database.json)"
-        - pushd tests
-        - TEST_URL="http://localhost" python3 -m unittest e2e.test_login
-        - popd
+       # - echo "Authentication Method => $(sudo jq -r ".authentication_method" /usr/local/submitty/config/database.json)"
+       # - TEST_URL="http://localhost" python3 -m unittest discover -v --start-directory tests
```
MasterOdin, line 193: If having Travis do all this work anyway, might as well have it still test PAM authentication on the login tests:

```suggestion
       - echo "Authentication Method => $(sudo jq -r ".authentication_method" /usr/local/submitty/config/database.json)"
       - pushd tests
       - TEST_URL="http://localhost" python3 -m unittest e2e.test_login
       - popd
```

.setup/testing/autograder.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
```
MasterOdin, line 1: Can we combine `.setup/testing/*` and `.setup/travis/*` files? Having them separate means we risk updating one and not the other until Travis is completely killed off which I'd like to avoid as it's added mental load, and I do not believe the files do that much that's different between the two. At worst, can probably gatekeep based on the `${TRAVIS}` env variable.

python_submitty_utils/tests/test_submitty_schema_validator.py
```diff
@@ -25,6 +25,7 @@ def tearDown(self):
 
     def test_valid_schemas(self):
         for entry in CONFIGS_DIR.iterdir():
+            print('testing ', entry)
```
MasterOdin, line 28: remove?

sbin/create_course.sh
```diff
@@ -293,7 +293,7 @@ if [[ $? -ne "0" ]] ; then
     exit
 fi
 
-python3 ${SUBMITTY_REPOSITORY_DIR}/migration/run_migrator.py -e course --course ${semester} ${course} migrate --initial
+sudo -E env "PATH=$PATH" python3 ${SUBMITTY_REPOSITORY_DIR}/migration/run_migrator.py -e course --course ${semester} ${course} migrate --initial
```
MasterOdin, line 296: Is this change necessary?

site/composer.lock
```diff
@@ -63,30 +63,34 @@
                 "markdown",
```
MasterOdin, line 63: not sure why you're updating this file in this PR if you're not also updating the composer.json file. Should revert this change unless there is a necessary change to composer.json.

.github/workflows/submitty_ci_2.yml
```diff
@@ -0,0 +1,216 @@
...
+              composer install --prefer-dist
+              popd
+          - name: Install Submitty python utils
+            run: |
+              export SUBMITTY_INSTALL_DIR=/usr/local/submitty 
+              export SUBMITTY_REPOSITORY=${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/Submitty
+              pushd ${SUBMITTY_REPOSITORY}/python_submitty_utils
+              pip3 install .
+              pip3 show submitty_utils
+              sudo apt-get install python3-setuptools
```
MasterOdin, line 133: Should either do `pip3 install -U setuptools` or `apt-get install python3-setuptools`. Doing both can lead to potential collisions / hard to debug issues as things are on different versions.

.github/workflows/submitty_ci_2.yml
```diff
@@ -0,0 +1,216 @@
...
+                ${{ runner.os }}-php-composer-
+          - name: Install PHP dependencies
+            run: |
+              pushd ${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/Submitty/site
+              composer install --prefer-dist
+              popd
+          - name: Install Submitty python utils
+            run: |
+              export SUBMITTY_INSTALL_DIR=/usr/local/submitty 
+              export SUBMITTY_REPOSITORY=${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/Submitty
```
MasterOdin, line 129: I'd rather see these exports just moved into the `env` map at the top of the file, with maybe a comment saying these variables should have same value as `SUBMITTY_INSTALL_DIR`, like:

```
env:
    # ...
    SUBMITTY_INSTALL_DIR: /usr/local/submitty
    # The following variables should all use a stem that matches SUBMITTY_INSTALL_DIR above
    SUBMITTY_REPOSITORY: /usr/local/submitty/GIT_CHECKOUT/Submitty
```

Otherwise, if we update one `SUBMITTY_REPOSITORY` definition, we've got to update multiple, and DRY.

.github/workflows/submitty_ci_2.yml
```diff
@@ -0,0 +1,216 @@
+name: Submitty CI-2
```
MasterOdin, line 1: Many of the steps here do multiple different things. We should break this up into individual run commands, leaving groupings together as makes sense. For example, we should run all pip installs in one run, all apt-gets in one run, etc., while then doing stuff like "setup and populate DB" would be its own step, running the setup scripts can be its own step, etc. I can give more concrete examples per run command if so desired.

Doing this will make it easier at a glance to tell what went wrong and debug. As GitHub does not have a sense of "pipeline steps" like Travis does, we really do just want to get fine-grained.

.github/workflows/submitty_ci_2.yml
```diff
@@ -0,0 +1,216 @@
...
+          - uses: nanasess/setup-chromedriver@master
+          - name: Run tests
+            run: |
+              sudo journalctl --rotate
+              sudo journalctl --vacuum-time=1s
+              sudo systemctl restart submitty_autograding_worker
+              sudo systemctl restart submitty_autograding_shipper
+              export DISPLAY=:99
+              chromedriver --url-base=/wd/hub &
+              sudo Xvfb -ac :99 -screen 0 1280x1024x24 > /dev/null 2>&1 & # optional
```
MasterOdin, line 206: We should not need `Xvfb` or to export `DISPLAY` as we're running headless.

MasterOdin (changes requested):

.github/workflows/submitty_ci_2.yml
shailpatels: When you say remove individual run commands do you mean get rid of the `run |` and replace it with a single line 
`run` ?

.setup/testing/autograder.sh
shailpatels: I'd like to keep them separate so when travis gets removed I can just delete the entire directory, it would also save me time in debugging trying to merge them back together which I don't think will really be worth it because the last two tests are integration and the git tests which I expect to be moved over pretty soon.

But yeah they are pretty similar maybe I can get away with defining 'travis' only blocks that can easily be removed once we're done with it.

.github/workflows/submitty_ci_2.yml
shailpatels: Yeah the `SUBMITTY_REPOSITORY` should be a top level `env` variable, I remember running into an issue where some of the bash scripts couldn't see those env variables unless they were defined with an export locally in the same run step, but that could have been a fluke from something else? I'll have to remove and re-test it

.github/workflows/submitty_ci_2.yml
MasterOdin: Changing:

```
- name: Install Submitty and populate database
            run: |
              PGPASSWORD=${PGPASSWORD} psql -d postgres -h localhost -U postgres -c "CREATE ROLE submitty_dbuser WITH SUPERUSER CREATEDB CREATEROLE LOGIN PASSWORD 'submitty_dbuser'"
              PGPASSWORD=${PGPASSWORD} psql -d postgres -h localhost -U submitty_dbuser -c "CREATE DATABASE submitty"
              sudo apt-get update
              sudo apt-get install libseccomp-dev
              sudo apt-get install libboost-all-dev
              sudo apt-get install apache2
              sudo apt-get install apache2-suexec-custom
              sudo apt-get install libapache2-mod-authnz-external
              sudo apt-get install libapache2-mod-authz-unixgroup
              sudo apt-get install libapache2-mod-wsgi-py3
              sudo apt-get install nginx
              sudo apt-get install php${PHP_VER}-fpm
              export SUBMITTY_INSTALL_DIR=/usr/local/submitty #need to export otherwise shell scripts don't have access
              export SUBMITTY_REPOSITORY=${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/Submitty
              pushd ${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/Submitty
              sudo -E env "PATH=$PATH" bash .setup/testing/autograder.sh
              sudo -E env "PATH=$PATH" bash .setup/testing/setup.sh
              popd
```

to be more like:

```
- name: Setup Database
  run: |
      PGPASSWORD=${PGPASSWORD} psql -d postgres -h localhost -U postgres -c "CREATE ROLE submitty_dbuser WITH SUPERUSER CREATEDB CREATEROLE LOGIN PASSWORD 'submitty_dbuser'"
      PGPASSWORD=${PGPASSWORD} psql -d postgres -h localhost -U submitty_dbuser -c "CREATE DATABASE submitty"

- name: Install apt-get packages
  run: |
      sudo apt-get update
      sudo apt-get install libseccomp-dev
      sudo apt-get install libboost-all-dev
      sudo apt-get install apache2
      sudo apt-get install apache2-suexec-custom
      sudo apt-get install libapache2-mod-authnz-external
      sudo apt-get install libapache2-mod-authz-unixgroup
      sudo apt-get install libapache2-mod-wsgi-py3
      sudo apt-get install nginx
      sudo apt-get install php${PHP_VER}-fpm

- name: Setup third party autograding dependencies
  run: sudo -E env "PATH=$PATH" bash .setup/testing/autograder.sh

- name: Populate Submitty
  run: sudo -E env "PATH=$PATH" bash .setup/testing/setup.sh
```

Each of these commands are individual from each other and do not rely on exports or anything from the previous step. Separating them makes it easier to say why a build failed (was it that the database failed to initialize? apt-get install error? etc) without having to dig into the full log for that line.

shailpatels: Ok makes sense, thanks

.setup/testing/autograder.sh
```diff
@@ -1,5 +1,27 @@
 #!/usr/bin/env bash
 
+
+# This function is from https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/bash/travis_retry.bash
+wget_retry() {
```
MasterOdin, line 5: Need to add:

```
  ANSI_RED="\033[31;1m"
  ANSI_RESET="\033[0m"
```

if we want to keep the color stuff.

MasterOdin (commented):
Would be good to make sure we're using the same PHP version throughout.

shailpatels:
> Would be good to make sure we're using the same PHP version throughout.

I'm planning on adding the specific php-version in a later PR (its one of the TODOs on this issue: #6212) 
it looks like actions/setup-node@v2.1.2 has been defaulting to PHP 8, which our codebase has been testing against and there are a couple of unit test and linting issues if it gets reverted to 7.2:
https://github.com/Submitty/Submitty/runs/1942280084?check_suite_focus=true

That'll require some php-code changes so I think should be apart of its own PR.

MasterOdin:
> > Would be good to make sure we're using the same PHP version throughout.
> 
> I'm planning on adding the specific php-version in a later PR (its one of the TODOs on this issue: #6212)
> it looks like actions/setup-node@v2.1.2 has been defaulting to PHP 8, which our codebase has been testing against and there are a couple of unit test and linting issues if it gets reverted to 7.2:
> https://github.com/Submitty/Submitty/runs/1942280084?check_suite_focus=true
> 
> That'll require some php-code changes so I think should be apart of its own PR.

That's not right. The run you link to is using PHP 8 which is causing the failures (see the header output of phpunit where it states `Runtime:       PHP 8.0.2`). Using PHP 7.4 (which is the default for `shivammathur/setup-php` if no version is specified) or PHP 7.2 does not have any lint / test failures. In looking into this, the proper syntax as well to specify the version from an environment variable is:

```
            - uses: shivammathur/setup-php@2.7.0
              with:
                  php-version: ${{ env.PHP_VER }}
```

doing `php-version: $PHP_VER` is incorrect and causes PHP8 to be installed, which as shown above causes test failures (which we should not bother dealing for now).

Additionally, if we specify `7.2` as the version, it'd probably be good to also specify `ubuntu-18.04` as the used version throughout for the OS. It looks like the default extension list for `shivammathur/setup-php` has shrunk a decent bit in `ubuntu-20.04` (which `ubuntu-latest` now points to) and will cause a handful of test failures for `7.2`. `7.4` has the same extensions on both runtimes.

PR #6214, opened 2021-02-18
[Bugfix:CourseMaterials] Handle failed course materials downloads

emaicus:
__Urgency: 3__
__Impact: 2__

In cases where a course material download was cancelled, the resultant temporary zip file was left abandoned in the `/tmp` directory. This PR adds the deletion of the zip file as a callback/shutdown function, preventing the issue.

### DIFF
--- a/site/app/controllers/course/CourseMaterialsController.php
+++ b/site/app/controllers/course/CourseMaterialsController.php
@@ -108,6 +108,9 @@ class CourseMaterialsController extends AbstractController {
         $zip_name = $temp_dir . "/" . $temp_name . ".zip";
         // replacing any whitespace with underscore char.
         $zip_file_name = preg_replace('/\s+/', '_', $dir_name) . ".zip";
+        // Always delete the zip file after script execution
+        register_shutdown_function('unlink', $zip_name);
+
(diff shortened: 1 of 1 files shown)


MasterOdin (approved):

MasterOdin:
Probably worth looking into https://github.com/maennchen/ZipStream-PHP for this since can still hit `/tmp` space issues if enough download requests are kicked off concurrently (I think).

shailpatels:
I haven't taken a close look yet but is this issue only with course materials? I remember the misc controller being mentioned in the last meeting or is it ok because its using `ZipStream` like @MasterOdin mentioned?

emaicus:
@shailpatels, that is correct, the misc controller is using `ZipStream`, which doesn't have the issue. I agree with @MasterOdin that we should look at using it here as well, but hopefully this serves as a good patch in the meantime.

shailpatels (approved):

hdrodz97 (approved):

Managed to reproduce the bug on master by dumping a 100MB file of data piped from `/dev/urandom` into a subfolder of `course_materials`, initiating the download of the folder, and exiting out of the browser as the download prepared. Whereas `master` doesn't clean up after itself, this PR does.

Issue #6209, opened 2021-02-16
Grade inquiry can leak other student's inquiries.


ghost:
Hi Prof. Cutler!

Bug: The grade inquiry feature can potentially show other student's inquiries and the TA responses.

How to reproduce:

1. Keep the gradable page open when inquiries are open.
2. When other student open inquiries, they will appear on the page as if they were your own (showing the other student's name).
3. TA responses will show up underneath as they answer.

I'm in `CSCI-2500` this semester and went to use the grade inquiry feature for an exam, when I suddenly saw my classmate's grade inquiry for the exam. I have kept the inquiry page open for about 15 minutes and am seeing new inquiries as they roll in. Speaking without any experience with submitty internals, I'm guessing the websocket that sends chat responses sends it to everyone with the page open (instead of the intended student only). This is a breach of privacy for obvious reasons.

Here is a screenshot of the bug: The first box is another student's grade inquiry. Below that is the TA's response to a (possibly) different inquiry.

![gradeinquiry](https://user-images.githubusercontent.com/70179370/108121009-1ff91e00-7070-11eb-83ed-8cb61ba7b7ed.jpg)

Thanks for all your hard work!

williamjallen:
I can confirm that I have seen the same behavior this semester a few times as well as once in Data Structures last semester.

Xystem4:
To add some more hopefully relevant information, I've had this happen to me multiple times in the past where I: opened an exam page once the grade was returned (but grade inquiries were not yet open), and kept the page open until inquiries were open (yes I've kept chrome tabs open for weeks before). Upon refreshing the page after inquiries were opened, I was able to see other students' inquiries (and responses from graders). At this point leaving entering the exam page from another tab would not be showing other students' inquiries.

I'm not too knowledgeable on the actual infrastructure behind any of this or if those steps are actually relevant in reproducing the issue, but the three times I've seen other students' inquiries it's always been after doing that.

williamschen23:
should have been closed by #6215

PR #6215, opened 2021-02-19
[Bugfix:TAGrading] add grade inquiry checks

shailpatels:

For some reason, WebSockets is leaking information by broadcasting new posts to unrelated students. Websockets should guarantee a unique page URL per student grade inquiry but if it doesn't a post id is sent to the client to render which doesn't have any server side checks to make sure the client can view that post.

### What is the new behavior?
Adds additional checks to make sure the current user has permissions to view the post requested through websockets. Modified the db query that gets a single post so now the `regrade_id` foreign_key needs to be given as well.

### Other information?
closes # #6209


site/app/controllers/student/GradeInquiryController.php
```diff
@@ -112,7 +112,11 @@ public function makeGradeInquiryPost($gradeable_id) {
 
         try {
             $regrade_post_id = $this->core->getQueries()->insertNewRegradePost($grade_inquiry_id, $user->getId(), $content, $gc_id);
-            $regrade_post = $this->core->getQueries()->getRegradePost($regrade_post_id);
+            $regrade_post = $this->core->getQueries()->getRegradePost($regrade_post_id, $grade_inquiry_id);
+            if ($regrade_post === null) {
```
MasterOdin, line 116: How does `$regrade_post` be null here? It seems to me that `getRegradePost($regrade_post_id, $grade_inquiry_id)` is just going to get always fetch the post that was just inserted right above as part of `insertNewRegradePost`?

site/public/js/grade-inquiry.js
```diff
@@ -74,6 +74,7 @@ function onGradeInquirySubmitClicked(button) {
 
   // prevent double submission
   form.data("submitted",true);
+  let gc_id = form.children('#gc_id').val();
```
MasterOdin, line 77:
```suggestion
  const gc_id = form.children('#gc_id').val();
```

site/app/controllers/student/GradeInquiryController.php
```diff
@@ -137,28 +141,55 @@ public function makeGradeInquiryPost($gradeable_id) {
...
+
+        $grade_inquiry = $graded_gradeable->getGradeInquiryByGcId($gc_id);
+        if (is_null($grade_inquiry)) {
+            return "";
+        }
+
+        $new_post = $this->core->getQueries()->getRegradePost($post_id, $grade_inquiry->getId());
+        if ($new_post === null) {
+            return "";
+        }
```
MasterOdin, line 186: It feels like we could do:

```suggestion
        $new_post = $this->core->getQueries()->getRegradePost($post_id);
        if ($new_post === null || $new_post['regrade_id'] !== $grade_inquiry->getId()) {
            return ""
        }
```

with no loss of functionality, while also not having to modify the signature of `getRegradePost`.

PR #6219, opened 2021-02-19
[Testing:Developer] Run github actions against PRs

shailpatels:

Forks do not run any of the tests on github actions

### What is the new behavior?
On PR's, forks should be testing against github actions

### Other information?
This is a branch off of the #6204 and should be merged after e2e tests have been merged


MasterOdin:
Do not use `pull_request_target` to run and test code. Just use `pull_request`. See the big warning on the [event page about it](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target):

> Warning: The pull_request_target event is granted a read/write repository token and can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch, and to help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered.

shailpatels:
> Do not use `pull_request_target` to run and test code. Just use `pull_request`. See the big warning on the [event page about it](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target):
> 
> > Warning: The pull_request_target event is granted a read/write repository token and can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch, and to help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered.

Huh I didn't see that page, I was actually reading this post from Github's blog:
https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/
, and my interpretation was that `pull_request_target` was supposed to be a _safer_ option for forks 


> In order to solve this, we’ve added a new pull_request_target event, which behaves in an almost identical way to the pull_request event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request.>

But I guess there are still risks with it, I'm a bit confused between the two doc pages, they seem like they're contradicting each other but I'll switch it over to a normal `pull_request` just to make sure

MasterOdin:
They do not contradict each other, it's just not super well written. Both the blog post and the doc page make mention the entire focus of this type is for making PR labels and comments. The "safety" / "trusted source" just refers to the fact that the workflow is not triggered from the merged commit that `pull_request` uses, rather the HEAD commit of the target branch, but then operates within the scope of the PR (so doing `actions/checkout` gets you the PR code). As such, it's safe to have a workflow that utilizes a write token and secrets (to do stuff like labels and comments) because an attacker cannot maliciously alter a workflow within their PR, rather it has to land into the target repository. However, because of that, it's not aimed at actually using it to run the PR code.

PR #6228, opened 2021-02-22
[Bugfix:TAGrading] Fix file opening bugs

### LINKED ISSUE #6226: Submitted file access errors for ta grading
**Describe the bug**
Instructor user is getting denied access to a png file with a '+' in the name.  Cannot view inline, or as popout, or do individual file download.  Can download the zip of all files successfully.

And if it has an '&' in the name, I get a "something went wrong" error.

DefinitiveAbove:

Cannot open files with a plus in the filename in TA grading and course materials.
Cannot view PDFs with an ampersand in the filename in the embedded viewer.

### What is the new behavior?
Fixes #6226 
Can open and download files with a plus in the filename in TA grading and course materials.
Can view PDFs with an ampersand in the filename in the embedded viewer.


shailpatels (changes requested):
Looks good tested files with `+`s `&` and combinations of other problematic chars and spaces and no file names (`.txt`s ). The PDF editor works again with strange chars as well now.

The download button still has the same issue when a file contains a `+` you get a permissions error instead of a prompt to download

shailpatels (approved):

MasterOdin:
Was this tested with the `"` and `'` characters? Or is it not possible to get a filename with those in the normal course of things?

shailpatels:
> Was this tested with the `"` and `'` characters? Or is it not possible to get a filename with those in the normal course of things?

I ssh'd into the VM and manually created a file with quotes in the submissions directory and it looks like double quotes break viewing and downloading files. Probably need to escape them to prevent a similar issue like in #6241

Single quotes work fine

We do prevent uploading with quotes in places with drag+drop upload so technically files with quotes should never hit
the ta grading interface or course materials
https://github.com/Submitty/Submitty/blob/ded95a29246e6c63931506b89a88c72ac9df297a/site/public/js/drag-and-drop.js#L914-L915

@DefinitiveAbove  the fix would probably be adding the `addslashes` function around the twig file that generates the JS function calls to open, expand, and download files. If it takes more work than that I wouldn't say it's worth it to support quotes in the offhand chance they somehow get submitted.

PR #6256, opened 2021-03-03
[Feature:TAGrading] limited grader access before due date

### LINKED ISSUE #6230: Access to student's submitted work prior to grading begins date
Currently limited access graders cannot view students submitted files until the grading opens date.  This was the original design, but has been an unnecessary burden.  Let's change this!  Limited access graders should be able to view their assigned students work as soon  as its available.  This will allow them to help during office hours if students have submitted preliminary work and want a one-on-one review.  It will also allow them to familiarize themselves with the problems & rubric before the submission deadline.

Open Question: We might want to restrict the start of grading (selecting marks) but I don't think this is actually a problem.  If a full/limited access grader incorrectly marks a student's early (not final) work before the deadline, we will get a version conflict that will need attention to be resolved.

DefinitiveAbove:

Limited graders can only access a gradeable flagged for limited access graders and above after grading opens.

### What is the new behavior?
Resolves #6230
Limited access graders can now access a gradeable flagged for limited access graders before grading opens

### DIFF
--- a/site/app/templates/grading/electronic/ta_status/StatusBase.twig
+++ b/site/app/templates/grading/electronic/ta_status/StatusBase.twig
@@ -208,21 +208,6 @@
                     <li>Grade Inquiries</li>
                 {% endif %}
             </ul>
-            {# {% if core.getUser().getGradingRegistrationSections()|length != 0 %}
-                {% if overall_average == null or overall_average.getCount() == 0 %}
-                    {% if team_assignment %}
-                        <a class="btn btn-primary"
-                           href="{{ grade_url }}" style="float: right;">
-                            My Next Ungraded Team
-                        </a>
-                    {% else %}
-                        <a class="btn btn-primary"
-                           href="{{ grade_url }}" style="float: right;">
-                            My Next Ungraded Student
-                        </a>
-                    {% endif %}
-                {% endif %}
-            {% endif %} #}
         {% endif %}
     </div>
     {# /Top Button #}
--- a/site/app/templates/grading/electronic/ta_status/StatusBase.twig
+++ b/site/app/templates/grading/electronic/ta_status/StatusBase.twig
@@ -208,7 +208,7 @@
                     <li>Grade Inquiries</li>
                 {% endif %}
             </ul>
-            {% if core.getUser().getGradingRegistrationSections()|length != 0 %}
+            {# {% if core.getUser().getGradingRegistrationSections()|length != 0 %}
                 {% if overall_average == null or overall_average.getCount() == 0 %}
                     {% if team_assignment %}
                         <a class="btn btn-primary"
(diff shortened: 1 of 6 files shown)


site/app/views/NavigationView.php
```diff
@@ -299,7 +299,7 @@ private function hasSubmitButton(Gradeable $gradeable): bool {
      */
     private function hasGradeButton(Gradeable $gradeable): bool {
         $now = $this->core->getDateTimeNow();
-        $date_limitation = $gradeable->getGradeStartDate() <= $now || $this->core->getUser()->getGroup() === User::GROUP_INSTRUCTOR || $this->core->getUser()->getGroup() === User::GROUP_FULL_ACCESS_GRADER;
+        $date_limitation = $gradeable->getGradeStartDate() <= $now || $this->core->getUser()->getGroup() === User::GROUP_INSTRUCTOR || $this->core->getUser()->getGroup() === User::GROUP_FULL_ACCESS_GRADER || $this->core->getUser()->getGroup() === User::GROUP_LIMITED_ACCESS_GRADER;
```
MasterOdin, line 302: Given its all numeric values under the hood, can simplify to a simpler less than or equal to:

```suggestion
        $date_limitation = $gradeable->getGradeStartDate() <= $now || $this->core->getUser()->getGroup() <= User::GROUP_LIMITED_ACCESS_GRADER;
```

site/app/templates/grading/electronic/ta_status/StatusBase.twig
```diff
@@ -208,7 +208,7 @@
                     <li>Grade Inquiries</li>
                 {% endif %}
             </ul>
-            {% if core.getUser().getGradingRegistrationSections()|length != 0 %}
+            {# {% if core.getUser().getGradingRegistrationSections()|length != 0 %}
```
MasterOdin, line 211: why is this commented out?

MasterOdin (changes requested):

site/app/templates/grading/electronic/ta_status/StatusBase.twig
DefinitiveAbove: When the button is clicked, it goes to a frog robot page - the button only appears when there are no students that have been graded, and I think it was designed for the old TA grading page (where you could call it with no parameters and it would give a ungraded student, although even without the frog robot the appear logic looks wrong too)
We talked about it a bit during the meeting and for now decided that we don't want the button (so maybe it should be deleted instead of commented out)

MasterOdin: Yeah, the main thing here for commenting out code is to give context for why it's commented out, when it might be uncommented out, link to a GH issue maybe, etc. Assume the student next semester that looks at this is not going to have the benefit of having been at the meeting that this was discussed.

However, if it's broken / dead code we no longer want / need, better to remove it altogether! 😄

MasterOdin (approved):

jeeva2812 (approved):
LGTM.

PR #6263, opened 2021-03-06
[Bugfix:Autograding] Make docker api timeouts explicit

emaicus:
This PR makes docker api timeouts explicit and adds  handling for a broader class of container creation errors.

### DIFF
--- a/autograder/autograder/execution_environments/container_network.py
+++ b/autograder/autograder/execution_environments/container_network.py
@@ -116,6 +116,7 @@ class Container():
             raise
         except Exception:
             self.log_function(f'ERROR: could not create container {self.full_name}')
+            client.close()
             raise
 
         self.log_meta(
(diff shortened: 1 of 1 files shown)


autograder/autograder/execution_environments/container_network.py
```diff
@@ -114,6 +114,10 @@ def create(self, execution_script, arguments, more_than_one):
             self.log_function(f'ERROR: The image {self.image} is not available on this worker')
             client.close()
             raise
+        except Exception:
```
MasterOdin, line 117: Can we catch more specific exceptions here for why the container failed to start?

hdrodz97: I had similar concerns, but the full exception stack trace is logged further up the chain.

hdrodz97 (approved):
Looks good to me!

autograder/autograder/execution_environments/container_network.py
bmcutler: Yes, we are intentionally adding a broad exception catch.  We experienced an unexpected network timeout exception on production and we didn't catch that error at a good place.  We don't think a specific exception catch here warranted right now, since we'd ALSO want to add the catch all exception.

PR #6264, opened 2021-03-06
[Testing:System] Add Vagrant & sample capabilities

hdrodz97:

When creating a Vagrant VM, all sample courses use `default` as their autograding capability and the default autograding worker only recognizes the `default` capability.

### What is the new behavior?

This PR adds capabilities to several gradeables in `more_autograding_examples` and configures the default autograding worker to recognize more capabilities when running under Vagrant.

### Other information?

This sets us up for testing of more sophisticated autograding workloads.

***The `autograding_workers.json` file is NOT updated or regenerated.***

The Vagrant sample courses should continue to function as long as they aren't rebuilt. Once they're rebuilt, many submissions will never get graded due to the `autograding_workers.json` file missing the required autograding capability. To fix this, you can either:

1. Modify `/usr/local/submitty/autograding_workers.json` so that `primary.capabilities` includes the following capabilities: `cpp`, `python`, `matlab`, `pdf-word-count`, `et-cetera`, `notebook`. Don't forget the double-quotes around each value! Run `/usr/local/submitty/.setup/INSTALL_SUBMITTY.sh` for the changes to take effect.
2. Run `vagrant destroy` followed by `vagrant up`. This is easier, but it takes much longer. Once it's up, just run `/usr/local/submitty/bin/grading_done.py` and make sure nothing's getting stuck.


jchulton:
Tested vagrant destroy vagrant up. Seemed to work fine

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -460,6 +463,16 @@ def write(x=''):
             }
         }
 
+        if args.setup_for_sample_courses:
+            worker_dict['primary']['capabilities'].extend([
+                'cpp',
+                'python',
+                'matlab',
+                'pdf-word-count',
+                'et-cetera',
```
MasterOdin, line 472: These two are not used. Can they be removed?

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -460,6 +463,16 @@ def write(x=''):
             }
         }
 
+        if args.setup_for_sample_courses:
+            worker_dict['primary']['capabilities'].extend([
+                'cpp',
+                'python',
+                'matlab',
+                'pdf-word-count',
```
MasterOdin, line 471: This does not look like it's used.

.setup/CONFIGURE_SUBMITTY.py
MasterOdin, line 472: This label looks like it's applied to gradeables that just use built-in Submitty capabilities? Do we need this as an example? Or could it be made more explicit as to why an sysadmin might want this as a capability tag?

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -460,6 +463,16 @@ def write(x=''):
             }
         }
 
+        if args.setup_for_sample_courses:
+            worker_dict['primary']['capabilities'].extend([
+                'cpp',
+                'python',
+                'matlab',
```
MasterOdin, line 470: This should be removed from the set of capabilities here. By default, the sample courses does not include matlab example as it requires [extra steps](https://submitty.org/developer/configuring_tie_in_programs#matlab) to enable. We should maybe just add to that doc page adding this tag after enabling matlab.

MasterOdin, line 470: This should be removed from the set of capabilities here. By default, the sample courses does not include matlab example as it requires [extra steps](https://submitty.org/developer/configuring_tie_in_programs#matlab) to enable. We should maybe just add to that doc page adding this tag after enabling matlab.

.setup/CONFIGURE_SUBMITTY.py
MasterOdin, line 471: this does not look like it's used?

.setup/CONFIGURE_SUBMITTY.py
MasterOdin, line 472: This label looks like it's applied to gradeables that just use built-in Submitty capabilities? Do we need this as an example? Or could it be made more explicit as to why an sysadmin might want this as a capability tag?

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -47,6 +47,9 @@ def get_input(question, default=""):
                                  formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 parser.add_argument('--debug', action='store_true', default=False, help='Configure Submitty to be in debug mode. '
                                                                         'This should not be used in production!')
+parser.add_argument('--setup-for-sample-courses', action='store_true', default=False,
```
MasterOdin, line 50: Given the changes to capability set, I would maybe just reterm this to `--setup-default-capabilities` or something as the list is really just stuff that comes with Submitty by default to even function.

.setup/CONFIGURE_SUBMITTY.py
bmcutler: agree...  only have the tags we use in the sample configs

.setup/CONFIGURE_SUBMITTY.py
bmcutler: its not default (don't want this on production)

MasterOdin: We may want to consider having this not be in `.setup/CONFIGURE_SUBMITTY.py` then as there's a chance someone will of course use this for their production installation.

But will not block this PR on that discussion.

MasterOdin (approved):

PR #6265, opened 2021-03-06
[Bugfix:Autograding] Properly clean up stale containers

emaicus:
Moves the `cleanup_stale_containers` function to use the `docker` package for python, hopefully making it less flaky.


hdrodz97 (approved):
Looks good, works as expected. Made two minor changes: one addressing linter complaints and another addressing a potential client leak situation.

PR #6274, opened 2021-03-10
[Testing:Developer] Make cypress login tests less flaky

shailpatels:

Cypress is supposed to clear each session between tests however sometimes a test fails because the user is still logged in between tests.

### What is the new behavior?
Group tests by whether a user is supposed to successfully login or not and now explicitly log out of Submitty after each test that can log in.

### DIFF
--- a/site/cypress/integration/login.spec.js
+++ b/site/cypress/integration/login.spec.js
@@ -2,93 +2,100 @@ import {buildUrl} from '../support/utils.js';
 
 
 describe('Test cases revolving around the logging in functionality of the site', () => {
-	it('should log in through root endpoint', () => {
-		const user = 'instructor';
-
-		//should hit the login form
-		cy.visit([]);
-		cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );
-
-		//fill out the login form and hit submit
-		cy.get('input[name=user_id]').type(user);
-		cy.get('input[name=password]').type(user);
-		cy.get('input[name=login]').click();
-
-		//should now be logged in as instructor and have a loggout button
-		cy.get('#logout .icon-title').should((val) => {
-			expect( val.text().trim() ).to.equal('Logout Quinn');
+	describe('Test cases where the user should succesfully login', () => {
+		afterEach(() => {
+			cy.logout();
 		});
 
-		cy.getCookies().then((cookies) => {
-			expect(cookies[2]).to.have.property('name', 'submitty_token');
-			expect(cookies[2]['value']).to.match(/^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/);
+		it('should log in through root endpoint', () => {
+			const user = 'instructor';
+
+			//should hit the login form
+			cy.visit('/');
+			cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );
+
+			//fill out the login form and hit submit
+			cy.get('input[name=user_id]').type(user);
+			cy.get('input[name=password]').type(user);
+			cy.get('input[name=login]').click();
+
+			//should now be logged in as instructor and have a loggout button
+			cy.get('#logout .icon-title').should((val) => {
+				expect( val.text().trim() ).to.equal('Logout Quinn');
+			});
+
+			cy.getCookies().then((cookies) => {
+				expect(cookies[2]).to.have.property('name', 'submitty_token');
+				expect(cookies[2]['value']).to.match(/^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/);
+			});
 		});
-	});
 
 
-	it('should login through login endpoint', () => {
-		const user = 'instructor';
+		it('should login through login endpoint', () => {
+			const user = 'instructor';
 
-		cy.visit('authentication/login');
-		cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );
-		cy.get('input[name=user_id]').type(user);
-		cy.get('input[name=password]').type(user);
-		cy.get('input[name=login]').click();
+			cy.visit('authentication/login');
+			cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );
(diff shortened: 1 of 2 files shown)


site/cypress/integration/login.spec.js
```diff
@@ -2,93 +2,100 @@ import {buildUrl} from '../support/utils.js';
...
+			cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );
 
-		//should now be logged in as instructor and have a loggout button
-		cy.get('#logout .icon-title').should((val) => {
-			expect( val.text().trim() ).to.equal('Logout Quinn');
-		});
+			//fill out the login form and hit submit
+			cy.get('input[name=user_id]').type(user);
+			cy.get('input[name=password]').type(user);
+			cy.get('input[name=login]').click();
```
MasterOdin, line 20: use `cy.login`?

site/cypress/integration/login.spec.js
```diff
@@ -2,93 +2,100 @@ import {buildUrl} from '../support/utils.js';
...
+	describe('Test cases where the user should succesfully login', () => {
+		afterEach(() => {
+			cy.logout();
+		});
 
-		//should hit the login form
-		cy.visit([]);
-		cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );
+		it('should log in through root endpoint', () => {
+			const user = 'instructor';
```
MasterOdin, line 11: I would move this to top-level of describe block as this is the same value in each test.

site/cypress/integration/login.spec.js
```diff
@@ -2,93 +2,100 @@ import {buildUrl} from '../support/utils.js';
...
-		cy.visit('authentication/login');
-		cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );
-		cy.get('input[name=user_id]').type(user);
-		cy.get('input[name=password]').type(user);
-		cy.get('input[name=login]').click();
+			cy.visit('authentication/login');
+			cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );
+			cy.get('input[name=user_id]').type(user);
+			cy.get('input[name=password]').type(user);
+			cy.get('input[name=login]').click();
```
MasterOdin, line 41: use `cy.login`?

site/cypress/integration/login.spec.js
```diff
@@ -2,93 +2,100 @@ import {buildUrl} from '../support/utils.js';
...
+			cy.visit(['sample', 'gradeable', 'open_homework']);
+			cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login?old=${encodeURIComponent(full_url)}`);
 
-		cy.get('#error-0').should((val) => {
-			expect( val.text().trim() ).to.equal('Could not login using that user id or password');
-		});
-	});
+			cy.get('input[name=user_id]').type(user);
+			cy.get('input[name=password]').type(user);
+			cy.get('input[name=login]').click();
```
MasterOdin, line 59: use `cy.login`?

MasterOdin (approved):
Slight changes requested, otherwise looks good and should hopefully help make things less flakey.

shailpatels:
@MasterOdin I moved both the inner `describe` blocks to the top level, but not sure if that's what you meant when you said 
> I would move this to top-level of describe block as this is the same value in each test.

MasterOdin:
Sorry that was vague, what I meant by that comment was to move the `const user = 'instructor';` as a construct within the describe block and to fall then within the scope of the tests, so that you only have this line once (instead of 3).

So something like:
```js
describe('Test cases revolving around the logging in functionality of the site', () => {
  describe('Test cases where the user should succesfully login', () => {
    const user = 'instructor';
    
    afterEach(() => {
      cy.logout();
    });

    it('should log in through root endpoint', () => {
      //should hit the login form
      cy.visit('/');
      cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );

      cy.login(user);

      //should now be logged in as instructor and have a loggout button
      cy.get('#logout .icon-title').should((val) => {
        expect( val.text().trim() ).to.equal('Logout Quinn');
      });

      cy.getCookies().then((cookies) => {
        expect(cookies[2]).to.have.property('name', 'submitty_token');
        expect(cookies[2]['value']).to.match(/^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/);
      });
    });

    it('should login through login endpoint', () => {
      cy.visit('authentication/login');
      cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login` );
      cy.login(user);

      cy.get('#logout .icon-title').should((val) => {
        expect( val.text().trim() ).to.equal('Logout Quinn');
      });
    });

    it('should redirect after logging in', () => {
      //try to visit a page not logged in, then log in and see where we are
      const full_url = buildUrl(['sample', 'gradeable', 'open_homework'], true);

      cy.visit(['sample', 'gradeable', 'open_homework']);
      cy.url().should('eq', `${Cypress.config('baseUrl')}/authentication/login?old=${encodeURIComponent(full_url)}`);

      cy.login(user);

      cy.url().should('eq', full_url);
    });


    it('should check if you can access a course', () => {
      cy.login('pearsr');
      cy.visit(['sample']);

      cy.get('.content').should('have.text', "\n    You don't have access to this course. \n    This is sample for Spring 2021. \n    If you think this is a mistake, please contact your instructor to gain access. \n    click  here  to back to homepage and see your courses list.\n");
    });
  });

  describe('Test cases where the user should not be able to login', () => {
    it('should reject bad passwords', () => {
      cy.visit([]);

      cy.get('input[name=user_id]').type('instructor');
      cy.get('input[name=password]').type('bad-password');
      cy.get('input[name=login]').click();

      cy.get('#error-0').should((val) => {
        expect( val.text().trim() ).to.equal('Could not login using that user id or password');
      });
    });


    it('should reject bad usernames', () => {
      cy.visit([]);

      cy.get('input[name=user_id]').type('bad-username');
      cy.get('input[name=password]').type('instructor');
      cy.get('input[name=login]').click();

      cy.get('#error-0').should((val) => {
        expect( val.text().trim() ).to.equal('Could not login using that user id or password');
      });
    });
  });
});
```

we keep the same structure, but just define `user` in one place.

MasterOdin (approved):

Issue #6210, opened 2021-02-17
Submitty Database Migration Error:  Database does not exist for master


erelsgl:
**Describe the bug**
During installation, after I answer all configuration questions, I get the following error:

`
Beginning Install Submitty Script
bash ${SUBMITTY_INSTALL_DIR}/.setup/INSTALL_SUBMITTY.sh clean skip_web_restart
CLONE OR UPDATE AnalysisTools...
    current version matches or exceeds minimum version v.18.06.00
CLONE OR UPDATE Lichen...
    current version matches or exceeds minimum version v20.05.00
CLONE OR UPDATE RainbowGrades...
    current version matches or exceeds minimum version v21.01.01
CLONE OR UPDATE Tutorial...
    current version matches or exceeds minimum version v20.10.01
CLONE OR UPDATE SysadminTools...
    current version matches or exceeds minimum version v20.01.00
Checking for system and database migrations
Submitty Database Migration Error:  Database does not exist for master
err_message
`

**Expected behavior**
I expected the install to complete successfully.

**To Reproduce**
1. I just run bash ./.setup/install_system.sh and answer the questions with mostly the default values.
2. See error

**Configuration**
 - OS: Ubuntu
 - Browser: irrelevant

erelsgl:
Note: `systemctl status postgresql` shows:

```
● postgresql.service - PostgreSQL RDBMS
   Loaded: loaded (/lib/systemd/system/postgresql.service; enabled; vendor preset: enabled)
   Active: active (exited) since Wed 2021-02-17 21:23:09 UTC; 2min 52s ago
  Process: 17225 ExecStart=/bin/true (code=exited, status=0/SUCCESS)
 Main PID: 17225 (code=exited, status=0/SUCCESS)

Feb 17 21:23:09 badkan2.com systemd[1]: Starting PostgreSQL RDBMS...
Feb 17 21:23:09 badkan2.com systemd[1]: Started PostgreSQL RDBMS.
```

MasterOdin:
For the postgres server, what databases exist in it?

erelsgl:
```
postgres=# \l
                                 List of databases
   Name    |      Owner      | Encoding | Collate |  Ctype  |   Access privileges
-----------+-----------------+----------+---------+---------+-----------------------
 postgres  | postgres        | UTF8     | C.UTF-8 | C.UTF-8 |
 submitty  | submitty_dbuser | UTF8     | C.UTF-8 | C.UTF-8 |
[...]

PR #6275, opened 2021-03-10
[Feature:System] Improve migrator DB error output

MasterOdin:

Related to #6210.

Right now, if we have issues connecting to the master database, we throw a generic looking exception message. This makes it harder to debug as it's unclear if it couldn't find the database, user does not have access, etc.

### What is the new behavior?

The exception message from sqlachemy is printed out to the user. This will contain appropriate error message such as `password authentication failed for "submitty_dbuser"`and the like to make it easier to debug.

### DIFF
--- a/migration/migrator/main.py
+++ b/migration/migrator/main.py
@@ -81,9 +81,12 @@ def status(args):
                     continue
                 print_status(database, environment, loop_args)
                 database.close()
-            except OperationalError:
+            except OperationalError as exc:
                 print(
-                    'Could not get database for migrations for {}'.format(environment)
+                    'Could not get database for migrations for {}:\n  {}'.format(
+                        environment,
+                        str(exc).split("\n")[0]
+                    )
                 )
         else:
             course_dir = Path(args.config.submitty['submitty_data_dir'], 'courses')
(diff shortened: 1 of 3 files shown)


migration/migrator/main.py
```diff
@@ -81,9 +81,12 @@ def status(args):
                     continue
                 print_status(database, environment, loop_args)
                 database.close()
-            except OperationalError:
+            except OperationalError as exc:
                 print(
-                    'Could not get database for migrations for {}'.format(environment)
+                    'Could not get database for migrations for {}:\n  {}'.format(
+                        environment,
+                        str(exc).split("\n")[0]
```
shailpatels, line 88: Is there a reason only the first line of the stack trace is printed? Wouldn't it be more helpful for debugging to show everything?

MasterOdin: sqlachemy produces the following string for the exception:
> (psycopg2.OperationalError) FATAL:  password authentication failed for user "submitty_dbuser1"
>
> (Background on this error at: http://sqlalche.me/e/13/e3q8)

I want to hide the final line as the link is not useful (and perhaps confusing) to a non-developer familiar with sqlachemy. I do want the first line though as it gives the specific error (the `(psycopg2.OperationalError)` bit do not necessarily care for, but not strongly enough to strip a strip method for it.

PR #6283, opened 2021-03-12
[Refactor:System] Remove deprecated emma code

MasterOdin:

* [x] Documentation has been updated/added if relevant: Submitty/submitty.github.io#325

emma last saw and update 15 years ago, and we had deprecated usage of it years ago because it was not compatible with newer versions of Java (Java 7+? Java 8+?). We have long since relied on JaCoCo for what emma was providing us, and so it's unlikely any instructor at this point is utilizing the emma code which is also totally untested by us.

### What is the new behavior?

Removes the emma code from the codebase. It's untested, unused, and unmaintained and no point in keeping it around.


bin/json_schemas/complete_config_schema.json
```diff
@@ -1454,11 +1430,25 @@
...
+                  "searchToken",
+                  "intComparison",
+                  "diff",
+                  "fileExists",
+                  "warnIfNotEmpty",
+                  "warnIfEmpty",
+                  "errorIfNotEmpty",
+                  "errorIfEmpty",
+                  "ImageDiff",
+                  "custom_validator"
```
MasterOdin, line 1449: Beyond the formatting change, I removed `EmmaInstrumentationGrader` and `EmmaCoverageReportGrader` from the list of allowed enums here.

PR #6293, opened 2021-03-14
[Testing:Developer] eslint cypress files

### LINKED ISSUE #6292: Setup eslint to run on cypress test suite files
eslint currently only runs against files under `src/js/`. This should be extended to also run for files under `cypress/`. We will probably want to stick a `.eslintrc.json` file in the cypress folder to add additional `env` variables (so that `describe` and `it` are allowed, etc.).

shailpatels:

* [X] Tests for the changes have been added/updated (if possible) Yes
* [X] Documentation has been updated/added if relevant N/A

eslint only runs against the JS files under site/public

### What is the new behavior?
eslint will now run against cypress files, automatically and can be manually run with the commands
```
npm run-script lint
npm run-script lint:fix
```
the following npm dev dependency is installed to make eslint play nice with cypress global definitions and specific syntax
[eslint-plugin-cypress](https://github.com/cypress-io/eslint-plugin-cypress)

### Other information?
closes #6292


shailpatels:
@MasterOdin I copied the `.eslintrc` file into the cypress directory but now all of the rules are duplicated between each rc file, is this the correct way to have multiple .eslintrc files in a project?

MasterOdin:
Eslint supports a concept of [cascading](https://eslint.org/docs/user-guide/configuring/configuration-files#cascading-and-hierarchy) where it will merge configs as it goes up the directory tree.

As such, you would be able to do:
```json
{
    "env": {
        "cypress/globals": true,
        "node": true,
        "browser": null,
        "jquery": null,
    },
    "parserOptions": {
        "sourceType": "module"
    },
    "plugins" : [
        "cypress"
    ],
}
```

in `site/cypress/.eslintrc.json` and this will merge (using Object.assign essentially for each top-level key) with `site/.eslintrc.json` to give us a final config that would then include our rules, setting the extends, etc. One thing to note is we set certain stuff in `env` to null within the cypress config so that it does not apply to it.

MasterOdin (approved):

PR #6296, opened 2021-03-16
[Bugfix:System] Pin sqlalchemy version

MasterOdin:

* [x] Documentation has been updated/added if relevant: https://github.com/Submitty/submitty.github.io/pull/326

Sqlalchemy changed behavior around how the `engine.has_table` function worked. This in-turn broke how the migrator was functioning.

### What is the new behavior?

Sets an upper-cap on the version of sqlalchemy that Submitty will install so that nobody is blocking as investigate how we might move off the `engine.has_table` function in a way that is backwards compatible.

### DIFF
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -139,7 +139,7 @@ jobs:
         - name: Install python libraries
           run : |
             python3 -m pip install --upgrade pip setuptools wheel
-            python3 -m pip install SQLAlchemy jsonschema jsonref pytz tzlocal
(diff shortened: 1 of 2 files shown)


.github/workflows/submitty_ci.yml
```diff
@@ -253,7 +253,6 @@ jobs:
               sudo pip3 install docker
               sudo pip3 install paramiko
               sudo pip3 install tzlocal
-              sudo pip3 install sqlalchemy
```
MasterOdin, line 256: this was a duplicate install line, removed it for the single install line on 270.

PR #6302, opened 2021-03-19
[Testing:Developer] Improve output for PR title action

MasterOdin:
This PR replaces the PR title GH action we were using with something that is a bit more Submitty specific, and maintained under our organization in https://github.com/Submitty/action-pr-title. The purpose of doing this is to make the error messages more friendly and easier to immediately act upon. The old action we used would give error messages that were always something like `Pull Request title "[Dependency:Test] Test" failed to pass match regex - /^(\[SYSADMIN ACTION\])?(\[(Bugfix|Feature|Refactor|Testing|Documentation|VPAT|UI\/UX):(Submission|Autograding|Forum|Notifications|TAGrading|InstructorUI|SubminiPolls|HelpQueue|CourseMaterials|Plagiarism|RainbowGrades|System|Developer|API)\] .{2,40}$|\[(DevDependency|Dependency)\] .{2,70})$/`

What did I fail on? The type? The module? The length?

The new action would return the following: `No allowed module for Dependency, expected [Dependency]. Got [Dependency:Test].`.

Much better! I know what I need to fix, easy peasy. This has error messages for allowed modules, allowed types, and minimum/maximum length for message. Because it's also an action that we maintain, it's easy to extend or modify as we see fit to suit our needs, without also futzing about regex strings in a yml file.


shailpatels (approved):
Looks good, covers all current pr title cases and error messages tell whats exactly wrong

PR #6308, opened 2021-03-22
[Feature:Calendar] A working (beta) calendar

zhangz22:

* [✅] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [✅] Tests for the changes have been added/updated (if possible)
* [✅] Documentation has been updated/added if relevant

### What is the new behavior?
1. A new item "Calendar" is added under "My profile" which will be a basic cross-course platform-wide component of Submitty
2. Reads the datebase each time the page is load. Reads from every course the user is in. It will not read database again unless user refreshes page manually
3. Gradeables listed on a montly calendar, with clickable url directly to submission page
4. Gradeable item color/style consistent with course page. Compatible with dark mode
5. Support switching month without reloading page

### Other information?
Calendar code can be reused by simply replacing the content in the passed-in JSON format data. Data format are described in comment.


site/public/css/calendar.css
```diff
@@ -0,0 +1,273 @@
...
+
+#next-month-switch
+{
+    text-align: left;
+}
+
+.cal-month-title
+{
+    font-size: 26px;
+    font-family: 'Source Sans Pro', sans-serif;
```
shailpatels, line 67: Why is the font family being set for this element here, toggling it on the inspector for mac firefox and chrome, it doesn't appear to do anything so I think it can be removed

site/public/css/calendar.css
```diff
@@ -0,0 +1,273 @@
...
+    font-style: normal;
+    border-radius: 50%;
+    padding-left: 3px;
+    padding-right: 3px;
+}
+
+/* the day number in each cell if the day belongs to prev/next month of the month the calendar is viewing */
+.cal-next-month-date
+{
+    color:#bfbfbf;
```
shailpatels, line 146: use `--standard-light-gray` from Submitty's colors.css instead

site/public/css/calendar.css
```diff
@@ -0,0 +1,273 @@
...
+.cal-gradeable-status-3:focus
+{
+    background-color: var(--focus-danger-red);
+    border: var(--focus-danger-red);
+}
+
+/* CLOSED  being graded by TA/Instructor */
+.cal-gradeable-status-4
+{
+    background-color: #a07436;
```
shailpatels, line 242: use `--standard-medium-orange` here instead

site/public/css/calendar.css
```diff
@@ -0,0 +1,273 @@
...
+}
+
+.cal-gradeable-status-4:hover
+{
+    background-color: #834609;
```
shailpatels, line 247: use `--warning-brown: #945600;` here

site/public/css/calendar.css
```diff
@@ -0,0 +1,273 @@
...
+    border: #834609;
+}
+
+.cal-gradeable-status-4:focus
+{
+    background-color: #cab99d;
```
shailpatels, line 253: use `--standard-beige` here instead

site/public/js/calendar.js
```diff
@@ -0,0 +1,195 @@
+/**
+ * Gets the previous month of a given month
+ * @param month : int the current month (1 as January and 12 as December)
+ * @param year : int the current year
+ * @returns {number[]} : array<int> {previous_month, year_of_previous_month}
+ */
+// eslint-disable-next-line no-unused-vars
+function prevMonth(month, year) {
```
shailpatels, line 8: you can implement prevMonth like so:
```js
//month is int 0=Jan, 11=Dec, 12=Jan, -1=Dec, ... 
prevMonth(month){
   let tmpDate = new Date();
   tmpDate.setMonth( month - 1 ); // date will handle looping for us
   return tmpDate.getMonth() //return month as int 0-11
}
```

Less math for us to worry about that way

site/public/js/calendar.js
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * Gets the next month of a given month
+ * @param month : int the current month (1 as January and 12 as December)
+ * @param year : int the current year
+ * @returns {number[]} : array<int> {next_month, year_of_next_month}
+ */
+// eslint-disable-next-line no-unused-vars
+function nextMonth(month, year) {
+    month = month + 1;
```
shailpatels, line 25: same thing here
```js
prevMonth(month){
   let tmpDate = new Date();
   tmpDate.setMonth( month + 1 ); // date will handle looping for us
   return tmpDate.getMonth() //return month as int 0-11
}
```

You can also set year and return it if that matters here

site/public/js/calendar.js
```diff
@@ -0,0 +1,195 @@
...
+ * Creates a HTML table cell that contains a date.
+ *
+ * @param year : int the year of the date
+ * @param month : int the month of the date (1 as January and 12 as December)
+ * @param day : int the date of the date (1 - 31)
+ * @param curr_view_month : int the current month that the calendar is viewing
+ * @returns {string} the HTML string containing the cell
+ */
+function displayDayCell(year, month, day, curr_view_month) {
+    const cell_date_str = `${year.toString()}-${month.toString().padStart(2, '0')}-${day.toString().padStart(2, '0')}`;
```
shailpatels, line 43: can do `cell_date_str` = new Date(year, month, day).toString()` here and use additional date format options instead of just toString if you want more pretty printing if this gets displayed to the user

site/public/js/calendar.js
```diff
@@ -0,0 +1,195 @@
...
+        // eslint-disable-next-line no-undef
+        if (day === curr_day && month === curr_month && year === curr_year) {
+            content += `<span class="cal-curr-month-date cal-day-title cal-today-title">${day}</span>`;
+        }
+        else {
+            content += `<span class="cal-curr-month-date cal-day-title">${day}</span>`;
+        }
+    }
+    else {
+        if (month > 12) {
```
shailpatels, line 58: why would month ever be greater than 12? Lets assume the caller of this function will pass a month representation of int from 0-11

site/app/templates/calendar/Calendar.twig
```diff
@@ -0,0 +1,70 @@
...
+                        <div id="{{ info.section_id }}" class="course-section-heading">
+                            <div class="course-section-title">
+                                {{ info.title }}
+                                {% if info.subtitle != "" %}
+                                    &nbsp;&nbsp;<em>{{ info.subtitle }}</em>
+                                {% endif %}
+                            </div>
+                        </div>
+
+                        <table class='table table-striped table-bordered persist-area' style='text-align: left'>
```
shailpatels, line 42: Lets make this page mobile-friendly from the start, the long list of gradeable dates after the calendar should be easy. For now, we'll just make sure the calendar fits on the screen for a mobile device and we can optimize it for useability later, I imagine it will need special functionality for mobile vs tablet.

Add the class `mobile-table` to this table and give it an id, something like ` id='calendar-list'` or whatever you prefer.

site/app/views/calendar/CalendarView.php
```diff
@@ -0,0 +1,129 @@
...
+        }
+        if ($year < 1970 || $year > 2100) {
+            $year = (int) date("Y");
+        }
+
+        $this->core->getOutput()->addInternalCss("navigation.css");
+        $this->core->getOutput()->addInternalCss('calendar.css');
+        $this->core->getOutput()->addInternalJs('calendar.js');
+        $this->core->getOutput()->enableMobileViewport();
+        $this->core->getOutput()->addBreadcrumb("Calendar");
```
shailpatels, line 115: load the following CSS file to support mobile tables:
```php
```

site/public/css/calendar.css
```diff
@@ -0,0 +1,273 @@
...
+{
+    background-color: var(--hover-good-green);
+    border: var(--hover-good-green);
+}
+
+.cal-gradeable-status-5:focus
+{
+    background-color: var(--disabled-good-green);
+    border: var(--disabled-good-green);
+}
```
shailpatels, line 273: throw this media query at the end of this css file to make tables look better on mobile:

```css
@media (max-width: 800px) {
    #calendar-list{
        margin: 0;
    }
    
    #calendar-list td:nth-of-type(1):before { content: "Course"; }
    #calendar-list td:nth-of-type(2):before { content: "Gradeable"; }
    #calendar-list td:nth-of-type(3):before { content: "Submission Closes"; }

    #calendar-list thead{
        display: none;
    }
} 
```

shailpatels (changes requested):

shailpatels:
Consider adding screenshots when making new UI changes:
image

image

shailpatels (changes requested):
WAVE is marking several contrast errors on the calendar 
image

See this page for more information:
https://submitty.org/developer/interface_design_style_guide/web_accessibility#all-visual-elements-need-to-adhere-to-a-minimum-contrast-ratio

I recommend you install the WAVE accessibility checker browser plugin to help debug accessibility errors, the main issues appear to be white fonts against bright backgrounds (white on bright green, gray on white etc) also keep in mind the font size and weight affect contrast as well.

Contrast errors and alerts are higher priority, you can look into the alerts but a lot of the time they're not relevant

shailpatels:
Impressive work so far! 
Besides minor code-style issues I expect this to be ready pretty soon :)

site/app/controllers/calendar/CalendarController.php
```diff
@@ -0,0 +1,77 @@
...
+use app\models\Course;
+use app\models\gradeable\Gradeable;
+use app\models\gradeable\GradeableList;
+use app\models\User;
+use app\views\calendar\CalendarView;
+use Exception;
+use Symfony\Component\Routing\Annotation\Route;
+
+class CalendarController extends AbstractController {
+    private $gradeables = [];
```
shailpatels, line 22: I think storing a list of gradeables and populating it from the db is something a model should do.
I would suggest the controller on `viewCalendar` will construct a new calendar object which will call `loadFromDb` from its constructor. Then you can just do $calendar->gradeable_list in `viewCalendar` when returning stuff to the view.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function viewCalendar($user_id = null): MultiResponse {
+        if ($this->gradeables == []) {
+            $this->loadFromDB($user_id);
+        }
+        $gradeable_list = new GradeableList($this->core, null, $this->gradeables);
+
+        return new MultiResponse(
+            JsonResponse::getSuccessResponse($gradeable_list->getGradeablesBySection()),
+            new WebResponse(CalendarView::class, 'showGradeableCalendar', $gradeable_list)
```
shailpatels, line 74: I think it makes more sense for the web response to call `showCalendar` instead of 
going through `showGradeableCalendar` first since `showCalendar` is whats calling the render twig function.

I would move showGradeableCalendar into a model since it looks like its structuring data a certain way and change the function to `createCalendarSections`.

You can then move that function to the model and pass the sections you need as a parameter from the webresponse:
```php
$calendar = new Calendar();
return new WebResponse(
   CalendarView::class, 'showGradeableCalendar'.
   $calendar->getGradeableList(),
   $calendar->getGradeablesByDate(),
   $calendar->getGradeablesBySections()
);
```

shailpatels (changes requested):

site/app/controllers/calendar/CalendarController.php
```diff
@@ -0,0 +1,77 @@
...
+class CalendarController extends AbstractController {
+    private $gradeables = [];
+
+    /**
+     * Loads from the database of all courses and get all gradeables information. Only load once unless the user
+     * refreshes the page.
+     *
+     * @param null $user_id
+     */
+    private function loadFromDB($user_id = null) {
```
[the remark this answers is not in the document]
zhangz22: Thanks for this suggestion. I'll remove this parameter in the next commit.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -0,0 +1,77 @@
...
+     * accessible through the side bar button in a global scope
+     *
+     * @Route("/calendar")
+     *
+     * @param string|null $user_id
+     * @return MultiResponse
+     * @throws Exception if a Gradeable failed to load from the database
+     * @see GlobalController::prep_user_sidebar
+     */
+    public function viewCalendar($user_id = null): MultiResponse {
```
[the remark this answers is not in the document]
zhangz22: Thanks for this suggestion. I'll remove this parameter in the next commit.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Loads from the database of all courses and get all gradeables information. Only load once unless the user
+     * refreshes the page.
+     *
+     * @param null $user_id
+     */
+    private function loadFromDB($user_id = null) {
+        $user = $this->core->getUser();
+        if (is_null($user_id) || $user->getAccessLevel() !== User::LEVEL_SUPERUSER) {
```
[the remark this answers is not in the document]
zhangz22: Thanks for this suggestion. I'll remove the $user_id parameter in the next commit.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -0,0 +1,77 @@
...
+            try {
+                $this->core->loadCourseConfig($course->getSemester(), $course->getTitle());
+                $this->core->loadCourseDatabase();
+                foreach ($this->core->getQueries()->getGradeableConfigs(null) as $gradeable) {
+                    /** @var Gradeable $gradeable */
+                    $this->gradeables["{$course->getSemester()}||{$course->getTitle()}||{$gradeable->getId()}"] = $gradeable;
+                }
+                $this->core->getCourseDB()->disconnect();
+            }
+            catch (Exception $e) {
```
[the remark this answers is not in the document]
zhangz22: Sure. I'll add the exception handler

site/app/controllers/calendar/CalendarController.php
```diff
@@ -0,0 +1,77 @@
...
+     *
+     * @Route("/calendar")
+     *
+     * @param string|null $user_id
+     * @return MultiResponse
+     * @throws Exception if a Gradeable failed to load from the database
+     * @see GlobalController::prep_user_sidebar
+     */
+    public function viewCalendar($user_id = null): MultiResponse {
+        if ($this->gradeables == []) {
```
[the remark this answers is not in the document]
zhangz22: Fix in the next commit

site/app/templates/calendar/Calendar.twig
```diff
@@ -0,0 +1,70 @@
...
+
+{# Begin Calendar #}
+<div class="content">
+    <h1>Calendar</h1>
+    <div class="cal-content">
+        <div class="cal-toolbar">
+            {# Replace by js function call #}
+        </div>
+
+        <div class="calendar" style='margin-bottom: 20px' id="full_calendar">
```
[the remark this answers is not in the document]
zhangz22: Sure. Fix in the next commit

site/app/controllers/calendar/CalendarController.php
```diff
@@ -0,0 +1,77 @@
...
+            $this->loadFromDB($user_id);
+        }
+        $gradeable_list = new GradeableList($this->core, null, $this->gradeables);
+
+        return new MultiResponse(
+            JsonResponse::getSuccessResponse($gradeable_list->getGradeablesBySection()),
```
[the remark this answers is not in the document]
zhangz22: Sure. Fix in the next commit

site/app/views/calendar/CalendarView.php
```diff
@@ -0,0 +1,129 @@
...
+     *              "title" => string (title shown at the top of the table),
+     *              "subtitle" =>  string (title shown at the top of the table, if any. Can be empty),
+     *              "section_id" => string (the id of the section. Will be used as the HTML id)
+     *              "gradeables" => array of ELEMENT. The structure of ELEMENT is described above
+     * @param string $empty_message string that will shown at the table of list section when no items are listed
+     * @return string
+     * @see NavigationView::gradeableSections
+     */
+    public function showCalendar(array $items_by_date, array $item_by_sections, string $empty_message): string {
+        $year = (isset($_GET['year']) && $_GET['year'] != "")  ?  (int) $_GET['year']  : (int) date("Y");
```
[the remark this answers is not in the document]
zhangz22: These $_GET parameters are not used in the program itself currently. It is just for debugging and testing. If I manually type a link .../calendar?month={m}&year={y} the page would jump to that month.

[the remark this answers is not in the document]
shailpatels: Ah ok, that makes sense. If its only for debugging I'd suggest removing it before getting merged but if this is setting up for a feature - (I could see users wanting the copy+paste specific calender dates), feel free to resolve this and associated comments regarding the `$_GET` params

site/app/views/calendar/CalendarView.php
```diff
@@ -0,0 +1,129 @@
...
+
+        $this->core->getOutput()->addInternalCss("navigation.css");
+        $this->core->getOutput()->addInternalCss('calendar.css');
+        $this->core->getOutput()->addInternalJs('calendar.js');
+        $this->core->getOutput()->enableMobileViewport();
+        $this->core->getOutput()->addBreadcrumb("Calendar");
+
+        return $this->core->getOutput()->renderTwigTemplate("calendar/Calendar.twig", [
+            "show_only_calendar" => 0,
+            "view_year" => $year,          // the year that the calendar is viewing
```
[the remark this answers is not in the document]
zhangz22: This is just a reserved option to allow the calendar to show a specific date by chaning the parameters `view_year `and `view_month`. `curr_month`, `curr_year` and `curr_day` is representing the real time of the user and the calendar will mark that date.

site/app/views/calendar/CalendarView.php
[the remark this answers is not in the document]
zhangz22: Yeah. This would be userful in potential features like jumping to a specific date, viewing a gradeable/course material in calendar, and so on.

site/app/templates/calendar/Calendar.twig
```diff
@@ -0,0 +1,70 @@
...
+                                    <th>Gradeable</th>
+                                    <th>Submission Closes</th>
+                                </tr>
+                            </thead>
+                            <tbody>
+                            {% for id, gradeable in info.gradeables %}
+                                <tr>
+                                    <td>{{ gradeable.course }}</td>
+                                    <td><a href="{{ gradeable.url }}">{{ gradeable.title }}</a></td>
+                                    <td>{{ (gradeable.submission != '') ? gradeable.submission|date('m/d/Y @ h:i A T') : '' }}</td>
```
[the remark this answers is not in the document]
zhangz22: Thanks for this suggestions. Surely a pre-configured date format is important and I didn't know there was a config. I'd like to ask is there a way also to enable this filter in Javascript files？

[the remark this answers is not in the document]
shailpatels: hmm I don't think there's a good way, the PHP DateTime reference has a ton of formatting options and JS doesn't really have an equivalent function. If possible you can construct the date and pass it from the server and embed it to JS. But other than that I think the only way is to construct the date in the expected format.

And above I just chose `getFormat('gradeable') ` as an example. You can add your own 'calendar' format in
`site/app/models/DateTimeFormat`. Getting the same format from this model everywhere in the calendar helps us in case we decide to change the formatting in the future, we only have to change 1 place.

site/app/controllers/calendar/CalendarController.php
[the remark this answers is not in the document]
shailpatels: Only resolve the conversation if you've made the actual commit to the branch and pushed. -GitHub will also show the conversation as outdated once a push has been made to indicate the file has been updated since my requested change is made.

This way we can quickly see if more work needs to be done or not in a PR :)

site/public/css/calendar.css
```diff
@@ -0,0 +1,346 @@
...
+    border: var(--hover-good-green);
+}
+
+a.cal-gradeable-status-5:focus {
+    color: var(--btn-text-black);
+    background-color: var(--disabled-good-green);
+    border: var(--disabled-good-green);
+}
+
+@media (max-width: 800px) {
```
shailpatels, line 334: This css media query was only effecting the list of gradeables below the calendar, I thought that was part of the final page but since it got removed you can remove this for now - sorry about that

shailpatels (changes requested):

site/public/js/calendar.js
```diff
@@ -0,0 +1,318 @@
...
+const monthNames = ['December', 'January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'];
+const monthNamesShort = ['Dec', 'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sept', 'Oct', 'Nov', 'Dev'];
+
+/**
+ * Gets the previous month of a given month
+ * @param month : int the current month (1 as January and 12 as December)
+ * @param year : int the current year
+ * @returns {number[]} : array<int> {previous_month, year_of_previous_month}
+ */
+// eslint-disable-next-line no-unused-vars
```
MasterOdin, line 11: You should be doing `/* exported prevMonth */` at the top of this file per the note at https://submitty.org/developer/coding_style_guide/javascript#style-rules and not explicitly disabling this rule. This applies to the other instances of `// eslint-disable-next-line` you have in this file for functions.

site/public/js/calendar.js
```diff
@@ -0,0 +1,318 @@
...
+        else if (month <= 0) {
+            month = month + 12;
+        }
+        content += `<span class="cal-next-month-date cal-day-title">${month}/${day}</span>`;
+    }
+    content += '</div>';
+
+    // List all gradeables of other items
+    content += '<div class="cal-cell-items-panel">';
+    // eslint-disable-next-line no-undef
```
MasterOdin, line 109: similar to above, but the inverse, you can use `/* global gradeables_by_date */` at the top of the file to indicate stuff that's coming from outside this script but is defined.

site/app/templates/calendar/Calendar.twig
```diff
@@ -0,0 +1,70 @@
+{% import _self as self %}
+
+<script>
+    let gradeables_by_date = {{ gradeables_by_date|json_encode|raw }};
+    let curr_year = {{ curr_year }};
+    let curr_month = {{ curr_month }};
+    let curr_day = {{ curr_day }};
```
MasterOdin, line 7: use `const` for these if they're not being re-assigned during execution.

site/app/models/gradeable/GradeableList.php
```diff
@@ -251,4 +252,34 @@ public function getGradeablesBySection(): array {
...
+        $gradeables = [];
+        // Load the gradeable information for each course
+        $courses = $core->getQueries()->getCourseForUserId($user->getId());
+        foreach ($courses as $course) {
+            /** @var \app\models\Course $course */
+            $core->loadCourseConfig($course->getSemester(), $course->getTitle());
+            $core->loadCourseDatabase();
+            foreach ($core->getQueries()->getGradeableConfigs(null) as $gradeable) {
+                /** @var Gradeable $gradeable */
+                $gradeables["{$course->getSemester()}||{$course->getTitle()}||{$gradeable->getId()}"] = $gradeable;
```
MasterOdin, line 277: Doing it this way with the `||` means that the code will currently break if the course title has a `||` in it as it'll affect the `explode` you do elsewhere. Could we use the course code instead which is better guaranteed to not have a `|` character in it?

site/app/models/gradeable/GradeableList.php
```diff
@@ -6,6 +6,7 @@
 use app\libraries\GradeableType;
 use app\models\AbstractModel;
 use app\models\User;
+use app\views\NavigationView;
```
MasterOdin, line 9:
```suggestion
```

unused, can delete.

site/app/models/CalendarInfo.php
```diff
@@ -0,0 +1,131 @@
...
+                // Put gradeables in by-date maps according to section (close/open)
+                $dueDate = ($gradeable->getType() === GradeableType::ELECTRONIC_FILE) ? $gradeable->getSubmissionDueDate()->format('Y-m-d') : '';
+                $info->items_by_date[$dueDate][] = $currGradeable;
+            }
+
+            // Put data of current section into the by-section map
+            $info->items_by_sections[] = $curr_section;
+        }
+
+        $info->empty_message = "There are currently no assignments posted.  Please check back later.";
```
MasterOdin, line 115: Do we need to define this string within this factory method, or can we just assign it to `private $empty_message = ''` on line 49 above? Or probably better yet just put this string within the twig template and get rid of this variable altogether.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -0,0 +1,43 @@
...
+     * @return WebResponse
+     * @see GlobalController::prep_user_sidebar
+     * @see CalendarView::showCalendar
+     */
+    public function viewCalendar(): WebResponse {
+        $user = $this->core->getUser();
+        try {
+            $gradeable_list = GradeableList::getAllGradeableListFromUserId($this->core, $user);
+        }
+        catch (Exception $e) {
```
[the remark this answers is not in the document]
zhangz22: Sorry I didn't know there was a generic exception handler. Should I just remove the catch block and let the function throw the exception, and then the exception handler would then handle it?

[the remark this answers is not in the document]
MasterOdin: Yeah, I would just remove it. For reference, the handler lives at https://github.com/Submitty/Submitty/blob/master/site/public/index.php#L37-L59.

site/public/js/calendar.js
zhangz22: I'd prefer to let the month use 1-12 since the associative arrays storing the gradeable info in php are using 'Y-m-d' formatted strings (which use 1-12 as month), and also php date object is using 1-12 as month... Just to keep them consistent and intuitive

site/public/css/calendar.css
zhangz22: Yes after a discussion in the meeting, the table was removed from the calendar page. But the code was still there and was disabled by a parameter. I think I would just keep those code as a hidden easter egg and in case some future feature that uses the calendar page can make use of the table, maybe

site/app/controllers/calendar/CalendarController.php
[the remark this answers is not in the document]
zhangz22: Sure.

site/app/models/CalendarInfo.php
zhangz22: This string may be changed in different factory methods. Currently CalendarInfo handles Gradeables, but in future commits the calendar will be used for course materials (Course Calendar feature) and this string will be contents related with course materials

site/public/js/calendar.js
```diff
@@ -0,0 +1,320 @@
...
+ * Creates a HTML table cell that contains a date.
+ *
+ * @param year : int the year of the date
+ * @param month : int the month of the date (1 as January and 12 as December)
+ * @param day : int the date of the date (1 - 31)
+ * @param curr_view_month : int the current month that the calendar is viewing
+ * @param view_semester : if the calendar is viewing the entire semester. If so, the day cell would show both the month and date
+ * @returns {string} the HTML string containing the cell
+ */
+function getDayCell(year, month, day, curr_view_month, view_semester=false) {
```
hdrodz97, line 79: Nit: `getDayCell` seems to imply we’re fetching data from somewhere rather than constructing a blob of HTML to output somewhere else, `generateDayCell` seems like a more appropriate name for this

site/public/js/calendar.js
```diff
@@ -0,0 +1,320 @@
...
+    return content;
+}
+
+/**
+ * Generates the title area for the calendar.
+ *
+ * @param title_area the title of the calendar (month+year/semester/...)
+ * @returns {string} the HTML code for the title area
+ */
+function getCalendarHeader(title_area) {
```
hdrodz97, line 145: (See `getDayCell` comment)

site/public/js/calendar.js
```diff
@@ -0,0 +1,320 @@
...
+}
+
+/**
+ * This function creates a table that shows the calendar.
+ *
+ * @param view_year : int year that the calendar is viewing
+ * @param view_month : int month that the calendar is viewing (1 as January and 12 as December)
+ * @returns {string} the HTML string contains the entire calendar table displaying view_month/view_year
+ */
+function getCalendarOfMonth(view_year, view_month) {
```
hdrodz97, line 173: (See `getDayCell` comment)

site/public/js/calendar.js
```diff
@@ -0,0 +1,320 @@
...
+
+/**
+ * Creates a calendar of the entire semester.
+ *
+ * @param start the start date of the semester in the format of YYYY-mm-dd
+ * @param end the end date of the semester in the format of YYYY-mm-dd
+ * @param semester_name the name of the semester
+ * @returns {string} the HTML string containing the cell
+ */
+function getFullCalendar(start, end, semester_name) {
```
hdrodz97, line 247: (See `getDayCell` comment)

hdrodz97 (changes requested):
Nice job, really liking how this is coming together so far! Two things apart from my general comments:

1. I found a bug in the tooltips regarding date/time parsing wherein the tooltips for every assignment say the assignment is due on “NaN/NaN/NaN @ NaN:NaN” (see the screenshot below). This only happens on Safari (v14); Chrome (on Ubuntu) and Firefox (on Ubuntu and macOS) aren’t reproducing it.
image
2. I like the use of color to indicate the status of various assignments, but I think there should be a secondary way to indicate this to accommodate for users with colorblindness. One idea I had was to include this information in the tooltip (something like having an extra line showing whether the gradeable is open or currently grading or what have you), but this wouldn’t work on mobile. I don’t really have any great ideas on this front, but it’s something worth looking into.

I also can’t comment too much on the PHP as I’m not too familiar with that region, but it looks like you’re already in good company for that 🙂

site/app/templates/calendar/Calendar.twig
```diff
@@ -0,0 +1,70 @@
...
+
+{# Begin Calendar #}
+<div class="content">
+    <h1>Calendar</h1>
+    <div class="cal-content">
+        <div class="cal-toolbar">
+        </div>
+
+        <div class="calendar" style='margin-bottom: 20px' id="full-calendar">
+            {# Replace by js function call #}
```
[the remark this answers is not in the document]
zhangz22: This <div> is supposed to be empty. The entire calendar is generated by js function and the js function will run once the page is load (the onload comment above) to replace this div to be the calendar

site/app/templates/calendar/Calendar.twig
```diff
@@ -0,0 +1,70 @@
...
+
+<script>
+    const gradeables_by_date = {{ gradeables_by_date|json_encode|raw }};
+    const curr_year = {{ curr_year }};
+    const curr_month = {{ curr_month }};
+    const curr_day = {{ curr_day }};
+
+    window.onload = function () {
+        loadCalendar({{ view_month }}, {{ view_year }});
+    }
```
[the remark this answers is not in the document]
zhangz22: EventListener is fixed.
I locate the NaN issue on macOS and it is due to the fact that parsing a string to a date object is implementation-dependent and there are inconsistency on the support for a date format between different browsers. I'll try to fix this issue

shailpatels:
@zhangz22 The issue with the `NaN` as a date on safari is most likely that safari only accepts 
ISO 8601 format which has a `T` in between the day and time parts, e.g : `YYYY-MM-DDTHH:MM:SS` and refuses to accept anything outside of that format in JS whereas firefox and chrome do not mind if the `T` is there or not.

hdrodz97 (approved):
Looks good to me!

site/public/js/calendar.js
shailpatels: ok, sounds good

shailpatels (approved):
Changes look good and should be ready to merge now

Issue #4432, opened 2019-09-05
Too many announcements in the forum


bmcutler:
**What problem are you trying to solve with Submitty**
If an instructor has made a couple dozen email announcements in the forum over the course of the semester, and leaves the star enabled for every post (they don't realize they can remove the star), those posts will be at the top of the forum and will drown out new threads.

**Describe the way you'd like to solve this problem**
Have some sort of hint to the instructor that they can un-announce the old announcements.  Instructor (and maybe all graders??) will have some extra messaging, suggestion, that they can accept (or decline), that will remove stars from threads that are at least 7 days old.

sak6lab:
Discussed having a modal that appears when there are more than 5 announcements that asks to remove stars from announcements that are more than 7 days old.

bmcutler:
Design revision (this is misused misunderstood in most classes, and let's avoid the annoying popups and manual announcement maintenance):

Let's add an 'expiration date' to every announcement thread (can be infinity (year 9999) for all existing threads)

For all new announcement threads lets default this expiration date to be 7 days from creation date.  Once "announcement" is selected, the instructor has a date picker UI to revise this date if desired.

Once the date passes, either 
  1) the star symbol (or whatever we agree on as the final symbol -- see issue #4520) disappears and the thread displays like a non announcement thread (now sorted chronologically)
OR
  2) the star symbol changes color(turns grey?) and the the thread drops  into chronological order.

1 might be difficult to accomplish?  What's stored in the database?  What does it look like it we try to edit this post?

2 might be less confusing?  Or more confusing?

PR #6310, opened 2021-03-24
[Feature:Forum] Announcement expiration date

jenaybarela:

Addresses #4432 

### What is the new behavior?
Announcements and pinned threads now have an expiration date. When that date passes the thread is no longer pinned.


migration/migrator/migrations/course/20210122001721_pinned_expiration.py
```diff
@@ -0,0 +1,36 @@
...
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    if not database.table_has_column('threads', 'pinned_expiration'):
+        database.execute("ALTER TABLE threads ADD IF NOT EXISTS pinned_expiration timestamp with time zone NOT null DEFAULT '1900-01-01 00:00:00'");
+        database.execute("UPDATE threads SET pinned_expiration = CAST(CASE WHEN pinned = true THEN '9998-01-01 00:00:00' ELSE '1900-01-01 00:00:00' END AS timestamp with time zone)");
```
BenKelly99, line 19: Would it be possible to leave the existing expiration date if the column already existed? Right now, if we have to rollback and then run this again, it will override all the instructor's expiration dates.

site/app/templates/forum/ThreadPostForm.twig
```diff
@@ -165,6 +165,12 @@
             <label class="inline-block" for="pinThread">{{ "Pin Thread" }}</label>
             <input type="checkbox" class="pinThread-checkbox" name="pinThread" id="pinThread" value="pinThread" data-ays-ignore="true"/>
         {% endif %}
+        {% if show_expiration and core.getUser().accessFullGrading() %}
+            <div class = "expiration inline-block" style="display: none;">
+              <label class="inline-block" for="expirationDate">{{ "Expiration Date" }}</label>
+              <input type="date" class="date_picker" name="expirationDate" id="expirationDate"/>
```
BenKelly99, line 171: Would it be better to have the value of this textbox default to 7 days in the future? That way the instructor knows that the default is 7 instead of it just setting it to 7 days when the box is empty.

site/public/js/forum.js
```diff
@@ -1897,6 +1904,7 @@ function updateThread(e) {
     data,
     success: function (response) {
       try {
+        alert(JSON.stringify(response));
```
BenKelly99, line 1907: Is this a debug statement? If so, please remove.

site/public/js/forum.js
```diff
@@ -798,12 +798,17 @@ function showEditPostForm(post_id, thread_id, shouldEditThread, render_markdown,
...
                 var thread_lock_date =  json.lock_thread_date;
                 var thread_status = json.thread_status;
+                var expiration = json.expiration;
                 $("#title").prop('disabled', false);
                 $(".edit_thread").show();
                 $('#label_lock_thread').show();
                 $("#title").val(thread_title);
                 $("#thread_status").val(thread_status);
                 $('#lock_thread_date').val(thread_lock_date);
+                if (Date.parse(expiration) > new Date()) {
```
BenKelly99, line 808: Have you tested this with conflicting timezones? Since the timezone is stored in the DB it should be fine but just to be sure.

BenKelly99 (changes requested):

BenKelly99:
Also if you haven't already, it probably worth documenting this feature on submitty.org

migration/migrator/migrations/course/20210122001721_pinned_expiration.py
jenaybarela: It shouldn't override any existing data since its surrounded by the if table has column statement.

site/app/templates/forum/ThreadPostForm.twig
jenaybarela: Fixed

migration/migrator/migrations/course/20210122001721_pinned_expiration.py
BenKelly99: Right, whoops haha

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -810,7 +810,12 @@ public function getThreadTitle(int $thread_id) {
     }
 
     public function setAnnouncement(int $thread_id, bool $onOff) {
-        $this->course_db->query("UPDATE threads SET pinned = ? WHERE id = ?", [$onOff, $thread_id]);
+        if ($onOff) {
+            $this->course_db->query("UPDATE threads SET pinned_expiration = ? WHERE id = ?", [date("Y-m-d H:i:s", strtotime('+7 days')), $thread_id]);
+        }
+        else {
+            $this->course_db->query("UPDATE threads SET pinned_expiration = ? WHERE id = ?", ['1900-01-01 00:00:00', $thread_id]);
+        }
```
MasterOdin, line 818:
```suggestion
        $expireTime = $onOff ? date("Y-m-d H:i:s", strtotime('+7 days')) : '1900-01-01 00:00:00';
        $this->course_db->query("UPDATE threads SET pinned_expiration = ? WHERE id = ?", [$expireTime, $thread_id]);
```

MasterOdin (commented):
Given this change, do we still need the `pinned` boolean in the `threads` table following the migration or can that be removed?

site/app/libraries/database/DatabaseQueries.php
jenaybarela: Yeah the pinned column can be removed.

jenaybarela:
> Given this change, do we still need the `pinned` boolean in the `threads` table following the migration or can that be removed?

No the pinned column can be removed.

site/public/js/forum.js
jenaybarela: I have tested with conflicting time zones it seems to work correctly.

BenKelly99 (approved):

bmcutler (changes requested):
I can't seem to edit the expiration date for an an announcement PR.
I enter a new date in the GUI, press "Update Post" but the date doesn't change.

jenaybarela:
> I can't seem to edit the expiration date for an an announcement PR.
> I enter a new date in the GUI, press "Update Post" but the date doesn't change.
> 
> [edit to add, I'm testing on Mac Chrome]

I'm attempting to reproduce this bug on Mac and Chrome but I didn't have the same issue. Could you could you provide screenshots?

bmcutler (approved):
Hmm, I can't reproduce the error now

PR #6311, opened 2021-03-24
[Bugfix:Autograding] More error handling for short-circuiting

hdrodz97:

If a gradeable is being rebuilt when there are submissions for it in the autograding queue, then there is a moment where the gradeable's complete config does not exist. This causes `try_short_circuit` to fail and potentially gets the shipper process to hang.

### What is the new behavior?

There are two new exception guards in the code: one for handling config loading specifically, and another that prints a generic error message for any unhandled errors such that we can check it out and handle the situation. In both cases, the behavior is the same: we give up on short-circuiting and let it attempt to go through the standard autograding procedure.


autograder/submitty_autograding_shipper.py
```diff
@@ -1299,18 +1311,29 @@ def try_short_circuit(config: dict, queue_file: str) -> bool:
...
+    # currently in the process of being re-built. In this case, we give up on trying to
+    # short-circuit and let the full autograder handle it. Best-case scenario, the config once
+    # again exists and we can grade it normally. Worst-case scenario, the config is still out of
+    # commission, but the grader is better-equipped to handle this scenario.
+    try:
+        with open(config_path) as fd:
+            config_obj = json.load(fd)
+        with open(gradeable_config_path) as fd:
+            gradeable_config_obj = json.load(fd)
+    except Exception as e:
```
MasterOdin, line 1324: Can we narrow the types of exceptions we expect here? Broadly catching any and all exceptions is not good practice as it ends up meaning we're going to run the risk of catching stuff we don't mean to. For example, if you mistyped `json.load` or `open` or whatever, it'll throw a `NameError` or an `AttributeError` which are both caught by `Exception` meaning we'll just constantly hit this except block even though the code is broken in a way we do not expect.

Instead, we should catch on as narrow a scope as possible. For example, within this block, we may expect we'd hit `OSError` and `json.JSONDecodeError` where the former covers still a wide swathe of conditions (cannot open, file permissions, it's a directory instead of a file, etc.), but it's narrower than `Exception` and limits only to OS related stuff. At best, doing this within dedicated code paths just strikes as "we don't actually know the side-effects of our code, let's catch it all".

I realize there's a lot of these `except Exception` catches within the autograding code, but please let's not add yet another that is definitely easy to avoid in this particular case.

autograder/submitty_autograding_shipper.py
```diff
@@ -635,9 +635,19 @@ def grade_queue_file(config, my_name, which_machine, which_untrusted, queue_file
...
     # Try to short-circuit this job. If it's possible, then great! Clean
     # everything up and return.
-    if try_short_circuit(config, queue_file):
-        grading_cleanup(config, my_name, queue_file, grading_file)
-        return
+    try:
+        if try_short_circuit(config, queue_file):
+            grading_cleanup(config, my_name, queue_file, grading_file)
+            return
+    except Exception as e:
```
MasterOdin, line 642: While I'd like to see this exception be narrowed, the underlying function calls within `try_short_circuit` and `grading_cleanup` do not do a great job of narrowing themselves on possible (and also have their own `exception Exception` clauses) so fine to punt on improving this for now.

MasterOdin (approved):
Code looks fine. Would be good to think on how to add tests around these changes to catch this sort of issue as race conditions are easy to re-introduce if not careful, but will not block on that given that it's breaking in production with some frequency. I would like to see the exception clause changed in the one spot though.

emaicus (approved):
I tested this and it works as expected.

autograder/submitty_autograding_shipper.py
hdrodz97: Narrowed it down to `FileNotFoundError` as that's the particular error that's not particularly a sign that something's gone wrong but rather a sign that the submission was a victim of unfortunate timing (being graded in the middle of a gradeable rebuild)

PR #6312, opened 2021-03-27
[Bugfix:Autograding] Avoid crash when listing containers

emaicus:
A race condition caused the `client.containers.list` to fail when a container was deleted during a call. This PR adds the `sparse=True` option, which tells the function to return containers without evaluation, preventing the call from failing. Subsequent errors are handled on our end (i.e., if we attempt to remove a nonexistent container).


MasterOdin:
In the case we get a a container list that includes removed containers, and we go to try and remove them, does failing to remove it because it's already been removed constitute an error condition (that we're guarding against, but should still be fixed at some point), or is it something that's fine?

emaicus:
@MasterOdin, yes, that is an acceptable case. I've added an `except` for `docker.errors.NotFound` which passes so that we aren't chatty if it happens.

PR #6326, opened 2021-04-07
[Feature:TAGrading] TA Grading Touch Panel Resize

DefinitiveAbove:

You cannot resize the TA grading panels when using only a touchscreen.

### What is the new behavior?
You can resize the TA grading panels with a touchscreen.

Tested on a Lenovo touchscreen running Brave 1.22.71 (Chromium 89.0.4389.114), and an iPad (iOS 14.2) - note: currently cannot scroll in panel selection popup on a touchscreen.

Steps for testing on a remote device:
1) Stop vagrant machine if running
2) In Vagrantfile, add a new line in config.vm.define under the site, websockets, and database ports - `ubuntu.vm.network 'forwarded_port', guest: 22, host: <port>, id: "ssh", host_ip: "0.0.0.0"`, replacing port with the port you want to expose for ssh.
3) Expose ports if necessary on device
4) Start vagrant machine with `vagrant up ssh` (or at the end of the Vagrantfile, comment out the if and end for ARGV.include?)
5) Retrieve SSH keys for vagrant machine from (`<SUBMITTY GIT REPO>/.vagrant/machines/<VM>/<VM BACKEND>/private_key`; on my machine the VM is ubuntu-18.04 and the VM backend is virtualbox)
6) Connect remote device to same network as local machine, use SSH utility such as Blink or Terminus (iOS apps, there are probably others for both iOS/Android/other devices) to connect to the port on the local machine with username root and private key from above; when you SSH, use parameter "-L" to forward site and websocket ports from local machine to remote device (ex. `ssh -L 1501:localhost:1501 -L 8443:localhost:8443 root@$SUBMITTY_HOST` , where $SUBMITTY_HOST is the local machine address)
     1) If you are on iOS, make sure that the SSH utility that you're using supports location tracking to keep the connection active in the background (for Blink, use `geo track`)
7) On remote device, navigate to localhost:1501


hdrodz97 (changes requested):
Works great on Firefox 87 and Chrome 89 on Ubuntu 20.04, as well as on Firefox 87 on Windows 10.

Edge 89 on Windows 10, however, has a "rubber scrolling" behavior similar to that of iOS, and that seems to kick in when dragging the resize handles. The panels still resize, but the entire page is also slowly moving at the same time, and when resizing panels horizontally, there is a chance that a user could unintentionally trigger the browser's "go back" gesture. Not sure if this also happens on iOS as I have no iOS devices to test.

I messed around a little bit with the CSS using Edge's dev tools, and I found that [adding `touch-action: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action) to the handles' CSS stopped this from happening.

Finally, in the "remote testing" instructions, there seems to be a small typo: I found I had to add an extra `-L` before the second tunnel description, so the full SSH command should be:
```sh
ssh -L 1501:localhost:1501 -L 8443:localhost:8443 root@$SUBMITTY_HOST
```
where `$SUBMITTY_HOST` is the IP address/hostname of the machine running the Submitty Vagrant box.

hdrodz97 (approved):
Perfect, works great on both Edge and Firefox on Windows, as well as on Chrome and Firefox on Ubuntu.

MasterOdin:
@DefinitiveAbove could you make a PR against https://github.com/Submitty/submitty.github.io and add a new section to https://submitty.org/developer/vm_install_using_vagrant with your above instructions on testing against a touch device?

hdrodz97:
I think I found something that might explain the inability to scroll in the panel selection popup. This isn't a bug introduced by this PR, but it bears mentioning anyway. If you click and drag on the faded background behind the popup window, you'll find that both the background and the popup window will be dragged along the window. Here's a screenshot of what that looks like:

image

Looks like the background has the `ui-draggable` and `ui-draggable-handle` classes associated with it when it probably shouldn't. My guess is that this is messing with the browser's handling of touch scrolling, since touch scrolling does work in the Settings panel which does not have the same "draggable background" bug.

PR #6340, opened 2021-04-17
[Refactor:Developer] Add simpler NavButton model

MasterOdin:

The `GlobalController` creates a bunch of instances of the `Button` model to set-up the left sidebar. For each instance that is created, there is some amount of boilerplate, namely that each have the same class, and that the id is usually `nav-sidebar-${title}`.

### What is the new behavior?

This adds a new `NavButton` model that removes a bit of boilerplate on the `Button` model as it was used in the `GlobalController` for setting up the sidebar.

### DIFF
--- a/site/tests/app/models/ButtonTester.php
+++ b/site/tests/app/models/ButtonTester.php
@@ -2,12 +2,12 @@
 
 namespace tests\app\models;
 
+use app\libraries\Core;
 use app\models\Button;
-use tests\BaseUnitTest;
 
-class ButtonTester extends BaseUnitTest {
+class ButtonTester extends \PHPUnit\Framework\TestCase {
     public function testDefaults() {
-        $button = new Button($this->createMockCore(), []);
+        $button = new Button(new Core(), []);
         $this->assertNull($button->getTitle());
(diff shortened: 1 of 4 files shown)


site/tests/app/models/ButtonTester.php
```diff
@@ -47,12 +47,12 @@ public function testOverrides() {
...
 
     public function testWrongTypeProgress() {
-        $button = new Button($this->createMockCore(), ['progress' => 'a']);
+        $button = new Button(new Core(), ['progress' => 'a']);
         $this->assertEquals(0, $button->getProgress());
     }
 
     public function testDate() {
-        $button = new Button($this->createMockCore(), []);
+        $button = new Button(new Core(), []);
```
MasterOdin, line 55: These changes are to remove the usage of the mock which is no longer necessary due to prior work on making the Core constructor have less initialization in it. This is mostly a slight optimization to speed-up running the test suite as it's quicker to make a `Core` object than it is to make a mock of it. This applies elsewhere in this file.

shailpatels (approved):

PR #6345, opened 2021-04-21
[Refactor:TAGrading] Subsection sort for checkpoint grading

vikrsri:

There is currently no way for graders to select whether they want to sort by subsection for checkpoint gradeables.

### What is the new behavior?
Now, graders can select their desired sorting method using a drop-down menu:
image

All sorting options have been moved to this drop-down menu.


site/app/templates/grading/simple/Display.twig
```diff
@@ -72,6 +72,32 @@
...
+            </script>
+        {% elseif sort == "first"  %}
+            <script>
+                $('select').val( "{{ grading_url }}?view={{ view_all ? 'all' : '' }}&sort=first" )
+            </script>
+        {% elseif sort == "last"  %}
+            <script>
+                $('select').val( "{{ grading_url }}?view={{ view_all ? 'all' : '' }}&sort=last" )
+            </script>
+        {% endif %}
```
MasterOdin, line 99: A couple of suggestions here:
1. We can use the shorthand ternary statement in twig to condense a few bits, so that `{{ cond ? 'foo' }}` is equivalent to `{{ cond ? 'foo' : '' }}`.
2. Let's inline our selection into the HTML, instead of having this separate JS, by doing `{{ sort == "foo" ? 'selected' }}`
3. Since the start of the URL for each option is the same (`{{ grading_url }}?view={{ view_all ? 'all' }}`), let's move this into the `onChange` handler to reduce having it in every option.
```suggestion
        <select
            id="subsection_sort"
            name="subsection_sorting"
            onChange="window.location.href='{{ grading_url }}&view={{ view_all ? 'all' }}&sort=' + this.value"
            style="font-size:15px;"
        >
            <option value="section_subsection" aria-label="sort by Section then Subsection" {{ sort == "section_subsection" ? 'selected' }}>
                Section then Subsection then User ID
            </option>
            <option value="id" aria-label="sort by ID" {{ sort == "id" ? 'selected' }}>
                Section then User ID
            </option>
            <option value="first" aria-label="sort by First Name" {{ sort == "first" ? 'selected' }}>
                Section then First Name then User ID
            </option>
            <option value="last" aria-label="sort by Last Name" {{ sort == "last" ? 'selected' }}>
                Section then Last Name then User ID
            </option>
        </select>
```

You'll want to double check the `onChange` clause as I freehand wrote that so it might not be perfectly right.

site/app/templates/grading/simple/Display.twig
```diff
@@ -94,19 +120,13 @@
                 <th style="width:20px">#</th>
                 <th style="width:70px" scope="col">Section</th>
                 <th style="width:90px; text-align: left" scope="col">
-                    <a href="{{ grading_url }}?view={{ view_all ? 'all' : '' }}&sort=id" aria-label="sort by ID">
-                        <span class="tooltiptext" title="sort by User ID">User ID </span><i class="fas fa-sort"></i>
-                    </a>
+                    <span class="tooltiptext" title="sort by User ID">User ID </span>
```
MasterOdin, line 123: Given that we've removed the sort on these headers, we should probably also remove the `tooltiptext` as well since it's no longer valid.

This carries for the changes below here as well.

site/app/controllers/grading/SimpleGraderController.php
```diff
@@ -124,6 +124,9 @@ public function gradePage($gradeable_id, $view = null, $sort = null) {
         elseif ($sort === "first") {
             $sort_key = "coalesce(NULLIF(u.user_preferred_firstname, ''), u.user_firstname)";
         }
+        elseif ($sort == "section_subsection") {
+            $sort_key = "u.registration_subsection";
+        }
         else {
             $sort_key = "coalesce(NULLIF(u.user_preferred_lastname, ''), u.user_lastname)";
         }
```
MasterOdin, line 132: If the intention is to make this the default sort method, than you should set it to be the last element in the `if/elseif/else` chain here. This also means you don't need to modify the `$sort` parameter.
```suggestion
        elseif ($sort === "last") {
            $sort_key = "coalesce(NULLIF(u.user_preferred_lastname, ''), u.user_lastname)";
        }
        else {
            $sort_key = "u.registration_subsection";
        }
```

MasterOdin (changes requested):

bmcutler:
@eddiekrystowski 
Can you please review this PR and take over all making required changes?  Thanks!

site/app/templates/grading/simple/Display.twig
```diff
@@ -108,19 +130,13 @@
...
-                    <a href="{{ grading_url }}?view={{ view_all ? 'all' : '' }}&sort=first" aria-label="sort by First Name">
-                        <span class="tooltiptext" title="sort by First Name">First Name </span><i class="fas fa-sort"></i>
-                    </a>
+                    <span title="sort by First Name">First Name </span>
                 </th>
                 <th style="width:110px; text-align: left" scope="col">
-                    <a href="{{ grading_url }}?view={{ view_all ? 'all' : '' }}&sort=last" aria-label="sort by Last Name">
-                        <span class="tooltiptext" title="sort by Last Name">Last Name </span><i class="fas fa-sort"></i>
-                    </a>
+                    <span title="sort by Last Name">Last Name </span>
```
MasterOdin, line 139: These span titles no longer make sense and should be removed or the links within the headers as a quick version of the dropdown restored.

MasterOdin (changes requested):

MasterOdin (approved):

PR #6353, opened 2021-04-26
[Bugfix:System] Fix dependency caching

### LINKED ISSUE #6352: Cache pip in e2e testing not working
Context: GitHub Action for e2e testing, the pip install was intended to be cached, but it isn't.

I don't think `Cache pip` is working.
For example if you see the GitHub action triggered by the latest commit e33b67e4cc4892d00edd894257e13f3408ed8826 :
Log: https://github.com/Submitty/Submitty/runs/2433159805?check_suite_focus=true#step:8:20
The log says:
```
Received 22 of 22 (100.0%), 0.0 MBs/sec
Cache Size: ~0 MB (22 B) <<< SEE HERE
/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/9ef3919a-4e7e-4d6e-b822-21ad22725245/cache.tzst -P -C /home/runner/work/Submitty/Submitty
Cache restored successfully
Cache restored from key: Linux-py-pip-
```

And you can see step 9 (install python dependencies) taking 1 minute:
https://github.com/Submitty/Submitty/runs/2433159805?check_suite_focus=true#step:9:56

_Originally posted by @wusatosi in https://github.com/Submitty/Submitty/issues/6212#issuecomment-826496640_

wusatosi:

* [ ] ~~Tests for the changes have been added/updated (if possible)~~ (NA)
* [ ] Documentation has been updated/added if relevant (I don't think this applies)

This pr relates to #6352, but is not a fix for it as this is not targeted for e2e action.

Currently, GitHub actions do not cache pip dependency for e2e as intended because the cache key is constant.

This is due to the current config using `hashFiles('**/requirements.txt')` to vary the cache key (line 225), which is constant as we do not use `requirments.txt` for pip dependency management (tracked by #3630)
https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/.github/workflows/submitty_ci.yml#L221-L228

### What is the new behavior?
Instead of relying on hashing `requirments.txt`, I use the current commit SHA as the key, so it always retrieves the dependency of the last CI run. This is applied for python-lint and python-test actions but not e2e action as we will wait for #3630 to resolve.

### Other information?
This does not change the fact that pip always uses the latest versions of packages for all actions, which we may need to alter  in the future.


wusatosi:
This is actually more complicated than I thought. After the fix, python-unit and python-lint has their proper dependency caching:
```
Received 0 of 11394434 (0.0%), 0.0 MBs/sec
Received 11394434 of 11394434 (100.0%), 7.4 MBs/sec
Cache Size: ~11 MB (11394434 B)
/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/18885c69-6dd1-4434-98ab-a8c8ac455e3f/cache.tzst -P -C /home/runner/work/Submitty/Submitty
Cache restored successfully
Cache restored from key: Linux-python-unit-pip-48907ecc4b8e3067b5a1f92699ae6b465d0eed76
```
(https://github.com/wusatosi/Submitty/runs/2439815836?check_suite_focus=true#step:4:21)

However, this is still not fixed for the e2e testing, its still not cached:
```
Received 22 of 22 (100.0%), 0.0 MBs/sec
Cache Size: ~0 MB (22 B)
/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/a5699d4d-9fe6-41ae-aa82-8bf70aff7a2a/cache.tzst -P -C /home/runner/work/Submitty/Submitty
Cache restored successfully
Cache restored from key: Linux-e2e-pip-48907ecc4b8e3067b5a1f92699ae6b465d0eed76
```

I think this is because a lot of dependencies are stored in the root's pip directory due to some dependency installation invoked using sudo. I am sure cache does not operate on sudo.

shailpatels:
@wusatosi If its not possible to cache the pip packages that are installed using `sudo` I think that's fine for now, caching for the unit tests and part of the e2e will give us some speedup anyways. Right now the e2e testing requires python packages installed for every user (some users who get created dynamically in the later setup scripts) so this the workaround to make sure certain dependencies are accessible for everyone. 

My concern with using something other than the requirements file or some other lock file is that say you have a commit to update a python package that introduces some breaking change (which has happened a couple of times already hence the issue for a requirements.txt in our install process), in this case, we would not want to use a cache'd version and force install the new package to make sure our CI can test its compatibility for us.

Maybe it would be better to try and use a requirements.txt to solve this issue, we could start small and only focus on whats 
used in the `.setup/` area since thats whats important for the CI setup. I would suggest trying
```
pip3 freeze > requirements.txt
pipreqs ${SUBMITTY_REPOSITORY}/.setup/
```

I'm not a python expert so not sure if this will produce helpful results but seeing the resulting output by printing it out 
or uploading as an artifact and then save it as a file in the project and use that as our lockfile hashkey

wusatosi:
> @wusatosi If its not possible to cache the pip packages that are installed using `sudo` I think that's fine for now, caching for the unit tests and part of the e2e will give us some speedup anyways. Right now the e2e testing requires python packages installed for every user (some users who get created dynamically in the later setup scripts) so this the workaround to make sure certain dependencies are accessible for everyone.
> 
> My concern with using something other than the requirements file or some other lock file is that say you have a commit to update a python package that introduces some breaking change (which has happened a couple of times already hence the issue for a requirements.txt in our install process), in this case, we would not want to use a cache'd version and force install the new package to make sure our CI can test its compatibility for us.
> 
> Maybe it would be better to try and use a requirements.txt to solve this issue, we could start small and only focus on whats
> used in the `.setup/` area since thats whats important for the CI setup. I would suggest trying
> 
> ```
> pip3 freeze > requirements.txt
> pipreqs ${SUBMITTY_REPOSITORY}/.setup/
> ```
> 
> I'm not a python expert so not sure if this will produce helpful results but seeing the resulting output by printing it out
> or uploading as an artifact and then save it as a file in the project and use that as our lockfile hashkey

You got a point.
I guess I will extract `requirements.txt` for the setup process from another PR, its more complicated than the `pip3 freeze > requirements.txt` because I need to update the setup scripts as well, which I have had a look at already, the install scripts are mostly concentrated here:
https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/.setup/install_system.sh#L177-L221

I will submit the PR only for optimizing `python-unit` and `python-lint` then.

wusatosi:
Ready for review for python-lint and python-unit actions.

shailpatels (changes requested):
I think a similar issue occurs if you use the `gihub.job` as part of the restore key, so 
github will always look for `python-lint` or `python-unit` but we don't want to always 
use a cache if we are updating/changing python dependencies and there doesn't appear to be any good fix around that without a requirements.txt

My suggestion is to use something like the branch name (this can be taken from the git ref: 
https://stackoverflow.com/questions/60300169/how-to-get-branch-name-on-github-action)

so that on the first commit a fresh install of python dependencies are done but on further pushes it will utilize the cache. This can give us some speedup while we are looking for a longer term fix with the requirements.txt

.github/workflows/submitty_ci.yml
```diff
@@ -222,9 +222,9 @@ jobs:
             uses: actions/cache@v2
             with:
               path: path ~/.cache/pip
-              key: ${{ runner.os }}-py-pip-${{ hashFiles('**/requirements.txt') }}
+              key: ${{ runner.os }}-${{ github.job }}-pip-${{ hashFiles('**/requirements.txt') }}
```
MasterOdin, line 225: Shouldn't this be `github.sha`?

wusatosi: No, the key should look like `ubuntu-e2e-pip-HASH`, so that we can restore the cache by `ubuntu-e2e-pip-`.

MasterOdin: Sorry, I should have specified that I mean that should it be `${{ hashFiles('**/requirements.txt') }}` -> `${{ github.sha }}`, given that we don't yet have a requirements.txt file?

wusatosi: Yes, I've reverted it, this is intentional, I will update this field as I make the pull request which adds the `requirments.txt` which is almost ready.

shailpatels (approved):
hmm I'm not sure if caching is branch-local - otherwise how would new branches get a cache hit. According to this if a cache miss occurs at the branch level upstream and parent branches and are searched next, which can be seen if you branch off of this PR it'll re-use the same cache.

https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

However, it does look like pip install will try and grab the latest stable one and overwrite whatever's cached if its too old and violates being a requirement, or at least that's what my understanding of it's behavior is. So probably not a concern if we perform more cache hits than we should since we're looking for the latest python packages for almost every python dependency anyways.

So I think this should be a good stepping stone until we get a requirements.txt so we can manage our dependencies in CI like in JS or PHP.

PR #6350, opened 2021-04-25
[Bugfix:HelpQueue] Fix JS error on opening page as student

MasterOdin:
### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)

### What is the current behavior?

Fixes #6349 

On loading the Office Hours queue page as a student, there are two possible JS errors that are thrown. The first is the one mentioned in the above issue. The second is building a binding to the notification alert on/off button. Both of these errors are a result of the elements not existing for students.

### What is the new behavior?

Wraps the code with a twig if statement such that the code is only triggered for a grader. This isn't necessarily the cleanest solution, which would be breaking the JS into its own file and only loading what is needed for a student vs grader, but quickest to fix the bug, and have the page cleanly load.

PR #6357, opened 2021-04-26
[Feature:HelpQueue] Pause timer UI

### LINKED ISSUE #5996: Add counter/timer indicating how long a student has been paused
The intention for the queue pause feature was for a short period of time (use the restroom), it is not to allow students to hold a place in the queue for hours.

Let's add a timer to the pause feature, allowing TAs to manually remove from the queue a student who is abusing the pause feature (TAs can determine policy on how long is "too long").

A3Digital:
* [x ] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)

Students can pause with no indicator for how long they have been paused.
Closes #5996 

### What is the new behavior?
A timer has been added that keeps track of long a student has been paused since they most recently entered a queue. If a student pauses and then later un-pauses, their timer is not reset, rather this time is stored to prevent student from unpausing and re-pausing to reset their pause time. Both the instructor and student can see the student's pause time.


site/app/models/OfficeHoursQueueModel.php
```diff
@@ -205,6 +212,14 @@ public function getCurrentQueueTimeIn() {
         return $this->current_queue_state['time_in'];
     }
 
+    public function getCurrentTimePaused() {
```
MasterOdin, line 215:
```suggestion
    public function getCurrentTimePaused(): int {
```

site/app/models/OfficeHoursQueueModel.php
```diff
@@ -205,6 +212,14 @@ public function getCurrentQueueTimeIn() {
...
+        return $this->current_queue_state['time_paused'];
+    }
+
+    public function getCurrentTimePausedStart() {
```
MasterOdin, line 219: Not sure if this returns a string or `\DateTime`?

site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -33,11 +33,12 @@
         $("#queue_filter_" + current_queues[i]).prop("checked", true);
         filter_choices[current_queues[i]] = true;
       }
-      if (filter_choices[current_queues[i]]) {
-          document.getElementById('queue_filter_' + current_queues[i]).parentElement.classList.remove("fully-transparent");
+      var queue_filter = document.getElementById('queue_filter_' + current_queues[i]);
+      if (filter_choices[current_queues[i]] && queue_filter != null) {
```
MasterOdin, line 37: With #6350, we should not need to check on `queue_filter != null` here or in the else if anymore.

Additionally, I would recommend writing the these lines (if we were to keep the null check) like:
```
      if (queue_filter == null) {
        continue;
      }
      if (filter_choices[current_queues[i]]) {
          queue_filter.parentElement.classList.remove("fully-transparent");
      }
      else {
          queue_filter.parentElement.classList.add("fully-transparent");
      }
```
So that we only have the conditional check on `queue_filter` in one place instead of two, or just wrapping the if/else inside an if check of `if (queue_filter != null)`

site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -33,11 +33,12 @@
         $("#queue_filter_" + current_queues[i]).prop("checked", true);
         filter_choices[current_queues[i]] = true;
       }
-      if (filter_choices[current_queues[i]]) {
-          document.getElementById('queue_filter_' + current_queues[i]).parentElement.classList.remove("fully-transparent");
+      var queue_filter = document.getElementById('queue_filter_' + current_queues[i]);
```
MasterOdin, line 36: Try and use [`let`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) and [`const`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const) for new code. In this case, since the variable is never re-assigned, we can use `const`.

```suggestion
      const queue_filter = document.getElementById('queue_filter_' + current_queues[i]);
```

MasterOdin (changes requested):

migration/migrator/data/course_tables.sql
```diff
@@ -767,7 +767,9 @@ CREATE TABLE public.queue (
     contact_info text,
     last_time_in_queue timestamp with time zone,
     time_help_start timestamp with time zone,
-    paused boolean DEFAULT false NOT NULL
+    paused boolean DEFAULT false NOT NULL,
```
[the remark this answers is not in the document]
A3Digital: No, this column no longer necessary. I was already looking into removing it and I was also told to leave that as a 2nd PR as it would be cleaner.

site/app/models/OfficeHoursQueueModel.php
```diff
@@ -136,6 +137,12 @@ public function timeToISO($time) {
         return date_format(date_create($time), "c");
     }
 
+    public function intToStringTimePaused($int) {
+        $m = $int / 60 % 60;
+        $s = $int % 60;
+        return $m . ":" . ($s < 10 ? "0" : "") . $s;
+    }
```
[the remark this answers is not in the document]
A3Digital: Oh okay, I'll work on this.

site/app/models/OfficeHoursQueueModel.php
A3Digital: Both of those functions are returning a string. I see how that may be misleading but I was trying to follow the naming scheme of the functions around it.

MasterOdin: Ah, alright, so we can append `: string` here as the return type hint.

site/app/templates/officeHoursQueue/QueueFooter.twig
A3Digital: I took a look at #6350 and it appears to disable my timer. Simply put, I noticed that filterQueues() was being called on the QueueStatus page as well so I simply re-used the already present timer code for the student's side. I'll probably just need to separate some of the timer code from filterQueues() so it can work as before.

site/app/models/OfficeHoursQueueModel.php
A3Digital: This change actually had to become `: ?string` as `time_paused_start` could be `NULL` which we had discussed about on another change above.

site/app/templates/officeHoursQueue/CurrentQueue.twig
```diff
@@ -62,6 +63,13 @@
           {% else %}
             <td>{{ viewer.timeToHM(entry['time_in']) }}</td>
           {% endif %}
+
+          {% if entry['time_paused_start'] == NULL or entry['current_state'] == 'being_helped' %}
+            <td>---</td>
+          {% else %}
+            <td class="help_timer" help_time="{{ viewer.timeToISO(entry['time_paused_start']) }}"></td>
```
MasterOdin, line 70: In testing this through multiple pauses and unpauses, this column (and also the similar value on the user's UI) was always displaying "∞": https://www.loom.com/share/0c7d7c658c6e487bbc96ffb47f30c965

A3Digital: Interesting. My first guess would be some time zone issue that I may have missed, but I already did some testing for this so this may not be the case.

Given that I haven't been able to recreate this, would you mine commenting out this piece of code right here? This should give us some insight on what is going on.
https://github.com/Submitty/Submitty/blob/37914d84cb47cd3ca483661cf29a89820fb87172/site/app/templates/officeHoursQueue/QueueFooter.twig#L128-L130

Also, could you try helping the student and see if that timer is working properly?

MasterOdin: On getting back into EST, it now works as expected (it showed inf when I was in UTC last week). My guess is that there is indeed some timezone bug, but could not easily reproduce at the moment. Resolving this for now.

migration/migrator/migrations/course/20210320080101_add_pause_timer.py
```diff
@@ -0,0 +1,8 @@
+
+def up(config, database, semester, course):
+    database.execute('ALTER TABLE queue ADD COLUMN IF NOT EXISTS time_paused INTEGER;')
+    database.execute('ALTER TABLE queue ADD COLUMN IF NOT EXISTS time_paused_start TIMESTAMP WITH TIME ZONE;')
```
MasterOdin, line 4: On migrating, if there are any existing paused students, on attempting to go to the office hours queue page, they get an exception:

```
TypeError (Code: 0) thrown in /usr/local/submitty/site/app/models/OfficeHoursQueueModel.php (Line 214) by:
        return $this->current_queue_state['time_paused'];


Message:
Return value of app\models\OfficeHoursQueueModel::getCurrentTimePaused() must be of the type integer, null returned

Stack Trace:
#0 /usr/local/submitty/site/vendor/twig/twig/src/Extension/CoreExtension.php(1499): app\models\OfficeHoursQueueModel->getCurrentTimePaused()
```

We will need to set an initial value for these columns for any rows where `paused = true`. For this, it's fine to just set the `time_paused_start` to be the current now timestamp.

MasterOdin (approved):
One final thing, but otherwise this is fine. Holding off on merging until @bmcutler can review the UI wording.

migration/migrator/migrations/course/20210320080101_add_pause_timer.py
A3Digital: This should be good to go now.

site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -91,6 +91,13 @@
           Pause position in queue
         {% endif %}
       </button>
+      <p>Total time paused since you entered the queue:</p>
+      {% if viewer.getCurrentTimePausedStart() == NULL %}
+        <p>{{viewer.intToStringTimePaused(viewer.getCurrentTimePaused())}}</p>
+      {% else %}
+        <p id="student_timer" help_time="{{ viewer.timeToISO(viewer.getCurrentTimePausedStart()) }}"></p>
+      {% endif %}
+      <p>WARNING: If you remain paused for too long, you may be removed by the course instructors to prevent students from unnecessarily holding their positions in the queue.</p>
```
bmcutler, line 100: please change text to:

<p>WARNING: If you remain paused for more than a few minutes, you may be removed by the teaching staff.  Please refer to the course syllabus for office hour queue usage policies.</p>

bmcutler (changes requested):
It looks broken to me.  [tested on mac chrome & mac safari]. 
When the student is paused & not being helped,

I see infinity symbol for time paused by both the student:
Screen Shot 2021-05-19 at 11 58 13 PM

and instructor:
Screen Shot 2021-05-19 at 11 58 49 PM

Also, please tweak the text as noted.

migration/migrator/migrations/course/20210320080101_add_pause_timer.py
```diff
@@ -0,0 +1,13 @@
...
+utc_string = native_utc_dt.strftime("%Y-%m-%d %H:%M:%S") + " UTC";
+
+def up(config, database, semester, course):
+    database.execute('ALTER TABLE queue ADD COLUMN IF NOT EXISTS time_paused INTEGER;')
+    database.execute('ALTER TABLE queue ADD COLUMN IF NOT EXISTS time_paused_start TIMESTAMP WITH TIME ZONE;')
+    database.execute('UPDATE queue SET time_paused = 0;')
+    database.execute('UPDATE queue SET time_paused_start = \'' + utc_string + '\' WHERE paused = true;')
+
+def down(config, database, semester, course):
+    database.execute('ALTER TABLE queue DROP COLUMN IF EXISTS time_paused;')
```
bmcutler, line 12: Do we need to delete these columns on the down migration?  Why can't we leave them?

A3Digital: My apologies, I thought this was standard for down migrations. There is no actual reason to delete them so I'll edit that out.

PR #6358, opened 2021-04-26
[Bugfix:Submission] Remove dropped courses on calendar

zhangz22:
Currently the check of whether a student has dropped a course (so they do not appear on the homepage) is performed in `HomePageController` instead of the query. It violates the specification of `getCourseForUserId()` query (`Retrieves all unarchived/archived courses (and details) that are accessible by $user_id`) since the function actually returns courses that are not accessible by the user. This PR moves the check to the query and let `getCourseForUserId()` handles all the checks.

* [✅] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [✅] Tests for the changes have been added/updated (if possible)
* [✅] Documentation has been updated/added if relevant

The database query `getCourseForUserId()` is used by the homepage and the calendar to list the course an user is registered. It states that the function should "`retrieve all unarchived/archived courses (and details) that are accessible by $user_id`", but its return value actually contains courses that a student once registered but later dropped (therefore not accessible). The filter to check if the student dropped a course is performed in `HomePageController`. 

### What is the new behavior?
The check of whether the student dropped a course is moved to `DatabaseQueries::getCourseForUserId()`
This change should also modify the bug that the calendar is showing dropped courses

### DIFF
--- a/site/app/controllers/HomePageController.php
+++ b/site/app/controllers/HomePageController.php
@@ -31,10 +31,14 @@ class HomePageController extends AbstractController {
      * @Route("/api/courses", methods={"GET"})
      * @Route("/home/courses", methods={"GET"})
      *
+     * @param bool|string $as_instructor
      * @param string|null $user_id
      * @return MultiResponse
      */
-    public function getCourses($user_id = null) {
+    public function getCourses($user_id = null, $as_instructor = false) {
+        if ($as_instructor === 'true') {
+            $as_instructor = true;
+        }
 
         $user = $this->core->getUser();
         if (is_null($user_id) || $user->getAccessLevel() !== User::LEVEL_SUPERUSER) {
@@ -43,6 +47,14 @@ class HomePageController extends AbstractController {
 
         $unarchived_courses = $this->core->getQueries()->getCourseForUserId($user_id);
         $archived_courses = $this->core->getQueries()->getCourseForUserId($user_id, true);
+
+        if ($as_instructor) {
+            foreach (['archived_courses', 'unarchived_courses'] as $var) {
+                $$var = array_filter($$var, function (Course $course) use ($user_id) {
+                    return $this->core->getQueries()->checkIsInstructorInCourse($user_id, $course->getTitle(), $course->getSemester());
+                });
+            }
+        }
 
         $callback = function (Course $course) {
             return $course->getCourseInfo();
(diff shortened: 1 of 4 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -3651,6 +3651,12 @@ public function getCourseForUserId($user_id, bool $archived = false): array {
             $course->loadDisplayName();
             $return[] = $course;
         }
+
+        // Callback to filter out any courses a student has dropped so they do not appear on the homepage.
+        $return = array_filter($return, function (Course $course) use ($user_id) {
+            return $this->checkStudentActiveInCourse($user_id, $course->getTitle(), $course->getSemester());
```
MasterOdin, line 3657: In moving this into the query function, we would also want to then just build it straight into the query itself, getting a bit more optimized in the number of queries we run on a page. In this case, we could add a `WHERE` clause of the effect `registration_section IS NOT NULL OR user_group<>4` and be equivalent to what this filter is doing. It should be fine to add this where clause in combination with the change the on the other file for the add course page.

site/app/controllers/HomePageController.php
```diff
@@ -32,13 +32,9 @@ public function __construct(Core $core) {
...
      *
      * @param string|null $user_id
-     * @param bool|string $as_instructor
      * @return MultiResponse
      */
-    public function getCourses($user_id = null, $as_instructor = false) {
-        if ($as_instructor === 'true') {
-            $as_instructor = true;
-        }
+    public function getCourses($user_id = null) {
```
MasterOdin, line 37: When removing a function parameter, it's good practice to make sure to analyze why it was there in the first place. My preferred method to do this is to look at a combination of git blame / git history (which GitHub provides buttons for in the upper right when viewing a file). You should get led to the PR that added it, and then also how it's used ideally.

In this case, the `$as_instructor` flag was used here:
https://github.com/Submitty/Submitty/blob/47d4d9af9db157009b254c22e37a30be3f9cde6a/site/app/templates/CreateCourseForm.twig#L111-L113

where we only wanted to get courses the user was an instructor in to allow copying metadata from. Removal of the flag introduces a bug where that list may now contain courses the person is both an instructor or TA or student.

So yeah, let's revert this change.

MasterOdin (changes requested):
Thanks for submitting this patch. Some small tweaks requested on the code, and it should be good to merge. Let me know if you've got any questions.

MasterOdin (approved):

PR #6370, opened 2021-04-28
[Bugfix:TAGrading] Fix using touch to scroll panel selector

DefinitiveAbove:

Panel selector in TA grading cannot be scrolled on a touchscreen.

### What is the new behavior?
Panel selector in TA grading can be scrolled on a touchscreen.
Changed panel to depend on Popup.twig


site/public/css/electronic.css
```diff
@@ -1298,7 +1298,7 @@ tr.is_publish {
 .layout-option-cont {
     display: grid;
     grid-template-columns: repeat(auto-fill, 250px);
-    width: 83vw;
+    width: 78vw;
```
hdrodz97, line 1301: Is there any reason this changed?

hdrodz97 (changes requested):
Works pretty well; touch scrolling in the grading panel selector on Firefox and Edge on Windows as well as on Chrome on Ubuntu.

Just want clarification on one thing in `electronic.css` before I approve.

site/public/css/electronic.css
DefinitiveAbove: After changing the popup to extend `Popup.twig`, it changed it slightly so if the width was too long, it would result in the inner container being larger it needed to be (the table was the correct size, but there was a horizontal drag bar that went out a bit to nothing). I tested it on Brave (Chromium, Windows) and Safari (iOS).

eddiekrystowski (approved):
Works great, tested on iPad with Chrome, Firefox, Safari, and Edge.

PR #6386, opened 2021-05-05
[Bugfix:TAGrading] Notebook image width scaling

### LINKED ISSUE #6383: Notebook submitted image scaled for display
The display of image files submitted as part of a students notebook gradeable works well for modest sized images.  But when the student uploads a very large scanned image, the display is awkward.  The "popout display" of the file is scaled to a reasonable size -- and the user can click + to see the full resolution image.  But the "inline view" of these submitted files shows the full resolution initially, with scrollbars.  Let's apply a reasonable height and width scaling to the initial inline view.

DefinitiveAbove:

Images in TA grading under "Submissions and Results Browser" and "Notebook" panels (+student notebook) that are from openFrame (file submission) are full size; large images will result in scroll bars.

### What is the new behavior?
Fixes #6383 
Images now have a max width of the parent container (no more horizontal bars)

### DIFF
--- a/site/public/js/server.js
+++ b/site/public/js/server.js
@@ -1177,16 +1177,22 @@ function openFrame(url, id, filename, ta_grading_interpret=false) {
 }
 
 function resizeFrame(id, max_height = 500, force_height=-1) {
+    let img = undefined;
+    img = $("iframe#" + id).contents().find("img");
     if($("iframe#" + id).contents().find("html").length !== 0) {
         $("iframe#" + id).contents().find("html").css("height", "inherit");
-        var img = $("iframe#" + id).contents().find("img");
+        img = $("iframe#" + id).contents().find("img");
         if(img) {
+            img.removeAttr("width");
+            img.removeAttr("height");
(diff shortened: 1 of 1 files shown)


shailpatels (changes requested):
This fixes the issues of large submissions having horizontal scrollbars but the height is still unrestricted, I think adding a max-height would be good.

Also on firefox inline viewing a PDF (and other tall images I assume) looks a bit weird. 
image

The display of the renderer is only a couple hundred pixels tall which makes it hard to view the pdf inline

DefinitiveAbove:
> This fixes the issues of large submissions having horizontal scrollbars but the height is still unrestricted, I think adding a max-height would be good.
> 
> Also on firefox inline viewing a PDF (and other tall images I assume) looks a bit weird.
> image
> 
> The display of the renderer is only a couple hundred pixels tall which makes it hard to view the pdf inline

resizeFrame should limit the height of the iframe, see new commits for FireFox fixes

shailpatels (approved):

PR #6391, opened 2021-05-10
[Refactor:System] Cleanup CI dependency

wusatosi:

* [x] Tests for the changes have been added/updated (if possible) (tested on local fork, if the CI passes this should be fine)

There are dependencies installed during the e2e CI pipeline that are not used(`cryptography`, `websocket_client`, and `pyPdf`), and there are duplications of dependencies as well (`selenium` and `python-pam`).

These dependencies are special to CI as they are not mentioned elsewhere in installation scripts.

### What is the new behavior?
The unused/duplicated dependencies are removed. 

### Other information?

The passing of the CI test should indicate this pr is working.

### DIFF
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -237,6 +237,7 @@ jobs:
               pip3 install tzlocal
               pip3 install jsonschema
               pip3 install jsonref
+              pip3 install psycopg2
               pip3 install docker
               pip3 install pyyaml
               pip3 install selenium
@@ -260,6 +261,7 @@ jobs:
               sudo pip3 install watchdog
               sudo pip3 install python-dateutil
               sudo pip3 install psutil
+              sudo pip3 install psycopg2
               sudo pip3 install python-pam
               sudo pip3 install "sqlalchemy<1.4.0"
 
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -231,19 +231,15 @@ jobs:
               sudo python3 -m pip install --upgrade pip setuptools wheel
               pip3 -V
               sudo pip3 -V
-              pip3 install cryptography
               pip3 install python-pam
               pip3 install "sqlalchemy<1.4.0"
               pip3 install selenium
               pip3 install tzlocal
               pip3 install jsonschema
               pip3 install jsonref
-              pip3 install websocket_client
-              pip3 install psycopg2
               pip3 install docker
               pip3 install pyyaml
               pip3 install selenium
-              pip3 install websocket_client
               pip3 install python-pam
               pip3 install python-dateutil
               pip3 install paramiko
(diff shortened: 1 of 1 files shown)


wusatosi:
oof, this did not fail when it runs on my fork, I will into this later and reopen the pull request once I am certain about it.

wusatosi:
> There is some flakiness in the failed test, and so re-running the suite may get it to pass.

Yeah, this makes sense, this specific test case did succeed on my fork, here:
https://github.com/wusatosi/Submitty/runs/2544604503?check_suite_focus=true#step:25:63
and the related code does not seem to use any other external libraries, I will make another commit to force the action to be re-executed.

shailpatels (dismissed):

wusatosi:
wait nvm I didn't intend to push that to this branch, I will revert it

wusatosi:
> Changes look good and works as described

Can you merge the pr? I have another change based on this PR that I need to submit.

.github/workflows/submitty_ci.yml
```diff
@@ -231,32 +231,29 @@ jobs:
               sudo python3 -m pip install --upgrade pip setuptools wheel
               pip3 -V
               sudo pip3 -V
-              pip3 install cryptography
+              # Testing specific
+              pip3 install selenium
+              pip3 install psycopg2
```
MasterOdin, line 236: This is a central dependency of Submitty as it's the database driver we use anywhere where we use sqlalchemy.

.github/workflows/submitty_ci.yml
```diff
@@ -231,32 +231,29 @@ jobs:
...
+              # General dependency
               pip3 install python-pam
               pip3 install "sqlalchemy<1.4.0"
-              pip3 install selenium
               pip3 install tzlocal
               pip3 install jsonschema
               pip3 install jsonref
-              pip3 install websocket_client
```
MasterOdin, line 241: This dependency is necessary for

https://github.com/Submitty/Submitty/blob/656f588461e028f3f7e596d1f486cf68646d4cc6/tests/e2e/base_testcase.py#L12

I'm guessing the E2E still works as some other dependency requires it as a transitive dependency?

MasterOdin (changes requested):

.github/workflows/submitty_ci.yml
wusatosi: Okay, I've missed it here. 
Can you check for me if the pyPdf is used?

.github/workflows/submitty_ci.yml
wusatosi: Understood

.github/workflows/submitty_ci.yml
MasterOdin: It's not. That would be a typo and can be safely removed, as we use its successor `pyPDF2`.

wusatosi: okay, that's the only dependency that matters then, I will only include the removal of pyPDF2.

PR #6399, opened 2021-05-12
[Refactor:Developer] Refactor python dep

### LINKED ISSUE #3630: Add requirements.txt for python packages

### LINKED ISSUE #6352: Cache pip in e2e testing not working

wusatosi:

### What is the behavior?

Currently, pip dependencies are not managed, they are installed explicitly by the setup scripts (install_systems.sh).
This pr collets dependencies into a requirements.txt (fixes #3630).

Topics related:
- [x] Migrator (`migration/migration_requirements.txt`)
- [x] Autograder (`autograder/autograder_requirements.txt`)
- [x] Submitty_utils (`python_submitty_utils/submitty_utils_requirements.txt`)
- [x] Submitty_scripts (`.setup/pip/system_requirements.txt`)

Related works:
- [x] Update CI (fixes #6352)
- [x] Fixes versioning
- [x] Update dependa-bot config

### Other info:
`python-unit` gets split into three separate steps as this prevents requirements.txt of one to be missing dependencies that are covered by requirement.txt of others.


.github/workflows/submitty_ci.yml
```diff
@@ -133,58 +133,106 @@ jobs:
...
-        needs: [eslint, php-lint-static, php-unit, python-lint, python-unit]
+        needs: [
+          eslint,
+          php-lint-static,
+          php-unit,
+          python-lint,
+          python-unit-migration,
+          python-unit-auto-grader,
+          python-unit-submitty-utils
+        ]
```
MasterOdin, line 230:
```suggestion
        needs:
          - eslint
          - php-lint-static
          - php-unit
          - python-lint
          - python-unit-migration
          - python-unit-auto-grader
          - python-unit-submitty-utils
```

.setup/pip/system_requirements.txt
```diff
@@ -0,0 +1,37 @@
+# Extract from install_system.sh, latest update May 7th
+
+python-pam==1.8.4
+PyYAML==3.12
+psycopg2-binary==2.8.6
+sqlalchemy<1.4.0
```
MasterOdin, line 6:
```suggestion
sqlalchemy==1.3.24
```

apply this in the other files.

MasterOdin (changes requested):
For the files within python module directories (autograder, migration, python_submitty_utils), please rename the files as just `requirements.txt` to follow general python package convention.

.github/workflows/submitty_ci.yml
wusatosi: Wow, didn't know this before! This is neat! Thanks 👍

wusatosi:
> For the files within python module directories (autograder, migration, python_submitty_utils), please rename the files as just `requirements.txt` to follow general python package convention.

Updated!

MasterOdin (commented):
One thing that concerns me is that we have the same dependencies listed in multiple places, and that realistically, we want to keep all of these dependency versions pinned to the same thing given we're not using virtualenvs. As such, given that its spread out, it makes it easy (and this will happen under this) that these versions will drift in some way if we only rely on human intent to keep them in-sync, especially as we might add a new shared dependency. It would be trivial to write a simple script that validates the `requirements.txt` files against the `system_requirements.txt` file to make sure all lines match and include it as a test step within the workflow, but I fear that dependabot will open a PR per python module (e.g. autograder, migration) per dependency version bump meaning such a test step would fail for dependabot. These are more musing at the moment, as I've not been hit by any sort of good idea on solving this, and would hold off on merging and containing this work until we've automated something here to remove the human element.

wusatosi:
> One thing that concerns me is that we have the same dependencies listed in multiple places, and that realistically, we want to keep all of these dependency versions pinned to the same thing given we're not using virtualenvs. As such, given that its spread out, it makes it easy (and this will happen under this) that these versions will drift in some way if we only rely on human intent to keep them in-sync, especially as we might add a new shared dependency. It would be trivial to write a simple script that validates the `requirements.txt` files against the `system_requirements.txt` file to make sure all lines match and include it as a test step within the workflow, but I fear that dependabot will open a PR per python module (e.g. autograder, migration) per dependency version bump meaning such a test step would fail for dependabot. These are more musing at the moment, as I've not been hit by any sort of good idea on solving this, and would hold off on merging and containing this work until we've automated something here to remove the human element.

There are two ideas that come to mind:
1. Have like a shared core requirements.txt, requirements.txt supports nested dependency file, since we have this monorepo, developers pulling this repo will have that central requirments.txt anyway, not "that" big a hurdle to look around and resolve the file paths.
2. Make a script that runs weekly/ before release that checks this dependency coherency, have it create an issue if inconsistency crops up. GitHub Actions supports corn jobs.

wusatosi:
I think I need to add submitty_util to the requirements.txt as well

MasterOdin:
Having thought about it, given that we currently do not support any component being installed individually outside of Submitty, and only as part of the Submitty install process, I think having just one requirements.txt in `.setup` is probably preferable here. Within the requirements.txt, just need to put comments to group requirements (e.g. `# Required by migrator` or `# Required by migrator and autograder`. That leaves dependabot only having to deal with one file, and avoids this whole issue in the first place.

Then, if we want to install the minimum dependency set, we can easily make a small wrapper around pip and that one requirements.txt file to pull out the dependencies we need. What do you think?

wusatosi:
> Having thought about it, given that we currently do not support any component being installed individually outside of Submitty, and only as part of the Submitty install process, I think having just one requirements.txt in `.setup` is probably preferable here. Within the requirements.txt, just need to put comments to group requirements (e.g. `# Required by migrator` or `# Required by migrator and autograder`. That leaves dependabot only having to deal with one file, and avoids this whole issue in the first place.
> 
> Then, if we want to install the minimum dependency set, we can easily make a small wrapper around pip and that one requirements.txt file to pull out the dependencies we need. What do you think?

I personally am against the idea of using the same requirements.txt for all cases, as IMO migrator and auto-grader are two distinctive utils that are separated from the main site. It is the mono-repo project structure that we have (that I actually want to separate) keeping them intact. 

I am working on an alternative method, as you can tell from the commit log. I have developed a script (17a2cdb5a306f530e248dbd4b7241fc9583a6ee7) that tests the dependency coherency across the requirements.txt. I have experimented with the "open an [issue](https://github.com/wusatosi/Submitty/issues/8) if there's inconsistency" method on my [local branch](https://github.com/wusatosi/Submitty/tree/invalid_dependency). I am not satisfied with this approach as the feedback loop is delayed.

The best way to resolve this is to fail all python unit tests if there's dependency inconsistency and leave a comment to the dependa-bot's pr notifying this issue so that maintainers can manually update the dependency across the project. Once updated, the maintainers can get feedback on the dependency coherency by looking at the unit test pipelines. This is built on the reasonable belief that we don't merge pr that fails unit tests and the fact that dependency coherency tests are trivial to execute.

MasterOdin:
> The best way to resolve this is to fail all python unit tests if there's dependency inconsistency and leave a comment to the dependa-bot's pr notifying this issue so that maintainers can manually update the dependency across the project. Once updated, the maintainers can get feedback on the dependency coherency by looking at the unit test pipelines. This is built on the reasonable belief that we don't merge pr that fails unit tests and the fact that dependency coherency tests are trivial to execute.

This adds overhead then to dealing with dependencies that I'd like to avoid. I appreciate that in a true, best-practices monorepo, we would want the individualized requirements.txt files, given that this is all going into a shared global state right now (as opposed to everything running under virtualenvs which would be better practice), having them separate just does not make total sense combined with the unfortunate maintenance overhead of it.

I would add that we could in-theory add an automated step that if there is a discrepancy reconcile it automatically and pushes a new commit with that, but given the general population of people who contribute to Submitty (undergrads at RPI, some number who have no prior Git experience), I'd like to avoid anything that mutates a git branch that's not done by the author themselves.

wusatosi:
> > The best way to resolve this is to fail all python unit tests if there's dependency inconsistency and leave a comment to the dependa-bot's pr notifying this issue so that maintainers can manually update the dependency across the project. Once updated, the maintainers can get feedback on the dependency coherency by looking at the unit test pipelines. This is built on the reasonable belief that we don't merge pr that fails unit tests and the fact that dependency coherency tests are trivial to execute.
> 
> This adds overhead then to dealing with dependencies that I'd like to avoid. I appreciate that in a true, best-practices monorepo, we would want the individualized requirements.txt files, given that this is all going into a shared global state right now (as opposed to everything running under virtualenvs which would be better practice), having them separate just does not make total sense combined with the unfortunate maintenance overhead of it.
> 
> I would add that we could in-theory add an automated step that if there is a discrepancy reconcile it automatically and pushes a new commit with that, but given the general population of people who contribute to Submitty (undergrads at RPI, some number who have no prior Git experience), I'd like to avoid anything that mutates a git branch that's not done by the author themselves.

yeah that's the most hassal free way, I am going to far with this. I will refactor to that approach, do I need to leave documentations on which module uses which dependencies? and auto fixing doesn't go well with dependa-bot pr as well, I presume.

wusatosi:
@MasterOdin This pr is ready.

.setup/pip/system_requirements.txt
```diff
@@ -0,0 +1,50 @@
...
+pyzbar==0.1.8
+pdf2image==1.14.0
+opencv-python==4.5.1.48
+numpy==1.19.5
+
+#python libraries for OCR for digit recognition
+onnxruntime==1.7.0
+
+# Email catcher
+nullsmtpd==0.5.0
```
MasterOdin, line 50: This is a vagrant dependency, let's move it into `.setup/pip/vagrant_requirements.txt`. Can also bump the version to 0.5.1 as 0.5.0 has a bug preventing it from working properly.

.setup/pip/system_requirements.txt
```diff
@@ -0,0 +1,50 @@
...
+# For Lichen / Plagiarism Detection
+parso==0.8.2
+
+# Python3 implementation of python-clang bindings (may not work < 6.0)
+clang==11.0
+
+#python libraries for QR bulk upload
+pyzbar==0.1.8
+pdf2image==1.14.0
+opencv-python==4.5.1.48
```
MasterOdin, line 43: We require an older version (`opencv-python==3.4.9.33`) so that it'll install cleanly on older versions of Ubuntu (18.04).

MasterOdin (changes requested):
Two small requested tweaks, otherwise I think this is just about ready to merge.

wusatosi:
I am currently on trip, I will update this PR once I am back home.
&nbsp;
&nbsp;
------------------&nbsp;Original&nbsp;------------------
From: &nbsp;"Matthew ***@***.***&gt;;
Date: &nbsp;Sun, May 23, 2021 05:22 AM
To: ***@***.***&gt;; 
Cc: ***@***.***&gt;; ***@***.***&gt;; 
Subject: &nbsp;Re: [Submitty/Submitty] [Refactor:Developer] Refactor python dep (#6399)

&nbsp;



 
@MasterOdin requested changes on this pull request.
 
Two small requested tweaks, otherwise I think this is just about ready to merge.

 
In .setup/pip/system_requirements.txt:
 &gt; +parso==0.8.2 + +# Python3 implementation of python-clang bindings (may not work < 6.0) +clang==11.0 + +#python libraries for QR bulk upload +pyzbar==0.1.8 +pdf2image==1.14.0 +opencv-python==4.5.1.48 +numpy==1.19.5 + +#python libraries for OCR for digit recognition +onnxruntime==1.7.0 + +# Email catcher +nullsmtpd==0.5.0  
This is a vagrant dependency, let's move it into .setup/pip/vagrant_requirements.txt. Can also bump the version to 0.5.1 as 0.5.0 has a bug preventing it from working properly.
 
 
In .setup/pip/system_requirements.txt:
 &gt; +PyPDF2==1.26.0 +distro==1.5.0 +jsonschema==3.2.0 +jsonref==0.2 +docker==5.0.0 + +# For Lichen / Plagiarism Detection +parso==0.8.2 + +# Python3 implementation of python-clang bindings (may not work < 6.0) +clang==11.0 + +#python libraries for QR bulk upload +pyzbar==0.1.8 +pdf2image==1.14.0 +opencv-python==4.5.1.48  
We require an older version (opencv-python==3.4.9.33) so that it'll install cleanly on older versions of Ubuntu (18.04).

.setup/pip/system_requirements.txt
wusatosi: @MasterOdin I am not very familiar with this package, how should I write the note?
Is `To pin this package to version 3.4.9.33 allows a clean install on older versions of Ubuntu (18.04)` sufficient?

.setup/pip/system_requirements.txt
wusatosi: resolved in f1e219b

.setup/pip/system_requirements.txt
wusatosi: Resolved in d810dc9, noted with:
```
# The use of this older version of `opencv-python` is deliberate
# This allows a clean install on older versions of Ubuntu (18.04).
```

wusatosi:
@MasterOdin This pr is ready to be merged.

PR #6400, opened 2021-05-12
[UI/UX:System] Improve Breadcrumb UI

williamjallen:
This update changes the breadcrumb separator to a proper icon instead of the original ">" character.  This update also adds the .centered-icon class to the global CSS file since centering FA icons is a common task which I don't believe has been added elsewhere in the CSS thus far.

The ">" character is currently inserted between locations in the breadcrumb path.

### What is the new behavior?
Inserts a small angle between locations in the breadcrumb path.  The icon looks somewhat cleaner than a text-based navigation cue.

### DIFF
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -138,9 +138,9 @@
                     <div id="breadcrumbs">
                         {% for b in breadcrumbs %}
                             {% if loop.index0 > 0 %}
-                              <div class="centered-icon">
-                                <i class="fas fa-angle-right"></i>
-                              </div>
+                                <span class="centered-icon">
+                                    <i class="fas fa-angle-right"></i>
+                                </span>
                             {% endif %}
                             <div class="breadcrumb">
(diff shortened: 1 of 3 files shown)


shailpatels:
Some of the e2e tests are failing because they're looking for specific text in the breadcrumb element. You're probably going to have to update the sidebar tests to handle the changes

Looks like its line 71 in `Submitty/tests/e2e/test_sidebar.py`

I would recommend starting off by trying to run the e2e tests locally:
https://submitty.org/developer/testing/site_end_to_end_tests

and you can print debug what text is being collected from the breadcrumbs, my guess is you would just need to change the target from `span` to the `div` you added but not sure how it'll handle the icon. 

Post in slack if you run into trouble trying to run the tests locally or have other questions

shailpatels (approved):
Changes look good and works in dark mode, tested on mac firefox, chrome, safari

@bmcutler If the updates look good visually to you should be good to merge

bmcutler:
👍

PR #6401, opened 2021-05-12
[UI/UX:TAGrading] Fix spacing issue between panels

### LINKED ISSUE #6398: Margin for Right Panel is Too Large
**Describe the bug**
The horizontal drag bar is not centered between the two panes on the grading interface.  See below:
before

**Expected behavior**
The drag bar should be centered horizontally between the two panes.

williamjallen:
An extra margin is added when the TA grading interface is split into two panes.
before

### What is the new behavior?
Fixes the gap around the edges of the grading interface when split into two panes.
after

### Other information:
Fixes #6398

### DIFF
--- a/site/public/js/ta-grading.js
+++ b/site/public/js/ta-grading.js
@@ -137,13 +137,13 @@ $(function () {
   if(localStorage.getItem('notebook-setting-file-submission-expand') == 'true') {
     let notebookPanel = $('#notebook-view');
     if(notebookPanel.length != 0) {
-      let notebookItems = notebookPanel.find('.openAllFilesubmissions'); 
+      let notebookItems = notebookPanel.find('.openAllFilesubmissions');
       for(var i = 0; i < notebookItems.length; i++) {
         notebookItems[i].onclick();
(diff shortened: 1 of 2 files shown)


site/public/js/ta-grading.js
```diff
@@ -406,7 +406,7 @@ function adjustGradingPanelHeader () {
   }
   // From the complete content remove the height occupied by navigation-bar and panel-header element
   // 6 is used for adding some space in the bottom
```
sbelsk, line 408: Would this comment not need to be modified or removed?

williamjallen: You are correct.  I had the change on my branch and then it got deleted amongst some other Git issues.

sbelsk (approved):
I tested this on Windows/Chrome and Firefox, and through the instructor, TA, and student peer grading UI. I noticed that the extra margin added is only in layout modes with a taller left, which the PR fixes. The page now looks alright in all of the cases I tested.
Chrome:
image

Firefox:
image

PR #6402, opened 2021-05-12
[UI/UX:TAGrading] Adjust hover animation for drag bars

williamjallen:
The drag bars between panes in the TA grading interface currently blur when hovered over.  The animation is slow and looks strange.

### What is the new behavior?
This PR changes the animation to make the drag bars change in size.  The animation is faster and more subtle but also clearly indicates to the user that they are able to be dragged.

### DIFF
--- a/site/public/css/electronic.css
+++ b/site/public/css/electronic.css
@@ -475,7 +475,7 @@ progress::-webkit-progress-value {
     transition: all 0.1s;
 }
 
-.two-panel-drag-bar:hover:after{
+.two-panel-drag-bar:hover:after {
   width: 9px;
 }
 
--- a/site/public/css/electronic.css
+++ b/site/public/css/electronic.css
@@ -457,7 +457,7 @@ progress::-webkit-progress-value {
     background-color: var(--submitty-logo-blue);
 }
 
-.panel-item-section-drag-bar:hover:after {
+.panel-item-section-drag-bar:hover:after, .panel-item-section-drag-bar:active:after {
   height: 9px;
 }
 
@@ -475,7 +475,7 @@ progress::-webkit-progress-value {
     transition: all 0.1s;
 }
 
-.two-panel-drag-bar:hover:after {
+.two-panel-drag-bar:hover:after, .two-panel-drag-bar:active:after  {
   width: 9px;
 }
 
--- a/site/public/css/electronic.css
+++ b/site/public/css/electronic.css
@@ -453,8 +453,12 @@ progress::-webkit-progress-value {
     left: 50%;
     transform: translate(-50%, -50%);
     border-radius: 50px;
-    background: var(--submitty-logo-blue);
-    transition: all 0.5s;
+    transition: all 0.1s;
+    background-color: var(--submitty-logo-blue);
+}
+
+.panel-item-section-drag-bar:hover:after {
+  height: 9px;
 }
 
 .two-panel-drag-bar:after {
@@ -468,12 +472,11 @@ progress::-webkit-progress-value {
     transform: translate(-50%, -50%);
     border-radius: 50px;
     background: var(--submitty-logo-blue);
-    transition: all 0.5s;
+    transition: all 0.1s;
 }
 
-.two-panel-drag-bar:hover:after,
-.panel-item-section-drag-bar:hover:after {
-    filter: blur(3px);
+.two-panel-drag-bar:hover:after{
+  width: 9px;
 }
 
 #regrade_inner_info .inner-container {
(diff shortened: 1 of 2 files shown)


williamjallen:
I am working to remove the most recent commit.  Please hold off on the review until after I figure out how to revert the change and propagate the changes to GitHub.

MasterOdin:
You can do `git reset` to unstage the last commit (add `--hard` to revert all changes as well, or can do a `git restore .` afterwards to get the same effect) and then do a `git push --force` to update the branch to have one less commit. I think that gets what you want?

williamjallen:
Great, thanks!  That seems to have worked so this should be ready to be reviewed.

MasterOdin:
In testing this, on resizing the pane, the scrollbars returned to their normal size and did not grow again until after I moved my cursor away from the scrollbar and then back onto it. Do people find value in the scrollbars modifying size or blur or whatever at all? I feel like the mouse cursor changing is good enough?

williamjallen:
I think it could go either way.  The cursor changing along with the presence/shape of the drag bar is probably enough to clearly indicate the usage to users.  I thought that having the drag bars change in size slightly would make them seem more interactive but it would not be a significant loss to remove it altogether.  I plan to do some thinking about this and perhaps revisit a few UI things on the TA grading interface later this summer.  Amongst the things I have in mind are ways of further reducing the space the drag bars take up by making it only a line or a dot or similar.

A3Digital (changes requested):
Changes look good and work as described.
Just a leftover comment and formatting that could be addressed.

Tested on Chrome and Firefox.

MasterOdin:
Yeah, I think the animation as-is is fine, it's just making it consistent as you move. For visual reference on what I'm referring to:


When I begin the drag animation, the bar decreases in size and the cursor goes back to regular cursor. When I finish the drag, the cursor becomes the "resize" icon again, but the bar does not change size except for if I move my mouse. I'm not saying this has to be solved within this PR, but just that we would want to track that as a UI bug (even if it's a very low priority one).

A3Digital (approved):

A3Digital:
The new edits work. The last thing that could be edited would be to keep the cursor the same while scrolling, as Matthew had suggested.

MasterOdin:
@williamjallen experimenting with the cursor icon, I think the following patch gets us what we want where the cursor icon remains constant through the drag/drop operation:

```diff
diff --git a/site/public/js/resizable-panels.js b/site/public/js/resizable-panels.js
index d0f0f5a21..1a32ef141 100644
--- a/site/public/js/resizable-panels.js
+++ b/site/public/js/resizable-panels.js
@@ -19,6 +19,7 @@
 function initializeResizablePanels (panelSel, dragBarSel, isHorizontalResize= false, callback = null) {
     // Select all the DOM elements for dragging in two-panel-mode
     const panelEle = document.querySelector(panelSel);
+
     const panelCont = panelEle.parentElement;
     const dragbar = document.querySelector(dragBarSel);

@@ -37,6 +38,12 @@ function initializeResizablePanels (panelSel, dragBarSel, isHorizontalResize= fa
         panelHeight = panelEle.getBoundingClientRect().height;
         panelWidth = panelEle.getBoundingClientRect().width;

+        document.documentElement.style.setProperty(
+            'cursor',
+            dragBarSel === '.two-panel-drag-bar' ? 'col-resize' : 'ns-resize',
+            'important',
+        );
+
         // Attach the listeners to `document`
         document.addEventListener('mousemove', mouseMoveHandler);
         document.addEventListener('mouseup', mouseUpHandler);
@@ -46,8 +53,8 @@ function initializeResizablePanels (panelSel, dragBarSel, isHorizontalResize= fa

     const mouseUpHandler = () => {
         // remove the dragging CSS props to go back to initial styling
-        dragbar.style.removeProperty('cursor');
-        document.body.style.removeProperty('cursor');
+        document.documentElement.style.removeProperty('cursor');
+
         document.body.style.removeProperty('user-select');
         document.body.style.removeProperty('pointer-events');
         dragbar.style.removeProperty('filter');
@@ -75,8 +82,6 @@ function initializeResizablePanels (panelSel, dragBarSel, isHorizontalResize= fa
             panelEle.style.width = `${updateValue}%`;
         }

-        // consistent mouse pointer during dragging
-        document.body.style.cursor = 'col-resize';
         // Disable text selection when dragging
         document.body.style.userSelect = 'none';
         document.body.style.pointerEvents = 'none';
```

The principle change here is that we're setting the cursor property on the `<html>` element instead of `<body>` and that we're setting the priority to `important` so that it overrides stuff underneath it.

PR #6408, opened 2021-05-13
[UI/UX:InstructorUI] Fix row header on Manage Graders page

sbelsk:
The category row in the table does not span all the columns when there are no graders in a category.
image

### What is the new behavior?
The row now extends to the 8th column well.
image


MasterOdin (approved):

MasterOdin:
Thanks for fixing this!

PR #6409, opened 2021-05-13
[UI/UX:TAGrading] Add text-wrapping to manual input

### LINKED ISSUE #6063: Numeric/Text gradeable Formatting
If we have a numeric/text gradeable with 2 numeric columns and 1 text column, when displayed 
on a wide screen, the number columns & text columns are the same width.

If we write a moderate amount of text in the text box (e.g. a short sentence) the text overflows 
and is not visible all at once.  

It would be preferable to have the number columns narrower, and the text column wider. 
And if the text is lengthy, to have the text wrap and the row get taller, so all of it is visible.

Screen Shot 2020-12-09 at 10 49 29 AM

williamjallen:
Text entry boxes currently do not expand when text is entered. 

### What is the new behavior?
This update changes the TA grading interface to make the contents of text boxes wrap/expand when the number of characters exceeds the size of the box.
Untitled


shailpatels:
Please make your PR titles more descriptive see our guidelines for making a PR title here: 
https://submitty.org/developer/how_to_contribute#how-to-make-a-pull-request-pr-to-submitty

If you're PR closes an issue you can add the comment "fixes #PR_NUMBER" and on merging the PR it will auto close the issue. See here for the complete list of 'magic' keywords:

https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

MasterOdin:
Have you tested this on different browsers to make sure they look and act the same when using contenteditable? Which browsers did you test on?

williamjallen:
I tested it on macOS/Safari latest and macOS/Chrome latest and it looked the same.  `contenteditable` attributes have been supported by all major browsers for some time now and, as far as I know, are the best way of doing this task because you can style a `<span>` as you normally would while including the functionality of an `<input>` and eliminating the need for any Javascript.

shailpatels (changes requested):
It's generally not a good idea to try and create your own interactive HTML element out of something like span or div. Screen readers can work with `input`s but won't be able to handle your modified `span` correctly since it doesn't indicate something that can be clicked. 

This is why one of the accessibility tests is failing, you can manually validate here:
https://validator.w3.org/#validate_by_input+with_options

williamjallen:
From [this](url) Stack Overflow post which led me to [this](https://www.w3.org/TR/html-aria/#docconformance) page, it appears that elements with the `contenteditable` attribute are implicitly given the `aria-readonly="false"` attribute.  I haven't done much accessibility stuff but I was under the impression that if `aria-readonly` is set to `false`, it will be appear like an input to a screen reader.  

If `contenditable` is not ideal, would a solution using JS to change the size based on the number of lines be better?  Another option could be to put a hidden input beneath this span just for accessibility purposes where the contents of the span and the input are synced with JS.  I'm not sure what other options there would be.  I can't think of any ways to do that with CSS alone.

shailpatels:
The first link you post is broken and yes the `aria` tags assist screen readers and other accessibility tools it's better to use native HTML elements whenever possible since they receive better support.

It sounds like you should replace the input with a `textarea` with a default of 1 row and resizeable false to get multiline since they're designed for it.

williamjallen:
Ok, I have made the requested changes.  I suppose that this is the simple way out, which makes it more accessible, but I contend that it is bad UI design because the user can still only see one line of text unless they expand the textarea.  Expanding a textarea for each user in a several hundred person class suddenly becomes a major hassle.  I was hoping for a Google Forms-style behavior where the field expands to accommodate text as necessary, up to a point.

MasterOdin:
You can dynamically set the height of the textareas based on content to achieve what you're looking for if we stick to textareas.

For example, this code makes it so that when you focus and type in the textarea it will automatically expand to show all lines, and then once you leave, it will minimize itself back down to one line:
```html
        <script type="text/javascript">
            function adjustHeight(el){
                el.style.height = (el.scrollHeight > el.clientHeight) ? (el.scrollHeight)+"px" : "30px";
            }
            function minimizeHeight(el) {
                el.style.height = '30px';
            }
        </script>
        <td class="option-small-input">
            <textarea class="option-small-box cell-grade"
                  tabindex="0" id="cell-{{ index }}-{{ text_start + loop.index0 }}"
                  role="textbox" contenteditable
                  value="{{ component_grade.getComment() }}"
                  onclick="this.select()"
                  onfocus="adjustHeight(this);"
                  onfocusout="minimizeHeight(this);"
                  onkeyup="adjustHeight(this);"
                  data-id="{{ component.getId() }}"
                  aria-label="{{component.getTitle()}} value for {{grade.getSubmitter().getUser().getDisplayedFirstName()}} {{grade.getSubmitter().getUser().getDisplayedLastName()}}"
                  rows="1"
                  style="resize: vertical;"
            ></textarea>
        </td>
```

shailpatels (approved):

Tested on MacOs Safari, Firefox and Chrome

PR #6411, opened 2021-05-13
[UI/UX:InstructorUI] Off-Center Error Banner

### LINKED ISSUE #6397: Late Submission Notice Placement is Off-Center
**Describe the bug**
The "Late Submission" notice on the grading interface seems to have been placed rather arbitrarily.  It is not aligned with anything.  The notice also takes up a considerable amount of vertical space.

Current behavior:
Untitled

**Expected behavior**
An easy fix would be to center the notification.  The ideal fix would be to move it somewhere where there is existing whitespace so that it does not take up any additional vertical space.

HenrikLam:
Fixes #6397 
Keeps the error message banner aligned in the middle for all views options. Header panel is also aligned now.

Also made the banner automatically size to fit the content, and also adding padding to the banner.

image
image

### What is the new behavior?
image

Automatically sizing the banner to fit the content
image

### DIFF
--- a/site/app/views/grading/ElectronicGraderView.php
+++ b/site/app/views/grading/ElectronicGraderView.php
@@ -443,7 +443,6 @@ HTML;
         if ($gradeable->isPeerGrading() && $this->core->getUser()->getGroup() == User::GROUP_STUDENT) {
             $peer = true;
         }
-
         //Each table column is represented as an array with the following entries:
         // width => how wide the column should be on the page, <td width=X>
         // title => displayed title in the table header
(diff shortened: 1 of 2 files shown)


sbelsk (approved):
I tested this on my machine (Windows/Chrome, Firefox) and it looks good. Works for "cancelled submission" and "no submission" cases in all of the view modes as well.
canelled submission
no_submission

bmcutler (changes requested):
Can we also horizontally center the bar of panel labels/symbols?  Currently they are still left justified.

Also, please be sure to test the panel layouts that have the left panel(s) full height and the buttons/banner above only the right stack of panel(s)
Screen Shot 2021-05-14 at 7 00 13 PM

And also test when the screen is narrow and the panel buttons lose the text and turn into icons only:
Screen Shot 2021-05-14 at 7 00 23 PM
And I note that the cancelled submission label currently has a fixed width, so it goes off screen to the right.

HenrikLam:
> Can we also horizontally center the bar of panel labels/symbols? Currently they are still left justified.

Do you happen to know in which class the panel symbols are left justified?

williamjallen (approved):

bmcutler (approved):
looks nice :)

PR #6412, opened 2021-05-13
[Feature:InstructorUI] Row Numbers in SQL Toolbox

### LINKED ISSUE #6372: Print row number in SQL Toolbox table

For a table with many rows, it can be tough to know which number row you're looking at.

A new column should be added to the left of the table that shows the row number. This should be calculated as we insert the rows into the DOM.

cjreed121:

Currently the SQL Toolbox only displays the SELECT query result without any row number.
Closes #6372 

### What is the new behavior?
A new column was created in SQL Toolbox which contains the row number for each line.

### DIFF
--- a/site/public/js/sql-toolbox.js
+++ b/site/public/js/sql-toolbox.js
@@ -44,6 +44,9 @@ async function runSqlQuery() {
 
         const header = document.createElement('thead');
         const header_row = document.createElement('tr');
+        const cell = document.createElement('td');
+        cell.innerText = 'Row #';
+        header_row.appendChild(cell);
         Object.keys(data[0]).forEach((col) => {
             const cell = document.createElement('td');
             cell.innerText = col;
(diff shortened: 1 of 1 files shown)


site/public/js/sql-toolbox.js
```diff
@@ -52,8 +55,13 @@ async function runSqlQuery() {
         header.appendChild(header_row);
         table.appendChild(header);
         const body = document.createElement('tbody');
+        let rownum = 1;
         data.forEach((row) => {
```
MasterOdin, line 59: You can amend this to be:

```suggestion
        data.forEach((row, idx) => {
```

and then on line 62 you could do `cell.textContent = idx + 1;` and not have and track the `rownum` variable.

MasterOdin (changes requested):

MasterOdin (approved):
Note: you do not need to re-request a review from me on making changes. I get an email whenever a commit or comment is made to a PR that I've reviewed.

PR #6413, opened 2021-05-13
[UI/UX:Forum] Fix forum date error

### LINKED ISSUE #6075: Editing post on Discussion Forum Displays "Invalid Date"
**Describe the bug**
When clicking the "Edit Post" button on the discussion forum, the popup displays "Editing a post by [User] on Invalid Date at Invalid Date"

**Expected behavior**
The Invalid Dates should be replaced with the appropriate dates if possible, which it usually should be.

**To Reproduce**
1. Create a post
2. Open the edit window
3. See error

**Configuration**
 - OS: OSX
 - Browser: Safari
 - Viewing as a limited access grader (probably not relevant)

**Screenshots**
image

**Additional context**
Also happens on my IOS phone in Chrome

williamjallen:
When editing posts on the discussion forum, the date format is incorrect.

### What is the new behavior?
This change fixes the dates on the discussion forum by correcting the date format to be a proper Javascript date.
Untitled

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -754,10 +754,11 @@ function showEditPostForm(post_id, thread_id, shouldEditThread, render_markdown,
             var anon = json.anon;
             var change_anon = json.change_anon;
             var user_id = escapeSpecialChars(json.user);
+            json.post_time = json.post_time.replace(" ", "T");
             var time = Date.parse(json.post_time);
             if(!time) {
                 // Timezone suffix ":00" might be missing
-                time = Date.parse(json.post_time+":00");
+                time = Date.parse(json.post_time+"00");
             }
             time = new Date(time);
             var categories_ids = json.categories_ids;


site/public/js/forum.js
```diff
@@ -754,10 +754,11 @@ function showEditPostForm(post_id, thread_id, shouldEditThread, render_markdown,
             var anon = json.anon;
             var change_anon = json.change_anon;
             var user_id = escapeSpecialChars(json.user);
+            json.post_time = json.post_time.replace(" ", "T");
```
MasterOdin, line 757: Where is this `post_time` coming from and can we update whatever format it is there to make `Date.parse` happy?

williamjallen: I'm not sure where it's coming from.  I tried to track it down but there isn't much to work with.  I'm not sure which file is the backend for this.  The only issue in the date is that there is a space between the date and the timestamp.  JS expects a T in front of the timestamp so I simply replaced the space with a T.  It could be a difference in how PHP and JS process dates.  The general format is exactly the same except for that.  PHP also adds a higher precision for the seconds than JS but JS seems to handle it just fine.

shailpatels:
The data comes from the backend function `public function getEditPostContent()` in `site/app/controllers/forum/ForumController.php`

The timestamp is just getting pulled from the database
```php
  $output['post_time'] = $result['timestamp'];                         
```

This error is MacOS safari specific which wants timestamps in  ISO 8601 format which requires a `T` separator within it, however other browsers Firefox and Chrome accept a space. This problem has appeared in multiple places and a long term solution would be for the backend to only return ISO 8601 strict timestamps so we don't have to deal  with front end parsing in different places

site/public/js/forum.js
MasterOdin: To help track this down, we can see that this is firing on making a request to `buildCourseUrl(['forum', 'posts', 'get']);` where that gives us a url path of `{_semester}/{_course}/forum/posts/get` within the code. Searching the codebase for that route leads us to the function [`getEditPostContent`](https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/site/app/controllers/forum/ForumController.php#L1079-L1104). Within that, we see then that the returned JSON object is coming from the [`getPost`](https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/site/app/libraries/database/DatabaseQueries.php#L724) database method, where timestamp is just the string as it comes out of the database. Scanning the database, the DB query is used in a handful of places, while this route is only used here, so we can safely do that replace within the controller method. As part of this, we can also properly amend the the `00` to it there as well. The easiest way would be to parse `$result['timestamp'];` into a `\DateTime` and then format it to something that can be directly used within the JS.

We want to avoid doing these sort of manipulations in JS as it's a weak link between PHP / JS and so prone to possible break down the road when a developer inadvertently changes the upstream function, as well as that we're trying to work around vendor quirks here which is a recipe for disaster.

PR #6414, opened 2021-05-13
[Feature:InstructorUI] CSV Export of SQL Toolbox Data

### LINKED ISSUE #6374: Allow CSV export of SQL Toolbox data

Right now, to get information out of the SQL toolbox, an instructor can just highlight and copy/paste the information from the HTML into some other source. However, on doing this, the information lacks any meaningful formatting when trying to paste it into a text document or excel.

Submitty should have a button that when pressed would export the data contained in the table as a CSV file. Ideally, this would all be done via JavaScript.

cjreed121:

Currently you can only view the data returned by the query. It can be difficult to extract this data to a different program.
Closes #6374 

### What is the new behavior?
There is now a Download CSV button which will download a CSV file of the data just ran by the query.
image

### DIFF
--- a/site/public/js/sql-toolbox.js
+++ b/site/public/js/sql-toolbox.js
@@ -81,14 +81,14 @@ async function downloadSqlResult(){
     let j;
     let csv = '';
     //Add headers to CSV string
-    for (i = 0; i < results.children.item(0).children.item(0).children.length; i++){
+    for (i = 1; i < results.children.item(0).children.item(0).children.length; i++){
         csv += `"${results.children.item(0).children.item(0).children.item(i).textContent}",`;
     }
     csv += '\n';
 
     //Add data to CSV string
     for (i = 0; i < results.children.item(1).children.length; i++){
-        for (j = 0; j < results.children.item(1).children.item(i).children.length; j++){
+        for (j = 1; j < results.children.item(1).children.item(i).children.length; j++){
             csv += `"${results.children.item(1).children.item(i).children.item(j).textContent.replaceAll('"', '""')}",`;
         }
         csv += '\n';
--- a/site/public/js/sql-toolbox.js
+++ b/site/public/js/sql-toolbox.js
@@ -82,14 +82,14 @@ async function downloadSqlResult(){
     let csv = '';
     //Add headers to CSV string
     for (i = 0; i < results.children.item(0).children.item(0).children.length; i++){
-        csv += `${results.children.item(0).children.item(0).children.item(i).textContent},`;
+        csv += `"${results.children.item(0).children.item(0).children.item(i).textContent}",`;
     }
     csv += '\n';
 
     //Add data to CSV string
     for (i = 0; i < results.children.item(1).children.length; i++){
         for (j = 0; j < results.children.item(1).children.item(i).children.length; j++){
-            csv += `${results.children.item(1).children.item(i).children.item(j).textContent},`;
+            csv += `"${results.children.item(1).children.item(i).children.item(j).textContent.replaceAll('"', '\'')}",`;
         }
         csv += '\n';
     }
--- a/site/public/js/sql-toolbox.js
+++ b/site/public/js/sql-toolbox.js
@@ -1,5 +1,6 @@
 /* global buildCourseUrl, csrfToken */
 /* exported runSqlQuery */
+/* exported downloadSqlResult */
 
 async function runSqlQuery() {
     document.getElementById('query-results').style.display = 'block';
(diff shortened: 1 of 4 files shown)


MasterOdin (commented):
You will need to consider how to handle cell values that might have newlines, commas, and quotation marks if rolling your own CSV download method on this.

MasterOdin:
Additionally, with merging #6412, you will want to account for and disregard the first column in every row.

site/public/js/sql-toolbox.js
```diff
@@ -73,3 +74,33 @@ async function runSqlQuery() {
...
+    let csv = '';
+    //Add headers to CSV string
+    for (i = 1; i < results.children.item(0).children.item(0).children.length; i++){
+        csv += `"${results.children.item(0).children.item(0).children.item(i).textContent}",`;
+    }
+    csv += '\n';
+
+    //Add data to CSV string
+    for (i = 0; i < results.children.item(1).children.length; i++){
+        for (j = 1; j < results.children.item(1).children.item(i).children.length; j++){
```
wusatosi, line 91: Chain may be too long

site/public/js/sql-toolbox.js
```diff
@@ -73,3 +74,33 @@ async function runSqlQuery() {
...
+
+async function downloadSqlResult(){
+    const results = document.getElementById('query-results');
+    let i;
+    let j;
+    let csv = '';
+    //Add headers to CSV string
+    for (i = 1; i < results.children.item(0).children.item(0).children.length; i++){
+        csv += `"${results.children.item(0).children.item(0).children.item(i).textContent}",`;
+    }
```
wusatosi, line 86: const header_elements = results.children.item(0).children.item(0).children;
    for (i = 1; i < header_elements.length; i++){
        csv += `"${header_elements.item(i).textContent}",`;
    }

site/public/js/sql-toolbox.js
```diff
@@ -73,3 +74,33 @@ async function runSqlQuery() {
         console.error(exc);
     }
 }
+
+async function downloadSqlResult(){
+    const results = document.getElementById('query-results');
+    let i;
+    let j;
```
wusatosi, line 81: Move these two declarations into the for loop, putting it outside is bit confusing IMO.

MasterOdin:
@cjreed121 Sorry about this, as I forgot this was opened when I merged #6421. With #6421 merged, you will need to migrate your change to fit the mjs model now employed on this page. Namely, this means that you'll need to move your function into this new file (`site/public/mjs/sql-toolbox.js`) and that you'll have to setup the button binding to be within the JS file instead of the twig template.

cjreed121:
@MasterOdin Thanks for heads up about #6421. The js-unit tests seem to be failing due to the tests not including my new button. I can add in the new button and make the tests pass but how would I go about actually testing my downloadSqlResult function? I see your tests look at the document but how would I test a file that gets downloaded?

MasterOdin:
Alright, so to immediately fix the failing test, you would need to append a new line here:

https://github.com/Submitty/Submitty/blob/ed323b5ff9d11806ab99cf1d10ed52283fe2e021/site/tests/mjs/sql-toolbox.spec.js#L12-L19

with `<button id='download-sql-btn' class="btn btn-primary">Download CSV</button>` underneath `<div id='run-sql-btn'>Submit</div>` (or wherever really).

The way the test suite here works is we set-up a custom mocked `document.body` that then all testcases use it.

To test your code, I would suggest that you split https://github.com/Submitty/Submitty/pull/6414/files#diff-577fca2b041809a35c7ba55d4755affad55d77cd0d95db9f2546d71d1b14a981R78-R96 into its own function where it accepts a HTML id as its parameter and then returns the CSV string.

To test, you would then do something like:

```
test('csv generation', () => {
	document.body.innerHTML = `
		<table id='table-foo'>
			<tr>
				<td>#</td>
				<td>column_1</td>
				<td>column_2</td>
			</tr>
			<tr>
				<td>1</td>
				<td>foo</td>
				<td>bar</td>
			</tr>
			<tr>
				<td>2</td>
				<td>"f'f"o'o"</td>
				<td>baz</td>
			</tr>
		</table>
	`;

	expect(generateCsv('table-foo')).toEqual(...);
});
```

For now, don't worry about testing the download code, I can handle adding tests for that given your relative newness to testing in general and I suspect it'll require some amount of goofy mocking.

For testing at large in JS in Submitty, https://jestjs.io/docs/getting-started is a good starting place.

williamjallen:
I have one comment about the UI:
You added a message about needing to run a query before you can download.  Instead of adding text (which they may not read) to tell users not to click the button which looks so inviting, perhaps you could disable the button until there are results to be downloaded.  Another option would be to hide the button altogether.  In addition, there is a lot of space on the right and two buttons on the left which makes the left look a little cluttered.  This is just my slight opinion here but I think the download button may look better if it is either on the right or somewhere in the query results section.

This PR may not be the best place to add it but I noticed that the column headers are not properly bolded on the output table.  That should at least be tracked so it can be resolved at some point.

site/public/mjs/sql-toolbox.js
```diff
@@ -25,6 +25,7 @@ export async function runSqlQuery() {
         if (json.status !== 'success') {
             error_mesage.textContent = json.message;
             error.style.display = 'block';
+            document.getElementById('download-sql-btn').disabled = true;
```
MasterOdin, line 28: Let's move this line to the top of the function. We should disable this button for so long as it takes us to get and generate the results to avoid any possible run conditions with downloading (i.e. if I hit "run query" and while the network request is in progress, I hit "download csv", what is going to get downloaded? The original dataset or possibly the new one as it loads in?).

Moving this to the top would also allow getting rid of the duplicate of this line in the exception catch as we know it's `true` except for when we get to the very bottom of this try.

williamjallen (approved):

PR #6416, opened 2021-05-13
[Bugfix:SubminiPolls] polls dir for new course/install

bmcutler:
PR #6180 included a migration to add the uploads/polls directory for existing courses, but missed adding this directory for new courses (and new installations).

### DIFF
--- a/migration/migrator/migrations/course/20210513192238_polls_directory_bugfix.py
+++ b/migration/migrator/migrations/course/20210513192238_polls_directory_bugfix.py
@@ -2,7 +2,7 @@ import shutil
 from pathlib import Path
 
 
-def up(config):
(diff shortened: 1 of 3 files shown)


bmcutler:
@sbelsk Tried to add you as reviewer, but I see I forgot to add you to the org first!  Whoops :)

sbelsk (approved):
I originally came across this problem (the missing directory) by trying to fix a bug where images attached to polls are not being saved properly. I reproduced the issue both with one of the courses on a new installation of Submitty, and with a new course I created, tested it with this PR, and the issue seemed to be solved (as well as the other bug! 😄). I poked around by printing out contents of directories to ensure that the `uploads/polls` directory exists.

Issue #6404, opened 2021-05-13
Migrate JS files to ts/mjs


MasterOdin:
**What problem are you trying to solve with Submitty**

Currently, all our JS usage within Submitty is using old style JS includes in the page, and we have our scope leak all over the place, having to also manually annotate files with a `/* exported */` and `/* global */` directives for eslint to understand what might be dropping through scripts.

**Describe the way you'd like to solve this problem**

A better way to handle this mess would be to make use of the module system that all modern browsers now support (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules), and is inline with how most people write JS these days. The JS files would get loaded as modules in the HTML, and then we have regular import / export statements within the files, giving us understanding of where functions / classes are coming from, while also allowing the IDE to provide more meaningful intellisense.

A side effect of this is that we can leverage the import / export syntax to also write proper unit testing for the JS files without resorting to any ugly shims or scope leakage in the test suite.

**Additional context**

This issue isn't necessarily to track the full scope of effort to do this project, but to gain feedback from developers on if we should do (or should not) do this, any potential pitfalls, etc. The actual migration process would most likely be more drawn out and take a while, doing it gradually overtime as it would require rewriting the hooks of the JS into the DOM that currently exist in a lot of places.

shailpatels:
I support moving to modules

I think a possible issue would be with the `addInternalJS` functions, would the versioned file be able to be used in the JS `import` statement?

wusatosi:
I see all the code related to #6421 has been moved to the TS folder, so is `ts` modules favored over `mjs` modules, and `mjs` modules have been deprecated? Do we need to update this issue? @MasterOdin

MasterOdin:
The basis for this was accomplished in #7285 with documentation in Submitty/submitty.github.io#392. Now it's just a matter of slowly transitioning the code over, but we don't need a meta issue to track that.

PR #6421, opened 2021-05-14
[Refactor:InstructorUI] Convert sql toolbox to use mjs

MasterOdin:

All of our JS usage is done as old-style scripts, where they're included in the page and populate the global scope, and being called directly. While this works, it is not easily testable and includes the necessity of inserting eslint directives in each file to declare globals and what is exported, which is a frustrating experience compared to "modern" JS.

### What is the new behavior?

This provides a proof-of-concept for #6404, using the new SQL Toolbox page as the testbed for it. This page was converted to use JS modules. As part of this, I also include tests for these new modules to demonstrate how that would work as well. These module files are stored in the `mjs` directory (I had wanted to use the `.mjs` extension as well, but TypeScript / VSCode does not support intellicode for it), and are loaded at the end of the page using the `type="module"` directive.

### Other information?

I've not yet finished writing the tests for sql-toolbox.js, so this should not be merged until that's done, but this PR is done enough to get review comments.

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -28,12 +28,19 @@
     "twig": "1.15.4"
   },
   "devDependencies": {
+    "@babel/core": "^7.14.2",
+    "@babel/preset-env": "^7.14.2",
+    "babel-jest": "^26.6.3",
     "cypress": "^7.3.0",
     "eslint": "^7.26.0",
-    "eslint-plugin-cypress": "^2.11.3"
+    "eslint-plugin-cypress": "^2.11.3",
+    "eslint-plugin-jest": "^24.3.6",
+    "jest": "^26.6.3",
+    "regenerator-runtime": "^0.13.7"
   },
   "scripts": {
-    "lint": "eslint \"public/js/**/*.js\" \"cypress/**/*.js\"",
-    "lint:fix": "eslint \"public/js/**/*.js\" \"cypress/**/*.js\" --fix"
+    "lint": "eslint \"public/js/**/*.js\" \"cypress/**/*.js\" \"public/mjs/**/*.js\" \"tests/mjs/**/*.js\"",
+    "lint:fix": "eslint \"public/js/**/*.js\" \"cypress/**/*.js\" \"public/mjs/**/*.js\" \"tests/mjs/**/*.js\" --fix",
+    "test": "jest"
   }
 }
--- a/site/public/mjs/sql-toolbox.js
+++ b/site/public/mjs/sql-toolbox.js
@@ -1,11 +1,11 @@
-/* global buildCourseUrl, csrfToken */
-/* exported runSqlQuery */
 
-async function runSqlQuery() {
+import { buildCourseUrl, getCsrfToken } from './server.js';
+
+export async function runSqlQuery() {
     document.getElementById('query-results').style.display = 'block';
 
     const form_data = new FormData();
-    form_data.append('csrf_token', csrfToken);
+    form_data.append('csrf_token', getCsrfToken());
     form_data.append('sql', document.querySelector('[name="sql"]').value);
 
     try {
@@ -24,7 +24,7 @@ async function runSqlQuery() {
         table.innerHTML = '';
 
         if (json.status !== 'success') {
-            error_mesage.innerText = json.message;
+            error_mesage.textContent = json.message;
             error.style.display = 'block';
             return;
         }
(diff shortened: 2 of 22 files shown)


site/package.json
```diff
@@ -28,12 +28,19 @@
     "twig": "1.15.4"
   },
   "devDependencies": {
+    "@babel/core": "^7.14.2",
```
MasterOdin, line 31: babel is added so that we can use the ES6 module directives in the test files. We could forgo it, and then use the old-style `const { buildCourseUrl } = require('../')` type syntax, but I wanted to keep the test files written in a similar fashion as the code it's testing for developer simplicity.

site/public/mjs/sql-toolbox.js
```diff
@@ -24,7 +24,7 @@ async function runSqlQuery() {
         table.innerHTML = '';
 
         if (json.status !== 'success') {
-            error_mesage.innerText = json.message;
+            error_mesage.textContent = json.message;
```
MasterOdin, line 27: JSDom (the HTML mocking library that jest uses) does not support `innerText`. While we could mock in support for it, better to just move to using `textContent` which is supported, and as we go rewrite any usage of `innerText` to `textContent` as well.

site/public/mjs/sql-toolbox.js
```diff
@@ -73,3 +73,9 @@ async function runSqlQuery() {
         console.error(exc);
     }
 }
+
+export function init() {
+    document.getElementById('run-sql-btn').addEventListener('click', () => runSqlQuery());
```
MasterOdin, line 78: JS modules are not able to be used directly within HTML like scripts are, so we have to manually build our event listener bindings. Not a huge deal, but something to be aware of in the transition.

shailpatels:
All these changes look really good to me and we should see  a lot of good improvements moving to modules,

Since we're only using babel to allow Jest to work with commonJs is it possible to use [esm](https://www.npmjs.com/package/esm) if so it could be a possible alternative since we're not actually using babel for building JS files for us and could have less setup

MasterOdin:
> Since we're only using babel to allow Jest to work with commonJs is it possible to use esm if so it could be a possible alternative since we're not actually using babel for building JS files for us and could have less setup

esm is not supported with jest without doing some really weird and ugly hacks that I'm not going to do. Jest itself has [experimental support for ES6 modules](https://jestjs.io/docs/ecmascript-modules), but it was throwing weird errors on trying to use it and debugging it was going to take more time than was worth it.

It should be noted that currently that a transitive dependency of jest itself ships with `@babel/core` and `babel-jest` anyway so it's not like we're actually bringing in extra stuff here. I'm keen on using jest though as it is the principal test framework I use at work and is one of the most (if not the most) popular testing framework at the moment.

MasterOdin:
Alright, I think this is ready for review and merge. I've updated the test suite so that both files are completely covered (not sure why codecov does not say that though). I'm happy enough with how this looks and works for now, and can refine it as it gets more usage by other people and other pages.

shailpatels:
Ah ok that makes sense, I didn't realize esm wasn't compatible with jest and yeah jest is the only JS testing framework I've used and it seems to be the most popular so I agree with using it too.

site/public/mjs/sql-toolbox.js
```diff
@@ -24,7 +24,7 @@ async function runSqlQuery() {
         table.innerHTML = '';
 
         if (json.status !== 'success') {
```
wusatosi, line 26: It's weird how codecov reporting a missing branch here.

site/public/mjs/sql-toolbox.js
```diff
@@ -36,7 +36,7 @@ async function runSqlQuery() {
         if (data.length === 0) {
```
wusatosi, line 36: Here as well, it's reporting missing branches, but as far as I can tell both branches are covered.

site/public/mjs/sql-toolbox.js
```diff
@@ -73,3 +73,9 @@ async function runSqlQuery() {
         console.error(exc);
```
wusatosi, line 73: It would be better to notify the user (even with `alert(exc)`)  as network/serialization failure is caught here.

shailpatels (approved):
I think this is ready to be merged, running jest locally after an npm install works as expected and this module should be a good example for future ones

site/public/mjs/sql-toolbox.js
shailpatels: Looks at the code coverage report locally it shows both branches coverage so maybe its a codecov specific issue

MasterOdin:
> I think this is ready to be merged, running jest locally after an npm install works as expected and this module should be a good example for future ones

Yeah, I plan to report to Codecov support on figuring this out on why the discrepancy between the two with it merged to master.

Issue #6390, opened 2021-05-10
Possible ways to speedup e2e testing?


wusatosi:
Currently, e2e testing runs for about an hour on each commit. This will worsen as we introduce more and more features to Submitty. I've had a look around to see all the logs. This is what I've found:

Top three actions ordered by time spent, cumulative to ~50 mins of each run.
| Order | Name                                 | Time spent | Related Scripts                    |
|-------|--------------------------------------|------------|------------------------------------|
| 1     | Run integration tests                | ~30min     | test cases under `tests/integrationTests/tests/`  driver:  `tests/integrationTests/lib.py`   |
| 2     | Set up sample course                 | ~15min     | `.setup/bin/setup_sample_courses.py` |
| 3     | Run selenium e2e tests with pam auth | ~10min     | test cases under `tests/e2e/*.py`    (using built in unittest driver)      |

# Proposing solution: (high-level overview) 
## Parallel testing
All the tests currently run sequentially, observing from the logs, there is no parallel testing implemented yet. Since testing are (suppose to be) completely isolated, this should be fairly easy to parallelize. 

### Integration testing:
For integration testing, we have a custom wrote driver. The main loop is implemented here:
https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/tests/integrationTests/lib.py#L92-L152
We could try to parallelize this process using modules like [multiprocessing](https://docs.python.org/3/library/multiprocessing.html). Essentially changing the big for-loop to a multiprocess for-loop.

### Selenium e2e tests
We currently use the built-in unittest module as the test suit driver. We could attempt to adopt other testing framework extensions that implement case-level parallelization. A possible candidate would be [testtools](https://pypi.org/project/testtools/) package described [on this StackOverflow post](https://stackoverflow.com/a/17059844), which can parallelize test cases that are based on `unittest.TestCase` that we already satisfy.
https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/tests/e2e/base_testcase.py#L28

## Optimize setup scripts:
Currently, the setup process requires ~20min to execute. The majority of time is spent on `Set up sample course`. This is a tricky one to improve. I have not run a good timing experiment to see what's spending the majority amount of time, the related file is [here](https://github.com/Submitty/Submitty/blob/656f588461e028f3f7e596d1f486cf68646d4cc6/.setup/bin/setup_sample_courses.py).

[...]

PR #6422, opened 2021-05-14
[Testing:Developer] Parallelize integration test suite

wusatosi:

Currently the integration testing driver executes every test module sequentially, spending ~30mins on each run 

### What is the new behavior?
- [x] Integration testing

This pr implement parallel testing for integration testing, it is a apart of effort for #6390.
This reduces time for each execution down to ~15mins, a 2x speed up.
This patch enables integration test to be run parallelly on a test module level.
This patch has no impact on client code.

### Other information?

I've also cleaned up the script a bit as well, concurrency notes are put in for future maintainers where appropriate.

### DIFF
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -176,6 +176,133 @@ jobs:
             files: migration/coverage.xml
             flags: migrator
 
+    integration-testing:
+        runs-on: ubuntu-18.04
+        services:
+          postgres:
+            image: postgres
+            env:
+              POSTGRES_PASSWORD: submitty_dbuser
+              POSTGRES_USER: postgres
+            options: >-
+              --health-cmd pg_isready
+              --health-interval 10s
+              --health-timeout 5s
+              --health-retries 5
+            ports:
+              - 5432:5432
+
+        steps:
+          - uses: actions/checkout@v2
+            with:
+              path: SUBMITTY_CPY/
+          - name: Copy Repo
+            run: |
+              # have to copy first, absolute paths not supported by actions/checkout@v2
+                sudo mkdir -p ${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/Submitty
+                sudo cp -R SUBMITTY_CPY/. ${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/Submitty
+                sudo chmod -R a+rwx  ${SUBMITTY_INSTALL_DIR}
+                sudo chmod -R a+rwx /tmp/
+
+          - uses: actions/setup-python@v2
+            with:
+              python-version: '3.6'
+          - name: Install python dependencies
+            run: |
+              python3 -m pip install --upgrade pip setuptools wheel
+              sudo python3 -m pip install --upgrade pip setuptools wheel
+              pip3 -V
+              sudo pip3 -V
+              pip3 install cryptography
+              pip3 install python-pam
+              pip3 install "sqlalchemy<1.4.0"
+              pip3 install selenium
+              pip3 install tzlocal
+              pip3 install jsonschema
+              pip3 install jsonref
+              pip3 install websocket_client
+              pip3 install psycopg2
+              pip3 install docker
+              pip3 install pyyaml
+              pip3 install selenium
+              pip3 install websocket_client
+              pip3 install python-pam
+              pip3 install python-dateutil
+              pip3 install paramiko
+              pip3 install pyzbar
+              pip3 install watchdog
+              #needs to be globally available for all users
+              sudo pip3 install docker
+              sudo pip3 install paramiko
+              sudo pip3 install tzlocal
+              sudo pip3 install jsonref
+              sudo pip3 install numpy
+              sudo pip3 install opencv-python
+              sudo pip3 install pyPdf
+              sudo apt-get install libzbar0
+              sudo pip3 install pyzbar
+              sudo pip3 install PyPDF2
+              sudo pip3 install pdf2image
+              sudo pip3 install onnxruntime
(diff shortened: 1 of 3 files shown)


shailpatels (approved):
Ran updated tests locally and looking at the output of GitHub actions this PR makes good improvements to the integration test suite. It does knock off about 10+ minutes on the integration tests step on GH actions which is great.

I looked over the code and it looks readable clear to me but I'm not an expert in python multiprocessing code, @MasterOdin do you want to take a quick look?

MasterOdin:
Looks like the test output is slightly borked:
```
--- BEGIN TEST MODULE TUTORIAL_12_SYSTEM_CALLS AT PROCESS 32404 ---
* Starting compilation...
* Starting prebuild for tutorial_12_system_calls... Done
* Finished compilation
* Starting testcase {modname}.{func.__name__}... PASSED
* Starting testcase {modname}.{func.__name__}... PASSED
* Starting testcase {modname}.{func.__name__}... PASSED
* Starting testcase {modname}.{func.__name__}... PASSED
* Starting testcase {modname}.{func.__name__}... PASSED
All testcases passed
--- END TEST MODULE TUTORIAL_12_SYSTEM_CALLS ---
```

wusatosi:
> Looks like the test output is slightly borked:
> 
> ```
> --- BEGIN TEST MODULE TUTORIAL_12_SYSTEM_CALLS AT PROCESS 32404 ---
> * Starting compilation...
> * Starting prebuild for tutorial_12_system_calls... Done
> * Finished compilation
> * Starting testcase {modname}.{func.__name__}... PASSED
> * Starting testcase {modname}.{func.__name__}... PASSED
> * Starting testcase {modname}.{func.__name__}... PASSED
> * Starting testcase {modname}.{func.__name__}... PASSED
> * Starting testcase {modname}.{func.__name__}... PASSED
> All testcases passed
> --- END TEST MODULE TUTORIAL_12_SYSTEM_CALLS ---
> ```

Missed an f on the template string, its now fixed

wusatosi:
> One gain for the integration test suite would be to move the compilation step of configure.out from a per-test instance to once per suite run. As part of the "prebuild" step, [build_main_configure.sh](https://github.com/Submitty/Submitty/blob/master/tests/integrationTests/build_main_configure.sh) is run where the compilation step of `configure.out` in it takes a number of seconds. This file is the same for all modules, and so could be safely hoisted to be compiled at the start of run.py to some temp location and then copied into place as needed for each test suite. This would save something like ~16s per module.

Speed up advise from matt (source: https://github.com/Submitty/Submitty/issues/6390#issuecomment-841946751)

wusatosi:
I also want to have a look into utilizing CMake cache, I presume this can increase our test speed by a large margin as well, and could also be used in the main e2e pipeline.

shailpatels:
> I also want to have a look into utilizing CMake cache, I presume this can increase our test speed by a large margin as well, and could also be used in the main e2e pipeline.

The CMake's cache is for storing environment variables, I'm not sure how this would speed of the tests? Unless you're thinking of something else

https://cmake.org/cmake/help/latest/variable/CACHE.html

wusatosi:
> > I also want to have a look into utilizing CMake cache, I presume this can increase our test speed by a large margin as well, and could also be used in the main e2e pipeline.
> 
> The CMake cache is for storing environment variables, I'm not sure how this would speed of the tests? Unless you're thinking of something else
> 
> 

okay, I thought cmake have compilation cache built in, I am refering to compilation cache, something similar to this: https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/

MasterOdin:
Two notes:
   
1. Please try and draw lines in what a PR does, keeping a PR focus as minimal as possible. This helps in two big ways: (1) makes it easier to review as the reviewer does not need to context switch (like right now I need to context switch between GH action and Python multiprocessing, very different things) and less code and (2) makes it easier to rollback code if somewhere where to break. For example, let's say the multiprocessing stuff broke something subtle and we revert it, we'd still want to keep the GH action changes and they're unrelated so could be kept.

1. > Speed up advise from matt (source: #6390 (comment))
   
   I don't see a change relating to this? I'd also like not to per the note above.

wusatosi:
> Two notes:
> 
> 1. Please try and draw lines in what a PR does, keeping a PR focus as minimal as possible. This helps in two big ways: (1) makes it easier to review as the reviewer does not need to context switch (like right now I need to context switch between GH action and Python multiprocessing, very different things) and less code and (2) makes it easier to rollback code if somewhere where to break. For example, let's say the multiprocessing stuff broke something subtle and we revert it, we'd still want to keep the GH action changes and they're unrelated so could be kept.
> 2. > Speed up advise from matt (source: #6390 (comment))
>    
>    
>    I don't see a change relating to this? I'd also like not to per the note above.

Understood, I was making notes to myself as I clean up my work yesterday, and you are right.

wusatosi:
CI Related code reverted.

wusatosi:
This is reverted back to commit `Logging fix`, bda8be4.
CI-related code will be opened in another PR.
Ready for review

MasterOdin (approved):
Works as expected, can still one individual modules and testcases.

Something to consider (but not necessarily fix yet) is that in the move to multiprocessing, our test output is going to get "jumbled" up such that we may have multiple tested modules printing out at the same time. Perhaps better would be to mimic what other test suites do (e.g. jest) where it implements a TUI that lists out each module name if it passed or not, and then a little text to indicate that it's "running", and then once a given module finishes, it updates the line for that module. The modules may finish in any order, and once a module does finish, we output the error. Hopefully that makes sense, but that is very much a "wouldn't that be swell" kind of thing, and not really a priority in anyway.

PR #6426, opened 2021-05-14
[Bugfix:InstructorUI] Add late days instead of overwriting

### LINKED ISSUE #6377: Late days added through "Late Days Allowed" page are not added to the total number of late days for a student
**Describe the bug**
The number of additional late days granted to a student through the "Late Days Allowed" dialog is not added to the initial number of late days specified in the course configuration but replaces it. This contradicts the description on the "Late Days Allowed" page.

**Expected behavior**
Since the "Late Days Allowed" page says: "Use this form to grant students additional late days (beyond the initial number specified in the course configuration).", the number of late days entered on this page should be added to the initial number of late days specified in the course configuration instead of replacing it.

**To Reproduce**
0. Check course configuration for the initial number of late days (see screenshot 1).
1. Navigate to https://submitty.cs.rpi.edu/courses/s21/csci4960/late_days
2. Select the student, datestamp, and enter the desired number of additional late days (see screenshot 2).
3. Navigate to the latest gradeable of the student, e.g., https://submitty.cs.rpi.edu/courses/s21/csci4960/gradeable/hw05/grading/grade?who_id=y3qrkXuR0f3zHhZ&sort=id&direction=ASC
4. See the number of late days entered in the "Late Days Allowed" actually reducing the total number of late days available to the student (see screenshot 3).

**Configuration**
 - Browser: Firefox 88.0 (64-bit)

**Screenshots**
image
image
image

**Additional context**
Either the description "Use this form to grant students additional late days (beyond the initial number specified in the course configuration)." should be changed to reflect the actual behavior or the algorithm of how additional late days are computed should be changed to match the description. The latter would make a lot more sense than just overriding the total late days (including the initial late days specified in the course configuration) every time a new value is entered through the "Late Days Allowed" page.

williamjallen:
When attempting to add late days to a student, the current behavior is to overwrite the number of late days with the number of late days entered.

### What is the new behavior?
This update clarifies that the number of late days is the number of *additional* late days and fixes the bug which resulted in the late days being overwritten each time.

### Other information?
Fixes #6377

### DIFF
--- a/site/app/controllers/admin/LateController.php
+++ b/site/app/controllers/admin/LateController.php
@@ -98,9 +98,7 @@ class LateController extends AbstractController {
 
             $date_time = DateUtils::parseDateTime($_POST['datestamp'], $this->core->getUser()->getUsableTimeZone());
 
-
-            $existing_user = $this->core->getQueries()->getLateDaysForUserOnTimestamp($_POST['user_id'], $date_time);
-            if (isset($existing_user)) {
+            if ($this->core->getQueries()->getLateDaysForUserOnTimestamp($_POST['user_id'], $date_time) !== null) {
                 $this->core->getQueries()->updateLateDays($_POST['user_id'], $date_time, $_POST['late_days'] + $existing_user->getAllowedLateDays());
             }
             else { // The user doesn't have any late days yet so we just assign them the number of late days specified.
--- a/site/app/controllers/admin/LateController.php
+++ b/site/app/controllers/admin/LateController.php
@@ -107,7 +107,7 @@ class LateController extends AbstractController {
                 $user_timestamp = DateUtils::parseDateTime($user->getSinceTimestamp(), $this->core->getUser()->getUsableTimeZone());
                 if ($user->getId() == $_POST['user_id'] && $user_timestamp == $date_time) {
                     $this->core->getQueries()->updateLateDays($_POST['user_id'], $date_time, $_POST['late_days'] + $user->getAllowedLateDays());
-                    $exists = True;
+                    $exists = true;
                     break;
                 }
             }
--- a/site/app/controllers/admin/LateController.php
+++ b/site/app/controllers/admin/LateController.php
@@ -98,21 +98,12 @@ class LateController extends AbstractController {
 
             $date_time = DateUtils::parseDateTime($_POST['datestamp'], $this->core->getUser()->getUsableTimeZone());
 
-            // This is by all means an ugly hack but this same process is completed elsewhere while loading this page
-            // and there is no way to get a specific user by ID and get the number of late days at the same time.
-            // Ultimately, the performance difference is rather negligible since only instructors load this page.
-            $users = $this->core->getQueries()->getUsersWithLateDays();
-            $exists = false;
-            foreach ($users as $user) {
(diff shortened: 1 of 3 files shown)


site/app/controllers/admin/LateController.php
```diff
@@ -98,7 +98,24 @@ public function updateLateDays($csv_option = null) {
 
             $date_time = DateUtils::parseDateTime($_POST['datestamp'], $this->core->getUser()->getUsableTimeZone());
 
-            $this->core->getQueries()->updateLateDays($_POST['user_id'], $date_time, $_POST['late_days']);
+            // This is by all means an ugly hack but this same process is completed elsewhere while loading this page
+            // and there is no way to get a specific user by ID and get the number of late days at the same time.
+            // Ultimately, the performance difference is rather negligible since only instructors load this page.
```
MasterOdin, line 103: Why not just add a new query function that gives you the specific user?

MasterOdin (changes requested):

site/app/controllers/admin/LateController.php
williamjallen: I could, but that requires modification of the database files.  I do not know SQL well enough to do that in a manner such that I am certain that there are no vulnerabilities or points of failure.  It is also worth noting that all of the users are accessed in multiple other places on this page and we can assume that there are is a conservative maximum of 1000 students being iterated through which is relatively small given that only instructors use this page and they only use it occasionally.

wusatosi:
This function is not covered by tests, can you add some tests for this function?

site/app/controllers/admin/LateController.php
MasterOdin: You should not worry about messing up the SQL in such a way that would introduce a security failure or point of failure. That will ideally get caught and ironed out in the code review process. I'm happy to give information on the SQL interface or practices or whatever you might need so feel free to ask on Slack. Really, the only big main thing to do is just always use escaped queries (i.e. with the `?` for your input and then pass the variables in as an array to `query()` method).

Anyway, here's a SQL function you can use to accomplish what you want:

```
    public function getLateDaysForUserOnTimestamp(string $user_id, \DateTime $since_timestamp): ?SimpleLateUser {
        $this->course_db->query(
            "
        SELECT u.user_id, user_firstname, user_preferred_firstname,
          user_lastname, user_preferred_lastname, allowed_late_days, since_timestamp
        FROM users AS u
        FULL OUTER JOIN late_days AS l
          ON u.user_id=l.user_id
        WHERE allowed_late_days IS NOT NULL
          AND u.user_id = ?
          AND since_timestamp = ?
        ORDER BY
          user_email ASC, since_timestamp DESC;",
          [$user_id, $since_timestamp)]
        );

        return $this->course_db->getRowCount() === 1
            ? new SimpleLateUser($this->core, $this->course_db->row())
            : null;
    }
```

The modification for this function from `getUsersWithLateDays` is the introduction of the two new WHERE clauses on `u.user_id` and `since_timestamp`, which we use to get a singular row from the DB that fit those criteria. The \PDO adapter handles converting the `\DateTime` object into something that slots into the query, and the postgres will handle timezone conversions on the comparison for us. If no such row exists, then we just return null.

williamjallen: Thanks!  I have made the requested changes.

Issue #5913, opened 2020-09-24
Polling: New Features


BenKelly99:
This is a list of all features that could be added to polling:

- [ ] Bulk import and export of polls. An instructor should be able to generate a JSON of all polls that they can use again in future semesters. For importing, it would be a bonus if there was a clean way to convert the old release times to new release times instead of just copying over the raw dates, since they are most likely quite old.

- [ ] All instructors to include a picture for the poll. For some questions, a picture or diagram is needed to ask the question. Allow the instructor to upload a picture to the poll, and then display it to students when they go to answer. Make sure to set a limit on the upload size for the photo.

- [ ] Integrate polling with rainbow grades so that summaries are posted on their automatically when reports are generated.

- [ ] Integrate polling with late days so instructors can use them to give students extra late days. Generally, instructors will have a set of cutoffs for when students earn them. E.g. a total of 10 gives 1 extra late day, a total of a 20 gives another, 35 gives another, and so on.

- [ ] Summary page for the student to view their results on all the polls. This might just be the rainbow grades part and so might not be necessary.

- [ ] Allow an instructor to view a student's poll activity. If an instructor wants to look up how a student is doing in polls to get a sense of their participation, there should be a way for instructors to do that. UI is flexible but could be a search bar on the polling page seen by instructors.

Harshalzzzzzzz:
@BenKelly99 
Dear Mentor ,
I would like to work on this issue, can you please guide me further on what needs to be done and how I can approach this issue and contribute to the project?
Thank you

BenKelly99:
Hi @Harshalzzzzzzz,
Between the two issues there's a lot of items that could be done right now. Does anything stick out to you as something you'd be interested in? If you choose a specific one I can point you in the right direction.

codeboy5:
@BenKelly99 I was interested in adding the 'bulk import/export' feature, can you help or guide a little as to how i should start ?

BenKelly99:
@codeboy5 Great! There will be a few things you need to do. The files to look at are PollController.php, PollModel.php, PollView.php, all the templates in the site/app/templates/polls/ folder, as well as the polling functions in DatabaseQueries.php (just use ctrl+F to find "poll"). For this issue, you should make a button that downloads all the polls in the database as a JSON. To get all the polls, you can call the getPolls() method in DatabaseQueries.php, which returns an array of PollModel objects. From there, you can convert them to a JSON, including the title, question, all the responses, etc.
[...]

PR #6433, opened 2021-05-17
[Feature:SubminiPolls] Poll Image Type and Size 

sbelsk:
Fixes one of the issues listed in #5913

The instructor's "Add Poll" and "Edit Poll" pages accept any file type as the optional image, with no error checking. The student's "View Poll" shows images properly, but shows a generic file icon if non-image files are attached to a poll.

### What is the new behavior?
The pop-up file browser in the "Add Poll" and "Edit Poll" pages accepts only image file types (JPEG, JPG, PNG, GIF), and only files under the size limit can be attached to a poll.

### DIFF
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -163,40 +163,20 @@ class PollController extends AbstractController {
         $poll_id = $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
         $file_path = null;
         if (isset($_FILES['image_file']) && $_FILES["image_file"]["name"] !== "") {
-            //$file = $_FILES["image_file"];
-
-            // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-            $errors = array();
             $maxsize = Utils::returnBytes(ini_get('upload_max_filesize'));
-            $acceptable = array(
-                'image/jpeg',
-                'image/jpg',
-                'image/gif',
-                'image/png'
-            );
-            if(($_FILES['image_file']['size'] >= $maxsize) || ($_FILES["image_file"]["size"] == 0)) {
+            $acceptable = array('image/jpeg','image/jpg','image/gif','image/png');
+            if (($_FILES['image_file']['size'] >= $maxsize) || ($_FILES["image_file"]["size"] == 0)) {
                 $this->core->addErrorMessage("File(s) uploaded too large.  Maximum size is " . ($max_size / 1024) . " kb.");
             }
-            if((!in_array($_FILES['image_file']['type'], $acceptable)) && (!empty($_FILES["image_file"]["type"]))) {
-                $this->core->addErrorMessage("Upload failed: Invalid file type. Only JPG, GIF and PNG types are accepted.");
+            elseif ((!in_array($_FILES['image_file']['type'], $acceptable)) && (!empty($_FILES["image_file"]["type"]))) {
+                $this->core->addErrorMessage("Upload failed: Invalid file type. Only JPG, JPEG, GIF and PNG types are accepted.");
             }
-            if(count($errors) === 0) {
+            else {
                 $file = $_FILES["image_file"];
                 $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $poll_id . "_" . $_FILES["image_file"]["name"]);
                 move_uploaded_file($file["tmp_name"], $file_path);
                 $this->core->getQueries()->setPollImage($poll_id, $file_path);
-            } else {
-                foreach($errors as $error) {
-                    echo '<script>alert("'.$error.'");</script>';
-                }
-                die(); //Ensure no more processing is done
             }
-            // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-
-            //$file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $poll_id . "_" . $_FILES["image_file"]["name"]);
-            //move_uploaded_file($file["tmp_name"], $file_path);
-            //$this->core->getQueries()->setPollImage($poll_id, $file_path);
         }
 
         return MultiResponse::RedirectOnlyResponse(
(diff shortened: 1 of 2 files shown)


sbelsk:
Comment: when a file larger than 10MB is uploaded, a different error handling bypasses my error checking for the file size, which I assume is the default way Submitty handles submission of files which are too large:
image

shailpatels:
> Comment: when a file larger than 10MB is uploaded, a different error handling bypasses my error checking for the file size, which I assume is the default way Submitty handles submission of files which are too large:

Yeah this would be hitting the default apache setting, if you want to test larger files you would need to configure the max file settings:

https://submitty.org/sysadmin/system_customization#allowing-large-student-file-upload-submissions

But as long as the error message is handled gracefully and the user knows they've uploaded too big of a file it should be fine

site/app/controllers/PollController.php
```diff
@@ -163,10 +163,20 @@ public function addNewPoll() {
...
-            $file = $_FILES["image_file"];
-            $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $poll_id . "_" . $_FILES["image_file"]["name"]);
-            move_uploaded_file($file["tmp_name"], $file_path);
-            $this->core->getQueries()->setPollImage($poll_id, $file_path);
+            $max_size = Utils::returnBytes(ini_get('upload_max_filesize'));
+            $acceptable = ['image/jpeg','image/jpg','image/gif','image/png'];
+            if (($_FILES['image_file']['size'] >= $max_size) || ($_FILES["image_file"]["size"] == 0)) {
+                $this->core->addErrorMessage("File(s) uploaded too large.  Maximum size is " . ($max_size / 1024) . " kb.");
+            }
+            elseif ((!in_array($_FILES['image_file']['type'], $acceptable)) && (!empty($_FILES["image_file"]["type"]))) {
```
shailpatels, line 171: For the file upload validation you should use 
`validateUploadedFiles` from `site/app/libraries/FileUtils.php` which will take each uploaded file and check if there are any errors like invalid name, file size, and other upload errors. 
It will return an associative array with the status of each file with the following information:

```
'name' : name of file
'type' : filetype
'error' : an array of error messages (empty if none)
'size' : size of file
'is_zip' : true if zip file
'success' : true if there were no errors
```

Once the uploaded file is validated you can use `isValidImage` also in `FileUtils.php` to check if its an acceptable image file type.

site/app/controllers/PollController.php
```diff
@@ -350,8 +360,18 @@ public function submitEdits(): RedirectResponse {
...
+                $this->core->addErrorMessage("Upload failed: Invalid file type. Only JPG, GIF and PNG types are accepted.");
+            }
+            else {
+                $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $_POST["poll_id"] . "_" . $_FILES["image_file"]["name"]);
+                move_uploaded_file($file["tmp_name"], $file_path);
```
shailpatels, line 373: same comment here, use above validation strategy and propagate errors if any back to the user

site/app/templates/polls/NewPollPage.twig
```diff
@@ -27,7 +27,7 @@
             <label for="image-file">
                 Optional Image:
             </label>
-            <input type="file" id="image-file" name="image_file"/>
+            <input type="file" id="image-file" name="image_file" accept="image/jpeg, image/jpg, image/gif, image/png"/>
```
shailpatels, line 30: You can put `accept="image/*"` instead of manually putting each image type, I think its cleaner since we're accepting every image type anyways

site/app/templates/polls/NewPollPage.twig
```diff
@@ -81,7 +81,7 @@
             <label for="image-file">
                 Optional Image:
             </label>
-            <input type="file" id="image-file" name="image_file"/>
+            <input type="file" id="image-file" name="image_file" accept="image/jpeg, image/jpg, image/gif, image/png" />
```
shailpatels, line 84: same comment as above here

shailpatels (changes requested):

site/app/controllers/PollController.php
sbelsk: Alright. I'm changing it use the two functions you mentioned, but I'm running into an issue with the function `validateUploadedFiles`. Is there a way to use it when we are just expecting one file to be uploaded? I can't make it work for this case, I'm getting `Warning: count(): Parameter must be an array or an object that implements Countable` that comes from the body of the function on line https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/site/app/libraries/FileUtils.php#L549 I'm trying to call the function with the argument `FileUtils::validateUploadedFiles($_FILES['image_file']);` when the contents of `$_FILES` are: 
image

shailpatels: Oh maybe that function was designed only for inputs with multiple file uploads but I'll take a look a look at it later

site/app/templates/polls/NewPollPage.twig
sbelsk: From what I saw, `isValidImage()` checks if the picture is of MIME type image, and if the MIME sub-type is one of those in `FileUtils::ALLOWED_IMAGE_TYPE`. I took a look at that constant at this line: https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/site/app/libraries/FileUtils.php#L15
So I figured that specifying for jpeg, jpg, gif, and png types instead of "image/*" would prevent potential conflicts between what files the pop-up file browser allows you to select and what the function `isValidImage()` accepts. Did I misunderstand, or maybe do you think that this should be changed anyways?

shailpatels: Yeah that's true, I'm looking at the list of image types now and its more than the 4 we support so keeping it the same would work with the backend validation whatever you choose

shailpatels:
Ok debugging this more it appears that if you change the file upload HTML input to
```html
<input type="file" id="image-file" name="image_file[]"/>
```
PHP will upload the file in an image array instead of a single file upload
Then you can do 
```php
FileUtils::validateUploadedFiles($_FILES["image_file"]);
```

to get the proper result, I'll make a PR fixing the validate file upload to handle single file structures

site/app/controllers/PollController.php
```diff
@@ -163,10 +163,25 @@ public function addNewPoll() {
         $poll_id = $this->core->getQueries()->addNewPoll($_POST["name"], $_POST["question"], $responses, $answers, $_POST["release_date"], $orders);
         $file_path = null;
         if (isset($_FILES['image_file']) && $_FILES["image_file"]["name"] !== "") {
-            $file = $_FILES["image_file"];
-            $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $poll_id . "_" . $_FILES["image_file"]["name"]);
-            move_uploaded_file($file["tmp_name"], $file_path);
-            $this->core->getQueries()->setPollImage($poll_id, $file_path);
+            $status = FileUtils::validateUploadedFiles($_FILES["image_file"]);
+            // validate the uploaded file size
+            if (array_key_exists("failed", $status)) {
```
shailpatels, line 168: The array key `failed` will only be set if 0 files are sent to be validated which is taken care of the `isset` above. You can check if anything went wrong with `if( ! $status['success'])` and then use `$status['error']` to get an error message.

site/app/controllers/PollController.php
```diff
@@ -163,10 +163,25 @@ public function addNewPoll() {
...
+                unlink($file_path);
```
shailpatels, line 169: `file_path` is never set to anything other than null here so I think this line can be removed

site/app/controllers/PollController.php
```diff
@@ -163,10 +163,25 @@ public function addNewPoll() {
...
+                $this->core->getOutput()->renderResultMessage("Failed to validate uploads " . $status["failed"], false);
+            }
+            else {
+                $file = $_FILES["image_file"];
+                $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $poll_id . "_" . $_FILES["image_file"]["name"]);
+                move_uploaded_file($file["tmp_name"], $file_path);
+                // validate the uploaded file type is indeed an image
+                if (!FileUtils::isValidImage($file_path)) {
```
shailpatels, line 177: I think checking that the file is a valid image should be done before copying it to the polls area, that way we can avoid the operation if its not an image file

site/app/controllers/PollController.php
```diff
@@ -352,6 +367,13 @@ public function submitEdits(): RedirectResponse {
             }
             $file_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "polls", "poll_image_" . $_POST["poll_id"] . "_" . $_FILES["image_file"]["name"]);
             move_uploaded_file($file["tmp_name"], $file_path);
+            $status = FileUtils::validateUploadedFiles($_FILES["image_file"]);
+            if (array_key_exists("failed", $status)) {
```
shailpatels, line 371: same comment, use `$status['success']` to check if there were any errors and return whatever is set in `$status['error']` if so.

shailpatels (changes requested):

shailpatels (approved):

PR #6434, opened 2021-05-17
[Feature:InstructorUI] Add schema info to SQL toolbox

### LINKED ISSUE #6418: SQL Toolbox on page DB Schema documentation
An instructor who knows SQL syntax, but doesn't know the exact details of the Submitty course DB has to go digging through our .sql file (currently 2200 lines long):
https://github.com/Submitty/Submitty/blob/master/migration/migrator/data/course_tables.sql
Or squint at an out of date entity relationship diagram
https://submitty.org/developer/database_design

Or they could type a query like this:
```
SELECT *
FROM pg_catalog.pg_tables
WHERE schemaname != 'pg_catalog' AND 
    schemaname != 'information_schema';
```
(from https://www.postgresqltutorial.com/postgresql-show-tables/)

And then to get the field names & types of a table:
```
select column_name, data_type from information_schema.columns where table_name = 'users';
```

Note, when I'm directly logged into a DB, I usually do
```
select * from users where false; 
```
just to remind myself of the exact column names --> it prints the column headings with no rows of data.  But in our SQL Toolbox, currently nothing gets printed as a result of this query, not even the column headings.

Instead it would be great to have up-to-date basic documentation on this page.
* Maybe right underneath the title "SQL Toolbox" is a button/expandable/link titled "Database Schema Documentation".  When you click on that button, you get a simple list of the 43(current # of) tables in the database.
* Each one of the table names is now clickable/expandable, and when you do, you get a simple list of the fields.

Alternatively we could just put a link to submitty.org, a link to the .sql file, and try to keep the ER diagram up to date.  But it does depend on the exact version of Submitty installed -- our field names & types change occasionally.  So generated/onserver docs would be better.  Docs could be generated on each click, or written to a markdown file when we INSTALL_SUBMITTY.sh.

...  I'm sure this type of simple docs have been made before, please use a standard tool if one exists :)
Thoughts @MasterOdin ?

cjreed121:

Currently, an instructor must know what tables and columns are in the database to effectively use the SQL Toolbox. They must look into the Submitty docs to discover what tables are in the database.
Closes #6418 

### What is the new behavior?
There is a button that will show all of the tables and columns in the database at the top of the SQL Toolbox.
image

image

image

### Other information?
PHP tests for this are still in progress.


site/app/controllers/admin/SqlToolboxController.php
```diff
@@ -24,7 +24,19 @@ class SqlToolboxController extends AbstractController {
      * @Route("/courses/{_semester}/{_course}/sql_toolbox", methods={"GET"})
      */
     public function showToolbox(): WebResponse {
-        return new WebResponse(SqlToolboxView::class, 'showToolbox');
+        $this->core->getCourseDB()->query("SELECT * FROM information_schema.columns WHERE table_schema='public'");
```
MasterOdin, line 27: We can sort the table names as part of our SQL query:

```suggestion
        $this->core->getCourseDB()->query("SELECT * FROM information_schema.columns WHERE table_schema='public' ORDER BY table_name ASC, ordinal_position ASC");
```

We also want to sort on `ordinal_position` as there's no guarantee that column order in this table is stable / guaranteed compared to what you'd see by looking at table metadata in `psql` or some other software.

site/app/controllers/admin/SqlToolboxController.php
```diff
@@ -24,7 +24,19 @@ class SqlToolboxController extends AbstractController {
...
+        $tables = $this->core->getCourseDB()->rows();
+        $organizedTables = [];
+        //Loop through and create a 2d array that holds all columns for each table name.
+        foreach ($tables as $table) {
+            if (!isset($organizedTables[$table['table_name']])) {
+                $organizedTables[$table['table_name']] = [];
+            }
+            array_push($organizedTables[$table['table_name']], $table['column_name']);
```
MasterOdin, line 35: Better to use:

```suggestion
            $organizedTables[$table['table_name']][] = $table['column_name'];
```

as this avoids a function call (`array_push`) that is unnecessary.

site/app/controllers/admin/SqlToolboxController.php
```diff
@@ -24,7 +24,19 @@ class SqlToolboxController extends AbstractController {
      * @Route("/courses/{_semester}/{_course}/sql_toolbox", methods={"GET"})
      */
     public function showToolbox(): WebResponse {
-        return new WebResponse(SqlToolboxView::class, 'showToolbox');
+        $this->core->getCourseDB()->query("SELECT * FROM information_schema.columns WHERE table_schema='public'");
+        $tables = $this->core->getCourseDB()->rows();
+        $organizedTables = [];
+        //Loop through and create a 2d array that holds all columns for each table name.
+        foreach ($tables as $table) {
```
MasterOdin, line 31: Since we don't use `$tables` anywhere except in the foreach, better to avoid defining it altogether.
```suggestion
        $organizedTables = [];
        //Loop through and create a 2d array that holds all columns for each table name.
        foreach ($this->core->getCourseDB()->rows() as $table) {
```

site/app/views/admin/SqlToolboxView.php
```diff
@@ -7,14 +7,16 @@
 use app\views\AbstractView;
 
 class SqlToolboxView extends AbstractView {
-    public function showToolbox(): string {
+    public function showToolbox($tables): string {
```
MasterOdin, line 10: Should always try to provide some sort of type hint if possible.
```suggestion
    public function showToolbox(array $tables): string {
```

site/app/templates/admin/SqlToolbox.twig
```diff
@@ -4,6 +4,28 @@
     <div>
       Use this toolbox to run a SELECT query. You cannot run any other type of query, and may only run a single query at a time.
       <br /><br />
+
+        <div>
+            <button id="sqlDatabaseSchema" class="btn btn-primary">Database Schema Documentation</button>
```
MasterOdin, line 9: Per our docs (https://submitty.org/developer/coding_style_guide/css), IDs and classes should be `kebab-case`, e.g. `sql-dashboard-schema`.

MasterOdin (changes requested):
I would like to see the controller just call something like:

```
$this->core->getQueries()->getCourseSchemaTables();
```

and that returns the `$organizedTables` array that you've defined and that then gets passed to the WebResponse as you have it.

cjreed121:
For the PHP Unit testing. I know I need to make more changes to SqlToolboxControllerTester.php but am not sure exactly sure what I am supposed to use. The assertEmpty on line 54 (my version) should be changed to an assertSame but should I just use similar code to what I have in my controller to test it against? Or what should I be checking it against? Also how do I resolve js unit not knowing what jQuery is?

site/public/mjs/sql-toolbox.js
```diff
@@ -77,6 +77,12 @@ export async function runSqlQuery() {
 
 export function init() {
     document.getElementById('run-sql-btn').addEventListener('click', () => runSqlQuery());
+    $('#sql-database-schema').on('click', () => {
+        $('#sql-database-schema-content').toggle();
+    });
+    $('.sql-database-table').on('click', (e) => {
+        $(e.target).next('.sql-database-columns').toggle();
+    });
```
MasterOdin, line 85:
```suggestion
    document.getElementById('sql-database-schema').addEventListener('click', () => {
        document.getElementById('sql-database-schema-content').style.display =
            document.getElementById('sql-database-schema-content').style.display === 'block'
                ? 'none'
                : 'block';
    });
    document.querySelectorAll('.sql-database-table').forEach((elem) => {
        elem.addEventListener('click', (e) => {
            e.target.nextElementSibling.style.display =
                e.target.nextElementSibling.style.display === 'block'
                    ? 'none'
                    : 'block';
        });
    });
```

Easy enough to rewrite this to not require jQuery, which I'd like to avoid as trying to import it both as a regular code and module import gets messy.

site/tests/app/controllers/admin/SqlToolboxControllerTester.php
```diff
@@ -41,6 +41,12 @@ public function tearDown(): void {
     }
 
     public function testShowToolbox(): void {
+        $this->setUpDatabase();
+
+        /*$courseDb = $this->core->getCourseDB();
+        $courseDb->expects($this->once())->method('query')->with("SELECT * FROM information_schema.columns WHERE table_schema='public'");
+        $courseDb->expects($this->once())->method('rows')->with()->willReturn();*/
```
MasterOdin, line 48: > The assertEmpty on line 54 (my version) should be changed to an assertSame but should I just use similar code to what I have in my controller to test it against? Or what should I be checking it against?

For this to work, as you're using `getCourseQueries`, you'll want to mock that, and not the course db. While you could do the latter, that expands the scope of the unit under test, which we want to avoid, making the test as targeted as possible against the smallest possible unit.

```
$tables = [
	// some snippet of data as expected in prod
];
$mock_queries = $this->createMock(DatabaseQueries::class);
$mock_queries->expects($this->once())->method('getCourseSchemaTables')->willReturn($tables);
$core->setQueries($mock_queries);
```

and then below, we can confirm that our table object was passed to our WebResponse parameters:

```
$this->assertSame($response->parameters, [$tables]);
```

cjreed121: Before I do the assertSame, I would need to run my array merger so the assertion can compare exactly, right? That is what I put in my latest commit which led it to pass the assertion in php unit. Or should I know the result of my array merger and hardcode that in to test against that?

cjreed121: Got it. Thanks.

bmcutler (changes requested):
Looks great!

Minor request, can you add the type (int, bool, timestamp) for each field as well?

Also, "categories_list" isn't a real Submitty data table, can we suppress the display of that one table?

bmcutler:
Nevermind, I just realized that categories list is a real table, it's for the forum :)
So ignore that request.

site/app/controllers/admin/SqlToolboxController.php
```diff
@@ -24,7 +24,15 @@ class SqlToolboxController extends AbstractController {
...
      */
     public function showToolbox(): WebResponse {
-        return new WebResponse(SqlToolboxView::class, 'showToolbox');
+        $organizedTables = [];
+        //Loop through and create a 2d array that holds all columns for each table name.
+        foreach ($this->core->getQueries()->getCourseSchemaTables() as $table) {
+            if (!isset($organizedTables[$table['table_name']])) {
+                $organizedTables[$table['table_name']] = [];
+            }
+            $organizedTables[$table['table_name']][] = $table['column_name'] . " - " . $table['data_type'];
```
MasterOdin, line 33: So, I think a better construct would be to do:
```php
            $organizedTables[$table['table_name']][] = [
                'name' => $table['column_name'],
                'type' => $table['data_type'],
            ];
```
and then in the twig template, we combine the columns to do `{{column.name}} - {{column.type}}`. The purpose behind this is in trying to make sure we do view stuff as close to the view layer as possible, and let the controllers and models interact with the data as close to its source as possible. This allows for easier augmentation of the data object with more information if needed here, as well as we could re-use the data model elsewhere more easily.

MasterOdin (approved):

MasterOdin:
For reference on how this currently looks:

Screen Shot 2021-05-30 at 10 09 48 AM

@bmcutler are you fine with how the data types are displayed?

bmcutler (approved):

PR #6435, opened 2021-05-17
[Bugfix:Forum] Fix edit categories button bug

### LINKED ISSUE #6049: Forum Edit Categories routing buggy
**Describe the bug**
From the full page forum screen, as instructor, going to the drop down "edit categories" doesn't work.
It launches a new tab which re-directs (?) back to the full page forum.

When a specific thread is selected, going to the drop down edit categories does work correctly.
However, it also launches itself in a new tab -- I dont remember if this was an intentional design.  
It seems like a new change.  I don't think any other page on our site launches itself in a new tab, 
and I think we should avoid doing so unless we need it for some good reason.

williamjallen:
Fixes #6049.

### What is the new behavior?
This change makes the "Edit Categories" menu option link to the correct page when viewing the wide forum view and prevents that menu option from opening new tabs on all views.  In addition, Bootstrap separators were fixed to be compatible with Bootstrap 2 and now display properly.

### DIFF
--- a/site/app/templates/forum/ForumBar.twig
+++ b/site/app/templates/forum/ForumBar.twig
@@ -87,29 +87,45 @@
 				{% if user_group <= more.required_rank %}
 					{% set onclick = (more.onclick[0]) ? 'onclick=' ~ more.onclick[1] %}
 					{% set optional_class = (more.optional_class[0]) ? 'class=' ~ more.optional_class[1] %}
-	          		<li class='dropdown-item' role="menuitem"><a id="{{ more.id }}" title="{{ more.title }}" {{ optional_class }} href="{{ more.link }}" {{ onclick }}>{{ more.display_text }}</a></li>
+	          		<li class='dropdown-item' role="menuitem">
+						<a id="{{ more.id }}" title="{{ more.title }}" {{ optional_class }} href="{{ more.link }}" {{ onclick }}>{{ more.display_text }}</a>
+					</li>
 	          	{% endif %}
 	        {% endfor %}
 	        {% if core.getUser().accessGrading() %}
-              <li class="dropdown-divider"></li>
-              <li title="Click to edit categories" class='dropdown-item' role="menuitem">
-            	<a href ="{{ manage_categories_url }}" title="Edit Forum Categories" target="_blank" >Edit Categories</a>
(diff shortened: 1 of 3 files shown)


site/app/templates/forum/ForumBar.twig
```diff
@@ -87,29 +87,45 @@
...
-	              {% if user_group <= core.getUser().accessGrading() %}
-	              	<li title="Sort posts by author's last name" id="alpha" class='dropdown-item' role="menuitem"><a href="#" class="key_to_click" tabindex="0" onclick="changeDisplayOptions('alpha')">Alphabetical</a></li>
-	              	<li title="Sort posts by author's registration section then last name" id="alpha_by_registration" class='dropdown-item' role="menuitem"><a href="#" class="key_to_click" tabindex="0" onclick="changeDisplayOptions('alpha_by_registration')">Alpha by Registration</a></li>
-	              	<li title="Sort posts by author's rotating section then last name" id="alpha_by_rotating" class='dropdown-item' role="menuitem"><a href="#" class="key_to_click" tabindex="0" onclick="changeDisplayOptions('alpha_by_rotating')">Alpha by Rotating</a></li>
-	              {% endif %}
+	          	<li class="divider"></li>
+	          	<li title="Click to toggle all post attachments" class='dropdown-item' role="menuitem">
+	          		<a href="#" class="key_to_click" tabindex="0" onclick="loadAllInlineImages()">
```
cjreed121, line 104: This doesn't actually toggle all attachments. It will open all of them but clicking on this again does not close all attachments. This would probably be a separate issue and PR though.

cjreed121 (approved):

site/app/templates/forum/ForumBar.twig
williamjallen: Good catch.  I agree that it should be another issue+PR.  I suspect that there might be other issues with this menu as well given how complex some of the code is.

MasterOdin: Opened #6475

MasterOdin (approved):
@williamjallen could you make a follow-up PR to normalize `site/app/templates/forum/ForumBar.twig` to only have space indentation, instead of the combination of tabs and spaces it currently has?

Issue #6036, opened 2020-12-01
Lecture Poll (Sub-mini) requests


bmcutler:
- [x] Instructor UI: have the date for a new poll default to today.  I think it currently starts blank and my browser is suggesting a date from a month ago.  (requires lots of clicks just to get to today.)

- [x] Instructor UI: have an indication on the main page if a poll doesn't have at least one correct answer selected.  Or maybe don't let me create a poll unless I select a correct answer.  We can discuss.

- [x] Instead of centering the title of the poll on the main page, let's left justify it.

- [ ] Let's have a way for the instructor to view the student view of the poll.  (Currently, all I can do is show the interface to edit the poll, which also has the correct answer(s) selected.)

- [x] Let's discuss:  an option for the instructor to "publish" the graph of the results (probably only once the poll is closed).  This should be an option, not automatic.  Sometimes, the poll has a 'correct' answer, and I want to re-use the poll...  so I don't want it published.  Sometimes, it is an opinion poll and I would like students to be able to see each others answers.  Sometimes we might light to publish the results while it is still open and accepting results.  This should probably be another column of checkboxes. 

... hopefully these aren't all repeats from other open issues :)

bmcutler:
The "Delete" button the instructor UI is rarely used, and can take up less room...  replace it with a trash can.
The "Edit poll" button the instructor UI could be replaced with the pencil icon.
Let's standardize the styling of this page to make it more similar to the all gradeables page instructor view.

wdturner:
I am happy to take a look, but was this intended to be assigned to me?

sbelsk:
I looked into some of the problems listed in the original comment:

1. I could not reproduce the issue of the date for the new poll defaulting to a date different than today.
I assume that it functions properly since the code used to select the date is used elsewhere on Submitty (late days, gradeables, forum) where that was no such issue.

2. The indication/message for having at least one correct answer selected was resolved in PR https://github.com/Submitty/Submitty/pull/6181

3. For the third point "Instead of centering the title of the poll on the main page, let's left justify it." -- what page is meant by "main page" (the instructor's Edit Poll page, student's View Poll page, or Poll summary for both students and instructors)?

bmcutler:
[...]

PR #6436, opened 2021-05-17
[UI/UX:SubminiPolls] Improve summary page UI

sbelsk:
Addresses issue https://github.com/Submitty/Submitty/issues/6036

The student view has the titles centered, while the instructor's are left justified. The titles in the instructor's page are forced to wrap earlier than necessary around the "edit" and "delete" icons.
before
past_student_view

### What is the new behavior?
Titles have been left-aligned and the widths of the columns have been adjusted. The titles in the instructor's page no longer wrap around unnecessarily.
after
student_page

### DIFF
--- a/site/app/templates/polls/AllPollsPageInstructor.twig
+++ b/site/app/templates/polls/AllPollsPageInstructor.twig
@@ -17,13 +17,17 @@
             </div>
         </div>
         <table id="today-table" class="table table-striped" style="width:100%;{{ dropdown_states['today'] ? '' : ' display: none;'}}">
-            <col style="width: 20%">
-            <col style="width: 20%">
-            <col style="width: 20%">
-            <col style="width: 20%">
-            <col style="width: 20%">
+			<col style="width: 5%">
+			<col style="width: 5%">
+			<col style="width: 45%">
+            <col style="width: 10%">
+            <col style="width: 10%">
+            <col style="width: 10%">
+            <col style="width: 15%">
             <thead>
                 <tr>
+                    <th scope="col"></th>
+                    <th scope="col"></th>
                     <th scope="col"></th>
                     <th scope="col">Visible</th>
                     <th scope="col">Accepting Answers</th>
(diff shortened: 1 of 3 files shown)


site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -17,13 +17,17 @@
...
+	    <col style="width: 5%">
+	    <col style="width: 5%">
+	    <col style="width: 45%">
+            <col style="width: 10%">
+            <col style="width: 10%">
+            <col style="width: 10%">
+            <col style="width: 15%">
             <thead>
                 <tr>
+                    <th scope="col"></th>
```
shailpatels, line 29: Just one small change, we should use `<td>` instead of `<tr>`  from [WAVE](https://wave.webaim.org/):

> #### What It Means
> A <th> (table header) contains no text.
> #### Why It Matters
> The <th> element helps associate table cells with the correct row/column headers. A <th> that contains no text may result in cells with missing or incorrect header information.
> #### How to Fix It
> If the table cell is a header, provide text within the cell that describes the column or row. If the cell is not a header or must remain empty (such as the top-left cell in a data table), make the cell a <td> rather than a <the>.

This error existed before on master, but this would be a good PR to fix the error.

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -72,14 +80,18 @@
...
+            <col style="width: 5%">
+            <col style="width: 30%">
             <col style="width: 15%">
+            <col style="width: 10%">
+            <col style="width: 10%">
+            <col style="width: 10%">
             <col style="width: 15%">
             <thead>
                 <tr>
+                    <th scope="col"></th>
```
shailpatels, line 93: same comment as above

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -129,12 +145,16 @@
...
-            <col style="width: 25%">
+            <col style="width: 5%">
+            <col style="width: 5%">
+            <col style="width: 50%">
+            <col style="width: 15%">
+            <col style="width: 10%">
+            <col style="width: 15%">
             <thead>
                 <tr>
+                    <th scope="col"></th>
```
shailpatels, line 156: same comment as above

shailpatels (changes requested):
Changes look good, tests on MacOS chrome, firefox, and safari and visuals look as described in PR 

1 small accessibility-related change requested  - other than should be ready to merge soon

site/app/templates/polls/AllPollsPageInstructor.twig
sbelsk: I tried to change all the relevant `<th>` tags to `<td>` like you suggested, and while it looks normal in light mode, in dark mode there are white lines only above the columns that still have the `<th>` tags. Do you think that's fine and that we should keep it that way, or should we leave it as it was before?
image

shailpatels: Yeah you can make them white by modifying the css class in `site/public/server.css` on ln 225 to be 
```css
.table tr, .table td[scope="col"]{
    border-top: 1px solid #ddd;
}
```
which should make the top the same color as the rest

sbelsk:
@shailpatels it looks like now after I changed the `<th>` that didn't have text in them to `<td>` like you suggested, the code fails the e2e tests because of the `scope="col"` in the  `<td>` elements (which were previously `<th>`). If I remove the `scope="col"`, the white line above the table in Dark Mode appears only above the `<th>` columns. Every time we try to solve one problem, another pops 😂. Any suggestions?

shailpatels:
ugh thats annoying I was hoping this would be a small fix, but in terms of accessibility its a small error, unfortunate we can't fix it but we can leave it as a todo for now since the selenium e2e tests are fragile and hard to update. 

How about you revert back to before the td was added and I can approve

shailpatels (approved):
Approved

MasterOdin (commented):
Fixed this myself to merge this, but something to check on future PRs.

PR #6437, opened 2021-05-17
[UI/UX:SubminiPolls] Fix Icons wrapping in Edit Poll

sbelsk:
In the instructor's Edit Poll/New Poll page, the "up", "down" and "delete" buttons for the responses wrap around when the width of the page is decreased. 
image

### What is the new behavior?
The text field adjusts to the width of the page and the buttons always stay on the same row to the right of the response in the Edit Poll page.
image


loadedskiffer (approved):
Everything looks good, she just used the calc() function to fix the wrapping and it appears to work fine.

MasterOdin (approved):

PR #6446, opened 2021-05-18
[UI/UX:SubminiPolls] Fix poll edit buttons in dark mode

### LINKED ISSUE #6444: Edit Poll UI in Dark Mode
**Describe the bug**
The "up" and "down" arrows in the Edit Poll page, when the page is in dark mode, are hard to see since they stay black.
image

**Expected behavior**
The buttons should change to white, like the trash can icon does.

**To Reproduce**
1. Switch website to dark mode
2. Go to the instructor's New Poll/Edit Poll page
3. Add new responses

**Configuration**
 - OS: Windows
 - Browser: Chrome, Firefox

williamjallen:
Fixes #6444

### What is the new behavior?
Makes the border around buttons on the edit polls UI turn white in dark mode.  This update also switches the arrows to proper icons.

### DIFF
--- a/site/app/templates/polls/NewPollPage.twig
+++ b/site/app/templates/polls/NewPollPage.twig
@@ -168,7 +168,7 @@
                     <i class="fa fa-lg fa-chevron-down move-arrow down"></i>
                 </div>
                 <div class="move-btn delete-btn">
-                    <i class="fa fa-lg fa-trash delete"></i>
+                    <i class="fa fa-lg fa-trash"></i>
                 </div>
                 <br/>
             </div>
--- a/site/app/templates/polls/NewPollPage.twig
+++ b/site/app/templates/polls/NewPollPage.twig
@@ -113,7 +113,7 @@
                             <i class="fa fa-lg fa-chevron-up"></i>
                         </div>
                         <div class="move-btn down-btn">
-                            <i class="fa fa-lg fa-chevron-down move-arrow down"></i>
+                            <i class="fa fa-lg fa-chevron-down"></i>
                         </div>
                         <div class="move-btn delete-btn">
                             <i class="fa fa-lg fa-trash delete"></i>
--- a/site/app/templates/polls/NewPollPage.twig
+++ b/site/app/templates/polls/NewPollPage.twig
@@ -110,7 +110,7 @@
                         <textarea aria-label="Reponse text" class="poll_response" name="response_{{response}}"style="resize: none; min-height: 50px; max-height: 300px; height: 50px; width: 80%;"
                         rows="10" cols="30">{{poll.getResponseString(response)}}</textarea>
                         <div class="move-btn up-btn">
-                            <i class="fa fa-lg fa-chevron-up move-arrow up"></i>
+                            <i class="fa fa-lg fa-chevron-up"></i>
                         </div>
                         <div class="move-btn down-btn">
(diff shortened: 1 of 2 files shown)


sbelsk (approved):
The new buttons looks good (on Windows/Chrome,Firefox) both in `Dark Mode` and `Dark Mode (Black)`. 
buttons
image

site/app/templates/polls/NewPollPage.twig
```diff
@@ -110,10 +110,10 @@
                         <textarea aria-label="Reponse text" class="poll_response" name="response_{{response}}"style="resize: none; min-height: 50px; max-height: 300px; height: 50px; width: 80%;"
                         rows="10" cols="30">{{poll.getResponseString(response)}}</textarea>
                         <div class="move-btn up-btn">
-                            <i class="move-arrow up"></i>
+                            <i class="fa fa-lg fa-chevron-up move-arrow up"></i>
```
MasterOdin, line 113: `move-arrow` and `up` are unused now, and can be removed:

```suggestion
                            <i class="fa fa-lg fa-chevron-up"></i>
```

site/app/templates/polls/NewPollPage.twig
```diff
@@ -110,10 +110,10 @@
...
                         </div>
                         <div class="move-btn down-btn">
-                            <i class="move-arrow down"></i>
+                            <i class="fa fa-lg fa-chevron-down move-arrow down"></i>
```
MasterOdin, line 116: Same here as with down
```suggestion
                            <i class="fa fa-lg fa-chevron-down move-arrow down"></i>
```

site/app/templates/polls/NewPollPage.twig
```diff
@@ -162,10 +162,10 @@
...
                 <div class="move-btn up-btn">
-                    <i class="move-arrow up"></i>
+                    <i class="fa fa-lg fa-chevron-up move-arrow up"></i>
                 </div>
                 <div class="move-btn down-btn">
-                    <i class="move-arrow down"></i>
+                    <i class="fa fa-lg fa-chevron-down move-arrow down"></i>
                 </div>
                 <div class="move-btn delete-btn">
                     <i class="fa fa-lg fa-trash delete"></i>
```
MasterOdin, line 171: Same here with delete:

```suggestion
                    <i class="fa fa-lg fa-trash"></i>
```

MasterOdin (changes requested):
Thanks for fixing this. Some small tweaks here, and then can be merged. On making changes to CSS files, especially on removing a class, we should make sure to clean-up any references to them left in the HTML.

site/app/templates/polls/NewPollPage.twig
williamjallen: Be careful.  I tried removing them and the JS for moving the fields up and down broke.  This file could use some serious refactoring but as it stands now, the JS targets classes so removing them is a problem.  The classes are being used as IDs here.

MasterOdin: I double checked these suggestions and nothing seemed broken? Only the `up-btn`, `down-btn`, and `delete-btn` classes are bound here. `move-arrow` is not referenced anywhere outside of these class definitions, and likewise the other classes.

williamjallen: Ok, looks good.

MasterOdin (approved):

PR #6447, opened 2021-05-18
[UI/UX:Forum] Fix forum post border in dark mode

### LINKED ISSUE #5874: Discussion Forum Shadows on Dark Mode
When using the normal color mode in Submitty, it uses shadows to visibly distinguish comments from each other. However, when using one of the Dark modes (normal or black), it removes these shadows (or they aren't visible) and the comments are no longer visibly separated.

Adding an outline to the comment boxes in dark mode (probably light grey or grey depending on which mode) to visibly separate the comment boxes.

Example:
darkmode
lightmode

eddiekrystowski:
Fixes #5874
Currently, there are no borders on discussion forum posts/threads in dark mode.

current-dark
current-dark(black)

### What is the new behavior?
Light gray borders have been added to each unimportant post in dark mode.

new-dark
new-dark(black)

### DIFF
--- a/site/public/css/forum.css
+++ b/site/public/css/forum.css
@@ -595,7 +595,7 @@
 }
 
 [data-theme="dark"] .post_box:not(.important) {
-    border:1px solid #bbb;
+    border: 1px solid #bbb;
 }
 
 .thread-box-full {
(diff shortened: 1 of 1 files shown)


MasterOdin (approved):
Thanks for fixing this. Changes work as expected.

PR #6449, opened 2021-05-18
[Bugfix:HelpQueue] Fix changing announcement breaking student page

### LINKED ISSUE #6448: Announcement update deletes page info
**Describe the bug**
Any time an announcement is edited, all of the important page elements on the student's side is deleted. This only happens if the announcement is being updated, as in it already exists. If an announcement was to be created for the first time, nothing would be deleted.

**Expected behavior**
Only for the announcement to be edited, nothing else should be deleted.

**To Reproduce**
1. Have at least one student page open on the Office Hours Queue.
2. Create an announcement, if one does not already exist.
3. Edit the announcement, either to something different or just to blank to remove it.
4. The student page will have all of the important page elements deleted now.

**Screenshots**
Before Announcment
After Announcement

**Additional context**
I believe the issue here is that the code is trying to delete the previous announcement element, however there is no previous announcement element since it is actually being replaced, thus the delete line just deletes the important information.

A3Digital:

Any time an announcement is updated, all student currently viewing the queue page have some of their page elements deleted.

### What is the new behavior?
The announcement is updated without deleting anything.
Fixes #6448 

### Other information?
Tested on Chrome and Firefox

### DIFF
--- a/site/app/templates/officeHoursQueue/QueueFooter.twig
+++ b/site/app/templates/officeHoursQueue/QueueFooter.twig
@@ -258,8 +258,7 @@
 
   function updateAnnouncement() {
       $.get("{{base_url}}/new_announcement", function(newAnnouncement) {
-          if ($('#announcement').length > 0){
-              $('#announcement').next().remove();
+          if ($('#announcement').length > 0) {
               $('#announcement').replaceWith(newAnnouncement);
           }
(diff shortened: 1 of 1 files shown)


MasterOdin (approved):
Looking at the git history, seems like that line has always been buggy, and so not sure why it was there in the first place. Oh well, thanks for fixing.

PR #6453, opened 2021-05-18
[Feature:System] Secondary Email Address

### LINKED ISSUE #6428: Secondary email address
Currently we only store 1 email address per user.  This email address is probably the students official school email.  Students can view, but cannot edit this email address from the "My Profile" page.
There are many reasons it would be convenient to allow students to specify a secondary email.  This secondary address should be optional, it can be specified/changed/deleted by the user.

Let's:
* Add two columns to the databases -- both the master & courses database, database trigger will need to be updated to keep data synchronized.  A string for the new/secondary email and a boolean for notify secondary email.
* Label this original email "Primary Email" in all places on the UI.  

On the "My Profile" page:
* Add the new email, labeled "Secondary Email".
* Allow users to edit their "Secondary Email" address (they still may not edit the primary), do some simple valid email string checks on this string data.
* Add a checkbox "Send email notifications to both Primary and Secondary Email Addresses".  By default this box is not checked.  When not checked, we only sent to primary.  It's not possible to opt out of emails to primary address.

From the "Manage Students" or "Manage Graders" pages:
* an instructor user may use "New" or "Edit" on another user and edit both the Primary and Secondary email addresses.
* Downloading/Importing users from .csv will need to change to include this extra column.

Note: The secondary address is by default empty string (or NULL -- let's decide).  The primary address may also be NULL (admin or testing accounts  often have empty/NULL email).  

And:
* Update the notifications functions that write to the emails database table to add both email addresses on the "To" line when requested.  I think it's better in this case to send 1 email with both addresses on the to line, instead of 2 separate emails.  We may need to modify the emails database table -- an additional column?  Or perhaps just store a string with 2 comma separate emails?
* Update the script that reads the emails database to put 2 emails on the To line.  Note, even if we just have a string with a comma, we will need to adjust the rate limit computation (emails sent per minute).  Sending 1 email to 2 people is the same as sending 2 emails to 1 person each.

cjreed121:

Right now, Submitty holds the default email address and will send email there only.
Closes #6428 

### What is the new behavior?
Submitty will allow users to input a secondary email address which can be used to receive Submitty messages in addition to their primary address. They will have the option to turn on and off the ability to receive email to their secondary address. Instructors will be able to see this secondary email address. All Submitty users can update their secondary email settings on their profile.


sbin/send_email.py
```diff
@@ -132,14 +133,12 @@ def construct_mail_client():
...
-    result = db.execute(
-        "SELECT emails.id, emails.user_id, users.user_email, emails.subject," +
-        " emails.body FROM emails INNER JOIN users ON" +
-        " emails.user_id = users.user_id WHERE" +
-        " emails.sent is NULL AND emails.error = ''" +
-        " ORDER BY id LIMIT 100;")
-
+    """Get an active queue of internal emails waiting to be sent."""
+    query = """SELECT id, user_id, email_address, subject, body FROM emails
+    WHERE email_address SIMILAR TO '%%@(%%.{}|{})' AND sent is NULL AND
```
shailpatels, line 138: This query should be prepared differently, a user could enter a string that breaks the query if they add single quotes to `EMAIL_INTERNAL_FORMAT`, not sure if its possible to trigger a SQL injection issue since the strings are expanded inside a similar to clause but it might be possible so generally, we should avoid constructing queries via concatenation and use parameterized queries which offload the query construction to sqlalchemy in this case.

sbin/send_email.py
```diff
@@ -153,6 +152,32 @@ def get_email_queue(db):
     return queued_emails
 
 
+def get_external_queue(db, num):
+    """Get an active queue of external emails waiting to be sent."""
+    query = """SELECT COUNT(*) FROM emails WHERE sent >= (NOW() - INTERVAL '1 hour') AND
+    email_address NOT SIMILAR TO '%%@(%%.{}|{})'"""
+    query = query.format(EMAIL_INTERNAL_FORMAT, EMAIL_INTERNAL_FORMAT)
```
shailpatels, line 159: same comments here as above

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -298,6 +299,7 @@ def get_input(question, default=""):
                 EMAIL_SERVER_PORT = int(get_input("What is the email server port?", defaults['email_server_port']))
             except ValueError:
                 EMAIL_SERVER_PORT = defaults['email_server_port']
+            EMAIL_INTERNAL_FORMAT = get_input("What is the internal email address format?", defaults['email_internal_format'])
```
shailpatels, line 302: I would suggest changing this to `EMAIL_DOMAIN` see comments below

site/app/controllers/UserProfileController.php
```diff
@@ -166,4 +167,36 @@ public function changeProfilePhoto() {
             }
         }
     }
+
+    /**
+     * @Route("/user_profile/change_secondary_email", methods={"POST"})
+     * @return JsonResponse
+     * @throws \ImagickException
```
shailpatels, line 174: why would this throw a `\ImagickException` ?

site/app/models/User.php
```diff
@@ -88,8 +92,12 @@ class User extends AbstractModel {
     protected $preferred_last_name = "";
     /** @prop @var  string The last name to be displayed by the system (either last name or preferred last name) */
     protected $displayed_last_name;
-    /** @prop @var string The email of the user */
+    /** @prop @var string The primary email of the user */
     protected $email;
+    /** @prop @var string The secondary email of the user */
+    protected $secondary_email;
+    /** @prop @var string The secondary email of the user */
```
shailpatels, line 99: This comment needs to be updated for `$email_both`

site/app/models/User.php
```diff
@@ -485,6 +495,15 @@ public static function validateUserData($field, $data): bool {
                 // validate email address against email RFCs
                 $validator = new EmailValidator();
                 return $validator->isValid($data, new RFCValidation());
+            case 'user_email_secondary':
```
shailpatels, line 498: Does this do the same thing as the above check? If so you can do

```php
case 'user_email':
case 'user_email_secondary':

...

return ...

```

site/app/templates/EditSecondaryEmailForm.twig
```diff
@@ -0,0 +1,43 @@
+{% extends 'generic/Popup.twig' %}
+{% block popup_id %}edit-secondary-email-form{% endblock %}
+{% block title %}Edit Secondary Email{% endblock %}
+{% block body %}
+    <p>Please fill in below to change your secondary email and configure whether or not you would like
+    us to send emails to it in addition to your normal email address.</p>
+    <div class="flex-col" id="user_settings_modal">
```
shailpatels, line 7: CSS Ids and classes should use dashes 
https://submitty.org/developer/coding_style_guide/css

site/app/templates/UserProfile.twig
```diff
@@ -28,6 +28,12 @@
                             <i class="fas fa-pencil-alt"></i> Edit Preferred Names
                         </a>
                     </span>
+                    <span>&nbsp;&nbsp;&nbsp;</span>
+                    <span>
+                        <a class="icon" tabindex="0" href="javascript:showUpdateSecondaryEmailForm()" aria-label="Edit user secondary email" title="Edit user secondary email">
```
shailpatels, line 33: This should probably be a button, `a` tags are reserved for navigation and buttons for everything else. Currently, you can not tab + open the modal with the keyboard which is an accessibility issue. Since the edit preferred names modal is setup the same way this can be done in a separate PR to fix both at the same time

site/public/js/user-profile.js
```diff
@@ -121,6 +128,41 @@ function updateUserProfilePhoto () {
...
+    else {
+        let data = new FormData();
+        data.append('csrf_token', $("#user-secondary-email-csrf").val());
+        data.append('secondary_email', second_email.val());
+        data.append('secondary_email_notify', second_email_notify.get(0).checked);
+        let url = buildUrl(['user_profile', 'change_secondary_email']);
+        $.ajax({
+            url,
+            type: "POST",
+            data,
```
shailpatels, line 146: I think on the success you can just refresh the page and let the twig repopulate the new fields instead of manually editing them, if we update the form in the future we would need to remember to update this logic as well. On error you can display the error message.

Also using the fetch API would clean up some of the code
```js
let response = await fetch(url , {
    method : 'POST',
    headers: {
      'Accept': 'application/json',
      'Content-Type': 'application/json'
    },
   body : { csrf_token : window.csrfToken, secondary_email .... }
});

response.json().then((data) => {
 ...
}).catch((err) => { });
```

site/public/js/userform.js
```diff
@@ -146,41 +146,49 @@ function checkValidEntries() {
...
-            setRedOrTransparent(input,valid_expression);
+            setRedOrTransparent(input, valid_expression);
             break;
-        }
+        case "user_email_secondary":
+            if (input.val() == '') {
+                input.css("background-color", "transparent");
+                break;
+            }
+            var valid_expression = /^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,253}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,253}[a-zA-Z0-9])?)*$/;
```
shailpatels, line 188: Both these cases appear to be doing the same thing so you can combine cases 

```js
case "user_email":
case "user_email_secondary" :
...
```

site/public/js/user-profile.js
```diff
@@ -121,6 +128,41 @@ function updateUserProfilePhoto () {
...
+            processData: false,
+            contentType: false,
```
shailpatels, line 148: Alternatively, you can set the form action to be the update secondary email endpoint and have the controller redirect the user to the current page since I don't think avoiding a refresh is super critical in this operation

site/app/controllers/UserProfileController.php
```diff
@@ -166,4 +167,36 @@ public function changeProfilePhoto() {
...
+        $user = $this->core->getUser();
+
+        if (!empty($_POST['secondary_email']) && !empty($_POST['secondary_email_notify'])) {
+            $secondaryEmail = trim($_POST['secondary_email']);
+            $secondaryEmailNotify = trim($_POST['secondary_email_notify']);
+
+            if ($user->validateUserData('user_email_secondary', $secondaryEmail) === true) {
+                $user->setSecondaryEmail($secondaryEmail);
+                $user->setEmailBoth($secondaryEmailNotify === "true");
+                $user->setUserUpdated(true);
```
shailpatels, line 186: Reading the docstring for the `user_updated` field

>This flag is set TRUE when a user edits their own preferred firstname.  When TRUE, preferred firstname
     *           is supposed to be locked from changes via student auto feed script.  Note that auto feed is still
     *           permitted to change (correct?) a user's legal firstname/lastname and email address.
     
    It sounds like it should only be set to true for when a user updates their preferred name, unless we also want a similar behavior for a secondary email which I don't think so since that information does not come from the autofeed source and can't clobber anything

site/public/js/user-profile.js
```diff
@@ -121,6 +128,41 @@ function updateUserProfilePhoto () {
...
+            success: function(res) {
+                const response = JSON.parse(res);
```
shailpatels, line 150: Entering a bad email address doesn't give any feedback to the user and only closes the modal, an error should be displayed if an invalid email is given or something else goes wrong

shailpatels (changes requested):

migration/migrator/migrations/master/20210517163053_secondary_email_address.py
```diff
@@ -0,0 +1,26 @@
...
+def up(config, database):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    database.execute("ALTER TABLE users ADD COLUMN IF NOT EXISTS user_email_secondary character varying(255) NOT NULL DEFAULT '';")
```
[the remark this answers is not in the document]
cjreed121: One of the migrations is for master while one is for course. I kept them pretty much identical so the sync user function would work well.

migration/migrator/migrations/master/20210518145110_trigger_update_user_second_email.py
```diff
@@ -0,0 +1,65 @@
...
+            END IF;
+            PERFORM dblink_exec(db_conn, query_string);
+        END LOOP;
+
+        -- All done.
+        RETURN NULL;
+    END;
+    $$ LANGUAGE plpgsql;"""
+    database.execute(sql)
+
```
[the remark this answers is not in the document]
cjreed121: This function already exists currently but I needed to modify it to support secondary email. Now it does but I have to overwrite the function here.

sbin/send_email.py
```diff
@@ -132,14 +133,12 @@ def construct_mail_client():
...
-        " emails.body FROM emails INNER JOIN users ON" +
-        " emails.user_id = users.user_id WHERE" +
-        " emails.sent is NULL AND emails.error = ''" +
-        " ORDER BY id LIMIT 100;")
-
+    """Get an active queue of internal emails waiting to be sent."""
+    query = """SELECT id, user_id, email_address, subject, body FROM emails
+    WHERE email_address SIMILAR TO '%%@(%%.{}|{})' AND sent is NULL AND
+    error = '' ORDER BY id LIMIT 100;"""
+    query = query.format(EMAIL_INTERNAL_FORMAT, EMAIL_INTERNAL_FORMAT)
```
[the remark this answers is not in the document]
cjreed121: I will call it email internal domain. I think it's important to say that this is a different rate limit. No matter what I will need to do string checks on the email address to check if it belongs to the internal address domain. You are thinking it would be better to do these string checks in the PHP?

site/app/controllers/UserProfileController.php
cjreed121: That's not supposed to be there. Will remove.

site/app/models/User.php
cjreed121: Yes it should do the exact same thing. I didn't think of doing it that way. I will change it to that.

site/app/templates/EditSecondaryEmailForm.twig
cjreed121: Yes I know Ids should use dashes. The reason I used specifically that is because the other two pop up forms in user profile used the exact same id. I kept it so I could keep any formatting already there. Should I change this for the others?

site/app/templates/UserProfile.twig
cjreed121: Alright. I'll create an issue for this and do that after this gets merged.

site/public/js/user-profile.js
cjreed121: One issue I see with this is if you refresh the page you won't be prompted with a success message but rather just a reload. You would see the fields updated but I think the success message is nice to have. Thoughts?

In terms of using the fetch API. I see that it could clean up some of the code but everywhere else in this file also uses Ajax. I would prefer to keep Ajax here to keep it consistent overall in this file.

site/public/js/user-profile.js
cjreed121: I will add an error field to the Ajax to handle this.

migration/migrator/migrations/master/20210517163053_secondary_email_address.py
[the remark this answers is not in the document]
shailpatels: Oh ok, I understand

sbin/send_email.py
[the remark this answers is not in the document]
shailpatels: I was suggesting to avoid doing any string comparisons, since the secondary email is its own property we could include a flag in the emails table to denote if the outgoing email is sending to a primary or secondary email.

site/app/templates/EditSecondaryEmailForm.twig
shailpatels: If you need the id to be a specific way to work with existing code then its fine for now but new code should try and use our style guides even older code does not follow it on the same pages.

site/public/js/user-profile.js
shailpatels: You can use the `$this->core->addSuccessMessage("...");` function which will add a green success banner when the page loads, its the PHP equivalent to the JS function you used.

sbin/send_email.py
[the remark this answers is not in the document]
cjreed121: We can't necessarily assume that a primary address is internal. No matter what we will have to do string comparisons to see if we are sending to an internal or external address.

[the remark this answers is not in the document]
shailpatels: I see, reading the issue I might of misunderstood the design goals and thought that secondary email implied external

shailpatels (changes requested):
There appears to be an issue creating a forum announcement,
creating a new thread and selecting the 'pin thread w/ email & notification' select box causes a red banner with the error message "Error parsing data. Please try again."

There is no additional information in the console, my notification settings were set to send myself a notification and email whenever I do the above actions

cjreed121:
> There appears to be an issue creating a forum announcement,
> creating a new thread and selecting the 'pin thread w/ email & notification' select box causes a red banner with the error message "Error parsing data. Please try again."
> 
> There is no additional information in the console, my notification settings were set to send myself a notification and email whenever I do the above actions

@shailpatels I have no issue creating a forum announcement on my system. I also made sure to turn on sending notifications to myself. I could not re-produce this error you are saying you got. Maybe try running a submitty install? That would happen if the database wasn't migrated.

sbin/send_email.py
cjreed121: Yeah I see the possibility of SQL injection here although it would be from sysadmin. Are there any good examples of parameterized queries in the code for sqlalchemy? I looked for docs on this but nothing was too helpful for getting this to work.

site/public/js/user-profile.js
cjreed121: Personally I feel the UI feels cleaner with the Ajax request rather than a full reload from a form request on a pop up. I think keeping it consistent with how preferred name would be good. Why do you think switching to the twig render is better?

HenrikLam (commented):
Seems to work as intended, couldn't reproduce the bug Shail mentioned.

sbin/send_email.py
```diff
@@ -132,14 +133,12 @@ def construct_mail_client():
...
-        "SELECT emails.id, emails.user_id, users.user_email, emails.subject," +
-        " emails.body FROM emails INNER JOIN users ON" +
-        " emails.user_id = users.user_id WHERE" +
-        " emails.sent is NULL AND emails.error = ''" +
-        " ORDER BY id LIMIT 100;")
-
+    """Get an active queue of internal emails waiting to be sent."""
+    query = """SELECT id, user_id, email_address, subject, body FROM emails
+    WHERE email_address SIMILAR TO :format AND sent is NULL AND
+    error = '' ORDER BY id LIMIT 100;"""
```
MasterOdin, line 139: Have you done any performance measuring on this query to see what the effects of adding `SIMILAR TO` would do to performance?

cjreed121: I just created a script to run the old one vs the new one 100 times and average the results. It seems this new script with the SIMILAR TO takes twice as long to run the query. The alternative would be to create another column will a boolean if this is an internal or external address and check the string in PHP. What do you think I should do here? Or do you have a different idea?

MasterOdin: When doing and discussing stress testing, it's important to provide concrete numbers in all this. How big was the table that you used when running the queries? Have you tried at different sizes so like 1000 rows, 10000 rows, 10,000,000 rows, etc., so that we measure both sizes that are within the ballpark of what Submitty might be facing, but also significantly more so we understand how this might scale.

When you say that the average time doubled, what does that mean? Did it double from taking 10ms to 20ms or 30s to 60s or 1.5min to 3min, etc.

cjreed121: Alright I went into more depth with my testing. Here are the results:

1,000 emails: 0.00035 normal and 0.00070 advanced
2,000 emails: 0.00070 normal and 0.0030 advanced
10,000 emails: 0.00065 normal and 0.00120 advanced
100,000 emails: 0.00070 normal and 0.0013 advanced

These results are in seconds. It seems the time stays about the same the more entries I put in. There is of course a limit of 100 so that probably is doing it. It seems that the advanced query takes about double the time but they are both a really small amount of time.

sbin/send_email.py
```diff
@@ -190,12 +212,18 @@ def send_email():
...
     db, metadata = setup_db()
     queued_emails = get_email_queue(db)
-    if len(queued_emails) == 0:
-        return
     mail_client = construct_mail_client()
 
     success_count = 0
 
+    if "'" in EMAIL_INTERNAL_DOMAIN:
+        return
```
MasterOdin, line 220: What's the purpose of this check? Attempting to block a SQL injection?

site/app/controllers/UserProfileController.php
```diff
@@ -10,6 +10,7 @@
 use app\libraries\response\RedirectResponse;
 use app\libraries\response\WebResponse;
 use app\views\grading\ImagesView;
+use phpDocumentor\Reflection\Types\Boolean;
```
MasterOdin, line 13: remove?

site/app/controllers/UserProfileController.php
```diff
@@ -166,4 +167,34 @@ public function changeProfilePhoto() {
             }
         }
     }
+
+    /**
+     * @Route("/user_profile/change_secondary_email", methods={"POST"})
+     * @return JsonResponse
+     */
+    public function changeSecondaryEmail() {
```
MasterOdin, line 175:
```suggestion
    public function changeSecondaryEmail(): JsonResponse {
```

site/public/js/user-profile.js
```diff
@@ -121,6 +128,50 @@ function updateUserProfilePhoto () {
...
+                if (response.status === "success") {
+                    const { data } = response;
+                    displaySuccessMessage(data.message);
+                    $('#secondary-email-row .value').text(data.secondary_email);
+                    $('#secondary-email-notify-row .value').text(data.secondary_email_notify);
+                    second_email.data('current-second-email', data.secondary_email);
+                    second_email_notify.data('current-second-email-notify', data.secondary_email_notify === "True" ? 1 : 0);
+                }
+                else if (response.status === "error") {
+                    displayErrorMessage("Some went wrong while updating secondary email address!");
```
MasterOdin, line 160: Typo:
```suggestion
                    displayErrorMessage("Something went wrong while updating secondary email address!");
```

Though, could we give a more meaningful error message in this case, like why updating the secondary email address failed (like bad validation, etc.)?

site/public/js/user-profile.js
```diff
@@ -121,6 +128,50 @@ function updateUserProfilePhoto () {
...
+                    second_email.val(second_email.data('current-second-email'));
+                    second_email_notify.prop('checked', second_email_notify.data('current-second-email-notify'));
+                }
+            },
+            error: function() {
+                // display error message
+                displayErrorMessage("Some went wrong while updating secondary email address!");
```
MasterOdin, line 167:
```suggestion
                displayErrorMessage("Something went wrong while updating secondary email address!");
```

site/public/js/user-profile.js
```diff
@@ -121,6 +128,50 @@ function updateUserProfilePhoto () {
...
+                    displaySuccessMessage(data.message);
+                    $('#secondary-email-row .value').text(data.secondary_email);
+                    $('#secondary-email-notify-row .value').text(data.secondary_email_notify);
+                    second_email.data('current-second-email', data.secondary_email);
+                    second_email_notify.data('current-second-email-notify', data.secondary_email_notify === "True" ? 1 : 0);
+                }
+                else if (response.status === "error") {
+                    displayErrorMessage("Some went wrong while updating secondary email address!");
+                    second_email.val(second_email.data('current-second-email'));
+                    second_email_notify.prop('checked', second_email_notify.data('current-second-email-notify'));
```
MasterOdin, line 162: Shouldn't reset the fields on error as that makes it harder for user to correct a typo or whatever they had wrong on submit.

MasterOdin (changes requested):

sbin/send_email.py
cjreed121: That was the idea before I added the parameterized queries. Will remove this.

site/app/controllers/UserProfileController.php
cjreed121: removed

site/public/js/user-profile.js
cjreed121: I have changed this to just simply display the error message the backend responds with.

site/public/js/user-profile.js
cjreed121: I was debating this myself. The reason I went for this was so it doesn't look like that is still saved in there. Maybe I should make a message appear saying UNSAVED changes without changing the actual data back. Thoughts?

sbin/send_email.py
MasterOdin: Alright, that's probably fine for performance, can re-address if we notice issues in production.

site/public/js/user-profile.js
MasterOdin: [`setCustomValidity`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity) may be a good option to show an error message tooltip on the input?

bmcutler (changes requested):
I've updated the text a bit.

Minor additional UI requests:

Can you make the box for entering the email a bit wider?  A 20 character wide email (not even that lengthy) causes the box to scroll horizontally -- at least on Mac Chrome.  Displaying 30+ characters without scrolling would be preferred.

Let's disable the checkbox when the secondary email is blank.  It doesn't make sense to enable the notifications to secondary if the box is blank.  Could also add that as a constraint on the database?  But not important.  I'm assuming the send email script is robust to blank primary and/or blank secondary email addresses -- I think we talked about that in a meeting?

Also, I echo Matt's feedback that the error handling is a little wonky.  It would be nice to disable the "Submit" button and have a simple red error text message on the page if the email string is invalid.

site/public/js/user-profile.js
cjreed121: I changed the input box to type email so the browser will not let you submit the form until that matches a valid email. I removed these lines as well since that really shouldn't be necessary.

bmcutler:
The checkbox can still be toggled when the secondary email is empty string.
But I'll go ahead and merge this anyway & open a new issue.

PR #6460, opened 2021-05-19
[Bugfix:HelpQueue] Fixed and Updated Notifications

### LINKED ISSUE #6237: Office Hours Queue doesn't get updated
**Describe the bug**
Office Hours queue notifications are enabled but when new students join the queue no notifications are shown and the queue is not even automatically updated to show the names of the newly added students. Those new names would only appear after I either forcefully reload the page or interact with some controls on the page (e.g., click the "Finish Helping" button for the student who is currently being helped).

**Expected behavior**
Notifications when enabled should appear immediately when a new student joins the queue. The queue should be automatically refreshed to immediately show the newly joined students.

**To Reproduce**
1. Navigate to the page: https://submitty.cs.rpi.edu/courses/s21/csci4380/office_hours_queue
2. Refresh the page until someone appears in the Office Hours Queue.
3. Click Help to start helping that student.
4. Another student who just joined the queue sends an email complaining that Submitty shows them that there are 0 people in front of them in the queue but they are not being helped. 
5. I have had no notifications in my browser from Submitty.
6. I check the Submitty page referenced above.
7. The page only shows the student being helped but not any of the newly added students.

**Configuration**
 - Browser: Firefox 85.0.2 (64-bit)

**Screenshots**
Here is the state of the queue when it doesn't show newly added students:
image

**Additional context**

A3Digital:
* [ x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)

Even with notifications enabled, a notification is never sent to the grader when a student joined the queue.

The current intention is that if notifications are enabled, any time a student joins a queue when the queue was previously empty, the grader would receive a notification.

Fixes #6237. Regarding this issue, the queue not updating was identified as most likely a bug in production that has since been fixed so this PR only addresses the fact that notifications are not working.

### What is the new behavior?
With notifications enabled, a notification is sent to the grader when a student is ready to be helped.

The new intention is that if notifications are enabled, any time the queue is updated and a student is ready to be helped when previously there was no student ready to be helped, the grader would receive a notification if they are not already helping a student.

To summarize, a student in the queue is considered ready to be helped if they are not paused and they are not already being helped. The reason for this distinction is that with the previous intention, if the only student in the queue was paused or if the only student in the queue was already being helped by another grader, then the notifications would effectively be disabled even though the graders are most likely waiting for a student to help.

### Other information?
Currently, the initial notification that is sent when notifications are enabled is a little wordy as it includes specifics as to when a notification will be sent. If the text could be shortened without sacrificing any information, this should most likely be done.


site/app/templates/officeHoursQueue/CurrentQueue.twig
```diff
@@ -23,19 +23,19 @@
                 <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
                 <input type="hidden" name="user_id" value="{{entry['user_id']}}"/>
                 {% if entry['paused'] %}
-                  <button type="submit" class="btn btn-primary help_btn" title="Student paused">
+                  <button type="submit" class="btn btn-primary help_btn" title="Student paused" ready="false">
```
MasterOdin, line 26: We should use `data-ready="false"` and `data-ready="true"` here. Generally, we want to avoid introducing non-standard attributes for a HTML tag, preferring to use the [data-* attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) to define custom data on elements.

MasterOdin (changes requested):

site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -149,13 +149,13 @@
       alert("This browser does not support desktop notification");
     }
     else if (Notification.permission === "granted") {
-      new Notification("Notifications have been enabled. You will be notified only when somebody joins the queue and there was nobody previously waiting for help.");
+      new Notification("Notifications have been enabled. You will be notified only if you are not helping a student and a student is now waiting for help when there was previously nobody waiting for help.");
```
sbelsk, line 152: Suggestions of making it shorter:
I think it's enough to say ```You will be notified when a student joins an empty queue```, but if we want to be specific with the wording and cover the edge case of paused students, we can maybe say ```You will be notified when a student joins a previously inactive queue.```

sbelsk (approved):
I tested this with multiple browsers on Windows/Chrome, Edge, Firefox and everything seems to work as described. I tested with paused students in the queue, removing/restoring students, and adding more students to the queue while helping a student.

site/app/templates/officeHoursQueue/QueueFooter.twig
A3Digital: I actually like the phrasing of "previously inactive queue" and I'll try to work that in there. 

Although, the reason I removed the phrase "when a student joins" is because the notification isn't only triggered when a student joins. It can also be triggered when a student un-pauses. This is why I changed that phrasing, so that it could be clear when the notification would be sent. But if you think this specificity is unnecessary and the previous phrasing works well, we can easily switch it back.

bmcutler:
@sbelsk Can you please make the minor wording change you suggested above and double check your review & approval?  Thanks

sbelsk (approved):

sbelsk:
@bmcutler Tested again, works as described.

Issue #6277, opened 2021-03-10
ta grading of notebook multiple choice grey is hard to read


bmcutler:
**What problem are you trying to solve with Submitty**
When ta grading a notebook gradeable, the multiple choice options are displayed in grey (the options are disabled since the TA isn't supposed to change the answers, of course).  But the grey is a bit hard to read. 

<img width="126" alt="Screen Shot 2021-05-17 at 10 11 00 AM" src="https://user-images.githubusercontent.com/5871417/118503006-3a3aaa80-b6f8-11eb-9176-4b502776cd05.png">

**Describe the way you'd like to solve this problem**
Let's change the CSS on the tagrading page for these multiple choice/radio buttons to make them really easy to see...  Some indication that they aren't editable would be good too..  [brainstorm/discuss?]

PR #6461, opened 2021-05-19
[UI/UX:TAGrading] Make disabled radio buttons more visible

sbelsk:
Fixes https://github.com/Submitty/Submitty/issues/6277
Checked disabled radio buttons are hard to see in light mode, and even more so in dark mode. The original issue addressed the buttons on the Notebook/TA Grading page, but this exists on the student's View Poll page when the poll is closed and still visible.
Notebook:
before_light before_dark 
Student View Poll:
student_before_light 

### What is the new behavior?
All disabled radio buttons on Submitty will now be more visible in every color theme
Notebook:
notebook_after_light notebook_after_dark
Student View Poll:
after_light after_dark after_darkbalsk

### Other information?
Tested on Widnows/Chrome,Edge,Firefox. It worked well on Chrome and Edge, but was ineffective on Firefox. Needs to be tested on other browsers as well (like Safari, etc.). If any reviewer knows how to improve it, please do so :)

### DIFF
--- a/site/public/css/global.css
+++ b/site/public/css/global.css
@@ -180,3 +180,15 @@ footer .footer-separator {
 #loading-bar-percentage{
     margin-left: 10px;
 }
+
+input[type="radio"][disabled=""]:checked:before {
+    content: "";
+    display: block;
+    position: relative;
+    top: 3px;
+    left: 3px;
+    width: 7px;
+    height: 7px;
+    border-radius: 100%;
+    background: var(--standard-medium-dark-gray);
+}


site/app/templates/notebook/Notebook.twig
```diff
@@ -125,7 +125,7 @@
                         {% for idx in choices_indices %}
                             <label for="multiple_choice_{{ num_multiple_choice }}_{{ loop.index0 }}" class="markdown">
 
-                                <input type="checkbox" class="key_to_click" tabindex="0"
+                                <input type="checkbox" class="key_to_click" class="disabled-radio" tabindex="0"
```
williamjallen, line 128: Does this new class need to be removed?

williamjallen (approved):
Although this currently violates the DRY principle, it seems that this is indeed the best way to accomplish this task after some discussion.  The background is different on the two pages and these are the only two places on the site where disabled radio buttons are used so I think it is ok to have different CSS for each one in case we want to change the style of one or the other at some future point in time.

site/public/css/electronic.css
```diff
@@ -1377,6 +1377,17 @@ tr.is_publish {
     padding: 10px;
 }
 
+input[type="radio"][disabled=""]:checked:before {
```
MasterOdin, line 1380: Let's stick this definition [here](https://github.com/Submitty/Submitty/blob/1d4866f3c88c333d84db90468409161376f2a22e/site/public/css/server.css#L169) and then it only needs to be defined once across the entire site, in case we decide to use disabled inputs elsewhere.

MasterOdin (changes requested):

site/public/css/electronic.css
sbelsk: I initially had it as global, but in a discussion with Barb she said that for now we should apply this to only these two instances. Do you still think it is better to have it more globally?

shailpatels (approved):
Disabled radio buttons look better on light and dark mode.

There appears to be a bug in firefox where radio buttons are not styled correctly.
https://bugzilla.mozilla.org/show_bug.cgi?id=182589

so there's not much we can do about that

site/public/css/electronic.css
MasterOdin: Yeah sure, I guess we can cross the bridge of making it global if we add a third place that we have disabled inputs.

PR #6462, opened 2021-05-19
[Testing:Developer] Add CI workflow to test vagrant up

shailpatels:

* [ ] Documentation has been updated/added if relevant TODO: will add documentation once merged to see UI to manually launch this workflow, currently there is no way to trigger unless on default branch

Vagrant up needs to be tested manually

### What is the new behavior?
Adds a workflow to run a vagrant up weekly, once merged into master should have the option to directly run against a branch for developers editing setup scripts. 

Currently runs on macOS which comes with vagrant  up and appears Github doesn't support virtualization in a Linux Runner

### DIFF
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -16,7 +16,7 @@ env:
 
 jobs:
     vagrant-up:
-        runs-on: ubuntu-18
+        runs-on: ubuntu-18.04
         steps:
           - uses: actions/checkout@v2
           - name: Vagrant up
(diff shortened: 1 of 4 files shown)


MasterOdin:
Thanks for doing this. It looks like we may have issues with rate limiting for Docker images, but not totally sure, and so will monitor and then we can re-asses if it's a problem.

I did modify this PR so that it'll run nightly (instead of weekly) since at the moment we're seeing a number of merges per day or so and I'd rather we get a warning sign of a corrupted vagrant up sooner rather than later. I also added a test to do a minimal validation that vagrant up was successful since it looks like if the `vagrant up` fails, it'll die, but exit with error code `0`, rendering the thing a "success".

PR #6463, opened 2021-05-19
[Feature:TAGrading] display message if no panel is selected

### LINKED ISSUE #6368: More intuitive visuals when no panel is selected in TA grading UI
**Describe the bug**
The first time a user enters the TA grading interface, no panels are selected for display.  The single panel is just grey.  This is nonintuitive for new users.  Especially for peer graders who will have little/no training prior to first use.  Let's add a subtle (grey) "Click above to select a panel for display" or similar message in this area if nothing is currently selected for display in one or multiple of the panels.

Screen Shot 2021-04-27 at 4 52 25 PM

Screen Shot 2021-04-27 at 4 54 09 PM

loadedskiffer:
Fixes #6368
Currently when someone opens the ta grading page for the first time, a blank grey panel is displayed with no instructions.
This could be confusing to first time users.
pr1_1

### What is the new behavior?
Now there is a message displayed that tells users to click on the buttons above to get a panel to display. I wasn't able to get the message to display in multiple panels but first time users will only have 1 panel displayed anyway. The message appears and disappears based on if the user has an active panel or not.
pr1_2
pr1_3
pr1_4

### Other information?
Although it the message won't display in every panel, Barb told me to go ahead with this PR and then if someone else thinks they can tackle it later, they can.

### DIFF
--- a/site/app/views/grading/ElectronicGraderView.php
+++ b/site/app/views/grading/ElectronicGraderView.php
@@ -918,7 +918,7 @@ HTML;
 
             $return .= <<<HTML
                 <div class="panels-container">
-                <h3 id="panel-instructions">Click above to select a panel for display</h3>
+                    <h3 id="panel-instructions">Click above to select a panel for display</h3>
                     <div class="two-panel-cont">
                          <div class="two-panel-item two-panel-left active">
                             <div class="panel-item-section left-top"></div>
--- a/site/app/views/grading/ElectronicGraderView.php
+++ b/site/app/views/grading/ElectronicGraderView.php
@@ -918,6 +918,7 @@ HTML;
 
             $return .= <<<HTML
                 <div class="panels-container">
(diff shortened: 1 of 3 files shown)


cjreed121 (approved):

MasterOdin (approved):
Please try and follow our [style guides](https://submitty.org/developer/coding_style_guide/). For HTML, you want to indent lines following opening tags (like the `<div>` above the `<h3>`).

Unfortunately, that particular JS file is not automatically linted and has to be hand-maintained to follow the rules, and that string literals are not either in PHP. But the gist is that for your particular change, there is a space between keyword and parentheses for the body, that there's a space between the closing parenthesis for the body and the opening curly brace, and that the ending brace of the `if` and `else` are on their own line. Since the file is not automatically linted, I'll handle making these changes in PRs, but just something to be aware of.

PR #6470, opened 2021-05-20
[UI/UX:Plagiarism] Add gradeable name to breadcrumb path

williamjallen:
When viewing a gradeable in Lichen, the breadcrumb path does not update to reflect the current location.  This also means that there are no ways to get back to the previous page via the UI.

### What is the new behavior?
This update adds an additional location in the breadcrumb path to make it easier to get back to the previous page.

### DIFF
--- a/site/app/views/admin/PlagiarismView.php
+++ b/site/app/views/admin/PlagiarismView.php
@@ -88,7 +88,8 @@ class PlagiarismView extends AbstractView {
     }
 
     public function showPlagiarismResult($semester, $course, $gradeable_id, $gradeable_title, $rankings) {
-        $this->core->getOutput()->addBreadcrumb('Plagiarism Detection', $this->core->buildCourseUrl(['plagiarism']));
+        $this->core->getOutput()->addBreadcrumb('Plagiarism', $this->core->buildCourseUrl(['plagiarism']));
+        $this->core->getOutput()->addBreadcrumb($gradeable_title, $this->core->buildCourseUrl([$gradeable_title]));
         $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('codemirror', 'codemirror.css'));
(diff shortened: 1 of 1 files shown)


site/app/views/admin/PlagiarismView.php
```diff
@@ -88,7 +88,8 @@ public function plagiarismMainPage($semester, $course, $gradeables_with_plagiari
     }
 
     public function showPlagiarismResult($semester, $course, $gradeable_id, $gradeable_title, $rankings) {
-        $this->core->getOutput()->addBreadcrumb('Plagiarism Detection', $this->core->buildCourseUrl(['plagiarism']));
+        $this->core->getOutput()->addBreadcrumb('Plagiarism', $this->core->buildCourseUrl(['plagiarism']));
+        $this->core->getOutput()->addBreadcrumb($gradeable_title, $this->core->buildCourseUrl([$gradeable_title]));
```
shailpatels, line 92: I believe links are only added to all breadcrumbs except the last most since that indicates the page you're currently on so you can removed the `$this->core->buildCourseUrl([$gradeable_title])` on ln 92 since it doesn't add a link:

image

shailpatels (changes requested):

site/app/views/admin/PlagiarismView.php
williamjallen: How is it any different from before though?  It seemed to me as if you added things to the breadcrumb and then the breadcrumb generation engine linkified all but the last one.  If I removed the gradeable name, I believe that it would stop at "Plagiarism Detection" with no link and the user would be unable to return to the previous page.

shailpatels: I mean just the build course url part

williamjallen: Oh, I see.  I thought that was necessary to link it in case there was a breadcrumb path even further down.  I'll remove it.

shailpatels (approved):

PR #6471, opened 2021-05-20
[Bugfix:SubminiPolls] past polls now in ascending order

loadedskiffer:
The future polls section shows the polls with the furthest out poll at the top and the closest poll at the bottom.
pr2_2

### What is the new behavior?
The future polls are now in ascending order with the closest poll appearing first.
pr2_1

### Other information?
I tested by making a ton of poles with different dates and it works perfectly.
I had to comment out @method array getAnswers() in the poll model because it was saying getAnswers was already defined.
It had no effect on anything but it got rid of the error

### DIFF
--- a/site/app/models/PollModel.php
+++ b/site/app/models/PollModel.php
@@ -12,7 +12,7 @@ use app\libraries\DateUtils;
  * @method int getId()
  * @method string getName()
  * @method string getQuestion()
- ///* @method array getAnswers()
+ * @method array getAnswers()
  * @method array getUserResponses()
  * @method string|null getImagePath()
  */
--- a/site/app/models/PollModel.php
+++ b/site/app/models/PollModel.php
@@ -12,7 +12,7 @@ use app\libraries\DateUtils;
  * @method int getId()
  * @method string getName()
  * @method string getQuestion()
- * @method array getAnswers()
+ ///* @method array getAnswers()
  * @method array getUserResponses()
  * @method string|null getImagePath()
  */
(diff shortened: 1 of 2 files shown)


site/app/models/PollModel.php
```diff
@@ -12,7 +12,7 @@
  * @method int getId()
  * @method string getName()
  * @method string getQuestion()
- * @method array getAnswers()
+ ///* @method array getAnswers()
```
shailpatels, line 15: should this be deleted?

sbelsk (approved):
I tested this on Windows/Chrome and the feature seems to be working fine. It's probably just me not knowing enough php to tell what the issue was with the function documentation in `PollModel.php`, but I was also wondering why they had a line there for the function `getImagePath()` which does not have a definition.

MasterOdin:
@sbelsk All models inherit from the `AbstractModel` class which in turn defines a so-called "magic function" ([`__call`](https://www.php.net/manual/en/language.oop5.overloading.php#object.call)). This function is called anytime you make a reference to a function on a model and there is no existing matching function, e.g., if you do `$poll->getImagePath()`, it will first check for a function `getImagePath` and when that is not found, it will run `__call`. This function then provides an easy wrapper around getting/setting property variables that are marked with the `@prop` annotation. So, by doing `/** @prop */ $image_path`, the `__call` magic method is mapped to provide `getImagePath` and `setImagePath($path)`.

More details in the docs at https://submitty.org/developer/developing_the_php_site/models.

PR #6472, opened 2021-05-20
[Bugfix:SubminiPolls] Format dates to match site

loadedskiffer:
Dates for future and past polls appear in YYYY-MM-DD format instead of MM/DD/YYYY which is used by the rest of the site.
pr3_3
pr3_4

### What is the new behavior?
Dates for future and past polls appear in MM/DD/YYYY format.
pr3_1
pr3_2

### Other information?
I added a lot of poles with different dates and it works every time. The date is still stored in the database in YYYY-MM-DD format because I think that is the only way Postgres stores date time objects, I just changed the way they're retrieved and displayed.


site/app/controllers/PollController.php
```diff
@@ -129,7 +129,7 @@ public function addNewPoll() {
                 new RedirectResponse($this->core->buildCourseUrl(['polls']))
             );
         }
-        $date = \DateTime::createFromFormat("Y-m-d", $_POST["release_date"]);
+        $date = \DateTime::createFromFormat("m/d/Y", $_POST["release_date"]);
```
shailpatels, line 132: We can pull a date config from the config so it works internationally and so we have the option to change all the formatting in 1 place in the future if needed.

E.g:
```php
$this->core->getConfig()->getDateTimeFormat()->getFormat('gradeable')
``` 
You can see a list of the different date formats in the file `site/app/models/DateTimeFormat.php` and choose whichever works, `gradeable` is a common one that is used in the site, if nothing works you can add a specific format for 'polls' and use that instead.

shailpatels (changes requested):

site/app/controllers/PollController.php
loadedskiffer: are you saying I should replace  $date = \DateTime::createFromFormat("m/d/Y", $_POST["release_date"]); with 
$date = \DateTime::createFromFormat($this->core->getConfig()->getDateTimeFormat()->getFormat('gradeable'), $_POST["release_date"]); ?

shailpatels: Yeah to clarify `$this->core->getConfig()->getDateTimeFormat()->getFormat('gradeable')` returns a date format (its type is a string) so you can use that wherever you have a hardcoded value, or pass it from PHP to twig. If the gradeable format doesn't work for you, you can pick a different one or create one specific for polls.

loadedskiffer: Ok thanks I didn't realize that, I'll go back and make some changes.

site/app/templates/polls/NewPollPage.twig
```diff
@@ -142,7 +142,7 @@
 <script>
     flatpickr('.poll-date', {
         allowInput: true,
-        dateFormat: "Y-m-d",
+        dateFormat: "m/d/Y",
```
shailpatels, line 145: You can pass the date format from the view file, in line 51 of `site/app/views/PollView.php`
```php
return $this->core->getOutput()->renderTwigTemplate("polls/NewPollPage.twig", [
            'csrf_token' => $this->core->getCsrfToken(),
            'base_url' => $this->core->buildCourseUrl() . '/polls',
            'date_format' => $this->core->getConfig()->getDateTimeFormat()->getFormat('poll')
          ]);
```

Then in the NewPollPage.twig you can access it by {{ date_format }}

Here's a quick overview of Twig templating if you're not familiar 
https://twig.symfony.com/doc/3.x/templates.html

The connection between controllers, models, views, and twig templates might be confusing at first so feel free to ask in slack if anything isn't clear.

shailpatels (changes requested):

shailpatels (approved):

PR #6474, opened 2021-05-21
[Feature:Submission] Gradeable Timer

### LINKED ISSUE #1865: Add "current time on server" dynamic clock to submission page

cjreed121:

Currently gradeables will only show the due date. Additionally if a student is taking a timed gradeable, they must time themselves.
Closes #1865 

### What is the new behavior?
The gradeable page now has a timer built on page. There is a timer for a count down until the normal gradeable due date if it has a due date. If the gradeable has a timer autograder configuration, a timer will appear counting down until you should submit your gradeable.
image

### Other information?
This is still a WIP but reviews could be helpful
There will be a separate PR to move autograder config.json data for timer into DB

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -2,6 +2,7 @@
 
 namespace app\libraries\database;
 
+use _HumbugBoxec8571fe8659\Nette\Utils\DateTime;
 use app\exceptions\DatabaseException;
 use app\exceptions\NotImplementedException;
 use app\exceptions\ValidationException;
@@ -4427,6 +4428,7 @@ AND gc_id IN (
               g_grade_locked_date AS grade_locked_date,
               g_min_grading_group AS min_grading_group,
               g_syllabus_bucket AS syllabus_bucket,
+              g_allowed_time AS allowed_time,
               eg.*,
               gc.*,
               (SELECT COUNT(*) AS cnt FROM regrade_requests WHERE g_id=g.g_id AND status = -1) AS active_regrade_request_count
@@ -4518,6 +4520,7 @@ AND gc_id IN (
 
             // Finally, create the gradeable
             $gradeable = new \app\models\gradeable\Gradeable($this->core, $row);
+            $gradeable->setAllowedTimeOverrides($this->getGradeableTimeOverride($gradeable->getId()));
 
             // Construct the components
             $component_properties = [
@@ -7175,5 +7178,10 @@ SQL;
       END) AS not_helped_count
 
 SQL;
+    }
+
+    private function getGradeableTimeOverride(string $gradeable_id): array {
+        $this->course_db->query('SELECT * FROM gradeable_allowed_time_override WHERE g_id=?', [$gradeable_id]);
+        return $this->course_db->rows();
     }
 }
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -4428,7 +4428,7 @@ AND gc_id IN (
               g_grade_locked_date AS grade_locked_date,
               g_min_grading_group AS min_grading_group,
               g_syllabus_bucket AS syllabus_bucket,
-              g_allowed_time AS allowed_time,
+              g_allowed_minutes AS allowed_minutes,
               eg.*,
               gc.*,
               (SELECT COUNT(*) AS cnt FROM regrade_requests WHERE g_id=g.g_id AND status = -1) AS active_regrade_request_count
@@ -4520,7 +4520,7 @@ AND gc_id IN (
 
             // Finally, create the gradeable
             $gradeable = new \app\models\gradeable\Gradeable($this->core, $row);
-            $gradeable->setAllowedTimeOverrides($this->getGradeableTimeOverride($gradeable->getId()));
+            $gradeable->setAllowedMinutesOverrides($this->getGradeableMinutesOverride($gradeable->getId()));
 
             // Construct the components
             $component_properties = [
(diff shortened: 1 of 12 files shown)


cjreed121:
In order to test this, the following must be done after pulling:

- Run Submitty install to run the database migrations

At this point the gradeable timer will work but not autograder configured timer until the following is done.

- Create a new gradeable in whichever course you would like (the timed config.json is not necessary right now)
- Connect to your course's database and open the gradeable table.
- Find the g_allowed_time column and give it an integer amount of minutes you would like all students to have

At this point the autograder configured timer should show up. This is calculated based off of the first time this gradeable has been opened by the specific user.

If you would like to give a specific user an override for time do the following:

- Connect to your course's database and open the gradeable_allowed_time_override table.
- Add a row where g_id is the gradeable id, user_id is the user's id, and allowed_time is the integer amount of minutes you would like this specific user to have on the gradeable

eddiekrystowski:
Both timers seem to work pretty well except for the fact that they are affected by changing the system time/timezone. Specifically with the gradeable timer, if you set the system clock so that the allotted time expires and then move it back, it will be stuck on "Past Due" until the page is refreshed. I'm not even sure what should happen in this case, since if you were doing that you're clearly aware you're messing with the timer. Also, if the progress bar goes yellow/red and then the system time is set back, the bar's progress will be moved backwards but the color remains the same. (staying yellow/red even if it should be green/yellow). In a way I think that is a good thing because it shows the user their antics don't actually affect the time left on their submission. If it is possible to a stop the timer from being manipulated by system time then these things wouldn't even be a problem. Other than that, I wasn't able to get anything to break.

The gradeable timer progress bar reacts well to resizing the page, and looks fine in all themes.

cjreed121:
> Both timers seem to work pretty well except for the fact that they are affected by changing the system time/timezone. Specifically with the gradeable timer, if you set the system clock so that the allotted time expires and then move it back, it will be stuck on "Past Due" until the page is refreshed. I'm not even sure what should happen in this case, since if you were doing that you're clearly aware you're messing with the timer. Also, if the progress bar goes yellow/red and then the system time is set back, the bar's progress will be moved backwards but the color remains the same. (staying yellow/red even if it should be green/yellow). In a way I think that is a good thing because it shows the user their antics don't actually affect the time left on their submission. If it is possible to a stop the timer from being manipulated by system time then these things wouldn't even be a problem. Other than that, I wasn't able to get anything to break.
> 
> The gradeable timer progress bar reacts well to resizing the page, and looks fine in all themes.

Thanks for testing this. System timezone should never change the timer since it uses UTC entirely. If you were able to get the timer to change while changing timezone please let me know as that should not happen. Yeah I see how the system clock can have an effect on the timer. I rely upon system time to create an accurate timer. I will have to see if there is any solution to system time changes.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2,6 +2,7 @@
 
 namespace app\libraries\database;
 
+use _HumbugBoxec8571fe8659\Nette\Utils\DateTime;
```
williamjallen, line 5: What is this for?

site/public/js/submission-page.js
```diff
@@ -7,3 +8,124 @@ function openActionsPopup(popup_css, element_id) {
...
+                curTime = data.current_time;
+                deadline = data.deadline;
+                updateTime();
+            }
+            else {
+                // eslint-disable-next-line no-undef
+                displayErrorMessage('Something went wrong while starting the timer');
+            }
+        },
+        error: function (err) {
```
williamjallen, line 60: What should be the expected behavior if the server cannot be contacted?  Should the progress bar continue moving forward?  Should the progress bar disappear?  Should some sort of message appear indicating that the user is now offline/unable to communicate with the server?  I'm personally in favor of the progress bar continuing based upon the system time and then displaying some sort of error message saying that the server can't be contacted.  This is also useful for other things such as the user turning off their internet or the server going down during a timed test.

site/public/js/submission-page.js
```diff
@@ -7,3 +8,124 @@ function openActionsPopup(popup_css, element_id) {
...
+function updateTime() {
+    if (Math.abs(Date.now() - lastTime) > 5000 || ticks_till_update <= 0) {
+        //we need to sync back up
+        syncWithServer();
+    }
+    else {
+        curTime += (Date.now()-lastTime);
+        ticks_till_update -= (Date.now()-lastTime);
+        lastTime = Date.now();
+        if (document.getElementById('gradeable-time-remaining-text') !== null) {
```
williamjallen, line 75: Any reason why you're using `document.getElementById()` instead of JQuery here?  (same thing for this whole function)

site/app/libraries/database/DatabaseQueries.php
cjreed121: Definitely not supposed to be there. Will remove that.

site/public/js/submission-page.js
cjreed121: There isn't necessarily a specific reason but there isn't much advantage or need to do so for this since I am simply getting elements by id with normal JavaScript.

site/public/js/submission-page.js
cjreed121: Right I forgot to finish the error handling part. I think I will add an argument into syncWithServer which will determine if this is a necessary call or not. Necessary means the first time its called or if the time was set off by more than 5 seconds. In this case I will have the timer error out since it could be off by a lot at this point or not even work. A non necessary call would just be the every 10 minute check which I would just continue counting. Once they connect to the network it will continue to work as normal.

williamjallen (approved):

bmcutler:
I tested this quickly.  Looks pretty good!

Layout comment (question?)
Maybe it's weird that the two timers separate when the screen is wide/font is small:
Screen Shot 2021-06-04 at 2 15 17 AM
But stack when the screen is narrow:
Screen Shot 2021-06-04 at 2 15 24 AM

Let's discuss the visual design details (color/size/shape) of the progress bar in the later PR -- 
perhaps the one that sets the allowed minutes in the DB from the config.json at gradeable build time.

bmcutler (approved):

PR #6479, opened 2021-05-21
[Bugfix:TAGrading] Added cookie to remember anonymous mode

loadedskiffer:
If a TA enables anonymous mode for a gradeable and then leaves the page, when they come back anonymous mode if off again. Anonymous mode is determined in the url eg. (anon_mode=1 or anon_mode=0).
pr4_2
pr4_3

### What is the new behavior?
A cookie is set when a TA enables anonymous mode so anonymous mode will always be turned on for that gradeable until they turn it off. The url is no longer used to convey information about anonymous mode.
pr4_1

### Other information?
I tested by  messing around with the gradeables in the sample course and toggling anonymous mode on and off. It worked for every assignment.

### DIFF
--- a/site/app/controllers/grading/ElectronicGraderController.php
+++ b/site/app/controllers/grading/ElectronicGraderController.php
@@ -1488,7 +1488,7 @@ class ElectronicGraderController extends AbstractController {
         $this->core->getOutput()->addInternalJs('grade-inquiry.js');
         $this->core->getOutput()->addInternalJs('websocket.js');
         $show_hidden = $this->core->getAccess()->canI("autograding.show_hidden_cases", ["gradeable" => $gradeable]);
-        $this->core->getOutput()->renderOutput(['grading', 'ElectronicGrader'], 'hwGradingPage', $gradeable, $graded_gradeable, $display_version, $progress, $show_hidden, $can_inquiry, $can_verify, $show_verify_all, $show_silent_edit, $late_status, $rollbackSubmission, $sort, $direction, $who_id, $solution_ta_notes, $submitter_itempool_map,$anon_mode_cookie);
+        $this->core->getOutput()->renderOutput(['grading', 'ElectronicGrader'], 'hwGradingPage', $gradeable, $graded_gradeable, $display_version, $progress, $show_hidden, $can_inquiry, $can_verify, $show_verify_all, $show_silent_edit, $late_status, $rollbackSubmission, $sort, $direction, $who_id, $solution_ta_notes, $submitter_itempool_map);
         $this->core->getOutput()->renderOutput(['grading', 'ElectronicGrader'], 'popupStudents');
         $this->core->getOutput()->renderOutput(['grading', 'ElectronicGrader'], 'popupMarkConflicts');
         $this->core->getOutput()->renderOutput(['grading', 'ElectronicGrader'], 'popupSettings');
--- a/site/app/controllers/grading/ElectronicGraderController.php
+++ b/site/app/controllers/grading/ElectronicGraderController.php
@@ -811,12 +811,10 @@ class ElectronicGraderController extends AbstractController {
             $this->core->addErrorMessage("You do not have permission to grade {$gradeable->getTitle()}");
             $this->core->redirect($this->core->buildCourseUrl());
         }
+        $anon_mode = false;
         if (array_key_exists('anon_mode', $_COOKIE)) {
             if ($_COOKIE['anon_mode'] === 'on') {
                 $anon_mode = true;
-            }
-            else {
-                $anon_mode = false;
             }
         }
         //Checks to see if the Grader has access to all users in the course,
--- a/site/app/controllers/grading/ElectronicGraderController.php
+++ b/site/app/controllers/grading/ElectronicGraderController.php
@@ -879,7 +879,6 @@ class ElectronicGraderController extends AbstractController {
                 }
             }
         }
-
         $show_all_sections_button = $can_show_all;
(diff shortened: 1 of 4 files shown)


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -806,6 +810,7 @@ public function detailsPage(Gradeable $gradeable, $graded_gradeables, $teamless_
             "team_gradeable_view_history" => $team_gradeable_view_history,
             "view_all" => $view_all,
             "anon_mode" => $anon_mode,
+            "anon_mode_cookie" => array_key_exists('anon_mode', $_COOKIE) ? $_COOKIE['anon_mode'] : 'off',
```
cjreed121, line 813: Couldn't you use only anon_mode since that is set based off of your cookie already?

cjreed121 (changes requested):

site/app/views/grading/ElectronicGraderView.php
loadedskiffer: The javascript needs access to the cookies value e.g const status = '{{ anon_mode_cookie }}. If I just passed anon_mode I wouldn't be able to update the cookie correctly.

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -517,7 +522,7 @@ public function detailsPage(Gradeable $gradeable, $graded_gradeables, $teamless_
             else {
                 $columns[]     = ["width" => "2%",  "title" => "",                 "function" => "index"];
                 $columns[]     = ["width" => "8%", "title" => "Section",          "function" => "section"];
-                if ($this->core->getUser()->getGroup() == User::GROUP_LIMITED_ACCESS_GRADER && $gradeable->getLimitedAccessBlind() == 2) {
+                if (($this->core->getUser()->getGroup() == User::GROUP_LIMITED_ACCESS_GRADER && $gradeable->getLimitedAccessBlind() == 2) || $anon_mode) {
```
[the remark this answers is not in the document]
loadedskiffer: you're right

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -443,6 +443,11 @@ public function detailsPage(Gradeable $gradeable, $graded_gradeables, $teamless_
         if ($gradeable->isPeerGrading() && $this->core->getUser()->getGroup() == User::GROUP_STUDENT) {
             $peer = true;
         }
+        if(array_key_exists('anon_mode', $_COOKIE)){
```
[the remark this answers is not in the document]
loadedskiffer: yup you're right

site/app/templates/grading/electronic/Details.twig
```diff
@@ -687,3 +686,19 @@
 {% endmacro %}
 
 {# /Column functions #}
+
+<script type="text/javascript">
```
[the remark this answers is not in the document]
loadedskiffer: when I put it in details.js I get an unused function warning and then it never gets called when the button is clicked so I'm not sure.

site/app/views/grading/ElectronicGraderView.php
cjreed121: You could pass anon_mode into your JavaScript instead though.

site/app/templates/grading/electronic/Details.twig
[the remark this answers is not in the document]
cjreed121: Missed the twig portion so this must remain in the twig file.

cjreed121:
You should also be able to remove the $anon_mode = false from the from details function argument since that takes in from url which is no longer necessary.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -811,7 +811,12 @@ public function showDetails($gradeable_id, $view = null, $sort = "id", $directio
             $this->core->addErrorMessage("You do not have permission to grade {$gradeable->getTitle()}");
             $this->core->redirect($this->core->buildCourseUrl());
         }
-
+        $anon_mode = false;
+        if (array_key_exists('anon_mode', $_COOKIE)) {
+            if ($_COOKIE['anon_mode'] === 'on') {
+                $anon_mode = true;
+            }
+        }
```
MasterOdin, line 819: We can make this a bit more succinct:
```suggestion
        $anon_mode = isset($_COOKIE['anon_mode']) && $_COOKIE['anon_mode'] === 'on';
```

While `array_key_exists` and `isset` are mostly the same, it's better to use `isset` unless you need some specific functionality of `array_key_exists` as it's a bit faster.

cjreed121 (approved):

site/app/templates/grading/electronic/Details.twig
```diff
@@ -687,3 +686,16 @@
 {% endmacro %}
 
 {# /Column functions #}
+<script>
+    function changeAnon() {
+        const status = '{{ anon_mode }}';
```
MasterOdin, line 691: Hm, when `anon_mode === true`, does this evaluate in twig to be `const status = '1'` and then for `anon_mode === false`, we get `const status = ''`? If possible, I think I would prefer doing something like  `const status = {{ anon_mode ? 1 : 0 }}` to make what's going on a bit less ambiguous if you don't know how PHP coerces boolean to strings.

MasterOdin (changes requested):
Small comment, just want to confirm what's going on, and then I think just slight tweak and good to merge.

site/app/templates/grading/electronic/Details.twig
loadedskiffer: Ok that makes sense, thanks

PR #6489, opened 2021-05-24
[Feature:SubminiPolls] Instructor Poll Preview

sbelsk:
Implements https://github.com/Submitty/Submitty/issues/6036

### What is the new behavior?
- [x] When editing or creating a new poll, the instructor can see the student view of the poll (The buttons in the popup modal are interactive but have no effect)
- [x] Attached images show up properly
- [x] Preview of the poll reflects the current values in the form without having to click "submit changes"

image

### DIFF
--- a/site/app/templates/polls/NewPollPage.twig
+++ b/site/app/templates/polls/NewPollPage.twig
@@ -334,7 +334,7 @@
         $('#poll-preview').css('display', 'block');
         $('#poll-preview-name').text($('#poll-name').val());
         $('#poll-preview-question').text($('#poll-question').val());
-        if ($('#poll-preview-image').attr('src') != "") {
+        if ($('#poll-preview-image').attr('src') != "placeholder/link") {
             $('#poll-preview-image-wrapper').show();
         }
         $('#poll-preview-answers').html("");
--- a/site/app/templates/polls/NewPollPage.twig
+++ b/site/app/templates/polls/NewPollPage.twig
@@ -52,6 +52,7 @@
             <span id="response-count-error" class="red-message polls-submit-error" hidden>Please make sure to add at least one response option.<br/></span>
             <span id="correct-response-count-error" class="red-message polls-submit-error" hidden>Please make sure to add at least one correct response.<br/></span>
             <button type="submit" id="poll-form-submit" class="btn btn-primary" disabled>Add Poll</button>
+            <button type="button" id="poll-preview-btn" class="btn btn-primary" onclick="previewPoll();" disabled>Preview</button>
         </form>
     {% else %}
         <h1> Edit Poll </h1>
@@ -134,6 +135,7 @@
             <span id="response-count-error" class="red-message polls-submit-error" hidden>Please make sure to add at least one response option.<br/></span>
             <span id="correct-response-count-error" class="red-message polls-submit-error" hidden>Please make sure to add at least one correct response.<br/></span>
             <button type="submit" id="poll-form-submit" class="btn btn-primary">Submit Changes</button>
+            <button type="button" id="poll-preview-btn" class="btn btn-primary" onclick="previewPoll();">Preview</button>
         </form>
     {% endif %}
 </div>
@@ -253,6 +255,7 @@
             $("#correct-response-count-error").show();
         } else {
             $("#poll-form-submit").prop("disabled", false);
+            $("#poll-preview-btn").prop("disabled", false);
             $(".polls-submit-error").hide();
         }
     }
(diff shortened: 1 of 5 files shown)


eddiekrystowski (approved):
Looks good! Nice work.

site/app/templates/polls/NewPollPage.twig
```diff
@@ -271,59 +274,92 @@
...
+        while ($("#response_"+count).length) {
+            $("#poll-preview-answers").append(`
+                <div class = "radio">
+                    <input type="radio" name="answer"/>
+                        <p>`+$("#response_"+count).val()+`</p>
+                    <br/>
+                    <br/>
+                </div>
+            `);
+            count++;
```
cjreed121, line 360: Could this be done with the actual student poll twig file rather than simulating one? This could allow for greater future maintainability so this doesn't need to be changed every time the student twig is changed.

site/app/templates/polls/PollPreview.twig
```diff
@@ -0,0 +1,31 @@
+{% extends 'generic/Popup.twig' %}
+{% block popup_id %}poll-preview{% endblock %}
+{% block title %}Student view of the poll:{% endblock %}
+{% block body %}
+<div>
+    <div class="content">
+        <h1 id="poll-preview-name">Poll Name</h1>
+        <div>
+            <p id="poll-preview-question"></p>
```
cjreed121, line 9: This does not support mark down like normal polls do. This is one of the benefits to switching to the normal student twig templates.

cjreed121 (changes requested):
I think this works well but I have left a few suggestions to better improve this feature.

site/app/templates/polls/NewPollPage.twig
sbelsk: I thought that this is what I would do initially, but the actual student poll twig file uses the saved values (php variables) of the poll object when rendered. Since I wanted the poll preview to update in real time when you update fields and values in the form (without having to click "submit changes"), I am updating the preview with JS, which is why I simulated the student view. If I just take the student twig file as is, it would not update in real time, and would not work at all when creating a new poll (I think that with the way the New Poll page is set up, a poll object is built only after you click the "submit" button which redirects you to the poll summary page. So when on the New Poll page, a poll doesn't exist until you click the submit, so the preview will have no data to display unless you go back and edit the poll you created). This might have been just my naive way of implementing this, so do you know of any better way of bypassing this? I agree that it would be ideal for this feature to share code for maintainability.

cjreed121: You could look into making an Ajax request that sends the question, responses, and image data to the server. The server in turn will render the Twig template and send back the HTML which you can load in your pop up.

MasterOdin: Yes, please just submit an AJAX request with the form data, and have it return the rendered twig template. As @cjreed121 indicates, doing it that way will keep the two pages in parity with each other, and just lower maintenance overhead. Otherwise, it's almost certainly the case we'll end up in the case where the student twig template gets updated for a bugfix and it's not carried over to this JS constructed modal.

sbelsk: I considered it, and came up with the following points:

- Using Ajax requests will require big modifications to the way polls work and will create other awkward problems. 
 For example, say a few fields in a form of an existing poll were changed, and Ajax requests have been sent to update the poll object's data in order for the poll preview would update in real time. Now, what if the user leaves the page without hitting the "submit changes" button? How would we restore the poll to the state it was before the changes? Save the whole history of a poll, or deal with saving a copy of every poll to handle this case?

- Using the actual student twig file would mean also disabling the fields in the form, since we don't want the user interacting with the preview having any effect. Manually disabling the fields on its own is something that would require maintenance in case other things would be added to the student page, so this does not really solve the problem you brought up, but rather moves it somewhere else.

- Finally, the polls feature in Submitty is supposed to be a rather simple feature (like its origin, the iclickers), so it is pretty mature right now and I don't think it will change much more in the future. In this case, possibly going through all the trouble I described above for maintainability is not worth it

cjreed121: Using these Ajax requests should not require a big modification. You won't actually have to store any of this data in the model. You would simply ask the server to render the form for you. You can create a controller function with an associated view that renders the form data you send in the Ajax. In terms of disabling the form, you could probably remove the submit button by adding a conditional in the Twig template for the student view of the poll. It is a pretty simple feature but if in a few months a change is made to the polls that may not follow to this leading instructors to think it looks one way while it looks a much different way.

shailpatels: Note you can also use the fetch API to send async requests which is a bit cleaner than AJAX, either is supported so its your preference on what's easier

PR #6496, opened 2021-05-25
[UI/UX:SubminiPolls] Make chevrons more visible in dark mode

### LINKED ISSUE #6452: Chevron Icons Hard to See in Dark Mode
**Describe the bug**
The chevron icons meant for collapsing or expanding the tables in the student's and instructor's Poll Summary page are always black which makes them hard to see under "Dark Mode", and impossible to see under "Dark Mode (Black)"
image
image

**Expected behavior**
The icons should turn white or black accordingly when the website theme is changed.

williamjallen:
The chevrons on the polls page are hard to see in dark mode.

### What is the new behavior?
Fixes #6452

Untitled

### DIFF
--- a/site/app/templates/polls/AllPollsPageInstructor.twig
+++ b/site/app/templates/polls/AllPollsPageInstructor.twig
@@ -10,7 +10,7 @@
     <hr>
 
     <div class="dropdown-wrapper">
-        <div onclick='updateDropdownStates($(this).find(".dropdown-arrow").hasClass("down"), "today_polls_dropdown", "{{ base_url }}")' class="dropdown-bar">
+        <div class="dropdown-bar">
             <h2> Today's Polls </h2>
             <div class="dropdown-btn">
                 <i class="fa fa-chevron-down"></i>
@@ -73,7 +73,7 @@
     <hr>
 
     <div class="dropdown-wrapper">
-        <div onclick='updateDropdownStates($(this).find(".dropdown-arrow").hasClass("down"), "old_polls_dropdown", "{{ base_url }}")' class="dropdown-bar">
+        <div class="dropdown-bar">
             <h2> Old Polls </h2>
             <div class="dropdown-btn">
                 <i class="fa fa-chevron-down"></i>
@@ -138,7 +138,7 @@
     <hr>
 
     <div class="dropdown-wrapper">
-        <div onclick='updateDropdownStates($(this).find(".dropdown-arrow").hasClass("down"), "future_polls_dropdown", "{{ base_url }}")' class="dropdown-bar">
+        <div class="dropdown-bar">
             <h2> Future Polls </h2>
             <div class="dropdown-btn">
                 <i class="fa fa-chevron-down"></i>
--- a/site/app/templates/polls/AllPollsPageInstructor.twig
+++ b/site/app/templates/polls/AllPollsPageInstructor.twig
@@ -13,7 +13,7 @@
         <div onclick='updateDropdownStates($(this).find(".dropdown-arrow").hasClass("down"), "today_polls_dropdown", "{{ base_url }}")' class="dropdown-bar">
             <h2> Today's Polls </h2>
             <div class="dropdown-btn">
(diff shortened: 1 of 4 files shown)


site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -13,7 +13,7 @@
         <div onclick='updateDropdownStates($(this).find(".dropdown-arrow").hasClass("down"), "today_polls_dropdown", "{{ base_url }}")' class="dropdown-bar">
```
sbelsk, line 13: Should this line be updated, now that `dropdown-arrow` has been removed? Same comment for line 76 and 141 in this file.

williamjallen: I believe that whole onclick event was unnecessary so I have updated my PR.  I don't believe that it was doing anything to begin with because it had single quotes around it which messes up the parsing on some browsers.

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -10,10 +10,10 @@
     <hr>
 
     <div class="dropdown-wrapper">
-        <div onclick='updateDropdownStates($(this).find(".dropdown-arrow").hasClass("down"), "today_polls_dropdown", "{{ base_url }}")' class="dropdown-bar">
+        <div class="dropdown-bar">
             <h2> Today's Polls </h2>
             <div class="dropdown-btn">
-                <i class="dropdown-arrow{{dropdown_states['today'] ? ' down' : ' right'}}"></i>
```
cjreed121, line 16: This looks nice in dark mode but removes a lot of functionality. In master, a cookie is created for future, today, and old which tell the website if the panels should be open or not as well as the chevrons on page load. I think this feature is still useful so the panels don't have to be hidden every time the page is loaded. I honestly think it can be kept the way it was before but only the color should be changed through CSS. Otherwise some refactoring may need to be down.

cjreed121 (changes requested):

site/app/templates/polls/AllPollsPageInstructor.twig
williamjallen: Ok, I'll update.  I didn't notice any differences so I just removed the code initially.

cjreed121: Yeah in order to see the difference you need to open/close the panels then refresh. If you compare the functionality of that to master you will see the difference.

williamjallen: PR updated.

cjreed121 (approved):

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -10,10 +10,10 @@
     <hr>
 
     <div class="dropdown-wrapper">
-        <div onclick='updateDropdownStates($(this).find(".dropdown-arrow").hasClass("down"), "today_polls_dropdown", "{{ base_url }}")' class="dropdown-bar">
+        <div  onclick="updateDropdownStates($(this).find('.dropdown-arrow').hasClass('down'), 'today_polls_dropdown', '{{ base_url }}')" class="dropdown-bar">
             <h2> Today's Polls </h2>
             <div class="dropdown-btn">
-                <i class="dropdown-arrow{{dropdown_states['today'] ? ' down' : ' right'}}"></i>
+                <i class="dropdown-arrow{{dropdown_states['today'] ? ' down' : ' right'}} fa fa-chevron-right"></i>
```
MasterOdin, line 16: With these changes, the `right` class no effect anymore, correct? Can we then simplify this to be `{{ dropdown_states['today'] ? 'down' }}` (and same as below for other categories).

Could then also remove the toggle on the `right` class from https://github.com/Submitty/Submitty/blob/FixChevronsDarkMode-Polls/site/public/js/polls-dropdown.js.

williamjallen: Yep, that is correct.  Changes have been made.

PR #6497, opened 2021-05-25
[Bugfix:Forum] Fix Attachment Button not Toggling 

### LINKED ISSUE #6445: Attachments Toggle Button not working as expected
**Describe the bug**
When on a thread inside the discussion forum, you can click attachments from the dropdown menu which is supposed to toggle all attachments on current thread. When clicked it opens all attachments and does nothing on repeated clicks.

**Expected behavior**
Since this is a toggle button, repeated clicks on this button should open or close the attachments.

**To Reproduce**
1. Open a thread in the discussion forum that contains a few attachments. (Example: Thread 8 in Sample)
2. Click attachments and all attachments will open
3. Click attachments again and nothing will happen

**Configuration**
 - OS: Linux
 - Browser: Chrome

eddiekrystowski:
Fixes #6445 
Currently, the Attachments dropdown button in discussion forum threads can only be used to open all attachments.

### What is the new behavior?
The Attachments drop down button toggles between showing and hiding all attachments. It keeps track of whether it has shown or hidden all last time, so if all are hidden and one is individually shown, then pressing the toggle will show all, not hide one and show the others.

### Other information?
When testing this PR, specifically try combinations such as showing/hiding all and then individually toggling an attachment and make sure they never get unlinked (clicking the Attachments button shows some and hides others, etc).

### DIFF
--- a/site/app/templates/forum/ForumBar.twig
+++ b/site/app/templates/forum/ForumBar.twig
@@ -101,7 +101,7 @@
 			{% if thread_exists and not is_full_threads_page %}
 	          	<li class="divider"></li>
 	          	<li title="Click to toggle all post attachments" class='dropdown-item' role="menuitem">
-	          		<a href="#" class="key_to_click" tabindex="0" onclick="loadAllInlineImages()">
+	          		<a href="#" id="toggle-attachments-button" class="key_to_click show-all" tabindex="0" onclick="loadAllInlineImages()">
    						Attachments <span class="attachment-badge badge">{{ total_attachments }}</span>
    					</a>
 				</li>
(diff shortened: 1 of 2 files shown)


sbelsk (commented):
I tested this on Windows/Chrome, Firefox from both the instructor's and the student's interface. The "Attachments" option in the "more" dropdown menu now works in both directions as requested, so when all the attachments are closed, clicking it will open all of them, and when all are open, clicking it again will now close all instead of doing nothing. However, when I open part of the attachments manually, I find that clicking the button will negate the states of all the attachments, so those that are closed will turn open, and those that are open will turn closed.

eddiekrystowski:
That's interesting, when I test on MacOS on Chrome this happens after manually opening one:
ezgif com-gif-maker

Can you provide the method you used to test? I don't think it should be an issue between browsers but maybe the sample post you tested with?

sbelsk (approved):
Alright, it was a problem on my end after all :) This works well and should be good to go

PR #6454, opened 2021-05-18
[UI/UX:InstructorUI] Fix user form contrast in dark mode

eddiekrystowski:
### What is the current behavior?
Fixes #5541
Currently, fields that fail validation on the user form in dark mode have contrast that makes the text illegible.

current-dark
current-dark(black)

### What is the new behavior?
The contrast has been adjusted so that the text is legible. The placeholder text uses a slightly off-white color to indicate it was not typed by the user.

new-dark
new-dark(black)

PR #6500, opened 2021-05-25
[UI/UX:InstructorUI] Fix Invalid Contrast on Admin Gradeable

### LINKED ISSUE #6499: Bad Contrast for Invalid Fields on Admin Gradeable in Dark Mode
**Describe the bug**
When creating/editing a gradeable in dark mode, the contrasts for editing the field is poor. Also, the invalid input is recognized by simply changing the background color, with no other indication that it is invalid which makes it less accessible.

**Expected behavior**
The contrast should be improved and the input should be marked with the invalid selector.

**To Reproduce**
1. Create a gradeable
2. Navigate to the Dates tab
3. Change TA Beta Testing Date to a date after Submission Open Date

**Configuration**
 - OS: MacOS
 - Browser: Chrome

**Screenshots**
image

eddiekrystowski:
Fixes #6499 
Same solution as #6454 
Currently, the contrast for invalid fields on the instructor create/edit gradeable page has poor contrast.

image

### What is the new behavior?
- Contrast is improved.
- Invalid fields are now marked with invalid selector to improve accessibility.
- Fixes a bug introduced in #6454 where dark mode styling was applied in light mode.

image

### DIFF
--- a/site/public/css/admin-gradeable.css
+++ b/site/public/css/admin-gradeable.css
@@ -18,6 +18,18 @@ input.readonly.peerstudents {
     width: 200px;
 }
 
+input[type=text]:invalid {
+    background-color: var(--alert-invalid-entry-pink);
+}
+
+[data-theme="dark"] input[type=text]:invalid {
+    background-color: var(--alert-danger-red);
+}
+
+[data-theme="dark"] input[type=text]:invalid::placeholder {
+    color: var(--alert-invalid-entry-pink);
+}
+
 select {
     margin-top:7px;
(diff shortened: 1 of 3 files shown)


loadedskiffer (approved):

MasterOdin:
It looks like we have this paradigm set in four different CSS files. At some point, it would be good to comb through these individual CSS files and combine them into one global include so we don't have to worry about it anymore (maybe in https://github.com/Submitty/Submitty/blob/master/site/public/css/input.css?).

PR #6504, opened 2021-05-25
[Bugfix:Plagiarism] Fix plagiarism config edit bug

williamjallen:
Currently an error may occur in some cases when you try to save edits to a gradeable from the plagiarism detection homepage:

before

### What is the new behavior?
This PR fixes the bug by adding additional checks.

### DIFF
--- a/site/app/controllers/admin/PlagiarismController.php
+++ b/site/app/controllers/admin/PlagiarismController.php
@@ -345,15 +345,11 @@ class PlagiarismController extends AbstractController {
             $old_array = json_decode($old_config, true);
             $regex_in_old = in_array("regex", $old_array);
             $regex_in_new = in_array("regex", $json_data);
-            $json_data["regex_updated"] = false;
-            if (
-                ($regex_in_old
-                && !$regex_in_new)
-                 || (!$regex_in_old
-                 && $regex_in_new)
-                 || ($old_array['regex'] != $json_data['regex'])
-            ) {
+            if ($regex_in_old || $regex_in_new) {
+                $json_data["regex_updated"] = false;
+                if (($regex_in_old && !$regex_in_new) || (!$regex_in_old && $regex_in_new) || ($old_array['regex'] != $json_data['regex'])) {
                     $json_data["regex_updated"] = true;
+                }
             }
         }
         if ($instructor_provided_code == true) {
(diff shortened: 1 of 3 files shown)


eddiekrystowski (approved):
Works well, I am unable to get that error now with these changes. Still much more work to do on this page, but it's a start.

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -345,15 +345,11 @@ public function saveNewPlagiarismConfiguration($new_or_edit, $gradeable_id = nul
             $old_array = json_decode($old_config, true);
             $regex_in_old = in_array("regex", $old_array);
             $regex_in_new = in_array("regex", $json_data);
```
MasterOdin, line 347: Unrelated to this change, but I think this is a bug here as `in_array` will always return false for both of these since these not flat arrays, but rather dictionaries. We should maybe be using `isset` to test for the existence of a key in the dictionary.

Investigating:
```php
<?php

$a = [
  'regex' => 'foo',
];

var_dump(in_array('regex', $a));  # prints bool(false)
var_dump(isset($a['regex']));     # prints bool(true)

$b = ['regex'];

var_dump(in_array('regex', $b));  # prints bool(true)
var_dump(isset($b['regex']));     # prints bool(false)
```

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -345,15 +345,11 @@ public function saveNewPlagiarismConfiguration($new_or_edit, $gradeable_id = nul
...
-            $json_data["regex_updated"] = false;
-            if (
-                ($regex_in_old
-                && !$regex_in_new)
-                 || (!$regex_in_old
-                 && $regex_in_new)
-                 || ($old_array['regex'] != $json_data['regex'])
-            ) {
+            if ($regex_in_old || $regex_in_new) {
+                $json_data["regex_updated"] = false;
```
MasterOdin, line 349: Is there harm of moving this assignment into where we construct the `$json_data` object on line 326 - 339?

MasterOdin (changes requested):

site/app/controllers/admin/PlagiarismController.php
williamjallen: Yep, good catch.  I have pushed a new change.

site/app/controllers/admin/PlagiarismController.php
williamjallen: I'm not sure how we would move the logic to lines 326-339.  Don't we still need to check if there is a regex?

PR #6507, opened 2021-05-25
[Bugfix:Plagiarism] Plagiarism configuration error checking

### LINKED ISSUE #6506: No Plagiarism Configuration Form Verification
**Describe the bug**
There seems to be no verification of inputs on the Plagiarism Detection Configuration page which can lead to weird behaviors. Just one example, if a configuration has the "Threshold/Maximum number of students" field saved as a negative number, the gradeable will just disappear from the list on the Plagiarism Detection home page. You are now unable to make a new configuration for the gradeable because one already exists, but you can't edit it either because it has disappeared.

**Expected behavior**
Some sort of form validation or more graceful handling of bad inputs.

**To Reproduce**
1. Go to Plagiarism Detection page
2. Create a new configuration for a gradeable
3. Set "Threshold/Maximum number of students" field to -2
4. Click Save Changes and return to the home page
5. Wait 5-10 seconds for the configuration to finish processing

**Configuration**
 - OS: MacOS
 - Browser: Chrome

williamjallen:
It is currently possible to input invalid values for some fields when creating and editing a gradeable.  (for example: inputting -1 into the sequence length field)

### What is the new behavior?
Fixes a few bugs which could lead to an invalid plagiarism configuration
Fixes #6506

### DIFF
--- a/site/app/controllers/admin/PlagiarismController.php
+++ b/site/app/controllers/admin/PlagiarismController.php
@@ -232,7 +232,7 @@ class PlagiarismController extends AbstractController {
         }
 
         $language = $_POST['language'];
-        if (isset($_POST['threshold']) && $_POST['threshold'] !== '' && $_POST['threshold'] >= 0) {
+        if (isset($_POST['threshold']) && $_POST['threshold'] !== '' && is_numeric($_POST['threshold']) && $_POST['threshold'] >= 0) {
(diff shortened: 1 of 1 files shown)


eddiekrystowski (approved):
Looks good, eventually we should also do this error checking on the JS side so the inputs can be visually marked as invalid for better accessibility.

PR #6513, opened 2021-05-26
[Bugfix:HelpQueue] individual queue usage stats page

loadedskiffer:
If you log in as an instructor and go to office hours queue => view queue stats => view student specific stats, the page doesn't load and throws an ugly error.
pr6_1

### What is the new behavior?
The page loads properly
pr6_2


eddiekrystowski (approved):
Works good. Did you perhaps mean to label this as Office Hours though? From what I could tell, this doesn't have to do with polls.

PR #6515, opened 2021-05-26
[Bugfix:System] Add guard on constraint drop

MasterOdin:

The migration to add the unique constraint to gradeable_data has an issue where the `DROP CONSTRAINT` would error out if `g_id_gd_team_id_unique` did not exist.

### What is the new behavior?

We use a `IF EXISTS` to guard on the constraint drop so that the migration will not error if it does not exist.


HenrikLam (approved):
The guard worked for me when doing migration where as the migration for course on master branch will result in error for the column not existing.

PR #6521, opened 2021-05-27
[Refactor:System] Remove Lock Date on Gradeables

cjreed121:

Right now a column exists called g_grade_locked_date. This doesn't have any functionality except to be able to be set. There is also no way an instructor could set this field.

### What is the new behavior?
This column has been removed from the database and site code.


bmcutler:
Please look through your line changes, you've made edits to other fields.
Also there's a stray 'z' on one line.

cjreed121:
> Please look through your line changes, you've made edits to other fields.
> Also there's a stray 'z' on one line.

Yes I missed the NOT NULL removal which I just fixed. I'm guessing the z you are referring to is on "timestamptz" which is intentional. "timestamptz" is timestamp with timezone in postgres. I can change it to "timestamp with timezone" if preferred.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5133,8 +5130,6 @@ public function createGradeable(\app\models\gradeable\Gradeable $gradeable) {
               g_ta_view_start_date,
               g_grade_start_date,
               g_grade_due_date,
-              g_grade_released_date,
```
MasterOdin, line 5136: we probably want to keep this line?

migration/migrator/data/course_tables.sql
```diff
@@ -281,12 +281,9 @@ CREATE TABLE public.gradeable (
     g_grade_start_date timestamp(6) with time zone NOT NULL,
     g_grade_due_date timestamp(6) with time zone NOT NULL,
     g_grade_released_date timestamp(6) with time zone NOT NULL,
-    g_grade_locked_date timestamp(6) with time zone,
     g_min_grading_group integer NOT NULL,
     g_syllabus_bucket character varying(255) NOT NULL,
     CONSTRAINT g_grade_due_date CHECK ((g_grade_due_date <= g_grade_released_date)),
-    CONSTRAINT g_grade_locked_date_max CHECK ((g_grade_locked_date <= '9999-03-01 00:00:00-05'::timestamp with time zone)),
-    CONSTRAINT g_grade_released_date CHECK ((g_grade_released_date <= g_grade_locked_date)),
```
MasterOdin, line 289: These constraints are not removed as part of the migration?

MasterOdin (changes requested):

site/app/libraries/database/DatabaseQueries.php
cjreed121: Yes. Not sure why that got removed but will fix it.

migration/migrator/data/course_tables.sql
cjreed121: Those got removed after running the migration that I have added here. Or do I need to add something else to the migration?

bmcutler (approved):

PR #6524, opened 2021-05-28
[UI/UX:TAGrading] Add Simple gradeable Legend

eddiekrystowski:
Currently, only gradeables with components have an icon legend. Simple gradeables just have text.

image
image

### What is the new behavior?
Simple gradeables now have a legend.
image
image


cjreed121 (approved):
Looks great

site/public/css/simple-grading.css
```diff
@@ -20,4 +20,27 @@
 
 .sticky-top label, .sticky-top .not-sticky {
     display: none !important;
-}
\ No newline at end of file
+}
+
+.simple-full-credit {
+    color: #149bdf;
```
williamjallen, line 26: Are these colors standardized? (if not, should they be?)

cjreed121: Yeah I noticed the same thing. I took a look and saw the javascript was using these color tags rather than a defined color variable. Now that this is in two places it could be something good to standardize.

eddiekrystowski: Oh yea, I meant to do that and forgot. I'll make those changes.

williamjallen (approved):

PR #6525, opened 2021-05-28
[Bugfix:Plagiarism] Fix bug when re-running plagiarism

williamjallen:
Currently, an error occurs in the lichen run log when re-running plagiarism detection on an existing gradeable.  The gradeable disappears from the list of configured gradeables but is unable to be reconfigured.  From the client side, it appears to simply disappear altogether.  This is due to an error in `sbin/run_lichen_plagiarism.py` which causes some files to be deleted and some to not be deleted.

The code currently has logic which attempts to not rehash files when re-running a gradeable.  This logic is partially broken and incomplete.

### What is the new behavior?
This PR removes unnecessary, buggy, logic from `sbin/run_lichen_plagiarism.py`.  When re-running plagiarism, the code now hashes and concatenates all of the submissions again, all the time.  Concatenating and hashing take less than 2% of the total running time on most test cases so there is no performance difference.

### DIFF
--- a/sbin/run_lichen_plagiarism.py
+++ b/sbin/run_lichen_plagiarism.py
@@ -61,9 +61,8 @@ def main():
         log_json = open(log_path, 'w+')
 
     start_time = time.time()
-    if (previous_hash == None or (previous_hash != config_hash and config_regex_changed)):
-        concat_res = subprocess.call(['/usr/local/submitty/Lichen/bin/concatenate_all.py', config_path ])
-        tok_res = subprocess.call(['/usr/local/submitty/Lichen/bin/tokenize_all.py', config_path ])
+    concat_res = subprocess.call(['/usr/local/submitty/Lichen/bin/concatenate_all.py', config_path ])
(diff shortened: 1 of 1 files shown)


loadedskiffer (approved):

MasterOdin:
Could you expand your PR description to include more detail about this change? What is the error that you're hitting? Is there specific conditions under which the error occurs? What was the if statement here trying to do, and why is removing it the right move? This context is useful if we ever need to go back in history to figure out why a change was made, and the original author is no longer available.

As a related note, this change looks like it renders the `regex_updated` field obsolete and so #6504 could be re-worked to just remove that entire block altogether on merging this. Why do we want to keep `regex_updated`?

williamjallen:
Changes made.  There were some other unnecessary bits of code which I removed as well.  I will investigate the issues with the other PR and re-work the logic as necessary.  Eventually, it would be good for the front end to be a little more interactive.  As part of that, I may rewrite a decent-sized chunk of this Python file as well as the function in PR #6504.

PR #6529, opened 2021-05-28
[Bugfix:TAGrading] Fix Mobile Scrolling on PDF Edit

### LINKED ISSUE #5348: Make sure TA Grading & PDF Annotation interface work on iPad
An instructor/TA should be able to do all of the rubric grading actions from an iPad interface (or other touch or pen based tablet computer).  For example, grading essays that involve annotation PDFs using the touch or pencil interface and select common marks or typing simple custom mark scores per student.

eddiekrystowski:
Closes #5348
On mobile, if you switch to the pencil tool and then back to the cursor, you will be unable to scroll. This makes your only option of scrolling to use the eraser tool which can be inconvenient.

### What is the new behavior?
* The cursor tool always allows scrolling of the page.
* Black is selected as the pen color in the toolbar the first time you go on the page (as opposed to a transparent box if ```localStorage.main_color``` is empty)

### Other information?
I believe this closes #5348 since the rest of the interface seems to work well on mobile.
Tested on iPad Pro with Apple Pencil

### DIFF
--- a/site/public/js/pdf/PDFInitToolbar.js
+++ b/site/public/js/pdf/PDFInitToolbar.js
@@ -188,7 +188,7 @@ function renderPDFToolbar() {
         document.getElementById("size_selector").addEventListener('click', sizeMenuToggle);
         document.addEventListener('colorchange', changeColor);
         let init_color = localStorage.getItem('main_color');
-        setColor(init_color ? init_color : "#000000");
+        setColor(init_color ? init_color : '#000000');
     }
 
     function colorMenuToggle(e){
--- a/site/public/js/pdf/PDFInitToolbar.js
+++ b/site/public/js/pdf/PDFInitToolbar.js
@@ -59,6 +59,7 @@ function renderPDFToolbar() {
                     PDFAnnotate.UI.enableEraser();
                     break;
                 case 'cursor':
+                    currentTool = 'cursor';
                     PDFAnnotate.UI.enableEdit();
                     break;
                 case 'clear':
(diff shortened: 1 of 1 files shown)


cjreed121 (approved):
Works as expected. Tested on iPad Pro.

site/public/js/pdf/PDFInitToolbar.js
```diff
@@ -57,8 +57,10 @@ function renderPDFToolbar() {
                 case 'eraser':
                     currentTool = 'eraser';
                     PDFAnnotate.UI.enableEraser();
+                    PDFAnnotate.UI.disableEdit();
```
MasterOdin, line 60: I'm still not clear on how we would hit an instance where on hitting `eraser`, `disableEdit` would not have already been called above?

site/public/js/pdf/PDFInitToolbar.js
```diff
@@ -187,7 +189,7 @@ function renderPDFToolbar() {
         document.getElementById("size_selector").addEventListener('click', sizeMenuToggle);
         document.addEventListener('colorchange', changeColor);
         let init_color = localStorage.getItem('main_color');
-        setColor(init_color);
+        setColor(init_color ? init_color : '#000000');
```
MasterOdin, line 192: if `init_color` is either a string or undefined, can do the following shorthand in JS:
```suggestion
        setColor(init_color || '#000000');
```

site/public/js/pdf/PDFInitToolbar.js
eddiekrystowski: I suppose you're right, I just added that line since I somehow managed to get to a state where scrolling with the eraser was leaving tiny marks where I first initiated the scroll. I haven't been able to reproduce it with/without that line so I can remove that line if you want.

MasterOdin: Please do.

PR #6530, opened 2021-05-29
[Testing:Developer] Skip vbguest in CI

MasterOdin:

The vagrant up CI job is failing due to some issue mounting the shared directory.

### What is the new behavior?

This may be related to some botched operation by vbguest when run in the context of the CI. To help here, I am adding a flag to the vagrantfile to allow skipping ensuring that vbguest is installed, and then using that in CI to skip it as well.


shailpatels (approved):
Seems strange that vbguest would be causing a mount issue but this appears to get through the initial vagrant up happier so should be good to merge once the `on push` is removed

PR #6537, opened 2021-05-31
[UI/UX:InstructorUI] "no assignments posted" Message Style

sbelsk:
The "no assignments posted" message that is displayed in the Gradeables page when there are no available gradeables is missing some padding
before_padding

### What is the new behavior?
Padding is added, and the message is now aligned with the "Gradeables" title.
image

### DIFF
--- a/site/app/templates/Navigation.twig
+++ b/site/app/templates/Navigation.twig
@@ -86,7 +86,7 @@
         {% endfor %}
         </div>
     {% else %}
-        <div class="container">
+        <div class="no-assignments-message">
             <p>There are currently no assignments posted.  Please check back later.</p>
         </div>
     {% endif %}
--- a/site/app/templates/Navigation.twig
+++ b/site/app/templates/Navigation.twig
@@ -86,7 +86,7 @@
         {% endfor %}
         </div>
     {% else %}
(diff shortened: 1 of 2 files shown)


williamjallen (changes requested):
Looks good on the wide desktop view but it still needs some work for other screen sizes.  You can probably just make the same changes to the other media queries.

https://user-images.githubusercontent.com/16820599/120367513-f2627880-c2de-11eb-9601-d1247b600f57.mov

williamjallen (approved):
Looks good.  The left-align issue still persists but it appears to come from the `container` class in Bootstrap.  This seems like a reasonable solution.

sbelsk:
> Looks good on the wide desktop view but it still needs some work for other screen sizes. You can probably just make the same changes to the other media queries.
> 
>  Screen.Recording.2021-06-01.at.1.38.11.PM.mov

The changes viewed in different screen sizes is due to the `container` class of the div that contains the text. Removing this class makes the tests fail, and modifying the test is probably unnecessary... For now, I think the cleanest way is just removing the left-padding so that the style would be consistent cross all screen widths. This PR now only adds padding to the bottom of the text.

PR #6538, opened 2021-06-01
[Feature:TAGrading] Student switching by itempool

### LINKED ISSUE #5977: TA grading -- next student with same itempool
Not sure if this is in an open issue or open PR, but it is a known request...  just bumping the priority.

When grading items from an itempool, a TA might like to grade all students who received version 1 first, then go back and grade all students who received version 2, etc.

Let's make a 3rd set of arrows that cycles through students given a specific problem from the item pool.

single arrow -- always go to the next student with a submission regardless of grading status.

double arrow -- go to the next student with incomplete grading (if a component is open, go to next someone who is incomplete in that component, else go to next person with anything incomplete)

new arrow (not sure what symbol to pick -- triple arrow??) -- go to next student with incomplete grading on the component currently open that was assigned the same problem from the component that is open.
[NOTE: this only makes sense if components are linked to itempools and a component is open and the open component has an itempool linked.  this set of arrows should disappear from UI, be disabled if not relevant]

DefinitiveAbove:

Can switch between students by prev/next student, or prev/next ungraded student.
Two set of arrows on hotbar (one for prev/next student, one for prev/next ungraded student)

### What is the new behavior?
Fixes #5977 
Can switch between students by prev/next student, prev/next ungraded student, prev/next student by itempool, and prev/next ungraded student by itempool.
Arrows for prev/next ungraded student (double arrows) removed - prev/next arrows now configurable under the TA grading settings menu (this also affects the left/right arrow shortcut).
image
image

Switching between students depends on open component (and associated itempool that it is linked to). When hovering over the prev/next arrow, it notes the itempool item and component that it will be checking when searching for the prev/next student. 

image


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1232,7 +1232,7 @@ private function amIBlindGrading($gradeable, $user, $peer) {
      *
      * @Route("/courses/{_semester}/{_course}/gradeable/{gradeable_id}/grading/grade")
      */
-    public function showGrading($gradeable_id, $who_id = '', $from = "", $to = null, $gradeable_version = null, $sort = "id", $direction = "ASC", $to_ungraded = null, $component_id = "-1", $anon_mode = false) {
+    public function showGrading($gradeable_id, $who_id = '', $from = "", $to = null, $gradeable_version = null, $sort = "id", $direction = "ASC", $to_ungraded = null, $component_id = "-1", $anon_mode = false, $to_same_itempool = false) {
```
williamjallen, line 1235: Some of the lines are getting a little long here.  Would it be better to break some of the longer lines?

site/app/models/GradingOrder.php
```diff
@@ -170,23 +196,51 @@ public function sort($type, $direction) {
      * Given the current submitter, get the previous submitter to grade.
      * Will skip students that do not need to be graded (eg no submission)
      * @param Submitter $submitter Current grading submitter
+     * @param bool $to_ungraded If the next submitter should be a submitter that is ungraded
+     * @param string|int $component_id Component ID to check (for ungraded/itempool)
+     * @param bool $to_same_itempool If the next submitter should have the same itempool item as the current submitter for the targeted component
      * @return Submitter Previous submitter to grade
      */
-    public function getPrevSubmitter(Submitter $submitter) {
+    public function getPrevSubmitter(Submitter $submitter, bool $to_ungraded = false, $component_id = "-1", $to_same_itempool = null) {
```
williamjallen, line 204: Should return types be listed for this function and the function on line 230?

site/public/js/ta-grading-keymap.js
```diff
@@ -13,6 +13,21 @@ var remapping = {
 };
 
 let settingsData = [{
+        id: "general-setting-list",
```
williamjallen, line 16: Should the indentation for this block be changed to match the surrounding code?

williamjallen (changes requested):

williamjallen:
In all, this looks good and the buttons work as intended.  The page seems slightly slower when it loads on this branch but I'm not sure if it's this PR or even fixable.

site/app/models/GradingOrder.php
DefinitiveAbove: I'm not sure if I was doing it wrong, but the PHP linter didn't like when I put multiple types to be returned

williamjallen: Is it possible to adjust the function such that it always returns one type?  There's a reason the linter doesn't like the possibility of returning different types and that's because if you're a programmer on the other end, using this function, it can be hard to deal with the result, especially if you are unfamiliar with the code.  If it's difficult/impossible to do without multiple return types, it wouldn't be the end of the world but it does make me a little uneasy when I look at it from a long-term maintainability point of view.

williamjallen (approved):

site/app/models/GradingOrder.php
```diff
@@ -170,23 +196,51 @@ public function sort($type, $direction) {
...
+     * @param bool $to_same_itempool If the next submitter should have the same itempool item as the current submitter for the targeted component
+     * @return Submitter|null Next submitter to grade
      */
-    public function getNextSubmitter(Submitter $submitter) {
+    public function getNextSubmitter(Submitter $submitter, bool $to_ungraded = false, $component_id = "-1", bool $to_same_itempool = false): ?Submitter {
+        $this->to_ungraded = $to_ungraded;
+        if ($to_ungraded) {
+            $this->initUsersNotFullyGraded($component_id);
+        }
+        error_log(gettype($component_id));
```
MasterOdin, line 235: Is this intended?

MasterOdin (changes requested):

site/app/models/GradingOrder.php
```diff
@@ -201,7 +201,7 @@ public function sort($type, $direction) {
      * @param bool $to_same_itempool If the next submitter should have the same itempool item as the current submitter for the targeted component
      * @return Submitter|null Previous submitter to grade
      */
-    public function getPrevSubmitter(Submitter $submitter, bool $to_ungraded = false, $component_id = "-1", $to_same_itempool = null): ?Submitter {
+    public function getPrevSubmitter(Submitter $submitter, bool $to_ungraded = false, $component_id = "-1", bool $to_same_itempool = false): ?Submitter {
```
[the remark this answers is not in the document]
MasterOdin: PHP 7.x unfortunately does not support union types, which `$component_id` is (`string|int`). However, would it be possible to replace the default value with `-1` instead of `"-1"` and get the same result?

williamjallen (approved):

PR #6546, opened 2021-06-01
[VPAT:TAGrading] Improve Accessibility on TA Grading Page

eddiekrystowski:
Currently, there are many WAVE errors on the TA Grading page:
* Missing form input labels
* Settings don't follow color scheme of site for similar menus

image
* Links with no text
### What is the new behavior?
* All inputs have hidden labels (except 4 off screen elements, couldn't track them down even with WAVE)
* Settings colors fixed

image
* Links have hidden text associated with them


site/app/templates/grading/electronic/NavigationBar.twig
```diff
@@ -7,6 +7,7 @@
                    data-href="{{ home_url }}"
                    id="main-page"
                 >
+                    <span hidden>Main Page</span>
```
MasterOdin, line 10: Is this for accessibility? I believe that the better option would be to use an [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) attribute on the `<a>` tag, and then maybe also add `aria-hidden="true"` on the `<i>` tag.

```suggestion
                   id="main-page"
                   aria-label="Main Page"
                >
```

See https://fontawesome.com/v5.15/how-to-use/on-the-web/other-topics/accessibility.

@shailpatels can confirm on this.

site/app/templates/grading/electronic/RubricPanel.twig
```diff
@@ -25,13 +25,14 @@
                     {# Verify all button #}
                     {% if show_verify_all %}
                         <div class="col-no-gutters">
+                            <label for="verify-all" tabIndex="0" style="display:none;">Verify All</label>
```
MasterOdin, line 28: This does not strike me as the right way to solve this. WAVE is flagging this input (and others) on needing a label to give assistance to screen readers on what the input does. By including `style="display: none"`, while the label exists and satisfies WAVE, it would still not appear for screen readers. Instead, we would want to "hide" the label by using a technique from [this page](https://webaim.org/techniques/css/invisiblecontent/), like [hiding it offscreen](https://webaim.org/techniques/css/invisiblecontent/#offscreen). This applies to all the other labels in this PR.

MasterOdin (changes requested):

site/app/templates/grading/electronic/NavigationBar.twig
eddiekrystowski: Good point. I had done that to make sure there weren't links with empty text, but I didn't realize aria-label would satisfy that condition.

MasterOdin (approved):
Looks fine, will merge once the conflicts are resolved.

PR #6548, opened 2021-06-01
[VPAT:InstructorUI] Improve Notebook Builder Accessibility

eddiekrystowski:
* Labels are missing on short answer inputs
* Contrast is very low on itempool and form settings containers
* Short answer inputs don't change text theme to match site theme

image

### What is the new behavior?
* Hidden labels have been added to all short answer inputs
* Itempool and form settings containers have had their colors adjusted
* Short answer inputs use eclipse in light mode and monokai in dark/black mode

image

### DIFF
--- a/site/public/css/colors.css
+++ b/site/public/css/colors.css
@@ -153,6 +153,10 @@
     --attachment-box-color: #f5f5f5;
     --file-upload-table-hover: #d1d1d1;
 
+    /*notebook builder*/
+    --notebook-builder-form-options-orange: #F7D8BA;
+    --notebook-builder-itempool-green: #CAF1DE;
+
 
         /*these differ from the colors on the posts for some reason?
         so I will put them here for reference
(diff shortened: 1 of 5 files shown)


site/public/css/colors.css
```diff
@@ -205,200 +209,204 @@
...
-      --box-shadow-slightly-transparent: rgba(255,255,255,0);
-      --no-image-available: #ADADAD;
-      --standard-input-background: #4a4a4a;
-      --checkbox-black: #5296db;
-      --overall-comment: #4E5353;
-      --alternating-row-color-light: #4d4848;
-      --grading-note: #A8A8A8;
-      --file-browser-blue: #ffffff;
-      --extra-credit-blue: #2DACF0;
+    /*master color*/
```
shailpatels, line 212: What are all these changes? If they're just spacing changes it should be reverted but scrolling quickly I don't see any real changes

shailpatels (changes requested):

site/public/css/colors.css
eddiekrystowski: Looks like somehow I made a bunch of indentation changes. I'll undo them.

shailpatels (approved):
Looks good, behavior appears the same of the notebook builder after these changes

PR #6551, opened 2021-06-02
[Bugfix:TAGrading] disable custom marks for Peers

loadedskiffer:
Currently, if a peer is grading a peer assignment they can edit the rubric and add custom marks.

### What is the new behavior?
Peer graders can no longer edit the rubric or add custom marks but if a non-peer grader is grading the same assignment, they can do both of those things.

Below is what a peer grader will see on the rubric panel when grading.
pr7_1

### Other information?
To test I configured peer assignments and tested the rubric for different graders and assignment types.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -28,9 +28,9 @@ exclude=
     sbin/check_everything.py,
     sbin/compare_reuploads.py,
     sbin/count_autograding_logs.py,
-    sbin/run_lichen_plagiarism.py,
     sbin/untrusted_canary.py,
     tests
 
 per-file-ignores =
     .github/workflows/pr_title_tester.py:E501,W605
+    sbin/run_lichen_plagiarism.py:E501
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,25 +1,45 @@
 version: 2
 updates:
-- package-ecosystem: composer
-  directory: "/site"
-  schedule:
-    interval: daily
-    time: "10:00"
-  open-pull-requests-limit: 10
-  labels:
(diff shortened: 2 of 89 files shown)


HenrikLam (changes requested):
I was still able to create custom marks when I'm peer grading as a student.

Steps
1. Create new gradeable with manual grading and peer component as instructor with all grade all option for peer matrix
2. Change all dates to the now option starting from the top row and going downward
3. Log onto aphacker and submit a sample file
4. Log onto student and click grade and go to the student with a file submitted and open a gradeable
5. Student is able to make a custom mark which is allowed to be saved
image

cjreed121 (changes requested):
You did all the changes to the front end to disable this feature but you must also do so to the backend PHP. I was able to add custom marks by manually sending requests to Submitty. This can work for any peer gradeable whether you see the option or not. In the controller you will have to check permissions before allowing the component to be saved if it has custom marks.

sbelsk (approved):
Seems to be working on my end. :)

HenrikLam (approved):
Looks good! Both edit rubric and custom marks are properly hidden for me now.

cjreed121 (changes requested):
Backend seems locked down this time but I have a suggestion to improve it. Also the front end gave me the option to set custom mark when I tried it this time. No idea why.

cjreed121:
I am still able to manually put in custom marks if I send raw requests to the site. I think you need to check more on the values of custom message and custom points rather than treating them like booleans.

cjreed121 (approved):

Issue #28, opened 2014-06-28
ifdef DEBUG statemtents


JFrei86:
Logging in the validator currently has
```c++
#ifdef DEBUG
    {Some form of error checked std::cout statement}
#endif
```
As well as in the runner and config. To improve usability with the rest of the system, we should move extra print statements to these #ifdef DEBUG lines and take some of the error statements out of the DEBUG preprocessor statements so that errors are more clear.

PR #6556, opened 2021-06-03
[Feature:Plagiarism] Implement right panel percent match

williamjallen:
Currently, the percent match for the user on the right panel of the plagiarism results page shows "undefined" for all users.

### What is the new behavior?
This PR implements logic to determine the list of matching users for the right panel of the plagiarism results page.  This PR depends upon PR #28 in the Lichen repository.  As part of the new logic, user 2 is now loaded much quicker than before by utilizing the new ranking files produced by Lichen repository PR #28. 

Untitled


shailpatels:
This branch appears to depend on  #6560 or vice versa in terms of what should get merged

williamjallen:
Yes, there are many issues with the plagiarism page currently and most changes are dependent upon one another.  These two PRs are two members of a long chain of PRs (see above commit history) and #6560 depends upon the work in this PR.  Both of them depend upon the work in the previous PR, and the work before that, and the work before that.  After these two PRs, are merged, the page should be functioning enough to have mostly independent PRs.

site/public/js/plagiarism.js
```diff
@@ -171,37 +172,41 @@ function requestAjaxData(url, f, es) {
 }
 
 function createRightUsersList(data, select = null) {
-    var position = 0;
-    var append_options;
+    let position = 0;
+    let append_options;
+    console.log(data);
+    data = JSON.parse(data);
+    console.log(data);
```
sbelsk, line 179: These `console.log` statements should be removed

bmcutler:
PR also includes some linting/phpstan fixes.

PR #6559, opened 2021-06-04
[Bugfix:TAGrading] fixed Anon mode bugs

loadedskiffer:
1. On the details page, The autograding & grading column are left justified in anon mode but centered in non-anon mode.

2. In 2 column mode, in non-anon mode, the layout seems to leave room for the student name (so the first row of buttons isn't center, it's got extra space on the left side

3. Team/student names were still displaying in 2 column mode if you enabled full screen mode.

### What is the new behavior?
1. On the details page, The autograding & grading column are now always centered.
pr8_1

2.In 2 column mode, the navbar is always centered.
pr8_2

3. Student/team names are never displayed in 2 column mode, regardless of full screen mode.

### Other information?
There were other issues with centering/spacing not mentioned in the issue that this PR is linked too. In order to fix all of the centering issues the only fix was to make the names appear above the navbar and not be in line with it. The div's were messing each other up and there was no good fix that could handle things like different screen size or long student names.
pr8_3
pr8_4

### DIFF
--- a/site/app/templates/grading/electronic/NavigationBar.twig
+++ b/site/app/templates/grading/electronic/NavigationBar.twig
@@ -107,9 +107,9 @@
 </div>
 <script>
     //if student/team name(s) are being displayed
-    if ({{ not anon_mode and not isBlind and (peer_blind_grading is same as('unblind') or peer_blind_grading is same as('single')) }}) {
-        $('.navigation-box').css({'position':'relative'});
-        $('.navigation-box').css({'float':'left'});
-        $('.navigation-box').css({'margin-left':'20%'});
-    }
+{#    {% if not anon_mode and not isBlind and (peer_blind_grading is same as('unblind') or peer_blind_grading is same as('single')) %}#}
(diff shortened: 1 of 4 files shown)


cjreed121 (approved):
Works as expected.

site/app/templates/grading/electronic/NavigationBar.twig
```diff
@@ -106,10 +106,4 @@
     </div>
 </div>
 <script>
-    //if student/team name(s) are being displayed
-    if ({{ not anon_mode and not isBlind and (peer_blind_grading is same as('unblind') or peer_blind_grading is same as('single')) }}) {
-        $('.navigation-box').css({'position':'relative'});
-        $('.navigation-box').css({'float':'left'});
-        $('.navigation-box').css({'margin-left':'20%'});
-    }
 </script>
```
MasterOdin, line 109: We can get rid of these `<script>` tags if there's nothing in them.

MasterOdin (changes requested):

PR #6560, opened 2021-06-04
[Feature:Plagiarism] Add coloring for common/provided code

williamjallen:
Regions marked "common" or "provided" in the matches.json exported from Lichen are currently colored yellow.

### What is the new behavior?
This PR adds the proper coloring for different types of matches as per the highlighting key.
Untitled


site/app/controllers/admin/PlagiarismController.php
```diff
@@ -68,9 +68,35 @@ private function getGradeablesFromPriorTerm() {
...
 
 
-    private function getRankings($gradeable_id) {
+    private function getOverallRankings($gradeable_id) {
         $course_path = $this->core->getConfig()->getCoursePath();
-        $file_path = $course_path . "/lichen/ranking/" . $gradeable_id . ".txt";
+        $file_path = $course_path . "/lichen/ranking/" . $gradeable_id . "/overall_ranking.txt";
+        if (!file_exists($file_path)) {
+            $this->core->addErrorMessage("Plagiarism Detection job is running for this gradeable.");
+            $this->core->redirect($this->core->buildCourseUrl(['plagiarism']));
```
shailpatels, line 76: I think you would want to return at the end of this if statement, otherwise the rest of the function will still execute

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -68,9 +68,35 @@ private function getGradeablesFromPriorTerm() {
...
+        }
+        $test = file_get_contents($file_path);
+        if (file_get_contents($file_path) == "") {
+            $this->core->addSuccessMessage("There are no matches(plagiarism) for the gradeable with current configuration");
+            //$this->core->redirect($this->core->buildCourseUrl(['plagiarism']));
```
shailpatels, line 81: should this line be uncommmented / deleted?

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -68,9 +68,35 @@ private function getGradeablesFromPriorTerm() {
...
+    }
+
+    /**
+     * Returns a ranking of users by percent match with user 1 (used for determining the rightmost dropdown list)
+     * @param string $gradeable_id
+     * @param string $user_id_1
+     * @param string $user_1_version
+     * @return array
+     */
+    private function getRankingsForUser($gradeable_id, $user_id_1, $user_1_version) {
```
shailpatels, line 97: Since this function already has the parameters annotated it would be good to add type hints to this and other functions added to make PHP enforce types better and improve static analysis

https://submitty.org/developer/coding_style_guide/php#type-declarations

site/public/js/plagiarism.js
```diff
@@ -171,37 +172,41 @@ function requestAjaxData(url, f, es) {
 }
 
 function createRightUsersList(data, select = null) {
-    var position = 0;
-    var append_options;
+    let position = 0;
+    let append_options;
+    console.log(data);
```
shailpatels, line 177: Are these console.logs needed

shailpatels (changes requested):

site/app/controllers/admin/PlagiarismController.php
williamjallen: I'll add a return statement just for good measure but `redirect` dies so the code should never pass this point.

williamjallen: On second thought, I'm not sure if adding a return statement is good practice here.  Thoughts?

MasterOdin: Good practice would be to not have this "helper" function mutate the response state, and rather just return some value, and then the redirect (and `addSuccessMessage` below) would happen in the caller site. The caller could then (eventually) return `RedirectResponse` for this failure case.

For a return value, I would probably suggest doing `: ?array` where if `$rankings === null`, then redirect, if `count($rankings) === 0`, show `successMessage`.

williamjallen: Wouldn't it defeat the purpose of having this function to move the error checking back to the caller site though?  The reason I made the function in the first place was to make sure that the same errors were treated the same way, with the same error text, which led to the same place in the code.

MasterOdin: Perhaps, though pretty usually the callers want to do different things on error conditions. For example, here, I would doubt the `ajaxGetSubmissionConcatenated` based on its name and return type (a stringified JSON object) that having a redirect would be an appropriate error condition.

site/public/js/plagiarism.js
```diff
@@ -171,37 +172,39 @@ function requestAjaxData(url, f, es) {
...
     $('[name="user_id_2"] option', form).eq(position).prop('selected', true);
     $('[name="user_id_2"]', form).change();
 }
 
 function createLeftUserVersionDropdown(version_data, active_version_user_1, max_matching_version, code_version_user_1) {
-    var append_options;
+    let append_options;
     $.each(version_data, function(i,version_to_append){
-        if(version_to_append == active_version_user_1 && version_to_append == max_matching_version){
+        console.log(version_to_append);
```
sbelsk, line 196: Need to be removed

bmcutler:
@williamjallen 
Please fix merge conflicts

williamjallen:
Fixed.  I think the issue was from where I made a slight change to a previous PR after this one was updated.

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -575,6 +582,13 @@ public function ajaxGetSubmissionConcatenated($gradeable_id, $user_id_1, $versio
 
         $rankings = $this->getOverallRankings($gradeable_id);
 
+        if (count($rankings) === 0 || $rankings === null) {
```
MasterOdin, line 585: This condition needs to be flipped so that it's `if ($rankings === null || count($rankings) === 0) {` as  `count(null)` will throw a PHP warning.

site/app/libraries/plagiarism/Interval.php
```diff
@@ -23,6 +26,10 @@ public function getStart(): int {
         return $this->start;
     }
 
+    public function getType(): string {
+        return $this->type;
+    }
```
MasterOdin, line 31: A usage of this function should be added to `site/tests/app/libraries/plagiarism/IntervalTester.php`

MasterOdin (changes requested):

site/app/controllers/admin/PlagiarismController.php
williamjallen: I will change this in the next PR.

PR #6566, opened 2021-06-07
[Feature:InstructorUI] allow instructor to toggle custom marks

loadedskiffer:
Custom marks can be added by anyone to any assignment (except if a peer is grading a peer assignment). 

### What is the new behavior?
When creating a new gradeable or editing one, an instructor has the option to choose if they want custom marks to be allowed for that assignment. 
pr9_1
pr9_2

### Other information?
To test:
-Create a new gradeable and select no for allow custom marks under the rubric panel
-make a submission
-grade the assignment
-see if you can add custom marks
-edit the gradeable and choose yes for allow custom marks
-try grading again


eddiekrystowski (approved):
Works well, tested on Mac Safari and Chrome

cjreed121 (approved):

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1791,7 +1791,17 @@ public function ajaxSaveGradedComponent($gradeable_id) {
...
+
+        //don't allow custom marks if they are disabled
+        if ($custom_message != null || $custom_points != null) {
+            if (!$gradeable->getAllowCustomMarks()) {
+                $this->core->getOutput()->renderJsonFail('Custom marks are disabled for this assignment');
+                return;
+            }
+        }
+
+        //don't allow peer graders to save custom marks no matter hoe gradeable is configured
```
bmcutler, line 1804: hoe -> how

bmcutler (changes requested):
Notes:

* If grading has begun and custom marks have been used, those marks will still factor into the points, students will still see those custom marks when they view the grade.

* But if a grader opens up the component and re-saves it, the custom mark is lost.

This is ok behavior, but let's put a warning next to the "Will graders be allowed to use custom marks ?" question.
Please add a simple message in red italics saying "WARNING: Grading has begun and custom marks have been used." IF there are any custom marks already saved in the database for this homework.

Note: We could delete all of the custom marks when the instructor checks this box, but is dangerous and could be a lot of undesired data loss.  So please don't do that!

site/app/controllers/grading/ElectronicGraderController.php
loadedskiffer: I wasn't able to find where custom marks are stored in the data base so I couldn't write a check for it. I instead checked if grading had begun and displayed the warning. If that's not satisfactory I can try to come up with something but I just wasted a ton of time trying to do a work around that didn't end up working.

loadedskiffer: Update - Chris helped me find it, it wasn't showing up in my data base because I'm an idiot, I'll fix soon.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5688,13 +5693,40 @@ public function getHasSubmission(Gradeable $gradeable, Submitter $submitter) {
...
+        foreach ($rows as $row) {
+            $this->course_db->query(
+                'SELECT * FROM gradeable_component_data WHERE gc_id=?',
+                [$row['gc_id']]
+            );
+            if ($this->course_db->row()) {
+                return true;
+            }
+        }
+        return false;
```
cjreed121, line 5720: This could be reduced down to 1 query.

cjreed121 (changes requested):

cjreed121 (approved):

migration/migrator/data/course_tables.sql
```diff
@@ -283,6 +283,7 @@ CREATE TABLE public.gradeable (
     g_grade_released_date timestamp(6) with time zone NOT NULL,
     g_min_grading_group integer NOT NULL,
     g_syllabus_bucket character varying(255) NOT NULL,
+    g_allow_custom_marks boolean NOT NULL,
```
bmcutler, line 286: This is missing the default value and the order doesn't match.

You should not manually make this edit, rather run the 'dump' command from step 3:
https://submitty.org/developer/migrations#writing-new-migrations

site/app/templates/admin/admin_gradeable/rubric/ElectronicFileRubric.twig
```diff
@@ -9,6 +9,21 @@
   <option value="5" {{ gradeable.getPrecision() == 5 ? 'selected' : '' }}>5</option>
   <option value="10" {{ gradeable.getPrecision() == 10 ? 'selected' : '' }}>10</option>
 </select>
+<br>
+<br>
+<fieldset>
+    <legend>
+    Will graders be allowed to use custom marks?
```
bmcutler, line 16: Let's only show this red message if there already are custom marks and the instructor is attempting to disable them.
(Never show the red message if the "yes" radio button is selected.)

bmcutler (changes requested):
A quick test was good...

Except as new issue #6607 I had trouble getting the message to go away...  I observed both an empty string custom mark in the database (the bug), but even when I got the mark to go away in the db the message remained.  Probably a testing confusion/mistake on my part.
(I think this code looks ok?)

bmcutler:
@loadedskiffer 
The logic for when to display the button seems to be broken.
It seems to appear whenever the yes/no is toggled and there is any grading completed.
It should only appear when custom mark is set to "no" and there is a custom mark that is empty. 
It's not just the issue @cjreed121 reports in issue #6607

loadedskiffer:
> @loadedskiffer
> The logic for when to display the warning text seems to be broken.
> It seems to appear whenever the yes/no is toggled and there is any grading completed.
> It should only appear when custom mark is set to "no" and there is a custom mark that is empty.
> It's not just the issue @cjreed121 reports in issue #6607

I'm not sure if it is a bug or it is intentional but If a TA grades a component and DOES NOT add a custom mark, a row is still added in the table where custom marks are stored which was breaking my query. However the row has an empty component_comment which isn't actually allowed for custom marks so the new query checks for an empty string and now works as expected. If Custom marks were originally allowed, and then a custom mark is used , and then an instructor attempts to turn off custom marks, the warning message is displayed.

migration/migrator/data/course_tables.sql
```diff
@@ -694,7 +695,8 @@ CREATE TABLE public.polls (
     question text NOT NULL,
     status text NOT NULL,
     release_date date NOT NULL,
-    image_path text
+    image_path text,
+    question_type text DEFAULT 'single-response'::text
```
cjreed121, line 699: These changes should not be in this PR

cjreed121 (changes requested):

bmcutler (approved):
this works great now, thanks!

Issue #6050, opened 2020-12-07
Polling - allow "select multiple" responses


bmcutler:
We should default to radio buttons, but allow the instructor to configure a 
specific question as a multiple response / checkbox button answer.

PR #6567, opened 2021-06-07
[Feature:SubminiPolls] Poll Multiple Response

sbelsk:

Closes https://github.com/Submitty/Submitty/issues/6050
Currently, polls are configured to collect a single response from students. The student interface displays radio buttons for the responses.

### What is the new behavior?
- polls can be configured to be either "single response" or "multiple response"
- in either poll configuration, the instructor can still specify more than one correct response
- import and export of polls of either kind is still supported
- instructor's poll results page reflects all answers selected by students in polls of both kinds
- student's poll summary page displays answer selected as before, and it displays the answers separated by commas if there are more than one

Instructor's view:
multi_response
Sudent's view:
multi_poll_student
poll-summary

### Other information?
This PR modifies the mechanism behind creating/editing polls and managing the students' answers. 
Currently, the calculation of late days earned with students' answers to polls is done in the RainbowGrades repo and is still implemented for single-response polls only. 
In the future, if we want this to be changed, we can also modify the instructor's new/edit poll page to specify the grading method, meaning how to determine whether the student’s answer should be marked full credit (i.e. choosing one of the correct answers to get full credit, or selecting all and only the correct answers, etc.)

### Modifications to the Database?
- Added column `question_type` to the `polls` table that can take value "single-response" or "multiple-response" 
- Previously, the rows `poll_responses` were unique combinations of `poll_id` `student_id`, since each student had a single response to each poll. Now, to save all of a student's responses to a poll, if it is of type multiple-response, there can be more than one row for each `poll_id` and `student_id` combination.


sbelsk:
I tentatively made it so that the poll type can be changed after it is initially created, but doing so after students have already begun submitting responses is undefined behavior (especially when changing from "multiple-response" to "single-response"). Should we display a warning message to instructors if they attempt to do this?

site/app/models/PollModel.php
```diff
@@ -23,6 +23,7 @@ class PollModel extends AbstractModel {
     protected $name;
     /** @prop-read string */
     protected $question;
+    protected $question_type; //can hold the value "single-response" or "multilple-response"
```
cjreed121, line 26: Add @prop-read and type here for comment

site/app/models/PollModel.php
```diff
@@ -58,6 +60,10 @@ public function getAnswers() {
         return $this->answers;
     }
 
+    public function getQuestionType() {
+        return $this->question_type;
+    }
+
```
cjreed121, line 66: This could be changed to a magic function/method like on lines 12-17.

site/app/libraries/PollUtils.php
```diff
@@ -26,6 +26,7 @@ public static function getPollExportData(array $polls): array {
                 "id" => $poll->getId(),
                 "name" => $poll->getName(),
                 "question" => $poll->getQuestion(),
+                "question-type" => $poll->getQuestionType(),
```
loadedskiffer, line 29: I think it would be best to replace the - in question-type to an  underscore to follow the naming convention.

site/app/controllers/PollController.php
```diff
@@ -341,7 +360,7 @@ public function submitEdits(): RedirectResponse {
             $this->core->addErrorMessage("Invalid Poll ID");
             return new RedirectResponse($returnUrl);
         }
-        if (!isset($_POST["response_count"]) || !isset($_POST["name"]) || !isset($_POST["question"]) || !isset($_POST["release_date"])) {
+        if (!isset($_POST["response_count"]) || !isset($_POST["name"]) || !isset($_POST["question"]) || !isset($_POST["question-type"]) || !isset($_POST["release_date"])) {
```
loadedskiffer, line 363: All other post variable use an underscore so question-type should too.

site/app/templates/polls/NewPollPage.twig
```diff
@@ -79,23 +89,39 @@
...
+                {% else %}
+                    <input type="radio" id="poll-type-single-response" name="question-type" value="single-response">Single response (radio buttons)</input>
+                {% endif %}
+            </label>
+            <label for="poll-type-multiple-response">
+                {% if poll.getQuestionType() == 'single-response'%}
+                    <input type="radio" id="poll-type-multiple-response" name="question-type" value="multiple-response">Multiple response (checkboxes)</input>
+                {% else %}
+                    <input type="radio" id="poll-type-multiple-response" name="question-type" value="multiple-response" checked="">Multiple response (checkboxes)</input>
+                {% endif %}
```
cjreed121, line 105: Rather than needing to write this 2 times you could do something like this:

`<input type="radio" id="poll-type-single-response" name="question-type" value="single-response" {% if poll.getQuestionType() == 'single-response' %} checked {% endif %}>Single response...</input>`

cjreed121 (changes requested):
Seems to work well. Just a few suggestions for the code though.

loadedskiffer (changes requested):
Besides these 2 small things, everything looks good. The feature works as expected.

site/app/libraries/PollUtils.php
sbelsk: You're right. I'll change that

loadedskiffer (approved):

site/app/templates/polls/NewPollPage.twig
sbelsk: Cool! I didn't know this was possible. I'll change this

cjreed121 (changes requested):

site/app/controllers/PollController.php
```diff
@@ -261,25 +261,44 @@ public function submitResponse() {
...
             $this->core->addErrorMessage("No answer given");
             return MultiResponse::RedirectOnlyResponse(
                 new RedirectResponse($this->core->buildCourseUrl(['polls']))
             );
         }
-        $poll = $this->core->getQueries()->getPoll($_POST["poll_id"]);
-        if ($poll == null) {
-            $this->core->addErrorMessage("Invalid Poll ID");
+        if ($poll->getQuestionType() != "single-response" && $poll->getQuestionType() != "multiple-response") {
+            $this->core->addErrorMessage("Invalid poll question type");
```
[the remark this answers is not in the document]
sbelsk: Good catch. I will add this.

site/app/controllers/PollController.php
```diff
@@ -261,25 +261,44 @@ public function submitResponse() {
...
             return MultiResponse::RedirectOnlyResponse(
                 new RedirectResponse($this->core->buildCourseUrl(['polls']))
             );
         }
         if ($poll->isOpen()) {
-            if ($_POST["answer"] == "-1") {
-                $this->core->getQueries()->deleteUserResponseIfExists($_POST["poll_id"]);
+            // FIXME: make this work for when we have only one answer, and also for when we have multiple responses
```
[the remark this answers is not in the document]
sbelsk: Yup, correct. Removing.

cjreed121 (approved):

migration/migrator/data/course_tables.sql
```diff
@@ -694,7 +694,8 @@ CREATE TABLE public.polls (
     question text NOT NULL,
     status text NOT NULL,
     release_date date NOT NULL,
-    image_path text
+    image_path text,
+    question_type text DEFAULT 'single-response'::text
```
MasterOdin, line 698: Better to use `varchar` here if the string is guaranteed to be under 255 characters, as it allows certain performance optimizations within the DB. I realize there's other fields within this table that are `text`, and they should probably be moved to `varchar` as well, except for "question", but no point adding more debt here.

site/app/controllers/PollController.php
```diff
@@ -117,7 +117,7 @@ public function showNewPollPage() {
      * @return MultiResponse
      */
     public function addNewPoll() {
-        if (!isset($_POST["response_count"]) || !isset($_POST["name"]) || !isset($_POST["question"]) || !isset($_POST["release_date"])) {
+        if (!isset($_POST["response_count"]) || !isset($_POST["name"]) || !isset($_POST["question"]) || !isset($_POST["question_type"]) || !isset($_POST["release_date"])) {
```
MasterOdin, line 120: Probably better to move to using a foreach here to make it a bit more succinct.
```php
$fields = ['response_count', 'name', 'question', 'question_type', 'release_date'];
foreach ($fields as $field) {
  if (!isset($_POST[$field])) {
    // error conditon
  }
}
```

site/app/controllers/PollController.php
```diff
@@ -497,6 +521,16 @@ public function importPollsFromJSON(): RedirectResponse {
...
             $question = $poll["question"];
+            /*  Polls that were exported before this feature was
+                implemented don't have this data. At the time, there
+                only existed questions of type single reponse. */
+            if (array_key_exists("question_type", $poll)) {
+                $question_type = $poll["question_type"];
+            }
+            else {
+                $question_type = "single-response";
+            }
```
MasterOdin, line 532:
```suggestion
            $question = array_key_exists("question_type", $poll) ? $poll['question_type'] : 'single-response';
```

site/app/templates/polls/NewPollPage.twig
```diff
@@ -6,14 +6,14 @@
             <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
             <input type="hidden" name="response_count" id="response-count" value="0"/>
 
+            <span class="option-title" id="poll-name-label">Poll Name:</span>
             <label for="poll-name">
```
MasterOdin, line 10: The label should be around the text ("Poll Name:") that the input belongs to for accessibility. Wrapping it around the input itself does not do anything.

See: https://www.w3.org/WAI/tutorials/forms/labels/

This applies to the other changes within this file.

site/app/controllers/PollController.php
```diff
@@ -341,7 +364,7 @@ public function submitEdits(): RedirectResponse {
             $this->core->addErrorMessage("Invalid Poll ID");
             return new RedirectResponse($returnUrl);
         }
-        if (!isset($_POST["response_count"]) || !isset($_POST["name"]) || !isset($_POST["question"]) || !isset($_POST["release_date"])) {
+        if (!isset($_POST["response_count"]) || !isset($_POST["name"]) || !isset($_POST["question"]) || !isset($_POST["question_type"]) || !isset($_POST["release_date"])) {
```
MasterOdin, line 367: can do a similar foreach here. Perhaps define the expected fields as a class property to iterate over so it's only defined in one place.

MasterOdin (changes requested):

migration/migrator/data/course_tables.sql
bmcutler: Sounds good.  This field will be < 255 characters yes.
I think we should add a couple other options for this value.

bmcutler (changes requested):
To tackle the question of grading multiple response, let's just expand to 6 question types:


single-response-single-correct
GUI allows instructor to pick only 1 answer as correct - radio buttons

single-response-multiple-correct [ CURRENT/DEFAULT OPTION ]
GUI allows instructor to pick 1 or many as correct - checkboxes
Rainbow grades will give student "credit" if they pick any correct option.

single-response-survey
GUI hides correct selection...  all answers are saved as correct to the database & file
(effectively the same as single-response-multiple-correct with all options manually selected as correct)

multiple-response-exact
GUI allows instructor to pick 1 or many as correct - checkboxes
RainbowGrades will only give student "credit" if they select exactly these options (no extra, no missing)

multiple-response-flexible
GUI allows instructor to pick 1 or many as correct - checkboxes
RainbowGrades will only give student "credit" if they select at least one correct option and no incorrect options.

multiple-response-survey
GUI hides correct selection...  all answers are saved as correct to the database & file


Probably display these to the instructor as vertical selection -- possibly a drop down menu?
And put a sentence on the GUI for the currently selected option explaining how it will be graded.

bmcutler:
Oh, I also meant to add...
I tested this and it looks good!!

I'm working on a PR to Rainbow Grades to handle the new poll response exported data (& be backwards compatible)
The RainbowGrades PR must be merged first, and we'll add a repo version upgrade too.

sbelsk:
I added the new poll types and changed the type of column in the migration as requested. The changes in the instructor's interface are only in the new/edit poll  pages (the wording/text on the page explaining each of the 6 types might need editing), and the student's interface has not changed.

bmcutler:
@sbelsk 
I've updated the text.  I realized you have duplicate code for for new/edit so I added another file to share/reduce duplicate code.  However it is now broken since we don't have an initial value for the poll type for a new poll.  Hopefully you can replair this out faster than me.

MasterOdin:
@bmcutler @sbelsk I've pushed a commit here to fix the form so that it now defaults to the first option in the form, should be all set there.

sbelsk:
@MasterOdin Thanks! I changed the default value for the "single-response-multiple-correct" option as requested by @bmcutler in the comment explaining the 6 types

PR #6572, opened 2021-06-08
[Bugfix:System] Fix setup script for pip install

wusatosi:

Vagrant up does not work on a fresh install.

The path to `system_requirements.txt` and `vagrant_requirements.txt` is incorrect in the setup script `install_system.sh`.
This bug is introduced by #6399.

Furthermore, there is a problem with install PyYAML that vagrant up will trigger. The details are discussed in #6565, this PR implements a temporary fix so that vagrant up will succeed. 

The scheduled `vagrant up` CI [fails](https://github.com/Submitty/Submitty/actions/runs/916847036) due to this issue.

### What is the new behavior?
This PR replaces the faulty requirements path with a correct one. 
This PR un-pined the version for PyYAML, this is a temporary fix, more discussion is in #6399.

### Other information?

This PR works as a manual trigger of `vagrant up` CI succeed in setting up python dependencies in [system_requirements](https://github.com/wusatosi/Submitty/runs/2771372918?check_suite_focus=true#step:3:4027) and [vagrant_requirments](https://github.com/wusatosi/Submitty/runs/2771372918?check_suite_focus=true#step:3:4056).

### DIFF
--- a/.setup/install_system.sh
+++ b/.setup/install_system.sh
@@ -182,10 +182,11 @@ source ${CURRENT_DIR}/distro_setup/setup_distro.sh
 #install DLL for zbar
 apt-get install libzbar0 --yes
 
-pip3 install -r ${SUBMITTY_REPOSITORY}/.setup/pip/system_requirements.txt
+pip install --upgrade pip
+pip3 install -r ${CURRENT_DIR}/.setup/pip/system_requirements.txt
 
 if [ ${VAGRANT} == 1 ]; then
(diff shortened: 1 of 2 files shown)


MasterOdin:
Closing this as vagrant is fixed in #6589 in a slightly different manner, without overwriting the apt installed python3-pip (which can lead to some weird side effects on apt upgrades) or unpinning the PyYAML version (which does not break vagrant). We did used to install pip manually outside of apt, but found that we ended up with version drift in prod as using `pip install -U ...` stuff wasn't something sysadmins did.

PR #6585, opened 2021-06-09
[Bugfix:TAGrading] Fix full panel view and file view height

DefinitiveAbove:

Viewing a file using the full panel icon does not work.
When a file is rapidly clicked twice (so it finishes loading while the iframe for the file is closed), the height is very small when opened back up. 
When using the Open/Close Submissions/Results buttons, instead of opening and closing all of them it inverts the open and closed state for the most part.

### What is the new behavior?
Viewing a file using the full panel icon now works.
When a file is rapidly clicked twice, an observer observes the iframe until it is opened up again to properly resize it.
When using the Open/Close Submissions/Results buttons, it fully opens/closes the child items.


HenrikLam (approved):
Seems good, all 3 bugs described seems to be fixed!

PR #6554, opened 2021-06-03
[Feature:InstructorUI] Add preview to markdown

eddiekrystowski:
* [x] Forum
* [x] Notebook Builder
* [x] Polls
* [x] Assignment Overall Comments
* [x] Office Hours Announcements
* [x] Grade Inquiries

### What is the current behavior?
You can't see what markdown will look like while writing it.

### What is the new behavior?
* There is a button that will toggle rendering the markdown preview.
* Markdown shortcut buttons have been added to
  * Edit Office Hours Announcement
  * New/Edit Poll
  * Grade Inquiries
* Markdown rendering has been added/fixed in
  * Grade Inquiries
  * Assignment Overall Comments
* Introduces ```Markdown.twig``` to help standardize display of markdown (will do a separate PR to use this everywhere, for now it is only in some spots).
* Introduces ```MarkdownArea.twig``` to help standardize markdown input fields.
---
## Forum
forum1
forum2
---
## Notebook Builder
notebookbuilder
---
## Polls
poll-preview
---
## Assignment Comments
comments-preview
---
## Office Hours Announcement
officehoursqueue-preview
---
## Grade Inquiries
gradeinquiries-preview

# Testing
Please navigate to each of the pages listed in the checklist and test:
* The preview feature
* The consistency of the markdown compared to other preview areas
* The form the preview is associated with still functions like normal

This is a large PR, so I'm sure I missed something, so do your best to break it.

PR #6595, opened 2021-06-10
[Feature:Forum] Better Markdown Code Boxes

### LINKED ISSUE #4208: Better Code Boxes for Forum
Looking at https://github.com/Submitty/Submitty/pull/4186 which was merged recently, I'd like to make a request to not wrap lines of code, but rather add a horizontal scroll bar. This is how I've seen most implementations for code viewers, and it's especially useful given the lack of line numbers. I think this is especially necessary for languages where whitespace is significant.

It may be that there was already a design discussion offline about this, and if so, I'd like to see a high level summary (as the original issue calls for scrolling) and then feel free to close this issue.

The wishlist:

* grey background for the block of code
* language specific syntax highlighting (perhaps a default per course, but allow author to specify/override)
* line numbers
* [probably] use horizontal scrolling for extra wide lines (don't wrap)

eddiekrystowski:
Closes #4208 
Code in markdown is just converted to monospace font and long lines wrap around to the next line
image

### What is the new behavior?
* Code is contained into CodeMirror code boxes
* Code has language specific syntax highlighting
* Code boxes have a light gray background
* Long lines enable scrolling inside of the code box instead of wrapping.
image
image

### Other Information?
Extension of #6554 which should be merged first.


site/app/templates/misc/Markdown.twig
```diff
@@ -0,0 +1,49 @@
...
+        });
+
+        //set timeout to ensure refresh happens after new content has been set
+        //using 1 ms works, but 100 ms is used to be safe
+        setTimeout(function() {
+            editor.refresh();
+        },100);
+    });
+
+</script>
```
MasterOdin, line 49: It may be worth looking into switching from Parsedown to [League\CommonMark](https://commonmark.thephpleague.com/) as it would make possible to attach our own custom renderers for code blocks so that we don't have to do this kind of janky JS DOM manipulation post-hoc.

eddiekrystowski: Okay, I'll look into that. Do you think I should try to incorporate that into this PR or do it in a separate one?

eddiekrystowski: @MasterOdin I have everything working on the new engine except for line numbers in the code boxes. I've been at this for a few days now, but it seems some of the documentation for League\CommonMark is outdated or incomplete, making it very difficult to figure out what exactly I should be doing, and how I can get extensions (like the syntax highlighter) to work alongside my own since they both deal with rendering code blocks. 

I could probably do DOM manipulation to include the line numbers, but that would defeat the whole point of switching to this new markdown engine, unless you think it would be worth it to switch anyways since this engine is getting updated more. Any suggestions?

MasterOdin: Would you be able to post your work in a branch so I can see it?

eddiekrystowski: Sorry for the delayed response. My implementation is now in the `commonmark` branch.

eddiekrystowski:
Implemented with new markdown engine in #6669

PR #6598, opened 2021-06-11
[Feature:Notifications] Add email pages for superusers

### LINKED ISSUE #4964: Sysadmin feature: Email all users -- global system announcement
The system administrator might like to contact all active users on the system via email (and/or notification).

Let's add a new web page, accessible only by admin users,
that will send an email (insert a row into the email queue) 
for all users on Submitty that are instructors or graders or 
non-null students for at least one active (non-archived) course.

This new page is the beginning of a set of pages that should be 
accessible only by admin level users.  

design discussion on access and on the details of this page.

HenrikLam:

Closes #4964
Sysadmins (superusers) cannot email users without directly interacting with the database.

### What is the new behavior?
Added a new page visible only to sysadmins where they can email active users(currently in a course).


williamjallen (changes requested):

site/app/templates/superuser/Email.twig
```diff
@@ -0,0 +1,23 @@
+<div id="superuser-email" class="content">
+    <h1> System Wide Email</h1>
+    <form>
+        <div class="checkbox-options">
+            Recipients:
+            <input type="checkbox" id="email-instructor" name="email-instructor" checked> Email to instructors </input>
+            <input type="checkbox" id="email-full-access" name="email-full-access" checked> Email to Full Access Graders </input>
+            <input type="checkbox" id="email-limited-access" name="email-limited-access" checked> Email to Limited Access Graders</input>
+            <input type="checkbox" id="email-student" name="email-student" checked> Email to students </input>
```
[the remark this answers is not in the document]
HenrikLam: I am not sure about how putting the checkboxes into an array works, but I wasn't able to access the values after doing name="email-checkbox[]" using javascript.

williamjallen (approved):
code looks good

site/app/controllers/EmailStatusController.php
```diff
@@ -33,4 +31,19 @@ public function getEmailStatusPage() {
...
+     */
+    public function getSuperuserEmailStatusPage(): WebResponse {
+        $results = [];
+        $emailStatuses = $this->core->getQueries()->getAllEmailStatuses();
+        return new WebResponse(
+            EmailStatusView::class,
+            'showSuperuserEmailStatus',
+            $emailStatuses
+        );
+    }
```
MasterOdin, line 48: Let's move this function into `site/app/controllers/superuser/SuperuserEmailController.php`?

site/app/libraries/NotificationFactory.php
```diff
@@ -221,8 +221,9 @@ public function sendNotifications(array $notifications): void {
     /**
      * prepare array of Email objects as param array
      * @param array $emails
+     * @param bool $forceSecondary
      */
-    public function sendEmails(array $emails): void {
+    public function sendEmails(array $emails, bool $forceSecondary = false): void {
```
MasterOdin, line 226: variables should be `snake_case` (`$force_secondary`), please fix throughout this PR.

site/app/controllers/superuser/SuperuserEmailController.php
```diff
@@ -0,0 +1,75 @@
...
+        return MultiResponse::webOnlyResponse(
+            new WebResponse(SuperuserEmailView::class, 'showEmailPage')
+        );
+    }
+    /**
+     * @Route("/superuser/email/send", methods={"POST"})
+     * @return JsonResponse
+     */
+    public function sendEmail(): JsonResponse {
+        if (!isset($_POST['emailContent']) || $_POST['emailContent'] == '') {
```
MasterOdin, line 39: Same as variables, POST keys (and the HMTL names in form) should be `snake_case`: `$_POST['email_content']`

site/app/models/SuperuserEmail.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace app\models;
+
+use app\libraries\Core;
+use app\models\Email;
```
MasterOdin, line 6: You don't need this `use`. It's implied that any class that's not given an absolute namespace is then using the namespace of this class.

MasterOdin (changes requested):

site/app/controllers/EmailStatusController.php
HenrikLam: Originally I intended `SuperuserEmailController.php` to be used for the email sending page while the `EmailStatusController.php` to be used for the email status page. Should I still move the function into `SuperuserEmailController.php`?

bmcutler (changes requested):
@HenrikLam 

I tested this, and it's looking good.  Almost ready to merge, but I have a few requested tweaks.

1. please look over & test the function I added to toggle the options for sending to users of different levels.

2. for the "email faculty" option, please change the query to send to all users of type faculty or superuser.  (currently it just grabs the faculty users)

3. if we unselect all checkboxes, please have a red popup error box indicating that no users are matched -- this should be similar to leaving the subject or body of the email message blank.  It should not attempt to send and should not delete your draft, it should wait for you to select at least one checbox.

4. if it's easy to do...  can we add a count of the number of users that were matched to the green success message?  E.g., "192 emails queued to be sent" or similar.

thanks!

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1126,6 +1126,52 @@ public function getGroupForUserInClass($semester, $course_name, $user_id) {
...
+        }
+        if ($faculty) {
+            $this->submitty_db->query(
+                "SELECT user_id FROM users WHERE user_access_level <= 2"
+            );
+            $results = $this->submitty_db->rows();
+            foreach ($results as $row) {
+                $result_rows[] = $row["user_id"];
+            }
+        }
```
MasterOdin, line 1171: What's the efficiency of this query for a very large number of users, especially when run multiple times and iterated through when showing the email page?

Is it cheaper to do all of the logic user group comparison logic + duplicate removal within the DB, like:

```php
        $args = [];
        $extra_join = '';
        $extra_where = '';
        if ($instructor) {
            $args[] = '1';
        }
        if ($fullAccess) {
            $args[] = '2';
        }
        if ($limitedAccess) {
            $args[] = '3';
        }
        if ($student) {
            $args[] = '4';
        }
        if ($faculty) {
            $extra_join = ' INNER JOIN users ON courses_users.user_id = users.user_id ';
            $extra_where = ' OR users.user_access_level <= 2';
        }
        $result_rows = [];
        $this->submitty_db->query(
            "SELECT DISTINCT courses_users.user_id as user_id
                FROM courses_users INNER JOIN courses 
                ON courses_users.semester = courses.semester AND courses_users.course = courses.course
                " . $extra_join . "
                WHERE courses.status = 1 AND user_group IN (" . implode(', ', $args) . ")" . $extra_where
        );
        return array_map(function ($row) { return $row['user_id']; }, $this->submitty_db->rows());
```

Though, also, could remove the `array_map in favor of modifying where the results are used to just use the associative array.

site/app/controllers/superuser/SuperuserEmailController.php
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * @Route("/superuser/email")
+     * @return MultiResponse
+     */
+    public function showEmailPage(): MultiResponse {
+        $num_faculty = count($this->core->getQueries()->getActiveUserIds(false, false, false, false, true));
+        $num_instructor = count($this->core->getQueries()->getActiveUserIds(true, false, false, false, false));
+        $num_full_access = count($this->core->getQueries()->getActiveUserIds(true, true, false, false, false)) - $num_instructor;
+        $num_limited_access = count($this->core->getQueries()->getActiveUserIds(true, true, true, false, false)) - $num_instructor - $num_full_access;
+        $num_student = count($this->core->getQueries()->getActiveUserIds(true, true, true, true, false)) - $num_instructor - $num_full_access - $num_limited_access;
```
MasterOdin, line 34: While convenient to re-use the query like this, this should be using a second DB query that just does the count within the query as that'll be quicker and use less memory.

site/app/views/email/EmailStatusView.php
```diff
@@ -9,14 +9,28 @@
 
 class EmailStatusView extends AbstractView {
     public function showEmailStatus($emailStatus) {
-        $this->core->getOutput()->addBreadcrumb("Email", $this->core->buildCourseUrl(["email"]));
+        $this->core->getOutput()->addBreadcrumb("Email Statuses", $this->core->buildCourseUrl(["email"]));
         $this->core->getOutput()->addVendorJs('bootstrap/js/bootstrap.bundle.min.js');
```
MasterOdin, line 13: What from bootstrap JS are we using on the email status page?

site/app/controllers/superuser/SuperuserEmailController.php
```diff
@@ -0,0 +1,88 @@
...
+            return JsonResponse::getFailResponse("Email content is empty.");
+        }
+        elseif (!isset($_POST['email_subject']) || $_POST['email_subject'] == '') {
+            return JsonResponse::getFailResponse("Email subject is empty.");
+        }
+        else {
+            // Because AJAX stringifies everthing
+            $email_instructor =  isset($_POST['email_instructor']) && $_POST['email_instructor'] == "true";
+            $email_full_access = isset($_POST['email_full_access']) && $_POST['email_full_access'] == "true";
+            $emailLimitedAccess = isset($_POST['email_limited_access']) && $_POST['email_limited_access'] == "true";
```
MasterOdin, line 62: snake_case: `$email_limited_access`

MasterOdin (changes requested):

site/app/templates/superuser/SuperuserEmailStatus.twig
```diff
@@ -0,0 +1,10 @@
+<div id="email" class="content">
```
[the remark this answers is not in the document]
HenrikLam: This is the same as `EmailStatus.twig`, and I am not sure if I should keep a seperate page for this. I initially was planning to make them different but after more design, I decided not to but though keeping the page might make later changes easier. Should I remove this file and make them use the same twig for now?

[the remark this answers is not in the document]
MasterOdin: Yeah, remove it. Easy enough to re-add later if it's decided that they're different enough to warrant separate pages.

site/app/libraries/database/DatabaseQueries.php
HenrikLam: Although I have very little knowledge of databases, I would imagine that my solution will be very inefficient for large amount of users, and will likely be much cheaper to use your solution. I have, however, modified `args` to initialize to `[-1]`, which is an invalid group in order to avoid a situation where the query becomes invalid when only faculty is selected. Is this a bad way to work around this situation?

site/app/controllers/superuser/SuperuserEmailController.php
HenrikLam: `public function countActiveUsersByGroup(): array {
        $this->submitty_db->query(
            "WITH A as (SELECT DISTINCT ON(user_id) user_id, user_group FROM courses_users JOIN courses
            on courses.course = courses_users.course
            and courses.semester = courses_users.semester
            WHERE courses.status = 1
            ORDER BY user_id, courses_users.user_group ASC)
            SELECT A.user_group, COUNT(A.user_group) FROM A GROUP BY A.user_group"
        );
        $result = [];
        foreach ($this->submitty_db->rows() as $row) {
            if ($row['user_group'] == 1) {
                $result['instructor'] = $row['count'];
            }
            elseif ($row['user_group'] == 2) {
                $result['full_access'] = $row['count'];
            }
            elseif ($row['user_group'] == 3) {
                $result['limited_access'] = $row['count'];
            }
            else {
                $result['student'] = $row['count'];
            }
        }
        $this->submitty_db->query(
            "SELECT COUNT(user_access_level) FROM users WHERE user_access_level <= 2"
        );
        $result['faculty'] = $this->submitty_db->row()['count'];
        return $result;
    }`
This is the query I've wrote for the counting. I separated the query for faculty however, because the results are from two different tables anyway and I don't know a good way to join them for counting. How does it look?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1126,6 +1126,84 @@ public function getGroupForUserInClass($semester, $course_name, $user_id) {
...
+                $result['student'] = $row['count'];
+            }
+        }
+        $this->submitty_db->query(
+            "SELECT COUNT(user_access_level) FROM users WHERE user_access_level <= 2"
+        );
+        $result['faculty'] = $this->submitty_db->row()['count'];
+        return $result;
+    }
+
```
HenrikLam, line 1206: @MasterOdin This is the query I've wrote for the counting. I separated the query for faculty however, because the results are from two different tables anyway and I don't know a good way to join them for counting. How does it look?

MasterOdin: This is fine as to get that to work would require another subquery anyway so not much performance would be gained in making it one query.

MasterOdin (approved):

PR #6599, opened 2021-06-11
[Refactor:CourseMaterials] Migrate Course Materials Config to DB

### LINKED ISSUE #6582: Refactor Course Materials (software)
To support Issue #6360 and other new requested course materials features.
And to make the code more robust and more maintainable...

We would like to refactor course materials:
* Move to a more standard model-view-controller code design
* Eliminate the  /var/local/submitty/courses/SEMESTER/COURSE/uploads/course_materials_file_data.json file and store this information in new database table(s).

This should be done carefully, with appropriate migrations, to not disrupt courses currently using the existing course materials functionality.

@emaicus or others, please add notes on refactor/design.

cjreed121:

Course Materials currently relies on a JSON file stored in the file system to determine permissions, dates, and more of the files in course materials.
Closes #6582

### What is the new behavior?
Now Course Materials will have a model and table in the database. This will improve future maintainability of the course materials as we can have a full MVC layout for it.

### Other information?
This is a work in progress. The database migration and model are correctly setup. Viewing the page works somewhat but the controller and view still need to be migrated to the new model.

### Progress
* [x] Create migration
* [x] Export SQL of migration
* [x] Create model
* [x] Refactor controller to support the model
* [x] Refactor view to support the model
* [x] Write tests


migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
```diff
@@ -0,0 +1,121 @@
...
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    # create tables here
+    database.execute(
+        """
+        CREATE TABLE IF NOT EXISTS course_materials (
+            url TEXT PRIMARY KEY,
```
shailpatels, line 24: I would suggest changing `url` to `path` or `file_path` since course materials can be links containing only a URL so it might be confusing what value this is supposed to represent 

Unless this is supposed to be the fully formed URL to access the resource from the server

site/app/controllers/AbstractController.php
```diff
@@ -280,4 +281,35 @@ protected function tryGetAutocheck(AutoGradedTestcase $testcase, string $autoche
...
     }
+
+    /**
+     * Gets a course material from its path.
+     * @param string $path
+     * @param bool $render_json true to render a JSEND response to the output in the failure/error case
+     * @return CourseMaterial|bool false in the fail/error case
+     */
+    public function tryGetCourseMaterial(string $path, bool $render_json = true) {
+        if ($path === '') {
```
shailpatels, line 292: You could use `file_exists` here to see if the path points to a valid file/dir which would be a stronger check and probably more common that someone passes a broken path then an empty string

site/app/models/CourseMaterial.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->setReleaseDate(DateUtils::parseDateTime($details['release_date'], $this->core->getUser()->getUsableTimeZone()));
+        $this->setHiddenFromStudents($details['hidden_from_students']);
+        $this->setPriority($details['priority']);
+        $this->setSectionLock($details['section_lock']);
+        if (isset($details['sections']) && $details['sections'][0] != null) {
+            $this->setSections($details['sections']);
+        }
+    }
+
+    public function isFile(): bool {
```
shailpatels, line 65: `isFile` can be simplied to

```php
return $this->getType() === self::FILE;
```

and similar thing for isLink

shailpatels (changes requested):

migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
```diff
@@ -0,0 +1,121 @@
...
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    # create tables here
+    database.execute(
+        """
+        CREATE TABLE IF NOT EXISTS course_materials (
+            path TEXT PRIMARY KEY,
```
MasterOdin, line 24: need to update the `course_db.sql` with this new name. Also, this should probably be `varchar(255)`, unless you're expecting very very long paths?

migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
```diff
@@ -0,0 +1,121 @@
+"""Migration for a given Submitty course database."""
+
+import json
+from pathlib import Path
+from sqlalchemy import insert
+
+def up(config, database, semester, course):
```
MasterOdin, line 7: It looks to me like this `up` migration would fail if we tried to run this twice (i.e. `migrate`, `rollback`, `migrate`). Your best bet might be to replace your `inserts` with [`upserts`](https://wiki.postgresql.org/wiki/UPSERT)?

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -177,115 +183,97 @@ public function modifyCourseMaterialsFileTimeStamp($filenames, $newdatatime) {
 
         //Check if the datetime is correct
         if (\DateTime::createFromFormat('Y-m-d H:i:sO', $new_data_time) === false) {
-            return $this->core->getOutput()->renderResultMessage("ERROR: Improperly formatted date", false);
+            return JsonResponse::getErrorResponse("Improperly formatted date");
         }
 
         $new_data_time = DateUtils::parseDateTime($new_data_time, $this->core->getUser()->getUsableTimeZone());
-        $new_data_time = DateUtils::dateTimeToString($new_data_time);
+        //$new_data_time = DateUtils::dateTimeToString($new_data_time);
```
MasterOdin, line 190: delete?

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -177,115 +183,97 @@ public function modifyCourseMaterialsFileTimeStamp($filenames, $newdatatime) {
...
+            return JsonResponse::getErrorResponse("Something went wrong while saving.");
+        }
+        return JsonResponse::getSuccessResponse("Time successfully set.");
     }
 
     /**
      * @Route("/courses/{_semester}/{_course}/course_materials/edit", methods={"POST"})
      * @AccessControl(role="INSTRUCTOR")
      */
     public function ajaxEditCourseMaterialsFiles() {
```
MasterOdin, line 229:
```suggestion
    public function ajaxEditCourseMaterialsFiles(): JsonResponse {
```

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -467,61 +449,52 @@ public function ajaxUploadCourseMaterialsFiles() {
...
-                            }
-                            else {
-                                $json[$path] = [
-                                    'release_datetime' => $release_time,
-                                    'hide_from_students' => $hide_from_students,
-                                    'external_link' => $is_external_link_file,
-                                    'sort_priority' => $sort_priority
-                                ];
-                            }
+                            $details['type'][$i] = $is_external_link_file ? 1 : 0;
```
MasterOdin, line 455: Use `CourseMaterial::LINK` and `CourseMaterial::FILE` instead of magic numbers. This applies to other places you have `1` and `0`.

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -467,61 +449,52 @@ public function ajaxUploadCourseMaterialsFiles() {
...
-                                $sections_exploded = @explode(",", $sections);
-                                if ($sections_exploded == null) {
-                                    $sections_exploded = [];
-                                }
-                                $json[$dst] = ['release_datetime' => $release_time, 'sections' => $sections_exploded, 'hide_from_students' => $hide_from_students, 'external_link' => $is_external_link_file, 'sort_priority' => $sort_priority];
-                            }
-                            else {
-                                $json[$dst] = ['release_datetime' => $release_time, 'hide_from_students' => $hide_from_students, 'external_link' => $is_external_link_file, 'sort_priority' => $sort_priority];
-                            }
+                            $details['type'][0] = $is_external_link_file ? 1 : 0;
```
MasterOdin, line 465: Like here

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7175,6 +7176,160 @@ public function getGradeableAccessTeam(
...
+                  course_material_id,
+                  section_id)
+                VALUES (?, ?)",
+                        $params
+                    );
+                }
+            }
+            return true;
+        }
+        catch (DatabaseException $e) {
```
MasterOdin, line 7263: why are you catching this exception?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7175,6 +7176,160 @@ public function getGradeableAccessTeam(
...
```
MasterOdin, line 7311: same here?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7175,6 +7176,160 @@ public function getGradeableAccessTeam(
...
+            return false;
+        }
+    }
+
+    public function deleteCourseMaterial(string $path): bool {
+        try {
+            $this->course_db->query("DELETE FROM course_materials WHERE path=?", [$path]);
+            return true;
+        }
+        catch (DatabaseException $e) {
```
MasterOdin, line 7321: and here?

MasterOdin (changes requested):

site/app/controllers/AbstractController.php
```diff
@@ -280,4 +281,35 @@ protected function tryGetAutocheck(AutoGradedTestcase $testcase, string $autoche
...
         }
     }
+
+    /**
+     * Gets a course material from its path.
+     * @param string $path
+     * @param bool $render_json true to render a JSEND response to the output in the failure/error case
+     * @return CourseMaterial|bool false in the fail/error case
+     */
+    public function tryGetCourseMaterial(string $path, bool $render_json = true) {
```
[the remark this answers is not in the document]
cjreed121: I did this for future use. I plan to add features after this PR and this would be very useful to have on all controllers.

site/app/controllers/AbstractController.php
```diff
@@ -280,4 +281,35 @@ protected function tryGetAutocheck(AutoGradedTestcase $testcase, string $autoche
...
+        if ($path === '') {
+            if ($render_json) {
+                $this->core->getOutput()->renderJsonFail('Missing path parameter');
```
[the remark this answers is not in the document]
cjreed121: Yeah that makes sense. I was just following other existing ones. I don't think tryGetCourseMaterial should necessarily throw an exception but rather give you value or false. The part try tells the developer that it is going to try and handle the exception and you don't. If there is an error or exception you would expect the value false. Thoughts?

migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
cjreed121: I was considering varchar(255) but wasn't sure if any instructor would ever have a subdirectory in addition to the normal directory that exceeds 255 chars. It is a limit we could establish and force. I know a 255 char path is already very long but if anyone was to hit it a warning would need to be shown. Thoughts?

migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
cjreed121: Yeah this is a good point. I just updated the queries to add a on conflict to do upserts. Can you take a look at the new statements?

site/app/libraries/database/DatabaseQueries.php
cjreed121: This is to answer this one and all the ones below. I did this so the controller could tell if the query was successful. The main reason I did this was for testing to mock those functions and return a true or false. Can you look at the CourseMaterialsControllerTester.php and see if there is a better way I can check and validate the whole DatabaseQueries functions?

MasterOdin: You could use:

https://github.com/Submitty/Submitty/blob/d76e545d5ec67e466265ba67d05d085607f60ae0/site/app/libraries/database/AbstractDatabase.php#L316-L327

to determine the affected number of rows? Depending on what you're trying to catch here (insert duplicate path?), I would suggest adding a check within the controller to protect against that condition  there, and the DB query operates expecting it to always succeed.

Catching the `DatabaseException` will catch both whatever normal condition you're expecting, but also if the query gets malformed, database dies unexpectedly, etc, which we definitely do not want to catch and is something we want to log.

cjreed121: This is how I currently verify if the argument is valid.

https://github.com/Submitty/Submitty/blob/7617456fe76d59e9a7187d76a30385e5d801c014/site/tests/app/controllers/course/CourseMaterialsControllerTester.php#L101-L120

This would make sure the information being passed into the function is correct which is what I want to do be doing. How can I switch this around so I don't have to use the true/false values with the catches? I am not necessarily looking to see the number of rows changed but rather what data went in. Or am I doing testing incorrectly?

cjreed121:
@MasterOdin To save this from needing another refactor soon, should I swap out the DatabaseQueries for doctrine as that is now merged?

migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
MasterOdin: Something to consider is that the maximum file path length on Windows is 255 characters. Given a zip of all course materials, if the path exceeds that value, then people on Windows would never be able to extract all files from that zip.

So we probably actually want to set the limit for max path _lower_ than 255 to help protect against that case, though I don't think we ever considered that in the original implementation.

migration/migrator/data/course_tables.sql
```diff
@@ -168,6 +168,30 @@ CREATE SEQUENCE public.categories_list_category_id_seq
 ALTER SEQUENCE public.categories_list_category_id_seq OWNED BY public.categories_list.category_id;
 
 
+--
+-- Name: course_materials; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.course_materials (
+    path text NOT NULL,
```
MasterOdin, line 176: A though occurs while looking through this PR, would we benefit here from having a regular serial primary key `id` here and make `path` a unique key on the table? That would allow not having to worry about dealing with serialization/unserialization errors on GET paths (e.g. like on deleting a course material).

cjreed121:
@MasterOdin I know I will need lint/static/unit tests changes still but can you take a look at the specific doctrine changes? I do have a working version but would like feedback to improve this since it's my first time in doctrine. For example on lines 243-256 in CourseMaterialsController, is there a better way to create the new ManyToOne relationship entities? Also I'm unsure of how much I need to define in the migration vs how much doctrine will just handle?

Lastly, will we be using some sort of AbstractEntity here? I think it could be useful for having magic methods for most of the getters/setters rather than needing to write them all.

migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
MasterOdin: It looks like you've re-added dropping the tables in the down part of the migration that it would render the upsert obsolete and you can just do a regular insert.

cjreed121: The plan is to not have the down migration. I had those for testing only. I do plan on sticking with the upserts.

site/app/entities/course/CourseMaterial.php
```diff
@@ -0,0 +1,175 @@
...
+        $this->setPath($details['path']);
+        $this->setReleaseDate($details['release_date']);
+        $this->setHiddenFromStudents($details['hidden_from_students']);
+        $this->setPriority($details['priority']);
+        $this->sections = new ArrayCollection();
+    }
+
+    /**
+     * @return Collection
+     */
```
MasterOdin, line 76: `Collection` in Doctrine is a "generic" type, which allows for specifying what it has by using the `<>` notation within the docstring. In this case, it would be `Collection<CourseMaterialSection>`, and this then allows phpstan to be aware of this info on static analysis.

site/app/entities/course/CourseMaterial.php
```diff
@@ -0,0 +1,175 @@
...
+        return $this->sections;
+    }
+
+    public function getPath(): string {
+        return $this->path;
+    }
+
+    /**
+     * @return int
+     */
```
MasterOdin, line 87: For these sorts of generic types, I would just leave off the docstring altogether as it provides no meaningful information that isn't just in the function signature itself.

site/app/libraries/Core.php
```diff
@@ -173,7 +173,7 @@ public function setSessionManager(SessionManager $manager) {
     private function createEntityManager(AbstractDatabase $database): EntityManager {
         $config = Setup::createAnnotationMetadataConfiguration(
             [FileUtils::joinPaths(__DIR__, '..', 'app', 'entities')],
-            false,
+            $this->config->isDebug(),
```
MasterOdin, line 176: Change makes sense, curious if you knew exactly what turning on debug mode here does versus regular behavior.

MasterOdin (commented):
I'm confused on what the vision for the CourseMaterials stuff is here as it looks like there's still a lot of dependency on file system traversal and using it as a point of truth to then update the DB and such. This is particularly true in the `listCourseMaterials` view function where it scans through the filesystem, and does some reconcillation with what's in the DB, but also totally optional if it is in the DB? I fear that this is introducing a lot of subtle complication to how this system works that's not going to be easy to maintain going forward.

With regards to the doctrine specific stuff, it looks mostly fine, with the main concern being that it looks like you call `->flush()` several times within given function. Do we need that explicit flush, or can we wait till the very end of the function and call flush once, allowing doctrine to potentially do deeper optimizations.

migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
MasterOdin: Given the nature of the change, I think dropping the tables is fine. Having thought about this more, this is an instance where we want to do a "true" rollback, where within the migration you:

1. create the tables
2. read the JSON into the tables

and so then in the rollback, you'd:

1. read the tables into the JSON file
2. drop the tables

as the source of truth shifts from JSON file to tables within this PR, and vice versa if we were to rollback.

site/app/views/course/CourseMaterialsView.php
```diff
@@ -176,12 +56,12 @@ public function listCourseMaterials($user) {
                 $priorities[$material] = 0.0;
             }
             $path = FileUtils::joinPaths($expected_path, $material);
-            $priorities[$material] += isset($json[$path]['sort_priority']) ? $json[$path]['sort_priority'] : $sort_default;
+            $priorities[$material] += isset($course_materials[$path]) ? $course_materials[$path]->getPriority() : $sort_default;
```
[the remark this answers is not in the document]
bmcutler: The direct editing of the directory structure of course materials was a nightmare -- if you didn't do it perfectly you bricked the course materials page for your course.  I no longer have hope of supporting this.
All course materials edits should be done via the website (or eventually an API?).  All details should be in the database.  

We are transitioning away from the .json file -- we won't delete it in this PR though (just in case we have bugs).  All information about all course materials items will be in the database.  Any remnants of the original course materials design (that displayed stuff in the file system, even if it wasn't in the .json file) can be deleted.

migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
bmcutler: I vote for not wasting time implementing the (nontrivial) down migration.  I vote for an empty down migration.  This is a big PR and a big change, but this change is LONG overdue.  We will not be looking backward once we switch to the database.
If we do make a mistake, we might lose metadata if we have to revert and bugfix and reapply, but I am ok with the potential data loss.  Importantly, we won't lose any of the course materials files.

site/app/views/course/CourseMaterialsView.php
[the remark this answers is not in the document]
cjreed121: Alright. I think I will remove directory traversal 100% and have the database tell PHP what to read from filesystem. I saw in the code how you could upload a file and course materials would auto configure the json for it but I think switching to doing it through the website is the better option.

site/app/libraries/Core.php
cjreed121: I forget exactly what I needed it for but after turning it on, things started to work for me. It might have been related to caching. I saw you made a PR for this. I'll remove the changes from this branch and pull master.

cjreed121:
The maximum number of times flush is called is 2 times in the ajax edit function. I tried to keep it at 1 but I was given errors about duplicate keys. It seemed that I had to call flush after the clear then I could start adding new sections.

shailpatels (changes requested):
I think there's an issue with duplicate file uploads, if you upload a file and then another with the same name you get a server error sent back to the console:

```
    <h1 style="margin-left: 20px; margin-top: 10px;">Server Error</h1>
    <div style="position: absolute; top: 144px; left: 362px; border: 1px dashed black; padding: 10px; font-family: monospace">
        Oh no! Something irrecoverable has happened...<br /><br />
        Doctrine\DBAL\Exception\UniqueConstraintViolationException (Code: 0) thrown in /usr/local/submitty/site/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractPostgreSQLDriver.php (Line 70) by:<br />
 return new UniqueConstraintViolationException($message, $exception);<br />
<br />
<br />
Message:<br />
An exception occurred while executing 'INSERT INTO course_materials (id, path, type, release_date, hidden_from_students, priority) VALUES (?, ?, ?, ?, ?, ?)' with params [230, "\/var\/local\/submitty\/courses\/f21\/sample\/uploads\/course_materials\/file5.txt", 0, "9998-01-01 00:00:00-0500", 0, 0]:<br />
<br />
SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint "course_materials_path_key"<br />
DETAIL: Key (path)=(/var/local/submitty/courses/f21/sample/uploads/course_materials/file5.txt) already exists.<br />
<br />
Stack Trace:<br />
#0 /usr/local/submitty/site/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php(182): Doctrine\DBAL\Driver\AbstractPostgreSQLDriver->convertException()<br />
#1 /usr/local/submitty/site/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php(159): Doctrine\DBAL\DBALException::wrapException()<br />
#2 /usr/local/submitty/site/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(2212): Doctrine\DBAL\DBALException::driverExceptionDuringQuery()<br />
#3 /usr/local/submitty/site/vendor/doctrine/dbal/lib/Doctrine/DBAL/Statement.php(183): Doctrine\DBAL\Connection->handleExceptionDuringQuery()<br />
#4 /usr/local/submitty/site/vendor/doctrine/orm/lib/Doctrine/ORM/Persisters/Entity/BasicEntityPersister.php(287): Doctrine\DBAL\Statement->execute()<br />
#5 /usr/local/submitty/site/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php(1122): Doctrine\ORM\Persisters\Entity\BasicEntityPersister->executeInserts()<br />
#6 /usr/local/submitty/site/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php(423): Doctrine\ORM\UnitOfWork->executeInserts()<br />
#7 /usr/local/submitty/site/vendor/doctrine/orm/lib/Doctrine/ORM/EntityManager.php(378): Doctrine\ORM\UnitOfWork->commit()<br />
#8 /usr/local/submitty/site/app/controllers/course/CourseMaterialsController.php(580): Doctrine\ORM\EntityManager->flush()<br />
#9 unknown file(unknown line): app\controllers\course\CourseMaterialsController->ajaxUploadCourseMaterialsFiles()<br />
#10 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(218): call_user_func_array()<br />
#11 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(196): app\libraries\routers\WebRouter->run()<br />
#12 /usr/local/submitty/site/public/index.php(129): app\libraries\routers\WebRouter::getWebResponse()<br />


                      ,--.    ,--.
                     (( O))--(( O))
                   ,'_`--'____`--'_`.
                  _:  ____________  :_      _____
                 | | ||::::::::::|| | |        \ \
                 | | ||::::::::::|| | |         \
                 | | ||::::::::::|| | |
                 |_| |/__________\| |_|
                   |________________|
                __..-'            `-..__
             .-| : .----------------. : |-.
           ,\ || | |\______________/| | || /.
          /`.\:| | ||  __  __  __  || | |;/,'\
         :`-._\;.| || '--''--''--' || |,:/_.-':
         |    :  | || .----------. || |  :    |
         |    |  | || '----SSt---' || |  |    |
         |    |  | ||   _   _   _  || |  |    |
         :,--.;  | ||  (_) (_) (_) || |  :,--.;
         (`-'|)  | ||______________|| |  (|`-')
          `--'   | |/______________\| |   `--'
                 |____________________|
                  `.________________,'
                   (_______)(_______)
                   (_______)(_______)
                   (_______)(_______)
                   (_______)(_______)
                  |        ||        |
                  '--------''--------'

```

This can also occur if a file already exists within a folder and an instructor uses optional paths and the file tree matches.

I believe course materials just overwrites the file, not sure if we want to keep that functionality if any professors use that as a method of updating files (@bmcutler ?) but otherwise should be a check against duplicate files being inserted to the db.

cjreed121:
@shailpatels Thanks for noting this issue. I made sure no new path will be created when there is an existing identical path already. I implemented so a warning is given rather than overwriting anything. If an instructor wants to change a file they can just delete the old one and re-upload. Otherwise it may be very easy to overwrite current files.

loadedskiffer (approved):
I set up a bunch of course materials, switched to this branch and then made the migrations and all of the data was correct. I couldn't break anything on the page and it looks good.

eddiekrystowski (approved):
Works good, I tested it with many different types of materials/structures and nothing broke, everything seemed to work as normal.

shailpatels (approved):

HenrikLam (approved):
Was not able to break it. Looks good to me!

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -465,63 +497,96 @@ public function ajaxUploadCourseMaterialsFiles() {
...
+        }
+        if ($dirs_to_make != null) {
+            $i = -1;
+            $new_paths = [];
+            foreach ($dirs_to_make as $dir) {
+                $cm = $this->core->getCourseEntityManager()->getRepository(CourseMaterial::class)->findOneBy(
+                    ['path' => $dir]
+                );
+                if ($cm == null && !in_array($dir, $new_paths)) {
+                    $details['type'][$i] = CourseMaterial::DIR;
```
MasterOdin, line 565: Why are you making `DIR` type course materials here, but not when you convert from the JSON file to DB structure?

site/app/libraries/CourseMaterialsUtils.php
```diff
@@ -11,67 +12,35 @@ class CourseMaterialsUtils {
...
      * @return bool Indicates if the file has been released or not
      * @throws MalformedDataException An error occurred parsing the file's release time data
      */
-    public static function isMaterialReleased(array $json, string $path): bool {
-        if (!array_key_exists($path, $json)) {
+    public static function isMaterialReleased(CourseMaterial $course_material): bool {
+        if ($course_material == null) {
             return false;
         }
         else {
```
MasterOdin, line 23: You can get rid of the `else` here and de-indent the body.

site/app/libraries/CourseMaterialsUtils.php
```diff
@@ -11,67 +12,35 @@ class CourseMaterialsUtils {
...
      * @param User $current_user the current user
      * @return bool Indicates if the file has been released or not
      */
-    public static function isSectionAllowed(array $json, string $path, User $current_user): bool {
-        if (!array_key_exists($path, $json)) {
+    public static function isSectionAllowed(CourseMaterial $course_material, User $current_user): bool {
+        if ($course_material == null) {
             return false;
         }
         else {
```
MasterOdin, line 42: You can get rid of the `else` here and de-indent the body.

MasterOdin (changes requested):

site/app/controllers/course/CourseMaterialsController.php
cjreed121: In the migration these types do get created at the end:

https://github.com/Submitty/Submitty/blob/f2ce5e630a7f169d8a5406124d47461c160018fc/migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py#L121-L141

migration/migrator/migrations/course/20210610100802_course_materials_json_to_db.py
```diff
@@ -0,0 +1,160 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("DROP TABLE IF EXISTS course_materials CASCADE")
+    database.execute("DROP TABLE IF EXISTS course_materials_sections CASCADE")
```
MasterOdin, line 159: Is this meant to be changed to not drop the tables? I'm confused as to the agreed upon direction of this migration.

MasterOdin (commented):
PR seems generally fine. I'm a bit concerned that we're still passing around paths to look up course materials where if the entire course material directory tree is now mapped to materials (including folders), we should be able to pass around course material IDs and then grab the path from the DB. This would allow us to avoid thinking about encoding / decoding paths as we pass it around the forms.

cjreed121:
> PR seems generally fine. I'm a bit concerned that we're still passing around paths to look up course materials where if the entire course material directory tree is now mapped to materials (including folders), we should be able to pass around course material IDs and then grab the path from the DB. This would allow us to avoid thinking about encoding / decoding paths as we pass it around the forms.
> 
> However, this can be done in a follow-up PR.

Are you talking about the display_file, read_file, etc routes where they use the full routes? If so I can definitely see how those can be shortened. I plan on making quite a few PRs after this with new features as I want this one to be strictly moving to DB. I'll have this be one of the PRs.

PR #6603, opened 2021-06-14
[Bugfix:InstructorUI] Change Late Day Timestamps to Dates

### LINKED ISSUE #6053: Late days allowed should not store effective dates with hours/minutes/seconds

cjreed121:

Currently late days are stored with a full timestamp.
Closes #6053

### What is the new behavior?
Late days are now only stored with a date. The precision down to the second is not needed for late days.

### Other information?
This does change all timestamps for late days into a date by casting it.

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -559,7 +559,7 @@ CREATE TABLE public.late_day_exceptions (
 CREATE TABLE public.late_days (
     user_id character varying(255) NOT NULL,
     allowed_late_days integer NOT NULL,
-    since_timestamp timestamp(6) with time zone NOT NULL
+    since_timestamp date NOT NULL
 );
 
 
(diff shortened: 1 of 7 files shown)


MasterOdin:
Has this been tested with a mix of timezones around a late day cutoff? So let's say an instructor in an EST institution sets the late day to '2021-06-16'. A student in PST submits at `2021-06-15 23:59:00` or a student in UTC submits at `2021-06-16 00:01:00`. I would expect the former to get a late day while the latter should not based on timezone conversions to EST.

cjreed121:
> Has this been tested with a mix of timezones around a late day cutoff? So let's say an instructor in an EST institution sets the late day to '2021-06-16'. A student in PST submits at `2021-06-15 23:59:00` or a student in UTC submits at `2021-06-16 00:01:00`. I would expect the former to get a late day while the latter should not based on timezone conversions to EST.

It seemed to be relying upon the user's timezone for late days. I just changed it in SimpleLateUser to entirely rely upon the system's timezone. When the DateTime object actually gets created it's time is set to 00:00:00 so any time during that day will be valid. Since it uses system time rather than a user's preference/their time, this should not be an issue.

PR #6604, opened 2021-06-14
[Bugfix:Plagiarism] Multiple submissions bugfix

williamjallen:
A PHP error currently occurs when Plagiarism is run on gradeables which have multiple submissions.

### What is the new behavior?
Bug was fixed by adding checking to make sure an index in an array existed before attempting to access that index.

### DIFF
--- a/site/app/libraries/plagiarism/PlagiarismUtils.php
+++ b/site/app/libraries/plagiarism/PlagiarismUtils.php
@@ -61,11 +61,13 @@ class PlagiarismUtils {
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


sbelsk:
I don't seem to have this error. How can I reproduce the issue this PR is supposed to solve?

williamjallen:
I can only reproduce it on one test case I created and it doesn't appear to affect anything else so I suppose I will close the PR..

sbelsk:
This issue popped up again and I was able to recreate it. This is being addressed and resolved in PR https://github.com/Submitty/Submitty/pull/6612, so I'm closing this one.

Issue #6568, opened 2021-06-07
Plagiarism Small Bugs Todo List


sbelsk:
**Describe the bug**
- [x] There is an off by one error in the front-end of the plagiarism results page where the last matching region/token of a type category (matches of type "match", "common", "provided") is not highlighted
![image](https://user-images.githubusercontent.com/71195502/121085997-9e5d0580-c7b0-11eb-9596-da7c9ac1ec1d.png)

- [x] We need to not make `matches.json` files for submissions that have no matches of type "match", even if they have match of type "provided" and/or "common"
- [ ] Fix/discuss percentages displayed for user 1 and user 2 on the side-by-side display
- [x] The "Last Run" and "Submissions" columns never display any data: ![image](https://user-images.githubusercontent.com/71195502/121201197-55559180-c842-11eb-8543-18f349778ebb.png)
- [x] The message displayed in this pop-up when clicking the "delete" button on a plagiarism run:
![image](https://user-images.githubusercontent.com/71195502/121201883-dca30500-c842-11eb-8c5b-2ce94f01d44f.png)
- [ ] The legend for the highlighted matches has no borders for the colors, so the color white for "unique matches" cannot be seen in light mode
![image](https://user-images.githubusercontent.com/71195502/121206495-82a43e80-c846-11eb-947a-b9807e20c217.png)

- [x] The new Plagiarism Detection Configuration page should not allow you to submit or create a new config with threshold = 1. Currently, if you do, the gradable completely disappears from the result summary page.

sbelsk:
- [ ] The "Last Run" and "Submissions" columns never display any data
![image](https://user-images.githubusercontent.com/71195502/121201197-55559180-c842-11eb-8543-18f349778ebb.png)

williamjallen:
@MasterOdin Not everything on this list has been completed.  Should this be reopened?

MasterOdin:
Can you make an issue for the remaining tasks? Having these omnibus issues makes it annoying to deal with

williamjallen:
Ok.  We just didn't want to create tens of small, trivial issues.  We'll separate the remaining tasks into new issues.

MasterOdin:
As I've said previously, would be good to try not to open these sort of omnibus issues where each bug / task is discrete and will have its own individual PR for it. This makes it easier to book-keep things, as well as understand the link between issue and PR at a glance.

PR #6605, opened 2021-06-14
[Bugfix:Plagiarism] Fix sequence length error checking

williamjallen:
It is currently possible to trigger an assertion failure in `compare_matches.cpp` from the UI by setting a sequence length of 1.

### What is the new behavior?
This PR changes the minimum allowable sequence length to 2 to bring the PHP into alignment with the C++ backend.

Fixes part of #6568

### DIFF
--- a/site/app/controllers/admin/PlagiarismController.php
+++ b/site/app/controllers/admin/PlagiarismController.php
@@ -296,7 +296,7 @@ class PlagiarismController extends AbstractController {
             $this->core->redirect($return_url);
         }
 
-        if (isset($_POST['sequence_length']) && $_POST['sequence_length'] !== '' && is_numeric($_POST['sequence_length']) && $_POST['sequence_length'] > 0) {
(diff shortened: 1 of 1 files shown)


site/app/controllers/admin/PlagiarismController.php
```diff
@@ -296,7 +296,7 @@ public function saveNewPlagiarismConfiguration($new_or_edit, $gradeable_id = nul
             $this->core->redirect($return_url);
         }
 
-        if (isset($_POST['sequence_length']) && $_POST['sequence_length'] !== '' && is_numeric($_POST['sequence_length']) && $_POST['sequence_length'] > 0) {
+        if (isset($_POST['sequence_length']) && $_POST['sequence_length'] !== '' && is_numeric($_POST['sequence_length']) && $_POST['sequence_length'] >= 2) {
             $sequence_length = $_POST['sequence_length'];
         }
         else {
```
sbelsk, line 302: Could we maybe add the minimum number required for the sequence length to the error message (line 303)?
Something like:
```php
$this->core->addErrorMessage("Invalid input provided for sequence length. The minimum threshold value is 2");

sbelsk (approved):

PR #6606, opened 2021-06-14
[Bugfix:Submission] Fix Penalty Badges Not Showing

### LINKED ISSUE #4631: TA Grading badge missing for penalty component
**Describe the bug**
When a TA graded component is negative, the badge is missing

(Not sure if it matters if the component is penalty-only, or just if the total score is < 0 for a mark that allows penaltys and positive total)

Screen Shot 2019-10-16 at 10 36 35 AM

eddiekrystowski:
Fixes #4631 
Currently, components that have only penalty points will not have their badge shown even if the student has earned the penalty points. For a rubric like this:
image
The result is this on the student's end:
image

### What is the new behavior?
Badges from penalty only components now show if the student has earned the penalty points. All other badges remain the same.
image

### DIFF
--- a/site/app/templates/functions/Badge.twig
+++ b/site/app/templates/functions/Badge.twig
@@ -62,7 +62,11 @@
         {% if extra_credit %}
             +{{ earned }}
         {% elseif earned < 0 %}
-            &minus;{{ earned|abs }} / {{ max }}
+            {% if max != 0 %}
+                &minus;{{ earned|abs }} / {{ max }}
+            {% else %}
+                &minus;{{ earned|abs }}
+            {% endif %}
         {% else %}
             {{ earned }} / {{ max }}
         {% endif %}
--- a/site/app/templates/functions/Badge.twig
+++ b/site/app/templates/functions/Badge.twig
@@ -81,16 +81,14 @@
         {% elseif max > 0 %}
             {# Normal case always shows #}
             true
-        {% elseif max < 0 %}
+        {% else %}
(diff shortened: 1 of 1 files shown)


loadedskiffer (approved):

bmcutler (changes requested):
Wait, what happened to the badge for extra credit?

It used to (not sure how long its been broken) show as "+3" for example when the total was nonzero & positive.  That is, if the denominator was zero, it did not show the " / 0" part and it also had a "+" to indicate extra credit.  

But if the numerator & denominator was zero it hid the badge completely (to not emphasize that no extra credit was awarded).  Similarly, I think for a penalty mark, if the numerator & denominator are zero, hide the badge.

If the numerator is negative, and the denominator is zero it should show "-3" (hiding the denominator).

And if the numerator is negative and the denominator is positive it should show "-3 / 2" (or similar)

eddiekrystowski:
Okay, I will make those changes to the formatting.

When I looked at it, extra credit worked by checking off extra credit on the component in the rubric and specifying extra credit points there. I believe this should still work in this PR (I will test tomorrow). Perhaps my naming of the component "Kind of extra credit" was misleading, as that component doesn't have extra credit activated.

What that component has is a zero denominator and non-zero non-extra-credit numerator (so 3/0 for example where the 3 points are regular points not extra credit), and so there is no badge shown, which is how it was before. Would you like me to have it show a badge for this case?

Issue #6590, opened 2021-06-09
Plagiarism Provided Code not Deleted


sbelsk:
**Describe the bug**
When a config is created with provided code, if after you have run it with the provided code once, you go to edit the config and remove the provided code, when you rerun lichen, the provided code files still exist and the output is the same as if you kept it enabled. 

**Expected behavior**
When you remove the provided code, the files the directories be removed.

**To Reproduce**
1. Create config with provided code
2. Run at least once
3. Edit the config to remove the provided code files
4. Rerun
5. See error

Issue #6561, opened 2021-06-04
Lichen Edit Configuration with Uploaded File


sbelsk:
**Describe the bug**
When you edit a configuration for plagiarism detection for an assignment, if previously there was a file attached under the "provided code" category, a message is displayed reassuring you that the file is still saved.
![currnt_provided](https://user-images.githubusercontent.com/71195502/120850729-d14f9100-c545-11eb-8d4a-8860a24cd199.png)
If you don't upload a new file, but keep the "yes" option selected, then when you try to save it, the next error message is displayed:
![image](https://user-images.githubusercontent.com/71195502/120850994-30150a80-c546-11eb-993a-f56e860a55cf.png)
the only way to save the configuration and leave the page is to address the error, so the user is forced to select the "no" option and discard the file.

**Expected behavior**
If a file is uploaded to a configuration, there should be a way for the user to edit the config to either: 
- upload a new file to a config that previously had no files uploaded to it
- discard an existing file and upload a new file instead of the old on
- discard the existing file without uploading a new one.

**To Reproduce**
1. Go to the "Plagiarism Detection" tab on the left tool bar
2. Click the "+ Configure New Gradeable for Plagiarism Detection" button, upload a file and create a new config
3. Go back and edit the config with the file
4. Click "Save configuration"
5. See error

PR #6612, opened 2021-06-15
[Feature:Plagiarism] Improve regex options

williamjallen:
It is currently impossible to select files from the `results` or `checkout` directories in a gradeable.  This can cause issues, especially when running Plagiarism on PDFs which have text extracted in the autograding process.

### What is the new behavior?
This PR adds the ability to apply regular expressions to the `submissions`, `results`, and `checkout` directories.  Also, a number of bugs were fixed along the way and refactors were performed, including an overhaul of the configuration form.  The new code is simpler, more organized, and more robust and introduces more error checking to make it easier to track down future bugs

### Other information?
Miscellaneous JavaScript was removed, and copied as-is from `server.js` to the relevant Twig files.  A future refactor PR will refactor the JS content and move it to a proper JS file.

This PR depends on Lichen PR https://github.com/Submitty/Lichen/pull/34


williamjallen:
Do you have any feedback about this @MasterOdin?  There are many things wrong with the rest of the code in this part of Submitty and we plan to do a significant code review/refactor once we get it working to some extent but we were wondering if you had any feedback about the changes made here.

sbelsk (approved):
Looks good. 
This PR actually closes both https://github.com/Submitty/Submitty/issues/6590 and https://github.com/Submitty/Submitty/issues/6561.
It's also worth mentioning that the "Prior Terms Gradeables" and the "submissions to be ignored" in the config form previously were not being used by the lichen concatenation, tokenization, hashing and matching -- and this PR kept it this way. The UI and the backend are still there in case we come back to it in the future.

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -489,6 +474,33 @@ public function reRunPlagiarism($gradeable_id) {
...
     }
 
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/plagiarism/configuration/new", methods={"GET"})
+     */
+    public function configureNewGradeableForPlagiarismForm() {
+        $semester = $this->core->getConfig()->getSemester();
+        $course = $this->core->getConfig()->getCourse();
+        $gradeable_with_submission = array_diff(scandir("/var/local/submitty/courses/$semester/$course/submissions/"), ['.', '..']);
```
MasterOdin, line 484: Could replace all the usages of `/var/local/submitty/courses/$semester/$course` with `$this->core->getConfig()->getCoursePath()`

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
```diff
@@ -22,167 +23,217 @@
...
-                        <label for="all_files_id">All Files</label>
-                    </span>
-                    <span>
-                        <input type="radio" id="regex_matching_files_id" value="regex_matching_files" name="file_option" {{ regex_matching_files }}>
-                        <label for="regex_matching_files_id">Regex matching files</label>
-                    </span>
-                    <input type="text" name="regex_to_select_files" class="low-margin-top" value="{{ regex }}"/>
+                    <div id="files_to_be_compared">
+                        <span class="option-alt">
+                            <a target=_blank href="https://https://submitty.org/instructor/course_management/plagiarism">How do I specify a regex expression?
```
MasterOdin, line 72: Looks like you have a double `https://` here.

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
```diff
@@ -22,167 +23,217 @@
...
+                                <option value="{{ sem }}">{{ sem }}</option>
+                            {% endfor %}
+                        </select>
+
+                        <select name="prev_course_{{ saved_prev_term_grad_length }}">
+                            <option value="">None</option>
+                        </select>
+                        <select name="prev_gradeable_{{ saved_prev_term_grad_length }}">
+                            <option value="">None</option>
+                        </select>#}
```
MasterOdin, line 198: Why is this commented out? Is it just temporary and will be re-added in a near-term future PR?

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
```diff
@@ -22,167 +23,217 @@
...
+                    </div>
+
+                    <div id="text_ignore_submission">
+                        {% if new_or_edit == "edit" %}
+                            {# {% for saved_ignore_submission in saved_config['ignore_submissions'] %}
+                                <div name="wrapper_ignore_submission_{{ loop.index0 }}">
+                                    <input type="text" name="ignore_submission_{{ loop.index0 }}" class="low-margin-top deletable-text" value="{{ saved_ignore_submission }}"/>
+                                    <i class="fa fa-trash delete-btn"></i>
+                                </div>
+                            {% endfor %} #}
```
MasterOdin, line 226: Same here with commented out code.

site/app/views/admin/PlagiarismView.php
```diff
@@ -109,83 +105,71 @@ public function showPlagiarismResult($semester, $course, $gradeable_id, $gradeab
...
+        $provided_code_filenames = [];
+        $version = "all_versions";
         $regex = "";
-        $language = ["python" => "selected", "java" => "", "plaintext" => "", "cpp" => "", "mips" => ""];
-        $provided_code = "";
-        $no_provided_code = "checked";
-        $provided_code_filename = "";
+        $regex_dirs = ["submissions"];
+        $language = array_fill_keys(PlagiarismUtils::getSupportedLanguages(), "");
+        $language["plaintext"] = "selected";
```
MasterOdin, line 121: Doing this here, wouldn't this mean that `$language["plaintext"]` is always `selected` then as it doesn't look like this gets reset anywhere?

site/app/views/admin/PlagiarismView.php
```diff
@@ -109,83 +105,71 @@ public function showPlagiarismResult($semester, $course, $gradeable_id, $gradeab
...
         $threshold = "5";
-        $sequence_length = "10";
+        $sequence_length = "4";
```
MasterOdin, line 123: Similar comment as above, why are we treating numeric values as strings?

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
```diff
@@ -22,167 +23,217 @@
...
                 </div>
             </div>
+            {##################################################################}
             <div class="plag-data-group">
-                <div class="plag-data-label">Threshold/Maximum number of students (more than this number of students with matching code will be considered common code):</div>
+                <div class="plag-data-label">Common Code Threshold:</div>
+                <label for="threshold">More than this number of students with matching code will be considered common code</label>
                 <div class="plag-data">
-                    <input type="text" name="threshold" value="{{ threshold }}" placeholder="(Required)"/>
+                    <input type="text" id="threshold" name="threshold" value="{{ threshold }}" placeholder="(Required)"/>
```
MasterOdin, line 112: This should be a `type="number"` input. Can also set a `min` value of `2` here to help guide users on what the minimum value is. Should also put it in the description as well.

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
```diff
@@ -22,167 +23,217 @@
...
                 </div>
             </div>
+            {##################################################################}
             <div class="plag-data-group">
                 <div class="plag-data-label">Sequence Length:</div>
                 <div class="plag-data">
                     <input type="text" name="sequence_length" value="{{ sequence_length }}" placeholder="(Required)"/>
```
MasterOdin, line 119: Same here, this is should be a `type="number"` input. Can also set a `min` here that makes sense.

MasterOdin (changes requested):
Not super familiar with this code, so comments are mostly just from a layman reading of the code.

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
williamjallen: Yes.  There is currently no complete, or nearly-complete backend for this, either in the Lichen codebase or in the PHP so we just commented this out for the moment instead of trying to get something together which only partially worked.  We'll hopefully be making PRs to implement these features soon.

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -214,105 +261,108 @@ public function showPlagiarismResult($gradeable_id) {
...
+
+        // Version
+        $version_option = $_POST['version_option'];
+        assert($version_option == "active_version" || $version_option == "all_versions");
+
+
+        // Regex
+        // TODO: Can we find a way to validate the regex here to tell the user their regex was invalid before feeding it to Lichen?
+        assert(isset($_POST["regex_dir"]) && isset($_POST["regex_to_select_files"]));
+        $regex_directories = $_POST["regex_dir"];
```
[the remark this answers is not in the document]
sbelsk: I'm pretty sure that the part in the concatenation code that uses the provided regex (around [this line](https://github.com/Submitty/Lichen/blob/0777e2594b0aa418b0c13f6b2cc48ecb8927b1ce/bin/concatenate_all.py#L85)) is only collecting the names of files that exist in each folder (`submissions`, `results`, `checkout`, including files in nested directories) and then filters that list of files based on the regex. If that it the case, then I think breaking out of the submission folder should not be possible

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -214,105 +261,108 @@ public function showPlagiarismResult($gradeable_id) {
...
             if ($_POST['prev_sem_' . $i] != "" && $_POST['prev_course_' . $i] != "" && $_POST['prev_gradeable_' . $i] != "") {
-                array_push($prev_term_gradeables, "/var/local/submitty/course/" . $_POST['prev_sem_' . $i] . "/" . $_POST['prev_course_' . $i] . "/submissions/" . $_POST['prev_gradeable_' . $i]);
+                array_push($prev_term_gradeables, FileUtils::joinPaths($course_path, $_POST['prev_sem_' . $i], $_POST['prev_course_' . $i], "submissions", $_POST['prev_gradeable_' . $i]));
             }
         }
 
-        $ignore_submissions = [];
+        // Submissions to ignore
         $ignore_submission_option = $_POST['ignore_submission_option'];
+        assert($ignore_submission_option == "ignore" || $ignore_submission_option == "no_ignore");
```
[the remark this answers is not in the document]
williamjallen: I wasn't sure how to properly handle this.  There can only ever be two values (`ignore` and `no_ignore`) and the frontend should know that.  If the value isn't one of those things, it's the programmer's fault, not the user's fault and this should theoretically never happen so I thought that throwing a PHP error would be appropriate.  Suggestions on what the correct thing to do here is?  If it isn't one of these two values, my thought was that we want to kill the script here before it goes on with the invalid data and starts corrupting things.

site/app/controllers/admin/PlagiarismController.php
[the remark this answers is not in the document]
MasterOdin: My concern was more about Submitty/Lichen#34 which changed how these fields were being used, and looks like @bmcutler raised similar concerns.

This line [https://github.com/Submitty/Lichen/pull/34/files#diff-30ca4b9b6281d4ace1b3c599e62652265d9ecf4c4180139cbb9556cae0e2e205R61] in the changeset looks like it would work alright to jump to a different course if I were to do `../../different_course/submissions` as a regex_dir?

site/app/controllers/admin/PlagiarismController.php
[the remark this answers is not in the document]
MasterOdin: Better would probably be to add an error message and just kick the user back to the form. We want to avoid having uncaught exceptions in the system as it's a poor UX for the end-user to commit an action and then see "An exception has occurred" screen, versus an error banner that states `Invalid ignore submission options, expected "ignore" or "no_ignore", got "<foo>".` Or something. A future improvement would be also saving state within the form when they're kicked back to it, but that can be a future thing.

site/app/controllers/admin/PlagiarismController.php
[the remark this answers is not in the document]
williamjallen: I just updated the Lichen PR to check for those three directories.  I believe that it just collects the files and then selects any of the files which match the pattern.  There is no way to match directories unless I am mistaken so I think any concerns have been addressed.

site/app/controllers/admin/PlagiarismController.php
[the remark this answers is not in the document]
williamjallen: Ok, will update.

PR #6614, opened 2021-06-16
[UI/UX:InstructorUI] Add borders to Plagiarism Highlight Key

### LINKED ISSUE #6610: Plagiarism highlighting key background needs work
The legend for the highlighted matches on the plagiarism results page has no borders for the colors, so the white color for "unique matches" cannot be seen in light mode

**Screenshots**

eddiekrystowski:
Fixes #6610 
image

There are no borders behind the icons in the plagiarism highlighting key, making white invisible on light mode.
### What is the new behavior?
There are now text shadows on light mode to help differentiate the icons from the background
image

### DIFF
--- a/site/app/templates/admin/PlagiarismHighlightingKey.twig
+++ b/site/app/templates/admin/PlagiarismHighlightingKey.twig
@@ -3,12 +3,12 @@
 {% block title %}Plagiarism Highlighting Key{% endblock %}
 {% block body %}
 <div>
-    <i style="color:white;" class="fas fa-square" aria-label="White Box"></i> Unique <br />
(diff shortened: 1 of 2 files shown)


williamjallen (changes requested):
This was my initial approach when I tackled this problem.  The issue is the space between the border and the top of the colored portion of the icon.  I tried putting another, slightly bigger, black, icon behind each icon but I had trouble making it look nice.  I also tried messing with the line height.  I'm not sure what the best way to tackle this issue is given that the icon has transparent regions above and below the "text" portion.

williamjallen (approved):
Great idea!  I like it a lot.

PR #6615, opened 2021-06-16
[Bugfix:Forum] Prevent spamming duplicate posts

### LINKED ISSUE #6613: Duplicate forum post submissions
While posting on a course forum, if you click any buttons that will submit a reply or a post multiple times, then multiple AJAX requests are sent to the server creating multiple of the same post/reply.

It was suggested by @williamjallen during meeting that we can disable the buttons once a submission button is clicked, and then reenabling it once the AJAX request is finished.

**Additional context**
Steps to recreate the bug

1. Go to the sample course on submitty and go to the discussion forum
2. Click create thread button
3. Enter arbitrary post title and post content and pick any categories
4. Click publish thread multiple times before the page is refreshed
5. See that duplicates of the same post is created on the discussion forum

The same issue is also seen similarly when replying to a thread.

eddiekrystowski:
Fixes #6613 
When submitting a post or creating a thread on the discussion forum, you can create multiple duplicate posts/threads by clicking the submit button multiple times before the page reloads.

### What is the new behavior?
The Submit Reply and Publish Thread buttons will now be disabled after being clicked on once, so that the form can't be submitted multiple times.

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -164,6 +164,8 @@ function publishFormWithAttachments(form, test_category, error_message, is_threa
     }
     var submit_url = form.attr('action');
 
+    form.find("[type=submit]").prop('disabled', true);
+
     $.ajax({
         url: submit_url,
         data: formData,


HenrikLam (approved):
Works as described!

PR #6616, opened 2021-06-16
[Feature:Plagiarism] Add ability to view run log to UI

williamjallen:
It is not currently possible to view the run log from the plagiarism UI.  This means that instructors are unable to check how far the algorithm has made it when running on large groups of assignments and are unable to report any errors that may occur.

### What is the new behavior?
This PR adds a button which lets instructors view the run log.

### Screenshots
Before:
before

After:
after

### DIFF
--- a/site/app/templates/plagiarism/Plagiarism.twig
+++ b/site/app/templates/plagiarism/Plagiarism.twig
@@ -111,8 +111,8 @@
             const url = buildCourseUrl(['plagiarism', 'gradeable', gradeable_id, 'log']) + `?gradeable_id=${gradeable_id}`;
             $.get(url, function (data) {
                 $(`<tr class="run-log-results">
-                <td colspan="7" style="text-align: left; width: inherit;">
-                    <pre style="width: 100%; white-space:pre-wrap; word-wrap: break-word; overflow: scroll;">` + data + `</pre>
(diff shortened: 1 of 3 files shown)


site/app/templates/plagiarism/Plagiarism.twig
```diff
@@ -3,7 +3,7 @@
     <div class="nav-buttons">
         <a class="btn btn-primary"
            href="{{ new_plagiarism_config_link }}"
-        >+ Configure New Gradeable for Plagiarism Detection</a>
+        ><i class="fas fa-plus"></i> Configure New Gradeable for Plagiarism Detection</a>
```
sbelsk, line 6: Is it just me, or should there be a tiny bit of space/padding between the + icon and the text on this button? 
image

sbelsk (changes requested):
Overall works well and displays the log as intended. Two small things I noticed when testing and playing with the UI:
- The box that appears with the contents of the log appears to always have horizontal and vertical scroll bars (I'm using Windows/Chrome)
- There is some whitespace wasted (especially between the "nightly rerun" and "run log" columns) with the way the widths of the columns are set, and I thought maybe we should try to maintain the more compact look this page had on master 
image

sbelsk (approved):
Looks better! Nice job.
image

PR #6617, opened 2021-06-16
[Bugfix:RainbowGrades] Rainbow Grades Allow Max Score Override 

### LINKED ISSUE #6351: Instructor UI for Rainbow Grades does not save/reload specified total for numeric gradeables
**Describe the bug**
If an instructor creates a numeric gradeable with 5 available points, but then edits this maximum number of points in the Rainbow Grades customization UI (to a lower number, effectively aiming to make this a partial extra credit assignment), the UI doesn't save, or doesn't reload this changed value.

**Expected behavior**
Either we should correctly save & reload the instructors override total, or not allow the instructor to edit this value from the UI.  

I vote for the former, but maybe the latter is better... The instructor can specify that some components on the numeric gradeable are extra credit -- However, if there is only a single component, or a components are partially extra credit this may not be simple to organize.

eddiekrystowski:
Fixes #6351 
Currently, when updating the max score field for a gradeable using Web Based Rainbow Grades Generation, the value is not saved in the UI when the page is reloaded. The value is just used correctly when rainbow grades is built, then is reset on the UI to the old value.

### What is the new behavior?
When the user enters any number that does not match the database's max score value, a warning icon shows up to let the user know that they have manually overridden it. A tooltip shows up giving more information when hovering over the warning icon. Any value that the user enters will persist on page reload.
image


eddiekrystowski:
@bmcutler The code for the tooltip message is in `site/app/templates/admin/RainbowCustomization.twig` on lines 140-144

bmcutler:
I quickly got alot of on screen errors when I started testing this PR:

Screen Shot 2021-06-18 at 10 39 10 AM

(sorry, not sure what specifically I did to cause this)

bmcutler:
Also, I changed the value, but the displayed actual max score and the override max score were equal in the tooltip.

bmcutler:
@eddiekrystowski 
I again got the same undefined offset error message (and a whole bunch of other errors) after I saved the rainbow grades page, navigated away and then came back.

bmcutler (approved):

PR #6619, opened 2021-06-16
[UI/UX:TAGrading] Fix spacing on Grading Panels Team List

eddiekrystowski:
The Team list takes up extra space at the top.
image

### What is the new behavior?
The space has been consolidated. Both sets of buttons in the toolbar now are always centered and move together, and the student name/team list will disappear when they overlap.
image

Teams with lots of members now show around 3-4 at a time and can be scrolled through.
Screen Recording 2021-06-23 at 3 57 23 PM

Also, a quick refactor has been done on `Details.twig` to move some inline table styling to `details.css`.

### Other information?
While this solution doesn't always center the toolbar on all screen sizes, I think it is the best solution since it optimizes the toolbar space up top which leaves the most amount of room for the panels which are the most important parts.


sbelsk (approved):

loadedskiffer (changes requested):
I think we should make it a drop down in order to conserve space and still have it centered at the same time.

loadedskiffer (approved):
Looks good to me and functions properly, Centering is preserved and minimal space is lost. 

> Just show everyone in the meeting tomorrow in case they don't like how the dropdown looks.

bmcutler (changes requested):
Sorry to say, but we've got to redo the layout of this feature on this page.  
The display of the student name or team is still taking up some of the precious vertical space.  The name needs to be on the left, and the buttons should be arranged in the remaining space on the right.  

The dropdown does not sufficiently fix the problem, there is still wasted vertical space  above the buttons due to the awkward layout/placement of the name(s).

Also, I don't really like the appearance of the dropdown.  
It's implying to me the viewer that I could select to focus grading on a different team member.  So this isn't quite the visual we were going for.

sbelsk (commented):
I don't know if this is the final way the names are going to be displayed, but just I took a lot at this and the page seems to be reacting well when adjusting the window size (it displays the names only if there is horizontal space for them). 
The only thing I noticed is that if you have your window relatively narrow before clicking "grade" on a specific team, when the grading interface loads, the text appears there by default even if there shouldn't be room for it:
image
If you resize the window it immediately fixes itself.

eddiekrystowski:
Good catch. I'll update that.

site/app/templates/grading/electronic/NavigationBar.twig
```diff
@@ -89,4 +98,11 @@
...
 </div>
-
+<script>
+    //if student/team name(s) are being displayed
+    {# if ({{ not anon_mode and not isBlind and (peer_blind_grading is same as('unblind') or peer_blind_grading is same as('single')) }}) {
+        $('.navigation-box').css({'position':'relative'});
+        $('.navigation-box').css({'float':'left'});
+        $('.navigation-box').css({'margin-left':'20%'});
+    } #}
+</script>
```
loadedskiffer: Pretty sure this commented out JS is here from when I was working on this and should be removed.

loadedskiffer (changes requested):
The Team names look great and nothing is broken but I'm pretty sure you didn't mean to include some of these changes in this PR.

loadedskiffer (approved):

bmcutler:
@eddiekrystowski or @williamjallen  
Please resolve conflicts

bmcutler (approved):
tested quickly, yes, I like this now :). Thanks!

bmcutler:
@eddiekrystowski or @williamjallen
Please resolve conflicts

bmcutler:
@eddiekrystowski
Please look into failing test case...

eddiekrystowski:
@bmcutler Just flaky e2e tests, all good now.

Issue #6618, opened 2021-06-16
Bootstrap Split Buttons not working


loadedskiffer:
**Describe the bug**
Currently if you follow bootstrap's documentation for making split buttons, they don't render properly and have no functionality.
This is due to edits being made to bootstrap dropdown-menu class. Upon looking further into this, there are other places where bootstrap classes have been edited strangely. This could have negative effects on adding new feautures going forward.
<img width="128" alt="issue3_1" src="https://user-images.githubusercontent.com/66340271/122275290-87559c00-ceb1-11eb-9130-154bef466945.PNG">
<img width="226" alt="issue3_2" src="https://user-images.githubusercontent.com/66340271/122275298-89b7f600-ceb1-11eb-887f-14c9cb74a661.PNG">

**Expected behavior**
If you exactly follow bootstrap's documentation for split buttons or any other component, they should render properly.
The edits to bootstrap's classes within submitty should most likely be rolled back in some way.
<img width="84" alt="issue3_3" src="https://user-images.githubusercontent.com/66340271/122275443-b23ff000-ceb1-11eb-9a64-745bfbd52433.PNG">

**To Reproduce**
1. Find the documentation for bootstrap 2.0 and find "split button dropdowns" under the components
2. Follow the instructions to implement them
3. See error
4. See line 818 of server.css and see edits 

**Configuration**
 - OS:  Windows
 - Browser chrome, edge

**Screenshots**

**Additional context**

eddiekrystowski:
Yes, I ran into this earlier and was going to mention it at the meeting but I forgot. 
It seems as if someone just copied over the bootstrap styling into server.css and then changed it for their specific needs. Fortunately, it looks like only one place uses it, `site/app/templates/forum/ForumBar.twig:85`, so it should be simple enough to change that class to something else, and then include the proper bootstrap classes in `bootstrap.css`.

PR #6620, opened 2021-06-16
[Feature:InstructorUI] Add regrade buttons

loadedskiffer:

Currently in order to regrade assignments, the instructor must do it from the command line with

sudo /usr/local/submitty/bin/regrade.py /var/local/submitty/courses/{{ semester}}/{{course}}/submissions/{{gradeable_id}}

The documentation for these commands are under instructor->assignment preparation->automated grading->batch regrade 
When any of the regrade commands are run, a file/files are created in /var/local/submitty/to_be_graded_queue

in order to see them there you must first stop your shipper and worker or they'll be graded right away.

Use:
-sudo systemctl stop submitty_autograding_shipper 
-sudo systemctl stop submitty_autograding_worker

An example of the contents of the file created may look like this
pr10_1

Once the file is created, the worker and shipper take care of the rest and the actual grading.

### What is the new behavior?
There are now multiple buttons available to the instructor (or TA's with instructor permissions) on the grading interface that allow them to regrade assignments.  
There are 4 options for regrading:
regrade active version - regrade the active version for one selected student who submitted a certain gradeable
regrade all versions - regrade every version for one selected student who submitted a certain gradeable
regrade all students active versions - regrade the active version for every student who submitted a certain gradeable
regrade all students all versions - regrade every version for every student who submitted a certain gradeable
pr10_2
pr10_3

These buttons have the same functionality as the command line option. They add the properly formatted files to the to_be_graded_queue directory so that the worker and shipper can process them properly.

### Other information?
Regrading is currently not available for assignments that use submissions from github because of known minor bugs in the way that those submissions are set up so the buttons are hidden for those assignments.

I realize the split buttons aren't the prettiest but the bootstrap ones won't render properly as of now #6618 and we still decided that we split buttons are the best option for this case.

To test:
1. First use the commands I listed above to manually regrade assignments and look at the files created in the queue
2. Now use the buttons and compare the files for all sorts of different gradeables
3. Change the autograding configuration for an assignment and see the results after using the regrade buttons


site/app/templates/grading/electronic/Details.twig
```diff
@@ -1,4 +1,20 @@
...
+        {% if row.graded_gradeable.getAutoGradedGradeable().getActiveVersion() > 0 %}
+            {% if is_team %}
+                {% set submission = submission | merge([{gradeable_id: row.graded_gradeable.gradeable.getId, submitter: row.graded_gradeable.getSubmitter().getTeam().getMemberUsers()[0].getId(), highest_version: row.graded_gradeable.getAutoGradedGradeable().getHighestVersion(), active_version: row.graded_gradeable.getAutoGradedGradeable().getActiveVersion() }]) %}
+            {% else %}
+                {% set submission = submission | merge([{gradeable_id: row.graded_gradeable.gradeable.getId, submitter: row.graded_gradeable.getSubmitter().getUser().getId(), highest_version: row.graded_gradeable.getAutoGradedGradeable().getHighestVersion(), active_version: row.graded_gradeable.getAutoGradedGradeable().getActiveVersion() }]) %}
+            {% endif %}
+        {% endif %}
+    {% endfor %}
+{% endfor %}
+<input type="hidden" id="submission" value="{{ submission|json_encode }}">
```
cjreed121, line 17: Why render all of the submissions to the user? This would also happen every time that page is loaded which isn't that efficient. Couldn't this be generated in the SubmissionController?

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -822,7 +825,11 @@ public function detailsPage(Gradeable $gradeable, $graded_gradeables, $teamless_
             "order_toggle_url" => $details_base_url . '?' .
                 http_build_query(['view' => $view_all ? 'all' : null, 'sort' => $sort === 'random' ? null : 'random', 'anon_mode' => $anon_mode]),
             "sort" => $sort,
-            "direction" => $direction
+            "direction" => $direction,
+            "csrf_token" => $this->core->getCsrfToken(),
```
cjreed121, line 829: The csrf token is already brought into the output. You can access this in Javascript by calling csrfToken since it is brought in from server.js.

site/public/js/drag-and-drop.js
```diff
@@ -848,6 +848,85 @@ function gatherInputAnswersByType(type){
...
+        url: submit_url,
+        data: formData,
+        processData: false,
+        xhr: function() {
+            var myXhr = $.ajaxSettings.xhr();
+            if(myXhr.upload){
+                myXhr.upload.addEventListener('progress',progress, false);
+            }
+            return myXhr;
+        },
```
cjreed121, line 891: This isn't a very large query. The progress bar may not be seen at all. Also my suggestion is to remove submissions from the user side making this even smaller. This probably won't be needed or is there another reason to keep it?

cjreed121 (changes requested):
The button seems to work. How does an instructor know when the assignment they are looking at has been regraded by the autograder? Also you should add some tests to your ajaxRegrade function.

williamjallen (changes requested):
Overall it looks good and works as intended but I have a few comments on the UI and I left a few comments on the code.

1. When in light mode, should this border turn black to match the buttons above?
Untitled
2. When you hover over the button part, should the dropdown appear?  Usually, the way split buttons work is that the dropdown part is used to make a selection to determine the contents of the button and clicking on the button actually triggers the action.  The split button for pull requests on GitHub is a good example of the usual behavior.  

https://user-images.githubusercontent.com/16820599/122401496-a81e0000-cf4a-11eb-9aa2-55e03513abc3.mov

3. When submissions are actively being regraded, should the button be disabled?

site/public/css/admin-gradeable.css
```diff
@@ -292,6 +292,42 @@ select[name="component-itempool"] {
...
+}
+
+.regrade-option {
+    border: 1px solid var(--standard-pale-gray);
+    width: 210px !important;
+}
+
+.regrade-option-details {
+    border: 1px solid var(--standard-pale-gray);
+    width: 320px !important;
```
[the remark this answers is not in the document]
loadedskiffer: I did it because I wanted the buttons to be the same size regardless of the text inside but I'm going to hold off until the bootstrap is fixed.

site/public/js/drag-and-drop.js
loadedskiffer: You're right I removed it

site/app/templates/grading/electronic/Details.twig
```diff
@@ -697,5 +708,61 @@
...
+        $('.regrade-btn').click(function(e) {
+            let count = 0;
+            {% for section in sections %}
+                {% for row in section.rows %}
+                    {% if row.graded_gradeable.getAutoGradedGradeable().getActiveVersion() > 0 %}
+                       count+=1;
+                    {% endif %}
+                {% endfor %}
+            {% endfor %}
+            if(count > 50) {
```
cjreed121, line 724: Take a look at what this creates.
image
This is a working solution but is going to add a lot of these calls. These calls should also not be needed to load the page (there could be hundreds). This feature is only designed to be used in rare cases so the performance of loading the page shouldn't drop for support of this. I think adding an Ajax call and having a backend route do the calculations would be a much better option here.

cjreed121 (changes requested):

site/app/controllers/student/SubmissionController.php
```diff
@@ -840,6 +841,160 @@ public function ajaxDeleteSplitItem($gradeable_id) {
...
+                $count += 1;
+            }
+        }
+        if ($regrade === 'true' || $regrade_all === 'true') {
+            $msg = $count . " submission(s) from " . $who_id . " added to queue for regrading";
+        }
+        else {
+            $msg = $count . " submissions added to queue for regrading";
+        }
+        return $this->core->getOutput()->renderResultMessage($msg, true);
```
cjreed121, line 996: This should probably be return JsonResponse::getSuccessResponse

site/public/js/drag-and-drop.js
```diff
@@ -848,6 +848,75 @@ function gatherInputAnswersByType(type){
...
+        headers : {
+            Accept: "application/json"
+        },
+        contentType: false,
+        type: 'POST',
+        success: function(data) {
+            try {
+                data = JSON.parse(data);
+                if (data['status'] === 'success') {
+                    window.location.href = return_url;
```
cjreed121, line 893: Wouldn't a window.location.reload() make more sense? It seems you are simply refreshing the page

cjreed121 (changes requested):

williamjallen (commented):
Aside from the issues Chris mentioned, I think the code looks good.

site/app/controllers/student/SubmissionController.php
loadedskiffer: This makes it so it displays the nice green success message at the top of the page which Barb wanted.

site/public/js/drag-and-drop.js
```diff
@@ -848,6 +848,73 @@ function gatherInputAnswersByType(type){
...
+                if (data['status'] === 'success') {
+                    window.location.reload();
+                }
+                else {
+                    if (data['message'] == "You do not have access to that page.") {
+                        window.location.reload();
+                    }
+                    else if(typeof data['code'] !== undefined && data['code'] === 302){
+
+                    }
```
cjreed121, line 899: Empty conditional?

cjreed121 (changes requested):

site/public/js/drag-and-drop.js
loadedskiffer: wow not sure how I overlooked that, I meant to delete it.

site/app/controllers/student/SubmissionController.php
```diff
@@ -840,6 +841,160 @@ public function ajaxDeleteSplitItem($gradeable_id) {
 
         return $this->uploadResult("Successfully deleted this PDF.");
     }
+    /**
+     * function for counting the number of submissions to be regraded
+     * @AccessControl(role="INSTRUCTOR")
+     * @Route("/courses/{_semester}/{_course}/gradeable/{gradeable_id}/regrade/count", methods={"POST"})
+     * @return JsonResponse
+     */
+    public function ajaxCountRegrade() {
```
cjreed121, line 850:
```suggestion
    public function ajaxCountRegrade(): JsonResponse {
```

cjreed121 (changes requested):
Really quick fix then I think it's good to go.

cjreed121 (approved):

williamjallen (approved):

site/app/templates/grading/electronic/Details.twig
```diff
@@ -697,5 +708,82 @@
...
+        $('.regrade-btn').click(function(e) {
+            ajaxCountRegrade(0,e,confirmMessage);
+         });
+    });
+
+    $(function() {
+        $('.regrade-btn-all').click(function(e) {
+            ajaxCountRegrade(1,e,confirmMessage);
+        });
+    });
```
MasterOdin, line 786: Why bind this click handler with JS versus just adding the click handler directly to the `<button>` and `<a>` tags directly?

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1138,11 +1143,28 @@ public function renderGradingPanelHeader(bool $isPeerPanel, bool $isStudentInfoP
...
+            $id = $graded_gradeable->getSubmitter()->getId();
+        }
+
         return $this->core->getOutput()->renderTwigTemplate("grading/electronic/AutogradingPanel.twig", [
             "version_instance" => $version_instance,
             "show_hidden_cases" => $show_hidden_cases,
+            "highest_version" =>  $graded_gradeable->getAutoGradedGradeable()->getHighestVersion(),
+            "max_submissions" => $gradeable->getAutogradingConfig()->getMaxSubmissions(),
+            "is_vcs" => $gradeable->isVcs(),
+            "gradeable_id" => $gradeable->getId(),
```
MasterOdin, line 1165: Do you need to pass in `$gradeable` here, or can we just do `$graded_gradeable->getGradeable()`?

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1138,11 +1143,28 @@ public function renderGradingPanelHeader(bool $isPeerPanel, bool $isStudentInfoP
      * @param bool $show_hidden_cases
      * @return string
      */
-    public function renderAutogradingPanel($version_instance, bool $show_hidden_cases) {
+    public function renderAutogradingPanel($version_instance, bool $show_hidden_cases, $graded_gradeable, $gradeable, $display_version) {
```
MasterOdin, line 1146: Add types for these new parameters.

site/public/js/drag-and-drop.js
```diff
@@ -848,6 +848,70 @@ function gatherInputAnswersByType(type){
...
+        contentType: false,
+        type: 'POST',
+        success: function(data) {
+            try {
+                data = JSON.parse(data);
+                if (data['status'] === 'success') {
+                    window.location.reload();
+                }
+                else {
+                    if (data['message'] == "You do not have access to that page.") {
```
MasterOdin, line 894: Is this error message possible to hit under normal usage?

site/app/templates/grading/electronic/Details.twig
```diff
@@ -697,5 +708,82 @@
...
+            },
+            contentType: false,
+            success: function(data) {
+                data = JSON.parse(data);
+                console.log(data);
+                if (data['status'] === 'success') {
+                    confirmMessage(e, data['data'], choice);
+                }
+                else {
+                    if (data['message'] === "You do not have access to that page.") {
```
MasterOdin, line 741: Similarly here, is this condition possible to hit?

site/app/templates/grading/electronic/Details.twig
```diff
@@ -697,5 +708,82 @@
...
+            data: formData,
+            processData: false,
+            type: 'POST',
+            headers : {
+                Accept: "application/json"
+            },
+            contentType: false,
+            success: function(data) {
+                data = JSON.parse(data);
+                console.log(data);
```
MasterOdin, line 736: remove

site/app/templates/grading/electronic/Details.twig
```diff
@@ -697,5 +708,82 @@
...
+                    else {
+                        alert("ERROR! Please contact administrator with following error:\n\n" + data['message']);
+                    }
+                }
+            }
+        });
+    }
+    function confirmMessage(e,count, choice) {
+        if(count > 50) {
+            e.preventDefault();
```
MasterOdin, line 753: Why prevent the default only for `count > 50`? What is the default that we're preventing here?

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -6,6 +6,7 @@
 use app\libraries\DateUtils;
 use app\libraries\FileUtils;
 use app\libraries\Utils;
+use app\models\gradeable\Component;
```
MasterOdin, line 9: This looks unused?

MasterOdin (changes requested):

site/app/templates/grading/electronic/Details.twig
loadedskiffer: I was having trouble passing the event as a parameter like that but I fixed it.

site/app/views/grading/ElectronicGraderView.php
loadedskiffer: Nope you're right

site/app/templates/grading/electronic/Details.twig
loadedskiffer: Barb wanted a confirm message to pop up if the instructor attempted to regrade over 50 submissions, to make sure they don't accidentally send a bunch of files to the to be graded queue

bmcutler:
@loadedskiffer 
please fix merge conflict, thanks :)

PR #6623, opened 2021-06-17
[Feature:Forum] After the fact email

### LINKED ISSUE #5785: After-the-fact forum announcement
An instructor makes a new thread on the forum, intending to make it an announcement and send an email, but forgets to do so when they submit the post.  Currently you can "pin" an existing thread...

Let's add a button to an existing *recent* thread (posted within the last 60 minutes) that is not currently pinned.  
button text: "Pin & email this thread"

edit to add:
Let's consider restrict this feature to:
* first post in a thread (definitely)
* only the original post author can chose to send the retroactive email (probably)
* only threads with no follow on posts (maybe not necessary?)

tagging @holzbh

HenrikLam:

Closes #5785

### What is the new behavior?
For posts created less than an hour ago by instructors or full access graders, the creator of the post can now make an announcement after already posting the thread with the pin & email this thread button.

image

For non-author and people with access less than a full-access grader, there is no change. The button also does not show up for posts older than 60 minutes old.


site/app/controllers/EmailStatusController.php
```diff
@@ -33,4 +32,19 @@ public function getEmailStatusPage() {
...
+     */
+    public function checkAnnouncement() {
+        if (isset($_GET["thread_id"])) {
+            $exists = $this->core->getQueries()->existsAnnouncementsId($_GET["thread_id"]);
+            return JsonResponse::getSuccessResponse([
+                "exists" => $exists
+            ]);
+        }
+        return JsonResponse::getFailResponse("No thread_id provided");
+    }
```
cjreed121, line 49: An announcement would be a forum post. Due to that I feel that this function should be in the ForumController instead. Or is there a specific reason it is in here?

site/app/controllers/forum/ForumController.php
```diff
@@ -369,6 +369,38 @@ public function publishThread() {
         return $this->core->getOutput()->renderJsonSuccess($result);
     }
 
+    /**
+     * @Route("/courses/{_semester}/{_course}/forum/make_announcement", methods={"POST"})
+     * @AccessControl(permission="forum.modify_announcement")
```
cjreed121, line 374: If only Full Access Graders and higher can call the other function why not just make the same here?

site/app/controllers/forum/ForumController.php
```diff
@@ -369,6 +369,38 @@ public function publishThread() {
...
+        $full_course_name = $this->core->getFullCourseName();
+        $thread_title = trim($_POST["title"]);
+        $thread_post_content = str_replace("\r", "", $_POST["thread_post_content"]);
+        $metadata = json_encode(['url' => $this->core->buildCourseUrl(['forum', 'threads', $_POST['id']]), 'thread_id' => $_POST['id']]);
+
+        $subject = "New Announcement: " . Notification::textShortner($thread_title);
+        $content = "An Instructor or Teaching Assistant made an announcement in:\n" . $full_course_name . "\n\n" . $thread_title . "\n\n" . $thread_post_content;
+        $event = ['component' => 'forum', 'metadata' => $metadata, 'content' => $content, 'subject' => $subject];
+        $this->core->getNotificationFactory()->onNewAnnouncement($event);
+    }
```
cjreed121, line 402: This should return a json response or something. You have it returning void while not providing any information to the user of what went wrong. You should also have some sort of success message.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -636,6 +636,19 @@ public function createPost($user, $content, $thread_id, $anonymous, $type, $firs
...
+    public function findPost($thread_id) {
+        $this->course_db->query("SELECT * from posts where thread_id = ?", [$thread_id]);
+        return $this->course_db->row();
+    }
+
+    public function existsAnnouncementsId($thread_id) {
+        $this->course_db->query("SELECT title from threads where id = ?", [$thread_id]);
+        $title = $this->course_db->row()['title'];
+        $announcement_subject = "[Submitty " . $this->core->getConfig()->getCourse() . "]: New Announcement: (" . $thread_id . ") " . $title;
+        $this->submitty_db->query("SELECT 1 from emails where subject = ?", [$announcement_subject]);
```
cjreed121, line 648: What is the purpose of this query?

site/public/js/forum.js
```diff
@@ -2104,3 +2104,30 @@ $(() => {
...
+    type: 'GET',
+    url: buildCourseUrl(['email', 'check_announcement']),
+    data: {'thread_id': id, "csrfToken": csrfToken},
+    success: function(res) {
+      const response = JSON.parse(res);
+      if (response.status === "success" && response["data"]["exists"] == false){
+        $('.pin-and-email-message').show();
+      }
+    }
+  });
```
cjreed121, line 2132: These should have some error handling whether the request failed or the backend reported an issue with input data.

cjreed121 (changes requested):
The button works and pins the post to the top. The button appeared for me when I created a new announcement though. I think one of your queries checks for emails and is conditional based off of that. Every minute 100 emails are sent all at once. This could lead to the button showing and an instructor clicking it while it's already an announcement. Also if a thread has a question mark on it's preview, after making it an announcement it still keeps that question mark. That should probably go away.

site/app/libraries/database/DatabaseQueries.php
HenrikLam: This was the way we came up with to check if the announcement was made, since there is no easy and clean way to check if the announcement was made at the moment without changing the database. We decided since the feature isn't going to be used too much, we should do this instead which was easier than making a migration.

site/app/controllers/EmailStatusController.php
HenrikLam: I put this here because this function involves checking with the email table in the database, which doesn't seemed too related to forum to me at the time. But the overall use of this function is very much related to ForumController and to be honest I wasn't sure where I should put this function. What do you think?

site/app/controllers/forum/ForumController.php
HenrikLam: I used forum.modify_announcement just to follow the overall trend of the rest of the file, there was no specific reason for using that over full_access_grader. Looking at access.php, it doesn't seem like there was much of a difference between the two.

site/app/controllers/forum/ForumController.php
HenrikLam: Should it simply return a json or should it display a message at the top too? Right now, the only indication that it succeeded is if the user checks to see if the button still exists, or if they go into the email status page if they are an instructor.

site/public/js/forum.js
HenrikLam: > The button works and pins the post to the top. The button appeared for me when I created a new announcement though. I think one of your queries checks for emails and is conditional based off of that. Every minute 100 emails are sent all at once. This could lead to the button showing and an instructor clicking it while it's already an announcement. Also if a thread has a question mark on it's preview, after making it an announcement it still keeps that question mark. That should probably go away.

This actually brings up an issue that I somehow didn't see before (for some reason I thought the announcement title includes the thread id). 

- An issue right now and possible ways to solve them: currently the database makes an announcement with only the thread title in the announcement and no thread id, meaning it is possible that the making the announcement button is incorrectly hidden

1. Option 1: We can change the announcement title to also include the thread id

2. Option 2: Assume the instructors/TA won't make a thread with duplicate title...

site/app/controllers/EmailStatusController.php
cjreed121: Typically controller functions should be isolated to where they are used. I see why you were thinking of going with this since it checks the emails but it wouldn't ever be used in the email status page. If someone wanted to go looking for this route it would make sense to find it in the ForumController rather than the EmailStatusController since it is used by the Forum.

site/app/controllers/forum/ForumController.php
cjreed121: Since the site is making a direct request here a display message at the top would probably be better.

site/public/js/forum.js
cjreed121: Yeah I think option 1 is definitely the better way to go compared to option 2. Option 2 could happen completely by accident. We need to make sure that when the page loads that it doesn't still show that button cause it may look like the instructor forgot to set it as an announcement. Now there is twice the emails.

migration/migrator/data/course_tables.sql
```diff
@@ -1061,6 +1062,8 @@ CREATE TABLE public.users (
     time_zone character varying DEFAULT 'NOT_SET/NOT_SET'::character varying NOT NULL,
     display_image_state character varying DEFAULT 'system'::character varying NOT NULL,
     registration_subsection character varying(255) DEFAULT ''::character varying NOT NULL,
+    user_email_secondary character varying(255) DEFAULT ''::character varying NOT NULL,
+    user_email_secondary_notify boolean DEFAULT false,
```
cjreed121, line 1066: These are not from this PR. Make sure to edit the SQL before you commit in case your database have migrations from other branches.

migration/migrator/data/submitty_db.sql
```diff
@@ -247,40 +247,40 @@ $$;
 CREATE FUNCTION public.sync_user() RETURNS trigger
     LANGUAGE plpgsql
     AS $$
-DECLARE
-    course_row RECORD;
-    db_conn VARCHAR;
-    query_string TEXT;
-    preferred_name_change_details TEXT;
-BEGIN
-    -- Check for changes in users.user_preferred_firstname and users.user_preferred_lastname.
```
cjreed121, line 256: Same as above for 250-284

migration/migrator/migrations/course/20210621125502_thread_announce.py
```diff
@@ -0,0 +1,33 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("ALTER TABLE threads DROP announced")
```
cjreed121, line 33: Typically we don't want a down migration to delete data when we can still run Submitty with that there. If for some reason Submitty needs to be rolled back it would be handy to not need to delete all that data. The down migration for this could just be pass since your up migrations does a check to make sure it doesn't already exist.

site/public/js/forum.js
```diff
@@ -2104,3 +2104,55 @@ $(() => {
...
+  $(".pin-and-email-message").attr('disabled','disabled');
+  $.ajax({
+    type: 'POST',
+    url: buildCourseUrl(['forum', 'make_announcement']),
+    data: {"id": id, 'csrf_token': window.csrfToken},
+    success: function(data){
+      pinAnnouncement(id, 1, window.csrfToken);
+      window.location.reload();
+    },
+  });
```
cjreed121, line 2119: This needs error handling through Ajax but also checks to see if the JsonResponse was Success or Fail.

cjreed121 (changes requested):

site/public/js/forum.js
HenrikLam: For some reason when I add the check on if JsonResponse was sucess or fail, the success message doesn't show up at the top... But I included the check anyway, because the user does have some kind of indication with how the button disappears and there is no error message.

site/app/controllers/forum/ForumController.php
```diff
@@ -369,6 +370,60 @@ public function publishThread() {
         return $this->core->getOutput()->renderJsonSuccess($result);
     }
 
+    /**
+     * @Route("/courses/{_semester}/{_course}/forum/make_announcement", methods={"POST"})
+     * @AccessControl(permission="forum.modify_announcement")
+     */
+    public function makeAnnouncement() {
```
cjreed121, line 377:
```suggestion
    public function makeAnnouncement(): JsonResponse {
```

site/app/controllers/forum/ForumController.php
```diff
@@ -369,6 +370,60 @@ public function publishThread() {
...
+        $this->core->getQueries()->updateResolveState($_POST['id'], 0);
+        return JsonResponse::getSuccessResponse("Announcement successfully queued for sending");
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/forum/check_announcement", methods={"GET"})
+     * @AccessControl(role="FULL_ACCESS_GRADER")
+     * @return JsonResponse
+     */
+    public function checkAnnouncement() {
```
cjreed121, line 417:
```suggestion
    public function checkAnnouncement(): JsonResponse {
```

cjreed121 (changes requested):

site/public/js/forum.js
```diff
@@ -2104,3 +2104,63 @@ $(() => {
...
+//Only used by the posters and only on recent posts (60 minutes since posted)
+function sendAnnouncement(thread_post_content, id){
+  $(".pin-and-email-message").attr('disabled','disabled');
+  $.ajax({
+    type: 'POST',
+    url: buildCourseUrl(['forum', 'make_announcement']),
+    data: {"id": id, 'csrf_token': window.csrfToken},
+    success: function(data){
+      if (JSON.parse(data)["status"] === "success"){ 
+        pinAnnouncement(id, 1, window.csrfToken);
```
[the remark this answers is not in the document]
HenrikLam: Yeah I see a couple places where I can use . instead, some of which which I haven't written myself. I will change the parts that I wrote, but avoid the others just in case I end up breaking something unknowingly. I am a bit confused about the CSRF part though since I see other ajax calls that pass it in forum.js but others that do not.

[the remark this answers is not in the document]
cjreed121: I'm not sure why they use CSRF in an argument. The server.js file provides the csrfToken variable everywhere needed.

cjreed121 (approved):

site/app/controllers/EmailStatusController.php
```diff
@@ -5,20 +5,19 @@
...
- * @AccessControl(role="INSTRUCTOR")
- */
 class EmailStatusController extends AbstractController {
     public function __construct(Core $core) {
         parent::__construct($core);
     }
 
     /**
      * @Route("/courses/{_semester}/{_course}/email_status")
+     * @AccessControl(role="INSTRUCTOR")
```
MasterOdin, line 20: Why move this from being a class method to a function method?

site/app/controllers/forum/ForumController.php
```diff
@@ -369,6 +370,60 @@ public function publishThread() {
...
+     * @AccessControl(permission="forum.modify_announcement")
+     */
+    public function makeAnnouncement(): JsonResponse {
+        if (!isset($_POST['id'])) {
+            $this->core->addErrorMessage("thread_id not provided");
+            return JsonResponse::getFailResponse("thread_id not provided");
+        }
+        // Check that the post is the first post of the thread
+        $thread_info = $this->core->getQueries()->findPost($_POST['id']);
+        if ($thread_info["parent_id"] != -1) {
```
MasterOdin, line 384: I think a PHP warning will be raised here if `$_POST['id']` does not point at an actual thread.

site/app/templates/forum/CreatePost.twig
```diff
@@ -130,3 +134,9 @@
 
     </form>
 {% endif %}
+
+<script>
+    $(document).ready(function() {
+        checkIfAnnounced({{ activeThread.id }});
+    });
+</script>
```
MasterOdin, line 142: why are we fetching this information via AJAX after page load, as opposed to just passing it into the twig template from the view?

site/app/controllers/forum/ForumController.php
MasterOdin: Though this query / check should be rewritten so that it also checks within the query that `parent_id = -1` as currently the query within `findPost` will return all posts within a thread (which could be a large amount, and waste memory to fetch) and that there's no guarantee that the first result in the result set will be the top post post in a given thread.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -781,6 +795,9 @@ public function createThread($markdown, $user, $title, $content, $anon, $prof_pi
 
         //Max id will be the most recent post
         $id = $this->course_db->rows()[0]["max_id"];
+        if ($announcement) {
+            $this->setAnnounced($id);
+        }
```
MasterOdin, line 800: Why run a separate query to update the thread as opposed to just setting the `announced` value within the `insert` above?

MasterOdin (changes requested):

site/app/controllers/EmailStatusController.php
HenrikLam: I originally wrote the thread query function in this controller but had a different access control for that function and forgot to change this back to being a class method when I moved the function, I will change it back.

site/public/js/forum.js
```diff
@@ -2103,3 +2103,63 @@ $(() => {
 function scrollThreadListTo(element){
   $(element).get(0).scrollIntoView({behavior: "smooth", block: "center"});
 }
+
+//Only used by the posters and only on recent posts (60 minutes since posted)
+function sendAnnouncement(thread_post_content, id){
```
MasterOdin, line 2108: The `thread_post_content` argument is unused here. Can it be removed?

site/public/js/forum.js
```diff
@@ -2103,3 +2103,63 @@ $(() => {
...
+  $(".pin-and-email-message").attr('disabled','disabled');
+  $.ajax({
+    type: 'POST',
+    url: buildCourseUrl(['forum', 'make_announcement']),
+    data: {"id": id, 'csrf_token': window.csrfToken},
+    success: function(data){
+      if (JSON.parse(data).status === "success"){
```
MasterOdin, line 2115: Good practice to wrap any `JSON.parse` with a try/catch since if the server returns a bad response (e.g. robot exception page), then this will just raise an exception without alerting the user anything has gone wrong.

site/public/js/forum.js
```diff
@@ -2103,3 +2103,63 @@ $(() => {
...
+}
+
+//Used to decide to hide or show the button to send announcements based on if an announcement is made already
+function checkIfAnnounced(id){
+  call = $.ajax({
+    type: 'GET',
+    url: buildCourseUrl(['forum', 'check_announcement']),
+    data: {'thread_id': id},
+    success: function(res) {
+      const response = JSON.parse(res);
```
MasterOdin, line 2136: same here.

site/app/views/forum/ForumThreadView.php
```diff
@@ -1121,7 +1121,8 @@ public function createPost($thread_id, $post, $unviewed_posts, $first, $reply_le
             "parent_id" => $parent_id,
             "render_markdown" => $markdown,
             "has_history" => $has_history,
-            "thread_previously_merged" => $merged_thread
+            "thread_previously_merged" => $merged_thread,
+            "thread_announced" => $this->core->getQueries()->existsAnnouncementsId($thread_id)
```
MasterOdin, line 1125: This query should happen within the controller and then pass in the value to the view.

MasterOdin (changes requested):

bmcutler:
@HenrikLam 
Sorry, I think I broke something when I attempted to resolve the merge conflicts with master branch.
I'm now getting an "Error parsing data.  Please try again." message.  (It was working before I did the merge.)
Please debug and then ask Chris to re-review.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -764,14 +779,14 @@ public function getUnviewedPosts($thread_id, $user_id) {
...
+    public function createThread($markdown, $user, $title, $content, $anon, $prof_pinned, $status, $hasAttachment, $categories_ids, $lock_thread_date, $expiration, $announcement) {
         $this->course_db->beginTransaction();
 
+        $now = $announcement ? new \DateTime("now") : null;
+
         try {
             //insert data
-            $this->course_db->query("INSERT INTO threads (title, created_by, pinned, status, deleted, merged_thread_id, merged_post_id, is_visible, lock_thread_date, pinned_expiration) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)", [$title, $user, $prof_pinned, $status, 0, -1, -1, true, $lock_thread_date, $expiration]);
-
+            $this->course_db->query("INSERT INTO threads (title, created_by, pinned, status, deleted, merged_thread_id, merged_post_id, is_visible, lock_thread_date, announced) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)", [$title, $user, $prof_pinned, $status, 0, -1, -1, true, $lock_thread_date, $expiration, $now]);
```
cjreed121, line 789: This needs to be fixed from merge conflict
```suggestion
            $this->course_db->query("INSERT INTO threads (title, created_by, pinned, status, deleted, merged_thread_id, merged_post_id, is_visible, lock_thread_date, pinned_expiration, announced) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)", [$title, $user, $prof_pinned, $status, 0, -1, -1, true, $lock_thread_date, $expiration, $now]);
```

cjreed121 (changes requested):

loadedskiffer (approved):
Could not break anything

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -764,14 +779,14 @@ public function getUnviewedPosts($thread_id, $user_id) {
         return $rows;
     }
 
-    public function createThread($markdown, $user, $title, $content, $anon, $prof_pinned, $status, $hasAttachment, $categories_ids, $lock_thread_date, $expiration) {
-
+    public function createThread($markdown, $user, $title, $content, $anon, $prof_pinned, $status, $hasAttachment, $categories_ids, $lock_thread_date, $expiration, $announcement) {
         $this->course_db->beginTransaction();
 
+        $now = $announcement ? new \DateTime("now") : null;
```
cjreed121, line 785: It might be better to use $this->core->getDateTimeNow() since it uses the config's timezone.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -794,6 +810,11 @@ public function createThread($markdown, $user, $title, $content, $anon, $prof_pi
         return ["thread_id" => $id, "post_id" => $post_id];
     }
 
+    public function setAnnounced($thread_id) {
+        $now = new \DateTime("now");
```
cjreed121, line 814: Same as above.

cjreed121 (changes requested):

cjreed121 (approved):

PR #6631, opened 2021-06-18
[Feature:Autograding] Gradeable Timer Autograder Config

cjreed121:

Currently, in order to support the gradeable progress timer on exams, the minutes must manually be entered into the database.

### What is the new behavior?
When the autograder configuration is built the data in the database gets set automatically.


bin/set_allowed_mins.py
```diff
@@ -0,0 +1,95 @@
...
+        db.execute(text(query), gradeable=GRADEABLE, userid=user['user'],
+                   minutes=user['allowed_minutes'])
+
+
+def main():
+    db, metadata = setup_db()
+    with open(CONFIG_FILE_PATH) as config_file:
+        json_string = config_file.read()
+    # Removes #'s
+    json_string = re.sub("(^|\n)#[^\n]*(?=\n)", "", json_string)
```
bmcutler, line 73: Why are there # in the json string?
We have C++ comments, but they are removed in the build process

bin/set_allowed_mins.py
```diff
@@ -0,0 +1,95 @@
...
+    CONFIG_FILE = json.loads(json_string)
+    timelimit_case = None
+    for testcase in CONFIG_FILE['testcases']:
+        if testcase['title'] == "Check Time Limit":
+            timelimit_case = testcase
+            break
+    if timelimit_case is None:
```
bmcutler, line 80: should be 'and' of these conditions, not the 'or'

bin/set_allowed_mins.py
```diff
@@ -0,0 +1,95 @@
...
+        for testcase in CONFIG_FILE['testcases']:
+            if 'validation' in testcase:
+                if len(testcase['validation']) > 0:
+                    if 'allowed_minutes' in testcase['validation'][0]:
+                        timelimit_case = testcase
+                        break
+
+    if timelimit_case is not None:
+        allowed_minutes = timelimit_case['validation'][0]['allowed_minutes']
+        override = timelimit_case['validation'][0]['override']
```
bmcutler, line 90: the override might be empty (no students overridden) or not present in the file, which would crash here?

bmcutler (changes requested):
Mostly good, I didn't test yet.  
I've noted (I believe) a few small bugs in the logic though.

bin/set_allowed_mins.py
cjreed121: The python crashes when trying to decode the JSON unless I do  this. I assume this is removing any comments found in the JSON.

bin/set_allowed_mins.py
cjreed121: The strategy here is to first see if there is a testcase named "Check Time Limit" and set the timelimit_case to that. If we have not found that then we will go looking again to see if any testcase has the allowed minutes field in case an instructor changed the name. I am not following your suggestion to change here?

bin/set_allowed_mins.py
bmcutler: Did you copy this line of code from somewhere?  There aren't # symbols in this json string that should be removed.  The # symbols in this file are markdown, and the markdown strings shouldn't be relevant to this script.  And I don't think this will match those # symbols since they aren't at the start of the line (immediately after the newline).

bin/set_allowed_mins.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def send_data(db, allowed_minutes, override):
+    query = """UPDATE gradeable SET g_allowed_minutes = :minutes
+               WHERE g_id=:gradeable"""
+    db.execute(text(query), minutes=allowed_minutes, gradeable=GRADEABLE)
+    query = """DELETE FROM gradeable_allowed_minutes_override WHERE g_id=:gradeable"""
+    db.execute(text(query), gradeable=GRADEABLE)
+    if override is not None:
+        for user in override:
```
bmcutler, line 64: What if the username is not a valid user in the course?  It appears we blindly add them to the database.  So it doesn't crash or error out -- that's good...  but it would be better if the database had a key constraint (please lets add one) on the user_id column to be a valid user.  And then this chunk of code would need to handle the error of a non existent user in the config file.  It would be nice if this error got spit back out to the build script STDOUT/STDERR -- whatever we are capturing for the webpage log.

Screen Shot 2021-07-23 at 1 04 32 AM

bmcutler (changes requested):

bmcutler (approved):

Issue #6639, opened 2021-06-21
Team history array bug


loadedskiffer:
**Describe the bug**
Ever since upgrading to ubuntu 20.04 every time I go to the Student index page for a team gradeable I get the PHP warning below
<img width="627" alt="issue4_1" src="https://user-images.githubusercontent.com/66340271/122811628-b9df0a80-d29e-11eb-9181-532dacfc5b60.PNG">

**Expected behavior**
The warning should not be occurring because nothing is actually broken, all information about the team and the gradeable is correct, there's just an error for some reason.

**To Reproduce**
1. upgrade to ubuntu 20.04
2. login as a grader and attempt to grade a team gradeable
4. See error on details page

**Configuration**
 - OS: Windows
 - Browser chrome, edge

eddiekrystowski:
I am not able to reproduce this bug on ubuntu 20.04, but it sounds a lot like that strange forum bug I was getting. Can you try deleting your Submitty cookies in Chrome (or a different browser) and see if it persists?

And with the fix from #6640, your version of the page looks the exact same as it should (no info missing)?

PR #6640, opened 2021-06-21
[Bugfix:TAGrading] Team History Array bug fix 

loadedskiffer:
fixes issue #6639 , there is a PHP warning occurring on the student index of team gradeables ever since upgrading to ubuntu 20.04

### What is the new behavior?
The warning is gone and the page loads properly.

### Other information?
To test:
Navigate to the student index page for a team gradeable after upgrading your VM

### DIFF
--- a/site/app/models/Team.php
+++ b/site/app/models/Team.php
@@ -202,6 +202,9 @@ class Team extends AbstractModel {
             $settings_file = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "submissions", $gradeable->getId(), $this->getId(), "user_assignment_settings.json");
             $this->assignment_settings = FileUtils::readJsonFile($settings_file);
         }
-        return $this->assignment_settings;
+        if (isset($this->assignment_settings) && is_array($this->assignment_settings)) {
+            return $this->assignment_settings;
+        }
+        return ["team_history" => null];
     }
 }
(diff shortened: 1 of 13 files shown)


MasterOdin (changes requested):
We should also modify `getAssignmentSettings` so that it always returns an array, instead of `array|false`.

loadedskiffer:
@MasterOdin so should getAssignmentSettings return an empty array instead of false ?

MasterOdin:
well, could also return `["team_history" => null]` and then could get rid of the null coalescing operator where you've added it. Either way, yeah, `getAssignmentSettings` should always return an array per its docstring.

site/app/models/Team.php
```diff
@@ -202,6 +202,9 @@ public function getAssignmentSettings(Gradeable $gradeable) {
             $settings_file = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "submissions", $gradeable->getId(), $this->getId(), "user_assignment_settings.json");
             $this->assignment_settings = FileUtils::readJsonFile($settings_file);
         }
-        return $this->assignment_settings;
+        if (isset($this->assignment_settings) && is_array($this->assignment_settings)) {
+            return $this->assignment_settings;
+        }
+        return ["team_history" => null];
```
MasterOdin, line 208:
```suggestion
        return $this->assignment_settings ? $this->assignment_settings : ["team_history" => null];
```

This would protect against both `false` and an empty array (`[]`).

MasterOdin (changes requested):

site/app/models/Team.php
loadedskiffer: That resolves to the same thing as $this->assignment_settings ? : ["team_history" => null]; right?

MasterOdin: Yeah, you're right. I always forget about the shorthand ternary statement within PHP.

MasterOdin (approved):

PR #6642, opened 2021-06-22
[Bugfix:Developer] Fix selector/indentation for stylelint

MasterOdin:

The setup of stylelint in #6630 did not have our two principle rules set:
* Use hyphens for selectors
* Use four spaces for indentation

### What is the new behavior?

Updates the rules for stylelint for selectors must not use underscores and instead use hyphens, and that indentation should be 4 spaces. Updates one of the not-ignored files to demonstrate the rules working, as well as unignored a file that was written to spec.

### Other information?

This also updates the `package-lock.json` file as it was not updated in #6630. When updating or adding a dependency, this file must be updated.


site/.stylelintrc.json
```diff
@@ -9,13 +14,11 @@
     "public/css/pdf/toolbar_fullscreen.css",
     "public/css/pdf/toolbar_embedded.css",
     "public/css/pdf/pdf_embedded.css",
-    "public/css/google/source_sans_pro.css",
-    "public/css/google/pt_sans.css",
+    "public/css/google/*",
```
MasterOdin, line 17: We can safely ignore this whole folder. These files are auto-generated when they're pulled from google's servers.

williamjallen (approved):
Looks good.  I did have one comment about the indentation though.  A decent number of files in Submitty currently have an indentation of 4 spaces while others have an indentation of two spaces.  I'll make a PR to standardize it to four spaces soon.  Four spaces vs two spaces is a negligible difference for CSS but if we ever move to SCSS or similar, we may want to consider switching to two spaces of indentation to prevent excess whitespace when nesting elements.

Issue #6532, opened 2021-05-30
Replace models with Doctrine entities


MasterOdin:
When the database layer was originally introduced back in 2016, it was a few hundred lines and was aiming to just collect common queries in one place instead of duplicating query logic across files. Models (in-theory) were then used to model the information coming out of the database, but in many places then (and now), raw array access was used instead. With functionality of Submitty ballooning to encompass more and more, and [DatabaseQueries.php](https://github.com/Submitty/Submitty/blob/master/site/app/libraries/database/DatabaseQueries.php) exceeding 7K lines of code, it is probably well past time to refactor it into something better.

One idea would be to replace our current models with [Doctrine](https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/index.html) entities/ORM ([object relation mapping](https://en.wikipedia.org/wiki/Object%E2%80%93relational_mapping)). In Doctrine, instead of explicitly defining SQL queries, we define annotations around entities, which we then combine with an "Entity Manager" to fetch info from database or write it.

For example, take the poll feature in Submitty. Under doctrine, we might have the following `Poll` entity:

```php
<?php

declare(strict_types=1);

namespace app\entities\poll;

use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity
 * @ORM\Table(name="polls")
 */
class Poll {
  /**
   * @ORM\Id
   * @ORM\Column(name="poll_id",type="integer")
   */
  private $id;

  /** @ORM\Column(name="name",type="text") */
  private $name;

  /** @ORM\Column(type="text") */
  private $question;

  /** @ORM\Column(type="text") */
  private $status;

  /** @ORM\Column(type="date") */
  private $release_date;

  /** @ORM\Column(type="text",nullable=true) */
  private $image_path;

  /**
   * @ORM\OneToMany(targetEntity="\app\entities\poll\Option",mappedBy="table")
   * @ORM\JoinColumn(name="poll_id", referencedColumnName="poll_id"),
   * @ORM\OrderBy({"order_id" = "ASC"})
   */
  private $options;

    /**
   * @ORM\OneToMany(targetEntity="\app\entities\poll\Response",mappedBy="table")
   * @ORM\JoinColumn(name="poll_id", referencedColumnName="poll_id"),
   * @ORM\OrderBy({"order_id" = "ASC"})
   */
  private $responses;

  public function __construct() {
    $this->options = new ArrayCollection();
[...]

PR #6643, opened 2021-06-22
[Refactor:Developer] Use doctrine for entity loading

MasterOdin:

* ~~[ ] Documentation has been updated/added if relevant~~, docs will be added in follow-up PR for Poll to give better real-world examples.

Starts on #6532

Right now we have a hand-rolled model setup for interacting with the database. This has led to an explosion of lines in `DatabaseQueries.php` as we've expanded the scope of Submitty, and also that the models are sometimes used and sometimes not, leading to the layer being a lot weaker than one would like.

### What is the new behavior?

This incorporates the `doctrine/orm` library into Submitty, with the aim of replacing our model layer with its entity layer. Instead of having to write our own queries to get and save information in the database, we would instead go through doctrine's entity manager which handles this all for us. This would allow us to reduce/remove DatabaseQueries, while also having a much strong model layer in the MVC setup.

As a first step, this uses the new entity manager to get the information for the our SQLToolbox page getting rid of the hand-spun query there. It's not necessarily the best example as this _adds_ code, but it is at the cost of a much strong model set-up, which benefits into writing the twig template instead of using array accesses.

### Other information?

The next PR will aim to replace our `Poll` model with a `Poll` entity which would act as a much fuller example of using the doctrine ORM.

This branch was originally started on-top of #6434, which is why there's all the commits before mine where I add doctrine.


site/app/entities/db/Table.php
```diff
@@ -0,0 +1,54 @@
...
+   * })
+   * @ORM\OrderBy({"position" = "ASC"})
+   */
+    protected $columns;
+
+    public function __construct() {
+        $this->columns = new ArrayCollection();
+    }
+
+    public function getName() {
```
williamjallen, line 44: Should this function have a docstring and a specified return type to help with static analysis?

site/app/entities/db/Table.php
```diff
@@ -0,0 +1,54 @@
...
+        return $this->name;
+    }
+
+  /**
+   * @return ArrayCollection<Column>
+   */
+    public function getColumns() {
```
williamjallen, line 51: Does this one need a return type?

site/app/libraries/Core.php
```diff
@@ -162,6 +170,22 @@ public function setSessionManager(SessionManager $manager) {
         $this->session_manager = $manager;
     }
 
+    private function createEntityManager(AbstractDatabase $database): EntityManager {
```
williamjallen, line 173: Does this need documentation above it for the parameter? (same with the rest of the following lines)

williamjallen (changes requested):
Looks good and seems to work as intended.  Some of the functions are missing documentation/return types which might be beneficial to add here given how often some of these files are going to be used.  Other than that, the code looks good.

site/app/entities/db/Table.php
MasterOdin: Yeah, it should, as well as putting a `@var string` on the class property as well.

site/app/entities/db/Table.php
MasterOdin: Yeah, fixed.

site/app/libraries/Core.php
MasterOdin: My view that when we use inline type hints, we should only include a docstring if it includes additional type information that cannot be expressed inline. This would include stuff like union types (e.g. `string|int`) or concrete array elements (e.g. `string[]`) or if there's a template class (e.g. `Collection<Column>` here). Otherwise, there's no need for the docstring and it just adds a duplicate place we have to keep the types updated for no real benefit.

williamjallen: Ok, seems reasonable enough then.

williamjallen (approved):

PR #6652, opened 2021-06-22
[Bugfix:TAGrading] Fix PDF breaking on no changes save

### LINKED ISSUE #6651: Broken PDFs after saving with no changes
**Describe the bug**
If a user saves the pdf without making any annotations on the pdf (can't have previous annotations made by this user) then upon reloading the page, the pdf will be broken and an error will appear in the console a number of times depending on how many pages there are.
This can also happen if a user opens up a pdf with someone else's annotations, tries to erase those annotations, and presses save.

**Expected behavior**
The pdf does not break after saving.

**To Reproduce**
1. Go to a gradeable with pdf submissions
2. Click on grade on any unannotated submission
2. Open the pdf for annotating
3. Immediately click save (without annotating anything)
4. Refresh page
5. Open pdf back up
4. See error (pdf won't load)

**Configuration**
 - OS: MacOS
 - Browser Chrome

**Screenshots**
image

eddiekrystowski:
Fixes #6651 
Currently, saving a pdf when the user has never made any annotations on the pdf will cause the pdf to break. Refreshing the page and trying to load the pdf after this action will result in the pdf not loading.

### What is the new behavior?
The pdf doesn't break if it is saved without the user having any annotations.


loadedskiffer (approved):
code is straight forward and functions as described.

site/app/controllers/pdf/PDFController.php
```diff
@@ -156,7 +156,7 @@ public function savePDFAnnotation(string $gradeable_id, string $target_dir) {
         $annotation_body = [
             'file_path' => $annotation_info['file_path'],
             'grader_id' => $grader_id,
-            'annotations' => json_decode($_POST['annotation_layer'], true)
+            'annotations' => isset($_POST['annotation_layer']) ? json_decode($_POST['annotation_layer'], true) : json_decode('{}')
```
MasterOdin, line 159: Better / quicker to just return the constant value than going through `json_decode`.
```suggestion
            'annotations' => isset($_POST['annotation_layer']) ? json_decode($_POST['annotation_layer'], true) : [];
```

site/public/js/pdf/PDFInitToolbar.js
```diff
@@ -151,13 +151,13 @@ function renderPDFToolbar() {
...
-        let annotation_layer = localStorage.getItem(`${window.RENDER_OPTIONS.documentId}/${GENERAL_INFORMATION.grader_id}/annotations`);
+        let annotation_layer = localStorage.getItem(`${window.RENDER_OPTIONS.documentId}/${GENERAL_INFORMATION.grader_id}/annotations`) || {};
         $.ajax({
             type: 'POST',
             url: url,
             data: {
-                annotation_layer,
-                GENERAL_INFORMATION,
+                'annotation_layer': annotation_layer,
+                'GENERAL_INFORMATION': GENERAL_INFORMATION,
```
MasterOdin, line 160: This change isn't necessary?

MasterOdin (changes requested):

site/public/js/pdf/PDFInitToolbar.js
eddiekrystowski: Yes, it was mostly for consistency since I saw that most other AJAX requests specified the keys.

MasterOdin:
Changes in this PR were included in #6727

PR #6660, opened 2021-06-23
[UI/UX:InstructorUI] Standardize Markdown with template

### LINKED ISSUE #5884: standardize markdown css for messages, etc. throughout submitty

eddiekrystowski:
Closes #5884
Currently, there are many places across the site where markdown is implemented using the markdown class in the parent container and then the markdown twig filter on the content, although many times one of the two is missing.

### What is the new behavior?

- All instances of markdown on the site now use `misc/Markdown.twig` which is a simple wrapper that applies both the class and the filter.
- A file path has been fixed on the student polls page which was preventing it from loading
- Notebook builder bug causing notebooks with markdown to not save has been fixed.


williamjallen (approved):

loadedskiffer (approved):
Formatting all looks good

sbelsk (commented):
I just resolved the merge conflicts. everything seems to be working fine, but I think that in the poll student page, the answers look a bit strange with the way the markdown template is pasted in since the template has the content in a div and the text is no longer inline next to the radio button
markdown_poll

sbelsk (approved):

PR #6661, opened 2021-06-23
[Feature:Plagiarism] Add matching users popup

williamjallen:
Yellow text indicates that a region matches somebody, but not the user on the right.  Many times, it may be useful for an instructor to view the list of other users who match the highlighted region.

### What is the new behavior?
This PR adds a popup containing a list of users who also match the clicked location when clicking on yellow- or red-highlighted text.  Clicking on one of the elements in the list loads that user into the right panel.  In addition, a small change was made to the behavior of the UI when clicking on orange text.  When clicking on orange text, the text turns red and the user 2 panel scrolls to the first instance of the matching text.

Untitled


sbelsk (changes requested):
The user popup works well when clicking on red highlighted regions, but doesn't seen to work for me when clicking on yellow regions. 
When I try to click on yellow regions, they turn blue and I can see a little gray square, but no popup with usernames:
image
The scrolling on the right panel for red-highlighted regions seems to be working well though.

sbelsk (approved):

PR #6663, opened 2021-06-23
[Refactor:System] Fix easily-fixable CSS linting issues

williamjallen:
PR https://github.com/Submitty/Submitty/pull/6630 added stylelint CSS linting to Submitty.

### What is the new behavior?
This is a follow-up PR to do an initial round of linting.  Most of the changes are simple things like fixing indentation, adding missing semicolons, and removing or adding whitespace as appropriate.

This PR touches almost every CSS file in Submitty and it is almost certain that there will be merge conflicts with existing PRs.  As such, all of the changes made in this PR are simple and do not require changing anything other than CSS files.  A number of future PRs will be created to address the remaining issues, a few at a time, to further reduce the length of the ignored files list.  Most issues involve class-name changes for classes which do not follow the specified naming rules and the resolution will involve changing non-CSS files.


site/public/css/colors.css
```diff
@@ -1,46 +1,39 @@
...
 
-    /*buttons, primarily but also can be used for notifications/anything
-    that requires this color scheme*/
+    /* might wanna change blue borders/hover text/sidebar text to submitty blue */
+
+    /* buttons, primarily but also can be used for notifications/anything
+    that requires this color scheme */
     --danger-red: #dc3545;
-    --focus-danger-red: #ff5555;
+    --focus-danger-red: #f55;
```
MasterOdin, line 10: Might be useful to run a quick poll to see if people like the short color codes vs long color codes. Personally I like the longer one, if just that's what any hex picker I've used gives me, and this is then a style I would violate consistently.

site/public/css/diff-viewer.css
```diff
@@ -1,10 +1,14 @@
...
-/*}*/
+/* .diff-container { */
+
+/*    background-color: var(--standard-pale-gray); */
+
+/*    border: 1px solid var(--standard-light-gray); */
+
+/*    width: 100%; */
+
+/* } */
```
MasterOdin, line 11: Spacing this out over multiple lines is weird, probably better to have it be a singular block comment:
```suggestion
/*
.diff-container {
    background-color: var(--standard-pale-gray);
    border: 1px solid var(--standard-light-gray);
    width: 100%;
}
*/
```

Though perhaps just get rid of it since it's been commented out a year, and there's no real explanation on it.

site/public/css/plagiarism.css
```diff
@@ -281,15 +283,21 @@ main#main.full-screen-mode .plagiarism-result-cont {
...
 
-    /*.plag-data-group {*/
-    /*    flex-direction: row;*/
-    /*}*/
+    /*    max-width: 150px; */
 
-    /*select[name="user_id_1"], a.toggle {*/
-    /*    max-width: 150px;*/
-    /*}*/
+    /* } */
```
MasterOdin, line 302: Same thing here as above, good to convert these to a singular block comment (or get rid of altogether?)

site/public/css/colors.css
williamjallen: I agree.  Should I add a rule to standardize it to always be the full 6 characters?

site/public/css/diff-viewer.css
williamjallen: Ok, I'll get rid of it then.

site/public/css/plagiarism.css
williamjallen: Yep, I'll get rid of this one too.  I was the one who commented it out in the first place and I guess I forgot to remove it.

site/public/css/colors.css
MasterOdin: Yes, please.

PR #6665, opened 2021-06-24
[Bugfix:TAGrading] Allow pen/text resize with pen tool

### LINKED ISSUE #4599: After selecting pen, cannot change pen size or text size
**Describe the bug**
When the pen is selected, cannot change the pen size or the text size.

**Expected behavior**
Can change the pen/text size even with the pen active.

**To Reproduce**
1. Open PDF for grading
2. Select pen
3. Attempt to change pen/text size

**Configuration**
 - Browser [e.g. chrome, safari] Firefox, Chrome

eddiekrystowski:
Fixes #4599 
Currently, on the pdf annotating interface, the pen/text size cannot be changed with the pen tool selected.
Also, any other inputs even in other tabs of the ta grading interface are disabled when the pen is selected.

### What is the new behavior?
The pen/text size can be adjusted with the pen tool selected, and all other inputs can be used with the pen tool selected.

### DIFF
--- a/site/public/js/pdf/PDFAnnotateEmbedded.js
+++ b/site/public/js/pdf/PDFAnnotateEmbedded.js
@@ -269,6 +269,19 @@ function render(gradeable_id, user_id, grader_id, file_name, file_path, page_num
                                 $('#save_status').css('color', 'red');
                             }
                         });
+                        document.getElementById(`pageContainer${page_id}`).addEventListener('mouseenter', () => {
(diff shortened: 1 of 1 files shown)


MasterOdin:
This feels like a hacky solution to fixing a bug that exists within the pdf-annotate.js library itself, where it needs to limit what mouse events it captures with a given tool, since this bug only manifests itself with the pen and no other tool.

loadedskiffer (approved):
Works as expected

PR #6666, opened 2021-06-24
[Feature:Plagiarism] Plagiarism ignore submissions

sbelsk:
There is no functional backend to the "ignored submissions" option in the configuration form for the plagiarism detection tool.

### What is the new behavior?
Instructors can specify which users to ignore when running the plagiarism matching algorithm.
image
The users in each category are queried from the database to compile a list of user id's, which is then written into the `config.json` of the gradeable.

### Other information?
This PR is related to, but independent from the Lichen PR https://github.com/Submitty/Lichen/pull/37, since the "ignore_submissions" field exists in the config.json in the current Submitty master branch. This PR writes user id's to ignore into that field in the config, and everything works normally if nothing is written into it.

### DIFF
--- a/site/public/css/plagiarism.css
+++ b/site/public/css/plagiarism.css
@@ -76,7 +76,7 @@ input[type=text]:invalid {
     color: var(--alert-invalid-entry-pink);
 }
 
-#prev_gradeable_div {
+#prev-gradeable-div {
     display: flex;
     justify-content: space-between;
     flex-wrap: wrap;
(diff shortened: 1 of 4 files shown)


site/public/css/plagiarism.css
```diff
@@ -97,14 +97,17 @@ input[type=text]:invalid {
...
-    position: relative;
-    padding-bottom: 30px;
+#current_code_file, #provided_code_file, #prev_gradeable_div {
+    display: none;
 }
 
-#text_ignore_submission, #current_code_file, #provided_code_file,
-#add_more_ignore, #prev_gradeable_div {
-    display: none;
+#ignore_others_list {
```
williamjallen, line 104: Submitty now mandates that IDs have words separated by dashes.  This file is currently ignored by the CSS linter but it's close to being able to be removed from the ignored list.  Can this be changed to be #ignore-others-list?

williamjallen (changes requested):

site/public/css/plagiarism.css
sbelsk: Yup! looking at the twig file and the css, there are many other instances where we are using underscores instead of dashes in IDs. I'll fix that.

williamjallen (approved):

PR #6669, opened 2021-06-24
[Refactor:InstructorUI] Change Markdown Engine to Commonmark

### LINKED ISSUE #4208: Better Code Boxes for Forum

eddiekrystowski:

Closes #4208 
Submitty uses [Parsedown](https://github.com/erusev/parsedown) as a markdown engine. It is no longer actively maintained or updated. It also lacks customization options.

### What is the new behavior?
The markdown engine has been changed to [League\CommonMark](https://commonmark.thephpleague.com/) which follows closely with the [CommonMark](https://commonmark.org/) markdown spec. This engine also allows the definition of custom renderers/extensions, making it highly more customizable.

Due to the customization potential, markdown code blocks on Submitty (not CodeMirror) are now styled with syntax highlighting, background color, automatic language detection, line numbers, and horizontal scrolling on large lines.

image
image


cjreed121 (approved):
Works great. Not sure if you are already planning to do this but make sure the docs get updated: https://submitty.org/student/communication/markdown. There seems to be some new/different syntax for Markdown with the new engine.

site/composer.json
```diff
@@ -29,7 +29,9 @@
     "cboden/ratchet": "0.4.3",
     "textalk/websocket": "1.5.3",
     "egulias/email-validator": "3.1.1",
-    "maennchen/zipstream-php": "^2.1"
+    "maennchen/zipstream-php": "^2.1",
+    "league/commonmark": "^1.6",
+    "spatie/commonmark-highlighter": "^2.1"
```
MasterOdin, line 34: Use specific versions here. As Submitty isn't a library, using ranges doesn't get us anything over specific values which dependabot handles bumping.

To resolve the merge conflict though, I would recommend doing:
```bash
git checkout master
git pull master
git checkout commonmark
git add .
git commit # commit merge
wget https://raw.githubusercontent.com/Submitty/Submitty/master/site/composer.lock -O composer.lock
wget https://raw.githubusercontent.com/Submitty/Submitty/master/site/composer.json -O composer.json
cd site
composer require "league/commonmark:1.6.5"
composer require "spatie/commonmark-highlighter:2.1.1"
```
(typing this free-hand, probably some error in wget and composer usage as I can never fully remember the former's flags and the latter specifying specific version using `:` or `@`.

MasterOdin: Additionally, can we remove the `erusev/parsedown` dependency?

MasterOdin (changes requested):

site/composer.json
eddiekrystowski: Thanks for the help on the merge conflict. I just removed the parsedown dependency.

MasterOdin (approved):

Issue #5495, opened 2020-06-08
Simple PHP Classes to test


MasterOdin:
Here is an unsorted list of classes that should be relatively easy to write unit tests for, with that all code should be concrete, and not require the need to mock anything. There's a handful of cases where the writer would have to create temporary files and such. The unit tests should cover all aspects of functionality (setters / getters, conditionals, etc.). 

- [ ] app/libraries/ForumUtils.php
- [ ] app/models/forum/Post.php
- [ ] app/models/forum/Thread.php
- [ ] app/models/notebook/SubmissionCodeBox.php
- [ ] app/models/notebook/SubmissionMultipleChoice.php
- [ ] app/models/notebook/SubmissionTextBox.php
- [ ] app/models/Notification.php
- [ ] app/models/SimpleGradeOverriddenUser.php
- [ ] app/models/SimpleLateUser.php
- [ ] app/models/SimpleStat.php
- [ ] app/models/Team.php

cjreed121:
@MasterOdin I took a look at ForumUtils.php and found that 2/3 functions in there are never used. It seems they were duplicated and placed in the actual ForumController and used there. Should I try to get rid of ForumUtils.php or extract out any Utility function from ForumController and put it in ForumUtils.php?

PR #6671, opened 2021-06-25
[Testing:Submission] Add Tests to SimpleStat and Team models

cjreed121:

SimpleStat and Team models have no PHPUnit testing.
Partially solves #5495 

### What is the new behavior?
There are now PHPUnit tests for those files.


site/app/libraries/Core.php
```diff
@@ -244,7 +244,7 @@ public function loadForum() {
             throw new \Exception("Need to load the config before we can create a forum instance.");
         }
 
-        $this->forum = new Forum($this);
+        //$this->forum = new Forum($this);
```
shailpatels, line 247: Should this be commented out? There are other mentions of a forum member in core which all seems unused, but I think that should be removed in its own PR instead of here.

shailpatels (changes requested):

site/app/libraries/Core.php
cjreed121: I'll remove from this PR and make a new PR for it. We discussed it should be turned off from being created in Core since it is never used. We are still keeping the code so a refactor can be done in the future.

shailpatels (approved):
Looks good

PR #6674, opened 2021-06-25
[Refactor:TAGrading] Clearer function and variable names

loadedskiffer:
There are 3 places where we felt we could improve the clarity of function/variable names for peer components/gradeables

gradeable->isPeerGrading(), checks if a gradeable has a peer component, could be confusing and make people think it is checking if the gradeable is being graded by a peer.

component->isPeer(), checks if a component is a peer component, pretty straight forward but could also make people think it means the component is being graded by a peer.

is_peer_grading, a variable that checks if a gradeable is being graded by a peer, could be confusing and make people think it does what the 2 functions above do.

### What is the new behavior?
gradeable->isPeerGrading() is now gradeable->hasPeerComponent()
component->isPeer() is now component->isPeerComponent()
is_peer_grading is now is_peer_grader


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1543,7 +1543,7 @@ public function getGradeableRubric(Gradeable $gradeable, User $grader) {
         }, array_filter($gradeable->getComponents(), function (Component $component) use ($gradeable) {
                 return $this->core->getAccess()->canI('grading.electronic.view_component', ['gradeable' => $gradeable, 'component' => $component]);
         }));
-        //return $grader->getGroup() === User::GROUP_INSTRUCTOR || ($component->isPeer() === ($grader->getGroup() === User::GROUP_STUDENT));
+        //return $grader->getGroup() === User::GROUP_INSTRUCTOR || ($component->isPeerComponent() === ($grader->getGroup() === User::GROUP_STUDENT));
```
williamjallen, line 1546: Do these lines need to be removed?

williamjallen (changes requested):
Changes look good, aside from one line which is currently commented out and may need to be removed.

williamjallen (approved):

Issue #6678, opened 2021-06-28
Removing Responses When Editing a Poll


sbelsk:
**Describe the bug**
As @MasterOdin pointed out, currently, editing a poll and removing a response that students may have already selected is undefined behavior and there are no warnings to instructors or error checking preventing it from happening.

1. Within the current implementation, deleting one of the response options will not delete the answers of student who selected that response. When loading **the student poll page, the frontend does not know how to handle an unknown response** (shows neither "no response" nor one of the existing response options) so it just displays nothing:
![remove_response_bug](https://user-images.githubusercontent.com/71195502/123681420-ce2f8400-d817-11eb-8c7e-f1a283b4c00b.png)

2. Another related issue is caused by the fact that after deleting poll response option(s), when submitting the "Edit Poll" form, the other response options don't get updated. 
The potential problem with this is that each response option in a poll gets an ID (consecutive integers), so if an instructor deleted an options and it does not happen to be the last one listed, then in the next time the poll would be edited, the checks for consecutive ID's in the poll backend would trigger an error:
![remove_response_bug2](https://user-images.githubusercontent.com/71195502/123682698-55312c00-d819-11eb-9a7a-25db4973df48.png)
As a result, **after deleting one of the answers in a poll, whether or not students have submitted a response yet, if that answer was not the last one listed, the poll is left in an uneditable state.**

**Expected behavior**
In a discussion with @bmcutler, we agreed we should totally prevent deletion of poll response options if anyone had already chosen them. The "Edit Poll" page would still display the "delete" icons next to the responses, and would have a popup telling the instructor to tell the students and/or TAs to switch their answers in order to be able to delete a response.

**To Reproduce**
Steps to reproduce the frontend undefined behavior:
1. Create a poll with more than one response option (type does not matter)
2. Have a student submit an answer to that poll
3. Edit the poll, and delete the responses the student has replied with
4. Reload the student view of the poll summary page or the page for that specific poll
5. See error

Steps to reproduce the uneditable poll behavior:
1. Create a poll with more than one response option (type does not matter)
2. Edit the newly created poll, deleting one of the response options which is not the last one
3. Go to edit the poll again and try to click "submit changes"

PR #6687, opened 2021-06-29
[Bugfix:SubminiPolls] Deleting poll responses bugfix

sbelsk:

fixes https://github.com/Submitty/Submitty/issues/6678

### What is the new behavior?
- The two bugs explained in the issue were fixed
- Minor changes to the New/Edit Poll page UI that make more sense with the implementation of deleting response options
- Fixed other small bugs that existed in polls found along the way:
   - in polls with only one response option, a student who already answered that poll would get a php error in the poll summary page
   - the new/edit poll page would allow instructors to submit responses with no text, which has been observed to lead to undefined behavior corrupting the poll's data
   - if you spam the submit button when creating a new poll, the poll may be duplicated

### Other information?
When testing, make sure to test all existing features of the SubminiPolls to see that everything still works as expected. 
Try to create and edit polls while students are submitting answers, and try swapping orders of response options, deleting and adding new ones, or anything you think may cause issues with the indices the responses are assigned.
Confirm that the desired behavior is also seen on the student's end.


cjreed121 (changes requested):
The code looks good. One comment though. Your check for making sure no one has responded only happens on front end. I was able to manually send a POST request to the backend and delete a response that someone had responded to as there are no checks in the edit poll function.

loadedskiffer (changes requested):
Adding a Response with no text has undefined behavior and corrupts the poll's data. There should be an empty string check for adding responses.

site/app/controllers/PollController.php
```diff
@@ -145,22 +145,45 @@ public function addNewPoll() {
...
+                    $answers[] = $_POST["option_id_" . $i];
+                }
+                $count++;
+            }
+            if ($i === 1000) {
+                // the indices assigned to responses don't have to be consecutive,
+                // so to make sure we don't get into an infinite loop, we stop checking
+                // after an arbitrary large number in case $response_count and the
+                // actual number of responses don't match
+                $this->core->addErrorMessage("Error occured in adding poll with the number of responses provided");
```
HenrikLam, line 182: Why not loop through the $_POST array instead of checking from 0 to 1000? Although I can't imagine why the instructor will need more than 1000 different response types, this solution could break if response_count is greater than 1000. It also seems like it might be difficult to manage in the future. Maybe you can even pass in an array of option_id_i's , response_i, and order_?

HenrikLam (changes requested):
Stops instructor from deleting a used response type for all poll types! Editing after deletion of an option seems to work fine too.

eddiekrystowski (changes requested):
I'm not sure what's causing it, but when I edit a poll, add a new response, and then immediately try to delete it, I get an error in the console.
Screen Shot 2021-06-30 at 10 21 31 PM

sbelsk:
> I'm not sure what's causing it, but when I edit a poll, add a new response, and then immediately try to delete it, I get an error in the console.
> Screen Shot 2021-06-30 at 10 21 31 PM

@eddiekrystowski Hm, I tried reproducing it and didn't get the error. I did encounter it previously when fixing bugs in this PR, this happens when the ajax request triggers some php error, so instead of returning the expected output, the html for the php error (the table with the stack call) is returned. Did you test on the most recent commit on this branch? Or perhaps you did other things that led to an edge case?

site/app/controllers/PollController.php
sbelsk: I tried to change the least amount of code in the way polls currently work to fix the bug in the PR, so I didn't change the way  the responses are stored in the `$_POST` array. If I were to iterate through it to extract the indices that are relevant to the responses, I would probably need to use some regex matching for the key strings. Would that be a cleaner solution?

loadedskiffer:
> > I'm not sure what's causing it, but when I edit a poll, add a new response, and then immediately try to delete it, I get an error in the console.
> > Screen Shot 2021-06-30 at 10 21 31 PM
> 
> @eddiekrystowski Hm, I tried reproducing it and didn't get the error. I did encounter it previously when fixing bugs in this PR, this happens when the ajax request triggers some php error, so instead of returning the expected output, the html for the php error (the table with the stack call) is returned. Did you test on the most recent commit on this branch? Or perhaps you did other things that led to an edge case?

@sbelsk @eddiekrystowski I couldn't reproduce this error either

site/app/controllers/PollController.php
HenrikLam: I think that would be a good!

site/app/controllers/PollController.php
```diff
@@ -472,6 +472,16 @@ public function submitEdits(): RedirectResponse {
...
+        if ($_POST["removed_responses"] !== "") {
+            $prev_responses = $poll->getResponses();
+            $removed_resp = explode(",", $_POST["removed_responses"]);
+            foreach ($removed_resp as $response_id) {
+                if (array_key_exists($response_id, $prev_responses)) {
+                    $this->core->addErrorMessage("Error occured in editing poll: attempt to delete response option that has already been submitted as an answer");
+                    return new RedirectResponse($returnUrl);
+                }
+            }
+        }
```
cjreed121: I can still send a back end request and delete an option that was not intended. When I made that request I also got quite a few errors too.

cjreed121 (changes requested):

eddiekrystowski:
> > I'm not sure what's causing it, but when I edit a poll, add a new response, and then immediately try to delete it, I get an error in the console.
> > Screen Shot 2021-06-30 at 10 21 31 PM
> 
> @eddiekrystowski Hm, I tried reproducing it and didn't get the error. I did encounter it previously when fixing bugs in this PR, this happens when the ajax request triggers some php error, so instead of returning the expected output, the html for the php error (the table with the stack call) is returned. Did you test on the most recent commit on this branch? Or perhaps you did other things that led to an edge case?

I tried it again with different polls since I deleted the other one when testing, and I haven't been able to reproduce it either. I had this poll sitting around for a while so perhaps there was something weird with it, but all the new polls I made I wasn't able to break.

eddiekrystowski (approved):

site/app/controllers/PollController.php
MasterOdin: The better way to handle this is leverage the fact that PHP can handle array names within forms. For example, doing:

```
<input type="hidden" name="foo[]" value="1" />
<input type="hidden" name="foo[]" value="10" />
```

will give you in PHP:
```
$_POST['foo'] = ['foo' => ['1', '10']];
```

You could then just iterate over the singular array for the non-sequential IDs, and doing regex or have this kind of iterator count check.

loadedskiffer (approved):

HenrikLam (approved):
Looks good to me!

PR #6688, opened 2021-06-29
[UI/UX:System] Improve System Updates Page

### LINKED ISSUE #5787: System Update page improvements
Superusers on Submitty have the system update tab.

Some ideas for improvements:

The release notes pulled from github are currently plaintext.

* We could make weblinks to the release notes on github, e.g., 
   `https://github.com/Submitty/Submitty/releases/tag/v20.07.02`
   Currently this sort of looks like markdown, so maybe that's easy?

* The bullet list of PRs could be augmented with links to those PRs on github...
  E.g.,  [Feature:InstructorUI] Notebook Builder 4.1 (#5762)
  should point to `https://github.com/Submitty/Submitty/pull/5762`

* Also, it would be nice to specifically search for an highlight items that require [SYSADMIN ACTION]
   We could search for that specific string?  
   Let's use color, a badge, and/or a font awesome icon to make sure that this is super visible.

* Similarly, would be good to highlight if a release has security updates!
   for example:
   https://github.com/Submitty/Submitty/releases/tag/v20.06.00

* Finally, I'm not sure what this page looks like when we are multiple releases behind.  
  Would be good to paste in / link to the release notes for ALL of those changes.  
  And if ANY of them (not just the last one) include a sysadmin action or security action 
  that is highlighted in an obvious way.

eddiekrystowski:
Closes #5787
Currently, the system update page for system administrators is missing:
* Markdown on release notes
* Links to PRs/prior release notes
* Special highlighting of sysadmin/security fixes
* Search/filter feature
* Previous release notes when multiple releases behind, currently it only shows the latest.

### What is the new behavior?
Currently, the system update page for system administrators now has all of the features mentioned above as well as:
* Ability to collapse/expand specific release notes
* Release notes load when the page loads, check for updates button has been removed

### Other Information?
To test:
* Go to `site/public/js/system-update.js` and add `current_tag = 'v20.10.00';` (or some other release) to the beginning of the `updateReleaseNotes` function.

### DIFF
--- a/site/.stylelintrc.json
+++ b/site/.stylelintrc.json
@@ -23,6 +23,7 @@
     "public/css/sidebar.css",
     "public/css/server.css",
     "public/css/sequence.css",
+    "public/css/system-update.css",
     "public/css/scrollable-sidebar.css",
     "public/css/rotatingsectionsform.css",
     "public/css/rainbow-customization.css",
--- a/site/.stylelintrc.json
+++ b/site/.stylelintrc.json
@@ -23,7 +23,6 @@
     "public/css/sidebar.css",
     "public/css/server.css",
     "public/css/sequence.css",
-    "public/css/system-update.css",
     "public/css/scrollable-sidebar.css",
     "public/css/rotatingsectionsform.css",
     "public/css/rainbow-customization.css",
--- a/site/app/controllers/HomePageController.php
+++ b/site/app/controllers/HomePageController.php
@@ -286,6 +286,7 @@ class HomePageController extends AbstractController {
         }
 
         $this->core->getOutput()->addInternalJs('system-update.js');
+        $this->core->getOutput()->addInternalCss('system-update.css');
         return new WebResponse(
             'HomePage',
             'showSystemUpdatePage',
--- a/site/app/controllers/MarkdownController.php
+++ b/site/app/controllers/MarkdownController.php
@@ -17,6 +17,14 @@ class MarkdownController extends AbstractController {
     }
 
     /**
+     * @Route("/markdown", methods={"POST"})
+     * @return WebResponse
+     */
+    public function displayMarkdown() {
+        return new WebResponse(MarkdownView::class, 'renderMarkdownPreview', true, $_POST['content']);
+    }
+
+    /**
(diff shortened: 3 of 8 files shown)


site/.stylelintrc.json
```diff
@@ -23,6 +23,7 @@
     "public/css/sidebar.css",
     "public/css/server.css",
     "public/css/sequence.css",
+    "public/css/system-update.css",
```
williamjallen, line 26: Is there a particular reason why this new file is being added to the ignored list?

williamjallen (changes requested):

site/.stylelintrc.json
eddiekrystowski: I added it since the way I inject styling into the markdown string returned from the server uses a regex to add classes based on the section title. For example, the section labeled SECURITY gets the `update-SECURITY` class, but the linter doesn't like caps in class names. I tried using \L to convert the matching group to lowercase in the regex, but had no luck.  I could probably do another step in processing after the markdown is added as html onto the page to make these classes lowercase, but I thought it might be too messy for what it's worth. Thoughts?

williamjallen: Hmm, interesting issue.  I believe the best option here may be to disable stylelint using inline ignores like so:
```
/* stylelint-disable */
a {}
/* stylelint-enable */
```

MasterOdin: JS' regex does not support the `\L` flag inline to change the casing. You need to apply it as a step on the captured group itself in the replace. The suggested way to do this given the multiple groups (and only want to lowercase one) is to use a function as the second argument to `.replace`, something like:

```
markdown_data = markdown_data.replace(/<li>(\[\w+:(\w+)\].+)(?=<\/li>)|<li>(\[([^\]]+)\].+)(?=<\/li>)/g, (match, p1, p2, p3, p4) => `<li class="release-item"><span class="update-${`${p2}${p4}`.toLowerCase()}">${p1}${p3}</span>`);
```

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_function_as_a_parameter for docs on it.

williamjallen (approved):

bmcutler:
looks great!

PR #6689, opened 2021-06-30
[Feature:TAGrading] prev/next assigned students

### LINKED ISSUE #4699: Access to gradeables outside of section
**Describe the bug**
When I finished grading a section, I pressed the arrow button to see the next student. Instead, I was supplied a student gradeable that is not in my section.

**Expected behavior**
I expected the next button to somehow signify that I had hit the edge of the section.

**To Reproduce**
1. Grade a section to 100%
2. Go to grade the last student in the section
3. Click the next student button

**Configuration**
 - OS: mac
 - Browser chrome

**Additional context**
This was newly formed RCOS sections
I am a full access grader

DefinitiveAbove:
When using the prev/next student arrows, full-access graders and instructors will always find the prev/next student from all of the grading sections.

### What is the new behavior?
Closes #4699
There is now a toggle in the TA grading settings to have the prev/next arrows strictly go to students in your assigned sections.
Toggle defaults to false.

Viewing all students or assigned sections in grading index is now linked to the TA grading toggle. Currently, it uses a cookie for the grading index that always has a consistent state with the TA grading toggle. As such, the grading index will now also default to only showing students in the assigned sections.


loadedskiffer (commented):
Where is the toggle ?

loadedskiffer (approved):
Buttons work as expected

bmcutler (changes requested):
As we discussed in the meeting, the word "force" here (both on the UI and in the code) is not very intuitive.

How about naming the variable navigate_assigned_students_only
With the longer UI text "Prev/Next buttons navigate through students in assigned registration/rotating sections only".  

This option should should be available only for instructors and full access graders.  It should be true by default.  And if unselected, it should apply to all gradeables and all courses.  (Stored in either the database or browser local storage/cookie.)

Also, please test that when the option is true, it does not visit students in the NULL section, but when false it does visit students in the NULL section.

loadedskiffer (commented):
the arrows work and the code looks good but what do you think about adding some sort of success message like "grading complete" when it  takes you back to the home page to eliminate confusion ? I feel like whenever someone gets redirected to a new page there is a little bit of confusion and a simple message could eliminate that.

loadedskiffer (changes requested):
logged in as ta2, originally the index page only shows me the submissions that I am responsible for grading. When I was at the end of my section and pressed the next arrow, I was brought back to the index page as expected but it displayed all of submissions for the whole class. I also think the message should be a little more concise.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1372,6 +1373,25 @@ public function showGrading(
...
+                else {
+                    $message .= " end";
+                }
+                $message .= " of";
+                if ($navigate_assigned_students_only !== "false") {
+                    $message .= " your assigned sections";
+                }
+                else {
+                    $message .= " the list";
+                }
```
loadedskiffer, line 1389: use ternary operator to condense some of this like $navigate_assigned_students_only !== "false" ?  $message .= " your assigned sections" : $message .= " the list";

loadedskiffer (changes requested):
Functionally works great! just a few tweaks in the code

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1372,6 +1373,25 @@ public function showGrading(
...
+                if ($filter !== 'default') {
+                    $message .= " (using filter '" . $filter . "')";
+                }
```
loadedskiffer, line 1392: use the ternary operator to condense some of this like, $navigate_assigned_students_only !== "false" ?  $message .= " your assigned sections" : $message .= " the list";

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1372,6 +1373,25 @@ public function showGrading(
...
+                $message .= ".";
+                $this->core->addSuccessMessage($message);
```
loadedskiffer, line 1394: use ternary operator to condense some of this, like $navigate_assigned_students_only !== "false" ?  $message .= " your assigned sections" : $message .= " the list";

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1372,6 +1373,25 @@ public function showGrading(
...
+                if ($to === 'prev') {
+                    $message .= " start";
+                }
+                else {
+                    $message .= " end";
+                }
+                $message .= " of";
+                if ($navigate_assigned_students_only !== "false") {
+                    $message .= " your assigned sections";
+                }
```
loadedskiffer, line 1386: use ternary operator to condense some of this like $navigate_assigned_students_only !== "false" ?  $message .= " your assigned sections" : $message .= " the list";

loadedskiffer (approved):

bmcutler (changes requested):
Overall this is looking really good.
I have a bunch of (hopefully small) requests though:

* Let's change the this button interface:
Screen Shot 2021-07-30 at 9 40 15 AM
to instead be labeled "Prev/Next buttons navigate through" and have drop down choices 
"all students" and "only students in assigned registration/rotation sections"

* It seems to not remember my setting of that drop down.  I personally will almost always want the "all students" option, but (when I go to a new gradeable?) it seems to switch back to the default "only students in assigned registration/rotation sections".  Is this setting per gradeable?  I think it should just be a single global setting.  Once I chose "all students" every gradeable/course should be all students until I switch it back.  [but that's just my opinion]

* I think when I choose "all students" on a team gradeable, we are not showing students in the null section who are not on a team on the index listing all teams & all students not on a team.

* [probably unrelated to this PR] I think our revisions to the sample course data no longer have students who are not on teams for our team gradeables?  Ideally we want examples of null students on a team, null students not on a team, and non-null students not on a team.

sbelsk (approved):
I tested it, and I really like the banner message indicating that reached the end/beginning and under which settings. It worked well with and without anonymous mode enabled, with "random order" and "default order", and with different arrow functionality settings.

cjreed121 (changes requested):
Seems to work well. I do have a few comments/changes about this though. On first details page load of this PR I noticed that I couldn't see anyone to grade. I could only see that after clicking view all sections. Normally view all is enabled by default. I imagine this is because the cookie has not been created. Maybe the cookie should get created in PHP on first page load? Cookies default value could be dependent on what group you are in. Otherwise every gradeable will be this way for instructors and possibly full access graders too.

Another comment is to remove the ?view=all from any link that takes you to this page as that won't be used anymore.

DefinitiveAbove:
> Seems to work well. I do have a few comments/changes about this though. On first details page load of this PR I noticed that I couldn't see anyone to grade. I could only see that after clicking view all sections. Normally view all is enabled by default. I imagine this is because the cookie has not been created. Maybe the cookie should get created in PHP on first page load? Cookies default value could be dependent on what group you are in. Otherwise every gradeable will be this way for instructors and possibly full access graders too.

This is intended behavior (when this PR is put in, when going to a grading page for the first time it should default all users to only show assigned sections for TA grading and the grading index); it should retain the last saved state (show assigned students or show all students) across all courses/gradeables though.

cjreed121:
> > Seems to work well. I do have a few comments/changes about this though. On first details page load of this PR I noticed that I couldn't see anyone to grade. I could only see that after clicking view all sections. Normally view all is enabled by default. I imagine this is because the cookie has not been created. Maybe the cookie should get created in PHP on first page load? Cookies default value could be dependent on what group you are in. Otherwise every gradeable will be this way for instructors and possibly full access graders too.
> 
> This is intended behavior (when going to a grading page for the first time after this PR is put in should default the user to `show assigned sections` for TA grading and the grading index); it should retain the last saved state (`show assigned students` or` show all students`) across all courses/gradeables though (fixed in commit below, wasn't working as intended earlier).

Changes definitely seem good. I found a few more ?view=all in the code:

- Around line 1159 of ElectronicGraderController.php
- The SimpleGraderController.php/SimpleGraderView.php/Display.twig still use the parameter view to control the view setting it seems.
- Checkpoint gradeables also seem to use the parameter view

cjreed121 (approved):

PR #6692, opened 2021-06-30
[Bugfix:Submission] Truncate Autograding Output

### LINKED ISSUE #6234: Autograding results view should truncate excessive STDOUT/STDERR output
When viewing the autograding results, if the student has a lot of debugging output and the entire contents of STDOUT is shown, opening an autograding testcase takes a long time and hangs the page.

The easiest way is if the output is too long for STDOUT/STDERR, only load the first so many lines and the user can additionally request a full download.  This should only be necessary for files with many, many newlines (10k+).

N/A

**Additional context**
While this is probably annoying for students when they do this by accident, it interrupts the flow of grading.

cjreed121:

Currently all lines of the autograding output will be shown to the Submitty UI. This could be 10k+ lines which can cause performance issues on the page.
Closes #6234 

### What is the new behavior?
Once the lines exceed a certain amount the page will not render the code at all and say you should download the file instead.


site/app/templates/autograding/AutoChecks.twig
```diff
@@ -64,6 +64,10 @@
                 <span onclick="openPopUp('{{ popup_css_file }}', '{{ element.title }}', {{ index }}, {{ check_index }}, {{ side }});" class="key_to_click" tabindex="0">
                     <i class="fas fa-window-restore" style="cursor: pointer;"></i>
                 </span>
+
+                <span onclick="downloadTestCaseResult({{ index }}, '{{ check.filename }}', {{ display_version }}, '{{ gradeable_id }}', '{{ who }}')" tabindex="0">
+                    <i class="fas fa-download" style="cursor: pointer;"></i>
+                </span>
```
eddiekrystowski, line 70: Although this button only downloads the student's output, it also appears on the expected program output diff element. It should just not show up at all for that element.

eddiekrystowski (changes requested):
Looks good, works as described.
Just had a quick change on the download button.

eddiekrystowski (approved):
Works well, backend prevents download of files the user does not have access to.

site/app/controllers/MiscController.php
```diff
@@ -227,6 +235,33 @@ public function downloadCourseFile($dir, $path) {
         readfile($path);
     }
 
+    /**
+     * @Route("/courses/{_semester}/{_course}/gradeable/{gradeable_id}/downloadTestCaseResult")
+     */
+    public function downloadTestCaseResult($gradeable_id, $version, $test_case, $file_name, $user_id) {
```
MasterOdin, line 241: Can you add types to this function?

site/app/controllers/MiscController.php
```diff
@@ -227,6 +235,33 @@ public function downloadCourseFile($dir, $path) {
...
+            if (!$this->core->getAccess()->canI("grading.electronic.grade", ["gradeable" => $gradeable, "graded_gradeable" => $graded_gradeable])) {
+                $this->core->addErrorMessage("You do not have peremission to download this file!");
+                return new RedirectResponse($this->core->buildCourseUrl(['gradeable', $gradeable_id]));
+            }
+        }
+        $autograde = $graded_gradeable->getAutoGradedGradeable()->getAutoGradedVersionInstance($version);
+
+        $name = "details/test" . sprintf("%02d", $test_case) . "/$file_name";
+        $files = $autograde->getResultsFiles();
+        $file_path = $files[$name]['path'];
```
MasterOdin, line 254: Need to account for `$name` not existing in `$files` for an invalid `$test_case` or `$file_name`. Should do an `isset` check that shows error message and redirects on that case.

site/app/libraries/DiffViewer.php
```diff
@@ -544,99 +544,104 @@ private function getDisplay($lines, $type = self::EXPECTED, $option = self::SPEC
...
-                // character highlighting
-                foreach ($this->diff[$type][$i] as $diff) {
-                    $html_orig = htmlentities(substr($lines[$i], $current, ($diff[0] - $current)));
-                    $test = str_replace("\0", "null", $html_orig);
-                    $html_orig_error = htmlentities(substr($lines[$i], $diff[0], ($diff[1] - $diff[0] + 1)));
-                    $test2 = str_replace("\0", "null", $html_orig_error);
-                    if ($option == self::SPECIAL_CHARS_ORIGINAL) {
-                        $html .= $html_orig;
-                        $html .= "<span class='highlight-char'>" . $html_orig_error . "</span>";
+        if (count($lines) > 5000) {
```
MasterOdin, line 547: We probably want to have this check above the insertion of blank lines above, so as to avoid the potential of a tons and tons of empty lines for specific diffs.

Also, please add a test case to https://github.com/Submitty/Submitty/blob/master/site/app/libraries/DiffViewer.php#L1 for these changes.

site/app/templates/autograding/AutoChecks.twig
```diff
@@ -64,6 +64,11 @@
                 <span onclick="openPopUp('{{ popup_css_file }}', '{{ element.title }}', {{ index }}, {{ check_index }}, {{ side }});" class="key_to_click" tabindex="0">
                     <i class="fas fa-window-restore" style="cursor: pointer;"></i>
                 </span>
+                {% if side == 0 %}
+                <span onclick="downloadTestCaseResult({{ index }}, '{{ check.filename }}', {{ display_version }}, '{{ gradeable_id }}', '{{ who }}')" tabindex="0">
```
MasterOdin, line 68: Should add a `title` value to this span to include some text on what this does for accessibility.

MasterOdin (changes requested):

site/app/libraries/DiffViewer.php
cjreed121: Can you clarify your first suggestion or give an example? I am not following what you are saying.

MasterOdin: Before showing this too long message, the diff viewer does:

https://github.com/Submitty/Submitty/blob/fc17a81168719161270ca828ceb39d80d7ab5474/site/app/libraries/DiffViewer.php#L533-L541

which strikes me that it is possible to have a diff where it's some X number of empty lines followed with "This file is too long to display. Please download it instead.". We should probably not show these empty lines at all in the case that we are just going to show the too long message, especially given that the number of empty lines I think is unbounded?

DefinitiveAbove (changes requested):

site/app/libraries/DiffViewer.php
```diff
@@ -530,21 +530,34 @@ private function getDisplay($lines, $type = self::EXPECTED, $option = self::SPEC
...
                 }
                 $html .= "\t</div>\n";
             }
         }
+        if (count($lines) + $num_blanks > 1000) {
+            if ($type === self::EXPECTED) {
+                $html = "<p style='color: black;'>This file has been truncated. Please contact instructor if you feel that you need the full file.</p>" . $html;
+            }
+            elseif ($type === self::ACTUAL) {
+                $html = "<p style='color: black;'>This file has been truncated. Please download it to see the full file.</p>" . $html;
```
MasterOdin, line 549: Why append `$html` here? Isn't this just going to be a thousand lines of blank content?

cjreed121: The diff viewer will only show the first 1000 lines. Once it hits that amount then it will still show 1000 lines but also the message above to let you know what happened. The content won't necessarily be blank but it could if that's what the program filled STDOUT with.

MasterOdin (approved):

bmcutler (changes requested):

site/app/libraries/DiffViewer.php
```diff
@@ -638,6 +643,15 @@ private function getDisplay($lines, $type = self::EXPECTED, $option = self::SPEC
                 $html .= "\t</div>\n";
             }
         }
+        if (count($lines) + $num_blanks > 1000) {
+            $html .= "<p>...</p>";
+            if ($type === self::EXPECTED) {
+                $html .= "<p style='color: red;'>This file has been truncated. Please contact instructor if you feel that you need the full file.</p>";
```
DefinitiveAbove, line 649: The message should also be displayed at the top of the file.

site/app/libraries/DiffViewer.php
```diff
@@ -638,6 +643,15 @@ private function getDisplay($lines, $type = self::EXPECTED, $option = self::SPEC
                 $html .= "\t</div>\n";
             }
         }
+        if (count($lines) + $num_blanks > 1000) {
```
DefinitiveAbove, line 646: This doesn't work for carriage returns (carriage returns don't add a new line number but still render as a new line).

DefinitiveAbove (changes requested):

site/app/libraries/DiffViewer.php
cjreed121: Carriage returns don't seem to work properly with the DiffViewer at all. If you render the DiffViewer with a bunch of them you get 1 line on the DiffViewer but a huge result box. I think supporting this would require a pretty large refactor of the DiffViewer. I have not seen an autograding config use carriage returns so it's not even expected to have them.

bmcutler:
@cjreed121 
Please add the truncation message at the top & the bottom of each truncated file
If it's not too difficult, please make the font color for that message bold and red or orange (right now with black text it is easy to scroll past and miss).  Please make sure it is clearly visible in both light more & dark mode.

bmcutler:
Whoops, didn't pull your changes, I see it is red text & at the bottom.  Can you make it bold too?

bmcutler:
Also, the font for the message is different at the top vs bottom (it is fixedwidth/courier at the bottom)
For the color red, use the --error-alert-dark-red for light mode (the color used for other autograding errors)
Currently it's simply using 'red'

bmcutler (approved):
thanks for the changes :)

PR #6693, opened 2021-06-30
[Feature:TAGrading] autograding queue statistics

### LINKED ISSUE #6580: Add stat for number of assignments in queue
Currently there is no easy way to see how many submissions are currently in the queue to be graded for a gradeable.

On the status page, above "Statistics for Completely Graded Assignments:" add another statistics section that displays 
what students currently have submissions in the queue.
issue2_1

**Additional context**

loadedskiffer:
closes #6580 
Currently there is no easy way to see how many submissions are currently in the queue to be graded for a gradeable.

### What is the new behavior?
2 stats have been added on the page, 1 for number of submissions in the queue and 1 for number of submissions being graded.
The stats are is hidden from peer graders
pr12_1

### Other information?
To test,
use the batch regrade commands <a href = "https://submitty.org/instructor/assignment_preparation/automated_grading/batch_regrade" >Here</a>
to quickly add assignments to the queue and then look at the stats on the statistics page for that gradeable.


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -477,6 +477,32 @@ public function showStatus($gradeable_id) {
...
+        $submissions_in_queue = 0;
+        $submissions_grading_in_progress = 0;
+        $order = new GradingOrder($this->core, $gradeable, $this->core->getUser(), true);
+        $order->sort("id", "ASC");
+        $graded_gradeables = [];
+        $user_ids = $this->core->getQueries()->getUsersOnTeamsForGradeable($gradeable); // Collect user ids so we know who isn't on a team
+        foreach ($order->getSortedGradedGradeables() as $g) {
+            $graded_gradeables[] = $g;
+            if ($gradeable->isTeamAssignment()) {
+                $user_ids = array_merge($user_ids, $g->getSubmitter()->getTeam()->getMemberUserIds());
```
cjreed121, line 490: Why do you create and add to user ids? These never seem to be used late in the function. Also all these additions should probably be put beneath the access control check since there is no point in running a pretty big loop until we know they have permission to see the page.

cjreed121 (changes requested):
I did a batch regrade and regraded every version of every student. I was told 179 submissions would be graded but only about 90 get reported on the site since your code seems to only look at latest version. I think you should make sure you count everything as there could be several times the amount that is currently reporting.

cjreed121 (approved):

bmcutler (changes requested):
It doesn't really make sense to have a denominator or percentage on these statistics. 
Just say for example that 17 submissions are in the queue for regrading (which could be 1 submission for 17 different students, or 17 submissions from 1 student).
I don't think it matters how many are currently being graded.  But if it is included, it should not have denominator or percentage either.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -494,6 +494,27 @@ public function showStatus($gradeable_id) {
             $peer = true;
         }
 
+        //get all graded gradeables for queue stats
+        $submissions_in_queue = 0;
+        $order = new GradingOrder($this->core, $gradeable, $this->core->getUser(), true);
+        $order->sort("id", "ASC");
+        $graded_gradeables = [];
+        foreach ($order->getSortedGradedGradeables() as $g) {
```
cjreed121, line 502: The getSortedGradedGradeables already returns an array of GradedGradeables. Why loop through them and add them when you could just set directory or just use this function in line 506?

cjreed121 (changes requested):

site/app/controllers/grading/ElectronicGraderController.php
cjreed121: Since you don't need the gradeables sorted you could do a more efficient call. Just directly call getGradedGradeables in DatabaseQueries. The users, team, and sort_keys can all be left as null.

cjreed121 (approved):

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -494,6 +494,23 @@ public function showStatus($gradeable_id) {
             $peer = true;
         }
 
+        //get all graded gradeables for queue stats
+        $submissions_in_queue = 0;
+        $gradeables[] = $gradeable;
+        $graded_gradeables = $this->core->getQueries()->getGradedGradeables($gradeables);
+        //check every submission (inactive and active) for every student
+        foreach ($graded_gradeables as $g) {
```
bmcutler, line 502: I'm worried this nested for loop might be unnecessarily expensive.
Would be better to look files in the queue directory for matches to this assignment.
But I think we can leave it for a future refactor, when eventually 
the queue is stored in the database instead of the file system.

PR #6695, opened 2021-07-01
[Testing:System] Update Cypress Access Check

cjreed121:

Right now Cypress tests have a hard-coded semester/year string.

### What is the new behavior?
Cypress generates the string every time it's run so when the next semester code starts the tests pass without needed a change on this.

### DIFF
--- a/site/cypress/integration/login.spec.js
+++ b/site/cypress/integration/login.spec.js
@@ -1,4 +1,4 @@
-import {buildUrl} from '../support/utils.js';
+import {buildUrl, getCurrentSemesterYearString} from '../support/utils.js';
 
 describe('Test cases revolving around the logging in functionality of the site', () => {
     describe('Test cases where the user should succesfully login', () => {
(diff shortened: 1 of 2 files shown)


site/cypress/integration/login.spec.js
```diff
@@ -52,7 +52,8 @@ describe('Test cases revolving around the logging in functionality of the site',
         it('should check if you can access a course', () => {
             cy.login('pearsr');
             cy.visit(['sample']);
-            cy.get('.content').should('have.text', '\n    You don\'t have access to this course. \n    This is sample for Spring 2021. \n    If you think this is a mistake, please contact your instructor to gain access. \n    click  here  to back to homepage and see your courses list.\n');
```
shailpatels, line 55: Thanks for catching this, I think making a utility function for getting the semester name plus string is overly specific, maybe returning just the semester name would let the function be more general in the future and the current date can be collected from anywhere.

But yeah I didn't catch the year when I copied and pasted the text, you could rewrite this line to be

```
cy.get('.content').contains('You don't have access to this course');
```

Which would also make it less fragile.

shailpatels (changes requested):

shailpatels (approved):

PR #6701, opened 2021-07-02
[Testing:SubminiPolls] Add unit test for poll model

sbelsk:

There are no tests implemented for the PollModel.php file

### What is the new behavior?
Added a test file that tests each function in the poll model using three different poll examples.
Tested and debugged.

### DIFF
--- a/site/tests/app/models/PollModelTester.php
+++ b/site/tests/app/models/PollModelTester.php
@@ -1,0 +1,197 @@
+<?php
+namespace tests\app\models;
+
+use app\libraries\Core;
+use app\models\PollModel;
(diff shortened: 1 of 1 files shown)


shailpatels (approved):

shailpatels:
For future reference, PHP arrays are by default associative so

```php
$test = [0,1,2];
var_dump($test);
```

returns

```
array(3) { [0]=> int(0) [1]=> int(1) [2]=> int(2) } 
```

which is the same result of
```php
$test = [0 => 0, 1 => 1, 2 => 2];
```

But since this is test code that won't be running on production I don't think its a big deal that it needs to be changed

PR #6708, opened 2021-07-06
[Refactor:SubminiPolls] Migrate polls to doctrine

MasterOdin:

The Submini Poll system uses the old system of models and manual DB queries.

### What is the new behavior?

This PR rewrites it to use doctrine and entity based system.

### Other information?

To do:
- [x] Remove poll queries from report controller
- [x] Rewrite poll model tests to entity
- [x] Write unit tests for code


sbelsk (changes requested):
From an initial review of this PR (I tested on this branch on an existing Submitty VM that I ran beforehand on main), I found the following two issues:
1. The "Release response histogram to students" dropdown in "edit poll" form is always set to "never" when the page is loaded, regardless of the configured option for that poll.
2. Whenever a student changes their answer to an open poll, the responses of all other students for that poll somehow get erased from the database. As a result, everyone else's responses get reset to "no response", and the count of the responses on the instructor page gets set to 1.

I'll continue testing once you fix those bugs.

sbelsk (changes requested):
The two previous error seem to have been fixed. I found 2 more bugs:
1. radio vs. checkbox on the student polls page don't display properly for the respective type of the poll. Single response polls should display radio buttons, and multiple response should be checkboxes. Also, when I was testing this on a poll that is a "single response" type, and submitting more than one response didn't fail (when I reloaded the poll page, the student response showed the two responses selected). There should be more checks in the backend for users submitting multiple responses.
2. Submitting "no response" on a poll of type "multiple response" triggers a php error "Invalid option id"

sbelsk (changes requested):
New changes look good. I tried to create a new poll and got the following error:
image

site/phpstan-baseline.neon
```diff
@@ -371,6 +360,26 @@ parameters:
 			count: 1
 			path: app/entities/email/EmailEntity.php
 
+		-
+			message: "#^Property app\\\\entities\\\\poll\\\\Option\\:\\:\\$id is never written, only read\\.$#"
+			count: 1
+			path: app/entities/poll/Option.php
+
+		-
+			message: "#^Property app\\\\entities\\\\poll\\\\Option\\:\\:\\$poll \\(app\\\\entities\\\\poll\\\\Poll\\) does not accept null\\.$#"
```
cjreed121, line 369: Could allow for type to be null so this can go away

site/.performance_warning_ignore.json
```diff
@@ -27,7 +27,4 @@
   "/courses/<semester>/<course>/course_materials/delete",
   "/courses/<semester>/<course>/course_materials/modify_timestamp",
   "/courses/<semester>/<course>/course_materials/edit",
```
cjreed121, line 29: Invalid JSON issue
```suggestion
  "/courses/<semester>/<course>/course_materials/edit"
```

cjreed121 (changes requested):
Code looks good and queries are minimal. Couple minor requested changes. Did a small test locally and seemed to work well.

migration/migrator/migrations/course/20221028222200_poll_response_migration_v2.py
```diff
@@ -0,0 +1,45 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("""
+    ALTER TABLE poll_responses DROP CONSTRAINT IF EXISTS poll_responses_id_seq;
```
DefinitiveAbove, line 18: May want to add indexes into the table on parts such as student_id in poll_responses (used in findAllByStudentIDWithAllOptions)

site/app/templates/polls/PollForm.twig
```diff
@@ -0,0 +1,403 @@
...
+            "placeholder" : "Enter question here...",
+            "preview_div_id" : "poll_preview",
+            "preview_div_name" : "poll_preview",
+            "onclick" : "previewPollMarkdown.call(this)",
+            "render_buttons" : true,
+            "min_height" : "100px"
+        } only %}
+        <br/><br/>
+
+        <span class="option-title" >
```
DefinitiveAbove, line 34: Code segment here looks to be identical to the code in PollType.twig - might want to put it into a separate twig file and include

DefinitiveAbove (changes requested):
Code looks good; some minor changes requested.

site/phpstan-baseline.neon
williamjallen: Isn't this deliberate though?  We never want nulls in this DB column (and we want Doctrine to know that), but we need to be able to delete an option without deleting a poll, right?  It definitely seems like something which should be fixed, but I'm not sure how to go about doing it best.

site/app/templates/polls/PollForm.twig
williamjallen: `PollType.twig` actually appears to be entirely unused, so I will delete it altogether.

migration/migrator/migrations/course/20221028222200_poll_response_migration_v2.py
williamjallen: Given that a large number of insert operations will be taking place as a class submits responses to a poll, and that there's already an index on the primary key which identifies all of the responses associated with a given poll, I'm somewhat wary of adding additional indices to the responses table.  Thoughts?

I have no hard data to reference unfortunately, so I'm definitely open to counterarguments on this matter.  Perhaps we could get this PR merged first, and then play around with the indexing sometime later down the road.

site/phpstan-baseline.neon
cjreed121: Good point. You could consider enabling orphanRemoval which makes the code simpler and would remove this error.

Here's an example from my course materials refactor:
https://github.com/Submitty/Submitty/blob/a882f5a032c251bf13ca358d11e24d0651d30821/site/app/entities/course/CourseMaterial.php#L66-L70
https://github.com/Submitty/Submitty/blob/a882f5a032c251bf13ca358d11e24d0651d30821/site/app/entities/course/CourseMaterial.php#L178-L180
https://github.com/Submitty/Submitty/blob/a882f5a032c251bf13ca358d11e24d0651d30821/site/app/controllers/course/CourseMaterialsController.php#L419-L423

williamjallen: Good call.  Updates made.

site/app/templates/polls/NewPollPage.twig
```diff
@@ -1,451 +0,0 @@
...
-                || (poll_type != "multiple-response"
-                    && ($('#poll-type-multiple-response-exact').is(':checked')
-                        || $('#poll-type-multiple-response-flexible').is(':checked')
-                        || $('#poll-type-multiple-response-survey').is(':checked')))) {
-                    // don't submit yet
-                    event.preventDefault();
-                    // send warning
-                    if (confirm("Changing the poll question type may lead to data of students' answers being " +
-                               "lost if students have already begun submitting responses to this poll. " +
-                               "Are you sure you want to change the type of this poll?")) {
```
sbelsk, line 438: We probably want this warning message back in case instructors accidentally change the poll type after responses have been submitted to it, since it's undefined behavior.

sbelsk (approved):
I tested and all looked good, except for the one comment I left about adding back the warning when changing the poll type. We could use other people to test is thoroughly as well, just in case.

site/app/templates/polls/NewPollPage.twig
williamjallen: The warning was actually there, but the logic wasn't being triggered correctly.  Changes made.

site/.performance_warning_ignore.json
```diff
@@ -26,5 +26,8 @@
   "/courses/<semester>/<course>/course_materials/upload",
   "/courses/<semester>/<course>/course_materials/delete",
   "/courses/<semester>/<course>/course_materials/modify_timestamp",
-  "/courses/<semester>/<course>/course_materials/edit"
+  "/courses/<semester>/<course>/course_materials/edit",
+  "/courses/<semester>/<course>/polls/newPoll",
+  "/courses/<semester>/<course>/polls/editPoll/submitEdits",
+  "/courses/<semester>/<course>/polls/deletePoll"
```
williamjallen, line 32: Getting Doctrine to group insert operations into one statement seems to be impractical.  Given that these are all instructor-side routes, with only a handful of duplicated queries, these should not be problematic at all.

cjreed121 (approved):
Code looks good to me. No issues from my testing. Great job!

PR #6717, opened 2021-07-06
[Bugfix:TAGrading] Custom mark arrows

### LINKED ISSUE #6694: Selection issue when setting a custom grade
Not a Submitty contributor, but wanted to point out some strange behavior I noticed when recently when grading on it.
**Describe the bug**
In the grading view, when assigning a number of points to the "Custom" option, there are up and down arrows that appear in the textbox, moving the number up or down by 0.5 at a time. The down arrow works fine, but after clicking the up arrow once, the down arrow is automatically selected again (regardless of cursor position still being over the up arrow). This means that if a user hovers over the up arrow and clicks twice, the score will go up 0.5 and then down 0.5. Moving the cursor again updates the selection to again be whichever arrow the cursor is over.

**Expected behavior**
I would expect the up arrow to be selected by default after clicking the up arrow, and to only change when the cursor is moved off of it.

**To Reproduce**
1. Open a gradeable, and look at the rubric view.
2. Move mouse over the textbox to the left of the "Custom" tag
3. click the upper arrow twice, and observe the score go up and then down, and the down arrow being the one now selected.

**Configuration**
 - OS: Windows 10 
 - Browser: Google Chrome Version 91.0.4472.114

**Additional context**
The re-selection (switching to down arrow) appears to happen on mouseup, as holding the mousebutton down on either arrow works properly, continually increasing/decreasing the score. Moving back and forth between the up and down arrows without letting go of the mouse also works properly, increasing/decreasing depending on the arrow currently below the cursor

loadedskiffer:

closes #6694

### What is the new behavior?
I had to disable the default arrows and add custom ones that don't glitch when the rubric refreshes and updates scores. 

NOTE:

### Other information?
to test:
first on master, mash the up arrow when trying to give someone a custom mark and see error
Now try on new branch

### DIFF
--- a/site/public/css/electronic.css
+++ b/site/public/css/electronic.css
@@ -37,12 +37,6 @@ aside {
     overflow-y: auto;
 }
 
-.custom-mark-number-input::-webkit-outer-spin-button,
-.custom-mark-number-input::-webkit-inner-spin-button {
-    -webkit-appearance: none;
-    margin: 0;
-}
-
 #electronic-gradeable-container {
     height: 100%;
     padding: 0;
@@ -714,6 +708,12 @@ progress::-webkit-progress-value {
     height: 22px;
     width: 22px;
     margin: 5px;
+}
+
+.custom-mark-number-input::-webkit-outer-spin-button,
+.custom-mark-number-input::-webkit-inner-spin-button {
+    -webkit-appearance: none;
+    margin: 0;
 }
 
 .ta-nav-wrapper {
--- a/site/public/css/electronic.css
+++ b/site/public/css/electronic.css
@@ -37,6 +37,12 @@ aside {
     overflow-y: auto;
 }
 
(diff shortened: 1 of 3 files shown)


HenrikLam:
Looks good and works as intended!

HenrikLam (approved):
Looks good and works as intended!

OwenMcGee:
Not sure if this is intentional (or considered important), but holding down click on the arrows no longer continues to increase/decrease for the duration that the mouse key is held (it just increases/decreases a single time, upon mouseup).

loadedskiffer:
This was not intentional, would you like me to add that functionality ?

OwenMcGee:
I honestly don't see it being used very much at all, but it is I think what I would _expect_ it to do. Either option seems fine to me, as long as it's intentional.

PR #6721, opened 2021-07-07
[Feature:System] Add plagiarism course to vagrant up

sbelsk:

### What is the new behavior?
This PR creates a course that will get created along with the other sample courses when "vagrant up" is run.
The course has pre-made gradeables with submissions that serve as test cases specifically designed to test the Plagiarism Detection Tool features. 
The files used for the submissions to the gradeables in this course are taken from the LichenTestData private repo, so having it cloned on your local machine before running vagrant is necessary for the creation of the course.

The code in this PR relies on the features introduced in PR https://github.com/Submitty/Submitty/pull/2216


shailpatels:
Where is the lichen test data directory located, we could pull it in during CI and have the full vagrant up test run with lichen installed

williamjallen:
> Where is the lichen test data directory located, we could pull it in during CI and have the full vagrant up test run with lichen installed

Oops, I didn't see this until just now since I was gone for the weekend.  The LichenTestData repo is a private repo, accessible only to members of the organization, so you have to clone it to the directory next to the Submitty/Submitty directory.  If it's there, it will be copied into the relevant GIT_CHECKOUT directory and the plagiarism course will be created.  Eventually we will probably want to add this to the CI, and add a number of Lichen tests but that will be in a future PR.

sbelsk:
@williamjallen I ran vagrant up last night on this branch and the course was successfully created together with its gradeables and submissions (yay!). There are two more things we still need to do before this PR is ready:
- For gradeables where we specify names of users to submit hardcoded files, we currently have the rest of the students in the course submit "empty" submissions, where no files are submitted to version 1 but the `user_assignment_settings.json` is created. Ideally when we hardcode the users we want only the specified names to submit to the gradeable.
- There is a specific testcase/gradeable (the ignore_submissions medium) that relies on the existence of the users `grader`, `grader2`, and `ta2`, but they are not listed in the "manage graders" page and their submissions are never made to that gradeable. Each of those users have their `.yml` files under `.setup/data/users` and they   exits in the sample course, so we need to find a way for them to appear in this course too.

williamjallen:
@shailpatels Do you know why these tests are failing by any chance?  I can't tell if this PR broke the test somehow or if this is just a flaky test.  It appears to be failing when it checks this URL:
Untitled

shailpatels:
@williamjallen the submission page tries to reload the page when it's the (autograding?) config 3 times which is kept track via the url.

This might just be bad luck with setting up the sample courses sometimes a student makes a submission and the page is broken when it shows no version selected.

I would suggest changing the url on that test to a different page that does not change since that test is checking if you try visiting a page not logged in, then after logging in you get sent to the page you originally entered.

Something like the docker UI, or course settings would be good, you can make a PR to handle that or I can this evening

sbelsk:
> @williamjallen I ran vagrant up last night on this branch and the course was successfully created together with its gradeables and submissions (yay!). There are two more things we still need to do before this PR is ready:
> 
> * For gradeables where we specify names of users to submit hardcoded files, we currently have the rest of the students in the course submit "empty" submissions, where no files are submitted to version 1 but the `user_assignment_settings.json` is created. Ideally when we hardcode the users we want only the specified names to submit to the gradeable.
> * There is a specific testcase/gradeable (the ignore_submissions medium) that relies on the existence of the users `grader`, `grader2`, and `ta2`, but they are not listed in the "manage graders" page and their submissions are never made to that gradeable. Each of those users have their `.yml` files under `.setup/data/users` and they   exits in the sample course, so we need to find a way for them to appear in this course too.

All of those have been fixed. I ran vagrant up last night test to it without the private LichenTestData repo cloned, and everything seemed ok.

williamjallen (approved):

PR #6723, opened 2021-07-08
[Feature:Submission] Add Spellcheck to Small CodeMirror

### LINKED ISSUE #4392: (Option to?) Disable capitalization & spellcheck for Notebook textboxes
A student is typing code in a Submitty notebook.
The browser autocapitalizes some of their typing...

"for (int i = 0; ..."  => . "for (int I = 0; ..."

Or spell corrects:

"std::cout << x << std::endl;"  => "std::count << x << std::end;"

The user can disable these in their browser preferences (but that would affect all webpages).

Let's consider disabling this for our pages (when appropriate -- e.g., when it's a code short answer box)
e.g.,

https://www.wufoo.com/html5/spellcheck-attribute/
https://stackoverflow.com/questions/11102729/how-to-prevent-auto-capitalization-in-ios-safari-input-tags

eddiekrystowski:
Fixes #4392 
Currently, small CodeMirror inputs have no spell checking. Small CodeMirror inputs are found on notebooks where the instructor might only want the student to enter a couple words, but not necessarily code.

### What is the new behavior?
Small CodeMirror inputs now have spell checking.


site/package.json
```diff
@@ -16,6 +16,7 @@
     "bootstrap": "4.6.0",
     "chosen-js": "1.8.7",
     "codemirror": "5.62.0",
+    "codemirror-spell-checker": "^1.1.2",
```
MasterOdin, line 19: We don't use ranges for dependencies (though there are some that do mistakenly at the moment). This is because Submitty is an app and not a library.

```suggestion
    "codemirror-spell-checker": "1.1.2",
```

MasterOdin (commented):
Hm, bit concerned that the dependency looks to be completely unmaintained, and also that it's got an issue named [Don't show numbers as spelling errors](https://github.com/sparksuite/codemirror-spell-checker/issues/32).

eddiekrystowski:
> Hm, bit concerned that the dependency looks to be completely unmaintained, and also that it's got an issue named [Don't show numbers as spelling errors](https://github.com/sparksuite/codemirror-spell-checker/issues/32). This isn't necessarily a bad thing, but this may be something we need to fork ourselves and maintain (similar to the pdf-annotate.js library), especially depending on how it works with CM6.

I agree. I also was taking a look at CM6, and it seems that they removed `defineMode` which is central to this dependency working. I'm sure there is some alternative method, but it won't work right out of the box.

MasterOdin:
Yeah, not sure how hard it is to migrate a CM plugin from 5 to 6. However, I wouldn't necessarily treat this as a blocker if there's no better spellcheck plugin available as it will probably be some time before we migrate to CM6 anyway.

cjreed121 (approved):
Works well

bmcutler (changes requested):
I tested this with Notebook Basic example.
Yes, we've got spell check on the short 1 line tall non code boxes, looks good.
But the multi-line, non-code (e.g. haiku) box does not have spell check.
Did we decide it would only work on the single line or am I using it wrong (incorrectly install/rebuild/reload)?

eddiekrystowski:
Good point, I've just made it so any plaintext CodeMirror should use spell-checking

bmcutler (approved):
yay great :)

PR #6726, opened 2021-07-08
[Bugfix:Developer] Fix Twig Errors on Gradeables/TA Grading

eddiekrystowski:
After #6713 which enforced strict variables in twig files while developing, there are many twig errors throughout the Gradeable/TA Grading pages.

### What is the new behavior?
All twig errors I could find by navigating from the Gradeables page have been resolved. It is possible I missed a couple pages/cases but the main functionalities of TA Grading should be working now.

### DIFF
--- a/site/app/controllers/admin/PlagiarismController.php
+++ b/site/app/controllers/admin/PlagiarismController.php
@@ -395,13 +395,19 @@ class PlagiarismController extends AbstractController {
 
 
         // Prior terms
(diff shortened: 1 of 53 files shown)


loadedskiffer (changes requested):
There is an exception being thrown on the status page for some peer assignments so I can't access the TA Grading interface for them. I know you can't fix everything with this pr but can you take a look at this one

loadedskiffer (approved):
Fixes a number of twig exceptions and should be a priority to be merged

PR #6727, opened 2021-07-08
[Bugfix:TAGrading] Fix Disappearing Text Annotations on PDF

### LINKED ISSUE #6699: Annotating PDFs doesn't work with typing text
**Describe the bug**
When attempting to annotate a student's PDF, typing text is either not working a tall or allows the text to be types but then all text disappears, depending on the browser. Clicking "Save" button has no effect.

**Expected behavior**
Text typed in the annotation node of grading PDFs should remain visible on the page and saved when the "Save" button is clicked.

**To Reproduce**
1. Open a student's PDF (I was grading article01 gradeable from CSCI-4963).
2. Click on "A" and then click anywhere on the PDF to start typing. The box appears and allows you to type (see screenshot 1).
3. When you click outside of the box to confirm the entry (or press Enter), the box disappears along with the text (screenshot 2).
4. Clicking the "Save" button says "Saved" but the text is still gone (screenshot 2).

**Configuration**
 - Browser: Google Chrome Version 91.0.4472.124 (Official Build) (64-bit)

**Screenshots**
image
image
image

**Additional context**
"Text size" combo box that can be accessed by clicking the 
image
icon doesn't have any font size selected by default. It just shows blank space.

eddiekrystowski:

Fixes #6699
Currently, if you create a text annotation without ever explicitly setting the text size, the text will disappear after the textbox loses focus. Then, upon saving and refreshing the page, all other text will disappear from the page. Now, even if you set the text size, text annotations are broken.

### What is the new behavior?
The text size will always have an initial value of 12 to prevent this. The pen size has also received an initial value of 5.

### Other information?
To reproduce the error for testing, clear localStorage before entering the gradeable, then enter the pdf annotator, create text, and observe it disappearing.
_In order to fix the pdf, you must go to 
`/var/local/submtity/<course path>/annotations/<gradeable id>` and delete the directory associated with the user whose pdf you are annotating._

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -1019,6 +1019,7 @@ CREATE TABLE public.threads (
     status integer DEFAULT 0 NOT NULL,
     lock_thread_date timestamp with time zone,
     pinned_expiration timestamp with time zone DEFAULT '1900-01-01 00:00:00-05'::timestamp with time zone NOT NULL,
+    announced timestamp(6) with time zone DEFAULT NULL::timestamp with time zone,
     CONSTRAINT threads_status_check CHECK ((status = ANY (ARRAY['-1'::integer, 0, 1])))
 );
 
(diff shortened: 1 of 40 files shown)


loadedskiffer (approved):
works as expected

PR #6728, opened 2021-07-09
[Testing:CourseMaterials] Fix flaky test

shailpatels:

The last test in the course materials cypress tests is flaky and often fails waiting for a page load.
This may be from a long test running and cypress slowing down the CI Job by recording everything

### What is the new behavior?
Force reload and add a wait to try and avoid getting stuck on a page

### DIFF
--- a/site/cypress/integration/courseMaterials.spec.js
+++ b/site/cypress/integration/courseMaterials.spec.js
@@ -7,6 +7,7 @@ describe('Test cases revolving around course material uploading and access contr
     before(() => {
         cy.visit('/');
         cy.login();
+        cy.wait(500);
         cy.visit(['sample', 'course_materials']);
     });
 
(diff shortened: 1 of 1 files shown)


site/cypress/integration/courseMaterials.spec.js
```diff
@@ -356,5 +356,6 @@ describe('Test cases revolving around course material uploading and access contr
 
         cy.get('.fa-trash').first().click();
         cy.get('.btn-danger').click();
+        cy.wait(1000);
```
MasterOdin, line 359: What does the wait here at the end of the test accomplish? All prior commands have executed at this point? Or is this just to protect on the teardown method?

shailpatels: Yeah it looks like there was an issue with loading the page after being finished and running the logout/login after each test.
The page not being response or cypress not seeing that its loaded, but it doesn't appear to fix it completely

PR #6729, opened 2021-07-09
[Bugfix:InstructorUI] Fix Various Twig Files

cjreed121:

If you edit a gradeable in a course with no full access or limited access graders, the page will not load. The sections page does not load at all. The email status page also does not load.

### What is the new behavior?
There is now a check to see if the proper key exists in the array in the twig file now. There are also key checks for the sections page and email status page.


HenrikLam (approved):
Logic for email status page looks good! The other files look good too.

PR #6730, opened 2021-07-09
[Bugfix:InstructorUI] Hidden file errors

### LINKED ISSUE #5924: Specify Files to Hide From Peer Graders

### LINKED ISSUE #6020: Specify files to hide from peers cannot handle spaces
**Describe the bug**
When adding more than 2 wildcards (exact matches, *s, [ ]s, ?s) if the user adds a space the third wildcard is removed from this list and not saved. This is an issue because the page does say "saved" but reloading the page shows that some of the data has been lost and this might cause instructors to think a file is getting hidden from peers when in reality its not.

**Expected behavior**
Not data should be lost when adding wildcards to the input string for excluding files and spaces should be handled correctly 

**To Reproduce**
1. Create new gradeable
2. Allow manual grading
3. Add peer grading component/question
4. Go to peer matrix tab
5. Add a string of 3 wildcards with spaces in the final
6. Hit enter to save, refresh the page
7. see error.

loadedskiffer:
closes #6020
closes #5924 

### What is the new behavior?
All wildcard / file names entered in the input box are saved and displayed correctly.

### Other information?

- As an instructor, edit a peer gradeable
- Go to the peer matrix panel
- Attempt to enter file names with spaces or wildcards with spaces in the "Enter any Wildcards..." box
- save and reload
- see error

### NOTE 
if you get a twig error "gradeable does not exist" when trying to get to the student index while logged in as a peer grader,
comment out lines 609-626 in app/templates/grading/electronic/details.twig. The total column for peer grading has been broken for a while but the recent changes to twig settings makes it so errors are thrown. There is another open PR that will fix this so just ignore it for now.

Since the error had to do with spaces, I made sure that the matching function properly matches file names with spaces and it works as expected.

### DIFF
--- a/site/app/templates/grading/electronic/Details.twig
+++ b/site/app/templates/grading/electronic/Details.twig
@@ -507,12 +507,14 @@
         {% set btn_class = "btn-primary" %}
     {% endif %}
     <td>
-        <a class="btn {{ btn_class }}" href="{{ context.grade_url }}?who_id={{ graded_gradeable.getSubmitter().getAnonId() }}&sort={{ context.sort }}&direction={{ context.direction }}">
-            {{ contents }}
-            {% if badge_count > 0 and graded_gradeable.getGradeable().isGradeInquiryPerComponentAllowed()%}
-                <span class="notification-badge">{{ badge_count }}</span>
-            {% endif %}
-        </a>
+        {% if badge_count is defined %}
+            <a class="btn {{ btn_class }}" href="{{ context.grade_url }}?who_id={{ graded_gradeable.getSubmitter().getAnonId() }}&sort={{ context.sort }}&direction={{ context.direction }}">
+                {{ contents }}
+                {% if badge_count > 0 and graded_gradeable.getGradeable().isGradeInquiryPerComponentAllowed()%}
+                    <span class="notification-badge">{{ badge_count }}</span>
+                {% endif %}
+            </a>
+        {% endif %}
     </td>
 {% endmacro %}
 
@@ -604,18 +606,18 @@
 
 {# Total Point Summary (peer) #}
 {% macro render_column_total_peer(context, section, graded_gradeable, index, info, column) %}
-    {# todo: test this (no way to create peer assignments currently) #}
-    {# todo: this is a lot of math for the view, we should move this into the Gradeable model #}
+     todo: test this (no way to create peer assignments currently)
+     todo: this is a lot of math for the view, we should move this into the Gradeable model
     {% set graded = gradeable.getGradedNonHiddenPoints() %}
     {% set components = gradeable.getComponentsGradedBy(core.getUser().getId()) %}
     {% for component in components %}
-        {# getScore is only the custom "mark" need to write a getTotalComponentScore and also make it clear or change name of Score #}
+         getScore is only the custom "mark" need to write a getTotalComponentScore and also make it clear or change name of Score
         {% set graded = graded + component.getScore() %}
     {% endfor %}
-    {#
+
         instead of autograding_score it should be total autograding possible
         I don't think total_peer_grading_points ever gets set...it should be set in the gradeable constructor
-    #}
+
     {% set total_possible = gradeable.getGradedNonHiddenPoints() + gradeable.getTotalPeerGradingPoints() %}
     <td>
         {% if gradeable.validateVersions() %}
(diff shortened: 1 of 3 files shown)


shailpatels:
Were you able to verify if the hidden file filters were working correctly? I am trying to view a peer gradeable from a student grader's perspective but get a frog robot for an undefined variable which is probably unrelated to this PR but I want to make sure adding a filter with a space correctly hides a file from student graders.

I set the "Open peer homework" into closed/being graded mode, then uploaded a file that should match the regex by aphacker, then logged in as browna to try and grade aphacker and to make sure the file did not appear but could not get past the grading index page without the error.

Here is the stack trace
```

app\exceptions\OutputException (Code: 0) thrown in /usr/local/submitty/site/app/libraries/Output.php (Line 344) by:


Message:
Variable "badge_count" does not exist. in /usr/local/submitty/site/app/templates/grading/electronic/Details.twig:512

Stack Trace:
#0 /usr/local/submitty/site/app/views/grading/ElectronicGraderView.php(832): app\libraries\Output->renderTwigTemplate()
#1 unknown file(unknown line): app\views\grading\ElectronicGraderView->detailsPage()
#2 /usr/local/submitty/site/app/libraries/Output.php(206): call_user_func_array()
#3 /usr/local/submitty/site/app/libraries/Output.php(175): app\libraries\Output->renderTemplate()
#4 /usr/local/submitty/site/app/controllers/grading/ElectronicGraderController.php(923): app\libraries\Output->renderOutput()
#5 unknown file(unknown line): app\controllers\grading\ElectronicGraderController->showDetails()
#6 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(218): call_user_func_array()
#7 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(196): app\libraries\routers\WebRouter->run()
#8 /usr/local/submitty/site/public/index.php(129): app\libraries\routers\WebRouter::getWebResponse()
```

loadedskiffer:
@shailpatels I just pushed another small fix that ensures the filter is matching all wildcards properly regardless of spacing. I wasn't getting the error you were getting but I think I fixed it so you should be able to look now. I logged in as a peer grader and I couldn't see any of the files that I excluded (my file.txt, *cats*,  bad file.txt). Now, regardless of spaces before/after commas and in between file names, they are matched and hidden.

shailpatels:
Sounds good, I'll take a look this evening and I saw your comment about the twig variable I expected it was unrelated and the error I saw was probably related.

site/app/templates/grading/electronic/Details.twig
```diff
@@ -604,18 +606,18 @@
 
 {# Total Point Summary (peer) #}
 {% macro render_column_total_peer(context, section, graded_gradeable, index, info, column) %}
-    {# todo: test this (no way to create peer assignments currently) #}
-    {# todo: this is a lot of math for the view, we should move this into the Gradeable model #}
+{#     todo: test this (no way to create peer assignments currently)#}
+{#     todo: this is a lot of math for the view, we should move this into the Gradeable model#}
```
shailpatels, line 610: are these changes meant to be included ?

shailpatels (approved):
Filtering files works as expected and are not displayed on the dom

site/app/templates/grading/electronic/Details.twig
loadedskiffer: I was trying to comment something out for testing and I guess I messed it up a bit, I'll remove those changes

PR #6731, opened 2021-07-09
[Bugfix:TAGrading] Add Panel Message for All Empty Panels

### LINKED ISSUE #6368: More intuitive visuals when no panel is selected in TA grading UI

eddiekrystowski:
Fixes #6368 
Currently, when there are empty panels when grading, a message telling the user to choose one from above only displays if the user is in single panel mode. All other modes don't show this message.

image
---
image
---
image

### What is the new behavior?
Now all panel layouts have this functionality

image
---
image
---
image

### Other Information
Also, a small fix was made to the header to stop the buttons from overlapping the student/team name on small screens.

### DIFF
--- a/site/app/templates/grading/electronic/NavigationBar.twig
+++ b/site/app/templates/grading/electronic/NavigationBar.twig
@@ -105,7 +105,7 @@
 
         const overlap_margin = 15;
         const overlapping = (panel_buttons_bbox.left - name_div_bbox.right) < overlap_margin;
-        if (overlapping) {
+        if (overlapping || taLayoutDet.isFullLeftColumnMode) {
             $('#grading-panel-student-name').hide()
         }
     });
--- a/site/app/templates/grading/electronic/NavigationBar.twig
+++ b/site/app/templates/grading/electronic/NavigationBar.twig
@@ -104,7 +104,7 @@
         const name_div_bbox = name_div[0].getClientRects()[0];
 
         const overlap_margin = 15;
-        const overlapping = (toolbar_bbox.left - name_div_bbox.right) < overlap_margin;
+        const overlapping = (panel_buttons_bbox.left - name_div_bbox.right) < overlap_margin;
         if (overlapping) {
             $('#grading-panel-student-name').hide()
         }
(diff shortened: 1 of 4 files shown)


loadedskiffer (changes requested):
There seems to be bug where if you load into the grading interface and there is no panel selected, the message is not displayed. If you select and then deselect a panel, the message is then displayed.

panel_bug

loadedskiffer (changes requested):
one more small bug, if I load into the grading interface with no panel selected and then choose a panel layout with a full nav bar (not a full height left column) the message displays on top of the panels.
panel_bug2

loadedskiffer (approved):

bmcutler (changes requested):
If I select a mode with full panel height on left, it initially does the layout without the user/team name details.  But then if I resize my browser window, the names show up and awkwardly overlap.  (Same bug that Miles noted?  Doesn't seem to be fully fixed)

Also, if I either refresh the browser window or advance to the "next" student with a blank/unselected panel, the panel gets filled in with content -- it forgets that it previously was blank.

eddiekrystowski:
@bmcutler From my testing it seems that panels getting filled when navigating between students also happens on master. Would you like me to make the panels stay blank?

bmcutler:
@eddiekrystowski
I don't remember what/when we made a decision was made for the initial/default behavior, 
but now that we have the message, let's start with each panel blank, and if the user unselects a panel for display let's 'remember' the blank.
(otherwise this new feature/message is almost useless since its visible for only fleetingly)
note: if we go from 3 panels to 4 panels, only one of the new panels should be blank/unselected

bmcutler (commented):
yay, change works well!

PR #6732, opened 2021-07-09
[SECURITY][Bugfix:TAGrading] Don't load or zip symlinks

cjreed121:
Right now symlinks can cause issues when downloading a zip file of a VCS gradeable. Symlinks can also be used to view their targeting directories in the TA grading file browser.

### What is the new behavior?
Symlinks are now skipped if they appear in the VCS checkout folder. Symlinks are also skipped in the FileUtils getAllFiles function to avoid attempting to follow symlinks from the TA page.


MasterOdin (approved):
Updated `FileUtils.php` tests to include symlinks to ensure they're properly filtered.

PR #6735, opened 2021-07-13
[Feature:Developer] Add new flag for quick clean on install

cjreed121:

There are quite a few developers here that use PHPStorm rather than the code watcher script. PHPStorm will copy the files directly into the site folder while vagrant will copy the changes into the GIT repo on the VM. The rsync command only finds differences between site folder and GIT repo site folder which will always be equivalent with PHPStorm and therefore never update. This causes composer, npm, and permissions to never get updated when using submitty_install.

### What is the new behavior?
There is a new new runtime flag "clean quick" which will do a much shorter version of the submitty_install clean. This would delete most files in site but not all like vendor or node modules to reduce install time. This would be run by PHPStorm users when they switch branches, encounter permission issues, or install a new package.

This also contains an addition to exclude from rsync for CSS linting as there is no reason to compare/send that.

Submitty docs will be updated for PHPStorm to easily use the submitty_install script with the flag.

### Other information
Testing the install scripts:
`submitty_install`: 30 seconds
`submitty_install clean`: 1 min 40 seconds
`submitty_install clean quick`: 50 seconds
This does add a little more time compared to submitty_install but it does make it work well. The alternative would to use code watcher with PHPStorm. The PHPStorm deployment is almost instantaneous while the code watcher takes a few seconds while the site updates. Just a few edits with code watcher would already make the new runtime flag worth running.

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -58,7 +58,7 @@ if [[ "$#" -ge 1 && "$1" != "test" && "$1" != "clean" && "$1" != "test_rainbow"
        && "$1" != "skip_web_restart" && "$1" != "disable_shipper_worker" ]]; then
     echo -e "Usage:"
     echo -e "   ./INSTALL_SUBMITTY.sh"
-    echo -e "   ./INSTALL_SUBMITTY.sh clean"
+    echo -e "   ./INSTALL_SUBMITTY.sh clean quick"
     echo -e "   ./INSTALL_SUBMITTY.sh clean test"
     echo -e "   ./INSTALL_SUBMITTY.sh clean skip_web_restart"
     echo -e "   ./INSTALL_SUBMITTY.sh clear test  <test_case_1>"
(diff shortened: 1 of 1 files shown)


MasterOdin (changes requested):
I'm fine adding a way to force `composer install` to run, but this should be behind a runtime flag, and not always run within vagrant. I have no interest in having `composer install` run every time I run the installer as I do not use PHPStorm hooked directly into the installed site.

For context, I added these checks in as I've (and also @bmcutler) done work on Submitty while offline and wanted to avoid composer/npm failures around networking.

cjreed121:
> I'm fine adding a way to force `composer install` to run, but this should be behind a runtime flag, and not always run within vagrant. I have no interest in having `composer install` run every time I run the installer as I do not use PHPStorm hooked directly into the installed site.
> 
> For context, I added these checks in as I've (and also @bmcutler) done work on Submitty while offline and wanted to avoid composer/npm failures around networking.

Yeah that sounds like a better option. I wasn't sure what the best way to go about this initially so I went with vagrant. Where would I put this runtime flag? Would this be something that goes on vagrant up or how does it work? I would also want to do this check everywhere the rsync command gets called as it also affects permissions, npm, and maybe something else.

MasterOdin:
The flag itself only needs to exist on the installer as the `vagrant up` is not affected by where PHPStorm is hooking in.

MasterOdin (approved):
Added a `shift` when we use the `quick` command so that `clean quick test` would work as expected. Otherwise, this looks fine. We probably want to add some documentation of using `quick` to the https://submitty.org/developer/phpstorm docs page?

PR #6737, opened 2021-07-13
[Bugfix:InstructorUI] Fix markdown editor for NB builder

### LINKED ISSUE #6736: Markdown widget is broken in NB builder
**Describe the bug**
Attempting to add a markdown section to a notebook using the notebook builder is broken.
It throws the following error in the console.
issue6_1
 

**Expected behavior**
Adding markdown should not produce any errors

**To Reproduce**
1.  login as instructor
2.  create notebook gradeable and use the notebook builder
2.  add a markdown section, add some text
3.  hit save and open console

**Configuration**
 - OS: Windows
 - Browser chrome, edge

loadedskiffer:
closes #6736 

### What is the new behavior?
Adding markdown to a notebook using the notebook builder works as expected.

### To Test
1. As instructor, make a notebook gradeable and use the notebook builder
2. try to add mark down and click save

### DIFF
--- a/site/app/templates/misc/MarkdownArea.twig
+++ b/site/app/templates/misc/MarkdownArea.twig
@@ -35,7 +35,7 @@
     {% include "misc/AccessibilityMessage.twig" %}
     <textarea
         id="{{ markdown_area_id }}"
-        class="fill-available {{ class is defined ? class : 'markdown-input'}}"
+        class="fill-available {{ class is defined ? class }}"
         name="{{ markdown_area_name is defined ? markdown_area_name : "markdown_area" }}"
         {% if placeholder is defined %}
             placeholder="{{ placeholder }}"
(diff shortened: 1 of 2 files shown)


MasterOdin (changes requested):
Proper fix looks to be to add `'class' => $data['class']` to https://github.com/Submitty/Submitty/blob/master/site/app/views/MarkdownView.php#L28. However, that function should probably be modified so that it does checking for the properties before attempting to pass them to the twig template, something like:

```php
$args = [];
$keys = [
	'markdown_area_id',
	'markdown_area_name',
	'markdown_area_value',
    'class',
    'onclick',
	'markdown_buttons_id',
    'other_textarea_attributes',
    'render_buttons',
    'placeholder',
    'preview_div_id',
    'textarea_onchange',
    'textarea_onkeydown',
    'textarea_onpaste',
    'textarea_maxlength',
    'data_previous_comment'
];
foreach ($keys as $key) {
    if (isset($data[$key])) {
        $args[$key] = $data[key];
    }
}
```

eddiekrystowski (approved):
Looks good, thanks for fixing this.

MasterOdin (approved):

PR #6741, opened 2021-07-14
[Feature:TAGrading] Rotate submitted images

### LINKED ISSUE #6276: Render notebook submitted images (and pdf?) within notebook
A student takes a quiz/test formatted as a notebook.
Some answers are textbook, some are multiple choice, some are uploaded images (png/jpg) or pdf.
Right now, images/pdf have to be viewed by the grader from within the file browser.
Sometimes they don't display well inline in the file browser and have to be popped out.
It's annoying to do the pop out if you're grading many students (have to manually close each window).

Let's display submitted images (and maybe pdf) inline in the notebook.
Both to the student (so they can confirm they uploaded good images)
And to the grader

Are there any security risks? 
what if the image is very large (high resolution)?
Might still want a popout or download option.

DefinitiveAbove:
You cannot rotate submitted images inline.

### What is the new behavior?
Closes #6276 
When viewing an image (file with a MIME type starting with `image`) in TA grading, icons appear above the image that allow you to rotate it clockwise and counterclockwise in increments of 90 degrees.
When rotating an image, it rotates other images that are loaded in through an iframe with the same URL as the image being rotated (mainly between the submission and notebook panel) (image rotation is client side only)


cjreed121 (commented):
I found a bug with this feature. I am on Linux using Chrome.

If you go into "Show file in full panel" and back the image will disappear and only show after rotating the image. On master this works fine.

Steps to reproduce:
1. Go to TA-grading on a gradeable  with an image submitted
2. Go into files and find the file that you submitted
3. Open the file and see it display on screen
4. Click the "Show file in full panel" button
5. Click the back button and see no image
6. The only way to fix it in this state seems to be to rotate it. Opening and closing dropdown does not seem to fix it.

cjreed121 (approved):
Works well. Tested on Linux Chrome

PR #6742, opened 2021-07-14
[Refactor:System] Stores Groups in Master DB

cjreed121:

Right now when creating a new course through the website, you must pick a course to base from. This simply finds a group to give the new course.

### What is the new behavior?
Rather than ask for a course, the site just asks for the group directly which comes from a new table. This new table will also be used in plagiarism.

### DIFF
--- a/migration/migrator/data/submitty_db.sql
+++ b/migration/migrator/data/submitty_db.sql
@@ -294,17 +294,7 @@ SET default_with_oids = false;
 CREATE TABLE public.courses (
     semester character varying(255) NOT NULL,
     course character varying(255) NOT NULL,
-    status smallint DEFAULT 1 NOT NULL
-);
-
-
---
--- Name: courses_groups; Type: TABLE; Schema: public; Owner: -
---
-
-CREATE TABLE public.courses_groups (
-    semester character varying(255) NOT NULL,
-    course character varying(255) NOT NULL,
+    status smallint DEFAULT 1 NOT NULL,
     group_name character varying(255) NOT NULL
 );
 
@@ -464,14 +454,6 @@ CREATE TABLE public.users (
 --
 
 ALTER TABLE ONLY public.emails ALTER COLUMN id SET DEFAULT nextval('public.emails_id_seq'::regclass);
-
-
---
--- Name: courses_groups courses_groups_pkey; Type: CONSTRAINT; Schema: public; Owner: -
---
-
-ALTER TABLE ONLY public.courses_groups
-    ADD CONSTRAINT courses_groups_pkey PRIMARY KEY (semester, course, group_name);
 
 
 --
--- a/migration/migrator/data/submitty_db.sql
+++ b/migration/migrator/data/submitty_db.sql
@@ -299,6 +299,17 @@ CREATE TABLE public.courses (
 
 
 --
+-- Name: courses_groups; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.courses_groups (
+    semester character varying(255) NOT NULL,
+    course character varying(255) NOT NULL,
+    group_name character varying(255) NOT NULL
+);
+
+
+--
 -- Name: courses_registration_sections; Type: TABLE; Schema: public; Owner: -
 --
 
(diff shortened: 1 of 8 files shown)


migration/migrator/data/submitty_db.sql
```diff
@@ -298,6 +298,17 @@ CREATE TABLE public.courses (
...
 
+--
+-- Name: courses_groups; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.courses_groups (
+    semester character varying(255) NOT NULL,
+    course character varying(255) NOT NULL,
+    group_name character varying(255) NOT NULL
+);
```
MasterOdin, line 309: Why have this be a separate table from `courses`?

sbelsk (commented):
For the plagiarism backend, it would be convenient if we had a function that takes a semester parameter and a course id parameter, and returns the group that corresponds with that course

migration/migrator/migrations/master/20210714135438_add_course_groups.py
```diff
@@ -0,0 +1,46 @@
...
+    :type database: migrator.db.Database
+    """
+    database.execute("ALTER TABLE courses ADD COLUMN IF NOT EXISTS group_name varchar(255);")
+    database.execute("ALTER TABLE courses ADD COLUMN IF NOT EXISTS owner_name varchar(255);")
+    database.execute("ALTER TABLE courses DROP CONSTRAINT IF EXISTS group_validate;")
+    database.execute("ALTER TABLE courses DROP CONSTRAINT IF EXISTS owner_validate;")
+    database.execute("ALTER TABLE courses ADD CONSTRAINT group_validate CHECK (group_name ~ '^[a-zA-Z0-9_-]*$')")
+    database.execute("ALTER TABLE courses ADD CONSTRAINT owner_validate CHECK (owner_name ~ '^[a-zA-Z0-9_-]*$')")
+    semester_dir = Path(config.submitty['submitty_data_dir'], 'courses')
+    for semester in semester_dir.iterdir():
```
bmcutler, line 22: This isn't safe to do on production.  We have directories from old terms that do not have corresponding Submitty databases / are not in the master database.

So we'll need to wrap this with some checks/exception handling.

migration/migrator/migrations/master/20210714135438_add_course_groups.py
```diff
@@ -0,0 +1,46 @@
...
+            query = "UPDATE courses SET group_name=:g, owner_name=:o WHERE semester = :s AND course = :c;"
+            params = {
+                's': semester.name,
+                'c': course.name,
+                'g': course.group(),
+                'o': course.owner()
+            }
+            database.session.execute(text(query), params)
+    # we should be able to force the not null constraint now
+    database.execute("ALTER TABLE courses ALTER COLUMN group_name SET NOT NULL;")
```
bmcutler, line 33: And perhaps the reverse is true, a course in the database that is missing a directory (if old course directories are not currently mounted on the file system or are inaccessible).  

Let's write this migration more defensively.  We can set the user and group to root if the directory does not exist or if the directory is inaccessible.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7355,6 +7355,15 @@ public function getGradeableAccessTeam(
         return $this->course_db->rows();
     }
 
+    public function getUserGroups(string $user_id): array {
```
bmcutler, line 7358: What is the purpose of this function?  It should only grab the group for courses in which that user has instructor access?

bmcutler (changes requested):
Just read through the code, I would still like to test it after these comments are addressed.

migration/migrator/migrations/master/20210714135438_add_course_groups.py
cjreed121: The UPDATE query is actually not dangerous or unsafe. If it does not find a course in the courses table then the query does nothing and moves on without any errors.

migration/migrator/migrations/master/20210714135438_add_course_groups.py
cjreed121: This is now properly set before we enforce the not null constraint.

site/app/libraries/database/DatabaseQueries.php
cjreed121: I added a check to make sure the user is an instructor of that course. The purpose of this function is to feed to the front end for faculty to choose the group they want to use.

bmcutler (approved):

PR #6745, opened 2021-07-15
[Bugfix:TAGrading] Prevent hidden file url hack

loadedskiffer:
A peer grader can do a url hack to download files that are hidden from them.

### What is the new behavior?
Hidden files are now totally protected from people without access to them.

### Other information?
Shail, follow the same steps you were using before


shailpatels (approved):
Looks good, the download and view file endpoints both correctly filter files if the user adds an extra space in their keywords and redirect if a user does not have permission.

PR #6746, opened 2021-07-15
[Bugfix:Autograding] Remove duplicate one_of field from json

### LINKED ISSUE #6257: Using one_of causing schema validation error
**Describe the bug**
When using the `one_of` property in a config, it can cause a validation error where a top-level `one_of` is left while we properly insert it into the validation object.

**Expected behavior**
No validation error, no top-level `one_of` key/value pair.

**To Reproduce**
Short config file:
```
{
    "testcases" : [ 
        {
            "type" : "FileCheck",
            "title" : "file",
            "points" : 0,
            "one_of" : true,
            "actual_file" : [ "file.pdf" ]
        },
    ]
}
```

This blows-up to something like:

```
{
    "details": "",
    "dispatcher_actions": [],
    "extra_credit": false,
    "hidden": false,
    "one_of": true,
    "points": 0,
    "pre_commands": [],
    "publish_actions": false,
    "release_hidden_details": false,
    "testcase_id": "test01",
    "testcase_label": "",
    "timestamped_stdout": false,
    "title": "file",
    "type": "FileCheck",
    "use_router": false,
    "validation": [
        {
            "actual_file": [
                "file.pdf"
            ],
            "deduction": 1.0,
            "description": "file.pdf",
            "method": "fileExists",
            "one_of": true,
            "show_actual": "always",
            "show_message": "always"
        }
    ],
    "view_testcase_message": true
}
```

Notice the `one_of` both at the top-level and within the validation object.

cjreed121:

Building a test case with the one_of field will cause a schema validation error as there will be two one_of fields in the inflated json.
Closes #6257 

### What is the new behavior?
The top level one_of field is removed when built.

### Other information?
I tested the same example provided in the issue before and after and this fixes it.

### DIFF
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -1250,12 +1250,13 @@ void General_Helper(nlohmann::json &single_testcase) {
 }
 
 void FileCheck_Helper(nlohmann::json &single_testcase) {
-  nlohmann::json::iterator f_itr,v_itr,m_itr,itr;
+  nlohmann::json::iterator f_itr,v_itr,m_itr,o_itr,itr;
 
   // Check the required fields for all test types
   f_itr = single_testcase.find("actual_file");
   v_itr = single_testcase.find("validation");
   m_itr = single_testcase.find("max_submissions");
+  o_itr = single_testcase.find("one_of");
 
   if (f_itr != single_testcase.end()) {
     // need to rewrite to use a validation
@@ -1279,6 +1280,7 @@ void FileCheck_Helper(nlohmann::json &single_testcase) {
     }
     single_testcase["validation"].push_back(v);
     single_testcase.erase(f_itr);
+    single_testcase.erase(o_itr);
   } else if (v_itr != single_testcase.end()) {
     // already has a validation
   } else {


MasterOdin:
Can you come up with a working config that uses the `one_of` key, and then setup a testcase in https://github.com/Submitty/Submitty/tree/master/tests/integrationTests/tests that includes the config, with a single testcase of just validating the schema (similar to https://github.com/Submitty/Submitty/tree/master/tests/integrationTests/tests/tutorial_16_docker_network_python)?

cjreed121:
> Can you come up with a working config that uses the `one_of` key, and then setup a testcase in https://github.com/Submitty/Submitty/tree/master/tests/integrationTests/tests that includes the config, with a single testcase of just validating the schema (similar to https://github.com/Submitty/Submitty/tree/master/tests/integrationTests/tests/tutorial_16_docker_network_python)?

@MasterOdin I can make one. What exactly does 'one_of' control in the autograding? I haven't really seen it before and there doesn't seem to be any documentation on itt.

MasterOdin:
By default, for the `fileExists` validation method passes only if all files in the `actual_file` array exists. Using the `one_of` flag allows toggling it so that it passes if at least one file exists.

A better example than the issue would be:

```
        {
            "type" : "FileCheck",
            "title" : "file",
            "points" : 0,
            "one_of" : true,
            "actual_file" : [ "foo.txt", "bar.txt" ]
        },
```

A test case for this would probably just be a simple config file with a test cases for all possible combinations:
1. `one_of: true` where there is no matching file
2. `one_of: true` where one file matches
3. `one_of: true` where all files match
4. `one_of: false` where no matching file
5. `one_of: false` where one matching file
6. `one_of: false` where no matching files

There would then only be one test function within the integration suite that uses this one config file and the submission to it would just include a handful of empty files that fulfill the above.

Would maybe also be good to include documentation on the method both within the complete_config.json schema file and on submitty.org?

cjreed121:
Alright. one_of makes much more sense when there is more than one filename. I'll write an integration test for this. Once tests are done I'll put in some documentation for this.

Also for case 6 in your comment: I'm assuming that is supposed to be where all files match?

cjreed121:
Where would something like this go in the submitty.org docs?

PR #6747, opened 2021-07-15
[Feature:Developer] Add flag to skip grading in setup script

MasterOdin:

Right now, for any submissions that are created will be run through the autograder after `setup_sample_courses.py` is run. This takes quite a long amount of time to do, and so we just don't create submissions under certain conditions (such as on CI) to avoid this.

### What is the new behavior?

Adds a few flag to the `setup_sample_courses.py` script that allows skipping the regrade step, so that autograding will not have anything to do. This flag can be used within CI so that the generated courses have submissions, but not waste time autograding.


MasterOdin:
@williamjallen here's a patch that I think does what you were asking for in webex, where the `setup_sample_courses.py` script will now create the submissions, but just not run them through the autograder.

To then get some of the ignored selenium tests to work, I think you'd then just selectively run the regrade script for a few specific people / gradeables, and then add a watch on `/var/local/submitty/to_be_graded_queue` (or whatever the folder is called) to wait on that folder being empty before continuing.

williamjallen (commented):
Looks good.  I'll run it in a few minutes to double-check that it works.

williamjallen (approved):
Seems to work

site/app/templates/grading/electronic/ta_status/StatusBase.twig
```diff
@@ -262,7 +262,7 @@
             {% if core.getUser().getGradingRegistrationSections()|length != 0 %}
                 {%  if team_assignment %}
                     <details class="btn btn-primary" id="details_button">
-                        <summary float="left">
```
bmcutler, line 265: Did you mean to include these changes in this PR?

williamjallen: @eddiekrystowski made those changes [here](https://github.com/Submitty/Submitty/pull/6747/commits/3831daaed1d44155873b356a08d62ea8b5b75445)  The changes were made to make it pass the tests I believe.

eddiekrystowski: Yes, it was failing an accessibility test since the float attribute is not allowed on the summary tag. Not sure why it was failing in this PR and not others.

MasterOdin: We were originally only checking accessibility on a  TA status page where there was no submissions is my guess.

PR #6749, opened 2021-07-15
[Feature:HelpQueue] regex for Office hours contact info

### LINKED ISSUE #5870: Improve Contact Information Field for Office Hours Queue
* Let's make the form box for typing contact information bigger.  It isn't big enough for a long webex URL.

* Also, let's add an option that will help enforce requirements for contact info.  It was suggested it be a regexp to ensure the student has typed in a webex url for example.  (or alternately require skype or whatever)

loadedskiffer:
closes #5870 

### What is the new behavior?
Instructors have the option to enter a regex pattern that all student's contact information must match. They can enter any pattern they want or choose from the few that are provided under "help with regex patterns".  

### Other information?
1. as instructor create a few different queues with different regex patterns
2. as a student attempt to join the queues with different contact info


site/app/templates/officeHoursQueue/QueueFilter.twig
```diff
@@ -2,6 +2,23 @@
...
+  <b>Help With Regex Patterns</b>
+    <a id="regex-help-icon" onclick="toggleRegexHelp()" onkeypress="toggleRegexHelp()" title="Help with regex patterns" aria-label="help with regex patterns" tabindex="0" style="padding:.5rem;">
+        <i class="fas fa-question"></i>
+    </a>
+    <div class="regex-help" style="display: none;">
+        Regex Patterns are to make sure your students enter the correct information
+        into the "contact information" field. Copy the patterns associated with the type of information you wish
+        your students to provide (copy the pound symbols too).<br>
+        <b>NOTE: REGEX PATTERNS ARE NOT REQUIRED</b><br>
+        <a href="https://www.php.net/manual/en/reference.pcre.pattern.syntax.php" target="_blank">Write Your Own Pattern</a><br>
```
williamjallen, line 14: Do we want to link to a page in the PHP documentation or would it be better to link a page on submitty.org?

site/app/templates/officeHoursQueue/QueueFilter.twig
```diff
@@ -10,7 +27,8 @@
     <a id="new_queue_rand_token" onclick="genRandCode('new_queue_token')" onkeypress="genRandCode('new_queue_token')" title="Generate random access code" aria-label="Generate random access code" tabindex="0" style="padding:.5rem;">
       <i class="fas fa-sync"></i>
     </a>
-    <button id="open_new_queue_btn" type="submit" class="btn btn-primary">Open New Queue</button>
+      <input type="text" id="regex" name="regex" placeholder="Contact Info Regex Pattern" aria-label="Regex" maxlength="2048">
```
williamjallen, line 30: should the max length be validated on the backend as well?

williamjallen (changes requested):
Seems to work as intended but I have a few concerns about the UI.
1.

To me, it looks a little odd with smaller bolded text directly above a bolded header.  In addition, it is not immediately clear that the question mark is clickable.  If I didn't see you click on the question mark during the daily meeting, I would have never known that it was an option.  

2.

Would it be better to expand the contact info box all the way to the right?  As it is now, it's hard to see the placeholder text and it is likely that many regex patterns will be longer and thus go beyond the right side of the input field.

site/app/templates/officeHoursQueue/QueueFilter.twig
loadedskiffer: I think it'd be better to link the php documentation because it has everything someone would need to figure out how to write their own pattern. It's super detailed and I think it'd be a waste of time to make our own documentation for regex.

site/app/templates/officeHoursQueue/QueueFilter.twig
loadedskiffer: I don't think it's necessary in this case, honestly do you think there should be a max-length attribute at all ?

loadedskiffer:
@williamjallen I made some changes to "Help with Regex." As far as expanding the contact info box, it always takes up the whole panel on my end so I don't think expanding the box will turn out great.
pr13_1

site/app/templates/officeHoursQueue/QueueFilter.twig
williamjallen: I'm not sure.  I assumed that it was there to prevent some error with the maximum length allowable for the regex or something.  If there is no max length, and the DB doesn't have a max length, I don't see any reason why it can't be removed

williamjallen (approved):

migration/migrator/data/course_tables.sql
```diff
@@ -13,20 +13,6 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
---
--- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: -
---
-
-CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
```
bmcutler, line 20: Why did this blob of the two .sql files get deleted?

loadedskiffer: something strange happened when I did a course dump, all fixed now.

bmcutler (changes requested):
I'm confused why these samples have the `#` symbol.   Why are we creating some new regex syntax?

I would expect to type: 
`*@gmail.com` for a gmail address or
`*@*` for any other email address (why do we need a special keyword for email?) or
`https://*.webex.com/meet/*` for a personal meeting room or 
`https://*.webex.com/*` for any webex or

This GUI is getting kinda messy.  I don't think the exansion of information is necessary, so I removed it and rearranged a bit.


More importantly...  

This feature is meaningless if the instructor doesn't check the "contact info" option from the course settings page.  Students won't even be able to enter contact info.  Don't show the contact info regex in the queue settings modal if that option isn't enabled from course settings.

Also, please make the contact info regex pattern inbox boxes (both of them) wider.   Pasting that widest webex sample in the box should visibly 'fit' with extra space.

site/app/templates/officeHoursQueue/QueueFilter.twig
```diff
@@ -10,7 +12,21 @@
...
+
+        <div class="regex-help">
+            <br>You may optionally specify a regular expression to ensure your students enter the requested
+            format for the contact information.  For example: <br>
+            <b>Gmail:&nbsp;&nbsp;</b> <tt>*@gmail.com</tt><br>
+            <b>Email:&nbsp;&nbsp;</b> <tt>*@*</tt><br>
+            <b>Webex Personal Room:&nbsp;&nbsp;</b> <tt>https://*.webex.com/meet/*</tt><br>
+            <b>Webex Meeting:&nbsp;&nbsp;</b> <tt>https://*.webex.com/*</tt><br>
+            <b>Skype link:&nbsp;&nbsp;</b> <tt>*join.skype*</tt><br>
+            <b>Zoom link:&nbsp;&nbsp;</b> <tt>*zoom.us*</tt><br><br>
```
eddiekrystowski, line 26: Try these regexes instead. Also, it may be a good idea to trim the input either when sending it to the server, or before the server attempts to match it to the regex (didn't look too thoroughly, you might already be doing this).
```suggestion
            <b>Gmail:&nbsp;&nbsp;</b> <tt>.+@gmail.com</tt><br>
            <b>Email:&nbsp;&nbsp;</b> <tt>.+@.+\..+</tt><br>
            <b>Webex Personal Room:&nbsp;&nbsp;</b> <tt>^https:\/\/.+.webex.com\/meet\/[^/]+$*</tt><br>
            <b>Webex Meeting:&nbsp;&nbsp;</b> <tt>^https:\/\/.+\.webex\.com\/.*</tt><br>
            <b>Skype link:&nbsp;&nbsp;</b> <tt>.+\.join\.skype.+</tt><br>
            <b>Zoom link:&nbsp;&nbsp;</b> <tt>.+\.zoom\.us.+</tt><br><br>
```

eddiekrystowski: Also, if you want to test any regexes, I recommend this site https://regexr.com/

loadedskiffer: thank you !

williamjallen (changes requested):

williamjallen (changes requested):
UI looks good overall but I still have a couple concerns.

1. The regex pattern box for the "modify old queue" section expands as much as possible but the regex pattern box for the new queue does not.  Since many of these regex patterns are quite long, I feel like it should expand to the maximum extent possible.  
Untitled

2. On smaller screens, (including common laptop sizes), the example regex options take up almost 1/3 of the vertical space.  I suspect that most people who are using this page are going to be opening/closing queues with the controls at the bottom so this adds another action (scrolling) which is required to get to the bottom of the modal.  Would it be better to have some sort of menu which unfolds when an arrow (clearly marked) is clicked?
Untitled

3. Should there be a little bit of space between the label and value for each of these?
Untitled

site/app/templates/officeHoursQueue/QueueFilter.twig
```diff
@@ -10,7 +12,21 @@
...
+
+        <div class="regex-help">
+            <br>You may optionally specify a regular expression to ensure your students enter the requested
+            format for the contact information.  For example: <br>
+            <b>Gmail:&nbsp;&nbsp;</b> <tt>.+@gmail.com</tt><br>
+            <b>Email:&nbsp;&nbsp;</b> <tt>.+@.+\..+</tt><br>
+            <b>Webex Personal Room:&nbsp;&nbsp;</b> <tt>^https:\/\/.+\.webex\.com\/meet\/.*</tt><br>
+            <b>Webex Meeting:&nbsp;&nbsp;</b> <tt>^https:\/\/.+\.webex\.com\/.*</tt><br>
+            <b>Skype link:&nbsp;&nbsp;</b> <tt>.*join.skype.com.*</tt><br>
+            <b>Zoom link:&nbsp;&nbsp;</b> <tt>.*zoom.us.*</tt><br><br>
```
eddiekrystowski, line 26:
```suggestion
            <b>Skype link:&nbsp;&nbsp;</b> <tt>.*join\.skype\.com.*</tt><br>
            <b>Zoom link:&nbsp;&nbsp;</b> <tt>.*zoom\.us.*</tt><br><br>
```
These periods need to be escaped otherwise they mean any character.

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -127,7 +127,18 @@ public function addPerson($queue_code) {
                 );
             }
             else {
-                $contact_info = $_POST['contact_info'];
+                //make sure contact information matches instructors regex pattern
+                $regex_pattern = $this->core->getQueries()->getQueueRegex($queue_code)[0]['regex_pattern'];
+                $regex_pattern = '#' . $regex_pattern . '#';
+                if ($regex_pattern !== '') {
```
sbelsk, line 133: If that queue has no regex pattern configured, then on line 131 `$regex_pattern` is equal to `''`. If on the next line you concatenate the `#` symbol to the beginning and the end of the string, should the if statement not check for `$regex_pattern !== '##'` ? (Or move the concatenation of the pound sign into the if statement) Maybe I'm misunderstanding

site/app/templates/officeHoursQueue/QueueFilter.twig
loadedskiffer: thanks for catching that

site/app/controllers/OfficeHoursQueueController.php
loadedskiffer: thanks for catching that. I just moved the concatenation

bmcutler:
got this twig error:

Screen Shot 2021-07-27 at 10 32 32 AM

bmcutler:
The above error happens when I try to add an "Office Hours Queue Welcome Message (Optional)"
@eddiekrystowski this might be something you worked on?

williamjallen:
I believe that @sbelsk found a way to fix that Twig error in one of her upcoming PRs

williamjallen:
I believe that it was made visible by https://github.com/Submitty/Submitty/pull/6713

eddiekrystowski:
I've pushed a commit to fix that twig error as well as another I found.

bmcutler:
Two modals is much better!

Requests:

Screen Shot 2021-07-27 at 10 38 31 AM

* let's not call it "old queue", let's call it "existing queue"

* please edit the title on the new popup modal "Queue Settings" should be "Edit Existing Queue"

* It would be nice to have just one "select queue" drop down, and have both new access code & new regex work with the same dropdown, but this is ok I guess(?).  If we leave it this way, let's make 2 subsection titles "Change Queue Access Code" and "Change Queue Regex Pattern".  Then the buttons can just be "Submit Change"

sbelsk:
> The above error happens when I try to add an "Office Hours Queue Welcome Message (Optional)"
> @eddiekrystowski this might be something you worked on?

@bmcutler Yes I noticed this too and I am fixing this in https://github.com/Submitty/Submitty/pull/6817. This was not introduced in this PR, but it is due to the wrong markdown template being included at the top of `site/app/templates/officeHoursQueue/QueueHeader.twig`.

williamjallen (approved):

bmcutler (approved):
yay! looks nice :)

PR #6752, opened 2021-07-16
[Feature:Submission] Timed gradeable popup submit 

### LINKED ISSUE #4596: Multiple "Submit" buttons on Notebook gradeables page
A lengthy quiz gradeable (that can be re-submitted multiple times) currently only has a green "Submit" button at the bottom of the page.  If it takes a whille to fill out the form/quiz, the student may lose their work.  

Let's encourage them to save/submit at specific points in the notebook.

Let's add a new option to the instructor config.json that will insert duplicate "Submit" buttons at specific points within the quiz.  All of the buttons do the same thing (save & submit all answers on the page).

It's important that the text on/near these buttons makes it clear that this counts as a submission.

Also, somewhere on the page we should indicate how many submissions the student has used and how many are remaining.

loadedskiffer:
closes #4596 

### What is the new behavior?
For timed gradeables, a pop up will appear in the upper right corner of the screen encouraging students to submit so they can save their progress. The popup has 3 buttons, submit, close and don't remind me again. The popup appears every time a set interval is reached, as of now the interval is either 20 percent of the total allowed time or 10 minutes, whichever is greater. The popup will never appear for gradeables with a time limit of less than 25 minutes. The popup will not appear if there is no new work to submit.

### Other information?
To test:

To manually make a gradeable a timed gradeable, go to the "gradeable" table in the database and set the allowed_minutes column for a notebook gradeable.

NOTE allowed minutes must be over 25 or the popup will never show up

1. Setup a timed notebook gradeable
2. attempt to complete the gradeable and view the popup.
3. for testing purposes, change the interval in notebook.twig to a much shorter amount of time so you don't have to wait 20 minutes to see it.
4. use all of the buttons


sbelsk (commented):
I didn't really know how to test this, I tried it on "Notebook Time Limit" in the development course but I didn't get popups to show up so I removed the twig "if" statements around the responsible code so I can't tell if this works with the right gradeables. Overall this looks very nice! 
I saw that after you click "submit" on the popup, the button gets disabled, but it shortly after gets enabled again before the page reloads and I managed to click it once more and made 2 submissions. Is there a way to keep the button disabled until the page reloads?
And one last thing, I think that popup would look nicer in light mode with a slightly lighter background color, but maybe it's just me :P

site/app/templates/notebook/Notebook.twig
```diff
@@ -262,6 +271,38 @@
...
+            counter++;
+            //interval is 20 percent of overall time allowed or 10 minutes, whichever is greater
+            let interval = ((({{ allowed_minutes * 60 }}) / 5) > 600 ? ({{ allowed_minutes * 60 }}) / 5 : 600);
+            if (counter > interval) {
+                if (sessionStorage.getItem("show_popup_submit") === 'yes') {
+                    $(".notebook-submit-popup").show();
+                    counter = 0;
+                }
+            }
+        }
```
cjreed121, line 292: This is not an accurate way to measure time. setInterval of 1000 ms looks like it will delay exactly 1 second but it is never really exact leading to your recorded time to go out of phase with the actual time. This also will not work if the page is refreshed as it resets counter back to 0. You are assuming that this script stays running the whole exam. Keep in mind the page is refreshed every time a submission is made.

Take a look at my gradeable timer PR or just the submission-page.js file to see how I made a more accurate timer.

cjreed121 (changes requested):

loadedskiffer:
@sbelsk I gave it a slightly lighter color and made it so the submit button stays disabled after clicking it. @cjreed121 The popup now relies on your gradeable timer.

site/app/templates/notebook/Notebook.twig
```diff
@@ -262,6 +271,44 @@
...
+    {% if is_timed and allowed_minutes > 25 %}
+        //interval is 20 percent of overall time allowed or 10 minutes, whichever is greater
+        let interval = Math.floor(((({{ allowed_minutes }}) / 5) > 10 ? ({{ allowed_minutes }}) / 5 : 10));
+        let loadTime = Date.now();
+       
+        function initializePopupTimer(){
+            sessionStorage.setItem("show_popup_submit", "yes");
+            syncWithServer(true);
+            count();
+        }
```
cjreed121, line 284: This function should not call syncWithServer. It was only designed to be called from updateTime. You calling it will make the gradeable timer start up twice and run two instances. If you want to run something after syncWithServer is run you can place it in the initializeTimer in submission-page.js. Or you could assume syncWithServer gets run since it will always be there if is_timed is true.

loadedskiffer: I just removed it, It was unnecessary in the first place.

site/app/templates/notebook/Notebook.twig
```diff
@@ -262,6 +271,43 @@
...
+
+        function count(){
+            const time_taken = Math.floor((curTime - loadTime)/1000);
+            const minutes_taken = Math.floor(time_taken / 60);
+            if (minutes_taken >= interval ) {
+                if (sessionStorage.getItem("show_popup_submit") === 'yes' && curTime < user_deadline && showPopup) {
+                    $(".notebook-submit-popup").show();
+                }
+            }
+            setTimeout(count,100);
```
cjreed121, line 294: This will run every 0.1 seconds. I had this for timer for accuracy but for something like this it could probably run once a second or maybe even every few seconds.

site/app/templates/notebook/Notebook.twig
```diff
@@ -262,6 +271,43 @@
 </div>
 
 <script>
+    //for displaying popup submit button
+    {% if is_timed and allowed_minutes > 25 %}
+        //interval is 20 percent of overall time allowed or 10 minutes, whichever is greater
+        let interval = Math.floor(((({{ allowed_minutes }}) / 5) > 10 ? ({{ allowed_minutes }}) / 5 : 10));
+        let loadTime = Date.now();
```
cjreed121, line 278: Maybe this should be set to curTime as that is the official time from server while Date.now() is client time? You use curTime later on and Date.now() which could be differing values. Of course you would expect them to be different in order to calculate how long it has been but these could differ by pretty large amounts if the client set a different time. You could even get a negative time.

site/app/templates/notebook/Notebook.twig
```diff
@@ -262,6 +271,43 @@
...
+                    $(".notebook-submit-popup").show();
+                }
+            }
+            setTimeout(count,100);
+        }
+
+        function closePopup() {
+            $(".notebook-submit-popup").hide();
+            //restart timer
+            loadTime = Date.now();
```
cjreed121, line 300: Change this to curTime as well if you go with my first comment.

cjreed121 (commented):
Works well but just a few more comments.

site/app/templates/notebook/Notebook.twig
loadedskiffer: when the page first loads curTime isn't set yet so I made it use Date.Now() once and then every time after that it uses curTime

cjreed121: If you keep it this way, it is possible that there is already a x minute gap between Date.now() and curTime. It is possible for you to not start this until curTime is set by integrating the initial function call into submission-page.js. In probably 99/100 cases this won't be a problem. Timezone would not affect this but if someone's time isn't synced then there could be some issues.

cjreed121 (approved):

site/app/templates/notebook/Notebook.twig
```diff
@@ -262,6 +271,41 @@
 </div>
 
 <script>
+    //for displaying popup submit button
+    //interval is 20 percent of overall time allowed or 10 minutes, whichever is greater
+    let interval = Math.floor(((({{ allowed_minutes }}) / 5) > 10 ? ({{ allowed_minutes }}) / 5 : 10));
+    let loadTime = 0;
+    let showPopup = false;
+
+    function initializePopupTimer(){
```
MasterOdin, line 280: When does this function fire? Always on the first time we open the submission page?

I'm trying to understand when the localStorage is getting set, and if it could be replaced with a local variable.

loadedskiffer: yes it fires the first time the user goes to the submission page, I used session storage so that the variable doesn't persist for a different gradeable.

MasterOdin: I'm not sure `sessionStorage` works the way you think it does?

For my understanding, let me know if the following is correct (with all navigation happening in the same page):
* User goes to notebook gradeable A, `initializePopupTimer` fires
* User navigates to a different page
* User goes to notebook gradeable A, `initializePopupTimer` fires
* User goes to a different page
* User goes to notebook gradeable B, `initializePopupTimer` fires

Assuming the above, on _any_ page load of the notebook gradeable, then `sessionStorage.setItem('show_popup_submit', 'yes')` happens resetting any previous value. Given this, I'm saying you can add a new `let showPopup = true;` variable above this function, and just use that to track the state and avoid `sessionStorage`.

loadedskiffer: but then wouldn't showPopup be set to true on every reload even if the user selects "don't show me again"

loadedskiffer: So if someone chooses don't remind me again and then they make a submission, the variable will be reset when the page reloads and the popup will show again

loadedskiffer: Disregard previous comments, I switched it to use local storage so that page reloads don't reset the value and different gradeables can have different values.

cjreed121 (approved):

sbelsk (approved):
Should work after @cjreed121's PR https://github.com/Submitty/Submitty/pull/6857 to fix the permissions on `set_allowed_mins.py` in the Submitty install script.

MasterOdin (changes requested):
You need to have some mechanism that clears out the stuff that's set in localStorage for this as it's not infinite, and could eventually fill up. My guess would be tying the behavior to "Submit" perhaps?

MasterOdin (approved):

PR #6760, opened 2021-07-18
[Testing:CourseMaterials] Skip flaky test on CI

shailpatels:

Currently the last test in course materials is flaky and attempts to fix it have not worked

### What is the new behavior?
The test will be skipped on CI but can be run locally 

The diff looks like a lot of changes because the test was wrapped in an if statement


MasterOdin:
Probably better to add https://www.npmjs.com/package/@cypress/skip-test and use that to skip?

PR #6764, opened 2021-07-19
[Bugfix:Developer] Fix generation of sample team IDs

cjreed121:

Right now, in the sample course, teams are generated with an incremental number ID per gradeable. There is a check in the setup_sample_courses script to make sure no duplicated IDs. This still produces a different set of IDs compared to what production would generate where the number ID is incremental across the entire course.

image

### What is the new behavior?
The setup script uses a similar method to the site code to generate the numeric IDs to model a production environment better. The numeric IDs are now always unique per course.

### Other information?
Below is a Vagrant Up workflow run on this branch to show that it works on a full vagrant up.
https://github.com/Submitty/Submitty/actions/runs/1045366739

### DIFF
--- a/.setup/bin/setup_sample_courses.py
+++ b/.setup/bin/setup_sample_courses.py
@@ -37,7 +37,7 @@ from tempfile import TemporaryDirectory
 from submitty_utils import dateutils
 
 from ruamel.yaml import YAML
-from sqlalchemy import create_engine, Table, MetaData, bindparam, select, join
+from sqlalchemy import create_engine, Table, MetaData, bindparam, select, join, func
 
 yaml = YAML(typ='safe')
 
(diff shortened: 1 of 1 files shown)


williamjallen (commented):
Would it be good to have a few missing teams to simulate users deleting their teams?  Otherwise looks good.

cjreed121:
> Would it be good to have a few missing teams to simulate users deleting their teams? Otherwise looks good.

I believe one of the gradeables has a few people not in a team. I don't necessarily think it needs to be simulated any more than it currently is.

williamjallen (approved):

bmcutler:
@cjreed121 and @shailpatels Please take a look at this failing course materials test.
(I reran it and it still failed)

PR #6770, opened 2021-07-19
[UI/UX:Autograding] Revise docker UI

### LINKED ISSUE #6697: Add functionality and revise design of "Docker UI"
Users of type "faculty" have access to the "Docker UI" interface from the sidebar.  (original issue #2426 and PR #5213)
Currently the UI allows you to view-only information about docker images available on the system, but the interface is not usefully organized.

Goals:

1. Re-organize the interface to be more useful.
2. Allow faculty users to add a new docker container image to the system.

------

The relevant information about the autograding worker machines and containers are stored in:
/usr/local/submitty/config/autograding_workers.json
/usr/local/submitty/config/autograding_containers.json 

See also:
https://submitty.org/developer/automated_grading
https://submitty.org/sysadmin/worker_installation

Each worker has one or more 'capabilities' and to satisfy each capability a machine must have one or more docker images installed.  

The current layout of the docker UI interface with 'cards' is does not scale well when we have multiple worker machines with slightly different and overlapping sets of capabilities and images.

Currently the sysadmin must edit these 2 files and then run 
/usr/local/submitty/.setup/INSTALL_SUBMITTY.sh to install new docker images and stop & start all remote workers.  Faculty users cannot add new images -- this is a inconvenience/block when developing autograding for a course gradeable.

--------

Requests:

1. At the top of the page list all of the workers and all of the capabilities and associations (it is not necessarily one-to-one).  Perhaps this can be a table with capabilities on one axis and workers on the other axis -- with checkboxes in each square if they are associated.

2. List all of the images installed on each worker machine.  Note: Currently the code looks at both the autograding_containers.json file and also runs a command on the machine to detect if the images are actually installed.  This command does not run on the remote workers so the information could be incomplete/incorrect.  For now let's just focus on what's in the file and assume that the installation command was successfully run.

3. To reduce clutter, perhaps let the user select a single worker machine or single capability and hide the images installed for other workers/capabilities.

4. Allow the user write access to ADD IMAGES to the list for a capability.  This will save the information to the autograding_containers.json file AND make a call to the system script to install new docker images on the worker machines as needed.

Security notes:

5. This should only be available to users with faculty or superuser access!

6. We should also verify that the script specified for the image is from a trusted dockerhub account.  We'll have to add a list of trusted dockerhub accounts on the system.  E.g., "submitty/", "submittyrpi/". "instructoraccount/", etc.

HenrikLam:

Closes #6697

### What is the new behavior?
- Lists out each worker machine's capabilities, showing if the worker machine is disabled in the autograding_workers.json.
- Lists out each image, its detail information, and its association with capabilities, allowing user to filter the images shown by the capabilities. Also shows if something went wrong with installing/updating an image.
- Allows user to add an image to a capability on the UI, subsequently triggering the system to update the docker images on each worker machines.
- Allows user to call an update on the docker images with a button on the UI in case there is an update or the previous update failed.

### Other information?
To test the adding image functionality and refreshing the docker images, you will need
1. Run `sudo systemctl stop submitty_daemon_jobs_handler` in submitty vm
2. Do `ps -ef | grep submitty_daemon`
3. Do `kill -9 p_id1 p_id2 ...` on all the processes listed by `ps -ef` except for the grep process
4. Run `sudo su -c '/usr/local/submitty/sbin/submitty_daemon_jobs/submitty_daemon_jobs.py' submitty_daemon`


site/public/js/docker_interface.js
```diff
@@ -17,3 +17,95 @@ function collapseSection(id,btn_id) {
...
+        data: {
+            'capability': capability,
+            'image': image,
+            csrf_token: csrfToken
+        },
+        success: function(data) {
+            console.log(data);
+            location.reload();
+        },
+        error: function(err) {
```
shailpatels, line 83: This error field will only ever get called if the ajax call throws a javascript error and not if the server responds with a 500 error or does not succeed. 

If given a 500 error the json response will become a frog robot and cannot be parsed, otherwise you'll get a standard response : https://submitty.org/developer/json_responses

shailpatels (changes requested):

site/app/controllers/DockerInterfaceController.php
```diff
@@ -75,4 +80,122 @@ public function showDockerInterface(): MultiResponse {
...
+
+        // check for proper format
+        $match = preg_match('/^[a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+\/[a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+:[a-zA-Z0-9][a-zA-Z0-9._-]{0,127}$/', $_POST['image']);
+        if ($match == 0) {
+            $this->core->addErrorMessage("Improper docker image name");
+            return JsonReponse::getErrorResponse("Improper docker image name");
+        }
+        $image_arr = explode(":", $_POST['image']);
+        // ping the dockerhub API to check if docker exists
+        $url = "https://registry.hub.docker.com/v1/repositories/".$image_arr[0]."/tags";
```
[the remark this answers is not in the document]
HenrikLam: When I tried to use file_get_contents, the function call returns false. Do you know why this might be?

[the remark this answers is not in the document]
shailpatels: It would depend on what type of request the docker API endpoint is expecting, so it might not work if it's not a plain get request when a resource pointing to the data you want like for example it is expecting json or wants a specific header set

That said I can't find more about the API endpoint since it looks like v1 is deprecated? If this lookup is being added now should the v2 API be used instead

site/app/controllers/DockerInterfaceController.php
```diff
@@ -75,4 +80,122 @@ public function showDockerInterface(): MultiResponse {
...
+     * @return bool
+     */
+    private function updateDocker() {
+        $now = $this->core->getDateTimeNow()->format('Ymd');
+        $docker_job_file = "/var/local/submitty/daemon_job_queue/docker" . $now . ".json";
+        $docker_data = [
+            "job" => "UpdateDockerImages"
+        ];
+
+        if (
```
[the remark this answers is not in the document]
HenrikLam: This was copied from other places on the site that tries to put a daemon job on the queue. `is_writable` will be true only if  `file_exists` is true, however, can be false also for other reasons like permission error. In the case of these reasons, the if statement will be short circuited and the code will not attempt to write to the file, which I think is what you were describing unless I am misunderstanding.

site/app/templates/admin/Docker.twig
```diff
@@ -5,79 +5,143 @@
     {% if docker_info is not null %}
         {# TODO: include worker information here in the future #}
         <p>Docker Version : {{ docker_info['ServerVersion'] }} </p>
+        <p>Updated on: {{last_updated}}</p>
+        {% if error_logs | length > 0 %}
+            <textarea disabled="disabled" style="resize: none; width: 100%">
```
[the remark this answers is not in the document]
HenrikLam: It will always be disabled, however, I wanted a way to highlight the information being displayed.

site/app/views/admin/DockerView.php
```diff
@@ -75,12 +81,107 @@ public function displayDockerPage(array $docker_data) {
...
+        $fail_images = [];
+        $error_logs = [];
+        
+        if (count($array_list) > 2) {
+            $last_ran = "sometime";
+            $most_recent = max($array_list);
+            $fd = fopen("/var/local/submitty/logs/docker/" . $most_recent, "r");
+            while (($buffer = fgets($fd)) !== false) {
+                $match = [];
+                if (preg_match("/^\[Last ran on: ([0-9 :-]{19})\]/", $buffer, $match)) {
```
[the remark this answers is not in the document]
HenrikLam: Not entirely sure how to handle errors here. For now I made the code break the loop if there's a `preg_error` and display that there was an error while parsing a file. Please let me know if you have a better idea.

site/app/controllers/DockerInterfaceController.php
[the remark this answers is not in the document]
HenrikLam: The problem with the v2 API is that it requires authentication to be used, after talking about this with Barb, we decided that we will use v1 for now.

site/app/controllers/DockerInterfaceController.php
```diff
@@ -75,4 +80,122 @@ public function showDockerInterface(): MultiResponse {
...
+    /**
+     * @Route("/admin/add_image", methods={"POST"})
+     * @return JsonResponse
+     */
+    public function addImage(): JsonResponse {
+        $user = $this->core->getUser();
+        if (is_null($user) || !$user->accessFaculty()) {
+            return new MultiResponse(
+                JsonResponse::getFailResponse("You don't have access to this endpoint."),
+                new WebResponse("Error", "errorPage", "You don't have access to this page.")
```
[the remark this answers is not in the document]
HenrikLam: Just to confirm, you are talking about the first parameter for `new WebResponse`, meaning the `"Error"`?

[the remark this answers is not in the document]
shailpatels: Yeah, "Error" is pointing to a class in the views directory

shailpatels (changes requested):
I'm getting the following PHP errors after checking out the branch and running a `submitty_install`

```
( ! )  Warning:  file_get_contents(/usr/local/submitty/config/autograding_workers.json):  failed to open stream: Permission denied in  /usr/local/submitty/site/app/libraries/FileUtils.php on line 272
--
1 | 0.0005 | 429280 | {main}(  ) | .../index.php:0
2 | 0.0663 | 12631992 | app\libraries\routers\WebRouter::getWebResponse(  ) | .../index.php:129
3 | 0.2459 | 16766952 | app\libraries\routers\WebRouter->run(  ) | .../WebRouter.php:196
4 | 0.2459 | 16767496 | call_user_func_array:{/usr/local/submitty/site/app/libraries/routers/WebRouter.php:218} (  ) | .../WebRouter.php:218
5 | 0.2459 | 16767504 | app\controllers\DockerInterfaceController->showDockerInterface(  ) | .../WebRouter.php:218
6 | 0.5001 | 16805224 | app\libraries\FileUtils::readJsonFile(  ) | .../DockerInterfaceController.php:71
7 | 0.5001 | 16805224 | file_get_contents (  ) | .../FileUtils.php:272
```

```
Warning: Invalid argument supplied for foreach() in /usr/local/submitty/site/app/views/admin/DockerView.php on line 88
--
1 | 0.0005 | 429280 | {main}(  ) | .../index.php:0
2 | 0.5006 | 16797184 | app\libraries\response\MultiResponse->render(  ) | .../index.php:140
3 | 0.5006 | 16797184 | app\libraries\response\WebResponse->render(  ) | .../MultiResponse.php:107
4 | 0.5006 | 16797184 | app\libraries\Output->renderOutput(  ) | .../WebResponse.php:43
5 | 0.5006 | 16797560 | app\libraries\Output->renderTemplate(  ) | .../Output.php:190
6 | 0.5009 | 16833736 | call_user_func_array:{/usr/local/submitty/site/app/libraries/Output.php:221} (  ) | .../Output.php:221
7 | 0.5009 | 16833752 | app\views\admin\DockerView->displayDockerPage(  ) | .../Output.php:221
```

shailpatels:
Here is a `stat` on that file on my VM
```
  File: /usr/local/submitty/config/autograding_workers.json
  Size: 293       	Blocks: 8          IO Block: 4096   regular file
Device: fd00h/64768d	Inode: 933139      Links: 1
Access: (0460/-r--rw----)  Uid: (    0/    root)   Gid: ( 1006/submitty_daemon)
Access: 2021-07-21 20:54:45.808036098 -0400
Modify: 2021-07-01 20:04:56.602925784 -0400
Change: 2021-07-01 20:04:56.602925784 -0400
 Birth: -
```

site/app/templates/admin/Docker.twig
```diff
@@ -5,79 +5,144 @@
...
-        <p class="black-message">These are the docker images that are installed and listed under autograding_containers.json</p>
-
-        <div class="docker-images-container">
-            {{ _self.renderDockerImage(autograding_containers['found']) }}
-        </div>
-    </div>
+    <h2 class="docker-header">Worker Machines</h2>
+    <table>
+        <tr>
+            <th></th>
```
shailpatels, line 24: What is the reason for this empty header?

site/app/controllers/DockerInterfaceController.php
```diff
@@ -75,4 +80,122 @@ public function showDockerInterface(): MultiResponse {
...
+            );
+        }
+        
+        if (!isset($_POST['image'])) {
+            $this->core->addErrorMessage("Image not set");
+            return JsonReponse::getErrorResponse("Image not set");
+        }
+        if (!isset($_POST['capability'])) {
+            $this->core->addErrorMessage("Capability not set");
+            return JsonReponse::getErrorResponse("Capability not set");
```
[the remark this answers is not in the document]
HenrikLam: Sorry, I am not sure what you mean by that.

sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
```diff
@@ -355,3 +355,24 @@ def run_job(self):
             subprocess.run(["sudo", "/usr/local/submitty/sbin/adduser_course.py", head_instructor, semester, course], stdout=output_file, stderr=output_file)
             if VERIFIED_ADMIN_USER != "":
                 subprocess.run(["sudo", "/usr/local/submitty/sbin/adduser_course.py", VERIFIED_ADMIN_USER, semester, course], stdout=output_file, stderr=output_file)
+
+
+class UpdateDockerImages(AbstractJob):
+    def run_job(self):
+        today = datetime.datetime.now()
+        log_path = os.path.join(DATA_DIR, "logs", "docker")
+        try:
```
shailpatels, line 364: Should this job trying to make the `logs` dir, I think most jobs assume they get created ahead of time in the install scripts

site/public/js/docker_interface.js
```diff
@@ -17,3 +17,114 @@ function collapseSection(id,btn_id) {
...
+        },
+        success: function(data) {
+            const json = JSON.parse(data);
+            if (json.status == 'success') {
+                localStorage.removeItem('capability');
+            }
+            else {
+                window.alert('Something went wrong. Please try again.');
+            }
+            location.reload();
```
shailpatels, line 93: I think having a location.reload at the end of a ajax call defeats the purpose of async JS, this could just be a native HTML form pointing to the addImage endpoint

shailpatels (changes requested):

shailpatels (changes requested):
image

For the table, I would recommend dropping the "Created Date" before each date since that information is included in the header. I think you can also drop the virtual size since its taking up too much room in a single cell. You can either move it to its own column or remove it entirely, I don't think anyone cared about the memory distinction

site/app/views/admin/DockerView.php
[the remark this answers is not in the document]
shailpatels: That should be fine, either displaying or logging it would work

site/app/templates/admin/Docker.twig
HenrikLam: This is an empty cell in the table, see screenshot. Just a personal preference in leaving it blank since that column seems intuitive, but I can also make it say Machine instead.
image

shailpatels (changes requested):
You will need to add HTML accessibility labels to the form controls for adding a capilbity,
also I think the button should say "Submit" and not "Send" since that's the general language across submitty

site/public/js/docker_interface.js
HenrikLam: The `location.reload` was left there to show the messages properly. For some reason the success/error messages doesn't show up properly for me on the site without reloading the page.

HenrikLam (commented):
> Here is a `stat` on that file on my VM
> 
> ```
>   File: /usr/local/submitty/config/autograding_workers.json
>   Size: 293       	Blocks: 8          IO Block: 4096   regular file
> Device: fd00h/64768d	Inode: 933139      Links: 1
> Access: (0460/-r--rw----)  Uid: (    0/    root)   Gid: ( 1006/submitty_daemon)
> Access: 2021-07-21 20:54:45.808036098 -0400
> Modify: 2021-07-01 20:04:56.602925784 -0400
> Change: 2021-07-01 20:04:56.602925784 -0400
>  Birth: -
> ```

Weird... It looks the stat looks the same as my vm but the page seems to work fine for me.

site/public/js/docker_interface.js
shailpatels: If the only reason is to display those messages and not update state then you can just pass the information back in the `JsonResponse` and use the JS `displaySuccessMessage` and `displayErrorMessage` to avoid the refresh.

The addErrorMessage on the PHP side will queue the message for the next page the user visits, its by design since the typical use would be to send a form request which would cause a page reload anyways and the user would see the messages then.

site/app/templates/admin/Docker.twig
shailpatels: I would suggest adding a table header for each column if it contains data, it may cause issues for screen readers otherwise or be less intuitive

HenrikLam:
> > Here is a `stat` on that file on my VM
> > ```
> >   File: /usr/local/submitty/config/autograding_workers.json
> >   Size: 293       	Blocks: 8          IO Block: 4096   regular file
> > Device: fd00h/64768d	Inode: 933139      Links: 1
> > Access: (0460/-r--rw----)  Uid: (    0/    root)   Gid: ( 1006/submitty_daemon)
> > Access: 2021-07-21 20:54:45.808036098 -0400
> > Modify: 2021-07-01 20:04:56.602925784 -0400
> > Change: 2021-07-01 20:04:56.602925784 -0400
> >  Birth: -
> > ```
> 
> Weird... It looks the stat looks the same as my vm but the page seems to work fine for me.

Turns out my read write permission was different in that mine was 660... I attempted to fix it by adding a line to `CONFIGURE_SUBMITTY.py`, is this only run during the creation of the vm?

HenrikLam:
> image
> 
> For the table, I would recommend dropping the "Created Date" before each date since that information is included in the header. I think you can also drop the virtual size since its taking up too much room in a single cell. You can either move it to its own column or remove it entirely, I don't think anyone cared about the memory distinction

After talking about this in the meeting, we've decided to remove the size column but keep the date created column as a way of distinguishing images in case of poor image naming.

shailpatels:
> After talking about this in the meeting, we've decided to remove the size column but keep the date created column as a way of distinguishing images in case of poor image naming.

Sorry I should have been more clear, I meant removing the "Created On: <Date>" that is in every row of the "Created Date" column. Since the header already explains what type of data is there, the table cells should only contain data, not additional titles

shailpatels:
The permissions error still exist for me, I think your migration fixes the permissions, but the file is owned by `submitty_daemon` which php doesn't have access to. I think you need to change the group to either 
`submitty_daemonphp` or `submitty_php`

HenrikLam:
@shailpatels I think the permission error should be fixed now! Let me know if it works for you.

eddiekrystowski (approved):
Works well, the page recognizes worker machines

site/app/templates/admin/Docker.twig
```diff
@@ -5,79 +5,141 @@
...
+                    <li>{{ name }}</li>
+                    {% endfor %}
+                </ul>
+                {% endif %}
+                </td>
+                <td> {{ value["created"] }} </td>
+                <td>
+                    {% if value["name"] ~ ':' ~ value["tag"] in image_to_capability|keys %}
+                        {% for capability in image_to_capability[value["name"] ~':'~ value["tag"]] %}
+                            <span class="badge" style="background-color:{{capability_to_color[capability]}}">{{capability}}</span>
```
shailpatels, line 92: The background colors with white font is causing WAVE contrast issues, the background should be darker. The ones I saw have issues are:

- #c3a2d2 (purple)
- #99B270 (yello-green)
- #6bb88f (green)
- #C8938D (brown)
- #cd98aa (pink)

site/app/templates/admin/Docker.twig
```diff
@@ -5,79 +5,141 @@
...
-    <div id="autograding-images" style="display:block;">
-        <p class="black-message">These are the docker images that are installed and listed under autograding_containers.json</p>
-
-        <div class="docker-images-container">
-            {{ _self.renderDockerImage(autograding_containers['found']) }}
-        </div>
-    </div>
+    <h2 class="docker-header">Worker Machines</h2>
+    <table class="machine-table">
+        <tr>
```
shailpatels, line 23: Ok I didn't look closely enough at the image of the table you pasted last review, and am changing my mind on adding that extra header - since the content to the right of the "Machine" header is actually data, not headers themselves its confusing that there is an actual column header on the same row.

I'm now thinking something like this better expresses the data you're trying to visualize

image

```html
<table class="machine-table">
    <tbody>
        <tr>
            <th>Machine</th>
            <th colspan="6" style="text-align:center;">Capability</th>
        </tr>
        <tr>
            <th></th>           
            <th>cpp</th>
            <th>default</th>
            <th>et-cetera</th>
       </tr>
```

And you can get the number of cols from twig and maybe the headers don't need to bold for the capibilities? Might look better 

I realize this PR has been open for a while however so this is lower priority can be ignored if there's not enough time or it can be discussed in the next meeting with @bmcutler and others.

site/app/templates/admin/Docker.twig
```diff
@@ -5,79 +5,141 @@
     {% if docker_info is not null %}
         {# TODO: include worker information here in the future #}
         <p>Docker Version : {{ docker_info['ServerVersion'] }} </p>
+        <p>Updated on: {{last_updated}}</p>
+        {% if error_logs | length > 0 %}
+            <textarea disabled="disabled" style="resize: none; width: 100%; height:25vh;">
```
shailpatels, line 10: I think you should put a title above the log display indicating these are logs and not asssociated with the above system information. 

Also HTML should not be used to visually style content, since the output will always be static a textarea is the wrong semantic element to use. You can use a `<pre>` tag which indicates pre-formatted text and will preserve any spacing from the logs and have a similar visual to a textarea's default font.

shailpatels (changes requested):
Last few requested changes are all accessibility related, if there is an urgency to get this PR merged, then I approve and a seperate issue can be made to get those fixes in later

site/app/templates/admin/Docker.twig
HenrikLam: Because the colors used for the badges are fixed for both light and dark mode, I had to change the table to be fixed in color scheme as well. This means I won't be using the variables specified in `colors.css`, is this an acceptable solution?

site/app/templates/admin/Docker.twig
HenrikLam: This is what I have changed it to, does this look okay?
image

site/app/templates/admin/Docker.twig
HenrikLam: I have made some changes to this area and this is what it looks like now, how does it look?
image

site/app/templates/admin/Docker.twig
shailpatels: @HenrikLam  I would suggest moving the num autograing works column next to the machine column and then having the capabilities to the right so the table looks more balanced but other then that I think it looks really good now

site/app/templates/admin/Docker.twig
shailpatels: Yeah I think thats fine, these badges seem unique to this page so if we need to reuse them in the future we work on getting them back into colors.css but for now it should be fine

HenrikLam: Actually... while the badges are unique to this page, the colors are copied from office hours queue, would that be cause for a small refactor?

site/app/templates/admin/Docker.twig
HenrikLam: The columns have been swapped!
image

shailpatels: Thanks for the quick update!

site/app/templates/admin/Docker.twig
shailpatels: I would say thats up to you, since you're generating the list of colors from php the most you could use the css `var(--your-color)` syntax and then move the hex codes to colors.css

Since it sounds like that touches office hours I think it should be in its own PR or you can make an issue for it

shailpatels (approved):

migration/migrator/migrations/system/20210722131039_change_autograding_workers_json_permission.py
```diff
@@ -0,0 +1,26 @@
...
+def up(config):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    workers_json = str(Path(config.submitty['submitty_install_dir'], 'config', 'autograding_workers.json'))
+    os.chmod(workers_json, 0o660)
+    os.system("chown submitty_php {}".format(workers_json))
```
MasterOdin, line 15: This chown looks like it'll be overwritten by `INSTALL_SUBMITTY_HELPER.sh` doing `shutil.chown(full_file_name, 'root',DAEMON_GID)`?

site/public/js/docker_interface.js
```diff
@@ -17,3 +17,122 @@ function collapseSection(id,btn_id) {
...
+function filterOnClick() {
+    const this_filter = $(this).data('capability');
+
+    $('.filter-buttons').each(function (){
+        $(this).addClass('fully-transparent');
+    });
+
+    if ($(this).hasClass('fully-transparent')) {
+        $(this).removeClass('fully-transparent');
+    }
```
MasterOdin, line 30: Do we need the `if` here, or can we just always run `$(this).removeClass('fully-transparent');` given that the result is that all elements that enter this function will not have `full-transparent` class.

site/public/js/docker_interface.js
```diff
@@ -17,3 +17,122 @@ function collapseSection(id,btn_id) {
...
+            console.error(err);
+            window.alert('Something went wrong. Please try again.');
+        },
+    });
+}
+
+$(document).ready(() => {
+    $('.filter-buttons').on('click', filterOnClick);
+    $('#show-all').on('click', showAll);
+    $('#add-field').val(localStorage.getItem('capability'));
```
MasterOdin, line 135: Why do we need the `localStorage` here? It looks like the only way to leave the page is the instructor themselves navigating away, at which point it feels like this `localStorage` would become stale.

migration/migrator/migrations/system/20210722131039_change_autograding_workers_json_permission.py
```diff
@@ -0,0 +1,26 @@
+"""Migration for the Submitty system."""
```
MasterOdin, line 1: Is there every a case where we would run this migration, but not the chmod / chown on this file in `INSTALL_SUBMITTY_HELPER`? If not, then the migration can be removed and just let the `INSTALL_SUBMITTY_HELPER` handle the permissions.

MasterOdin (changes requested):

migration/migrator/migrations/system/20210722131039_change_autograding_workers_json_permission.py
HenrikLam: I have moved it to CONFIGURE_SUBMITTY.py

HenrikLam:
I edited the CONFIGURE_SUBMITTY.py for the necessary permissions on the json files, I tested this by calling `python3 ${SUBMITTY_REPOSITORY}/.setup/CONFIGURE_SUBMITTY.py --debug --setup-for-sample-courses` in my local vm and the permissions seem to be correct for me.

migration/migrator/migrations/system/20210722131039_change_autograding_workers_json_permission.py
HenrikLam: Barb said she would see where the permission change should happen so we shouldn't merge this yet.

bmcutler (approved):
Looks good for initial testing on the production machine.

Minor note: It looks like the color scheme was only tested in dark mode?
Light mode doesn't match the table schemes on our other pages:

Screen Shot 2021-08-04 at 11 48 20 PM

PR #6771, opened 2021-07-19
[Bugfix:TAGrading] PDF Repair

### LINKED ISSUE #6699: Annotating PDFs doesn't work with typing text

### LINKED ISSUE #6700: Annotations disappear from the PDF when atempting to move them
**Describe the bug**
When attempting to annotate a student's PDF, typing text is either not working a tall or allows the text to be types but then all text disappears, depending on the browser. Clicking "Save" button does not restore lost annotations.

**Expected behavior**
Annotations that were moved around the PDFs should remain visible on the page and saved when the "Save" button is clicked.

**To Reproduce**
1. Open a student's PDF (I was grading article01 gradeable from CSCI-4963).
2. Add some annotations, e.g., underlining some text. See screenshot 1.
3. Click on an arrow button and then attempt to drag the underlining to another part of the PDF. Release the mouse button and observe the underlining disappear. In other words, the arrow erases the annotation instead of moving it. See screenshot 2.
4. For some other annotation, the effect was slightly different. Click on an arrow button and then attempt to drag the underlining to another part of the PDF. The annotation would not move but a small blue highlighted artifact appears on the PDF (see screenshot 3 the artifact is circled with red pen).

**Configuration**
 - Browser: Google Chrome Version 91.0.4472.124 (Official Build) (64-bit)

**Screenshots**
image
image
image

**Additional context**
I was not able to understand what accounts for the difference between the behavior described in 3 and 4. Maybe it is the zoom level? Overall, the PDF annotation feature feels very clunky, hard to use, and unreliable. I had several cases when it actually didn't save my annotations despite me clicking the Save button and Submitty saying that everything was saved. It is a highly wanted feature, I just wish it was debugged better, so that it is more intuitive to use and more reliable too.

Another thought that maybe if it is hard to get the annotation feature to work correctly inside Submitty, it would be possible to allow graders use some external tool (like Acrobat) and then allow them to replace the student's PDF with its annotated version in the submission? I'm just trying to see what can be done so that I could grade a bunch of student PDFs quickly and hopefully with little effort.

eddiekrystowski:
Fixes #6699 
Fixes #6700 
Currently, various actions on the PDF annotation page can cause PDFs to get into a broken state where either the PDF won't load, or weird behavior occurs such as annotations disappearing.
Also, there is a bug where Submitty is giving graders the student view of the pdf, causing their grading to not save.

### What is the new behavior?
Bugfixes:
* Submitty gives the correct version of the pdf view, so graders' annotations are properly saved.
* If Submitty detects any faulty annotations from the current user upon loading, it will attempt give a prompt to the user asking if they would like to try and repair them.
  * If they accept, Submitty will try to reset properties with null values to a reasonable default value.
  * If no default value is found, the annotation will be deleted.
  * Submitty will save the repaired annotations and refresh the page.
 
image
* If the user declines to do a repair, a button will be added to allow them to do it again at any time, and a red danger message will show above the pdf

image

 * If Submitty detects any faulty annotations from NOT the current user, it will alert the current user and temporarily fix them so that the user can make their own annotations.
   * In order to actually repair these annotations, the user who created them must visit the page and agree to repair them. (see above)
 
image

Other Improvements: 
* Pen color/size selector menus will close when a click occurs outside of them.
* More checking has been done to help prevent faulty annotations from being created
* Improved speed of pdf toolbar actions
* Allow custom setting of zoom level
  * Zoom level is remembered across different students
* Added rotate left button
* Added Clear button 
  * Only clears your annotations (saves and refreshes page)
* Added button to hide/show other users' annotations
* Allows graders to specify a custom color from a color picker

### Testing
The easiest way to test is to modify the annotations in localStorage.
1. Log in (I will use 'instructor')
2. Go to pdf annotation page.
3. Make a text annotation
4. Open console
5. Enter `let annotations = JSON.parse(localStorage.getItem('<PDF NAME>.pdf/<CURRENT USER>/annotations'))`
6. Make a modification to the annotations array to set a property on an annotation to null. For example if there is only a single text annotation, `a[0].size = null`, or `a[0].content = null`
7. Save the annotations to storage by doing 
`localStorage.setItem('<PDF NAME>.pdf/<CURRENT USER>/annotations', JSON.stringify(annotations))`
8. Press the save button to save the faulty annotations to the server.
9. Refresh the page to see repair popup


loadedskiffer (changes requested):
Seems to work fine but a few small comments,
1. I don't think it's necessary to have +/- buttons and up/down arrows to change the zoom level, one should be removed.
2. The save button seems to stick when clicked and I couldn't really tell if it was working at first.
3. The clear button works as intended but I don't like how it closes the pdf when you click it, is there any way to make it reopen the pdf after the refresh ?

eddiekrystowski:
The reason I left the +/- buttons is because this page needs to be easily accessible to users on mobile devices, and the up/down arrows only show up when hovering on the zoom level which mobile users can't do.

Good point about the button, the save operation can take a while sometimes. I made it change colors when clicked / when new changes are added to make it more responsive.

Also, I thought I had to refresh the page after clearing, but it turns out the annotations keep track of which user they belong to, so I just removed the refresh.

Thanks for the feedback!

loadedskiffer (changes requested):
New changes look good but I think I found a couple bugs with changing the color.

1. Sometimes when you change the selected color and then try to draw, the color is black instead of the chosen color.
2. When choosing a custom color, the annotations appear in the correct color but the little box that indicates what color you are using displays the wrong color.

loadedskiffer:
Did you get rid of the custom color picker?

eddiekrystowski:
No, maybe the Twig caching is acting strangely. Maybe try a hard refresh or a clean install?

loadedskiffer (approved):

PR #6772, opened 2021-07-20
[Bugfix:InstructorUI] Clarify late days allowed

williamjallen:
Issue https://github.com/Submitty/Submitty/issues/6377 describes an instructor's confusion with the late days allowed page.  

### What is the new behavior?
The text on the page was edited to clarify that the number of late days provided will be the new number of late days, not an additional number of late days to add.  To make this behavior more obvious, and remove the need to look up the number of existing late days for a user, this PR also loads the existing number of late days as the initial value on the late days field.

https://user-images.githubusercontent.com/16820599/126254531-b5a9e841-5a8d-4423-b3ba-19f2a6c7e373.mov

### Other information?
Fixes https://github.com/Submitty/Submitty/issues/6377


site/app/templates/admin/LateDays.twig
```diff
@@ -18,8 +18,8 @@
                 <input class="option-input" type="text" name="datestamp" id="datestamp">
```
shailpatels, line 18: I would suggest changing "Datestamp" to something more clear if an instructor doesn't read the line at the top, something "Time when to grant/activate latedays"

site/app/templates/admin/LateDays.twig
```diff
@@ -81,10 +81,26 @@
     {% endif %}
 </div>
 <script>
+    const students_and_late_days = JSON.parse(`{
+    {% for user in users %}
+        "{{ user.getId() }}":"{{ user.getAllowedLateDays() }}",
+    {% endfor %}
+    "":""
```
shailpatels, line 88: Is `"":""` used for anything specific? If there are no users in the latedays granted list 
JSON.parse("{}") is  a valid default

shailpatels (changes requested):

site/app/templates/admin/LateDays.twig
williamjallen: `JSON.parse()` doesn't like dangling commas for some reason.  Since the previous statement adds a comma after each value, I added a placeholder at the end which does not have the trailing comma.

site/app/templates/admin/LateDays.twig
williamjallen: I agree. Does it look more clear to you now?

shailpatels (approved):

bmcutler (changes requested):
I'm confused what late day value is being initialized into the box and why.

I've set the default number of days for the term to be '5' (on course settings)
And entered the data below.  It always sorts with the most recent date at the top (maybe it would make more sense to sort in the other direction).  Note: it doesn't seem to matter what order the data is entered (that's good).

However, right now it always defaults to '10', the oldest date, even though today is the 22nd and the current total number of days available to this student as of today is '3'.

Perhaps it should look at the date typed into the second box and grab the total number on that day?

Or maybe just remove this attempt to put a value in the box, since its confusing and complicated.

Screen Shot 2021-07-22 at 10 46 16 AM

williamjallen:
Ohh, I see.  I didn't think about a person having late days awarded on different days so I was just trying to load the late days for a person.  I'll change it.

shailpatels (approved):
Reapproving, html looks good and removing the JS does simplifiy things although it was my assumption that the conversastion in #6377 mentioned loading the number of late days a student was granted was part of the requested design

bmcutler (changes requested):
Looking good. 
Let's add a little bit more documentation on this page.

After this sentence:
"The default number of late days available to each student at the start of term is configured on the Course Settings page."
Let's display the value for that setting...  
"The initial allowed late days for each student in this course is currently set to X."

Then add a new paragraph.
Thanks :)

bmcutler (approved):
@williamjallen 
Please fix the failing cypress test.

PR #6774, opened 2021-07-20
[Dependency] Remove Java from primary installation

eddiekrystowski:
Java 8 is included in `setup_distro.sh` and the images are included in `CONFIGURE_SUBMITTY.py`. All grading being done with java is in containers so it doesn't need to be included in the base setup.

### What is the new behavior?
Java 8 is removed from `setup_distro.sh` and `CONFIGURE_SUBMITTY.py`

### DIFF
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -481,8 +481,6 @@ if not args.worker:
             "default": [
                           "submitty/clang:6.0",
                           "submitty/autograding-default:latest",
-                          "submitty/java:8",
-                          "submitty/java:11",
                           "submitty/python:3.6",
                           "submittyrpi/csci1200:default"
                        ]
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -481,6 +481,7 @@ if not args.worker:
             "default": [
                           "submitty/clang:6.0",
                           "submitty/autograding-default:latest",
+                          "submitty/java:11",
                           "submitty/python:3.6",
                           "submittyrpi/csci1200:default"
                        ]
--- a/.setup/distro_setup/ubuntu/18.04/setup_distro.sh
+++ b/.setup/distro_setup/ubuntu/18.04/setup_distro.sh
@@ -82,11 +82,6 @@ apt-get install -qqy cmake
 # for Lichen (Plagiarism Detection)
 apt-get install -qqy python-clang-6.0
 
-# Install OpenJDK8 Non-Interactively
-echo "installing java8"
-apt-get install -qqy openjdk-8-jdk
-update-java-alternatives --set java-1.8.0-openjdk-amd64
-
 # Install Image Magick for image comparison, etc.
 apt-get install -qqy imagemagick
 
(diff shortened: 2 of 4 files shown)


eddiekrystowski:
@MasterOdin I think you had mentioned that I would just need to remove java from the distro setup, but I'm not sure if I did it correctly or if there is anything else I have to do, so it'd be great if you could review!

.setup/distro_setup/ubuntu/18.04/setup_distro.sh
```diff
@@ -82,11 +82,6 @@ apt-get install -qqy cmake
 # for Lichen (Plagiarism Detection)
 apt-get install -qqy python-clang-6.0
 
-# Install OpenJDK8 Non-Interactively
-echo "installing java8"
-apt-get install -qqy openjdk-8-jdk
-update-java-alternatives --set java-1.8.0-openjdk-amd64
```
MasterOdin, line 88: Let's leave Java for 16.04 and 18.04. 16.04 is deprecated and fully unsupported / unmaintained, and 18.04 is close enough to that status as well that better to just leave it as-is.

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -481,8 +481,6 @@ def write(x=''):
             "default": [
                           "submitty/clang:6.0",
                           "submitty/autograding-default:latest",
-                          "submitty/java:8",
-                          "submitty/java:11",
```
MasterOdin, line 485: How long does it take to download `submitty/java:11`?

eddiekrystowski: Not very long at all

MasterOdin: Let's leave submitty/java:11 in the default list I think for now, given its ubiquitous nature in CS education and the cost of it is minimal. Can definitely kill off java:8 though.

MasterOdin:
@eddiekrystowski For reference, when I setup Ubuntu 20.04, I added guards around these two bits of installer code that rely on Java being installed or that would only be useful if we had Java.

* https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/.setup/INSTALL_SUBMITTY_HELPER.sh#L388-L413
* https://github.com/Submitty/Submitty/blob/2a1575dcd20382aa83f33f71f0adc9ddbced1f54/.setup/install_system.sh#L297-L341

Thinking about it, this does have an issue where on installing java, the instructor will not have access to our tools without also manually running that block. There is a low priority task to move the second block into the install helper so that we can keep those up-to-date that would resolve that, but it's not something need to worry about yet.

MasterOdin (approved):

MasterOdin:
Before merging, @eddiekrystowski could you manually trigger the vagrant up workflow (https://github.com/Submitty/Submitty/actions/workflows/vagrant_up.yaml) to run through? Whenever we touch installer scripts, we want to run that to ensure we're not breaking the basic install.

eddiekrystowski:
Thanks for the help, I'll mention you once it's done.

.setup/CONFIGURE_SUBMITTY.py
bmcutler: No, we are still using java 8 in classes.

bmcutler: But our classes are running out of custom containers with jacoco and dafny
submittyrpi/csci2600:java_tools_dafny

So...  I'll not fight about this.

PR #6776, opened 2021-07-20
[Bugfix:InstructorUI] Fix Graders Page Twig

cjreed121:
There is a twig error when loading the graders page on a brand new course.

### What is the new behavior?
Variable is checked to exist before it's value is used and the page loads properly now.

### DIFF
--- a/site/app/templates/admin/users/GraderList.twig
+++ b/site/app/templates/admin/users/GraderList.twig
@@ -54,7 +54,7 @@
 </div>
 {% include('admin/users/UserForm.twig') %}
 {% include('admin/users/GraderListForm.twig') %}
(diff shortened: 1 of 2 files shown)


shailpatels (changes requested):
The variable `num_rotating_sections` never appears to be passed to `RegistrationSectionsForm.twig`

The only place `RegistrationSectionsForm` is used is within `site/app/template/admin/users/GraderList.twig` which does not have an reference of `num_rotating_sections` so it will not show up in the included twig files' context either.

GraderList is only called from `site/app/views/admin/UsersView.php` but is not given a `num_rotating_sections`, it is given a `rot_sections` which is probably an array of rotating sections. I think you might be able to use that instead

PR #6778, opened 2021-07-20
[Testing:Developer] Use constant url for cypress login tests

williamjallen:
The Cypress login tests currently depend upon the sample course having a fixed configuration.  This can cause issues such as the issues exhibited by https://github.com/Submitty/Submitty/pull/6721.

### What is the new behavior?
This PR makes a slight modification to the path used to verify that the user is logged in by checking to see that the instructor has access to the course config page.


MasterOdin (approved):
While this is probably not something that'll ever change (and has existed for for 5 years?), good to move to stuff that's not dependent on creation scripts.

PR #6780, opened 2021-07-20
[Feature:Developer] Add polls to sample course

sbelsk:

Fixes: https://github.com/Submitty/Submitty/issues/6791
By default, when running `vagrant up`, Submini Polls are disabled and have to be manually enabled in the course settings in order to access the polls pages.

### What is the new behavior?
Polls are enabled and a few sample polls are loaded automatically into the sample course, some of which have responses from students. Adding polls to the sample course is the first step towards fully testing the polls functionality in the e2e tests.


williamjallen (approved):

.setup/bin/setup_sample_courses.py
```diff
@@ -1217,6 +1218,85 @@ def add_sample_forum_data(self):
...
+                                  order_id=i,
+                                  poll_id=poll["id"],
+                                  response=poll_response,
+                                  correct=(i in poll["correct_responses"]))
+                i += 1
+
+        # generate responses to the polls
+        poll_responses_data = []
+        # poll1: for each self.users make a random number (0-5) of responses
+        poll1_response_ids = [0, 1, 2, 3, 4]
```
MasterOdin, line 1275: Can make this less hard-coded and more dependent on underlying data
```suggestion
        poll1_response_ids = list(range(len(polls_data[0]['responses'])))
```

.setup/bin/setup_sample_courses.py
```diff
@@ -1217,6 +1218,85 @@ def add_sample_forum_data(self):
...
+        for user in self.users:
+            random_responses = random.sample(poll1_response_ids, random.randint(0, 5))
```
MasterOdin, line 1277: Same here, though I don't think you want the upper limit to be `5` here given that there's only 5 options with `0 <= option_id <= 4`, so you'd want `random.randint(0, 4)`?

```suggestion
            random_responses = random.sample(poll1_response_ids, random.randint(0, len(polls_data[0]['responses']) - 1))
```

.setup/bin/setup_sample_courses.py
```diff
@@ -1217,6 +1218,85 @@ def add_sample_forum_data(self):
...
+                    "student_id": user.id,
+                    "option_id": response_id
+                })
+        # poll2: take a large portion of self.users and make each submit one random response
+        for user in self.users:
+            if random.random() < 0.8:
+                poll_responses_data.append({
+                    "poll_id": polls_data[1]["id"],
+                    "student_id": user.id,
+                    "option_id": random.randint(0, 3)
```
MasterOdin, line 1290: and here:
```suggestion
                    "option_id": random.randint(0, len(polls_data[1]['responses']) - 1)
```

.setup/bin/setup_sample_courses.py
```diff
@@ -1217,6 +1218,85 @@ def add_sample_forum_data(self):
...
+                              question_type=poll["question_type"])
+            i = 0
+            for poll_response in poll["responses"]:
+                self.conn.execute(poll_options_table.insert(),
+                                  option_id=i,
+                                  order_id=i,
+                                  poll_id=poll["id"],
+                                  response=poll_response,
+                                  correct=(i in poll["correct_responses"]))
+                i += 1
```
MasterOdin, line 1270: Given that `poll_response` is used just once and so not saving a bunch of typing on doing it instead of `poll["responses"][i]`, better (in my opinion) to just have this be a regular counting indexed loop as allows getting rid of the two surrounding lines.
```suggestion
            for i in range(len(poll["responses"])):
                self.conn.execute(poll_options_table.insert(),
                                  option_id=i,
                                  order_id=i,
                                  poll_id=poll["id"],
                                  response=poll["responses"][i],
                                  correct=(i in poll["correct_responses"]))
```

MasterOdin (changes requested):
Minor tweaks. The big thing is trying to avoid hardcoding values from data into a script as much as possible. While it's convenient, it can lead to off-by-one errors and such later on as the underlying data might be changed without someone realizing it was hardcoded within the the setup script.

.setup/bin/setup_sample_courses.py
sbelsk: What I'm trying to do is to generate a random subset of the set [0, 1, 2, 3, 4] which represents the id's of the possible responses for the multiple response poll. The function `random.sample()` takes the container and the number of elements to select, and returns a random subset of elements from the given container. I wanted number of elements in the subset to also be random (every student submits a different number of responses), and it can vary from 0 responses to all of the responses, which is 5. So this should be good but without the `- 1`

MasterOdin (approved):

PR #6781, opened 2021-07-20
[Testing:Plagiarism] Add plagiarism UI tests

williamjallen:
There are currently no end-to-end tests for the plagiarism detection interface.

### What is the new behavior?
This PR adds a few basic tests to the plagiarism detection interface.  The configuration form is tested, along with the main page, to ensure that it is possible to create, edit, re-run, and delete gradeable configurations as one would expect.  This PR makes no attempt to verify the correctness of Lichen or of the plagiarism results interface.


williamjallen:
This PR is significantly out of date and the vast majority of the tests would have to be rewritten to be useful.  I will open a new one once the UI stabilizes and I'm able to make Cypress work reliably.

PR #6789, opened 2021-07-21
[Feature:InstructorUI] Usability Revisions for New Course Page

### LINKED ISSUE #6779: Usability Revisions to New Course Web UI

A few small requests for this page:

1. After the text
"Please follow course code conventions established by your school and department."
Let's list all of the previous course codes that the current user has been an instructor for.
E.g.
"My Prior Course Codes: csci1200, csci4530, csci4550"

2. Let's add the GNU/Linux usernames of all users in the currently selected group.
Probably place this underneath the dropdown menu.

3. Oh, that dropdown menu should probably have and be initialized with an 
<unselected> option to prevent defaulting  to whatever is first on that list.

4. Let's add some error checking & clear on page warning messages for:
- If we already have a course for the currently selected semester + course code
- If the selected user group does not have the current user in the group AND submitty_daemon AND submitty_php AND submitty_cgi AND submitty_admin
I think the create course script will fail with these errors, but this page doesn't capture these errors (it just submits a request to the job queue).  Don't change the asynchronous request, just redundantly check for those conditions on this page.

cjreed121:

Refer to issue below for list of revisions needed
Closes #6779 

### What is the new behavior?
Prior courses are listed for the user about to create a course. Usernames of users in the Linux group are listed below the select dropdown. The group dropdown starts with a "Please select an option" value.  There are more error checks to insure there is no prior course with same semester/title and the proper users are in the group before creating a course there.


bmcutler:
@cjreed121 
Please proofread my changes...
Can you also make 2 additional changes?

* The wrapping of the group users is buggy on specific screen widths.  Is it because there are no spaces after the comma for the group members? or is the width of a textbox incorrect?
Screen Shot 2021-08-04 at 9 55 28 AM

* And let's place the "My Prior Course Codes" (now blue) underneath the data entry box for the course code (on the right side), similar to the placement of the "Users in this group" message.

Thanks

bmcutler (approved):

PR #6796, opened 2021-07-22
[Testing:InstructorUI] Add tests for late days allowed

williamjallen:
There are currently no tests for the late days allowed page

### What is the new behavior?
This PR adds a few end-to-end tests to the late days allowed page to ensure that the basic functionality still works.  The test add, modify, and delete late days for a user.


HenrikLam (approved):
Looks good and works for me!

PR #6800, opened 2021-07-22
[Bugfix:Submission] Refactor Notebook Autosave

### LINKED ISSUE #5351: Adding a new question midexam to a notebook gradeable can break autosave
**Describe the bug**

If a professor adds a new question to a notebook gradable before an old one. And if a student has already filled out the old one. The local storage autosave will restore the old answer to the new textbox. the autosave uses relative location, so if the relative location changes the the restore is broken

More info in this pr #5323 

**Expected behavior**

The restore should know which question to restore to and not just use relative position on the page

**To Reproduce**
1. Fill in answers to all notebook gradable questions
2. Add a new question of the same type before an old question
3. rebuild the gradable and refresh the submission page

**Configuration**

Linux/Firefox (but it should be the same for every os/browsers)

loadedskiffer:
closes #5351 
Autosaving notebook answers currently relies on numbered id's for each question which is what leads to the issue above.

### What is the new behavior?
Autosaving now uses the question filename (determined in the config.json) as a key and since each question is required to have a unique filename, there are no issues.

### Other information?
To Test:
1. Test a lot because this is a total refactor of how autosave works
2. Fill out a notebook gradeable
3. Add a question to the config, reupload the config
4. refresh the notebook
5. If on master, see that the answers after the new question you added are wrong
6. if on new branch, see that the answers are correct


site/public/js/drag-and-drop.js
```diff
@@ -822,7 +822,7 @@ function gatherInputAnswersByType(type){
     {
         var inputs = $("[id^="+type+"_]");
     }
-
+    //console.log(inputs);
```
williamjallen, line 825: should this be removed?

site/public/js/drag-and-drop.js
```diff
@@ -832,6 +832,7 @@ function gatherInputAnswersByType(type){
         var key = "";
         var value = "";
         if(type == "codebox"){
+            //console.log(this_input_answer);
```
williamjallen, line 835: remove?

site/public/js/drag-and-drop.js
```diff
@@ -846,7 +847,7 @@ function gatherInputAnswersByType(type){
         }
         input_answers[key].push(value);
     }
-
+    //console.log(input_answers);
```
williamjallen, line 850: remove?

williamjallen (changes requested):

williamjallen (approved):

site/public/js/gradeable-notebook.js
```diff
@@ -94,43 +129,47 @@ function saveNotebookToLocal() {
...
+            $('.short_answer').each(function(){
+                $(this).children('div').each(function(){
+                    //match file names
+                    if ($(this).attr('class') === filename) {
+                        //set input
+                        const editor = ($(this)[0]).querySelector('.CodeMirror').CodeMirror;
+                        editor.setValue(answers);
+                    }
+                });
+            });
```
eddiekrystowski, line 172:
```suggestion
            $(`.short_answer > div[data-filename="${filename}"]`).each(function(){
              //set input
              const editor = ($(this)[0]).querySelector('.CodeMirror').CodeMirror;
              editor.setValue(answers);
            });
```
This can also be condensed. This selector selects all `div` elements with the `data-filename` attribute equal to the filename that are direct children of an element with the `short-answer` class (direct children because we used `>` in between the element queries.

site/public/js/gradeable-notebook.js
```diff
@@ -94,43 +129,47 @@ function saveNotebookToLocal() {
...
+                                //check proper inputs
+                                for (let i = 0; i < answers.length; i++) {
+                                    if ($(this)[0].defaultValue === answers[i]) {
+                                        $(this).prop('checked', true);
+                                    }
+                                }
+                            });
+                        });
+                    }
+                });
```
eddiekrystowski, line 156:
```suggestion
            $(`fieldset.mc[data-filename="${filename}"] input`).each(function() {
                //check proper inputs
                for (let i = 0; i < answers.length; i++) {
                    if ($(this)[0].defaultValue === answers[i]) {
                        $(this).prop('checked', true);
                    }
                }
            });
```
This can be condensed to remove unnecessary looping. Note that you can use jQuery's [.find()](https://api.jquery.com/find/) method which is similar to .children() except children only searches immediate children while find searches all children no matter how many levels there are. 

This selector will select all `input` elements that are a child (not necessarily direct because we separated by a space, not a `>`) of all `fieldset`s with the class `mc` and the `data-filename` attribute equal to the filename.

site/app/templates/notebook/Notebook.twig
```diff
@@ -61,6 +61,7 @@
                 <div
                     id="codebox_{{ num_codeboxes }}"
                     name="codebox_{{ num_codeboxes }}"
+                    class="{{ cell.filename }}"
```
eddiekrystowski, line 64: The filename probably shouldn't be stored in the class because we would probably never need to style these. Also, if we tried, we would have to escape the period in the filename which would be a mess. I recommend storing it in a data attribute.
```suggestion
                    data-filename="{{ cell.filename }}"
```

site/app/templates/notebook/Notebook.twig
```diff
@@ -130,6 +131,7 @@
 
                 <fieldset id="mc_field_{{ num_multiple_choice }}"
                           class="mc"
+                          name="{{ cell.filename }}"
```
eddiekrystowski, line 134: Same suggestion as above, can change to `data-filename` instead of name attribute

eddiekrystowski (changes requested):

eddiekrystowski:
@loadedskiffer I just freehanded those jquery selectors, I think they are right but may need to be tweaked.

site/public/js/gradeable-notebook.js
```diff
@@ -79,13 +79,49 @@ function notebookAutosaveKey() {
...
+    //loop through multiple choice questions and save answers
+    $('.multiple_choice').each(function(){
+        let file_name='';
+        $(this).children('fieldset').each(function(){
+            file_name = $(this).data('filename');
+            if (file_name) {
+                const answers = [];
+                $(this).children('label').each(function(){
+                    //grab selected answers
+                    $(this).children('input').each(function(){
```
eddiekrystowski, line 94:
```suggestion
                $(this).find('input').each(function(){
```
Small suggestion here, using find allows us to skip the looping through the labels. This is also good, because it doesn't tie the javascript code to a specific html structure.
Right now this code only works if the html structure is
```
<fieldset>
    <label>
        <input>
        <input>
        <input>
    </label>
    <label>
        <input>
        <input>
        <input>
    </label>
</fieldset>
```
but using find, it can be like this
```
<fieldset>
    ... (anything, any number of levels, could even be that input is a direct child!)
            <input>
</fieldset>
```

site/public/js/gradeable-notebook.js
```diff
@@ -79,13 +79,49 @@ function notebookAutosaveKey() {
...
+                        if ($(this)[0].checked) {
+                            answers.push($(this)[0].defaultValue);
+                        }
+                    });
+                });
+                mc_inputs.push([file_name, answers]);
```
eddiekrystowski, line 100: I would recommend using an object to store the filename and value instead of an array. As someone taking a look at this code for the first time, I was confused about what `inputs.multiple_choice[id][0]` was vs `inputs.multiple_choice[id][1]`. In javascript the keys will be implied if not specified, so if you do
```suggestion
                mc_inputs.push({file_name, answers});
```
then this would make an object like this:
```
{
    file_name: file_name,
    answers: answers
}
```
And then when loading, you can use the object instead of the array for more clarity

site/public/js/gradeable-notebook.js
```diff
@@ -79,13 +79,49 @@ function notebookAutosaveKey() {
...
+            }
+        });
+    });
+
+    //save short answers
+    $('.short_answer').each(function(){
+        $(this).children('div').each(function(){
```
eddiekrystowski, line 107: Same as earlier, if someone were to add another level of div nesting outside of the answer, then this code stops working. Plus, targeting the `div` element is risky because it is so common, someone is bound to eventually mess up the structure. Instead, I would recommend this:
```suggestion
        $(this).find('div[name^="codebox_"]').each(function(){
```
This targets all divs that are children of the short answer (at any level), that also have a name attribute that starts with `codebox_`. This way our selector is more specific about the divs it wants so there is less chance of it pulling in garbage and the code becomes clearer.

site/public/js/gradeable-notebook.js
```diff
@@ -79,13 +79,49 @@ function notebookAutosaveKey() {
...
+            let file_name = $(this).data('filename');
+            if (file_name) {
+                let value = '';
+                file_name = file_name.trim();
+                //grab input
+                const editor = ($(this)[0]).querySelector('.CodeMirror').CodeMirror;
+                value = editor.getValue();
```
eddiekrystowski, line 114:
```suggestion
            const file_name = $(this).data('filename').trim();
            if (file_name) {
                //grab input
                const editor = $(this).find('.CodeMirror').CodeMirror;
                const value = editor.getValue();
```
Just a few changes to use const where possible. 
Also, you can use `find` as a JQuery alternative to `querySelector` which removes the need to convert to an HTMLElement first.

site/public/js/gradeable-notebook.js
```diff
@@ -79,13 +79,49 @@ function notebookAutosaveKey() {
...
+                short_answer_inputs.push([file_name, value]);
```
eddiekrystowski, line 115: See earlier suggestion about storing in object

site/public/js/gradeable-notebook.js
```diff
@@ -94,43 +130,54 @@ function saveNotebookToLocal() {
...
-            const index = id.substring('multiple_choice_'.length);
-            $(`#mc_field_${index} :input`).each((_index, element) => {
-                $(element).prop('checked', values.includes(element.value)).change();
+        //to prevent data loss when these changes get installed on production
+        const not_found = [];
+
+        //restore multiple choice
+        for (const id in inputs.multiple_choice) {
+            const filename = inputs.multiple_choice[id][0];
+            const answers = inputs.multiple_choice[id][1];
```
eddiekrystowski, line 144: You can also change this line to use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to extract the properties from the object (change {} to [] if using array) (totally not necessary though, just something cool)
```suggestion
            const {filename, answers} = inputs.multiple_choice[id];
```

site/public/js/gradeable-notebook.js
```diff
@@ -94,43 +130,54 @@ function saveNotebookToLocal() {
...
-        for (const id in state.short_answer) {
-            const answer = state.short_answer[id][0];
-            // Restore the answer and trigger change events (for setting button states)
-            // (see https://stackoverflow.com/questions/4672505/why-does-the-jquery-change-event-not-trigger-when-i-set-the-value-of-a-select-us)
-            $(`#${id}`).val(answer).trigger('input');
+
+        //restore short answers
+        for (const id in inputs.short_answer) {
+            const filename = inputs.short_answer[id][0];
+            const answers = inputs.short_answer[id][1];
```
eddiekrystowski, line 163: Same suggestion as the multiple choice (be careful, you use `value` before with short answers but `answers` here, the variable name needs to match the key name for the destructuring to work on objects)

site/public/js/gradeable-notebook.js
```diff
@@ -94,43 +130,54 @@ function saveNotebookToLocal() {
...
+            let found = false;
+            $(`.short_answer > div[data-filename="${filename}"]`).each(function(){
```
eddiekrystowski, line 165: Now that I think about it, this doesn't even need to be a `.each()` loop anymore, because there should never be more than one question with the same answer file (pretty sure that's not allowed). In that case, you can just capture the value ```const question = $(`.short_answer > div[data-filename="${filename}"]`)``` and continue like normal using `question` instead of `this` and setting found `if (question.length > 0)` (JQuery way to see if the selection actually found anything). Can probably do the same thing with multiple choice

site/public/js/gradeable-notebook.js
```diff
@@ -94,43 +130,54 @@ function saveNotebookToLocal() {
...
+                //set input
+                found = true;
+                const editor = ($(this)[0]).querySelector('.CodeMirror').CodeMirror;
```
eddiekrystowski, line 168: Same as before, can use find to skip this awkward conversion
```suggestion
                const editor = $(this).find('.CodeMirror').CodeMirror;
```

site/public/js/gradeable-notebook.js
```diff
@@ -94,43 +130,54 @@ function saveNotebookToLocal() {
...
+
+        if (not_found.length > 0) {
+            for (const id in not_found) {
+                console.log(not_found[id]);
             }
-            const cm = codebox.CodeMirror;
-            // This automatically triggers the event handler for the clear and
-            // recent buttons.
-            cm.setValue(answer);
+            alert(`Answer(s) could not be restored you will have to copy and paste them in the proper place, the answers are in your console right now, press f12 to open it. Answers: ${not_found}`);
```
eddiekrystowski, line 180: The instructions for how to open the console vary greatly between browsers, so I'm not sure if it is the best spot to put for users, especially because it can be very intimidating if they have never seen/used it before. Perhaps you could just create an element on the page and just insert the text blob there? It doesn't have to be pretty. Maybe try something like 
```
//create container to hold old answers
const old_answers_div = document.createElement('div');
old_answers_div.id = 'old-answers';
old_answers_div.classList.add ('box ', 'red-background');

//create header text to warn user
const old_answers_header = document.createElement('h4');
old_answers_header.innerHTML = "Answer(s) could not be restored you will have to copy and paste them in the proper place";
old_answers_header.classList.add('red-message');
//add header to container
old_answers_div.appendChild(old_answers_header);

// for ... of loop loops through all elements in array, while for ... in loop loops through
// all indexes/keys in arrays/objects
for (const answer of not_found) {
    const answer_text = document.createElement('p');
    answer_text.innerHTML = answer;
    old_answers_div.appendChild(answer_text);
}
$(old_answers_div).insertAfter('#gradeable-info');
```
This would put the answers in a div at the top of the page just below the gradeable info. Other styling may be necessary

eddiekrystowski (changes requested):
Looking better, I have some more feedback about code style/extensibility for the future and DOM traversal safety.

site/public/js/gradeable-notebook.js
loadedskiffer: For some reason I could only get this to work using the way I had it before, that's how it was done previously as well.

site/public/js/gradeable-notebook.js
loadedskiffer: good catch, I couldn't get rid of the .each() loop for multiple choice because I need it to go through each input though

site/public/js/gradeable-notebook.js
```diff
@@ -94,43 +128,65 @@ function saveNotebookToLocal() {
...
+            //add header to container
+            old_answers_div.appendChild(old_answers_header);
+
+            // for ... of loop loops through all elements in array, while for ... in loop loops through
+            // all indexes/keys in arrays/objects
+            for (const i in not_found) {
+                const answer_text = document.createElement('p');
+                answer_text.innerHTML = not_found[i];
+                old_answers_div.appendChild(answer_text);
             }
```
eddiekrystowski, line 188:
```suggestion
            old_answers_div.classList.add('box');
            old_answers_div.style.backgroundColor = 'var(--alert-background-danger-red)';

            //create header text to warn user
            const old_answers_header = document.createElement('h4');
            old_answers_header.style.color = 'var(--focus-danger-red)';
            old_answers_header.innerHTML = 'Answer(s) could not be restored you will have to copy and paste them in the proper place';
            //add header to container
            old_answers_div.appendChild(old_answers_header);

            // for ... of loop loops through all elements in array, while for ... in loop loops through
            // all indexes/keys in arrays/objects
            const old_answers_list = document.createElement('ol');
            old_answers_list.style.marginLeft = '3em';
            for (const i in not_found) {
                const answer_text = document.createElement('li');
                answer_text.innerHTML = not_found[i];
                old_answers_list.appendChild(answer_text);
            }
            old_answers_div.appendChild(old_answers_list);
```
Upon actually testing the code I gave you, I realized it is very ugly. Try this instead.

eddiekrystowski (changes requested):
Restoring multiple choice answers doesn't correctly set the button status underneath. For example, if on my most recent submission I choose A, then I switch to B and refresh the page, the Use Most Recent Submission should be enabled, but it's not.

Other than that, it seems to work well, I just gave a suggestion to fix the styling of that div at the top of the page.

eddiekrystowski (approved):

site/public/js/gradeable-notebook.js
```diff
@@ -130,46 +129,70 @@ function saveNotebookToLocal() {
...
+        if (not_found.length > 0) {
+            const old_answers_div = document.createElement('div');
+            old_answers_div.id = 'old-answers';
+            old_answers_div.classList.add('box');
+            old_answers_div.style.backgroundColor = 'var(--alert-background-danger-red)';
+
+            //create header text to warn user
+            const old_answers_header = document.createElement('h4');
+            old_answers_header.style.color = 'var(--focus-danger-red)';
+            old_answers_header.innerHTML = 'Answer(s) could not be restored you will have to copy and paste them in the proper place';
```
williamjallen, line 181: I believe that the text here isn't grammatically correct.  Would adding a period after restored be better?
Suggestion: `Answer(s) could not be restored. You will have to copy and paste them in the proper place.`

williamjallen (changes requested):
One minor wording suggestion, otherwise looks good.

williamjallen (approved):

PR #6807, opened 2021-07-23
[Bugfix:TAGrading] Fix old annotations not showing

DefinitiveAbove:
Annotations are currently saved and loaded using the URL encoded file path.

### What is the new behavior?
Annotations are now saved and loaded using the original file path.

### Other information?
If any annotations were made during the period that the URL encoded file path was used for saving/loading, they will no longer be used.


eddiekrystowski (approved):
Works well, thanks for fixing this!

PR #6809, opened 2021-07-23
[Refactor:CourseMaterials] View and Twig Refactor

cjreed121:

This is a PR to fix a few issues in #6599. The page did not handle sorting correctly. The migration also had errors on very early years.

### What is the new behavior?
The old migration is fixed. A new migration was written to support storing links in the database rather than a json file. The view and twig was rewritten such that sorting works as expected.

### Other information?
An old migration was rewritten to fix issues in it.


HenrikLam:
When I try to access the course material page I got this error.
image
Is it perhaps a problem with a previously uploaded url having no title?

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -355,229 +355,236 @@ public function ajaxUploadCourseMaterialsFiles(): JsonResponse {
...
+                            $entries = [];
+                            $disallowed_folders = [".svn", ".git", ".idea", "__macosx"];
+                            $disallowed_files = ['.ds_store'];
+                            for ($i = 0; $i < $zip->numFiles; $i++) {
+                                $entries[] = $zip->getNameIndex($i);
                             }
-                            if (substr($name, -1) !== '/') {
-                                foreach ($disallowed_files as $file) {
-                                    if (basename($name) === $file) {
+                            $entries = array_filter($entries, function ($entry) use ($disallowed_folders, $disallowed_files) {
```
HenrikLam, line 471: Should we alert the user if something is disallowed?

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -355,229 +355,236 @@ public function ajaxUploadCourseMaterialsFiles(): JsonResponse {
...
+                                if (substr($name, -1) !== '/') {
+                                    foreach ($disallowed_files as $file) {
+                                        if (basename($name) === $file) {
+                                            return false;
+                                        }
+                                    }
+                                }
+                                return true;
+                            });
+                            $zfiles = array_filter($entries, function ($entry) {
```
HenrikLam, line 487: Same as above

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -355,229 +355,236 @@ public function ajaxUploadCourseMaterialsFiles(): JsonResponse {
...
-                }
-                // Is this really an error we should fail on?
-                if (!@unlink($uploaded_files[1]["tmp_name"][$j])) {
-                    return JsonResponse::getErrorResponse("Failed to delete the uploaded file {$uploaded_files[1]['name'][$j]} from temporary storage.");
+                    else {
+                        return JsonResponse::getErrorResponse("The tmp file '{$uploaded_files[1]['name'][$j]}' was not properly uploaded.");
+                    }
+                    // Is this really an error we should fail on?
+                    if (!@unlink($uploaded_files[1]["tmp_name"][$j])) {
+                        return JsonResponse::getErrorResponse("Failed to delete the uploaded file {$uploaded_files[1]['name'][$j]} from temporary storage.");
```
HenrikLam, line 544: My opinion is that perhaps this isn't something we should fail on, but instead hide from the user. Perhaps this is instead something the sys admin should be concerned with since the user will have very limited information on why this is failing.

site/app/templates/course/CourseMaterials.twig
```diff
@@ -44,150 +45,129 @@
...
-                        if(new Date(fileDT).getTime() <= new Date(currentDT).getTime()){
-                            return green;
-                        } else if(new Date(fileDT).getTime() >= new Date(neverDT).getTime()){
-                            return red;
-                        } else {
-                            return yellow;
+                    <script>
+                        function determineRelease(inputID){
+                            var now = new Date();
+
```
HenrikLam, line 104: Generally the use of `var` is discouraged. Perhaps we can change these twig functions to use `const` or `let` instead?

HenrikLam (changes requested):

loadedskiffer (changes requested):
Sorting priority doesn't seem to be working for folders, changing their priority isn't changing where they are displaying

site/app/controllers/course/CourseMaterialsController.php
cjreed121: I don't think so. These are things that should typically be not included in an upload. For example, people on a Mac would never see the .ds_store file but would get alerted every time they upload a zip. I think it is best to keep it how it is. Unless you have some suggestion here?

loadedskiffer (approved):

HenrikLam:
It seems that the sorting rank for folders still do not work properly. All folders have rank 99 and all files are rank 0 while the link has rank 5.
image

williamjallen (approved):

HenrikLam (changes requested):
Everything is functioning as intended, however, the submit full change button for folders is enabled just by clicking the field for changing release date and does not require any input. Changing the limit to section option and then changing it back to what it was originally does also enable the button. The button also stays as enabled even after exiting the modal pop up. Also the hide from students option has no effect on the button.

cjreed121:
> Everything is functioning as intended, however, the submit full change button for folders is enabled just by clicking the field for changing release date and does not require any input. Changing the limit to section option and then changing it back to what it was originally does also enable the button. The button also stays as enabled even after exiting the modal pop up. Also the hide from students option has no effect on the button.

I was able to fix all except the release date change. That seems like a much tougher fix since due to the flatpickr there. I may leave that for now and maybe there can be a fix later. I am assuming that once you change those values you do plan to make a full change.

PR #6811, opened 2021-07-23
[Refactor:CourseMaterials] Use ZipStream for bulk downloading

### LINKED ISSUE #6229: Use zipstream for course materials download folder
As noted in PR #6214 

Instead of writing a (potentially large) zip file to /tmp, let's use the zip stream library for course materials download folder...  like we do in other parts of the site.

loadedskiffer:
closes #6229 

### What is the new behavior?
Course materials zip download now uses the ZipStream library 

### Other information?
To test:
As instructor upload folders to course materials
As student download the folders

### DIFF
--- a/site/app/controllers/course/CourseMaterialsController.php
+++ b/site/app/controllers/course/CourseMaterialsController.php
@@ -84,22 +84,25 @@ class CourseMaterialsController extends AbstractController {
         }
 
         $temp_dir = "/tmp";
-        // makes a random zip file name on the server
         $temp_name = uniqid($this->core->getUser()->getId(), true);
         $zip_name = $temp_dir . "/" . $temp_name . ".zip";
-        // replacing any whitespace with underscore char.
+         //replacing any whitespace with underscore char.
         $zip_file_name = preg_replace('/\s+/', '_', $dir_name) . ".zip";
-        // Always delete the zip file after script execution
-        register_shutdown_function('unlink', $zip_name);
 
-        $zip = new \ZipArchive();
-        $zip->open($zip_name, \ZipArchive::CREATE | \ZipArchive::OVERWRITE);
         $isFolderEmptyForMe = true;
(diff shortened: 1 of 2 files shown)


cjreed121 (approved):

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -129,13 +133,14 @@ public function downloadCourseMaterialZip($dir_name, $path) {
...
         }
-        $zip->close();
+        $zip_stream->finish();
         header("Content-type: application/zip");
         header("Content-Disposition: attachment; filename=$zip_file_name");
         header("Content-length: " . filesize($zip_name));
         header("Pragma: no-cache");
         header("Expires: 0");
         readfile("$zip_name");
+        unlink($zip_name);
```
MasterOdin, line 143: Is this already handled by `register_shutdown_function('unlink', $zip_name);` on line 92?

site/public/js/server.js
```diff
@@ -659,6 +659,7 @@ function check_server(url) {
 }
 
 function downloadFile(path, dir) {
+    console.log(['download'] + `?dir=${encodeURIComponent(dir)}&path=${encodeURIComponent(path)}`);
```
MasterOdin, line 662: can be removed?

site/app/controllers/course/CourseMaterialsController.php
loadedskiffer: I realized lines 137-143 are unnecessary when using zip stream and were causing problems with some downloads, they have been removed

MasterOdin (approved):
Could you also submit a PR to cleanup:

https://github.com/Submitty/Submitty/blob/01d292e2b854ce3a9a98e24d70919c17af43e1e7/site/app/controllers/MiscController.php#L352-L358?

PR #6812, opened 2021-07-23
[Refactor:Notifications] Email Status Pagination

### LINKED ISSUE #6767: sysadmin webpage does not load if we have too many emails...
**Describe the bug**
Production system with over 1 years worth of emails. fails to load superuser email status page:
https://REDACTED_SYSTEM_URL/superuser/email_status

Screen Shot 2021-07-19 at 10 41 02 AM

**Expected behavior**

Yes, we do need to resolve issue #3326, and cleanup the oldest emails in the table...

But, if the data in this table is massive, we should still handle this more gracefully.  Is the problem that the query result itself is too large?  Or is the problem that the page is attempting to html format the display of the results of this query?

1. If the query result is too big, let's add a cap to the size of this query, grabbing the most recent 10,000 emails or something.

2. If the html formatting of the query is too big, let's only attempt to html display the most recent values in this query result.  Perhaps displaying only 1000 emails per page, and then a "next page" option.  (We might want/need to look through more than the 1000 most recent emails.)

3. Is this also potentially a problem for a single course?  We may need to employ these limits for the instructor page for the email status for a single course.  I can't tell if this is a problem because the course page data isn't very big yet on our production system, since the column for a single course is recently added and the courses are small this summer.  Let's make sure to generate a large amount of test data ...

HenrikLam:

Closes #6767

Currently, all email statuses are gathered and listed all at once. On production where there are many emails sent over the years, the system will run out of ram trying to load all the necessary information.

### What is the new behavior?
The page now separates the email statuses and the email statuses are loaded asynchronously when a new page is selected. The database queries are also moved to use doctrine instead.


HenrikLam:
Hi @MasterOdin @cjreed121, please let me know how the doctrine stuff look.

site/app/controllers/EmailStatusController.php
```diff
@@ -15,34 +17,77 @@ public function __construct(Core $core) {
...
+        );
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/email_status_page", methods={"GET"})
+     * @Route("/api/courses/{_semester}/{_course}/email_status_page", methods={"GET"})
+     * @AccessControl(role="INSTRUCTOR")
+     * @return array
+     */
+    public function getEmailStatusesByPage(): array {
```
cjreed121, line 44: I am confused what this function is supposed to be returning. You return a Json response which contains HTML from the rendered Twig. You specify that it will return an array though. If this were to be a web and an API route then it should be strictly a normal json output. I think this should be the normal route which returns a WebResponse. If you want to add this to an API route then that should simply return a JSON containing information about the email statuses per page but not the HTML.

site/app/controllers/EmailStatusController.php
```diff
@@ -15,34 +17,77 @@ public function __construct(Core $core) {
...
+        );
+    }
+
+    /**
+     * @Route("/admin/email_status_page", methods={"GET"})
+     * @Route("/api/admin/email_status_page", methods={"GET"})
+     * @AccessControl(level="SUPERUSER")
+     * @return array
+     */
+    public function getSuperuserEmailStatusesByPage(): array {
```
cjreed121, line 81: Why is the first one /superuser but the second one /admin for the routes?

site/app/entities/email/EmailEntity.php
```diff
@@ -0,0 +1,113 @@
...
+ * @ORM\Table(name="emails")
+ * @method string getUserId()
+ * @method string getSubject()
+ * @method string getBody()
+ * @method \DateTime getCreated()
+ * @method \DateTime getSent()
+ * @method string getError()
+ * @method string getEmailAddress()
+ * @method string getSemester()
+ * @method string getCourse()
```
cjreed121, line 19: Can be removed

site/app/entities/email/EmailEntity.php
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue
+     * @var int
+     */
+    private $id;
+    /**
+     * @ORM\Column(type="string")
+     * @prop-read string
```
cjreed121, line 31: `@prop-read` only works for models in the code base. This should be `@var` instead.

site/app/controllers/EmailStatusController.php
```diff
@@ -15,34 +17,77 @@ public function __construct(Core $core) {
...
+        $page = $_GET['page'] ?? 1;
+        $result = new EmailStatusModel($this->core, $this->core->getSubmittyEntityManager()->getRepository(EmailEntity::class)->getEmailsByPage($page));
```
cjreed121, line 83: Is a model still necessary? Could this be moved to a utility or just put here or in the view raw?

site/app/repositories/email/EmailRepository.php
```diff
@@ -0,0 +1,87 @@
...
+use Doctrine\ORM\Tools\Pagination\Paginator;
+
+class EmailRepository extends EntityRepository {
+    const PAGE_SIZE = 1000;
+    const MAX_SUBJECTS_PER_PAGE = 10;
+
+    public function getEmailsByPage(int $page, $semester = null, $course = null): Paginator {
+        $entity = EmailEntity::class;
+        $course_specific = ($semester && $course) ? "WHERE e.semester = '{$semester}' AND e.course = '{$course}' " : "";
+        $dql = "SELECT e FROM {$entity} e {$course_specific} ORDER BY e.created DESC";
```
cjreed121, line 17: You should do something like :course_specific and then when you create the query use the setParamter function on course_specific. A good example is in #6708.

migration/migrator/migrations/master/20210726142358_change_email_id_type.py
```diff
@@ -0,0 +1,25 @@
...
+def up(config, database):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    database.execute("ALTER TABLE emails ALTER COLUMN id SET DATA TYPE bigint;")
```
cjreed121, line 13: I feel this belongs in a separate PR

site/app/repositories/email/EmailRepository.php
```diff
@@ -0,0 +1,87 @@
...
+        }
+        $q = $this->_em->createQuery($dql);
+        $query_res = $q->getResult();
+        $page = 1;
+        $count = 0;
+        $subject = 0;
+        foreach ($query_res as $email) {
+            $count += $email[1];
+            $subject += 1;
+            if ($count > self::PAGE_SIZE || $subject >= self::MAX_SUBJECTS_PER_PAGE) {
```
cjreed121, line 43: What is $count is 3x the size of PAGE_SIZE shouldn't $page be increased more than 1?

site/app/repositories/email/EmailRepository.php
```diff
@@ -0,0 +1,87 @@
...
+                $page += 1;
+                $count = 0;
+                $subject = 0;
+            }
+        }
+        return $page;
```
cjreed121, line 49: Is there any way this could maybe be counted in DQL then only divided by the page size in PHP?

site/app/controllers/EmailStatusController.php
HenrikLam: I had thought removing it would be a bit complicated at first, but it wasn't that bad. I've removed the model class now.

site/app/repositories/email/EmailRepository.php
HenrikLam: When I first thought of this, I thought that that if it increases by more than 1 then we will have to split the same email (same subject, same created timestamp) into multiple pages which might cause confusion, so I decided to make 10,000 a softer limit for making it a new page.

site/app/repositories/email/EmailRepository.php
HenrikLam: @MasterOdin After discussing the logic on this query, we couldn't come up with a way to fit this pagination logic into the dql. We were wondering if you have any ideas on a cleaner way to do this in the dql instead.

site/app/controllers/EmailStatusController.php
```diff
@@ -15,34 +16,80 @@ public function __construct(Core $core) {
...
      */
     public function getEmailStatusPage() {
         $semester = $this->core->getConfig()->getSemester();
         $course = $this->core->getConfig()->getCourse();
-        $result = $this->core->getQueries()->getEmailStatusWithCourse($semester, $course);
-        return MultiResponse::webOnlyResponse(
-            new WebResponse(
+
+        /** @phpstan-ignore-next-line */
+        $num_page = $this->core->getSubmittyEntityManager()->getRepository(EmailEntity::class)->getPageNum($semester, $course);
```
cjreed121, line 28: Rather than having phpstan ignore this line you can type hint. You could do something like:

```suggestion
        /** @var EmailRepository $repository */
        $repository = $this->core->getSubmittyEntityManager()->getRepository(EmailEntity::class);
        $num_page = $repository->getPageNum($semester, $course);
```
You would need to add "use EmailRepository" here as well.

site/app/controllers/EmailStatusController.php
```diff
@@ -15,34 +16,80 @@ public function __construct(Core $core) {
...
+     * @Route("/api/courses/{_semester}/{_course}/email_status_page", methods={"GET"})
+     * @AccessControl(role="INSTRUCTOR")
+     * @return array
+     */
+    public function getEmailStatusesByPage(): array {
+        $semester = $this->core->getConfig()->getSemester();
+        $course = $this->core->getConfig()->getCourse();
+        $page = isset($_POST['page']) ? $_POST['page'] : 1;
+        /** @phpstan-ignore-next-line */
+        $result = $this->core->getSubmittyEntityManager()->getRepository(EmailEntity::class)->getEmailsByPage($page, $semester, $course);
```
cjreed121, line 49: Same here

site/app/controllers/EmailStatusController.php
```diff
@@ -15,34 +16,80 @@ public function __construct(Core $core) {
...
+
+        return $this->core->getOutput()->renderJsonSuccess(
+            $this->core->getOutput()->renderTemplate(
                 EmailStatusView::class,
-                'showEmailStatus',
+                'renderStatusPage',
                 $result
             )
         );
     }
```
cjreed121, line 58: Why is this a json response of a rendered template?

site/app/controllers/EmailStatusController.php
```diff
@@ -15,34 +16,80 @@ public function __construct(Core $core) {
...
 
     /**
-     * @Route("/courses/{_semester}/{_course}/email_status")
+     * @Route("/courses/{_semester}/{_course}/email_status", methods={"GET"})
      * @AccessControl(role="INSTRUCTOR")
-     * @return MultiResponse
+     * @return WebResponse
      */
     public function getEmailStatusPage() {
```
cjreed121, line 23:
```suggestion
    public function getEmailStatusPage(): WebResponse {
```

site/app/views/email/EmailStatusView.php
```diff
@@ -8,23 +8,69 @@
 use app\models\EmailStatusModel;
 
 class EmailStatusView extends AbstractView {
-    public function showEmailStatus($email_status) {
+    public function showEmailStatusPage($num_page, $load_page_url) {
```
cjreed121, line 11: Add types

site/app/views/email/EmailStatusView.php
```diff
@@ -8,23 +8,69 @@
...
-        $this->core->getOutput()->addVendorJs('bootstrap/js/bootstrap.bundle.min.js');
-        $this->core->getOutput()->addInternalCss('email-status.css');
-        return $this->core->getOutput()->renderTwigTemplate("EmailStatusPage.twig", [
-            "subjects" => $email_status->getSubjects(),
-            "pending" => $email_status->getPending(),
-            "successes" => $email_status->getSuccesses(),
-            "errors" => $email_status->getErrors(),
-            "courses" => $email_status->getCourses()
+
+    public function renderStatusPage($data) {
```
cjreed121, line 27: Add type

cjreed121 (changes requested):

site/app/controllers/EmailStatusController.php
```diff
@@ -3,46 +3,98 @@
...
+            EmailStatusView::class,
+            'showEmailStatusPage',
+            $num_page,
+            $this->core->buildCourseUrl(["email_status_page"])
+        );
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/email_status_page", methods={"GET"})
+     * @Route("/api/courses/{_semester}/{_course}/email_status_page", methods={"GET"})
```
cjreed121, line 41: Does it make sense to have this as API if it is a web response?

site/app/repositories/email/EmailRepository.php
```diff
@@ -0,0 +1,91 @@
...
+namespace app\repositories\email;
+
+use app\models\EmailStatusModel;
+use Doctrine\ORM\EntityRepository;
+
+class EmailRepository extends EntityRepository {
+    const PAGE_SIZE = 5000;
+    const MAX_SUBJECTS_PER_PAGE = 10;
+
+    public function getEmailsByPage(int $page, $semester = null, $course = null): iterable {
```
cjreed121, line 12: You are returning an array of iterable objects. Shouldn't return type be array then? Technically an array is iterable but array is the more specific option.

cjreed121 (changes requested):

site/app/controllers/EmailStatusController.php
cjreed121: @HenrikLam. Can you resolve this?

cjreed121 (approved):

Issue #6813, opened 2021-07-25
Add docs for setting up nginx for sysadmin


MasterOdin:
On https://submitty.org/sysadmin/installation, we document setting up apache, but do not do the same for nginx for the websockets. We should add a brief section on it, and include details on changing the websocket port (after

lalithaar:
I've submitted a pull request adding the required documentation [here](https://github.com/Submitty/submitty.github.io/pull/651). As this is my first open-source contribution, I would appreciate any feedback or suggestions for improvement. Please let me know if further changes are needed!"

PR #6814, opened 2021-07-25
[Feature:System] Remove hardcoding of websocket port

MasterOdin:

The websocket port within Submitty is hardcoded in the code to use `8443`. This caused collisions if trying to use Ubuntu 18.04 and 20.04 VMs at the same time (let alone additional worker machines).

### What is the new behavior?

This reworks the code to allow the websocket to be set to arbitrary values, being set within the `/usr/local/submitty/config/submitty.json` file, and then that's read in when loading the page. If the value is not set, it'll falback to the default value of `8443` like before. 

### Other information?

Documentation for this will be done as part of #6813.

### DIFF
--- a/site/public/js/server.js
+++ b/site/public/js/server.js
@@ -1,13 +1,4 @@
-var csrfToken = undefined;
-
-window.addEventListener("load", function() {
-  for (const elem in document.body.dataset) {
-    window[elem] = document.body.dataset[elem];
-  }
-});
-
 window.addEventListener("resize", checkSidebarCollapse);
-
 
 ////////////Begin: Removed redundant link in breadcrumbs////////////////////////
 //See this pr for why we might want to remove this code at some point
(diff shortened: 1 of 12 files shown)


williamjallen (approved):
I have reviewed the code twice and tested the branch on my machine.  I have not tested having two VMs running yet but I expect that to work appropriately.

site/public/js/server.js
```diff
@@ -1,14 +1,5 @@
-var csrfToken = undefined;
```
bmcutler, line 1: is this edit related to this PR?

MasterOdin: Not strictly needed, but as I was adding a new variable to be loaded into `window`, I took the opportunity to clean up how these were being set (the `var csrfToken = undefined` is technically doing `window.csrfToken = undefiend`) so that they're not available immediately on page load, instead of having to wait for the onload function below that line to run.

PR #6815, opened 2021-07-25
[Feature:System] installation on Mac M1/ARM using UTM

bmcutler:
### What is the new behavior?

Since virtualbox is not available on the Mac M1/ARM hardware, this is work-in-progress changes to the installation to instead use the UTM virtual machine platform.  

This does not use vagrant (yet?), but instead the developer should manually follow the instructions in
README_utm_arm.txt

### Remaining issues / Comments:

* Maybe the network is flakey?  Sometimes the scripts fail on downloading a docker image.  I added a check for the arm architecture and allow installation to proceed even if the docker download fails.

* I had to comment out some of the installation of analysis tools w/ Haskell bit - maybe this can be fixed?  not urgent.

* The message of the day code is commented out - something was different, possible necessary because this is not a vagrant initialized VM

* The C++ json library usage in autograding is inconsistent on x86 vs arm.  This code has been minimally refactored.


.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -489,6 +494,10 @@ echo -e "Compile and install analysis tools"
 
 mkdir -p ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisTools
 
+# WIP - HASKELL BINARY IS NOT AVAILABLE ARM 64
+if [ ${UTM_ARM} == 0 ]; then
```
MasterOdin, line 498: We should be able to download these binaries, though is this comment indicating that cannot run these on ARM? Or did you not try?

.setup/distro_setup/setup_distro.sh
```diff
@@ -39,7 +39,7 @@ fi
 # check if set, else use default /etc/motd
 # we expect that SUBMISSION_URL and GIT_URL to be set above
 
-if [ ${VAGRANT} == 1 ]; then
+if [ ${VAGRANT} == 1] && [ ${UTM_ARM} == 0]; then
```
MasterOdin, line 42: For my curiosity, what was the error you hit here? I'm curious if UTM has a weird build of Ubuntu MOTD stuff, or what might be going on here.

.setup/install_system.sh
```diff
@@ -666,7 +686,7 @@ if [ ${WORKER} == 0 ]; then
         echo "Creating submitty master database"
         su postgres -c "psql -c \"CREATE DATABASE submitty WITH OWNER ${DB_USER}\""
         su postgres -c "psql submitty -c \"ALTER SCHEMA public OWNER TO ${DB_USER}\""
-        python3 ${SUBMITTY_REPOSITORY}/migration/run_migrator.py -e master -e system migrate --initial
+        python3 ${SUBMITTY_REPOSITORY}/migration/run_migrator.py -c ${SUBMITTY_INSTALL_DIR}/config -e master -e system migrate --initial
```
MasterOdin, line 689: Is this change necessary? the `run_migrator.py` script should be automatically picking up the config directory, and I'm a bit worried on why it wouldn't work under arm.

grading/load_config_json.cpp
```diff
@@ -123,37 +123,58 @@ void AddGlobalDefaults(nlohmann::json &whole_config) {
     whole_config["gradeable_message"] = whole_config.value("assignment_message","");
```
MasterOdin, line 123: are these changes meant for the arm work? Is this because getting a newer version of the `nlohmann::json` library?

.setup/INSTALL_SUBMITTY_HELPER.sh
bmcutler: The comment below was the error message
"only 64-bit (x86_64) Linux binary is available"

I haven't further investigated yet

.setup/install_system.sh
bmcutler: I don't think this was necessary, I was incrementally debugging and I think I was running something from the wrong directory so the default didn't work.  It shouldn't hurt, right?  It might make future installation debugging easier?

grading/load_config_json.cpp
bmcutler: Yes, I hate these changes.
I think it's a different version of nlohmann::json.  
Or perhaps a different version of the compiler, or both.

.setup/distro_setup/setup_distro.sh
bmcutler: I'll get back to you when I retest this.  I might be because it wasn't inside of vagrant?  Or because I was re-running something.

I installed Ubuntu 20.04 Arm Server
https://mac.getutm.app/gallery/ubuntu-20-04

grading/load_config_json.cpp
bmcutler: In PR #6833, I've got us properly using the most recent version of the the nlohmann json library.  And I think it's using the same version of compiler on both x86 and arm machines, but unfortunately I'm getting null/type errors with the original version of the code only on arm.  Neither version of the code is elegant -- lots of this should be refactor eventually.

grading/load_config_json.cpp
```diff
@@ -123,37 +130,48 @@ void AddGlobalDefaults(nlohmann::json &whole_config) {
...
+  json_set_default(whole_config,"gradeable_message",std::string(""));
 
-  whole_config["early_submission_incentive"] = whole_config.value("early_submission_incentive", nlohmann::json::object());
-  whole_config["early_submission_incentive"]["message"] = whole_config["early_submission_incentive"].value("message", "");
-  whole_config["early_submission_incentive"]["minimum_days_early"] = whole_config["early_submission_incentive"].value("minimum_days_early", 0);
-  whole_config["early_submission_incentive"]["minimum_points"] = whole_config["early_submission_incentive"].value("minimum_points", 0);
-  whole_config["early_submission_incentive"]["test_cases"] = whole_config["early_submission_incentive"].value("test_cases", std::vector<int>());
-
-  whole_config["max_submission_size"] = whole_config.value("max_submission_size",MAX_SUBMISSION_SIZE);
+  if (whole_config.find("load_gradeable_message") == whole_config.end()) {
```
MasterOdin, line 136: Could we have here `json_set_default(whole_config, "load_gradeable_message", nlohmann::json::object());`, and so we have just the else branch, a la:

```c++
  json_set_default(whole_config, "load_gradeable_message", nlohmann::json::object());
  json_set_default(whole_config["load_gradeable_message"],"message",std::string(""));
  json_set_default(whole_config["load_gradeable_message"],"first_time_only",false);
```

The doc for [`json::object`](https://nlohmann.github.io/json/api/basic_json/object/).

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -59,7 +60,15 @@ def update_docker_images(user, host, worker, autograding_workers, autograding_co
             except Exception as e:
               print(f"ERROR: Could not pull {image}")
               traceback.print_exc()
-              success = False
+
+              # check for machine
+              if platform.machine() == "aarch64":
+                  # TEMPORARY: docker pull often fails on ARM installation
+                  print("WARNING: SKIPPING DOCKER PULL ERROR")
```
MasterOdin, line 67: Could we include more specifics on what that error is. Given that hitting this requires a specific architecture not everyone has access to, more details would be appreciated just so might know if this were to get fixed at the engine level or something.

MasterOdin:
Running https://github.com/Submitty/Submitty/actions/runs/1227019586 to test changes against Vagrant installs

PR #6819, opened 2021-07-26
[Feature:System] Add Option to Spin Up VM-Worker Pair

### LINKED ISSUE #4744: Automate Additional Portions of Worker Machine Installation and Use
- [ ] Automate the creation of the ```submitty``` user on worker machine installation.
- [ ] Automate adding the ```submitty``` user to the appropriate groups on worker machine installation.
- [ ] Add a check to ```install_all_workers.py``` that flags improper ```submitty``` user specification (e.g. makes certain that the ```submitty``` user is in the correct groups.

eddiekrystowski:

Closes #4744 
Currently, developers are required to manually install worker machines using [the worker installation instructions](https://submitty.org/sysadmin/worker_installation).

### What is the new behavior?
A vagrant VM-worker pair can be set up by running `WORKER_PAIR=1 vagrant up`.
To set up a worker machine on a specific machine that already exists, run `install_worker.sh` on that machine. This will perform steps 1-7 of  [the worker installation instructions](https://submitty.org/sysadmin/worker_installation), with `submitty` as the `SUPERVISOR_USER`. By default, RPI installations are included, but you can run `install_worker.sh --no-rpi` to not include these.

### DIFF
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -210,7 +210,8 @@ print('Hit enter to use default in []')
 print()
 
 if args.worker:
-    SUPERVISOR_USER = get_input('What is the id for your submitty user?', defaults['supervisor_user'])
+    SUPERVISOR_USER = get_input('What is the id for your submitty user?', 'submitty')
+    print('SUPERVISOR USER : {}'.format(SUPERVISOR_USER)) 
 else:
     DATABASE_HOST = get_input('What is the database host?', defaults['database_host'])
     print()
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -51,6 +51,7 @@ parser.add_argument('--setup-for-sample-courses', action='store_true', default=F
                     help="Sets up Submitty for use with the sample courses. This is a Vagrant convenience "
                          "flag and should not be used in production!")
 parser.add_argument('--worker', action='store_true', default=False, help='Configure Submitty with autograding only')
+parser.add_argument('--worker-helper', default=False, help='Configure Submitty alongside a worker VM')
 parser.add_argument('--install-dir', default='/usr/local/submitty', help='Set the install directory for Submitty')
 parser.add_argument('--data-dir', default='/var/local/submitty', help='Set the data directory for Submitty')
 
@@ -461,10 +462,19 @@ if not args.worker:
                 "capabilities": ["default"],
                 "address": "localhost",
                 "username": "",
-                "num_autograding_workers": NUM_GRADING_SCHEDULER_WORKERS,
+                "num_autograding_workers": 0,
                 "enabled" : True
             }
         }
+
+        if args.worker_helper:
+            worker_dict["submitty-worker"] = {
+                "capabilities": ['default'],
+                "address": "192.168.1.8",
+                "username": "submitty",
+                "num_autograding_workers": 5,
+                "enabled": True
+            }
 
         if args.setup_for_sample_courses:
             worker_dict['primary']['capabilities'].extend([
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -51,7 +51,7 @@ parser.add_argument('--setup-for-sample-courses', action='store_true', default=F
                     help="Sets up Submitty for use with the sample courses. This is a Vagrant convenience "
                          "flag and should not be used in production!")
 parser.add_argument('--worker', action='store_true', default=False, help='Configure Submitty with autograding only')
-parser.add_argument('--worker-helper', default=False, help='Configure Submitty alongside a worker VM')
+parser.add_argument('--worker-pair', action='store_true', default=False, help='Configure Submitty alongside a worker VM')
 parser.add_argument('--install-dir', default='/usr/local/submitty', help='Set the install directory for Submitty')
 parser.add_argument('--data-dir', default='/var/local/submitty', help='Set the data directory for Submitty')
 
@@ -467,7 +467,7 @@ if not args.worker:
             }
         }
 
-        if args.worker_helper:
+        if args.worker_pair:
             worker_dict["submitty-worker"] = {
                 "capabilities": ['default'],
                 "address": "192.168.1.8",
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -462,7 +462,7 @@ if not args.worker:
                 "capabilities": ["default"],
                 "address": "localhost",
                 "username": "",
-                "num_autograding_workers": 0,
+                "num_autograding_workers": NUM_GRADING_SCHEDULER_WORKERS,
                 "enabled" : True
             }
         }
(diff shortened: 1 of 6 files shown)


eddiekrystowski:
@MasterOdin  Right now, I have the IP address of the worker machine (only in the VM-worker pair) hard coded as `192.169.1.8`. As far as I understand, the vagrant machines will use their own private network so a collision with any other machine using this IP shouldn't be an issue, but I'm not sure about this, any feedback would be appreciated!

HenrikLam (approved):
Both `vagrant up` with and without the `WORKER_PAIR=1` flag worked for me.

.setup/INSTALL_SUBMITTY_HELPER_BIN.sh
```diff
@@ -82,7 +82,7 @@ chown -R root:"${DAEMON_GROUP}" ${SUBMITTY_INSTALL_DIR}/autograder
 chmod -R 750 ${SUBMITTY_INSTALL_DIR}/autograder
 
 if [ "${WORKER}" == 1 ]; then
-    chown -R root:${SUPERVISOR_USER} ${SUBMITTY_INSTALL_DIR}/sbin/shipper_utils
+    chown -R ${SUPERVISOR_USER} ${SUBMITTY_INSTALL_DIR}/sbin/shipper_utils
```
MasterOdin, line 85: Please use `user:group` notation here as we do everywhere else, even if the intention is that it'll be `${SUPERVISOR_USER}:${SUPERVISOR_USER}`. This keeps things consistent, and also very explicit on what the permissions are.

.setup/install_system.sh
```diff
@@ -164,6 +169,11 @@ else
 fi
 
 if [ ${WORKER} == 1 ]; then
+    # Setting it up to allow SSH as root by default
+    mkdir -p -m 700 /root/.ssh
+    cp /home/vagrant/.ssh/authorized_keys /root/.ssh
+
+    sed -i -e "s/PermitRootLogin prohibit-password/PermitRootLogin yes/g" /etc/ssh/sshd_config
```
MasterOdin, line 176: These should be wrapped in a `if [ ${VAGRANT} == 1]; then` or else it'll break running this on regular machines we want to setup as a worker.

.setup/install_worker.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+
+# Usage:
+#   install_worker.sh [--no-rpi]
+
+# Read through the flags passed to the script reading them in and setting
+# appropriate bash variables, breaking out of this once we hit something we
+# don't recognize as a flag
+
+export NO_RPI=0
```
MasterOdin, line 10: This should be setup to operate in the same fashion as `install_system.sh` where we should only use `rpi.sh` if they pass the `EXTRA=rpi` environment variable to it. The stuff installed there is not used at all within development, and given our move to Docker, is likely to get phased out altogether.

.setup/install_system.sh
```diff
@@ -611,8 +628,8 @@ echo Beginning Submitty Setup
 
 #If in worker mode, run configure with --worker option.
 if [ ${WORKER} == 1 ]; then
-    echo  Running configure submitty in worker mode
-    python3 ${SUBMITTY_REPOSITORY}/.setup/CONFIGURE_SUBMITTY.py --worker
+    echo "Running configure submitty in worker mode"
+    echo "submitty" | python3 ${SUBMITTY_REPOSITORY}/.setup/CONFIGURE_SUBMITTY.py --worker
```
MasterOdin, line 632: This argument should only be being passed in when `${VAGRANT} == 1` is true.

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -468,13 +470,28 @@ def write(x=''):
...
                 'python',
                 'et-cetera',
                 'notebook',
             ])
+            worker_dict['submitty-worker']['capabilities'].extend([
+                'cpp',
+                'python',
+                'et-cetera',
+                'notebook',
+            ])
```
MasterOdin, line 494: Shouldn't this be under the `if args.worker_pair` branch? Why would we want to set the capabilities for the machine if it doesn't exist?

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -50,6 +50,7 @@ def get_input(question, default=""):
                     help="Sets up Submitty for use with the sample courses. This is a Vagrant convenience "
                          "flag and should not be used in production!")
 parser.add_argument('--worker', action='store_true', default=False, help='Configure Submitty with autograding only')
+parser.add_argument('--worker-pair', default=False, help='Configure Submitty alongside a worker VM')
```
MasterOdin, line 53: Make the help string more explicit that this should only be run when using vagrant.

Vagrantfile
```diff
@@ -50,6 +65,16 @@ Vagrant.configure(2) do |config|
     ubuntu.vm.network 'forwarded_port', guest: 1501, host: 1501   # site
     ubuntu.vm.network 'forwarded_port', guest: 8433, host: 8433   # Websockets
     ubuntu.vm.network 'forwarded_port', guest: 5432, host: 16432  # database
+    ubuntu.vm.provision 'shell', inline: $script
+  end
+
+  config.vm.define 'submitty-worker', autostart: autostart_worker do |ubuntu|
+    ubuntu.vm.box = 'bento/ubuntu-20.04'
+    # If this IP address changes, it must be changed in install_system.sh and 
+    # CONFIGURE_SUBMITTY.pyto allow the ssh connection
```
MasterOdin, line 74:
```suggestion
    # CONFIGURE_SUBMITTY.py to allow the ssh connection
```

Vagrantfile
```diff
@@ -50,6 +65,16 @@ Vagrant.configure(2) do |config|
...
+    ubuntu.vm.network "private_network", ip: "172.18.2.8"
+    ubuntu.vm.network 'forwarded_port', guest: 22, host: 2220, id: 'ssh'
+    ubuntu.vm.provision 'shell', inline: $worker_script
   end
```
MasterOdin, line 78: Move this block after the `ubuntu-20.04` block. We should be grouping related blocks (`ubuntu-18.04` and `ubuntu-20.04` in this case) next to each other to enforce changes to one should be made to the other.

MasterOdin (changes requested):

HenrikLam:
I should also add that while the worker machines do get spun up with the flag, I haven't been able to stop the daemon workers on the worker machine using the commands.

.setup/install_system.sh
eddiekrystowski: I agree, I am just not sure what to do since I don't pass the vagrant flag to `install_system.sh` in `install_worker.sh` because I don't want any of the other vagrant setup to happen, just this special setup for the worker, which I understand wouldn't work for a regular machine. Should I just pass another flag like `--vagrant-worker` and then replace this with 
`if [ ${VAGRANT_WORKER} == 1]; then`? Seems like it might be a bit much though.

Vagrantfile
eddiekrystowski: Good idea, I'll group `ubuntu-20.04` and `ubuntu-18.04` together at the bottom since the `ubuntu-20.04` needs `submitty-worker` to already be provisioned to set up ssh keys for submitty_daemon.

.setup/install_system.sh
MasterOdin: I would suggest that we reuse the vagrant flag and for stuff that would then get triggered you add an additional `${WORKER} == 0` in addition to the existing vagrant check.

This would perfectly capture the logic of "we are provisioning a regular vagrant VM" vs "we are provisioning a vagrant worker" without introducing a third flag.

.setup/install_worker.sh
eddiekrystowski: I think I've done this correctly in my new commit using `${@}` to pass on the positional arguments, but I'm not sure.

PR #6829, opened 2021-07-27
[Bugfix:Plagiarism] Adjust animation for plagiarism drag bar

williamjallen:
The hover animation for the drag bar on the plagiarism interface currently uses a blur effect on hover to indicate that it can be dragged.  This is different from the behavior on the TA Grading Interface which causes the drag bar to increase in size to indicate that it is draggable.

### What is the new behavior?
This PR makes a slight modification to the way the drag bar on the plagiarism results page works to bring it in line with the TA Grading Interface.

https://user-images.githubusercontent.com/16820599/127198309-976758dc-05bc-448c-b075-40485d3504c7.mov

### Other information?
It would be nice if all of the drag bars could share more code.  Unfortunately, differences between the styling on the pages which use the resizable panels means that code sharing is somewhat impractical.

### DIFF
--- a/site/public/css/plagiarism.css
+++ b/site/public/css/plagiarism.css
@@ -246,14 +246,14 @@ main#main.full-screen-mode .plagiarism-result-cont {
 
 .plag-drag-bar {
     height: 100%;
-    width: 20px;
+    width: 14px;
     position: relative;
     cursor: col-resize;
 }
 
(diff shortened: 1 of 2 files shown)


eddiekrystowski (approved):
Looks good, I agree it would be good to share more code, but as you mentioned TA grading is so different that I think this is fine for now.

PR #6830, opened 2021-07-27
[Refactor:Notifications] Change email id type

HenrikLam:

The current database stores the email id as integer

### What is the new behavior?
The id is now stored as bigint instead.


bmcutler (changes requested):
Looks like there are a number of changes to the .sql file unrelated to email table.

PR #6832, opened 2021-07-27
[Bugfix:Forum] Fix Button Alignment in Empty Forum

### LINKED ISSUE #6629: Create post button on empty forum is misaligned
**Describe the bug**
The "create thread" button is misaligned when there are no posts in a forum.

**To Reproduce**
1. Open forum with no posts
2. See error

**Screenshots**
Untitled

eddiekrystowski:
Fixes #6629 
The Create Thread button is overlapping other items when the forum is empty.
Untitled

### What is the new behavior?
The Create Thread button doesn't overlap anything.
image

### DIFF
--- a/site/app/templates/forum/ForumBar.twig
+++ b/site/app/templates/forum/ForumBar.twig
@@ -1,6 +1,6 @@
 {% set user_group = core.getUser().getGroup() %}
 
-{% if show_more %}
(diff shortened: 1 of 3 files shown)


williamjallen (approved):
Looks good.  This solution is much better than the one I originally attempted with margins.

PR #6777, opened 2021-07-20
[Feature:Plagiarism] Implement prior term gradeables

sbelsk:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
- The "Prior term gradeables" field in the configuration form has no effect
- Many known bugs in the highlighting of regions in the result page
- The `PlagiarismView.php` file has many calculations and logic code segments

### What is the new behavior?
Fixes https://github.com/Submitty/Submitty/issues/2488
- The "Prior term gradeables" field in the configuration form loads other gradeables which share the same UNIX group with the current course
- The result page has been refactored to support the new feature, including a complete rewrite of `plagiarism.js`
- The logic in the view has been moved to the controller and the necessary adjustments to the configuration page twig file have been made
- A small change to the appearance of the plagiarism result page

### Other information?
This PR is dependent on Lichen PR https://github.com/Submitty/Lichen/pull/42 which implements the Lichen backend support for the prior term gradeables feature.

This PR also relies on the infrastructure set by PR https://github.com/Submitty/Submitty/pull/6742 to get courses in the same group, so this PR has to be merged afterwards.

PR #6836, opened 2021-07-28
[Refactor:Plagiarism] Improve other gradeables UI

williamjallen:
### What is the new behavior?
PR https://github.com/Submitty/Submitty/pull/6777 introduced new functionality to Lichen/plagiarism.  After discussion, a few improvements were suggested.  This PR is a follow-up PR which makes a few tweaks to the UI and makes a change to the permissions checking requested in https://github.com/Submitty/Submitty/pull/6777.

In full, this PR:
- Adds error-checking to prevent the current gradeable from being compared with itself and for duplicate values entered for that field in the form
- Changes the wording/descriptive text of the other gradeables subheading
- Excludes courses with the group "root" from being used as the source for other gradeables, regardless of the group attached to the current course
- Adds links to submitty.org documentation where appropriate 

### Screenshots:
Before:
before
After:
after

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -7478,7 +7478,20 @@ SQL;
         return $this->submitty_db->rows();
     }
 
+    public function getGroupForCourse(string $semester, string $course): string {
+        $this->submitty_db->query(
+            'SELECT group_name FROM courses WHERE semester = ? AND course = ?',
+            [$semester, $course]
+        );
+        return count($this->submitty_db->rows()) === 1 ? $this->submitty_db->rows()[0]['group_name'] : '';
+    }
+
     public function getOtherCoursesWithSameGroup(string $semester, string $course): array {
+        // some courses are archived and assigned the group "root" on production.  We want to exclude these.
+        if ($this->getGroupForCourse($semester, $course) === 'root') {
+            return [];
+        }
+
         $this->submitty_db->query(
             'SELECT c2.course, c2.semester FROM courses c1 INNER JOIN courses c2 ON c1.group_name = c2.group_name
                    WHERE c1.semester = ? AND c1.course = ?',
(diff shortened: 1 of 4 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7478,7 +7478,20 @@ public function getUserGroups(string $user_id): array {
...
+        $this->submitty_db->query(
+            'SELECT group_name FROM courses WHERE semester = ? AND course = ?',
+            [$semester, $course]
+        );
+        return count($this->submitty_db->rows()) === 1 ? $this->submitty_db->rows()[0]['group_name'] : '';
+    }
+
     public function getOtherCoursesWithSameGroup(string $semester, string $course): array {
+        // some courses are archived and assigned the group "root" on production.  We want to exclude these.
+        if ($this->getGroupForCourse($semester, $course) === 'root') {
```
sbelsk, line 7491: if `getGroupForCourse` returns an empty string, it can mean that either the DB is corrupted and there are multiple rows with the same semester and course, or that it could not find the semester and course you passed to it. To cover both of those cases, would it not be better to also check on line 7491 that the return is not `''` and handle the error in some way?

sbelsk: @cjreed121 Does line 7491 in `getOtherCoursesWithSameGroup` where we check for the group name being equal to 'root' affect the expected behavior for this function in other places this is used on Submitty (like on the page where instructors can create new courses)?

cjreed121: This would not affect my page. I use the getUserGroups function. I'll go ahead and modify that to what I need in my PR to avoid seeing root as an available group. Also, rather than having the getGroupForCourse function you could just modify the SQL in this function to never select anything that has the root group.

williamjallen: Do you have suggestions on how that could be done?  I wasn't sure how to do that in SQL and I figured that the other function could be useful at some point so I went ahead and added it but it might be better to write it in one query.

cjreed121: Add this to the end of the query `AND c1.group_name != 'root'`. That will return 0 rows if the current course is root group.

sbelsk (approved):

PR #6837, opened 2021-07-28
[Bugfix:Submission] View/Download PDF VCS gradeable

eddiekrystowski:
When viewing your grade for a git gradeable that has an annotated pdf, you can't view or download it.
image

### What is the new behavior?
You can view/download annotated pdfs on a git gradeable.
image

Also, the download annotations button when annotating pdf gradeables is now hidden for peer graders.

### Other Information?
Also includes a fix for some twig errors I was getting regarding grade inquiries.


sbelsk (approved):
Works as described

site/app/controllers/pdf/PDFController.php
```diff
@@ -71,7 +71,7 @@ public function downloadStudentPDF(string $gradeable_id, string $filename, strin
         $annotation_jsons = [];
 
         $latest_timestamp = filemtime($path);
-        $md5_path = md5($anon_path);
+        $md5_path = md5($path);
```
MasterOdin, line 74: These changes render the `$anon_path` variable redundant and could then just remove them.

However, this makes me worried that this is causing user info to get leaked as part of peer grading from this change. What do the URLs look like?

MasterOdin (changes requested):

site/app/controllers/pdf/PDFController.php
eddiekrystowski: image
This is the URL for the TA Grading page when grading bauchg.

MasterOdin: I meant the URL on https://github.com/Submitty/Submitty/blob/git-pdf/site/app/templates/autograding/TAResults.twig#L78.

Though it looks like we have the same `file.path` and `file.anon_path` usage on the line above. What's the difference between those two values, and why would we want to have both?

eddiekrystowski: Oh sorry, I was confusing myself with my other PR on downloading annotations on the annotation page.

The reason I made these changes was because for whatever reason, using `md5($anon_path)` doesn't give the correct file path for pdfs submitted through VCS. I tried investigating whether it was an issue with the checkout vs submission folder, but the issue was the file name of the annotations wasn't right. When I switched to using `$path` instead, both vcs and regular submissions worked fine.

I'm not sure why we have both, the url looks like this `http://localhost:1511/courses/f21/sample/gradeable/peer_pdf/download_pdf?filename=words_881.pdf&path=/var/local/submitty/courses/f21/sample/submissions/peer_pdf/aphacker/1/words_881.pdf&anon_path=/var/local/submitty/courses/f21/sample/submissions/peer_pdf/DF654JKL45QER34/1/words_881.pdf`

And as for the TAResults page that you linked, the urls do expose the student's id in `file.path`, but I don't think that's an issue since the student themselves are the only ones that can access that button (and this is how it was before as well). However, I just realized that this will also happen on the pdf annotation page as a result of #6848, so I'll have to disable that button for peer graders

eddiekrystowski: Nevermind, looks like this has broken the view/download functionality for non-vcs pdfs. This PR was made in the middle of some filename issues happening on pdfs, so now that those are resolved I'll mark this as a draft and give it another go.

eddiekrystowski: It may also be a good idea to revert #6848 until I fix the student id being exposed in the url for peers

MasterOdin: You mean #6846?

eddiekrystowski: Sorry didn’t see this, yes I meant #6846

Eceptonsu (changes requested):
Both buttons were not shown on my end when I test it

bmcutler:
duplicated by PR #8012 
closing.

PR #6839, opened 2021-07-28
[Feature:Plagiarism] Refresh run log while open

williamjallen:
When viewing the run log to see the progress for a given run, the only way to refresh the log is to close and re-open the log.  This can be annoying for instructors who are monitoring the progress of a slow run.

### What is the new behavior?
This PR adds automatic refreshing to the run log so that it's possible to monitor the progress of a run without having to close and re-open the run log.  In addition, the scroll position will remain stuck to the bottom of the log after refreshing unless the user scrolls away.

### DIFF
--- a/site/app/templates/plagiarism/Plagiarism.twig
+++ b/site/app/templates/plagiarism/Plagiarism.twig
@@ -139,20 +139,42 @@
         originalTableWidth = $(".plag-table-cont").width();
     });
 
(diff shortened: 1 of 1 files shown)


sbelsk (approved):
Works well! I'm excited for this PR :) This also solves the case when previously if you spammed the "view log" button before the first request loaded, you would get multiple divs with contents of the log.

PR #6849, opened 2021-07-30
[Bugfix:TAGrading] Fix peer feedback not saving

DefinitiveAbove:
Peer feedback does not save as intended.

### What is the new behavior?
Peer grading now saves and displays the last selected option correctly.
When an error occurs while saving, it now displays "Error Saving Changes" in place of "All Changes Saved".


site/public/js/server.js
```diff
@@ -1375,7 +1375,11 @@ function peerFeedbackUpload(grader_id, user_id, g_id, feedback){
         contentType: false,
         success: function(data) {
             try {
```
shailpatels, line 1377: Is the try catch needed here, I don't think the `$.html` function throws any errors and if the `$.getJSON` function fails it would hit the `error` function

shailpatels (changes requested):

shailpatels (approved):
Looks good, saving and viewing peer feedback works with peer gradeables and peer team gradeables

PR #6824, opened 2021-07-27
[Testing:SubminiPolls] Add polls cypress tests

sbelsk:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
There are no tests for polls currently, except for unit tests for the polls model and bulk export of polls (poll utils).

### What is the new behavior?
Polls now have an e2e test that runs in GitHub actions to test basic features on the instructor and student pages. 

### Other information?
The test operates on the sample course and relies on the existence of the polls that are set up automatically with the course when running `vagrant up` (implemented in https://github.com/Submitty/Submitty/pull/6780).

PR #6865, opened 2021-08-02
[Testing:System] Specify Timezone in e2e Test

cjreed121:

Right now the setup.sh for testing sets the timezone for the site as America/New_York while the rest of the e2e testing suite has no set timezone. This creates a conflict in timezones when storing/viewing data.

### What is the new behavior?
The database (postgres) and testing machine are both set to America/New_York in the e2e test. This is needed for #6824 to pass.

### Other information?
This has been tested in #6824 and has been proven to work during a time where a conflict would happen.

An alternative solution to this would be to remove the specified timezone from the setup.sh and have all of the timezones align to their automatic timezone.

### DIFF
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -249,6 +249,7 @@ jobs:
             env:
               POSTGRES_PASSWORD: submitty_dbuser
               POSTGRES_USER: postgres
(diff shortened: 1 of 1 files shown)


sbelsk (approved):
Solves the weird time zone bug in my polls cypress test PR. Thanks!

PR #6867, opened 2021-08-02
[Feature:Developer] Add install lichen script alias

sbelsk:

Currently, when developing/working on the Lichen repo, to install lichen you have to run `bash /usr/local/submitty/GIT_CHECKOUT/Lichen/install_lichen.sh` (or `install_submitty` which calls the latter but installs many other things too) inside the vagrant terminal.

### What is the new behavior?
An additional saved variable and alias were added to make Lichen development more convenient :)


.setup/install_system.sh
```diff
@@ -135,6 +139,8 @@ export DAEMONCGI_GROUP=${DAEMONCGI_GROUP}
 alias submitty_help=\"echo -e '${INSTALL_HELP}'\"
 alias install_submitty='/usr/local/submitty/.setup/INSTALL_SUBMITTY.sh'
 alias submitty_install='/usr/local/submitty/.setup/INSTALL_SUBMITTY.sh'
+alias install_lichen='sudo bash /usr/local/submitty/GIT_CHECKOUT/Lichen/install_lichen.sh'
+alias lichen_install='sudo bash /usr/local/submitty/GIT_CHECKOUT/Lichen/install_lichen.sh'
```
MasterOdin, line 143: You don't need `sudo` here. These aliases are only available to the root user.

MasterOdin (changes requested):

.setup/install_system.sh
sbelsk: I tried it without the `sudo` first and it spit out an error about missing permissions, so I added it...

MasterOdin: What was the error message?

sbelsk: Hm, I don't remember but I can try testing it again. Btw, is the only way to test this script to do `destroy` and `up` in vagrant, or is it possible to somehow rerun it?

MasterOdin: You can just directly edit `/root/.bashrc` and `source /root/.bashrc` to test the aliases.

MasterOdin (approved):

PR #6870, opened 2021-08-03
[Testing:CourseMaterials] Skip other flaky course materials test

shailpatels:

* [X] Documentation has been updated/added if relevant (N/A)

Currently, there are still course materials tests on CI. Sometimes the tests fail on a pull_request action but not on a push not sure why this is. There may be an issue with long cypress tests or the way the course materials page is set up (multiple requests seem to respond with a server error without a page refresh which slows down the tests)

### What is the new behavior?
This PR skips them on CI, those tests can still be run locally while the issue behind them is investigated.


williamjallen:
I've been looking into this issue off and on for a few days now and I think I have identified at least one potential issue but it may not be the only issue.  Right now, the login/logout functions navigate to the login/logout buttons and click them.  The issue seems to be that Cypress doesn't wait for the page to load after they are clicked.  The easy fix, waiting for the page to load by using `cy.get('some_element').should('be.visible')`, fails inexplicably and I am unable to find a way to make it work.  This issue is currently resolved elsewhere by adding a few `cy.wait()` commands but that is not ideal and ultimately leads to flaky tests such as here...

williamjallen:
I am able to reproduce the flakiness of this test on my local machine by running the test multiple times back-to-back without a break.

williamjallen:
My attempts to fix this issue are on the branch `fix-course-materials-test` if that's any use to you.

bmcutler:
@shailpatels 
please resolve merge conflicts

shailpatels:
> I am able to reproduce the flakiness of this test on my local machine by running the test multiple times back-to-back without a break.

Oh that's interesting, I'll try that as well. It's hard to tell what the issue is, especially since sometimes the push succeeds and not the PR and vice versa. I think longer tests take more memory when cypress is recording which slows the site down so adding more waits isn't ideal but I'm not 100% sure on this.

williamjallen:
> Oh that's interesting, I'll try that as well. It's hard to tell what the issue is, especially since sometimes the push succeeds and not the PR and vice versa. I think longer tests take more memory when cypress is recording which slows the site down so adding more waits isn't ideal but I'm not 100% sure on this.

That was the direction I was going too.  If you notice, cypress is really struggling to make videos for each of the tests.  I looked at a 3-minute video of a test I wrote for the TA grading interface a few days ago and it had a total of three frames.  I'm wondering if it would be worthwhile to remove the video recording  for the moment, and just keep the screenshots.  I've been playing around with PR https://github.com/Submitty/Submitty/pull/6880/ and tried turning off the video recording but it didn't help.  The linked PR is similar in that it runs just fine locally but hasn't passed the CI a single time, always failing on one particular test for no apparent reason.

PR #6871, opened 2021-08-03
[Bugfix:SubminiPolls] Fix results chart for edited polls

### LINKED ISSUE #6868: Submini Poll results are incorrectly represented in the histogram
**Describe the bug**
Submini Poll results incorrectly show the number of responses for some (most of the) options. If this is indeed a bug, it is a critical one because some instructors are making their decisions based on the results of student surveys. If these results are reported incorrectly, it affects which option is taken as the majority student vote. Thus, this bug would completely invalidate any voting results that were collected in any surveys or elections using Submini Polls.

**Expected behavior**
Submini Poll results should correctly show the number of responses for all the options.

**To Reproduce**
1. Navigate to the main Submini Poll page. In my specific example, this is course CSCI2600. The corresponding URL is: https://submitty.cs.rpi.edu/courses/u21/csci2600/polls . Click "View Results" to open the results page. See screenshot 1.
2. Hover over the options in the histogram to see that the option with the highest number of responses is "No" (36 responses). See screenshot 2.
3. Navigate to the SQL Toolbox page (https://submitty.cs.rpi.edu/courses/u21/csci2600/sql_toolbox). Enter the SQL query to get the data on responses from the database:
```
SELECT po.option_id, po.response, COUNT(*) AS responses_count, (SELECT COUNT(*) FROM poll_responses WHERE poll_id = 4) AS total_responses FROM poll_responses pr JOIN poll_options po ON pr.option_id = po.option_id AND pr.poll_id = po.poll_id WHERE pr.poll_id = 4 GROUP BY po.option_id, po.response;
```
4. Observe the output of the query which clearly shows that the option with the highest number of responses was not "No" but "Yes but only if this question is optional and extra credit." See screenshot 3.

**Configuration**
 - Browser: Google Chrome Version 92.0.4515.107 (Official Build) (64-bit)

**Screenshots**
image
image
image

**Additional context**
It looks like the problem might be with the database query Submitty runs internally when instead of joining on the option_id, data is joined on the order_id field. See the screenshot below:
image

MasterOdin:

Fixes #6868 

If you edit a poll and re-arrange the responses, then the poll results would display incorrectly. This was because we were using the `option_id` for the response in graphing for plotly. An example would be a poll with three options (`Yes`, `No`, `Maybe`) where after poll creation and people have put in responses, we switch the ordering of `Yes` and `No`. This means that `Yes` would have `{option_id: 0, order_id: 1}` and `No` would have `{option_id: 1, order_id: 0}`.

After filling out the responses for a number of students, we might have the following data we were plugging into the graph:

```
tick_text = {
	  0: "No"
	  1: "Yes"
	  2: "Maybe"
};

tick_val = {
	  0: 1
	  1: 0
	  2: 2
};

y_data = {
	  0: 1
	  1: 3
	  2: 1
}
```

where plotly would sort itself on the `tick_val` under the hood, but then break the implicit ordering we had in `y_data` that depended on plotly putting things in the order as we gave it (i.e. the ticks being in order of `1, 0, 2`).

### What is the new behavior?

This PR fixes the poll results so that it displays correctly. Instead of trying to do some complicated logic with `tickvals` or `ticktext`, we can just use the [basic bar chart API](https://plotly.com/javascript/bar-charts/#basic-bar-chart) to give plotly an object that has both x,y data and plotly graphs it as is, giving us both the expected graphing result as well as simpler code overall.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -7401,9 +7401,23 @@ AND gc_id IN (
 
     public function getResults($poll_id) {
         $results = [];
-        foreach ($this->getResponses($poll_id) as $option_id => $answer) {
-            $this->course_db->query("SELECT * FROM poll_responses where poll_id = ? and option_id = ?", [$poll_id, $option_id]);
-            $results[$option_id] = count($this->course_db->rows());
+        $query = <<<SQL
+SELECT
+    po.option_id,
+    pr.count
+FROM
+    poll_options AS po
+    LEFT JOIN (
+        SELECT option_id, COUNT(*) AS count FROM poll_responses WHERE poll_id = ? GROUP BY option_id
+    ) AS pr ON pr.option_id = po.option_id
+WHERE
+    po.poll_id = ?
+ORDER BY
+    po.order_id ASC
+SQL;
+        $this->course_db->query($query, [$poll_id, $poll_id]);
+        foreach ($this->course_db->rows() as $row) {
+            $results[$row["option_id"]] = $row['count'];
         }
         return $results;
     }
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -7404,7 +7404,7 @@ AND gc_id IN (
         $query = <<<SQL
 SELECT
     po.option_id,
-    pr.count
+    COALESCE(pr.count, 0) AS count
 FROM
     poll_options AS po
     LEFT JOIN (
--- a/site/app/templates/polls/ViewPollResults.twig
+++ b/site/app/templates/polls/ViewPollResults.twig
@@ -1,28 +1,18 @@
 <script>
     window.onload = function() {
-        const xPoints = [];
-        const xPointsIndexes = [];
-        const yPoints = [];
-        let string;
-        {% for answer, number in results %}
-            string = "{{ poll.getResponseString(answer) }}";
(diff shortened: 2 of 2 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7401,9 +7401,24 @@ public function deletePoll($poll_id) {
...
+    ) AS pr ON pr.option_id = po.option_id
+WHERE
+    po.poll_id = ?
+ORDER BY
+    po.order_id ASC
+SQL,
+            [$poll_id, $poll_id]
+        );
+        foreach ($this->course_db->rows() as $row) {
+            $results[$row["option_id"]] = $row['count'];
```
MasterOdin, line 7421: This change isn't necessary to fix the bug, but was just an optimization on the SQL query that existed before, reducing the amount of queries we run from `N+1` where `N` is number of options to `1`.

sbelsk (commented):
I tested it and I'm still getting an issue with the results graph if I swap the order of the reponses: 
image
image
image
I made 4 students manually submit responses, and the graph should show 1 response for "answer 2" and 3 reponses for "answer 3"

Another comment is that with this fix, answers to the poll that have no student responses don't show up on the "view results" graph, and instructors might like seeing an option there even if no students chose it.

MasterOdin:
@sbelsk You're right, I failed to account for the case where an option has no responses. This should be fixed now.

sbelsk (approved):

PR #6873, opened 2021-08-03
[Refactor:Plagiarism] Move logic out of view

sbelsk:
Currently the `plagiarismMainPage` function in the plagiarism view contains all the logic of compiling the data for the table in the plagiarism main page. 

### What is the new behavior?
The logic has been moved from the plagiarism view to the plagiarism controller.

### Other information?
No breaking changes.


site/app/views/admin/PlagiarismView.php
```diff
@@ -8,72 +8,8 @@
 
 class PlagiarismView extends AbstractView {
 
-    public function plagiarismMainPage($all_configurations, $refresh_page, $nightly_rerun_info) {
+    public function plagiarismMainPage($plagiarism_result_info, $refresh_page) {
```
cjreed121, line 11: Add types to function arguments since you are already editing it.

cjreed121 (changes requested):

site/app/views/admin/PlagiarismView.php
williamjallen: Might as well add types to all of the functions in this file as well.  We didn't do it in the last refactor because we were waiting to do it in this PR.

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -346,14 +347,73 @@ public function plagiarismMainPage(string $refresh_page = "NO_REFRESH"): WebResp
...
+                }
+                $plagiarism_row['rerun_plagiarism_link'] = $this->core->buildCourseUrl(["plagiarism", "gradeable", $plagiarism_row["id"], "rerun"]) . "?config_id={$plagiarism_row["config_id"]}";
+                $plagiarism_row['edit_plagiarism_link'] = $this->core->buildCourseUrl(["plagiarism", "configuration", "edit"]) . "?gradeable_id={$plagiarism_row["id"]}&config_id={$plagiarism_row["config_id"]}";
+                $plagiarism_row['nightly_rerun_link'] = $this->core->buildCourseUrl(["plagiarism", "gradeable", $plagiarism_row["id"], "nightly_rerun"]) . "?config_id={$plagiarism_row["config_id"]}";
+            }
+            $plagiarism_result_info[] = $plagiarism_row;
+        }
 
         return new WebResponse(
             ['admin', 'Plagiarism'],
```
cjreed121, line 413: This could be replaced with PlagiarismView::class

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -346,14 +347,73 @@ public function plagiarismMainPage(string $refresh_page = "NO_REFRESH"): WebResp
...
+                    $submissions += count(array_diff(scandir(FileUtils::joinPaths($this->getConfigDirectoryPath($gradeable['g_id'], $gradeable['g_config_version']), "users", $student)), ['.', '..']));
+                }
+                $students = count($students);
+            }
+            else {
+                $timestamp = "N/A";
+                $students = "N/A";
+                $submissions = "N/A";
+            }
+            $plagiarism_row['timestamp'] = $timestamp;
```
shailpatels, line 375: same comment as above with php array syntax

shailpatels (changes requested):
Since a lot of this code already existed and is just being moved to a different area you can ignore the PHP style comments if needed

I think getting the date format should be the only major concern for maintaining this code in the future

site/app/templates/plagiarism/Plagiarism.twig
```diff
@@ -83,7 +83,7 @@
 {% macro renderPlagiarismRow(row) %}
     <tr id="{{ row['id'] }}_{{ row['config_id'] }}_table_row">
         <td>
-            {% if row['ranking_available'] is defined %}
+            {% if row['ranking_available'] == true %}
```
williamjallen, line 86: I don't believe the `== true` part is necessary here

williamjallen (approved):

cjreed121 (approved):

PR #6880, opened 2021-08-03
[Testing:TAGrading] Add e2e tests to TA Grading Interface

williamjallen:
There are currently only a very minimal number of e2e tests for the TA Grading Interface.  This has caused issues many times in the past when new features and bug fixes accidentally break working features.  The TA Grading Interface is extremely complex and it is unreasonable to try to test everything every time a change is made.

### What is the new behavior?
This PR adds a few basic e2e tests to the TA Grading Interface in an initial attempt to reduce the number of regressions introduced.  Additional unit tests should be added in the future to complement these tests since the vast number of edge cases in the TA Grading Interface means that it is unreasonable to write e2e tests for each and every possible situation.

### Other information?
On my MacBook Pro, these tests take about 3 minutes to complete.


HenrikLam (commented):
Seems to work pretty well, however, I did encounter an error when I first tried to test it where the cypress test was waiting for the log in page to come but the test timed out before the page fully loaded. This bug, however, disappeared when I ran the test again so I believe it's just a problem with the flakiness of the testing.

williamjallen:
The login/logout functions themselves are a little flaky because they don't wait for the page to load after the click correctly.  I believe that that is one of the main issues behind the flakiness of the course materials tests, and is probably something that should be addressed eventually.  I looked into it but came away empty-handed...

HenrikLam:
Yeah, if the flakiness is not unique to this PR I would say it should be addressed eventually in a different PR. It does seem to be consistently passing on my machine though and I am happy with how it looks currently. Should I approve it now or should I approve it after you take it off of draft state?

williamjallen:
I moved it into draft mode temporarily while I try to figure out why it's failing on the last test occasionally.  I haven't been able to reproduce it on my local machine, unlike the flakiness of the course materials tests, but I don't want to introduce yet another bad test.  If you approve, you're welcome to approve.  I'm just going to run it on the CI a few more times and maybe tinker with it a little bit to see if I can make it more solid.

HenrikLam:
In that case I will wait and test it again after it comes out of draft then.

williamjallen:
The tests in this PR are too long and are not reliable enough.  I have created #7486 with a completely different approach in an attempt to address the issues.

PR #6895, opened 2021-08-05
[Bugfix:CourseMaterials] catch migration directory exception

bmcutler:
revisions to the migration originally added in #6599

### DIFF
--- a/migration/migrator/migrations/course/20210723135404_course_materials_link_to_db.py
+++ b/migration/migrator/migrations/course/20210723135404_course_materials_link_to_db.py
@@ -31,12 +31,15 @@ def up(config, database, semester, course):
                     url = None
                     url_title = None
                     if 'external_link' in itemvalue and itemvalue['external_link'] is True:
-                        f = open(itemkey)
-                        data = json.load(f)
-                        url = data['url']
-                        url_title = data['name']
-                        if url is not None and url_title is None:
-                            url_title = url
+                        try:
+                            f = open(itemkey)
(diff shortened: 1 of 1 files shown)


MasterOdin:
What was the particular exception getting thrown?

bmcutler:
I sent an example to Chris over Slack DM.

cjreed121 (approved):
In some courses, a few course materials were given the external link flag as true when they were not links. This must have been a bug before my refactor. Production successfully migrated with this change as it would skip over any course material with this issue since it could not find the JSON file.

MasterOdin:
So we could avoid the exception by wrapping that code:
```
if os.path.exists(itemkey):
    # rest of logic
```
?

cjreed121:
> So we could avoid the exception by wrapping that code:
> 
> ```
> if os.path.exists(itemkey):
>     # rest of logic
> ```
> 
> ?
> 
> The worst case right now is that there'll be some X external link course materials that have been inserted into the database as `(path, url, url_title) VALUES (itemkey, null, null)` and I guess they're broken and the instructor would have to manually clean them up if we cared?
> 
> Or better to have just skipped them? Not that I guess it matters since I'm guessing this migration was already run to completion.

url and url_title will be null and not used unless the type of the course material is a link. Any directory that managed to get set as a link were actually converted back to a directory due to the structure of my first migration. I guess in worst case there is a few extra queries setting those fields to null in this migration. I'll go and switch it to avoid these unnecessary DB queries.

PR #6907, opened 2021-08-05
[Feature:CourseMaterials] Edit Course Materials Links

### LINKED ISSUE #5934: Edit link url for course material link
If the instructor typos the url, when uploading a course material link item,
currently they have to delete the item and re-add the item.

Let's let the edit the link url.

cjreed121:

Once you create a new link on the course materials page there is no way to edit it. The only effective way of editing it is to delete the link and create a new one.
Closes #5934

### What is the new behavior?
If you click the edit button on a link you are given additional fields to edit the title or url of the link.


site/public/js/server.js
```diff
@@ -306,6 +306,20 @@ function newEditCourseMaterialsForm(path, dir, this_file_section, this_hide_from
         $("#all-sections-showing-yes", form).prop('checked',false);
         $("#all-sections-showing-no", form).prop('checked',true);
     }
+    if (is_link === "1") {
```
HenrikLam, line 309: Is there a reason to use `"1"` over `"true"`? Would `"true"` be more readable?

HenrikLam:
If you create 2 links of different names, then change 1 of the link to have the same link, the link that was change seems to disappear (even if the link url is different). The old link also seems to linger and the user can no longer use that name within that directory despite it not being displayed anymore. 
image

site/public/js/server.js
cjreed121: That would be because the value is actually "1" and not "true".

HenrikLam (commented):
Works well, but the bug where if you have a slash as first character for directory name still exists.

HenrikLam (approved):
Seems to work well

PR #6908, opened 2021-08-05
[Feature:Plagiarism] Add ability to click on right panel mark

williamjallen:
It is currently possible to click a mark on the left panel which highlights a matching region (or several) in red but impossible to do the reverse and click a mark on the right panel.  In addition, when clicking on a user in the matching users popup, the mark which was clicked is not highlighted after the colors refresh and it may be difficult for the instructor to find it.

### What is the new behavior?
This PR addresses both of the above issues.  It is now possible to click regions on either the right or left pane with proper highlighting, and the UI will now automatically highlight the clicked region after the colors reload when clicking a user in the matching users popup.

### DIFF
--- a/site/public/js/plagiarism.js
+++ b/site/public/js/plagiarism.js
@@ -385,6 +385,10 @@ function refreshColorInfo(state) {
                             'type': interval.type,
                             'matching_positions': mp_text_marks,
                             'others': interval.others,
+                            'start_line': interval.start_line - 1,
+                            'end_line': interval.start_line - 1,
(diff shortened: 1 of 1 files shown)


sbelsk (commented):
Clicking on the right panel woks well and it highlights the matching region(s) on the left. Sometimes when I switch the right user after clicking on a region in the left user panel and selecting one of the other users in the popup, the matching regions don't turn red, though I couldn't really see any pattern for when it would highlight the matching regions in orange or red.

sbelsk (approved):
works as described.

PR #6915, opened 2021-08-09
[Bugfix:System] Include Archived Courses in Course Codes

cjreed121:

If you go to the course creation page, you will only see current non-archived courses in your prior course codes.

### What is the new behavior?
All courses that you are in will be shown whether it is archived or not.

### Other information?
I made an archived course and saw it disappear from the page. After the fix it is showing correctly.

### DIFF
--- a/site/app/controllers/HomePageController.php
+++ b/site/app/controllers/HomePageController.php
@@ -275,7 +275,10 @@ class HomePageController extends AbstractController {
                 $this->core->getQueries()->getAllTerms(),
                 $this->core->getUser()->getAccessLevel() === User::LEVEL_SUPERUSER,
                 $this->core->getCsrfToken(),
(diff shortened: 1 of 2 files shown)


site/app/controllers/HomePageController.php
```diff
@@ -275,7 +275,10 @@ public function createCoursePage() {
                 $this->core->getQueries()->getAllTerms(),
                 $this->core->getUser()->getAccessLevel() === User::LEVEL_SUPERUSER,
                 $this->core->getCsrfToken(),
-                $this->core->getQueries()->getCourseForUserId($this->core->getUser()->getId())
+                array_merge(
```
bmcutler, line 278: why do we we hav an array merge?  why do we have a function call that gets the course codes only for archived courses?  wouldn't it be better to change the function to be more useful?

PR #6922, opened 2021-08-10
[Feature:Developer] Add VCS gradeable to sample course

sbelsk:

Currently there doesn't exist a sample gradeable where students submit by committing files to a version control system (VCS) repository.

### What is the new behavior?
This PR adds such a gradeable with sample submissions, to help with development and future testing of the feature.

### Other information?
The current state of the PR:
- A gradeable (with autograding) is created with the under the right settings
- The repositories that belong to the gradeable for all the users in the course are generated succesfully
- Sample files are committed to the students' repos


sbelsk:
@tkoz0 I saw that you've been looking into finishing this PR but also eventually gut stuck. Would you mind sharing a log of what you've looked into, so that whoever works on it next can pick up where you left off? Feel free to update the PR body and list any current issues that still need to be addressed/fixed.

IDzyre (commented):
VCS gradeables have a different type of 'form' config created,`
        if self.is_repository:
            form_json['date_due'] = dateutils.write_submitty_date(self.submission_due_date)
            form_json['upload_type'] = 'repository'
            form_json['subdirectory'] = '{$gradeable_id}/{$user_id}'
            return form_json
`

After adding these lines (after the config_path is set around line 2030), the gradeable is able to 'attempt' to find the files. Some of the 'students' repos have files in them, some of them don't,.

prestoncarman:
Partially related to this change, but more of a high level question about using zip files for the submissions. Is there a reason to using a binary format? This means you can not see contents of the file until you download and unzip it. Could someone explain the reasoning behind this?

bmcutler:
@prestoncarman 

Before we had "submit by VCS" all submissions were by drag & drop.  But for classes with larger projects that had code or other files in subdirectories, you cannot "drag & drop" a directory tree and preserve that hierarchy.  (as I understood it back then it was impossible to do this on web for security reasons.  so unless the answer is different today...)

So instead we instructed students to zip up the files and any directory structure, and that directory structure (if any) would be created on the server and the compilation/testing expecting the hierarchy would work correctly. 

If you are instead asking about how a few of our sample/test autograding examples are zip files rather than raw text files?  Well this feature also helped how we how we tested/created submissions with the multiple file examples.  To make sure that when we randomize which file each student in the sample course submits we have a complete file set submission.   (the functionality may have been expanded since those initial examples were created.)

bmcutler (approved):
We aren't getting full credit on any autograding, but we'll debug that in a future PR.
Also, we cannot view the files through the TA grading interface -- we have a permissions error.

bmcutler:
nvmd about the 2nd point, I was mistaken about the permissions error.  I think TA grading is (probably) ok.

PR #5969, opened 2020-10-22
[Feature:InstructorUI] Turn off gradeable due/release date

jportorreal00:
### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?

#5846 
Currently we are able to turn off submission due dates, but not release dates. The placement of gradeables on the homepage is also less than ideal when it comes to the gradeable placement. Gradebles with no due date end up in the "RELEASED" section.

### What is the new behavior?
Turn on or off due dates and release dates. The placement of of gradeables is now improved based on the recommendations in the issue.

### Other information?

~~This could be a breaking change since the "NOT NULL" constraint was dropped in the database. I found a few places where we rely on these dates existing so I have been changing them.~~
The database now holds an additional boolean for storing whether or not the gradeable has a due date. If a gradeable has a due date then it cannot have a release date. Instructors must add a due date and release date later if they plan to release grades overall. Grades can still be accessed and grade inquiries can be made during the open period if a gradeable has no due date. This would be for a work as you go class.

PR #6924, opened 2021-08-11
[Bugfix:Developer] Fix KeyError in setup script

MasterOdin:

`setup_sample_courses.py` was erroring out for me due to the new plagarisism course not having the `eg_has_due_date` or `eg_has_release_date` keys for any of the electronic gradeables.

### What is the new behavior?

This modifies the setup script to allow these fields to be optional and defaulting to true if they're omitted. This is inline with the vast majority of how we use this particular field in the various course yml files.


williamjallen (approved):
Code looks good.  I'll test it tonight or tomorrow night when I finish one of the things I'm working on.  Caused by: https://github.com/Submitty/Submitty/pull/5969 being merged after newer PRs which added the plagiarism course.

PR #6927, opened 2021-08-11
[Refactor:Plagiarism] Reduce plag. results page load time

williamjallen:
Currently, loading the plagiarism results page takes a substantial amount of time.  Most of this time is not spent loading what should be the most expensive parts of the page, but rather making repeated database queries to determine the first and last name associated with each submission.

### What is the new behavior?
This PR combines the repeated database queries into a single query.  In testing, this more than halved the overall page load time for a gradeable with 75 submissions with results.  In the case of a large class of 300 students, each making 5-10 submissions, the load time improvements could be even greater than the ones observed in testing.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -7181,7 +7181,7 @@ AND gc_id IN (
         }
 
         $users = [];
-        foreach ($details_list as &$details) {
+        foreach ($details_list as $details) {
             $users[$details['user_id']] = new User($this->core, $details);
         }
 
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -153,6 +153,16 @@ class DatabaseQueries {
     }
 
     /**
+     * Gets a user from the database given a numeric user_id.
+     *
+     * @param int $numeric_id
+     * @return User|null
+     */
+    public function getUserByNumericId(int $numeric_id): ?User {
+        return $this->getUser($numeric_id, true);
+    }
+
+    /**
      * Gets a list of first and last names from the database given an array of IDs.
      *
      * @param array $user_id_list
@@ -162,26 +172,6 @@ class DatabaseQueries {
         return $this->getUsers($user_id_list);
     }
 
-    /**
-     * @param int $numeric_id
-     * @return User|null
-     */
-    public function getUserByNumericId(int $numeric_id): ?User {
-        return $this->getUser($numeric_id, true);
-    }
-
-    /**
-     * @param int|string $id
-     * @return User|null
-     */
-    public function getUserByIdOrNumericId($id): ?User {
-        $ret = $this->getUser($id);
-        if ($ret === null) {
-            return $this->getUser($id, true);
-        }
-
-        return $ret;
-    }
     /**
      * Retrieves all students from a course and their virtual attendance
      * information such as logs for the course materials page, and logs
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -7177,18 +7177,18 @@ AND gc_id IN (
             );
 
             $updated_details_list = [];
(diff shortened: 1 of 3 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -144,16 +144,37 @@ public function updateSubmittyUserTimeZone(User $user, string $time_zone) {
...
+    public function getUserByNumericId(int $numeric_id): ?User {
         return $this->getUser($numeric_id, true);
     }
 
-    public function getUserByIdOrNumericId($id) {
+    /**
+     * @param int|string $id
+     * @return User|null
+     */
+    public function getUserByIdOrNumericId($id): ?User {
```
cjreed121, line 177: I know you didn't pick this function name but does it make sense to keep it like this. This calls getUser with $id and no is_numeric field which means it will always be by normal id and never numeric id.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7122,20 +7163,29 @@ private function getUser($user_id, bool $is_numeric = false): ?User {
...
+                    $details = array_merge($details, $user);
                 }
-                $details = array_merge($details, $user);
+                $i++;
             }
         }
 
-        return new User($this->core, $details);
+        $users = [];
+        foreach ($details_list as &$details) {
```
cjreed121, line 7184: I don't think you need the & here as you make no edits to $details.
```suggestion
        foreach ($details_list as $details) {
```

cjreed121 (changes requested):
Looks good overall but I just have 2 comments

site/app/libraries/database/DatabaseQueries.php
williamjallen: I was a little confused about the purpose of this function as well.  If I'm not mistaken, this function does get a user by either the normal user ID or the numeric ID.  The function essentially just runs the query again with the is_numeric flag if the regular ID couldn't be found.  I believe the only place this function is called is here: https://github.com/Submitty/Submitty/blob/1704fd8fd4a82010d8a8d904c35b8b2d28be6f31/site/app/views/submission/HomeworkView.php#L637

Perhaps the best thing to do here would be to just rewrite the place where this function is called and delete the function altogether.  It would be better to use `is_numeric()` to make the determination about what type of ID it is instead of just running the query twice...

site/app/libraries/database/DatabaseQueries.php
williamjallen: Yep, you're right.  That's a remnant from a previous attempt I made to fix this issue.  I'll delete it.

site/app/libraries/database/DatabaseQueries.php
cjreed121: I think I misread the code first initially. It definitely works but yeah I agree that there really isn't a need for it to have 2 queries when you can quickly check if it's numeric without the database.

williamjallen: I'll make the changes described above and remove the function.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7176,21 +7176,26 @@ private function getUsers(array $user_id_list, bool $is_numeric = false): ?array
...
+                 }
+                 $details_list = $updated_details_list;
+             }
         }
 
         $users = [];
         foreach ($details_list as $details) {
+            if (isset($users[$details['user_id']])) {
+                $test = $details['user_id'];
+            }
```
cjreed121, line 7198: Purpose of this?

cjreed121 (changes requested):

site/app/libraries/database/DatabaseQueries.php
williamjallen: Good catch.  I'll remove it.  I was using it for debugging to see if duplicate entries existed in the array and forgot to remove it.

cjreed121 (approved):

PR #6943, opened 2021-08-16
[Bugfix:TAGrading] Fix Shaky Overall Comments on Firefox

### LINKED ISSUE #6856: Text area for entering comments in the instructior grading interface is "shaken" on typing every character
**Describe the bug**
When entering grading comments in the "Overall Comment" box using the instructor grading interface, the box "shakes" after entering every single character.

**Expected behavior**
The "Overall Comment" box should stay fixed as the user is entering text. 

**To Reproduce**
1. Navigate to the "Grading Interface" of a particular student in the instructor's interface. In my specific example, this is gradeble p01 in CSCI4963.
2. Make sure the Rubric panel is visible. Scroll that pane; all the way down until the "Overall Comment" box is visible. Start typing some text in that box.
3. Observe that the box repositions itself by adjusting the state of the scroll bars on every single character entry. This produces teh effect of the box being "shaken" as shown in the animated screenshot 1.

**Configuration**
 - Browser: Mozilla Firefox 90.0.2 (64-bit)

**Screenshots**
rnBpqnmJAV

**Additional context**
None.

eddiekrystowski:
Fixes #6856 
Currently, when using the rubric panel on TA Grading to enter an overall comment on Firefox, if you are using a layout that makes this window small (such as 4 panel layout), then every time you type in the box, it shakes.

### What is the new behavior?
The box doesn't shake. However, Firefox does automatically scroll to focus to the line you are typing in. I tried various workarounds to fix this but ultimately they would be very hacky and figured it would be best to let the browser continue its default behavior.

### DIFF
--- a/site/app/templates/misc/MarkdownArea.twig
+++ b/site/app/templates/misc/MarkdownArea.twig
@@ -59,7 +59,7 @@
             {{ other_textarea_attributes }}
         {% endif %}
     >{{ markdown_area_value is defined ? markdown_area_value : '' }}</textarea>
-    <div id="{{ preview_div_id }}" class="fill-available markdown-preview" style="min-height:{{ min_height is defined ? min_height : "0px" }}" hidden></div>
+    <div id="{{ preview_div_id }}" class="fill-available markdown-preview" style="min-height:{{ min_height is defined ? min_height : "0px" }}; max-height:{{ max_height is defined ? max_height : "none" }}" hidden></div>
 </div>
 
 <script>
--- a/site/public/templates/grading/OverallComment.twig
+++ b/site/public/templates/grading/OverallComment.twig
@@ -39,7 +39,7 @@ Required Parameters:
             onclick : "previewOverallCommentMarkdown.call(this, '" ~ (overall_comment["logged_in_user"]["user_id"]|escape) ~ "')", 
             render_buttons : true,
             min_height : "100px",
-            textarea_onkeydown : "auto_grow(this)",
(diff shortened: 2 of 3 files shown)


cjreed121:
When I type in the text box, it still has a shaky feel of the textbox getting bigger then smaller really quick. It definitely seems better than what it was in the issue. Did you experience the textbox growing then shrinking really quick?

HenrikLam:
I tested this on windows Firefox, and the textbox also shakes for me. Also is there a reason why the `MarkdownArea.twig` file is duplicated between two directories?

eddiekrystowski:
https://user-images.githubusercontent.com/28243927/129632928-287527d2-f842-48a2-b4ea-99d74c577b0b.mov

This is what happens when I type in the text box on Mac/Firefox. Can you show what it looks like on your end? (can just show in webex too). If the behavior that happens looks the same as in the original issue, check to make sure the textbox doesn't have `onkeydown="autogrow(this)"`. If that is the case, you may need to do a hard refresh since I deleted that because it was causing the jittering.

The reason `MarkdownArea.twig` is in two places is because TAGrading templates are served with TwigJS, while other areas on the site aren't, so we need one copy for each.

cjreed121 (approved):
I forgot that this would be a Twig JS related issue and didn't clear the cache. After doing so it works well.

HenrikLam (approved):
After clearing the cache, the fix seems to have worked for me.

PR #6945, opened 2021-08-16
[Feature:Calendar] Calendar Items

### LINKED ISSUE #6920: Add messages/notes to calendar
Right now the Submitty calendar only shows gradeables. It would be great if instructors could add messages/notes (important dates, announcements, etc) to the calendar which work towards #5411.

On the course calendar page, an instructor is able to add these messages/notes to the calendar.

cjreed121:

Right now the calendar only contains gradeables.
Closes #6920

### What is the new behavior?
In addition to the main calendar, there are individual course calendars. Items (announcements, important info, notes) can be put on the calendar. Only instructors have permission to put items on their specific course calendar from the course calendar page. Items will show on both the course and global calendar.

### Other information?
This is still a work in progress


eddiekrystowski:
If I enter `Test message <a href="www.google.com">test link</a>` I get this on the calendar.
image

Some tags like `<em>` give the desired effect (italicizing text). I don't know if it is even desirable to allow HTML here? 
Also, after this happens, the pencil is unclickable

This is the HTML of that calendar cell after I do this
image

cjreed121:
I just added a check to make sure the text string does not contain any HTML tags. If so it will not create the calendar item and alert the user. Any existing HTML tags would still show but this is still in development so you would just need to delete that manually if the button is still broken for you.

eddiekrystowski (dismissed):
Looks good, I think it would be a good idea to remove the new item button from the student's page but that can easily be done in another PR.

site/app/views/calendar/CalendarView.php
```diff
@@ -48,9 +53,12 @@ public function showCalendar(CalendarInfo $info): string {
...
             "curr_day" => date("d"),   // the current date
             'date_format' => $this->core->getConfig()->getDateTimeFormat()->getFormat('gradeable'),
-            "gradeables_by_date" => $info->getInfoByDate(),
-            "gradeables_by_section" => $info->getInfoBySections(),
+            "gradeables_by_date" => $info->getGradeablesByDate(),
+            "gradeables_by_section" => $info->getGradeablesBySections(),
             "empty_message" => $info->getEmptyMessage(),
+            "in_course" => $in_course,
+            "items_by_date" => $info->getItemsByDate(),
+            "is_instructor" => $this->core->getUser()->getGroup() === 1
```
williamjallen, line 61: Should the constant 1 be replaced with a constant value which better describes what it is?  If I recall correctly, the constant is `User::GROUP_INSTRUCTOR`.

site/public/js/calendar.js
```diff
@@ -104,6 +109,67 @@ function generateCalendarItem(item) {
         </a>`;
 }
 
+function generateItemHtml(item, date) {
+    let styleString = '';
+    switch (item['type']) {
+        case 'announcement':
+            styleString = 'style="font-weight: bold;"';
```
williamjallen, line 116: what happens if someone comes along later and decides elsewhere in the code that they want to put an inline style?  It might be better to use a CSS class to indicate this.  Bootstrap has the `.font-weight-bold` and `.font-italic` classes which might be helpful here.

site/public/js/calendar.js
```diff
@@ -104,6 +109,67 @@ function generateCalendarItem(item) {
...
+            styleString = 'style="font-style: italic;"';
+            break;
+    }
+    if (isGlobal) {
+        return `<p ${styleString}>${item['course']}: ${item['text']}`;
+    }
+    else {
+        if (isInstructor) {
+            const string =  `<p ${styleString}>${item['text']} <a data-id="${item['id']}" data-type="${item['type']}" data-text="${window.btoa(item['text'])}" data-date="${date}" class="calendar-item-edit-button"><i class="fas fa-pencil-alt"></i></a></p>`;
+            /*document.getElementById(item['id']).onclick = function () {
```
williamjallen, line 128: unnecessary comment?

site/public/js/calendar.js
```diff
@@ -104,6 +109,67 @@ function generateCalendarItem(item) {
...
+    }
+}
+
+function editCalendarItemForm(itemType, itemText, itemId, date) {
+    $('#calendar-item-type-edit').val(itemType);
+    $('#calendar-item-text-edit').val(itemText);
+    $('#edit-picker-edit').val(date);
+    $('#calendar-item-id').val(itemId);
+
+    $('#edit-calendar-item-form').css('display', 'block');
```
williamjallen, line 145: It would be equivalent to use jQuery's `.toggle()` method here, which also avoids accidentally overriding someone's CSS in the future if the display property was set to something like `inline-block`.  Minor change though so if it's not easy to implement elsewhere, don't worry about it.

williamjallen (changes requested):

site/public/js/calendar.js
cjreed121: This code is going to be removed as I realized there already exists some code for styling items.

site/public/js/calendar.js
cjreed121: I'm going to use .show() here instead as there is no reason this button should hide for whatever reason.

williamjallen:
This PR still seems to need some work, with multiple failing tests.  Is there anything in particular you're looking for in a review here?

EdenBendheim:
> This PR still seems to need some work, with multiple failing tests. Is there anything in particular you're looking for in a review here?

Yea for now we just need to test to make sure nothing that can be input with a comment will break the calendar. I checked out the tests that we are failing but I'm not quite sure I understand them.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +31,125 @@ class CalendarController extends AbstractController {
...
+        $courses = $this->core->getQueries()->getCourseForUserId($user->getId());
+
+        $gradeables_of_user = GradeableUtils::getAllGradeableListFromUserId($this->core, $user, $courses, $calendar_messages);
+
+        return new WebResponse(CalendarView::class, 'showCalendar', CalendarInfo::loadGradeableCalendarInfo($this->core, $gradeables_of_user, $courses, $calendar_messages));
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/calendar")
+     * @return WebResponse
```
williamjallen, line 46: This PHPDoc line is unnecessary since the return value already has a native PHP type declaration

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +31,125 @@ class CalendarController extends AbstractController {
...
+                break;
+            }
+        }
+        $gradeables = GradeableUtils::getGradeablesFromUserAndCourse($this->core, $calendar_messages);
+
+        return new WebResponse(CalendarView::class, 'showCalendar', CalendarInfo::loadGradeableCalendarInfo($this->core, $gradeables, $oneCourse, $calendar_messages), true);
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/calendar/newItem", methods={"POST"})
```
williamjallen, line 67: I personally dislike camel case routes, and they generally go against the rest of the routing scheme used across Submitty.  Perhaps the route `/courses/{_semester}/{_course}/calendar/items/new` or similar would be a good alternative.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +31,125 @@ class CalendarController extends AbstractController {
...
+     * @AccessControl(role="INSTRUCTOR")
+     */
+    public function createMessage(): RedirectResponse {
+        $type = $_POST['type'];
+        $date = $_POST['date'];
+        $text = $_POST['text'];
```
williamjallen, line 73: You should first check to make sure these fields exist in the `$_POST` array before blindly accessing them like this.  If the fields don't exist, return an error of some kind.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +31,125 @@ class CalendarController extends AbstractController {
...
+
+        $this->core->getCourseEntityManager()->persist($calendar_item);
+        $this->core->getCourseEntityManager()->flush();
+
+        $this->core->addSuccessMessage("Calendar item successfully added");
+        return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/calendar/editItem", methods={"POST"})
```
williamjallen, line 105: This route could also be overhauled.  What about something along the lines of `/courses/{_semester}/{_course}/calendar/items/{id}/edit`?

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +31,125 @@ class CalendarController extends AbstractController {
...
+     * @AccessControl(role="INSTRUCTOR")
+     */
+    public function editMessage(): RedirectResponse {
+        $type = $_POST['type'];
+        $date = $_POST['date'];
+        $text = $_POST['text'];
+        $id = $_POST['id'];
```
williamjallen, line 112: Same comment here about checking for these fields before blindly accessing them

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +31,125 @@ class CalendarController extends AbstractController {
...
+        catch (\InvalidArgumentException  $e) {
+            $this->core->addErrorMessage("That is not a valid calendar item type");
+            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+        }
+        $this->core->getCourseEntityManager()->flush();
+        return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/calendar/deleteItem", methods={"POST"})
```
williamjallen, line 140: Same here.  What about the route `/courses/{_semester}/{_course}/calendar/items/{id}/delete`?

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +31,125 @@ class CalendarController extends AbstractController {
...
+     * @AccessControl(role="INSTRUCTOR")
+     */
+    public function deleteMessage(): JsonResponse {
+        $id = $_POST['id'];
```
williamjallen, line 144: Same comment about checking for the existence of the field before blindly accessing it here.

site/app/entities/calendar/CalendarItem.php
williamjallen, line 1: I believe that all of the PHPDoc annotations are redundant in this file.

site/app/models/CalendarInfo.php
```diff
@@ -14,6 +14,7 @@
  * Container of information used to fill the calendar.
  */
 class CalendarInfo extends AbstractModel {
+    const COLORS = ['#c3a2d2','#99b270','#cd98aa','#6bb88f','#c8938d','#6b9fb8','#c39e83','#98a3cd','#8ac78e','#b39b61','#6eb9aa','#b4be79','#94a2cc','#80be79','#b48b64','#b9b26e','#83a0c3','#ada5d4','#e57fcf','#c0c246'];
```
williamjallen, line 17: This line is very long and should be reformatted.

Should the colors be listed here anyway?  Why not do this kind of logic in the relevant view?  No need to clutter the model with this kind of UI business.

Additionally, where did all of these colors come from?  Is there a way to share code with another part of the codebase somehow?

site/app/models/CalendarInfo.php
```diff
@@ -102,10 +114,12 @@ public static function loadGradeableCalendarInfo(Core $core, array $gradeables_o
                     'status_note' => ($submit_btn === null) ? "N/A" : $submit_btn->getTitle(),
                     'grading_open' => $gradeable->getGradeStartDate() !== null ? $gradeable->getGradeStartDate()->format($date_format) : '',
                     'grading_due' => $gradeable->getGradeDueDate() !== null ? $gradeable->getGradeDueDate()->format($date_format) : '',
-                    'class' => ($submit_btn === null) ? "" : explode(' ', $submit_btn->getClass())[1],
+                    'class' => /*($submit_btn === null) ? "" : explode(' ', $submit_btn->getClass())[1]*/ 'btn-nav calendar-color-blue',
```
williamjallen, line 117: I assume this commented out code should be eliminated?

site/app/models/CalendarInfo.php
```diff
@@ -119,21 +133,48 @@ public static function loadGradeableCalendarInfo(Core $core, array $gradeables_o
...
+                        'show_due' => false,
+                        'submission' => '',
+                        'status_note' => '',
+                        'color' => $info->colors[$course->getSemester() . $course->getTitle()],
+                        'type' => 'item'
+                    ];
+                    $info->items_by_date[$date][] = $curItem;
+                }
+                catch (\Exception $e) { //Empty catch to skip this item
+                }
```
williamjallen, line 156: Is there a more specific exception we can catch here?  Why is this try-catch necessary in the first place?

site/app/templates/calendar/Calendar.twig
```diff
@@ -5,17 +5,33 @@
...
+    const isGlobal = {% if not in_course %}true{% else %}false{% endif %};
+    const isInstructor = {% if is_instructor %}true{% else %}false{% endif %};
 
     window.addEventListener("load", function () {
         loadCalendar({{ view_month }}, {{ view_year }});
     });
+
+    function openNewItemModal() {
+        $('#new-calendar-item-form').css('display', 'block');
+    }
```
williamjallen, line 17: This should be moved to the relevant `.js` file.

site/app/templates/calendar/Calendar.twig
```diff
@@ -5,17 +5,33 @@
...
 </script>
 
 {# Begin Calendar #}
 <div class="content">
     <h1>Calendar</h1>
+{#    <div class="cal-legend">
+        {% for course, color in colors %}
+            <p>{{ course }}: {{ color }}</p>
+        {% endfor %}
+    </div> #}
```
williamjallen, line 27: This commented out code should presumably be deleted

site/public/css/calendar.css
```diff
@@ -252,6 +255,11 @@ a.cal-gradeable-status-link:focus {
     border: none;
 }
 
+.calendar-color-blue {
+    background-color: #218ffd;
+    color: var(--text-black) !important;
```
williamjallen, line 260: Is the `!important` necessary here?

site/public/js/calendar.js
```diff
@@ -104,6 +101,64 @@ function generateCalendarItem(item) {
...
+    else {
+        if (isInstructor) {
+            const string =  `<p ${styleString}>${item['text']} <a data-id="${item['id']}" data-type="${item['type']}" data-text="${window.btoa(item['text'])}" data-date="${date}" class="calendar-item-edit-button"><i class="fas fa-pencil-alt"></i></a></p>`;
+            return string;
+        }
+        else {
+            return `<p ${styleString}>${item['text']}</p>`;
+        }
+    }
+}*/
```
williamjallen, line 126: Should this be deleted?

williamjallen (changes requested):

site/app/entities/calendar/CalendarItem.php
```diff
@@ -88,7 +89,7 @@ public function setText(string $text): void {
     }
 
     /**
-     * @param int $type
+     * @param int $type Takes in either 0 or 1
```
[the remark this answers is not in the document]
LittleSilver33: Got it. I thought I was making it clear but I now see how it could be the opposite.

site/app/controllers/GlobalController.php
```diff
@@ -185,6 +185,16 @@ public function prep_course_sidebar(&$sidebar_buttons, $unread_notifications_cou
...
+        if ($this->core->getConfig()->isDebug()) {
+            $sidebar_buttons[] = new Button($this->core, [
+                "href" => $this->core->buildCourseUrl(['calendar']),
+                "title" => "Course Calendar",
+                "class" => "nav-row",
+                "id" => "nav-sidebar-calendar",
+                "icon" => "fa-calendar"
+            ]);
+        }
+
```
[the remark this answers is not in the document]
LittleSilver33: I also saw this code when I was reviewing it last night and I was pretty sure that it would only be for development stuff given the isDebug() portion. I opted to leave it but if we decide on changing it I believe it's a quick fix.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +31,125 @@ class CalendarController extends AbstractController {
...
+     * @return WebResponse
+     */
+    public function viewCourseCalendar(): WebResponse {
+        $calendar_messages = [];
+        $user = $this->core->getUser();
+        $courses = $this->core->getQueries()->getCourseForUserId($user->getId());
+        $name = $this->core->getConfig()->getCourse();
+        // Initialize $oneCourse and get the specific course for the course calendar
+        $oneCourse = $courses[0];
+        foreach ($courses as $course) {
```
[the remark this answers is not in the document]
LittleSilver33: I don't know if there is a way to only get a specific course. I tried to find one, but I didn't see a way to get it. I'll take a look again and see if I can find anything.

site/app/models/CalendarInfo.php
LittleSilver33: I honestly don't know what this code does either. It is referenced once from what I can tell (basically these are the colors that the gradeables are given for each course). When I commented it out and created new gradeables though, it didn't actually break anything (which really surprised me). As for where the colors came from, this is before my time so I can only speculate but my hunch is that these are colors that Chris (and probably other people) agreed are distinct enough to be the valid colors for gradeables. If you have any idea what this does please reach out but I'll see if I can find other ways to get colors for gradeables.

cjreed121: I believe I took these colors from office hours. I didn't worry about sharing the code at the time since this was still a draft PR but that would probably be good to do.

site/app/models/CalendarInfo.php
LittleSilver33: So I think that this try-catch is just in case the current calendar item doesn't work for some reason. I really don't think that this try-catch is really necessary but maybe I should make a different way to check if the calendar item is working (although I'm really not sure how to do that without a try-catch). Currently I'm going to delete it but let me know if you want this done differently.

site/app/models/CalendarInfo.php
LittleSilver33: I'll take a look and see how I can edit this. Basically the colors gets assigned in here so that it can be accessed through each course. I don't know if this is the best way to do it. As for how the colors came to be, it is before my time. My guess would be that Chris probably got input from people who decided on these colors. Not totally sure.

site/app/models/CalendarInfo.php
LittleSilver33: Okay so update. I tried to remove it and I get a Parsing Error but the error comes from outside of the try-catch yet somehow doesn't cause problems when the try-catch is in the code. I really have no idea why this is a problem and why the code is working when the try-catch is there. I don't see what this changes. If you have any idea please let me know, I'll take another look with fresh eyes to try to figure it out again.

site/app/models/CalendarInfo.php
```diff
@@ -70,7 +70,7 @@ public static function loadGradeableCalendarInfo(Core $core, array $gradeables_o
         $i = 1;
         /** @var Course $course */
         foreach ($courses as $course) {
-            $info->colors[$course->getSemester() . $course->getTitle()] = "var(--category-color-".$i.")";
+            $info->colors[$course->getSemester() . $course->getTitle()] = "var(--category-color-" . $i . ")";
```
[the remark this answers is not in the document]
LittleSilver33: Got it!

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +32,179 @@ class CalendarController extends AbstractController {
...
+            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+        }
+        $calendar_item->setDate(new \DateTime($date));
+        try {
+            $calendar_item->setText($text);
+        }
+        catch (\InvalidArgumentException $e) {
+            $this->core->addErrorMessage("Text exceeds 255 characters which is not allowed");
+            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+        }
```
williamjallen, line 116: Same here.  You technically don't know that the argument is invalid because it's more than 255 characters long.  All you know is that `setText()` received an invalid argument.  Yes, the length limit is currently the only limitation, but we want to remain open to future development in which more error cases could be added.

Just return the error message in the exception.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +32,179 @@ class CalendarController extends AbstractController {
...
+        else {
+            $this->core->addErrorMessage("Invalid or incorrect text given");
+            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+        }
+
+        if (strip_tags($text) !== $text) {
+            $this->core->addErrorMessage("HTML cannot be used in this text");
+            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+        }
```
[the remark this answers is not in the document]
williamjallen: If we do want a check like this, it might be better to move the logic to the model's `setText()` function.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +32,179 @@ class CalendarController extends AbstractController {
...
+        if (isset($_POST['type'])) {
+            $type = $_POST['type'];
+        }
+        else {
+            $this->core->addErrorMessage("Invalid or incorrect type given");
+            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+        }
+
+        if (isset($_POST['date'])) {
+            $date = $_POST['date'];
```
williamjallen, line 140: Do we want to also check that it's a valid date by converting it to a `DateTime` object?

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +32,179 @@ class CalendarController extends AbstractController {
...
+        if (strip_tags($text) !== $text) {
+            $this->core->addErrorMessage("HTML cannot be used in this text");
+            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+        }
+
+        $calendar_item = $this->core->getCourseEntityManager()->getRepository(CalendarItem::class)
+            ->findOneBy(['id' => $id]);
+
+        if ($calendar_item === null) {
+            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
```
williamjallen, line 172: Do we want to add an error message saying why the redirect happened?  If we don't have an error message, it might confuse users who don't understand that an error condition occurred.

site/app/entities/calendar/CalendarItem.php
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * @ORM\Column(type="date")
+     * @var DateTime
+     */
+    protected $date;
+
+    /**
+     * @return int
+     */
+    public function getId(): int {
```
williamjallen, line 48: We don't need a PHPDoc comment if there's already a PHP type declaration.

site/app/entities/calendar/CalendarItem.php
```diff
@@ -0,0 +1,115 @@
...
+            throw new \InvalidArgumentException();
+        }
+        $this->text = $text;
+    }
+
+    public function setType(int $type): void {
+        if ($type === self::ANNOUNCEMENT || $type === self::TEXT) {
+            $this->type = $type;
+        }
+        throw new \InvalidArgumentException();
```
williamjallen, line 83: What exception occurred?  You should specify an exception message.  Apply this throughout the rest of this file.

site/app/entities/calendar/CalendarItem.php
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    public function getTypeString(): string {
+        switch ($this->type) {
+            case self::TEXT:
+                return 'text';
+            case self::ANNOUNCEMENT:
+                return 'ann';
+        }
+        throw new \Exception();
```
williamjallen, line 113: It's a bad idea to just throw a blank exception.  Is there a more specific exception you can use?  A `RuntimeException` comes to mind, but there might be other good options.

site/app/models/CalendarInfo.php
```diff
@@ -43,10 +43,12 @@ class CalendarInfo extends AbstractModel {
      * "title"      => string (title shown at the top of the table),
      * "subtitle"   => string (title shown at the top of the table, if any. Can be empty),
      * "section_id" => string (the id of the section. Will be used as the HTML id)
-     * "gradeables" => array. The structure of this array is same as the element of value of $items_by_date
+     * "gradeables" => array. The structure of this array is same as the element of value of $gradeables_by_date
      */
     private $items_by_sections = [];
 
+    private $colors = [];
```
williamjallen, line 50: A PHPDoc comment specifying the type of this parameter should be added.

site/app/models/CalendarInfo.php
```diff
@@ -59,13 +61,22 @@ class CalendarInfo extends AbstractModel {
...
         $date_format = $core->getConfig()->getDateTimeFormat()->getFormat('gradeable');
 
         $gradeable_list = new GradeableList($core, $core->getUser(), $gradeables_of_user["gradeables"]);
 
-        // get the gradeables from the GradeableList and group them by section
+        $i = 1;
+        /** @var Course $course */
+        foreach ($courses as $course) {
+            $info->colors[$course->getSemester() . $course->getTitle()] = "var(--category-color-$i)";
+            if ($i < 21) {
```
williamjallen, line 74: Where does the magic number 21 come from?  Constants like this are generally a bad thing for code readability.

site/app/models/gradeable/GradeableUtils.php
```diff
@@ -69,20 +73,17 @@ public static function getGradeablesFromCourse(Core $core, Course $course): arra
      * The method loads from the database of all courses and get all gradeables information.
      * Only load once unless the user refreshes the page.
      *
-     * @param Core $core
-     * @param User $user The user to filter gradeables by
      * @return array<string, array<string, Gradeable>|array<string, GradedGradeable>|array<string, Button>>
      * @throws \Exception if a Gradeable failed to load from the database
      */
-    public static function getAllGradeableListFromUserId(Core $core, User $user): array {
+    public static function getAllGradeableListFromUserId(Core $core, User $user, array $courses, array &$calendar_messages): array {
```
williamjallen, line 79: Is there a reason why you're passing `$calendar_messages` by reference here?  References are rather dangerous to use because it isn't immediately obvious to the user that their array might be getting updated without their knowledge.

site/app/models/gradeable/GradeableUtils.php
```diff
@@ -91,4 +92,19 @@ public static function getAllGradeableListFromUserId(Core $core, User $user): ar
...
         return ["gradeables" => $gradeables, "graded_gradeables" => $graded_gradeables, "submit_btns" => $submit_btns];
     }
+
+    /**
+     * This function assumes that you are calling it where Core has a defined course
+     *
+     * @return \app\models\Button[][]|Gradeable[][]|GradedGradeable[][]|array[]
+     * @throws \Exception
+     */
+    public static function getGradeablesFromUserAndCourse(Core $core, array &$calendar_messages): array {
```
williamjallen, line 102: Same comment here about passing `$calendar_messages` by reference.

site/app/models/gradeable/GradeableUtils.php
```diff
@@ -91,4 +92,19 @@ public static function getAllGradeableListFromUserId(Core $core, User $user): ar
         $core->getConfig()->setCourseLoaded(false);
         return ["gradeables" => $gradeables, "graded_gradeables" => $graded_gradeables, "submit_btns" => $submit_btns];
     }
+
+    /**
+     * This function assumes that you are calling it where Core has a defined course
+     *
+     * @return \app\models\Button[][]|Gradeable[][]|GradedGradeable[][]|array[]
+     * @throws \Exception
```
williamjallen, line 100: There's no particular need to add an `@throws` tag for a top-level `Exception`.  It's better to use it if there's a specific exception type potentially being thrown.

site/app/models/gradeable/GradeableUtils.php
```diff
@@ -91,4 +92,19 @@ public static function getAllGradeableListFromUserId(Core $core, User $user): ar
...
+     */
+    public static function getGradeablesFromUserAndCourse(Core $core, array &$calendar_messages): array {
+        $gradeables_of_course = self::getGradeablesFromCourse($core, $core->getConfig()->getSemester(), $core->getConfig()->getCourse(), $calendar_messages, false);
+        $gradeables = $gradeables_of_course["gradeables"];
+        $graded_gradeables = $gradeables_of_course["graded_gradeables"];
+        $submit_btns = $gradeables_of_course["submit_btns"];
+
+        return ["gradeables" => $gradeables, "graded_gradeables" => $graded_gradeables, "submit_btns" => $submit_btns];
```
williamjallen, line 108: This line is pretty long. It might be better to wrap it for readability purposes.

site/public/css/calendar.css
williamjallen: Is there a way to change the CSS to avoid using the `!important`?  Overrides like this can get messy really fast if someone every decides to try to change the CSS again.

williamjallen (changes requested):

site/app/controllers/calendar/CalendarController.php
```diff
@@ -27,8 +32,179 @@ class CalendarController extends AbstractController {
...
+    /**
+     * @Route("/courses/{_semester}/{_course}/calendar")
+     */
+    public function viewCourseCalendar(): WebResponse {
+        $calendar_messages = [];
+        $user = $this->core->getUser();
+        $courses = $this->core->getQueries()->getCourseForUserId($user->getId());
+        $name = $this->core->getConfig()->getCourse();
+        // Initialize $oneCourse and get the specific course for the course calendar
+        $oneCourse = $courses[0];
```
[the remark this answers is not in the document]
LittleSilver33: I thought about that but it isn't possible to view the course calendar page for a given course without being a part of any courses. At least, that's what I figured. I can of course set it to something else or check if the value exists.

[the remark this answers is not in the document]
williamjallen: That seems reasonable.  I don't think there's a need to handle anything gracefully because we'd want an error to show up in the logs if someone was able to access it somehow.

site/app/models/CalendarInfo.php
EdenBendheim: Its for the colors that calendar currently implements:
image

site/app/models/gradeable/GradeableUtils.php
EdenBendheim: It does need to be passed reference in order to be changed in "getGradeablesFromCourse". Added a NOTE to clarify.

williamjallen (changes requested):
In addition to the code changes suggested below, I also have a few UI comments:

1. I think there should be a bit of space between the new item button and the calendar:
image
2. All of these input fields should line up.  I think the message text field should also be larger to accommodate a longer message without the user having to scroll.
image
3. There is no background for the announcement text, so it is not obvious where it is, or even that the user should pay attention to it.
image
4. aaaaaaaaaaaaaaa I thought I selected it and couldn't deselect it or something!  Also, why is the border on the right side thinner?
image
5. How is an important thing different from a note?
image
6. How many notes do I have here?  It's pretty confusing.  I also created a duplicate note (note5) on accident but I can't edit it or delete it.  That's pretty annoying.
image
7. If you click submit with no data in the input fields, it will create a blank item.  You probably want to add a check to make sure the note isn't empty.

sbelsk (changes requested):
Functionaly-wise, the changes look good. A few very minor UI details I noticed that should probably be addressed:
1. When hovering on the items I created in the calendar, the text I see looks like:
image
instead of the course name. (Though that seems to be an issue with other calendar objects, like the gradeables).
2. When I drag my cursor across the border of an announcement, the cursor changes for just a split of a second, as if the border is clickable. If we can figure out what causes it and remove it, that would be nice.

EdenBendheim (commented):
Took a glance, looks good!

EdenBendheim (approved):

PR #6946, opened 2021-08-16
[Feature:InstructorUI] Autograding Status UI

### LINKED ISSUE #6765: Grading Done Web GUI
Currently, instructors can monitor the autograding queue from the ssh command line using the 
grading_done.py script.  The script prints stats once every 5 seconds.  As you leave the script running over 
time, you can estimate the progress and pace.

Instructors must have limited sudo access in order to run this script.
Users who do not have ssh access, or who are not comfortable using using the command line
are not able to check the grading queue progress details.

Let's make a webpage version of this data.
Same information, and also tracking the progress and pace if the page is left open for a period of time.
[NOTE - we will need to be careful, this page might be expensive & a potential large load on the file system since the queue is currently stored as files.  We do eventually want to move the queue to the database, but we aren't there yet.]
It will have some overlap with the work-in-progress docker info page
(capabilities & worker machines)

HenrikLam:
Closes #6765

### What is the new behavior?
New page that allows user to monitor the status of autograding, each tables are updated continuously.

### Other information?

You can test by hitting the regrade button on each gradeable, or by making submissions on a gradeable. There also seems to be gradeable being batch submitted when you do vagrant up. You can also do `sudo systemctl stop submitty_autograding_shipper`
`sudo systemctl stop submitty_autograding_worker` to stop the system from finishing grading the files to inspect the UI closely.


site/app/controllers/AutogradingStatusController.php
```diff
@@ -0,0 +1,72 @@
...
+use app\views\AutogradingStatusView;
+use app\views\ErrorView;
+use Symfony\Component\Routing\Annotation\Route;
+
+class AutogradingStatusController extends AbstractController {
+    /**
+     * @Route("/autograding_status", methods={"GET"})
+     * @return WebResponse | MultiResponse
+     */
+    public function getGradingDonePage() {
```
cjreed121, line 20: You could give this a return type of ResponseInterface.

site/app/controllers/AutogradingStatusController.php
```diff
@@ -0,0 +1,72 @@
...
+        $instructors = $this->core->getQueries()->getActiveUserIds(true, false, false, false, true);
+        $user_id = $this->core->getUser()->getId();
+        if (!in_array($user_id, $instructors)) {
```
cjreed121, line 23: Potentially a quicker check could be to call getInstructorLevelAccessCourse and see if there are any results rather than collecting all instructors and faculty.

site/app/controllers/AutogradingStatusController.php
```diff
@@ -0,0 +1,72 @@
...
+        return new WebResponse(
+            AutogradingStatusView::class,
+            'displayPage',
+            $this->getAutogradingInfo()
+        );
+    }
+
+    /**
+     * Used to continuous in the page's continuous updates
+     * @Route("autograding_status/get_update", methods={"GET"})
```
cjreed121, line 39: Put a / in front of route as all others have that

site/app/libraries/QueueItem.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace app\libraries;
+
+/**
+ * Class QueueItem
+ * Helper class used to interpret the files read in the autograding json files
+ */
+class QueueItem {
+    /** @prop-read @var array */
```
cjreed121, line 10: `@prop-read` does not do anything here as this is not a model. This probably could just be a model rather than a library since it just stores data for you.

site/app/views/AutogradingStatusView.php
```diff
@@ -0,0 +1,18 @@
...
+
+class AutogradingStatusView extends AbstractView {
+    public function displayPage($progress) {
+        $this->core->getOutput()->addInternalJs("autograding-status.js");
+        $this->core->getOutput()->addInternalCss("autograding-status.css");
+        $this->core->getOutput()->addInternalCss('table.css');
+        return $this->core->getOutput()->renderTwigTemplate("AutogradingStatus.twig", [
+            "progress" => $progress
+        ]);
+    }
```
cjreed121, line 17: This should get a breadcrumb because top left just says Submitty

site/public/js/autograding-status.js
```diff
@@ -0,0 +1,132 @@
...
+let interval = null;
+
+function updateTable() {
+    clearInterval(interval);
+    $.ajax({
+        // eslint-disable-next-line no-undef
+        url: buildUrl(['autograding_status', 'get_update']),
+        type: 'GET',
+        // eslint-disable-next-line no-undef
+        data: {'csrf_token': csrfToken},
```
cjreed121, line 15: Rather than telling eslint to ignore these at the top you can put:

/* global buildUrl, csrfToken */

That will tell eslint that it is not in current scope but is global when the browser executes the code.

Also csrfToken is not necessary here as this is a get request. Csrf tokens are only required on post requests.

site/public/js/autograding-status.js
```diff
@@ -0,0 +1,132 @@
...
+        success: function(response) {
+            try {
+                const data = $('#data');
+                let json = JSON.parse(response);
+                if (json.status !== 'success') {
+                    // eslint-disable-next-line no-undef
+                    displayErrorMessage('This login session has expired, please log in again to continue to receive updates');
```
cjreed121, line 22: You can simply just add this to the global list as mentioned above.

cjreed121 (changes requested):

site/app/libraries/GradingQueue.php
```diff
@@ -209,4 +211,220 @@ public function getGradingCount() {
...
+
+            if ($regrade) {
+                $course_info[$file_segments[0] . "__" . $file_segments[1]][$file_segments[2]]["regrade"] += 1;
+            }
+            else {
+                $course_info[$file_segments[0] . "__" . $file_segments[1]][$file_segments[2]]["interactive"] += 1;
+            }
+        }
+        catch (\Exception $e) {
+            $error .= "Could not read for {$queue_or_grading_file}: {$e}";
```
williamjallen, line 421: Do we want the entire stack trace here?

site/app/libraries/GradingQueue.php
```diff
@@ -209,4 +211,220 @@ public function getGradingCount() {
...
+        $regrade = false;
+
+        $elapsed_time = -1;
+
+        $job_file = basename($queue_or_grading_file);
+
+        try {
+            $entry = new QueueItem($queue_or_grading_file, $epoch_time, $is_grading);
+            $regrade = $entry->isRegrade();
+            if ($regrade) {
```
williamjallen, line 340: Is this temporary variable necessary if we're only using it in one spot?

site/app/libraries/GradingQueue.php
```diff
@@ -209,4 +211,220 @@ public function getGradingCount() {
...
+            }
+            else {
+                if ($is_grading) {
+                    $detailed_queue_counts["interactive_ongoing"] += 1;
+                }
+                else {
+                    $detailed_queue_counts["interactive"] += 1;
+                }
+            }
+            $capability = "default";
```
williamjallen, line 356: Can we be certain that there will always be a capability named "default"?  (I'm not too familiar with how this works yet)

site/app/libraries/GradingQueue.php
```diff
@@ -209,4 +211,220 @@ public function getGradingCount() {
...
+                }
+            }
+            else {
+                $max_time = $entry->getQueueObj()["max_possible_grading_time"];
+                $full_machine = $entry->getGradingQueueObj()['machine'];
+
+                $grading_machine = "NONE";
+                foreach ($open_autograding_workers_json as $machine => $details) {
+                    $m = $details["address"];
+                    if ($details["username"] != "") {
```
williamjallen, line 389: `!==` might be better here

site/app/libraries/GradingQueue.php
```diff
@@ -209,4 +211,220 @@ public function getGradingCount() {
...
+                        $m = "{$details['username']}@{$m}";
+                    }
+                    if ($full_machine == $m) {
```
williamjallen, line 392: `===` might be better here

site/app/libraries/QueueItem.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
williamjallen, line 1: Should a few basic unit tests be added for this file?

site/app/templates/AutogradingStatus.twig
```diff
@@ -0,0 +1,192 @@
...
+                <th rowspan="3" class="col-md-2 right-boarder">
+                    TIME STAMP
+                </th>
+                <th colspan="2" style="text-align: left;"  class="col-md-4">
+                    INTERATIVE
+                </th>
+                <th colspan="2" style="text-align: left;"  class="col-md-4 right-boarder">
+                    REGRADE
+                </th>
+                <th class="col-md-2 right-boarder" colspan="{{ progress["machine_grading_counts"]|keys|length }}" style="text-align: left;">
```
williamjallen, line 102: The bootstrap class `.text-left` would do the same thing and remove the need for an inline style.  Same goes for a number of other places in this file.

site/app/templates/AutogradingStatus.twig
```diff
@@ -0,0 +1,192 @@
...
+<div class="content table-content custom-scrollbar">
+    <h1>
+        Autograding Monitor
+    </h1>
+    <div id="data" data-machine-num="{{progress["machine_grading_counts"]|keys|length}}" data-capability-num="{{progress["capability_queue_counts"]|keys|length}}" hidden>
+    </div>
+    <table class="table table-striped table-bordered" id="autograding-status-table">
+        <thead>
+            <tr>
+                <th rowspan="3" class="col-md-2 right-boarder">
```
williamjallen, line 93: Do we need a custom border class here (and elsewhere in this table) if we have the `.table-bordered` bootstrap class?

site/app/templates/AutogradingStatus.twig
```diff
@@ -0,0 +1,192 @@
...
+                    {{ progress["queue_counts"]["interactive"] }}
+                </td>
+                <td>
+                    {{ progress["queue_counts"]["regrade_ongoing"] }}
+                </td>
+                <td class="right-boarder">
+                    {{ progress["queue_counts"]["regrade"] }}
+                </td>
+                {% for machine, count in progress["machine_grading_counts"] %}
+                {% if machine == progress["machine_grading_counts"]|keys|last %}
```
williamjallen, line 174: It might be cleaner if this section was indented a little bit.

site/public/css/autograding-status.css
```diff
@@ -0,0 +1,22 @@
...
+#course-table > *,
+#machine-table > * {
+    text-align: left;
+}
+
+#toggle-btn {
+    margin: 1em 0 1em 0;
+}
+
+#course-table td:nth-of-type(1)::before { content: "Semester"; }
```
williamjallen, line 18: What purpose are these trying to serve?  It seems to me like we're writing HTML in a CSS file here...

site/public/css/server.css
```diff
@@ -2029,3 +2029,15 @@ header > * {
 .mermaid-element {
     width: 100%;
 }
+
+.custom-scrollbar::-webkit-scrollbar {
```
williamjallen, line 2033: Have you tested these to make sure they work on all browsers?  Is there any reason why these are necessary instead of just relying on native OS scrollbars?  Scrollbars are tricky because they look, and act, differently on different systems and browsers.

site/public/js/autograding-status.js
```diff
@@ -0,0 +1,132 @@
...
+                    return;
+                }
+
+                json = json.data;
+
+                // Check to see if any of the machine and capability info are outdated, if so, refresh the page
+                if (data.data('machine-num') != Object.keys(json.machine_grading_counts).length) {
```
williamjallen, line 29: `!==` here perhaps, as well as on line 32?

site/public/js/autograding-status.js
```diff
@@ -0,0 +1,132 @@
...
+                new_row.insertCell().innerHTML = json.queue_counts.interactive_ongoing;
+                new_cell = new_row.insertCell();
+                new_cell.innerHTML = json.queue_counts.interactive;
+                new_cell.className = 'right-boarder';
+                new_row.insertCell().innerHTML = json.queue_counts.regrade_ongoing;
+                new_cell = new_row.insertCell();
+                new_cell.innerHTML = json.queue_counts.regrade;
+                new_cell.className = 'right-boarder';
+                Object.keys(json.machine_grading_counts).forEach((key, i) => {
+                    if (i == Object.keys(json.machine_grading_counts).length - 1) {
```
williamjallen, line 84: `===`?

site/public/js/autograding-status.js
```diff
@@ -0,0 +1,132 @@
...
+                    new_row.insertCell().innerHTML = json.capability_queue_counts[key];
+                });
+                // Check if old logs should be removed to make room for new logs
+                if ($('#autograding-status-table tbody tr').length > max_log) {
+                    // +3 to account for the thead
+                    table.deleteRow(max_log + 3);
+                }
+
+                // Queue this function to be run again after specified delay
+                interval = setInterval(updateTable, refresh_freq);
```
williamjallen, line 103: Instead of using setInterval, it might be better to use setTimeout() and then call that recursively, waiting for requests to come back before sending out the next one.  That way, if requests are taking a long time, we don't spam the server with requests.

site/public/js/autograding-status.js
```diff
@@ -0,0 +1,132 @@
...
+    else {
+        updateTable();
+        // eslint-disable-next-line no-undef
+        displaySuccessMessage('Update has been resumed');
+        $(this).text('Pause Update');
+    }
+}
+
+$(document).ready(() => {
+    interval = setInterval(updateTable, refresh_freq);
```
williamjallen, line 129: Same comment as above regarding setInterval()

williamjallen (changes requested):
Code looks good in general, although I left a few comments throughout.  I would ideally like to see a few unit tests for this.  It's a new page and it would be good to ensure that it stays functional instead of attempting to retrofit tests to the code later.  Perhaps a couple e2e tests to check the page update toggle, as well as basic checking to make sure access controls work as expected for each type of user.

site/app/controllers/AutogradingStatusController.php
```diff
@@ -0,0 +1,72 @@
...
+use app\libraries\response\MultiResponse;
+use app\libraries\response\WebResponse;
+use app\libraries\response\JsonResponse;
+use app\libraries\routers\AccessControl;
+use app\libraries\GradingQueue;
+use app\views\AutogradingStatusView;
+use app\views\ErrorView;
+use Symfony\Component\Routing\Annotation\Route;
+
+class AutogradingStatusController extends AbstractController {
```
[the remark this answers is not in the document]
HenrikLam: As far as I know,`@AccessControl` only checks your access for the current course. However, this page doesn't exist as a course page so `AccessControl` would not work.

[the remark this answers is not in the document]
williamjallen: Hmm, ok.  I'm still a little wary of this because someone could easily come in later and make changes, not knowing that access control had to be done in each function, and unknowingly leave a security vulnerability open.  It's tough though so if there's no better option I'll approve it the way it is.

site/app/libraries/GradingQueue.php
HenrikLam: An entire stack trace might be a bit excessive, I assume you are referring to the appending of `$e`?

site/app/libraries/GradingQueue.php
HenrikLam: Which variable are you referring to? Both `$entry` and `$regrade` are used later in the code.

site/app/libraries/GradingQueue.php
HenrikLam: Default should exist from the submitty setup script, it is possible that someone get rid of it if they go into the system itself however.

site/app/templates/AutogradingStatus.twig
HenrikLam: The border is for the borders between the columns, which does not exist between every column. For example, between the machines and capabilities, each capability is a column but for the sake of association, I did not put a border between them.
image

site/app/libraries/GradingQueue.php
williamjallen: Yeah, I was unclear.  Currently it will print a large stack trace.  What I meant to suggest is that it be switched to `$e->getMessage()` to only print the message and not the entire stack trace.

site/app/templates/AutogradingStatus.twig
HenrikLam: Good point, I never know how twig should look like styling wise. I have seen both indenting and not indenting done but I was never sure.

site/public/css/autograding-status.css
HenrikLam: This allows the table to be displayed properly when the windows become too small. Without these lines, the table displayed will consist of `f21`, `sample`, `grading_homework`, and so on without any of the column names,
image

site/public/css/server.css
HenrikLam: It seems to work for both Chrome and Firefox, I am only using it because I hate the way the native windows scrollbar looks.

site/app/libraries/GradingQueue.php
williamjallen: I was referring to `$regrade` here but I see that it's used in more places than I thought initially so it looks ok.

site/public/js/autograding-status.js
HenrikLam: I considered using `setTimeout()`, but then the button for pausing the timeout becomes tricky, or at least I don't know of a way to get rid of things queued to run with `setTimeout()`. With setInterval, however, I can use the `clearInterval()`.

williamjallen: An easy way to do this would be to use the button to add/remove an attribute somewhere and then check for the existence of that attribute before calling `setTimeout()` again.

site/public/css/autograding-status.css
williamjallen: Oh, ok, sounds good then.  Might want to leave a comment near the table saying that this has been done, just in case someone comes along later and wants to edit this table.

site/app/controllers/AutogradingStatusController.php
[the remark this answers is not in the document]
HenrikLam: Yeah I agree it isn't the greatest solution. I should also note that something similar is also done in the docker UI page, not to justify this but as a note of where else this is happening.

site/public/js/autograding-status.js
HenrikLam: What are the advantages of using `setTimeout()` over `setInterval()`, right now `setInterval()` also allows us to wait for the response to come back before sending another request.

site/app/libraries/QueueItem.php
HenrikLam: Unit tests might be a little tricky since it is reading the queue files which are constantly being grabbed by the `autograding_workers`, it could be done if we first manually stop the shippers and workers though.

cjreed121: Unit tests should not rely on autograding or the website to be functional. Unit tests are run before our e2e tests. Instead you can create very basic json files and call the constructor with that.

site/public/js/autograding-status.js
williamjallen: It would just mean that you wouldn't have to always be clearing/resetting the interval.  If I understand the code correctly, this code essentially uses `setInterval()` to do the job of `setTimeout()`.  `setTimeout()` only runs once after a set amount of time, while the current code sets the interval and runs the function once which clears the interval and then re-sets it.

site/app/controllers/AutogradingStatusController.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Used to continuous in the page's continuous updates
+     * @Route("/autograding_status/get_update", methods={"GET"})
+     * @return JsonResponse
+     */
+    public function getProgress(): JsonResponse {
+        $instructors = $this->core->getQueries()->getActiveUserIds(true, false, false, false, true);
+        $user_id = $this->core->getUser()->getId();
+        if (!in_array($user_id, $instructors)) {
```
cjreed121, line 44: This can also be shortened like mentioned as above.

site/app/controllers/GlobalController.php
```diff
@@ -382,12 +388,16 @@ public function prep_user_sidebar(&$sidebar_buttons) {
             ]);
         }
 
-        // --------------------------------------------------------------------------
-        // FACULTY & SUPERUSERS ONLY
-        if ($this->core->getUser()->accessFaculty()) {
+        // Create the line for all faculties, superusers, and instructors
+        $instructors = $this->core->getQueries()->getActiveUserIds(true, false, false, false, true);
+        if ($this->core->getUser()->accessFaculty() || in_array($this->core->getUser()->getId(), $instructors)) {
```
cjreed121, line 393: Here as well

site/app/libraries/GradingQueue.php
```diff
@@ -209,4 +212,219 @@ public function getGradingCount() {
...
+            "capability_queue_counts" => $capability_queue_counts,
+            "queue_counts" => $queue_counts,
+            "num_autograding_workers" => $workers,
+            "course_info" => $course_info,
+            "ongoing_job_info" => $ongoing_job_info
+        ];
+    }
+
+    // Helper function used to interpret the job files and update the count variables appropriately
+    private function updateDetailedQueueCount(string $queue_or_grading_file, bool $is_grading, int $epoch_time, array &$detailed_queue_counts, array &$capability_queue_counts, array &$machine_grading_counts, array $open_autograding_workers_json, array &$course_info): array {
```
cjreed121, line 328: Maybe make this function header multiple lines? Also where it gets called too seems long so maybe do the same there too.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -3846,6 +3846,16 @@ public function getCourseForUserId($user_id, bool $archived = false): array {
         return $return;
     }
 
+    /**
+     * Get all courses where the user with the specified user_id is assigned as an instructor
+     * @param string $user_id
+     * @return array
+     */
+    public function getInstructorLevelAccessCourse(string $user_id): array {
+        $this->submitty_db->query("SELECT DISTINCT semester, course FROM courses_users WHERE user_id=? AND user_group=1", [$user_id]);
```
cjreed121, line 3855: Why select distinct? This table has a defined primary key on semester, course, and user_id so there is no way for there to be multiple identical entries per user.

site/app/views/AutogradingStatusView.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace app\views;
+
+use app\views\AbstractView;
+use app\models\User;
+use app\models\AutogradingStatusController;
+
+class AutogradingStatusView extends AbstractView {
+    public function displayPage($progress) {
```
cjreed121, line 10: You could put a type hint on $progress.

site/public/js/autograding-status.js
```diff
@@ -0,0 +1,126 @@
...
+// Change this variable to change the max number of entries in the table before the oldest entries are replaced
+const max_log = 60;
+
+let time_id = -1;
+
+function updateTable() {
+    $.ajax({
+        url: buildUrl(['autograding_status', 'get_update']),
+        type: 'GET',
+        data: {'csrf_token': csrfToken},
```
cjreed121, line 13: The csrf token is not necessary as this is a GET request

cjreed121 (changes requested):

site/app/libraries/GradingQueue.php
HenrikLam: Is there a way to make the linter happy without using a line for each parameter? I opted for this way unoriginally because I thought it will become too big with one line each for each parameter.

williamjallen: I don't know of any way to satisfy the linter without either putting them all on one line or each on its own line.

site/public/css/server.css
williamjallen: Ok, I'm still a little wary of it but it does seem to work as expected on safari and chrome.  It does look a little weird in light mode though.  You might want to have the background color update with the theme.

williamjallen (commented):
Most major changes have been addressed beyond the comments Chris left so I think it's close to being ready to merge once those are addressed.  I do still have a couple concerns about relying upon the existence of the "default" capability but perhaps we can discuss that in the meeting tomorrow.  I would also really like to see a little bit of testing where possible to check for things like making sure that students can't access this page for example, as well as some unit testing.  It doesn't necessarily have to be done in this PR but I just want to make sure that we aren't adding yet another page which ultimately doesn't get tests run against it on a regular basis.

Also, the pause updates button doesn't seem to work for me.  Could it be that something was forgotten somewhere when `setInterval()` was changed to `setTimeout()`?

site/app/libraries/GradingQueue.php
cjreed121: I'm pretty sure you would have to put all arguments on separate line. I think that would still look much better than having 1 very long line.

site/app/controllers/GlobalController.php
```diff
@@ -445,7 +447,7 @@ public function prep_user_sidebar(&$sidebar_buttons) {
 
         // --------------------------------------------------------------------------
         // INSTRUCTOR IN ANY COURSE
-        if (in_array($this->core->getUser()->getId(), $instructors)) {
+        if (!empty($this->core->getQueries()->getInstructorLevelAccessCourse($this->core->getUser()->getId()))) {
```
cjreed121: No reason to run this query again if you can just store it in a variable from your query above on 394.

cjreed121 (changes requested):

HenrikLam:
I left out the flag to stop the update thinking stopTimeout would be enough, but it seems sometimes this breaks when we are unlucky with the timing. It seems to work consistently for me now after adding the flag.

cjreed121 (changes requested):

site/app/controllers/AutogradingStatusController.php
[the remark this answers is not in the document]
cjreed121: Maybe just add a comment saying this is a security check. I have seen other places in the code do that when a manual security check was needed.

site/app/controllers/AutogradingStatusController.php
```diff
@@ -0,0 +1,70 @@
...
+        $info = $this->getAutogradingInfo();
+        return JsonResponse::getSuccessResponse($info);
+    }
+
+    // Uses the GradingQueue class to get all the info from the necessary files
+    private function getAutogradingInfo(): array {
+        $gq = new GradingQueue(null, null, $this->core->getConfig()->getSubmittyPath());
+        $info = $gq->getAutogradingInfo($this->core->getConfig()->getConfigPath());
+        $course = [];
+        $tmp = $this->core->getQueries()->getInstructorLevelAccessCourse($this->core->getUser()->getId());
```
cjreed121, line 54: Is there a better name for this besides $tmp?

cjreed121 (changes requested):
Looks good. About to approve but just one last code style comment.

cjreed121 (approved):

williamjallen (approved):

PR #6966, opened 2021-08-22
[Refactor:Plagiarism] Move plagiarism to database

williamjallen:
### What is the current behavior?
Submitty's plagiarism detection system currently only operates on files.  Configuration data is stored in `config.json` files which are then passed to Lichen for processing.  Loading, storing, and managing `config.json` files is somewhat tricky and is a weakness in Plagiarism due to the amount of error checking required.  In addition, it is difficult to implement several upcoming features such as configuration templates and nightly reruns with the current system without having to either introduce new global Lichen files or iterate over all of the `config.json` files.

### What is the new behavior?
This PR moves all configuration info to the database.  When a Lichen run is started, the configuration information is pulled from the database and written to a `config.json` file which is then passed to Lichen.  Moving the configuration information to the database means that it will be much simpler to implement the nightly rerun feature, and the future configuration template feature, as well as slightly speeding up the time it takes to load the plagiarism main page and the plagiarism configuration form.

### Other information?
This is a breaking change which depends upon https://github.com/Submitty/Lichen/pull/61 to work properly.

PR #6975, opened 2021-08-24
[Feature:Plagiarism] Add ability to import existing configs

williamjallen:
Currently, the only way plagiarism configurations may be created is to manually create a configuration and let it run.  This can take a significant amount of time for instructors who may have tens of gradeables they wish to run with similar, or identical, plagiarism configurations.  

### What is the new behavior?
This PR adds the ability to download `config.json` files from the main plagiarism interface, as well as adding the ability to import existing configurations or upload custom `config.json` files.  

Default manual entry option:
ss1

Upload config.json option:
ss2

Import existing config option:
ss3

Plagiarism main page:
ss4

### Other information?
This PR depends upon the changes introduced in https://github.com/Submitty/Submitty/pull/6966 and should be merged only after that PR is merged.


.setup/pip/system_requirements.txt
```diff
@@ -42,7 +42,7 @@ PyPDF2==1.26.0
 distro==1.6.0
 jsonschema==3.2.0
 jsonref==0.2
-docker==5.0.1
+docker==5.0.0
```
MasterOdin, line 45: intentional change?

migration/migrator/data/course_tables.sql
```diff
@@ -13,20 +13,6 @@ SET xmloption = content;
 SET client_min_messages = warning;
```
MasterOdin, line 13: These changes were generated from the migrator `dump` function?

sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
```diff
@@ -168,18 +169,22 @@ def run_job(self):
         course = self.job_details['course']
         gradeable = self.job_details['gradeable']
         config_id = self.job_details['config_id']
+        config_data = self.job_details['config_data']
 
         # error checking
         # prevent against backwards crawling
-        if '..' in semester or '..' in course or '..' in gradeable or '..' in config_id:
```
MasterOdin, line 174: You're removing the `'..'` check here as `config_id` should only be of type `int`. It would be good to add a type check to that effect for `config_id`, in the case someone figures out how to dump arbitrary files into the job directory, they could make `config_id` whatever type or value they wanted.

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -282,35 +285,134 @@ private function saveNewProvidedCode(string $temporary_file_path, string $filena
...
+            ];
+        }
+
+        $lichen_job_file = $this->getQueuePath($gradeable_id, $config_id);
+
+        if (file_exists($lichen_job_file) && !is_writable($lichen_job_file)) {
+            throw new Exception("Error: Failed to create lichen job. Try again");
+        }
+
+        if (!file_put_contents($lichen_job_file, json_encode($lichen_job_data, JSON_PRETTY_PRINT))) {
```
MasterOdin, line 379: Use `FileUtils::writeJsonFile($lichen_job_file, $lichen_job_data)` here.

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -533,161 +636,280 @@ public function saveNewPlagiarismConfiguration(string $new_or_edit, string $grad
...
+        if ($new_or_edit === "new") {
+            try {
+                $config_id = $em->getRepository(PlagiarismConfig::class)
+                    ->findOneBy(["gradeable_id" => $gradeable_id], ["config_id" => "DESC"]);
+                if ($config_id === null) {
+                    $config_id = 1;
+                }
+                else {
+                    $config_id = $config_id->getConfigID() + 1;
+                }
```
MasterOdin, line 663: What is the relation of this `$config_id` versus the one that gets passed into this function?

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -533,161 +636,280 @@ public function saveNewPlagiarismConfiguration(string $new_or_edit, string $grad
...
+                elseif (!isset($data["prior_term_gradeables"])) {
+                    $error_message = "Error: Invalid or missing prior_term_gradeables field";
+                }
+                elseif (!isset($data["ignore_submissions"])) {
+                    $error_message = "Error: Invalid or missing ignore_submissions field";
+                }
 
+                if ($error_message !== "") {
+                    throw new Exception($error_message);
```
MasterOdin, line 711: What other exceptions might we hit within this try/catch? The validation errors seems like something a human might accidentally run into, but other exceptions (e.g. coming from the DB) should probably not be caught so that we (the developers) get an error report of it happening as it shouldn't ever happen.

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -533,161 +636,280 @@ public function saveNewPlagiarismConfiguration(string $new_or_edit, string $grad
...
+                    in_array("results", $data["regex_dirs"]),
+                    in_array("checkout", $data["regex_dirs"]),
+                    $data["language"],
+                    $data["threshold"],
+                    $data["sequence_length"],
+                    $data["prior_term_gradeables"],
+                    $data["ignore_submissions"]
+                );
+                $em->persist($new_config);
+                $em->flush();
```
MasterOdin, line 729: The `->flush()` method should be ideally called as few times as possible, unless there's something specific why you need the DB changes to happen now, as opposed after batching all of them together. It's not clear to me why we couldn't just call this once at the end of the function.

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -864,8 +995,8 @@ public function configurePlagiarismForm(): WebResponse {
         $config["regex_dirs"] = ["submissions"];
         $config["language"] = array_fill_keys(PlagiarismUtils::getSupportedLanguages(), "");
         $config["language"]["plaintext"] = "selected";
-        $config["threshold"] = 5;
-        $config["sequence_length"] = 4;
+        $config["threshold"] = 15;
+        $config["sequence_length"] = 10;
```
MasterOdin, line 999: Can we have these values as class constants with an explanation for why they're the values they are, even if it's just a reference to look in another file.

site/app/entities/plagiarism/PlagiarismConfig.php
```diff
@@ -0,0 +1,336 @@
...
+        $this->setLanguage($language);
+        $this->setThreshold($threshold);
+        $this->setSequenceLength($sequence_length);
+        $this->setOtherGradeables($other_gradeables);
+        $this->setIgnoredSubmissions($ignored_submissions);
+    }
+
+    /**
+     * @return int
+     */
```
MasterOdin, line 148: These docstrings are not useful and are already covered by the type hint on the function. Unless there's meaningful documentation within the comment, there's (in my opinion) no reason to have it as it just means there's two places instead of one we need to update if we were to change the variable's type for some reason.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7615,4 +7615,26 @@ private function getGradeableMinutesOverride(string $gradeable_id): array {
...
+    /**
+     * Gets the total number of submissions made to a given gradeable
+     *
+     * @param string $gradeable_id
+     * @return int
+     */
+    public function getTotalSubmissionsToGradeable(string $gradeable_id): int {
+        $this->course_db->query('SELECT * FROM electronic_gradeable_data WHERE g_id=?', [$gradeable_id]);
+        return count($this->course_db->rows());
+    }
```
MasterOdin, line 7639: You should be doing the `count` within the query itself, not within PHP. This will save on both memory and speed.

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
```diff
@@ -219,6 +262,42 @@
         }
     });
 
+    // UPLOAD CONFIG ///////////////////////////////////////////////////////////
+    function reloadConfigs(gradeable_id) {
+        $("#import-config-config-id").empty();
+        $.each(gradeables_with_plag_configs[gradeable_id], function(indeex, item) {
```
MasterOdin, line 268:
```suggestion
        $.each(gradeables_with_plag_configs[gradeable_id], function(index, item) {
```

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
```diff
@@ -307,38 +386,35 @@
...
+            $('[name="sequence_length"]', form).val('14');
         }
-        else if ($('[name="language"]', form).val() == "plaintext") {
-            $('[name="sequence_length"]', form).val('4');
+        else if ($('[name="language"]', form).val() === "plaintext") {
+            $('[name="sequence_length"]', form).val('14');
         }
-        else if ($('[name="language"]', form).val() == "mips") {
+        else if ($('[name="language"]', form).val() === "mips") {
             $('[name="sequence_length"]', form).val('5');
```
MasterOdin, line 410: Similar to a comment above, it would be good to document as to how these values are getting chosen somewhere.

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
```diff
@@ -307,38 +386,35 @@
...
         }
+    }
+
+    $(document).ready(function() {
```
MasterOdin, line 414: A style thing as we've not made a commitment to it via a style guide, but the going within JS is to use [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) as they're more succinct and don't have the `this` trap (unless you actually do need `this` within the function).

MasterOdin (changes requested):

site/app/controllers/admin/PlagiarismController.php
williamjallen: This particular piece of code is something which has been bothering me for some time but I'm not sure how to do it better.  If this is an edit to an existing config, a `config_id` will be passed to the function.  If this is a new config, we need to create a new `config_id` which is what this code does.

migration/migrator/data/course_tables.sql
williamjallen: yes.  I haven't modified any of the dumped files except by running the migrator as far as I am aware.  I've seen similar changes in other PRs I've reviewed.  I don't know enough SQL to know what purpose these serve or what may be causing the issue.

.setup/pip/system_requirements.txt
williamjallen: Nope, not intentional.  Will revert.  I suspect this may come from when I merged master in and then reverted the change.

site/app/controllers/admin/PlagiarismController.php
williamjallen: Isn't this the only `->flush()` method around?  I thought it would be best to put the `->flush()`near where the only other writing DB interactions occur.  If I'm not mistaken, this is the last (and only) place where writing DB operations occur.

site/app/controllers/admin/PlagiarismController.php
williamjallen: Hmm, tricky.  Perhaps there is a better way to do it but I moved a decent bit of validation to `PlagiarismConfig.php` so that as much input validation as possible is done in that one central place.  The main thing I was trying to catch here were potential validation errors (such as when an invalid `config.json` is uploaded) which should be referred back to the user.  Thoughts?

MasterOdin: Use more specific exception classes, in this case [`ValidationException`](https://github.com/Submitty/Submitty/blob/master/site/app/exceptions/ValidationException.php) seems appropriate. Then where the catches are you would catch just these types of exceptions.

Generally, in PHP (and other languages with typed exceptions), you want to avoid throwing a broad exception or catching a broad exception, preferring to make things as scoped as possible. Python is a better example of why to use narrow exceptions, where any error including syntax errors are thrown exceptions.

site/app/controllers/admin/PlagiarismController.php
MasterOdin: I would suggest moving the `flush()` here to be done after the if/else branching. Calling it having not entered either branch will be a no-op, but otherwise does we want, reducing the number of times we have this line by one. We also don't want to wrap the `flush` in a try/catch since if it fails, that would be a fatal error we want to abort on.

sbelsk (changes requested):
The export feature works well. When I try to test the feature to create a configuration based off of an existing one, (selecting a certain gradeable and config number), the new one that gets created is for a completely different gradeable. And same thing for uploading configs, I upload a config for "Open Team Homework", and a config for "Autograde and TA Homework (C System Calls)" gets created instead. The error checking for config upload seems to work well though.

migration/migrator/data/course_tables.sql
MasterOdin: The changes are fine. It's just shuffling around defaults with the newest versions of postgres. The stuff removed was mostly just compatibility with much older (e.g. postgres 8) versions we don't need to worry about.

williamjallen:
> The export feature works well. When I try to test the feature to create a configuration based off of an existing one, (selecting a certain gradeable and config number), the new one that gets created is for a completely different gradeable. And same thing for uploading configs, I upload a config for "Open Team Homework", and a config for "Autograde and TA Homework (C System Calls)" gets created instead. The error checking for config upload seems to work well though.

Are you certain that this is not the result of user error?  It seems to work as expected for me.  If this was in fact some sort of misunderstanding about how it works, I wonder if there's perhaps a better way to make this obvious on the UI.  The issue here is that if you import a config (using either of the two new methods), it will take the *settings* from the imported config and make a new configuration associated with the selected gradeable.  If I had to guess, you probably left the default gradeable selected ("Autograde and TA Homework (C System Calls)" is the default for the sample course).  This is a tricky interface which obviously caused you some trouble which may be annoying for instructors as well but I can't think of any clearer ways to do it.  Presumably, instructors will learn if they do the wrong thing a few times, so in the worst-case scenario we can just leave it as-is.

site/app/templates/plagiarism/PlagiarismConfigurationForm.twig
```diff
@@ -48,143 +48,185 @@
...
+                        <label for="manual-config-id">Manual configuration</label>
+                    </div>
+                    <div class="plag-data radio-label-pair" style="margin-left: 0;">
+                        <input type="radio" id="upload-config-id" value="upload_config" name="config_option">
+                        <label for="upload-config-id">Upload configuration file</label>
+                    </div>
+                    <div class="plag-data radio-label-pair" style="margin-left: 0;">
+                        <input type="radio" id="import-config-id" value="import_config" name="config_option">
+                        <label for="import-config-id">Use existing configuration</label>
+                    </div>
```
sbelsk, line 65: Perhaps different wording on the UI form might help?
```suggestion
                    <div class="plag-data-label">Parameter Configuration:</div>
                    <div class="plag-data radio-label-pair" style="margin-left: 0;">
                        <input type="radio" id="manual-config-id" value="manual_config" name="config_option" checked>
                        <label for="manual-config-id">Manual configuration</label>
                    </div>
                    <div class="plag-data radio-label-pair" style="margin-left: 0;">
                        <input type="radio" id="upload-config-id" value="upload_config" name="config_option">
                        <label for="upload-config-id">Upload configuration file</label>
                    </div>
                    <div class="plag-data radio-label-pair" style="margin-left: 0;">
                        <input type="radio" id="import-config-id" value="import_config" name="config_option">
                        <label for="import-config-id">Use existing configuration</label>
                    </div>
```
or:
```suggestion
                    <div class="plag-data-label">Configuration Source:</div>
                    <div class="plag-data radio-label-pair" style="margin-left: 0;">
                        <input type="radio" id="manual-config-id" value="manual_config" name="config_option" checked>
                        <label for="manual-config-id">Manual configuration</label>
                    </div>
                    <div class="plag-data radio-label-pair" style="margin-left: 0;">
                        <input type="radio" id="upload-config-id" value="upload_config" name="config_option">
                        <label for="upload-config-id">Upload configuration parameters file</label>
                    </div>
                    <div class="plag-data radio-label-pair" style="margin-left: 0;">
                        <input type="radio" id="import-config-id" value="import_config" name="config_option">
                        <label for="import-config-id">Use parameters from existing configuration</label>
                    </div>
```

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -629,129 +679,262 @@ public function saveNewPlagiarismConfiguration(string $new_or_edit, string $grad
...
+                }
+                elseif (!isset($data["sequence_length"])) {
+                    $error_message = "Error: Invalid or missing sequence_length field";
+                }
+                elseif (!isset($data["prior_term_gradeables"])) {
+                    $error_message = "Error: Invalid or missing prior_term_gradeables field";
+                }
+                elseif (!isset($data["ignore_submissions"])) {
+                    $error_message = "Error: Invalid or missing ignore_submissions field";
+                }
```
sbelsk, line 737: How about something like this to make this shorter:
```suggestion
                $error_message = "";
                $keys = ["version", "regex", "regex_dirs", "language", "threshold", "sequence_length", "prior_term_gradeables", "ignore_submissions"];
                for ($keys as $key) {
                    if (!isset($data[$key])) {
                        $error_message .= "Error: Invalid or missing {$key} field";
                    }
                }
```

sbelsk (approved):

PR #6985, opened 2021-08-29
[Bugfix:InstructorUI] Fix Null Date Issue

cjreed121:

In some cases when changing dates, the backend may try to save dates as null which is not allowed.

### What is the new behavior?
This will make sure no dates are null and also makes sure that the correct information is sent to the backend. When you turn on/off gradeable due date the submission due date field may not get sent which is now fixed.

### DIFF
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -1161,6 +1161,10 @@ class AdminGradeableController extends AbstractController {
             }
         }
 
+        if (!$gradeable->hasDueDate() && $gradeable->hasReleaseDate()) {
+            $gradeable->setHasReleaseDate(false);
+        }
+
         // Set the dates last just in case the request contained parameters that
         //  affect date validation
         if ($date_set) {
--- a/site/public/js/admin-gradeable-updates.js
+++ b/site/public/js/admin-gradeable-updates.js
@@ -108,6 +108,7 @@ function updateDueDate() {
         cont1.hide();
         cont2.hide();
         cont3.hide();
+        $('#has_release_date_no').prop('checked', true);
     }
     else {
         cont.show();
@@ -138,8 +139,10 @@ $(document).ready(function () {
 
     ajaxCheckBuildStatus();
     $('input:not(#random-peer-graders-list,#number_to_peer_grade),select,textarea').change(function () {
-        if ($(this).hasClass('date-radio')) {
+        if ($(this).hasClass('date-radio') && is_electronic) {
             updateDueDate();
+        }
+        if ($(this).hasClass('date-radio')) {
             updateReleaseDate();
         }
         if ($(this).hasClass('ignore')) {
--- a/site/public/js/admin-gradeable-updates.js
+++ b/site/public/js/admin-gradeable-updates.js
@@ -98,6 +98,37 @@ function updateGradeableErrorCallback(message, response_data) {
     updateErrorMessage();
 }
 
+function updateDueDate() {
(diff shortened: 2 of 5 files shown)


site/public/js/admin-gradeable-updates.js
```diff
@@ -107,6 +138,8 @@ $(document).ready(function () {
 
     ajaxCheckBuildStatus();
     $('input:not(#random-peer-graders-list,#number_to_peer_grade),select,textarea').change(function () {
+        updateDueDate();
+        updateReleaseDate();
```
MasterOdin, line 142: Is there any risk to having this run on changing the inputs for non-date fields? Do we even want that, and could we make this only operate on the date fields / `has_due_date_no`?

cjreed121: I'm not sure if there is any risk but I can wrap it in an if statement to make sure we are targeting those specific radio buttons since that is when it matters. I just wanted to make sure these were called before the data was being sent.

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1161,6 +1161,10 @@ private function updateGradeable(Gradeable $gradeable, $details) {
             }
         }
 
+        if (!$gradeable->hasDueDate() && $gradeable->hasReleaseDate()) {
```
MasterOdin, line 1164: Does the front-end allow an instructor to configure their gradeable in this way?

cjreed121: The front end will hide the release date option once there is no due date. If both were selected yes at the same time and you only checked the due date to no then this case can exist I believe. I guess that could be handled in the JavaScript instead if you think that is a better option.

shailpatels (changes requested):
When creating a new checkpoint or numeric gradeable the dates shown to the user are less then what should be I think, for example I am only shown the TA beta testing date on this checkpoint gradeable, : 
image

but pressing the show all dates button the TA/instructor manual grading due date is editable but not shown at first
Neither is the Grades Released Date shown at first

If I then select no for "will this assignment have a due date" I get the same error message 
```
Failed to parse response from server: [object Object] admin-gradeable-updates.js:591:25
    error http://localhost:1511/js/admin-gradeable-updates.js?v=1630377922:591
    jQuery 8
    ajaxUpdateGradeableProperty http://localhost:1511/js/admin-gradeable-updates.js?v=1630377922:567
    <anonymous> http://localhost:1511/js/admin-gradeable-updates.js?v=1630377922:217
    jQuery 9
    <anonymous> http://localhost:1511/js/admin-gradeable-updates.js?v=1630377922:140
    jQuery 13
```
The same issue occurs with numeric gradeables as well, and creating gradeables with no due date never gets saved.

I tested on Chrome and Firfox on MacOs

site/app/controllers/admin/AdminGradeableController.php
MasterOdin: It's good to have a check like this on the backend (in case they do some POST hack or whatever), but the front-end should also be updated so that they can't have `hasDueDate` be false and `hasReleaseDate` true since it's just going to confuse the instructor when it gets auto-changed on save.

shailpatels (approved):
Looks good, I wasn't able to see the saving issue or any console errors after messing with having and not having due dates on different types of gradeables.

Checkpoint and numeric gradeables now have their option to have a due date disabled, I'm guessing to prevent creating a configuration that doesn't make sense

Electronic gradeables look like they're working as expected with and without due dates

Issue #6626, opened 2021-06-18
Create New Gradeable button isn't disabled when it is unable to be clicked


williamjallen:
**Describe the bug**
When no gradeable type is selected, the "Create New Gradeable" button is active.  When clicking on the button, an alert appears to tell the user to select a gradeable type.  

**Expected behavior**
It would be better to disable the button until all of the requirements are met to prevent any confusion.

**Screenshots**
![Untitled](https://user-images.githubusercontent.com/16820599/122563744-92740d80-d012-11eb-89c0-2c24a541d856.png)

Devtinator:
I would love to work on this.

sonalika2001:
Hi @williamjallen, I'd like to work on this issue. This is my first time contributing to open source. However, I've worked on frontend development and I fully understand the issue at hand. Please guide me as to how to get started with setting up my system for the same and where to look for this page in the code. Thank you!

Gulsonish:
Hey @williamjallen , Please check the pull request, I think I have fixed your mentioned bug. please reply me if it correct.

Gulsonish:
Can You guide me @williamjallen with the error like which folder or file is connected with that .twig file for that error or suggest me connected files for that issue, it would be easy to fix the issue in very less time.

PR #6986, opened 2021-08-29
This code fixes "Create New Gradeable" button problem

Gulsonish:
Bug Fixed : disable the button until all of the requirements are met


williamjallen (changes requested):
A few comments:
1. All pull requests be named according to the rules listed [here](https://submitty.org/developer/how_to_contribute).  An example of one good prefix you could add is `[UI/UX:InstructorUI]`.
2. In the other information section, you should list the issue this closes by using one of the GitHub keywords listed [here](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests).
3. Since this is a UI-only change, can you please add some screenshots or screen recordings of before/after so it's clear what exactly this PR changes?
4. While I think this is a good start, it doesn't really fix the issue mentioned in https://github.com/Submitty/Submitty/issues/6626.  The issue here is that the error checking doesn't enable/disable the button.  If you were to make changes which contain invalid input, this PR would enable the button anyway because you made changes.

Gulsonish:
Can You guide me @williamjallen with the error like which folder or file is connected with that .twig file for that error or suggest me connected files for that issue, it would be easy to fix the issue in very less time.

PR #6998, opened 2021-08-31
[Bugfix:Forum] Add Table Support for Markdown

eddiekrystowski:
Currently, tables aren't rendered in markdown
Screen Shot 2021-08-31 at 6 45 02 PM

### What is the new behavior?
Tables now work as expected in markdown
image

### DIFF
--- a/site/app/libraries/Output.php
+++ b/site/app/libraries/Output.php
@@ -8,6 +8,7 @@ use app\libraries\FileUtils;
 use app\models\Breadcrumb;
 use app\views\ErrorView;
 use League\CommonMark\Extension\Autolink\AutolinkExtension;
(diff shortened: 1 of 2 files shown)


williamjallen (approved):
Looks good and seems to work as expected.  My only comment would be about the border color on the table.  For the few example tables I made while testing, the borders seemed a little distracting.  Would a dark gray work better?  See below for an example of what GitHub does for MD tables:

|this|is|a|table|
|-|-|-|-|
|a|b|c|d|

MasterOdin:
@eddiekrystowski Do you plan to make the suggested color change, or should we merge this as is?

eddiekrystowski:
I've just made the change, should be all set to merge now.

PR #7009, opened 2021-09-02
[Bugfix:System] Fix Grade Inquiry Date Validation

cjreed121:

Courses that have grade inquiries enabled on gradeables will produce a date validation error when trying to load the gradeables page.

### What is the new behavior?
The error is now gone and date validation is handled properly for grade inquiries.

### DIFF
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -1088,6 +1088,8 @@ class AdminGradeableController extends AbstractController {
             }
         }
 
+        $regrade_modified = false;
+
         // Apply other new values for all properties submitted
         foreach ($details as $prop => $post_val) {
             // Convert boolean values into booleans
@@ -1143,6 +1145,12 @@ class AdminGradeableController extends AbstractController {
                 }
             }
 
+            if ($prop === 'regrade_allowed') {
+                if ($post_val !== $gradeable->isRegradeAllowed()) {
+                    $regrade_modified = true;
+                }
+            }
+
             // Try to set the property
             try {
                 //convert the property name to a setter name
@@ -1169,7 +1177,7 @@ class AdminGradeableController extends AbstractController {
         //  affect date validation
         if ($date_set) {
             try {
-                $gradeable->setDates($dates);
+                $gradeable->setDates($dates, $regrade_modified);
                 $updated_properties = $gradeable->getDateStrings(false);
             }
             catch (ValidationException $e) {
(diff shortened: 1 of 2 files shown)


site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1088,6 +1088,8 @@ private function updateGradeable(Gradeable $gradeable, $details) {
             }
         }
 
+        $regrade_modified = false;
```
williamjallen, line 1091: Is there any particular reason why this is defined here, ~50 lines from where it is next used?

site/app/models/gradeable/Gradeable.php
```diff
@@ -756,7 +760,7 @@ private function assertDates(array $dates) {
      * @param \DateTime[] $dates Array of dates, indexed by property name
      * @return \DateTime[] Array of dates, indexed by property name
      */
-    private function coerceDates(array $dates) {
+    private function coerceDates(array $dates, $regrade_modified = false) {
```
williamjallen, line 763: Does this new parameter need a type?

site/app/models/gradeable/Gradeable.php
```diff
@@ -692,7 +692,7 @@ private static function validateDateSet(array $date_properties, array $date_valu
      * Gets the dates that require validation for the gradeable's current configuration.
      * @return string[] array of date property names that need validation
      */
-    private function getDateValidationSet() {
+    private function getDateValidationSet($regrade_modified = false) {
```
williamjallen, line 695: Does this new parameter need a type?

williamjallen (changes requested):

site/app/controllers/admin/AdminGradeableController.php
cjreed121: This is the closest out of for loop scope location. I wish it could be closer but I don't want to repeatedly set the variable and want it to have a default value of false.

site/app/models/gradeable/Gradeable.php
cjreed121: It definitely should. I'll go ahead and fix both.

site/app/controllers/admin/AdminGradeableController.php
williamjallen: Oh yeah, I see now.  Perhaps it should have a comment indicating what purpose it serves?

cjreed121: Sure I'll go ahead and throw that in.

williamjallen (approved):

PR #7011, opened 2021-09-03
[Bugfix:System] Add Warning Message to Build

### LINKED ISSUE #6982: Instructors (without sudo) cannot build gradeables from command line
**Describe the bug**
A change this summer that writes notebook gradeable allowed time into the database requires the database password, so instructors can no longer (re-)build gradeables from the command line.  This disrupts the workflow for some power-users.
Note: The build works ok from the web gui because the submitty_daemon user does have access to the password.  

**Expected behavior**
Unfortunately we can't easily/securely allow instructors access to the database password, but we can just skip (with a clear warning, when applicable) this update when gradeables are (re)built from the command line.  Let's modify the code to attempt to get the database password from the file, and if it fails workaround the update of the notebook timings.    Print a big warning if this gradeable has allowed timings and we are skipping the database update.  The core autograding functionality should still be successfully built.  The warning should instruct the user to build from the webpage to properly update this data.  

Alternatively, we could grant instructors limited sudo to execute this build process as sudo with access to the database...  but I don't think that's necessary.  Maybe eventually this is a task for the API...

cjreed121:

If a non-sudo unix user wanted to rebuild the gradeables in the course, the script would fail since they can't access the database password.
Closes #6982

### What is the new behavior?
There is an extra check to see if the user can access the file. If not and they are trying to build a gradeable with allowed minutes, a warning will appear. Warning message put below:

WARNING: You do not have access to set allowed minutes from CLI. Please use website to set that.

### DIFF
--- a/bin/set_allowed_mins.py
+++ b/bin/set_allowed_mins.py
@@ -10,9 +10,6 @@ import datetime
 import os
 import sys
 import json
-import errno
-
-HAS_ERROR = False
 
 try:
     CONFIG_PATH = os.path.join(
@@ -21,22 +18,10 @@ try:
     with open(os.path.join(CONFIG_PATH, 'submitty.json')) as open_file:
         SUBMITTY_CONFIG = json.load(open_file)
 
-    try:
-        with open(os.path.join(CONFIG_PATH, 'database.json')) as open_file:
-            DATABASE_CONFIG = json.load(open_file)
-    except IOError as e:
-        if e.errno == errno.EACCES:
-            HAS_ERROR = True
-
 except Exception as config_fail_error:
     print("[{}] ERROR: CORE SUBMITTY CONFIGURATION ERROR {}".format(
         str(datetime.datetime.now()), str(config_fail_error)))
     sys.exit(1)
-
-if not HAS_ERROR:
-    DB_HOST = DATABASE_CONFIG['database_host']
-    DB_USER = DATABASE_CONFIG['database_user']
-    DB_PASSWORD = DATABASE_CONFIG['database_password']
 
 CONFIG_FILE_PATH = sys.argv[1]
 SEMESTER = sys.argv[2]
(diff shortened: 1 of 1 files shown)


bmcutler:
Sorry, this is not the fix I was looking for.  
Rather than exiting the build, build should proceed, but skip the database write.

cjreed121:
> Sorry, this is not the fix I was looking for.
> Rather than exiting the build, build should proceed, but skip the database write.

What I changed should flag a warning if the gradeable that it is building contains allowed minutes. It is only a warning and it should continue running the rest of the build. I tested it on my vagrant and it worked as expected. Did it not do that for you?

bin/set_allowed_mins.py
```diff
@@ -18,17 +21,22 @@
     with open(os.path.join(CONFIG_PATH, 'submitty.json')) as open_file:
         SUBMITTY_CONFIG = json.load(open_file)
 
-    with open(os.path.join(CONFIG_PATH, 'database.json')) as open_file:
-        DATABASE_CONFIG = json.load(open_file)
+    try:
+        with open(os.path.join(CONFIG_PATH, 'database.json')) as open_file:
+            DATABASE_CONFIG = json.load(open_file)
+    except IOError as e:
+        if e.errno == errno.EACCES:
```
MasterOdin, line 28: Why would other `IOError`s be allowed to happen and the script keeps going? If _any_ exception occurs on reading the script, we need to abort since we cannot load the database details.

Perhaps more appropriate would be:

```python
    except IOError as e:
        if e.errno == errno.EACCES:
            HAS_ERROR = True
        else:
            raise
```

such that we re-raise the exception in other cases and then let the script crash out below? But I'm not well-versed enough on types of `IOError` to know whether or not we should care enough about the possible reasons it gets thrown or just group all `IOError` under an umbrella of "could not open file due to permissions".

MasterOdin (changes requested):

bmcutler:
@cjreed121 please see Matt's comments

MasterOdin (approved):

PR #7016, opened 2021-09-03
[Bugfix:CourseMaterials] Make Course Materials items clickable

### LINKED ISSUE #6992: Course Materials items are not clickable
**Describe the bug**
When hovering the mouse over Course Materials items, the pointer changes to a hand hinting that the item is clickable but when you click it nothing happens.

**Expected behavior**
Course Materials items should open when you click on them.

**To Reproduce**
1. Navigate to the Course Materials page. In my specific example, this is course CSCI-2500. The corresponding URL is https://submitty.cs.rpi.edu/courses/f21/csci2500/course_materials 
2. Expand Labs and then Lab 01.
3. Hover the mouse over the name of the file "lab01.pdf". The mouse pointer changes to a hand but when you click the mouse, nothing happens. See the screenshot. Clicking on the floppy icon and the download button works as expected. The problem is only when clicking the name of the file itself.

**Configuration**
 - Browser: Mozilla Firefox 91.0.2 (64-bit)

**Screenshots**
1nKo4CtjzW

**Additional context**
N/A

loadedskiffer:
closes #6992 

### What is the new behavior?
Clicking the name of a course material pops it open in a new tab.
The name of the item and the icon are now effectively one element.

### Other information?
 I had to do some weird stuff with JavaScript to get the colors to change on hover correctly because of global settings and because we need the icon and the text to act as one element.


site/app/templates/course/CourseMaterials.twig
```diff
@@ -64,11 +64,12 @@
...
                 {% else %}
                     <i class="fas fa-file" style="vertical-align: text-bottom;"></i>
-                    {{ name }}
                 {% endif %}
                 {% set extension = name|split('.')|last|lower %}
                 {% if '.' ~ extension in ['.pdf', '.jpg', '.jpeg', '.c', '.cpp', '.s', '.twig', '.py', '.java', '.png', '.txt', '.h', '.html', '.php', '.js', '.sql', '.sh', '.md', '.csv', '.salsa', '.erl', '.oz', '.pl', '.hs', '.gif'] %}
-                    <a target="_blank" href="{{ display_file_url }}?course_material_id={{ course_material.getId }}" aria-label="Pop up {{ name }} in a new window" tabindex="0"><i class="fas fa-window-restore" title="Pop up the file in a new window"></i></a>
+                    <a class="popout-item" onmouseover="hoverColor(this)" onmouseout="defaultColor()" target="_blank" href="{{ display_file_url }}?course_material_id={{ course_material.getId }}" aria-label="Pop up {{ name }} in a new window" tabindex="0" style="text-decoration: none;">{{ name }} <i class="fas fa-window-restore" title="Pop up the file in a new window"></i></a>
```
MasterOdin, line 70: You can avoid the JS here by adding the following CSS to `course-materials.css`:

```css
a.popout-item:hover *,
a.popout-item > .fas:hover {
    color: var(--external-link-color);
}
```

So, the first line of the new rule is that we apply the color to the `a` tag itself and all of its children. This gets us the color when we hover over the name of the item. With just the first line, then hovering over the icon itself gives a different color. This is due to the specificity of the following:

https://github.com/Submitty/Submitty/blob/5b5b58ec85374f11f204598eae7397213d75bb06/site/public/css/links.css#L19-L22

That's where the second rule comes in above, which has greater specificity than the latter rule, so gives us that nice external link blue on hover now for both the text and the icon. We can now safely get rid of the JS as well.

MasterOdin (changes requested):

site/public/css/course-materials.css
```diff
@@ -14,3 +14,8 @@
     color: var(--standard-dark-green);
     float: none;
 }
+
+a.popout-item:hover *,
+a.popout-item:hover {
```
MasterOdin, line 19: Looking at this, I think we want the selectors to be:
```suggestion
a.popout-item:hover,
a.popout-item:hover *,
a.popout-item > .fas:hover {
```

You're right we need the `a.popout-item:hover`, which I missed in my original suggestion and just happened to overlapped the color in `.file-viewer a:hover` so I didn't see that we needed this.

MasterOdin (changes requested):

site/public/css/course-materials.css
loadedskiffer: yup it looks good now

MasterOdin (approved):

PR #7020, opened 2021-09-04
[Feature:System] Add web socket failure banner message

### LINKED ISSUE #7018: Warning banner that websockets are malfunctioning
A recurring problem in system administration has been a misconfigured or crashed websocket server.
Unfortunately, this problem is not immediately apparent to a quick test of the system.
It is only apparent to a user awaiting dynamic update on the queue or ta grading pages.

If the client cannot connect to the websocket server, let's add a warning banner on the top of pages that use websockets.  Draft message something approximately like "Warning:  Failed to connect to websockets server.  This page will not dynamically update.  Manually reload this page to see updates.  Notify the sysadmin INSERT SYSADMIN EMAIL ADDRESS."

cjreed121:

Currently the only way to tell that your browser did not connect to the web socket server is by looking in the Javascript console. There is also an error in the websocket.js file that I fixed in this PR. When the socket tries to reconnect it does to with a page name of the full URL rather than the actual page name it initially tried to connect with.
Closes #7018 

### What is the new behavior?
If there was a connection failure, a message will appear saying that the websocket server is not working right now and to contact sysadmin. If while on the same page instance the connection opens back up then the message will go away and continue as expected. Now the socket will reconnect with the appropriate page name rather than the URL.

### DIFF
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -223,6 +223,9 @@
                 {{ system_message }}
             </div>
             {% endif %}
+            <div class="system-message warning" id="socket-server-system-message" hidden>
+                You are not connected to socket server. You will not receive live updates. Please notify system administrator if issue persists.
+            </div>
             <div id='messages'>
                 <!-- If the below <div> definition changes, update the design in server.js in displayMessage function -->
                 {% for message in messages %}
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -225,7 +225,7 @@
             {% endif %}
             <div class="system-message warning" id="socket-server-system-message" hidden>
                 Warning: Failed to connect to websocket server. This page will not dynamically update.
-                Manually reload this page to see updates.{% if sysadmin_email is defined %} Notify the
+                Manually reload this page to see updates.{% if sysadmin_email is not empty %} Notify the
                 sysadmin at <a href='mailto:{{ sysadmin_email }}'>{{ sysadmin_email }}</a>.{% endif %}
             </div>
             <div id='messages'>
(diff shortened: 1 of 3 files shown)


site/app/templates/GlobalHeader.twig
```diff
@@ -223,6 +223,10 @@
                 {{ system_message }}
             </div>
             {% endif %}
+            <div class="system-message warning" id="socket-server-system-message" hidden>
+                Warning: Failed to connect to websocket server. This page will not dynamically update.
+                Manually reload this page to see updates. Notify the sysadmin {{ sysadmin_email }}.
```
MasterOdin, line 228: We should check that `sysadmin_email` is non-empty before showing the "Notify the sysadmin" bit. Additionally, I would say change the final sentence te:

```
Notify the sysadmin at <a href='mailto:{{ sysadmin_email }}'>{{ sysadmin_email }}</a>.
```

To make use of the mailto tag, minimally for screen readers if not others.

MasterOdin (changes requested):

site/app/templates/GlobalHeader.twig
```diff
@@ -223,6 +223,11 @@
                 {{ system_message }}
             </div>
             {% endif %}
+            <div class="system-message warning" id="socket-server-system-message" hidden>
+                Warning: Failed to connect to websocket server. This page will not dynamically update.
+                Manually reload this page to see updates.{% if sysadmin_email is defined %} Notify the
```
MasterOdin, line 228: I believe we want to use the [`empty`](https://twig.symfony.com/doc/3.x/tests/empty.html) test clause: `sysadmin_email is empty`

From my understanding `$this->core->getConfig()->getSysAdminEmail()` should always minimally return the empty string.

MasterOdin (changes requested):

MasterOdin (approved):

PR #7023, opened 2021-09-06
[Bugfix:TAGrading] Fix anon mark statistics

### LINKED ISSUE #6810: Other users with mark are identified by hash instead of name
**Describe the bug**
When viewing the rubric panel in the TA grading interface, it is possible to click an icon on the right side of each mark to see the other users who also received that mark.  When not in blinded grading or anonymous mode, the names of the users who received the mark are still displayed as random strings instead of the name or user ID of the user.

**Expected behavior**
When not in a grading mode which hides student names, the random string should be replaced with either the name or user ID of the student.

**To Reproduce**
1. Go to the TA grading interface
2. Start grading a student and switch to the rubric pane
3. Click the users icon next to any mark

**Configuration**
N/A

**Screenshots**
Untitled

joekrystowski:
Fixes #6810 
Previously, when displaying users who got the same mark, users were kept anonymous whether anonymous mode was on or not.

image

Now, the anonymity of the users will be displayed depending on Anon mode and gradable settings for limited access graders.

Screen Shot 2021-09-05 at 9 21 35 PM


cjreed121 (changes requested):
This definitely seems to fix the issue visually. The one issue I see with this PR is the links that are created for each user go by their actual ID rather than their anonymous ID. When clicking on that URL it takes you to that person's page for grading but try clicking the next/previous student buttons. You will see that it creates a Submitty error. This will happen whenever you visit the grading page by actual ID rather than anonymous ID. I would try to see if there is a way you can use their anonymous ID in the link while keeping the text of the link their actual ID.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -2833,6 +2833,8 @@ private function getMarkStats(Mark $mark, User $grader) {
         // Show all submitters if grader has permissions, otherwise just section submitters
         $submitter_ids = ($grader->accessFullGrading() ? $all_submitter_ids : $section_submitter_ids);
 
+        $submitter_anon_ids = ($anon != 'unblind') ? $submitter_ids : $submitter_anon_ids = $this->core->getQueries()->getAnonId($submitter_ids);
```
cjreed121, line 2836: Why do you set $submitter_anon_ids again in the last part of the ternary operator? You could remove setting that to this below because the result of the ternary operator will also set $submitter_anon_ids.
```suggestion
        $submitter_anon_ids = ($anon != 'unblind') ? $submitter_ids : $this->core->getQueries()->getAnonId($submitter_ids);
```

cjreed121 (changes requested):
Seems to work well now. Just 1 more code style change.

site/app/controllers/grading/ElectronicGraderController.php
joekrystowski: > Why do you set $submitter_anon_ids again in the last part of the ternary operator? You could remove setting that to this below because the result of the ternary operator will also set $submitter_anon_ids.

Oops, no idea why I did that.

cjreed121 (approved):

PR #7030, opened 2021-09-06
[Feature:InstructorUI] Add Remove column to Peer matrix

### LINKED ISSUE #5782: Admin Edit Peer Grading Matrix Trash Icon
This would be an improvement upon the feature mentioned in Issue #5364 and implemented in PR's #5416 and #5722 .

At present, clicking the pencil icon for a grader on the Peer Matrix table shows a popup that allows removing the grader entirely. Instead of the option being in the popup, we can create a trash icon columns in the Peer Matrix table, similar to the existing pencil icon column for removing the grader.

On a related note, we can also improve the UI by not allowing instructors to be added as a grader or a student into the Peer Matrix.

loadedskiffer:
closes #5782

### What is the new behavior?
I added a "remove" column on the peer matrix for removing a grader.
pr15_1
The actual functionality of removing a grader hasn't changed, it's just easier to do so now.


HenrikLam:
Minor css suggestion but can we also add the borders to `.table th`?

HenrikLam (changes requested):
Works well, could we change the alert message for removing a grader though? It says that the changes have been made before we press submit even though this isn't reflected on the user's end which I thought was a bit confusing. Maybe something like a simple 

> Are you sure you want to remove $grader_to_edit from the grading matrix


>

HenrikLam (changes requested):

HenrikLam (approved):
Looks good to me!

PR #7032, opened 2021-09-07
[Testing:Autograding] Autograding status test

### LINKED ISSUE #7007: Creating tests for grading status page
Currently the grading status has no testing to ensure correctness of information displayed 

Create test data for autograding and use cypress or selenium to perform e2e testing. Ideally this will be added to the github actions.

**Additional context**
Will need to disable the daemon jobs to ensure consistent results.

HenrikLam:

Closes #7007

There are no tests for the autograding status page. Also the was a bug that I found that attempts to read a non existent entry in a JSON.

### What is the new behavior?
Test implemented in cypress, JSON bug fixed.


site/cypress/integration/autograding_status_1.spec.js
```diff
@@ -0,0 +1,97 @@
...
+            cy.wait(500);
+            cy.visit(autograding_status_path);
+        });
+
+        it('Should start at pause update', () => {
+            cy.get('#toggle-btn').should('have.text', 'Pause Update').click().should('have.text', 'Resume Update');
+            cy.get('.alert-success').invoke('text').should('contain', 'Update has been stopped');
+
+            // Check that the table isn't gaining new entries
+            cy.wait(5000);
```
williamjallen, line 21: Is this a long enough wait?

site/cypress/integration/autograding_status_1.spec.js
```diff
@@ -0,0 +1,97 @@
...
+            cy.get('#autograding-status-table tbody tr').eq(1).should('not.exist');
+
+            cy.get('#toggle-btn').should('have.text', 'Resume Update').click().should('have.text', 'Pause Update');
+            cy.get('.alert-success').invoke('text').should('contain', 'Update has been resumed');
+
+            // Check that the table is gaining new entries
+            cy.wait(5000);
```
williamjallen, line 28: Same comment here about the wait time

site/cypress/integration/autograding_status_1.spec.js
```diff
@@ -0,0 +1,97 @@
...
+            cy.get('#autograding-status-table tbody tr td').eq(3).then(element => cy.get(element).should('contain', '0'));
+            cy.get('#autograding-status-table tbody tr td').eq(4).then(element => cy.get(element).should('contain', '101'));
+            cy.get('#autograding-status-table tbody tr td').eq(5).then(element => cy.get(element).should('contain', '0'));
+            cy.get('#autograding-status-table tbody tr td').eq(6).then(element => cy.get(element).should('contain', '0'));
+            cy.get('#autograding-status-table tbody tr td').eq(7).then(element => cy.get(element).should('contain', '0'));
+            cy.get('#autograding-status-table tbody tr td').eq(8).then(element => cy.get(element).should('contain', '0'));
+            cy.get('#autograding-status-table tbody tr td').eq(9).then(element => cy.get(element).should('contain', '0'));
+            cy.get('#autograding-status-table tbody tr td').eq(10).then(element => cy.get(element).should('contain', '102'));
+
+            cy.logout();
```
williamjallen, line 94: You might think about making a function which runs after each test and logs out the user instead of having to do it in each test.

site/cypress/integration/autograding_status_2.spec.js
```diff
@@ -0,0 +1,20 @@
...
+        before(() => {
+            cy.visit('/');
+            cy.login();
+            cy.wait(500);
+            cy.visit(autograding_status_path);
+        });
+
+        it('should show the gradeables being grading', () => {
+            cy.viewport(1920,1200);
+            cy.get('#machine-table tbody tr').eq(4).should('exist');
```
williamjallen, line 17: Should the user be logged out eventually?

williamjallen (changes requested):
Looks good in general.  Another test you could add is a test to check that the various permissions levels work correctly.

site/cypress/integration/autograding_status_1.spec.js
```diff
@@ -91,7 +94,15 @@ skipOn(Cypress.env('run_area') === 'CI', () => {
...
+        });
+
+        it('should only allow instructor level users', () => {
+            // attempt to visit page as student
+            cy.visit([]);
+            cy.login('student');
+            cy.visit(autograding_status_path);
+            cy.get('#autograding-status-table').should('not.exist');
+            cy.visit('/');
         });
```
williamjallen: Should checks be added for graders and TAs?  Are there any differences between regular instructors and faculty which should be tested?

williamjallen (changes requested):

site/cypress/integration/autograding_status_1.spec.js
HenrikLam: Right now the page is available for users that are made an instructor for any course. However, there is no users that are instructors and not faculty in our set up right now (?).

williamjallen: Ok, that seems like something to add to the sample course in the future but not necessarily in this PR then.  Perhaps just check the access permissions for graders and TAs and then I think it should be good to go.

williamjallen (approved):

PR #7036, opened 2021-09-07
[Bugfix:Forum] Fix Error Creating Categories on Firefox

### LINKED ISSUE #6991: Problems Adding Forum Categories
**Describe the bug**
When adding forum categories in Firefox (windows 10 64-bit, 91.0.2), after trying to add a category a brief error `Something went wrong while trying to add a new category. Please try again` shows up as a popup box, but quickly disappears due to the page reloading. The category is not added.

In Chrome (Windows 10 64-bit, 92.0.4515.159), after trying to add a category, there is no error but the new category does not appear (while still on the Manage Categories page). If the page is manually reloaded, the new category is there.

**Expected behavior**
Regardless of browser, clicking Add Category should successfully add the new category. Ideally it would show up immediately after clicking -- the page seems to refresh.

**To Reproduce**
1. Go to a course's Course Settings
2. Click Customize Categories
3. Type in a new category name (`HW1` for example)
4. Click Add Category
5. See error in Firefox or lack of new category until refresh in Chrome.

eddiekrystowski:
Fixes #6991 
Currently, attempting to create forum categories on firefox fails. An alert saying "Something went wrong" shows up briefly before the page is reloaded. The category is not created.

Also, when changing the color of a category in any browser, even when successful, it shows "Successfully updated category "null"!"

### What is the new behavior?
* Creating forum categories now works properly in firefox
* Categories page DOM traversal has been slightly refactored to prevent accidental breaking in the future
* The correct category name is shown when the color is updated
* The category rank is now given an initial value in the database, before it was NULL until the categories were rearranged.

### DIFF
--- a/site/app/templates/forum/ShowCategories.twig
+++ b/site/app/templates/forum/ShowCategories.twig
@@ -16,7 +16,7 @@
         <input type="hidden" name="csrf_token" value="{{ csrf_token }}" />
         <span  style="float: right;">
                 <input id="new_category_text" aria-label="New Category" placeholder="New Category" style="resize:none;" type="text" name="new_category" maxlength="50" onkeyup="checkInputMaxLength($(this))" />
-                <button type="button" title="Add new category" onclick="addNewCategory('{{ csrf_token }}'); window.location.reload();" style="margin-left:10px;" class="btn btn-primary btn-sm key_to_click" tabindex="0">
+                <button type="button" title="Add new category" onclick="addNewCategory('{{ csrf_token }}');" style="margin-left:10px;" class="btn btn-primary btn-sm key_to_click" tabindex="0">
                     <i class="fas fa-plus-circle fa-1x"></i> Add category
                 </button>
             </span>
--- a/site/app/templates/forum/ShowCategories.twig
+++ b/site/app/templates/forum/ShowCategories.twig
@@ -30,16 +30,21 @@
         <ul id="ui-category-list" style="padding-left: 0em;">
             {# TODO: scrollbar #}
             {% for category in categories %}
-                <li id="categorylistitem-{{ category.category_id }}" class="category-sortable catergory-item-box" style="color: {{ category.color }};">
+                <li id="categorylistitem-{{ category.category_id }}" 
+                    class="category-sortable category-item-box" 
+                    style="color: {{ category.color }};"
+                    data-category_desc={{ category.category_desc }}
+                    data-category_id={{ category.category_id }}
+                >
                     <i class="fas fa-bars handle" aria-label="Drag to reorder" title="Drag to reorder"></i>
                     <span class="categorylistitem-desc">
-                            <span style="word-wrap: break-word;">{{ category.category_desc }}</span>
-                            <a class="post_button" title="Edit Category Description" aria-label="Edit Category Description"><i class="fas fa-edit edit-category-button"></i></a>
+                            <span class="categorylistitem-desc-text" style="word-wrap: break-word;">{{ category.category_desc }}</span>
+                            <a class="post_button" title="Edit Category Description" aria-label="Edit Category Description"><i class="fas fa-edit category-button edit-category-desc-button"></i></a>
                         </span>
                     <span class="categorylistitem-editdesc" style="display: none;">
-                            <input type="text" aria-label="New Description of Category" placeholder="New Description of Category" class="edit-catergoty-name" style="padding: 0;" maxlength="50" onkeyup="checkInputMaxLength($(this))">
-                            <a class="post_button" title="Save Changes" aria-label="Save Changes"><i class="fas fa-check edit-category-button"></i></a>
-                            <a class="post_button" title="Cancel Changes" aria-label="Cancel Changes"><i class="fas fa-times edit-category-button"></i></a>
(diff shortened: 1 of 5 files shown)


jenaybarela:
I ran into a bug while testing. When changing the color of the categories it causes the name to be displayed wrong. When you refresh the name is back to normal.

https://user-images.githubusercontent.com/25368899/135346867-5ed8cc2c-c20b-47d5-af1b-acd0a2f03676.mp4

jenaybarela (changes requested):
I ran into a bug while testing. When changing the color of the categories it causes the name to be displayed wrong. When you refresh the name is back to normal. Other than that it looks good.

https://user-images.githubusercontent.com/25368899/135347367-7c029937-48fa-4f9e-bf2f-e1e3ee0af9b6.mp4

site/app/templates/forum/ShowCategories.twig
```diff
@@ -67,33 +72,40 @@
...
             deleteCategory(category_id, category_desc, '{{ csrf_token }}');
         });
-        $("#ui-category-list").find(".fa-edit").click(function() {
-            var item = $(this).parent().parent().parent();
-            var category_desc = item.find(".categorylistitem-desc span").text().trim();
+
+        $(".edit-category-desc-button").click(function() {
+            const item = getItem(this);
+            const category_desc = item.data('category_desc');
+            console.log(item, category_desc)
```
cjreed121, line 90: This should be removed.

site/public/js/forum.js
```diff
@@ -1321,15 +1323,16 @@ function deleteCategory(category_id, category_desc, csrf_token){
     })
 }
 
-function editCategory(category_id, category_desc, category_color, csrf_token) {
+function editCategory(category_id, category_desc, category_color, changed, csrf_token) {
+    console.log(arguments);
```
cjreed121, line 1327: This should be removed.

site/app/templates/forum/ShowCategories.twig
```diff
@@ -67,33 +72,40 @@
...
 
-        $("#ui-category-list").find(".fa-check").click(function() {
-            var item = $(this).parent().parent().parent();
-            var category_id = parseInt(item.attr('id').split("-")[1]);
-            var category_desc_original = item.find(".categorylistitem-desc span").text().trim();
-            var category_desc = item.find("input").val().trim();
+        $(".save-desc-button").click(function() {
+            const item = getItem(this);
+            const category_id = item.data('category_id');
+            const category_desc_original = item.data('category_desc');//item.find(".categorylistitem-desc span").text().trim();
```
cjreed121, line 105: Should we remove that extra commented code?

cjreed121 (changes requested):
Definitely seems to fix the issue.

I found a bug when deleting and then creating a new category. Normally the rank will be 0, 1, 2, 3. If you delete any item besides the last rank and add a new item you will get duplicate ranks in the database. I'm guessing this is related to how the ranks are generated and might be unrelated to this PR. I think it wasn't really noticed before since the rank always defaulted to null. My forum rewrite might also just end up tackling this as it's not really a priority since it doesn't cause anything to break.

Also in addition to the bug @jenaybarela found: If you press the edit button the same issue will happen. It seems to be cutting off at the first space.

site/app/templates/forum/ShowCategories.twig
eddiekrystowski: Ah yes, sorry I forgot to clean up once I was done.

eddiekrystowski:
> I think it wasn't really noticed before since the rank always defaulted to null. My forum rewrite might also just end up tackling this as it's not really a priority since it doesn't cause anything to break.

@cjreed121 I think it may be best to just leave for the forum refactor since this page is likely to get a good amount of changes (its quite messy right now). If you want, I can either revert the rank behavior to the previous NULL behavior, or just leave it as is and let you take a stab at it in the refactor.

cjreed121:
> > I think it wasn't really noticed before since the rank always defaulted to null. My forum rewrite might also just end up tackling this as it's not really a priority since it doesn't cause anything to break.
> 
> @cjreed121 I think it may be best to just leave for the forum refactor since this page is likely to get a good amount of changes (its quite messy right now). If you want, I can either revert the rank behavior to the previous NULL behavior, or just leave it as is and let you take a stab at it in the refactor.

Sounds good. You can leave your changes as I think it is better for them to have some default value other than NULL.

cjreed121 (changes requested):
I found a tiny UI bug. If you edit the name of a category, save it, and then change the color, the name will revert back to the original name. If you refresh it will correct this and the new name will show. Tested on Linux Firefox.

eddiekrystowski:
> I found a tiny UI bug. If you edit the name of a category, save it, and then change the color, the name will revert back to the original name. If you refresh it will correct this and the new name will show. Tested on Linux Firefox.

Ah yes, thanks for testing this. This page is in dire need of a real refactor, these bugs are mainly coming from data attributes being set/read instead of the actual value of the category name, so in places that it is not set properly, it reads old values.

Should be all good now.

cjreed121 (approved):
Looks good. Just make sure to fix the tests.

PR #7041, opened 2021-09-08
[Feature:SubminiPolls] Add student histogram visibility option

BenKelly99:
Add option for instructors to enable student to be able to see the histogram of responses for a poll. This option is toggleable for each poll and has three options:

1. The students can see the histogram as long as the poll is visible.
2. The students can see the histogram as long as the poll has ended.
3. (Default) The students cannot see the histogram.

A button to view these results is added, if available, when a student selects a poll and clicks "View Poll."

The setting for histogram visibility is included in exporting and importing of poll data.

Tests have been modified and added appropriately. Running unit tests gives no errors that are related to changes in this PR.

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -778,7 +778,8 @@ CREATE TABLE public.polls (
     status text NOT NULL,
     release_date date NOT NULL,
     image_path text,
-    question_type character varying(35) DEFAULT 'single-response-multiple-correct'::character varying
+    question_type character varying(35) DEFAULT 'single-response-multiple-correct'::character varying,
+    student_histogram_release_setting text NOT NULL
 );
 
 
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -779,7 +779,7 @@ CREATE TABLE public.polls (
     release_date date NOT NULL,
     image_path text,
     question_type character varying(35) DEFAULT 'single-response-multiple-correct'::character varying,
-    student_histogram_release_setting text NOT NULL
+    release_histogram character varying(10) NOT NULL
 );
 
 
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -422,9 +422,7 @@ class PollController extends AbstractController {
         }
         if (!in_array($_POST["student_histogram_release_setting"], PollUtils::getStudentHistogramReleaseSettings())) {
             $this->core->addErrorMessage("Invalid student histogram release setting");
-            return MultiResponse::RedirectOnlyResponse(
-                new RedirectResponse($this->core->buildCourseUrl(['polls']))
-            );
+            return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
         $file_path = null;
         if (isset($_FILES['image_file']) && $_FILES["image_file"]["name"] !== "") {
--- a/site/app/controllers/PollController.php
+++ b/site/app/controllers/PollController.php
@@ -145,7 +145,7 @@ class PollController extends AbstractController {
                 new RedirectResponse($this->core->buildCourseUrl(['polls']))
             );
         }
-        if (!in_array($_POST["release_histogram"], PollUtils::getHistogramReleases())) {
+        if (!in_array($_POST["release_histogram"], PollUtils::getReleaseHistogramSettings())) {
             $this->core->addErrorMessage("Invalid student histogram release setting");
             return MultiResponse::RedirectOnlyResponse(
                 new RedirectResponse($this->core->buildCourseUrl(['polls']))
@@ -420,7 +420,7 @@ class PollController extends AbstractController {
             $this->core->addErrorMessage("Invalid poll question type");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
-        if (!in_array($_POST["release_histogram"], PollUtils::getHistogramReleases())) {
+        if (!in_array($_POST["release_histogram"], PollUtils::getReleaseHistogramSettings())) {
             $this->core->addErrorMessage("Invalid student histogram release setting");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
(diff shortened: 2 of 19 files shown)


site/app/controllers/PollController.php
```diff
@@ -538,6 +547,13 @@ public function viewResults($poll_id) {
             );
         }
         $poll = $this->core->getQueries()->getPoll($poll_id);
+
+        if (!$this->core->getUser()->accessAdmin() && !$poll->isStudentHistogramAvailable()) {
+            return MultiResponse::RedirectOnlyResponse(
+                new RedirectResponse($this->core->buildCourseUrl(['polls']))
+            );
+        }
```
sbelsk, line 555: I tested trying to view the poll histogram page from a student after deleting the poll as an instructor and got the error: 
image
I believe you also want to check if `$poll` is not null here, since this is what the function `getPoll` returns when no row in the DB matches the given id. This way, this if statement will also cover this case and students won't see a fog robot page, but would instead be redirected to the main poll page.

sbelsk (changes requested):
Looks and works really nice! I tried to play around with changing the settings, closing and opening the polls, changing the student responses etc. and everything worked as intended (except one tiny thing I left an inline comment about). I do have three main things to say:
- I noticed that when a poll is configured to always display the result histogram and when the poll is open, the two buttons for "submit" and "see class results" are below each other with no spacing: 
crowded_buttons
would it be better to have them next to each other?
- Again if a poll is configured to always have the histogram visible to students, and say a student wants to submit their answer to the poll and then look at the histogram. They have select their answer, click "submit", be redirected to the home page, click the poll again, and only then click to see the results. I wonder if there is any better place we can put the button so that they could access the histogram page without having that many page loads/clicks...
- It would be nice if we could add a few basic test cases for this new feature in the cypress test that was added recently: `Submitty/site/cypress/integration/polls.spec.js`

site/app/controllers/PollController.php
BenKelly99: Fixed.

BenKelly99:
Fixed the issues you brought up besides cypress tests, which Barb said we don't need to worry about.

BenKelly99:
Also re: having students get kicked out to main polls page, i think we'll want to change it so that submitting submits the answer via an AJAX call instead of a redirecting post request. But that will be a separate PR. @sbelsk

site/app/templates/polls/NewPollPage.twig
```diff
@@ -40,9 +40,21 @@
                 Expected release date:
             </label>
             <input id="poll-date" class="poll-date" name="release_date" autocomplete="off" type="text" size="10" value=""/>
-            <br/>
 
+            <br/><br/>
+            
+            <label for="student-histogram-release-setting" class="option-title">
+                Release response histogram to students
```
sbelsk, line 47: To be consistent with the formatting of the rest of this page
```suggestion
                Release response histogram to students:
```

site/app/templates/polls/NewPollPage.twig
```diff
@@ -112,6 +124,17 @@
                 Expected release date:
             </label>
             <input id="poll-date" class="poll-date" name="release_date" autocomplete="off" type="text" size="10" value="{{ poll.getReleaseDate() }}"/>
+            <br/><br/>
+
+            <label for="student-histogram-release-setting" class="option-title">
+                Release response histogram to students
```
sbelsk, line 130: Same as above
```suggestion
                Release response histogram to students:
```

sbelsk (changes requested):
Fixes look good. I noticed two more small things when testing again:
1. The font of the text on the buttons on the instructor page now look different 
font_chnaged
2. When a poll is configured to always release response histogram to students, on the instructor page, the icon on the "View Results" button is locked when the poll is not visible: 
image, not sure if this is intended behavior, but I can see how it might be confusing for instructors.

BenKelly99:
Change made to the text, the view results being locked when the poll isn't visible is intended, since students shouldn't be able to view the results for a poll when that poll is not available to be viewed.

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -171,7 +187,17 @@
...
                         <td id="poll_{{ poll.getId() }}_responses"> {{ poll.getUserResponses()|length }} </td>
-                        <td><a href="{{base_url}}/viewResults/{{ poll.getId() }}" class="btn btn-primary">View Results</a></td>
+                        <td>
+                            <a href="{{base_url}}/viewResults/{{ poll.getId() }}" class="btn btn-primary">
+                                {% if poll.isStudentHistogramAvailable() %}
+                                    <i class="fas fa-lock-open">
+                                {% else %}
+                                    <i class="fas fa-lock">
+                                {% endif %}
+                                    &nbsp;View Results
```
sbelsk, line 197: I think you missed the fix for the font in the future polls table, but the fix did work on the other tables :)

BenKelly99:
@sbelsk Good catch! Updated now.

sbelsk (approved):

migration/migrator/data/course_tables.sql
```diff
@@ -778,7 +778,8 @@ CREATE TABLE public.polls (
     status text NOT NULL,
     release_date date NOT NULL,
     image_path text,
-    question_type character varying(35) DEFAULT 'single-response-multiple-correct'::character varying
+    question_type character varying(35) DEFAULT 'single-response-multiple-correct'::character varying,
+    student_histogram_release_setting text NOT NULL
```
MasterOdin, line 782: Use `character varying` if what we're going to be storing is a fairly finite length string.

MasterOdin: Also I think we can rename this to `release_histogram` to save on characters throughout the codebase. We're not going to have some toggle for releasing histogram to other user groups as far as I'm aware.

MasterOdin (changes requested):

BenKelly99:
@MasterOdin Updated the names of the field and functions.

PR #7042, opened 2021-09-08
[Refactor:System] Add cache for annotations

DefinitiveAbove:
Annotations loaded by the WebRouter are fully parsed from files on each request.

### What is the new behavior?
Annotations loaded by the WebRouter are now cached.

### Other information?
Caching is currently done with files; it may be better to use a different cache type for both this and Twig caching - see https://symfony.com/doc/current/components/cache.html#available-cache-adapters for caching options with annotations


site/composer.json
```diff
@@ -29,6 +29,7 @@
     "php-ds/php-ds": "1.3.0",
     "ramsey/uuid": "4.2.1",
     "spatie/commonmark-highlighter": "2.1.1",
+    "symfony/cache": "^4.4",
```
MasterOdin, line 32: You a set value, in this case `4.4.30`.

Because Submitty is an application that's never imported we don't really gain anything by defining a range of allowed versions of a dependency, versus having a set value makes it clear what version of a dependency we're using (and in the case of symfony that all versions are the same).

site/tests/app/libraries/routers/WebRouterTester.php
```diff
@@ -3,13 +3,20 @@
...
 use Symfony\Component\HttpFoundation\Request;
 
 /**
  * @runTestsInSeparateProcesses
  */
 class WebRouterTester extends BaseUnitTest {
+    public function createMockCore($config_values = [], $user_config = [], $queries = [], $access = []) {
+        $config_values['debug'] = true;
+        $core = parent::createMockCore($config_values, $user_config, $queries, $access);
+        return $core;
```
MasterOdin, line 17:
```suggestion
        return parent::createMockCore($config_values, $user_config, $queries, $access);
```

site/tests/app/libraries/routers/WebRouterTester.php
```diff
@@ -3,13 +3,20 @@
 namespace tests\app\libraries\routers;
 
 use app\libraries\routers\WebRouter;
+use app\models\Config;
```
MasterOdin, line 6: This is used or can it be removed?

```suggestion
```

MasterOdin (changes requested):

MasterOdin (approved):
The changes here are fine, and are definitely best practices of what to do given the docs.

However, whenever you make a performance type change, it's good to come up with some way to measure the change. What's the page load difference from using the file system cache vs the in-memory cache? This is even more important if we were to move from the file system to something like memcached. The latter will be faster, but if it's only a speed-up of say 0.01ms, then it's not worth the time spent setting it up and dealing with memcached as a dependency.

For PHP, measuring this kind of information is by going through [xdebug](https://xdebug.org/).

PR #7048, opened 2021-09-08
[Feature:Submission] Add alert to explain locked gradeable

### LINKED ISSUE #7037: When a gradeable is "Locked" add a hover or popup explaining why
Gradeaable 2 is locked until student has earned minimum points on Gradeable 1.
But student doesn't know why.

Add hover text and/or a. popup to explain which gradeable, and what point threshold is necessary to unlock.

loadedskiffer:
closes #7037

### What is the new behavior?
If a gradeable is locked and a user clicks on it, an alert pops up that  tells them what prerequisite they must complete in order to gain access to it. There is also a tooltip that pops up with the same message if the user hovers over.
pr18_1

### Other information?
To test:
1. Lock a gradeable by adding a prerequisite
2. Test the button


site/app/models/gradeable/Gradeable.php
```diff
@@ -2206,4 +2206,18 @@ public function isLocked(string $user_id): bool {
...
     }
+    /**
+     * Returns prerequisite for a gradeable
+     *
+     * @return string
+     */
+    public function getPrerequisite(): string {
+        if ($this->depends_on !== null && $this->depends_on_points !== null) {
+            $dependent_gradeable = $this->core->getQueries()->getGradeableConfig($this->depends_on);
+            return $dependent_gradeable->getId();
```
sbelsk, line 2217: Should the gradeable title be displayed to students as opposed to the gradeable id?

loadedskiffer: ya that makes more sense

williamjallen (commented):
Is a click the best trigger for this?  Would a tooltip be better perhaps?

loadedskiffer:
> Is a click the best trigger for this? Would a tooltip be better perhaps?

@bmcutler what do you think, I know you specifically requested the click but I think having the tooltip as well would be good.

williamjallen (changes requested):
Looks good in general but I have some concerns about the UI.  In my opinion, it would look better if it were either a real tooltip (using the native browser tooltip functionality) or a simulated tooltip which appeared under the mouse pointer.  I think it could work other ways as well and I'm open to different ideas but those are my initial thoughts.  An alternative would be to use something similar to the tags which appear when you hover over a sidebar item when the sidebar is collapsed.  Perhaps this could share code with it?
Untitled

loadedskiffer:
> Looks good in general but I have some concerns about the UI. In my opinion, it would look better if it were either a real tooltip (using the native browser tooltip functionality) or a simulated tooltip which appeared under the mouse pointer. I think it could work other ways as well and I'm open to different ideas but those are my initial thoughts. An alternative would be to use something similar to the tags which appear when you hover over a sidebar item when the sidebar is collapsed. Perhaps this could share code with it?
> Untitled

I decided to go with the native tooltip, I made a custom one originally because I wanted to make it more visible and blatantly obvious but I guess the regular tooltip works just as well.

williamjallen:
How hard would it be to repurpose the label (not sure what to call it) which appears when you hover over elements on the sidebar when it is closed for a message here?  I think the native tooltip as is now implemented is reasonable but I also agree with your thinking that making it more obvious by adding something to the side of the button might be helpful.
Untitled

loadedskiffer:
> How hard would it be to repurpose the label (not sure what to call it) which appears when you hover over elements on the sidebar when it is closed for a message here? I think the native tooltip as is now implemented is reasonable but I also agree with your thinking that making it more obvious by adding something to the side of the button might be helpful.
> Untitled

it looks like it would be pretty nontrivial, the more I look at the native tooltip the more I like it anyways, I think that tooltip and the alert is definitely enough to let people know what's going on.

williamjallen (approved):

PR #7051, opened 2021-09-09
[Bugfix:InstructorUI] dark mode in table background 

### LINKED ISSUE #7040: Out of place white colored background in dark mode.
**Describe the bug**
The Numeric/Text gradeable table have an obnoxious white background when in dark mode.

**Expected behavior**
The background should fit the theme of dark mode. This can be done with `table-striped` class, or just giving it a transparent border.

**Configuration**

**Screenshots**
image

Gulsonish:
Please check if the PR fulfills your requirements:

Fixes #7040


site/public/css/server.css
```diff
@@ -244,6 +244,14 @@ button,
 /****************************
  * TABLE STYLES
  ***************************/
+
+@media (prefers-color-scheme: dark) {
```
HenrikLam, line 248: This solution is not preferred, because there is a settings page where the user can specify that they want the system's preferred color scheme. This looks like it will do the check again and may override the user's selection on their profile page. Another possible solution would be to just add the `table-striped` class to the tables on the page which will do the checking of which color scheme the user wants for you.

Gulsonish: Only I have to add table-striped class in table, right??

And "the tables on **the page which will do the checking of** which color scheme the user wants for you" I am confused with this sentence. You mean NumericRubric.twig file you are talking about, right??

Gulsonish: I am facing these kind of problem as I am unable to see changes and effects in API because I could not instal submitty locally to check API's operations because of issue on my computer.

HenrikLam: > Only I have to add table-striped class in table, right??
> 
> And "the tables on **the page which will do the checking of** which color scheme the user wants for you" I am confused with this sentence. You mean NumericRubric.twig file you are talking about, right??

Yup, though the class was actually `table-stripe`, my mistake. And turns out the problem was slightly more complicated than I had thought. Currently, the rows get cloned when we try to create more rows. However, the count of the row also gets cloned and each row ends up having an "even" count and thus the background color is not properly alternated. Instead of cloning the row, we can instead just insert a row using javascript.

HenrikLam (changes requested):
See comment above

Gulsonish:
> See comment above

1. problem was coming because there is no configured or added css with this table-stripe class name in any .css file instead they have table-striped class.
2. Don't worry cloning is obviously an not a desired outcome as we have to add a row or delete as per our requirements. so, I have to add buttons for new row creation or deletion both, right??

HenrikLam:
> > See comment above
> 
> 1. problem was coming because there is no configured or added css with this table-stripe class name in any .css file instead they have table-striped class. So, I have to add with in .css file according to color.css file.
> 2. Don't worry cloning is obviously not a desired outcome as we have to add a row or delete as per our requirements. so, I have to add buttons for new row creation or deletion both, right??
>    Screenshot 2021-09-09 003341

No need to add `table-striped`, just use the already existing `table-stripe`. No need to add a button either, what I meant was instead of using the clone function, we can instead to a regular insert that gives the inserted row all the required classes and attributes.

Gulsonish:
@HenrikLam Please check new PR #7103 and I apologise for long time gap, actually I was busy little bit, and I was trying to solve this problem in prefered way but It couldn't be possible as I had to create a variable for new row insertion which can be appended in that table. Please Check and reply me if any issue is there.

PR #7073, opened 2021-09-14
[Feature:Plagiarism] Plagiarism anonymous mode

williamjallen:
Currently, it is difficult for instructors to report bugs in the plagiarism results page due to confidentiality matters, primarily involving the association of code/plagiarism results with specific names.

### What is the new behavior?
This pull request introduces a new "anonymous mode" feature to the plagiarism results interface which allows the instructor to toggle between showing names and random hashes.

Untitled

### Other information?
This is not meant to be a completely anonymous mode such as is the case for blinded grading.  Since instructors can always turn this off, we only care about the visual display of data being anonymized so that instructors can more freely share screenshots and/or show students accused of cheating the evidence against them with fewer confidentiality concerns.

### DIFF
--- a/site/app/templates/plagiarism/PlagiarismResult.twig
+++ b/site/app/templates/plagiarism/PlagiarismResult.twig
@@ -17,9 +17,7 @@
         <div class="plag-flex-row">
             <span class="plag-flex-item">
                 <select id="user-1-dropdown-list" name="user_id_1" class="user1-select">
-                    {% for ranking in rankings %}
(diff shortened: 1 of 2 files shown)


sbelsk (changes requested):
Seems very nice! Worked well on normal gradeables configurations, as well as on configs with other gradeables, but for team gradeables it shows an extra "0" character in every dropdown item: 
image

sbelsk (approved):
Looks good now. One thing is, when the page is refreshed, or when you re-visit a results page for a config you have seen and toggled anon mode on, then by default the page goes back to anon mode off. Do we want for this to be saved on local storage maybe?

sbelsk (approved):
Works well.

PR #7076, opened 2021-09-14
[Bugfix:Forum] Fix edit post markdown preview

### LINKED ISSUE #7003: Turning off markdown while previewing renders the text box unable to be edited
**Describe the bug**
When previewing markdown, the preview remains even when markdown is turned off.  This renders the text box unable to be edited until markdown is turned back on and preview is turned off.  

**Expected behavior**
When turning off markdown, make sure the text box gets switched back to a text box instead of the markdown preview.

**To Reproduce**
1. Go to the forum and begin to make a new post
2. Turn on markdown (if it isn't already on)
3. Preview markdown
4. Turn off markdown by clicking the blue markdown icon
5. See error

**Configuration**
 - macOS/Chrome latest

**Screenshots**
Untitled

### LINKED ISSUE #7035: Incorrect markdown preview in Discussion Forum thread
**Describe the bug**
If you enable markdown preview on the forum, cancel the edit window, then open the edit window for another post on the same thread, the preview for the previous message will be displayed instead.

**Expected behavior**
The markdown preview for the message currently being edited should be shown. 

**To Reproduce**
1. Enable markdown preview on a forum post
2. Cancel the edit window
3. Begin editing a different message in the same thread

**Configuration**
 - OS: MacOS 10.15.7
 - Browser Chrome

joekrystowski:
Fixes #7003 and Fixes #7035
Currently, toggling out of markdown while markdown preview is on will leave the text unable to be edited, because it is still in preview mode. Additionally, if you cancel the edit popup while in preview mode and open the edit menu for a different message, the markdown preview for the previous message will be shown.

### What is the new behavior?
Now, when you toggle out of markdown the preview button will be toggled if it is currently in preview mode. Additionally, if you cancel the edit window while in preview mode, you will also be taken out of preview mode before canceling, ensuring that when you reopen an edit popup you will be in edit mode. This also follows the behavior of the update post button.

### DIFF
--- a/site/app/templates/forum/ThreadPostForm.twig
+++ b/site/app/templates/forum/ThreadPostForm.twig
@@ -189,12 +189,7 @@
             </div>
         {% endif %}
         {% if show_cancel_edit_form is defined and show_cancel_edit_form %}
-        <a onclick="
-                 $(`#markdown_buttons_0`).find('.preview-button').trigger('click');
-                 $('#edit-user-post').css('display', 'none');
-                 $(this).closest('.thread-post-form').find('[name=thread_post_content]').val('');
-                 $('#title').val('');
-                 $(this).closest('form').trigger('reinitialize.areYouSure');"
+        <a onclick="cancelEditPostForum.call()"
                      class="btn btn-default close-button key_to_click" tabindex="0">Cancel</a>
         {% endif %}
         <input type="submit" name="post" value="{{ submit_label }}" class="btn btn-primary inline-block" />
--- a/site/app/templates/forum/ThreadPostForm.twig
+++ b/site/app/templates/forum/ThreadPostForm.twig
@@ -189,7 +189,9 @@
             </div>
         {% endif %}
         {% if show_cancel_edit_form is defined and show_cancel_edit_form %}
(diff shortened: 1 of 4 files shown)


site/app/templates/forum/ThreadPostForm.twig
```diff
@@ -189,10 +189,7 @@
             </div>
         {% endif %}
         {% if show_cancel_edit_form is defined and show_cancel_edit_form %}
-        <a onclick="$('#edit-user-post').css('display', 'none');
-                 $(this).closest('.thread-post-form').find('[name=thread_post_content]').val('');
-                 $('#title').val('');
-                 $(this).closest('form').trigger('reinitialize.areYouSure');"
+        <a onclick="cancelEditPostForum.call()"
```
eddiekrystowski, line 192: Works as intended, so I'll approve.
Just a small code suggestion:
```suggestion
        <a onclick="cancelEditPostForum()"
```
`<function>.call()` is only necessary when you want to invoke a function with a specific calling context (`this` value). In this case, it doesn't matter, so you can just call the function directly.

eddiekrystowski (approved):

PR #7079, opened 2021-09-15
[Bugfix:TAGrading] Add Markdown to Solution Notes

eddiekrystowski:
Currently, markdown cannot be used in solution notes, and newline rendering is broken, leading to large ugly elements like this:
image

### What is the new behavior?
* Markdown has been added to the page
* The page has been streamlined
  * No more edit button, you just toggle the markdown preview/edit mode like on other pages
  * Components that already have solutions will start in preview mode, others without a solution will start in edit mode
  * Cancel button and Start Now (which shows for components with no solution) buttons have been removed
* As a result of markdown being added, the newline issue has been resolved

image


site/app/templates/misc/MarkdownArea.twig
```diff
@@ -41,7 +41,8 @@
             placeholder="{{ placeholder }}"
         {% endif %}
         rows="10" cols="30"
-        maxlength="5000"
+        {# 524288 characters is the default max length for text input fields #}
+        maxlength="{{no_maxlength is defined and no_maxlength ? '524288' : '5000'}}"
```
listeph, line 45: Is 5000 chars as the standard not enough? google says that's 1.5 pages single spaced. Just wondering because maybe we shouldn't introduce an new variable just to do this unless it's needed

site/app/templates/misc/MarkdownArea.twig
```diff
@@ -25,7 +25,7 @@
         <button title="Insert a code segment" type="button" onclick="addMarkdownCode(0, $('#{{ markdown_area_id }}'))" class="btn btn-default btn-markdown key_to_click"  tabindex="0">Code <i class="fas fa-code fa-1x"></i></button>
         <button title="Insert bold text" type="button" onclick="addMarkdownCode(2, $('#{{ markdown_area_id }}'))" class="btn btn-default btn-markdown key_to_click"  tabindex="0">Bold <i class="fas fa-bold fa-1x"></i></button>
         <button title="Insert italic text" type="button" onclick="addMarkdownCode(3, $('#{{ markdown_area_id }}'))" class="btn btn-default btn-markdown key_to_click"  tabindex="0">Italics <i class="fas fa-italic fa-1x"></i></button>
-        <button title="Preview Markdown" type="button" onclick="{{ onclick }}" class="btn btn-default btn-markdown key_to_click" tabindex="0">Preview <i class="fas fa-eye fa-1x"></i></button>
+        <button title="Preview Markdown" type="button" onclick="{{ onclick }}" class="btn btn-default btn-markdown key_to_click" tabindex="0" data-initialize_preview="{{initialize_preview is defined ? initialize_preview : false}}">Preview <i class="fas fa-eye fa-1x"></i></button>
```
listeph, line 28: I think this can be `initialize_preview == true` , but I also think it's better to do this outside the .twig when you do `include "misc/MarkdownArea.twig"`

site/app/templates/grading/electronic/SolutionTaNotesPanel.twig
```diff
@@ -26,42 +26,32 @@
...
-                >Edit</button>
-            </div>
-            <div class="solution-edit hide" id="sol-textbox-cont-{{ component_id }}-edit">
+            <div class="solution-edit" id="sol-textbox-cont-{{ component_id }}-edit">
                 <label for="textbox-solution-{{ component_id }}" tabIndex="0" class="screen-reader">Solution for {{ component_title }}</label>
-                <textarea id="textbox-solution-{{ component_id }}"
-                          class="solution-ta-notes-textbox"
-                          placeholder='Start writing the solution for "{{ component_title }}" component.'
-                >{{ solution_notes }}</textarea>
+                {% set initialize_preview = not is_first_edit %}
```
listeph, line 40: Is this needed? Can you directly do `"initialize_preview" : is_first_edit == true`

site/app/templates/misc/MarkdownArea.twig
listeph: why is the variable name `data-initialize_preview` have a dash mixed with an underscore?

site/public/css/electronic.css
```diff
@@ -1462,7 +1462,7 @@ tr.is_publish {
 }
 
 [id^=textbox-solution-] {
-    min-height: 120px;
+    min-height: 150px !important;
```
listeph, line 1465: does this work without the `!important`? I think using `!important` [might be bad for future maintainability](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#how_is_specificity_calculated), but someone else might know about this better than I do

site/app/templates/misc/MarkdownArea.twig
```diff
@@ -65,5 +66,6 @@
 <script>
     $('#{{markdown_area_id}}').ready(function() {
         enableTabsInTextArea("#{{markdown_area_id}}");
-    })
+    });
+    $('#{{markdown_area_id}}').closest('.markdown-area').find('[data-initialize_preview="1"]').trigger('click');
```
listeph, line 70: based on [this explanation of `.ready()`](http://tutorials.jenkov.com/jquery/document-ready.html ), I think this line of code should be in there to make sure the button has loaded first?

listeph (changes requested):
I have a few suggestions, but overall some good changes so far! :)
- Have we discussed the streamlined UI with the rest of the Submitty team yet? I think maybe on a page refresh / fresh load, finding the white secondary "Edit" button (as opposed to the blue primary "Update" button) might be unintuitive.
- This is unrelated to your changes, but I noticed while testing that I get an error toast when I delete everything and then press "update"

site/public/js/solution-ta-notes.js
```diff
@@ -1,5 +1,5 @@
-/* global buildCourseUrl, displaySuccessMessage, displayErrorMessage, csrfToken */
-/* exported updateSolutionTaNotes, showSolutionTextboxCont, cancelEditingSolution */
+/* global buildCourseUrl, displaySuccessMessage, displayErrorMessage, csrfToken, previewMarkdown */
+/* exported updateSolutionTaNotes, showSolutionTextboxCont, cancelEditingSolution, previewSolutionNotesMarkdown */
```
listeph, line 2: You deleted some functions so you should probably delete them from the comment too. I'm curious for my own knowledge -- do these comments do anything?

site/app/templates/misc/MarkdownArea.twig
eddiekrystowski: The reason I removed this limit for this page was because 
1) The original page didn't have a limit to the size of solutions
2) If an instructor has a large code snippet that they want to insert as a solution (or maybe they want to show multiple solutions for the same problem), they are able to paste it in the box even with the maxlength=5000, but then they would be unable to type any more characters (only backspace) which would be very confusing.
3) I think the instructor should just be able to specify a solution of any length, unlike the forum where we want to stop students from being able to, say, post their entire homework code.

Do you think these reasons are good? More than happy to discuss further if not. Just to clarify, all other places that use this `MarkdownArea.twig` template right now still have a limit of 5000 characters.

site/app/templates/misc/MarkdownArea.twig
eddiekrystowski: Technically, `.ready()` is only necessary for scripts that are in the `<head>` tag or earlier in the `<body>` tag than the elements they try to reference, since the DOM loads in a linear fashion so it is guaranteed that the top part of the twig template would render before the script tag is executed.

But I do agree, it definitely doesn't hurt to add, plus I already added it on the line above. Thanks for catching this!

site/app/templates/misc/MarkdownArea.twig
listeph: Okay sounds goodd! You're right with code snippets maybe it'll exceed 5000 chars

site/app/templates/misc/MarkdownArea.twig
listeph: Ohh I see, thanks for sharing about the DOM loading! You can make the call for what's best, or leave it as is then.

site/app/templates/misc/MarkdownArea.twig
listeph: Nitpick / up to you, but this could be more concise using `[|default](https://twig.symfony.com/doc/2.x/filters/default.html)` like this: `{{no_maxlength|default(null) ? '524288' : '5000'}}`

eddiekrystowski: Oh cool, I wasn't aware of this filter! Thanks for the suggestion.

site/app/templates/misc/MarkdownArea.twig
eddiekrystowski: Oops, good catch. I'll fix that

eddiekrystowski:
@listeph Sorry for taking so long to address your changes, I think I've resolved all of them. I agree that the Edit/Preview button was a bit easy to miss so I've made it a primary button like the update button, let me know what you think!

listeph (approved):
Works, yay for markdown!! 🎉 🎉 

Nitpick, but in the future the UI might need improvement though because having the "Edit/Preview" button _and_ the "Update" button might still be kinda confusing. Even though I get why doing it this way made the most sense with re-using the `MarkdownArea.twig` format.

I mentioned it to the team during the Wed meeting and they suggested maybe in the future we could have our Markdown.twig file something more like Github's UI as an example, where "preview" is a separate tab.
Screen Shot 2021-09-24 at 3 32 14 PM
Screen Shot 2021-09-24 at 3 32 18 PM

eddiekrystowski:
I agree, I've thought about having another tab, just haven't gotten around to it. But if it's something people think would be beneficial I can definitely look into it!

PR #7084, opened 2021-09-15
[Feature:HelpQueue] Make queue access code optional

### LINKED ISSUE #7072: Allow Office hour queues to not require access code
Some courses want the access code for office hours queue to be optional.

 if we set access code = empty string, don’t ask student for access code for that queue

loadedskiffer:
closes #7072

### What is the new behavior?
An instructor can choose to not have an access code for a queue, in which case they leave the access code input blank. Students can then join the queue without an access code.

### Other information?
To test:
1. Open a new queue and leave the access code blank
2. join the queue without an access code
3. attempt to join queues that do have required access codes without entering one


site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -57,20 +57,16 @@ public function openQueue() {
...
-            $this->core->addErrorMessage("Missing secret code");
-            return MultiResponse::RedirectOnlyResponse(
-                new RedirectResponse($this->core->buildCourseUrl(['office_hours_queue']))
-            );
-        }
-
         //Replace whitespace with "_"
         $queue_code = trim($_POST['code']);
-        $token = trim($_POST['token']);
+        $token = empty($_POST['token']) ? '' : trim($_POST['token']);
```
shailpatels, line 62: I think `isset` would be the better check here since it only checks to make sure a value is set, whereas `empty` checks for a value that is truthy to empty (0,false,empty string).

Because of that the value `0` cannot be used as a secret

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -57,20 +57,16 @@ public function openQueue() {
...
 
         $re = '/^[\sa-zA-Z0-9_\-]+$/m';
         preg_match_all($re, $queue_code, $matches_code, PREG_SET_ORDER, 0);
-        preg_match_all($re, $token, $matches_token, PREG_SET_ORDER, 0);
+        $matches_token = 1;
+        if ($token != "") {
```
shailpatels, line 67: Should use `!==` prevent PHP from type casting in the comparison

site/app/templates/officeHoursQueue/NewQueue.twig
```diff
@@ -7,10 +7,11 @@
         <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
         <input type="text" id="new_queue_code" name="code" placeholder="New Queue Name" aria-label="New Queue Name" required="required" maxlength="2048" style="margin-top: 1rem;"><br>
         <div class="access-code-input" style="margin-top: 1rem;">
-            <input type="text" id="new_queue_token" name="token" placeholder="Access Code" aria-label="Access Code" required="required" maxlength="2048">
+            <input type="text" id="new_queue_token" name="token" placeholder="Access Code" aria-label="Access Code" maxlength="2048">
```
shailpatels, line 10: HTML Ids should use dashes not underscores 
https://submitty.org/developer/coding_style_guide/css

site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -15,7 +15,7 @@
       </select>
     </div>
     <label for="token_box">Queue Access Code:</label>
-    <input type="text" name="token" id="token_box" placeholder="Queue Access Code" style="margin:.5rem" autocomplete="on" required="required">
+    <input type="text" name="token" id="token_box" placeholder="Queue Access Code" style="margin:.5rem" autocomplete="on">
```
shailpatels, line 18: Same comment as above

shailpatels (changes requested):

shailpatels (changes requested):
The optional access code works well,
however creating a new q and setting an access code does not get saved correctly and show up as having no access code.

loadedskiffer:
> The optional access code works well,
> however creating a new q and setting an access code does not get saved correctly and show up as having no access code.

Whoops when I switched from empty() to isset() I forgot to change the logic

shailpatels (approved):
Looks good, and works as described

Tested on firefox,chrome macos

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -57,20 +57,16 @@ public function openQueue() {
...
-            $this->core->addErrorMessage("Missing secret code");
-            return MultiResponse::RedirectOnlyResponse(
-                new RedirectResponse($this->core->buildCourseUrl(['office_hours_queue']))
-            );
-        }
-
         //Replace whitespace with "_"
         $queue_code = trim($_POST['code']);
-        $token = trim($_POST['token']);
+        $token = isset($_POST['token']) ? trim($_POST['token']) : '';
```
MasterOdin, line 62: Easier to use null coalescing operator within the `trim` to avoid the ternary.
```suggestion
        $token = trim($_POST['token'] ?? '');
```

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -143,7 +132,7 @@ public function addPerson($queue_code) {
         }
 
         $queue_code = trim($queue_code);
-        $token = trim($_POST['token']);
+        $token = isset($_POST['token']) ? trim($_POST['token']) : '';
```
MasterOdin, line 135: Same here:
```
        $token = trim($_POST['token'] ?? '');
```

PR #7094, opened 2021-09-16
[Refactor:System] Simplify the PHP websocket code

MasterOdin:
### What is the new behavior?

Looking at the PHP code to figure out what could be the root cause the reported warning messages about undefined index reference when closing the socket, found there was a number of opportunities to simplify the code, strengthen types, etc. Behavior should be largely unchanged, with the exception of some improved error handling around authentication.

### Other information?

Should fix this error reported by @bmcutler:
image

### DIFF
--- a/site/app/libraries/socket/Client.php
+++ b/site/app/libraries/socket/Client.php
@@ -16,7 +16,12 @@ class Client extends WebSocket\Client {
         ]);
     }
 
-    public function send($payload, $opcode = 'text', $masked = true): void {
-        parent::send(json_encode($payload), $opcode, $masked);
+    /**
+     * Send a JSON encoded payload
+     *
+     * @param mixed $json
+     */
+    public function json_send($json): void {
+        $this->send(json_encode($json));
     }
 }
--- a/site/app/libraries/socket/Server.php
+++ b/site/app/libraries/socket/Server.php
@@ -13,22 +13,22 @@ class Server implements MessageComponentInterface {
 
     // Holds the mapping between pages that have open socket clients and those clients
     /** @var array */
-    private array $clients = [];
+    private $clients = [];
 
     // Holds the mapping between Connection Objects IDs (key) and user current course&page (value)
     /** @var array */
-    private array $pages = [];
+    private $pages = [];
 
     // Holds the mapping between Connection Objects IDs (key) and User_ID (value)
     /** @var array */
-    private array $sessions = [];
+    private $sessions = [];
 
     // Holds the mapping between User_ID (key) and Connection objects (value)
     /** @var array  */
-    private array $users = [];
+    private $users = [];
 
     /** @var Core */
-    private Core $core;
+    private $core;
 
     public function __construct(Core $core) {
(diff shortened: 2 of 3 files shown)


site/app/libraries/socket/Client.php
```diff
@@ -16,7 +16,12 @@ public function __construct(Core $core) {
         ]);
     }
 
-    public function send($payload, $opcode = 'text', $masked = true): void {
```
MasterOdin, line 19: Overriding the `send` function was causing an error in my client where `$payload` should only ever be a string. Not sure why phpstan wasn't flagging this as an error in the build. Given the likelihood the parent type may become more strongly typed in the future and this becomes a proper error, better to just avoid it, and add a new function that expects to `json_encode` its payload before calling `send`.

site/app/libraries/socket/Server.php
```diff
@@ -91,108 +84,42 @@ private function broadcast(ConnectionInterface $from, string $content, string $p
...
-     * User_ID refers to the user_id of the user who is making a connection to the socket from a web page
-     * For example: if instructor has opened a socket connection from Forum threads page, user_id is instructor
-     */
-
-    /**
-     * Fetches Connection object/s of a given User_ID
-     * @param string $user_id
-     * @return bool|ConnectionInterface
-     */
-    private function getSocketClients(string $user_id) {
```
MasterOdin, line 107: This function is no longer used, so removed.

site/app/libraries/socket/Server.php
```diff
@@ -91,108 +84,42 @@ private function broadcast(ConnectionInterface $from, string $content, string $p
...
-            $this->users[$user_id] = [$conn];
-        }
-    }
-
-    /**
-     * Deletes Connection object associativity with User_ID
-     * @param ConnectionInterface $conn
-     * @return void
-     */
-    private function removeSocketClient(ConnectionInterface $conn): void {
```
MasterOdin, line 151: This function was only being called in one place (`onClose`), so just moved its code into that caller.

site/app/libraries/socket/Server.php
```diff
@@ -91,108 +84,42 @@ private function broadcast(ConnectionInterface $from, string $content, string $p
...
-        $user_id = $this->getSocketUserID($conn);
-        if ($user_id) {
-            unset($this->sessions[$conn->resourceId]);
-            foreach ($this->users[$user_id] as $client) {
-                if ($client === $conn) {
-                    unset($client);
-                    break;
-                }
-            }
-            unset($this->users[$user_id]);
```
MasterOdin, line 161: This was the cause of the notice/warnings we were noticing in prod. If a user opened two websocket connections, and then closed one, this line runs even though `$this->users[$user_id]` is non-empty as it holds a reference to the second socket. When the second socket goes to close, `$this->users[$user_id]` has been unset, so we get the notice/warning.

HenrikLam (approved):
Looks good to me. Added some type declarations, can you check if this will cause any problem?

PR #7109, opened 2021-09-19
[Bugfix:Submission] QR/OCR bulk upload fixes

shailpatels:

OpenCV function change in the version used  in submitty and qr_prefix was not getting sent correctly

### What is the new behavior?
This PR includes some fixes related to bulk uploads
The opencv findContours function had its return typed changed in 3.4 for Ubuntu 20 and is now handled correctly
The qr_prefix parameter was not getting sent correctly to the backend
Submitting a split item by numeric user id was broken


HenrikLam (approved):
Looks good, QR scanning doesn't seem to have errors in logs anymore when the expected range is correct.

PR #7115, opened 2021-09-20
[Bugfix:InstructorUI] Fix NaN bug when changing date field

### LINKED ISSUE #7054: excused absence extensions NaN
**Describe the bug**
when I fill out this form I should be able to specific days of extension OR new due date.
But if I only specify days of extension, and try to leave new due date blank, it switches 
my days of extension to NaN.   I see this on Mac Chrome -- if that is relevant (haven't 
tested other browsers.

Screen Shot 2021-09-09 at 9 03 36 PM
Screen Shot 2021-09-09 at 9 03 44 PM

williamjallen:
Fixes #7054

### What is the new behavior?
Fixes bug such that the number of late days specified manually will only be overwritten when a valid date is selected.  Perhaps @bmcutler has thoughts here?  An alternative would be to overwrite the manually specified late days when the date field is not empty, even if it has an invalid date in it.

### DIFF
--- a/site/public/js/extensions.js
+++ b/site/public/js/extensions.js
@@ -58,7 +58,9 @@ function setLateDays() {
     const new_date = new Date($('#late-calendar').val());
     const old_date = new Date($('#due-date').data('date'));
     const diff = (new_date.getTime() - old_date.getTime()) / (1000 * 3600 * 24);
-    document.getElementById('late-days').value = diff;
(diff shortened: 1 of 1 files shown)


shailpatels (approved):
Looks good, tabbing away from the calendar input without selecting does not overwrite with NaN,
changing the date manually works like normal

Issue #6415, opened 2021-05-13
Replace all usages of `new Date` and `Date.parse` with date library


MasterOdin:
**What problem are you trying to solve with Submitty**
We use the builtin JS date library in a number of places. While this mostly works, we keep getting bit by browser implementation specific stuff and pitfalls because it's not a very developer friendly interface. This is especially a bit hairy as we then do bits of math with these dates, which is equally error prone.

**Describe the way you'd like to solve this problem**
Instead, we should use a library like [luxon](https://moment.github.io/luxon/) which abstracts this stuff (as well as takes advantage of the `Intl` library for better international support). This would also provide the same interface across all browsers so that we're confident we're not going to hit a bug where the date displays properly on Chrome, but not on Safari or FireFox.

**Describe any potential alternatives you'd tried to solve the problem**
There's a number of libraries we should consider, including but not limited to:
- luxon
- date-fns
- DayJS

Whoever does this should look around the ecosystem to see what people are using and weigh the various options.

wusatosi:
This is all the places I can find:

https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/site/public/js/latedays.js#L8-L9
https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/site/public/js/latedays.js#L33-L40
https://github.com/Submitty/Submitty/blob/1361619858aaec89157bea5ab79b37b721b120ce/site/public/js/extensions.js#L7
https://github.com/Submitty/Submitty/blob/1361619858aaec89157bea5ab79b37b721b120ce/site/public/js/extensions.js#L57-L62
https://github.com/Submitty/Submitty/blob/1361619858aaec89157bea5ab79b37b721b120ce/site/public/js/confetti.js#L48
https://github.com/Submitty/Submitty/blob/8339b0df017bacfce7f68ee72f9ea3eea6615628/site/public/js/calendar.js#L52-L55
https://github.com/Submitty/Submitty/blob/8339b0df017bacfce7f68ee72f9ea3eea6615628/site/public/js/calendar.js#L184-L186
https://github.com/Submitty/Submitty/blob/8339b0df017bacfce7f68ee72f9ea3eea6615628/site/public/js/calendar.js#L205-L216
https://github.com/Submitty/Submitty/blob/dbee91d58f4918b5c9fbc60bae2e5989ac959f73/site/public/js/simple-grading.js#L274
[...]

PR #7123, opened 2021-09-22
[Bugfix:Forum] Missing expiration datepicker fix

jenaybarela:

Currently on firefox the expiration date picker does not show up due to an issue with date.parse not working with "-"  in the date

### What is the new behavior?
The new behavior replaces the "-" with "/". This is only a temporary fix when issue #6415 is implemented this will also be fixed.

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -788,7 +788,8 @@ function showEditPostForm(post_id, thread_id, shouldEditThread, render_markdown,
                 var thread_title = json.title;
                 var thread_lock_date =  json.lock_thread_date;
                 var thread_status = json.thread_status;
-                var expiration = json.expiration;
+                var expiration = json.expiration.replace("-", "/");
+                expiration = expiration.replace("-", "/");
                 $("#title").prop('disabled', false);
                 $(".edit_thread").show();
                 $('#label_lock_thread').show();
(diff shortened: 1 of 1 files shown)


site/public/js/forum.js
```diff
@@ -788,7 +788,8 @@ function showEditPostForm(post_id, thread_id, shouldEditThread, render_markdown,
                 var thread_title = json.title;
                 var thread_lock_date =  json.lock_thread_date;
                 var thread_status = json.thread_status;
-                var expiration = json.expiration;
+                var expiration = json.expiration.replace("-", "/");
+                expiration = expiration.replace("-", "/");
```
MasterOdin, line 792: You can use the `g` flag on regex to match all elements in a replacement. Given that there's only two `/` characters and no more (unless we have an invalid date), should be fine to do:
```suggestion
                const expiration = json.expiration.replace(/-/g, "/");
```

jenaybarela: I didn't use the `g` flag since there is another `-` that shouldn't get changed. 
Here is how the date is stored in the database `2021-09-27 00:00:00-04`

MasterOdin: Oh okay, well this is certainly a mess, and all the more reason to move away from `Date.parse`.

PR #7126, opened 2021-09-22
[Bugfix:TAGrading] Remove empty comments from DB

### LINKED ISSUE #7113: Editing the ta grading overall comment to be empty should delete entry from database
**Describe the bug**
If you write an overall comment for a student.  And it saves.  But then go back and delete the overall comment (set it to empty string)...  it does save correctly, but the row is still in the database.   

**Expected behavior**
I'm not sure if the student sees the empty message from the author of that empty string comment.  (It would be confusing if it was displayed.  Better yet, we should remove that row from the database.

DefinitiveAbove:
When an overall comment is replaced to be empty, the empty comment remains in the database.

### What is the new behavior?
Fixes #7113 
When an overall comment is replaced to be empty, the empty comment is removed from the database.


williamjallen (approved):
Works as described.  Blank entries are removed from the DB.

williamjallen:
I missed it in my review but should this PR also include a migration which removes existing blank entries or do we not want to touch existing data?

MasterOdin:
I would say to have a migration to remove them. It's "bad" data that as the issue indicates, could have some issue where we show the blank comment on the UI to students and cause some confusion.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5790,6 +5790,11 @@ private function updateOverallComments(TaGradedGradeable $ta_graded_gradeable) {
 
     private function updateOverallComment(TaGradedGradeable $ta_graded_gradeable, $comment, $grader_id) {
         $g_id = $ta_graded_gradeable->getGradedGradeable()->getGradeable()->getId();
+        if ($comment === "") {
+            $this->deleteOverallComment($g_id, $grader_id, $ta_graded_gradeable->getGradedGradeable()->getGradeable()->isTeamAssignment());
```
MasterOdin, line 5794: While this change looks good, I feel like I'm missing something on what the `deleteOverallComment` function does:

https://github.com/Submitty/Submitty/blob/05c533bde9a64ad48c37cb67c55f6b0bb6a83e1c/site/app/libraries/database/DatabaseQueries.php#L5825-L5833

I don't think we want to delete the entire gradeable_data row on removing the TA comment...?

DefinitiveAbove: I changed it to remove the data in gd_overall_comment inside of gradeable_data; I might be missing it in the code but it looks like the gd_overall_comment column inside of gradeable_data is unused - would it be better to make a migration to also remove gd_overall_comment inside of gradeable_data?

bmcutler: If I'm remember correctly, this is a deprecated column.  
Previously (a term or two ago) overall comment was not tied to a specific grader.
Now it is tied to a specific grader -- allowing us to have multiple overall comments from different graders and peer graders.
It is probably safe to remove at this point.  We could do it in this PR, or maybe make a separate PR?

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -3047,7 +3047,7 @@ public function ajaxClearPeerMarks($gradeable_id) {
             $ta_graded_gradeable->deleteGradedComponent($component, $this->core->getQueries()->getUserById($peer_id));
         }
         $ta_graded_gradeable->removeOverallComment($peer_id);
-        $this->core->getQueries()->deleteOverallComment($gradeable_id, $peer_id, $gradeable->isTeamAssignment());
```
MasterOdin, line 3050: What is the behavior change here? Why was the original code setup to remove from `gradeable_data` in addition to `gradeable_data_overall_comment`? I'm guessing we still want that removal from `gradeable_data` as we're clearing a peer grading here, so the data is no longer valid.

MasterOdin (approved):

PR #7127, opened 2021-09-22
[Refactor:System] General performance improvements

DefinitiveAbove:
### What is the new behavior?
Performance improvements on AbstractModel, GradingOrder, and Component.
While this is aimed at improving performance on TA grading pages, it also will likely improve performance on other pages as well (the improvements to AbstractModel should improve performance on many pages).

On AbstractModel, a cache (not persistent) is added in to cache function types/name conversions, specifically to avoid redundant calls to Utils::startsWith and AbstractModel's convertName method.

On GradingOrder, getSortedGradedGradeables search is improved to avoid unnecessarily using getSubmitterIndex each time.

On Component, the call to getPrecision for a gradeable is moved outside of the for loop to avoid redundant calls.

Test results for improvements:

Grading index for gradeable "grading_homework" inside of the "sample" course as user "submitty-admin" on view all sections:
Cumulative load time before improvements: 2900 ms
Cumulative load time after improvements: 1168 ms

Grading page for user "aphacker" on gradeable "grading_homework" in the "sample" course as user "submitty-admin":
Cumulative load time before improvements: 1440 ms
Cumulative load time after improvements: 784 ms

Tested using Xdebug profiler, with debug mode disabled for Submitty and caches loaded for pages (each page loaded once after reload site before profiler was run); read profiler info with WinCacheGrind.

### Other information?
On the grading page for TA grading, more improvements can be made with improving late day calculations.


site/app/models/GradingOrder.php
```diff
@@ -496,18 +496,31 @@ public function getSortedGradedGradeables() {
...
+                }
+
+                $count++;
             }
         }
+        foreach ($items as $item => $val) {
+            $unsorted[] = $val;
+        }
+
         //Since the array's elements were not added in the same order as the indices, sort to fix it
```
williamjallen, line 523: does this comment need to be removed?

site/app/models/GradingOrder.php
```diff
@@ -496,18 +496,31 @@ public function getSortedGradedGradeables() {
...
-                $unsorted[] = $gg;
-            }
-            else {
-                $gg_idx[$idx] = $gg;
+            $items[$gg->getSubmitter()->getId()] = $gg;
+        }
+        $count = 0;
+
+        //Iterate through all sections and their submitters to find this one
+        foreach ($this->section_submitters as $name => $section) {
```
williamjallen, line 506: I'm a little confused about what this code is trying to do.  Is there a particular reason why it's being done this way?  From my naive point of view, not entirely understanding the code but just looking at the structure of it, it seems odd that the best way to do this would be to look things up in linear time repeatedly with a double-nested loop structure.

williamjallen (changes requested):

site/app/models/GradingOrder.php
```diff
@@ -496,18 +496,30 @@ public function getSortedGradedGradeables() {
...
+                $count++;
             }
         }
-        //Since the array's elements were not added in the same order as the indices, sort to fix it
-        ksort($gg_idx);
+        foreach ($items as $item => $val) {
+            $unsorted[] = $val;
+        }
+
         return array_merge($gg_idx, $unsorted);
```
MasterOdin, line 523: Can use [`array_values`](https://www.php.net/manual/en/function.array-values.php) to avoid the foreach. I'd also suggest doing this all inline to avoid any extra memory usage if these are big arrays.

```suggestion
        return array_merge($gg_idx, array_values($items));
```

site/app/models/GradingOrder.php
```diff
@@ -496,18 +496,30 @@ public function getSortedGradedGradeables() {
...
+        $count = 0;
+
+        //Iterate through all sections and their submitters to find this one
+        foreach ($this->section_submitters as $name => $section) {
+            for ($i = 0; $i < count($section); $i++) {
+                $testSub = $section[$i]->getId();
+
+                //Found them
+                if (isset($items[$testSub])) {
+                    $gg_idx[$count] = $items[$testSub];
```
MasterOdin, line 512: Do we care about the gaps in `$gg_idx` between indexes, or can it just be purely sequential, so like `$gg_idx[] = $items[$testSub];` would work fine.

MasterOdin (commented):
Changes to `AbstractModel` look fine. There's some small nits to `GradingOrder`, but I don't have enough context on full getting the changes to loops if we can do them more efficiently.

williamjallen (approved):

PR #7128, opened 2021-09-23
[Feature:HelpQueue] Refactor queue contact info

loadedskiffer:
Currently Enabling Contact Information for Queues is a global setting so if it's on all queues require contact information and vice versa.

### What is the new behavior?
Contact Information can now be turned on and off from queue to queue so a class can have some queues that require it and some that don't.
From the instructor side:
There is a check box to require contact info
pr19_1
If it is checked the option to specify the regex pattern appears
pr19_2
The instructor can also change the setting from the edit panel
pr19_3
From the student side:
If the student selects a queue that requires contact info, the box for them to input it will appear, otherwise it's hidden.
pr19_4
pr19_5

### Other information?
To Test:
Configure multiple queues with different contact info settings
try joining and editing them


site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -452,6 +452,29 @@ public function changeRegex($queue_code) {
...
     }
 
+    /**
+     * @Route("/courses/{_semester}/{_course}/office_hours_queue/{queue_code}/change_contact_information", methods={"POST"})
+     * @AccessControl(role="LIMITED_ACCESS_GRADER")
+     * @return MultiResponse
+     */
+    public function changeContactInformation($queue_code) {
+        if (empty($queue_code)) {
+            $this->core->addErrorMessage("Missing queue name");
```
shailpatels, line 462: Is queue code the same thing as queue name or should this message be updated

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -452,6 +452,29 @@ public function changeRegex($queue_code) {
...
+            return MultiResponse::RedirectOnlyResponse(
```
shailpatels, line 463: Can just do `return new RedicrectResponse(...)` instead of calling the multiresponse since we've deprecated that function

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -452,6 +452,29 @@ public function changeRegex($queue_code) {
         );
     }
 
+    /**
+     * @Route("/courses/{_semester}/{_course}/office_hours_queue/{queue_code}/change_contact_information", methods={"POST"})
+     * @AccessControl(role="LIMITED_ACCESS_GRADER")
+     * @return MultiResponse
+     */
+    public function changeContactInformation($queue_code) {
+        if (empty($queue_code)) {
```
shailpatels, line 461: Should this be `empty` or `isset` 
https://stackoverflow.com/a/1219550

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -452,6 +452,29 @@ public function changeRegex($queue_code) {
...
+            $this->core->addErrorMessage("Missing queue name");
+            return MultiResponse::RedirectOnlyResponse(
+                new RedirectResponse($this->core->buildCourseUrl(['office_hours_queue']))
+            );
+        }
+
+        $contact_information = $_POST['contact_information'] == "true";
```
shailpatels, line 468: Should use 3 equals signs for PHP strict comparison

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -4304,7 +4304,7 @@ public function getUsersNotificationSettings(array $user_ids) {
      * Gets All Parent Authors who this user responded to
      *
      * @param string $post_author_id current_user_id
-     * @param string $post_id        the parent post id
+     * @param string $post_id        the parent post idgetAllQueues
```
shailpatels, line 4307: why is getAllQueues appended to `id` here?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -6241,6 +6241,11 @@ public function getQueueId($queue_code) {
         return $this->course_db->rows()[0]['id'];
     }
 
+    public function getQueueHasContactInformation($queue_code) {
+        $this->course_db->query("select * from queue_settings where code = ?;", [$queue_code]);
```
shailpatels, line 6245: Is this query guaranteed to always return something, should there be a check for the number of results returned or ability to handle a null response

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -6423,6 +6428,9 @@ public function changeQueueRegex($regex_pattern, $queue_code) {
         $this->course_db->query("UPDATE queue_settings SET regex_pattern = ? WHERE code = ?", [$regex_pattern, $queue_code]);
     }
 
+    public function changeQueueContactInformation($contact_information, $queue_code) {
```
shailpatels, line 6431: Would recommend adding PHP type hints to newly added functions whenever possible 

https://submitty.org/developer/coding_style_guide/php#type-declarations

shailpatels (changes requested):
There's a lot of code related to gradeable prerequisite logic, should that be removed from this branch?

site/app/controllers/OfficeHoursQueueController.php
loadedskiffer: They are the same thing

site/app/controllers/OfficeHoursQueueController.php
loadedskiffer: yup should be !isset

site/app/libraries/database/DatabaseQueries.php
loadedskiffer: whoops I think I was trying to do a ctrl f and accidentally typed in the file instead of the search box

site/app/libraries/database/DatabaseQueries.php
loadedskiffer: The column has a default value of true and there is no way to ever make it something other than true/false so it's guaranteed to return something but I guess a check can't hurt

shailpatels: Ah ok if there's a default value then I expect it to be safe, wasn't sure if this was a value that could ever be unset

site/app/controllers/OfficeHoursQueueController.php
shailpatels: Ok sounds good

site/app/controllers/OfficeHoursQueueController.php
loadedskiffer: > There's a lot of code related to gradeable prerequisite logic, should that be removed from this branch?

Everything should be good now, can you please take another look when you get a chance?

shailpatels (changes requested):
I am unable to open a queue that does not have an access code, any attempt gives an error banner with ` Missing secret code `

loadedskiffer:
> I am unable to open a queue that does not have an access code, any attempt gives an error banner with `Missing secret code`

for some reason not all of the changes that I made in my PR where I made access codes optional were reflected in this branch, should be all good now.

shailpatels (approved):
Looks good and works as described

PR #7136, opened 2021-09-25
[Bugfix:SubminiPolls] Fix doubly-escaped HTML in hist title

### LINKED ISSUE #7098: Polls histogram encoded characters in title
**Describe the bug**
Double quotes in the title are encoded in the Submini histogram.

Screen Shot 2021-09-17 at 1 44 04 AM

Screen Shot 2021-09-17 at 1 44 19 AM

williamjallen:
HTML special characters are currently doubly-escaped in the title of the poll results histogram which leads to the issue described in #7098.

### What is the new behavior?
This PR fixes #7098 by safely un-encoding the title of the poll before passing it to the plotting library which encodes it again.  Care was taken to mitigate possible opportunities for XSS vulnerabilities by using `DOMParser().parseFromString` to decode the string after Twig renders it without sacrificing usability.

### DIFF
--- a/site/app/templates/polls/ViewPollResults.twig
+++ b/site/app/templates/polls/ViewPollResults.twig
@@ -11,8 +11,9 @@
             data[0].x.push("{{ poll.getResponseString(answer) }}");
             data[0].y.push({{ number }});
         {% endfor %}
+        const title_text = new DOMParser().parseFromString("{{ poll.getName() }}", "text/html");
         const layout = {
-          title: "{{ poll.getName() }}",
+          title: title_text.documentElement.textContent,
         }
         Plotly.newPlot("chartContainer", data, layout);
     }


BenKelly99 (approved):
LGTM

site/app/templates/polls/ViewPollResults.twig
```diff
@@ -11,8 +11,9 @@
             data[0].x.push("{{ poll.getResponseString(answer) }}");
             data[0].y.push({{ number }});
         {% endfor %}
+        const title_text = new DOMParser().parseFromString("{{ poll.getName() }}", "text/html");
```
shailpatels, line 14: If this is down in a `.twig` file why not use a PHP function to decode/unescape the poll name. You can do this here or in the `.getName` function - but I wouldn't recommend that since its in the model and there is no associated view file which would be the best place for it.

You can use https://www.php.net/manual/en/function.htmlspecialchars-decode.php if only HTML special chars are escaped (I'm not sure the actual encoding method used for polls - assuming its that from the PR description)

shailpatels (changes requested):

site/app/templates/polls/ViewPollResults.twig
williamjallen: The issue here is that if a PHP or Twig decoding method was used, this code would be susceptible to an XSS attack.  The method used here was done deliberately to make the string be valid JS before it is decoded.  Unless I missed something, I believe that the current method should be safe in all modern browsers.

shailpatels: I don't see how using twig to decode would cause an XSS attack (unless you would use twig's raw filter but that wouldn't be the correct decoding method for this case anyways)

shailpatels (approved):
Looks good and correctly decodes html special chars

site/app/templates/polls/ViewPollResults.twig
williamjallen: Ok.  If the decoding were to be done on the server-side, how would it work such that plotly doesn't try to encode it again?  The only way I could think of was by using the `|raw` filter after un-escaping every character which is bad for the aforementioned reasons.

PR #7137, opened 2021-09-25
[UI/UX:System] BugFix runaway markdown table borders

williamjallen:
#6998 added support for tables in markdown.  Unfortunately, a small portion of the CSS included in that PR applies to all `<td>` elements site-wide which leads to tables having thicker borders as shown below:

Untitled

### What is the new behavior?
Fixes the runaway CSS selector.
Untitled

### DIFF
--- a/site/public/css/markdown.css
+++ b/site/public/css/markdown.css
@@ -85,7 +85,7 @@
     overflow-x: scroll;
 }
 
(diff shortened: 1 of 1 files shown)


eddiekrystowski (approved):
Oops, thanks for catching this!

PR #7141, opened 2021-09-26
[Refactor:System] Lint several CSS files

williamjallen:
When CSS linting was first introduced, most files were ignored entirely.  PR #6663 fixed some of the issues but many files still contained so many errors that they had to be ignored entirely.

### What is the new behavior?
In this PR, I made a number of deeper changes which allowed for multiple files to be removed from the ignored list. Most notably, `server.css` has been linted fully, except for a handful of places where specific errors had to be ignored due to the complexity of the code.  Further Pull Requests will further reduce the number of completely ignored files.


site/public/css/global.css
```diff
@@ -80,10 +80,6 @@ nav a > i.f {
     padding-left: 7px;
 }
 
-.breadcrumb_heading {
-    font-size: 24px;
-}
-
```
eddiekrystowski, line 86: The removal of this makes breadcrumbs look strange:
image

site/public/css/server.css
```diff
@@ -1619,15 +1576,6 @@ end of styles used in the admin gradeable page
...
 
-.wrapper_layout_header > th {
-    padding: 4px;
-    border: 1px solid var(--text-black);
-}
-
-.wrapper-upload .wrapper_layout_header {
-    background-color: var(--standard-pale-gray);
-}
-
```
eddiekrystowski, line 1630: I think this class is still in use in `site/app/templates/admin/UploadWrapperForm.twig`, so perhaps this should be 
```
.wrapper-layout-header {
    color: var(--text-black);
}
```
so that the table works on dark mode. (of course, change the class to use dashes as well)
(Sorry, GitHub doesn't let me insert a suggestion on deleted lines)

eddiekrystowski (changes requested):
Looks mostly good! Just found a few tiny things. Grepped classes that were deleted and all (except for those mentioned in my review) weren't in use. Didn't see any typos either.

Thanks for doing this!

site/public/css/global.css
williamjallen: Good catch.  I used PHPStorm to look for each of these and apparently the indexing doesn't work as well as I'd hope.  Corrections made.

eddiekrystowski (approved):

PR #7142, opened 2021-09-27
[Bugfix:Submission] Fix seating chart in dark mode

### LINKED ISSUE #7125: Seating assignment banner font is hard to read in dark mode
**Describe the bug**
In dark mode, the font color in the seating assignment banner is white, even though the banner itself has a light green background color. This makes the text quite hard to read.

**Expected behavior**
Either the font color should be switched to be darker for this banner only or the background color of the banner should be switched to be darker. Latter is probably more preferable since it is dark mode.

**To Reproduce**
1. Switch to dark mode
2. View seating assignment banner in a course with seating assignments
3. See error

**Configuration**
 - OS: Windows
 - Browser: Firefox

**Screenshots**
image

**Additional context**
N/A

loadedskiffer:
closes #7125 
pr20_1

### What is the new behavior?
The table is clearly visible in light and dark mode
pr20_2
pr20_3

### Other information?
To test:
I couldn't figure out how to actually upload a seating chart so I just edited a few lines in Navigation.twig and Roomseating.twig to get the table to show up

### DIFF
--- a/site/app/templates/navigation/RoomSeating.twig
+++ b/site/app/templates/navigation/RoomSeating.twig
@@ -5,12 +5,12 @@
     {% endif %}
     <tr><td>
         <table border=0 cellpadding=5 cellspacing=0>
-            <tr><td colspan=2>{{ core.getDisplayedCourseName() }}{{ gradeable_title != null ? ': '~gradeable_title : ''}}</td></tr>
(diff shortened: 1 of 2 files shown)


sbelsk (commented):
Works as described. Do we not want the borders of this table to show anymore?

loadedskiffer:
> Works as described. Do we not want the borders of this table to show anymore?

I added them, not sure why they went away in the first place but they're back

PR #7144, opened 2021-09-27
[UI/UX:InstructorUI] Add Tabs to Markdown Area

eddiekrystowski:
-------
Currently, the areas to enter markdown on the site use a button to switch between Edit/Preview mode. It looks like this:
image
image
 
----
### What is the new behavior?
It was discussed that it would be better to separate the modes away from the button and into tabs, similarly to how GitHub works. This has multiple advantages, the primary ones being it is much easier for the user to see what mode they are in, and on pages like the TA Solution Notes where some buttons to submit updated information could be confused with the button for switching markdown modes.

Now the UI looks like this:
image
image

Also, additional comments at the top of `site/app/templates/misc/MarkdownArea.twig` and its twin `site/public/templates/misc/MarkdownArea.twig` have been added to help document the proper way to update/use these files.

----
### Other information?
When testing make sure to check **all** places that markdown input is used (or split it up with someone 😄, I've done my own testing but it's a lot to check so I may have missed something). The locations (on the site) are:
* Forum
* Polls
* Office Hours Queue Announcement
* Notebook Builder (often overlooked)
* Solution Notes panel in TA Grading
* Grade Inquiries
* Rubric Panel Overall Comment in TA Grading


site/public/templates/grading/OverallComment.twig
```diff
@@ -31,13 +31,13 @@ Required Parameters:
...
             placeholder : "Overall message for student about the gradeable...",
             preview_div_id : "overall-comment-markdown-preview-" ~ (overall_comment["logged_in_user"]["user_id"]|escape),
             preview_div_name : "overall-comment-markdown-preview-" ~ (overall_comment["logged_in_user"]["user_id"]|escape),
             preview_button_id : "overall-comment-markdown-preview-button",
-            onclick : "previewOverallCommentMarkdown.call(this, '" ~ (overall_comment["logged_in_user"]["user_id"]|escape) ~ "')", 
-            render_buttons : true,
+            onclick : "previewOverallCommentMarkdown.bind(this, '" ~ (overall_comment["logged_in_user"]["user_id"]|escape) ~ "')", 
+            render_header : true,
             min_height : "100px",
             max_height : "150px",
```
williamjallen, line 42: Should the min/max height parameters (elsewhere too, not just here) be measured in `em` units instead of `px` units?  If you measure it in `em` units you can control roughly how many lines tall the box is.  The counterargument here is that everything else is measured in pixels so why not do it here.  I'm not sure what the best option is here but I wanted to throw it out there just in case it was useful.

williamjallen (changes requested):
Looks really really good overall, in terms of both UI and code.  I do have a few comments about the UI and I've left a few comments throughout the code.

1. Should the text here be updated when in preview mode?  I'm not sure what the best behavior is but it initially seemed confusing because using tab to indent didn't work (for obvious reasons since we're in preview mode)
Untitled

2. Pressing tab in edit mode on the forum (both with and without markdown) actually adds two tabs, although that may be a forum issue, not an issue with this PR

3. Pressing the Link/Code/etc buttons adds text to the box, even in preview mode, although the changes are not displayed until you switch modes again.

4. Pressing the Link/Code/etc buttons causes the text box to lose focus.  This is a rather minor issue, and potentially not worth the effort to address, but it can be annoying if you're trying to add things inline.  See how GitHub does it for an example of what I expected the behavior to be.

5. I'm not sure what the intended behavior is but it was unclear to me whether the page worked properly and saved the solution notes.  I'm not sure if this is a markdown issue or a UI issue in general in this case.  I expected the markdown area to go away altogether when the update button was pressed, leaving rendered markdown and a button to click to update it.

6. You mentioned that the TA grading rubric panel has markdown but I didn't see the markdown preview or any of the control buttons when I was testing it.  Perhaps I was not using it correctly?

site/public/templates/misc/MarkdownArea.twig
```diff
@@ -1,72 +1,114 @@
...
+            <a target="_blank" href="https://submitty.org/student/communication/markdown">
+                <i style="font-style:normal;" class="fa-question-circle"></i>
+            </a>
+            <button type="button" title="Insert a link" onclick="addMarkdownCode(1, $('#{{ markdown_area_id }}'))" class="btn btn-default btn-markdown btn-markdown-link key_to_click"  tabindex="0">Link <i class="fas fa-link fa-1x"></i></button>
+            <button title="Insert a code segment" type="button" onclick="addMarkdownCode(0, $('#{{ markdown_area_id }}'))" class="btn btn-default btn-markdown btn-markdown-code key_to_click"  tabindex="0">Code <i class="fas fa-code fa-1x"></i></button>
+            <button title="Insert bold text" type="button" onclick="addMarkdownCode(2, $('#{{ markdown_area_id }}'))" class="btn btn-default btn-markdown btn-markdown-bold key_to_click"  tabindex="0">Bold <i class="fas fa-bold fa-1x"></i></button>
+            <button title="Insert italic text" type="button" onclick="addMarkdownCode(3, $('#{{ markdown_area_id }}'))" class="btn btn-default btn-markdown btn-markdown-italic key_to_click"  tabindex="0">Italics <i class="fas fa-italic fa-1x"></i></button>
```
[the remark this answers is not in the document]
eddiekrystowski: They are defined in the docstring of the function itself, but I agree, the function is a bit unclear and also the values themselves are quite arbitrary. In my newest commit I've rewritten this function to be clearer.

site/public/templates/misc/MarkdownArea.twig
```diff
@@ -1,72 +1,114 @@
...
+            {% endif %}
+            {% if textarea_onkeydown is defined %}
+                onkeydown="{{ textarea_onkeydown }}"
+            {% endif %}
+            {% if textarea_onpaste is defined %}
+                onpaste="{{ textarea_onpaste }}"
+            {% endif %}
+            {% if textarea_maxlength is defined %}
+                maxlength="{{ textarea_maxlength }}"
+            {% endif %}
```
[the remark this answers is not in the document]
eddiekrystowski: Good catch, I had combined `no_maxlength` and `textarea_maxlength` into a single expression and forgot to remove this part.

site/public/css/markdown.css
```diff
@@ -90,6 +92,89 @@
...
+    border: 1px solid var(--standard-light-gray);
+}
+
+/* header (.markdown-area-header + children)*/
+.markdown-area-header {
+    display: flex;
+    flex-direction: row;
+    background-color: var(--standard-pale-gray);
+    padding: 0 10px 0 10px;
+    border-radius: 6px 6px 0 0;
```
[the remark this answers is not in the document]
eddiekrystowski: I think I can shorten the padding, but I'm not sure if I can shorted the border-radius since according to [this](https://www.w3schools.com/cssref/css3_pr_border-radius.asp) there seems to be no way to style both the top left and top right corners without 4 values.

site/public/css/markdown.css
```diff
@@ -90,6 +92,89 @@
...
+    padding: 16px;
+    background: var(--default-white);
+    border-radius: 0 0 6px 6px;
+}
+
+.markdown-area textarea {
+    resize: vertical;
+    min-height: 100px;
+    max-height: 300px;
+    width: -webkit-fill-available;
```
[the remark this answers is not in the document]
eddiekrystowski: That width attribute is actually redundant since the `markdown-area` element will always have the `fill-available` class on it which does the same thing and properly works for all browsers (has multiple fallbacks), so I've just removed the width style here.

site/public/templates/grading/OverallComment.twig
eddiekrystowski: You're right about `em` being a better solution. It was simply a case of me converting the various forms throughout the site to use this common template, and it just so happens that those pages used `px` to define sizes so I did too since it was just easier. I'll add that conversion to my list of things to do later

eddiekrystowski:
Thanks for the in depth review! 

> 1. Should the text here be updated when in preview mode?  I'm not sure what the best behavior is but it initially seemed confusing because using tab to indent didn't work (for obvious reasons since we're in preview mode)

Good idea, I've changed this text to go away in preview mode.

> 2. Pressing tab in edit mode on the forum (both with and without markdown) actually adds two tabs, although that may be a forum issue, not an issue with this PR

I've been meaning to track down this issue, the problem was that the forum was calling `enableTabsInTextarea` multiple times on a single textarea. This is done by default for every `markdown-area` input, so I've removed the calls to this function elsewhere.

> 3. Pressing the Link/Code/etc buttons adds text to the box, even in preview mode, although the changes are not displayed until you switch modes again.

Good catch, this should be solved now.

> 4. Pressing the Link/Code/etc buttons causes the text box to lose focus.  This is a rather minor issue, and potentially not worth the effort to address, but it can be annoying if you're trying to add things inline.  See how GitHub does it for an example of what I expected the behavior to be.

I've updated it to restore focus to the textarea after pressing these buttons. I don't think it exactly mimics GitHub since GitHub tries to be smart about placing the preset around things instead of just inserting, but it should be an improvement regardless.

> 5. I'm not sure what the intended behavior is but it was unclear to me whether the page worked properly and saved the solution notes.  I'm not sure if this is a markdown issue or a UI issue in general in this case.  I expected the markdown area to go away altogether when the update button was pressed, leaving rendered markdown and a button to click to update it.

Clicking the Update button should produce a green success message at the top of the page saying something to the effect that the solution has been saved. Although I do see that this could be a source of confusion... perhaps `Update` is no longer a good name? Maybe just `Save`?

> 6. You mentioned that the TA grading rubric panel has markdown but I didn't see the markdown preview or any of the control buttons when I was testing it.  Perhaps I was not using it correctly?

I was referring to the Overall Comment section (located at the bottom of the rubric panel). I think this should always be on, so if you don't see it at all I would certainly be interested in seeing that in action since this page specifically has been temperamental with regard to markdown in the past, so I wouldn't be surprised if you are encountering some edge case.

eddiekrystowski:
@williamjallen I actually have a question, I noticed [these lines](https://github.com/Submitty/Submitty/blob/master/site/public/css/markdown.css#L117-L119) at the bottom of `markdown.css` which I presume were due to the runaway markdown selector fixed in #7137. If that is the case, are these lines safe to be deleted now that the selector has been fixed? I've checked the page myself, I just am not familiar with how that page should actually look.

williamjallen:
Code changes look good.  I'll do a more thorough final review of the code and functionality tonight before hitting approve.

I saw the last few lines in markdown.css when I was making the PR you mentioned and was confused about what they did so I left them there.  Perhaps it was an attempt at a fix for the table borders issue as you mentioned?  I'm not familiar with how that page should look either.

Regarding the overall comment section, I see this when I go [here](http://localhost:1511/courses/f21/sample/gradeable/grades_released_homework_onlyta/grading/grade?who_id=DF654JKL45QER34&sort=id&direction=ASC):
Untitled

HenrikLam:
There seems to be a delay when you click the preview tab, I assume this is the markdown rendering. However, when you spam between write and preview, the change in the tab is sometimes delayed, especially when you spam click the two buttons in an alternating fashion which might be annoying. Is it possible to load the preview before the tab switch to reduce the delay?

HenrikLam:
There seems to be an x scroll bar in the preview textbox but not one for write textbox. Is this intentional? We should make the two behave identically if possible.

HenrikLam:
Is there currently a way to both bold and italicize the text? That sounds like it could be useful at times.

site/public/js/grade-inquiry.js
```diff
@@ -223,11 +223,10 @@ function newDiscussionRender(discussion) {
...
 
-function previewInquiryMarkdown() {
+function previewInquiryMarkdown(mode) {
     const markdown_textarea = $(this).closest('.markdown-area').find('[name="replyTextArea"]');
     const preview_element = $('#inquiry_preview');
-    const preview_button = $(this);
     const inquiry_content = markdown_textarea.val();
 
-    previewMarkdown(markdown_textarea, preview_element, preview_button, { content: inquiry_content });
+    previewMarkdown(mode, markdown_textarea, preview_element, { content: inquiry_content });
```
HenrikLam, line 231: Instead of having a `previewMarkdown` function for the different places. Could we instead have a wrapper function that takes the `markdown_textarea` and `preview_element` variables and call `previewMarkdown` in that wrapper function instead so we won't have to edit all these functions across the system?

HenrikLam (changes requested):

williamjallen:
> There seems to be an x scroll bar in the preview textbox but not one for write textbox. Is this intentional? We should make the two behave identically if possible.

Just my opinion but I'd disagree here.  Take GitHub for an example.  The box I wrote this message in didn't scroll on the x-axis, even though many of the lines were wider than the width of the text box.  With the below code snippet, the preview (and what is ultimately rendered) does scroll.  I'd argue that it is best to disallow horizontal scrolling while editing and only allow horizontal scrolling for long text or long code in the markdown preview and rendered output.

Example of a long line which breaks (and should break in my opinion) in both the edit and preview modes:
ksjflksjdhflksjdhaflkjhflksjdhfldskjfhlkdsjhfksahjdflkjshdflkjdhsfkjashdflkjahsflkjhasdlkfjhaslkjfhalskjdhflkasjdhflkajshdlfkjhasdlkfjhalskdjhflaksjdhflkasjhdflkajhsflka

Example of a long line of code which breaks in edit mode and scrolls in preview mode:
```
ksjflksjdhflksjdhaflkjhflksjdhfldskjfhlkdsjhfksahjdflkjshdflkjdhsfkjashdflkjahsflkjhasdlkfjhaslkjfhalskjdhflkasjdhflkajshdlfkjhasdlkfjhalskdjhflaksjdhflkasjhdflkajhsflka
```

I believe that this is also the behavior of the markdown rendered in Submitty but I may be incorrect.

HenrikLam:
> I believe that this is also the behavior of the markdown rendered in Submitty but I may be incorrect.

Right now the normal text does wrap around on submitty. However, the code block has its own x scroll bar which I am ok with. However, this exists together with the x scroll bar of the textbox itself, and it doesn't seem to do anything.
image

williamjallen:
Oh, I see.  So, the issue is not that the rendered output is scrolling but rather than an unscrollable horizontal scrollbar appears on the main box.  I agree that that should be fixed then.

eddiekrystowski:
> > I believe that this is also the behavior of the markdown rendered in Submitty but I may be incorrect.
> 
> Right now the normal text does wrap around on submitty. However, the code block has its own x scroll bar which I am ok with. However, this exists together with the x scroll bar of the textbox itself, which doesn't seem to do anything.

Ah yes, thanks for bringing this up. I test on Mac which hides scrollbars like these by default so I don't see those. I'll be sure to hide those. The only place that there should be horizontal scroll bars are on long code snippets.

HenrikLam:
> It should be possible by doing any of the following

Oh I see, I must have done something incorrectly when I tried it the first time then. It seems to work for me now!

sbelsk (changes requested):
The new markdown interface looks great! I tested the new markdown on all the places on Submitty you listed.

1. On the grade inquiry interface, I typed a message and clicked "preview", and then submitted it while the text was still in preview mode. I could see my messaged was sent, but the textbox wasn't cleared because I was still on the "preview" tab. After I switched to the "write" tab, it  cleared.
Screenshot_1
2. On the edit poll page, the markdown interface does not show up, and I just see a text box (works fine on the new poll page though)
Screenshot_2

3. The overall comment in TA grading is still appearing to have the old markdown interface
Screenshot_3
4.  Suggestion: since the banner on the office queue page (the field you can specify on the course settings page) renders markdown, is it worth to put the markdown interface there too?
Screenshot_4
image

eddiekrystowski:
Thanks for testing extensively!

> 1. On the grade inquiry interface, I typed a message and clicked "preview", and then submitted it while the text was still in preview mode. I could see my messaged was sent, but the textbox wasn't cleared because I was still on the "preview" tab. After I switched to the "write" tab, it  cleared.

Oops, good point, I'll look into that along with the other grade inquiry markdown bug Barb mentioned in Slack.

> 2. On the edit poll page, the markdown interface does not show up, and I just see a text box (works fine on the new poll page though)

Definitely didn't forget about the edit poll page (I did)

> 3. The overall comment in TA grading is still appearing to have the old markdown interface
>    Screenshot_3

I think this is because the page uses TwigJS so a hard refresh and cache clear can be necessary (inspect element, then right click on refresh button and hit Empty Cache and Hard Reload), let me know if this doesn't fix it.

> 4. Suggestion: since the banner on the office queue page (the field you can specify on the course settings page) renders markdown, is it worth to put the markdown interface there too?

Great point, I'll look into that, should be easy enough.

eddiekrystowski:
Thanks for all of the testing everyone, I will summarize what changes I have made since there have been a lot of suggestions:

@HenrikLam 
* There is now an indicator that the markdown is loading when switching over to preview mode. 
  * In all cases, a blue spinner will show up next to the preview 
  * If the loading of the preview takes longer than 300ms, `Loading...` will show up in the preview text area. This can be seen by loading the TA Solution Notes page with pre-existing solutions.

https://user-images.githubusercontent.com/28243927/135769421-3b3a2756-2392-4414-8f5c-9c9f9ca72858.mov

https://user-images.githubusercontent.com/28243927/135769909-d216e762-3873-44b8-a586-e0a6d851d462.mov


* All `preview...Markdown` functions have been consolidated into a general `previewMarkdown` function in `server.js`. This function has also been updated to not make an ajax request when switching back to edit/write mode, which should speed up tab switching.
* Unnecessary horizontal scrollbars have been removed. (At least it looks this way on Mac Chrome, let me know if they are still present on Windows)
* Considering ways to update markdown preview as the user is typing to reduce load times, but I'll leave that out of this PR since it is already quite large.

@sbelsk 
* When submitting an action on the grade inquiry page, the mode is now automatically switched back to edit/write mode so you shouldn't see the old preview anymore.
* The Edit Poll page has been updated to use the new markdown area.
* Some areas on the Course Settings page (the ones that I knew supported markdown) now can be edited with the standard markdown area interface.


Other:
* `MarkdownController.php` and `MarkdownView.php` have been modified to remove unnecessary routes/functions.
* Grader overall comments are now displayed with markdown (before they were just able to be written in markdown but the markdown was not actually being rendered when the student viewed them)
* Change `Update` button in TA Solution Notes page to `Save`
  * This button is white will turn blue when the solution has been altered from the previous solution

williamjallen (changes requested):
Code changes look good.  I do have a few more UI comments though:

1. At least to me, having the spinner appear next to the preview tab text seems a little awkward.  I feel like it would be better to remove the "Loading..." text and just put the spinner in the content pane while content is loading.

2. When clicking the control buttons (Link, Code, etc), the cursor is placed at the end of the text in the textbox, not after the text which was just added.  If you add some text and then try to add a link in the middle, the link template is correctly added in the middle but the cursor position is moved to the end of the text box.

3. Is the extra box around the preview area necessary?  I feel like the rendered output could just be placed directly under the header now and it would look fine.
Untitled

4. I know in my last review I suggested some changes to make the buttons on the right not add text to the text area when in preview mode.  Now that I've spent some more time playing around with it, I wonder if it's even necessary for them to be visible when in preview mode.  If you look at GitHub, you can see that switching to preview mode hides all of the editing options completely.  Would it be better to do something similar here?

5. Did we decide whether long text should break or not?  In GitHub, long strings of text break in preview mode when they reach the end of the text area but long lines of code do not.  Should it be the same here?
Untitled

eddiekrystowski:
Thanks for looking through all of the code again!

> 1. At least to me, having the spinner appear next to the preview tab text seems a little awkward.  I feel like it would be better to remove the "Loading..." text and just put the spinner in the content pane while content is loading.

Good point, I'll do that.

> 2. When clicking the control buttons (Link, Code, etc), the cursor is placed at the end of the text in the textbox, not after the text which was just added.  If you add some text and then try to add a link in the middle, the link template is correctly added in the middle but the cursor position is moved to the end of the text box.

Thanks for catching this.

> 3. Is the extra box around the preview area necessary?  I feel like the rendered output could just be placed directly under the header now and it would look fine.
> 
> Untitled
> 

At first I envisioned there might be a time where someone would want to insert something into the "body" of the markdown area (such as something more besides the Accessibility message), but now I don't think that would really be necessary since this template is meant to be a compact piece of a form on any page. So I agree, I'll just remove the outer box.

> 4. I know in my last review I suggested some changes to make the buttons on the right not add text to the text area when in preview mode.  Now that I've spent some more time playing around with it, I wonder if it's even necessary for them to be visible when in preview mode.  If you look at GitHub, you can see that switching to preview mode hides all of the editing options completely.  Would it be better to do something similar here?

True, that can be done to avoid confusion. Should be easy.

> 5. Did we decide whether long text should break or not?  In GitHub, long strings of text break in preview mode when they reach the end of the text area but long lines of code do not.  Should it be the same here?
> 
> Untitled
> 

Oh yea, it definitely should, somehow that must have gotten broken.

williamjallen (commented):
Changes look good, I have tested it as thoroughly as I possibly can and I have not found any new ways to break it or achieve any undesirable functionality.  I'll approve this PR as soon as the tests pass.

As a completely trivial point which is likely not worth fixing, I did notice that the markdown re-renders (and displays the spinner) if you click the preview tab again while in preview mode.

site/public/js/server.js
```diff
@@ -1666,7 +1666,13 @@ function addMarkdownCode(type){
     let insert = '';
     switch (type) {
         case 'code':
-            insert = '```\ncode\n```';
+            const current = text[cursor];
+            const last_newline_pos = text.substring(0, cursor).split('').lastIndexOf('\n');
+            if (text.substring(last_newline_pos, cursor).length !== 1) {
+                insert = '\n';
+            }
+            console.log('current', current);
```
williamjallen:
```suggestion
```

Was this being logged deliberately?

eddiekrystowski: No, I added that just to test, my most recent commit should have gotten rid of that

eddiekrystowski: Just fiddling with getting e2e tests to pass now

williamjallen: Yeah, my bad.  GitHub didn't refresh or something and I didn't see the removal of the console log until after I submitted it.  Changes look good and seem to work as expected.

HenrikLam:
Found another quality of life issue where a new line is not inserted after the code block, but since this is a very small change I made a commit to the branch myself (hope I didn't mess anything up).

HenrikLam (approved):
Looks good!

sbelsk (approved):

williamjallen (approved):

PR #7154, opened 2021-09-29
[Feature:Submission] Add Switch to Most Recent button

loadedskiffer:

There has been some confusion about the disabled submit button, when not viewing the most recent version of a gradeable. Currently if not viewing the most recent submission the submit button text changes to "Switch to most recent version to make a submission" and is disabled. Clicking it does not actually switch to the most recent version which is what was causing confusion.
pr22_2

### What is the new behavior?
The submit button now always says submit but is disabled when not viewing the most recent submission, there is also a tooltip that says "switch to most recent version to submit". When the button is disabled, there is another button beside it that allows the user to easily switch to their most recent submission.

pr22_1

### DIFF
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -352,8 +352,7 @@
                     "content" : upload_message
                 } only %}
             </div>
-            <button type="button" id="submit" class="submit-gradeable btn btn-success">
-{#                {{ viewing_inactive_version ? 'Switch To Most Recent Version To Make New Submission' : 'Submit' }}#}
+            <button type="button" id="submit" class="submit-gradeable btn btn-success" {% if viewing_inactive_version %} title="Switch to most recent version to submit"{% endif %}>
                 Submit
             </button>
         {% if viewing_inactive_version %}
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -355,11 +355,11 @@
             <button type="button" id="submit" class="submit-gradeable btn btn-success" {% if viewing_inactive_version %} title="Switch to most recent version to submit"{% endif %}>
                 Submit
             </button>
-        {% if viewing_inactive_version %}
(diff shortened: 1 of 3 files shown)


shailpatels:
I would suggest this feature be added to notebook gradeables if possible since you are prevented from submitting on previous versions like normal submission gradeables

loadedskiffer:
> I would suggest this feature be added to notebook gradeables if possible since you are prevented from submitting on previous versions like normal submission gradeables

@shailpatels I'm not sure what you mean, the feature is implemented for notebook gradeables, this is a screenshot of notebook basic in the development course.
pr22_3

shailpatels:
Ah sorry about that! Disregard that comment then

shailpatels (approved):
Looks good and works as expected - the separate buttons make it more clear

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -352,11 +352,16 @@
...
             </div>
-            <button type="button" id="submit" class="submit-gradeable btn btn-success">
-                {{ viewing_inactive_version ? 'Switch To Most Recent Version To Make New Submission' : 'Submit' }}
+            <button type="button" id="submit" class="submit-gradeable btn btn-success" {% if viewing_inactive_version %} title="Switch to most recent version to submit"{% endif %}>
+                Submit
+            </button>
+        {% if viewing_inactive_version %}
+            <button type="button" id="switch_to_most_recent" class="btn btn-success" onclick="switchToMostRecent('{{ recent_version_url }}')">
+                Switch to Most Recent Version
             </button>
```
MasterOdin, line 361: I think we want to use a `a` tag here instead of a button for better accessibility as it's an anchor to another page, versus doing some action on the current page.

```suggestion
            <a href='{{ recent_version_url }}' class='btn btn-success' id='switch-to-most-recent'>
                Switch to Most Recent Version
            </a>
```

This would then allow getting rid of the added JS as well.

MasterOdin (changes requested):

loadedskiffer:
I decided to make the button blue because I think it makes more sense.
pr22_4

MasterOdin (approved):

PR #7159, opened 2021-09-29
[Bugfix:TAGrading] Fix download button duplication

DefinitiveAbove:
Going in and out of a PDF annotation duplicates the download button in the PDF toolbar.

### What is the new behavior?
Going in and out of a PDF annotation no longer duplicates the download button in the PDF toolbar.

### DIFF
--- a/site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
+++ b/site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
@@ -50,6 +50,10 @@
     //render download annotations button on toolbar
     $(function() {
         const download_annotations_button = document.getElementById('download-annotations-btn-template').content.cloneNode(true);
-        $('#pdf-buttons').append(download_annotations_button);
+        if ($('#pdf-buttons').has('#download-annotations-btn').length) {
+            $('#pdf-buttons', '#download-annotations-btn').replaceWith(download_annotations_button);
+        } else {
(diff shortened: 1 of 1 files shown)


sbelsk (approved):
Works as described. Thanks for noticing and fixing this!

site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
```diff
@@ -50,6 +50,10 @@
     //render download annotations button on toolbar
     $(function() {
         const download_annotations_button = document.getElementById('download-annotations-btn-template').content.cloneNode(true);
-        $('#pdf-buttons').append(download_annotations_button);
+        if ($('#pdf-buttons').has('#download-annotations-btn').length) {
+            $('#pdf-buttons', '#download-annotations-btn').replaceWith(download_annotations_button);
+        } else {
+            $('#pdf-buttons').append(download_annotations_button);
+        }
```
MasterOdin, line 57: Why do the replace? The button content does not seem to change and so it's just in essence a null-op?

```suggestion
        if (!$('#pdf-buttons').has('#download-annotations-btn').length) {
            const download_annotations_button = document.getElementById('download-annotations-btn-template').content.cloneNode(true);
            $('#pdf-buttons').append(download_annotations_button);
        }
```

MasterOdin (changes requested):

PR #7168, opened 2021-09-30
[Refactor:CourseMaterials] New Course Material URL

cjreed121:

Right now when you click a course material you are taken to a link that includes the course material ID. This can cause issues if instructors want a more permanent URL.

### What is the new behavior?
Now instead of using these links with IDs, clicking on the course material will take you to a new URL. This route is formatted where the relative path needed from /var/local/submitty/courses/SEMESTER/COURSE/uploads/course_materials will be in the URL rather than as a GET parameter.

### Other information?
This also changes a quick conditional error where graders couldn't see all course materials. It's a 2 character fix so I didn't make a full PR for it.

### DIFF
--- a/site/app/views/course/CourseMaterialsView.php
+++ b/site/app/views/course/CourseMaterialsView.php
@@ -24,14 +24,19 @@ class CourseMaterialsView extends AbstractView {
         $directory_priorities = [];
         $seen = [];
         $folder_ids = [];
+        $links = [];
+        $base_view_url = $this->core->buildCourseUrl(['course_material']);
 
         /** @var CourseMaterial $course_material */
         foreach ($course_materials_db as $course_material) {
+            $rel_path = substr($course_material->getPath(), strlen($base_course_material_path) + 1);
             if ($course_material->isDir()) {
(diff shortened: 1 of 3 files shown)


site/app/views/course/CourseMaterialsView.php
```diff
@@ -77,7 +88,7 @@ public function listCourseMaterials(array $course_materials_db) {
             if ($course_material->isDir()) {
                 continue;
             }
-            if ($this->core->getUser()->getGroup() != 1 && $course_material->getReleaseDate() > $date_now) {
+            if ($this->core->getUser()->getGroup() > 3 && $course_material->getReleaseDate() > $date_now) {
```
shailpatels, line 91: I think using `if( ! $this->core->getUser()->accessFullGrading() )` is a more readable check

https://github.com/Submitty/Submitty/blob/master/site/app/models/User.php#L349-L351

shailpatels (changes requested):
Adding a space to either the optional path or filename when uploading a course material prevents viewing it (most likely since the space is getting converted to a + in the url)

shailpatels:
@cjreed121 Can you fix merge conflicts?

DefinitiveAbove (approved):

PR #7171, opened 2021-10-02
[Testing:Plagiarism] Add plagiarism unit tests

sbelsk:
There exist currently no unit tests for the new `PlagiarismConfig.php` entity interface, and the existing test for `Interval.php` does not encompass much due do to changes made to it over the summer.

### What is the new behavior?
A new unit test file was created for `PlagiarismConfig.php`, and the tester for `Interval.php` has been made more extensive.


site/tests/app/libraries/plagiarism/IntervalTester.php
```diff
@@ -7,18 +7,85 @@
...
+        $this->my_interval[0]->updateEnd(22);
+        $this->assertEquals($this->my_interval[0]->getEnd(), 22);
+
+        $this->my_interval[0]->updateType("specific-match");
+        $this->assertEquals($this->my_interval[0]->getType(), "specific-match");
+    }
+
+    public function testOtherMatchingPositions() {
+        // TODO: should the interval class still support input matching
+        // regions with no start and end positions?
```
williamjallen, line 48: Yes, we still need to support this.  It's perhaps a little confusing the way we have it currently but this is how the matching users popup list is generated.  If you look in `Interval.php` on line 68, it's adding a user+version+source gradeable which is a match but doesn't have start and end positions because the user+version+source gradeable is not the current user on the right and thus doesn't need them.

williamjallen (changes requested):

bmcutler:
@williamjallen 
Please review this PR :)

williamjallen (approved):
Changes seem reasonable to me.  We can make any further iterative improvements in future PRs.

PR #7176, opened 2021-10-03
[Bugfix:TAGrading] Allow for empty solution notes

### LINKED ISSUE #7096: Can't Delete All Text in Solution Notes
**Describe the bug**
Won't let user save if they delete all text in solution notes textbox

**Expected behavior**
this should be ok

**To Reproduce**
1. sign in as instructor or TA
2. click "Grade" for a gradeable that has manual grading enabled
2.5. (this is unintuitive, but you may need to click "View All" button if it says "No grading to be done! :)" but you actually do ... that's a separate issue)
3. Click "Grade" for a random student / team
4. navigate to "Solutions/Notes" tab up top
5. write something in solutions notes and save, then delete and try to save
6. you will get error toast "Something went wrong while updating the solution..." even though this should be valid

joekrystowski:
You can not save if you delete all of the text in the solution notes.

### What is the new behavior?
Fixes #7096 Allows for solution notes to be saved when empty.

### DIFF
--- a/site/app/controllers/grading/ElectronicGraderController.php
+++ b/site/app/controllers/grading/ElectronicGraderController.php
@@ -2967,9 +2967,6 @@ class ElectronicGraderController extends AbstractController {
         if (!$gradeable) {
             $error = "Invalid Gradeable ID given!";
         }
-        elseif (empty($solution_text)) {
-            $error = "Please provide some non-empty solution";
-        }
         elseif ($componentItempoolInfo['is_linked'] && empty($itempool_item)) {
             //Itempool must be non-empty when component is linked with the itempool
             $error = 'This component expects only non-empty itempool-item!';


williamjallen (changes requested):
While this is a good starting place, I believe that it leaves empty solution notes in the database.  Ideally, blank solution notes should be deleted from the database.

joekrystowski:
> While this is a good starting place, I believe that it leaves empty solution notes in the database. Ideally, blank solution notes should be deleted from the database.

I thought about that as well, but for the solution notes the database keeps record of all past entries as well. So if I were to save then delete the entry that would effectively have the same behavior as not allowing blank entries (the original issue) because the solution would revert to what was there before.

williamjallen:
Oh, ok, makes sense.  Changes look good then.

williamjallen (approved):

MasterOdin:
Is there any UI differences when displaying a solution note that is empty but had been previously filled, and a solution note that had never been filled?

joekrystowski:
> > Yes. When a solution note has never been filled it there will be a header above the solution box that tells you there is no solution. When a solution note has been filled before but is empty now it does not do that.
> 
> 
> 
> Is that behavior desirable? Why would we want to treat them differently? This may be something you discussed with @bmcutler, just want to document this in case we ever go back and look at this PR in the future.

The general understanding I got when we discussed this in the meeting was that this feature was rarely used so a simple fix should suffice for now. I agree at some point it would be nice to have it at least emulate a state of no solution (even though there are entries in the database) when the solution box is empty.

williamjallen:
Another possible option here would be to pretend that there are no solution notes when the most recent solution notes are blank.  e.g., It would be impossible to tell that there used to be solution notes from the client-side.

PR #7177, opened 2021-10-03
[Bugfix:System] Fix db dumps to work on postgresql 10

MasterOdin:

The DB dump in #6966 introduced syntax in the DB dumps that did not work on postgresql 10, which is used by Ubuntu 18.04. This new syntax is a new default setting (`default_table_access_method`) and then an alternative term (`FUNCTION` instead of `PROCEDURE`).

### What is the new behavior?

The `SET` for default arg of `default_table_access_method` has been removed and we now now use `PROCEDURE` for dumps. This is done automatically as part of migrator, and tests added for it. There's no change in behavior of how the dumps work between postgresql 10 or postgresql 12 (Ubuntu 20.04).

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -120,7 +120,6 @@ $_$;
 
 SET default_tablespace = '';
 
-SET default_table_access_method = heap;
 
 --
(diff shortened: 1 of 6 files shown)


MasterOdin:
Vagrant Up test run: https://github.com/Submitty/Submitty/actions/runs/1301051698

PR #7178, opened 2021-10-03
[Bugfix:Submission] Fix incorrect late day message

### LINKED ISSUE #7120: Gradeable view has incorrect message for late days
**Describe the bug**
When a gradeable is submitted late, a banner at he top of the page describes the status of the gradeable with a message. There is a case where too many late days were attempted to be used. There were enough late days left, but not enough late days allowed for the given gradeable. The message given is incorrect for this case.

**Expected behavior**
The message should say:
Your active version was submitted __ day(s) after the deadline, and you would be charged __ late day(s) for this assignment, but your instructor specified that __ late day(s) may be used for this assignment.

**To Reproduce**
1.  Give a student n+1 late days
2.  Create a gradeable that allows n late days to be used
2.  Submit the assignment n + 1 days late

**Configuration**
 - OS: Windows
 - Chrome

**Screenshots**
Current:
late_day_message_late_this_assignment_old

Expected:
late_day_message_late_this_assignment_new

**Additional context**

joekrystowski:
Currently, if you have used more late days than there are allowed on the assignment, but you have that many late days, you will be given the wrong error message.
Screen Shot 2021-10-03 at 4 35 55 PM

### What is the new behavior?
Fixes #7120 Under the above conditions, the user will see the correct error message that accurately describes the problem with their submission. 
Screen Shot 2021-10-03 at 4 36 06 PM


jportorreal00 (changes requested):
This should technically fix this issue for this specific late day message for this case. However, there are some other issues that happen for each of the cases. There shoulod probably be other variable changes as well.

`$late_days_remaining` gives us the number of late days remaining at the end of **ALL** gradeables and late day events.

`$late_day_budget` gives us the number of late days we are allowed to use for this specific assignment.

`$late_days_remaining` will not always be accurate since the number of late days remaining at the end of all gradeables is not guaranteed to be the same number as the number pf late days remaining after a specific gradeable event. So in this case we may want to use `$late_day_info->getLateDaysRemaining()`.

tkoz0 (approved):
I tested this and it looks good.

PR #7179, opened 2021-10-03
[Refactor:Forum] Remove unused models

MasterOdin:

A forum refactor was started in #3760 over 2 years ago (05/2019). We never picked up the thread again on completing this refactor, and the original developer has long since moved on.

### What is the new behavior?

The code is removed as it is unused, and as we would want any refactor to use doctrine, is unlikely to be useful going forward. As such, the code has been fully removed.


cjreed121 (approved):
Doctrine entities are almost done so this fits well.

Issue #7017, opened 2021-09-04
Visual cue for pinned items that will expire soon


bmcutler:
**What problem are you trying to solve with Submitty**
Some announcements are intended to be "permanent" (e.g., here's the webex link for office hours for the entire term).
Vs. some announcements are temporary (e.g., here's info about this week's homework) and should expire in 7 days (the default).

It's not obvious that the dates were expiration dates were set correctly unless you click in to edit the post.  We might be surprised by a disappearing intended-to-be-permanent annoucenment.

**Describe the way you'd like to solve this problem**
Let's have 2 colors of pin icons for announcements.  I suggest:
gold pin = will expire in 7 or fewer days (the default)
submitty blue pin = expiration date is more than 7 days from now (date was edited from default)

While we do this, I also suggest that we change the color of the personal pin = bookmark.  
I suggest a red bookmark icon (I think this is common in other applications)

Thoughts?

PR #7184, opened 2021-10-04
[Feature:Forum] Adding visual cue for expiring threads

jenaybarela:

Addresses #7017 

### What is the new behavior?
When a pinned thread will expire in 7 or fewer days the pin will be gold otherwise it will be blue


shailpatels (changes requested):
I would suggest modifying the tool tip for expiring threads to indicate they are expiring so ther's some way for the user to figure out why the color is different.

Also has the design been discussed in a meeting? I think its confusing that expiring pins are yellow which is the default now

PR #7187, opened 2021-10-05
[Bugfix:CourseMaterials] Fix Extra Link Text

cjreed121:

If you have a link or a file that is not does not match the listed extensions, extra text will appear.

### What is the new behavior?
An extra condition check will happen to make sure we aren't looking at a name. In the case we can now display the name.

### DIFF
--- a/site/app/templates/course/CourseMaterials.twig
+++ b/site/app/templates/course/CourseMaterials.twig
@@ -64,12 +64,11 @@
                     <a href="{{ course_material.getUrl }}" onclick="markViewed([{{ course_material.getId }}])" onauxclick="markViewed([{{ course_material.getId }}])">{{ course_material.getUrlTitle }}</a>
                 {% else %}
                     <i class="fas fa-file" style="vertical-align: text-bottom;"></i>
+                    {{ name }}
                 {% endif %}
                 {% set extension = name|split('.')|last|lower %}
                 {% if '.' ~ extension in ['.pdf', '.jpg', '.jpeg', '.c', '.cpp', '.s', '.twig', '.py', '.java', '.png', '.txt', '.h', '.html', '.php', '.js', '.sql', '.sh', '.md', '.csv', '.salsa', '.erl', '.oz', '.pl', '.hs', '.gif'] %}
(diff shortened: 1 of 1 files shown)


loadedskiffer (approved):

site/app/templates/course/CourseMaterials.twig
```diff
@@ -64,12 +64,11 @@
                     <a href="{{ course_material.getUrl }}" onclick="markViewed([{{ course_material.getId }}])" onauxclick="markViewed([{{ course_material.getId }}])">{{ course_material.getUrlTitle }}</a>
                 {% else %}
                     <i class="fas fa-file" style="vertical-align: text-bottom;"></i>
+                    {{ name }}
                 {% endif %}
                 {% set extension = name|split('.')|last|lower %}
                 {% if '.' ~ extension in ['.pdf', '.jpg', '.jpeg', '.c', '.cpp', '.s', '.twig', '.py', '.java', '.png', '.txt', '.h', '.html', '.php', '.js', '.sql', '.sh', '.md', '.csv', '.salsa', '.erl', '.oz', '.pl', '.hs', '.gif'] %}
                     <a class="popout-item" target="_blank" href="{{ display_file_url }}?course_material_id={{ course_material.getId }}" aria-label="Pop up {{ name }} in a new window" tabindex="0" style="text-decoration: none;">{{ name }} <i class="fas fa-window-restore" title="Pop up the file in a new window"></i></a>
```
MasterOdin, line 71: It seems like if we enter this if statement, we will output `{{ name }}` as part of the `a` tag, and so if we hit the `else` above and this `if`, then we output `{{ name }}` twice.

Additionally, having stared at this whole thing, should we tuck all the lines between 69-75 under the `else` branch of the `if course_material.isLink` check?

MasterOdin (changes requested):
Could you include a screenshot of the bug you're fixing? I'm unclear what "extra characters" means in this context. Like is it extra spaces? The name repeated?

site/app/templates/course/CourseMaterials.twig
cjreed121: I forgot to push my last commit so now it should be up to date. The name issue no longer occurs. Yeah I'll try to tuck the whole thing in the else to make it a bit better organized and update the PR text.

loadedskiffer (approved):

PR #7195, opened 2021-10-07
[Bugfix:TAGrading] Fix checkpoint marks filling incorrectly

### LINKED ISSUE #7124: Checkpoint marks fill incorrectly
**Describe the bug**
If you are in the grading page for a checkpoint gradeable, it may appear as some people got a checkpoint but when you refresh they were never given. Look in additional context for extra info about this.

**Expected behavior**
Only checkpoints that were actually given should mark on other devices/windows.

**To Reproduce**
1. Log into the grader and instructor account (2 separate windows or devices at same time)
2. Go into the grading page of the Grading Lab gradeable [http://localhost:1511/courses/f21/sample/gradeable/grading_lab/grading?view=](http://localhost:1511/courses/f21/sample/gradeable/grading_lab/grading?view=)
3. You will see the grader only has access to sections 4 and 5 while instructor can see all sections
4. On the instructor account grade one of the first people in section 1
5. You should see the marks also show on the grader account's window but in their section 4 (section 4 already has a few marks so make sure to do a few tests if it looks like nothing changed)

**Additional context**
The main reason this seems to be happening is because the cells have a numeric ID that grows from what they can see. The socket server sends messages based off of this cell ID (unrelated to their user ID) to mark that as full, half, or empty.

So if 2 graders see different sections and grade at same time, they will see each others marks in their own section while none of those marks are actually reported. A refresh will of course fix this issue but a grader may see a slow build up of random marks over time until they refresh.

There should be either different or additional data sent over the socket server so this issue does not happen.

DanJanikowski:

Fixes #7124 for both checkpoint and numeric gradeables

### What is the new behavior?

Checkpoints will now only fill for the students whose grades were updated. If the student does not appear in the other graders table then no cells will visibly change.


site/public/js/simple-grading.js
```diff
@@ -667,34 +670,34 @@ function setupSimpleGrading(action) {
...
-                input_cell[0].selectionEnd - input_cell[0].selectionStart == 0))) {
-            event.preventDefault();
-            movement("right");
-        }
-        else if(event.code === "ArrowDown") {
-            event.preventDefault();
-            movement("down");
-        }
-    });
+    // // default key movement
```
DefinitiveAbove, line 673: If this isn't needed anymore it might be better to just remove it

DefinitiveAbove (changes requested):
Looks good to me; just one comment about the commented out code.

site/public/js/simple-grading.js
DanJanikowski: Ended up readding this, realized there's a minor bug with arrows keys on checkpoint gradeables.

DefinitiveAbove (approved):

PR #7229, opened 2021-10-11
[Feature:Plagiarism] Indicator for new run available 

williamjallen:
Many Lichen runs take a significant amount of time to complete meaning that instructors may start plagiarism runs and then return to them later.  In addition, a future PR will implement the nightly-rerun feature which will automatically run plagiarism detection in the background at low usage hours.  In both of these cases, it can be helpful to know which runs have been viewed and which ones haven't.

### What is the new behavior?
All accesses of the plagiarism results page are now logged in the database and runs which have not yet been accessed by the current user are given a different style in the plagiarism main page UI.  In addition, the plagiarism main page UI has been completely overhauled to reduce the amount of horizontal space needed for each row of the table.

Untitled

In addition to the new run indicator, I forgot to include data about the state of the provided code for each run in the last PR.  In hindsight, it would have probably been better to create a separate PR to add it but it is so intertwined with the code of this PR that I have left it in.

Once this PR is merged, all of the necessary data will have been moved to the database and all future changes will be able to be fully backwards compatible.  This means that the feature flag will be able to be removed from plagiarism sometime shortly after this PR is merged.


sbelsk (approved):
I really like this new UI! Works as described.

PR #7237, opened 2021-10-13
[Feature:TAGrading] Numerical Gradable redesign

### LINKED ISSUE #6953: Numeric/Text is unable to save changes; editing is broken for text fields
**Describe the bug**
Numeric/Text gradeable fails to save the contents of the text field. When attempting to edit the contents of the Text field, arrows do not work, so there is no way to move the insertion point to make edits. You can only select all contents and delete it and then start typing over again.

**Expected behavior**
The contents of all fields should be saved. Arrows should work to move the insertion point inside the Text field.

**To Reproduce**
1. Navigate to the gradeable page of a Numeric/Text gradeable. In my specific example, this is course CSCI-4963. The gradeable is flipped01. The corresponding URL is https://submitty.cs.rpi.edu/courses/u21/csci4963/gradeable/flipped01/grading?view=all The configuration of the gradeable is shown in screenshot 1.
2. Enter the value for the numeric field and the Text field and attempt to move to another row. Submitty marks the text field with red and shows an error message. See screenshot 2. The error message doesn't really tell me what I did wrong.
3. Attempt to click in the text field and press arrow to the left. Instead of moving the insertion point to the desired word that needs to be edited, it moves the focus to another field and triggers the error message. Clicking in the middle of the text also doesn't move the insertion point but instead selects the entire text. Pressing right arrow has no effect whatsoever. See the animated gif.

**Configuration**
 - Browser: Mozilla Firefox 90.0.2 (64-bit)

**Screenshots**
image
image
WaqzbfSizI

**Additional context**
N/A

HenrikLam:

Closes #6953

### What is the new behavior?
There is now an overhead sticky text field for editing the numerical and string entries. You can also double click a cell to be able to move your cursor around the text (to mimic the functionality of google sheets). Also UI change to have a background around the sticky text fields.


HenrikLam:
Made a couple changes since the last draft, I moved the buttons settings button and view statistics button to what I think makes sense (let me know what you think). Also made the edit zone also visible, but is disabled when the page is first loaded. It is enabled after you select a cell to edit. Also made the box bigger so it fits the whole content width since I moved the buttons.

shailpatels (changes requested):
It looks like the socket connection is being closed immediately after connecting on the simple grading page:

```
WebSocket: Closed websocket.js:70:29
```

This is preventing the live updates if multiple users are on that page.

Also if two users are on the same page and they make an update, the second user's screen is not updated but if they change the same cell they get an error 

```
Error: activeTab is undefined 7 undefined
Save error: displayed stale data (0) does not match database (1) server.js:941:25
```

shailpatels (changes requested):
This interface shows up for simple grading checkpoint entry, I don't think there's any reason to include this functionality in that page since you can't enter any numbers/text anyways? 

image

Also is there a reason the fixed position panel is not aligned to the top of the screen? It also does not fit the width of the main column containing the entry boxes

image

HenrikLam:
> It looks like the socket connection is being closed immediately after connecting on the simple grading page:
> 
> ```
> WebSocket: Closed websocket.js:70:29
> ```
> 
> This is preventing the live updates if multiple users are on that page.
> 
> Also if two users are on the same page and they make an update, the second user's screen is not updated but if they change the same cell they get an error
> 
> ```
> Error: activeTab is undefined 7 undefined
> Save error: displayed stale data (0) does not match database (1) server.js:941:25
> ```

Do you know what could have caused the WebSocket issue? Also if you've tested it on your local machine, could you also test the master branch? For some reason the test is still failing for me on my local machine after a vagrant destroy and up due to a save error.

site/app/templates/grading/simple/Display.twig
```diff
@@ -13,18 +13,7 @@
...
-            {% if action == 'lab' %}
-                <div>
-                    <p class="not-sticky">Undo/Redo Changes</p>
-                    <button id="checkpoint-undo" class="btn btn-primary key_to_click" tabindex="0" onclick="checkpointRollTo('{{ gradeable.getId() }}', -1)" disabled>Undo <i class="fas fa-undo"></i></button>
-                    <button id="checkpoint-redo" class="btn btn-primary key_to_click" tabindex="0" onclick="checkpointRollTo('{{ gradeable.getId() }}', 1)" disabled>Redo <i class="fas fa-redo"></i></button>
-                </div>
-            {% endif %}
-        </div>
-
+	<div class="column-wrapper" style="display: inline-block; width: 100%;">
```
williamjallen, line 16: Should the inline CSS be moved to an external stylesheet?

site/app/templates/grading/simple/Display.twig
```diff
@@ -109,18 +98,36 @@
...
-            <label for="csvUpload">Upload CSV
-                <input class="csvButtonUpload" type="file" id="csvUpload" accept=".csv, .txt">
-            </label>
-            <p>
-                The CSV file should be formatted as such:
-                <br />
-                user id,first name,last name,grade1,grade2,...,total points earned,text1,text2,...
-            </p>
-        </div>
+		<p style="margin-bottom: 1vh">
```
williamjallen, line 101: In general, using `vh` units for this type of thing is challenging to work with in the future.  It would be better to use `em`, `rem`, or `px` to accomplish this task unless I'm misunderstanding why `vh` was chosen.  Also, this is another thing which could be moved to an external stylesheet.

site/app/templates/grading/simple/Display.twig
```diff
@@ -109,18 +98,36 @@
...
+			The CSV file should be formatted as such:
+			<br />
+			user id,first name,last name,grade1,grade2,...,total points earned,text1,text2,...
+		</p>
+		<label for="csvUpload">Upload CSV
+			<input class="csvButtonUpload" type="file" id="csvUpload" accept=".csv, .txt">
```
williamjallen, line 107: Can the ID and class be converted to be of the form `csv-upload` as per the Submitty conventions [here](https://submitty.org/developer/coding_style_guide/css)?

site/app/templates/grading/simple/Display.twig
```diff
@@ -109,18 +98,36 @@
...
+		</label>
     {% endif %}
-
+	<div class="column-wrapper" style="display: inline-block; width: 100%;">
```
williamjallen, line 110: same comment here about inline CSS.

site/public/css/simple-grading.css
```diff
@@ -24,6 +25,18 @@
...
 }
 
+.sticky-top .sticky-box {
+    border: 3px solid var(--background-blue);
+    padding: 1em;
+    background-color: var(--default-white);
+}
+
+.edit-zone {
+    width: 70vw !important;
```
williamjallen, line 35: Is `vw` the best type of units to use here?  Any reason this can't be sized with percents relative to the parent element?

site/public/js/simple-grading.js
```diff
@@ -670,29 +670,31 @@ function setupSimpleGrading(action) {
...
-        else if(event.code === "ArrowRight" && (!input_cell.length || (
-                input_cell[0].selectionEnd == input_cell[0].value.length &&
-                input_cell[0].selectionEnd - input_cell[0].selectionStart == 0))) {
-            event.preventDefault();
-            movement("right");
-        }
-        else if(event.code === "ArrowDown") {
-            event.preventDefault();
-            movement("down");
+        if ($(event.target).attr('clicked') == null || $(event.target).attr('clicked') < 2) {
```
williamjallen, line 676: Might want to use `===` here

site/public/js/simple-grading.js
```diff
@@ -670,29 +670,31 @@ function setupSimpleGrading(action) {
...
+            if(event.code === "ArrowLeft" && (!input_cell.length || (
+                    input_cell[0].selectionStart == 0 &&
+                    input_cell[0].selectionEnd - input_cell[0].selectionStart == 0))) {
```
williamjallen, line 679: Same here about `===`

site/public/js/simple-grading.js
```diff
@@ -670,29 +670,31 @@ function setupSimpleGrading(action) {
...
+                movement("left");
+            }
+            else if(event.code === "ArrowUp") {
+                event.preventDefault();
+                movement("up");
+            }
+            // if there is no selection OR there is a selection to the far right with 0 length
+            else if(event.code === "ArrowRight" && (!input_cell.length || (
+                    input_cell[0].selectionEnd == input_cell[0].value.length &&
+                    input_cell[0].selectionEnd - input_cell[0].selectionStart == 0))) {
```
williamjallen, line 690: Same comment here about `===`

HenrikLam:
> This interface shows up for simple grading checkpoint entry, I don't think there's any reason to include this functionality in that page since you can't enter any numbers/text anyways?
> 
> image
> 
> Also is there a reason the fixed position panel is not aligned to the top of the screen? It also does not fit the width of the main column containing the entry boxes
> 
> image

The div on the top should be aligned to the top, I agree with that and I will fix that soon. However, making the div span the whole content is giving me issue since it doesn't seem to update when I collapse the sidebar...

williamjallen (changes requested):

site/app/templates/grading/simple/Display.twig
HenrikLam: I would argue this isn't necessary since these styles aren't share for every column-wrapper though it is used by multiple.

shailpatels:
Ok sounds good, @HenrikLam the styling of that panel can be fixed later - smaller priority compared to removing it from simple gradeable + checkpoint gradeable 

I will try and see if the websocket issue exists on master although the test seems to passing on CI now ?

HenrikLam:
@shailpatels It seems to have passed once I moved the buttons back to a structure like how it used to be before this PR, and the testing seems to have passed. It seems a bit strange that this would be affecting the websocket so I am a bit concerned about the state of our testing for this gradeable.

shailpatels:
Hmm yeah the only thing I can think of is that there was a conflicting id or the ui changes were blocking someone the test was trying to click?

jportorreal00 (approved):

williamjallen (changes requested):
I found a bunch of issues.  There may be more issues but it's hard to tell what may or may not be related so I'll keep the list relatively concise.  Once these issues are fixed, I'll re-review and may have more comments.

1. The page misbehaves when it is scrolled.  When using trackpads and mice with smooth scrolling, the page will gradually come to a stop after the page is scrolled.  This is what triggered the shaking I believe.

https://user-images.githubusercontent.com/16820599/142924882-689eb586-1072-4f96-af25-e2530188446d.mov

2. I entered text into the first numeric column but the rightmost box turned red.  Furthermore, I could not make the red background go away by doing any action, except by refreshing the page.
Untitled

3. I entered text in a box and then reloaded the page and the text was gone.  Even more concerning is that the text was moved to a box for a different user.  I think the best way to describe it is with a screen recording:
https://user-images.githubusercontent.com/16820599/142925793-805cd049-7dd0-49e8-ab99-f07551aa8f3a.mov

4. I was unable to delete text from a box.  Refreshing the page led to the text reappearing.
https://user-images.githubusercontent.com/16820599/142926028-d9330279-db52-408a-9e9f-7463e473095b.mov

5. When I click on a numeric box, I have to delete the 0 and then replace it with the desired number.  I would prefer for the box to be cleared when the box is selected for the first time (and still has a blue background).  An alternative would be to have the 0 be a placeholder until the user enters a number.  
Untitled

6. It's possible to enter text into a numeric field and no warnings appear to indicate that the page is now in an invalid state.
Untitled

7. I'm not sure if this is related to this PR or not but is this menu relevant to the UI here?  In addition, I'd prefer to see a setting somewhere in this menu which allows users to change the hotkeys (or turn hotkeys off) so that the cursor position can be adjusted with the arrow keys.
Untitled

8. Again, not sure if it's entirely related to this PR or not but the stats menu appears to be broken.  I've tinkered with the boxes a bit and I have not seen anything but 0s in any of the entries in the stats menu.
Untitled

williamjallen:
Closing as per discussion at today's mergeathon.  Too much of the underlying code has changed for reviving this PR to be viable.

PR #7245, opened 2021-10-13
[Feature:Submission] Add late day cache logic

jportorreal00:

Current late days use previous late day events as context in order to determine the #of late days remaining and # of late days used.

### What is the new behavior?
Late day information will now be stored in a table. (Table is similar to Late day usage table). This will make accessing late day information a lot faster since we will not have to calculate it from scratch every time.

### Other information?

Tested Using late Day Usage table. Late days remaining/late days used should line up.

Ex:
image


DefinitiveAbove (approved):

BenKelly99 (approved):
Verified that late days seemed to work as they used to

Issue #6042, opened 2020-12-03
office hours queue statistics improvements


bmcutler:
small formatting requests:
left justify the days of the week, queue names, weeks.
right justify the number of interactions, number of students, other numbers.
don't display sub-seconds for the average help/wait time etc.

don't include clear outliers in the maximum or average help/wait time.  (discuss how to do this.)

allow instructors to view the detailed office hour information for a specific student.  Would like to know what days the student went for help (early in term, or recently), what days of the week, what queue, how long they were helped, and by what ta/mentor.  
Either search by username, or click on the name from the "view student specific stats" page.

micpap25:
Linking to #10372

micpap25:
Noting this typo while i'm here since it's on the same page 
![image](https://github.com/Submitty/Submitty/assets/34077605/74acdf49-6b34-4110-941b-e41e0a9eaa94)

PR #7252, opened 2021-10-14
[Feature:HelpQueue] Switch queue stats to use percentiles

elihschiff:

When I look at the office hours queue stats, outliers make the data less useful. This has been bugging me all semester. For example, in data structures at RPI this semester we have a 2min help queue. Our avg help time is about 3min, with our max help time being almost 50min. This is most likely caused by somebody forgetting to stop helping a student or somebody messing up and thinking they are helping a student in the regular queue.

### What is the new behavior?

This switches from using min/max/avg to 5th percentile/95th percentile/median (aka 50th percentile)
I believe this data is a lot more useful. I did leave in max wait time as well as I think that is a useful metric and it is less likely to be biased by people forgetting as mentors should be removing students who are not in the room when their name is called.

I am also open to feedback if 5/95 are not the best numbers to use here, or any other ideas people may have to address this problem.

Also addresses the `don't include clear outliers in the maximum or average help/wait time` section of #6042
image


williamjallen:
I like the changes here.  My only concern is that the table is now quite wide.

Personally, I think that it would be ok to have the headers be a little misleading and just say "max help time" like it did before when it's actually the 95th percentile, likewise for the other stats.  Otherwise, it just makes the page a little busy in my opinion.  A tooltip on the headers could contain more information on the exact calculations if needed.

For the minimum values, I personally don't see any need for the 5th percentile cutoff.  If TAs/mentors are watching the queue, those values should be relatively close to 0 for the large majority of the time.

PR #7259, opened 2021-10-18
[Bugfix:InstructorUI] Fix sorting of student activity

### LINKED ISSUE #7206: student activity dashboard column sorting is weird
**Describe the bug**
The sorting of empty/NULL values relative to non-NULL values on the Student Activity Dashboard is weird for some (all?) columns.

E.g.,  We sort a column by last access (students who accessed somethign the first day and dropped ...  student who accessed it today).  A student who has never accessed something (has last access date set to NULL) should be placed next to students who only accessed something the first day of class.  They are instead placed next to students who accessed it today.  

Please sanity check all columns and all sorting (increasing & decreasing) and all filtering.

joekrystowski:
Currently, empty/NULL values on the student activity dashboard are treated as the highest values when sorting. So, for example, a student with no polls answered will be ranked above someone with 10 polls answered, even though they should be below those with 1 poll answered.

### What is the new behavior?
Now, empty/NULL values are treated as the lowest values when sorting. Fixes #7206


loadedskiffer (approved):
all columns are sorted correctly

MasterOdin:
Sorry for the merge conflict here @joekrystowski! Are you able to resolve? The files you edited were moved to `site/ts` and `site/tests/ts`, and you'll need to re-apply the changes onto them.

site/public/mjs/activity-dashboard.js
```diff
@@ -69,8 +69,14 @@ export function comparator (row1, row2, n, dir) {
 }
 
 // if n == 0 or n == 8
-// returns true if x < y, digits < strings < empty strings
+// returns true if x < y, empty strings < digits < strings
 export function helper (x, y, i) {
+    if (x != '' && y == '') {
+        return false;
+    }
+    else if (x == '' && y != '') {
```
williamjallen, line 77: It would be better to use the strict (`===` and `!==`) operators here instead of `==` and `!=`

williamjallen (changes requested):

bmcutler:
@williamjallen 
Can you rereview & take over making the necessary changes to get this PR merged?

PR #7263, opened 2021-10-18
[Feature:SubminiPolls] Refactor viewing page

BenKelly99:
Currently, the results page and view poll page are separate pages and the view poll page is hard to view for instructors.

### What is the new behavior?
Now, they are combined into one page viewable by students are instructors alike.


eddiekrystowski:
@BenKelly99 I've added changes so that the graph floats to the right of the answers (when they are expanded), and will resize until the screen is small enough where it will wrap to the next line. Let me know if there is anything you want changed or if I missed anything! I tested on mobile and with different screen sizes and it fits well.

BenKelly99:
This looks great, but is strangely failing python linting tests. Anyone seen this?

sbelsk (commented):
Great work! I tested it and I have a few comments:
1. On the student page for viewing the poll, I think it might be slightly confusing that only the button for collapsing/displaying the poll info is there when the histogram is not released. What do you think about only displaying the two buttons if the histogram is released to the students, and otherwise display no buttons at all (and then only thing displayed to students would be the poll info)?
2. Is there a way to make histogram as wide as possible, or at least centered, when it is displayed below the poll question on narrow screens? Right now it looks a little squished:
image
and there still appears to be some space to the right of it 
image
3. Testing it in dark mode I noticed that it might be nicer to have a little bit of spacing between the poll info and the histogram panels
image

sbelsk:
By the way, I believe this PR will close the last checkmark on the list of feature requests in https://github.com/Submitty/Submitty/issues/6036

BenKelly99:
@sbelsk Should be ready for re-review. Rather than adjusting the spacing between the graph and the poll info, I've updated the graph to display properly in dark mode.

sbelsk (changes requested):
The recent fixes look very good! I found one more thing that should probably be addressed before I approve this, and I have another request. 

* When the poll responses a longer than just a few words, it seems that the response lines are being wrapped around even when the histogram is not toggled:
short_lines
if you do toggle the histogram, then some of the text gets covered:
short_lines_covered
and even when the histogram is not available for display for students, the text still gets wrapped:
short_line_even_when_not_visible


* It would be nice of the page remembered the preference for each the panels being open or closed when you refreshed the page. This should work per poll, for both student and instructors.

kraduc (changes requested):
In addition to the comments above, some additions may be warranted regarding the behavior of polls that contain images. 

- When a student uses the view poll page, the image associated with the poll is not visible with any combination of the toggles.

Student view in master branch:
master

Student view of same poll with new changes: 
pr

- The instructor view also does not show the image, but this appears to be consistent with the master branch. It might be worth considering if the instructor poll view should also include the image in the refactoring.

bmcutler:
@Viyerelu23333
In addition to reviewing this, can you please try to finish the issues noted above with layout, etc.

Viyerelu23333:
> if you do toggle the histogram, then some of the text gets covered

scr0

There is a `margin-right` applied to `#poll-info`.

---

> ...even when the histogram is not available for display for students, the text still gets wrapped.

scr1

This is handled by `flex` automatically.  For better readability, I applied `max-width: 73%` to the `#poll-info`.

---

I could not reproduce this issue on my end: see the screenshots above.

> When a student uses the view poll page, the image associated with the poll is not visible with any combination of the toggles.

Viyerelu23333 (approved):
Looks good to me and works on my end.

bmcutler:
@sbelsk If you review this and find problems, please open an issue and @Viyerelu23333 will address them in a new PR.

PR #7279, opened 2021-10-22
[Feature:HelpQueue] Send message to queue via sockets

### LINKED ISSUE #7092: Push message to all students waiting in the queue
Office hours are remote, and office hours are ending soon, but there is a long line of students who unfortunately won't be helped.  The TAs close the queue, and will soon empty the queue.  But they would like to push a message to all students before they do that -- apologizing for removing them without help and giving directions to other resources to ask for help (e.g., time & location for next office hours)

Option to type a message, and have it delivered to the student page via websocket.  

Question: Does that message disappear on forced page refresh?

loadedskiffer:

closes #7092

### What is the new behavior?
TA's and Mentors now have the ability to send a message to a selected queue. They can also clear the message and see all queues that have messages.

pr24_4

The "View Messages" button displays a table with queues and their corresponding messages/sent times.

The message is delivered to all students in the selected queue. Right now the message is hardcoded to last for 2 hours. If the student then leaves the queue or is kicked out, they can no longer see the message. If they rejoin, the message will be there. This feature is meant to be followed by a feature that allows the queue staff to also send the message via email so that if the queue closes, students can still see the message. There will also be a follow up feature to allow queue staff to easily send a message to an individual student.

pr24_2

When a message is sent, all students in the queue will receive a browser notification (if they allow it)
pr24_3

### Other information?
pr24_5


DefinitiveAbove (changes requested):
Current functionality looks good to me for the most part (had a few minor comments); I had a few comments below that might be good for this PR:

* I think that a notification/alert should be created/ran after receiving a message for the queue - if the queue is especially long, there might be students who don't necessarily have the queue page in the foreground, and doing so should immediately bring the message to their attention.
* It might be a good idea to add the date/time that the message was sent out into the message header that is displayed on the page in case content in the message contains time-sensitive information.
* This might be outside the scope of the issue that this is fixing, but it might be better to store the message+timestamp in the database to help avoid the issue of new users entering the queue not receiving a message.

site/app/templates/officeHoursQueue/QueueSocketMessage.twig
```diff
@@ -0,0 +1,24 @@
...
+            {% for queue in viewer.getAllQueues() %}
+                <option value="{{queue['code']}}">{{queue['code']}}</option>
+            {% endfor %}
+        </select>
+        <br>
+        <br>
+        <textarea name="socket-message" id="socket-msg" rows="15" cols="50" placeholder="Message"></textarea>
+        <br>
+        <br>
+        Note: This message will only be delivered if the student has the queue page open <br>
```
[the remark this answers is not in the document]
loadedskiffer: How'd you get the text to appear above the horizontal line?

loadedskiffer:
@DefinitiveAbove How would you suggest making the alert, I'm not sure how to shift focus to the tab if they don't have it in the foreground

site/app/templates/officeHoursQueue/QueueSocketMessage.twig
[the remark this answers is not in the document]
loadedskiffer: @DefinitiveAbove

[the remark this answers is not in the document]
DefinitiveAbove: I put it in the div with the form-title class; you might have to change the `block title` to `block title_tag` to get it to work

DefinitiveAbove:
> @DefinitiveAbove How would you suggest making the alert, I'm not sure how to shift focus to the tab if they don't have it in the foreground

I think the behavior is browser dependent (on current browser versions alerts may not pull focus to the current tab anymore), but two ways that I could think of to alert the user is either using the alert() function or using the Notifications API.

For the alert() function, I tested it on Chromium 95 and while it doesn't pull focus to the tab, it displays a small dot next to the icon - 
image

For the Notifications API, the user has to allow permissions (you can make the request permissions thing come up in code), but if you send a notification (on Windows with default settings at least), it shows the notification natively -
image

I'm not sure which would be better in this case though (maybe Notification first and fallback to alert if permission denied?)

loadedskiffer:
@DefinitiveAbove I used the Notification API to add alerts, I tested it on edge, chrome and Firefox on Windows and it works as expected.

site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -190,8 +201,86 @@
...
+        window.socketClient = new WebSocketClient();
+        window.socketClient.onmessage = (msg) => {
+            switch (msg.type) {
+                case "queue_message":
+                    queueMessage(msg.message, msg.queue, msg.course);
+                    break;
+            }
+        };
+        window.socketClient.open('office_hours_queue');
+    }
```
cjreed121, line 272: We already have a websocket for this page. Could you add this type of message to that current one? Having 2 sockets for the same page seems unnecessary.

site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -190,8 +201,86 @@
...
+                    queue: queue,
+                    time: now.getTime(),
+                    formatted_time: now.toLocaleString(),
+                    course: course
+                }
+                localStorage.setItem('QueueMessageExpire', JSON.stringify(item));
+                localStorage.setItem('QueueMessage', message);
+                $('.queue-message').css('display', 'block');
+                $('#queue-message-msg').html(message);
+                $('#queue-message-msg-time').html(now.toLocaleString());
```
cjreed121, line 239: By setting raw html here and in 215-216, this could opens a XSS vulnerability by setting raw script tags. I tested it and was able to run some random javascript code.

cjreed121 (changes requested):
I pointed out a few issues within the code itself. Another issue becomes apparent with sockets which is that any message sent on the socket server gets broadcasted to all other clients. Anyone can send these socket messages which introduces an issue with messages here. This isn't an issue on other pages since the socket message only tells the page to fetch from the website rather than retrieving content from the socket message. I was able to start sending messages to other users from a student account. Storing these messages in the database could resolve this.

loadedskiffer:
@cjreed121 I made the changes to use the database to display messages. Now the websocket just triggers a page reload like in the other instances. I had to use local storage on the TA side in order to remove the message from the database after 80 minutes. (Just an arbitrary value we decided on in meeting) I also had to use Local storage on the student side in order to still display a message after they've been removed from a queue.

site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -311,5 +313,37 @@
       });
   }
 
+  function updateMessage(remove) {
+      location.reload();
```
cjreed121, line 317: What if instead of reloading the page, we make a network request to quickly fetch the queue message and have javascript render it in. This is how the office hour queue works. It helps reduce server load and provide a smoother experience to user as it's less sudden reloads. Just make sure that the code you use to set the content sets the text and not actual HTML to prevent an XSS issue.

site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -190,8 +222,56 @@
...
+                    let notification = new Notification("New Message From Office Hours Queue Staff");
+                }
+                // Otherwise, we need to ask the user for permission
+                else if (Notification.permission !== 'denied' || Notification.permission === "default") {
+                    Notification.requestPermission(function (permission) {
+                        if (permission === "granted") {
+                            let notification = new Notification("New Message From Office Hours Queue Staff");
+                        }
+                    });
+                //if notification permissions are denied
```
cjreed121, line 238: I was able to get prompted but before I could do anything the page reloaded and the pop up asking for permission went away. Removing refresh could help here for example. Also this seems like it will keep asking permission every time (maybe it won't but it definitely looks like it will). I wonder if maybe we ask for permission of notifications as soon as they load the office hours queue page. If they denied it, maybe we put a button on the page asking them if they ever change their mind they can request again. Notifications could also be used for maybe lets say when you are picked for help or when you are removed from the queue (probably another PR but having the notification change would help there).

cjreed121 (changes requested):
I think it would be great if we could avoid local storage for the message. It seems a bit strange to use between classes. My suggestions should allow you to remove the need for a TA to remove the message. I don't think a student who was removed would need to see the message. If they rejoin it would show it for them and if the time ended I'm guessing the message won't be important any more.

site/app/templates/officeHoursQueue/QueueStatus.twig
loadedskiffer: I got rid of the page reload problem. This will not ask for your permission every time, if you allow notifications it will never ask you again and if you deny notifications it will never ask you again. I like the idea of adding a button to allow users to change notification preferences and sending notifications for those other 2 scenarios I will make a quick PR for that after this one.

cjreed121 (commented):
1 potential issue I see is that queue admins have no way to see if there is a current queue message visible. Could that also be displayed on the grader queue page? Also lets say a course uses 1 office hour queue and the staff change shift, the new staff might be stuck with a previous incorrect queue message. Maybe a clear message could be implemented to revoke that message rather than needing to send a new one? Other than those I think this PR looks great.

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -531,7 +531,37 @@ public function changeContactInformation($queue_code) {
...
+            $this->core->addSuccessMessage("Message cleared");
+        }
+        else {
+            if (empty($_POST['socket-message'])) {
+                $this->core->addErrorMessage("Missing message");
+                return new RedirectResponse($this->core->buildCourseUrl(['office_hours_queue']));
+            }
+            $message = trim($_POST['socket-message']);
+            $this->core->getQueries()->setQueueMessage($code, $message);
+            $this->sendSocketMessage(['type' => 'update_message', 'queue_code' => $code, 'alert' => true, 'message' => $message, 'course' => $this->core->getDisplayedCourseName()]);
```
cjreed121, line 559: I think we should avoid adding any content like a message in the socket message. The reason is because any user can broadcast a message to all clients and now create notifications with arbitrary content.

site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -311,5 +313,19 @@
...
   }
 
+  function updateMessage(queue_code, alert, message, course) {
+    {% if viewer.inQueue() %}
+        let current_code = '{{ viewer.getCurrentQueueCode() }}'
+        if (current_code === queue_code) {
+            if (alert) {
+                queueMessage(message, course);
+            }
+            updateQueueStatus();
```
cjreed121, line 323: A solution I see to this problem is to create an update queue message function. You can create an endpoint returning json info of the latest queue message. Then the function will call that endpoint and know the message which can be passed to the queue message function. I don't necessarily think course is necessary. If you click on the notification I believe it takes you directly to that page.

site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -190,8 +204,29 @@
   </div>
 {% endif %}
 <script>
-    function toggleSwitchQueueForm(){
+
+    function queueMessage(message, course) {
+        if (Notification.permission === "granted") {
+            // If it's okay let's create a notification
+            let notification = new Notification("SUBMITTY(" + course + ") - New Message From Office Hours Queue Staff: " + message);
```
cjreed121, line 211: If you look at [https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification](https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification) you can see there is another constructor parameter. You can pass a body to the notification rather than using the title to hold the message. There is also some other fields that could let you customize this nicely.

cjreed121 (changes requested):

migration/migrator/migrations/course/20211104173201_queue_message.py
```diff
@@ -0,0 +1,33 @@
...
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
```
jportorreal00, line 32: Should this column be dropped on rollbacks?

site/app/models/OfficeHoursQueueModel.php
```diff
@@ -318,4 +318,36 @@ public function weekNumToDate($weeknum, $yearnum): string {
...
+                if ($elapsed_time->h < 2) {
+                    return $row['message'];
+                }
+                else {
+                    return null;
+                }
+            }
+            else {
+                return null;
+            }
```
jportorreal00, line 340: This is very nit-picky, but I think this chain of if statements could be flattened to one. Then you can just default to return null at the end of the function.

jportorreal00:
From what I remember, the function of the queue message is really nice. I like the addition of the time sent.

A few notes for the UI experience:
image

1. It feels a bit cluttered with the 3 sections open. If you want to have the "View Messages" button I definitely think it should toggle to "Hide Messages" once clicked
a. One idea is to show the table at all times and potentially have a "Clear" button within the row. Similar to the UI on the Open Queue table.
b. If there are no queue messages, there should probably be a row that says "no queue messages to show" 
2. When I click the "Send Message" button while updating the queue message, the modal closes. I didn't expect that since the UI didn't feel like I wasn't pressing a "submit" button. The flow might get a bit inconvenient for professors trying to update multiple queue messages. 

Really nice work!

migration/migrator/migrations/course/20211104173201_queue_message.py
loadedskiffer: I think that'd make sense, I also dropped the message_sent_time on rollback

site/app/models/OfficeHoursQueueModel.php
loadedskiffer: you're right, that was an extremely ugly if statement, thanks

loadedskiffer:
@jportorreal00 I made it so the button changes to "Hide Messages", I made it so theres a row that says no messages when there aren't any. As far as the modal closing, since the button submits a form to send the message, it reloads the page I'm not sure if there is a way to keep the modal open.

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -521,6 +521,53 @@ public function changeContactInformation($queue_code) {
...
```
cjreed121, line 549: `message` and `course` could be removed here as they aren't used.

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -521,6 +521,53 @@ public function changeContactInformation($queue_code) {
...
+            $this->core->addSuccessMessage("Message Sent To Queue");
+        }
+
+        return new RedirectResponse($this->core->buildCourseUrl(['office_hours_queue']));
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/office_hours_queue/get_queue_message", methods={"POST"})
```
cjreed121, line 557: This would probably make more sense as a GET request as there is no data we are changing or creating.

cjreed121 (changes requested):

site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -311,5 +313,46 @@
...
+   * @return {Promise} Rejects except when the response returns status 'success'
+   */
+  function ajaxGetQueueMessage(queue_code) {
+      return new Promise(function (resolve, reject) {
+          $.getJSON({
+              type: "GET",
+              async: false,
+              url: buildCourseUrl(['office_hours_queue', 'get_queue_message']),
+              data: {
+                  'csrf_token': csrfToken,
```
cjreed121, line 342: Since this is a GET request the csrf token can be removed.

cjreed121 (changes requested):

cjreed121 (approved):
Looks / works great!

site/app/templates/officeHoursQueue/QueueSocketMessage.twig
```diff
@@ -0,0 +1,86 @@
...
+                        <td>{{queue['message']}}</td>
+                        <td>{{ queue['message_sent_time'] }}</td>
+                    </tr>
+                {% endif %}
+            {% endfor %}
+            {% if count == 0 %}
+                <tr>
+                    <td>No messages to display</td>
+                    <td></td>
+                    <td></td>
```
jportorreal00, line 66: `<td colspan="3">No messages to display</td>` might look a but better

jportorreal00:
> @jportorreal00 I made it so the button changes to "Hide Messages", I made it so theres a row that says no messages when there aren't any. As far as the modal closing, since the button submits a form to send the message, it reloads the page I'm not sure if there is a way to keep the modal open.

Thanks for taking a look. I do see what you mean about the form. For now I think this is good. 

One last note. I was re-testing today and saw all of my old/stale messages. I had to go through and clear them one by one. Not even sure if that will even be an issue, but there could probably be a "stale message flush" (automatically? or even a button for manual flushing).

In spite of that I think the PR is good to go. Instructor feedback will be good to have so there can be any helpful updates.

jportorreal00 (approved):
Left a few small notes. They can be revisited after. Feature looks really good!

DefinitiveAbove (approved):
Looks good to me!

PR #7285, opened 2021-10-25
[Feature:Developer] Add esbuild for bundling & TypeScript

shailpatels:

* [X] Documentation has been updated/added if relevant (https://github.com/Submitty/submitty.github.io/pull/392)

Submitty appends a version to output files but when using es6 import syntax (`import foo from './bar'`),  duplicate files are shown in the browser source list.

### What is the new behavior?
esbuild has been added to bundle files so that all imports are resolved into a single file which works with our versioning method. 

There is now a `ts` folder that includes javascript modules and typescript files, the output is then placed into an `mjs` folder. This is so the existing PHP code of using `addInternalJS` does not need to be changed.

typescript is now supported if developers choose to write modules in it

### Other information?
No breaking changes, will add documentation for the new developer updates


site/public/ts/sql-toolbox.js
```diff
@@ -1,4 +1,4 @@
-import { buildCourseUrl, getCsrfToken } from './server.js';
+import { buildCourseUrl, getCsrfToken } from './server.ts';
```
MasterOdin, line 1: Is the file extension necessary?

.setup/install_submitty/install_site.sh
```diff
@@ -61,7 +61,7 @@ mv /tmp/index.html ${SUBMITTY_INSTALL_DIR}/site/public
 # copy the website from the repo. We don't need the tests directory in production and then
 # we don't want vendor as if it exists, it was generated locally for testing purposes, so
 # we don't want it
-result=$(rsync -rtz -i --exclude 'tests' --exclude '/site/cache' --exclude '/site/vendor' --exclude 'site/node_modules/' --exclude '/site/phpstan.neon' --exclude '/site/phpstan-baseline.neon' --exclude '/site/.eslintrc.json' --exclude '/site/.eslintignore' --exclude '/site/cypress/' --exclude '/site/cypress.json' --exclude '/site/jest.config.js' --exclude '/site/.babelrc.json' ${SUBMITTY_REPOSITORY}/site ${SUBMITTY_INSTALL_DIR})
+result=$(rsync -rtz -i --exclude 'tests' --exclude '/site/cache' --exclude '/site/vendor' --exclude 'site/node_modules/' --exclude '/site/phpstan.neon' --exclude '/site/phpstan-baseline.neon' --exclude '/site/.eslintrc.json' --exclude '/site/.eslintignore' --exclude '/site/cypress/' --exclude '/site/cypress.json' --exclude '/site/jest.config.js' --exclude '/site/.babelrc.json' --exclude '/site/public/mjs' ${SUBMITTY_REPOSITORY}/site ${SUBMITTY_INSTALL_DIR})
```
MasterOdin, line 64: Resolve the merge conflict here.

MasterOdin (changes requested):
Can you move the files from `site/public/ts` to `site/ts`? We are not going to make available the files via http calls, and it would also make more sense of the `tests` folder mimicking top-level folder structure.

Should also rename `site/tests/mjs` to `site/tests/ts`.

site/public/ts/sql-toolbox.js
shailpatels: jest was complaining when it was on `.js` but yeah not needed so removed

shailpatels:
Made requested changes and eslint should now be setup to support .ts files.

MasterOdin:
@shailpatels looks like something is not building correctly, causing the e2e suite to stall out.

shailpatels:
yeah saw that last night, I think the site install changes might broke something and the site is serving 502s? Trying to debug today

shailpatels:
@MasterOdin Fixed issue with e2e tests, should be set now

site/.build.js
```diff
@@ -0,0 +1,29 @@
+const fs = require('fs');
+const module_path = './ts/';
```
MasterOdin, line 2: Let's use `path.join` to help ensure this is cross-OS compatible:

```suggestion
const path = require('path');
const module_path = path.join(__dirname, 'ts');
```

site/.build.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+getAllFiles(module_path);
+
+require('esbuild').build({
+    entryPoints: files,
+    bundle: true,
+    minify: true,
+    sourcemap: true,
+    outdir: './public/mjs',
```
MasterOdin, line 28:
```suggestion
    outdir: path.join(__dirname, 'public', 'mjs'),
```

site/.build.js
```diff
@@ -0,0 +1,29 @@
...
+        const filename = file['name'];
+        if (!filename.includes('.js') && !filename.includes('.ts')){
+            return;
+        }
+
+        files.push(path + filename);
+    });
+}
+
+getAllFiles(module_path);
```
MasterOdin, line 21: Let's rewrite this as:

```suggestion
function getAllFiles(dir) {
    return fs.readdirSync(dir, { withFileTypes: true }).reduce((acc, entry) => {
        if (!entry.isDirectory() && entry.name.endsWith('.ts') || entry.name.endsWith('.js')) {
            acc.push(path.join(dir, entry.name));
        }
        
        return acc;
    }, []);
}

const files = getAllFiles(module_path);
```

Better to have a pure function than one that has side-effects. Let's also not introduce recursion until we need it, especially with regards to handling `entryPoints` below.

site/ts/server.ts
```diff
@@ -4,7 +4,7 @@
  * @param {string[]} parts
```
MasterOdin, line 4: We don't need to mark this file as an entryPoint right for esbuild as we're not directly including this anywhere. It's always implicitly bundled by the other scripts, and esbuild handles that automatically for us.

As such, let's move this file under `site/ts/utils/` directory and only bundle files at the top level as "entryPoints". I can see us adding additional files like `server.ts` that are included across multiple entryPoints, but are not entryPoints themselves.

MasterOdin (changes requested):
One last nit, otherwise I think good to merge.

MasterOdin (approved):

PR #7287, opened 2021-10-25
[Feature:Submission] Create Late Day Cache view

jportorreal00:

Currently, there is no bulk late day cache view

### What is the new behavior?
This creates a late day cache view in order to visualize the cache for all students in the course.

When the cache is being recalculated or flushed, the table and buttons will be greyed out and disabled. There will also be a status message present to show that the table is being updated.
image

image

### Other information?
 no
 tested along with all other late day views (mobile and not mobile)

Prerequisite: #7338


tkoz0 (approved):
I have tested the changes and given a thorough look at the code. It looks good to me and works as expected.

PR #7296, opened 2021-10-27
[Feature:TAGrading] Add attachment upload

### LINKED ISSUE #7134: Grader uploads a .pdf (or other file) as feedback from the grading process
For some courses, for some assignments, it may be appropriate/desirable to not use Submitty's manual grading rubric, but instead to prepare a detailed document outside of Submitty.  This file might be a pdf or image or text.

Similar to the overall comment box, let's allow any grader to attach one or more files to the student's assignment.  The student will be able to view/download these files in the same place they can see the annotated .pdfs.

DefinitiveAbove:
There is no attachment upload.

### What is the new behavior?
Closes #7134 
You can now upload attachments during grading. When grading a student, graders can now attach files that will appear below the overall comment in both the gradeable and the grading area.
image

image

image


sbelsk (changes requested):
Functionality:
1. Nice to interact with from the Instructor/TA interface, attachments show up properly on the student end :) 
2. There seems to be an issue with released peer gradeables. For example, I tried opening the "Released Peer PDF Homework" from a student account got the following error:
image
I also tried releasing the "Closed Peer Team Homework Grading" and logging in to view it from a student account that had a grade, and got the same error.
3. Question: do we want peer graders to be able to upload attachments in the overall comments section?

UI Suggestions:
1. Should we have the display of attachment look more similar to the existing UI for annotated pdfs? 
image
2. I think the box in the TA grading where you click to upload takes a lot of space currently, what if we did something similar to the way the discussion forum handles attachments, so instead of having 5 rows of text, we would just have one row with a single button that says "upload attachment" (no text necessary to clarify that there are currently no attachments, they just show up if you upload them and then you have the option to delete them)? image

sbelsk (commented):
Changes look good. One minor thing I noticed when I was testing thing again, is that if the text in overall comment is changed when the grades are released for a certain gradeable, then the button that says "VIEW GRADE" on the main page from the student view turns green to indicate some new action (this is an existing feature). Do we want this to be triggered as well if there are changes related to attachments (e.g., new file has been uploaded, file was changed, deleted, etc)?

sbelsk (changes requested):
Tested it again, when a new file is uploaded the student's button turns green as desired. We should also do this when attachments are deleted, which I believe would go here:

https://github.com/Submitty/Submitty/blob/ac5fdf3fd5ac7559df2b44f04d1b4c1cb0ace413/site/app/controllers/grading/ElectronicGraderController.php#L1119-L1123

DefinitiveAbove:
> Tested it again, when a new file is uploaded the student's button turns green as desired. We should also do this when attachments are deleted, which I believe would go here:
> 
> 
> 
> https://github.com/Submitty/Submitty/blob/ac5fdf3fd5ac7559df2b44f04d1b4c1cb0ace413/site/app/controllers/grading/ElectronicGraderController.php#L1119-L1123

I don't think we want to do it when deleting attachments since deleting attachments means that there is less info; a user who is expecting new information for the gradeable when the button turns green may be confused if they click on the gradeable and there's nothing new, but rather information is just lost (since there is no display showing which attachments were deleted since last access).

sbelsk (approved):

PR #7307, opened 2021-10-31
[Bugfix:Submission] Fix Line Wrapping on Notebook Inputs

eddiekrystowski:
Currently, lines aren't wrapping properly on notebook gradeable inputs. In inputs with regular text (not code), the text should be wrapping, but it is not.
image

### What is the new behavior?
The short answer inputs on notebooks will have wrapping enabled when no language is specified (plaintext). Otherwise, if a language is specified, there will be no line wrapping.
image


shailpatels (approved):
Looks good,
tested on mac firefox,chrome and safari

PR #7308, opened 2021-11-01
[Bugfix:HelpQueue] Don't show switching if no queue is open

loadedskiffer:

Currently, if there is only 1 open queue and a student is in that queue, the option to switch queues still displays even though there are no other open queues to switch into.

### What is the new behavior?
The form doesn't display if there is only 1 open queue, unless the queue that is open is not the queue the student is in. For example It could be the case that the student is in queue A, but we close that queue and open queue B.  They should be able to switch to queue B, even if there is only 1 queue option.

### DIFF
--- a/site/app/templates/officeHoursQueue/QueueStatus.twig
+++ b/site/app/templates/officeHoursQueue/QueueStatus.twig
@@ -96,8 +96,15 @@
       <input type="hidden" name="queue_code" value="{{ viewer.getCurrentQueueCode() }}"/>
       <button id="leave_queue" type="submit" class="btn btn-danger">Leave Queue</button>
     </form>
+      {% set my_queue = 'code' %}
       {% set open_queues = viewer.getAllQueues()|filter(queue => queue['open'] == 'true') %}
-      {% if open_queues|length > 1  %}
+      {% if open_queues|length == 1 %}
+          {% for queue in viewer.getAllQueues()|filter(queue => queue['open'] == 'true') %}
+            {% set my_queue = queue['code'] %}
+          {% endfor %}
+      {% endif %}
+
+      {% if open_queues|length > 1  or my_queue != viewer.getCurrentQueueCode() %}
       <form method="post" action="{{ base_url }}/{{ viewer.getCurrentQueueCode() | url_encode }}/switch">
           <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
           <input type="hidden" id="name_box" name="name" maxlength="20" value="{{viewer.getName()}}" style="margin:.5rem" required="required">
--- a/site/app/templates/officeHoursQueue/QueueStatus.twig
+++ b/site/app/templates/officeHoursQueue/QueueStatus.twig
@@ -103,7 +103,6 @@
             {% set my_queue = queue['code'] %}
           {% endfor %}
       {% endif %}
-
       {% if open_queues|length > 1  or my_queue != viewer.getCurrentQueueCode() %}
       <form method="post" action="{{ base_url }}/{{ viewer.getCurrentQueueCode() | url_encode }}/switch">
           <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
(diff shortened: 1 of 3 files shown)


site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -96,6 +96,14 @@
       <input type="hidden" name="queue_code" value="{{ viewer.getCurrentQueueCode() }}"/>
       <button id="leave_queue" type="submit" class="btn btn-danger">Leave Queue</button>
     </form>
+      {% set my_queue = 'code' %}
+      {% set open_queues = viewer.getAllQueues()|filter(queue => queue['open'] == 'true') %}
```
shailpatels, line 100: Is this something that should be handled on the view?

If the student should only ever interact with queues that are set to open then there is no reason to send the entire list of qs to the twig page.

I think adding a function like `getAllOpenQueues` to the model and  then filtering at the database level for open queues only is a better solution

shailpatels (changes requested):

site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -96,6 +96,14 @@
...
       <button id="leave_queue" type="submit" class="btn btn-danger">Leave Queue</button>
     </form>
+      {% set my_queue = 'code' %}
+      {% set open_queues = viewer.getAllOpenQueues()%}
+      {% if open_queues|length == 1 %}
+          {% for queue in viewer.getAllOpenQueues() %}
+            {% set my_queue = queue['code'] %}
+          {% endfor %}
+      {% endif %}
+      {% if open_queues|length > 1  or my_queue != viewer.getCurrentQueueCode() %}
```
shailpatels, line 106: I don't think the twig code above is needed. But I don't fully understand what its trying to do,
it looks like if there's exactly one open queue then set `my_queue` to the last open queue's code? That doesn't seem right - I'm guessing the `== 1` should be a `> 1`

You could do something like instead
```php
{% if open_queues|length > 1 or (viewer.getAllOpenqueues() | last)['code'] != viewer.getAllOpenQueues()) %}
```

If this if block is supposed to render the switch queues form, wouldn't the logic just be 
`if open_queues|length > 1` ? why do you need to compare queue codes

shailpatels (changes requested):

site/app/templates/officeHoursQueue/QueueStatus.twig
loadedskiffer: There is an edge case where the student is in queue A, but we close that queue and open queue B.  They should be able to switch to queue B, even if there is only 1 queue option.  (So testing if there is <= 1 queues open is not the right check in this case).

shailpatels: ok

shailpatels:
@loadedskiffer Added a comment with your explination

shailpatels (approved):
Looks good

site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -96,6 +96,18 @@
...
       <button id="leave_queue" type="submit" class="btn btn-danger">Leave Queue</button>
     </form>
+
+      {# Display the switch queue form if there is more than 1 open queue or the queue code we got does not match the 1 open queue 
+      e.g if an instructor closes the queue we are in and opens are new one#}
+
+      {% set my_queue = 'code' %}
+      {% set open_queues = viewer.getAllOpenQueues()%}
+      {% if open_queues|length == 1 %}
+          {% for queue in viewer.getAllOpenQueues() %}
```
MasterOdin, line 106: Why not re-use `open_queues` here and avoid a SQL call?

```suggestion
          {% for queue in open_queues %}
```

site/app/templates/officeHoursQueue/QueueStatus.twig
```diff
@@ -109,7 +121,7 @@
               </label>
               <select id="switch_queue_code" class="form-control" name="code" required="required" onchange="toggleSwitchQueueForm()">
                   <option value="">Queue Name</option>
-                  {% for queue in viewer.getAllQueues()|filter(queue => queue['open'] == 'true') %}
+                  {% for queue in viewer.getAllOpenQueues()%}
```
MasterOdin, line 124: Same here, can you re-use `open_queues` and avoid a SQL call?

```suggestion
                  {% for queue in open_queues %}
```

MasterOdin (changes requested):

MasterOdin (approved):

PR #7317, opened 2021-11-02
[Bugfix:InstructorUI] Fix duplicate null section on CSV upload

### LINKED ISSUE #7316: Duplicate NULL Registration Sections on Classlist Upload
To reproduce, try uploading a classlist CSV from instructor UI in "Manage Students" sidebar tab, with "NULL" registration section specified in the CSV. An example classlist CSV file with "NULL" registration section is [here](https://docs.google.com/spreadsheets/d/1hyubCKktMUGXLyTkgxQkd4cjTengNjEzeQjkhak2VN0/edit?usp=sharing).

The buggy behavior is that this adds a "NULL" registration section to the registration sections table database, rather than just using the null row in the table (the expected behavior). This causes two "NULL" registration sections to exist for the course.
Screen Shot 2021-11-02 at 2 28 36 AM
Screen Shot 2021-11-02 at 2 23 46 AM

listeph:

### What is the current / new behavior?
Fixes #7316

(Already described the bug and expected behavior with screenshots attached in the issue, so I won't repeat for conciseness).

### How to test
1. Log in as 'instructor' and create a new blank course for testing
2. go to "Manage Students" from course's sidebar
3. try uploading a classlist CSV file with "NULL" registration section such as [this](https://docs.google.com/spreadsheets/d/1hyubCKktMUGXLyTkgxQkd4cjTengNjEzeQjkhak2VN0/edit?usp=sharing)
4. Check in "Manage Sections" from sidebar now that only one NULL registration section exists in the course. You can also check the actual database more legitimately by doing `SELECT * FROM sections_registration` or `SELECT user_id, registration_section FROM users;` on your test course in pg_admin.


site/app/controllers/admin/UsersController.php
```diff
@@ -1000,14 +937,30 @@ public function uploadUserList($list_type = "classlist") {
...
-                $bad_row_details[$row_num + 1][] = $list_type === 'classlist'
-                    ? 'Registration section'
-                    : 'Grader-group';
-                if (!in_array('row4_validation', $bad_columns)) {
-                    $bad_columns[] = 'row4_validation';
+            /* Check registration for appropriate format. Allowed characters - A-Z,a-z,_,- .
+            Registration section is optional for graders, so automatically validate if not set.*/
+            if (isset($vals[$registration_section_idx]) && strtolower($vals[$registration_section_idx]) === "null") {
+                $vals[$registration_section_idx] = null;
+                $uploaded_data[$row_num][$registration_section_idx] = null;
```
listeph, line 944: So once again, this is the 1 line change for bug fix.

Explanation: changes to `$vals` (which is an alias to a row in `$uploaded_data`) doesn't reflect in the actual `$uploaded_data`. So later, when we do (I am abbreviating the logic in line 1070 of `UserController.php`):
`foreach ($uploaded_data as $row) {
      ...
      $this->core->getQueries()->insertNewRegistrationSection($row[$registration_section_idx]);
      ...
}`
it still inserts unchanged value, a.k.a. "NULL" string, into the database.

williamjallen (commented):
Code looks good

williamjallen (approved):

site/app/controllers/admin/UsersController.php
MasterOdin: Looking at the code, it looks like we set a value of `$vals` a second time during this loop, probably with the intention that'll stick till later?

It may be better to adjust the foreach to be:

```
foreach ($uploaded_data as $row_num => $vals) {
    // ... existing code
    $uploaded_data[$row_num] = $vals;
}
```

This ensures we update the row for `$uploaded_data` after manipulating it, without having to update both arrays each time.

An alternative would be to pass-by-reference in the foreach itself:
```
foreach ($uploaded_data as $row_num => &$vals) {
    // ... rest of the code
}
unset($vals);
```

where we pass-by-reference the value for `$vals` and so changing it will change the original data. I prefer to avoid this as requiring the `unset` after the `foreach` is somewhat unexpected for new people (see warning on https://www.php.net/manual/en/control-structures.foreach.php), and it's just a bit of a pitfall versus the above, and I don't think the data is so large we need to save on memory here.

MasterOdin (changes requested):

site/app/controllers/admin/UsersController.php
listeph: Good suggestion, and thanks for sharing about the pass-by-reference didn't know that was possible! I will request for your re-review once I test just to be safe.

MasterOdin (approved):

listeph:
Manually tested functionality again to be safe ✅

PR #7318, opened 2021-11-02
[Bugfix:TAGrading] Hide custom marks from peer graders

loadedskiffer:

Currently peer graders can still see the option to add custom marks however when they try to use them they get a permission error.

### What is the new behavior?
Peer graders don't have the option to use custom marks.


sbelsk (approved):
Works as described.

PR #7319, opened 2021-11-02
[Bugfix:Forum] Fixing CSS issues on pin borders 

jenaybarela:

Screen Shot 2021-10-13 at 10 44 40 AM

Currently the borders on the pins are missing for the instructors and the tooltips have not been updated to say gold pins are expiring soon.
### What is the new behavior?
Screen Shot 2021-11-02 at 12 05 58 PM
The boarder has been fixed, the tooltip text has been updated, and the bookmark color has been changed to red.

### DIFF
--- a/site/app/templates/forum/CreatePost.twig
+++ b/site/app/templates/forum/CreatePost.twig
@@ -5,16 +5,16 @@
             {% if userGroup <= 2 and activeThreadAnnouncement %}
 
                 {% set expiringVisual = (expiring) ? 'active-thread-remove-expiring-announcement' : 'active-thread-remove-announcement' %}
-                <a class="{{expiringVisual}} key_to_click" tabindex="0" onClick="alterAnnouncement({{ activeThread.id }}, 'Are you sure you want to unpin this thread?', 0, '{{ csrf_token }}')" title="Pinned announcement -- will expire soon" aria-label="Pinned announcement -- will expire soon"><i class="fas fa-thumbtack"></i></a>
+                <a class="{{expiringVisual}} key_to_click" tabindex="0" onClick="alterAnnouncement({{ activeThread.id }}, 'Are you sure you want to unpin this thread?', 0, '{{ csrf_token }}')" title="Pinned thread -- will expire soon" aria-label="Pinned thread -- will expire soon"><i class="fas fa-thumbtack"></i></a>
 
             {% elseif activeThreadAnnouncement %}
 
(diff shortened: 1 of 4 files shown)


sbelsk (commented):
Colors looks very nice :) They only thing is, I think the UI never actually uses the word "announcement" (even though the backend does), but it instead uses "thread" everywhere. So perhaps the text on the pages should agree with the terminology used in the the tooltips for consistency?

sbelsk (approved):

site/app/templates/forum/CreatePost.twig
```diff
@@ -5,7 +5,7 @@
             {% if userGroup <= 2 and activeThreadAnnouncement %}
 
                 {% set expiringVisual = (expiring) ? 'active-thread-remove-expiring-announcement' : 'active-thread-remove-announcement' %}
-                <a class="{{expiringVisual}} key_to_click" tabindex="0" onClick="alterAnnouncement({{ activeThread.id }}, 'Are you sure you want to unpin this thread?', 0, '{{ csrf_token }}')" title="Unpin Thread" aria-label="Unpin Thread"><i class="fas fa-thumbtack reverse_golden_hover"></i></a>
+                <a class="{{expiringVisual}} key_to_click" tabindex="0" onClick="alterAnnouncement({{ activeThread.id }}, 'Are you sure you want to unpin this thread?', 0, '{{ csrf_token }}')" title="Pinned thread -- will expire soon" aria-label="Pinned thread -- will expire soon"><i class="fas fa-thumbtack"></i></a>
```
MasterOdin, line 8: Per the [W3C spec](https://www.w3.org/TR/WCAG20-TECHS/H33.html) for accessibility for `a` tags titles:

> The title attribute is used to provide additional information to help clarify or further describe the purpose of a link.

Let's revert this change.

MasterOdin (changes requested):

site/app/templates/forum/CreatePost.twig
jenaybarela: Should I just revert the aria-label or the title as well?

MasterOdin: Both aria-label and title should be used, where the former is going to be what's read aloud by a screen reader and the latter used when mousing over. Ideally, both should describe what the link is going to _do_ with possible additional information, as you did in follow-up commits :+1:

Thanks!

MasterOdin (approved):

PR #7327, opened 2021-11-04
[Bugfix:Submission] Ignore empty preferred names

benjaminrsherman:

Currently the database query for identifying a student's name on the leaderboard will return an empty string if the student has never set a preferred name.

### What is the new behavior?
This change updates the database query to ignore empty strings in the user's preferred name, therefore falling back to the user's base name if they did not set a preferred one.


benjaminrsherman:
I'm not sure how to correctly format SQL queries, so please let me know if there's something I should change regarding that.

bmcutler (approved):
Seems to work!

PR #7333, opened 2021-11-07
[Bugfix:System] Specify date format when creating terms

dpshelio:
-------

* [ ] Tests for the changes have been added/updated (if possible) *Not sure how to test this bash script beyond what I mention below*

At the moment the `create_term.sh` script tries to use the dates as required in the input of the script.
However, the script doesn't consider the datestyle setting the database may had.

### What is the new behavior?

Behaviour is the same, however now it's more specific as it converts the input date specifying the format.
It also avoids problems when systems may have a DD/MM/YYYY configuration and accepts it by mistake for some input dates.

### Other information?

This won't change any current usage to the people that uses this script unchanged. And it will also work for people with other locales (I'm not sure how postgresql sets that default, but I'm in Europe and MM/DD/YYYY is not the default).

This was tested as now it works on my system:
#### Before change:

```bash
/usr/local/submitty/sbin/create_term.sh 05 fake22 06/21/2022 08/21/2022
```

```
ERROR: Failed to INSERT new term into master DB
ERROR:  date/time field value out of range: "06/21/2022"
LINE 3: VALUES ('05', 'fake22', '06/21/2022', '08/21/2022');
                                ^
HINT:  Perhaps you need a different "datestyle" setting.
```

#### After change:

```bash
/usr/local/submitty/sbin/create_term.sh 05 fake22 06/21/2022 08/21/2022
```

```
'05' term has been INSERTed into the master DB.
You may now create courses for the '05' term.
```

```sql
submitty=# select * from terms;
```

```
 term_id |  name  | start_date |  end_date  
---------+--------+------------+------------
 05      | fake22 | 2022-06-21 | 2022-08-21
(1 row)
```

Results are as expected! 🎉 

It took me a while to find this out, thanks go to [this answer in StackOverflow](https://stackoverflow.com/a/34761058/1087595).

### DIFF
--- a/sbin/create_term.sh
+++ b/sbin/create_term.sh
@@ -54,7 +54,7 @@ name=$2
 start=$3
 end=$4
 
-# Validate that start and end dates are properly formatted.
+# Validate that start and end dates are properly formatted (MM/DD/YYY).
 regex='^(0[0-9]|1[0-2])/(0[0-9]|[1-2][0-9]|3[0-1])/20[0-9]{2}$'
 
 if ! [[ $start =~ $regex ]] ; then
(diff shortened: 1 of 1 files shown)


MasterOdin (approved):
Thanks for the bugfix!

PR #7343, opened 2021-11-09
[UI/UX:InstructorUI] Colors refactor 

jenaybarela:

The colors file has become unmanageable. There are many duplicates which can make it difficult when trying to pick colors to use. This often causes people to just add new colors rather then look through the current ones to find one that works.

### What is the new behavior?
This removes duplicates from the dark mode section.


williamjallen:
Is there ever a need to use RGB colors in Submitty?  e.g., should we add a rule to the CSS linter to require hex colors?  I'm not sure if it's a good idea, just a thought.

sbelsk (changes requested):
Clicking around, things seem alright.

I did notice one thing, where in dark mode, when hovering on top of any dark-grey buttons (`btn-default`) anywhere on Submitty, their color turns almost white which makes the text very hard/impossible to read. for exmaple,
before hovering:
image
after hovering (my mouse is on the button, can't be seen in the screenshot):
image

sbelsk (approved):

PR #7344, opened 2021-11-10
[Bugfix:Submission] Upload bulk/student with previous files

### LINKED ISSUE #7190: drag & drop files individual vs bulk
**Describe the bug**
As instructor, go to a bulk upload gradeable.  

First, make a submission as yourself select "Normal Submission" - e.g., a file named foo.cpp.
After submitting, your files from your last submission are diplayed in the drop & drop area.

Next, go to do a bulk upload select "Bulk PDF Upload".
foo.cpp, is still in the drag & drop area.  If you press "submit", you get a javascript error / console frog robot.

You need to manually delete foo.cpp every time, and drag in the bulk pdfs you want to submit.

**Expected behavior**
Your personally submitted files should never display in the box if you select "Make Submission for student" or "Bulk PDF Upload"

shailpatels:

If an instructor uploads a normal submission file and then switches to either bulk or submit for student mode, the 
the previous submission needs to be deleted every time.

### What is the new behavior?
Files submitted on normal submission are auto removed when switching modes and restored when switching back to normal submission mode.

### Other information?
Note: Since there is no distinction between files submitted for a student, bulk upload and normal submission - previous files are no only shown for normal submission. Unsure if this is ideal behavior or if displaying previous files is needed for each mode @bmcutler ?

The display also shows a warning banner when changing modes:

image

image

image

closes #7190

### DIFF
--- a/site/app/templates/submission/homework/BulkUploadBox.twig
+++ b/site/app/templates/submission/homework/BulkUploadBox.twig
@@ -26,82 +26,82 @@
         <tbody>
             {% for file in files %}
                 {% if file.cover_image is not empty %}
-                {# Display files uploaded only on bulk upload mode #}
+                    {# Display files uploaded only on bulk upload mode #}
 
-                {% if file["valid"] == false %}
-                    <tr class="tr tr-vertically-centered" style="background-color:pink;" id="row-{{ loop.index}}">
-                {% else %}
-                    <tr class="tr tr-vertically-centered" id="row-{{ loop.index }}" >
-                {% endif %}
-                    <td>{{ loop.index }}</td>
-                    <td>{{ file.clean_timestamp }}</td>
-                    <td>
-                        {{ file.filename_full }}
-                        <br>
-                        <div class="scrollable-image" >
-                            <img src="{{ file.cover_image.url }}" width="100%" alt = "{{ file.filename_full }} preview"/>
-                        </div>
-                    </td>
-                    <td>
-                        <a href="javascript:openFile('{{ file.url_full }}');" aria-label="View full PDF in a new window">
-                            <i class="fas fa-window-restore"></i>
-			            </a>
-                    </td>
-                    <td id="input_area">
-                        <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
-                        <div id="users_{{ loop.index }}">
-                            <input type="text" id="bulk_user_id_{{ loop.index }}" aria-label = "Enter User ID to assign"
-                                value ="{{ file["id"] }}"
-                                {# we should only submit on enter for normal bulk uploads or on the last team upload box #}
-                                {% if not team_assignment %}
-                                    onkeydown="uploadOnEnter( {{ loop.index }}, event );"
-                                {% else %}
-                                    onkeydown="focusNextOnEnter({{ loop.index }}, 0)"
-                                {% endif %}
-                            />
-
-
-                            {# Display any similar matches if using OCR #}
-                            {% if use_ocr %}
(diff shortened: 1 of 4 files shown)


site/ts/submission.ts
```diff
@@ -0,0 +1,51 @@
+interface Window{
```
shailpatels, line 1: @MasterOdin Is the recommended way of accessing globals in TypeScript? these functions and variables are either defined in a regular JS file or sent through twig and placed within a `<script>` tag

site/app/templates/submission/homework/BulkUploadBox.twig
```diff
@@ -25,6 +25,9 @@
         </thead>
         <tbody>
             {% for file in files %}
+                {% if file.cover_image is not empty %}
+                {# Display files uploaded only on bulk upload mode #}
```
jportorreal00, line 29: A little nitpicky, but I think adding an indentation for the inner is a bit more readable.

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -374,6 +378,26 @@
...
+                            }
+
+                            const warning_banner = document.getElementById('submission-mode-warning');
+                            if (!warning_banner.hasChildNodes()){
+                                warning_banner.appendChild( document.createElement('h2') );
+                                warning_banner.firstChild.classList.add('warning');
+                            }
+
+                            message = selected_radio_id === 'radio-bulk' ? 'Warning: Submitting files for bulk upload!' :
+                                'Warning: Submitting files for a student!';
```
jportorreal00, line 396: The message "Warning" carries a lot of context with it. In this case I don't really feel like its much of a negative. Maybe something like "Notice", "Note", or "Attention" could make it feel like the instructor isn't doing anything wrong.

jportorreal00 (commented):
@shailpatels It looks really nice. I like the placement of the banner and the logic checks out. Refreshing, submitting, and switching tabs works as expected. Just a few nots I left! Just nitpicking

jportorreal00 (approved):

site/ts/submission.ts
MasterOdin: I am not thrilled about this. Can you document that it's an open `TODO` to remove this hack as we move more stuff into TS / modules.

site/app/templates/submission/homework/SubmitBox.twig
MasterOdin: Warning here is fine. This is an action with a decent amount of concern (e.g. you're submitting for a student, did you mean to, is it the right student, etc.) versus a regular upload.

site/ts/submission.ts
shailpatels: Yeah that makes sense and a large amount of code related to submission is just in the JS global scope. What would the alternative be for variables set in a twig file, instead of inline JS maybe have HTML with data attributes instead?

site/app/templates/submission/homework/SubmitBox.twig
shailpatels: I agree, a warning is the right phrasing for submission for a student to make the instructor make sure they don't accidentally wreck someone's submission history like by uploading to the wrong person or accidentally charging a student a late day.

For bulk upload its less of a concern so I can see that saying something like "You are uploading in bulk upload mode" instead of saying warning

site/ts/submission.ts
MasterOdin: Yeah, use data attributes. It's not perfect, but it's better than inline JS as it minimally allows us to also write the code in TS.

PR #7354, opened 2021-11-12
[Bugfix:Developer] Fix Doctrine debug mode type issues

williamjallen:
A PHP error currently occurs in some cases when Doctrine queries with DateTime objects are run in debug mode.  This behavior was likely introduced in https://github.com/Submitty/Submitty/pull/7295.

### What is the new behavior?
Resolves issue by formatting DateTime objects appropriately before displaying.  In addition, this PR fixes a secondary issue where the `$params` parameter in `formatQuery()` is null and thus unable to be iterated over.

### DIFF
--- a/site/app/libraries/database/DatabaseUtils.php
+++ b/site/app/libraries/database/DatabaseUtils.php
@@ -7,9 +7,9 @@ namespace app\libraries\database;
 use app\libraries\DateUtils;
 
 class DatabaseUtils {
-    public static function formatQuery($sql, $params): string {
-        if (is_null($params)) {
-            return '';
+    public static function formatQuery(string $sql, ?array $params): string {
+        if ($params === null) {
+            return $sql;
         }
 
         foreach ($params as $param) {
--- a/site/app/libraries/database/DatabaseUtils.php
+++ b/site/app/libraries/database/DatabaseUtils.php
@@ -11,7 +11,7 @@ class DatabaseUtils {
         if (is_null($params)) {
             return '';
         }
-        
+
         foreach ($params as $param) {
             if (gettype($param) == 'object' && get_class($param) == get_class(new \DateTime())) {
                 $param = DateUtils::dateTimeToString($param);
--- a/site/app/libraries/database/DatabaseUtils.php
+++ b/site/app/libraries/database/DatabaseUtils.php
@@ -4,9 +4,18 @@ declare(strict_types=1);
 
 namespace app\libraries\database;
 
+use app\libraries\DateUtils;
+
 class DatabaseUtils {
     public static function formatQuery($sql, $params): string {
+        if (is_null($params)) {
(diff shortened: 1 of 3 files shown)


MasterOdin (changes requested):
Add testcases to https://github.com/Submitty/Submitty/blob/master/site/tests/app/libraries/database/DatabaseUtilsTester.php for the changed behavior.

site/app/libraries/database/DatabaseUtils.php
```diff
@@ -4,9 +4,18 @@
 
 namespace app\libraries\database;
 
+use app\libraries\DateUtils;
+
 class DatabaseUtils {
     public static function formatQuery($sql, $params): string {
+        if (is_null($params)) {
```
MasterOdin, line 11: How do we hit this case? When/how is `$params` null?

williamjallen: I am unsure.  I fixed the issue with DateTime objects and another error about null `$params` popped up and I couldn't find a source of the issue so I just wrapped it in the `is_null` and moved on.

MasterOdin: This was a thrown exception?

williamjallen: This was the error: Untitled
(line 15 is the start of the for-loop in this example)

site/app/libraries/database/DatabaseUtils.php
```diff
@@ -4,9 +4,18 @@
...
+            return '';
+        }
```
MasterOdin, line 13: So given the above, let's make this change:

```suggestion
    public static function formatQuery(string $sql, ?array $params): string {
        if ($params === null) {
            return $sql;
        }
```

I forgot to add types in my original PR, so let's add them here, where this should be the expected shape given the bit about null. Also using `=== null` is a micro-optimization over `is_null`, but worth preferring as it avoids the overhead of a function call to do the exact same thing as equality. Finally, given null parameters, we should still have a string for the query to run, so should just return `$sql` as is.

MasterOdin (changes requested):

MasterOdin:
@williamjallen you still need to add testcases to https://github.com/Submitty/Submitty/blob/master/site/tests/app/libraries/database/DatabaseUtilsTester.php to cover the changes in behavior, good to merge after that.

williamjallen:
I have some homework due tonight but I plan to have the tests finished by tomorrow

MasterOdin:
@williamjallen Sorry for losing track on this, I made a commit https://github.com/Submitty/Submitty/pull/7354/commits/dd8c14e7b13fbfc36851e79f0d3b5320c744ca31 to add a missing test case (`null`), just something to consider for the next time tests are written to ensure full test coverage.

PR #7371, opened 2021-11-18
[Bugfix:SubminiPolls] Add default release date to new poll

BenKelly99:
Currently, the release date field is blank for new polls

### What is the new behavior?
Now, the default release date will be the current date

### DIFF
--- a/site/app/templates/polls/NewPollPage.twig
+++ b/site/app/templates/polls/NewPollPage.twig
@@ -400,10 +400,10 @@
     $(document).ready(function() {
         {% if not poll is defined %}
             const today = new Date();
-            const year = today.getFullYear().toString().padStart(4, '0')
-            const month = (today.getMonth() + 1).toString().padStart(2, '0')
-            const day =  today.getDate().toString().padStart(2, '0')
-            const date_string = year + "-" + month + "-" + day
+            const year = today.getFullYear().toString().padStart(4, '0');
+            const month = (today.getMonth() + 1).toString().padStart(2, '0');
+            const day =  today.getDate().toString().padStart(2, '0');
+            const date_string = year + "-" + month + "-" + day;
             $("#poll-date").flatpickr({
                 defaultDate:date_string
             })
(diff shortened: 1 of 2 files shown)


HenrikLam (changes requested):
Looks good but looks like tests needs to be changed.

shailpatels:
Test cases revolving around polls functionality -- Should verify making, editing, deleting poll works as expected (failed)

It looks like cypress test trys to create a poll but then gets an invalid date error which fails to create it

sbelsk (commented):
I think the js-lint error in `site/cypress/integration/polls.spec.js` can be fixed using something like
```javascript
const date_string = `${year}-${month}-${day}`;
```
instead of the older way of performing the string concatenation

BenKelly99:
Thanks Shelly! I had updated another location but forgot the cypress tests.

BenKelly99:
@HenrikLam Should be ready for re-review

BenKelly99:
@shailpatels is there a way to access the screenshots/video the test takes during its run? not sure how to see those

shailpatels:
> @shailpatels is there a way to access the screenshots/video the test takes during its run? not sure how to see those

@BenKelly99 You can see them in the artifacts section below the github action run, at the bottom of the page

here is an example
https://github.com/Submitty/Submitty/actions/runs/1517986775

You can get to this by clicking on the 'details' link under any job and then pressing 'summary' on the left

BenKelly99:
@sbelsk @HenrikLam tests now pass, should be an easy review now.

HenrikLam (approved):
Looks good to me!

site/app/templates/polls/NewPollPage.twig
```diff
@@ -398,6 +398,16 @@
...
     $(document).ready(function() {
+        {% if poll is not defined %}
+            const today = new Date();
+            const year = today.getFullYear().toString().padStart(4, '0');
+            const month = (today.getMonth() + 1).toString().padStart(2, '0');
+            const day = today.getDate().toString().padStart(2, '0');
+            const date_string = `${year}-${month}-${day}`;
+            $("#poll-date").flatpickr({
+                defaultDate: date_string
+            });
```
MasterOdin, line 409: Ideally we'd be able to combine this flatpickr call with https://github.com/Submitty/Submitty/blob/96750b883980eeb41bb69fc179856012fe755e62/site/app/templates/polls/NewPollPage.twig#L188-L194 to avoid destroying and then recreating the flatpickr instance on the element, but this is probably simpler for now.

PR #1698, opened 2018-02-06
Install Dafny, Pylint, Python unittest etc

bmcutler:
New tools used by Principles of Software & Computer Science 1

PR #7396, opened 2021-11-29
[Bugfix:Developer] Fix Python Linting

williamjallen:
A recent issue caused Python linting to begin failing on master.  After some digging, I was unable to determine the original source of the issue and have simply addressed the errors in this PR.  

Example of a PR with failing CI: https://github.com/Submitty/Submitty/runs/4351911193?check_suite_focus=true

The most recent commit to Submitty (all checks succeeded): https://github.com/Submitty/Submitty/pull/7385/checks

### What is the new behavior?
Fixes linting.

### DIFF
--- a/autograder/autograder/insert_database_version_data.py
+++ b/autograder/autograder/insert_database_version_data.py
@@ -127,7 +127,7 @@ def insert_into_database(config, semester, course, gradeable_id, user_id, team_i
     us to do an update here (as the PHP also deals with the active version for us), but in case
     we're using some other method of grading, we'll insert the row and whoever called the script
     will need to handle the active version afterwards.
-    """
+    """   # noqa: B018
     if is_team is True:
         result = db.execute(select([func.count()]).select_from(data_table)
                             .where(data_table.c.g_id == bindparam('g_id'))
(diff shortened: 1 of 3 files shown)


MasterOdin:
The error is arising because we install flake8 and flake8-bugbear without any version constraints within the CI:

https://github.com/Submitty/Submitty/blob/bfcbbfdfcad0081caff84ca1442ed8453b3445b9/.github/workflows/submitty_ci.yml#L174

flake8-bugbear released a new version ([`21.11.281`](https://github.com/PyCQA/flake8-bugbear/releases/tag/21.11.28)) which includes:

> Add B018 check to find useless declarations

This is a reason why putting version constraints on things is good, and why we should probably do that here, putting them into a pip requirements file similar to system or vagrant.

MasterOdin (approved):

williamjallen:
We have `pylint` in `.setup/pip/system_requirements.txt`.  Should these two dependencies also be added to that file or should there be a separate file for developer resources like these that aren't strictly necessary for a standard Submitty installation?

MasterOdin:
`pylint` is in the system requirements as it was required by "Principles of Software" course back in s2018 semester (https://github.com/Submitty/Submitty/pull/1698).

It's not strictly required to actually do anything within Submitty itself, and as such might be good to remove it if we can get that course into using a Docker image...

PR #7406, opened 2021-12-01
[Bugfix:Developer] Enable Cypress test retries

williamjallen:
Cypress tests appear to fail at random on a limited number pull requests.  Most of the time the issue may be attributed to flakiness in the tests or testing environment and simply rerunning the tests often fixes the issue.  Cypress provides a configuration option to rerun tests a specified number of times on failure, succeeding if any of the attempts succeeds.

### What is the new behavior?
This PR enables Cypress test retries with an initial number of 2 retries per failed test.  If, after retrying twice, the test has not succeeded, Cypress concludes that the test has failed, otherwise the test passes.  This setting is only enabled when using `cypress run`, not using the graphical UI opened by the command `cypress open`.  `cypress run` is the command used in the CI to run all of the tests in sequence while `cypress open` is the standard way to write and verify tests locally.

### DIFF
--- a/site/cypress.json
+++ b/site/cypress.json
@@ -4,5 +4,9 @@
     "*/*.spec.js",
 		"*.spec.js"
 	],
-	"pageLoadTimeout" : 120000
+	"pageLoadTimeout" : 120000,
+  "retries": {
+    "runMode": 2,
+    "openMode": 0
+  }
 }


jportorreal00:
@williamjallen I probably wasn't there for this conversation, but who decided that 2 retries was best here? 2 is usually my go to, but it can be a bit too much if the tests take a long time to run. I think it should be fine, but I was just curious!

williamjallen:
> @williamjallen I probably wasn't there for this conversation, but who decided that 2 retries was best here? 2 is usually my go to, but it can be a bit too much if the tests take a long time to run. I think it should be fine, but I was just curious!

It was completely arbitrary.  2 is what is used in the Cypress documentation so I kept it at that.  I don't think it will be too much of an issue.  The key is for our tests to be as reliable as possible and the longest of the test cases is less than a minute long usually so retrying them should have almost no effect on the overall running time of our 50+ minute CI given the potential reduction in flakiness.  The tests are not re-run when using `cypress open` as is typically used locally so this should only really affect the CI running time.

MasterOdin (approved):

PR #7407, opened 2021-12-01
[Feature:TAGrading] Popup panels

DefinitiveAbove:
You cannot pop out panels for TA grading in new windows.

### What is the new behavior?
You can now pop out panels for TA grading in new windows.
When panels are popped out they cannot be used in the main window.
Panels are associated with the window that initially opened it, and have a shared ID with the window that initially opened it.
Panels will sync up with the last opened grading page from the specified browser page.

image


loadedskiffer (commented):
I've tested as an instructor, ta and mentor on a few different gradeables and when I try to pop out a panel I just get the following screen.
popout_1

williamjallen:
Would it be possible to add some screenshots of the existing+new UI to the PR description?  It makes it easier to reference in the future if things break or if reviewers want to know what was added.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1283,6 +1283,144 @@ private function amIBlindGrading($gradeable, $user, $peer) {
...
+     * @param $who_id This is the user we wish to view, this field will only be passed on when the request originates
+     *                  on the grading index page
+     * @param $from This is the user that was being viewed when a navigation button was clicked on the TA grading
+     *                  interface.  Knowing who we were viewing allows us to decide who to view next.
+     * @param $to Used to determine the direction to move in, either 'prev' or 'next'
+     * @param $to_ungraded Should the next student we go to be the next submission or next ungraded submission?
+     *
+     * @Route("/courses/{_semester}/{_course}/gradeable/{gradeable_id}/grading/grade/popup")
+     */
+    public function showGradingPopup($gradeable_id, $who_id = '', $gradeable_version = null, $anon_mode = false, $panel = "", $window_id = -1) {
```
williamjallen, line 1298: Types should be added to each of the parameters and to the return type.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1283,6 +1283,144 @@ private function amIBlindGrading($gradeable, $user, $peer) {
...
+            $anon_mode = true;
+        }
+        /** @var Gradeable $gradeable */
+
+        $gradeable = $this->tryGetGradeable($gradeable_id, false);
+        if ($gradeable === false) {
+            $this->core->addErrorMessage('Invalid Gradeable!');
+            $this->core->redirect($this->core->buildCourseUrl());
+        }
+        $peer = $gradeable->hasPeerComponent() && $this->core->getUser()->getGroup() == User::GROUP_STUDENT;
```
williamjallen, line 1309: It would be better to use `===` here instead of `==`.  (same thing applies to line 1311)

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1283,6 +1283,144 @@ private function amIBlindGrading($gradeable, $user, $peer) {
...
+                $prevVersionInstance = $graded_gradeable->getAutoGradedGradeable()->getAutoGradedVersionInstance($previousVersion);
+                if ($prevVersionInstance == null) {
+                    $rollbackSubmission = -1;
+                    break;
+                }
+                $lateInfo = LateDays::fromUser($this->core, $late_days_user)->getLateDayInfoByGradeable($gradeable);
+                $daysLate = $prevVersionInstance->getDaysLate();
+
+                // If this version is a good submission then it the rollback Submision
+                if ($lateInfo == null || ($lateInfo->getStatus($daysLate) == LateDayInfo::STATUS_GOOD)) {
```
williamjallen, line 1389: Should use `===` here as well

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1283,6 +1283,144 @@ private function amIBlindGrading($gradeable, $user, $peer) {
...
+        $this->core->getOutput()->addInternalCss('electronic.css');
+
+        $this->core->getOutput()->addInternalJs('forum.js');
+        $this->core->getOutput()->addInternalCss('grade-inquiry.css');
+        $this->core->getOutput()->addInternalJs('grade-inquiry.js');
+        $this->core->getOutput()->addInternalJs('websocket.js');
+        $show_hidden = $this->core->getAccess()->canI("autograding.show_hidden_cases", ["gradeable" => $gradeable]);
+        $this->core->getOutput()->renderOutput(['grading', 'ElectronicGrader'], 'hwGradingPopup', $gradeable, $graded_gradeable, $panel, $display_version, $show_hidden, $can_inquiry, $can_verify, $show_verify_all, $show_silent_edit, $rollbackSubmission, $who_id, $solution_ta_notes, $submitter_itempool_map, $anon_mode, $blind_grading, $window_id);
+        $this->core->getOutput()->renderOutput(['grading', 'ElectronicGrader'], 'popupStudents');
+        $this->core->getOutput()->renderOutput(['grading', 'ElectronicGrader'], 'popupMarkConflicts');
```
williamjallen, line 1421: I'm not too familiar with this part of the codebase so there may be no better way to do it without significant changes but ideally this type of code would be moved to a view instead of a controller.

site/app/libraries/Output.php
```diff
@@ -398,6 +406,32 @@ private function getView($class) {
         return $this->loaded_views[$class];
     }
 
+    public function overrideHeader($view, string $function, ...$args) {
```
williamjallen, line 409: What is the general purpose/idea behind the changes to this file?

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -891,6 +891,159 @@ public function randomizeButtonWarning(Gradeable $gradeable) {
         ]);
     }
 
+    public function hwGradingPopup(Gradeable $gradeable, GradedGradeable $graded_gradeable, string $type, int $display_version, bool $show_hidden_cases, bool $can_inquiry, bool $can_verify, bool $show_verify_all, bool $show_silent_edit, $rollbackSubmission, $from, array $solution_ta_notes, array $submitter_itempool_map, $anon_mode, $blind_grading, $window_id) {
```
williamjallen, line 894: The list of parameters is quite long here.  It would be better to break it such that each parameter is on a separate line.  In addition, a return type should be added.

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1170,7 +1327,7 @@ public function renderNavigationBar(GradedGradeable $graded_gradeable, float $pr
         ]);
     }
 
-    public function renderGradingPanelHeader(bool $isPeerPanel, bool $isStudentInfoPanel, bool $isDiscussionPanel, bool $isRegradePanel, bool $is_notebook, string $error_color, string $error_message): string {
+    public function renderGradingPanelHeader(bool $isPeerPanel, bool $isStudentInfoPanel, bool $isDiscussionPanel, bool $isRegradePanel, bool $is_notebook, string $error_color, string $error_message, $version, $anon_mode, $submitter_id, $gradeable_id): string {
```
williamjallen, line 1330: new parameters should have types specified like the other parameters

site/public/js/ta-grading.js
```diff
@@ -108,86 +100,128 @@ $(function () {
...
+        }
+      }
+    })
 
-  // Progress bar value
-  let value = $(".progressbar").val() ? $(".progressbar").val() : 0;
-  $(".progress-value").html("<b>" + value + '%</b>');
+    // Progress bar value
+    let value = $(".progressbar").val() ? $(".progressbar").val() : 0;
+    $(".progress-value").html("<b>" + value + '%</b>');
```
williamjallen, line 130: Would be cleaner to write this as `<b>${value} %</b>` (surrounded by backticks--I can't get them to show up in MD).  I'm surprised that eslint didn't catch this.  Usually it does.

site/public/js/ta-grading.js
```diff
@@ -108,86 +100,128 @@ $(function () {
...
+      if (panelSpanId) {
+        const panelId = panelSpanId.split(/(_|-)btn/)[0];
+        if (position === "popup") {
+          openPopup(panelId);
+          setPanelsVisibilities(panelId, false, position);
+        } else {
+          setPanelsVisibilities(panelId, null, position);
+          checkNotebookScroll();
+        }
+        $('select#' + panelId + '_select').hide();
```
williamjallen, line 176: Could rewrite this as `select#${panelId}_select` as noted before

site/public/js/ta-grading.js
```diff
@@ -108,86 +100,128 @@ $(function () {
...
-  // panel position selector change event
-  $(".grade-panel .panel-position-cont").change(function() {
-    let panelSpanId = $(this).parent().attr('id');
-    let position = $(this).val();
-    if (panelSpanId) {
-      const panelId = panelSpanId.split(/(_|-)btn/)[0];
-      setPanelsVisibilities(panelId, null, position);
-      $('select#' + panelId + '_select').hide();
-      checkNotebookScroll();
+    if(localStorage.getItem('notebook-setting-file-submission-expand') == 'true') {
```
williamjallen, line 184: Should use `===` here

site/public/js/ta-grading.js
```diff
@@ -108,86 +100,128 @@ $(function () {
...
+      let notebookPanel = $('#notebook-view');
+      if(notebookPanel.length != 0) {
```
williamjallen, line 186: Should use `!==` here

site/public/js/ta-grading.js
```diff
@@ -108,86 +100,128 @@ $(function () {
...
+        let notebookItems = notebookPanel.find('.openAllFilesubmissions');
+        for(var i = 0; i < notebookItems.length; i++) {
```
williamjallen, line 188: Would be better to use `let` instead of `var`

site/public/js/ta-grading.js
```diff
@@ -805,10 +847,10 @@ function setPanelsVisibilities (ele, forceVisible=null, position=null) {
   panelElements.forEach((panel) => {
     let id_str = document.getElementById(panel.str + "_btn") ? "#" + panel.str + "_btn" : "#" + panel.str + "-btn";
     if (panel.str === ele) {
-      const eleVisibility = forceVisible !== null ? forceVisible : !$("#" + panel.str).is(":visible");
+      let eleVisibility = forceVisible !== null ? forceVisible : !$("#" + panel.str).is(":visible");
```
williamjallen, line 850: Any particular reason for this change?  I may have missed it but I don't see this variable being modified anywhere.

site/public/js/ta-grading.js
```diff
@@ -1654,3 +1696,74 @@ function collapseFile(panel = "submission"){
     $(fileFullPanelOptions[panel]["fileView"]).hide();
   });
 }
+
+// https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability
+function storageAvailable(type) {
+  var storage;
+  try {
+      storage = window[type];
+      var x = '__storage_test__';
```
williamjallen, line 1705: should use `const` instead of `var`

williamjallen (changes requested):
I just briefly looked through the code and have made comments on a few locations.  Overall it looks pretty good but I'm a little bit worried about how the panels are being displayed and the changes to `Output.php`.  Is there any technical reason why it's better to do it this way instead of simply rendering regular Submitty pages with no header or footer and a full-page view of the pane?

PR #7416, opened 2021-12-06
[Feature:InstructorUI] Autograding Stacktrace

### LINKED ISSUE #4398: Improve Autograding Error messages when requested Docker image was not available
**Describe the bug**
Right now we have to manually pull the docker image to the worker machine that is doing the autograding.  But if we forget to do this step, the error message is currently confusing.  (it's only in the stack traces log, if you think to look there.) 

**Expected behavior**
So let's get this error message to a location visible to the instructor/ta who's preparing the autograding, debugging instantiation of this gradeable, etc.

HenrikLam:

Closes #4398

There is currently no way of accessing the autograding stacktrace except by doing ssh into the submitty system. This PR adds a section to display the 7 most recent autograding stacktrace files onto the autograding status page for instructors.

### What is the new behavior?
image
Each stacktrace is separated from its originating file. There is a refresh button to manually refresh the stacktrace. The auto refreshing of this page does not apply to the stacktrace since changes should happen less frequently for this section.


migration/migrator/migrations/system/20211121152144_autograding_stacktrace_chgrp.py
```diff
@@ -0,0 +1,26 @@
...
+
+def down(config):
+    """
+    Run down migration (rollback).
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    os.system("chgrp -R {} {}".format(config.submitty_users['course_builders_group'], str(Path(config.submitty['submitty_data_dir'], 'logs', 'autograding_stack_traces'))))
+    os.system("chgrp -R {} {}".format(config.submitty_users['course_builders_group'], str(Path(config.submitty['submitty_data_dir'], 'logs', 'autograding'))))
```
MasterOdin, line 26: You don't need to use a migration to change the group of these directories. Just change the `chown` lines for these directories in `.setup/INSTALL_SUBMITTY_HELPER.sh`.

However, this does change the group from `course_builders` to `daemonphp_group`  meaning instructors would no longer be able to access these directories?

HenrikLam: Yes, the idea was that the instructors will use the autograding status page to view such information since ssh-ing into the submitty machine isn't a popular practice among instructors anyway. Changing this to `daemonphp_group` will allow the php script to read the logs which was not allowed when the group was `course_builders`.

site/app/controllers/AutogradingStatusController.php
```diff
@@ -44,6 +45,36 @@ public function getProgress(): JsonResponse {
...
+     */
+    public function getErrorStackTrace(): JsonResponse {
+        if (empty($this->core->getQueries()->getInstructorLevelAccessCourse($this->core->getUser()->getId()))) {
+            return JsonResponse::getFailResponse("You do not have access to this endpoint.");
+        }
+        $stack_trace_path = FileUtils::joinPaths($this->core->getConfig()->getSubmittyPath(), "logs", "autograding_stack_traces");
+        if (!is_readable($stack_trace_path)) {
+            return JsonResponse::getFailResponse("Could not access the stack trace path.");
+        }
+        $files = scandir($stack_trace_path, SCANDIR_SORT_DESCENDING);
```
MasterOdin, line 61: If we only show the last 7 stack traces, why do we want to return all of them here? What happens if an instructor has hundreds of stack traces that in total weigh several GB? We should probably do the limiting of number of returned files _here_ instead of within the JS code and avoid sending unnecessary bytes over the wire, slowing down load of the page.

MasterOdin (changes requested):

site/app/controllers/AutogradingStatusController.php
```diff
@@ -44,6 +45,36 @@ public function getProgress(): JsonResponse {
...
 
+    /**
+     * Attempts to read the stack trace and find any error message
+     * @Route("/autograding_status/get_stack", methods={"GET"})
+     * @return JsonResponse
+     */
+    public function getErrorStackTrace(): JsonResponse {
+        if (empty($this->core->getQueries()->getInstructorLevelAccessCourse($this->core->getUser()->getId()))) {
+            return JsonResponse::getFailResponse("You do not have access to this endpoint.");
+        }
```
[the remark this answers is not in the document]
HenrikLam: I believe there is a difference between the two in that the query checks for if the user is assigned an instructor in any course while the access controls checks the system wide access level. I believe this whole page was designed to be allowed access to TAs (usually grad TAs) who are granted instructor level access to only the course.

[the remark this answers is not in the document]
williamjallen: This issue has come up a few times in the last year or so.  Would it be reasonable to change the behavior such that it returns the *current* access level?

williamjallen (changes requested):
The code looks good for the most part but I have a few UI concerns.  In light mode, the contrast is rather poor.  In addition, the scrollbars also have poor contrast/coloring in light mode.

HenrikLam:
@williamjallen I came up with two options for the color, one is to make it like the code box in the rest of the site and the other one is to just make it display the page and not distinguish the content from the stacktrace. Also I made some changes to the scrollbar. I have attached the screenshots, let me  know what you think or if you have any suggestions!

**Code block style**
Dark Mode:
image

Light Mode:
image

**No Style**
Dark Mode:
image

Light Mode:
image

Viyerelu23333 (approved):
Looks good to me and works on my end.

PR #7434, opened 2021-12-10
[Feature:SubminiPolls] Display correct answer in poll results

### LINKED ISSUE #7352: Show correct answer to poll after it is closed
Currently there is no way for a student to view the correct answer of a poll after it is closed.
There is a setting that allows students to see the histogram but that doesn't specifically say the correct answer.

We should show the histogram but instead of blue, color the correct answer green so it is clear which one is correct.

tlasich:
- [ ] Tests for the changes have been added/updated (if possible)
- [ ] Documentation has been updated/added if relevant

Closes #7352 

All bars in the Submini Poll Results bar chart are blue regardless of correctness. Students cannot easily decipher which answer is correct.

### What is the new behavior?
Poll responses that are denoted as correct by the instructor now have green bars to indicate that they are the correct answer. This behavior has been verified for polls with a single or multiple correct options.

### Other information?
This does cause surveys to be completely green as there are no "incorrect" answers. This can be easily fixed if the desired bar color for surveys is blue.

I tested the new functionality by creating mock data in my local database for single answer, multiple answer, and survey polls.

submitty_single_correct_poll
submitty_multiple_correct_poll
submitty_survey_poll


site/app/templates/polls/ViewPollResults.twig
```diff
@@ -5,11 +5,15 @@
...
             y: [],
+            marker: {
+              color: [],
+            },
           }
         ];
-        {% for answer, number in results %}
-            data[0].x.push("{{ poll.getResponseString(answer)|replace({"\n": " ", "\r": " ", "\t": " "}) }}");
-            data[0].y.push({{ number }});
+        {% for response, attributes in results %}
```
MasterOdin, line 13: If we're renaming this, let's call it what it is, `option_id`?

MasterOdin (approved):
One small nit, otherwise looks good and will merge after the change is commited.

bmcutler:
Thanks for this contribution :)

However, we need to do a bit more work before we can merge this feature.  The instructor needs to have control over when and if the correct answer(s) are displayed.  I've added a comment on the original issue.

site/app/templates/polls/ViewPollResults.twig
```diff
@@ -5,11 +5,15 @@
...
+            },
           }
         ];
-        {% for answer, number in results %}
-            data[0].x.push("{{ poll.getResponseString(answer)|replace({"\n": " ", "\r": " ", "\t": " "}) }}");
-            data[0].y.push({{ number }});
+        {% for option_id, attributes in results %}
+            data[0].x.push("{{ poll.getResponseString(option_id)|replace({"\n": " ", "\r": " ", "\t": " "}) }}");
+            data[0].y.push({{ attributes['count'] }});
+            data[0].marker.color.push("{{ attributes['is_correct'] ? '#2ca02c' : '#1f77b4' }}");
```
williamjallen, line 16: Are raw hex codes the best option here?

Viyerelu23333:
Superseded by #7595

PR #6482, opened 2021-05-22
[Feature:System] Add support for ldap authentication

MasterOdin:
### Please check if the PR fulfills these requirements:

* [x] The PR title and message follows our [guidelines](https://submitty.org/developer/how_to_contribute)
* [x] Tests for the changes have been added/updated (if possible)
* [x] Documentation has been updated/added if relevant: https://github.com/Submitty/submitty.github.io/pull/385

### What is the current behavior?

Closes #2054 

Submitty does not currently support using LDAP to authenticate users.

### What is the new behavior?

Adds LDAP support for Submitty.

PR #7444, opened 2021-12-14
[Feature:System] Add support for ldap authentication

MasterOdin:

The PR #6482 was reverted in #7442 as it introduced a bug where the migration would place the `authentication.json` file with the wrong permissions, rendering it unreadable, and prevented the system from being used.

### What is the new behavior?

This PR redoes #6482, while fixing the bug with permissions, setting it as part of the migration.

### Other information?

I tested by doing a `vagrant up` from master, and then re-running #6482 and was able to reproduce bug where authentication.json file was not readable. I then made a patch to the migration file fixing that, and on re-running the migration, the file was properly created and was readable. I then re-tested using `vagrant up` on master, and confirmed that applying this PR would not break the site.


migration/migrator/migrations/system/20211214211432_authentication_method.py
```diff
@@ -0,0 +1,61 @@
...
+
+    with (config.config_path / 'database.json').open('r+') as db_file:
+        db_info = json.load(db_file, object_pairs_hook=OrderedDict)
+        db_file.seek(0)
+        db_file.truncate()
+        del db_info['authentication_method']
+        json.dump(db_info, db_file, indent=4)
+    chown(authentication_file, 'root', config.submitty_users['daemonphp_group'])
+    chmod(authentication_file, 0o440)
+
```
MasterOdin, line 37: This is the only change in this PR from the original one. This matches the chmod / chown that is done in `CONFIGURE.py`.

williamjallen (approved):

shailpatels (approved):
Looks good, did a vagrant up before and after this PR and no longer see the authentication error shown in the previous merged branch.

@MasterOdin Do you think we should run the login e2e tests and or check for site errors script on our nightly vagrant up for a stronger check the developer instance is good to avoid issues like this in the future?

MasterOdin:
I don't think either would have caught this issue, it only cropped up on running the migration on a machine that had not previously run the migration. Ideally we'd implement improved testing for migrations, but I'm not sure how that'd look.

Issue #7230, opened 2021-10-11
Unify behavior on showing submission answer header when no notes


MasterOdin:
> Is there any UI differences when displaying a solution note that is empty but had been previously filled, and a solution note that had never been filled?

Yes. When a solution note has never been filled it there will be a header above the solution box that tells you there is no solution. When a solution note has been filled before but is empty now it does not do that.

_Originally posted by @joekrystowski in https://github.com/Submitty/Submitty/issues/7176#issuecomment-939148374_

-----

We should unify the design here to show the same header in the case where no notes have been added or that the most recent revision of the notes is blank. There's not really any functional difference between the two and why we would want to special case them.

PR #7487, opened 2021-12-26
[Refactor:TAGrading] Refactor Solution Notes pane UI

williamjallen:
There are currently a number of small functionality issues with the solution notes pane of the TA grading interface, as well as a handful of locations where code was modified multiple times and is no longer used.

Before:
before

### What is the new behavior?

This PR fixes a number of UI bugs on the solution notes pane of the TA grading interface and removes a quantity of dead code.  #7176 fixed many of the issues but there were a few more left to be fixed.  Notably, the text of the solution notes no longer shows outside the markdown text box as it used to prior to #7176.

After:
after

### Other information?
Closes https://github.com/Submitty/Submitty/issues/7230

Tests for this interface will be added once https://github.com/Submitty/Submitty/pull/7486 is merged.


HenrikLam (commented):
One minor adjustment I would like is to disable the save change button again after it is pressed. Currently the button stays active after we press save change. Additionally, we may also check if the content is different from what is already saved before 
deciding if the button should be active, but this might not be necessary. 
SaveStaysActive

Other than this, this PR looks good.

HenrikLam (approved):
Looks good to me

DefinitiveAbove (changes requested):
Functionality looks good to me, minor change w/removing disabled class on buttons

Issue #7039, opened 2021-09-07
Removal of margin for card in peer matrix tab


HenrikLam:
**Current behavior**
Currently, the cards in the peer matrix tab seems to be misaligned with the table. This is due to the inherent padding that comes with the `card` class. Let's add a new class to the card that overrides the padding or get rid of the padding if we can not override it.

**Configuration**
 - OS: [e.g. Windows, Linux, ...]

**Screenshots**
![image](https://user-images.githubusercontent.com/49821332/132424025-b1d79151-c804-48b5-b52d-eed69861e15f.png)

williamjallen:
I suspect that something with Bootstrap has been changed recently.  I am seeing new UI changes on the plagiarism interface which look like Bootstrap updates to me.  I looked for the source of the issue but couldn't find anything which may have caused this.  I believe the issue affects anything which uses the Bootstrap grid (which is the only part of Bootstrap currently updated by dependency PRs), most notably when it affects padding/margins and table borders.

PR #7490, opened 2021-12-27
[Refactor:InstructorUI] Refactor peer matrix UI

williamjallen:
https://github.com/Submitty/Submitty/issues/7039 describes an issue with the margins on the primary columns of the peer matrix interface in the instructor UI.  While working on the issue described, a handful of other minor UI issues were observed.

Before:

### What is the new behavior?
This PR refactors the peer matrix UI to include a radio button option for choosing between "all grade all" and manually specifying number of others to grade, as well as removing excess styling code and removing excess margins.  Once this PR is merged, I will write a number of Cypress tests to cover this part of the system.

After:
after

### Other information?
Closes https://github.com/Submitty/Submitty/issues/7039


williamjallen:
@HenrikLam I have updated the layout of the checkboxes and radio buttons.  I'm still not entirely happy with the layout but I think it's an improvement at least.  I also tinkered with making the table the full width of the content pane but in my opinion, it looks rather ugly and cramped since it isn't aligned with the rest of the content on the page.  It does waste some space how it currently is, but since the primary element of the table is a list of user IDs which wrap, it isn't too big of an issue (in my opinion).  I have attached a screenshot of the full-width table.

Thoughts?

Untitled

HenrikLam (changes requested):
I am happy with how the UI looks after the changes, I do have one minor nitpick on the number input field being able to go below 0 though. This is probably not a big problem since instructors won't purposefully put in a negative number. Also is it intended that you can put 0 in that field? It seems to only clear the matrix which doesn't seem very useful...

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -201,21 +212,20 @@
...
-        let form = $("#admin-gradeable-edit-peers-form");
         let grader_id = $("#admin-gradeable-edit-peers-title").text().substr(18);
         let fd = new FormData();
         fd.append('csrf_token', $("#csrf-token").val());
         fd.append('grader_id', grader_id);
         fd.append('remove_grader', $('#removed-all').text());
         let curr_student_ids = [];
         $(".edit_user_ids :input.peerstudents").each(function() {
-            if ($(this).val() != "") curr_student_ids.push($(this).val());
+            if ($(this).val() !== "") curr_student_ids.push($(this).val());
```
HenrikLam, line 222: Is this convention allowed in JS files? Although this is not checked in twig files, perhaps we should follow the same convention throughout our files.

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -201,21 +212,20 @@
...
         });
         fd.append('curr_student_ids', JSON.stringify(curr_student_ids));
         // set to check for duplicate user entries
         let add_student_ids = new Set();
         $(".add_user_ids").each(function() {
-            if ($(this).val() != "") add_student_ids.add($(this).val());
+            if ($(this).val() !== "") add_student_ids.add($(this).val());
```
HenrikLam, line 228: Same as above.

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -328,7 +342,7 @@
         // set to check for duplicate user entries
         let add_student_ids = new Set();
         $(".add_user_ids", form).each(function() {
-            if ($(this).val() != "") add_student_ids.add($(this).val());
+            if ($(this).val() !== "") add_student_ids.add($(this).val());
```
HenrikLam, line 345: Sorry, one more here...

HenrikLam (approved):
Looks good to me.

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -1,81 +1,92 @@
...
-                         <input type="button" name="download_peer_csv" class="btn btn-default key_to_click ignore" id="download_peer_csv" tabindex="0" value="Download Peer CSV" onclick="downloadPeerDataAsCsv('peer-table')"></input>
-                    </div>
-                    <br>
+    </div>
+    <br>
+    <div class="row">
+        <div class="col-md-12 pl-0">
+            <div>
+                Other Controls:
+                <div class="">
```
MasterOdin, line 34: Let's get rid of the `class=""`

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -84,7 +95,7 @@
     }
 
     function removeStudent(i, $grader_to_edit) {
-        var form = $("#admin-gradeable-edit-peers-form");
+        let form = $("#admin-gradeable-edit-peers-form");
```
MasterOdin, line 98: Use `const` if not redefining the variable. This is true of most of the usages of `let` in this diff.

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -1,81 +1,92 @@
 <div id="peer-control">
-    <h1> Options for Peer Matrix </h1>
-        <div class="row">
-            <div class="card col-md-6" style="margin:0px">
-                <div class="content" style="margin:0px">Upload a csv file dictating grader-student pairs for peer grading. The csv file must contain a labelled 'student' column and a 'grader' column: <BR><hr>
-                    <input class="custom-file-input" type="file" name="peer_graders_list" id="peer_graders_list">
-                </div>
+    <h2> Options for Peer Matrix </h2>
```
MasterOdin, line 2: Having the additional spaces surrounding the text in the header strikes me as...weird.
```suggestion
    <h2>Options for Peer Matrix</h2>
```

MasterOdin (changes requested):

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -1,81 +1,92 @@
...
-            <div class="card col-md-12" style="margin:0px">
-                <div class="content"style="margin:0px">
-                    <p>Other Controls: </p>
-                    <div class="btn-wrapper" >
-                        <input type="button" name="clear_peer_matrix" class="btn btn-default key_to_click ignore" id="clear_peer_matrix" tabindex="0" value="Clear Peer Matrix"></input>
-                         <input type="button" name="download_peer_csv" class="btn btn-default key_to_click ignore" id="download_peer_csv" tabindex="0" value="Download Peer CSV" onclick="downloadPeerDataAsCsv('peer-table')"></input>
-                    </div>
-                    <br>
+    </div>
+    <br>
```
MasterOdin, line 29: Not asking for a change on this as it's not a huge thing, but if we wanted to strictly follow best accessibility / web practices, we should only use the `<br>` tag for line breaks in text where the break is significant (e.g. if writing a poem or an address). We should __not__ be using them to put margins between elements, rather relying on CSS `margin` to do this. Similar, for `<input>` tags we want on individual lines should be wrapped in a `<div>`. While it'd be nice to fix this, if it's a non-trivial amount of effort, fine to not deal with this.

PR #7543, opened 2022-01-17
[Bugfix:TAGrading] Always show grade inquiry in index

### LINKED ISSUE #7461: grade override + grade inquiry
**Describe the bug**
When a student has an unresolved grade inquiry AND the instructor overrides their grade,
we cannot visually see in the UI which student has the unresolved grade inquiry
It displays the "override" status on the index of all students -- let's make a visual for both...

DefinitiveAbove:
On the TA grading index, it will not show that a user has a grade inquiry if their grade is overridden or if they have no submission for the assignment.

### What is the new behavior?
Closes #7461 
It will always show if a user has a grade inquiry on the button in the grading index.

### DIFF
--- a/site/app/templates/grading/electronic/Details.twig
+++ b/site/app/templates/grading/electronic/Details.twig
@@ -429,11 +429,6 @@
     {% elseif graded_gradeable.getAutoGradedGradeable().getActiveVersion() == 0 %}
         {% set contents = "Cancelled Submission" %}
         {% set btn_class = "btn-default" %}
-        {% if graded_gradeable.hasActiveRegradeRequest() %}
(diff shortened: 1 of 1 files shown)


williamjallen:
Would it be possible to edit a few before/after screenshots which show the changes in this PR?  I believe that I have tested it but it would be helpful to make sure that I've checked all of the relevant spots.

shailpatels (approved):
Looks good, here's a screenshot of the button when a student has a grade override + grade inq.

image

I'm not sure what the part about "showing a grade inq if a student has no submission" in the PR description, I don't think its possible to submit a grade inq without a submission first

PR #7567, opened 2022-01-24
[Feature:Autograding] WIP: Add syscalls for Rust compilation and improve Docker image regex

benjaminrsherman:
Autograding tests trying to run `cargo` fail on the `statx` syscall.

### What is the new behavior?
Autograding tests can now use `cargo`.  Due to issues when copying files from the compilation stage to the run stage, `cargo run` will recompile programs even if they were built in a previous testcase.  The example Rust gradeable avoids this issue by just copying the built binary from the compilation stage.

### Other Information
The existing Docker regex doesn't allow official Docker images to be imported (they don't belong to an organization).  The new regex makes the organization component of the image URI optional.

### DIFF
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -509,7 +509,8 @@ if not args.worker:
                           "submitty/autograding-default:latest",
                           "submitty/java:11",
                           "submitty/python:3.6",
-                          "submittyrpi/csci1200:default"
+                          "submittyrpi/csci1200:default",
+                          "rust:1.58-slim"
                        ]
         }
 
--- a/site/app/controllers/DockerInterfaceController.php
+++ b/site/app/controllers/DockerInterfaceController.php
@@ -80,7 +80,7 @@ class DockerInterfaceController extends AbstractController {
         }
 
         // check for proper format
-        $match = preg_match('/^[a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+\/[a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+:[a-zA-Z0-9][a-zA-Z0-9._-]{0,127}$/', $_POST['image']);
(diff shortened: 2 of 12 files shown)


site/app/controllers/DockerInterfaceController.php
```diff
@@ -80,7 +80,7 @@ public function addImage() {
         }
 
         // check for proper format
-        $match = preg_match('/^[a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+\/[a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+:[a-zA-Z0-9][a-zA-Z0-9._-]{0,127}$/', $_POST['image']);
+        $match = preg_match('/^([a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+\/)?[a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+:[a-zA-Z0-9][a-zA-Z0-9._-]{0,127}$/', $_POST['image']);
```
MasterOdin, line 83: Do you know what would happen if an instructor used this container for grading as well? Would it "work" or just error in some way?

The worry I have is that these official images will "work", but because they lack seccomp and stuff, then we're not properly safeguarding execution, and instructors would not realize this.

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -509,7 +509,8 @@ def write(x=''):
                           "submitty/autograding-default:latest",
                           "submitty/java:11",
                           "submitty/python:3.6",
-                          "submittyrpi/csci1200:default"
+                          "submittyrpi/csci1200:default",
+                          "rust:1.58-slim"
```
MasterOdin, line 513: Let's move this so that it's added in `.setup/install_system.sh` only on vagrant systems, and not have it be installed by default for all Submitty installations.

site/app/controllers/DockerInterfaceController.php
benjaminrsherman: Since the Rust example gradeable runs, I don't think there will be an issue running the containers.  As for seccomp I _think_ that the safeguards will apply since `cargo` fails to run in the Rust container without the `statx` syscall allowed, but I'm not sure what exactly Submitty does to safeguard execution so there might be more constraints which don't get applied to the containers.

What would be the difference between an official image vs a "Submitty-blessed" one?

PR #7579, opened 2022-01-26
[Bugfix:Forum] Fix Markdown Line Number Positioning

eddiekrystowski:
The line numbers on code blocks do not line up properly with the code block.
image

### What is the new behavior?
The line numbers are properly aligned.
image

### Other information?
Another issue that was reported seemed to have duplicate/runaway line numbers like in the screenshot below but I have been unable to reproduce this aspect of the bug.
image


sbelsk (approved):
Works as described.

HenrikLam (approved):
The aspect of the bug that you weren't able to reproduce also seems to be fixed by this PR. Looks good to me!

Issue #7547, opened 2022-01-18
Bad line wrap for long poll responses


bmcutler:
**Describe the bug**
if a poll response is lengthy, it wraps poorly:
<img width="754" alt="Screen Shot 2022-01-18 at 5 24 19 PM" src="https://user-images.githubusercontent.com/5871417/150028589-73090ef5-12b4-4879-8d42-6c37d9e71129.png">

PR #7580, opened 2022-01-26
[Bugfix:SubminiPolls] Fix wrapping of long responses

sbelsk:
fixes https://github.com/Submitty/Submitty/issues/7547

### What is the new behavior?
Long responses now wrap around such that the first line of the response stays to the right of the radio box.
image


HenrikLam (approved):
Looks good to me.

PR #7585, opened 2022-01-29
[Bugfix:Plagiarism] Fix dates

williamjallen:
Various places in the plagiarism interface use the `g_grade_due_date` database column as the data source for gradeable due dates.  This is the wrong column and does not display the submission due date as seen by the students

### What is the new behavior?
Fixes the bug.  In addition, the behavior when dealing with gradeables with no due dates has been improved such that no due date is displayed on the interface.

New interface:
Untitled

### DIFF
--- a/site/app/controllers/admin/PlagiarismController.php
+++ b/site/app/controllers/admin/PlagiarismController.php
@@ -447,7 +447,7 @@ class PlagiarismController extends AbstractController {
             $configuration = [
                 "g_id" => $config->getGradeableID(),
                 "g_title" => $all_gradeables[$config->getGradeableID()],
-                "g_grade_due_date" => $this->core->getQueries()->getDateForGradeableById($config->getGradeableID()),
+                "due_date" => $this->core->getQueries()->getDueDateForGradeableById($config->getGradeableID()),
                 "g_config_version" => $config->getConfigID(),
                 "last_run_timestamp" => $config->getLastRunTimestamp(),
                 "has_been_viewed" => $config->userHasAccessed($user_id)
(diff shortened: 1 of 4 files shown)


site/app/templates/plagiarism/Plagiarism.twig
```diff
@@ -92,7 +92,7 @@
                 {{ row['title'] }}
             {% endif %}
             <br/>
-            Due: {{ row['duedate'] }}
+            {% if row['duedate'] != null %} Due: {{ row['duedate'] }} {% endif %}
```
MasterOdin, line 95: Shouldn't this be `row['due_date']`?

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -558,7 +562,7 @@ public function plagiarismMainPage(string $refresh_page = "NO_REFRESH"): WebResp
                 'title' => $gradeable['g_title'],
                 'id' => $gradeable['g_id'],
                 'config_id' => $gradeable['g_config_version'],
-                'duedate' => $gradeable['g_grade_due_date']->format($gradeable_date_format),
+                'duedate' => $gradeable['due_date'] == null ? null : $gradeable['due_date']->format($gradeable_date_format),
```
shailpatels, line 565: Could rewrite this as `$gradeable['due_date']->format($gradeable_date_format) ?? null`

with the https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op operator

site/app/controllers/admin/PlagiarismController.php
```diff
@@ -1016,12 +1020,19 @@ public function configurePlagiarismForm(): WebResponse {
                 unset($gradeable_ids_titles[$i]);
                 continue;
             }
-            $duedate = $this->core->getQueries()->getDateForGradeableById($gradeable_id_title['g_id']);
-            $gradeable_ids_titles[$i]['g_grade_due_date'] = $duedate->format($this->core->getConfig()->getDateTimeFormat()->getFormat('late_days_allowed'));
+            $duedate = $this->core->getQueries()->getDueDateForGradeableById($gradeable_id_title['g_id']);
+            $gradeable_ids_titles[$i]['due_date'] = $duedate == null ? 'no due date' : $duedate->format($this->core->getConfig()->getDateTimeFormat()->getFormat('late_days_allowed'));
```
shailpatels, line 1024: same comment as above

shailpatels (changes requested):

site/app/controllers/admin/PlagiarismController.php
williamjallen: Apologies for the late reply.  Unless I'm misunderstanding, this solution doesn't work here because we're trying to make sure we don't call `->format()` on something which is null, not checking to see if `->format()` returns null.  I believe that recent versions of PHP include the `?->` operator, which would be useful for this exact use-case, but it does not work here due to our usage of PHP7.x.

site/app/controllers/admin/PlagiarismController.php
williamjallen: (See above response)

PR #7594, opened 2022-01-31
[Bugfix:CourseMaterials] Fix hiding empty folders and showing new tags

### LINKED ISSUE #7372: New Course Materials only shows NEW tag one layer up
When a new document is added to course materials, there is a "NEW" tag visible on the document itself and its parent folder. However, if this document is nested within multiple folders, there is no visible "NEW" tag to students on the "higher up" folders.

Expected behavior: if there is any new document within a folder, then there will be a "NEW" tag on that folder (even if the document is deeply nested within the folder).

### LINKED ISSUE #7533: Fix behavior of hideEmptyCourseMaterialFolders
**Describe the bug**
When the "Course Materials" page is reloaded, empty course folders briefly appear before disappearing. This does not look good.

**Expected behavior**
Empty course folders should not briefly appear like this.

**To Reproduce**
1. Go to a "Course Materials" page.
2. Look at the file system.
3. Refresh the page to do again.

**Configuration**
 - OS: Windows
 - Browser" Chrome

**Screenshots**

**Additional context**
I believe this is caused because hideEmptyCourseMaterialFolders is called too late in the code.
I am not so good with web exploitation, but I'm concerned this could cause a greater issue if one were to click the "download folder" button on an empty folder.

DefinitiveAbove:
Empty folders will briefly appear when loading the course materials page before disappearing.
New tags on a file/folder only show up on the file/folder and direct parent folder.

### What is the new behavior?
Closes #7533 
Empty folders will not show up when loading the course materials page.
Closes #7372 
New tags on a file/folder extend all the way up to the root folder on the course materials page.

### DIFF
--- a/site/app/views/course/CourseMaterialsView.php
+++ b/site/app/views/course/CourseMaterialsView.php
@@ -148,8 +148,9 @@ class CourseMaterialsView extends AbstractView {
                 if ($this->removeEmptyFolders($course_material)) {
                     unset($course_materials[$path]);
                 }
-            } else {
-                return false;   
+            }
+            else {
+                return false;
             }
         }
         return true;
--- a/site/app/views/course/CourseMaterialsView.php
+++ b/site/app/views/course/CourseMaterialsView.php
@@ -121,6 +121,8 @@ class CourseMaterialsView extends AbstractView {
                 [$file_name => $course_material] + array_slice($path_to_place, $index, null, true);
         }
 
+        $this->removeEmptyFolders($final_structure);
+
         $this->setSeen($final_structure, $seen, $base_course_material_path);
 
         return $this->core->getOutput()->renderTwigTemplate("course/CourseMaterials.twig", [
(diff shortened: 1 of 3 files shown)


shailpatels (approved):
Looks good, fixes the bug as described

loadedskiffer (approved):

site/app/views/course/CourseMaterialsView.php
```diff
@@ -140,13 +142,28 @@ public function listCourseMaterials(array $course_materials_db) {
...
 
+    private function removeEmptyFolders(array &$course_materials): bool {
+        foreach ($course_materials as $path => $course_material) {
+            if (is_array($course_material)) {
+                if ($this->removeEmptyFolders($course_material)) {
+                    unset($course_materials[$path]);
+                }
+            }
+            else {
+                return false;
```
MasterOdin, line 153: Why the `return false;` here? Does `$course_materials` here have an implicit sort that it has all folders first, followed by files? Is it important that we enforce this sort here in some way? Can we just remove the whole `else` branch, and just have it always iterate over all `$course_materials` here? I don't think any instructor has such a long list of course materials that we need to worry about performance optimizations within this loop at the cost of having some implicit necessary structure that is easy to overlook.

site/app/views/course/CourseMaterialsView.php
```diff
@@ -140,13 +142,28 @@ public function listCourseMaterials(array $course_materials_db) {
         ]);
     }
 
+    private function removeEmptyFolders(array &$course_materials): bool {
```
MasterOdin, line 145: Can we remove the `bool` here and just make this `void`? Unless we plan to utilize the returned value in some way, better to have less code.

MasterOdin (dismissed):

site/app/views/course/CourseMaterialsView.php
DefinitiveAbove: The bool/return false is needed for the return call on line 148 to determine if the folder was empty/only had empty folders in it

PR #7610, opened 2022-02-04
[Bugfix:HelpQueue] stop queue occupancy table replicating

loadedskiffer:

Currently web sockets in the queue are producing a bug that makes the queue occupancy table/buttons replicate. 
pr28_1

### What is the new behavior?
The table/buttons do not replicate.

### Other information?
To replicate bug:
1. Join a queue in one window
2. In another window as an instructor, do something to trigger a web socket refresh (open a queue, close a queue, remove a student)
3. See that on the students queue page, the table/buttons are replicated


williamjallen (commented):
Is there any particular reason why it is necessary to have duplicate HTML here?

loadedskiffer:
> Is there any particular reason why it is necessary to have duplicate HTML here?

Due to the way that the websockets refresh the page If I were to put the table in QueueStaus.twig I'm pretty positive I would need to have duplicate html but I think this is a better solution

williamjallen:
Untitled
The table-replication seems to have been fixed, but I was able to make it duplicate the "Close Table" buttons using the same methods that caused the original table replication bug.

loadedskiffer:
> 

I'm not sure what's happening here, no matter what I do I can't reproduce it on this branch or on master.

williamjallen:
Interesting.  I just checked again to make sure, but I still see the buttons being duplicated.  The table replication has been fixed it seems.  I double-checked and I believe that I have the latest changes.

I have attached a screen recording of the behavior I am observing.  If necessary, we could discuss/debug this together at the next meeting.

I don't think it's related but in case it's helpful, the student in the right pane in the screen recording is Joe Student (who is already in the queue when the page is loaded) and the student in the left pane in the screen recording is Ben Bitdiddle (not in the queue initially)

https://user-images.githubusercontent.com/16820599/153617984-31446ae4-bb1c-4c09-93e0-f382f1777b22.mov

tkoz0 (commented):
I tested this and observe the queue occupancy table and buttons completely disappearing on the student's page when the instructor opens or closes a queue.

tkoz0:
I continued testing this further (with multiple students) and I am not finding a situation where the button replicates. Every time I have the instructor make an update to a queue, the queue occupancy info completely disappears for all the students.

cjreed121 (approved):
Tested locally and seems to fix the issue.

PR #7623, opened 2022-02-07
[Refactor:Developer] Typescript config updates

DefinitiveAbove:
### What is the new behavior?
Enabled strict checking for Typescript (and updated existing files to meet checking)
Allow use of variables/functions from other Javascript functions in Typescript functions
Added Twig type library
Fix .build.js not going into directories

### DIFF
--- a/site/.build.js
+++ b/site/.build.js
@@ -4,7 +4,9 @@ const module_path = path.join(__dirname, 'ts');
 
 function getAllFiles(dir) {
     return fs.readdirSync(dir, { withFileTypes: true }).reduce((acc, entry) => {
-        if (!entry.isDirectory() && entry.name.endsWith('.ts') || entry.name.endsWith('.js')) {
+        if (entry.isDirectory()) {
+            acc.push(...getAllFiles(path.join(dir, entry.name)));
+        } else if (entry.name.endsWith('.ts') || entry.name.endsWith('.js')) {
             acc.push(path.join(dir, entry.name));
         }
         
--- a/site/.build.js
+++ b/site/.build.js
@@ -4,9 +4,7 @@ const module_path = path.join(__dirname, 'ts');
 
 function getAllFiles(dir) {
     return fs.readdirSync(dir, { withFileTypes: true }).reduce((acc, entry) => {
-        if (entry.isDirectory()) {
-            acc.push(...getAllFiles(path.join(dir, entry.name)));
-        } else if (entry.name.endsWith('.ts') || entry.name.endsWith('.js')) {
+        if (!entry.isDirectory() && entry.name.endsWith('.ts') || entry.name.endsWith('.js')) {
             acc.push(path.join(dir, entry.name));
         }
         
--- a/site/ts/submission.ts
+++ b/site/ts/submission.ts
@@ -3,8 +3,8 @@ declare global {
     interface Window{
         file_array: File[][];
         num_submission_boxes: number;
-        deleteFiles(Number): void;
-        addFile(File, Number, Boolean):void;
+        deleteFiles(part: Number): void;
+        addFile(file: File, part: Number, check_duplicate_zip: Boolean):void;
         loadPreviousFilesOnDropBoxes():void;
     }
 }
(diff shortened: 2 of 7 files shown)


site/.build.js
```diff
@@ -4,7 +4,9 @@ const module_path = path.join(__dirname, 'ts');
 
 function getAllFiles(dir) {
     return fs.readdirSync(dir, { withFileTypes: true }).reduce((acc, entry) => {
-        if (!entry.isDirectory() && entry.name.endsWith('.ts') || entry.name.endsWith('.js')) {
+        if (entry.isDirectory()) {
+            acc.push(...getAllFiles(path.join(dir, entry.name)));
```
MasterOdin, line 8: Why do we need to be entering directories? Our entry points are only at the top level of the `ts` folder.

site/ts/submission.ts
```diff
@@ -3,8 +3,8 @@ declare global {
     interface Window{
         file_array: File[][];
         num_submission_boxes: number;
-        deleteFiles(Number): void;
-        addFile(File, Number, Boolean):void;
+        deleteFiles(part: Number): void;
+        addFile(file: File, part: Number, check_duplicate_zip: Boolean):void;
```
MasterOdin, line 7:
```suggestion
        deleteFiles(part: number): void;
        addFile(file: File, part: number, check_duplicate_zip: boolean): void;
```

site/ts/utils/server.ts
```diff
@@ -14,7 +14,7 @@ export function buildCourseUrl(parts: string[] = []) : string{
  * @returns string
  */
 export function getCsrfToken() : string{
-    return document.body.dataset.csrfToken;
+    return document.body.dataset.csrfToken ?? "";
```
MasterOdin, line 17: return document.body.dataset.csrfToken ?? '';

site/tsconfig.json
```diff
@@ -1,3 +1,8 @@
 {
-    "include" : ["./ts/**/*.ts"]
+    "include" : ["./ts/**/*.ts", "./public/js/**/*.js"],
+    "compilerOptions": {
+        "strict": true,
+        "allowJs": true,
+        "noEmit": true
+    }
 }
```
MasterOdin, line 8: Should add a step to the `.github` action that runs `tsc` to validate our typescript usage.

MasterOdin (changes requested):

site/.build.js
shailpatels: @DefinitiveAbove Here is a blurb talking about the TS folder structure https://submitty.org/developer/developing_the_php_site/javascript#creating-new-modules.

The idea is that entry points (top level files) are bundled and will have a corresponding `js` file in the `mjs` directory, and any files in a subdirectory of `ts` will use the es6 module syntax `import` and webpack will just bundle everything together into a single file instead of needing to send out multiple

site/tsconfig.json
```diff
@@ -1,3 +1,8 @@
 {
-    "include" : ["./ts/**/*.ts"]
+    "include" : ["./ts/**/*.ts", "./public/js/**/*.js"],
+    "compilerOptions": {
+        "strict": true,
+        "allowJs": true,
+        "noEmit": true
```
shailpatels, line 6: You can remove the `-noEmit` from the "package.json" `build` script since its being added here now

site/tsconfig.json
shailpatels: @HenrikLam Probably best place to add it is the existing `js-lint` job

https://github.com/Submitty/Submitty/blob/2e30229699e49f376e4f7b2028b52846072f3796/.github/workflows/submitty_ci.yml#L59-L61

and just change too

```
- name: Run eslint & tsc
- run: |
    npm run eslint
    npm run build
```

shailpatels (approved):
I'm happy with the changes after eslint and Matt's comments are resolved

site/tsconfig.json
MasterOdin: Please put these as two separate steps in the build so that you can tell at a glance what run script failed before you even expand the tab for the full details.

MasterOdin (approved):

PR #7628, opened 2022-02-09
[Refactor:Submission] Move instructor submission logic to TS

shailpatels:

The majority of instructor submission logic in a twig file instead of its own js file prevents good separation and utilizing js module features. Also if an instructor changes a submission mode in one gradeable, the change will carry over to other gradeables 

### What is the new behavior?
This PR refactors the submission logic for handling the state of the instructor submission settings out of a twig file and into a module.

It also restricts saving the state to the current browser session + gradable id so if an instructor changes any options and goes to a new gradeable, the settings between gradeables will not affect each other.

The module `submission.ts` has been renamed to `instructor-submission.ts` since all the logic in it is specific to instructors and should only be loaded when a user has instructor-level privileges. 

### Other information?
No changes in behavior other then preventing submission state mode from effecting multiple gradeable ids


bmcutler:
@shailpatels 
please fix merge conflicts

DefinitiveAbove (approved):

PR #7631, opened 2022-02-09
[Refactor:Developer] Remove invalid CSS properties

williamjallen:
https://github.com/Submitty/Submitty/pull/7630 is currently failing the `css-stylelint` test because there are a couple places in the codebase that use invalid CSS `filter` properties.  

### What is the new behavior?
This PR removes the offending properties.

### DIFF
--- a/site/.stylelintrc.json
+++ b/site/.stylelintrc.json
@@ -8,6 +8,7 @@
   },
   "ignoreFiles": [
     "vendor/**",
+    "tests/report/**",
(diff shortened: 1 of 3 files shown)


MasterOdin (approved):
Looks like `filter: alpha(opacity=80)` and `filter: progid:dximagetransform.microsoft.gradient(enabled = false);` were for compatibility with old IE versions, which are long since dead and we don't support, so :+1: on removing.

PR #7632, opened 2022-02-09
[Refactor:Forum] Doctrine Forum Entities

cjreed121:

There are no models / entities for the forum right now.

### What is the new behavior?
This PR adds the Doctrine entities for the forum. They aren't used anywhere yet but future PRs will be able to use these rather than DatabaseQueries. There will likely be changes to these for eager loading but this PR is mainly for structure.

I tested these with a temporary route and they seem to work well.


MasterOdin (approved):
Looks fine, but hard to do a full review of this until it's actually hooked up to examine performance, etc.

PR #7633, opened 2022-02-09
[Bugfix:TAGrading] Improve late day banner visuals

### LINKED ISSUE #7581: Late vs Bad Gradeable Visuals Unclear
**Describe the bug**
At the moment a gradeable can have a "late" banner if:
1. The version displayed was submitted after the due date and is still on time (due to late days/extensions)
OR
2. The version displayed was submitted after the due date and will be in the "Bad" state (too many late days/extensions used) AND there IS an earlier version that could be counted as "Late"/"On Time"
OR
3. The version displayed was submitted after the due date and will be in the "Bad" state (too many late days/extensions used) AND there IS NOT an earlier version that could be counted as "Late"/"On Time"

Currently if case 1 or 2 is true, when viewing the submission from the grading interface, the background is orange and an orange banner is displayed up top stating `Late (Roll-back to on time submission -X)` where X is a version number of a submission prior to the due date.

If case 3 is true, then instead the grading interface will have a red background and a red banner that states `Late (No on time submissions available)`

**Expected behavior**
This section needs to be discussed with @bmcutler 

Case 1 behaves as expected.

However, in cases 2 AND 3, we should display the red banner, since we are trying to indicate to the grader that until something changes (extension applied, student reverts to an earlier version if applicable), the grader probably does not need to grade this submission. Whether or not an earlier version exists to roll back to should not impact the grader's choice, since the student, not the grader should be the one to decide to select a different active version.

For case 2 we can still display the text about `(Roll-back to on time submission -X)`.

For cases 2 and 3 it may also be worth changing the word `Late` to something more descriptive to distinguish it from the orange banner. (Orange and red can be difficult for some people to differentiate.)

**To Reproduce**
1. Create a new gradeable with a deadline in the future. Make sure maximum late days allowed for the gradeable is 0 and that submissions will be accepted after the deadline.
2. On one user, make a submission to the gradeable
3. Close the gradeable (set the due date to Now)
4. On the same user, make a second submission to the gradeable
5. On a different user, make a submission to the gradeable
6. As a grader, view both submissions from the grading interface
7. See the difference

DefinitiveAbove:
* #7581 

### What is the new behavior?
Closes #7581 
Shows `Late Submission` in orange when selected submission is late but has enough late days/extensions to cover it.

Shows `Bad Submission (submitter has valid submission - Version #X)` in red when selected submission is late and does not have enough late days/extensions to cover it, but has a previous submission that is valid.

Shows `Bad Submission (no valid submission available)` in red when selected submission is late and does not have enough late days/extensions to cover it, and does not have a previous submission that is valid.

### DIFF
--- a/site/app/views/grading/ElectronicGraderView.php
+++ b/site/app/views/grading/ElectronicGraderView.php
@@ -894,7 +894,7 @@ HTML;
 
     //The student not in section variable indicates that an full access grader is viewing a student that is not in their
     //assigned section. canViewWholeGradeable determines whether hidden testcases can be viewed.
-    public function hwGradingPage(Gradeable $gradeable, GradedGradeable $graded_gradeable, int $display_version, float $progress, bool $show_hidden_cases, bool $can_inquiry, bool $can_verify, bool $show_verify_all, bool $show_silent_edit, string $late_status, $rollbackSubmission, $sort, $direction, $from, array $solution_ta_notes, array $submitter_itempool_map, $anon_mode, $blind_grading) {
+    public function hwGradingPage(Gradeable $gradeable, GradedGradeable $graded_gradeable, int $display_version, float $progress, bool $show_hidden_cases, bool $can_inquiry, bool $can_verify, bool $show_verify_all, bool $show_silent_edit, int $late_status, $rollbackSubmission, $sort, $direction, $from, array $solution_ta_notes, array $submitter_itempool_map, $anon_mode, $blind_grading) {
         $this->core->getOutput()->addInternalCss('admin-gradeable.css');
         $isPeerPanel = false;
         $isStudentInfoPanel = true;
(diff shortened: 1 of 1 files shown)


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -963,17 +963,25 @@ public function hwGradingPage(Gradeable $gradeable, GradedGradeable $graded_grad
...
-                "message" => "Late Submission (Rollback to on-time submission - " . $rollbackSubmission . ")"
+                "message" => "Late Submission"
             ];
         }
-        elseif ($late_status != LateDayInfo::STATUS_GOOD && $late_status != LateDayInfo::STATUS_LATE) {
-            $error_message = [
-                "color" => "var(--standard-red-orange)", // fire engine red
-                "message" => "Late Submission (No on time submission available)"
-            ];
+        elseif ($late_status == LateDayInfo::STATUS_BAD) {
```
williamjallen, line 972:
```suggestion
        elseif ($late_status === LateDayInfo::STATUS_BAD) {
```

williamjallen (changes requested):
Besides the minor code change I requested, this PR seems to work as expected.

williamjallen:
This isn't necessarily something has to be done in this PR but since reproducing each of these cases requires a decent bit of manual work, perhaps it would be beneficial to write a small number of Cypress tests for this.

williamjallen (approved):

PR #7642, opened 2022-02-14
[Feature:Autograding] Docker UI OS version

### LINKED ISSUE #7546: Add "operating system version" to the Docker Interface UI

Currently we display a column for the "Docker Version" of each worker machine.

Let's also add a column for the "OS version" of each worker machine.   Currently we have some workers that are Ubuntu 18.04 and some that are Ubuntu 20.04.  It would be nice to have them labeled and not have to guess.

HenrikLam:

Closes #7546
Currently the page does not display information on the system of the worker machines.
image

### What is the new behavior?
Adds a new OS information column to the worker machines table.
image

### Other information?

To test this, you will need to press the update dockers and machines button located at the top of the page to trigger a log file to be written.
I was not able to test the page nor the log writing with worker machines present, which could be an introduction of potential bugs.


sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -155,7 +154,7 @@ def parse_arguments():
 
 def update_machine(machine,stats,args):
 
-    print(f"UPDATE MACHINE: {machine}")
+    os.system(f"echo UPDATE MACHINE: {machine}")
```
MasterOdin, line 157: Why change this from `print` to using `os.system`?

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -185,8 +184,6 @@ def update_machine(machine,stats,args):
     if success == False:
         print(f"ERROR: Failed to pull one or more required docker images on {machine}")
         return False
-
-    print(f"finished updating machine: {machine}")
```
MasterOdin, line 189: Why remove this print?

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -83,22 +84,20 @@ def run_commands_on_worker(user, host, commands, operation='unspecified operatio
         return True
     else:
         success = False
-        timed_out = False
```
DefinitiveAbove, line 86: It doesn't look like this is used anywhere (outside of being set to true)

DefinitiveAbove (commented):
Functionality looks good for the most part; I tested by merging `vagrant-worker-bugfix` into this branch - I'm not sure if this is related at all to this PR but `Docker Version` on the `primary` machine outputs `Error`.

sbin/shipper_utils/update_and_install_workers.py
HenrikLam: The change here is due to the linux command being executed before the print statement here by the OS (or so I believe). By changing the print to an os.system, this seems to ensure that the `UPDATE MACHINE` line is printed before everything to ensure the log parsing is done properly.

MasterOdin: Yeah, this is because of how python works where it buffers text before writing it out to `stdout` and `stderr` so that the prints more efficient at the cost that it doesn't necessarily output immediately. `os.system` circumvents this buffering because of how it calls down to the underlying OS, at the expense that we also cannot do anything with its output or know that it errored really.

The better solution would be to make sure we just don't use `os.system` anywhere, and instead [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run) (or equivalent `subprocess` method) which has the added benefit that it forces us to consider error cases where a command fails for some reason. This will also help us avoid these sorts of race conditions with prints when we're trying to parse the outputted log file.

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -65,13 +65,14 @@ def update_docker_images(user, host, worker, autograding_workers, autograding_co
         docker_images_obj = client.images.list()
         #print the details of the image
         get_docker_info.printDockerInfo()
-        
+        os.system('lsb_release -a')
```
MasterOdin, line 68: From above, we can replace this with
```suggestion
        subprocess.run(['lsb_release', '-a'], capture_output=True, check=True)
```

Note, this change does mean that if the command exits with a non-zero exit code, then an exception will be raised (instead of it being silently ignored like `os.system`), but that's probably what we want here instead of having a broken log file that we only notice when we go to parse it.

MasterOdin (changes requested):

HenrikLam:
To whoever that will take over this PR. When I was trying to test this PR, I was not able to set up my worker machine because of complications with my laptop. What happened previously was that python and the OS had different buffers and that caused the output in the log file to be out of order when worker machines are involved. I have included what might be a working solution but I have no way to test it. The first step to taking over this PR will be to set up worker machine in your device. Instructions on this can be found [here](https://submitty.org/developer/getting_started/worker_vm). To trigger the log file to be output, go to DockerUI page and hit the update button at the top of the page.

Viyerelu23333 (approved):

tkoz0 (approved):
I tested this with a worker machine and it works as expected. The code looks good.

Viyerelu23333 (approved):

PR #7648, opened 2022-02-15
[Feature:Developer] Minimal mode toggle for output

DefinitiveAbove:
Output.php always has header/sidebar with global header/footer enabled.

### What is the new behavior?
There is an additional toggle in `Output.php` called `render_full` - specifies if a page should be rendered in full (with header/sidebar) or if it should be rendered minimally (for popups).


williamjallen (commented):
Code looks good overall, but I have a minor nitpick about the name of this feature.  "render full" is on by default, so any page which utilizes this function will likely do so by calling `setRenderFull(false)`.  If I were reading through code, having no prior knowledge of the existence of this command, I feel like the name "setRenderFull" (or more generally, the concept of "render full" mode) doesn't effectively explain what this mode does.

There are two main issues here:
1. Is it better to have a "render full" mode which gets turned off, or is it better to have a "minimal mode" that gets turned on?
2. In my opinion, "render full" doesn't adequately convey information about what this command does.  A non-comprehensive list of potential alternatives I came up with: "minimal mode", "full page content", "useUIHeader"/"useUIFooter", "content only mode".  My suggestion is "content only mode" which could have `setContentOnly()`, `isContentOnly()`.

If you feel strongly that "render full" is the best name for this, I have no problem approving that.  Just some food for thought.

MasterOdin:
Why do want this? Is the plan to add more pages that would use some minimal setup? Why are some elements (like system messages) not hidden in these popups?

DefinitiveAbove:
> Why do want this? Is the plan to add more pages that would use some minimal setup? Why are some elements (like system messages) not hidden in these popups?
> 
> It would be good to document these things here in case we ever look back to this PR months or years from now.

I'm planning on using this page for having popups in TA grading for each the panels; I left the messages in for if there are messages that are relevant to the popup at hand, but if it's better to not have them I can remove them.

site/app/controllers/MiscController.php
```diff
@@ -181,6 +179,8 @@ public function displayFile($dir = null, $path = null, $gradeable_id = null, $us
                 $this->core->getOutput()->renderOutput('Misc', 'displayCode', $file_type, $corrected_name, $contents);
             }
             else {
+                $this->core->getOutput()->useHeader(false);
+                $this->core->getOutput()->useFooter(false);
```
williamjallen, line 183: What purpose did this change serve?  (Just to be clear, I'm not requesting changes here.  I just didn't see the reasoning behind it immediately and wanted to understand)

site/app/libraries/Output.php
```diff
@@ -549,6 +551,14 @@ public function useFooter($bool = true) {
...
     }
 
+    public function setContentOnly($bool = false) {
+        $this->content_only = $bool;
+    }
+
+    public function isContentOnly() {
+        return $this->content_only;
+    }
+
```
williamjallen, line 561: It would be good to add type hints here to match the surrounding code

site/app/controllers/MiscController.php
DefinitiveAbove: I think I was thinking that the file was a raw file that was being outputted but that's not the case; I changed it to also use content_only here

williamjallen (approved):

MasterOdin (approved):

PR #7649, opened 2022-02-15
[Refactor:TAGrading] Autograding/Student Info panels to TS

### LINKED ISSUE #7689: Calling TypeScript functions from Twig/Event Listeners
Currently, any inputs/items that need events that depend on TypeScript code either has the event listener for the item added inside of the TypeScript file that has the function to be called or in a Twig file under a module type script tag. While it isn't much of an issue now, as more files are migrated over to TypeScript this could lead to worse code readability when there are TypeScript files that are depended on by multiple Twig files that need different event listeners (putting them in the top-level module when it is depended on by multiple Twig files means adding event listeners that are unnecessary for some of the files or having multiple init functions in the module, and putting them in the Twig file requires a separate script tag to add them and loses the TypeScript checks). Additionally, importing from module files in Twig means that you have to get the timestamp of the file over to Twig if you don't want to request the file twice.

Below are some potential ideas for trying to keep it a bit more organized:

Attach objects from TypeScript onto the `window` global variable:

+ Can call functions directly in non-module JS/in HTML in Twig; very similar to current style of many Twig files
- No TypeScript coverage in Twig files
- Potentially clutter `window` global variable (one way to reduce this clutter could be to have each top-level module only put one object on the `window` global variable and have all of the necessary functions/variables for that module in that object)

Individual .ts files for each Twig file that are all entry points:

+ Potentially full TypeScript coverage (if no calls in Twig files)
- Potential bloat from redundant bundled dependencies, many files may lead to high overhead

Individual .ts files for each Twig file that are exported into a single entry point:

+ Potentially full TypeScript coverage (if no calls in Twig files)
- File is bloated for most pages

It should also be noted that to have full TypeScript coverage, the data from Twig has to be sent through other methods (attributes in HTML/AJAX calls).

DefinitiveAbove:
Autograding/student info panels are in Javascript.

### What is the new behavior?
Refactor autograding/student info panels to TS modules.
Change format of ESBuild to `esm` (allows calling from outside of built JS), enabled code splitting for TS.
Closes #7689 - Added new folder under the TS folder which are all entry points - `twig` contains TS files that are directly related to Twig files (init/functions for a single Twig file should be made in a TS file specifically for that Twig file under the `twig` folder; code not directly associated with that Twig file should be moved into a support file not under the `twig` folder)

Added a function to be used in Twig files called `add_twig_module_js` - adds a .js file to Output.php from the /mjs/twig folder; this is mainly for cases where there are nested Twig files to avoid having to find and add the .js file from each Twig file if applicable.


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1197,6 +1197,7 @@ public function renderAutogradingPanel($version_instance, bool $show_hidden_case
         $this->core->getOutput()->addInternalJs('submission-page.js');
         $this->core->getOutput()->addInternalJs('drag-and-drop.js');
         $this->core->getOutput()->addVendorJs('bootstrap/js/bootstrap.bundle.min.js');
+        $this->core->getOutput()->addModuleJs("grading.js");
```
shailpatels, line 1200: Should this be `addInternalModuleJs` instead?

The `addModuleJS` function won't timestamp your given URL and also won't search the `mjs` directory for the file and causing a JS error.
Also should adding this file be in both the autograding and submission panel I think it would be sending this file over twice

shailpatels (changes requested):

shailpatels:
I'm happy with the change for using `esm` as the output format, all modern browsers should support it and its a better method than defining an interface and putting variables/functions in the global `window` object

shailpatels (approved):

cjreed121 (approved):
The pages seems to retain the same functionality. Looks good to me.

.setup/install_submitty/install_site.sh
```diff
@@ -38,6 +38,17 @@ set_permissions () {
     esac
 }
 
+set_mjs_permission () {
+    for file in $1/*; do
+        if [ -d "$file" ]; then
+            chmod 551 $file
```
shailpatels, line 44: Is there a specific reason for using `551`, I think only the php site user needs execute permissions on JS so there wouldn't be a need for the user or group to execute it?

site/.build.js
```diff
@@ -2,22 +2,36 @@ const fs = require('fs');
 const path = require('path');
 const module_path = path.join(__dirname, 'ts');
 
-function getAllFiles(dir) {
+function getTwigFiles(dir) {
```
shailpatels, line 5: Maybe this function name should be more descriptive to make sure people know this is collecting the files under `ts/twig` and not under `app`

site/app/libraries/Output.php
```diff
@@ -514,8 +518,12 @@ public function addCss(string $url): void {
         $this->css->add($url);
     }
 
+    public function addInternalModuleTwigJs(string $file) {
+        $this->addModuleJs($this->timestampResource($file, 'mjs/twig'));
+    }
+
     public function addInternalModuleJs(string $file) {
-        $this->addModuleJs($this->timestampResource($file, 'mjs'));
+        $this->addModuleJs($this->timestampResource($file, 'mjs/module'));
```
shailpatels, line 526: Is there a reason to move the existing js and ts files into a module directory under `ts`? Those files were intended to be the top level entry points and any files in a subdirectory would be bundled into those entry points.

The `mjs` typically indicates ecma script module so I think the second dir is redundant

shailpatels (changes requested):

.setup/install_submitty/install_site.sh
DefinitiveAbove: The chmod here is only setting the permissions to `551` if it is a directory (if it is a file it calls `set_permissions` like before); I'm not sure what permissions child directories need though and I just assumed it based on the permissions set on the `mjs` directory previously.

site/ts/twig/grading/electronic/AutogradingPanel.ts
```diff
@@ -0,0 +1,57 @@
+import { getCsrfToken } from '../../../utils/server';
```
MasterOdin, line 1: Why put this file under `twig/` directory? All of these files are being applied to twig files that this hierarchy doesn't make sense to me.

DefinitiveAbove: I put this file under the `twig` directory since there are functions and event listeners specific to that twig file; the main reason that I made the `twig` directory is to keep the event listeners per twig page more organized, since HTML attributes such as `onclick` can't be used anymore without putting the functions in TypeScript on a global variable. An `mjs` file should be under the `twig` folder if there are event listeners/functions that are for a specific Twig file; anything that has to be reused by multiple Twig files/TypeScript files or isn't related to a Twig file should be put under either a top entry point or as a utility. While you can have all of the twig TS files be top level entry points, I think that it might lead to unnecessary clutter on the top level and makes the TS files that the Twig files are using less clear.

While having multiple ways to import the TS files isn't ideal, I added the way to add them in Twig files since it avoids issues with having to figure out what TS files they want to import in cases where there are Twig files that `include` other Twig files (especially when the Twig files are conditionally imported).

One way that might be less confusing is to have only Twig TS files be entry points and to have them only be imported from Twig files since shared code can just be imported through TypeScript - this would make it easier to debug since you would just have to look for the associated Twig file to find its TS file and would be able to see everything that Twig file uses from the TS file.

site/.build.js
```diff
@@ -2,22 +2,36 @@ const fs = require('fs');
 const path = require('path');
 const module_path = path.join(__dirname, 'ts');
 
-function getAllFiles(dir) {
+function getTwigTSFiles(dir) {
     return fs.readdirSync(dir, { withFileTypes: true }).reduce((acc, entry) => {
-        if (!entry.isDirectory() && entry.name.endsWith('.ts') || entry.name.endsWith('.js')) {
+        if (entry.isDirectory()) {
```
shailpatels, line 7: This is related to Matt's comment about the file structure but I think this basically assumes every file under `twig` is an entry point since they will all be given to esbuild to bundle - even if a file is only used as a utility. Not sure if this is what we want since it might cause additional files in the `mjs` dir, but if every file in that directory is an entry point then maybe we can just bring it all up to the top level like @MasterOdin suggested ? 

Not sure how much mirroring the file structure helps when figuring out which file is associated with which twig template since at the end of the day its the developer who decides which `mjs` files are delivered to the user

shailpatels (approved):
Approving this for moving some of the inline JS code to their own files. Functionality for student info and autograding panels looks the same and didn't see any issues with any other gradeable types

Not sure what direction to take regarding the organization of `twig` js files - personally I'd prefer them as just top level files but am fine with either or

MasterOdin:
Not thrilled with this, but in the interest of not blocking on my own super limited time at the moment, merging. As we move forward with migrating to `mjs`, can reconsider the approach.

PR #7657, opened 2022-02-16
[Bugfix:CourseMaterials] Hide empty folders bugfix

DefinitiveAbove:
Hiding empty folders does not work as intended.

### What is the new behavior?
Hiding empty folders works as intended.

### DIFF
--- a/site/app/views/course/CourseMaterialsView.php
+++ b/site/app/views/course/CourseMaterialsView.php
@@ -155,17 +155,16 @@ class CourseMaterialsView extends AbstractView {
     }
 
     private function removeEmptyFolders(array &$course_materials): bool {
+        $is_empty = true;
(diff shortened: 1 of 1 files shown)


MasterOdin (approved):
@DefinitiveAbove Please provide more context in your PR messages to explain the what and why this PR exists. The PR should be able to stand up in isolation (or alongside a linked issue) and make sense to a new developer (who reasonably understands Submitty) who does not have the benefit of the Spring 2022 weekly meetings or Slack context. The context of "Hide empty folders bugfix" along with what you had originally put "It did not work as intended" is not sufficient detail for this developer to know exactly what the bug was, how to reproduce it, etc. without also then having to dive into understanding what the old code did and how the new code changes it. I've rewritten the original message to contain more details on what I'd like to see.

PR #7668, opened 2022-02-23
[Bugfix:Developer] Fix Vagrant worker setup error

DefinitiveAbove:
The setup script for running a worker machine in Vagrant fails as it was missing a necessary environment variable, and we were not properly echoing a string into the configure script, causing it to hang.

### What is the new behavior?
The setup script for running a worker machine does not fail.


.setup/install_worker.sh
```diff
@@ -8,6 +8,7 @@
 
 GIT_PATH=/usr/local/submitty/GIT_CHECKOUT/Submitty
 SUPERVISOR_USER=submitty
+NO_RPI=0
```
MasterOdin, line 11: Is running the `rpi.sh` script strictly necessary for having the vagrant worker operate properly or can we omit it altogether from this script, similar to how it's omitted in `.setup/install_system.sh` and is up to the end user to run the script (or not).

DefinitiveAbove: We talked about this a bit during the meeting and decided on leaving the rpi.sh script in with the flag in this file set to 0 by default.

MasterOdin: Could you document as to _why_ it was decided this way? A good mantra here is "If it's not written down, it did not happen", and that in a year, are the people at that meeting going to remember why they agreed upon doing this? I don't particularly care if we always run `rpi.sh` for the worker for testing, I just want this documented somewhere.

DefinitiveAbove: Discussed more during a different meeting - we decided on removing the flag/script from this file (since they're RPI specific packages)

MasterOdin (approved):

PR #7676, opened 2022-02-26
[SYSADMIN ACTION][Feature:System] Auto-rewrite trailing slash

### LINKED ISSUE #7253: URL routing flexibility

The url to the main page of my course, and the forum for my course are:
https://submitty.cs.rpi.edu/courses/f21/csci1200
https://submitty.cs.rpi.edu/courses/f21/csci1200/forum

But if I manually type in these urls and add a trailing slash, e.g.,
`https://submitty.cs.rpi.edu/courses/f21/csci1200/`
`https://submitty.cs.rpi.edu/courses/f21/csci1200/forum/`
I get re-directed to the main submitty page (it is treated as an invalid url/route)

Can we make the routing a little more flexible and go to the intuitive page?

wusatosi:

* [x] Tests for the changes have been added/updated
* [x] Documentation has been updated/added if relevant: https://github.com/Submitty/submitty.github.io/pull/407

See https://submitty.org/sysadmin/version_notes/v22.03.00

If the URL contains a trailing slash, the user gets redirected to the login page because the current system cannot path match the URL.
For example:
```/courses/f21/csci1200/ -> /authentication/login```

### What is the new behavior?
This PR provides more flexibility to the URL as this instructs Apache to redirect the request to exclude the trailing slash via 301.
For example: `/courses/f21/csci1200/ -> /courses/f21/csci1200`

### Other information?
Closes #7253.

### CODE REVIEW

**Review by MasterOdin — CHANGES_REQUESTED** [review]

File: .setup/apache/submitty.conf [hunk 1]
```diff
@@ -40,6 +40,13 @@
         Require all granted
 
         RewriteEngine On
+        
+        # Routing Flexibility: Removes the trailing slash if it is supplied
+        # Example: /courses/f21/csci1200/ -> /courses/f21/csci1200
+        # See issue #7253
```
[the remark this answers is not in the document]
wusatosi: resolved in f93a5f8

MasterOdin (approved):

MasterOdin:
@bmcutler I've assigned you to the PR to handle merging it, along with the version notes for it in https://github.com/Submitty/submitty.github.io/pull/407

.setup/apache/submitty.conf
```diff
@@ -40,6 +40,12 @@
         Require all granted
 
         RewriteEngine On
+        
+        # Routing Flexibility: Removes the trailing slash if it is supplied
+        # Example: /courses/f21/csci1200/ -> /courses/f21/csci1200
+        RewriteCond %{REQUEST_FILENAME} !-d
+        RewriteCond %{REQUEST_URI} /(.*)/$
+        RewriteRule ^ /%1 [R=301,L]
```
MasterOdin, line 48: Hm, this actually seems to break git. I'm not sure why it seems git auto-adds a trailing slash, so we may need to special case this so to ignore any URIs that include a `/git/` path?

MasterOdin (changes requested):

.setup/apache/submitty.conf
wusatosi: > Given the redirect that's currently happening, I wonder if we could put in place a patch to `site/public/index.php` to have similar behavior on the redirect?

At that point we should just follow [the guide](https://symfony.com/doc/3.3/routing/redirect_trailing_slash.html) refereed in https://github.com/Submitty/Submitty/issues/7253#issuecomment-977482083 or make a 404 handler if applicable

MasterOdin: That'd also be fine.

wusatosi: okay the guide I've found previously refers to 3.x version of symfony, apparently this [should be enabled](https://symfony.com/doc/current/routing.html#redirecting-urls-with-trailing-slashes) in newer version ([>4.0](https://symfony.com/doc/4.1/routing/redirect_trailing_slash.html)) by default out of the box? (also mentioned in https://github.com/symfony/symfony/issues/39875). I wonder why this doesn't it work for us.

MasterOdin: Weird, now the git tests are passing. I wonder if it was just a temporary hiccup? This PR might be good to go without any further changes, though it still might be better to move this into the PHP code if it's not too much trouble as then don't need to have upgrade instructions for sysadmins...

wusatosi: will implementing this behavior in `WebRouter.php` a cleaner alternative?

MasterOdin: I think so?

PR #7679, opened 2022-02-27
[Bugfix:Developer] Reorder ldap setup in CI

### LINKED ISSUE #7678: LDAP not fully installed in CI environment
**Describe the bug**
ldapadd not installed in Submitty CI pipeline in `e2e` job. 
This results in script (`setup_sample_courses.py`) faulting, [outputing](https://github.com/Submitty/Submitty/runs/5342783900?check_suite_focus=true#step:21:18):
```
sh: 1: ldapadd: not found
sh: 1: ldapadd: not found
sh: 1: ldapadd: not found
...
```

**To Reproduce**
1. See recently triggered e2e pipeline
2. Look at step `Set up sample course`
3. Expand log
4. See fault

**Screenshots**
image

**Additional context**

wusatosi:

Fixes #7678 .
`Set up sample course` faults because `ldapadd` (part of the open-ldap) is not present at the time of execution.

### What is the new behavior?
Reordered `Setup Idap` in front of the `Set up sample course`, so `ldapadd` will be available when setup sample course is invoked.

### Other information?

Do we need to refactor `setup_sample_courses.py` so that it fails when external calls fail?

### DIFF
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -372,6 +372,11 @@ jobs:
               cd $SUBMITTY_REPOSITORY
               sudo -E env "PATH=$PATH" bash .setup/testing/autograder.sh
 
+          - name: Setup ldap
+            run: |
+              cd $SUBMITTY_REPOSITORY
+              sudo -E env "PATH=$PATH" bash .setup/testing/setup_ldap.sh
+
           - name: Configure Test suite
             run: |
               cd $SUBMITTY_REPOSITORY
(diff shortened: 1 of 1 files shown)


shailpatels (approved):
Thanks for catching this, output now looks like the ldapp system call is working in CI now

MasterOdin:
Thanks for this, seems like the E2E test suite has not been testing Ldap (or Database) auth methods, or else we should have caught this was going on. I've fixed the CI suite in #7686 (I think) and once confirmed that this causes the ldap tests to fail as expected, I'll merge this PR, and then merge master onto #7686 which should then pass to complete the loop.

This is a good reason to not use `os.system` in scripts as it doesn't not error on invalid exit codes, and this is my mistake. 😬

PR #7680, opened 2022-02-27
[Bugfix:Forum] Mark current thread as read

### LINKED ISSUE #7275: First post selected in discussion forum does not get marked as read
**Describe the bug**
When first opening the discussion forum page, selecting a post from the list (which currently is in wide mode, taking up the whole screen) will open that post, but will not mark it as read on the left sidebar. Clicking off of the post onto another post also does not mark the post as read. Only clicking on the post again (with or without clicking to another post first) in the left sidebar will mark it as read.

**Expected behavior**
I would expect the post to become marked as read the first time it is clicked on, from the wide list. I'm not a submitty contributor so I don't know if the current behavior is intended or not, but I personally at least find it odd.

**To Reproduce**
1. Open the discussion forum
2. While the list of posts is still taking the entire screen, find a post that is unread
3. Open the unread post, and observe it still being highlighted as if it were unread
4. Switch to a different post, and observe the original post still being highlighted as if it were unread

**Configuration**
 - OS: Windows 10
 - Browser: chrome

**Screenshots**
An example of this in action is shown below

Wide list:

After clicking on post 44:

After then clicking on post 43:

After clicking on post 44 again:

[ editied to remove content from active course ]

HeatMint:
Fixes #7275 
Currently, users who is viewing discussion forum thread via link/URL or href saw the first thread they view highlighted as unread.
submitty1

### What is the new behavior?
Users who are viewing the thread in discussion forum via link/URL or href are now expected to see the corresponding thread marked as read and highlighted correctly.
submitty2

### Other information?
Tested locally.


site/app/templates/forum/ShowForumThreads.twig
```diff
@@ -14,6 +14,12 @@
         $("form").areYouSure();
         loadThreadHandler();
         initSocketClient();
+
+        //delete new_thread Class to mark posts as seen in frontend for users who access posts via href or url
+        thread_id = window.location.pathname.substring(window.location.pathname.lastIndexOf('/') + 1);
```
shailpatels, line 19: Thanks for the PR @HeatMint ,

I think we want to avoid parsing the URL to get the current thread incase this changes in the future. As a suggestion you can access the current selected thread with the following jquery:

```js
$('data#current-thread').val()
```

Once you have the thread id you can just find the element by id using 

```js
$(`#thread_box_link_${thread_id} .thread_box`).removeClass('new_thread');
```

shailpatels (changes requested):

shailpatels (approved):
Looks good, thanks for the contribution!

PR #7695, opened 2022-03-02
[Testing:Developer] General Cypress test improvements

DefinitiveAbove:
Some of the course material tests in Cypress were very flaky, and the `visit` command is missing a return leading to undefined behavior.

### What is the new behavior?
Added return for `visit` command overwrite (https://docs.cypress.io/api/cypress-api/custom-commands#Overwrite-visit-command) - breaks some of the tests requiring some of the fixes listed below.

Added `waitPageChange` helper command - waits for the page to change once to a new page (navigation/refresh) before continuing code execution - used in login/logout and multiple other locations where appropriate to avoid race conditions.

Added `force` onto multiple timestamp-related field entries (tests would fail depending on screen size because the date picker would be covering the typing box).

Added `logout` to afterEach for test cases in general (`visit` fix led to Cypress resetting cookies - signing out most of the time, but not always (some sort of bug having to do when the test before the current test fails, leading to Cypress not signing out)).

Fixes focused around the course materials tests to decrease inconsistencies/race conditions, removed waits/reloads where unneeded after fixes.

### Other information?
Test that doesn't run in CI is failing - `Pre-autograding test` - `Should show newly added autograding jobs` (`autograding_status_1.spec.js`); not sure what the correct numbers are supposed to be since the numbers keep changing on my local instance.


site/cypress/integration/autograding_status_1.spec.js
```diff
@@ -15,7 +15,7 @@ skipOn(Cypress.env('run_area') === 'CI', () => {
         });
 
         afterEach(() => {
-            cy.logout();
+            cy.logout(true);
         });
```
MasterOdin, line 19: Do we even need to logout? Per the cypress docs, it looks like it automatically clears localStorage, cookies, and sessions before each test, which would automatically force a logout. We are left with a bunch of "dangling" sessions between each test, but whatever, that's not really something to worry about.

DefinitiveAbove: Ideally there would be no need to logout, but sometimes it doesn't actually clear everything between test cases (which leads to an error when it tries to login when it's already logged in) - I'm not exactly sure why it happens but I know that it happens sometimes when an XHR request spills over from the previous test case - https://github.com/cypress-io/cypress/issues/216 + https://github.com/cypress-io/cypress/issues/686.

site/cypress/integration/late_days_allowed.spec.js
```diff
@@ -29,21 +40,21 @@ describe('Test cases involving the late days allowed page', () => {
 
         it('check invalid datestamp', () => {
             cy.get('#user_id').type('bitdiddle');
-            cy.get('#datestamp').type("this/isn't/a/date");
+            cy.get('#datestamp').type("this/isn't/a/date", {'force': true});
```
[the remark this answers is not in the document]
DefinitiveAbove: I've been testing using `npx cypress open` (so no automatic video saving) and can't reproduce it right now, but the issue has to do with the `user_id` box right before the datestamp - if it happens to wait long enough on the user id box, an autofill pops up with username suggestions, which can be covering the next field and causes the test to fail.

image

site/cypress/integration/autograding_status_1.spec.js
MasterOdin: Hm, alright. Can you move this `afterEach` to live in `site/cypress/support/index.js` so that it's just defined once (though I don't remember if that'll operate after each test, or after each outer describe...)?

site/cypress/integration/late_days_allowed.spec.js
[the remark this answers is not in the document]
MasterOdin: Ah, okay. I guess we'd want to send a `.blur()` event to the previous input line and that would hopefully equally resolve it?

site/cypress/integration/autograding_status_1.spec.js
DefinitiveAbove: The `api/auth.spec.js` tests don't log in or out of the site using the standard method; I wasn't able to find a way to have a global afterEach that excludes certain test files.

MasterOdin: Given that they don't really ever make use of anything out of cypress, I'd be in-favor of moving them out of cypress then if it would simplify other things.

We can move them into `site/tests/api` and then use regular `jest` and such on running them.

site/cypress/support/index.js
```diff
@@ -16,3 +16,7 @@
 import './commands';
 
 require('@cypress/skip-test/support');
+
+afterEach(() => {
+    cy.logout(true);
+});
```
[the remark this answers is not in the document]
DefinitiveAbove: Leaving tests in a logged-in state works most of the time but can occasionally be flaky; see https://github.com/Submitty/Submitty/pull/7695#discussion_r818181484 for more details.

I added in a list that lets certain tests be excluded from logging out, though it only excludes based on the full path of the current test title and not the file name itself (Cypress.spec reports differently when running all tests versus running each test individually so I don't think there's a way to check the filename without a hacky workaround, and I don't think there will be a case where there will be two tests with the same title paths).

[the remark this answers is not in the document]
MasterOdin: Could we make the logout function have a conditional (activated via parameter) such that it can check if we're logged in first before triggering the action, else do nothing? I think that might be a bit cleaner than using text ignores that could be easy to inadvertently break.

shailpatels (approved):
Thanks for this, reran the e2e tests a couple times and the cypress portion seems stable. 

The manual `login` and `logout` were previously added because there was an issue where the existing user would remain logged in between tests - this is supposed to be prevented by cypress clearing the cookie/session state between each one. Maybe something was fixed in either cypress or github actions or submitty? Either way good to know its working and should save time

bmcutler:
@DefinitiveAbove 
Please resolve merge conflicts

PR #7707, opened 2022-03-06
[Bugfix:Developer] Do not install haskell on non-x86_64

MasterOdin:

We install haskell to help developing analysis tools within Vagrant. Unfortunately, this is only available on x86_64 architecture, and not yet available on running on other architectures like arm64 (necessary for M1).

### What is the new behavior?

This disables downloading and installing haskell / stack on non-x86_64 machines, fixing running `install_system.sh` on M1 vagrant (or any other architecture). While an alternative might be to migrate to GHCup to get this to work, as we look to migrate off AnalysisTools anyway, I don't think it's worth spending the time trying to get the toolchain to work as we look to remove it altogether soon anyway.

### DIFF
--- a/.setup/install_system.sh
+++ b/.setup/install_system.sh
@@ -218,7 +218,8 @@ fi
 # STACK SETUP
 #################
 
-if [ ${VAGRANT} == 1 ] && [ ${WORKER} == 0 ]; then
+# stack is not available for non-x86_64 systems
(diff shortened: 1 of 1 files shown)


MasterOdin:
Vagrant: https://github.com/Submitty/Submitty/actions/runs/1941309538

PR #7708, opened 2022-03-06
[Bugfix:System] Get arm64 sources for docker

MasterOdin:

Right now, we only fetch sources for `amd64` for the docker apt repository. This prevents installing it on arm64 (M1).

### What is the new behavior?

Adds fetching `arm64` to the docker repository we're fetching from, so that the same line works on both amd64 or arm64. By specifying both, there's a few kb more download on fetching the repository, but we avoid doing any sort of complicated logic to make sure we're fetching the right arch name using flags.


MasterOdin:
Vagrant: https://github.com/Submitty/Submitty/actions/runs/1941331418

PR #7711, opened 2022-03-07
[UI/UX:System] Add Close Features to Pop Up

### LINKED ISSUE #7637: Add "X" Button To Settings Panel in Grading Index
Would like an "X" button at the top of the Settings panel in the grading index

Often times I would click on the wrench icon by accident, then need to scroll down to the bottom of the Settings panel before I can close it.

Other times I only need to access the settings close to the top, but still require to scroll down to the bottom to close the panel

Screen Shot 2022-02-13 at 11 08 03 PM

cjreed121:

Currently the only way to close a pop up is to click the close button at the bottom. If the pop up is long you may have to scroll down to see the close button.
Closes #7637

### What is the new behavior?
This PR adds an X at the top right corner of all pop ups. It also adds the ability to click anywhere not in the pop up and it will get closed.

### Pictures
Light Mode:
image

Dark Mode:
image

### DIFF
--- a/site/app/templates/generic/Popup.twig
+++ b/site/app/templates/generic/Popup.twig
@@ -9,10 +9,11 @@
                     <div class="form-title">
                         {% block title_tag %}
                             <h1>{% block title %}Untitled Form{% endblock %}</h1>
-                            <a
+                            <button
                                 onclick="closePopup('{{ block('popup_id') }}')"
                                 class="btn btn-default close-button key_to_click"
-                            >Close</a>
+                                tabindex="-1"
+                            >Close</button>
                         {% endblock %}
                     </div>
                 {% endblock %}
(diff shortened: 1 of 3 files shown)


HenrikLam (commented):
I am wondering if the X on the top right is necessary since all the pop ups already have a close button on the bottom. Perhaps we could remove the button on the bottom instead since the X is always present? Code looks good to me.

shailpatels:
I think this needs a design discussion with the team, personally I'm not a fan of the X since it doesn't really match the rest of the Submitty design - it also adds duplicate functionality 

I think a better way forward would be to have the close button always present in the modal and have it `fixed` to the page, whereas the content is the normal scrollable window

Here is an example mock up, by moving the buttons panel outside of of the `<div class=form-body>`

image


This only issue is that it might be confusing to user since some browsers (maybe just mac) like to hide scrollbars when not scrolling

shailpatels (changes requested):
There's a bit off centering with the new top close button

image

This can probably be fixed if you change the the `a` tag to a `<button class="btn btn-default"`

Elements that do not perform navigation shouldn't be an `a` tag anyways

shailpatels (approved):
Looks good and works as described

Screenshot of example display
image

Issue #7652, opened 2022-02-15
Add SAML 2.0 authentication support


MasterOdin:
RPI is moving towards having SAML be the suggested login mechanism for applications as it allows them to enforce SSO with login requirements to the RPI network. At some point, Submitty will be required to tie into this setup, or risk losing its ability to operate outside the firewall.

RPI uses the [Shibboleth](https://en.wikipedia.org/wiki/Shibboleth_(software)) as its identity provider that we need to integrate with, but the hope is that we can have a fairly generic SAML 2.0 authentication method. 

Some resources:
* https://shibboleth.atlassian.net/wiki/spaces/CONCEPT/overview explains the core concepts of how SAML works, where the SP would be Submitty, and IdP is RPI.
* https://github.com/winstonhong/Shibboleth-SAML-IdP-and-SP is a possible demo app, where we'd look to replace the SP side of things
    * https://github.com/winstonhong/Shibboleth-SAML-IdP-and-SP/blob/e89d06c91c72b2072f546ae0e9acc139583ad41d/LDAP-Dockerized-CentOS/users.ldif can be modified to provide a handful of users to authenticate with for testing for Submitty.
    * https://github.com/CSCfi/shibboleth-idp-dockerized provides a more up-to-date version of shibboleth, not sure how important that is
* Possible PHP libraries:
    * https://www.lightsaml.com/
    * https://github.com/onelogin/php-saml
    * https://simplesamlphp.org/

The workflow would look something like:
* Document setting up a IdP instance in Vagrant
* Get it working on Submitty
    * Create new SamlAuthentication class under https://github.com/Submitty/Submitty/tree/master/site/app/authentication
    * Will we need to create a new "step2" login endpoint such that user hits "login on Submitty" -> "gets redirected to SAML IdP" -> "logins on IdP" -> "redirects back to Submitty"?
    * Will need to document steps of getting SP metadata to provide to IdP
    * Configuration details to library we use?
    * What questions would we need to ask to RPI admins for production (e.g. assertions we can get)? 
* Get details for integrating with RPI from sysadmins
    * Will need to generate metadata.xml from Submitty to give to sysadmins, plus get details on the SAML assertions structure

sama8:
Sharing, if relevant -> [SAML Jackson](https://github.com/boxyhq/jackson)

cjreed121:
[...]

PR #7721, opened 2022-03-08
[Feature:Submission] Add VCS Auth Tokens

cjreed121:

When you do HTTPS Git through a Submitty hosted repository you must log in with your username and password. This will become an issue with #7652 coming in as Submitty won't be able to directly verify passwords in this mode.

### What is the new behavior?
There is a new page where you can create and manage your Git Auth Tokens. Upon creation of a token you are given its value once as it is hashed before reaching the database. If you forget the token you can easily create another and revoke the previous.

Tokens will have an expiration value of either null or a specific date/time. Null allows for the token to be indefinite (Not sure if we want to allow this?).

Once you create an auth token you can go to git and when it asks for your password provide your token instead of your password.

This PR does not end support for password authentication over git rather it adds the option to use tokens.

Screenshots:
image
image

### Other information?
Documentation PR: https://github.com/Submitty/submitty.github.io/pull/425


site/app/controllers/student/GitAuthController.php
```diff
@@ -0,0 +1,91 @@
...
+            $hashed_token,
+            $name,
+            $expiration
+        );
+        $em = $this->core->getSubmittyEntityManager();
+        $em->persist($auth_token);
+        $em->flush();
+        $id = $auth_token->getId();
+
+        $em = $this->core->getSubmittyEntityManager();
```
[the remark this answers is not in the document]
cjreed121: Yeah we definitely can. I'll be sure to remove the second one.

site/app/views/GitAuthView.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace app\views;
+
+use app\entities\GitAuthToken;
+use app\libraries\FileUtils;
+
+class GitAuthView extends AbstractView {
+    public function showGitAuthPage(array $tokens, ?array $token = null) {
```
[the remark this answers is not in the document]
cjreed121: Yeah that makes sense. I'm still not 100% settled on what the UI will be like right now so the structure may change. Before I mark this PR as ready I'll be sure whatever method doesn't have a weird loop like this.

site/app/views/submission/HomeworkView.php
```diff
@@ -505,7 +505,9 @@ public function getRandomIndices(int $array_length, string $student_id, string $
             'allowed_minutes' => $gradeable->getUserAllowedTime($this->core->getUser()),
             'can_student_submit' => $canStudentSubmit,
             'is_grader_view' => false,
-            'recent_version_url' => $recent_version_url
+            'recent_version_url' => $recent_version_url,
+            'git_auth_token_url' => $this->core->buildUrl(['git_auth_tokens']),
+            'git_auth_token_required' => false
```
cjreed121, line 510: This (referring to git_auth_token_required) along with 49 on TeamView.php would be eventually replaced with something like:

`$this->core->getAuthentication() instanceof SamlAuthentication`

Right now there won't be any requirement to use these but with SAML auth these will be necessary.

DefinitiveAbove (changes requested):

site/app/controllers/student/GitAuthController.php
```diff
@@ -0,0 +1,99 @@
...
+            $tokens,
+            $auth_token,
+            $token
+        );
+    }
+
+    /**
+     * @Route("/git_auth_tokens/revoke", methods={"POST"})
+     */
+    public function revokeToken(): RedirectResponse {
```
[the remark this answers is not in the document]
cjreed121: Yes there should. Thanks for catching that.

site/app/templates/GitAuthTokens.twig
```diff
@@ -0,0 +1,63 @@
...
+                {% else %}
+                    {% if token.isExpired(current_time) %}
+                        <td>EXPIRED on {{ token.getExpiration()|date('m/d/Y @ h:i A T') }}</td>
+                    {% else %}
+                        <td>{{ token.getExpiration()|date('m/d/Y @ h:i A T') }}</td>
+                    {% endif %}
+                {% endif %}
+                <td>
+                    <form action="/git_auth_tokens/revoke" method="POST">
+                        <input type="text" hidden name="csrf_token" value="{{ csrf_token }}">
```
shailpatels, line 50: `input type="hidden"` is the value preferred for form inputs that won't be shown to users to prevent browsers from doing weird autofill behavior and screenreaders from trying to speak those inputs

site/app/templates/GitAuthTokens.twig
```diff
@@ -0,0 +1,63 @@
...
+                        <input type="number" hidden name="id" value="{{ token.getId() }}">
```
shailpatels, line 51: Same for this input if it shouldn't be displayed to the user

site/app/templates/GitAuthTokens.twig
```diff
@@ -0,0 +1,63 @@
...
+                        <input type="submit" class="btn btn-danger" value="Revoke">
```
shailpatels, line 52: You would need either an HTML label for this form input or an aria attribute
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints

shailpatels (changes requested):
One small fix there is a contrast issue with the message after creating a token

image

The black on dark blue should be a higher contrast like white text on dark blue background.
You can check this with an accebility checker like WAVE https://submitty.org/developer/software_and_system_design/web_accessibility#examine-your-interface-design-with-wave

Or on firefox you can inspect element on the black text and select its color and will give a warning if the contrainst does not meet web standards.

image

shailpatels (approved):
Install and usage worked as described, didn't take a super close look at the code since it looks like that was done already but looked good overall

Was able to clone a VCS with a token generated and tested on windows. I did see git using windows credential manager for saving the token locally so that should prevent users from having to keep entering the token over and over again.

image

There are some accessibility issues I mentioned but this PR is good to merge if needed sooner and those can be fixed in a separate PR or made into an issue

shailpatels:
As mentioned in last week's meeting but we could consider combing this page with the API tokens page since they both have similar functions and would remove an extra icon on the left for instructors

Instead of two different pages, I think generating an API token can just be the top section in a 'tokens' page since there will only be one, and the user-generated git tokens can be listed underneath them

api
git

PR #7745, opened 2022-03-14
[Bugfix:Submission] Fix submit as student/bulk upload button

DefinitiveAbove:
Using the `Make Submission for a Student` or `Bulk PDF Upload` buttons do not work and throw an error.
image

### What is the new behavior?
Fixes submit as student/bulk upload buttons; issue was caused when `submitty_ocr` feature flag is disabled.

### Other information?
Can test by disabling debug mode and the `submitty_ocr` feature flag (for debug, edit `config/database.json` at `/usr/local/submitty/config/database.json` and set `debugging_enabled` to false); make sure that the elements with ids `toggle-id-scan`/`use-ocr` do not exist on the submission page.

### DIFF
--- a/site/ts/instructor-submission.ts
+++ b/site/ts/instructor-submission.ts
@@ -82,7 +82,7 @@ function changeSubmissionMode(event: Event){
     const qrUploadOpts = document.getElementById('qr-split-opts');
     const numericUploadOpts = document.getElementById('numeric-split-opts');
     const useQRCheckBox = document.getElementById('use-qr') as HTMLInputElement;
-    const useScanIdsCheckBox = document.getElementById('use-ocr') as HTMLInputElement;
+    const useScanIdsCheckBox = document.getElementById('use-ocr') as HTMLInputElement | null;
     const scanIdsOpts = document.getElementById('toggle-id-scan');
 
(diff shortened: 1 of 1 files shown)


shailpatels (approved):
Thanks for making a quick fix for this!

site/ts/instructor-submission.ts
```diff
@@ -40,7 +40,7 @@ function init(){
     const qrPrefixInput = document.getElementById('qr_prefix') as HTMLInputElement;
     const qrSuffixInput = document.getElementById('qr_suffix') as HTMLInputElement;
     const useQRCheckBox = document.getElementById('use-qr') as HTMLInputElement;
-    const useScanIdsCheckBox = document.getElementById('use-ocr') as HTMLInputElement;
+    const useScanIdsCheckBox = document.getElementById('use-ocr') as HTMLInputElement | null;
```
MasterOdin, line 43: This change is essentially the same as removing the type assertion altogether and just having:
```
    const useScanIdsCheckBox = document.getElementById('use-ocr');
```

DefinitiveAbove: document.getElementById provides `HTMLElement | null` - HTMLInputElement provides some additional properties such as the `checked` property so just having `document.getElementById('use-ocr')` throws an error with the checker.

MasterOdin: Fair

PR #7746, opened 2022-03-14
[Bugfix:TAGrading] Fix TA grading notebook bugs

### LINKED ISSUE #7734: Notebook boxes in TA grading initially empty
**Describe the bug**
When I'm in the TA grading interface, viewing the Notebook panel,
For all textboxes (code or text),
The textbox initially looks empty, until I click on each box, then the students text appears.
Without clicking, I cannot tell if the student left it blank, or if the text is just not rendering.

I've reported this before -- and I think it was fixed (apologies if I'm misremembering)
But it is broken in the latest release.  One Mac Chrome at least.

DefinitiveAbove:
* #7734 

### What is the new behavior?
Closes #7734 
Fixes TA grading notebook autoscroll by adding a brief transition period to the scroll.

Attempted to fix issue with empty notebook boxes by adding and enabling the CodeMirror autoreload addon (https://codemirror.net/doc/manual.html#addon_autorefresh) where CodeMirror boxes are enabled.

### Other information?
Was unable to reproduce empty notebook boxes on Chrome/Edge/Firefox on Windows 10; needs testing on Chrome on macOS.

### DIFF
--- a/site/public/js/ta-grading.js
+++ b/site/public/js/ta-grading.js
@@ -160,9 +160,6 @@ $(function () {
       checkNotebookScroll();
     }
   });
-  notebookScrollLoad();
-
-  checkNotebookScroll();
 
   if(localStorage.getItem('notebook-setting-file-submission-expand') == 'true') {
     let notebookPanel = $('#notebook-view');
@@ -188,6 +185,9 @@ $(function () {
   // calling it for the first time i.e initializing
   adjustGradingPanelHeader();
   resizeObserver.observe(document.getElementById('grading-panel-header'));
+
+  notebookScrollLoad();
+  checkNotebookScroll();
 });
 
 function changeStudentArrowTooltips(data) {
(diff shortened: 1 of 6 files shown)


site/public/js/ta-grading.js
```diff
@@ -302,7 +302,7 @@ function notebookScrollLoad() {
     }
     if (element !== null) {
       if (element.length !== 0) {
-        notebookView.scrollTop(element.offset().top - notebookView.offset().top + notebookView.scrollTop());
+        notebookView.animate({scrollTop: (element.offset().top - notebookView.offset().top + notebookView.scrollTop())}, 100);
```
MasterOdin, line 305: Do you know why using `.scrollTop` was not working and `.animate` does? I would think both should work with the only difference being that the latter looks smoother.

cjreed121: I just took a look at this. It seems it might be trying to scrollTop too early so it doesn't work. If you go in the JS console and call the notebookScrollLoad function it actually fixes itself. I was able to get it worked by moving the call to this function down a couple of lines. I'm not too familiar with the TA grading code so I don't know the best place to move the call to but I figure making this get called after the loading is done would probably be better than hoping the animate waits long enough.

cjreed121 (changes requested):
I just tested it and it seems to fix the issue described. See comment above though.

site/public/js/ta-grading.js
DefinitiveAbove: Thanks for checking - I tested some more and it looks like `adjustGradingPanelHeader` was causing the browser to reset the scroll position inside of the notebook view, so the delay was unneeded.

cjreed121 (approved):
Tested on Mac Chrome. Works well and both issues seem to be gone.

PR #7764, opened 2022-03-17
[UI/UX:CourseMaterials] show sections in table format

### LINKED ISSUE #7418: course materials restrict by section popup editor
**Describe the bug**
The sections wrap poorly in this modal/popup:

Screen Shot 2021-12-07 at 1 26 04 AM

There shouldn't be a line break between "Section 6" and the checkbox.
Also, should the checkbox be to the left of the label?  Should try to be consistent across the site.

jeninyg:

Closes #7418

### What is the new behavior?
sections

### DIFF
--- a/site/app/templates/course/EditCourseMaterialsForm.twig
+++ b/site/app/templates/course/EditCourseMaterialsForm.twig
@@ -16,7 +16,7 @@
                 Yes
             </label>
             <br>
-            <span id = "show-some-section-selection-edit" style="display: none">
+            <div id = "show-some-section-selection-edit" style="display: none">
             <br>
             <table>
             {% for section in reg_sections %}
@@ -36,7 +36,7 @@
                 {% endif %}
             {% endfor %}
             </table>
-            </span>
+            </div>
             <script>
                 function toggleRegistrationSectionsEdit(){
                     $("#show-some-section-selection-edit").toggle();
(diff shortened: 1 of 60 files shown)


site/app/templates/course/EditCourseMaterialsForm.twig
```diff
@@ -17,21 +17,25 @@
             </label>
             <br>
             <span id = "show-some-section-selection-edit" style="display: none">
```
shailpatels, line 19: You're getting web validator errors because in HTML a `span` is considered an inline element but a `table` is a block element. If you change this from a `span` to a `div` it should work fine and stop complaining

See more information here
https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements

shailpatels (changes requested):
Make sure to take a look at the docs on the PR title to fix that step of the CI

https://submitty.org/developer/getting_started/how_to_contribute#how-to-make-a-pull-request-pr-to-submitty

site/app/templates/course/EditCourseMaterialsForm.twig
jeninyg: Thanks!

shailpatels (approved):
Looks good, thanks for the PR!

checkboxes now follow columns

shailpatels (changes requested):
The recent changes appear to have introduced a bug where section names with different lengths is breaking the columns

image

shailpatels (approved):
Looks good , columns now line up together and widths support longer section names

image

image

PR #7771, opened 2022-03-20
[Feature:Submission] Notify Late Team Submission

### LINKED ISSUE #7733: Late team submission alerts
When there are team assignments given to students, we allow team members to submit late assignments. We treat these late submissions differently based on the late days and extensions given to each user. Because of this it is possible for a team member to submit a *new* submission that makes another team member have a BAD status for their gradeable. This should be strongly discouraged with an alert and it should notify all affected team members.

In order to encourage students to prevent these submissions, we should add an alert on the HomeowekView page. We should check the LateDayCache to see if there are any team members that will be affected negatively.

N/A

**Additional context**
N/A

jportorreal00:

Team members are allowed to submit new gradeable versions that may potentially leave their team members with a BAD late day status

### What is the new behavior?
Team members that get charged new late days or get left with a bad status on a gradable are now notified.

### Other information?
Closes #7733


ch3nnyj:
I have started trying to review your pr, however when I tried to install your branch there was an error regarding pip version being outdated. Is it possible for you to update it so I can review it? 
Screenshot 2023-01-30 at 4 58 01 PM

jportorreal00:
@ch3nnyj i just pulled from main to get the most recent .setup/pip data. please lmk if that works

ch3nnyj:
@jportorreal00 It still seems to be the same error, "Could not find a version that satisfies the requirement opencv-python==3". For some reason its seems to be the the branch does not have the correct versions associated with opencv-python==3.4.9.33

04vmatsibekker:
I think this code change looks well, but I think it might be good in the future to have to get the team member who doesn't have a late day's permission before the person submits the assignment.

bmcutler:
We'll leave the PR as is, without blocking submission awaiting approval of anyone else.
Submitty will leave team dynamics and course policies to the instructor/teaching staff to handle outside of Submitty and err on the side of collecting all student work (active version can be modified later).

PR #7778, opened 2022-03-23
[Bugfix:Submission] Cancelled submission bugfixes

DefinitiveAbove:
When going to a cancelled submission, the page will reload three times when trying to access "version 0".
Users cannot submit on "version 0" of a cancelled submission.
Cancelling a notebook submission throws errors when viewing "version 0".

### What is the new behavior?
Autograding results are not rendered on "version 0", fixing the page reload issue.
Users can now submit on "version 0".
Viewing "version 0" of a cancelled notebook submission no longer throws errors.


shailpatels (approved):
Looks good, fixes both issues described

PR #7781, opened 2022-03-24
[Bugfix:System] Fix sidebar cookie bug

williamjallen:
Currently, if you: visit a course homepage, visit a subpage of that course, close the sidebar, navigate back to the homepage of the course, open the sidebar, and reload the page, the sidebar will remain closed.

@KCony reported this issue with the sidebar via slack.  The following screen recording is from that bug report.
firefox_tntajgsqxe

### What is the new behavior?
The sidebar is now able to be opened/closed from any page.  This means that it is no longer possible to have page-specific sidebar settings, but it makes the general UI more user-friendly.

Note: I was unable to reproduce the issue after making these changes but due to the finicky nature of cookies, it is possible that these changes are not enough and there may be edge cases where the sidebar still does not behave properly.

### DIFF
--- a/site/public/js/server.js
+++ b/site/public/js/server.js
@@ -1282,13 +1282,12 @@ $.fn.isInViewport = function() {                                        // jQuer
 };
 
 function checkSidebarCollapse() {
-    var size = $(document.body).width();
-    if (size < 1150) {
-        document.cookie = "collapse_sidebar=true;";
+    if ($(document.body).width() < 1150) {
+        document.cookie = "collapse_sidebar=true;path=/";
         $("aside").toggleClass("collapsed", true);
     }
     else{
-        document.cookie = "collapse_sidebar=false;";
+        document.cookie = "collapse_sidebar=false;path=/";
         $("aside").toggleClass("collapsed", false);
     }
 }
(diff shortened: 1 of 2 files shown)


site/public/js/server.js
```diff
@@ -1357,11 +1356,6 @@ $(document).ready(function() {
             }
         }
     });
-
-    //If they make their screen too small, collapse the sidebar to allow more horizontal space
-    $(document.body).resize(function() {
-        checkSidebarCollapse();
-    });
```
MasterOdin, line 1364: To confirm, you're removing this because we have `window.addEventListener("resize", checkSidebarCollapse);` at the top of this file and that the two of them were accomplishing the same thing, and we only need the one. 

I think we might want to move that event listener attachment into the `$(document).ready(function() {` as I think calling `checkSidebarCollapse` before the DOM has loaded (or at least the sidebar of the DOM) will cause an error to be thrown when calling `$("aside").toggleClass` as `$("aside")` will return `null` and then calling `toggleClass` on `null`.

williamjallen: Good comments.  I thought they were accomplishing the same thing but when I removed the event listener at the top of the file, everything broke, while removing the above event listener seemed to have no effect.  I'll move the event listener from the top of the file into the `document.ready()`.

DefinitiveAbove (approved):

PR #7799, opened 2022-04-02
[Bugfix:TAGrading] Fix turning off random order

### LINKED ISSUE #7283: Visiting the grading stats page turns off random order
**Describe the bug**
If random order is turned on, visiting the grading stats page and then returning to the grading index turns random mode off.

**Expected behavior**
Random order mode should remain on when visiting the grading stats page and returning.

**To Reproduce**
1. Visit a gradeable and turn random order on
2. Visit the grading stats page
3. Visit grading index again

poorna2152:

Fixes #7283

### What is the new behavior?
Setting Random order on and visiting to Grading Stat page and going back via Grading details button shows the Random order turned on.


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -53,7 +53,8 @@ public function statusPage(
         string $section_type,
         int $regrade_requests,
         bool $show_warnings,
-        int $submissions_in_queue
+        int $submissions_in_queue,
+        $sort
```
williamjallen, line 57: A type hint for the `$sort` parameter should be added here to match the other parameters.

site/app/templates/grading/electronic/ta_status/StatusBase.twig
```diff
@@ -286,7 +286,14 @@
         }
 
         // change the url so if the user refreshes the page, they stay on the same tab
-        window.history.replaceState("", "", buildCourseUrl(['gradeable', '{{ gradeable_id }}', 'grading', 'status']) + `?nav_tab=${tab}`);
+        let searchParams = new URLSearchParams(window.location.search);
```
williamjallen, line 289:
```suggestion
        const searchParams = new URLSearchParams(window.location.search);
```

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -483,7 +483,7 @@ public function ajaxVerifyComponent($gradeable_id, $verify_all = false) {
      * @AccessControl(role="LIMITED_ACCESS_GRADER")
      * @Route("/courses/{_semester}/{_course}/gradeable/{gradeable_id}/grading/status")
      */
-    public function showStatus($gradeable_id) {
+    public function showStatus($gradeable_id, $sort = null) {
```
williamjallen, line 486: It would be good to add types to both parameters here

williamjallen (approved):
This appears to work as expected so I will approve.  It's possible that something like cookies or local storage could be used here to eliminate the need to mess with the URL at all, similar to the way the anonymous mode toggle works.  Perhaps it's best to merge this as-is and simply address that issue in the future if we run into problems.

site/app/templates/grading/electronic/Details.twig
```diff
@@ -711,6 +711,17 @@
         location.reload();
     }
 
+    function changeSortOrder() {
+        const status = {{ sort == 'random' ? 1 : 0}};
+        if (status) {
```
williamjallen, line 716:
```suggestion
        if ({{ sort == 'random' ? true : false}}) {
```

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -838,7 +838,7 @@ public function showStatus($gradeable_id) {
      * Shows the list of submitters
      * @Route("/courses/{_semester}/{_course}/gradeable/{gradeable_id}/grading/details")
      */
-    public function showDetails($gradeable_id, $sort = "id", $direction = "ASC") {
+    public function showDetails($gradeable_id, $direction = "ASC") {
```
williamjallen, line 841: since you're modifying the line anyway, it might be good to go ahead and add type hints here for the parameters and return value.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -859,6 +859,7 @@ public function showDetails($gradeable_id, $sort = "id", $direction = "ASC") {
             $this->core->redirect($this->core->buildCourseUrl());
         }
         $anon_mode = isset($_COOKIE['anon_mode']) && $_COOKIE['anon_mode'] === 'on';
+        $sort = isset($_COOKIE['sort']) ? $_COOKIE['sort'] : "id";
```
williamjallen, line 862: Just for consistency with the other strings on this line:
```suggestion
        $sort = isset($_COOKIE['sort']) ? $_COOKIE['sort'] : 'id';
```

site/app/templates/grading/electronic/Details.twig
```diff
@@ -42,7 +42,7 @@
                 </a>
                 {% endif %}
                 <a class="btn btn-default"
-                   href="{{ order_toggle_url }}">
+                   href="" onclick="changeSortOrder()">
```
williamjallen, line 45: It would be good to change this to a `<button>`, as per [this](https://github.com/Submitty/Submitty/issues/7819) issue since it no longer acts as a link.

williamjallen (changes requested):
These changes seem to work as intended.  I have a few minor code details I've pointed out below but I will approve after that is done.

site/app/templates/grading/electronic/Details.twig
```diff
@@ -712,8 +712,7 @@
     }
 
     function changeSortOrder() {
-        const status = {{ sort == 'random' ? 1 : 0}};
-        if (status) {
+        if ("{{ sort == 'random' ? true : false }}") {
```
williamjallen, line 715: Is there a reason why you put quotes around it here?  You should just be able to put the JS constants `true` and `false` there instead of relying upon string inference.

williamjallen (changes requested):

site/app/templates/grading/electronic/Details.twig
williamjallen: Interesting.  Given that it messes up the styling, I'd say it's ok to leave as-is.  You can make another PR to style the buttons properly if you want.  If you don't want to do that, I'd at least comment on https://github.com/Submitty/Submitty/issues/7819 and state that they styling is different (and why, if possible) so we can track it in the future if someone wants to work on it.

site/app/templates/grading/electronic/Details.twig
williamjallen: ah, I think the issue is that the Twig constants `true` and `false` are used.  Try this:
`if ({{ sort == 'random' ? "true" : "false" }}) {`

williamjallen (approved):
Changes look good, assuming tests pass

PR #7800, opened 2022-04-02
[Bugfix:TAGrading] Fix broken Home link in TA grading

### LINKED ISSUE #7796: Homepage link in TA grading header broken
**Describe the bug**
When in TA grading, the first link to the homepage is broken.

**To Reproduce**
1. Sign into Submitty as "student".
2. Go to the sample course and select "Gradeables" on the sidebar.
3. Click GRADE on "Closed Peer Team Homework"
4. Grade any team
5. Nothing will happen when you click on the "Submitty" link on the left side of the header.
image

poorna2152:

Closes #7796
### What is the new behavior?
Redirects to the home page

### DIFF
--- a/site/public/js/server.js
+++ b/site/public/js/server.js
@@ -4,6 +4,7 @@ window.addEventListener("resize", checkSidebarCollapse);
 //See this pr for why we might want to remove this code at some point
 //https://github.com/Submitty/Submitty/pull/5071
 window.addEventListener("resize", function(){
+  loadInBreadcrumbLinks();
   adjustBreadcrumbLinks();
 });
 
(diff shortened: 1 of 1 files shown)


shailpatels (changes requested):
It looks like this bug was caused when the `resize` js event is fired before the document is ready

The fix looks good and works after window resizing between mobile and desktop, thanks for the PR !

shailpatels (approved):
It looks like this bug was caused when the `resize` js event is fired before the document is ready

The fix looks good and works after window resizing between mobile and desktop, thanks for the PR !

PR #7804, opened 2022-04-04
[Refactor:Submission] Submission file hiding in UI

DefinitiveAbove:
`hide_submitted_files` is a gradeable config option.

A scanned exam/bulk upload gradeable is separated from a `electronic_hw` gradeable based on the gradeable type set initially.

"Make new submissions and access all prior versions?" under "What are students allowed to do?" for electronic gradeables does not work as expected for student access to prior versions; there is also a gradeable autograding config option which hides the gradeable version as part of it (`hide_version_and_test_details`.

### What is the new behavior?
Student download (can students view/download submitted files?) is added as an option for a gradeable that is configurable in the UI. By default, this is set to the opposite of whether the gradeable was previously a bulk upload/scanned exam option. `hide_submitted_files` is removed as an option and will no longer affect whether gradeables will display submitted files. This also affects previous gradeables.

Scanned exam/bulk upload is removed as a gradeable parameter; in the frontend, the gradeable type instead dynamically changes from a student gradeable (`electronic_hw` - "Students will submit one or more files by direct upload to the Submitty website") to a scanned exam/bulk upload (`electronic_bulk` - "TA/Instructor will (bulk) upload scanned .pdf for online manual grading"); if "View the gradeable on the course home page?" is set to "When grades are released", it will be a bulk upload gradeable; otherwise, it will be a student gradeable.

Fixed already-existing UI version hiding option ("Make new submissions and access all prior versions?") to properly hide other versions from the student, and only show the active version to them if applicable. `hide_version_and_test_details` is renamed to `hide_test_details` and no longer affects whether the version is shown or hidden. This also affects previous gradeables.

### Other information?
Functionality that hid individual images and bulk upload data of PDFs from student view has been removed; future images/bulk upload data will have a prefix of `.` to hide them, but current files will become visible - this was discussed and decided on because the current implementation of hiding/restricting access to certain files for bulk upload gradeables is unclear.

`hide_version_and_test_details` will still hide test details but will no longer hide the version.

eg_scanned_exam under the electronic_gradeable table is being removed - the field wasn't necessarily accurate since instructors would sometimes select one option and then change the uploaded gradeable config to match the other option.

When reverting this PR, eg_scanned_exam in the database is set based on if "View the gradeable on the course home page?" is set to "When grades are released".


shailpatels (changes requested):
I left some comments but biggest issue is with the change of removing `scanned_exam` as a database stored type. I think the new method to base it off of if students can download submissions after grades are released isn't the best method 

If an instructor prevents students from downloading submission until after grading is complete it doesn't necessarily imply a gradeable is a bulk uploaded gradeable. 

I'm not sure if any of this has been discussed but was there an issue for having bulk upload use better defaults/config to let students only download after grades have been released? A scanned exam / bulk upload is technically only an electronic gradeable that has that flag set to true on creation - it should still have all the other settings and configs of a normal electronic gradeable 

I guess the overall question is why does the scanned gradeable flag need to be removed and silently calculated based on configs instead of just explicitly set and have the gradeable use good defaults introduced with the rest of this PR to get the expected behavior of an exam

shailpatels:
Should be noted this PR required rebuilding gradeables

site/app/models/gradeable/Gradeable.php
```diff
@@ -52,14 +52,14 @@
...
  * @method bool isTaGrading()
- * @method bool isScannedExam()
- * @method void setScannedExam($scanned_exam)
  * @method bool isStudentView()
  * @method void setStudentView($can_student_view)
  * @method bool isStudentViewAfterGrades()
  * @method void setStudentViewAfterGrades($can_student_view_after_grades)
  * @method bool isStudentSubmit()
  * @method void setStudentSubmit($can_student_submit)
+ * @method bool canStudentDownload()
```
shailpatels, line 61: I think you would have to manually define this function instead of using the magic getter/setter stuff - or either add it as a special prefix to look out for 

https://github.com/Submitty/Submitty/blob/master/site/app/models/AbstractModel.php#L130

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -157,8 +157,8 @@ private function editPage(Gradeable $gradeable, $semester, $course, $nav_tab = 0
 
         $type_string = 'UNKNOWN';
         if ($gradeable->getType() === GradeableType::ELECTRONIC_FILE) {
-            if ($gradeable->isScannedExam()) {
-                $type_string = self::gradeable_type_strings['electronic_exam'];
+            if ($gradeable->isStudentView() && $gradeable->isStudentViewAfterGrades()) {
```
shailpatels, line 160: I think it would make sense for a function in the gradeable model that returns this so people know what this combination of checks is supposed to represent

something like:
```
isBulkUpload(){
   return $this->isStudentView() && $this->isStudentViewAfterGrades();
}
```

shailpatels (changes requested):

grading/load_config_json.cpp
```diff
@@ -139,11 +139,8 @@ void AddGlobalDefaults(nlohmann::json &whole_config) {
 
   whole_config["required_capabilities"] = whole_config.value("required_capabilities","default");
 
-  // Configure defaults for hide_submitted_files
-  whole_config["hide_submitted_files"] = whole_config.value("hide_submitted_files", false);
-
-  // Configure defaults for hide_version_and_test_details
-  whole_config["hide_version_and_test_details"] = whole_config.value("hide_version_and_test_details", false);
+  // Configure defaults for hide_test_details
+  whole_config["hide_test_details"] = whole_config.value("hide_test_details", whole_config.value("hide_version_and_test_details", false));
```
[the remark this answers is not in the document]
DefinitiveAbove: We discussed this today and decided on keeping it so `hide_version_and_test_details` would act like `hide_test_details` (sinde `hide_version_and_test_details` is likely being used on some active gradeables still).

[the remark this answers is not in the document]
shailpatels: Ok so keeping it for backwards compatibility but expect both values to have the same effect?

site/app/controllers/MiscController.php
```diff
@@ -427,17 +416,7 @@ public function downloadSubmissionZip($gradeable_id, $user_id, $version, $is_ano
...
-                        // we can grab all files regardless of type.
-                        if ($gradeable->isScannedExam()) {
-                            if (mime_content_type($file_path) === 'application/pdf') {
-                                $zip_stream->addFileFromPath($folder_name . "/" . $relative_path, $file_path);
-                            }
-                        }
-                        else {
-                            $zip_stream->addFileFromPath($folder_name . "/" . $relative_path, $file_path);
-                        }
+                        $zip_stream->addFileFromPath($folder_name . "/" . $relative_path, $file_path);
```
williamjallen, line 419: It might be better to use `FileUtils::joinPaths()` here instead of concatenating strings manually.

site/app/controllers/student/SubmissionController.php
```diff
@@ -1888,8 +1888,12 @@ public function showBulkStats($gradeable_id) {
             }
             $files = scandir($user_path);
             $num_files = count($files) - 3;
-            $json_path = $user_path . "/" . $num_files . "/bulk_upload_data.json";
+            $json_path = $user_path . "/" . $num_files . "/.bulk_upload_data.json";
             if (!file_exists($json_path)) {
+                $json_path = $user_path . "/" . $num_files . "/bulk_upload_data.json";
```
williamjallen, line 1893: Same comment here about joining path components

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -219,28 +229,47 @@
     }
 
     function onStudentViewChange() {
+        if ($('#radio_ef_student_upload').is(':checked') || $('#radio_ef_bulk_upload').is(':checked')) {
+            if ($('#yes_student_view_after_grades').is(':checked')) {
+                $('#radio_ef_student_upload').prop('checked', false);
+                $('#radio_ef_bulk_upload').prop('checked', true);
+                let gradeableTypeString = $('#gradeable-type-string');
```
williamjallen, line 236:
```suggestion
                const gradeableTypeString = $('#gradeable-type-string');
```

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -219,28 +229,47 @@
...
+                gradeableTypeString.html(gradeableTypeString.attr('data-bulk'));
+            } else {
+                $('#radio_ef_bulk_upload').prop('checked', false);
+                $('#radio_ef_student_upload').prop('checked', true);
+                let gradeableTypeString = $('#gradeable-type-string');
```
williamjallen, line 241:
```suggestion
                const gradeableTypeString = $('#gradeable-type-string');
```

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -219,28 +229,47 @@
...
             //$('#student_submit_download_view').show();
             disableElementChildren('#student_submit_view', !$('#no_student_view_after_grades').is(':checked'));
+            disableElementChildren('#student_download_view', !$('#no_student_view_after_grades').is(':checked'));
         }
         else {
             // $('#student_submit_download_view').hide();
             disableElementChildren('#student_submit_view', true);
+            disableElementChildren('#student_download_view', true);
             let studentSubmit = $('#no_student_submit');
+            let studentDownload = $('#no_student_download');
```
williamjallen, line 255:
```suggestion
            const studentSubmit = $('#no_student_submit');
            const studentDownload = $('#no_student_download');
```

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -219,28 +229,47 @@
...
 
             $('input[name=student_view_after_grades]').prop('checked', false);
+            $('input[name=student_download]').prop('checked', false);
         }
     }
 
     function onStudentSubmitChange() {
         let cont = $('#due_date_options_container');
-        let exam = {{ is_exam ? "true": "false" }};
+        let bulk_upload = {{ is_bulk_upload ? "true": "false" }};
```
williamjallen, line 269:
```suggestion
        const bulk_upload = {{ is_bulk_upload ? "true": "false" }};
```

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -308,6 +337,9 @@
             let elem = $('#yes_student_view');
             elem.prop('checked', true);
             elem.change();
+            let download = $('#no_student_download');
+            download.prop('checked', true);
+            download.change();
             let sub = $('#no_student_submit');
```
williamjallen, line 343:
```suggestion
            const elem = $('#yes_student_view');
            elem.prop('checked', true);
            elem.change();
            const download = $('#no_student_download');
            download.prop('checked', true);
            download.change();
            const sub = $('#no_student_submit');
```

site/app/templates/admin/admin_gradeable/AdminGradeableCreate.twig
```diff
@@ -89,7 +89,7 @@
             <div class="option-alt"><a target=_blank href="http://submitty.org/instructor/create_edit_gradeable#types-of-gradeables">What is the type of the gradeable <i style="font-style:normal;" class="fa-question-circle"></i></a></div>
             <span class="required_type" >(Required)</span>
         {% else %}
-            <fieldset><legend><b style='font-size:18px'>Gradeable type</b>: <span id="gradeable-type-string">{{ type_string }}</span></legend>
+            <fieldset><legend><b style='font-size:18px'>Gradeable type</b>: <span id="gradeable-type-string" data-hw="{{ gradeable_type_strings['electronic_hw'] | e('html_attr') }}" data-bulk="{{ gradeable_type_strings['electronic_bulk'] | e ('html_attr') }}">{{ type_string }}</span></legend>
```
williamjallen, line 92: Is there supposed to be a space in `e ('html_attr')`?  It's used both with and without the space in this line.

site/app/templates/admin/admin_gradeable/AdminGradeableCreate.twig
```diff
@@ -571,15 +571,15 @@
             $('#radio_electronic_file').prop('checked', true);
             onGradeableTypeChange();
             let vcs = $('input[name=vcs]');
-            let scannedExam = $('input[name=scanned_exam]');
+            let bulkUpload = $('input[name=bulk_upload]');
```
williamjallen, line 574:
```suggestion
            const bulkUpload = $('input[name=bulk_upload]');
```

williamjallen (commented):
Code looks good overall, and would be fine to merge as-is.  I noted a few stylistic changes below, but they do not necessarily have to be made for this to be merged.

williamjallen (approved):
Code looks good.

tkoz0 (approved):
Works as expected with a test bulk upload gradeable.

migration/migrator/migrations/course/20220330154141_update_electronic_gradeable_student_access_options.py
```diff
@@ -0,0 +1,40 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    # database.execute("ALTER TABLE electronic_gradeable ADD IF NOT EXISTS eg_student_download boolean DEFAULT true NOT NULL")
```
cjreed121, line 18: Should we remove the commented code?

site/app/controllers/MiscController.php
```diff
@@ -427,17 +416,7 @@ public function downloadSubmissionZip($gradeable_id, $user_id, $version, $is_ano
...
-                        // we can grab all files regardless of type.
-                        if ($gradeable->isScannedExam()) {
-                            if (mime_content_type($file_path) === 'application/pdf') {
-                                $zip_stream->addFileFromPath($folder_name . "/" . $relative_path, $file_path);
-                            }
-                        }
-                        else {
-                            $zip_stream->addFileFromPath($folder_name . "/" . $relative_path, $file_path);
-                        }
+                        $zip_stream->addFileFromPath(FileUtils::joinPaths($folder_name, $relative_path), $file_path);
```
cjreed121, line 419: Why now download all the images? Can't we still check if it's a bulk gradeable?

more_autograding_examples/notebook_basic/config/config.json
```diff
@@ -1,8 +1,5 @@
 {
 	"required_capabilities": "notebook",
-
```
cjreed121, line 3: Don't we want notebooks to not show submitted files as it may be misleading?

cjreed121 (changes requested):

site/app/controllers/MiscController.php
DefinitiveAbove: Back when this PR was initially made I think this was discussed and we decided to remove this to remove potential confusion with files being silently removed in the backend but given the size of images in the zip it might be good to bring this up to see if we want to change this behavior.

Eceptonsu (commented):
Code looks fine and the changes work as expected.
When "Make new submissions and access all prior versions?" is selected as "No", the "Will this assignment have a due date?" option in the date section disappears. Is this intentional?

PR #7805, opened 2022-04-04
[Bugfix:TAGrading] PDF page assignment bugfixes

### LINKED ISSUE #7735: Specify page bulk upload pdf page correspondence to rubric component is broken
**Describe the bug**
This broke sometime during 2020/2021.  But if we attempt to specify pdf page numbers for rubric components there are lots of buggy popups and the interface is broken.

DefinitiveAbove:
* #7735 

### What is the new behavior?
Closes #7735

Fixes error popup on edit gradeable page when the PDF page number is edited.

Fixes PDF page assignment scrolling (with submission browser + rubric in TA grading).

### Other information?
This can be tested by: 
0. Get a multi-page PDF.
1. Create a gradeable with the type `Students will submit one or more files by direct upload to the Submitty website`
2. Edit the gradeable and go to the rubric tab.
3. Under `And do you expect each specific problem/item/component to appear on a specific page number of the PDF document?`, select `Yes`.
4. Click on the component, and on the right side of the title line enter a valid page number for your multi-page PDF.
5. Add more components and set their page numbers as needed.
6. Submit the PDF to the gradeable and go to grade the PDF (gradeables -> grade -> <find user ID> -> Grade (under TA grading column)
7. If not in multi-panel mode, click on the panel selector: 
image 
and select `side-by-side` under `Two panel options`.
8. Select and open the `Files` panel in one column and the `Rubric` panel in the other column.
9. Open one of the components in the `Rubric` panel by clicking on the component title. The PDF and page should automatically open in the `Files` panel.

If you submit with multiple accounts, switching between them with the rubric and files panels active should automatically open the PDF to the correct page after the rubric component is opened.

### DIFF
--- a/site/public/js/pdf/PDFAnnotateEmbedded.js
+++ b/site/public/js/pdf/PDFAnnotateEmbedded.js
@@ -276,7 +276,7 @@ function render(gradeable_id, user_id, grader_id, file_name, file_path, page_num
                             // scroll to page on load
                             const initialPage = $(`#pageContainer${page_id}`);
                             if (initialPage.length) {
-                                $('#file-content').animate({scrollTop: initialPage[0].offsetTop}, 500);
+                                $('#submission_browser').scrollTop(initialPage[0].offsetTop);
                             }
                         }
                         document.getElementById(`pageContainer${page_id}`).addEventListener('pointerdown', () => {
--- a/site/public/js/ta-grading-rubric.js
+++ b/site/public/js/ta-grading-rubric.js
@@ -2254,7 +2254,7 @@ function setPdfPageAssignment(page) {
         page = 1;
     }
 
-    return closeAllComponents(true)
+    return closeAllComponents(true, true)
         .then(function () {
             return ajaxSaveComponentPages(getGradeableId(), {'page': page});
         })
@@ -2668,11 +2668,11 @@ function scrollToPage(page_num){
             let page = $("#pageContainer" + page_num);
             if($("#file-view").is(":visible")){
                 if(page.length) {
-                    $('#file-content').animate({scrollTop: page[0].offsetTop}, 500);
+                    $('#submission_browser').scrollTop(page[0].offsetTop);
                 }
             }
             else {
-                expandFile("upload.pdf", files[i].getAttribute("file-url"), page_num-1);
+                viewFileFullPanel("upload.pdf", files[i].getAttribute("file-url"), page_num-1);
             }
         }
     }
--- a/site/public/js/ta-grading-rubric.js
+++ b/site/public/js/ta-grading-rubric.js
@@ -2665,6 +2665,7 @@ function scrollToPage(page_num){
     let files = $(".openable-element-submissions");
     for(let i = 0; i < files.length; i++){
         if(files[i].innerText.trim() == "upload.pdf"){
+            page_num = Math.min($("#viewer > .page").length, page_num);
             let page = $("#pageContainer" + page_num);
             if($("#file-view").is(":visible")){
                 if(page.length) {


williamjallen (commented):
Things generally look good here and the feature is already quite broken, so anything is better than what currently exists.  I do have a few comments about the functionality that should probably be addressed eventually, but don't necessarily have to be addressed in this PR.  

1. I'd ideally like to see a little more robustness when it comes to non-bulk-upload submissions.  Unless I was using the feature improperly, I was unable to make the page scroll to the correct location when selecting the option for students to select their own page numbers.
2. When using both page numbering systems, I was unable to make the page scroll to the correct location when multiple different file names were involved.  Ideally, I'd like to see a bit more robustness in how PDF submissions are handled in this regard.
3. When the instructor enters an invalid page number (larger than the PDF document), the page doesn't handle it well.  I'm not sure what the best behavior is here, but perhaps the document should be scrolled to the last page if the page number specified exceeds the total number of pages.
4. Perhaps the page number associated with each component could be added to the grading UI so it is clear to the graders which page the PDF should scroll to.


Additionally, since this is a feature which is prone to breaking in the future, would it be reasonable to add a few simple Cypress tests for this?  I'm not entirely sure how feasible it would be, or if there are gradeables with the appropriate configuration in the sample course, but it would help ensure that this feature doesn't get broken again in the future.

shailpatels (approved):
There is a bit of delay when switching to the next student and auto opening the rubrcic then PDF page but overall was able to get the workflow down with a bulk upload gradeable and auto opening it 

Approving if we want to get this merged quicker and call it an incremental improvement and focus on polishing it up in later PRs

@DefinitiveAbove I think we should update the documentation page since its still using the old TA grading interface I think - making sure instructors use the two pannel setup for this probably should be written down some where if its required

DefinitiveAbove:
> Things generally look good here and the feature is already quite broken, so anything is better than what currently exists. I do have a few comments about the functionality that should probably be addressed eventually, but don't necessarily have to be addressed in this PR.
> 
> 1. I'd ideally like to see a little more robustness when it comes to non-bulk-upload submissions.  Unless I was using the feature improperly, I was unable to make the page scroll to the correct location when selecting the option for students to select their own page numbers.
> 2. When using both page numbering systems, I was unable to make the page scroll to the correct location when multiple different file names were involved.  Ideally, I'd like to see a bit more robustness in how PDF submissions are handled in this regard.
> 3. When the instructor enters an invalid page number (larger than the PDF document), the page doesn't handle it well.  I'm not sure what the best behavior is here, but perhaps the document should be scrolled to the last page if the page number specified exceeds the total number of pages.
> 4. Perhaps the page number associated with each component could be added to the grading UI so it is clear to the graders which page the PDF should scroll to.
> 
> Additionally, since this is a feature which is prone to breaking in the future, would it be reasonable to add a few simple Cypress tests for this? I'm not entirely sure how feasible it would be, or if there are gradeables with the appropriate configuration in the sample course, but it would help ensure that this feature doesn't get broken again in the future.

For 1, it looks like only the storage of student submitted page numbers is implemented - there's no code for actually displaying that. There's a quick fix that can be done with just putting it into graded_component, but I think it might be better to add this in a different PR. 4 is also related to this - page numbers assigned for all submissions do show up on the rubric next to the title of each component, but the student submitted page numbers do not.

For 2, it is currently hardcoded to only accept `upload.pdf` - I think this should be changed to automatically work when there is a single PDF in a submission (especially for student submitted page numbers), but when multiple PDFs are in a submission, I think it shouldn't automatically open and have some sort of warning indicating that there are multiple PDF files.

3 should be fixed with the new changes.

> There is a bit of delay when switching to the next student and auto opening the rubrcic then PDF page but overall was able to get the workflow down with a bulk upload gradeable and auto opening it
> 
> Approving if we want to get this merged quicker and call it an incremental improvement and focus on polishing it up in later PRs
> 
> @DefinitiveAbove I think we should update the documentation page since its still using the old TA grading interface I think - making sure instructors use the two pannel setup for this probably should be written down some where if its required

The delay is because the rubric is loaded through AJAX through multiple parts (loading the rubric itself, and loading the individual component), and then the actual file has to load in the files pane afterwards. 

I think that the current fixes should be merged in and the other things made in a future PR since this fixes the main functionality for instructors; I can look into updating any of the old TA grading interface.

Issue #7474, opened 2021-12-22
column for registration type


bmcutler:
**What problem are you trying to solve with Submitty**
Most students are registered in the course for a grade.
But we also have a small number of students who might be auditing the course (ungraded).
Or students who are still on the registration but have been given the status of withdrawn or late drop (again ungraded).

**Describe the way you'd like to solve this problem**
1. Let's add another column to the users table of the database to handle this.

      field name = 'registration_type'
      type = character varying(255)
      default value = 'graded'
      other values 'audit' or 'withdrawn' or ...  could add other things eventually
      hmm...  this might be convenient to use for teaching staff who are not officially registered but are testing

2. let's display this value in the manage students/graders -> edit student/graders panel
    and allow editing of the value on that panel
    and make sure we can upload files to initialize this data in bulk

3.  It should also be exported through rainbow grades.  Rainbow grades can then use this information for display and statistics.

4.  TA grading statistics might also want to use this information.

**Describe any potential alternatives you'd tried to solve the problem**

Alternately we could have an enum instead of a string.  But that would be less convenient to expand functionality.

bmcutler:
@tkoz0 
The registration type column has been added to the database and is visible on the manage students page.
The final piece is to get this data exported with the other information from "Grade Reports" -> "Generate Grade Summaries" which write a .json file for each student to 
/var/local/submitty/courses/SEMESTER/COURSE/reports/all_grades/USER.json
We'd like to have an additional field "registration_type" : "graded" (for example) 
added to each user file.

PR #7808, opened 2022-04-06
[Feature:InstructorUI] Course Users Registration Type

listeph:

Progress on #7474 

### What is the new behavior?
- Adds a `registration_type` column to course_db's `users` table, which is constrained to values 'graded' (default), 'audit', 'withdrawn', and 'staff'
- Displays new registration type value in Managing Students tab (Instructor UI). Registration type is not displayed for Managing Graders because they would all just be "staff". sample
- Allows editing of registration type on edit student form. If the user's `group` is changed to be a grader or instructor, then the dropdown disappears because the registration type will default to 'staff'.  edit user form
- Enables the bulk-upload of (optional) registration type column in Managing Students. If the column is left empty for a new user, the registration type will default to 'graded'. For the bulk-upload of Managing Graders, users are automatically defaulted to 'staff' registration type.


migration/migrator/migrations/course/20220316135223_add_users_registration_type_column.py
```diff
@@ -0,0 +1,36 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    pass
```
jportorreal00, line 36: may want to drop the row and revert the constraint changes here

site/app/templates/admin/users/StudentList.twig
```diff
@@ -21,14 +21,15 @@
...
                     <th>View Grades</th>
+                    <th>Registration Type</th>
                     <th>Edit Student</th>
                     <th>Delete Student</th>
                 </tr>
             </thead>
             {% for registration, students in sections %}
                 <tbody id="section-{{ registration }}">
                     <tr class="info">
-                        <th class="section-break" colspan="10">Students Enrolled in Registration Section {{ registration }}</td>
```
jportorreal00, line 31: i feel like this value has been changed a few times. Is there any way to set it to a value? Like `{{ header|length }}` or something passed in from the View? Not necessary but could save someone from future grief.

site/app/templates/admin/users/UserForm.twig
```diff
@@ -94,6 +94,14 @@
                 <option value="4">Student</option>
             </select>
         </label>
+        <label for="registration_type" class="full-width">
+            Registration Type:
+            <select name="registration_type" id="registration_type">
+                <option value="graded">graded</option>
```
jportorreal00, line 100: Also nitpicky, but we could possibly refactor this like the other input options to use something like `{% for reg in reg_types %}`

jportorreal00 (approved):

migration/migrator/migrations/course/20220316135223_add_users_registration_type_column.py
listeph: I was thinking about that too but it says in the migration documentation 

> In general, your down migration should only undo the necessary changes to make earlier versions of the Submitty software work. For example, if your migration is adding a column to an existing table in the database, it is probably not necessary or desirable to delete that data in the down function. We don’t want to lose any data since we likely plan to return to this state in the future. Your down function may be empty.

`

And then my `up` function drops the constraint for in case anyone changes it

PR #7811, opened 2022-04-10
[Refactor:Autograding] Make configure.out global

cjreed121:
When building a gradeable, the build_homework_function.sh will compile configure.out from scratch when it has no gradeable specific dependencies. This can take over 30 seconds to compile.

### What is the new behavior?
The compilation is moved to the system install script so it can update when Submitty updates but not be recompiled unnecessarily between gradeable builds. The configure.out will now be stored in /usr/local/submitty/bin/configure.out rather than the gradeable build location.

This significantly reduces the time needed to do setup sample course in CI going from total of about 1 hour 10 min down to about 45 minutes. It also greatly reduces the time to do a vagrant up from about 1 hour 40 min down to about 45 min (see https://github.com/Submitty/Submitty/runs/5828167885?check_suite_focus=true).

### Other information?
Ideally the configure.out would only be recompiled when the grading code changes during an install. Due to the current setup of the install script with fill in variables, this is tough to do checks on. Not sure if anyone has any suggestions to check for changes. Worst case we are looking at adding about 30 seconds to a full install script run but reducing every gradeable build by the same 30 seconds.


shailpatels:
Which specific variables are required to be filled in? Instead of having a `g++` command inside the bash script, you could create a `makefile` to compile the configure.out binary then you could just run `make` to perform a recompile and it would be able to check if any of the dependencies requires a recompile automatically

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -477,13 +477,29 @@ g++ ${SUBMITTY_INSTALL_DIR}/src/grading/system_call_check.cpp -o ${SUBMITTY_INST
 # build the helper program for calculating early submission incentive extensions
 g++ ${SUBMITTY_INSTALL_DIR}/bin/calculate_extensions.cpp -lboost_system -lboost_filesystem -std=c++11 -Wall -g -o ${SUBMITTY_INSTALL_DIR}/bin/calculate_extensions.out
 
+GRADINGCODE=${SUBMITTY_INSTALL_DIR}/src/grading
+JSONCODE=${SUBMITTY_INSTALL_DIR}/vendor/include
+
+# Create the complete/build config using main_configure
+g++ ${GRADINGCODE}/main_configure.cpp ${GRADINGCODE}/load_config_json.cpp ${GRADINGCODE}/execute.cpp \
```
williamjallen, line 484: I'm not at all knowledgeable about this part of the codebase so I don't know how much performance matters, but is there any reason to not add some sort of optimization flag here?

cjreed121: Good point. I did some quick performance testing to see if it would be worth it. Before/after is comparing adding the -O3 flag so before is without any optimizations.

Compilation time:
Before: 26 seconds
After: 42.8 seconds (and it threw a couple of warnings)

Running time (of only configure.out):
Before: 0.006 seconds
After: 0.007 seconds

Note that these aren't averaged so the running time is probably incorrect but the point I'm trying to make is that the values are so close it's probably not worth the extra compilation time unless we were building thousands of gradeables.

bmcutler (approved):

PR #7812, opened 2022-04-11
[Refactor:TAGrading] Panel TypeScript refactors

DefinitiveAbove:
Discussion Forum/Peer/Regrade/Solution TA Notes/Submission panels are in JavaScript.

### What is the new behavior?
Discussion Forum/Peer/Regrade/Solution TA Notes/Submission panels have been migrated to TypeScript.

### DIFF
--- a/site/app/templates/grading/electronic/DiscussionForumPanel.twig
+++ b/site/app/templates/grading/electronic/DiscussionForumPanel.twig
@@ -1,19 +1,9 @@
 {% import _self as self %}
+{{ add_twig_module_js('grading/electronic/DiscussionForumPanel.js') }}
 <div id="discussion_browser" class="draggable rubric_panel" style="display:block;">
-        <div class="draggable_content" style="width:100%">
-        	<div id="posts_list" style="padding:20px;margin-top:10px;max-height: 100%; overflow: initial;" >
-            	{{ discussion_forum_content|raw }}
-            </div>
-        </div>
-    
+	<div class="draggable_content" style="width:100%">
+		<div id="posts_list" style="padding:20px;margin-top:10px;max-height: 100%; overflow: initial;" >
+			{{ discussion_forum_content|raw }}
+		</div>
+	</div>
 </div>
-
-<script>
-
-	
-		$('#posts_list').css('min-width', '-webkit-fill-available');
-		$(function() {
-			generateCodeMirrorBlocks(document);
-		});
-
-</script>
--- a/site/ts/twig/grading/electronic/SubmissionPanel.ts
+++ b/site/ts/twig/grading/electronic/SubmissionPanel.ts
@@ -1,0 +1,88 @@
+import { urlDecode } from '../../../utils/server';
+
+// expand all files in Submissions and Results section
+function openAll(click_class: string, class_modifier: string) {
+    const toClose = $(`#div_viewer_${$(`.${click_class}${class_modifier}`).attr('data-viewer_id')}`).hasClass('open');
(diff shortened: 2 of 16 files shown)


site/ts/twig/grading/electronic/SubmissionPanel.ts
```diff
@@ -0,0 +1,88 @@
...
+    const toClose = $(`#div_viewer_${$(`.${click_class}${class_modifier}`).attr('data-viewer_id')}`).hasClass('open');
+    $('#submission_browser').find(`.${click_class}${class_modifier}`).each(function () {
+        // Check that the file is not a PDF before clicking on it
+        const viewerID = $(this).attr('data-viewer_id');
+        if (($(this).parent().hasClass('file-viewer') && $(`#file_viewer_${viewerID}`).hasClass('shown') === toClose) ||
+            ($(this).parent().hasClass('div-viewer') && $(`#div_viewer_${viewerID}`).hasClass('open') === toClose)) {
+            const innerText = Object.values($(this))[0].innerText;
+            if (innerText.slice(-4) !== '.pdf') {
+                $(this).click();
+            }
```
williamjallen, line 14: Is this the best way to check if it's a PDF and open the element?

site/ts/twig/grading/electronic/SubmissionPanel.ts
```diff
@@ -0,0 +1,88 @@
...
+    $('#file-container .div-viewer .file-viewer-dir').on('click', (event: JQuery.TriggeredEvent) => {
+        const target = $(event.currentTarget).parent();
+        const id = target.attr('data-viewer-id')!;
+
+        openDiv(id);
+        updateCookies();
+    });
+
+    const currentCodeStyle = localStorage.getItem('theme');
+    const currentCodeStyleRadio = (currentCodeStyle == null || currentCodeStyle == 'light') ? 'style_light' : 'style_dark';
```
williamjallen, line 85:
```suggestion
    const currentCodeStyleRadio = (currentCodeStyle === null || currentCodeStyle === 'light') ? 'style_light' : 'style_dark';
```

williamjallen (changes requested):

site/app/templates/grading/electronic/DiscussionForumPanel.twig
```diff
@@ -1,19 +1,9 @@
...
-		$('#posts_list').css('min-width', '-webkit-fill-available');
-		$(function() {
-			generateCodeMirrorBlocks(document);
-		});
-
-</script>
\ No newline at end of file
+	<div class="draggable_content" style="width:100%">
+		<div id="posts_list" style="padding:20px;margin-top:10px;max-height: 100%; overflow: initial;" >
+			{{ discussion_forum_content|raw }}
```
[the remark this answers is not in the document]
DefinitiveAbove: `discussion_forum_content` contains raw HTML that would get parsed by the .twig file - I think it would be better to move the HTML that is generated over to Twig instead, but it might be better for a different PR since it needs a refactor in the PHP code.

site/ts/twig/grading/electronic/SubmissionPanel.ts
DefinitiveAbove: I think the best way to do this would be to additionally pass the MIME type over, but I think that might be better in a separate PR since this PR focuses mainly on updating the Twig/TS files.

williamjallen (commented):
Code looks good, but I haven't had a chance to test it yet.  Just putting this here as an approval in case someone else tests and approves.

bmcutler:
@DefinitiveAbove 
Please resolve conflicts in this older PR

PR #7813, opened 2022-04-11
[Feature:System] Markdown Login Screen Customization

cjreed121:
There is no way to customize the login screen right now.

### What is the new behavior?
A login.md file can be created inside /usr/local/submitty/config which will be rendered on the login screen. By default `# Login` is rendered but the system administrator can choose to customize this however they like.

### Other information?
Documentation PR: https://github.com/Submitty/submitty.github.io/pull/420

### DIFF
--- a/site/app/templates/Authentication.twig
+++ b/site/app/templates/Authentication.twig
@@ -1,5 +1,9 @@
 <div class="content shadow" id="login-box">
-    <h1 id="login-guest">Login</h1>
+    <div id="login-guest">
+        {% include "misc/Markdown.twig" with {
+            "content": login_content
+        } only %}
+    </div>
     <!--
     <em>Note: Registered students have not yet been uploaded to the Submitty course databases.  Please check back later.</em>
     -->
--- a/site/app/views/AuthenticationView.php
+++ b/site/app/views/AuthenticationView.php
@@ -2,6 +2,8 @@
 
 namespace app\views;
 
+use app\libraries\FileUtils;
+
 class AuthenticationView extends AbstractView {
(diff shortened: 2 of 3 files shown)


shailpatels:
Suggestion: Instead of checking if the `login.md` file exists maybe the install site script can automatically generate it and set the correct permissions by default - that way the sysadmins don't need to create the file themselves or set permissions and can instead directly change the file if needed.

cjreed121:
> Suggestion: Instead of checking if the `login.md` file exists maybe the install site script can automatically generate it and set the correct permissions by default - that way the sysadmins don't need to create the file themselves or set permissions and can instead directly change the file if needed.

It was definitely an option I was considering. The footer links, for example, require that a sysadmin manually create the file and set permissions so I kind of just followed the same practice here. In the meeting, I brought it up and it was decided to stick with the sysadmins having to create the file.

shailpatels:
Ah ok that makes sense if there's already a similar feature to this then yeah I agree using the same process here would be best too

shailpatels (approved):
Looks good and was able to follow the documentation to create a custom login header

shailpatels:
Are local images supposed to be supported for this ? I couldn't create an image in the same directory as `login.md` and if it load - looks like users will need to only display ones that are publically available

cjreed121:
> Are local images supposed to be supported for this ? I couldn't create an image in the same directory as `login.md` and get it to load - images loaded via a URL worked fine

The intention is that you would use a full URL. I also think it's best we stay away from loading arbitrary file names from the config folder.

PR #7817, opened 2022-04-15
[Bugfix:Submission] Fix TAs Submission Type Select

cjreed121:
Full access graders / TAs are shown the option to make a submission for another student or bulk upload but choosing those options does not do anything. The code to make those functional is only sent to instructors.

### What is the new behavior?
The code is now also sent to full access graders. In SubmitBox.twig on line 67, it can be seen that being a full access grader controls whether or not you would receive the submission type buttons.

This PR also renames instructor-submission.ts to ta-submission.ts as the functionality is not limited to instructors only anymore.

### DIFF
--- a/site/app/views/submission/HomeworkView.php
+++ b/site/app/views/submission/HomeworkView.php
@@ -45,7 +45,7 @@ class HomeworkView extends AbstractView {
         $is_team_assignment = $gradeable->isTeamAssignment();
 
         if ($this->core->getUser()->accessFullGrading()) {
-            $this->core->getOutput()->addInternalModuleJs('ta-submission.js');
+            $this->core->getOutput()->addInternalModuleJs('grader-submission.js');
         }
 
         // Only show the late banner if the submission has a due date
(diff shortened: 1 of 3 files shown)


shailpatels (approved):
Looks good, 

I think maybe `grader-submission.js` would be a better name for the file to cover what its functionality is

cjreed121:
> Looks good,
> 
> I think maybe `grader-submission.ts` would be a better name for the file to cover what its functionality is

I was considering grader-submission.ts but I was kind of thinking against it because usually when we say grader we are referring to limited access grader, full access grader, and instructor permission levels. This functionality is not given to limited access graders so the name could be misleading. Something like full-access-grader-submission.ts seems way too long so my next thought was just ta-submission.ts. Thoughts?

shailpatels:
I think you run into a similar issue where `ta-submission.ts` implies its specific to TAs but this is shared logic that'll run if an instructor is submitting for a student or if uploading/assigning bulk upload exams

Since its just a filename I'm not too picky - the comment at the top of that file can be updated to better reflect what it's doing

cjreed121:
We decided on `grader-submission.ts` in the meeting so I'm renaming it to that.

PR #7818, opened 2022-04-17
[Bugfix:System] Patch authentication.json

cjreed121:
Running CONFIGURE_SUBMITTY.py on older systems may crash. (Specifically any system created before PR mentioned below that is running up to date code)

PR #7444 contained a migration which moved the authentication method into authentication.json and added an ldap_options field. This migration set the default value of ldap_options to [] when it should have been {} because CONFIGURE_SUBMITTY.py expects ldap_options to be a dictionary rather than a list.

### What is the new behavior?
CONFIGURE_SUBMITTY.py does not crash.

This PR adds a system migration to check if [] is the value of ldap_options in authentication.json and if so, changes it to {}. When CONFIGURE_SUBMITTY.py gets run, the url, uid, and bind values would get filled in.

### Other information?
Another option besides this migration could be to modify the CONFIGURE_SUBMITTY.py script to check for [] and change it to a {} but we would always have to keep that code there for backwards compatability. So I think this PR is the better option.

### DIFF
--- a/migration/migrator/migrations/system/20220417174632_patch_authentication_method.py
+++ b/migration/migrator/migrations/system/20220417174632_patch_authentication_method.py
@@ -12,7 +12,7 @@ def up(config):
     authentication_file = config.config_path / 'authentication.json'
     with authentication_file.open('r+') as auth_file:
         auth_info = json.load(auth_file, object_pairs_hook=OrderedDict)
-        if auth_info['ldap_options'] == []:
+        if not isinstance(auth_info['ldap_options'], dict):
             auth_info['ldap_options'] = {}
             auth_file.seek(0)
             auth_file.truncate()
--- a/migration/migrator/migrations/system/20220417174632_patch_authentication_method.py
+++ b/migration/migrator/migrations/system/20220417174632_patch_authentication_method.py
@@ -1,0 +1,29 @@
+"""Migration for the Submitty system."""
+from collections import OrderedDict
+import json
+
+def up(config):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    authentication_file = config.config_path / 'authentication.json'
+    with authentication_file.open('r+') as auth_file:
+        auth_info = json.load(auth_file, object_pairs_hook=OrderedDict)
+        if auth_info['ldap_options'] == []:
+            auth_info['ldap_options'] = {}
+            auth_file.seek(0)
+            auth_file.truncate()
+            auth_file.write(json.dumps(auth_info, indent=4))
+
+
+def down(config):
+    """
+    Run down migration (rollback).
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    pass


cjreed121:
For reference, this was the error thrown when CONFIGURE_SUBMITTY.py was crashing:
```
  File ".setup/CONFIGURE_SUBMITTY.py", line 319, in <module>
    'url': default_auth_options.get('url', ''),
AttributeError: 'list' object has no attribute 'get'
```

migration/migrator/migrations/system/20220417174632_patch_authentication_method.py
```diff
@@ -0,0 +1,29 @@
...
+import json
+
+def up(config):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    authentication_file = config.config_path / 'authentication.json'
```
shailpatels, line 12: Should this be 
```python
authentication_file = config.config_path + '/authentication.json' 
```

instead?

cjreed121: I'm not really sure to be honest. I had just copied that line from the original migration that created the authentication file as shown: https://github.com/Submitty/Submitty/blob/d78f916ce2419dae839f2a041d7887c6cb2d0c19/migration/migrator/migrations/system/20211214211432_authentication_method.py#L17

I'm not too familiar with Python so I wasn't sure if this was some idiomatic method of creating the path but I saw it worked.

MasterOdin: `config.config_path` is an instance of [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html). It defines an [overload of the slash operator](https://docs.python.org/3/library/pathlib.html#operators) to make it easier to do path joining. Various equivalent methods for this are:

```python
from pathlib import Path

a = Path('foo')

b = a / 'bar' / 'baz'
c = Path(a, 'bar', 'baz')
d = Path.joinpath('bar', 'baz')

assert b == c == d  # foo/bar/baz
```

Trying to do `config.config_path + '/authentication.json'` will throw error: `TypeError: unsupported operand type(s) for +: 'PosixPath' and 'str'`.

migration/migrator/migrations/system/20220417174632_patch_authentication_method.py
```diff
@@ -0,0 +1,29 @@
...
+    with authentication_file.open('r+') as auth_file:
+        auth_info = json.load(auth_file, object_pairs_hook=OrderedDict)
+        if auth_info['ldap_options'] == []:
```
MasterOdin, line 15: Better to do [`isinstance](https://docs.python.org/3/library/functions.html#isinstance) check for type as if `auth['ldap_options']` was defined to a list with elements, then this equality would fail. Additionally, since we want the value to be a `dict` and everything else is wrong, I'd make this more "broad" and just do:

```suggestion
        if not isinstance(auth_info['ldap_options'], dict):
```

MasterOdin (changes requested):

MasterOdin (approved):

Issue #1, opened 2014-02-17
Make a survey


JFrei86:
Get the opinion of the Data Structures students and CS1 Students.
       Focus Opinion (Design, Sophistication of Algorithms, etc...)

PR #7823, opened 2022-04-19
[Feature:CourseMaterials] move files, different display_name views

jeninyg:

Instructors can upload file, but once they do so, there is no way to change file name/file location
### What is the new behavior?
Now instructors can change the file name/location, and display a different file name (display_name) to non-owners of file.

New UI for both file/URL changes
Edit

downloading file will name the file as display_name.


site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -389,29 +389,74 @@ public function ajaxEditCourseMaterialsFiles(bool $flush = true): JsonResponse {
...
+            }
+        }
+
+        if (isset($_POST['display_name'])) {
+            $display_name = $_POST['display_name'];
+            $path = $course_material->getPath();
+            $dirs = explode("/", $path);
+            array_pop($dirs);
+            $path = implode("/", $dirs);
+            if ($course_material->isLink() && $display_name !== $course_material->getDisplayName()) {
```
shailpatels, line 440: Is the display_name changing the location of where the file is getting stored? Is it not possible to just always keep the path the same and just visually change the text to be whatever the display name is? Or is the real filepath supposed to be hidden from users?

site/app/repositories/course/CourseMaterialRepository.php
```diff
@@ -13,4 +13,12 @@ public function getCourseMaterials(): array {
         $dql = 'SELECT c FROM app\entities\course\CourseMaterial c ORDER BY c.priority ASC, c.path ASC';
         return $this->_em->createQuery($dql)->getResult();
     }
+    public function findDisplayName(string $dir, string $display_name): array {
```
shailpatels, line 16: Why does the display name need to be queried  ? If the display name is getting updated to the database then I think you would just need to update the entity so that whenever it gets constructed whatever will be in the DB will populate the PHP object.

shailpatels (changes requested):
Is there a reason why instructors cannot see the displayed name they selected? I think it would only add confusion if students are seeing a filename that instructors do not right away without clicking the edit button on each.

site/app/controllers/course/CourseMaterialsController.php
jeninyg: The display name was to solve the problem where professors wanted to name their files differently than what is being shown to students. So they would see file1-v2 and students would see file1. Display_name used to be called url_title as we didn't want to show the whole link, but a description of the link. Now it serves two purposes. For files, it is the name students/ta/mentors see. For links, it is the link title. I changed the path when a url is being edited here as that was the original behavior. 

This is what editing a link previously looked liked. I changed it so editing a link and file would have same UI. 
urllink

site/app/repositories/course/CourseMaterialRepository.php
jeninyg: The display name gets queried when a professor wants to update the display name of a file. I wanted to make sure there wasn't another file having the same display name in the same directory. However, it doesn't work very well as for a lot of files, the display_name would be empty. This is because of two reasons: 1) files are uploaded before this change. 2) I don't update display name when uploading a file. I didn't think to implement this because it wouldn't fix #1.

Do you think it would be better to get rid of this check, add more constraints in the findDisplayName query, update display_name when new file is uploaded, perform migration...

I don't know if this will be a widely used feature (to have two different names for a file). But it does give professors the ability to rename files and move them to a different folder.

site/app/controllers/course/CourseMaterialsController.php
shailpatels: Right but why would the real location of a URL or file need to be changed if only the display name is getting updated?

site/app/repositories/course/CourseMaterialRepository.php
shailpatels: I don't think there's a need to make sure display names are unique, they can be whatever the user wants since the filepaths are already checked to be unique.

Also why would changing the display name change the file location? If I were a user I would expect my file to be in the same location and the display name shouldn't effect that, I think that would just be confusing.


Again I think that display name should be a purely visual change that doesn't effect the real file location in anyway

site/app/controllers/course/CourseMaterialsController.php
jeninyg: This is only the case for links. I'm not sure what the original design decisions behind this was. For links, the path column would be `/var/local/submitty/courses/s22/sample/uploads/course_materials/link-display_name`, the url column would have the link, and the display_name (originally url_title) column would store the display_name.
table

site/app/repositories/course/CourseMaterialRepository.php
jeninyg: Ok. I just pushed code to remove this check. I also realized that I grab the path when users changed the display name no matter what. I changed it so path is only grabbed if the course material being edited was a link. 

Because we enforce enforce display_name (of a link) to be unique within a folder, changing the path simplifies that. 

This is the original behavior:
https://github.com/Submitty/Submitty/blob/77b1607440127cd9beac61b0e78fedfd1c24f47f/site/app/controllers/course/CourseMaterialsController.php#L392

shailpatels:
On the Wednesday meeting it was discussed the display name should only be a visual change allowing students to see a more clear name then whatever the original file upload is - so I think the logic to move file and URL locations should be removed to simplify things and work with the original use case. 

Refactoring any existing code that deals with moving URLs also would be beneficial to simply some of the course materials logic if possible

site/app/templates/course/EditCourseMaterialsForm.twig
```diff
@@ -3,6 +3,17 @@
 {% block title %}Edit Course Materials{% endblock %}
 {% block body %}
     <div id="material-edit-form" class="flex-col flex-col-space" data-directory = "Check">
+    
+        <label id="file_path-label">
```
shailpatels, line 7: New CSS Ids should use dashes instead of underscores, see 
https://submitty.org/developer/software_and_system_design/coding_style_guide/css

But if this touching older JS / HTML and it would be too hard to change disregard

site/public/js/drag-and-drop.js
```diff
@@ -1146,10 +1146,45 @@ function handleDownloadImages(csrf_token) {
     });
 }
 
+function isValidFileName(file_name) {
```
shailpatels, line 1149: I think this function should be placed under `site/ts/twig/courseMaterials/upload.ts` or something, that way it can be used in other places and it would get the typescript type checking + ability to be tested/linted. 

You can then include it to the twig file via `{{ add_twig_module_js('courseMaterials/upload.js') }}`

Here is an example PR https://github.com/Submitty/Submitty/pull/7649 that uses it, the twig integration with typescript was developed by @DefinitiveAbove so he can also help out if you run into issues

site/public/js/drag-and-drop.js
```diff
@@ -1146,10 +1146,45 @@ function handleDownloadImages(csrf_token) {
...
+        return false;
+    }
+    else if (file_name.indexOf("<") != -1 || file_name.indexOf(">") != -1) {
+        alert("ERROR! You may not use angle brackets in your filename: " + file_name);
+        return false;
+    }
+    return true;
+}
+
+function shouldReplaceFileIfDup(file_name, target_path) {
```
shailpatels, line 1165: Same comment as above with this function

shailpatels (changes requested):
There are some PHP edits where to logic with the optional filepaths have changed? Are those bugfixes related to this PR or do they change some functionality - maybe it would be good to add a description of any of those changes to the PR description

site/app/templates/course/CourseMaterials.twig
```diff
@@ -152,7 +155,7 @@
                 </a>
                 <a onclick="downloadCourseMaterialZip('{{ folder_ids[folder_path] }}')" aria-label="Download the folder: {{ name }}" class="key_to_click" tabindex="0"><i class="fas fa-download" title="Download the folder"></i></a>
                 {% if user_group == 1 %}
-                    <a class="key_to_click" tabindex="0" onclick='newEditCourseMaterialsFolderForm("{{ folder_path }}", "{{ directory_priorities[folder_path] }}")'> <i class="fas fa-pencil-alt black-btn" aria-hidden="true" style="font-size: 16px; margin: 5px;"></i></a>
+                    <a class="key_to_click" tabindex="0" onclick='newEditCourseMaterialsFolderForm("{{ folder_ids[folder_path] }}", "{{ directory_priorities[folder_path] }}")'> <i class="fas fa-pencil-alt black-btn" aria-hidden="true" style="font-size: 16px; margin: 5px;"></i></a>
```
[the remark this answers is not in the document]
04vmatsibekker: It seems this change is important for downloading courseMaterials folder, if I change it back, the folder won't download

PR #7825, opened 2022-04-20
[Feature:Submission] Added Time-Zone Picker Warning Message

### LINKED ISSUE #5501: Datetime pickers time zone warning message
Date time pickers rely on the logged in user to have their system time zone set to match their submitty time zone, otherwise entering times is very confusing.

Since this shouldn't affect many users, we don't need to over engineer a solution.  If a user's system time is different than the server time then we should add a warning message to the date time picker.

annawxng:
Closes #5501
Added a warning message that pops up when the user selects a time-zone differing from their system time-zone. This is to ensure that the rest of the time-zone functionalities, such as entering times, execute properly.

Here is an example of the warning message with the dark theme and light theme:

theme for time zone


shailpatels (changes requested):
Thanks for the contribution @annawxng,

The github diff is marking every file as changed - this usually happens when your editor or git setup changes line endings from linux to windows (LF -> CRLF)

You should use the command `dos2unix` to revert the line endings, or change them via your editor -whichever is faster.

You can also set your git config to prevent auto converting with 
```
git config --global core.autocrlf false
```

shailpatels:
Ohh sorry @annawxng - looking back at this the large changes are because the `node_module` folder is checked in. These dependencies shouldn't be part of the git file system since they get installed by the install scripts locally.

I believe you can do `git rm -r --cached node_modules` to get rid of them and commit the removal to prevent merging them in

annawxng:
> Ohh sorry @annawxng - looking back at this the large changes are because the `node_module` folder is checked in. These dependencies shouldn't be part of the git file system since they get installed by the install scripts locally.
> 
> I believe you can do `git rm -r --cached node_modules` to get rid of them and commit the removal to prevent merging them in

I ran the command and it showed in the terminal that it removed all of the files, but for some reason, the change isn't showing up in the git diff and I cannot commit it. Is there an alternative way to fix this?

shailpatels:
It looks like you did a top level `npm install` or something similar in that case you can just delete the top level `node_modules` directory, the `package.json` and `package-lock.json`

The `node_modules` directory is only installed under `site` so if you want to install JS dependencies you need to under that directory

annawxng:
> It looks like you did a top level `npm install` or something similar in that case you can just delete the top level `node_modules` directory, the `package.json` and `package-lock.json`
> 
> The `node_modules` directory is only installed under `site` so if you want to install JS dependencies you need to under that directory

Sounds good!, just removed them!

site/app/views/UserProfileView.php
```diff
@@ -62,4 +62,4 @@ public function showUserProfile(
             'user_utc_offset' => $user_utc_offset
         ]);
     }
-}
+}
```
shailpatels: The whitespace changes should be removed - I think our linters should fix the newlines at the end of file issue

see 

https://submitty.org/developer/testing/linting_static_analysis#python-linting and 

https://submitty.org/developer/testing/linting_static_analysis#php-linting

site/public/js/user-profile.js
```diff
@@ -39,6 +39,20 @@ function getAvailableTimeZones() {
     return $('#time_zone_selector_label').data('available_time_zones').split(',')
 }
 
+/**
```
shailpatels, line 42: Generally good to leave a link to the source of where a code snippet was from in the doccomments 

It looks like this is from https://stackoverflow.com/questions/9149556/how-to-get-utc-offset-in-javascript-analog-of-timezoneinfo-getutcoffset-in-c

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -12,7 +12,6 @@
 import tzlocal
 import tempfile
```
shailpatels, line 14: Probably want to get rid of the whitespace change here, see 

https://submitty.org/developer/testing/linting_static_analysis#php-linting
and
https://submitty.org/developer/testing/linting_static_analysis#php-linting

shailpatels (changes requested):
Overall looks good, just a small nitpicks and should be ready to merge soon.

If you look under `site/app/tempaltes/GlobalHeader.twig` there's a similar function for displaying messages from the PHP side - it would be nice to get support for the warning messages there too since you're already adding the CSS to support it.
https://github.com/Submitty/Submitty/blob/master/site/app/templates/GlobalHeader.twig#L238-L252

        <div id='messages'>
            
            {% for message in messages %}
                <div id='{{ message.type }}-{{ message.key }}' class="inner-message alert alert-{{ message.type }}">
                    <span>
                        {% if message.type == "error" %}
                        <i class="fas fa-times-circle"></i>
                        {% else %}
                        <i class="fas fa-check-circle"></i>
                        {% endif %}
                        {{ message.error | nl2br }}
                    </span>
                    <a class="fas fa-times key_to_click" tabindex="0" onClick="removeMessagePopup('{{ message.type }}-{{ message.key }}');"></a>
                </div>
            {% endfor %}
        </div>

Second it looks like the notifications on the userprofile page have the success/error icon black/white but the rest of the site have them as the same color as the rest of the notification. I believe this is an issue before your PR but if its something that's quick to fix it would be nice to make consistent 

here's what it looks like when you upload an assignment for example
image

site/public/js/user-profile.js
annawxng: Done!

shailpatels (commented):
The black/white icon not matching the rest of the notifications is an existing issue and can be fixed in a separate issue/PR

shailpatels (approved):
Changes looks good and works as described.

Thank you for the contribution !

PR #7856, opened 2022-04-29
[Bugfix:TAGrading] Fix Grade Button Without Release Date

cjreed121:
When on the gradeables page as a grader, the grade button for an assignment without a release date will still show the database's release date value.
image

### What is the new behavior?
The button will only say grade if the gradeable has no release date.


shailpatels (approved):
Looks good, one small comment is that when a gradeable is in the Future / Beta category the gradeable only says "Submit" instead of "Alpha Submit / Beta Submit" like the rest of the gradeables. But I don't think its a big deal if this should be merged quickly 

image

jeninyg (approved):

PR #7859, opened 2022-04-29
[SYSADMIN ACTION][SECURITY][Bugfix:System] Use UNIX Socket for Websockets

cjreed121:
The PHP websocket server runs on a local port which is then proxied via NGINX.

### What is the new behavior?
The PHP websocket server now runs on a unix socket which gets proxied via NGINX.

### Other information?
Documentation PR: https://github.com/Submitty/submitty.github.io/pull/424


bmcutler:
reviewed offline

PR #7860, opened 2022-04-29
[Bugfix:System] Prevent Websocket Crashes

cjreed121:
Certain messages can be sent to the websocket server that can cause it to crash and restart.

### What is the new behavior?
Additional checks along with a try/catch have been added to prevent websocket crashes.

### Other information?
An example message that could cause a crash is sending a non-string page value.


bmcutler:
Reviewed offline

PR #7863, opened 2022-04-29
[Bugfix:TAGrading] Links to buttons

### LINKED ISSUE #7819: Anchor tags should not be used for non-navigation buttons
**Describe the bug**
A number of places throughout Submitty use anchor tags and are stylized as buttons but do not perform navigation and instead rely on an `onclick` event. This can confuse screen readers which might interpret the element as a link. Browsers may also fire events differently based on if an element is an anchor tag or a `button`

**Expected behavior**
Elements that perform an action and do not change the page should use the `button` tag, the `button` tag also automatically fires `onclick` when pressed via keyboard which would remove the use of the `key_to_click` class being used everywhere 

**Additional context**
https://css-tricks.com/buttons-vs-links/
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#onclick_events

jeninyg:

Help With #7819


shailpatels (approved):
Looks good 

Thanks for this !

shailpatels:
You will need to update the selenium tests to use the new selector - but if you're interested it would be a good learning experience to move the test to cypress

bmcutler:
@jeninyg 
Can you please make the PR message more informative?  
Please explain the purpose of this PR (copy paste from the issue if you like).  
Putting a link to the original issue is not enough.

Also, does this close the issue?  Does it fix all instances of this problem?  Or just some of them?   
(Github seems to be ready to close the issue -- but maybe that is stale from when you wrote "Fixes" and now that it says "helps with" it wont?). - @shailpatels please confirm

shailpatels:
@bmcutler  This PR fixes a couple of instances but not all of them, this section from Mozilla is a good overview of why not to use links as buttons - the original issue also has some extra context

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#onclick_events

bmcutler:
@immortalcodes @shailpatels 
Is PR #7906 duplicating these changes?

Should we close this PR in favor of 7906?
If we don't close it, @immortalcodes, please fix this PR message, etc as discussed above.

shailpatels:
The changes are different this one is only on user profile page, I still need to update the tests to pass with the changes. Was hoping to move them to cypress at the same time but the migration to 10x got in the way

PR #7874, opened 2022-05-05
[Refactor:System] Quote variables in `build_homework_function.sh`

williamjallen:
`build_homework_function.sh` does not quote bash variables properly.  This can result in [unintended consequences](https://unix.stackexchange.com/a/171347/195445) if variables contain spaces, or other special characters.

### What is the new behavior?
Variables in `build_homework_function.sh` have been quoted.

### DIFF
--- a/bin/build_homework_function.sh
+++ b/bin/build_homework_function.sh
@@ -4,10 +4,10 @@
 
 function clean_homework {
     # which assignment to cleanup
-    semester=$1
-    course=$2
-    assignment=$3
-    course_dir=$SUBMITTY_DATA_DIR/courses/$semester/$course
+    semester="$1"
+    course="$2"
+    assignment="$3"
+    course_dir="$SUBMITTY_DATA_DIR/courses/$semester/$course"
 
     echo "clean old build: $course_dir $3"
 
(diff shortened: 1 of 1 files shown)


bin/build_homework_function.sh
```diff
@@ -13,78 +12,77 @@ function clean_homework {
...
-    find $course_dir/custom_validation_code/          -exec chgrp -f ${course_group} {} \;
-    find $course_dir/config/build/            -type d -exec chmod -f ug+rwx,g+s,o= {} \;
-    find $course_dir/config/build/            -type f -exec chmod -f ug+rw,o= {} \;
-    find $course_dir/config/build/            -exec chgrp -f ${course_group} {} \;
-    find $course_dir/config/complete_config/  -type d -exec chmod -f ug+rwx,g+s,o= {} \;
-    find $course_dir/config/complete_config/  -type f -exec chmod -f ug+rw,o= {} \;
-    find $course_dir/config/complete_config/  -exec chgrp -f ${course_group} {} \;
+    find "$course_dir/build/"                   -type d -exec chmod -f ug+rwx,g+s,o= {} \;
+    find "$course_dir/build/"                   -type f -exec chmod -f ug+rw,o= {} \;
+    find "$course_dir/build/"                           -exec chgrp -f ${course_group} {} \;
```
cjreed121, line 61: Might be good to also quote ${course_group}. Also for all other instances in this file.

bin/build_homework_function.sh
```diff
@@ -3,7 +3,6 @@
 ##########################################################################
 
 function clean_homework {
-
     # which assignment to cleanup
     semester=$1
```
cjreed121, line 7: Lines 7-10 should probably get quoted as well.

bin/build_homework_function.sh
```diff
@@ -145,29 +143,29 @@ function build_homework {
...
         echo -e "\nFailed to run cpp preprocessor on ${course_dir}/build/${assignment}/config.json"
         popd > /dev/null
         exit 1
     fi
 
     # Run the complete config json through a python json syntax checker.
-    python3 ${GRADINGCODE}/json_syntax_checker.py complete_config.json
+    python3 "${GRADINGCODE}/json_syntax_checker.py" complete_config.json
     py_res=$?
     if (( $py_res != 0 )); then
```
cjreed121, line 170: Quote here too

bin/build_homework_function.sh
```diff
@@ -176,7 +174,7 @@ function build_homework {
     fi
 
     # Add allowed minutes in database from config if exists
-    python3 ${SUBMITTY_INSTALL_DIR}/bin/set_allowed_mins.py ${hw_build_path}/complete_config.json ${semester} ${course} ${assignment}
+    python3 "${SUBMITTY_INSTALL_DIR}/bin/set_allowed_mins.py" "${hw_build_path}/complete_config.json" "${semester}" "${course}" "${assignment}"
     set_minutes=$?
 
     if ((set_minutes != 0)); then
```
cjreed121, line 180: Quote here too with dollar sign

bin/build_homework_function.sh
```diff
@@ -186,17 +184,17 @@ function build_homework {
...
+        "${GRADINGCODE}/empty_custom_function.cpp" \
         -I${JSONCODE} \
         -pthread -g -std=c++11 -lseccomp -o configure.out
 
 
-    ./configure.out complete_config.json ${course_dir}/config/build/build_${assignment}.json $assignment
+    ./configure.out complete_config.json "${course_dir}/config/build/build_${assignment}.json" "$assignment"
     configure_res=$?
 
     if (( $configure_res != 0 )); then
```
cjreed121, line 200: Quote here too

cjreed121 (changes requested):
Looks like a good change. I commented a couple of places I think you may have missed.

bin/build_homework_function.sh
cjreed121: Is there a reason you didn't do this one?

williamjallen: Are you mentioning the one on (new) line 180?  I'm by no means an expert at shell scripting, but my understanding is that line 180 compares the actual string "set_minutes" to 0, not the value of the variable `$set_minutes`.  Whether or not this line performs the task it was meant to do is a good question that I'm not entirely sure about.

cjreed121: I added this code in https://github.com/Submitty/Submitty/pull/6631. The intention was that the variable set_minutes is compared with 0. This does seem to work although I definitely think it should be corrected. I must have made a typo but it worked as expected while testing. This is what I'm suggesting to change it to:
```suggestion
    if (("$set_minutes" != 0)); then
```

williamjallen: Thanks for the clarification!

cjreed121 (approved):
Looks good to me

PR #7884, opened 2022-05-12
[Refactor:System] Quote variables in create_course.sh

williamjallen:
There are many improperly-quoted variables in Submitty's installation scripts.  This can result in [unintended consequences](https://unix.stackexchange.com/a/171347/195445) if variables contain spaces, or other special characters.

### What is the new behavior?
This PR continues the work begun in #7874 to improve the robustness of Submitty's various shell scripts by properly quoting the variables in `create_course.sh`.

### DIFF
--- a/sbin/create_course.sh
+++ b/sbin/create_course.sh
@@ -39,7 +39,7 @@ if [ -d "${DATABASE_HOST}" ]; then
 fi
 
 # Check that Submitty Master DB exists.
-PGPASSWORD=${DATABASE_PASS} psql ${CONN_STRING} -lqt | cut -d \| -f 1 | grep -qw submitty
+PGPASSWORD="${DATABASE_PASS}" psql ${CONN_STRING} -lqt | cut -d \| -f 1 | grep -qw submitty
 if [[ "$?" -ne "0" ]] ; then
     echo "ERROR: Submitty master database doesn't exist."
     exit
@@ -47,7 +47,7 @@ fi
 
 #Ensure that tables exist within Submitty Master DB.
 sql="SELECT count(*) FROM pg_tables WHERE schemaname='public' AND tablename IN ('terms','courses','courses_users','sessions','users');"
-table_count=`PGPASSWORD=${DATABASE_PASS} psql ${CONN_STRING} -d submitty -tAc "${sql}"`
+table_count=`PGPASSWORD="${DATABASE_PASS}" psql ${CONN_STRING} -d submitty -tAc "${sql}"`
 if [[ "$table_count" -ne "5" ]] ; then
     echo "ERROR: Submitty Master DB is invalid."
     exit
--- a/sbin/create_course.sh
+++ b/sbin/create_course.sh
@@ -14,56 +14,56 @@ fi
 
 CONF_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/../config
 
-SUBMITTY_REPOSITORY_DIR=$(jq -r '.submitty_repository' ${CONF_DIR}/submitty.json)
-SUBMITTY_INSTALL_DIR=$(jq -r '.submitty_install_dir' ${CONF_DIR}/submitty.json)
-SUBMITTY_DATA_DIR=$(jq -r '.submitty_data_dir' ${CONF_DIR}/submitty.json)
-SUBMISSION_URL=$(jq -r '.submission_url' ${CONF_DIR}/submitty.json)
+SUBMITTY_REPOSITORY_DIR=$(jq -r '.submitty_repository' "${CONF_DIR}/submitty.json")
+SUBMITTY_INSTALL_DIR=$(jq -r '.submitty_install_dir' "${CONF_DIR}/submitty.json")
+SUBMITTY_DATA_DIR=$(jq -r '.submitty_data_dir' "${CONF_DIR}/submitty.json")
+SUBMISSION_URL=$(jq -r '.submission_url' "${CONF_DIR}/submitty.json")
 
-PHP_USER=$(jq -r '.php_user' ${CONF_DIR}/submitty_users.json)
-DAEMON_USER=$(jq -r '.daemon_user' ${CONF_DIR}/submitty_users.json)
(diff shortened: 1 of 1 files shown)


sbin/create_course.sh
```diff
@@ -14,56 +14,56 @@ fi
...
 PGPASSWORD=${DATABASE_PASS} psql ${CONN_STRING} -lqt | cut -d \| -f 1 | grep -qw submitty
-if [[ $? -ne "0" ]] ; then
+if [[ "$?" -ne "0" ]] ; then
     echo "ERROR: Submitty master database doesn't exist."
     exit
 fi
 
 #Ensure that tables exist within Submitty Master DB.
 sql="SELECT count(*) FROM pg_tables WHERE schemaname='public' AND tablename IN ('terms','courses','courses_users','sessions','users');"
 table_count=`PGPASSWORD=${DATABASE_PASS} psql ${CONN_STRING} -d submitty -tAc "${sql}"`
```
Viyerelu23333, line 50: Consider using `$(...)` notation ([SC2006](https://www.shellcheck.net/wiki/SC2006)) and quoting the `${DATABASE_PASS}` ([SC2086](https://www.shellcheck.net/wiki/SC2086))?

```suggestion
table_count=$(PGPASSWORD="${DATABASE_PASS}" psql ${CONN_STRING} -d submitty -tAc "${sql}")
```

sbin/create_course.sh
```diff
@@ -14,56 +14,56 @@ fi
...
 ########################################################################################################################
 
 CONN_STRING="-h ${DATABASE_HOST} -U ${DATABASE_USER}"
-if [ -d ${DATABASE_HOST} ]; then
+if [ -d "${DATABASE_HOST}" ]; then
     CONN_EXTRA="${CONN_STRING} -p ${DATABASE_PORT}"
 fi
 
 # Check that Submitty Master DB exists.
 PGPASSWORD=${DATABASE_PASS} psql ${CONN_STRING} -lqt | cut -d \| -f 1 | grep -qw submitty
```
Viyerelu23333, line 42: Considering quoting the `${DATABASE_PASS}` ([SC2086](https://www.shellcheck.net/wiki/SC2086))?

```suggestion
PGPASSWORD="${DATABASE_PASS}" psql ${CONN_STRING} -lqt | cut -d \| -f 1 | grep -qw submitty
```

Viyerelu23333, line 42: Consider quoting the `${DATABASE_PASS}` ([SC2086](https://www.shellcheck.net/wiki/SC2086))?

```suggestion
PGPASSWORD="${DATABASE_PASS}" psql ${CONN_STRING} -lqt | cut -d \| -f 1 | grep -qw submitty
```

sbin/create_course.sh
```diff
@@ -14,56 +14,56 @@ fi
...
+DATABASE_USER=$(jq -r '.database_user' "${CONF_DIR}/database.json")
+DATABASE_PASS=$(jq -r '.database_password' "${CONF_DIR}/database.json")
 
 ########################################################################################################################
 ########################################################################################################################
 
 CONN_STRING="-h ${DATABASE_HOST} -U ${DATABASE_USER}"
-if [ -d ${DATABASE_HOST} ]; then
+if [ -d "${DATABASE_HOST}" ]; then
     CONN_EXTRA="${CONN_STRING} -p ${DATABASE_PORT}"
```
Viyerelu23333, line 38: `${CONN_EXTRA}` is never used...

Viyerelu23333, line 38: `${CONN_EXTRA}` is never used...

Viyerelu23333 (commented):
Actually, tests with `[[ cond ]]` [will not expand the filenames](https://tldp.org/LDP/abs/html/testconstructs.html#AEN3206) when using `bash`.  But it works either way.  

Looks good to me.

sbin/create_course.sh
williamjallen: If you see the previous commit, I actually had to unquote `DATABASE_PASS` since it's meant to be a number of CLI arguments.  It's a really really hacky solution, but I wanted to minimize the changes and I don't believe there are any security risks coming from it.  

As far as the SC2006 goes, I'm planning to go through and make another PR to fix other things like that.  There are a few tricky changes that are necessary in some cases (including for other errors) so I want to get the unquoted variables out of the way before beginning that PR.

sbin/create_course.sh
williamjallen: Same comment about quoting `DATABASE_PASS` here

sbin/create_course.sh
williamjallen: Same as with the instances of SC2006 in this file, I'll come back through with a second PR (hopefully when I add shellcheck linting) to fix more complicated things like that.  I think `CONN_EXTRA` is safe to remove here, but it can be risky to remove things like that without being very thorough in reading the code.  I specifically wanted to be strict in only quoting variables in this PR, making only the most minimal edits possible to minimize the risk of breakages.

I'm not opposed to the idea of going ahead and fully linting the entire file, but it can be a little more risky.  Thoughts on this?

sbin/create_course.sh
Viyerelu23333: Yeah, I saw your last commit.  I think you are referring to `CONN_STRING` rather than `DATABASE_PASS`...  The definitions of the variables are:

- `DATABASE_PASS`: [L31](../blob/main/sbin/create_course.sh#L31)
  `DATABASE_PASS=$(jq -r '.database_password' "${CONF_DIR}/database.json")`
- `CONN_STRING`: [L36](../blob/main/sbin/create_course.sh#L36)
  `CONN_STRING="-h ${DATABASE_HOST} -U ${DATABASE_USER}"`

`CONN_STRING` does have some parameters that will pass into `psql`... Quoting it will definitively introduce problems.  

<hr/>
All SC2006, SC2115 related comments are marked as resolved.

sbin/create_course.sh
Viyerelu23333: Actually, nowhere in this repository reads `CONN_EXTRA`... 

Maybe it could be safely deleted in your second PR if the CI passes?

Viyerelu23333 (approved):

PR #7888, opened 2022-05-14
[Bugfix:InstructorUI] Fixed Hard Coded colors

### LINKED ISSUE #7568: Colors used for numeric gradeables are hard coded
**Describe the bug**
Within the 'simple-grading.js' file there are a bunch of uses of hard coded colors for numeric gradeables instead of using 'colors.css'. This mainly causes problems with the dark mode version of the site. When a number is entered, and the field is updated the background color of the cell will change to the light mode version color.

The uses in 'simple-grading.js' I found were on line numbers: 420, 466, 557, 560, 564, 905, 907

immortalcodes:
Fixes #7568  : Colors are hard coded  for numeric gradeables instead of using 'colors.css'. This mainly causes problems with the dark mode version of the site. 

### What is the new behavior?
Standard names for the colors are used as described in colors.css instead of hashcodes , also the color '#ff7777'(not available in colors.css) is replaced with '#ff9999'(available in colors.css).

### DIFF
--- a/site/public/js/simple-grading.js
+++ b/site/public/js/simple-grading.js
@@ -947,7 +947,7 @@ function numericSocketHandler(elem_id, anon_id, value, total) {
     elem.data('origval', value);
     elem.attr('data-origval', value);
     elem.val(value);
-    elem.css("background-color", "white");
+    elem.css("background-color", "--always-default-white");
     if(value == 0) {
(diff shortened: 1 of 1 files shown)


immortalcodes:
Is there something on my part that needs to be done?

PR #7889, opened 2022-05-14
[Feature:Plagiarism] Add color customization

williamjallen:
The current color scheme for the plagiarism interface has poor contrast and is somewhat distracting when trying to compare two submissions.

Current interface:
before

### What is the new behavior?
The default colors have been adjusted to improve the contrast.  Additionally, the ability to customize the colors has been added.  Colors are chosen using a standard HTML5 color picker, and are saved in local storage for use across multiple sessions and configurations.

New interface with default colors:
after

Highlighting key with customizable colors:
plag_highlighting_key


shailpatels (changes requested):
I think you want to use the `input` event instead of onchange for color pickers, for color pickers onchange is inconsistent, especially on mac which opens up the native color picker and does not close it after a selection. Also if a user sets a color and then immediately closes the model the color might not get saved or set correctly.

Eceptonsu (changes requested):
The "reset color" button updated the colors for the first time after I changed the colors, but when I did it for the second time, the update is not instant, a refresh is needed at this point.

williamjallen:
Good catch.  Changes have been made.

Eceptonsu (approved):

PR #7894, opened 2022-05-17
[Bugfix:TAGrading] Lab navigation for checkpoint fix

### LINKED ISSUE #7761: Arrow key navigation functionality broken for checkpoint gradeables
After a checkpoint mark box is clicked, if the user tries to navigate vertically with the arrow keys, the selection will move 2 boxes at once.  The problem is located in the simple-grading.js file. Something to do with code needed for moving both the focus and cursor within the numeric gradable cell. When both of these are moved for checkpoints then we shift 2 cells at once.

Eceptonsu:
Fixes #7761
If the user tries to navigate vertically with the arrow keys through the lab checkpoints, the selection will move 2 cells at once. The arrow keys also do not work in the null section.

### What is the new behavior?
The arrow keys now correctly move 1 cell at a time when the user wants to navigate vertically. They also function correctly in null sections.

https://user-images.githubusercontent.com/79226246/168719906-8704f0c1-415f-4a53-a6d1-2bb32955a3cc.mp4

### DIFF
--- a/site/public/js/simple-grading.js
+++ b/site/public/js/simple-grading.js
@@ -717,6 +717,11 @@ function setupSimpleGrading(action) {
                     }
 
                     if (direction == "down") {
+                        // Check if cell needs to move beyond null section
+                        if (new_selector_array[1] === "") {
+                            break;
+                        }
+
                         // Check if cell needs to move to next section
                         new_selector_array[1] += 1;
                         new_selector_array[2] = 0;
--- a/site/public/js/simple-grading.js
+++ b/site/public/js/simple-grading.js
@@ -648,7 +648,10 @@ function setupSimpleGrading(action) {
         if(prev_cell.length) {
             // ids have the format cell-#SECTION-ROW#-COL#
             var new_selector_array = prev_cell.attr("id").split("-");
-            new_selector_array[1] = parseInt(new_selector_array[1]);
+
+            if (new_selector_array[1].length) {
+                new_selector_array[1] = parseInt(new_selector_array[1]);
+            }
             new_selector_array[2] = parseInt(new_selector_array[2]);
             new_selector_array[3] = parseInt(new_selector_array[3]);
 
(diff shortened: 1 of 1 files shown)


site/public/js/simple-grading.js
```diff
@@ -658,10 +661,19 @@ function setupSimpleGrading(action) {
             else if (direction == "left") new_selector_array[3] -= 1;
             else if (direction == "right") new_selector_array[3] += 1;
 
-            // TODO: cannot move to NULL section (all cells with id 'cell--#-#')
             if (new_selector_array[2] < 0 && direction == "up") {
-                // // Selection needs to move to above section, if one exists
                 new_selector_array[2] += 1;
+                // // Selection needs to move to above the null section
+                if (new_selector_array[1] == "") {
```
williamjallen, line 667: It is generally better to use `===` instead of `==` in JS to prevent type inferences from causing problems
```suggestion
                if (new_selector_array[1] === "") {
```

site/public/js/simple-grading.js
```diff
@@ -658,10 +661,19 @@ function setupSimpleGrading(action) {
...
+                    new_selector_array[1] = 1;
+                    while (true) {
+                        let temp_cell = $("#" + new_selector_array.join("-"));
```
williamjallen, line 670: `const` should be used instead of `let` unless you are reassigning the variable. (`var` should be avoided altogether)
```suggestion
                        const temp_cell = $("#" + new_selector_array.join("-"));
```

site/public/js/simple-grading.js
```diff
@@ -708,7 +720,9 @@ function setupSimpleGrading(action) {
                         // Check if cell needs to move to next section
                         new_selector_array[1] += 1;
                         new_selector_array[2] = 0;
-                        new_selector_array[3] = 0;
+
+                        // Check if cell needs to move to null section
+                        if (tries == 1 && new_selector_array[1] != "") {new_selector_array[1] = "";}
```
williamjallen, line 725: Same thing here about `===` and `!==`
```suggestion
                        if (tries= == 1 && new_selector_array[1] !== "") {new_selector_array[1] = "";}
```

williamjallen: additionally, it would be more readable to format this such that it isn't on one line.

williamjallen (commented):
After reaching the bottom, the focus seems to wrap around to the top again when scrolling down.  The reverse is not true though, so if you are at the top and try to scroll up, it does not go to the bottom.  Should this behavior be standardized in both directions?  Personally, I think it shouldn't wrap in either direction, but if someone proposed a good use case for it, I wouldn't have a problem with it.

williamjallen (approved):

PR #7900, opened 2022-05-17
[Feature:System] Update OS Versions after Installation

### LINKED ISSUE #7744: submitty_install does not update docker log file
**Describe the bug**
Doing vagrant up or submitty_install does not write to the docker log files  and currently, the only way to write to the log is to push the update button. This is because the log is written by the script used by the button.

**Expected behavior**
A log file should be written whenever we interact with the script to pull docker images.

**To Reproduce**
1.  Do vagrant up or delete the logs from `/var/local/submitty/logs` and then do submitty_install clean
2.  See that there is no docker log in either the DockerUI page or in `/var/local/submitty/logs/docker`

**Configuration**

**Screenshots**

**Additional context**

Viyerelu23333:

Closes #7744 

Currently, instructors need to manually click “[Update dockers and machines](https://submitty.cs.rpi.edu/admin/update_docker)” button to retrieve the Docker/OS versions.  

### What is the new behavior?

[DockerUI page](https://submitty.cs.rpi.edu/admin/docker) could display the versions after installations and updates.  

### Other information?

This PR changes the [installation script](files).  At the end of the installation/update, it will dispatch a daemon job to update all versions (identical to pushing the update button on the [DockerUI page](https://submitty.cs.rpi.edu/admin/docker)).  

Not sure if this is a good implementation: according to the [original issue](../issues/7744), it should be updated every time the [`update_and_install_workers.py`](../blob/master/sbin/shipper_utils/update_and_install_workers.py) is called.  So far, it is only called by the [installation script](../blob/master/.setup/INSTALL_SUBMITTY_HELPER.sh#L925) and the [DockerUI controller](../blob/master/site/app/controllers/DockerInterfaceController.php#L170).

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -933,4 +933,10 @@ else
         python3 -u ${SUBMITTY_INSTALL_DIR}/sbin/restart_shipper_and_all_workers.py
         echo -e -n "Done restarting shipper & workers\n\n"
     fi
+
+    # Update OS / Docker versions for all machines
+    echo "Update OS and Docker versions for all machines"
+    DATE_YMD=`date +%Y%m%d`
+    echo -e '{\n  "job": "UpdateDockerImages"\n}' > ${SUBMITTY_DATA_DIR}/daemon_job_queue/docker${DATE_YMD}.json
+    echo "Dispatched version updating job"
 fi
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -936,7 +936,7 @@ else
 
     # Update OS / Docker versions for all machines
     echo "Update OS and Docker versions for all machines"
-    DATE_YMD=`date +%Y%m%d`
+    DATE_YMD=$(date +%Y%m%d)
     echo -e '{\n  "job": "UpdateDockerImages"\n}' > "${SUBMITTY_DATA_DIR}/daemon_job_queue/docker${DATE_YMD}.json"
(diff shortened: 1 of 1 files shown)


.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -933,4 +933,10 @@ else
         python3 -u ${SUBMITTY_INSTALL_DIR}/sbin/restart_shipper_and_all_workers.py
         echo -e -n "Done restarting shipper & workers\n\n"
     fi
+
+    # Update OS / Docker versions for all machines
+    echo "Update OS and Docker versions for all machines"
+    DATE_YMD=`date +%Y%m%d`
+    echo -e '{\n  "job": "UpdateDockerImages"\n}' > ${SUBMITTY_DATA_DIR}/daemon_job_queue/docker${DATE_YMD}.json
```
williamjallen, line 940: It is good practice to quote variables in shell scripts.  It's probably not too big of a risk here, but it can cause a bunch of [unexpected problems](https://unix.stackexchange.com/a/171347/195445) if there are spaces or other special characters in the variable.
```suggestion
    echo -e '{\n  "job": "UpdateDockerImages"\n}' > "${SUBMITTY_DATA_DIR}/daemon_job_queue/docker${DATE_YMD}.json"
```

williamjallen (changes requested):

williamjallen (changes requested):

williamjallen (commented):
Code looks good but I'll leave the final approval to Barb or someone else who knows what the intended functionality is.

Viyerelu23333:
A better implementation is in #7932

PR #7902, opened 2022-05-18
[Bugfix:Submission] Bulk submission options

### LINKED ISSUE #7881: Bulk upload expected page count
**Describe the bug**
The expected page count field is shown when split by QR mode is turned off.
image

**Expected behavior**
The expected page count field should only be shown when QR mode is turned on. When splitting by page number the page count would always be correct or the split wouldn't work correctly. It only makes sense to check the page count when doing it by QR code.

**To Reproduce**
1. Go to submit any gradeable in the sample course as an instructor or full access grader
2. See expected page count input box when QR is turned off
3. Turn QR code on and see that expected page count disappears

Eceptonsu:
Fixes #7881
The expected page count input field is only shown when split by QR mode is turned off.
The use_ocr option will not be remembered like other options, the check boxes' behaviors are not consistent.

### What is the new behavior?
The expected page count input field is shown when only  QR mode is turned on.
The use_ocr option will be remembered after the submission option changes and the page refreshes.


cjreed121 (changes requested):
I feel like a much simpler solution would be to move the expected-num-box div in SubmitBox.twig into the qr-split-opts div rather than having to change all of display style settings. Was there a reason you chose this method?

site/ts/grader-submission.ts
```diff
@@ -68,6 +69,10 @@ function init(){
     if (prevQRSuffix){
         qrSuffixInput.value = prevQRSuffix;
     }
+
+    if (prevScanSetting){
+        useScanIdsCheckBox!.checked = prevScanSetting === 'true';
+    }
```
cjreed121, line 75: I think I understand what this is doing but it's not clear from looking at the PR description. The PR description makes it seems like you are only fixing QR codes when you are doing more.

cjreed121 (changes requested):
Make sure to also update the Cypress test so the GitHub actions e2e test passes

site/ts/grader-submission.ts
shailpatels: It looks like the "scan user IDs" checkbox wasn't being loaded if already checked and switching between modes - but I agree would be useful to have in PR description to make sure others know full scope of changes

shailpatels (approved):
Fixes issue and works as described.

Thanks!

cjreed121 (approved):

PR #7903, opened 2022-05-18
[SECURITY][Bugfix:System] Prevent generate_repos injection

cjreed121:
Using os.system in the generate repos script with variable fill ins could lead to possible command injection.

### What is the new behavior?
subprocess.run is used instead with an array of parameters so command injection is not possible.

### DIFF
--- a/bin/generate_repos.py
+++ b/bin/generate_repos.py
@@ -62,7 +62,7 @@ def add_empty_commit(folder,which_branch):
             '-c', 'user.name=submitty',
             '-c', 'user.email=submitty@example.com',
             'commit', '--allow-empty',
-            '-m initial empty commit',
+            '-m', 'initial empty commit',
             '--author=submitty <submitty@example.com>'])
         subprocess.run(['git', 'push', 'origin', which_branch])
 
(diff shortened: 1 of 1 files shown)


bin/generate_repos.py
```diff
@@ -55,14 +55,16 @@ def add_empty_commit(folder,which_branch):
...
-                  "-c user.name=submitty -c user.email=submitty@example.com commit " +
-                  "--allow-empty -m 'initial empty commit' " +
-                  "--author='submitty <submitty@example.com>'")
-        os.system(f'git push origin {which_branch}')
+        subprocess.run(['git', 'checkout', '-b', which_branch])
+        subprocess.run(['git',
+            '-c', 'user.name=submitty',
+            '-c', 'user.email=submitty@example.com',
+            'commit', '--allow-empty',
+            '-m initial empty commit',
```
williamjallen, line 65: Should this not be two different arguments?

```suggestion
            '-m', 'initial empty commit',
```

cjreed121: Both ways work. I had changed it to the current way to resolve a bug I thought it was causing but it seems I can change it back.

PR #7905, opened 2022-05-18
[Bugfix:System] Add Access Control to 2 Routes

cjreed121:
There was 2 routes that did not have proper access control.

### What is the new behavior?
These routes now make sure the user is authorized to use those routes before running the rest of the controller's code.

### DIFF
--- a/site/app/controllers/DockerInterfaceController.php
+++ b/site/app/controllers/DockerInterfaceController.php
@@ -153,9 +153,16 @@ class DockerInterfaceController extends AbstractController {
 
     /**
      * @Route("/admin/update_docker", methods={"GET"})
-     * @return JsonResponse
+     * @return JsonResponse | MultiResponse
      */
     public function updateDockerCall() {
+        $user = $this->core->getUser();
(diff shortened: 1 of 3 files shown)


williamjallen (approved):

cjreed121:
@shailpatels Since we are only checking faculty we can check that in the top level pages using annotations. We of course can't check instructor, full access, etc in top level though. To show an example my PR https://github.com/Submitty/Submitty/pull/7822 does this. The reason I didn't go ahead and put that here is because the rest of the file does it in this format and I was just making a quick fix rather than a refactor. I'd say we should probably make a separate PR to convert all cases like this and fix to AccessControl annotation.

shailpatels:
Oh ok, at one point there was a bug where top level routes without a course and semester wouldn't be able to get access checked, I think that was fixed with the superuser account introduction but the docker UI was written before that and that's why the functions are manually checking the account without the annotation. 

I don't think it would be a refactor? Wouldn't it just be adding the top-level annotation + `use` and deleting the manual checks?

PR #7906, opened 2022-05-18
[Bugfix:InstructorUI] Correction of Tags

immortalcodes:
### What is the current behaviour?
Implements  #7819 : Anchor tags were being used instead of button tags by relying on onclick event which created confusions in interpretation by screen readers and different browsers. 

### What is the new behavior?
All the instances of anchor tag which used onclick event have been updated with button tag in the entire codebase.


immortalcodes:
@shailpatels , what could I do to correct the checks.

site/app/templates/autograding/Attachments.twig
```diff
@@ -5,10 +5,10 @@
...
             {{ file.name }}</a> &nbsp;
         <a id = 'open_file_{{ file.name|url_encode }}' onclick='popOutSubmittedFile("{{ file.name|url_encode }}", "{{ file.path|url_encode }}")' aria-label="Pop up the file in a new window" class="key_to_click" tabindex="0"><i class="fas fa-window-restore" title="Pop up the file in a new window"></i></a>
-        <a onclick='downloadFile("{{ file.path|url_encode }}", "attachments")' aria-label="Download the file" class="key_to_click" tabindex="0"><i class="fas fa-download" title="Download the file"></i></a>
+        <button onclick='downloadFile("{{ file.path|url_encode }}", "attachments")' aria-label="Download the file" class="key_to_click" tabindex="0"><i class="fas fa-download" title="Download the file"></i></button>
         {% if can_modify %}
-        <a onclick='deleteAttachment(this, "{{ file.name|url_encode }}")' aria-label="Delete the file" class="key_to_click" tabindex="0"><i class="fas fa-trash" title="Delete the file"></i></a>
+        <button onclick='deleteAttachment(this, "{{ file.name|url_encode }}")' aria-label="Delete the file" class="key_to_click" tabindex="0"><i class="fas fa-trash" title="Delete the file"></i></button>
```
shailpatels, line 10: You should also be able to delete the `key_to_click` class since it was only used to add an event listener to manually trigger the `onclick` event on `a` tags. But the button element will automatically fire it 

see this function here which does the logic for adding the events
https://github.com/Submitty/Submitty/blob/master/site/public/js/server.js#L1414-L1424

shailpatels (changes requested):

site/app/templates/autograding/Attachments.twig
immortalcodes: I will do as suggested.

site/app/templates/course/CourseMaterials.twig
```diff
@@ -9,13 +9,13 @@
     {% if user_group == 1 %}
         <div style="float: right; margin-bottom: 20px;">
             {# button to open all of the subfolders and cookie it#}
-            <a onclick='setCookie("foldersOpen",openAllDivForCourseMaterials());' class="btn btn-primary key_to_click" tabindex="0">Open/Close All Folders</a>
+            <button onclick='setCookie("foldersOpen",openAllDivForCourseMaterials());' class="btn btn-primary key_to_click" tabindex="0">Open/Close All Folders</button>
```
shailpatels, line 12: I'm seeing some visual changes in the course materials pages I think this. I think there are some issues where some of the buttons don't have any text within the button so maybe some of the icons need to be refactored so that they are nested within the button instead of containing the anchor tag

image

site/app/templates/course/CourseMaterials.twig
```diff
@@ -61,7 +61,7 @@
             <div class="file-viewer">
                 {% if course_material.isLink %}
                     <i class="fas fa-link" style="vertical-align: text-bottom;"></i>
-                    <a href="{{ course_material.getUrl }}" onclick="markViewed([{{ course_material.getId }}])" onauxclick="markViewed([{{ course_material.getId }}])">{{ course_material.getUrlTitle }}</a>
+                    <button href="{{ course_material.getUrl }}" onclick="markViewed([{{ course_material.getId }}])" onauxclick="markViewed([{{ course_material.getId }}])">{{ course_material.getUrlTitle }}</button>
```
shailpatels, line 64: This should not be converted to a `button`, instructors may also post links in the course materials page - this is what is causing the cypress e2e error since the link is no longer functional.

image

shailpatels (changes requested):

Issue #5291, opened 2020-04-23
Remove python2 from default Submitty installation


MasterOdin:
Python2 was end-of-life at the start of January 2020, and its final release came out a few days ago. Submitty itself also has no core dependency on python2 and is only used for a number of sample gradeables. We should upgrade the sample gradeables to python2 (and remove the python2 example from the choice of language example and just use C, C++, and Python3), and then remove dealing with python2 from the installation scripts.

- [ ] Update more_autograding_examples to be python3 (except choice_of_language)
- [ ] Remove python2 example from choice_of_language
- [ ] Update any example in Tutorial using python2
- [ ] Update any supporting Submitty library to use python3 by default
- [ ] Remove python2 from installation scripts

bmcutler:
@MasterOdin Can we close this now?

MasterOdin:
Re-opening as we still depend on python2 being installed for sample course as it relies upon the `python_simple_homework` assignment for its testing, and that python2 is still included in the installation script.

PR #7913, opened 2022-05-20
[Testing:Developer] CI version updates

williamjallen:
The GitHub CI currently uses Ubuntu 18.04 and Python 3.6 to run our tests.  The production machines and developer environment use Ubuntu 20.04 and Python 3.8.

### What is the new behavior?
The GitHub CI has been updated to match the versions of Ubuntu and Python used in the developer environment and in production.  As a result of these changes, a number of tests were disabled.


williamjallen:
This is related to https://github.com/Submitty/Submitty/issues/5291

MasterOdin:
@williamjallen any idea on why the auto grading integration tests are failing on the updated python version?

williamjallen:
Some of the autograding tests are failing because the assignments are run using Python 2, which is included in Ubuntu 18.04, but not 20.04.  That should be relatively straightforward to fix.  A few other tests are failing due to some issue with Java, as well as a few which are failing the memory debugging tests.  I have looked at both of those in some depth but I haven't figured out the source of the issue yet.  Thoughts on what might be causing this behavior?

Barb said to make a PR so we can track it, which is what this is.

williamjallen:
Are you sure that Ubuntu 18.04 comes with Java 8?  Everything I can find seems to indicate that Ubuntu 18.04 and 20.04 both include Open JDK 11.  I'm not particularly knowledgeable about Java so I could be wrong here though.

MasterOdin:
The machines that are provided by GitHub come pre-installed with some amount of software, and so since we're not setting a specific Java version, we are relying on that default. See these two documents for what's installed by default on 18.04 vs 20.04:
* https://github.com/actions/virtual-environments/blob/ubuntu20/20220515.1/images/linux/Ubuntu1804-Readme.md
* https://github.com/actions/virtual-environments/blob/ubuntu20/20220515.1/images/linux/Ubuntu2004-Readme.md

williamjallen:
It appears that the python unit tests are failing on https://github.com/Submitty/Submitty/pull/7922 because Python 3.6 (the default in Ubuntu 18.04) is not supported by Numpy.  I believe this PR will resolve that issue.

bmcutler (approved):
@williamjallen will open a new issue with a checklist of the autograding tests that must be repaired & re-enabled.

PR #7918, opened 2022-05-22
[Bugfix:Submission] Bulk submission info collision

### LINKED ISSUE #4657: Page count for bulk upload with QR code fails if used after regular bulk upload
**Describe the bug**
When testing the bulk upload for page ranges, I first uploaded a random PDF as a regular bulk upload and selected a page count of 1. I then deleted all uploaded PDFs, and switched to the QR mode, where I then uploaded the `01_qr_six_6page.pdf` test file and selected value of `4`. However, after all were processed, the `Page Count` column red `1` for all PDFs. No matter how many times I deleted all PDFs, the count column was always `1`. Viewing the full PDF button (and not just the cover image picture) showed that the PDF had 6 pages.

**Expected behavior**
Expected count to accurately reflect number of pages in the uploaded PDF, or prevent someone from going to the bulk upload by PDF if it's not supposed to be used after a regular bulk upload.

**To Reproduce**
1. Upload via non-qr code bulk upload with value of 1
2. Upload PDF via qr code bulk upload with any selected value of pages

**Configuration**
 - Browser [e.g. chrome, safari] chrome

### LINKED ISSUE #7882: Bulk upload file name collision
**Describe the bug**
When resubmitting a bulk upload with the same file name, the oldest file with that name will be used. This could cause issues if you set the wrong prefix and go and correct it later on. The ID will never be correct in that case until you change the file name.

**Expected behavior**
The newest file with that name should be used so this issue doesn't happen.

**To Reproduce**
1. Go to the submission page of the Grading Homework PDF gradeable in the sample course
2. Do a bulk upload submission with `02_qr_prefix_small_errors.pdf` inside of `more_autograding_examples/bulk_upload_pdfs/submissions` and split by QR code. Don't set a prefix yet!
3. Look at the user IDs in the unassigned PDFs section and notice that there is a prefix of `sampledata_`.
4. Click the delete all PDFs button
5. Do another bulk upload submission with same file but set the prefix to be `sampledata_`.
6. Look at the user IDs again and notice the prefix is still there when it should have been removed.

**Additional context**
The code that needs to be fixed is here:
https://github.com/Submitty/Submitty/blob/d67db099d042377b8108476f0fb7ab6b0a600db9/site/app/views/submission/HomeworkView.php#L551-L553
The issue is how PHP combines arrays with duplicate keys. The duplicate key being the file name.

Eceptonsu:
Fixes #7882
Fixes #4657
After a resubmission for bulk upload, the page count column is not correctly updated if changing from split-by-page to split-by-QR. Moreover, the prefixes for the user ID are not cleared after a resubmission with specified QR code prefixes. The issue is PHP arrays concatenation preserves the original keys.

### What is the new behavior?
After a resubmission for bulk upload with changed settings, the relevant info will be correctly updated and shown in the PDF preview section after a refresh when the file processing is finished.


shailpatels (approved):
I think this should fix the two issues described. Overwriting previous files with newer ones should cover most use cases. 

There might be a use case where instructors upload the same file twice with different parameters without submitting any split items in between. This case is probably rare and the long term strategy is to use unique filenames in the backend to prevent overwriting any information.

For example, if a user uploads a QR bulk upload followed by a regular bulk upload none of the QR parsed information is displayed - but again this probably a rare use case and for an incremental improvement this PR should be good to merge

PR #7925, opened 2022-05-23
[Refactor:Developer] Mark Ubuntu 18.04 as unsupported

MasterOdin:

We currently mark Ubuntu 18.04 as being supported for running Submitty. This version of Ubuntu has a number of old packages (e.g. pip) that prevent us from keeping some stuff about the project up-to-date, as well as use newer language features. 

### What is the new behavior?

Running Submitty on Ubuntu 18.04 is now considered unsupported and a "at your own risk". Once we do stuff like update some pip dependencies, it will be definitely broken out of the box. I believe all RPI machines at this point have moved to 20.04, so this shouldn't break anything for us, and at this point, there's been plenty of time to update to 20.04 for any interested parties. This would ideally go hand in hand with supporting 22.04 in the near future.

### DIFF
--- a/.github/workflows/vagrant_up.yaml
+++ b/.github/workflows/vagrant_up.yaml
@@ -1,6 +1,6 @@
 name: Vagrant Up
 
-on: 
+on:
   schedule:
     - cron: '0 0 * * *'
   workflow_dispatch:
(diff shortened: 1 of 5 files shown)


MasterOdin:
We should ideally get #7913 merged before we merge this so that we're not testing stuff on the now unsupported 18.04 target.

shailpatels (approved):
I think this is good to go when we are ready to drop ubuntu 18

As for the comment about the specific pip installs this PR 
https://github.com/Submitty/Submitty/pull/7723/files should upgrade OpenCV 

and then getting a prebuilt version of cryptography would be really nice to avoid building it from source and getting whatever dependencies it needs for building which can be done separately like you said and  removing that comment then

PR #7932, opened 2022-05-24
[Feature:System] Stop the worker if it has no space

### LINKED ISSUE #7315: Useful error when worker machine partition has no available space
**Describe the bug**
One of our worker machines ran out of space on / (root).
Causing autograding jobs got stuck on the worker machine.
The jobs just froze in the queue.  No error was returned back to the primary machine.
Running /usr/local/submitty/sbin/restart_all_shippers_and_workers.py script ran successfully, 
but some of the jobs (the jobs re-assigned to the broken machine) immediately got stuck again.

**Expected behavior**
Besides getting stuck, it would be helpful to have an error message and/or a timeout happen.

NOTE: The machine ran out of space because /var/lib/docker/overlay2 was large (20GB) and the machine had been mispartitioned with a significantly smaller than intended amount of space available for /var.

Maybe the restart_all_shippers_and_workers.py script should do basic file system checks on return warnings if a partition is > 80% full?

### LINKED ISSUE #7744: submitty_install does not update docker log file

Viyerelu23333:

This PR closes #7315, closes #7744, closes #7900

According to the [original issue](../issues/7315), the worker will get stuck if there is no space left.  And running [`restart_all...py`](../blob/main/sbin/restart_shipper_and_all_workers.py) script will not produce errors.

### What is the new behavior?

Before_Restarting
Restarting
After_Restarting

The worker will check the disk usage when it launches.  The thresholds for Warning and Error are 70% and 93%, respectively.  If the worker's root partition usage meets the thresholds, it will

- leave a warning message on the log file;
- exit, if the usage meets the error threshold.

The [`restart_all...py`](../blob/main/sbin/restart_shipper_and_all_workers.py) script will verify the status of the service after restarting.  If some workers report `inactive`, then it displays a message.  

### Other information?
- Daemon Job Dispatcher
  A bash script could create the daemon job in the CLI, not through the PHP pages.

- Testing
  The testing could be done with a worker machine, and `dd` or `lvreduce` commands.  To fill up the worker machine,
  1. Run `df -h` to retrieve the current usage;
  2. Run `dd if=/dev/zero of=awesomeBigFile bs=1G count=n` to fill the disk with an `n` GiB file;
  3. Alternatively, run `lvreduce -r -L -${n}G /dev/mapper/${AWESOME_LVM_VOLUME}` to reduce the LVM volume by `${n}` GB, make sure to substitute `${AWESOME_LVM_VOLUME}` to your actual volume;
  4. Run `python3 ${SUBMITTY_INSTALL_DIR}/sbin/restart_shipper_and_all_workers.py` on the primary machine.

  Note that there is a difference between `df -h` and `shutil.disk_usage('/')`.  On my end, `df -h` percentage is greater than the latter one.


Viyerelu23333:
The CI fails because of the `jq` version: Ubuntu 18.04 only has `jq-1.5`, but the script uses a feature introduced in `jq-1.6` (`$ARGS.named`).

cjreed121 (approved):
Seems to work well. Reviewed PHP/twig code.

bmcutler (changes requested):
Please rename create_daemon_job.sh to update_worker_health.sh (is that the right name?)

The other tasks are not likely going to be scheduled as cron jobs -- and they are not similar enough to require this general purpose script.  Then this script can be considerably simplified (args etc).

tkoz0:
It would be nice to have a more detailed error message than `Some workers are inactive, check workers' logs.` when a worker is stopped. This message does not help a whole lot with finding the problem.

bmcutler (approved):

bmcutler (approved):

Issue #7289, opened 2021-10-27
Migrate away from PyPDF2 dependency 


shailpatels:
**Describe the bug**
PyPDF2 is a dependency for bulk uploads for reading and writing PDF files but is no longer maintained since 2016
We have seen valid PDF files not being excepted by PyPDF2

**Expected behavior**
Find alternate package that supports PDF reading and writing in python

**To Reproduce**
N/A

**Configuration**
N/A

atharva1051:
I assume this issue has been resolved in #7938 .

Good to see PyPDF2 again under maintenance.
Anyway if you were to move away from PyPDF2 I can suggest few alternatives. 

1.  **_PyPDF4_** is an improved version of PyPDF2 with bug fixes and additional features. It provides similar functionality for reading, writing, and manipulating PDF files.

2.  **_PDFMiner_** is a library for extracting text, images, and metadata from PDF files. It offers more advanced features for analyzing PDF content, such as converting PDF pages to images or extracting tables.

3. _**pdfrw**_ is a library that allows you to read and write PDF files in a low-level, programmatic way. It provides direct access to the PDF file structure, which can be useful for specific PDF manipulation tasks.

PR #7938, opened 2022-05-25
[Dependency] PyPDF2 version update

akshatbatra:
Related to #7289, some PDF files are not accepted by PyPDF2.

### What is the new behavior?
Active maintenance of PyPDF2 has resulted in it being more dependable while working with corrupt PDFs.
PyPDF2 is now able to work with files such as [broken.pdf](https://github.com/Submitty/Submitty/files/7422603/broken.pdf) with `strict=False` set, it can now deal with xref table not being present at the specified location.
Required version of PyPDF2 has been updated to 1.28.2 and its deprecated usage has been replaced with latest recommendations.


cjreed121:
This is a duplicate of https://github.com/Submitty/Submitty/pull/7926. One of these should be closed.

bmcutler:
Yes please, make a PR to the other repo :)

Eceptonsu (approved):
It works well for me even though it is a bit slow.

shailpatels (approved):
Thanks for the fix and good to hear that pypdf2 is under active maintenance again!

Tested with regular and bulk upload and both appeared to work as expected

sbin/submitty_daemon_jobs/submitty_jobs/bulk_qr_split.py
```diff
@@ -33,15 +33,15 @@ def main(args):
 
     try:
         os.chdir(split_path)
-        pdfPages = PdfFileReader(filename)
+        pdfPages = PdfFileReader(filename, strict=False)
         pdf_writer = PdfFileWriter()
```
MasterOdin, line 37: We'll need to update the class names while we're at it as the `PdfFile*` version has been deprecated in favor of the shorter names.

```suggestion
        pdfPages = PdfReader(filename, strict=False)
        pdf_writer = PdfWriter()
```

MasterOdin (changes requested):

MasterOdin (approved):

MasterOdin:
Before merging this, we need to improve how we manage python dependencies so that they get updated as part of running `INSTALL_SUBMITTY_HELPER.sh` instead of on `install_system.sh`, as otherwise all code that relies on the new version of PyPDF2 will break until manually run `pip install -U PyPDF2` (with appropriate chmod).

PR #7940, opened 2022-05-26
[SYSADMIN ACTION][Feature:System] HTTP/2 Support

### LINKED ISSUE #3850: Setup http/2 on the server
As we split up our files to be more modular, users will have to download more files for each page. The best way to optimize load times is to take advantage of http2's multiplexing and avoid bundling.

I'm pretty sure we just need to make a few changes to `submitty.conf` to setup http2. We should be sure to modify both `submitty.conf` and `submitty_http.conf` in the .setup directory to help new sysadmins install Submitty

Viyerelu23333:

* [x] Documentation has been updated/added if relevant (https://github.com/Submitty/submitty.github.io/pull/438/)

Closes #3850 

Currently, the Apache2 serves all content in HTTP/1.1.  And the server uses `prefork` as the multiprocessing module.  The module is introduced by `mod_php`, which is not used in production.  

### What is the new behavior?

Most browsers only support HTTP/2 when the TLS is enabled.  The script will also enable HTTPS feature in the development environment.  If a browser does not support HTTP/2, then the server will serve contents in HTTP/1.1.

Developers could use `dev-upgrade-h2.sh` to change the HTTP version:

- `dev-upgrade-h2.sh up`: Upgrade
  After upgrading, the server uses TLS and HTTP/2; or HTTP/1.1 + TLS if a browser does not support `h2`.

- `dev-upgrade-h2.sh down`: Downgrade
  The server uses HTTP/1.1, or HTTP/2 over TCP (`h2c`) if a browser supports `h2c`.

`mod_php` will be disabled and another multiprocessing module (`mpm_event`) will be used in Apache2.  This is effective before upgrading/downgrading.

### Other information?

- NGINX
  The NGINX server still serves WebSocket traffics with HTTP/1.1.

  **Note that** most browsers could not trust WebSocket traffics with self-signed certificates.  If you are dealing with WebSocket related features, there are some workarounds:
    - Copy the certs from VM and trust it on your host system; OR
    - Trust/Ignore the certificates on browsers; OR
    - Downgrade to HTTP/1.1 using `.setup/dev-upgrade-h2.sh down`.

  Trusting/ignoring certificates on browsers:
    - For Chrome, start Chrome with `--ignore-certificate-errors` option.
    - For Firefox...
      1. trust the certificate when the warning pops up;
      2. type `https://localhost:8443/ws` and select trust again.

  This will not impact the production as it is equipped with standard certificates from the trusted CA.

- Performance Tuning, Production
  See documentations above.


shailpatels:
See the following lines in the `git_tests.sh` script and see if changing the `http` to `https` will pass the github CI

https://github.com/Submitty/Submitty/blob/main/tests/git_test.sh#L8
https://github.com/Submitty/Submitty/blob/main/tests/git_test.sh#L46

Viyerelu23333:
> See the following lines in the `git_tests.sh` script and see if changing the `http` to `https` will pass the github CI
> 
> https://github.com/Submitty/Submitty/blob/main/tests/git_test.sh#L8 https://github.com/Submitty/Submitty/blob/main/tests/git_test.sh#L46
> 
> hopefully that's the only issue and its a quickfix
> 
> Here is an example of redirecting from HTTP to https automatically which might be beneficial https://www.namecheap.com/support/knowledgebase/article.aspx/9821/38/apache-redirect-to-https/
> 
> However I'm not sure how git will handle the redirect maybe it will follow it when trying a pull or maybe not. But using https in the git tests should be the way forward since that's what users would use on production anyways

The commit above attempts to fix the git_test issue, but it failed again with an authentication error.  I did not reproduce the issue with vagrant VM...  The certificate is trusted by the system.  

I just realized that Apache cannot serve both HTTP and HTTPS on the same port (1511)...

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -916,6 +916,12 @@ if [ "${WORKER}" == 0 ]; then
     fi
 fi
 
+# Upgrade to http/2 with TLS for non-worker Vagrant VMs
```
shailpatels, line 919: `install_submitty` could be run multiple times in the same VM, especially frequently for developers working outside the site development code. Is it safe to run the dev-upgrade scripts multiple times with it? Would it make sense to move this to `install_system` if possible? So it will only run with the initial vagrant up

Viyerelu23333: Thank you for reviewing.  Most of the codes are safe to rerun -- it will remove the changes then re-applying -- except that the certificate might expire in 730 days since the script launched.  I just added an expiration check.

shailpatels (changes requested):
Sorry for the delay in reviewing, I think upgrading an existing vagrant instance to https is working well just some edits need to be done for a new vagrant up.

For the vagrant up job here
https://github.com/Submitty/Submitty/runs/7225673608?check_suite_focus=true#step:4:11

it appears to fail with the error message
```
    ubuntu-20.04: /usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/dev-upgrade-h2.sh: line 32: SUBMITTY_INSTALL_DIR: parameter null or not set
```

That value is definitely set earlier in the `INSTALL_SUBMITTY_HELPER` script so maybe its the way you're invoking it? I think maybe trying the syntax `"${SUBMITTY_REPOSITORY}/.setup/dev-upgrade-h2.sh" up` and manually passing any other env vars needed could work better since the `SUBMITTY_INSTALL_DIR` isn't being exported in the  `INSTALL_SUBMITTY_HELPER` script

shailpatels:
See here docs for manually running a workflow for debugging : https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

shailpatels (approved):
Looks good, I think this is ready to be merged!

Here is a vagrant up run for this: https://github.com/Submitty/Submitty/actions/runs/2694743160

I would also recommend updating the documentation site for https, and making sure developers know they must specify `https` when visiting the site or they'll get a bad response. Since its a self signed cert they'll also get browser warnings like this first time so probably good to mention that too
Capture

 This is probably this is the page/section you want to update
https://submitty.org/developer/getting_started/vm_install_using_vagrant#using-your-submitty-developer-vm

.setup/install_system.sh
```diff
@@ -435,6 +435,13 @@ if [ ${WORKER} == 0 ]; then
 
     a2enmod include actions cgi suexec authnz_external headers ssl proxy_fcgi rewrite proxy_http proxy_wstunnel
 
+    # Disable mod_php and prefork based multiprocessing module
+    # as Submitty uses php-fpm to handle the PHP requests
+    a2dismod php${PHP_VERSION} mpm_prefork
+    # Enable event based multiprocessing module and http/2 module
+    # this will enable http/2 on any directory, if possible
+    a2enmod mpm_event http2
```
cjreed121, line 443: Won't this then require that servers have SSL/TLS enabled already? I'm only pointing this out cause it would effect new vagrant machines. Maybe this should be wrapped in a vagrant == 0?

cjreed121 (changes requested):

.setup/install_system.sh
Viyerelu23333: I'll remove these lines, as the upgrade now is a part of sysadmin actions.  Thank you for reviewing.

.setup/dev-upgrade-h2.sh
```diff
@@ -0,0 +1,335 @@
...
+P_CERT="/etc/ssl/selfcert"
+
+# generate certificate + key to the domain below
+S_DOMAIN="localhost"
+
+# should force update the certificates
+B_FORCE=0
+
+# Submitty system-defined env vars and checks
+SUBMITTY_INSTALL_DIR=${SUBMITTY_INSTALL_DIR:?}
```
04vmatsibekker, line 32: image

cjreed121 (approved):

PR #7946, opened 2022-05-26
[Bugfix:TAGrading] Numeric grading fixes

Eceptonsu:
Users are able to enter strings and negative integers into numeric cells which will cause save errors in the back end without any feedback. At that point, a page refresh is needed. In addition, input greater than the max_clamp value for the particular component is also allowed, which will produce the same situation. 
For the comment, users can not delete text for the cell it is filled.

### What is the new behavior?
Only positive scores that are less than the max_clamp value for the component are allowed in numeric cells.
Users are able to delete text in comment cells.

### DIFF
--- a/site/app/templates/grading/simple/Display.twig
+++ b/site/app/templates/grading/simple/Display.twig
@@ -301,6 +301,7 @@
                     value="{{ component_score }}"
                     onclick="this.select()"
                     onchange=" this.style.backgroundColor = 'var(--default-white)';"
+                    onkeydown="javascript: return ['Backspace','Delete','ArrowLeft','ArrowRight'].includes(event.code) ? true : !isNaN(Number(event.key)) && event.code!=='Space'"
                     {% if component_grade is not null and component_grade.getGrader() != null %}
                         data-grader="{{ component_grade.getGrader().getId() }}"
                     {% endif %}
--- a/site/app/templates/grading/simple/Display.twig
+++ b/site/app/templates/grading/simple/Display.twig
@@ -122,7 +122,7 @@
     {% endif %}
 
     {# This is a data table #}
-    <table id="data-table" data-gradeable="{{ gradeable.getId() }}" data-current-grader="{{ user_id }}" data-max-clamps="{{ max_clamps }}" class="table table-striped table-bordered persist-area" style="table-layout: fixed;">
+    <table id="data-table" data-gradeable="{{ gradeable.getId() }}" data-current-grader="{{ user_id }}" class="table table-striped table-bordered persist-area" style="table-layout: fixed;">
         <caption />
         {# Table header #}
         <thead>
@@ -285,6 +285,7 @@
     {% if components_numeric|length > 0 %}
         {% set total = 0 %}
         {% for component in grade.getGradeable().getComponents()|filter((component) => not component.isText()) %}
+            {% set max_clamp = component.getUpperClamp() %}
             {% set ta_grade = grade.getTaGradedGradeable() %}
             {% set component_grade = ta_grade is not null ? ta_grade.getGradedComponent(component) : null %}
             {% set component_score = component_grade.getScore() | default(0) %}
@@ -298,6 +299,7 @@
                     id="cell-{{ section_id }}-{{ index }}-{{ loop.index0 }}"
                     data-id="{{ component.getId() }}"
                     data-origval="{{ component_score }}"
+                    data-maxclamp="{{ max_clamp }}"
                     value="{{ component_score }}"
                     onclick="this.select()"
                     onchange=" this.style.backgroundColor = 'var(--default-white)';"
--- a/site/app/templates/grading/simple/Display.twig
+++ b/site/app/templates/grading/simple/Display.twig
@@ -122,7 +122,7 @@
     {% endif %}
 
     {# This is a data table #}
-    <table id="data-table" data-current-grader="{{ user_id }}" class="table table-striped table-bordered persist-area" style="table-layout: fixed;">
+    <table id="data-table" data-gradeable="{{ gradeable.getId() }}" data-current-grader="{{ user_id }}" class="table table-striped table-bordered persist-area" style="table-layout: fixed;">
         <caption />
         {# Table header #}
         <thead>
--- a/site/app/templates/grading/simple/Display.twig
+++ b/site/app/templates/grading/simple/Display.twig
@@ -122,7 +122,7 @@
     {% endif %}
 
     {# This is a data table #}
-    <table id="data-table" data-gradeable="{{ gradeable.getId() }}" data-current-grader="{{ user_id }}" class="table table-striped table-bordered persist-area" style="table-layout: fixed;">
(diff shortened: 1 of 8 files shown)


site/app/templates/grading/simple/Display.twig
```diff
@@ -292,15 +292,16 @@
             <td class="option-small-input">
                 <input
                     class="option-small-box cell-grade" tabindex="0"
-                    style="text-align: center; {{ component_score == 0 ? "color: var(--standard-light-medium-gray);" : "" }}
+                    style="text-align: center; width: 120px; {{ component_score == 0 ? "color: var(--standard-light-medium-gray);" : "" }}
                         {{ component_score == 0 ? "background-color: var(--alert-background-blue);" : "background-color: var(--default-white);" }}"
-                    type="text"
+                    type="number"
```
poorna2152, line 297: Setting `type="number"` would only allow to insert integer values in to the input field. 
I think user should also be able to enter floats as well.( I assume floats also be accepted based on [this ](https://github.com/Submitty/Submitty/blob/main/site/public/js/simple-grading.js#L436)). You could also use the attribute `min=0 ` to ensure that negative values are not inserted.

site/public/js/simple-grading.js
```diff
@@ -413,23 +413,61 @@ function setupCheckboxCells() {
...
+        var row_el = $("tr#row-" + split_id[1] + "-" + split_id[2]);
+        console.log(elem.data('origval'));
+        console.log(elem.val());
+
+        var scores = {};
+        var old_scores = {};
+        var total = 0;
+
+        var exceed_max = false;
+        const numbers = /^[0-9]+$/;
```
poorna2152, line 426: Regex only works for positive integers

site/public/js/simple-grading.js
```diff
@@ -413,23 +413,61 @@ function setupCheckboxCells() {
...
+        var split_id = elem.attr("id").split("-");
+        var row_el = $("tr#row-" + split_id[1] + "-" + split_id[2]);
+        console.log(elem.data('origval'));
+        console.log(elem.val());
+
+        var scores = {};
+        var old_scores = {};
+        var total = 0;
+
+        var exceed_max = false;
```
poorna2152, line 425: Is this variable used.

site/public/js/simple-grading.js
```diff
@@ -413,23 +413,55 @@ function setupCheckboxCells() {
...
+        const numbers = /^[0-9]*\.?[0-9]*$/;
+
+        if(this.tagName.toLowerCase() === 'input') {
+            // Empty input is ok for comment but not numeric cells
+            if(!this.value) {
+                this.value = 0;
+            }
+            else if(!this.value.match(numbers)) {
+                alert('Score should be a positive number');
+                this.value = 0;
```
poorna2152, line 433: return the function from here.

site/public/js/simple-grading.js
```diff
@@ -413,23 +413,55 @@ function setupCheckboxCells() {
...
+        var old_scores = {};
+        var total = 0;
+
+        let value = this.value;
+        const numbers = /^[0-9]*\.?[0-9]*$/;
+
+        if(this.tagName.toLowerCase() === 'input') {
+            // Empty input is ok for comment but not numeric cells
+            if(!this.value) {
+                this.value = 0;
```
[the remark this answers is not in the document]
Eceptonsu: For empty input, I think it makes sense to change from an empty string to 0, and the value needs to be saved so I did not make the function returns here.

site/public/js/simple-grading.js
Eceptonsu: Same as above.

site/public/js/simple-grading.js
```diff
@@ -413,23 +413,55 @@ function setupCheckboxCells() {
...
+            }
+            else if(!this.value.match(numbers)) {
+                alert('Score should be a positive number');
+                this.value = 0;
+            }
+            // Input greater than the max_clamp for the component is not allowed
+            else {
+                submitAJAX(
+                    buildCourseUrl(['gradeable', row_el.data('gradeable'), 'maxclamp']),
```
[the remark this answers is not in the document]
Eceptonsu: Based on how the backend is written, I think at least one check is needed for both requests. But I have made some changes to make the code more concise.

cjreed121 (changes requested):
Seems to work but I think it could be handled more efficiently. Rather than sending a request to check if a value is valid, just store the max clamp values in the page and have JS check them before making the request to save. The save PHP code must also perform the check before attempting to save to the database.

site/app/templates/grading/simple/Display.twig
```diff
@@ -292,15 +292,16 @@
...
                     id="cell-{{ section_id }}-{{ index }}-{{ loop.index0 }}"
                     data-id="{{ component.getId() }}"
                     data-origval="{{ component_score }}"
                     value="{{ component_score }}"
```
cjreed121, line 301: Why not store the max clamp in a place like this? I think the split will work but I think putting data attributes feels cleaner.

site/public/js/simple-grading.js
```diff
@@ -413,23 +413,42 @@ function setupCheckboxCells() {
...
+            }
+            else if(!this.value.match(numbers)) {
+                alert('Score should be a positive number');
+                this.value = 0;
+            }
+            // Input greater than the max_clamp for the component is not allowed
+            else {
+                let max_clamps = localStorage.getItem('max_clamp').split('-');
+                if(max_clamps[split_id[3]] && max_clamps[split_id[3]] < this.value) {
+                    alert('Score should be less than the max clamp value: ' + max_clamps[split_id[3]]);
```
cjreed121, line 439: This should probably be "less than or equal to" rather than just "less than"

cjreed121 (changes requested):
I also noticed an issue. If I had a max clamp of 5 and put in a number greater than 5 but less than 10 it would do an alert. But if I did 10 there was no alert but there was a message that went in the JS console.

Eceptonsu:
> 

Thanks for pointing out, changes have been made.

site/app/templates/grading/simple/Display.twig
```diff
@@ -122,7 +122,7 @@
     {% endif %}
 
     {# This is a data table #}
-    <table id="data-table" data-current-grader="{{ user_id }}" class="table table-striped table-bordered persist-area" style="table-layout: fixed;">
+    <table id="data-table" data-gradeable="{{ gradeable.getId() }}" data-current-grader="{{ user_id }}" class="table table-striped table-bordered persist-area" style="table-layout: fixed;">
```
cjreed121, line 125: Why add data-gradeable?

cjreed121 (changes requested):
After this last requested change I think this should be good to go.

site/app/templates/grading/simple/Display.twig
Eceptonsu: Forgot to delete this change from previous commits, should be good now.

cjreed121 (approved):

PR #7880, opened 2022-05-10
[Refactor:System] Quote variables in install scripts

williamjallen:
### What is the current behavior?
There are many improperly-quoted variables in Submitty's installation scripts.  This can result in [unintended consequences](https://unix.stackexchange.com/a/171347/195445) if variables contain spaces, or other special characters.

### What is the new behavior?
This PR continues the work begun in #7874 to improve the robustness of Submitty's various shell scripts by properly quoting the variables in most of the installation scripts.

PR #7957, opened 2022-05-27
[Feature:Autograding] Add cloning of AnalysisToolsTS

poorna2152:
### What is the new behavior?
Clones the AnalysisToolsTS repository
### Other information?

Tested using the `submitty_install`

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -675,14 +675,14 @@ chmod -R 555 ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisTools
 
 echo -e "Build and install analysis tools ts"
 
-mkdir -p ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+mkdir -p "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
 
-ANALYSIS_TOOLS_TS_REPO=${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/AnalysisToolsTS/
+ANALYSIS_TOOLS_TS_REPO="${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/AnalysisToolsTS/"
 
 # Copy cloned files to AnalysisToolsTS directory
-rsync -rtz ${ANALYSIS_TOOLS_TS_REPO} ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+rsync -rtz "${ANALYSIS_TOOLS_TS_REPO}" "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
 
-pushd ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+pushd "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
 
 # # install npm packages
 npm install
@@ -693,8 +693,8 @@ npm run build
 popd > /dev/null
 
 # # change permissions
-chown -R ${DAEMON_USER}:${COURSE_BUILDERS_GROUP} ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
-chmod -R 555 ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+chown -R "${DAEMON_USER}":"${COURSE_BUILDERS_GROUP}" "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
+chmod -R 555 "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
 
 #####################################
 # Add read & traverse permissions for RainbowGrades and vendor repos
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -685,10 +685,10 @@ rsync -rtz "${ANALYSIS_TOOLS_TS_REPO}" "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysis
 pushd "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
 
 # # install npm packages
-npm install
+sudo npm install
 
 # # build project
-npm run build
+sudo npm run build
 
 popd > /dev/null
 
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -669,6 +669,32 @@ popd > /dev/null
 chown -R ${DAEMON_USER}:${COURSE_BUILDERS_GROUP} ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisTools
 chmod -R 555 ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisTools
 
+################################################################################################################
+################################################################################################################
+# BUILD AND INSTALL ANALYSIS TOOLS TS
+
+echo -e "Build and install analysis tools ts"
+
+mkdir -p ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+
+ANALYSIS_TOOLS_TS_REPO=${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/AnalysisToolsTS/
+
+# Copy cloned files to AnalysisToolsTS directory
+rsync -rtz ${ANALYSIS_TOOLS_TS_REPO} ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+
+# pushd ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+
+# # install npm packages
+# npm install
+
+# # build project
+# npm run build
+
+popd > /dev/null
+
+# # change permissions
+chown -R ${DAEMON_USER}:${COURSE_BUILDERS_GROUP} ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+chmod -R 555 ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
 
 #####################################
 # Add read & traverse permissions for RainbowGrades and vendor repos
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -682,13 +682,13 @@ ANALYSIS_TOOLS_TS_REPO=${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/AnalysisToolsTS/
 # Copy cloned files to AnalysisToolsTS directory
 rsync -rtz ${ANALYSIS_TOOLS_TS_REPO} ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
 
-# pushd ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+pushd ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
 
 # # install npm packages
-# npm install
+npm install
 
 # # build project
-# npm run build
+npm run build
 
 popd > /dev/null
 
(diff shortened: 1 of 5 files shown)


.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
 chown -R ${DAEMON_USER}:${COURSE_BUILDERS_GROUP} ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisTools
 chmod -R 555 ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisTools
 
+################################################################################################################
+################################################################################################################
+# BUILD AND INSTALL ANALYSIS TOOLS TS
+
+echo -e "Build and install analysis tools ts"
+
+mkdir -p ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
```
williamjallen, line 678:
```suggestion
mkdir -p "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
```

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
...
+
+ANALYSIS_TOOLS_TS_REPO=${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/AnalysisToolsTS/
```
williamjallen, line 680:
```suggestion
ANALYSIS_TOOLS_TS_REPO="${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/AnalysisToolsTS/"
```

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
...
+# # install npm packages
+# npm install
+
+# # build project
+# npm run build
+
+popd > /dev/null
+
+# # change permissions
+chown -R ${DAEMON_USER}:${COURSE_BUILDERS_GROUP} ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
```
williamjallen, line 696:
```suggestion
chown -R "${DAEMON_USER}":"${COURSE_BUILDERS_GROUP}" "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
```

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
...
+chmod -R 555 ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
```
williamjallen, line 697:
```suggestion
chmod -R 555 "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
```

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
...
+
+# pushd ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+
+# # install npm packages
+# npm install
+
+# # build project
+# npm run build
+
+popd > /dev/null
```
williamjallen, line 693: If `pushd` is commented out, this should be commented out as well I believe

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
...
+# BUILD AND INSTALL ANALYSIS TOOLS TS
+
+echo -e "Build and install analysis tools ts"
+
+mkdir -p ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+
+ANALYSIS_TOOLS_TS_REPO=${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/AnalysisToolsTS/
+
+# Copy cloned files to AnalysisToolsTS directory
+rsync -rtz ${ANALYSIS_TOOLS_TS_REPO} ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
```
williamjallen, line 683:
```suggestion
rsync -rtz "${ANALYSIS_TOOLS_TS_REPO}" "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
```

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
...
+
+# pushd ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS
+
+# # install npm packages
+# npm install
+
+# # build project
+# npm run build
```
williamjallen, line 691: Is there any particular reason why we don't want to go ahead and build the project here, even if we aren't using it for anything yet?  This isn't necessarily something that I'm requesting be done in this PR, just a general question.

If the installation scripts aren't run, I think it would be good to at least put some sort of placeholder message here to indicate that installation was skipped.

williamjallen (changes requested):
In general, these changes look good with the exception of a handful of small technical issues.  You'll need to add the new repository to the list of directories to share and then run `vagrant up` again to ensure that this directory is shared correctly.

See this for the relevant lines that need to be changed: https://github.com/Submitty/Submitty/blob/724cecf75896aab12b5931de12a3dbdc1b530778/Vagrantfile#L71

.setup/INSTALL_SUBMITTY_HELPER.sh
poorna2152: Is the quotations needed. I copied the AnalysisTools build procedure. https://github.com/Submitty/Submitty/blob/main/.setup/INSTALL_SUBMITTY_HELPER.sh#L587-L627

.setup/INSTALL_SUBMITTY_HELPER.sh
poorna2152: Uncommented. Tested the procedure by running `submitty_install`

.setup/INSTALL_SUBMITTY_HELPER.sh
williamjallen: Yes, it is good practice to quote variables in shell scripts to prevent command injection.  I actually have [another PR ](https://github.com/Submitty/Submitty/pull/7880) open right now to quote a bunch of variables in this file.

williamjallen (changes requested):
Changes look good in general but I ran into a small hangup while installing.  I ran `install_submitty` without the AnalysisToolsTS repository cloned locally and everything worked fine.  I then ran `vagrant halt`, cloned the AnalysisToolsTS repository, ran `vagrant up`, and ran `install_submitty` again.  It produced this error:
image

poorna2152:
I didnt come across this. I updated the pr to use `sudo npm install.` Can you check it now.

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
...
+# Copy cloned files to AnalysisToolsTS directory
+rsync -rtz "${ANALYSIS_TOOLS_TS_REPO}" "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
+
+pushd "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
+
+# # install npm packages
+sudo npm install
+
+# # build project
+sudo npm run build
```
williamjallen, line 691: In general, we should probably not be using `sudo` when installing dependencies since doing so allows the dependencies we're installing to have a lot of unnecessary permissions (including, in the extreme case, potentially deleting all the files on the system).  I think this could be resolved by changing the permissions on the relevant files but I'm not sure what the correct permissions should be.  Perhaps @bmcutler can advise here?

williamjallen (changes requested):

.setup/INSTALL_SUBMITTY_HELPER.sh
poorna2152: Update to run the `npm install` from `daemon user`. Gave daemon user `rwx `permissions to the directory `SubmittyAnalysisToolsTS`.  rwx permissions are needed due to creation of new directories in `npm install`.

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
...
+
+# # change permissions
+chown -R "${DAEMON_USER}":"${COURSE_BUILDERS_GROUP}" "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
+chmod -R 755 "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
+
+# # install npm packages
+su - ${DAEMON_USER} -c "cd ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS && npm install"
+
+# # build project
+su - ${DAEMON_USER} -c "cd ${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS && npm run build"
```
williamjallen, line 695: These variables should ideally be quoted.  Additionally, is there a reason why the `cd` is in these statements?  I think it would be better to have separate `pushd` and `popd` statements above and below each of the `su` statements, unless there's a reason for the current behavior.

williamjallen (changes requested):
After these few small changes are addressed, I believe that this PR will be ready to be merged.

.setup/INSTALL_SUBMITTY_HELPER.sh
poorna2152: Using `su -` changes the currently residing directory that to the switched users home. Thus the `cd` was needed there. . Changed the command to use `su` so the directories will not get changed.

Is it okay to use `su` instead of `su -` (from security point).  Because following article says.
https://www.geeksforgeeks.org/difference-between-su-and-su-command-in-linux/

"_su command does not create a new User Environment (in the simple term they pretends to be the target user) but su - creates a totally new User Environment (in the simple term they are actually the target user) that we can check with environment variables like:_"

poorna2152: Also based on the meeting yesterday removed the addition of `submitty_count_ts` command in execute.cpp. Will add this in a seperate PR.

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -669,6 +669,32 @@ popd > /dev/null
...
+# BUILD AND INSTALL ANALYSIS TOOLS TS
+
+echo -e "Build and install analysis tools ts"
+
+mkdir -p "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
+
+ANALYSIS_TOOLS_TS_REPO="${SUBMITTY_INSTALL_DIR}/GIT_CHECKOUT/AnalysisToolsTS/"
+
+# Copy cloned files to AnalysisToolsTS directory
+rsync -rtz "${ANALYSIS_TOOLS_TS_REPO}" "${SUBMITTY_INSTALL_DIR}/SubmittyAnalysisToolsTS"
```
williamjallen, line 683: Is there a reason why we're copying into `SubmittyAnalysisToolsTS` instead of `AnalysisToolsTS`?

williamjallen (changes requested):
Sorry for the somewhat late reply.  Will approve once this last comment is resolved.

.setup/INSTALL_SUBMITTY_HELPER.sh
poorna2152: I was replicating the `AnalysisTools` behaviour. AnalysisTools was copying some files to SubmittyAnalysisTools. https://github.com/Submitty/Submitty/blob/main/.setup/INSTALL_SUBMITTY_HELPER.sh#L636
Should it be changed.

williamjallen: I suppose that would be a question for Barb.  Personally, I think `AnalysisToolsTS` is already getting long, and the relation to Submitty is implied, but perhaps Barb is of a different opinion.

bmcutler: Let's leave it for now.  We can revisit the name when we eventually get rid of the original AnalysisTools

williamjallen (approved):

bmcutler (approved):

PR #7964, opened 2022-05-29
[Bugfix:InstructorUI] Table in dark mode

### LINKED ISSUE #7040: Out of place white colored background in dark mode.

immortalcodes:
Fixes #7040 
The Numeric/Text gradeable table have an obnoxious white background when in dark mode.

### What is the new behavior?
The table is displayed in proper colors .
Screenshot_20220530_004916
### Other information?
I have tested the result on local environment and there were some confusion regarding duplication of rows in previous conversations that also seems to be resolved.


shailpatels (changes requested):
Thanks for the fix !

Could also you also update the table for numeric gradeable types which also has a rubric panel with a similar issue. It would be nice to fix up both areas while making these changes

Example of it :
image

Adding the `table-striped` CSS class should fix it as well

immortalcodes:
I have fixed and tested the same.
image

williamjallen (approved):

PR #7972, opened 2022-05-30
[UI/UX:Submission] Change "Grade My Repository" button color

williamjallen:
The Git submission interface is currently somewhat confusing.  The "Grade My Repository" and "Create and Manage Git Auth Tokens" buttons are both primary buttons, and are close to each other.

image

### What is the new behavior?
The "Grade My Repository" button has been changed to a green color to match the style of the "Submit" button for regular gradeables.  Additionally, the "upload message" seen on regular upload gradeables has been added along with a small heading to further clarify the interface.  The words on the authentication tokens button have also been adjusted.

image

### DIFF
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -170,7 +170,14 @@
             {% endif %}
 
         {% endif %}
-        <input type="submit" id="submit" class="submit-gradeable btn btn-primary" value="Grade My Repository" />
+
+        <h3>Grade your repository:</h3>
+        <div class="upload-message">
+            {% include "misc/Markdown.twig" with {
+                "content" : upload_message
+            } only %}
+        </div>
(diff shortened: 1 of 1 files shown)


cjreed121 (changes requested):
Your change of the authentication token button seems to be gone. Looks like the merge of main back in seemed to revert that.

I think putting the message there is a good idea but the message is now a little misleading since the button is not named "Submit". We could change the green button to be named "Submit" but keep the h3 tag that says "Grade your repository"? Or add a specific vcs_upload_message? Not sure if there are any other better options here.

bmcutler:
The vcs button text is not changed by this PR.
We chose "Grade my repository" intentionally -- submission by vcs is a 2 step process, pushing your commits & then requesting grading.

bmcutler:
@williamjallen please update this PR as we discussed last week.
Change "By clicking 'submit'... " to "By submitting..."

cjreed121 (approved):

PR #7984, opened 2022-06-01
[Bugfix:InstructorUI] Fixed Edit Gradeables page

### LINKED ISSUE #6859: Tabs on edit gradeable cover up footer
**Describe the bug**
On the create/edit gradeable page (especially when viewed on a small screen), there is extra space above the div with the content of the page, and if there is enough content displayed to fill the whole screen, that div hides the Submitty footer (either partially or fully, depends on the length of the page)
Screenshot_10

**Expected behavior**
The space above the page content div should be consistent with the rest of the pages on the site, and if the page is too long, the user should be able to scroll in order to see the footer.

immortalcodes:
Fixes #6859, Earlier on create/edit gradeable page the side navigation panel contained scroll and if the content on the page increased in the size then the footer went hidden under the content.

### What is the new behavior?
The scroll from the side navigation is removed and is made consistent with general user interface of the application.

### Other information?
Here is the screenshot for same
image

### DIFF
--- a/site/public/css/electronic.css
+++ b/site/public/css/electronic.css
@@ -27,7 +27,7 @@ main#main.full-screen-mode .content {
 
 #wrapper {
     padding-top: 30px;
-    min-height: 0;
+    min-height: max-content;
 }
 
 aside {
--- a/site/public/css/electronic.css
+++ b/site/public/css/electronic.css
@@ -26,7 +26,6 @@ main#main.full-screen-mode .content {
 }
 
 #wrapper {
-    padding-top: 30px;
     min-height: max-content;
 }
 
--- a/site/public/css/global.css
+++ b/site/public/css/global.css
@@ -111,6 +111,7 @@ nav a > i.f {
 #wrapper {
     display: flex;
     flex-grow: 1;
+    min-height: max-content;
 }
 
 main {


site/public/css/electronic.css
```diff
@@ -27,7 +27,7 @@ main#main.full-screen-mode .content {
 
 #wrapper {
     padding-top: 30px;
-    min-height: 0;
```
shailpatels, line 30: Question: Is this min-height in this file needed ? The `global.css` will put the `#wrapper` CSS rule on every page with the `min-height: max-content;`. 

When I see it on firefox it seems the duplicate rules get ignored and the others are applied. Not sure if this is the same on other browsers. Would just removing the `min-height` in the electronic.css cause any issues instead?

Eceptonsu (commented):
Works for me on Chrome, but I do agree that only having `the min-height: max-content;` in `global.css` will do the work.

site/public/css/electronic.css
immortalcodes: @Eceptonsu @shailpatels  I also though that global.css must do the work but ..
image
Here in the screenshot (firefox browser) , It derives the css component from electronic.css and without it , the property didn't worked.
Hence, addition there was done.

Eceptonsu (changes requested):
I think the original issue #6859 also mentioned that there is extra space between the header and the content div, maybe this should be fixed too.

immortalcodes:
@Eceptonsu Thanks for reminding , I have removed that too.

Eceptonsu (approved):

Issue #7958, opened 2022-05-27
Add Functionality to Cancel a Loading Test Case


MBon1:
**What problem are you trying to solve with Submitty**
After choosing to "Show Details" on a test case in a gradable, the user must wait until the output is displayed.  Especially for larger outputs or when using a slow machine or have a slow Internet connection, this loading can take a noticeable amount of time.  

**Describe the way you'd like to solve this problem**
When the user selects to "Show Details" on a test case, allow for the loading to be aborted, either by making the "Loading..." text be clickable or by adding additional clickable text to perform this action.  After Loading has been cancelled, sets text back to "Show Details".  

**Additional context**
Current "Loading..." Text
![image](https://user-images.githubusercontent.com/20442222/170777806-ac59f2a0-1cbb-4a55-9ed5-e5d294365c9b.png)

Possible New "Loading..." Text (with "Cancel" option)
![image](https://user-images.githubusercontent.com/20442222/170777921-af082c5d-b0ad-4c57-ab0b-7de2df617c3e.png)

PR #7986, opened 2022-06-01
[Feature:Submission] Cancel Loading Test Case

MBon1:
- Feature #7958 .  
  - Currently, when selecting to "Show Details" on a gradable's test case, the user must wait until the output is displayed.  There is no way to cancel this load request.  

### What is the new behavior?
After selecting "Show Details" on a test case, the user will be able to click "Cancel Loading" to abort the loading of the test case's output.  After the load is cancelled, the user will again be given the option to "Show Details" for that test case.  Aborting the loading of one test case will not affect any other test cases.  

### Other information?
As a part of my testing, using various network speeds (modified through the inspector), I began loading and then cancelled a single test case.  Through the inspector's network tab, I was able to see that the request had been aborted.  I then did this with two test cases, one with only a single test case being aborted and another where both test cases were aborted.  Lastly, I tried rapidly selecting "Show Details" and "Cancel Loading" repeatedly on a single test case to try to get incorrect text to display and incorrect functionality to occur.  These tests were done both on Google Chrome and Microsoft Edge and produced the expected behavior on both.


site/public/js/server.js
```diff
@@ -29,6 +29,14 @@ function adjustBreadcrumbLinks(){
 }
 ////////////End: Removed redundant link in breadcrumbs//////////////////////////
 
+/* GLOBAL VARIABLES */
+
+// Map to store the XMLHttpRequest object returned by getJSON when getting the 
+// output to display in loadTestcaseOutput()
+// Key: orig_div_name   Value: XMLHttpRequest object
+var output_xhrs = new Map();
```
williamjallen, line 37: Why do we need a global variable here?  Global variables are generally a very very very bad thing in terms of future development.

site/public/js/server.js
```diff
@@ -133,33 +141,79 @@ function changeDiffView(div_name, gradeable_id, who_id, version, index, autochec
...
+/**
+ * Collapses Test Case Output.  
+ *
+ * @param {string} div_name - id of div containing the output of the test case
+ * @param {string} index - index of test case
+ * @param {object} loadingTools - loading tools used by span containing the text that 
+ *                                correspond to the state of the test case (show, hide, loading)
+ */
+function CollapseTestcaseOutput(div_name, index, loadingTools)
+{
```
williamjallen, line 154: The opening curly brace should be on the same line as the function declaration, as per the Submitty [JS style guide](https://submitty.org/developer/software_and_system_design/coding_style_guide/javascript).  This file is not linted by `eslint`, but the same rules should be applied when possible.  The specific eslint rule we use is the `brace-style` [rule](https://eslint.org/docs/rules/brace-style).

site/public/js/server.js
```diff
@@ -133,33 +141,79 @@ function changeDiffView(div_name, gradeable_id, who_id, version, index, autochec
...
-        toggleDiv(orig_div_name);
-
-        loadingTools.find("span").hide();
-        loadingTools.find(".loading-tools-show").show();
+    // Checks if output for this div is being loaded
+    if(output_xhrs.has(orig_div_name))
+    {
+        // Checks if xhr is defined and has not succeeded yet
+        // If so, abort loading
+        var xhr = output_xhrs.get(orig_div_name);
```
williamjallen, line 186: Since `xhr` isn't going to be reassigned, `const` can be used instead of `let`.  `var` should always be avoided in JS code.
```suggestion
        const xhr = output_xhrs.get(orig_div_name);
```

site/public/js/server.js
```diff
@@ -133,33 +141,79 @@ function changeDiffView(div_name, gradeable_id, who_id, version, index, autochec
...
-    if($(div_name).is(":visible")){
-        $("#show_char_"+index).toggle();
-        $(div_name).empty();
-        toggleDiv(orig_div_name);
-
-        loadingTools.find("span").hide();
-        loadingTools.find(".loading-tools-show").show();
+    // Checks if output for this div is being loaded
+    if(output_xhrs.has(orig_div_name))
+    {
```
williamjallen, line 183: The same comment about brace style applies here.

site/public/js/server.js
```diff
@@ -133,33 +141,79 @@ function changeDiffView(div_name, gradeable_id, who_id, version, index, autochec
...
+        // Checks if xhr is defined and has not succeeded yet
+        // If so, abort loading
+        var xhr = output_xhrs.get(orig_div_name);
+        if(xhr && xhr.readyState != 4){
```
williamjallen, line 187: It is generally better to use the [strict equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality) operators in JS code to prevent unexpected behavior as a result of the default JS type conversion.
```suggestion
        if(xhr && xhr.readyState !== 4){
```

site/public/js/server.js
```diff
@@ -133,33 +141,79 @@ function changeDiffView(div_name, gradeable_id, who_id, version, index, autochec
...
-    else{
+    // Checks if test case output is visible
+    // Output is visible - Collapse test case output
+    if($(div_name).is(":visible"))
+    {
+        CollapseTestcaseOutput(orig_div_name, index, loadingTools);
+    }
+    // Output is not visible - Expand test case output
+    else
+    {
```
williamjallen, line 200: Same comments here about brace style

site/public/js/server.js
```diff
@@ -169,11 +223,19 @@ function loadTestcaseOutput(div_name, gradeable_id, who_id, index, version = '')
...
-                console.log(e);
-                displayAjaxError(e);
+                if(output_xhrs.has(orig_div_name) && output_xhrs.get(orig_div_name).readyState == 0){
+                    output_xhrs.delete(orig_div_name);
+                    CollapseTestcaseOutput(orig_div_name, index, loadingTools);
+                    console.log("JSON Load Aborted");
+                }
+                else
+                {
+                    alert("Could not load diff, please refresh the page and try again.");
```
williamjallen, line 233: Same comment about brace style

site/public/js/server.js
```diff
@@ -169,11 +223,19 @@ function loadTestcaseOutput(div_name, gradeable_id, who_id, index, version = '')
...
+                    console.log(e);
+                    displayAjaxError(e);
+                }
             }
-        })
+        }))
```
williamjallen, line 238:
```suggestion
        }));
```

site/app/templates/autograding/AutoResults.twig
```diff
@@ -64,15 +64,15 @@
...
                                 <span class="loading-tools-hide" style="color: var(--standard-deep-blue); text-decoration: underline;" hidden>
                                     Hide Details
                                 </span>
                                 <span class="loading-tools-show" style="color: var(--standard-deep-blue); text-decoration: underline;">
                                     Show Details
                                 </span>
-                                <span class="loading-tools-in-progress" aria-label="Loading Details for {{testcase.name}}" hidden>
-                                    <i class="fas -notch fa-spin save-button"></i>Loading...
+                                <span class="loading-tools-in-progress" style="text-decoration: underline;" aria-label="Loading Details for {{testcase.name}}" hidden>
+                                    <i class="fas -notch fa-spin save-button"></i>Cancel Loading</u>
```
williamjallen, line 75: I might have missed it but I don't see an opening `<u>` tag to match the closing tag here.  Is this an intentional change?

site/public/js/server.js
```diff
@@ -169,11 +223,19 @@ function loadTestcaseOutput(div_name, gradeable_id, who_id, index, version = '')
                 enableKeyToClick();
             },
             error: function(e) {
-                alert("Could not load diff, please refresh the page and try again.");
-                console.log(e);
-                displayAjaxError(e);
+                if(output_xhrs.has(orig_div_name) && output_xhrs.get(orig_div_name).readyState == 0){
```
williamjallen, line 226: Same comment about strict equality operator here
```suggestion
                if(output_xhrs.has(orig_div_name) && output_xhrs.get(orig_div_name).readyState === 0){
```

williamjallen (changes requested):

site/app/templates/autograding/AutoResults.twig
MBon1: This was an unintentional change.  The `<\u>` tag has been removed.

site/public/js/server.js
MBon1: Opening curly brace has been moved to the end of the preceding line.

bmcutler:
@williamjallen @cjreed121 
Can one of you please re-review this so we can merge it soon, thanks

site/public/js/server.js
```diff
@@ -150,14 +150,14 @@ function changeDiffView(div_name, gradeable_id, who_id, version, index, autochec
  * @param {object} loadingTools - loading tools used by span containing the text that 
  *                                correspond to the state of the test case (show, hide, loading)
  */
-function CollapseTestcaseOutput(div_name, index, loadingTools){
-    $("#show_char_" + index).toggle();
-    $("#" + div_name).empty();
-    toggleDiv(div_name);
+// function CollapseTestcaseOutput(div_name, index, loadingTools){
```
williamjallen, line 153: Is there a reason why these are commented out?

MBon1: I made this commit to show someone my code, so the commented out section was a byproduct of my work flow.  It was  removed in my latest commit.

site/public/js/server.js
MBon1: The global variable and functions using it have been moved to their own JavaScript file.

site/public/js/testcase-output.js
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Collapses Test Case Output.
+ *
+ * @param {string} div_name - id of div containing the output of the test case
+ * @param {string} index - index of test case
+ * @param {object} loadingTools - loading tools used by span containing the text that
+ *                                correspond to the state of the test case (show, hide, loading)
+ */
+function CollapseTestcaseOutput(div_name, index, loadingTools) {
+    $(`show_char_${index}`).toggle();
```
williamjallen, line 15: Is this supposed to be selecting a class/id?

williamjallen (changes requested):

site/public/js/testcase-output.js
MBon1: Yes.  The `$(selector)` is the jQuery Selector Syntax for selecting HTML elements.

site/public/js/testcase-output.js
```diff
@@ -0,0 +1,102 @@
+// Map to store the XMLHttpRequest object returned by getJSON when getting the
+// output to display in loadTestcaseOutput()
+// Key: orig_div_name   Value: XMLHttpRequest object
+const output_xhrs = new Map();
```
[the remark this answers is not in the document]
MBon1: I can change it to use `loading_testcases_xml_http_requests` instead of `output_xhrs` .

site/public/js/testcase-output.js
```diff
@@ -0,0 +1,102 @@
...
+    if ($(div_name).is(':visible')) {
+        CollapseTestcaseOutput(orig_div_name, index, loadingTools);
+    }
+    // Output is not visible - Expand test case output
+    else {
+        $(`#show_char_${index}`).toggle();
+        // eslint-disable-next-line no-undef
+        const url = `${buildCourseUrl(['gradeable', gradeable_id, 'grading', 'student_output'])}?who_id=${who_id}&index=${index}&version=${version}`;
+
+        loadingTools.find('span').hide();
```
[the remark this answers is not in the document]
MBon1: I can make this change, but for my own understanding, what would be the advantage of searching for elements by their classes rather than searching for `span` ?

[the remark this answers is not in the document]
williamjallen: Selecting something by its class is much more specific than searching for a generic `span` tag.  On this page, there are probably dozens or hundreds of `span` tags overall.  What if someone in the future came along and added a second `span` tag to the parent of the `span` you are selecting?  It would probably break your code, and/or cause a headache for that future developer, since they may need to make their own changes in addition to re-writing your code.

site/public/js/testcase-output.js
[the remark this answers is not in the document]
williamjallen: Sure

site/public/js/testcase-output.js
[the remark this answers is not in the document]
MBon1: Okay.  That makes sense.  Thank you.

site/public/js/testcase-output.js
```diff
@@ -0,0 +1,103 @@
...
+                enableKeyToClick();
+            },
+            error: function(e) {
+                if (loading_testcases_xml_http_requests.has(orig_div_name) && loading_testcases_xml_http_requests.get(orig_div_name).readyState === 0) {
+                    loading_testcases_xml_http_requests.delete(orig_div_name);
+                    CollapseTestcaseOutput(orig_div_name, index, loadingTools);
+                    console.log('JSON Load Aborted');
+                }
+                else {
+                    alert('Could not load diff, please refresh the page and try again.');
```
poorna2152, line 95: If there is a case where this `else` path gets executed wouldn't you want to call the `ColllapseTestcaseOutput` function and also delete the `request` from the map if it exists. I thought that the deleting request and call to `CollapseTestcaseoutput` should be done outside `if-else` statement.

MBon1: If this else is reached, then the test case could not be loaded due to an error and the page would need to be refreshed. Since the page will be refreshed, we would not need to remove the request from the map since when the page is changed or refreshed, global objects do not persist, so it's only relevant to the current page.

poorna2152: Ah yes that makes sense

poorna2152 (approved):
Tested the changes by merging main (didnt push it). Changes work as expected.

bmcutler:
@williamjallen please rereview. thanks :)

williamjallen (approved):
Code looks good and functions as expected.  My only minor nitpick is that the text is black, which might not make it clear to the user that it is clickable.

(p.s.: sorry about the late re-review--I took last week off to recover from some burnout and I wanted to give it a thorough review before approving)

PR #7997, opened 2022-06-03
[Bugfix:System] autoupdate Python dependencies

shailpatels:

* [ ] Documentation has been updated/added if relevant (Will do after review)

Currently, pip installation that is done in the initial Submitty installation or vagrant up are only ever done once via `install_system.sh`.

### What is the new behavior?
The pip install has been moved out and into their own file called `update_system.sh` which is now invoked with the `INSTALL_SUBMITTY` script 

### Other information?
Currently, I only moved over the pip installation, will look into OS dependencies next and then specific installs like DrMemory, Haskell Stack - etc.


bmcutler (commented):
On M1/ARM I had to comment out 3 lines in .setup/pip/system_requirements.txt:
(these exact versions are not available for ARM)

#opencv-python==3.4.9.33
#onnx==1.9.0
#onnxruntime==1.3.0

Then it worked ok.

PR #8000, opened 2022-06-04
[Bugfix:InstructorUI] Fixed PDF Annotation Toolbar

### LINKED ISSUE #7323: PDF Annotation Toolbar Breaks on Horizontal Scroll
**Describe the bug**
While being zoomed in on a PDF in the grading interface, when trying to scroll to the right, the toolbar doesn't stay at the top as expected which makes some of the buttons no longer accessible (see screenshots)

before scrolling:
pdf_anontate_before

after scrolling:
pdf_anotate_after

immortalcodes:
 Closes #7323:While being zoomed in on a PDF in the grading interface, when trying to scroll to the right,the tools available gets hidden and inaccessible.  

### What is the new behavior?
Zoomed in on a PDF in the grading interface, when trying to scroll to the right, the position of the toolbar doesn't change and hence the tools don't hide and are always available.
### Other information?
- [x] Tested locally


akshatbatra:
Annotation toolbar now overlays the document, hiding some of its contents along with hiding the top portion of vertical scrollbar.

immortalcodes:
It doesn't hide the document unless you zoom it and scroll down and then to show some part , some part will be hidden. Am I getting something wrong ? May be u could attach some screenshots for the same.

akshatbatra:
Submitty-panel

immortalcodes:
@akshatbatra I have done some changes , could you verify if the issue is happening again?

akshatbatra (commented):
Does the added function run only once (on page load) ? If that's the case then resizing the window can lead to unexpected behavior, you can fix it by adding a resize event listener. But if that's not the case then the changes seem good.

akshatbatra (commented):
`mouseover` is a working hack here. I think there is an extra `$( ... )` enclosing the event listener.  I think most browsers don't support `onresize` attribute for all HTML elements, it's only supported for `window` object. You can get rid of the `onresize`. if you want to apply a fix that aligns more with the problem, then you can go with `ResizeObserver`. But IE isn't happy with that.

site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
```diff
@@ -53,4 +53,11 @@
             $('#pdf-buttons').append(document.getElementById('download-annotations-btn-template').content.cloneNode(true));
         }
     });
+
+    $($('#submission_browser').on("mouseover",function addOffset(){
```
akshatbatra, line 57:
```suggestion
    const offSetter = new ResizeObserver(entries => {
        entries.forEach(entry => {
            let height = $(".sticky-file-info").outerHeight();
            $("#viewer").css("paddingTop", height);
        });
    });
    offSetter.observe(document.querySelector(".sticky-file-info"));
```

williamjallen (commented):
I'm a little confused as to why this PR is using JS at all.  Unless I'm misunderstanding the issue, I believe that it can be resolved entirely with CSS styling.

immortalcodes:
I tried doing it with CSS right from the start, but there were a lot of problems. I implemented a lot of solutions but each one had it's own issue when tested with resizing of the window, different resolutions or zooming(in/out) the pdf . Sometimes the toolbar overlapped the thing or sometimes the resizing created problems.
So I used JS to add a dynamic offset for the annotation pdf which adjusts itself on resizing or change of screen resolution.
I think I have tried most possible options but then still if there are some more suggestions I can give it a shot.

Eceptonsu (commented):
Works for me.

PR #8002, opened 2022-06-05
[SECURITY][Feature:TAGrading] Unique ids for different gradeables

### LINKED ISSUE #7569: Student anonymous id's are not unique across gradeables
This issue is mainly related to PR #7195. In order to fix the issue of incorrect checkpoint and numeric gradeable socket updates, student anon_id's had to be sent over the sockets. The anonymous ID's tied to each student are generated one time and used for each gradeable.

To make it more secure it might be better to generate a new unique ID for each student per gradeable.

akshatbatra:
Fixes #7569, the anonymous id tied to each student is generated one time and the same is used for every gradeable.

### What is the new behavior?
To make it more secure, anonymous ids are now unique across gradeables.

Credit (guidance/suggestions): @cjreed121, @Eceptonsu


cjreed121 (changes requested):
This does seem to work but I think the change we are looking for is more than just simple grading. We probably want unique anon ids across all gradeables (including electronic).

If you look in the `users` table in one of the course databases you will see each student has one specific anonymous ID. Maybe instead we can create a database table where each row defines a user_id, gradeable, and anonymous_id with a unique index of user_id and gradeable.

migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
```diff
@@ -0,0 +1,34 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("CREATE TABLE IF NOT EXISTS gradeable_anon (user_id character varying NOT NULL, g_id character varying(255) NOT NULL, anon_id character varying(255));")
```
cjreed121, line 17: Can you add foreign keys for the user id and gradeable id

site/app/controllers/AbstractController.php
```diff
@@ -129,7 +129,7 @@ protected function tryGetMark(Component $component, string $mark_id, bool $rende
      * @param bool $render_json true to render a JSEND response to the output in the failure/error case
      * @return string|bool false in the fail/error case
      */
-    protected function tryGetSubmitterIdFromAnonId(string $anon_id, bool $render_json = true) {
+    protected function tryGetSubmitterIdFromAnonId(string $anon_id, string $g_id = null, bool $render_json = true) {
```
cjreed121, line 132: Would there be a case where we want g_id as null?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2820,37 +2820,61 @@ public function getAllSectionsForGradeable($gradeable) {
...
+                    JOIN (
+                        SELECT gd.gd_{$type}_id, gcmd.gcm_id
+                        FROM gradeable_component_mark_data AS gcmd
+                            JOIN gradeable_data gd ON gd.gd_id=gcmd.gd_id
+                    ) as gcmd
+                    ON gcmd.gd_{$type}_id=u.{$type}_id
+                WHERE gcmd.gcm_id = ?",
+                $params
+            );
+        }
```
cjreed121, line 2875: Since these queries are quite similar, you could probably just concatenate strings to form this query to reduce the amount of duplicated code. There are a couple other queries in this file that do something like that.

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1556,7 +1558,7 @@ public function renderPeerPanel(GradedGradeable $graded_gradeable, int $display_
         return $this->core->getOutput()->renderTwigTemplate("grading/electronic/PeerPanel.twig", [
                 "gradeable_id" => $gradeable->getId(),
                 "is_ta_grading" => $gradeable->isTaGrading(),
-                "anon_id" => $graded_gradeable->getSubmitter()->getAnonId(),
+                "anon_id" => $graded_gradeable->getSubmitter()->isTeam() ? $graded_gradeable->getSubmitter()->getAnonId() : $graded_gradeable->getSubmitter()->getAnonId($graded_gradeable->getGradeableId()),
```
cjreed121, line 1561: This ternary operator of if its a team do 2 separate things seems to be used quite often. Could we perhaps add a function in the Submitter class to get the respective team or anon id based on which it is to reduce these a bit?

cjreed121 (changes requested):

migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
```diff
@@ -0,0 +1,62 @@
...
+                REFERENCES gradeable(g_id)
+                ON UPDATE CASCADE
+        );
+        """
+    )
+    database_file = config.config_path / 'database.json'
+    with(database_file).open('r') as db_file:
+        db_info = json.load(db_file, object_pairs_hook=OrderedDict)
+        database.execute("GRANT SELECT, INSERT, UPDATE, DELETE ON gradeable_anon TO {}".format(db_info['database_course_user']))
+
```
cjreed121, line 37: Ideally you shouldn't have to do this but I know it won't work otherwise. I just made an issue https://github.com/Submitty/Submitty/issues/8209 to track this. This is definitely correct for now but if that issue gets resolved before this goes in then I'd remove it.

migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
akshatbatra: Done

site/app/controllers/AbstractController.php
akshatbatra: Currently, there is no such place, so, default value has been removed.

site/app/libraries/database/DatabaseQueries.php
akshatbatra: String concatenation is used now.

site/app/views/grading/ElectronicGraderView.php
akshatbatra: A check inside `Submitter::getAnonId()` was already added, so all these ternary operators have been removed and replaced with a call similar to `$graded_gradeable->getSubmitter()->getAnonId($graded_gradeable->getGradeableId())`.

migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
akshatbatra: Okay :thumbsup:

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -4417,14 +4424,55 @@ public function mergeThread($parent_thread_id, $child_thread_id, &$message, &$ch
...
+                $this->core->getQueries()->getUserById($user_id)->getAnonId($g_id);
+                continue;
+            }
+            try {
+                $params = [$user_id, $g_id, $anon_id];
+                $this->course_db->query("INSERT INTO gradeable_anon(user_id, g_id, anon_id) VALUES (?, ?, ?)", $params);
+                $this->course_db->commit();
+            }
+            catch (DatabaseException $dbException) {
+                $this->course_db->rollback();
```
cjreed121, line 4447: Why commit and possibly do a rollback if you never start a transaction?

site/app/models/User.php
```diff
@@ -461,10 +454,19 @@ public function setGradingRegistrationSections($sections) {
...
+    public function getAnonId($g_id) {
+        if ($g_id === "") {
+            return "";
+        }
+        $anon_id = $this->core->getQueries()->getAnonId($this->id, $g_id);
+        $anon_id = empty($anon_id) ? null : $anon_id[$this->getId()];
+        if ($anon_id === null) {
             $alpha = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890";
-            $anon_ids = $this->core->getQueries()->getAllAnonIds();
+            $anon_ids = $this->core->getQueries()->getAllAnonIdsByGradeable();
```
cjreed121, line 469: Why not pass $g_id into this function?

cjreed121 (changes requested):
Also take a look at the `sbin/anonymize.py` script and make sure it works with these new changes.

site/app/libraries/database/DatabaseQueries.php
akshatbatra: Oh, it was left there when I copied the template from one of the functions above. This has been removed.

site/app/models/User.php
akshatbatra: If `$g_id` isn't passed then all `anon_ids` in the table are returned. I intended to keep the `anon_ids` unique across gradeables, but thinking about it again I don't think it is necessary, it's just increasing the number of ids to loop over and thus increasing the computation time, so it has now been modified to send `$g_id` too.

bmcutler:
@cjreed121 @akshatbatra 
anonymize.py should just be removed.  That was an old script from 2017 (renamed in 2018) when we first added peer grading.  It is not needed or used and is just clutter.

migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
```diff
@@ -0,0 +1,62 @@
...
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute(
+        """
+        CREATE TABLE IF NOT EXISTS gradeable_anon (
+            user_id character varying NOT NULL,
+            g_id character varying(255) NOT NULL,
+            anon_id character varying(255),
+            FOREIGN KEY(user_id)
```
cjreed121, line 24: Should we also add a UNIQUE key here of g_id and anon_d? I know one of the scripts makes sure to not repeat an anon_id but I don't see how it could be harmful to make our database also guarantee it.

cjreed121 (commented):
Looks good to me. Just 1 last comment.

migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
akshatbatra: I don't think `g_id` is unique, all the users in `gradeable_anon` have same `g_id`. Also, after modifying User.php to send `$g_id` i.e. `$anon_ids = $this->core->getQueries()->getAllAnonIdsByGradeable($g_id)`. `anon_id` is not guaranteed to be unique when whole table is considered, it is only unique in the context of a particular gradeable, practically it can be expected to be unique across the table but not theoretically.

cjreed121: I was referring that there can be a unique key/constraint consisting of BOTH g_id and anon_id. I know there can definitely be multiple g_id's and possibly multiple anon_id's but there should never be a multiple rows where the g_id and anon_id are both equal.

cjreed121: Here is an example of this https://github.com/Submitty/Submitty/blob/2590ba5ea747dda50b5e1f7e862de9f582292e72/migration/migrator/data/course_tables.sql#L1853

akshatbatra: Oh, okay, my bad, will do it.

cjreed121 (approved):

bmcutler:
@akshatbatra 
Please fix merge conflicts

williamjallen (commented):
Given that `sbin/anonymize.py` has been deleted, the corresponding entry in `.flake8` should be deleted as well.

Disregard the previous review.  The GitHub interface appears to be a little glitchy with its handling of saved text in the review comment box...

site/app/controllers/pdf/PDFController.php
```diff
@@ -172,13 +172,13 @@ public function savePDFAnnotation(string $gradeable_id, string $target_dir) {
...
 
-    public function getAnonPath($file_path) {
+    public function getAnonPath($file_path, $g_id) {
         $file_path_parts = explode("/", $file_path);
         $anon_path = "";
         for ($index = 1; $index < count($file_path_parts); $index++) {
             if ($index === 9) {
                 $user_id = $file_path_parts[$index];
-                $anon_ids = $this->core->getQueries()->getUserFromAnon($user_id);
+                $anon_ids = $this->core->getQueries()->getUserFromAnon($user_id, $g_id);
```
Eceptonsu, line 181: Should we consider anon team id here

akshatbatra: Looking at some related code, it feels like it should be there. Surprising that no unexpected behavior was ever reported due to this on production machine.

Eceptonsu (approved):

PR #8012, opened 2022-06-07
[Bugfix:Submission] Download Annotated PDF

Eceptonsu:
Duplicates of #6837 
When viewing your grade for a git gradeable that has an annotated pdf, you can view it but not download it.
After downloading the annotated pdf in the grading interface,  the user's annotations will disappear after clicking the show/hide all annotation button.

### What is the new behavior?
Annotated pdfs on a git gradeable can now be viewed and downloaded. 
The download button when annotating pdf gradeables is now hidden for peer graders.
Users' annotations will always be there no matter what actions they take.

### Other information?
Some fixes were based on PR#6837 by eddiekrystowski


tkoz0 (commented):
I'm noticing some buggy behavior in the grading interface. As instructor in the sample course, do the following steps:
1. Open the "Grading Homework PDF Grading" assignment and open the PDF file for a student's submission
2. Make an annotation with the pen tool and save it
3. Click download to open the new window with the annotated PDF, then close it
4. Click hide other annotations and then click show all annotations. The instructor's annotations disappear.
I would expect the instructor's annotations to stay visible here, but they disappear. If you go back to the file list and open the student's PDF again, the annotations reappear.

Eceptonsu (commented):
Thanks for pointing it out, I fixed the issue in the latest commit. This PR is also related to downloading the annotated PDF on the submission page for VCS submissions, hope that part works fine.

tkoz0 (approved):
I have tested the changes and it looks like everything works as expected.

PR #8014, opened 2022-06-09
[Bugfix:Forum] Validate Forum Thread IDs

### LINKED ISSUE #7955: Discussion Forum -- error in /var/local/submitty/logs/site_errors
This error was found in the site_errors log.
I don't have a report from a user about a problem, so I can't say how exactly it was produced.

-----------------------

00:09:15 05/24/2022 - FATAL ERROR
app\exceptions\DatabaseException (Code: 0) thrown in /usr/local/submitty/site/app/libraries/database/AbstractDatabase.php (Line 188) by:
            throw new DatabaseException($pdoException->getMessage(), $query, $parameters);

Message:
SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type integer: ""

Stack Trace:
#0 /usr/local/submitty/site/app/libraries/database/DatabaseQueries.php(794): app\libraries\database\AbstractDatabase->query()
#1 /usr/local/submitty/site/app/views/forum/ForumThreadView.php(713): app\libraries\database\DatabaseQueries->getFirstPostForThread()
#2 /usr/local/submitty/site/app/views/forum/ForumThreadView.php(607): app\views\forum\ForumThreadView->displayThreadList()
#3 unknown file(unknown line): app\views\forum\ForumThreadView->showAlteredDisplayList()
#4 /usr/local/submitty/site/app/libraries/Output.php(228): call_user_func_array()
#5 /usr/local/submitty/site/app/controllers/forum/ForumController.php(927): app\libraries\Output->renderTemplate()
#6 unknown file(unknown line): app\controllers\forum\ForumController->getSingleThread()
#7 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(241): call_user_func_array()
#8 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(219): app\libraries\routers\WebRouter->run()
#9 /usr/local/submitty/site/public/index.php(131): app\libraries\routers\WebRouter::getWebResponse()
Extra Details:
	query: SELECT * FROM posts WHERE parent_id = -1 AND thread_id = ?
	parameters:
		

URL: https://submitty.cs.rpi.edu//courses/u22/csci2600/forum/threads/single

MBon1:

- Fixes #7955 
  - If a thread is queried in the Discussion Form using an invalid (a null or non-integer value) or non-existent thread id, a server error would be displayed prompting for the page to be refreshed.

### What is the new behavior?
Now, the thread id is checked when a thread is being retrieved to make sure that it and its corresponding thread exists.  If the thread id is invalid or the thread does not exist, a "fail" JSON response case will be rendered that says that the id is not valid.  

### Other information?

I had two pages opened to the discussion forum.  One of these pages had the Inspector Console panel opened and the other with with the Inspector Network panel  opened.  In the Console, the following commands were entered: 
  - Commands that Rendered a "successful" JSON response case
    - window.socketClient.send({type: "new_thread", thread_id: "1"})
    - window.socketClient.send({type: "new_thread", thread_id: 3})
  - Commands that Rendered a "fail" JSON response case
    - window.socketClient.send({type: "new_thread", thread_id: "100"})
    - window.socketClient.send({type: "new_thread", thread_id: ""})
    - window.socketClient.send({type: "new_thread", thread_id: "thread id 1"})

After each command was entered, I checked the page with the Inspector Network panel opened to see if the "Error parsing new post" message was displayed onto the page and the command's corresponding request in the Inspector Network panel.  If a valid thread id was used in the command (the two commands under "Commands that Rendered a 'successful' JSON response case"), I checked if the request's preview had a line saying that the status was "success".  If an invalid or non-existent thread id was used (the commands under "Commands that Rendered a 'fail' JSON response case"), I check to see that the "Error parsing new post" message does not display and a line saying that the status was "fail" in the command's corresponding request's preview.

### DIFF
--- a/site/app/controllers/forum/ForumController.php
+++ b/site/app/controllers/forum/ForumController.php
@@ -916,6 +916,18 @@ class ForumController extends AbstractController {
      * @Route("/courses/{_semester}/{_course}/forum/threads/single", methods={"POST"})
      */
     public function getSingleThread() {
+        // Checks if thread id is null. If so, render "fail" json response case informing that thread id is null.
+        if ($_POST['thread_id'] == null) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NULL ID)", $_POST['thread_id']);
+        }
+        // Checks if thread id is not an integer value. If so, render "fail" json response case informing that thread id is not an integer value.
+        if (!ctype_digit($_POST['thread_id'])) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NON-INTEGER ID)", $_POST['thread_id']);
+        }
+        // Checks if thread id does not exist. If so, render "fail" json response case informing that the thread does not exist.
+        if (!$this->core->getQueries()->existsThread($_POST['thread_id'])) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NON-EXISTANT ID)", $_POST['thread_id']);
+        }
         $thread_id = $_POST['thread_id'];
         $thread = $this->core->getQueries()->getThread($thread_id);
         $categories_ids = $this->core->getQueries()->getCategoriesIdForThread($thread_id);
--- a/site/app/controllers/forum/ForumController.php
+++ b/site/app/controllers/forum/ForumController.php
@@ -916,24 +916,20 @@ class ForumController extends AbstractController {
      * @Route("/courses/{_semester}/{_course}/forum/threads/single", methods={"POST"})
      */
     public function getSingleThread() {
-        // Checks if thread id is undefined. If so, render "fail" json response case informing that thread id is not set.
-        if (!isset($_POST['thread_id'])) {
-            return $this->core->getOutput()->renderJsonFail("Invalid thread id (UNSET ID)");
-        }
-        // Checks if thread id is null. If so, render "fail" json response case informing that thread id is null.
-        if ($_POST['thread_id'] == null) {
-            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NULL ID)");
+        $thread_id = $_POST['thread_id'];
+        // Checks if thread id is empty. If so, render "fail" json response case informing that thread id is empty.
+        if (empty($thread_id)) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (EMPTY ID)");
         }
         // Checks if thread id is not an integer value. If so, render "fail" json response case informing that thread id is not an integer value.
-        if (!(is_int($_POST['thread_id']) || ctype_digit($_POST['thread_id']))) {
-            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NON-INTEGER)");
+        if (!(is_int($thread_id) || ctype_digit($_POST['thread_id']))) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NON-INTEGER ID)");
         }
-        // Checks if thread id does not exist. If so, render "fail" json response case informing that the thread does not exist.
-        if (!$this->core->getQueries()->existsThread($_POST['thread_id'])) {
+        $thread = $this->core->getQueries()->getThread($thread_id);
+        // Checks if no threads were found. If so, render "fail" json response case informing that the no threads were found with the given ID.
+        if (!(count($thread) > 0)) {
             return $this->core->getOutput()->renderJsonFail("Invalid thread id (NON-EXISTENT ID)");
         }
-        $thread_id = $_POST['thread_id'];
-        $thread = $this->core->getQueries()->getThread($thread_id);
         $categories_ids = $this->core->getQueries()->getCategoriesIdForThread($thread_id);
         $show_deleted = $this->showDeleted();
         $currentCourse = $this->core->getConfig()->getCourse();
--- a/site/app/controllers/forum/ForumController.php
+++ b/site/app/controllers/forum/ForumController.php
@@ -918,15 +918,15 @@ class ForumController extends AbstractController {
     public function getSingleThread() {
         // Checks if thread id is null. If so, render "fail" json response case informing that thread id is null.
         if ($_POST['thread_id'] == null) {
-            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NULL ID)", $_POST['thread_id']);
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NULL ID)");
         }
         // Checks if thread id is not an integer value. If so, render "fail" json response case informing that thread id is not an integer value.
-        if (!ctype_digit($_POST['thread_id'])) {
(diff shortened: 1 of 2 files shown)


site/app/controllers/forum/ForumController.php
```diff
@@ -916,6 +916,18 @@ public function getThreads($page_number = null) {
      * @Route("/courses/{_semester}/{_course}/forum/threads/single", methods={"POST"})
      */
     public function getSingleThread() {
+        // Checks if thread id is null. If so, render "fail" json response case informing that thread id is null.
+        if ($_POST['thread_id'] == null) {
```
akshatbatra, line 920: I've been told that using === is always better.

site/app/controllers/forum/ForumController.php
```diff
@@ -916,6 +916,18 @@ public function getThreads($page_number = null) {
...
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NULL ID)", $_POST['thread_id']);
+        }
+        // Checks if thread id is not an integer value. If so, render "fail" json response case informing that thread id is not an integer value.
+        if (!ctype_digit($_POST['thread_id'])) {
```
akshatbatra, line 924: Should there be !(is_int($_POST['thread_id']) || ctype_digit($_POST['thread_id'])) ?

site/app/controllers/forum/ForumController.php
```diff
@@ -916,6 +916,18 @@ public function getThreads($page_number = null) {
...
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NON-INTEGER ID)", $_POST['thread_id']);
+        }
+        // Checks if thread id does not exist. If so, render "fail" json response case informing that the thread does not exist.
+        if (!$this->core->getQueries()->existsThread($_POST['thread_id'])) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NON-EXISTANT ID)", $_POST['thread_id']);
```
akshatbatra, line 929: EXISTENT*

akshatbatra (changes requested):
You may also want to have a look at `showThreads($thread_id = null, $option = 'tree')`, it also isn't handling non-existent thread ids gracefully.

site/app/views/forum/ForumThreadView.php
```diff
@@ -710,6 +710,10 @@ public function displayThreadList($threads, $filtering, &$activeThreadAnnounceme
         $thread_content = [];
 
         foreach ($threads as $thread) {
+            // Checks if thread ID is null. If so, skip this threads.
+            if ($thread["id"] == null) {
```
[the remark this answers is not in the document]
MBon1: Changed `==` to `===`.

MBon1:
> Also, do we really need to return $_POST['thread_id'] as error data? I would rather avoid returning a malformed string as it is to the client.

That makes sense.  I guess return $_POST['thread_id'] would be extraneous information for the client.

site/app/controllers/forum/ForumController.php
MBon1: While testing, it looks like if I used `===` instead of `==`, the null value would get caught when when the function checks if the thread id is an integer, so I think it might be better to use `==` in this case.

MBon1: I also added a check to see if the thread id is unset using `isset`.  Interestingly, `isset` did not catch either of the aforementioned thread ids used when testing for null.

cjreed121: You could look into using the `empty` function

site/app/controllers/forum/ForumController.php
```diff
@@ -916,6 +916,22 @@ public function getThreads($page_number = null) {
...
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NULL ID)");
+        }
+        // Checks if thread id is not an integer value. If so, render "fail" json response case informing that thread id is not an integer value.
+        if (!(is_int($_POST['thread_id']) || ctype_digit($_POST['thread_id']))) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NON-INTEGER)");
+        }
+        // Checks if thread id does not exist. If so, render "fail" json response case informing that the thread does not exist.
+        if (!$this->core->getQueries()->existsThread($_POST['thread_id'])) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NON-EXISTENT ID)");
+        }
```
cjreed121, line 934: You could probably just check the result of the below getThread function call rather than doing this.

cjreed121 (changes requested):

site/app/controllers/forum/ForumController.php
MBon1: I looked into `empty`, and it handles of both `$_POST['thread_id']) == null` and  `!isset($_POST['thread_id']))`.  I'll change it to use `empty`.  Thank you.

site/app/controllers/forum/ForumController.php
```diff
@@ -916,6 +916,22 @@ public function getThreads($page_number = null) {
...
+        // Checks if thread id is undefined. If so, render "fail" json response case informing that thread id is not set.
+        if (!isset($_POST['thread_id'])) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (UNSET ID)");
+        }
+        // Checks if thread id is null. If so, render "fail" json response case informing that thread id is null.
+        if ($_POST['thread_id'] == null) {
+            return $this->core->getOutput()->renderJsonFail("Invalid thread id (NULL ID)");
+        }
+        // Checks if thread id is not an integer value. If so, render "fail" json response case informing that thread id is not an integer value.
+        if (!(is_int($_POST['thread_id']) || ctype_digit($_POST['thread_id']))) {
```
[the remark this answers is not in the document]
MBon1: I'm torn with using only `is_int`.  I'm not exactly sure if it can be assume that `$_Post`  should only contain integers since strings composed of only numeric characters can currently be used interchangeably with their corresponding integer equivalent (i.e. using `"1"` vs `1`) when querying for a thread.

site/app/controllers/forum/ForumController.php
MBon1: Okay.  I'll use `!(count($thread) > 1)` instead of `!$this->core->getQueries()->existsThread($thread_id)`.  Thank you.

site/app/views/forum/ForumThreadView.php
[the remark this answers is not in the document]
MBon1: I decided to use `empty` instead to check if the id is also set.

cjreed121 (approved):

bmcutler:
@akshatbatra 
Can you re-review and let us know if you approve?  Thanks!

akshatbatra (approved):

akshatbatra:
Things look good now.

PR #8018, opened 2022-06-09
[Bugfix:TAGrading] Annotated PDF Instructor/Student View

Eceptonsu:
Students are able to view the annotation from other students and the instructor.
There are also uncaught syntax errors hanging in the console, see the following picture.
image

### What is the new behavior?
Students can not view others' annotations while the instructor still can view everything through the **Show All Annotations** button. Both **Download** and  **Show All Annotations** buttons are disabled for students. See the following pictures.

Student interface
image

Instructor interface
image

Removed the error from the console.


site/app/controllers/pdf/PDFController.php
```diff
@@ -214,6 +214,7 @@ public function showGraderPDFEmbedded(string $gradeable_id) {
         }
         $is_peer_grader = false;
         $grader_id = $this->core->getUser()->getId();
+        $is_peer_grader = false;
```
tkoz0, line 217: Why is this line duplicated?

tkoz0 (changes requested):
I tested it and it looks good. I only commented about 1 line of code, otherwise I think it's fine.

site/app/controllers/pdf/PDFController.php
Eceptonsu: Fixed now, thanks for pointing out

PR #7338, opened 2021-11-08
[Feature:Submission] Late Day database triggers

jportorreal00:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
Late days are currently calculated within the LateDays object via php. Current late days use previous late day events as context in order to determine the #of late days remaining and # of late days used.

### What is the new behavior?
Late day information will now be stored in a table. (Table is similar to Late day usage table). This will make accessing late day information a lot faster since we will not have to calculate it from scratch every time.

Late days will now be automatically reset and recalculated using triggers within the databases. Cases where this will happen:

- submit new gradeable
- gradeable version change
- drop submission
- late day (allowed table) addition
- late day (allowed table) deletion
- gradeable late days allowed change
- gradeable due date change
- gradeable due date disabling/enabling
- gradeable late days disabling/enabling
- gradeable deletion
- add new late day extension
- edit existing late day extension
- delete late day extension
- change default late days (this is the only **manual** table reset: inside the php code)

### Other information?

I tested locally on two courses. I ran over each of the scenarios listed and they work as expected (using pgAdmin)

The changes to twig files were due to the changes in the Late Day Usage data. Ex:
image

PR #8022, opened 2022-06-09
Late days 7954

### LINKED ISSUE #7954: Late Days - Complex Usage Confusion
Steps to produce bugs: 
NOTE: In this example today is 5/27, adjust your dates as needed when you reproduce

In a brand new course (e.g., the blank course)...

* Set "Initial Allowed Late Days" to 5 and "Default late days" to 3 on the "Course Settings" page.  Under "Manage sections" add a registration section.  Add the "student" user to the course and place them in the registration section you just created.

Screen Shot 2022-05-26 at 11 18 42 PM

* Create a gradeable that allows students to submit using up to 3 late days, with deadline 2 days ago (5/25 in this example), use the student account to submit to that gradeable today (5/27) -- and be "charged" 2 late days.  

Screen Shot 2022-05-27 at 11 06 44 AM

This is the students view on "My Late Days/Exceptions", and below it the instructors view in "Late Days Allowed" -> "Late Days Forensics".  These are correct.

Screen Shot 2022-05-27 at 11 06 20 AM

Screen Shot 2022-05-27 at 11 07 01 AM

* Use the "Late Days Allowed" page to give that student an extra late day for the semester effective tomorrow (5/28 in this example).  The design SHOULD BE that we do this by entering '6' for that student.  
[ NOTE: This part of the design is definitely confusing.  But the design was it should be the new semester total allowed (whether they used days on prior homeworks or not). ]

Screen Shot 2022-05-27 at 11 07 26 AM

Screen Shot 2022-05-27 at 11 07 34 AM

* The student view is confusing & buggy:

Screen Shot 2022-05-27 at 11 07 55 AM

Tomorrow's date should not have any info in the "max number allowed for this assignment" column (it's not an assignment).  The change in late days should be "+1" for the tomorrow (5/28) date row.  I'm actually not sure where the "+5" is coming from.  The remaining late days for the tomorrow (5/28) row should be "4"... not "8".

* The instructor view on "Late Days Allowed" -> "Late Days Forensics", has the same errors as above.
 
Screen Shot 2022-05-27 at 11 08 07 AM

* Now, changing Course Settings to 10 initial allowed late days for the semester:

Screen Shot 2022-05-27 at 11 09 14 AM

* The student & instructor views: 

Screen Shot 2022-05-27 at 11 08 32 AM

Screen Shot 2022-05-27 at 11 08 51 AM

Again, we have the same stray '6'  in the student view.   Because the student is allowed to use a total of 6 days for the semester, this new total specified for tomorrow is actually a loss in total late days (their peers can use 10).  It might not be what the instructor wanted (or maybe they do want to penalize this student by taking away late days for another reason), but it is the original functionality design.  The student has used 2, so they have 4 remaining for future homeworks (so that number is correct).  

* However, if we change Course Settings initial allowed days from 10 back to 5 initial allowed late days for the semester, we get a different computation than we had before:

Screen Shot 2022-05-27 at 11 09 32 AM

Screen Shot 2022-05-27 at 11 09 50 AM

Again the stray '6' should not be there in the tomorrow (5/28) row, but it looks like the remaining late days are now correct -- but they are different from what we had above the same setting, initial late days allowed = 5.

tkoz0:

In #7954, a sequence of steps is described to produce wrong information in the `late_day_cache` table. This is caused by incorrectly using existing data from `late_day_cache` to compute the rest of the cache.

### What is the new behavior?
The incorrect calculation is fixed. Additionally, the confusing number on the student view (for a late day update applied to them) is removed.

### Other information?
This change requires a database migration. This was tested by repeating similar steps as described in #7954, inspecting the database tables, and deleting/recalculating the `late_day_cache` at various points during steps used to reproduce the issue.

Closes #7954


migration/migrator/migrations/course/20220609173520_fix_late_day_bug.py
```diff
@@ -0,0 +1,89 @@
+"""Migration for a given Submitty course database."""
+
+
+def up(config, database, semester, course):
```
williamjallen, line 4: I believe that there should still be a `down()` function in all migrations, regardless of whether it is a no-op -- someone feel free to correct me if this is incorrect.  Regardless, unless there's a good reason to not include one, there should be a down migration which restores the database to the previous state, just in case we ever need to revert this change at some point.

williamjallen (changes requested):

migration/migrator/migrations/course/20220609173520_fix_late_day_bug.py
tkoz0: At the meeting with @bmcutler we discussed this and I believe we concluded a down migration was not necessary. In this case, it's a function change so any code that depends on this function will use one that was updated to work correctly. I believe this is why it's not necessary to have a down migration, but we can discuss further if this is a mistake.

williamjallen (approved):
Changes work as described.

Eceptonsu (approved):

migration/migrator/data/course_tables.sql
```diff
@@ -1671,7 +1656,7 @@ CREATE TABLE public.users (
     user_email_secondary character varying(255) DEFAULT ''::character varying NOT NULL,
     user_email_secondary_notify boolean DEFAULT false,
     registration_type character varying(255) DEFAULT 'graded'::character varying,
-    CONSTRAINT check_registration_type CHECK (((registration_type)::text = ANY ((ARRAY['graded'::character varying, 'audit'::character varying, 'withdrawn'::character varying, 'staff'::character varying])::text[]))),
+    CONSTRAINT check_registration_type CHECK (((registration_type)::text = ANY (ARRAY[('graded'::character varying)::text, ('audit'::character varying)::text, ('withdrawn'::character varying)::text, ('staff'::character varying)::text]))),
```
bmcutler, line 1659: I don't think this change should be in this PR

migration/migrator/data/course_tables.sql
```diff
@@ -13,20 +13,6 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
---
--- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: -
---
-
-CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
```
bmcutler, line 20: Why are these lines removed?

migration/migrator/data/course_tables.sql
```diff
@@ -41,7 +27,6 @@ CREATE TYPE public.notifications_component AS ENUM (
 
 SET default_tablespace = '';
 
-SET default_with_oids = false;
```
bmcutler, line 44: why is this removed in this PR?
(I dont know where it's used, I see it was added 2 months ago with PR #7338

bmcutler (changes requested):

migration/migrator/migrations/course/20220609173520_fix_late_day_bug.py
```diff
@@ -0,0 +1,89 @@
...
+            WHERE ldc.user_id = user_id
+            ORDER BY ldc.late_day_date DESC, ldc.g_id DESC NULLS LAST
+            LIMIT 1
+        ) LOOP
+            late_days_remaining = var_row.late_days_remaining;
+            latestDate = var_row.late_day_date;
+        END LOOP;
+        
+        -- Get the number of late days charged up to this point
+        late_days_used = (SELECT COALESCE(SUM(-ldc.late_days_change), 0)
```
jportorreal00, line 44: Changes look good! All over a missing negative sign *facepalm*

PR #8030, opened 2022-06-12
[Feature:System] Add support for SAML authentication

### LINKED ISSUE #1803: Implement CAS authentication
We currently have a login system in place which allows login/logout, but I think it'd be neat if we could plug directly in to the RPI CAS system. Although they don't say what the CAS system is, it _looks_ like a re-skin of JASIG CAS (judging from [this js file](https://cas-auth.rpi.edu/cas/js/cas.js)). Luckily, JASIG CAS was very pluggable in my limited usage of it a few years ago.

pros:
- logout comes for free! (https://cas-auth.rpi.edu/cas/logout)
- easier for school to reset passwords
- [JASIG CAS](https://wiki.jasig.org/display/CAS/Home) is a trusted project from MIT with an [official](https://wiki.jasig.org/display/CASC/phpCAS) php client library to work with it.

cons:
- do teachers have an account through our CAS system? If not, probably not worth it then because then we'd still need to maintain an admin login panel or sthg
- CAS does authentication, not authorization. We still need a way to identify students vs teachers/graders. I assume this already happens somewhere in the code, but it might need to be reworked if we implement this.

Also note that since JASIG CAS is an open source project, this still scales well to other schools and possibly with other deployment strategies.

### LINKED ISSUE #7652: Add SAML 2.0 authentication support

cjreed121:

Submitty does not support SAML authentication.

Closes #7652
Closes #1803 as CAS wouldn't be needed if SAML is used

### What is the new behavior?
Submitty now offers SAML authentication as an option.

### Other information?
Testing will rely upon https://github.com/Submitty/docker-test-saml-idp to act as a mock SAML IdP.

This uses https://github.com/onelogin/php-saml to handle the SAML requests and assertions.

This is not ready for merge but feedback would be great.


site/cgi-bin/saml_check.cgi
```diff
@@ -0,0 +1,28 @@
...
+    print()
+
+    arguments = cgi.FieldStorage()
+    valid = False
+    if 'username' in arguments:
+        valid = valid_username(arguments['username'].value)
+    print(json.dumps({"valid": valid}))
+
+def valid_username(username):
+    if re.match('^[a-z]{2,6}[0-9]{0,2}$', username):
```
cjreed121, line 22: Ideally this would instead call something abstract that the sysadmin can define. This is just here for testing for now.

migration/migrator/migrations/system/20220501171554_add_saml_options.py
```diff
@@ -0,0 +1,28 @@
...
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    authentication_file = config.config_path / 'authentication.json'
+    with authentication_file.open('r+') as auth_file:
+        auth_info = json.load(auth_file, object_pairs_hook=OrderedDict)
+        if not 'saml_options' in auth_info:
+            auth_info['saml_options'] = {}
```
Viyerelu23333, line 16: Maybe adding the keys with default value here?  As the PHP might throw an error at [`site/app/models/Config.php@L343`](https://github.com/Submitty/Submitty/pull/8030/files#diff-ecb12eba4b1d073362c00e86f6a3ee5d70bf982eada6d72ff244abec74792940R343)

migration/migrator/migrations/master/20220322164933_add_saml_mapped_users.py
```diff
@@ -0,0 +1,68 @@
...
+    database.execute(
+    """
+    CREATE OR REPLACE FUNCTION saml_mapping_check() RETURNS trigger
+        LANGUAGE plpgsql
+        AS $$
+        BEGIN
+            IF (SELECT count(*) FROM saml_mapped_users WHERE NEW.user_id = user_id) = 2
+            THEN
+                IF (SELECT count(*) FROM saml_mapped_users WHERE NEW.user_id = user_id AND user_id = saml_id) > 0
+                THEN
```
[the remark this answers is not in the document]
cjreed121: Ah yes I have to run another migration dump. Thanks for reminding me.

migration/migrator/migrations/system/20220501171554_add_saml_options.py
cjreed121: If you go and run CONFIGURE_SUBMITTY.py it would then fill those values in. Also if you wanted to just raw edit this config file then you could the put the values in. I don't think it's necessary to have them here and this is similar to how LDAP was done.

williamjallen (commented):
Code looks good overall.  I noted a couple minor things but it otherwise looks fine,

site/app/authentication/SamlAuthentication.php
```diff
@@ -0,0 +1,115 @@
...
+    public function authenticate(): bool {
+        if (isset($_POST['user_id']) && isset($_SESSION['Authenticated_User_Id'])) {
+            // need to check if SAML_Authenticated_User has permission to log in as user_id
+            $authorized_users = array_map(function ($user) {
+                return $user['user_id'];
+            }, $this->core->getQueries()->getSAMLAuthorizedUserIDs($_SESSION['Authenticated_User_Id']));
+            unset($_SESSION['Authenticated_User_Id']);
+            if (in_array($_POST['user_id'], $authorized_users)) {
+                $this->user_id = $_POST['user_id'];
+                $_POST['RelayState'] = $_SESSION['RelayState'];
```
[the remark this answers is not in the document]
cjreed121: The only way that code will get executed is if a user passes through 102-104 which will require that those get set. If `$_SESSION['Authenticated_User_Id']` is set then the other should be too.

bmcutler (approved):
minor requests:  (open an issue if not addressed here)

* on the saml management page, sort the list of mapped proxy users by saml username, then submitty username

* dark mode for the saml login.md has poor contrast for links.

PR #8035, opened 2022-06-13
[Feature:Autograding] Add comment counting

### LINKED ISSUE #2979: Feature Request: some automated grading of comments
counting lines of comments?
sloc count?

poorna2152:

Closes #2979 

### Other information?

[PR#435](https://github.com/Submitty/submitty.github.io/pull/435) adding documentation


bin/comment_count.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+if [ "$#" -eq 0 ]; then
+  echo "Pass at least one directory or file" >&2
+  exit 1
+fi
+mapfile -t csv < <( cloc --csv --quiet "$@" 2>/dev/null)
```
MasterOdin, line 6: From the docs, this program is described as counting the number of comments, however,  `cloc` counts number of _comment lines_ in a file and this is an important distinction to make in the docs at least. For example, given the cpp code block:

```cpp
/*
this is a multiline
comment
that goes for a few lines
*/

// this is a single line comment
```

`cloc` will return a comment line count of `6`. If we explicitly want the number of comments vs comment line count, then I think we'll need another tool.

.setup/update_system.sh
```diff
@@ -27,6 +27,10 @@ fi
 #install DLL for zbar
 apt-get install libzbar0 --yes
 
+#libraries for comment counting :
+#install cloc
+apt install cloc --yes
```
MasterOdin, line 32: Per the [Debian Administrator's Handbook on the apt tools](https://debian-handbook.info/browse/stable/sect.apt-get.html):
> Both tools are built on top of the same library and are thus very close, but the default behavior of **apt** has been improved for interactive use and to actually do what most users expect. The APT developers reserve the right to change the public interface of this tool to further improve it. On the opposite, the public interface of **apt-get** is well defined and will not change in any backwards incompatible way. It is thus the tool that you want to use when you need to script package installation requests.

so we should do instead:
```suggestion
apt-get install cloc --yes
```

bin/comment_count.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
MasterOdin, line 1: As a note, I'd recommend writing this file in python (where use `subprocess.run` around `cloc --csv --quiet`, and then rest of code is just python) so that:
1. it's more portable (I cannot run this on my macOS terminal as it doesn't have `mapfile` function, which is a bash 4+ builtin and mac comes with zsh (default) and bash 3)
2. it's simpler for new students / programmers to edit and doesn't require futzing with bash's weird string handling.

bin/comment_count.sh
poorna2152: I went over some [similar counting tools](https://github.com/AlDanial/cloc#other-counters-). They dont seem to have a functionality to match our use case. 
What I could see as an option is to get the cloc tool's output and seperately go over the source file and counting multi-line comments, and substract the values. (multi-line comments in most of the languages seems to use `/* */`. So effectively check for these symbols) Does that sound as a feasible option.

williamjallen: I proposed this idea in the meeting but it didn't seem to go anywhere: What about using tree-sitter to count comments?  We'd be limited by the number of languages tree-sitter supports, but I don't expect that to be an issue given the courses which would likely use this utility the most.  It would take a few dozen lines of code at most to support basic comment-counting functionality, and would allow us to change the behavior as necessary.

poorna2152: Tested this using `tree-sitter` it seems to be possible to do this.

.setup/install_submitty/install_bin.sh
```diff
@@ -42,6 +42,9 @@ done
 chown root:root ${SUBMITTY_INSTALL_DIR}/bin/generate_repos.py
 chmod 500 ${SUBMITTY_INSTALL_DIR}/bin/generate_repos.py
 
+chown ${DAEMON_USER}:${COURSE_BUILDERS_GROUP} ${SUBMITTY_INSTALL_DIR}/bin/comment_count.sh
+chmod 555 ${SUBMITTY_INSTALL_DIR}/bin/comment_count.sh
```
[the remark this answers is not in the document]
poorna2152: This is needed because this is an autograding command

poorna2152:
When invoking the `comment_count.py` file it was failing due to `RLIMIT_NPROC` being 0.  So had to increase the `RLIMIT_NPROC` size to get it working. Added the setting of `RLIMIT_NPROC` to the example config file in the more_autograding_examples directory.

bmcutler (changes requested):
Looks great!

We should also add this new example to the integration test case suite.
We'll also want example submissions with fewer than the target line count and more than the target line count. (to make sure the score & the messages are correct.)

Viyerelu23333 (approved):
Tested and worked well on my end.

bmcutler (approved):

PR #8036, opened 2022-06-13
[Bugfix:System] Create the course only if the term exists

### LINKED ISSUE #7214: Creating course with invalid term does not throw error
While `create_course.sh` does a not fail execution when the class exists, it does not completely confirm it was set up correctly either. (Is that to many nots?) In my case, the term was missing, so it build the class, but could not associate it with a term. After creating the term and running the script a second time, the output suggests the course is set up, but is missing the link between course and term.

_Originally posted by @prestoncarman in https://github.com/Submitty/Submitty/issues/7160#issuecomment-934020166_

akshatbatra:
Fixes #7214, running `create_course.sh` with an invalid term creates the course even after failing to link it to a term, running `create_course.sh` after creating the term suggests that the course is already set up and doesn't link the term with the course.

### What is the new behavior?
Providing an invalid term doesn't create the course and throws an error suggesting to create the term first.

### DIFF
--- a/sbin/create_course.sh
+++ b/sbin/create_course.sh
@@ -117,6 +117,13 @@ fi
 #       additional instructors and/or head TAs who need read/write
 #       access to these files
 
+TERM_EXISTS=$(PGPASSWORD=${DATABASE_PASS} psql ${CONN_STRING} -d submitty -AXtc "SELECT COUNT(*) FROM terms WHERE term_id='${semester}'")
+
+if [[ "$TERM_EXISTS" -eq "0" ]] ; then
+    echo "ERROR: Provided term ${semester} doesn't exist."
+    echo "       To fix, try creating the term first by running 'create_term.sh'."
+    exit
+fi
 
 # FIXME: add some error checking on the $semester and $course
 #        variables


sbin/create_course.sh
```diff
@@ -117,6 +117,13 @@ fi
 #       additional instructors and/or head TAs who need read/write
 #       access to these files
 
+TERM_EXISTS=$(PGPASSWORD=${DATABASE_PASS} psql ${CONN_STRING} -d submitty -AXtc "SELECT COUNT(*) FROM terms WHERE term_id='${semester}'")
+
+if [[ "$TERM_EXISTS" -eq "0" ]] ; then
+    echo "ERROR: Provided term ${semester} doesn't exist."
+    echo "       To fix, try creating the term first by running 'create_term.sh'."
```
cjreed121, line 124: I don't think we need to specify the fix here so you can probably remove this line.

cjreed121 (commented):
Change looks good. Just 1 comment about the output.

sbin/create_course.sh
akshatbatra: Done

cjreed121 (approved):
Looks good to me and works as expected

PR #8041, opened 2022-06-14
[Feature:InstructorUI] Added registration section to home page

### LINKED ISSUE #7248: Display registration section to student users
Currently a student in a class who has access to Submitty only knows that are in a valid (non-NULL) section, 
but cannot determine which registration section they are part of.  (If we have multiple sections, they cannot know if 
they are in the "correct" section, or if 

Let's add a display the registration section somewhere.  
Probably on the main page listing all courses?

NOTE: We still want to keep the rotating section secret/hidden from the student.  They do not need to know that information.

tkoz0:

The home page does not list registration sections.

### What is the new behavior?
Registration sections are now listed on the home page. Below are examples of how they appear:

On desktop:
image

On mobile:
image

The section information is not displayed for users in the NULL section (most likely the course staff).

### Other information?
Closes #7248

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -4083,7 +4083,7 @@ SQL;
         }
 
         $query = <<<SQL
-SELECT t.name AS term_name, u.semester, u.course, u.user_group
+SELECT t.name AS term_name, u.semester, u.course, u.user_group, u.registration_section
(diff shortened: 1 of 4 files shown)


bmcutler:
@shailpatels 
I requested nothing instead of null

shailpatels (approved):
@bmcutler ah ok if that's the intended behavior then everything else looks good

PR #8046, opened 2022-06-14
[Feature:Autograding] Autograding command json file

### LINKED ISSUE #1256: Separate white listing of programs into 2 lists: all submitty & per university
Also the shortcuts for paths for options.

Specifically these functions from file:
https://github.com/Submitty/Submitty/blob/master/grading/execute.cpp
bool system_program(const std::string &program, std::string &full_path_executable) {
std::string validate_option(const std::string &program, const std::string &option) {

poorna2152:

Closes #1256


Viyerelu23333 (commented):
The code looks good to me and worked on my end.  The only concern might be the security — 

Every time the runner is called, it pulls the dictionary.  If the json is changed after building the course, it might impact the autograding results.

Also tested on the worker machine.

grading/execute.cpp
```diff
@@ -39,150 +39,56 @@ extern const int CPU_TO_WALLCLOCK_TIME_BUFFER;  // defined in default_config.h
...
 
+#define ALLOWED_COMMANDS_CUSTOM SUBMITTY_INSTALL_DIRECTORY + "/config/autograding_allowed_commands_custom.json"
+#define ALLOWED_COMMANDS_DEFAULT SUBMITTY_INSTALL_DIRECTORY + "/config/autograding_allowed_commands_default.json"
 
 // =====================================================================================
 // =====================================================================================
 
+std::string replace_placeholder(std::string value) {
+    std::string placeholder = "{SUBMITTY_INSTALL_DIRECTORY}";
+    int index = value.find(placeholder);
```
Viyerelu23333, line 50:
```suggestion
    std::string::size_type index = value.find(placeholder);
```

Viyerelu23333:
I am trying to test the installation path substitution... and Submitty reported errors in the docker autograding (Tutorial/05).  

<details>

```
LOGGING BEGIN my_runner.out
Log entry for container0:

Running User Code...
4.0K ./.submit.timestamp  4.0K ./queue_file.json
4.0K ./buggy3.cpp	  4.0K ./user_assignment_settings.json
 15M ./my_runner.out
========================================================
TEST test01
TITLE Check for goto and auto
Execution has no graphics display
IN EXECUTE:  'submitty_count token -l c Goto *.cpp 1>STDOUT_0.txt 2>STDERR_0.txt'
identified 0 dispatcher actions
terminate called after throwing an instance of 'nlohmann::detail::parse_error'
  what():  [json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal
terminate called after throwing an instance of 'nlohmann::detail::parse_error'
  what():  [json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal
```

</details>

For the jailed sandbox autograding (Tutorial/07), it seems the substitution works in the building phase.  However, when running the actual autograding, the substitution seems to be messed up.  

<details>

Building:

```diff
...
INSTALL assignment config: /usr/local/submitty/GIT_CHECKOUT/Tutorial/examples/07_loop_depth/config
        destination:       /var/local/submitty/courses/f22/tutorial/build/07_loop_depth
FILENAME /usr/local/submitty/bin/configure.out
Inflating testcases
+ replaced as /usr/local/submitty/SubmittyAnalysisTools/count
Scanning dependencies of target custom_functions
[ 10%] Building CXX object CMakeFiles/custom_functions.dir/usr/local/submitty/src/grading/seccomp_functions.cpp.o
...
```

Running:

```diff
IN EXECUTE:  'submitty_count -l python depth loop *.py 1>STDOUT.txt 2>STDERR.txt'
identified 0 dispatcher actions
+ replaced as /usr/local/submitty/SubmittyAnalysisTools/count
childPID = 497070
PARENT PROCESS START: 
+ replaced as /usr/local/submitty/SubmittyAnalysisTools/count
PARSE COMMAND LINE submitty_count -l python depth loop *.py 1>STDOUT.txt 2>STDERR.txt
TOKEN   0 IS 'submitty_count'
TOKEN   1 IS '-l'
TOKEN   2 IS 'python'
TOKEN   3 IS 'depth'
TOKEN   4 IS 'loop'
TOKEN   5 IS '*.py'
TOKEN   6 IS '1>STDOUT.txt'
TOKEN   7 IS '2>STDERR.txt'
+ replaced as /usr/local/submitty/SubmittyAnalysisTools/count
CHECK FOR UNESCAPED WILDCARD '-l'
CHECK FOR UNESCAPED WILDCARD 'python'
CHECK FOR UNESCAPED WILDCARD 'depth'
CHECK FOR UNESCAPED WILDCARD 'loop'
CHECK FOR UNESCAPED WILDCARD '*.py'
  directory: 
  file_pattern *.py
   MATCHED!  'solution.py'


- MY PROGRAM: '/usr/local/submittyAnalysisTools/count'
MY STDIN:  ''
MY STDOUT:  'STDOUT.txt'
MY STDERR:  'STDERR.txt'
MY ARGS (5) :'-l' 'python' 'depth' 'loop' 'solution.py' 

Child exited, status=1
```

</details>

Not sure if these two cases relate to this PR.

grading/execute.cpp
```diff
@@ -39,150 +39,51 @@ extern const int CPU_TO_WALLCLOCK_TIME_BUFFER;  // defined in default_config.h
...
-    full_path_executable = itr->second;
-    return true;
-  }
-
-  // did they already use the full path name?
-  for (itr = allowed_system_programs.begin(); itr != allowed_system_programs.end(); itr++) {
-    if (itr->second == program) {
-      full_path_executable = program;
-      return true;
+bool system_program(const std::string &program, std::string &full_path_executable, const bool running_in_docker)
```
akshatbatra, line 57: Will `system_program` be called more than once during the execution of this program? If yes, then it might be good to define `allowed_system_programs` early on and use the same on every call instead of creating a new json object on every call.

grading/execute.cpp
```diff
@@ -39,150 +39,51 @@ extern const int CPU_TO_WALLCLOCK_TIME_BUFFER;  // defined in default_config.h
...
+        full_path_executable = replace_placeholder(allowed_system_programs[program]);
+        return true;
     }
-  }
 
-  // not an allowed system program
-  return false;
+    for (auto& x : allowed_system_programs.items())
+    {
+        if (replace_placeholder(x.value()) == program) {
```
akshatbatra, line 79: Wouldn't it be better to move the creation of `autograding_allowed_commands_default.json` to some Python setup script where `SUBMITTY_INSTALL_DIRECTORY` is defined? That way `replace_placeholder` will not be needed at all.

grading/execute.cpp
poorna2152: I also had the same idea, but I thought since this is a secure piece of information we wouldn't want a object passing around functions. Not sure about my argument though.

grading/execute.cpp
poorna2152: Yeah. I could use this [filling_variable_replacement](https://github.com/Submitty/Submitty/blob/main/.setup/INSTALL_SUBMITTY_HELPER.sh#L219). Does that sound good

grading/execute.cpp
akshatbatra: That shouldn't be a problem, but this argument is convincing enough to err on the side of caution.

grading/execute.cpp
```diff
@@ -70,13 +61,13 @@ bool system_program(const std::string &program, std::string &full_path_executabl
         allowed_system_programs["php"] = "/usr/bin/php";
     }
     if (allowed_system_programs.contains(program)) {
-        full_path_executable = replace_placeholder(allowed_system_programs[program]);
+        full_path_executable = allowed_system_programs[program];
         return true;
     }
 
     for (auto& x : allowed_system_programs.items())
```
akshatbatra, line 68: Just one more comment considering that this program should be as performant as possible. This loop is encountered only if `allowed_system_programs.contains(program)` returns false, would it be a good idea to wrap this/these loop(s) in a conditional such that if `program` contains `/` then we can directly skip to this loop instead of first running `allowed_system_programs.contains(program)`? Because I don't think if `/` is present in `program` then it will be one of the keys in JSON. Thoughts?

poorna2152: If I understand correctly you are suggesting `if (program.length() > 0 && program[0] == "/") ` check right

akshatbatra: Right, something like:
`if (program.length() > 0 && program[0] == "/") {
    for (auto& x : allowed_system_programs.items())
    {
        if (x.value() == program) {
           full_path_executable = program;
           return true;
        }
    }
}
else {
    if (allowed_system_programs.contains(program)) {
        full_path_executable = allowed_system_programs[program];
        return true;
    }
}`

poorna2152: Updated

akshatbatra (approved):
Haven't tested it, but code looks good now.

Viyerelu23333 (approved):
Looks good to me and worked on my end

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -451,8 +451,31 @@ find "${SUBMITTY_INSTALL_DIR}/vendor" -type f -exec chmod 444 {} \;
 
 echo -e "Copy the grading code"
 
-# copy the files from the repo
-rsync -rtz "${SUBMITTY_REPOSITORY}/grading" "${SUBMITTY_INSTALL_DIR}/src"
+# copy the files from the repo excluding autograding_allowed_commands_default.json
+sync_exclude=autograding_allowed_commands_default.json
+rsync -rtz --exclude "${sync_exclude}" "${SUBMITTY_REPOSITORY}/grading" "${SUBMITTY_INSTALL_DIR}/src"
```
bmcutler, line 456: unnecessary complication, don't bother to exclude this file from the rsync.

bmcutler (changes requested):
Overall this looks very good!  Thanks!

Just some requests to help the long term maintainability:

* Let's allow C/C++ comments in both of the allowed_autograding_commands files.  This will allow us to inline document the requester/purpose for each allowed command and to one-by-one or block comment out commands that we are phasing out/deprecating.  These comments can be stripped out using the preprocessor -- hopefully not too messy to organize :)

   - Allowing comments will let us put a note at the top of the default file that this will be installed to <insert location>.   (related to the unnecessary complication of excluding from rsync)


* Let's be consistent about the naming of the files & the variables within the code.  Instead of 'allowed_system_program' 'default_allowed_commands' 'custom_allowed_commands' 'autograding_allowed_commands_default', etc...  let's in all places use the term 'allowed_autograding_commands', 'allowed_autograding_commands_default', and 'allowed_autograding_commands_custom'.

grading/execute.cpp
```diff
@@ -39,150 +39,44 @@ extern const int CPU_TO_WALLCLOCK_TIME_BUFFER;  // defined in default_config.h
...
+        nlohmann::json allowed_system_programs_custom;
+        custom_sstr >> allowed_system_programs_custom;
+        allowed_system_programs.update(allowed_system_programs_custom);
     }
-  }
-
-  // not an allowed system program
-  return false;
+    if(running_in_docker){
+        allowed_system_programs["bash"] = "/bin/bash";
```
bmcutler, line 60: maybe not in this PR (future PR?)
But it would be good to put this list of commands into the json file(s) as well
Either a separate json with commands for docker containers only, or make the json structure more complex with a boolean per flag about whether it available only inside docker container autograding.  
A more complex json structure would have the benefit of easy extensibility for future features.

poorna2152: Items now have a structure such as this.

```
"php":                     { "path": "/usr/bin/php", "onlyDocker": true }
"matlab":                  { "path": "/usr/local/bin/matlab" },
```

poorna2152:
Added comments in json with the ignore_comment functionality when parsing a json string in nlohmann json.
https://json.nlohmann.me/features/comments/

saumyaborwankar (approved):
Tested it. Works as expected.

bmcutler (approved):
finally :)

PR #8051, opened 2022-06-15
Added file size cap and file format criteria

akshatbatra:
There is no client side limit on profile image size and a whole lot of unneeded file formats are supported opening doors for file inclusion and crafted SVGs.

### What is the new behavior?
A client side limit of 20MB has been imposed and only JPG, GIF, PNG and WEBP file formats are supported now.

### DIFF
--- a/site/app/controllers/UserProfileController.php
+++ b/site/app/controllers/UserProfileController.php
@@ -145,6 +145,11 @@ class UserProfileController extends AbstractController {
         else {
             $meta = explode('.', $_FILES['user_image']['name']);
             $extension = $meta[1];
+            $supported_ext = ["jpg", "jpeg", "gif", "png", "webp"];
+            $supp_mime_types = ["image/jpeg", "image/gif", "image/png", "image/webp"];
+            if (!(in_array($extension, $supported_ext) && in_array(mime_content_type($_FILES['user_image']['tmp_name']), $supp_mime_types))) {
+                return JsonResponse::getErrorResponse('File format not supported.');
+            }
 
             // Save image for user
             $result = $user->setDisplayImage($extension, $_FILES['user_image']['tmp_name']);
--- a/site/public/js/user-profile.js
+++ b/site/public/js/user-profile.js
@@ -273,4 +273,10 @@ $(document).ready(function() {
     // Set time zone drop down boxes to the user's time zone (only after other JS has finished loading)
     let user_time_zone =  $('#time_zone_selector_label').data('user_time_zone');
     $('[value="' + user_time_zone + '"]').prop('selected', true);
+    $('#user-image-button').bind('change', function() {
+        if ((this.files[0].size/1048576)>20.0){
+            alert("Selected file's size exceeds 20 MB");
+            $('#user-image-button').val('');
+        }
+    });
 });
--- a/site/app/templates/EditProfilePhotoForm.twig
+++ b/site/app/templates/EditProfilePhotoForm.twig
@@ -17,7 +17,10 @@
         <label for="user-image-button">
         <h2>My Profile Photo</h2>
         </label>
-        <input type="file" name="user_image" id="user-image-button" accept="image/*">
+        <br>
+        <input type="file" name="user_image" id="user-image-button" accept=".jpg, .jpeg, .gif, .png, .webp">
+        <br>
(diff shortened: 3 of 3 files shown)


site/public/js/user-profile.js
```diff
@@ -273,4 +273,10 @@ $(document).ready(function() {
     // Set time zone drop down boxes to the user's time zone (only after other JS has finished loading)
     let user_time_zone =  $('#time_zone_selector_label').data('user_time_zone');
     $('[value="' + user_time_zone + '"]').prop('selected', true);
+    $('#user-image-button').bind('change', function() {
+        if ((this.files[0].size/1048576)>20.0){
+            alert("Selected file's size exceeds 20 MB");
+            $('#user-image-button').val('');
+        }
+    });
```
williamjallen, line 281: Is this being enforced on the server side somewhere besides the default upload size settings?

williamjallen (commented):
I have one comment below about the code.  Other than that, the changes look decent.  I question whether we want to support animated GIFs though.  Thoughts?

site/public/js/user-profile.js
williamjallen: I also wonder whether 20 MB is an appropriate limit for a profile picture.  If everyone in a 300-person class uploaded a 20 MB file, going to the page which lists student profiles would result in a very very large download.  Would a more appropriate limit be 2MB?  Maybe 5MB?  It would be better to have an even lower (<1MB) limit in my opinion, but it would likely require most students to compress their images somewhat which may be desirable.

akshatbatra: I agree, animated GIFs don't seem good in educational settings. I tested with some animated ones and they lost their animated nature, but they also got kind of corrupted. If a static GIF is uploaded it shows up fine. 

Image dimensions are being resized server side, due to this no image will probably have 20MB size. I agree that 20MB is a bit large, I relied on server-side resize for compression. I think having to compress image yourself can be a hassle and deteriorates user experience. It should definitely be decreased but I'm also not sure about the exact number.

MBon1 (approved):
The code looks good.  I successfully edited a profile photo using a GIF, JPG, PNG, and WEBP.  I also tried using an SVG file and a PNG that exceeded the 5 MB limit, and in these case, I was successfully prevented from setting the profile photo to either of these files.

site/app/controllers/UserProfileController.php
```diff
@@ -145,6 +145,11 @@ public function changeProfilePhoto() {
         else {
             $meta = explode('.', $_FILES['user_image']['name']);
             $extension = $meta[1];
```
shailpatels, line 147: While you're editing this code I think it would be a good change to edit how the file type is being read and instead use some of the helper functions in Submitty's `site/app/libraries/FileUtils.php` , mainly the `isValidImage` and `validateUploadedFiles` functions

shailpatels (changes requested):

site/app/controllers/UserProfileController.php
akshatbatra: Thanks for pointing this out, didn't know that these functions existed.

cjreed121 (changes requested):
Front end size validation will definitely provide a good user experience but I think we should also add the same check to the backend PHP.

site/app/controllers/UserProfileController.php
```diff
@@ -143,11 +144,13 @@ public function changeProfilePhoto() {
             return JsonResponse::getErrorResponse('No image uploaded to update the profile photo');
         }
         else {
-            $meta = explode('.', $_FILES['user_image']['name']);
-            $extension = $meta[1];
+            preg_match("/^.*\.(jpg|jpeg|png|gif)$/i", $_FILES['user_image']['name'], $extension);
+            if (!(FileUtils::isValidImage($_FILES['user_image']['tmp_name']) && FileUtils::validateUploadedFiles($_FILES['user_image'])[0]['success'] && (count($extension) >= 2) && $_FILES['user_image']['size'] <= 5 * 1048576)) {
+                return JsonResponse::getErrorResponse("Something's wrong with the uploaded file.");
+            }
```
cjreed121, line 151: If I have a file such as image.png.file for example, it is considered valid by this. Is the intention of this to be valid?

PR #8054, opened 2022-06-16
[SECURITY][Feature:Submission] Cap on number of profile photos

akshatbatra:
Any user can upload virtually unlimited profile photos.

### What is the new behavior?
In order to prevent resource abuse, a cap has been imposed on the number of profile photos a user can upload.

### DIFF
--- a/site/app/controllers/UserProfileController.php
+++ b/site/app/controllers/UserProfileController.php
@@ -152,7 +152,7 @@ class UserProfileController extends AbstractController {
             if ($result === 2) {
                 return JsonResponse::getErrorResponse('You have exhausted the quota for number of profile photos, kindly contact the system administrator to resolve this.');
             }
- 
+
             if (!$result) {
                 return JsonResponse::getErrorResponse('Something went wrong while updating your profile photo.');
             }
--- a/site/app/controllers/UserProfileController.php
+++ b/site/app/controllers/UserProfileController.php
@@ -9,6 +9,7 @@ use app\libraries\response\JsonResponse;
 use app\libraries\response\MultiResponse;
 use app\libraries\response\RedirectResponse;
 use app\libraries\response\WebResponse;
+use app\models\User;
 use Symfony\Component\Routing\Annotation\Route;
 
 /**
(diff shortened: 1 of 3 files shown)


akshatbatra:
It's 50, 2 is the return code if quota has been exhausted.

immortalcodes:
Oh, okay. Is there any specific reason for making it  50 and it would be better if you could mention the 'number' in the comments.

williamjallen:
Just a comment: should there be a cap on the total number of profile photos allowed, or should there be a limit on the frequency with which the profile picture can be changed?  I know it's relatively common for larger sites to enforce restrictions such as only being able to change your profile picture once per month.  Personally, I think that something like allowing users to change their profile picture up to 5 times in a 6-month span would be more effective than simply a hard limit on the number of times it can be changed.

bmcutler:
I'm ok with 50, this should be way more than necessary and the sysadmin can bulk clear out very old files if necessary.

MBon1:
Other than other than the user's current profile photo, why do the former profile photos need to be stored?  Or is this for the case when a used uploads too many profile photos in quick succession?

site/app/controllers/UserProfileController.php
```diff
@@ -149,6 +149,9 @@ public function changeProfilePhoto() {
             // Save image for user
             $result = $user->setDisplayImage($extension, $_FILES['user_image']['tmp_name']);
             $display_image = $user->getDisplayImage();
+            if ($result === 2) {
```
cjreed121, line 152: If I was just looking at this code without seeing the rest of the code in the PR I would have no idea what 2 means. Maybe add some constants in the User model. So instead of 2 you could say User::PROFILE_PIC_QUOTA_EXHAUSTED for example. It's of course fine for the value of the constant to be an int but it provides more clarity when looking at the code.

If you look at the top of User.php you'll see some examples of constants currently used.

cjreed121 (changes requested):

site/app/models/User.php
```diff
@@ -313,15 +316,18 @@ public function setDisplayImage(string $image_extension, string $tmp_file_path):
...
+            if ($exception->getCode() === self::PROFILE_IMG_QUOTA_EXHAUSTED) {
+                return self::PROFILE_IMG_QUOTA_EXHAUSTED;
+            }
         }
 
         // Update the DB to 'preferred'
         if ($image_saved && $this->core->getQueries()->updateUserDisplayImageState($this->id, 'preferred')) {
             $this->display_image_state = 'preferred';
-            return true;
+            return 1;
```
cjreed121, line 327: Could you also use constants for these ints too

site/app/controllers/UserProfileController.php
```diff
@@ -149,6 +149,9 @@ public function changeProfilePhoto() {
             // Save image for user
             $result = $user->setDisplayImage($extension, $_FILES['user_image']['tmp_name']);
             $display_image = $user->getDisplayImage();
+            if ($result === \app\models\User::PROFILE_IMG_QUOTA_EXHAUSTED) {
```
cjreed121, line 152: You should be able to import the User model with a "use" statement and then this could be shortened to just User::PROFILE_IMG_QUOTE_EXHAUSTED

cjreed121 (changes requested):

cjreed121 (approved):
Looks good

PR #8061, opened 2022-06-18
[Bugfix:TAGrading] Remove Stack Trace in Team PDF Grading

### LINKED ISSUE #8043: Unwanted Notice below the annotation toolbar in Team PDF Grading
**Describe the bug**

As could be seen in the above image there is an unwanted notice and call stack information that should not be visible to user.
This issue occurs only in the Team PDF grading (not in individual PDF grading). 

**Expected behavior**
The information highlighted in the red box should not be there.

**To Reproduce**
1. In InstructorUI
2. Go to Sample Course
3. Any Team PDF Gradeable
4. Grade

Eceptonsu:
Closes #8043
There is an unwanted stack trace in team PDF grading that should not be visible to the user.

### What is the new behavior?
Removes the stack trace in team PDF grading.

### DIFF
--- a/site/app/controllers/pdf/PDFController.php
+++ b/site/app/controllers/pdf/PDFController.php
@@ -178,8 +178,13 @@ class PDFController extends AbstractController {
         for ($index = 1; $index < count($file_path_parts); $index++) {
             if ($index == 9) {
                 $user_id = $file_path_parts[$index];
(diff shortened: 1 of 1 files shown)


Viyerelu23333 (commented):
Tested the team pdf gradeable and looks good to me.

Issue #8034, opened 2022-06-13
Forum page title does not reflect current selected thread title


williamjallen:
**Describe the bug**
The page title does not change when switching between forum threads.

**Expected behavior**
The page title should reflect the title of the thread which is currently selected.

**To Reproduce**
1. Open the discussion for any course
2. Click on any post and observe that the page title matches the post title
3. Click on a different post and observe that the page title does not change

**Configuration**
N/A

**Screenshots**
<img width="1202" alt="first" src="https://user-images.githubusercontent.com/16820599/173361356-5a56696c-aeea-4741-af1f-4c496889c03d.png">
<img width="1201" alt="second" src="https://user-images.githubusercontent.com/16820599/173361368-9c53ae58-fb55-422f-9eab-2c6b61b09092.png">

chandankumar1307:
Hello @williamjallen 
I would love to fix this issue. I think I understand what the issue is here. Would love to talk more about if you could assign me this issue.
I would love to be a part of Submitty's community.
Thank you

kevin-pease:
@chandankumar1307 Are you working on this issue? If not, I'd like to pick up this issue! :)

williamjallen:
Closed by #8062

PR #8062, opened 2022-06-20
[Bugfix:Forum] Forum Title now Display Current Thread

immortalcodes:

The page title does not change when switching between different forum threads.

### What is the new behavior?
The Forum page title now display info about the current thread.

Solves #8034

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -1677,9 +1677,9 @@ function loadThreadHandler(){
           return;
         }
         event.preventDefault();
-        var obj = this;
-        var thread_id = $(obj).data("thread_id");
-        var thread_title = $(obj).data("thread_title");
+        let obj = this;
+        let thread_id = $(obj).data("thread_id");
+        let thread_title = $(obj).data("thread_title");
 
         var url = buildCourseUrl(['forum', 'threads', thread_id]);
         $.ajax({


site/public/js/forum.js
```diff
@@ -1679,6 +1679,7 @@ function loadThreadHandler(){
         event.preventDefault();
         var obj = this;
         var thread_id = $(obj).data("thread_id");
+        var thread_title = $(obj).data("thread_title");
```
williamjallen, line 1682: It's good practice to use `let` and `const` in JS files.  The rest of this file hasn't been linted since `eslint` was added which is why the other variables are defined with the `var` keyword.
```suggestion
        const thread_title = $(obj).data("thread_title");
```

site/public/js/forum.js
```diff
@@ -1712,7 +1713,13 @@ function loadThreadHandler(){
...
                 $('#posts_list').empty().html(JSON.parse(json.data.html));
                 window.history.pushState({"pageTitle":document.title},"", url);
-
+                //Updates the title and breadcrumb
+                $(document).attr("title", thread_title);
+                if(thread_title.length > 25){
+                  $("h1.breadcrumb-heading").text(thread_title.slice(0,25)+"...");  
+                }else{
+                  $("h1.breadcrumb-heading").text(thread_title);
+                }
```
williamjallen, line 1722: This file isn't linted by `eslint`, but `eslint` generally enforces the suggested spacing rules.
```suggestion
                if (thread_title.length > 25) {
                  $('h1.breadcrumb-heading').text(thread_title.slice(0,25) + '...');  
                }
                else {
                  $('h1.breadcrumb-heading').text(thread_title);
                }
```

williamjallen (changes requested):

immortalcodes:
I have changed the `var` to `let` , I guess `const` might give error as the values needs to be updated with the changing threads.
I tried adding linting to the document but I think the file is excluded from linting for some reason , I could forcelint if that seems right.

williamjallen:
Changes look good. I'll review the functionality tonight if possible.

Regarding linting, I'm currently in the process of linting Submitty's JS files.  It's tricky, especially given that merge conflicts are common after large linting PRs get merge so `forum.js` is still ignored.  Since there's no automatic linting for files like this, I try to enforce the linting rules manually when reviewing.

williamjallen (approved):
Changes work as described

PR #8077, opened 2022-06-21
[Feature:System] Parallelize worker updates

### LINKED ISSUE #8060: Parallelize worker software/system updates
When INSTALL_SUBMITTY.sh is run on the primary machine, it kicks off updates on each of the worker machines.
These updates may include costly/slow recompilation of software, downloads of new packages, docker images, etc.
Currently all of these updates are done one machine at a time, in serial.
The system is partially offline until all updates are completed, and this is the sum of the time of all updates.

Instead, we would like do the worker updates in parallel.  So the offline time is equal to the slowest machine, rather than the sum.

OTHER NOTES:

This should be tested with a variety of worker machine states:

* A machine that is intentionally disabled (the machine is set to "enabled" : false in the /usr/local/submitty/config/autograding_workers.json file)

* A machine is very temporarily unreachable due to a very short network outage (< 1 minute) -- in this case we should re-try the connection a short time later.  If a machine is not reachable after multiple attempts in a modest amount of time (> a few minutes) then the machine should be skipped and the end result of the command should be a warning/error.

* A machine that was reached initially in the update, but has crashed or otherwise become unresponsive.  This machine should should be skipped and the end result of the update command should be a warning/error.

* A machine that is marked "enabled" : true but cannot be reached at all.  There was/is code in the update scripts to automatically mark this machine as "enabled" : false -- but I'm not sure that feature is functional, and I'm not sure if that's the design we want to use.   This machine should should be skipped and the end result of the update command should be a warning/error.

tkoz0:

Worker updates are run 1 at a time.

### What is the new behavior?
Worker updates are run in parallel so they can complete faster.

### Other information?
Closes #8060


tkoz0:
@bmcutler I looked at worker the `restart_shipper_and_all_workers.py` file and its main `for` loop could be parallelized similarly to how I parallelized `update_and_install_workers.py`, but it's a fast command needing only 1 or 2 seconds per worker so I don't think we really need to parallelize. As for what can/should be combined/separated in these scripts, I looked at them and I'm not sure and I think it needs some discussion before I work on significant changes in this part of Submitty. For now, should I open this PR for review or should I wait for further discussion?

Viyerelu23333:
Tested on my end and the multi-threading works fine.  The only problem is the DockerUI — it depends on the logs in `logs/docker` and threading makes the logs out of order.

A possible solution is to use a list to store the logs in one place and use `threading.Lock` to lock the  output sections...

shailpatels (approved):
Changes look good - sorry for the delay had some trouble after my test env got in a bad state last week.

But overall seeing the parallel updates work and the existing docker UI page appears to be working the same with uploading and removing images.

I did see that the worker machine does not have its system info present but that maybe due to the output being incomplete / parsed wrong. Its just listed as its default "not yet retrieved" as commented in one of the test files which should be ok for now

Something else is that the script to run commands on a worker can run forever and we don't have a timeout concept, I noticed this a couple of times when I was running multiple VMs and one would hang causing the install process to never complete. I think this is out of scope here but probably should be an issue for future work.

The behavior where if there is an error causes the entire install script to exit early, this is the existing behavior (something i also looked at) so just noting here that won't change as part of this PR 

But other than that worker updates look like they run in parallel well!

shailpatels:
Example of the parallel output:
```
Starting thread for machine: primary. Full output will print when all threads complete.
Starting thread for machine: worker-1. Full output will print when all threads complete.
Description:	Ubuntu 22.04.4 LTS

worker-1: worker daemon is inactive
============Detailed Command Output for worker-1 ============
... rest of output ...
```

PR #8081, opened 2022-06-22
[Dependency] Update Linux packages

akshatbatra:
Linux packages are installed at distro setup and not updated afterwards. Outdated packages often pose security risks.

### What is the new behavior?
Package sources list is updated and packages are upgraded in update_system.sh.


.setup/update_system.sh
```diff
@@ -29,6 +29,9 @@ apt-get install libzbar0 --yes
 
 ##source ${CURRENT_DIR}/distro_setup/setup_distro.sh
 
+# update packages
+apt-get update && apt-get upgrade
```
shailpatels, line 33: I think you would want to put a `-y` after the apt-get upgrade incase it asks you for a prompt to install any upgradeable packages

akshatbatra:
An opt-in flag would be equivalent to not having it at all, because it anyways is a one-line command. It was discussed in the meeting that at some places Submitty may be handled by professors and not professional sys admins and it's easy to forget to upgrade. For now, I've added skip_apt_upgrade. It could be problematic due to being inclined to upgrading. Should it be handled somewhere in CONFIGURE_SUBMITTY.py , so that there can be a default user-defined configuration of whether to upgrade or not which can then be overridden using a flag if needed.

.setup/update_system.sh
```diff
@@ -23,9 +23,28 @@ if [ -d "${CURRENT_DIR}/../.vagrant" ]; then
     VAGRANT=1
 fi
 
+APT=true
+if [ ${CI} == true ]; then
+    APT=false
+fi
```
MasterOdin, line 29: Why disable this on CI in particular?

akshatbatra: e2e tests were failing and I couldn't figure out the issue so tried disabling it on CI as advised but that didn't work, using bash instead of source did, I've reverted this.

MasterOdin:
Yeah, we've had weird issues in the past when using `source` vs `bash`, mostly in having variables bleed from one script to the next, and it hard to track some of the logic of where they're set / used.

akshatbatra:
Thanks for explaining it. I spent a lot of time scratching my head to figure out that this was the issue, ended up learning the hard way.

shailpatels (changes requested):
<strike>Make sure to also update the usage with the new optional command 
https://github.com/Submitty/Submitty/blob/main/.setup/INSTALL_SUBMITTY_HELPER.sh#L81-L95

Also I think the two dashes in the front should be removed so its just `skip_apt_upgrade` and consistent with the other optional commands 
</strike>

This review is no longer relevant after moving to a config file

shailpatels:
I also saw the following prompt in the middle of the install:

```
At least one file of /etc/pam.d/common-{auth,account,password,session} was modified locally. Please indicate if local changes should be discarded and stick with the standard configuration. Otherwise you will have to configure the authentication system by yourself.
```

This required me to manually hit enter, maybe the `DEBIAN_FRONTEND` env var isn't being set correctly or needs to be reset. @MasterOdin do you know if this is an issue or if the PAM config has a specific way they should be upgraded?

akshatbatra:
@shailpatels
I'm not able to reproduce this. Did you run `INSTALL_SUBMITTY.sh` with some particular arguments?
Could `apt-get install -qqy libpam-passwdqc` in `.setup/distro_setup/ubuntu/20.04/setup_distro.sh` be causing this?

shailpatels:
It looked like it was related to the pam authentication libraries used. Maybe I had a bad configuration with mine? I'll try using a fresh VM install to see if I can get that again

MasterOdin:
For Vagrant, we modify the PAM installation at:

https://github.com/Submitty/Submitty/blob/f5054d0069a188fb394b1eafa4a3e9dea2d3b189/.setup/install_system.sh#L465-L466

We do this to allow short weak passwords, and this is definitely not something I'd necessarily expect on a production system. However, depending on how a sysadmin has configured their system, I would certainly expect these sorts of things to be asked from time to time when doing an apt-get update. I suspect @shailpatels vagrant is now borked in that trying to recreate sample courses won't work if he discarded the local changes.

To make this non-interactive, we would need to use `Dpkg::Options` (see https://stackoverflow.com/questions/33370297/apt-get-update-non-interactive), but again, there's risks inherent with bypassing this, and could lead to broken systems depending on what about local vs remote changed.

.setup/update_system.sh
```diff
@@ -23,9 +23,25 @@ if [ -d "${CURRENT_DIR}/../.vagrant" ]; then
     VAGRANT=1
 fi
 
+APT=true
+
+# check if apt upgrades have been disabled
+for flag in "$@"; do
```
MasterOdin, line 29: as opposed to being a runtime flag, I would suggest making it be a flag that gets set in a JSON file, as I think either a sysadmin would always want this or never want it, and not just sometimes not want it.

akshatbatra: The configuration is now in a JSON file with default behavior of not doing upgrades, if enabled, default config is not to override prompts, if it is set to true then the default behavior is to keep the old configuration file. I've kept the flag, maybe this can come handy if one wants to run `INSTALL_SUBMITTY.sh` such that it's guaranteed to be prompt-less at least from the side of `update_system.sh` irrespective of the configuration in `update_settings.json`. `CONFIGURE_SUBMITTY.py` silently creates `update_settings.json`, otherwise, it has to be manually created, this ensures that only those sys admins use this who really want it. If `update_settings.json` doesn't exist, upgrades don't happen.

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -690,6 +690,24 @@ def write(x=''):
...
+
+UPDATE_SETTINGS_JSON = os.path.join(CONFIG_INSTALL_DIR, 'update_settings.json')
+
+if not os.path.isfile(UPDATE_SETTINGS_JSON):
+    config = OrderedDict()
+    config['upgrade_linux_packages'] = False
+    config['linux_packages_upgrade_type'] = "upgrade" #can be dist-upgrade
+    config['autohandle_apt_prompts'] = False
+    #Refer https://stackoverflow.com/questions/33370297/apt-get-update-non-interactive
+    config['autohandle_apt_prompts_option'] = '--force-confold'
```
shailpatels, line 704: I'm not familiar with this option but if it can avoid cmd/gui prompts then I'm fine with it, I would think exporting the variable `export DEBIAN_FRONTEND=noninteractive` would also force it to not show any prompts

MasterOdin: Doing `DEBIAN_FRONTEND=noninteractive` does not stop apt from showing prompts on how to deal with conflicts in the config files for packages.

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -199,6 +200,7 @@ def __call__(self, parser, namespace, values, option_string=None):
     'course_code_requirements': "Please follow your school's convention for course code.",
     'sys_admin_email': '',
     'sys_admin_url': '',
+    'upgrade_linux_packages': False,
```
shailpatels, line 203: Does this option require a migration to the existing installations of submitty's `config/submitty.json` file?

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -311,6 +318,7 @@ def __call__(self, parser, namespace, values, option_string=None):
 
     SYS_ADMIN_EMAIL = get_input("What is the email for system administration?", defaults['sys_admin_email'])
     SYS_ADMIN_URL = get_input("Where to report problems with Submitty (url for help link)?", defaults['sys_admin_url'])
+    UPGRADE_LINUX_PACKAGES = get_input("Should Linux packages be updated with Submitty install?", defaults['upgrade_linux_packages'])
```
shailpatels, line 321: I think this should say something like "Should Linux packages be automatically updated with each Submitty update install" to emphasize this is not a 1 time thing

shailpatels (changes requested):

.setup/CONFIGURE_SUBMITTY.py
akshatbatra: Forgot that, I've added the migration.

.setup/CONFIGURE_SUBMITTY.py
akshatbatra: Thanks for pointing that, the text has been tweaked.

williamjallen:
Closing as per discussion with @bmcutler.  A new page will be added to the documentation with information for sysadmins, including linux package updates.

bmcutler:
instead added basic documentation on updating Linux System:
https://submitty.org/sysadmin/installation/update_server

PR #8086, opened 2022-06-23
[Feature:Autograding] Environment variables in config.json

### LINKED ISSUE #2976: Allow instructors to specify environment variables
Within the config.json for the autograder, it'd be potentially beneficial to allow instructors to specify environment variables to be set within the autograding environment.

poorna2152:

Closes #2976 

### What is the new behavior?
Allows adding enviroment_variables to config.json.e.g.:
```
{
   "enviroment_variables" : {
	"TEST_USER": "test"
   }
}
```


bmcutler (commented):
I haven't tested this, but it looks good!

But can I ask that int this we clean up the existing hardcoded environment variables in the code to be default values in this structure?
If the instructor doesn't specify this in their config.json, those hard coded values are inserted into this json and set by new set_env_variables function.

If the instructor does specify any of those hard coded environment variables (possibly with different values) those values are correctly used instead.

Also, let's add an example to more-autograding_examples.  We don't necessarily need a new example, but maybe just add it to an existing example.

Thanks!

poorna2152:
Skipped adding,
1. `DYNAMORIO_CONFIGDIR` : [link](https://github.com/Submitty/Submitty/blob/main/grading/main_runner.cpp#L150)
2. `DISPLAY`:  [link]( https://github.com/Submitty/Submitty/blob/main/grading/execute.cpp#L1293)

Because they were added much earlier in the process and I was adding env_variables just before executing the command.

prestoncarman:
Is there a way to test this? Or by adding it to the example does it get tested?

bmcutler (approved):
nice work
sorry it took so long to review & merge

PR #8093, opened 2022-06-26
[Bugfix:CourseMaterials] Remove redundant folders 

akshatbatra:
The folders that have no files in the course materials page are not shown in the frontend but they persist inside the filesystem.

Steps to reproduce:
1.) Create a folder by specifying an optional subdirectory such as "outer" in "Upload Course Materials" and upload a file inside it.
2.) Delete that file using the interface, the folder will no more be shown as it has no files.
3.) Go to the relevant course_materials folder in the system e.g. `/var/local/submitty/courses/s22/blank/uploads/course_materials`
4.) See that the folder that was invisible on the interface is still there in the system.

### What is the new behavior?
The redundant empty folders are removed from the file system and also removed using entity manager. This also applies to nested folders such as outer/inner/file.txt. This prevents long empty folder trees from existing in the course materials.

### Other information?
Note for the reviewer: I've tested but it's easy to overlook a special case. So please test for various scenarios especially involving nested folders.

### DIFF
--- a/site/app/controllers/course/CourseMaterialsController.php
+++ b/site/app/controllers/course/CourseMaterialsController.php
@@ -145,7 +145,7 @@ class CourseMaterialsController extends AbstractController {
             FileUtils::getTopEmptyDir($path, $base_path, $empty_folders);
             if (count($empty_folders) > 0) {
                 foreach ($empty_folders as $folder) {
-                    $success = $success & rmdir($folder);
+                    $success = $success && rmdir($folder);
                     if ($success === false) {
                         break;
                     }
--- a/site/app/controllers/course/CourseMaterialsController.php
+++ b/site/app/controllers/course/CourseMaterialsController.php
@@ -137,7 +137,7 @@ class CourseMaterialsController extends AbstractController {
         else {
             $success = unlink($path);
         }
-        $base_path = $this->core->getConfig()->getCoursePath() . "/uploads/course_materials";
+        $base_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), "uploads", "course_materials");
         // delete the topmost parent folder that's empty (contains no files)
         if (pathinfo($path, PATHINFO_DIRNAME) !== $base_path) {
             $empty_folders = [];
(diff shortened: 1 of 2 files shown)


tkoz0 (approved):
Looks good.

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -137,7 +137,22 @@ public function deleteCourseMaterial($id) {
...
+            if (count($empty_folders) > 0) {
+                $path = $empty_folders[0];
+                $success = $success && FileUtils::recursiveRmdir($path);
+                foreach ($all_files as $file) {
+                    if (str_starts_with($file->getPath(), $path)) {
+                        $this->core->getCourseEntityManager()->remove($file);
+                    }
+                }
+                $this->core->getCourseEntityManager()->flush();
+            }
```
cjreed121, line 154: Why remove all files again? Doesn't this get done in 127-131?

cjreed121 (changes requested):

site/app/controllers/course/CourseMaterialsController.php
akshatbatra: `$path` here is different (equal to `$empty_folders[0]`), this hasn't been removed previously.

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -137,7 +137,22 @@ public function deleteCourseMaterial($id) {
         else {
             $success = unlink($path);
         }
-
+        $base_path = $this->core->getConfig()->getCoursePath() . "/uploads/course_materials";
```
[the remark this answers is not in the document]
akshatbatra: Done

site/app/controllers/course/CourseMaterialsController.php
cjreed121: I see.  Thanks for clarification. Couldn't this be simplified to just remove the single top directory? We know it's empty so there shouldn't be any course materials under it.

akshatbatra: The definition of 'empty' here is having no files, I think it could contain empty subfolders inside it, so, to remove them iteration is necessary.

cjreed121 (approved):

PR #8103, opened 2022-06-28
[Bugfix:Autograding] Mermaid Sequence Diagram Fix

Eceptonsu:
The mermaid sequence diagrams for network gradeables have messages overlapping with each other most likely due to broken y values for the elements in the SVG.

### What is the new behavior?
The messages in the mermaid sequence diagrams now have line breaks and other elements have correct y values.
image

### Other information?
This is a temporary fix by adding delays.

### DIFF
--- a/grading/python/submitty_router.py
+++ b/grading/python/submitty_router.py
@@ -111,7 +111,7 @@ class submitty_router():
         outfile.write(f'{start}: {message}\n')
       else:
         str_lines = [str(x) for x in message_lines ]
-        outfile.write(f'{start}: {"<br>".join(str_lines)}\n')
+        outfile.write(f'{start}: {"<br/>".join(str_lines)}\n')
 
       if diagram_label is not None and obj['diagram_label'].strip() != '':
         outfile.write(f'Note over {sender},{recipient}: {diagram_label}\n')
--- a/grading/python/submitty_router.py
+++ b/grading/python/submitty_router.py
@@ -111,7 +111,7 @@ class submitty_router():
         outfile.write(f'{start}: {message}\n')
       else:
         str_lines = [str(x) for x in message_lines ]
-        outfile.write(f'{start}: {"<br/>".join(str_lines)}\n')
+        outfile.write(f'{start}: {"<br>".join(str_lines)}\n')
 
       if diagram_label is not None and obj['diagram_label'].strip() != '':
         outfile.write(f'Note over {sender},{recipient}: {diagram_label}\n')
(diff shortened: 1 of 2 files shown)


grading/python/submitty_router.py
```diff
@@ -111,7 +111,7 @@ def write_sequence_file(self, obj, status, message_type):
         outfile.write(f'{start}: {message}\n')
       else:
         str_lines = [str(x) for x in message_lines ]
-        outfile.write(f'{start}: {"<br>".join(str_lines)}\n')
+        outfile.write(f'{start}: {"<br/>".join(str_lines)}\n')
```
williamjallen, line 114: This change shouldn't be necessary since `<br>` tags don't need to be closed.

site/app/templates/autograding/AutoChecks.twig
```diff
@@ -125,7 +125,8 @@
                   "useMaxWidth" : false
                 };
                 mermaid.initialize(config);
-                mermaid.init( { "useMaxWidth" : false }, "#div_{{ index }}_{{ check_index }}");
+                // TODO: This is a TEMPERARY fix, use callback
```
williamjallen, line 128:
```suggestion
                // TODO: This is a TEMPORARY fix, use callback
```

williamjallen (changes requested):
I noted a couple minor things.  Otherwise, things look good.  How difficult would it be to use a callback as noted in the comment instead of this (quite hacky) approach?  I'm not familiar with this part of the codebase, but it seems like it would likely be a rather simple change.

It's possible that all you need to do is put the `mermaid.init()` call inside a `$( document ).ready()` function so it doesn't get run until the page finishes loading.

grading/python/submitty_router.py
williamjallen: Oops, forgot the suggestion (and apparently can't edit in afterwards)
```suggestion
        outfile.write(f'{start}: {"<br>".join(str_lines)}\n')
```

Eceptonsu:
> I noted a couple minor things. Otherwise, things look good. How difficult would it be to use a callback as noted in the comment instead of this (quite hacky) approach? I'm not familiar with this part of the codebase, but it seems like it would likely be a rather simple change.
> 
> It's possible that all you need to do is put the `mermaid.init()` call inside a `$( document ).ready()` function so it doesn't get run until the page finishes loading.

Thanks for reminding me of this approach, it worked perfectly fine.

williamjallen (approved):

PR #8118, opened 2022-06-30
[Refactor:TAGrading] PDF console errors

Eceptonsu:
In the grading interface, there are console errors related to already defined identifiers when trying to view multiple pdfs in a session. This prevents later pdfs from loading.
There are also warning related to undefined variables.

### What is the new behavior?
Removes the corresponding console errors and users can view multiple pdfs in a session.

### DIFF
--- a/site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
+++ b/site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
@@ -67,17 +67,14 @@
     same could not be achived through CSS as it cause the tools to hide when moved right in panel.
     This has been done taking care of resizing of panel ,platform(device),overlapping and scrolling of panels */
 
-    try {
+    if(typeof offSetter === "undefined") {
         const offSetter = new ResizeObserver(entries => {
             entries.forEach(entry => {
                 let height = $(".sticky-file-info").outerHeight();
-                $("#viewer").css("paddingTop", height);
+                $("#viewer").css("padding-top", height);
             });
         });
         offSetter.observe(document.querySelector(".sticky-file-info"));
-    }
-    catch (e) {
-        // Prevent defining the variable again
     }
 
 </script>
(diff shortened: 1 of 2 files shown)


site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
```diff
@@ -67,12 +67,17 @@
...
+                $("#viewer").css("paddingTop", height);
+            });
         });
-    });
-    offSetter.observe(document.querySelector(".sticky-file-info"));
+        offSetter.observe(document.querySelector(".sticky-file-info"));
+    }
+    catch (e) {
+        // Prevent defining the variable again
+    }
```
williamjallen, line 81: Is there a reason why we need these global variables in the first place?  Ideally this could be better fixed by eliminating the global variables.  If the global variables are absolutely necessary, it would be better to use [`typeof` to check if a variable is defined](https://stackoverflow.com/questions/5113374/javascript-check-if-variable-exists-is-defined-initialized) instead of a try-catch.  As it is now, if another exception is thrown in that block of code, nothing will ever be printed in the console, which will make future development challenging.

site/app/templates/grading/electronic/PDFAnnotationEmbedded.twig
```diff
@@ -67,12 +67,17 @@
...
 
-    const offSetter = new ResizeObserver(entries => {
-        entries.forEach(entry => {
-            let height = $(".sticky-file-info").outerHeight();
-            $("#viewer").css("paddingTop", height);
+    try {
+        const offSetter = new ResizeObserver(entries => {
+            entries.forEach(entry => {
+                let height = $(".sticky-file-info").outerHeight();
+                $("#viewer").css("paddingTop", height);
```
williamjallen, line 74: Is this line doing anything productive?  I believe that the correct way to set the padding on the top would be 
`$("#viewer").css("padding-top", height);`.

williamjallen (changes requested):

Viyerelu23333 (commented):
Looks good to me once William's comments are addressed.  Also tested the `typeof` method and worked on my end.

Issue #8044, opened 2022-06-14
PWA for Submitty


immortalcodes:
**What problem are you trying to solve with Submitty**
Progressive Web App for Submitty
PWAs offer a great solution with their ability to leverage the resource of a website and provide a cross platform experience on any device with almost all features a mobile app has to offer , like push notifications etc.

**Describe the way you'd like to solve this problem**
Creating a PWA for Submitty shall be done in form of checkpoints:
- [ ] Make Submitty installable

This is a dynamic feature request issue and will be added with more checkpoints in future.
Mention the checkpoints in the pull request when linking them to this issue.

cjreed121:
Also see #6269

PR #8121, opened 2022-07-01
[Feature:System] Installable PWA for Submitty

immortalcodes:

Adds a feature as requested in #8044 

### What is the new behavior?
Progressive Web APP for Submitty.
Now the submitty can be installed much like an app on any platform(Mac,Android,ios,linux,etc). To install the application visit the user profile and click on the install app button present over there.

### Other information?
It does not break any feature or prevent you from doing anything that you were able to do earlier.This is in development phase , as there are a lot of feature that we wish to add to make the experince much more enhanced.


site/public/index.php
```diff
@@ -143,3 +143,4 @@ function error_handler() {
 }
 
 $core->getOutput()->displayOutput();
+?>
```
shailpatels, line 146: Was this change intentional? The ending tag for `php` is optional and I think in some cases of how we send the buffer of output we wouldn't want an ending tag here

site/public/manifest.json
```diff
@@ -0,0 +1,32 @@
...
+    {
+    "src": "img/moorthy_duck.ico",
+    "sizes": "72x72 96x96 128x128 256x256"
+    }
+  ],
+
+  "shortcuts": [
+    {
+      "name":"Submitty",
+      "url":"The url you would like to open when the user chooses this shortcut. This must be a URL local to your PWA. For example: If my start_url is /, this URL must be something like /shortcut",
```
shailpatels, line 27: I think these defaults should be updated or removed

site/app/templates/UserProfile.twig
```diff
@@ -132,9 +132,36 @@
...
+       
+        e.preventDefault();
+
+        storedPrompt = e;
+        pwaBtn[0].style.display = 'block';
+        pwaBtn[0].addEventListener('click', (e) => {
+         pwaBtn[0].style.display = 'none';
+         storedPrompt.prompt();
+         storedPrompt.userChoice.then((choiceResult) => {
+         if (choiceResult.outcome === 'accepted') {
```
shailpatels, line 155: Is this if statement required? I don't think this action would need to be logged but it doesn't seem clear what this logic is doing

package.json
```diff
@@ -0,0 +1,5 @@
+{
```
shailpatels, line 1: Is this top-level package.json and package-lock files required? Why wouldn't this package be installed under the `site` directory?

node_modules/.package-lock.json
```diff
@@ -0,0 +1,38 @@
+{
```
shailpatels, line 1: Same comment as the other package-lock files for why this is here and checked in?

shailpatels (changes requested):
A couple of comments on the files checked in and is it possible to use the existing submitty duck image in `site/public/img` and the favicon already there? Is there something specific about the new ones checked in?

Also be sure to take a look at https://submitty.org/developer/testing/linting_static_analysis#php-linting for fixing the failing PHP lint action

immortalcodes:
> Is it possible to use the existing submitty duck image in `site/public/img` and the favicon already there? Is there something specific about the new ones checked in?
They should work fine for most of the resolutions , I just added the new onces because they are bigger in pixels so they won't have problems in adjuting for large screen sizes

site/public/index.php
immortalcodes: This was something I added while trying differnent approaches and I thought it might not create problem to have it there , I will correct this .

site/app/templates/UserProfile.twig
immortalcodes: The upper two lines hides the `add pwa` button and the if statement just logs the successful installation of the pwa in the console. The log in the console could be removed if suggested

immortalcodes:
I have removed the workbox components , they are not necessary as of now and could be added in the later pull requests.

poorna2152:
I tested this on my machine. Looks good to me. Tested on ubuntu 20.04 browser google chrome.
When I first visited the site the install icon was there on the addressbar. When I installed it and visited the site again the icon was missing. Is that the feature. I also wanted to test the installable on Firefox. Similarly I didn't get the install icon on search bar I guess it is also because the I installed the Submitty previously.

site/app/templates/UserProfile.twig
shailpatels: I was referring to the `console.log` in the if/else statement. I don't think that is necessary since the user will most likely never see that and they should see visual feedback from the browser when they either accept or deny the prompt to install the PWA

shailpatels:
> > Is it possible to use the existing submitty duck image in `site/public/img` and the favicon already there? Is there something specific about the new ones checked in?
> 
> They should work fine for most of the resolutions , I just added the new onces because they are bigger in pixels so they won't have problems in adjuting for large screen sizes

Does the specific png resolution matter, the one here is svg so it should scale to whatever if PWAs can support using that as the icon instead of a png
https://github.com/Submitty/Submitty/blob/main/site/public/img/moorthy_duck/00-original.svg

immortalcodes:
> > > Is it possible to use the existing submitty duck image in `site/public/img` and the favicon already there? Is there something specific about the new ones checked in?
> > 
> > 
> > They should work fine for most of the resolutions , I just added the new onces because they are bigger in pixels so they won't have problems in adjuting for large screen sizes
> 
> Does the specific png resolution matter, the one here is svg so it should scale to whatever if PWAs can support using that as the icon instead of a png https://github.com/Submitty/Submitty/blob/main/site/public/img/moorthy_duck/00-original.svg

It's a rule that for a PWA one must have a 512x512 png in the icons.

shailpatels (changes requested):
This PR is looking good and is close to be being merged, I think the last thing that needs to be done is on the user profile page. When a user has already installed the PWA I think it would be good to show that on the page. This can be done with a simple message like "The Submitty PWA has already been installed, see https://support.google.com/chrome/answer/9658361 to uninstall it", that way the user knows its installed.

The second is that the `beforeinstallprompt` event is unsupported by browsers that to do not have PWA support like firefox and safari. I think we should leave a message in that case like "Submitty supports an installable PWA which requires Chrome" so user's know the feature exists. 

I'm not sure if its possible to detect if the specific event is supported or not so maybe the only way is to check the user agent string to display this extra info

shailpatels (changes requested):
Is the new favicon file added different then the one already at `site/public/favicon.ico` ? Or are both needed for the PWA.

Also I was able to test using an SVG with the following manifest files in order to reuse the existing logo. Also it can support multiple sizes like the favicon too if that's needed. I think that would be the better solution since its a smaller file and we can use multiple resolutions if needed in the future

```
{
      "src": "img/moorthy_duck/00-original.svg",
      "sizes": "512x512 256x256 128x128",
      "type": "image/svg+xml",
      "purpose": "any"
    },
```

Viyerelu23333:
I also tested on my end.  Windows Firefox with plugin (PWAsForFirefox) does not prompt for installation in the address bar (perhaps it needs HTTPS?), but works fine for manual installation.  Chrome prompts the icon in the address bar, and it is fully functional.  For Safari in iOS and iPadOS, there are no prompts for installation; but works fine after adding bookmarks to the home screen.  The only problem for Safari is that, the PNG alpha becomes black, which might be an aesthetic issue.

immortalcodes:
I have added few labels that display  different information in different cases :
When installed in chrome:
image

When not installed in chrome:
image


When in other browsers:
image


I was looking to move the js part to some module but then I encountered problems related to SAML after merging the main hence I am working in the detached head for now.
Regarding the SVG part, I guess it might create a problem to use it in android/ios (it was stated in Lighthouse documentation) but if it works then we can replace it.

shailpatels (changes requested):
The latest changes seem to think that any chromium-based browser already has the PWA installed already, instead of showing the button to install which may block users from easily installing it.

I think you would need something like this to detect the app
https://web.dev/get-installed-related-apps/

shailpatels:
> Regarding the SVG part, I guess it might create a problem to use it in android/ios (it was stated in Lighthouse documentation) but if it works then we can replace it.

Could you post the lighthouse error that was causing this? I didn't see anything when running locally but haven't tested on the target device

immortalcodes:
> > Regarding the SVG part, I guess it might create a problem to use it in android/ios (it was stated in Lighthouse documentation) but if it works then we can replace it.
> 
> Could you post the lighthouse error message related to this? I didn't see anything when running locally but haven't tested on the target device

image

shailpatels:
Discussed on 8/5: I wasnt able to reproduce the issue of not having the install button when the PWA was not installed on chrome, my best guess is I was between branches when the PWA was installed causing some issues.

I would say the only thing left is to move the JS into their own `.ts` file so we can perform code checking and linting on it and to see if the `moorthy_duck.png` that is already under the `img` folder can be resized to the expected PWA dimensions instead of converting a SVG to png

shailpatels (approved):
I think this looks good to merge !

Great work :)

PR #8122, opened 2022-07-01
[Testing:CourseMaterials] Add sample course materials

### LINKED ISSUE #6794: Course Materials are not being tested for accessibility
The Course Materials page is not created by default with the sample course, so it does not get tested by the html validators for accessibility.

Course Materials should be enabled in the sample course when running "vagrant up" with a bunch of sample data and files uploaded under different settings. The goal is to feature the different functionalities of the page to increase the coverage of the tests.

akshatbatra:
Fixes #6794, currently there are no sample course materials, so the html checker only runs against an empty course materials page (without any folder/files).

### What is the new behavior?
Sample course materials have been added and are pre-uploaded for the course named 'sample'.

Credit: Thanks @cjreed121 for guidance.


poorna2152:
@akshatbatra
I keep getting this error when doing cypress tests. I only did a `submitty_install` on your branch. Am I missing something on setting this up. (`submitty_install` doesnt seem to be running the `setup_sample_courses.py` is that the case)

Screenshot from 2022-07-07 06-57-54 (1)

akshatbatra:
Yes, I think you need to run `sudo bash /usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/bin/recreate_sample_courses.sh`. @poorna2152

poorna2152:
> Yes, I think you need to run `sudo bash /usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/bin/recreate_sample_courses.sh`.
Did this but still gives me the same error (above) when running cypress tests and when open the sample course from browser it gives me the same error.

akshatbatra:
@poorna2152 
I believe it's been fixed now.

akshatbatra:
There was an issue in deleting the materials, this fixes that.

poorna2152:
Hi @akshatbatra 
Cypress tests are still failing for me.
Screenshot from 2022-07-10 18-29-16
This seems to be trying to click the delete button of a wrong element. Can you check if the behavior is the same for you.

poorna2152:
Hi,
@akshatbatra 
I still have Cypress tests failing. It seems to be clicking the wrong delete button. Can you check if the behavior is the same for you.
Screenshot from 2022-07-10 18-29-16

akshatbatra:
Hi @poorna2152 
I re-ran the tests on my system and they are passing.
Did any of the previous test runs timed out?
From what I observe in the screenshot, first two tests are failing and the state of the page is that it contains folder and files that aren't supposed to be there and should have been cleaned by the previous tests. I think you need to have the exact state (only containing the sample course materials) before beginning a test run. If any run times out, it most likely sets the stage for the upcoming ones to fail (this is at least true for course materials tests).
Before running the tests you need to delete the folder, files and links other than "outer", "words" and "words_1463.pdf". If even a single test fails due to timeout, you should stop the run and go to the page manually and reset the state to only contain sample course materials.

This should fix the issue, if it still persists then it may be the case that running the tests in a bad state resulted in deletion of a material that is expected to be there by the tests. A quick fix for this is to delete all the materials and upload `sample_course_materials.zip` with "Unzip" box checked, also, remember to set the release date to any past date.

akshatbatra:
I just noticed that the sixth test is failing, will commit a fix soon.

poorna2152:
This resolved the issue

poorna2152 (approved):
Changes looks good to me works as expected. e2e test case `test_click_sidebar_links_instructor_blank` in `test_sidebar`  was failing for me locally. I dont think it is related to any of the changes here.

PR #8132, opened 2022-07-03
[Testing:Developer] Add eqeqeq eslint rule

williamjallen:

The Javascript `==` and `!=` operators are generally not recommended because they can behave in [unexpected ways](https://dorey.github.io/JavaScript-Equality-Table/).  Our `eslint` configuration does not currently specify which operators are preferred.  I have recently observed inconsistencies in Javascript code style relating to the equality operator while reviewing other PRs.

### What is the new behavior?
This PR adds the [`eqeqeq`](https://eslint.org/docs/latest/rules/eqeqeq) and rule to our `eslint` configuration.

A documentation PR to update [this page](https://submitty.org/developer/software_and_system_design/coding_style_guide/javascript) will be forthcoming.

### DIFF
--- a/site/public/js/admin-team-form.js
+++ b/site/public/js/admin-team-form.js
@@ -127,7 +127,7 @@ function adminTeamForm(new_team, who_id, reg_section, rot_section, user_assignme
                     team_history_tbody.append(getRowBound(curr_json_entry.admin_user, `Added ${curr_json_entry.first_user}`));
                 }
                 // eslint-disable-next-line eqeqeq
-                else if (curr_json_entry.action == 'admin_create' || curr_json_entry.action == 'admin_add_user') {
+                else if (curr_json_entry.action == 'admin_create' || curr_json_entry.action == 'admin_add_user'){
                     team_history_tbody.append(getRowBound(curr_json_entry.admin_user, `Added ${curr_json_entry.added_user}`));
                 }
                 // eslint-disable-next-line eqeqeq
@@ -135,7 +135,7 @@ function adminTeamForm(new_team, who_id, reg_section, rot_section, user_assignme
                     team_history_tbody.append(getRowBound(curr_json_entry.user, 'Created Team'));
                 }
                 // eslint-disable-next-line eqeqeq
-                else if (user_assignment_setting_json.team_history[j].action == 'admin_remove_user') {
+                else if (user_assignment_setting_json.team_history[j].action == 'admin_remove_user'){
                     team_history_tbody.append(getRowBound(curr_json_entry.admin_user, `Removed ${curr_json_entry.removed_user}`));
                 }
                 // eslint-disable-next-line eqeqeq
@@ -192,7 +192,7 @@ function adminTeamForm(new_team, who_id, reg_section, rot_section, user_assignme
     }
 }
 
-function getTeamHistoryTableRowString(isAfterLockDate, date, user, action) {
+function getTeamHistoryTableRowString(isAfterLockDate, date, user, action){
     return `<tr ${isAfterLockDate ? 'class="admin-team-history-after-lock"' : ''} tabIndex="-1">
         <td class="admin-team-history-td" tabIndex="0">${user}</td>
         <td class="admin-team-history-td" tabIndex="0">${action}</td>
@@ -200,11 +200,11 @@ function getTeamHistoryTableRowString(isAfterLockDate, date, user, action) {
     </tr>`;
 }
 
-function getTeamFormLabelString(for_prefix, text, user_num ) {
+function getTeamFormLabelString(for_prefix, text, user_num ){
     return `<label tabIndex="0" for="${for_prefix + user_num}" style="display:none;">${`${text} ${user_num}`}</label>`;
 }
 
-function getTeamFormReadOnlyInputString(id_prefix, name_prefix, value, user_num, class_string='') {
+function getTeamFormReadOnlyInputString(id_prefix, name_prefix, value, user_num, class_string=''){
     return `<input 
                 tabIndex="0" 
                 id="${id_prefix + user_num}" 
@@ -216,7 +216,7 @@ function getTeamFormReadOnlyInputString(id_prefix, name_prefix, value, user_num,
             />`;
 }
 
-function getTeamFormInputString(id_prefix, name_prefix, user_num) {
+function getTeamFormInputString(id_prefix, name_prefix, user_num){
     return `<input 
                 tabIndex="0" 
                 id="${id_prefix + user_num}" 
@@ -225,7 +225,7 @@ function getTeamFormInputString(id_prefix, name_prefix, user_num) {
             />`;
 }
 
-function getTeamFormButtonString(id_prefix, button_class, text, user_num, onclick, ...onclickArgs) {
+function getTeamFormButtonString(id_prefix, button_class, text, user_num, onclick, ...onclickArgs){
     return `<button 
                 tabIndex="0"
                 id="${id_prefix + user_num}"
@@ -237,7 +237,7 @@ function getTeamFormButtonString(id_prefix, button_class, text, user_num, onclic
             </button>`;
 }
 
-function getTeamFormAddMoreUsersButtonString(user_num) {
+function getTeamFormAddMoreUsersButtonString(user_num){
     return `<button 
                 id="admin_team_form_add_more_users_button"
                 onclick="addTeamMemberInput(this, ${user_num});" 
@@ -250,7 +250,7 @@ function getTeamFormAddMoreUsersButtonString(user_num) {
             </button>`;
 }
 
-function getTeamFormMultipleInvitesWarningString() {
+function getTeamFormMultipleInvitesWarningString(){
     return `<div id="multiple-invites-warning" class="red-message" style="margin-top:3px;width:90%">
                 *Pending members highlighted in pink/red have invites to multiple teams.
             </div>`;
--- a/site/public/js/admin-team-form.js
+++ b/site/public/js/admin-team-form.js
@@ -107,6 +107,7 @@ function adminTeamForm(new_team, who_id, reg_section, rot_section, user_assignme
             $(`#user_id_${i}`).autocomplete('option', 'appendTo', form);
         }
 
+        // eslint-disable-next-line eqeqeq
         if (user_assignment_setting_json != false) {
             const team_history_len = user_assignment_setting_json.team_history.length;
             team_history_tbody.append(getTeamHistoryTableRowString('', user_assignment_setting_json.team_history[0].time, 'N/A', 'Team Formed'));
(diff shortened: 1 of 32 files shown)


MasterOdin:
Could we split this into two PRs, one for `eqeqeq` and one for `space-before-blocks`? I think the latter can be merged pretty instantly as it's just style changes, but the former I think we can do some amount of code changes / longer PR review where we currently just ignore the rule.

williamjallen:
Alright, will do.  Do you have suggestions for other good rules that should be added?  I looked through the list and didn't see anything that stood out, but I'm not familiar with what rules are considered to be common practice/useful.

MasterOdin:
Looking at #8133, I would recommend adding [`@typescript-eslint/type-annotation-spacing`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/type-annotation-spacing.md). As we grow the usage of modules, I would also recommend adding the `import/order` rule from [eslint-plugin-import](https://www.npmjs.com/package/eslint-plugin-import) with following definition:

```js
    'import/order': [
      'error',
      {
        alphabetize: { caseInsensitive: true, order: 'asc' },
        groups: [
          ['builtin', 'external'],
          'parent',
          'sibling',
          'index',
          'object',
        ],
        'newlines-between': 'always',
      },
    ],
```

Otherwise, not sure there's anything in particular that jumps out at me, but some amount of this is just growing the style guide stuff organically as we see people differing on style, and going "can we mandate particular style here?"

site/public/js/admin-team-form.js
```diff
@@ -166,6 +176,7 @@ function adminTeamForm(new_team, who_id, reg_section, rot_section, user_assignme
                 break;
             }
         }
+        // eslint-disable-next-line eqeqeq
         if (found || $(this).val() == '') {
```
MasterOdin, line 180: We could replace this with `===` only if we're confident that we're not relying on entering this if statement when `$(this).val()` is `0` or `[]` as those are both `== ''`

site/public/js/autograding-status.js
```diff
@@ -42,10 +42,12 @@ function updateTable() {
                         new_row.insertCell().innerHTML = course_name[1];
                         new_row.insertCell().innerHTML = key2;
                         const int_box = new_row.insertCell();
+                        // eslint-disable-next-line eqeqeq
                         if (info.interactive != 0) {
```
MasterOdin, line 46: Similar to above, we can replace these only if confident that it cannot be `''` or `[]`, so probably safe to just leave these as disabled for now.

MasterOdin (changes requested):

MasterOdin:
Another rule would be [`prefer-arrow-callback`](https://eslint.org/docs/latest/rules/prefer-arrow-callback) as the standard in most code bases in my experience these days are in using arrow functions.

williamjallen:
`prefer-arrow-callback` is [listed](https://github.com/Submitty/Submitty/blob/a350062895539e5c52e5e2ea3529d7a7f80ff704/site/.eslintrc.json#L31) in our `.eslintrc.json` file but unless I'm misunderstanding the usage of this rule, it doesn't seem to be enforced.  Example: https://github.com/Submitty/Submitty/blob/a350062895539e5c52e5e2ea3529d7a7f80ff704/site/public/js/directory.js#L11

I'll poke around in the `eslint` config some more tomorrow to figure out what's wrong.  I'll also look over the changes in this PR to resolve anything which can be done safely such as the ones requested in an earlier comment.

MasterOdin:
Ah, it's because of the usage of `this` within the anonymous function. I missed that. Solution there would be to work to drop jQuery as a dependency I suppose.

tkoz0:
Why is it that you are leaving quite a few `==` and `!=` as ignored? Is it because of the need to test the changes?

site/public/js/confetti.js
```diff
@@ -16,6 +16,7 @@ function addConfetti() {
         }
     });
     window.addEventListener('keypress', (e) => {
+        // eslint-disable-next-line eqeqeq
         if (e.code === 'Enter' && canvas.style.display != 'none') {
```
MasterOdin, line 20:
```suggestion
        if (e.code === 'Enter' && canvas.style.display !== 'none') {
```

site/public/js/system-update.js
```diff
@@ -140,6 +140,7 @@ function toggleAllReleases(toggleAllButton) {
     //  if toggleAllButton has collapsed - will collapse all
     //  if toggleAllButton doesn't have collapsed - will expand all
     $('.btn-toggle-release').each( (index, button) => {
+        // eslint-disable-next-line eqeqeq
         if ($(toggleAllButton).hasClass('collapsed') != $(button).hasClass('collapsed')) {
```
MasterOdin, line 144:
```suggestion
        if ($(toggleAllButton).hasClass('collapsed') !== $(button).hasClass('collapsed')) {
```

[`$.hasClass`](https://api.jquery.com/hasclass/) only returns a boolean.

site/ts/activity-dashboard.js
```diff
@@ -71,12 +72,15 @@ export function comparator (row1, row2, n, dir) {
...
+    // eslint-disable-next-line eqeqeq
     if (x != '' && y == '') {
         return false;
     }
+    // eslint-disable-next-line eqeqeq
     else if (x == '' && y != '') {
         return true;
     }
+    // eslint-disable-next-line eqeqeq
     if (i == 0 || i == 8) {
```
MasterOdin, line 84:
```suggestion
    if (i == 0 || i === 8) {
```

site/ts/activity-dashboard.js
```diff
@@ -90,13 +94,14 @@ export function helper (x, y, i) {
             return x < y;
         }
     }
+    // eslint-disable-next-line eqeqeq
     else if ((i <= 7 && i >= 4) || i == 9) {
```
MasterOdin, line 98:
```suggestion
    else if ((i <= 7 && i >= 4) || i === 9) {
```

site/ts/activity-dashboard.js
```diff
@@ -209,9 +220,11 @@ export function columnOnClick(n) {
     }
 
     for (i = 0; i < 10; i++) {
+        // eslint-disable-next-line eqeqeq
         if (i != n && $(`#${i}`).children('i').hasClass('fa-angle-up')) {
```
MasterOdin, line 224: All usages of this function pass `n` as a number.

```suggestion
        if (i !== n && $(`#${i}`).children('i').hasClass('fa-angle-up')) {
```

site/ts/activity-dashboard.js
```diff
@@ -209,9 +220,11 @@ export function columnOnClick(n) {
...
             $(`#${i}`).children('i').removeClass('fa-angle-up');
         }
+        // eslint-disable-next-line eqeqeq
         else if (i != n && $(`#${i}`).children('i').hasClass('fa-angle-down')) {
```
MasterOdin, line 228: Same here

```suggestion
        // eslint-disable-next-line eqeqeq
        else if (i !== n && $(`#${i}`).children('i').hasClass('fa-angle-down')) {
```

MasterOdin (changes requested):

site/public/js/user-profile.js
```diff
@@ -284,6 +284,7 @@ $(document).ready(() => {
 
                     // Check user's current time zone, give a warning message if the user's current time zone differs from systems' time-zone
                     const offset = getCurrentUTCOffset();
+                    // eslint-disable-next-line eqeqeq
                     if (response.data.utc_offset != offset) {
```
MBon1, line 288: I was able to get the same result when changing the time zone in the user profile after replacing `response.data.utc_offset != offset` with `response.data.utc_offset !== offset`.

site/public/js/docker_interface.js
```diff
@@ -31,6 +31,7 @@ function filterOnClick() {
         const this_row = $(this);
         let hide = true;
         $(this).find('.badge').each(function () {
+            // eslint-disable-next-line eqeqeq
             if ($(this).text() == this_filter) {
```
MBon1, line 35: In the jQuery documentations, `.data` can return any type.  However, in my testing, it looks like in this function (`filterOnClick`), `.data` is only returning a `string`, the selected filter's name.

site/public/js/forum.js
```diff
@@ -95,7 +95,7 @@ function uploadImageAttachments(attachment_box) {
         const filename = target.attr('fname');
         // eslint-disable-next-line no-undef
         for (let j = 0; j < file_array[part-1].length; j++) {
-            // eslint-disable-next-line no-undef
+            // eslint-disable-next-line no-undef, eqeqeq
             if (file_array[part-1][j].name == filename) {
```
MBon1, line 99: I was able to get the same result when uploading an attachment to a forum post after replacing `file_array[part-1][j].name == filename` with `file_array[part-1][j].name === filename`.  When getting the type of `file_array[part-1][j].name` and `filename`, `string` is returned for both.

site/public/js/forum.js
```diff
@@ -966,6 +974,7 @@ function readThreadStatusValues() {
 
 function dynamicScrollLoadPage(element, atEnd) {
     const load_page = $(element).data(atEnd?'next_page':'prev_page');
+    // eslint-disable-next-line eqeqeq
     if (load_page == 0) {
```
MBon1, line 978: The same behavior was produced when scrolling through the discussion forum posts after replacing `load_page == 0` with `load_page === 0`.  When getting the type of `load_page`, `number` is returned.

site/public/js/forum.js
```diff
@@ -985,6 +994,7 @@ function dynamicScrollLoadPage(element, atEnd) {
         load_page_callback = function(content, count) {
             spinner_down.hide();
             arrow_down.before(content);
+            // eslint-disable-next-line eqeqeq
             if (count == 0) {
```
MBon1, line 998: The same behavior was produced when scrolling through the discussion forum posts after replacing `count == 0` with `count === 0`.  When getting the type of `count`, `number` is returned.

site/public/js/forum.js
```diff
@@ -1006,6 +1016,7 @@ function dynamicScrollLoadPage(element, atEnd) {
         load_page_callback = function(content, count) {
             spinner_up.hide();
             arrow_up.after(content);
+            // eslint-disable-next-line eqeqeq
             if (count == 0) {
```
MBon1, line 1020: While I was unable to trigger this `else` condition, the same behavior would likely occur if `count == 0` is replaced with `count === 0` as the code here is almost identical to the preceding condition.

MBon1 (changes requested):
Including my previous comments, there were a couple of places that I found using `===` instead of `==` could work.  The files include `forum.js`, `docker_interface.js`, and `user-profile.js`.

williamjallen:
> Why is it that you are leaving quite a few `==` and `!=` as ignored? Is it because of the need to test the changes?

Yes.  This PR actually turned out to not be as bad as I expected, but many linting PRs are difficult to test so we usually ignore any pre-existing errors which can't be corrected easily.  In some cases such as the previous requested changes, we can be certain that the change won't break anything.  In other cases, it is sometimes difficult to determine what type a variable might be so we simply ignore the changes.

MasterOdin (approved):
Yeah, I suspect we could resolve each of these by going through one by one, but in the sake of progress, and not letting more of these errors enter the codebase, I would say merge this now, and the remaining ignored errors can be slowly cleaned up as people touch the containing code, or some entrepreneurial soul wishes to clean it up separately.

MBon1 (approved):

PR #8138, opened 2022-07-04
[Documentation:Developer] Remove hack step from UTM README

MasterOdin:

When setting up the VM using UTM, it was necessary to "hack" what dependencies got installed as the old versions of those dependencies were not compatible with arm64 architecture.

### What is the new behavior?

Now that we've moved to Ubuntu 20.04, and so can better support installation of newer versions of these dependencies which have arm64 support, it's no longer necessary to run the hack step that was listed in the UTM README. As such, they've been removed.

I'm not sure if `libgl1-mesa-glx` is still something that's required to be installed or not though.

### Other information?

I'm not sure who other than @bmcutler has used these instructions.

### DIFF
--- a/README_utm_arm.txt
+++ b/README_utm_arm.txt
@@ -157,17 +157,7 @@ On an M1 Mac laptop, we cannot use virtual box, so follow these instructions ins
    sudo umount /usr/local/submitty/GIT_CHECKOUT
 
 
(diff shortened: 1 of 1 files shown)


bmcutler:
oops, forgot you had made this.  Duplicate, handled in PR #8761.

PR #8142, opened 2022-07-04
[Bugfix:Developer] Update minimum PHP version for composer

MasterOdin:

* [x] Documentation has been updated/added if relevant: https://github.com/Submitty/submitty.github.io/pull/440

We had set the `config.platform.php` value in the `composer.json` file to 7.2 so that no matter what version of PHP the host system was using, it would only try to use packages that were 7.2 compatible. Given that we've dropped Ubuntu 18.04 and our new minimum version is 7.4, we can bump the value.

### What is the new behavior?

Set the value of `config.platform.php` to 7.4 to match our new minimum supported version. This will allow installing and updating packages to versions that did not support 7.2 (e.g. phpunit 9.x or twig 3.x). 

### Other information?

Vagrant up run: https://github.com/Submitty/Submitty/actions/runs/2611293412

### DIFF
--- a/site/composer.json
+++ b/site/composer.json
@@ -2,7 +2,7 @@
   "name": "submitty/submitty",
   "config": {
     "platform": {
-      "php": "7.2"
+      "php": "7.4"
     },
     "sort-packages": true,
     "allow-plugins": {


akshatbatra:
Didn't know that minimum version is 7.4
[Core.php Line 147](https://github.com/Submitty/Submitty/blob/af6554eafc8a8a5b547eceb82a6f2383ded19522/site/app/libraries/Core.php#L147) has an error message referring to 7.2, maybe that should also be updated. I was recently misguided by that error message.

PR #8150, opened 2022-07-06
[UI/UX:CourseMaterials] Info messages to ensure correct usage

akshatbatra:
1.) Every user with grading access can view unreleased course materials. 
2.) Hiding a course material isn't a strong protection for preventing students from accessing the material as it can be accessed if the course material ID is known which seemingly is a serial number, if last uploaded course material ID is 119 then next is 120.

### What is the new behavior?
1.) A note has been added to make instructors aware that release dates only apply to students.
2.) In order to make sure that instructors don't use "Hide course materials" as an alternative to setting release dates, an information icon has been added which (on hover) gives an explanation for this.

### Current
old
### New
new


bmcutler:
Stylistically, I suggest a message in red text similar to this message on the "Manage Students" -> (pencil icon) to "Edit Student" modal:

Screen Shot 2022-07-06 at 12 53 19 PM

Add this text if the option is selected.

immortalcodes:
What is the exact purpose of Hiding the course material? (It should be such that students should not be able to access the course material).
Wouldn't implementing a solution where students are not able to guess the id or even after guessing the id their access is not granted be a better solution?

akshatbatra:
The initial proposal was to assign random strings as course materials IDs but that was dismissed as it could make links look weird, links with plain numbers look simpler. The design behind this feature is that the material should be accessible to everybody who has the link, so implementing an access check is out of question. This is a kind of insecure way of hiding course materials but this was not meant to be secure to begin with, therefore, only the warning message has been added so that instructors don't end up using it as an alternative to release date feature.

immortalcodes (approved):
Everything works fine as described

PR #8153, opened 2022-07-07
[Feature:Autograding] Max Submission Size String

### LINKED ISSUE #5712: max_submission_size not correctly carried into build_xxx.json file
**Describe the bug**
Instructors may set the max submission size using the `max_submission_size` flag in a `config.json`.  This information is not being properly carried into the `<course_dir>/config/build/xxx.json`.  Consider the screenshot below in which the `max_submission_size` has been set to 5000000000 bytes.  After gradeable rebuild the information correctly landed in the complete_config directory json, but not the build directory json.

**Expected behavior**
The `max_submission_size` should be identical in both of these configuration files.

**Screenshots**
max_submission_size

**Additional context**
This value gets parsed and used as part of the submission page, so an erroneous value results in students not being able to upload files as large as the instructor has specified.

Viyerelu23333:

Closes #5712 

The issue was fixed in #6815.

### What is the new behavior?
This PR adds a new parser for stringified `max_submission_size`.

### Other information?
- Accepted format: `"\d+ ?([KMGT]i?)B?"`
  - `SI` format (`KB, K`) will be converted using base 1000
  - `IEC-i` format (`KiB, Ki`) will be converted using base 1024
- Testing
  - Manually compile the `configure.out` using `g++`, follow `INSTALL_SUBMITTY_HELPER.sh#L522-568`
  - Run the `configure.out`
  - Use `grep` to parse the output (`jq` cannot deal with `uint64_t`)
  - See details below
- Examples
  - `2333` will not be parsed, and stays as numeric `2333`
  - `"2333"` will be parsed as byte, to numeric `2333`
  - `"233 kib"` will be parsed as IEC-i `Kibibyte`, `238592`
  - `"233 mb"` will be parsed as SI `Megabyte`, `233000000`
  - `"-233k"` will cause `invalid_argument("Wrong format")`
  - `"99999990ti"` will cause `overflow_error()`

<details>

```bash
cd ${SUBMITTY_INSTALL_DIR}
# Manually compile the `configure.out`
GRADINGCODE="${SUBMITTY_INSTALL_DIR}/src/grading"
JSONCODE="${SUBMITTY_INSTALL_DIR}/vendor/include"
g++ "${GRADINGCODE}/main_configure.cpp" \
    "${GRADINGCODE}/load_config_json.cpp" \
    "${GRADINGCODE}/execute.cpp" \
    "${GRADINGCODE}/TestCase.cpp" \
    "${GRADINGCODE}/error_message.cpp" \
    "${GRADINGCODE}/window_utils.cpp" \
    "${GRADINGCODE}/dispatch.cpp" \
    "${GRADINGCODE}/change.cpp" \
    "${GRADINGCODE}/difference.cpp" \
    "${GRADINGCODE}/tokenSearch.cpp" \
    "${GRADINGCODE}/tokens.cpp" \
    "${GRADINGCODE}/clean.cpp" \
    "${GRADINGCODE}/execute_limits.cpp" \
    "${GRADINGCODE}/seccomp_functions.cpp" \
    "${GRADINGCODE}/empty_custom_function.cpp" \
    "-I${JSONCODE}" \
    -pthread -std=c++11 -lseccomp -o "${SUBMITTY_INSTALL_DIR}/bin/configure.out"

# Run the `configure.out` and get output
./bin/configure.out more_autograding_examples/c_failure_messages/config/config.json ./config_out.json c_failure_messages \
&& grep "max_submission_size" config_out.json
```

</details>

### DIFF
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -32,16 +32,18 @@ uint64_t parse_file_size(std::string size_str) {
   // Accepted units, all Bs could be omitted:
   //    Base 1000: B, KB,  MB,  GB,  TB   (SI)
   //    Base 1024: B, KiB, MiB, GiB, TiB  (IEC-i)
-  const std::regex accepted { "[1-9]\\d*([kmgt]i?)?b?" };
+  const std::regex accepted { "\\d+([kmgt]i?)?b?" };
   // check if it is well formatted
   if (!std::regex_match(size_str, accepted)) {
     std::cout << "Got string: " << size_str << std::endl;
-    throw std::invalid_argument("Wrong format: accepted format is [1-9]\\d*([KMGT]i?)?B?");
+    throw std::invalid_argument("Wrong format: accepted format is \\d+([KMGT]i?)?B?");
   }
 
   // Extract numeric file size, will cause exception if the number is larger than 64-bit
   size_t unit_begin  = 0;
   uint64_t file_size = std::stoull(size_str, &unit_begin);
+  if (!file_size)
+    throw std::invalid_argument("Wrong file size: 0 byte");
 
   // Parse the units
   const std::string suffices { "bkmgt" };
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -1890,24 +1890,3 @@ nlohmann::json ValidateANotebook(const nlohmann::json& notebook, const nlohmann:
     }
     return complete;
 }
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -32,21 +32,16 @@ uint64_t parse_file_size(std::string size_str) {
   // Accepted units, all Bs could be omitted:
   //    Base 1000: B, KB,  MB,  GB,  TB   (SI)
   //    Base 1024: B, KiB, MiB, GiB, TiB  (IEC-i)
-  const std::regex accepted { "\\d+([kmgt]i?)?b?" };
+  const std::regex accepted { "[1-9]\\d*([kmgt]i?)?b?" };
+  // check if it is well formatted
+  if (!std::regex_match(size_str, accepted)) {
+    std::cout << "Got string: " << size_str << std::endl;
+    throw std::invalid_argument("Wrong format: accepted format is [1-9]\\d*([KMGT]i?)?B?");
+  }
 
   // Extract numeric file size, will cause exception if the number is larger than 64-bit
   size_t unit_begin  = 0;
   uint64_t file_size = std::stoull(size_str, &unit_begin);
-  // check if it is 0
-  if (file_size == 0) {
-    std::cout << "Got parsed value " << file_size << " from string: " << size_str << std::endl;
-    throw std::invalid_argument("Wrong file size");
-  }
-  // check if it is well formatted
-  if (!std::regex_match(size_str, accepted)) {
-    std::cout << "Got string " << size_str << std::endl;
-    throw std::invalid_argument("Wrong format: accepted format is \\d+([KMGT]i?)?B?");
-  }
 
   // Parse the units
   const std::string suffices { "bkmgt" };
@@ -61,9 +56,9 @@ uint64_t parse_file_size(std::string size_str) {
   // Return the multiplied size, throw if overflowed
   uint64_t multiplier  = static_cast<uint64_t>(std::pow(base, power));
   uint64_t parsed_size = file_size * multiplier;
-  if (file_size != parsed_size / multiplier) {
+  if (file_size != parsed_size / multiplier)
     throw std::overflow_error("Total file size overflowed");
-  }
+
   return parsed_size;
 }
 
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -22,7 +22,7 @@ void json_set_default(nlohmann::json &whole_config, std::string field, const T& 
 
 uint64_t parse_file_size(std::string size_str) {
   // Trim whitespaces
-  size_str.erase(remove_if(size_str.begin(), size_str.end(), isspace), size_str.end());
+  size_str.erase(std::remove_if(size_str.begin(), size_str.end(), isspace), size_str.end());
 
   // Convert string to lowercase, note that it cannot handle UTF-8 characters
   for (char& c : size_str)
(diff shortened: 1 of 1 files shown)


grading/load_config_json.cpp
```diff
@@ -20,6 +20,40 @@ void json_set_default(nlohmann::json &whole_config, std::string field, const T&
...
+
+  // Convert string to lowercase, note that it cannot handle UTF-8 characters
+  for (char& c : size_str)
+    c = std::tolower(c);
+
+  // Extract numeric file size, will cause exception if the number is larger than 64-bit
+  size_t unit_begin = 0;
+  uint64_t file_size = std::stoull(size_str, &unit_begin);
+  // check if it is 0 or contains invalid chars     ====whitelist====
+  if (file_size == 0 || size_str.find_first_not_of("01234567890bkmgti") == std::string::npos) {
```
akshatbatra, line 35: I haven't touched C++ from a long time, just here out of curiosity. Could you explain how this line is working? If this is doing a whitelist check then shouldn't there be `!= std::string::npos` ?

Viyerelu23333: Thanks for catching that.  Will fix in the next commit.

grading/load_config_json.cpp
```diff
@@ -20,6 +20,40 @@ void json_set_default(nlohmann::json &whole_config, std::string field, const T&
...
+  // Trim whitespaces
+  size_str.erase(std::remove_if(size_str.begin(), size_str.end(), isspace), size_str.end());
+
+  // Convert string to lowercase, note that it cannot handle UTF-8 characters
+  for (char& c : size_str)
+    c = std::tolower(c);
+
+  // Extract numeric file size, will cause exception if the number is larger than 64-bit
+  size_t unit_begin = 0;
+  uint64_t file_size = std::stoull(size_str, &unit_begin);
```
poorna2152, line 33: If the `size_str` starts with a letter this line is going to give an error, which is not caught. (example: `k1234`)

grading/load_config_json.cpp
```diff
@@ -20,6 +20,40 @@ void json_set_default(nlohmann::json &whole_config, std::string field, const T&
...
+  if (file_size == 0 || size_str.find_first_not_of("01234567890bkmgti") != std::string::npos) {
+    std::cout << "Got parsed value " << file_size << " from string: " << size_str << std::endl;
+    throw std::invalid_argument("Wrong file size");
+  }
+
+  const std::string suffices = "bkmgt";
+  // Calculate l-shifts
+  size_t lshift = 0;
+  if (unit_begin != size_str.size() // handle strings without unit, treat as byte
+      && (lshift = suffices.find(size_str[unit_begin])) == std::string::npos) {
```
poorna2152, line 44: I think something like `1234 bkmgt` can be parsed through here. It is should give an error.

grading/load_config_json.cpp
Viyerelu23333: I think in this case, the program throws `invalid_argument()` from `std::stoull` as it could not be parsed from `pos = 0`...

grading/load_config_json.cpp
Viyerelu23333: This will be parsed as `1234`, as only `b` is considered as the unit.

poorna2152: Yes but shouldn’t this throw an error because the input has an invalid suffix

grading/load_config_json.cpp
poorna2152: Yes the commad throws an error. Is it possible to use something like `regex_match` function in c++ to test the input value.  It would solve the below problem as well

grading/load_config_json.cpp
```diff
@@ -20,6 +21,52 @@ void json_set_default(nlohmann::json &whole_config, std::string field, const T&
...
+  // Accepted units, all Bs could be omitted:
+  //    Base 1000: B, KB,  MB,  GB,  TB   (SI)
+  //    Base 1024: B, KiB, MiB, GiB, TiB  (IEC-i)
+  const std::regex accepted { "\\d+([kmgt]i?)?b?" };
+
+  // Extract numeric file size, will cause exception if the number is larger than 64-bit
+  size_t unit_begin  = 0;
+  uint64_t file_size = std::stoull(size_str, &unit_begin);
+  // check if it is 0
+  if (file_size == 0) {
```
poorna2152, line 41: Can we use the regex to test whether the file_size is 0 without having a seperate check for that. Is there a reason to have a seperate check for this. May a regex like `[1-9]\d*([kmgt]i?)?b?` this.(Not sure whether this is the correct one.)

grading/load_config_json.cpp
```diff
@@ -20,6 +21,52 @@ void json_set_default(nlohmann::json &whole_config, std::string field, const T&
...
+    std::cout << "Got parsed value " << file_size << " from string: " << size_str << std::endl;
+    throw std::invalid_argument("Wrong file size");
+  }
+  // check if it is well formatted
+  if (!std::regex_match(size_str, accepted)) {
```
poorna2152, line 46: Can we have this check just after conversion to lower_case, so that we dont get an implicit error from `stdoull` when the file_size starts with a string. (eg.`k123`)

grading/load_config_json.cpp
Viyerelu23333: Thank you, I think this is the correct one

poorna2152: Sorry I just noticed this. Something like `0123` is going to fail using this regex. Not sure whether this should be accepted.

Viyerelu23333: Yes, when the number starts with `0`, it is considered as base 8 for `stoull`.  As we are only dealing with base 10, I think it should be fine here.

poorna2152: I tested this. `stoull` for me seems to be giving the output without `0`( eg. `0123` gives `123`)

Viyerelu23333: Maybe try 077?

poorna2152: That also gives me 77.

poorna2152: Did a google search `stoull` default base is base 10. You can change it using the 3rd parameter. I guess you can go with the previous regex and the check for file size zero. Sorry about that change

Viyerelu23333: Oh, my bad, from the documentation, it uses base 10 by default...  I will revert the changes.

poorna2152: Thanks, sorry again for the confusion

PR #8156, opened 2022-07-11
[Refactor:System] Switch to opencv-python-headless

### LINKED ISSUE #8152: Rebuild gradeable stuck on `in rebuild queue`
**Describe the bug**
When `Rebuild Gradeable`  button of a Gradeable is clicked build status is stuck on ` is in the rebuild queue`.

**Expected behavior**
Build status should go into  `Gradeable build complete` after some time.

**To Reproduce**
1. Log in as an instructor
2. Go to a edit gradeable page for any gradeable
3. Click the `Submissions\Autograding` tab
4. Click the `Rebuild Gradeable` button

**Additional context**
Managed to solve this using command `apt-get install libgl1` as suggested by @Viyerelu23333

shailpatels:

opencv-python includes GUI modules that on ubuntu require additional apt packages to be installed

### What is the new behavior?
Since the GUI modules are never used in a server environment this PR switches to using the headless version which includes all core modules but not the GUI for a smaller install and no additional ubuntu dependencies. A migration is included to install the headless for current systems and VMs 

Closes #8152


migration/migrator/migrations/system/20220710211447_switch_opencv_headless.py
```diff
@@ -0,0 +1,22 @@
...
+
+def up(config):
+    #get existing installed packages and check if opencv-python is there
+    #https://stackoverflow.com/questions/44210656/how-to-check-if-a-module-is-installed-in-python-and-if-not-install-it-within-t    
+    installed = {pkg.key for pkg in pkg_resources.working_set}
+    if 'opencv-python' in installed:
+        print("Uninstalling opencv-python")
+        os.system("pip3 uninstall opencv-python --yes --no-input")
+
+    os.system("pip3 install pip install opencv-python-headless==4.6.0.66")
```
MasterOdin, line 13: So a couple things:

1. Can we use `subprocess.check_call` instead of `os.system` so that if these lines fail to do what we want, we're not potentially left in an "invalid" state where we still use `opencv-python` instead of `opencv-python-headless`.
1. I'd suggest using `python3 -m pip ...` to invoke pip as opposed to using a bare `pip3 ...`. It's the recommended way to use it according to [pip's docs](https://pip.pypa.io/en/stable/getting-started/#install-a-package-from-github). It's mainly to accomdiate systems where you may have multiple python installations, and avoid calling the wrong pip, but I think it's a good habit to have nevertheless

MasterOdin, line 13: I think the double `pip3 install pip install` will probably not work...

```suggestion
    os.system("python3 -m pip install opencv-python-headless==4.6.0.66")
```

MasterOdin (changes requested):

migration/migrator/migrations/system/20220710211447_switch_opencv_headless.py
shailpatels: oh whoops added that in twice when adding the specific version

shailpatels:
@MasterOdin I can make those changes - I'll remake this PR I realize I branched off of master originally instead of main and pulling in main led to some weird history

shailpatels:
closing for https://github.com/Submitty/Submitty/pull/8157

PR #8157, opened 2022-07-11
[Refactor:System] Switch to opencv-python-headless

### LINKED ISSUE #8152: Rebuild gradeable stuck on `in rebuild queue`

shailpatels:

opencv-python includes GUI modules that on ubuntu require additional apt packages to be installed

### What is the new behavior?
Since the GUI modules are never used in a server environment this PR switches to using the headless version which includes all core modules but not the GUI for a smaller install and no additional ubuntu dependencies. A migration is included to install the headless for current systems and VMs 

Closes #8152


migration/migrator/migrations/system/20220710211447_switch_opencv_headless.py
```diff
@@ -0,0 +1,31 @@
...
+    #https://stackoverflow.com/questions/44210656/how-to-check-if-a-module-is-installed-in-python-and-if-not-install-it-within-t    
+    installed = {pkg.key for pkg in pkg_resources.working_set}
+    if 'opencv-python' in installed:
+        print("Uninstalling opencv-python")
+        try:
+            subprocess.check_call("python3 -m pip uninstall opencv-python --yes --no-input", shell=True, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            print("Failed to uninstall opencv-python, quitting")
+            print(e.output)
+            return
```
shailpatels, line 16: I'm just `returning` here to exit the migration early and let the rest of the install to continue - but would a better strategy be to raise the exception again to let the user know something went wrong @MasterOdin ?

MasterOdin: What's the error that gets raised if the system call fails? Assuming it's informative enough, I would not bother with a try/catch and just let it throw. We definitely don't want to return early as it'll cause this migration to be marked as run.

shailpatels: I think most of the time it'll be a generic exception - for example using `pip3` instead of `pip` in the command just throws the subprocess exception `subprocess.CalledProcessError: Command 'python3 -m pip3 uninstall opencv-python --yes --no-input' returned non-zero exit status 1. `

I have it printing stderr to stdout which would return a more helpful output of something like `/usr/bin/python3: No module named pip3` above the stacktrace though

MasterOdin: I would maybe then do:

```python
print("Failed to uninstall opencv-python", file=sys.stderr)
raise e
```

Ideally this exception shouldn't be raised that I don't think we need to spend too much time bikeshedding on the ideal messaging to the end user.

Viyerelu23333 (approved):

PR #8161, opened 2022-07-11
[Testing:TAGrading] Multiple pdfs annotations gradable

Eceptonsu:
There are no gradables that test multiple pdf submissions with multiple annotations.

### What is the new behavior?
Set up a gradable that uses multiple pdf submissions with multiple annotations.

### Other information?
The course uses an existing gradable - Released Peer PDF Homework


.setup/bin/setup_sample_courses.py
```diff
@@ -980,6 +1019,47 @@ def create(self):
                                         src = os.path.join(gradeable.lichen_sample_path, gradeable.plagiarism_submissions.pop())
                                         dst = os.path.join(submission_path, str(version))
                                         create_gradeable_submission(src, dst)
+                                elif gradeable.annotated_pdf is True:
+                                    if version == versions_to_submit:
+                                        annotation_version_path = os.path.join(annotation_path, str(versions_to_submit))
+                                        if not os.path.exists(annotation_version_path):
+                                            os.makedirs(annotation_version_path)
+                                            os.system("chown -R submitty_php:{}_tas_www {}".format(self.code, annotation_version_path))
```
williamjallen, line 1027: `os.chown()` might be better than `os.system()` here.

.setup/bin/setup_sample_courses.py
```diff
@@ -980,6 +1019,47 @@ def create(self):
...
+
+                                        if version == versions_to_submit:
+                                            annotations = random.sample(gradeable.annotations, random.randint(1, len(gradeable.annotations)))
+                                            graders = random.sample(assigned_graders, len(annotations)-1)
+                                            # Make sure instructor is responsible for one of the annotations
+                                            graders.append("instructor")
+
+                                            anon_dst = os.path.join(dst, submission).split("/")
+                                            anon_dst[9] = user.anon_id
+                                            anon_dst = "/".join(anon_dst) # has the user id in the file path being anonymous
```
williamjallen, line 1054: Is there a better way to write this?

.setup/bin/setup_sample_courses.py
```diff
@@ -980,6 +1019,47 @@ def create(self):
...
+                                            
+                                            for i in range(len(annotations)):
+                                                annotation_src = os.path.join(gradeable.annotation_path, annotations[i])
+                                                annotation_dst = os.path.join(annotation_path, str(version))
+                                                encoded_path = hashlib.md5(anon_dst.encode()).hexdigest()
+                                                # the file name has the format of ENCODED-ANON-SUBMISSION-PATH_GRADER.json
+                                                annotation_file_name = str(encoded_path) + "_" + graders[i] + ".json"
```
williamjallen, line 1061: Could use an f-string here for simplicity

williamjallen (changes requested):

.setup/bin/setup_sample_courses.py
```diff
@@ -643,15 +672,15 @@ def create_ldap(self):
...
-shadowWarning: 0""")
+            dn: uid={self.id},ou=users,dc=vagrant,dc=local
+            objectClass: top
+            objectClass: account
+            objectClass: shadowAccount
+            uid: {self.id}
+            userPassword: {self.id}
+            shadowLastChange: 0
+            shadowMax: 0
+            shadowWarning: 0""")
```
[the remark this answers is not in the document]
Eceptonsu: The original code was not indented. Basically made it a bit nicer-looking.

.setup/bin/setup_sample_courses.py
Eceptonsu: I was kinda confused about this part because `os.chown()` requires uid and gid, and in this case it seems a specified group is needed. I basically looked at other parts of this file when I wrote this line.

.setup/bin/setup_sample_courses.py
Eceptonsu: I followed the same logic as `getAnonPath()` in other PHP files so I'm not sure if there is a better way.

.setup/bin/setup_sample_courses.py
williamjallen: Given that this line appears elsewhere in this script, and it's already a pretty hacky script which only ever gets run in the development environment, it probably isn't worth the effort to change I suppose.

.setup/bin/setup_sample_courses.py
williamjallen: Part of the point of using `os.path.join()` is that this could theoretically be run on systems which have separators other than `/`.  Thus, it is a little odd to split the path on slashes after using a function to join the path for you.  I suppose this works well enough in the default development environment, so is probably not worth the time to change.

.setup/bin/setup_sample_courses.py
```diff
@@ -1034,8 +1030,10 @@ def create(self):
                                     ]).where(
                                         peer_assign.columns.user_id == user.id
                                     )
-                                    for res in self.conn.execute(stmt):  
+                                    for res in self.conn.execute(stmt):
+                                        #print(res)  
                                         assigned_graders.append(res[1])
+                                    #print(assigned_graders)
```
williamjallen, line 1036: should this commented out code be removed?

williamjallen (changes requested):

.setup/bin/setup_sample_courses.py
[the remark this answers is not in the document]
Eceptonsu: Oh my bad, will change it back now.

Viyerelu23333 (approved):
Looks good to me and worked on my end

PR #8172, opened 2022-07-13
[UI/UX:CourseMaterials] Text tweak in upload form

akshatbatra:
Text was left unchanged in upload course materials form in #8150.

### What is the new behavior?
It has been tweaked to be consistent with edit course material form and edit folder form.


bmcutler (approved):
thanks for fixing this.  Too bad the code is so repetitive to have this text 3 times :(

Issue #7437, opened 2021-12-13
Fix Student Information Column Horizontally to Allow for Easy Marking in Grid Layout


bnidevs:
**What problem are you trying to solve with Submitty**
Student column moves with the grid during grading, so sometimes I have to go back and forth between the student column and the column I want to mark, to make sure I am marking the same row.

I have to do either that, or zoom out very far to the point where it's hard to read the text onscreen.

I am suggesting fixing the student information column to the left side where it is currently, but only horizontally.
So it will scroll vertically with the rest of the grading grid, but stay fixed horizontally.

**Describe the way you'd like to solve this problem**
I think my way of fixing it would be to place the grading grid in it's own table, separate from the student column, but sync the vertical scroll value of both, then remove the `overflow-x` of the student column

**Additional context**
I have attached a video of the current implementation vs what I hope it will look like
https://user-images.githubusercontent.com/33227410/145770278-a15a552f-ed60-4dfc-872c-a3ed858d2995.mov

[youtube video](https://youtu.be/hHXek_VHYec)

PR #8173, opened 2022-07-14
[Feature:TAGrading] Fixed Columns for easy grading

immortalcodes:

Adds the feature as requested in #7437 
Student column moves with the grid during grading,which casuse difficulties in grading as TA/Instructor have to go back and forth between the student column and the column  to mark and  make sure that they are marking the same row.

### What is the new behavior?
Now the student and other info columns remain fixed in the place which makes the grading easier.
image


site/public/css/simple-grading.css
```diff
@@ -29,6 +29,49 @@ input[type="number"]::-webkit-outer-spin-button {
...
+  }
+tr>th:nth-child(5),tr>td:nth-child(5)
+ {
+    position: sticky;
+    left: 16.75rem;
+    border-right: 1px solid var(--standard-medium-gray);
+  }
+  
+  tr:nth-child(odd) td {
+    background: white;
```
williamjallen, line 68: Should this be fixed to white, or should it be changed to a variable listed in `colors.css` like the ones used elsewhere in these changes?

williamjallen (dismissed):

shailpatels (changes requested):
I tested this on windows firefox and chrome and got a strange issue where the animation and color of selecting a checkbox doesn't show anymore.

On main:
before

After installing the branc
after
h

In both browsers I can connect to the websockets server and I do not see any console errors. Are you able to check off students? I tested with a new gradeable created under the sample course

Eceptonsu (approved):
Works for me.

tkoz0:
I'm seeing unexpected behavior with several checkpoints. In this image, I have selected checkpoint 1 with the arrow keys by moving left, but it is hidden behind the student information.
image
I also want to ask what the keyboard controls are supposed to be because I don't think it's clear or stated anywhere.

Eceptonsu:
> I'm seeing unexpected behavior with several checkpoints. In this image, I have selected checkpoint 1 with the arrow keys by moving left, but it is hidden behind the student information. image I also want to ask what the keyboard controls are supposed to be because I don't think it's clear or stated anywhere.

The keyboard works for me. But one thing I noticed is that the title of the table is not at the center with respect to the window, it is located at the center of the entire table.
image
image

immortalcodes:
Actually, the thing is that the fixed columns are overlapping the scrollable columns, this might not be entirely a problem since scrolling through the touchpad works well, there are possible solutions through javascript but looking for something simpler.

immortalcodes:
image
I tried a lot of things but I guess the fix would look  something  like this.
In this, the upper title gets fixed on the left side of the table. 
The problem is with the column span which makes the title scrollable which otherwise would remain fixed.

shailpatels (approved):
Looks good, latest changes move the registration section name to the left side and not centered which I don't think is important since it remains visible the entire time

PR #8179, opened 2022-07-14
[Feature:Calendar] Add Calendar Sidebar Button Out of Debug

MBon1:

 - Currently, the calendar button on the sidebar is only viewable when debugging is enabled.
### What is the new behavior?
 - Even when debugging is not enabled, the calendar button will now appear on the sidebar.
### Other information?
**How did you test**
 - In config/database.json, set the “debugging_enabled” field to false. 
- Log into Submitty as a student, TA, and instructor. Check if each can see the "Calendar" button in the sidebar and if selecting the button sends the user to the Calendar page.
- For each user, check if the "Calendar" button is still visible and usable after going into a course and a Gradeable.


Eceptonsu (approved):

site/app/controllers/GlobalController.php
```diff
@@ -381,15 +381,13 @@ public function prep_user_sidebar(&$sidebar_buttons) {
...
 
-        if ($this->core->getConfig()->isDebug()) {
             $sidebar_buttons[] = new Button($this->core, [
                 "href" => $this->core->buildUrl(['calendar']),
                 "title" => "Calendar",
                 "class" => "nav-row",
                 "id" => "nav-sidebar-calendar",
                 "icon" => "fa-calendar"
             ]);
-        }
```
tkoz0, line 391: Looks good except you should adjust the indentation here.

tkoz0 (changes requested):

tkoz0 (approved):
Looks good

PR #8176, opened 2022-07-14
[Testing:TAGrading] Multi team pdf annotations gradable

Eceptonsu:
### What is the current behavior?
There are no peer team gradables that test multiple pdf submissions with multiple annotations.

### What is the new behavior?
Set up a peer team gradable that uses multiple pdf submissions with multiple annotations.

### Other information?
The name of the gradable is Released Peer Team PDF Homework

PR #8190, opened 2022-07-18
[Refactor:TAGrading] Grading results annotations

Eceptonsu:
When having multiple annotated pdfs, students with different resolutions will see different layouts for the grading results interface, see the image below.
a49a01d47527cfd3acda46b530b5211

### What is the new behavior?
Fixes the layout issue.


site/app/templates/autograding/TAResults.twig
```diff
@@ -77,37 +77,27 @@ $( document ).ready(function() {
...
                                         <a class="btn btn-success key_to_click" style="margin-right:20px;" tabindex="0" onclick="openUrl('{{ student_pdf_view_url }}?filename={{ file.name }}&path={{ file.path|url_encode }}&anon_path={{ file.anon_path|url_encode }}')">View Popup <i class="fas fa-window-restore"></i></a>
-                                        <button class = 'btn btn-default key_to_click' onclick='downloadStudentAnnotations("{{ student_pdf_download_url }}?filename={{ file.name }}&path={{ file.path|url_encode }}&anon_path={{ file.anon_path|url_encode }}", "{{file.path|url_encode}}", "annotated_pdfs")' aria-label="Download the file"> Download
+                                        <button class = 'btn btn-default key_to_click' style="margin-right:20px;" onclick='downloadStudentAnnotations("{{ student_pdf_download_url }}?filename={{ file.name }}&path={{ file.path|url_encode }}&anon_path={{ file.anon_path|url_encode }}", "{{file.path|url_encode}}", "annotated_pdfs")' aria-label="Download the file"> Download
                                         <i class="fas fa-download" title="Download the file"></i></button>
                                     </p>
```
williamjallen, line 95: There's a bunch of inline CSS here, which could be improved by moving it to an external CSS stylesheet.

site/public/css/server.css
```diff
@@ -628,6 +628,24 @@ td > a.active-sort {
...
+    float: right;
+    word-break: break-word;
+    display: inline-block;
+}
+
+.no-border-box .col-md-4 .offset-md-4 {
+    float: center;
+    word-break: break-word;
+    display: inline-block;
+}
```
tkoz0, line 641: The `float` property is conflicting in both of these. VSCode also warns me that `display: inline-block` is ignored because of specifying a value for `float`. I think it would be better to separate these 3 classes differently and avoid specifying something like 2 different values for `float` in 2 different places for the same class.

tkoz0 (changes requested):
I'm still seeing some strange behavior when resizing (the buttons switching between being laid out horizontally/vertically):
image
Also, not exactly related to this PR but I'm noticing that when I go on the "Grading PDF Peer Homework" in the sample course to add annotations, I am unable to save them. I was trying to do this before realizing this was a page style change, not a annotations functionality change.

site/public/css/server.css
Eceptonsu: The saving issue is due to the permission of the annotation folder. I just fixed it in #8176 . Thanks for letting me know.

tkoz0 (approved):
Works as expected now

bmcutler:
@Eceptonsu Please fix merge conflicts :)

PR #8200, opened 2022-07-20
[Feature:System] Add incremental build for tsc

cjreed121:
Right now tsc and esbuild are called for all typescript/mjs code. esbuild is pretty fast but tsc takes a lot of time since it has to typecheck all files from scratch.

### What is the new behavior?
Incremental building was added for tsc to speedup the typecheck time (Results listed below). This adds an `incremental_build` folder inside of `/usr/local/submitty/site` where any incremental build files get stored. This will help developers primarily so they don't have to wait as long for this script to finish.

### Other information?
This lists the time taken to run just a site install of Submitty:
- Before incremental: 21 seconds
- After incremental: 6 seconds

NOTE: You won't feel the effects of the speedup until you have run it at least once because it needs to create the first initial incremental build. After that all builds will be significantly faster.

### DIFF
--- a/site/tsconfig.json
+++ b/site/tsconfig.json
@@ -3,6 +3,8 @@
     "compilerOptions": {
         "strict": true,
         "allowJs": true,
-        "noEmit": true
+        "noEmit": true,
+        "incremental": true,
+        "outDir": "incremental_build"
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ coverage.xml
 /site/cypress/videos/
 /site/public/mjs/
 /site/report/
+/site/incremental_build/
 
 db_update.php
 .setup/data/random_users.txt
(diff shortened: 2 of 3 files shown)


shailpatels (approved):
Looks good, thanks for this !

site/tsconfig.json
```diff
@@ -3,6 +3,8 @@
     "compilerOptions": {
         "strict": true,
         "allowJs": true,
-        "noEmit": true
+        "noEmit": true,
+        "incremental": true,
+        "outDir": "incremental_build"
```
MasterOdin, line 8: Please add this folder to gitignore (`/site/incremental_build/`) since we don't need that checked in for someone who runs tsc outside of the install script.

MasterOdin (changes requested):

site/tsconfig.json
cjreed121: Thanks for catching that

MasterOdin (approved):

PR #8204, opened 2022-07-22
[Bugfix:TAGrading] View/Download Annotations

Eceptonsu:
The view/download functionality for annotated pdfs does not function correctly when in the view grade and grading interface. Annotations created by one user can only be downloaded by the same user. The bug is brought up by #8018. 

### What is the new behavior?
All annotations can be viewed/downloaded.


cjreed121 (approved):
Looks good to me and fixes bug

PR #8211, opened 2022-07-25
[Refactor:Developer] Remove Hardcoded Arguments

Viyerelu23333:

`setup_sample_courses.py` will use hardcoded `SUBMITTY_INSTALL_DIR`, `SUBMITTY_DATA_DIR`, and hardcoded database connections.

### What is the new behavior?

`setup_sample_courses.py` will read the configuration in `SUBMITTY_INSTALL_DIR` from user-defined argument.


tkoz0:
I looked over the code and it looks good, but I did not run it for testing. I have stuff I need for docker autograding so I don't want to recreate sample courses.

sbin/create_course.sh
```diff
@@ -35,10 +35,7 @@ echo $DATABASE_COURSE_USER
 ########################################################################################################################
 ########################################################################################################################
 
-CONN_STRING="-h ${DATABASE_HOST} -U ${DATABASE_USER}"
-if [ -d "${DATABASE_HOST}" ]; then
-    CONN_EXTRA="${CONN_STRING} -p ${DATABASE_PORT}"
-fi
+CONN_STRING="-h ${DATABASE_HOST} -U ${DATABASE_USER} -p ${DATABASE_PORT}"
```
cjreed121, line 39: Does this still work fine if we use a unix socket or regular network connection? As well for the setup_sample_courses script too.

Viyerelu23333: Yes — the default `host` parameter for `psql` is `/var/run/postgresql`, and it also supports URL.
For `setup_sample_courses.py`, the [scheme](https://www.postgresql.org/docs/current/libpq-connect.html#id-1.7.3.8.3.6) `user:pass@host:port/dbname` needs the socket `host` to be percent-encoded.  So `/dbname?parameters=` seems to be more flexible.

bmcutler:
@Viyerelu23333 
Please resolve conflicts

bmcutler (approved):

.setup/bin/setup_sample_courses.py
```diff
@@ -494,6 +512,8 @@ def parse_args():
     parser.add_argument("--courses_path", default=os.path.join(SETUP_DATA_PATH, "courses"),
                         help="Path to the folder that contains .yml files to use for course creation. Defaults to "
                              "../data/courses")
+    parser.add_argument("--install_dir", type=str, default="/usr/local/submitty", help="install path of submitty")
+    parser.add_argument("--data_dir", type=str, default="/var/local/submitty", help="data path of submitty")
```
MasterOdin, line 516: Can use the values of the globals as the defaults so that way don't have two separate places to edit if they change.

```suggestion
    parser.add_argument("--install_dir", type=str, default=SUBMITTY_INSTALL_DIR, help="install path of submitty")
    parser.add_argument("--data_dir", type=str, default=SUBMITTY_DATA_DIR, help="data path of submitty")
```

PR #8216, opened 2022-07-27
[Feature:CourseMaterials] Display folder common properties

### LINKED ISSUE #8201: Display common properties for course materials folders

When recursive updates are applied using edit folder form pop-up on the course materials page, the updated properties are set only for the files inside the folder/subfolders. The properties aren't updated for the folder itself in course materials table in the database. The reason that it is like this is that if the properties are set for folder then unexpected behavior can be encountered if an individual file/folder inside the folder is edited, as then the state of the parent folder (or folders up in the hierarchy) will be inaccurate.

**Video demonstration:**

https://user-images.githubusercontent.com/51257141/180068678-aaf05b5e-a20f-4a1f-a2f9-b6527367701a.mov

A database migration can be made which will give meaning to the existing folder entries by updating them with the common props, this will make sure that the existing course materials also display the common properties.
Changes in `CourseMaterialsController.php`: Each time a file/folder is uploaded/edited, all of its parent folders will be checked for any changes in common props and necessary updates will be made in the database.

**Describe any potential alternative solutions**

This can be handled solely in `CourseMaterialsView.php`. All the folders/files inside a folder can be checked for common props. If the database doesn't reflect the same common props for the folder entry in the table then that can be updated. Only the common props will be kept and rendered as part of edit folder form. The downside is that this will do the same computation on every page reload, it will still run the common props finding code irrespective of the need of database updates.

### LINKED ISSUE #8262: course materials - Hidden from Students null
**Describe the bug**

Stackdump when accessing course materials from a previous term:

15:22:16 08/05/2022 - FATAL ERROR
TypeError (Code: 0) thrown in /usr/local/submitty/site/app/entities/course/CourseMaterial.php (Line 131) by:
        return $this->hidden_from_students;

Message:
Return value of app\entities\course\CourseMaterial::isHiddenFromStudents() must be of the type bool, null returned

Stack Trace:
#0 /usr/local/submitty/site/app/views/course/CourseMaterialsView.php(37): app\entities\course\CourseMaterial->isHiddenFromStudents()
#1 unknown file(unknown line): app\views\course\CourseMaterialsView->listCourseMaterials()
#2 /usr/local/submitty/site/app/libraries/Output.php(229): call_user_func_array()
#3 /usr/local/submitty/site/app/libraries/Output.php(198): app\libraries\Output->renderTemplate()
#4 /usr/local/submitty/site/app/libraries/response/WebResponse.php(43): app\libraries\Output->renderOutput()
#5 /usr/local/submitty/site/public/index.php(142): app\libraries\response\WebResponse->render()

URL: https://submitty.cs.rpi.edu//courses/u20/XXX/course_materials

reported by @holzbh

akshatbatra:

The edit folder form in course materials doesn't show any properties (sections, release date, hidden from students) as marked. The only way to look for properties that are common throughout the folder is to open edit form of each file separately.

### What is the new behavior?
Common properties are now marked in edit folder form on course materials page. This is a client side oriented solution.

## Video demonstration:

https://user-images.githubusercontent.com/51257141/181287968-1807f23c-ba44-47d6-8f13-57a14bf6aabc.mov

The video shows that common section 2 is shown checked in the edit folder popup, hide materials checkbox is only checked when the property is same for all the files. Release date is not shown as it is not common.

### Other information?
Closes #8201 and fixes #8262
Also, a new implementation of hide course materials warning (#8150) has been introduced, this removes the repetition of same warning message in different templates and increases maintainability.


bmcutler:
In the case where there are common settings, but also different settings, it might be nice to visually indicate that some fields are different.
In this example:
Screen Shot 2022-08-01 at 12 34 51 PM

2 is checked blue because both files are for section 2.
But maybe 1 and 3 should be checked in grey, because one of the files is also available for section 1 and the other is also for section 3.

bmcutler:
Also (continuing the example above) if I were to change the date at the folder level (currently the files have different dates), it would remove access to sections 1 & 3 -- a visual indication that these do not match would be helpful to avoid unintentional, destructive edits.

poorna2152:
@akshatbatra Came across #8243 when testing this PR.

site/app/templates/course/EditCourseMaterialsFolderForm.twig
```diff
@@ -43,20 +43,18 @@
...
             <em>Note: Hidden course materials are omitted from the course materials directory listing; however, these materials can still be accessed via the URL.</em>
-            <span id="edit-folder-form-hide-warning" class="red-message full-width" style="display: none;">
-                Warning: Students can view the hidden course material by guessing the corresponding course material ID (a simple number).  It is recommended to instead use the release date feature if it is necessary to prevent students from accessing course materials.
-            </span>
         </label>
         <script>
             function handleHideMaterialsChange() {
                 enableFullUpdate();
-                $('#hide-folder-materials-checkbox-edit:visible ~ #edit-folder-form-hide-warning').toggle();
+                toggleHideMaterialsWarning('hide-folder-materials-checkbox-edit');
```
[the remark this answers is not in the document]
akshatbatra: In other two cases, only the warning had to be shown, but here `enableFullUpdate()` is also present. It looked neat and more readable to keep handling of this one here instead of adding a conditional inside `toggleHideMaterialsWarning` to also call `enableFullUpdate()`.

akshatbatra:
Sections that are selected for some materials now have gray color.
A solution for destructive sections edits has been added:
Gray checkbox represents a partial section, if it is kept unchanged then materials that have that section will persist it. Clicking a partial section checkbox makes it blue (apply for all), clicking it again makes it white (remove for all).
Also, #8243 has been fixed: If yes is selected but no section is checked then `Restrict to at least one section or select 'No'.` alert is shown.

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -395,7 +401,7 @@ public function ajaxEditCourseMaterialsFiles(bool $flush = true): JsonResponse {
                 }
 
                 foreach ($course_material->getSections() as $section) {
-                    if (!in_array($section->getSectionId(), $keep_ids)) {
+                    if (!in_array($section->getSectionId(), $keep_ids) && !in_array($section->getSectionId(), $partial_sections)) {
```
poorna2152, line 404: Changes look good. One thing I didnt understand is why you need the partialSections here. I thought partial sections is solely a front end thing and it wont get inserted into the database. If that is the case do you need the check here. I might be missing something.

akshatbatra: This takes care of the destructive nature of folder edits that was present before.
Assume a folder X having materials A and B.
A: Section 1, Section 2
B: Section 2
Now, if you were to edit folder X and set section 2, then the earlier behavior was that Section 1 would get unset for material A. However, now this will not be the case as sections that are gray colored don't get unset.
But what if you want to have destructive behavior (unset Section 1 and only keep Section 2) ?:
Just click on that section twice, first time it will turn from gray to blue and second time from blue to white (unchecked), now it is not a partial section and will be removed from the materials present inside the folder.

poorna2152: Ah makes sense!.

poorna2152 (approved):
Looks good and changes work as expected

cjreed121 (changes requested):
I noticed a possible issue. If I have a folder with two files inside with one section restricted and the other not, if I try to edit the folder and change the date a JS alert is thrown at me. Is this intentional? The message isn't very clear what I'm supposed to do if it is intentional.

site/app/templates/course/CourseMaterials.twig
```diff
@@ -80,18 +80,18 @@
...
                     {% endif %}
-                    <a class="key_to_click" tabindex="0" onclick='newEditCourseMaterialsForm("{{ course_material.getId }}", "{{ course_material.getPriority }}", {{course_material.getSections|map(s => s.getSectionId)|json_encode|raw}}, "{{course_material.isHiddenFromStudents}}", "{{ course_material.getReleaseDate|date(date_format) }}", "{{ course_material.isLink }}", "{{ url_title }}", "{{ url_url }}")'> <i class="fas fa-pencil-alt black-btn" aria-hidden="true" style="font-size: 16px; margin: 5px;"></i></a>
+                    <a class="key_to_click" tabindex="0" onclick='newEditCourseMaterialsForm(this)' data-id="{{ course_material.getId }}" data-priority="{{ course_material.getPriority }}" data-sections={{course_material.getSections|map(s => s.getSectionId)|json_encode|raw}} data-hidden-state="{{course_material.isHiddenFromStudents}}" data-release-time="{{ course_material.getReleaseDate|date(date_format) }}" data-is-link="{{ course_material.isLink }}" data-link-title="{{ url_title }}" data-link-url="{{ url_url }}"> <i class="fas fa-pencil-alt black-btn" aria-hidden="true" style="font-size: 16px; margin: 5px;"></i></a>
```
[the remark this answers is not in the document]
akshatbatra: This line has been broken down as suggested.

Thanks for testing thoroughly, the alert you faced is intentional but it shouldn't have been thrown, so, it was indeed an issue. It was happening because partial sections weren't getting recognized when applying recursive updates, so that alert said that you should select 'No' radio button when asked about whether you want to restrict the folder to some sections. Please see if you can reproduce that issue now, the alert message has also been tweaked. A couple of Cypress tests have been added. Also, support for partial 'Hide from Students' setting in folder edit form wasn't available earlier, it has been implemented too.

site/app/templates/course/CourseMaterials.twig
```diff
@@ -235,9 +280,86 @@
...
+                    hiddenState = partiallyHidden;
+                }
+                editButton.attr('data-hidden-state', hiddenState);
+            }
+
+            function callFolderSetter(elem) {
+                if (topFolderMatcher.test(elem.id)) {
+                    folderSetter($(elem));
+                }
+            }
```
cjreed121, line 355: Could you add a couple comments on these methods to explain what they're doing.

cjreed121 (changes requested):
Works well for me. Just a quick change for comments for future readability and I think this is good for merge.

cjreed121 (approved):

Issue #7866, opened 2022-04-30
Add "Expand All/Collapse All" Functionality to Gradeable Results


bnidevs:
**What problem are you trying to solve with Submitty**
There is no easy way to open ALL of the gradeable sections other than by pressing "Show Details" for every single section of the gradeable, which becomes frustrating because the rest of the sections shift downward when sections above them are opened.

**Describe the way you'd like to solve this problem**
Add a similar "Show Details" text on the Autograding/Grading Total bar displaying "Expand All/Collapse All" which will open all the sections and close all the sections at the same time.

If some but not all of the sections are opened, "Collapse All" will be displayed and will "Hide Details" on all open sections. "Expand All will only be displayed if all sections are closed. (This is subject to UX/UI discussion, it can be the other way around, but this is just one proposal).

Here is the current appearance:

<img width="1430" alt="Screen Shot 2022-04-29 at 9 15 11 PM" src="https://user-images.githubusercontent.com/33227410/166085037-4450dbb7-9d8c-4669-9ae5-1fae4121fa7c.png">

Here is desired appearance and functionality:

<img width="1444" alt="Screen Shot 2022-04-29 at 9 14 58 PM" src="https://user-images.githubusercontent.com/33227410/166085041-6e96ab80-1cb1-4dce-a197-a8d6ae1a4361.png">

<img width="1432" alt="Screen Shot 2022-04-29 at 9 15 56 PM" src="https://user-images.githubusercontent.com/33227410/166085099-4153c892-a152-414b-9cd0-f54bb250dd01.png">

(Answers to autograder in attached photos are redacted for academic integrity)

akshay2742:
Hello, I'm interested to work on this issue, can you assign me to work upon this

AMMAR-62:
Hey! @akshay2742 , are you working on this issue?? if not may I take up this issue?? @bnidevs please give an update on this

akshay2742:
Still it is not assigned to me

bnidevs:
why are you mentioning me, its pretty clearly already assigned

AMMAR-62:
Sorry for mentioning you! I am still a beginner to open source, don't know the best practices to follow here

PR #8217, opened 2022-07-27
[Feature:Submission] Expand/Collapse All Test Cases

MBon1:

- Feature #7866
  - Currently, there is no way to expand all gradeable test cases other than manually selecting each test case to open them.

### What is the new behavior?
- If a gradeable has at least one test case that can display additional details, then an "Expand All Test Cases" toggle and a "Collapse All Test Cases" toggle will be visible above the test cases.
- When selected, the "Expand All Test Cases" toggle will open all test cases.
- When selected, the "Collapse All Test Cases" toggle will close all test cases.
- If all test cases are opened (either through the toggle or manually), the "Expand All Test Cases" toggle's text will be greyed out until one test case is closed.
- If all test cases are closed (either through the toggle or manually), the "Collapse All Test Cases" toggle's text will be greyed out until one test case is opened.

### Other information?
**How did you test**
In a gradeable that has a couple of test cases with additional details (i.e. `sample > Leaderboard` or `sample > Autograder Hidden and Extra Credit (C++ Hidden Tests)` ) : 
 - Select the "Expand All Test Cases" toggle and check that all test cases are successfully opened.
 - Select the "Collapse All test Cases" toggle and check that all test cases successfully close.
 - Select the "Expand All Test Cases" toggle, and as the test cases are loading, select "Collapse All test Cases".  Check that all test cases were aborted.
 - Manually open each test case in the gradeable and check that the "Expand All Test Cases" is greyed out and the "Collapse All Test Cases" is interactable.
 - Manually close each test case in the gradeable and check that the "Collapse All Test Cases" is greyed out and the "Expand All Test Cases" is interactable.
 - Manually open each test case in the gradeable then as they are loading, cancel each of the test case from loading.  Check that the "Collapse All Test Cases" is greyed out and the "Expand All Test Cases" is interactable.  This was tested on `sample > Leaderboard` as there are only two test cases in this gradeable.

In a gradeable that does not have a test case, check that the "Expand/Collapse All Test Cases" toggles are not displayed.

**Screenshots**
image

image

image


tkoz0 (commented):
I tested this and it works well. The only question I had was wondering if it makes sense to be able to be able to do both expand/collapse all when some of the cases are shown. For example, if I open 5 of 10 test cases and then decide I want to close them all, I can't click 1 button to do that. I could close all 5 individually or open the other 5 and then collapse all 10.

MBon1:
> I tested this and it works well. The only question I had was wondering if it makes sense to be able to be able to do both expand/collapse all when some of the cases are shown. For example, if I open 5 of 10 test cases and then decide I want to close them all, I can't click 1 button to do that. I could close all 5 individually or open the other 5 and then collapse all 10.

In that case, the user could select "Expand All Test Cases" then "Collapse All Test Cases", but regardless, I see your point.  It may be more useful to have both the "Expand All Test Cases" and "Collapse All Test Cases" available.

site/public/js/testcase-output.js
```diff
@@ -20,26 +26,73 @@ function CollapseTestcaseOutput(div_name, index, loadingTools) {
...
  *
- * @global {loading_testcases_xml_http_requests} : Map (Key - string [orig_div_name]; Value - XMLHttpRequest object)
+ * @param {object} set_all_loadingTools - loading tools used by span containing the text that
+ *                                        correspond to expand/collapse all test cases (show, hide)
+ * @param {Array.<JSON>} test_cases - array of test case objects encoded using JSON
+ * @param {boolean} show_hidden - should hidden test cases be displayed
+ * @param {boolean} show_hidden_details - should hidden details be displayed
+ * @param {boolean} are_test_cases_expanded - should the call check if all test cases have been expanded
+ */
+function ChcekStatesOfAllTestcases(set_all_loadingTools, test_cases, show_hidden, show_hidden_details, are_test_cases_expanded) {
```
poorna2152, line 44: spelling `CheckStatesOfAllTestcases`

MBon1: Spelling fixed.

poorna2152:
@MBon1 When I tested the functionality on a gradeable where the title of the validation contains a `'` this didn't seem to work. Following error was shown in the console. Can you check if you can reproduce this.  In the screenshot below the second title `C++'s execution`  contains a `'` then this doesn't work but when the `'` is removed this works. I'm guessing this has something to do with the encoded `test_cases` array passed to function `loadTestcaseOutput`.
Screenshot from 2022-08-07 09-56-32

tkoz0 (approved):
Works great for me

williamjallen (commented):
Should some e2e tests be added to make sure this doesn't break in the future?

site/public/js/testcase-output.js
```diff
@@ -82,22 +173,62 @@ function loadTestcaseOutput(div_name, gradeable_id, who_id, index, version = '')
...
+    // Expand/Collapse all test cases
+    for (let index = 0; index < num_test_cases; index++) {
+        // Check if test case has loading tools (can be expanded/collapsed)
+        if ($(`#tc_${index}`).find('.loading-tools').length === 1) {
+            const test_case_div_name = `testcase_${index}`;
+            const isTestCaseLoadingToolsShowVisible = $(`#tc_${index}`).find('.loading-tools').find('.loading-tools-show').is(':visible');
+
+            // Check if test case should be expanded/collapsed
+            if ((expand_all && isTestCaseLoadingToolsShowVisible) || (!expand_all && !isTestCaseLoadingToolsShowVisible)) {
+                loadTestcaseOutput(test_case_div_name, gradeable_id, who_id, index, num_test_cases, version);
```
poorna2152, line 230: Execution comes to this function when `ExpandAll` or `CollapseAll` is clicked. Through this function `CheckStatesOfAllTestcases()` function is called for each test case and that tests state of each test case. Is that needed because you know after ExpandAll/CollapseAll is clicked state of the all the test cases is going to be same. You could pass some sort of a boolean to stop that function from executing.

poorna2152 (approved):
Works as expected and changes look good to me

bmcutler:
@williamjallen
We discussed e2e tests for this feature in the meeting and this may need more work because autograding is skipped for github actions e2e testing.

PR #8031, opened 2022-06-12
[SYSADMIN ACTION][SECURITY][Bugfix:System] Add Course User for Postgres

cjreed121:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
Submitty has a single database user which is a postgres superuser. This is used for all queries on the site but it gives too much permission for features such as SQL toolbox.

### What is the new behavior?
A new postgres user was added that is strictly for courses and is limited to basic SQL commands on the public schema of each course database.

### Other information?
Sysadmin action documentation not complete yet. Will update PR once complete.

PR #8226, opened 2022-07-29
[Bugfix:Developer] Default privileges for course db user

### LINKED ISSUE #8209: Add default permissions for course database user
After PR https://github.com/Submitty/Submitty/pull/8031, a course database user was added and all permissions were set for current tables. Any new table will still need permissions set.

We should make a migration to set the default permissions on the schema so that the course database user will have the correct permission for any new table. Make sure to do this for tables AND sequences.

The alternative is that we are stuck setting permissions on every migration that adds a new table/sequence. PR https://github.com/Submitty/Submitty/pull/8002 had to do this but it shouldn't be necessary.

akshatbatra:
After PR #8031, a course database user was added and all permissions (SELECT, INSERT, UPDATE, DELETE) were set for current tables. But new tables still need permissions to be set when they are added, PR #8002 had to do this.

### What is the new behavior?
 A migration has been added to set the default permissions on the public schema so that the course database user will have the correct permissions for newly added tables too, this change has also been made for sequences.

### Other information?
Closes #8209


migration/migrator/migrations/course/20220729110000_course_dbuser_default_privileges.py
```diff
@@ -0,0 +1,42 @@
...
+    :type course: str
+    """
+    database_file = config.config_path / 'database.json'
+    with(database_file).open('r') as db_file:
+        db_info = json.load(db_file, object_pairs_hook=OrderedDict)
+        database.execute("ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO {}".format(db_info['database_course_user']))
+        database.execute("ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, UPDATE ON SEQUENCES TO {}".format(db_info['database_course_user']))
+
+
+def down(config, database, semester, course):
```
bmcutler, line 25: Down migration is unnecessary (may unnecessarily cause problems).  Just leave the down migration empty.

williamjallen:
The linting errors should be fixed once main is merged into this PR.

PR #8242, opened 2022-08-01
[Testing:Developer] Fix Python linting errors

williamjallen:
A recent update to our Python linter, `flake8`, seems to have caused many of the recent CI jobs to fail.

### What is the new behavior?
Fixes the linting errors.

### DIFF
--- a/bin/set_allowed_mins.py
+++ b/bin/set_allowed_mins.py
@@ -64,8 +64,8 @@ def send_data(db, allowed_minutes, override):
     db.execute(text(query), gradeable=GRADEABLE)
     if override is not None:
         for user in override:
-            query = """INSERT INTO gradeable_allowed_minutes_override (g_id, user_id, allowed_minutes)  # noqa: E501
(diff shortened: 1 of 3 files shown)


bin/set_allowed_mins.py
```diff
@@ -64,7 +64,7 @@ def send_data(db, allowed_minutes, override):
     db.execute(text(query), gradeable=GRADEABLE)
     if override is not None:
         for user in override:
-            query = """INSERT INTO gradeable_allowed_minutes_override (g_id, user_id, allowed_minutes)
+            query = """INSERT INTO gradeable_allowed_minutes_override (g_id, user_id, allowed_minutes)  # noqa: E501
```
bmcutler, line 67: can you put a comment within a multi-line string?
this seems hazardous

williamjallen: Yep, good catch.  I changed the comment to use single quotes and then reverted it without thinking.  I believe that I have fixed the issue.

bmcutler (approved):

PR #8249, opened 2022-08-02
[Bugfix:Forum] Grey Scrollbar Up Arrow at Top of Forum

MBon1:

 - If the scroll bar is at the upper end of the thread list, the up arrow on the scroll bar is not greyed out.
image

### What is the new behavior?
 - If the scroll bar is at the upper end of the thread list, the up arrow on the scroll bar will now be greyed out.
image

### Other information?
**How did you test**
 - Go to a "Discussion Forum" page (I tested in the Discussion Forum in "sample" course)
 - Scroll to the top of the thread list and check if the up arrow on the scrollbar is greyed out. 
 - Select a Forum Thread. Scroll to the top of the thread list and check if the up arrow on the scrollbar is greyed out.


poorna2152 (approved):
Works as expected

PR #8251, opened 2022-08-02
[Feature:Autograding] Tolerance check in diff validator

### LINKED ISSUE #1362: autograde compare a file of numbers with a tolerance
For example, a matlab output file with a vector or matrix of floating point numbers.  Reasonable partial credit if the size of the vector or matrix is correct, and the numbers are within instructor specified tolerances.

poorna2152:

Closes #1362 

### What is the new behavior?
Adds a parameter to `diff` validator called `tolerance`.  Tolerance could be set to a specified number. If the numerical values in the student submissions deviates from expected output but is still within the specified tolerance student gets full marks. (Similar to ignoreWhitespaces)

edit to add image of larger example:
Screenshot 2024-07-01 at 11 55 49 PM


bmcutler:
I ran these tests with the example provided and I don't see a difference between the different grading methods so its hard to understand what the differences are.  Maybe the example needs some work, or maybe the feature isn't working as I expect.

poorna2152:
for the example added in `more_autograding_examples`,
Tolerance
Screenshot from 2022-08-18 19-38-54

Buggy
Screenshot from 2022-08-18 19-36-46

bmcutler:
This PR appears to have good progress, but it's not ready to go right now.

As an example, this diff highlighting visual is confusing me:
Screenshot 2024-06-28 at 1 30 09 PM

(and it would confuse a student about what was wrong and needed to be fixed).

I'm going to expand and improve the example to make a more clear list of what needs to be addressed.

bmcutler (approved):
I think this is ok to merge now
(well once we get the Cypress tests fixed)

PR #8254, opened 2022-08-03
[Feature:Submission] Expected String Rework

Eceptonsu:
Closes #6722
Instructors can only specify answers by providing a file and using the "expected_file" option on the assignment config.

### What is the new behavior?
Allowing the usage of "expected_string" instead, whereby instructors can just enter the correct answer in a string.

### Other information?
This PR is based on @eddiekrystowski's previous work.


grading/load_config_json.cpp
```diff
@@ -1093,6 +1093,17 @@ void InflateTestcase(nlohmann::json &single_testcase, nlohmann::json &whole_conf
...
+      }
+      if (grader.find("expected_string") != grader.end()) {
+        itr2 = grader.find("show_expected");
+        if (itr2 == grader.end()) {
+          grader["show_expected"] = "always";
+        } else {
+          assert (validShowValue(*itr2));
+        }
+        grader["use_expected_string"] = true;
+        assert(grader.find("expected_file") == grader.end());
```
Viyerelu23333, line 1106: Adding an error message here would be helpful

```suggestion
        assert(grader.find("expected_file") == grader.end() && "You cannot specify both expected_file and expected_string");
```

grading/TestCase.cpp
```diff
@@ -188,8 +189,30 @@ bool openStudentFile(const TestCase &tc, const nlohmann::json &j, std::string &s
...
+    filename = "test_output/" + expected_file;
+    std::ofstream expected_file_stream(filename);
+    expected_file_stream << expected_string;
+    expected_file_stream.close();
+  } else {
+    filename = getOutputContainingFolderPath(tc, expected_file) + expected_file;
+  }
+  
   if (filename == "") {
     messages.push_back(std::make_pair(MESSAGE_FAILURE,"ERROR!  EXPECTED FILENAME MISSING"));
```
Viyerelu23333, line 217: If both `expected_string` and `expected_file` are missing, the `filename` will be `test_output/AUTOGENERATED_`; so the `if` statement below cannot catch the error.

Viyerelu23333 (approved):

PR #8256, opened 2022-08-03
[Dependency] update AnalysisToolsTS v22.08.01

bmcutler:


poorna2152:
Updated installation for AnalysisToolsTS `v22.08.01`

poorna2152:
Do we need a migration here to remove existing node-tree-sitter files

poorna2152 (approved):

PR #8261, opened 2022-08-05
[Bugfix:Developer] Anon id migration fix

akshatbatra:
#8002 was recently merged, `anon_id` column was removed from the users table and `setup_sample_courses.py` was modified to directly set `anon_id` in `gradeable_anon`. Accompanying migration extracted `anon_id` from users table to insert inside `gradeable_anon` but there may be users with null `anon_id` when migration runs, so running `submitty_install` currently throws an error.

### What is the new behavior?
A check has been added to move `anon_id` only if the column exists and is not null. An issue identified by Jerry (@Viyerelu23333) has also been fixed.

### DIFF
--- a/migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
+++ b/migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
@@ -37,19 +37,10 @@ def up(config, database, semester, course):
         database.execute("GRANT SELECT, INSERT, UPDATE, DELETE ON gradeable_anon TO {}".format(db_info['database_course_user']))
 
     database.execute("""
-        DO
-        $do$
-        BEGIN
-        IF EXISTS (SELECT column_name
-                   FROM information_schema.columns
-                   WHERE table_name='users' and column_name='anon_id') THEN
-               INSERT INTO gradeable_anon (
-                    SELECT u.user_id, g_id, u.anon_id
-                    FROM gradeable g JOIN users u ON 1=1 WHERE NOT EXISTS (SELECT 1 FROM gradeable_anon WHERE user_id=u.user_id AND g_id=g.g_id) AND u.anon_id IS NOT NULL
-                );
-            END IF;
-        END
-        $do$
+        INSERT INTO gradeable_anon (
+            SELECT u.user_id, g_id, u.anon_id
+            FROM gradeable g JOIN users u ON 1=1 WHERE NOT EXISTS (SELECT 1 FROM gradeable_anon WHERE user_id=u.user_id AND g_id=g.g_id) AND u.anon_id IS NOT NULL
+        );
     """)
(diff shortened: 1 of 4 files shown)


migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
```diff
@@ -36,28 +36,22 @@ def up(config, database, semester, course):
...
-    :param database: Object for interacting with given database for environment
-    :type database: migrator.db.Database
-    :param semester: Semester of the course being migrated
-    :type semester: str
-    :param course: Code of course being migrated
-    :type course: str
-    """
-    database.execute("ALTER TABLE users ADD COLUMN IF NOT EXISTS anon_id character varying(255);")
-    database.execute("UPDATE users u SET anon_id=(SELECT ga.anon_id FROM gradeable_anon ga WHERE ga.user_id=u.user_id LIMIT 1);")
+    pass
```
cjreed121, line 57: Why remove these? If the PR gets reverted for some reason and this down migration was run, the old code wouldn't work.

cjreed121 (changes requested):

migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
akshatbatra: I was not feeling sure about side-effects of running down migration, as gradeable specific `anon_id` could be used in file paths and switching back to old model would use different `anon_id` than what was used to create the file. But not being able to use Submitty at all is a bigger problem than some less probable side-effects, so, I've added it back.

migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
```diff
@@ -36,15 +36,22 @@ def up(config, database, semester, course):
...
+        IF EXISTS (SELECT column_name
+                   FROM information_schema.columns
+                   WHERE table_name='users' and column_name='anon_id') THEN
+               INSERT INTO gradeable_anon (
+                    SELECT u.user_id, g_id, u.anon_id
+                    FROM gradeable g JOIN users u ON 1=1 WHERE NOT EXISTS (SELECT 1 FROM gradeable_anon WHERE user_id=u.user_id AND g_id=g.g_id) AND u.anon_id IS NOT NULL
+                );
+            END IF;
+        END
+        $do$
```
cjreed121, line 52: Since we can guarantee the column exists, can we simplify this like how it was before?

cjreed121 (changes requested):

migration/migrator/migrations/course/20220612122726_gradeable_specific_anon_ids.py
akshatbatra: Done

cjreed121 (approved):

PR #8264, opened 2022-08-07
[Bugfix:Autograding] Success, information messages not shown

### LINKED ISSUE #5621: Custom Validator success message not shown
**Describe the bug**
Messages of type "success" and "information" are not shown if the testcase has full points and there is no warning/error, even if `show_message` is set to "always".

**Expected behavior**
Do show success/information messages.

**To Reproduce**
1. Create a testcase with a custom validator that returns a result as shown [here](https://submitty.org/instructor/assignment_configuration/validation#success) (so status is success, full points, and a success-message).
2. Run an evaluation.
3. See the successful testcase with no message.

**Additional context**
The problem happens during evaluation, as the `results.json` already has no message for the autochecks.
If the score is 0 instead of 1, the message is shown.
If the message is in an array, it is not shown.
If the message is in an array together with a warning/error, it is shown.
If the message has type warning/error instead, it is shown.

The documentation says "Status influences both the color of a message and when it will be displayed to the student.", but it doesn't specify that further. However, if the current behavior was intended, the examples [here](https://submitty.org/instructor/assignment_configuration/validation#success) would make no sense as the messages would never be shown.

poorna2152:

Fixes #5621 

### What is the new behavior?
Success and information messages shown when `show_messages: always` is set in config.json. Custom validator success and information messages shown.

### DIFF
--- a/grading/main_validator.cpp
+++ b/grading/main_validator.cpp
@@ -199,7 +199,7 @@ double ValidateAutoCheck(const TestCase &my_testcase, int which_autocheck, nlohm
     }
 
     std::string fm = tcg.value("failure_message","");
-    if (!test_case_success) {
+    if (!test_case_success || show_message) {
       bool failure_message_already_added = false;
       if (FN==0) {
         for (int m = 0; m < messages.size(); m++) {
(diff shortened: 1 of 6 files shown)


bmcutler (approved):
Thanks!  looks good!

PR #8269, opened 2022-08-08
[Testing:Developer] Automated ShellCheck linting

williamjallen:

Submitty's shell scripts currently have no static analysis/style enforcement tools being run on them on a regular basis.

### What is the new behavior?
This PR adds a new CI job which runs ShellCheck on Submitty's shell scripts.  ShellCheck does not support a list of ignored files like most linters so a somewhat hacky attempt at recreating the same functionality was devised in a short Python script. For checking individual files, the command `shellcheck <filename>` can be run.

### Other information?
An associated documentation PR will be created soon.


MasterOdin (commented):
What would be the penalty of running shellcheck per file vs once for all files, as that might allow using something like find / xargs, instead of using bash 4+ features.

Alternatively, could write the script in a python wrapper if desire to make this easier to run out of the box for people who's OS doesn't come with bash 4 (e.g. macOS).

williamjallen:
I don't think there would be much of a downside to running ShellCheck more than once.  The only thing I can think of is that the information message (see screenshot for an example) at the bottom of the run would potentially be duplicated if multiple files have errors.  
image

I wasn't able to get recursive path globbing to work properly with find / xargs.  If you have suggestions regarding how to do it cleanly, that would definitely be a better option.

Perhaps if there isn't a clean way to rewrite in a Bash <4-compatible way I should just rewrite in Python..

Viyerelu23333:
Not pretty sure if this would fit into the problem, but I think it is possible with `find`.  Might require some modifications to the `.shellcheckignore`.

```bash
# find files ending with '.sh', and prune the contents in ./tests or ./.git or ./site
find . \! \( -type d \( -path './tests' -o -path './.git' -o -path './site' \) -prune \) -name '*.sh'
# also it's possible to use \! -path './path/to/sh/file' to ignore files, but quite bulky here
```

Then, get the unique lines from the `find`, comparing to `.shellcheckignore`

```bash
# compare sorted files and suppress the unique lines from file 2 and both file 
# (only print unique lines for file 1, the find command above)
comm -23 <(COMMAND_ABOVE | sort) <(cat .shellcheckignore | sort)
```

Finally, run the `shellcheck`

```bash
COMMAND_ABOVE | xargs shellcheck
```

williamjallen:
That solution is definitely possible, but it fails to provide support for a list of files similar to other ignore files such as `.rsyncignore`, `.eslintignore`, etc.  It's likely that additional scripts that we want to ignore will be added in the future so it's much easier to modify a list of paths than modify a somewhat complex script.

run_shellcheck.py
```diff
@@ -0,0 +1,26 @@
...
+
+all_shell_scripts = set(Path('.').rglob('*.sh'))
+
+shell_scripts_to_check = [str(x) for x in all_shell_scripts.difference(ignored_files)]
+
+print(['shellcheck'] + shell_scripts_to_check)
+
+return_code = 0
+for script in shell_scripts_to_check:
+    process = subprocess.run(['shellcheck', script], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
Viyerelu23333, line 21:
```suggestion
    process = subprocess.run(['shellcheck', '-Calways', script], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```

Maybe also print out standard error here?  Also consider adding color output?

Viyerelu23333 (approved):
Looks good to me, and worked on my end

run_shellcheck.py
williamjallen: When I tried adding STDERR, ShellCheck only printed newlines to stderr for both successful and unsuccessful runs.  I'll see if I can add it such that the interleaving with STDOUT remains sensible.  I also tried adding color output but I was unable to get it to work after several hours of work on it.  If you have suggestions, I'm definitely open to trying them.

MasterOdin: Adding `-Calways` as suggested works for me in getting it to have color. I don't think there's any way to do it otherwise.

> When I tried adding STDERR, ShellCheck only printed newlines to stderr for both successful and unsuccessful runs. I'll see if I can add it such that the interleaving with STDOUT remains sensible. 

It looks like shellcheck only outputs to stdout in all cases, and so I think makes sense to just ignore stderr completely.

We should wrap `process.stdout.print` in a check that it's not empty as well so that we don't get an empty line per checked file that's good.

bmcutler:
@williamjallen 
Looks like there is still unaddressed feedback from @MasterOdin

MasterOdin (approved):

MasterOdin:
It would probably be good to add a page under https://submitty.org/developer/software_and_system_design/coding_style_guide/index for shell code style guide that mentions we follow the shellcheck guidelines, how to run the linter, and just mention some of the more common things that might trip people up (e.g. having to always use `${...}` or whatever the rule is).

PR #8274, opened 2022-08-09
[Bugfix:CourseMaterials] Overlapping Buttons in UI

### LINKED ISSUE #8248: Mispalced buttons in Moblie UI for Course Material Page
**Describe the bug**
The buttons in the mobile view of the course material page overlap .

**Expected behaviour**
The buttons should not overlap.
**To Reproduce**
Steps to reproduce the behaviour:
1. Visit the course material page
2. In the inspect window, click the mobile view
3. See error

**Screenshots**
image

immortalcodes:

Fixes #8248 
Currently, the buttons overlap each other.
### What is the new behavior?
After the addition of the margin, they don't overlap.
image


site/public/css/course-materials.css
```diff
@@ -32,3 +32,7 @@ a.popout-item > .fas:hover {
     display: flex;
     flex-wrap: nowrap;
 }
+
+.btn.btn-primary {
+    margin-top: 4px;
+}
```
williamjallen, line 38: This seems like the wrong fix for this problem.  What if there was a different primary button somewhere else on the page?  It would also get this new margin.  It would be better to add a new class to each of the buttons, or better yet, use flexbox properties to do the wrapping cleanly.

williamjallen (changes requested):

shailpatels (approved):
Looks good, changes work as described

PR #8276, opened 2022-08-11
[Feature:Autograding] Add submitty_count_ts command

### LINKED ISSUE #1458: Write more test cases for submitty_count
For all supported languages
For node/token/function

Put them in more_autograding_examples?

### LINKED ISSUE #1673: Use a different parsing backend for Python static analysis
We currently use [language-python](https://hackage.haskell.org/package/language-python) as the backend for Python static analysis. This package is no longer actively maintained and does not support some modern Python 3 features (see #1613 ). Furthermore, this package is not included in the latest [Stackage LTS set](https://www.stackage.org/lts-10.3), preventing the project from updating and gaining access to bug-fixes and library ecosystem improvements.

This coincides with the upcoming integration of the Common AST backend, which will also require dispatch to an external tool for parsing.

### LINKED ISSUE #4701: F-Strings break function call counts
**Describe the bug**
When a submitted program written in python contains an f-string, the program is unable to count function calls in the program.
**Expected behavior**
1. Write a program which contains f-strings
2. Submit the program
3. The number of calls of a given function should be counted
**To Reproduce**
1. Submit a program which uses f-strings and a given number of calls of a function 
2. Have submitty attempt to count this
3. See error

**Configuration**
 - OS: macOS
 - Browser: Safari/Chrome

**Screenshots**
https://imgur.com/a/at9hKES

**Submitty Error**
[91mParser error: UnexpectedToken (StringToken {token_span = SpanCoLinear {span_filename = "/var/local/submitty/autograding_tmp/untrusted00/tmp/TMP_WORK/test02/part2/check2.py", span_row = 22, span_start_column = 8, span_end_column = 34}, token_literal = "'{len(s_lname)} last names'"})

poorna2152:

Closes #4701 
Closes #1673 
Closes #1458


more_autograding_examples/cpp_count_ts/submissions/buggy.cpp
```diff
@@ -0,0 +1,14 @@
...
+  for (auto i : nums)
+    result += i;
+  return result;
+}
+
+int main() {
+  std::vector<int> nums = {2, 21, 52, 49, 38, 80};
+  std::cout << "Sum = " << sum(nums) << std::endl;
+  return 0;
+}
```
bmcutler, line 14: @poorna2152 

Every file in our repositories should have a newline at the end of the last line -- unless there is a specific need for something different for a file.  Many developers use editors that automatically add these missing newlines back in for some/most file types (so if they edit any of these files in the future, this change will be committed and it will be confusing if they didn't change anything else of substance in the file.)  

Github UI is labeling these missing characters in this PR and some of your other PRs too.

poorna2152: Thanks for pointing that out. Fixed the issue in this PR. Will fix the issues in other PRs as well

tests/integrationTests/tests/cpp_count_ts/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        os.mkdir(os.path.join(test.testcase_path, "assignment_config"))
+    except OSError:
+        pass
+    try:
+        data_path = os.path.join(test.testcase_path, "data")
+        if os.path.isdir(data_path):
+            shutil.rmtree(data_path)
+        os.mkdir(data_path)
+    except OSError:
+        pass
```
williamjallen, line 28: What is the intended purpose of these try-except statement?

tests/integrationTests/tests/cpp_count_ts/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+                     os.path.join(test.testcase_path, "assignment_config")])
+
+
+############################################################################
+
+def cleanup(test):
+    subprocess.call(["rm"] + ["-rf"] +
+            glob.glob(os.path.join(test.testcase_path, "data", "test*")))
+    subprocess.call(["rm"] + ["-f"] +
+            glob.glob(os.path.join(test.testcase_path, "data", "*")))
```
williamjallen, line 40: I believe that `shutil.rmtree()` would be a cleaner solution here

tests/integrationTests/tests/cpp_count_ts/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    try:
+        data_path = os.path.join(test.testcase_path, "data")
+        if os.path.isdir(data_path):
+            shutil.rmtree(data_path)
+        os.mkdir(data_path)
+    except OSError:
+        pass
+    subprocess.call(["cp",
+                     os.path.join(SAMPLE_ASSIGNMENT_CONFIG, "config.json"),
+                     os.path.join(test.testcase_path, "assignment_config")])
```
williamjallen, line 31: It would be better to use `shutil.copyfile()` here.

tests/integrationTests/tests/cpp_count_ts/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    except Exception:
+        traceback.print_exc()
+        raise
+
+@testcase
+def correct(test):
+    cleanup(test)
+    subprocess.call(["cp",
+                     os.path.join(SAMPLE_SUBMISSIONS, "solution.cpp"),
+                     os.path.join(test.testcase_path, "data")])
```
williamjallen, line 58: Same comment here about using `shutil.copyfile()`  (apply throughout this file)

williamjallen (commented):
I noted a couple code style things below, but everything else looks good

tests/integrationTests/tests/cpp_count_ts/__init__.py
poorna2152: One clarification. I was basically copying the implementation of the existing integration tests and changing the `__init__.py ` to match this test case. I noticed none of the tests were using `shutil.copyfile()`. I am not sure if there is a specific reason for that. Should I go ahead and make this change?

williamjallen: My understanding is that `subprocess.call` was used as something of a hack to mimic the CLI when prototyping the tests and then never removed.  I think it's probably best to not mimic the existing tests in this case.

tests/integrationTests/tests/cpp_count_ts/__init__.py
poorna2152: These files are not cleaned up when a test is finished since they are not specific to test runs.

poorna2152: Yes that makes sense. Updated the code

williamjallen (approved):

PR #8284, opened 2022-08-12
[Feature:Submission] Session management page

### LINKED ISSUE #8258: Manage sessions page for all users
A user can't retrieve information about their active sessions.

A "Manage Sessions" page will be created listing all the active sessions for a user, if the user wishes to terminate any of the sessions, they will be able to do so by simply clicking a button in front of that session.

**Additional context**
Once this is done, work on #8257 can be started.

akshatbatra:

There is no way for users to have a look at their active sessions and terminate any dangling sessions.

### What is the new behavior?
"Manage Sessions" page has been added inside profile page. "Single session" feature has also been added that allows a user to keep only one active session at a time.

Screen Shot 2023-06-23 at 2 56 44 PM

Screen Shot 2023-06-23 at 2 57 14 PM

  #### Location:
Navigate to "My Profile", "Manage Sessions" button is situated at the bottom of the page.

### Other information?
Closes #8258

In order to enable browser and platform detection, `submitty_install` needs to be run which will setup the cache needed by `browscap-php`.


migration/migrator/migrations/master/20220811134614_user_session_management.py
```diff
@@ -0,0 +1,39 @@
...
+        ALTER TABLE sessions
+        DROP COLUMN IF EXISTS session_created,
+        DROP COLUMN IF EXISTS browser_name,
+        DROP COLUMN IF EXISTS browser_version,
+        DROP COLUMN IF EXISTS platform;
+    """)
+    database.execute("""
+        ALTER TABLE users
+        DROP COLUMN IF EXISTS enforce_secure_session;
+    """)
```
williamjallen, line 39: Why are we deleting data here?  We usually try to keep data in case we have to roll back migrations for unrelated reasons.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2989,17 +2989,60 @@ public function getSession($session_id) {
...
 
     /**
-     * @todo: write phpdoc
+     * Fetch the session creation timestamp and other information for all active sessions of the given user
+     * @param string $user_id
+     *
+     * @return array
+     */
+    public function getSessionsInfoByUser($user_id) {
+        $this->submitty_db->query("SELECT session_id, csrf_token, session_created, browser_name, browser_version, platform FROM sessions WHERE user_id=? ORDER BY session_created", [$user_id]);
```
williamjallen, line 2998: It would be more readable to break this into several lines (see other functions in this file)

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2989,17 +2989,60 @@ public function getSession($session_id) {
     }
 
     /**
-     * @todo: write phpdoc
+     * Fetch the session creation timestamp and other information for all active sessions of the given user
+     * @param string $user_id
+     *
+     * @return array
+     */
+    public function getSessionsInfoByUser($user_id) {
```
williamjallen, line 2997: It would be good to add type hints here

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2989,17 +2989,60 @@ public function getSession($session_id) {
...
+        $this->submitty_db->query("SELECT enforce_secure_session FROM users WHERE user_id=?", [$user_id]);
+        return $this->submitty_db->rows()[0]['enforce_secure_session'];
+    }
+
+    /**
+     * Terminate all the sessions of current user except the one corresponding to the provided session_id
+     * @param string $session_id
+     * @param string $user_id
+     */
+    public function removeUserSessionsExcept($session_id, $user_id = null) {
```
williamjallen, line 3028: It would be good to add type hints here too

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2989,17 +2989,60 @@ public function getSession($session_id) {
...
+    /**
+     * Create a new login session for a user along with the session expiration timestamp which can be extended later on
      *
      * @param string $session_id
      * @param string $user_id
      * @param string $csrf_token
+     * @param array $user_agent
      */
-    public function newSession($session_id, $user_id, $csrf_token) {
+    public function newSession($session_id, $user_id, $csrf_token, $user_agent) {
```
williamjallen, line 3041: Same comment here about type hints

site/app/models/User.php
```diff
@@ -108,6 +108,8 @@ class User extends AbstractModel {
     protected $group;
     /** @prop @var int The access level of the user (ex: superuser, faculty, user) */
     protected $access_level;
+    /** @prop @var bool Should the user only have one active session at a time? */
+    protected $enforce_secure_session;
```
williamjallen, line 112: Is this the best name for this functionality?  When reading through the rest of this PR, I wasn't sure what this variable did.  Perhaps something like `$enforce_single_session` would be more readable.

site/app/templates/ManageSessions.twig
```diff
@@ -0,0 +1,43 @@
...
+                    <td>{{ session['platform'] }}</td>
+                    <td>
+                    {% if current_session_id == session['session_id'] %}
+                        <form action="authentication/logout">
+                            <input type="submit" class="btn btn-primary" value="Logout">
+                        </form>
+                    {% else %}
+                        <form method='POST' class="other-session" onsubmit="return terminateSession(this, '{{ csrf_token }}')">
+                            <input type="hidden" name="session_id" value="{{ session['session_id'] }}">
+                            <input type="submit" class="btn btn-primary" value="Terminate">
```
williamjallen, line 35: I think the interface is currently a little confusing.  What is the difference between "Logout" and "Terminate"?  They ultimately do the same thing: log the user out of a session.  Perhaps the text for all of the buttons should read "Logout" and some text can be added somewhere else to indicate that a session is the current one.

site/public/js/manage-sessions.js
```diff
@@ -0,0 +1,139 @@
+function removeRowTrigger(elem) {
+    const tr_elem = elem.parent().parent();
+    const table_elem = tr_elem.parent();
+    tr_elem.remove();
+    table_elem.trigger('change');
+}
+
+// eslint-disable-next-line no-unused-vars
```
williamjallen, line 8: It would be better to add an ignore line to the top of the file.  See [here](https://submitty.org/developer/software_and_system_design/coding_style_guide/javascript#style-rules) for more information.  The same comment applies elsewhere throughout this file.

williamjallen (changes requested):
In addition to the specific comments I've noted below, this is a new page which is mostly independent of the rest of Submitty.  It would be better to use Doctrine for managing the database and TypeScript instead of the current JavaScript files.  Given that the PR is mostly complete already, switching to these is not required, but it's something to consider.

It would also be good to add some Cypress tests for the new interface.

MasterOdin:
As opposed to requiring sysadmin action, could we use https://github.com/browscap/browscap-php instead, and add the step of grabbing / compiling the browscap.ini file or whatever into our install script?

akshatbatra:
@williamjallen 
Thanks for reviewing and providing valuable suggestions. I've done some changes locally. I'll also modify the code to use Doctrine and Typescript instead of JS and will also add Cypress tests and will then push.

site/app/controllers/ManageSessionsController.php
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public function showSessionsPage() {
+        return new WebResponse('ManageSessions', 'showSessionsPage');
+    }
+
+    /**
+     * Terminate a session
+     *
+     * @Route("/manage_sessions/terminate", methods={"POST"})
+     * @return JsonResponse
```
MasterOdin, line 37: For these various methods, what's the benefit to having this be a JsonResponse vs just having it be a regular redirect response back to the page?

Given that Submitty isn't a SPA, we shouldn't have AJAX around stuff that doesn't really need it as it requires additional JS code and such.

akshatbatra:
@MasterOdin
Using browscap-php seems like a good option, I haven't tested it but it should be able to use the local `browscap.ini`.
OR
I think another option can be to use [ini_set](https://www.php.net/manual/en/function.ini-set.php) to set the browscap value before `get_browser` call (grabbing the `browscap.ini` in install script will still be needed). Which option would be preferable?

site/app/controllers/ManageSessionsController.php
akshatbatra: There's no strong particular reason, I thought that AJAX enhances the user experience so that's why kept it this way....I'll change it to be a normal response.

MasterOdin: Given that I doubt any end user would particularly care or complain about using AJAX or regular HTTP post, I would opt for the path that uses the least amount of code, which in this case is without AJAX, allowing to remove the new JS file altogether.

akshatbatra: My motivation to prefer AJAX over regular POST comes from writings like this [ComputerWeekly article](https://www.computerweekly.com/news/2240081622/Use-Ajax-skills-to-enhance-the-online-user-experience). But given that Submitty isn't a leisure product but a necessity, it feels right to keep things simple.

MasterOdin:
I think the browscap-php solution is best, especially given this [note](https://browscap.org/#Using-Browscap) on their site:
> We only officially support using the [browscap-php](https://github.com/browscap/browscap-php) library. Other methods of loading the files are not supported, but we'll do our best to help when time permits.

This isn't code that should run that often, and so isn't quite performance critical so having this run in userland vs as a c-extension shouldn't matter.

akshatbatra:
This is how the page looks now:
session_management_page
`Session` is a Doctrine entity now and Cypress tests have been added.
`browscap-php` is cache oriented, there is apparently no straightforward API to work with a plain `browscap.ini`, you have to convert it to cache. Due to PHP version constraint, an older version (v4.3.0) of `browscap-php` had to be installed (see [browscap-php on packagist.org](https://packagist.org/packages/browscap/browscap-php)). 
A dependency conflict ([screenshot](https://user-images.githubusercontent.com/51257141/185998652-0658c404-3db4-44ad-9e2a-5ce2031bf3f3.jpg)) was encountered, consequently `doctrine/cache` had to be downgraded (v2.2.0 to v1.13.0). Please see if that's the right thing to do.

MasterOdin:
I merged #8142 which should fix the PHP version issue you were running into when trying to install the latest version of `browscap-php`. Could you please try again to do that?

akshatbatra:
The issue seems to persist:

php-v7 4

MasterOdin:
Why are you using 5.1 vs 7.1 for the browscap library?

MasterOdin:
If you set:

https://github.com/Submitty/Submitty/blob/4709eba6b6526d7a7c5c2080cc1bb720bdd6e376/site/composer.json#L5

to be `7.4.3`, does that allow installing the latest version (7.1) of the dependency?

akshatbatra:
Changing it to `7.4.3` abates the problem because now it only wants to update `symfony/console` (v4.4.42 => v5.4.11).

browscap-composer

See [output](https://github.com/Submitty/Submitty/files/9400774/php-v7.4.3.txt) of running `composer require browscap/browscap-php -W`.

MasterOdin:
Alright, so the best we can do is to use `browscap/browscap-php@^5` in that it supports all our top-level dependencies, but will require downgrading the transitive `doctrine/cache` dependency. Given it's a transitive dependency, having it be downgraded should be fine, and it doesn't look like there's any open CVE reports for the given version. The other conflicts from your screenshot should have been resolved by updating the PHP version in the composer.json from my PR, and we shouldn't then need to explicitly set 7.4.3 as well.

Thanks for going through this process of trying different versions, I really appreciate it!

site/app/controllers/ManageSessionsController.php
```diff
@@ -0,0 +1,111 @@
...
+            if ($this->core->getCurrentSessionId() === $session_id) {
+                return new RedirectResponse($this->core->buildUrl(["authentication", "logout"]));
+            }
+            $this->core->getSubmittyEntityManager()->remove($session);
+            $this->core->getSubmittyEntityManager()->flush();
+            $this->core->addSuccessMessage("Session terminated successfully.");
+        }
+        else {
+            $this->core->addErrorMessage("You don't have permission to terminate that session.");
+        }
```
cjreed121, line 72: Can we make the error the same whether the session doesn't exist or it doesn't belong to user? This could let a user start guessing other session IDs. There isn't anything harmful that could be done with a session ID without the JWT secret but I think it's safer to just give the same error message or edit the logic so there is only one return with an error.

site/app/entities/Session.php
```diff
@@ -0,0 +1,112 @@
...
+    public function getPlatform(): string {
+        return $this->platform;
+    }
+
+    public function isCurrent(string $current_session_id): bool {
+        return $this->session_id === $current_session_id;
+    }
+
+    public function updateSessionExpiration(\DateTime $current_dt) {
+        $this->session_expires = $current_dt->add(\DateInterval::createFromDateString('336 hours'));
```
cjreed121, line 110: 336 hours is 2 weeks but it might not be immediately clear upon reading the code. Could you either put 2 weeks in the string or put a comment saying it's 2 weeks to be a bit more readable.

cjreed121 (changes requested):
Looks great. Thanks for using Doctrine. I have a couple code review comments. I haven't done a test on my VM yet since I happen to be reinstalling right now but I'll do a full test soon.

site/app/entities/Session.php
MasterOdin: Looking at https://www.php.net/manual/en/datetime.formats.relative.php, I wonder if we could also replace `336 hours` with `2 weeks` or `14 days`.

site/app/models/User.php
akshatbatra: As suggested, it has been changed to `$enforce_single_session`.

site/app/templates/ManageSessions.twig
akshatbatra: It says 'Logout' instead of 'Terminate' now.

site/public/js/manage-sessions.js
akshatbatra: Regular POST requests are sent now instead of AJAX so this file has been removed.

site/app/controllers/ManageSessionsController.php
akshatbatra: It has been changed now to say 'Session doesn't exist' for both cases.

site/app/entities/Session.php
akshatbatra: It has been tweaked to be `2 weeks`.

cjreed121:
> Alright, so the best we can do is to use `browscap/browscap-php@^5` in that it supports all our top-level dependencies, but will require downgrading the transitive `doctrine/cache` dependency. Given it's a transitive dependency, having it be downgraded should be fine, and it doesn't look like there's any open CVE reports for the given version. The other conflicts from your screenshot should have been resolved by updating the PHP version in the composer.json from my PR, and we shouldn't then need to explicitly set 7.4.3 as well.
> 
> The alternative is spending time to upgrade symfony + twig, which while I think is well worth it, is probably outside the immediate scope, and better to get this done and merged, and then can separately focus on those.
> 
> Thanks for going through this process of trying different versions, I really appreciate it!

Should we make an issue for getting those upgraded?

cjreed121:
A couple design thoughts:

- Should this page have its own sidebar link? It is definitely something that could be linked to from the profile page I feel.
- Should the action button to logout also be a danger button rather than a primary button? A feel it being a primary button makes me want to click it.

cjreed121:
In our meeting it was determined that we should remove this new page and authentication tokens page from the sidebar. We can just add two links on the profile page to point to these pages since they really won't be used that often. Let's also change the logout buttons to either danger or secondary buttons.

Finally, I'm going to make an issue for possibly adding some more functionality to this table. The new functionality is to show other SAML proxy users here too if you have access to them. This is nothing that would have to happen in this PR but it's just an idea for the future brought up in the meeting.

williamjallen:
I am in favor of replacing the buttons with danger buttons.

akshatbatra:
Links to both the pages have been moved from sidebar to profile page. 'Logout' button has been tweaked to be a danger button.

cjreed121 (commented):
I think this is just about ready to merge. Can you update this submitty docs page https://submitty.org/student/account/authentication_tokens to say that a user has to go to my profile first? Also I think we usually prefer to pin dependencies to exact version rather than ^ version in the composer.json. Could you change that as well?

cjreed121 (approved):

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -624,6 +624,12 @@ popd > /dev/null
 # COPY THE 1.0 Grading Website if not in worker mode
 if [ "${WORKER}" == 0 ]; then
     bash "${SUBMITTY_REPOSITORY}/.setup/install_submitty/install_site.sh"
+
+    if [ "${CI}" != true ]; then
+        echo -e "Checking for and fetching latest browscap.ini"
+        # fetch and convert browscap.ini to cache, may take some time on initial setup or if there's an update
+        ${SUBMITTY_INSTALL_DIR}/sbin/update_browscap.php "${PHP_USER}"
+    fi
```
MasterOdin, line 632: Let's move this block into `.setup/install_submitty/install_site.sh` to keep all the "site installation" code centralized.

MasterOdin (changes requested):
Some small tweaks suggested, and then good to merge.

Good job on this :+1:

sbin/update_browscap.php
```diff
@@ -0,0 +1,28 @@
...
+$filesystem = new \League\Flysystem\Filesystem($file_cache);
+$cache = new \MatthiasMullie\Scrapbook\Psr16\SimpleCache(
+    new \MatthiasMullie\Scrapbook\Adapters\Flysystem($filesystem)
+);
+$logger = new \Psr\Log\NullLogger();
+$bc = new \BrowscapPHP\BrowscapUpdater($cache, $logger);
+// create an active/warm cache
+$bc->update(\BrowscapPHP\Helper\IniLoaderInterface::PHP_INI);
+
+`chown -R ${cache_user}:${cache_user} ${cache_dir};`;
```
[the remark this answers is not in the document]
akshatbatra: After spending some time on it, I've identified the cause of this unexpected behavior: `shell_exec` is disabled but only in `/etc/php/7.4/fpm/php.ini` not in `/etc/php/7.4/cli/php.ini`, no function is.

.setup/INSTALL_SUBMITTY_HELPER.sh
akshatbatra: This has been moved to `.setup/install_submitty/install_site.sh`, but now every time `.setup/bin/code_watcher.py` detects a change, it checks for latest `browscap.ini`. Please see if that's right.
A possible solution can be to send an argument (or exporting a variable) to/for `.setup/install_submitty/install_site.sh` when invoked via `.setup/INSTALL_SUBMITTY_HELPER_SITE.sh` to disable the check for `browscap.ini` update.

akshatbatra: Now, `browscap` update is only checked if it's sent as an argument to `.setup/install_submitty/install_site.sh`. This is only done in `INSTALL_SUBMITTY_HELPER.sh`; this way, update check is avoided if `install_site.sh` runs separately at any other place.

cjreed121 (changes requested):
#8620 will likely get merged before this and will probably cause some merge conflicts here.

site/app/controllers/ManageSessionsController.php
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @Route("/manage_sessions", methods={"GET"})
+     * @return WebResponse
+     */
+    public function showSessionsPage(): WebResponse {
+        /** @var SessionRepository $repo */
+        $repo = $this->core->getSubmittyEntityManager()->getRepository(Session::class);
+        $user_sessions = $repo->getAllByUser($this->core->getUser()->getId());
+        return new WebResponse(
+            'ManageSessions',
```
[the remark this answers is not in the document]
akshatbatra: `ManageSessions` has been replaced with `ManageSessionsView::class`.

site/app/controllers/ManageSessionsController.php
```diff
@@ -0,0 +1,97 @@
...
+     * Terminate all sessions except current
+     *
+     * @Route("/manage_sessions/logout_all", methods={"POST"})
+     * @return RedirectResponse
+     */
+    public function logoutAllExceptCurrent(): RedirectResponse {
+        /** @var SessionRepository $repo */
+        $repo = $this->core->getSubmittyEntityManager()->getRepository(Session::class);
+        $repo->removeUserSessionsExcept($this->core->getUser()->getId(), $this->core->getCurrentSessionId());
+        return new RedirectResponse($this->core->buildUrl(["manage_sessions"]));
```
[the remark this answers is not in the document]
akshatbatra: It has been added.

site/app/libraries/DateUtils.php
```diff
@@ -222,11 +222,11 @@ public static function getUTCOffset(string $time_zone): string {
      * to the given format.
      *
      * @param User $user
-     * @param string $time_stamp
+     * @param \DateTime|string $time_stamp
      * @param string $format
      * @return string
      */
-    public static function convertTimeStamp(User $user, string $time_stamp, string $format): string {
```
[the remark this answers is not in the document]
akshatbatra: This change was needed at some point during development; it got left out when the related function call was reverted.

site/app/libraries/SessionManager.php
```diff
@@ -40,17 +46,27 @@ public function __construct(Core $core) {
...
+        $em = $this->core->getSubmittyEntityManager();
+        /** @var SessionRepository $repo */
+        $repo =  $em->getRepository(Session::class);
+        $expressionBuilder = Criteria::expr();
+        $this->session = $repo->matching(new Criteria(
+            $expressionBuilder->andX(
+                $expressionBuilder->eq("session_id", $session_id),
+                $expressionBuilder->gt("session_expires", $this->core->getDateTimeNow())
+            )
+        ))->first();
```
cjreed121, line 58: Could we instead put this logic inside the repository? Then we can just have a function like `getActiveSessionById` or something.

cjreed121 (changes requested):
Thanks for addressing the merge conflicts

migration/migrator/migrations/master/20220811134614_user_session_management.py
```diff
@@ -0,0 +1,29 @@
...
+        ALTER TABLE sessions
+        ADD COLUMN IF NOT EXISTS session_created timestamp(0) with time zone DEFAULT NULL,
+        ADD COLUMN IF NOT EXISTS browser_name character varying(50) DEFAULT 'Unknown',
+        ADD COLUMN IF NOT EXISTS browser_version character varying(15) DEFAULT '',
+        ADD COLUMN IF NOT EXISTS platform character varying(50) DEFAULT 'Unknown';
+    """)
+    database.execute("""
+        ALTER TABLE users
+        ADD COLUMN IF NOT EXISTS enforce_single_session boolean DEFAULT false;
+    """)
```
williamjallen, line 23: These SQL commands should be run as part of a single `database.execute()` call.  That way if one of them fails everything will be rolled back as part of a single transaction.

williamjallen (changes requested):

migration/migrator/migrations/master/20220811134614_user_session_management.py
ziesski: Enclosed it using BEGIN and COMMIT. Thanks for the suggestion

migration/migrator/migrations/master/20220811134614_user_session_management.py
```diff
@@ -0,0 +1,35 @@
...
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+
+    database.execute("""
+        BEGIN;
```
williamjallen, line 14: You don't need to include `BEGIN` and `COMMIT` in the migration.  The migrator does that for you.  You just need to make sure all of the commands are in a single `database.execute()` call.

williamjallen (changes requested):

04vmatsibekker (commented):
When testing this out and enabling single session, I would have thought that the browser I was on would be the only one I could be logged in on until I logged out of that browser, but what I found was that sure only 1 session could be run at a time, but that when I tried to log onto other browsers the previous browser would get logged out. 

Also when I had two browsers open and I exited out of one browser I was sort of expecting to have my number of sessions to decrease by one, but since I was technically still logged in - meaning if I went back to the site I wouldn't have to log in again I was automatically instructor - I still had two sessions open. Perhaps it could be nice in the future for when someone just exits out of a page a little pop-up shows up asking if they want to log out. 

Also in your description of your PR it seemed like there would be a manage sessions button on the sidebar of the page, it turns out I have to go to my profile page, maybe it would be good if that was made a little more clear on the PR.

akshatbatra:
1.) The purpose of "Enforce Single Session" is to get rid of unnecessary dangling sessions that arise due to a user logging in from different browsers/computers. If a session remains active on a system that is not the legit user's primary system then it increases the surface for a compromise, this feature's sole purpose is to mitigate that. As far as linking session to a particular browser is concerned, a similar idea popped up back during the design phase but it was discarded due to being too intrusive as identifying a specific browser requires to store more information about user's system than a user might want to share, privacy implications have to be taken into account. Moreover, linking to a particular system/browser may lead to inconvenience in case the user needs to log in from a non-primary system in case the primary system is not available (whatever the reason may be).

2.) I suppose that dormant sessions are automatically pruned after a certain time period. Submitty is very frequently used by some users - multiple times a day. Closing a Submitty browser window has to be a quick process. However, a configurable (option to choose Ask every time/Never ask) popup window can be implemented, this can be done in another PR.

3.) The sidebar link was the initial proposition, but the location of the button was changed in [this commit](https://github.com/Submitty/Submitty/pull/8284/commits/384d980ed5d946ddee689eb8836215badfb9bc4e), see [this comment](https://github.com/Submitty/Submitty/pull/8284#issuecomment-1238443427) for context. The description has been updated to reflect the current location.

Thanks for the really great analysis.

migration/migrator/migrations/master/20220811134614_user_session_management.py
ziesski: made the changes :)

ziesski:
> 

Thanks for your follow up, we discussed it during meeting today, and ended up in similar conclusion.
For part 3, indeed, your notes on initial post "Location" clearly shows that it was inside profile page, but in the middle of the conversation there is one prototype image posted with separate left sidebar column, which in turn led some of our developers confused. I attached screenshot of the current version in your original post for clarity. I hope you find this edit appropriate :)

migration/migrator/migrations/master/20220811134614_user_session_management.py
ziesski: Thanks for your insight :)

ziesski:
> 1.) The purpose of "Enforce Single Session" is to get rid of unnecessary dangling sessions that arise due to a user logging in from different browsers/computers. If a session remains active on a system that is not the legit user's primary system then it increases the surface for a compromise, this feature's sole purpose is to mitigate that. As far as linking session to a particular browser is concerned, a similar idea popped up back during the design phase but it was discarded due to being too intrusive as identifying a specific browser requires to store more information about user's system than a user might want to share, privacy implications have to be taken into account. Moreover, linking to a particular system/browser may lead to inconvenience in case the user needs to log in from a non-primary system in case the primary system is not available (whatever the reason may be).
> 
> 2.) I suppose that dormant sessions are automatically pruned after a certain time period. Submitty is very frequently used by some users - multiple times a day. Closing a Submitty browser window has to be a quick process. However, a configurable (option to choose Ask every time/Never ask) popup window can be implemented, this can be done in another PR.
> 
> 3.) The sidebar link was the initial proposition, but the location of the button was changed in [this commit](https://github.com/Submitty/Submitty/pull/8284/commits/384d980ed5d946ddee689eb8836215badfb9bc4e), see [this comment](https://github.com/Submitty/Submitty/pull/8284#issuecomment-1238443427) for context. The description has been updated to reflect the current location.
> 
> Thanks for the really great analysis.

Thanks for your follow up, we discussed it during meeting today, and ended up in similar conclusion.
For part 3, indeed, your notes on initial post "Location" clearly shows that it was inside profile page, but in the middle of the conversation there is one prototype image posted with separate left sidebar column, which in turn led some of our developers confused. I attached screenshot of the current version in your original post for clarity. I hope you find this edit appropriate :)

musaabimran (approved):

musaabimran (approved):
Good to merge!

04vmatsibekker:
Once tests pass I'll approve.

cjreed121 (approved):
Works great

PR #8298, opened 2022-08-16
[Testing:Submission] Expected String Test Cases

Eceptonsu:
There are no test cases for the expected string functionality created in PR #8254 

### What is the new behavior?
Add a set of test cases for the expected string functionality


DefinitiveAbove (commented):
Looks good when tested with #8254

PR #8306, opened 2022-08-19
[DevDependency] Upgrade cypress to 10.6

shailpatels:

Cypress introduced some breaking changes between 9.x and 10.x

### What is the new behavior?
The migration to cypress has been run by upgrading to 10.6 and running `npx cypress open`, the file structure has been changed so that tests are now under `e2e` instead of `integration`, there is now a `cypress.config.ts` to configure it. The github actions for Submitty has also been updated to work with the newer version.

### Other information

Originally the PR made by dependabot got stale and conflicted with main so  I recreated it manually since its limited to 10prs right now


williamjallen (approved):
Changes look good.  Would moving the cypress tests to the `tests` directory be a good idea while you're making changes?  As it is now, cypress has its own independent directory which can be confusing for new developers.

shailpatels:
> Changes look good. Would moving the cypress tests to the `tests` directory be a good idea while you're making changes? As it is now, cypress has its own independent directory which can be confusing for new developers.

Yeah as far as I can tell there shouldn't be an issue with that, the cypress config would have to live in the site directory next to the `package-lock` files but the `cypress` directory with tests should be able to be moved wherever.

Maybe you can bring up the topic in the daily meetings to see if the team would want it in the `tests` directory instead

PR #8316, opened 2022-08-22
[Bugfix:TAGrading] Fix missing gradeable_id in display file

Viyerelu23333:

In the file viewing page, the missing `gradeable_id` in the `iframe` source causes a permission error.

### What is the new behavior?
`gradeable_id` is added for TA grading page.

### DIFF
--- a/site/public/js/ta-grading.js
+++ b/site/public/js/ta-grading.js
@@ -506,7 +506,7 @@ function adjustGradingPanelHeader () {
       height += $(this).outerHeight(true);
     }
   });
-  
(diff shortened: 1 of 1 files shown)


akshatbatra:
Thanks for fixing this.
For the record: This issue was introduced by #8002.

[server.js Line 810](https://github.com/Submitty/Submitty/blob/main/site/public/js/server.js#L810) also points to the same route so `gradeable_id` may be needed there too.

Please let me know if I can be of any help.

PR #8333, opened 2022-08-24
[Refactor:Autograding] Cleanup `make` output log

Viyerelu23333:

`log_make_output.txt` is never generated and used; so when building the gradeable, make output is always `null`.

### What is the new behavior?

`log_make_output.txt` is removed.  The config interface won't display the make output.


tkoz0 (approved):
Tested and looks good.

PR #8339, opened 2022-08-27
[Bugfix:Submission] Adding and deleting users

### LINKED ISSUE #8338: Anon Ids missing for recently added students
**Describe the bug**
If we add a student to the course after one or more gradeables have been created, the student will not have an anon id for that gradeable.  This causes weird bugs...

* If the newly added student tries to access the gradeable [ notebook gradeable does not render, irrelevant errors about teams, etc. ]

* When a grader goes to view the grading for that assignment, the newly added student is not listed in the list of all students in the course.

The anon_id table is missing any anon id for the newly added students for these existing gradeables.

akshatbatra:

If a student is added to a course after one or more gradeables have been created, the student will not have an anon id for that gradeable, this causes several bugs.

### What is the new behavior?
Whenever a user is created, `gradeable_anon` is populated with anon ids for existing gradeables.

Another issue related to deleting a user was identified:
delete_user

It is impossible to delete a user even though they didn't have any record of interactions; to fix this, a migration has been created, the function isn't new, a minor change in an already existing function has been done.

A migration has been added that adds the missing anon ids, the same migration also adds a trigger to automatically insert required rows in `gradeable_anon` table whenever insertion happens in `users` (course) table.

Running the migration is all that is needed, but for insurance, `anonymize.py` has been added, this is not to be confused with the script that existed before PR 8002 as most of it has been modified.

### Other information?

Fixes #8338 
This is another one in the series of bugs introduced by #8002, hopefully this string ends with this PR.

### DIFF
--- a/sbin/adduser_course.py
+++ b/sbin/adduser_course.py
@@ -7,6 +7,8 @@ will be an instructor.
 
 import argparse
 import json
+import random
+import string
 from os import path
 import sys
 from sqlalchemy import create_engine, MetaData, Table, bindparam, and_
@@ -20,6 +22,8 @@ DATABASE_HOST = DATABASE_DETAILS['database_host']
 DATABASE_PORT = DATABASE_DETAILS['database_port']
 DATABASE_USER = DATABASE_DETAILS['database_user']
 DATABASE_PASS = DATABASE_DETAILS['database_password']
+DATABASE_COURSE_USER = DATABASE_DETAILS['database_course_user']
+DATABASE_COURSE_PASS = DATABASE_DETAILS['database_course_password']
 
 
 def parse_args():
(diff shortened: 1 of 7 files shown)


sbin/adduser_course.py
```diff
@@ -107,6 +111,46 @@ def main():
             registration_section=registration_section
         )
 
+    # function taken from setup_sample_courses
```
cjreed121, line 114: Won't the database do all of this for us now?

cjreed121 (changes requested):

sbin/adduser_course.py
akshatbatra: Thanks for pointing that, missed to revert these changes after adding the trigger.

PR #8378, opened 2022-09-09
[Bugfix:System] Fix LDAP install on vagrant

cjreed121:
When the vagrant system is installed, LDAP doesn't get fully set up because it is missing a couple commands. So when the setup_sample_courses script tries to add the users it fails on them and there are then 0 ldap users. There was also a bash syntax error from #6815.

### What is the new behavior?
The testing LDAP and vagrant LDAP now use the same script and it works locally. This should also help us detect problems like this earlier for LDAP. The bash syntax issue has been fixed.

### Other information?
I tested locally with LDAP and SAML and they both worked.

### DIFF
--- a/.setup/distro_setup/setup_distro.sh
+++ b/.setup/distro_setup/setup_distro.sh
@@ -41,7 +41,7 @@ fi
 
 # SEE GITHUB ISSUE #7885 - https://github.com/Submitty/Submitty/issues/7885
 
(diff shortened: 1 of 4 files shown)


MasterOdin (approved):

cjreed121:
This should be good now. I also did another vagrant up and it worked as expected now.

PR #6624, opened 2021-06-17
[Refactor:Forum] Add proper bootstrap dropdown support

eddiekrystowski:
### What is the current behavior?
Fixes #6618
Currently, the bootstrap dropdown-menu class is partially implemented in `server.css`, and has been modified greatly so it is difficult to use in other areas (can't just simply use dropdown-menu class like bootstrap docs suggests).

### What is the new behavior?
Dropdown related classes have been added to `bootstrap.css` and Submitty specific overrides have been specified in `server.css`. The only place previously using this class (More dropdown on discussion forum) has been adjusted to use the proper layout. 

Now, following [this simple format](https://getbootstrap.com/docs/4.0/components/dropdowns/#single-button-dropdowns), dropdown menus can easily be added.

Also, there is now support for split buttons and [split button dropdowns](https://getbootstrap.com/docs/4.0/components/dropdowns/#single-button-dropdowns).

PR #3475, opened 2019-03-24
Forum bar update

andrewaikens87:
This contains the re-design of the discussion forum bar from the discussion during the hackathon. I have provided some images below of the new design for each page.

Create thread:
createThreadView

Instructor view:
instructorView

Student view:
studentView

Search view:
searchView

Stats view:
statsView

PR #8380, opened 2022-09-10
[Bugfix:Forum] Remove dead code causing JS syntax issue

williamjallen:
When viewing a forum with merged threads in anything but the wide view, the Twig variable `is_full_threads_page` is left undefined, which results in a JS syntax error due to the ternary expression on line 18.

### What is the new behavior?
The `dropdownLabel` ID does not appear anywhere in the codebase except for this location, so I have removed all of the related logic, including the buggy line.

### DIFF
--- a/site/app/templates/forum/ForumBar.twig
+++ b/site/app/templates/forum/ForumBar.twig
@@ -2,28 +2,13 @@
 
 {% if show_more is defined and show_more %}
 	<script>
-
 		$(document).ready(function() {
 		    let filterOption = '{{ more_data[0]["filter_option"] }}';
 		    if (filterOption) {
                 $('#' + filterOption).parent().removeClass("active");
                 $('#' + filterOption).addClass("active");
             }
-			let del = $('#delete');
-			let merge  = $('#merge_thread');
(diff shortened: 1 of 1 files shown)


MasterOdin:
@williamjallen can you provide any context on why this code existed in the first place? Why is it better to fully remove it, vs fixing what it's supposed to be doing?

williamjallen:
The faulty ternary expression originated in #6713.  The command `git log -S "dropdownLabel"` indicates that the string `dropdownLabel` occurs in the diff three times.  In particular, it was introduced in #3475 and eliminated in #6624 during a refactor of the dropdown menu boxed in the attached screenshot.  The sorting functionality appears to work properly so I am not aware of any issues which need to be fixed as opposed to just removing the dead code.

The undefined Twig variable which prompted this, `is_full_threads_page`, is only ever set to true or apparently left undefined.  The definition (or not) of `is_full_threads_page` is effectively used as a boolean in logic later in the same file, which prevents its removal altogether.

image

DefinitiveAbove (approved):

PR #8381, opened 2022-09-10
[UI/UX:Submission] Add Contrast to Submission Interface

### LINKED ISSUE #8250: Poor contrast on submission interface
**Describe the bug**
When submissions are not allowed, a dark gray banner with black text is displayed on the submission interface.  This banner has poor contrast and does not match the surrounding interface in light mode.

image

**Expected behavior**
The contrast should be improved by changing the color of the background for this banner.  Personally, I think the usual Submitty blue we use for other banners would be acceptable, but other colors might be viable as well.

Gage-Zahn:
### Overview

When an assignment is no longer open for submissions, the text color of the message is set to white to add contrast.

Closes #8250

182425179-3f2201ad-303e-4da1-9a14-09ae51b3ea24

### What is the new behavior?

After


bmcutler:
Hi @Gage-Zahn 
Thanks for your contribution

However, we don't want to use raw rgb in the twig files, instead we want to use named colors from:
./site/public/css/colors.css

Also be sure to test both light mode & dark mode, this is where many of our contrast bugs arise.

Gage-Zahn:
I made a change so that it uses the colors in colors.css 

In Light mode:
image

In Dark mode:
image

williamjallen:
These changes look somewhat better.  I recommend using WAVE (see [here](https://submitty.org/developer/software_and_system_design/web_accessibility#examine-your-interface-design-with-wave) for more info) to make sure the contrast in dark mode and light mode is ok.

Gage-Zahn:
### Before
**Dark Mode**
image

**Light Mode**
image

These both pass the minimum ratio of 4.5:1. Optimally Dark Mode would have better contrast.

### After
Switching --standard-light-gray for --subtle-grey-background results in this:

**Dark Mode**
image

**Light Mode**
image

shailpatels (approved):
Changes look good and pass the contrast requirements on WAVE for light and dark modes.

Thanks for the contribution!

Issue #8319, opened 2022-08-22
Registration Type -- continued


bmcutler:
**What problem are you trying to solve with Submitty**

Following on Issue #7474, and PRs #7808 and #8026.

- [x] Previous work added the registration_type column in the course database users table.
    But we need to ALSO add this column redundantly to the master submitty_db courses_users table.
    Like other fields that are duplicated from the master table to the course table (e.g., registration_section, manual 
    registration) edits to the registration type should only be made to the master database, and then auto copied to the 
    relevant courses database.

- [ ]  Update the new student / edit student web GUI so that the edit doesn't happen to the course db, but to the master DB.

- [ ]  Update the new student / edit student  and new grader / edit grader web GUI so the registration type is always visible (currently its only there for the student types).  A new grader should default to 'staff' and a new student should default to 'graded', but any value should be selectable for any user.

- [ ]  Update the list of types in the dropdown menu to add 'staff', and change 'withdrawn' to be 'late drop / withdrawn'.  

- [x] Update the autofeed to automatically set the registration type field for students who have 'audit' and 'late drop / withdrawn' status.

This will be a few PRs.  Tagging / assigning to multiple people who have worked on this / reviewed this previously.

bmcutler:
I think some (all?) of these items may be complete.

PR #8389, opened 2022-09-12
[Feature:System] Master DB Migration / Registration Type

pbailie:
Submitty Master Database Migration Requested In #8319, Part 1 
- New column in `courses_users` table: `registration_type VARCHAR(255) DEFAULT 'graded'`
- New constraint `check_registration_type` only allows text values `'graded'`, `'audit'`, `'withdrawn'`, or `'staff'`.
    - This mirrors the check constraint on a course DB's column `users.registration_type`.
- Update trigger function so `courses_users.registration_type` is copied to the relevant course database `users.registration_type`.
- Provide rollback migration.
- This migration is required for testing autofeed script updates related to #8319


cjreed121 (changes requested):
The master db needs to get dumped to provide an updated SQL file. Also I'm guessing we would want to change how the registration type is updated if an instructor does it to update the master db rather than just the course db?

pbailie:
> The master db needs to get dumped to provide an updated SQL file.

Updating `submitty_db.sql` is no problem.  I'll provide that soon.

> Also I'm guessing we would want to change how the registration type is updated if an instructor does it to update the master db rather than just the course db?

Now that I look at it again, somehow the UPDATE sync for `registration_type` is missing.  I wrote it, tested it, and somehow it didn't make it into the commit.  I'll get that fixed, too.

pbailie:
> Now that I look at it again, somehow the UPDATE sync for registration_type is missing. I wrote it, tested it, and somehow it didn't make it into the commit. I'll get that fixed, too.

Hold up...  The `registration_type` update is in the commit on line 59 between `rotating_section` and `manual_registration`. https://github.com/Submitty/Submitty/blob/89da05b044866cd555661dba6d3219564ab0ce57/migration/migrator/migrations/master/20220908163945_student_registration_type.py#L55-L62

cjreed121:
@tkoz0 is supposed to add some web backend changes here and then I'll review again

tkoz0:
@cjreed121 which changes were those? I don't recall planning to make additional changes to this PR

cjreed121 (changes requested):
You can also drop `registration_type` from line 1099 of DatabaseQueries.php since the trigger should do that for us now.

migration/migrator/migrations/master/20220908163945_student_registration_type.py
```diff
@@ -0,0 +1,120 @@
...
+    Run down migration (rollback).
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+
+    database.execute("ALTER TABLE IF EXISTS courses_users DROP CONSTRAINT IF EXISTS check_registration_type")
+    database.execute("ALTER TABLE IF EXISTS courses_users DROP COLUMN IF EXISTS registration_type")
```
cjreed121, line 83: Do we want to drop these? Usually we like to keep these in place to avoid losing data.

cjreed121 (changes requested):
PHP code looks good. Didn't test locally though yet. One more comment on migration.

migration/migrator/migrations/master/20220908163945_student_registration_type.py
pbailie: This is to prevent a collision.  These columns *should not* exist prior to this migration, but if they did -- the migration will throw an ugly looking exception when trying to create the columns.

pbailie: For clarity...  Upon closer examination -- the drop statements in question are not about collisions in the 'up' migration, and are not in the 'up' migration at all.  They are applied to the 'down' migration.

But for future reference, `IF NOT EXISTS` is not a universally available condition within an `ALTER TABLE` command.  e.g. Something like `ALTER TABLE IF EXISTS courses_users ADD CONSTRAINT IF NOT EXISTS check_registration_type ...` will fail as a Postgresql syntax error.

PR #8391, opened 2022-09-13
[UI/UX:CourseMaterials] Change course materials icon

williamjallen:
#8362 changed the style of the `fa-copy` icon as part of the addition of a new "sharp" family of icons.  The new copy icon looks like [this](https://fontawesome.com/icons/copy?s=solid&f=classic), which is a change from the "two documents" style which existed previously. 

### What is the new behavior?
The previous "two documents" look does not appear anywhere in the current set of FA free icons so I have replaced the course materials icon with a single "file" icon (see screenshot for the new look).  Other possibilities for the replacement icon are [`fa-folder`](https://fontawesome.com/icons/folder?f=classic&s=solid) and [`fa-folder-open`](https://fontawesome.com/icons/folder-open?f=classic&s=solid).

image

### DIFF
--- a/site/app/controllers/GlobalController.php
+++ b/site/app/controllers/GlobalController.php
@@ -172,7 +172,7 @@ class GlobalController extends AbstractController {
             $sidebar_buttons[] = new NavButton($this->core, [
                 "href" => $this->core->buildCourseUrl(['course_materials']),
                 "title" => "Course Materials",
-                "icon" => "fa-copy"
+                "icon" => "fa-file"
             ]);
         }
 


shailpatels (approved):
Thanks for catching this,

I think a possible replacement could also be the folder tree - that _feels_ like course materials to me https://fontawesome.com/icons/folder-tree?s=solid&f=classic, not sure how it'll look on smaller screens though

but the single file probably conveys the same meaning pretty well

williamjallen:
For reference, here's what the `folder-tree` icon would look like:
image

I agree that it conveys the idea well, but I think it's too small when placed next to the other icons on the sidebar.

PR #8394, opened 2022-09-14
[UI/UX:HelpQueue] Office hour staff text color

billyvande:
Text color is different in dark mode. 

### What is the new behavior?
Since the objective is to make text color same. Now, office hour text has same color in dark mode.

### DIFF
--- a/site/app/templates/officeHoursQueue/QueueHeader.twig
+++ b/site/app/templates/officeHoursQueue/QueueHeader.twig
@@ -45,7 +45,7 @@
         {% for queue in viewer.getAllQueues() %}
             {% if queue['open'] == 0 %}
 
-                <label tabindex="0" class="btn"  style="word-wrap: break-word; border-color: {{ viewer.getColorFromCode(queue['code']) }}; background: {{ viewer.getColorFromCode(queue['code']) }}; color: inherit;">{{queue['code']}}
+                <label tabindex="0" class="btn filter-buttons"  style="word-wrap: break-word; border-color: {{ viewer.getColorFromCode(queue['code']) }}; background: {{ viewer.getColorFromCode(queue['code']) }}; color: inherit;">{{queue['code']}}
                     <input
                             type="checkbox"
                             title="Toggle filter for: {{queue['code']|upper}}"
--- a/site/app/templates/officeHoursQueue/QueueHeader.twig
+++ b/site/app/templates/officeHoursQueue/QueueHeader.twig
@@ -27,7 +27,7 @@
         {% for queue in viewer.getAllQueues() %}
             {% if queue['open'] != 0 %}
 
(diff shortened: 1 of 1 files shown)


cjreed121 (changes requested):
Rather than adding hard coded CSS. We should be able to match up the classes on both. You'll see that one of the labels has classes `btn` and `filter-buttons` while the other just has `btn`. Let's keep it consistent so they both have `btn` and `filter-buttons`.

billyvande:
Thanks for corrections. This is my first time contributing to an open source. 
I update Office hours header and match the class.

site/app/templates/officeHoursQueue/QueueHeader.twig
```diff
@@ -27,7 +27,7 @@
         {% for queue in viewer.getAllQueues() %}
             {% if queue['open'] != 0 %}
 
-                <label tabindex="0" class="btn filter-buttons"  style="word-wrap: break-word; border-color: {{ viewer.getColorFromCode(queue['code']) }}; background: {{ viewer.getColorFromCode(queue['code']) }};">{{queue['code']}}
+                <label tabindex="0" class="btn filter-buttons"  style="word-wrap: break-word; border-color: {{ viewer.getColorFromCode(queue['code']) }}; background: {{ viewer.getColorFromCode(queue['code']) }}; color: inherit;">{{queue['code']}}
```
cjreed121, line 30: Can you remove `color: inherit` here

site/app/templates/officeHoursQueue/QueueHeader.twig
```diff
@@ -45,7 +45,7 @@
         {% for queue in viewer.getAllQueues() %}
             {% if queue['open'] == 0 %}
 
-                <label tabindex="0" class="btn"  style="word-wrap: break-word; border-color: {{ viewer.getColorFromCode(queue['code']) }}; background: {{ viewer.getColorFromCode(queue['code']) }}; color: inherit;">{{queue['code']}}
+                <label tabindex="0" class="btn filter-buttons"  style="word-wrap: break-word; border-color: {{ viewer.getColorFromCode(queue['code']) }}; background: {{ viewer.getColorFromCode(queue['code']) }}; color: inherit;">{{queue['code']}}
```
cjreed121, line 48: Also remove here too

cjreed121 (changes requested):
I used WAVE to check accessibility and there was some contrast errors with this. If you can quickly make the changes I've noted it should fix them and I think this is then good to merge.

site/app/templates/officeHoursQueue/QueueHeader.twig
billyvande: yes, sure!

billyvande:
Before : 
image
After : 
image

I wish my PR solve the issues.

cjreed121:
Please leave the title as is so it complies with https://submitty.org/developer/getting_started/how_to_contribute

PR #8407, opened 2022-09-17
[Bugfix:CourseMaterials] Remove null hidden from student

cjreed121:
Null values are allowed in the `hidden_from_students` column in course materials. A recent change has made it so the page expects that this column has a non-null value.

### What is the new behavior?
This PR adds a migration to set all folders which would have the null values to have the value false. It also adds a database constraint to ensure this value is never null again.

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -656,7 +656,7 @@ CREATE TABLE public.course_materials (
     path character varying(255),
     type smallint NOT NULL,
     release_date timestamp with time zone,
-    hidden_from_students boolean,
+    hidden_from_students boolean NOT NULL,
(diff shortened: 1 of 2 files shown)


akshatbatra:
This is issue #8262, I think this bug appeared as a result of #8150, back then, I assumed that folder rows in the database record meaningful information in all of their columns as is the case with files. The change which is the cause of this issue has been reverted in #8216. `hidden_from_students` will not be required to be `NOT NULL` once #8216 is merged.

cjreed121:
Alright. Didn't realize it was already being fixed in that PR. I just saw a report from someone else about this issue so I went to go see if I could make a quick fix. Thanks for pointing it out. I'll close this and see if I can approve that other PR.

PR #8414, opened 2022-09-18
[Documentation:Developer] Update UTM ARM instructions

jssunray34:
I updated the README to clarify some of the steps for M1 Mac users.

### DIFF
--- a/README_utm_arm.txt
+++ b/README_utm_arm.txt
@@ -19,7 +19,7 @@ On an M1 Mac laptop, we cannot use virtual box, so follow these instructions ins
 
    And also create directory:
    vendor/
-   in GIT_CHEXKOUT directory and checkout from:
+   in GIT_CHECKOUT directory and checkout from:
    git clone https://github.com/nlohmann/json.git
    to this path:
    vendor/nlohmann/json/
--- a/README_utm_arm.txt
+++ b/README_utm_arm.txt
@@ -39,12 +39,10 @@ On an M1 Mac laptop, we cannot use virtual box, so follow these instructions ins
    Current version: 3.5.1
 
 
-3. Download and save the Ubuntu 20.04 ARM Server ISO
-   https://mac.getutm.app/gallery/ubuntu-20-04 *
-   https://cdimage.ubuntu.com/releases/20.04.4/release/
+3. Go to: https://cdimage.ubuntu.com/releases/20.04.4/release/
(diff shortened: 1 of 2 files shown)


README_utm_arm.txt
```diff
@@ -40,7 +40,7 @@ On an M1 Mac laptop, we cannot use virtual box, so follow these instructions ins
 
 
 3. Download and save the Ubuntu 20.04 ARM Server ISO
-   https://mac.getutm.app/gallery/ubuntu-20-04
+   https://mac.getutm.app/gallery/ubuntu-20-04 (click "Alternative and previous releases")
```
MasterOdin, line 43: Going to this URL, I could not see anything that said "Alternative and previous releases", and the ISO was under the "Downloads" header.

jssunray34: I updated the instructions to clarify, thanks for pointing this out.

README_utm_arm.txt
```diff
@@ -40,9 +40,11 @@ On an M1 Mac laptop, we cannot use virtual box, so follow these instructions ins
 
 
 3. Download and save the Ubuntu 20.04 ARM Server ISO
-   https://mac.getutm.app/gallery/ubuntu-20-04
+   https://mac.getutm.app/gallery/ubuntu-20-04 *
    https://cdimage.ubuntu.com/releases/20.04.4/release/
 
+   * click "Ubuntu Server for ARM", then click "Alternative and previous releases"
```
MasterOdin, line 46: Okay, so I see now, I never clicked the download link on the `mac.getutm.app` page and didn't realize it took you to Ubuntu's site.

Did you end up using any other part of that utm page other than the link into Ubuntu's site for the download? Going to "Alternative and previous releases" seems to only present amd64 releases and not the arm64 as we need.

Going to `https://cdimage.ubuntu.com/releases/20.04.4/release/` does give us the ARM server ISO pretty immediately.

My recommendation would be to rewrite this entire section to just have the one link and avoid the clickthrough from UTM's site:
```suggestion
   https://cdimage.ubuntu.com/releases/20.04/release/
```

and then we could include the `https://mac.getutm.app/gallery/ubuntu-20-04` at the top of this page as just a reference to it.

jssunray34: I didn't download from the UTM site, so I agree we should just have people use the first link. I've updated the README accordingly.

MasterOdin (approved):

MasterOdin:
Thanks for improving our docs!

PR #8464, opened 2022-10-01
[UI/UX:Submission] Clarification for Preferred Name Order 

fatimaelfasi:

Just provided a suggestion to make name order clearer. No actual change to code functionality.

### DIFF
--- a/site/app/templates/EditNameForm.twig
+++ b/site/app/templates/EditNameForm.twig
@@ -12,7 +12,7 @@
         <input id="user-preferred-name-csrf" type="hidden" name="csrf_token" value="{{ csrf_token }}" />
         <h2>My Preferred Name</h2>
         <div class="edit_user_row">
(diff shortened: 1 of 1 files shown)


bmcutler:
Sorry, this is a 'hack' and is not going to satisfy the needs described in the issue.  We don't want to ask students or administrators to incorrectly swap the students names.  The family name has specific meaning legally.  Usually names are sorted by family name whether it is displayed as the first or last name when combined with the given name.  Furthermore, we get data from the university registrar with 2 name fields, and it would be incorrect and confusing to swap values for students who prefer to have their names displayed in the other order.

fatimaelfasi:
@bmcutler Thank you for the response. I am just beginning to learn about open source programs and this was my first attempt at a contribution, so I truly appreciate the feedback. I had not considered that the info is fed in from university records.

PR #8467, opened 2022-10-01
[Bugfix:InstructorUI] Allow non-ascii characters in user names

### LINKED ISSUE #8466: Allow foreign names with Latin characters to be accepted by User validation
Currently, first and last names with special latin characters outside of the ASCII subset are rejected by Submitty's user validation.
The current workaround when faced with this issue is to change the user's name in the system to use only basic English letters.
By allowing these special characters to pass validation, it can make for a better user experience for users with foreign names.

skara9:
First and last names with special Latin alphabet letters, such as letters with accents or umlauts,  are rejected by user validation.
Resolves #8466

### What is the new behavior?
The validation for User first and last names has been updated to include a subset of Latin-1 characters. This means names with these letters will no longer cause issues with user accounts on Submitty.

### DIFF
--- a/.setup/bin/versions.sh
+++ b/.setup/bin/versions.sh
@@ -7,8 +7,8 @@
 # SUBMITTY REPOS
 export AnalysisTools_Version=v22.03.00
 export AnalysisToolsTS_Version=v22.08.01
-export Lichen_Version=v22.05.00
-export RainbowGrades_Version=v22.07.00
+export Lichen_Version=v22.10.00
+export RainbowGrades_Version=v22.10.00
 export Tutorial_Version=v22.06.00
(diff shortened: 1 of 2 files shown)


williamjallen:
Hi @skara9, thanks for this contribution!  Can you maybe fill out the pull request body a little more?  In particular, it would be good to add some information about what Latin characters are, why this change is helpful, and why the changes in this PR are the best way to fix the issue in case anyone needs to reference this change in the future.

Although not strictly necessary for this PR to be merged, it might also be good to update the relevant tests to prevent this feature regressing in the future.

MasterOdin (approved):

MasterOdin:
Thanks for the bug fix!

PR #8470, opened 2022-10-02
[UI/UX:HelpQueue] Updated office hour queue button color

kevinochoa8266:
Text color is not the same in light mode as it is in dark mode.

### What is the new behavior?
Updated to have the same color text.

### Other information?
Used the CSS Inherit Keyword to inherit color from the parent.

### DIFF
--- a/site/app/templates/officeHoursQueue/QueueHeader.twig
+++ b/site/app/templates/officeHoursQueue/QueueHeader.twig
@@ -27,7 +27,7 @@
         {% for queue in viewer.getAllQueues() %}
             {% if queue['open'] != 0 %}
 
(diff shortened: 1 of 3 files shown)


shailpatels (changes requested):
Thanks for the PR @kevinochoa8266!

If you could add an image for UI changes in the future that would be helpful for reviews. Currently, the black text is needed on the light queue filter buttons for accessibility reasons

image

See our documentation here for accessibility https://submitty.org/developer/software_and_system_design/web_accessibility

Basically the contrast between the white text on the pastel colors is too low. You can use a plugin like WAVE which will show errors like contrast.

However when a queue is being filtered out, having the text become white so its visible on the black background is a good fix

image

So I think the behavior we want is to inherit colors when the queue is being filtered out and not having a color be filled in, but when the button does contain color it should use default black text. Take a look at the `fully-transparent` CSS class being defined that might a good place to put the change instead.

shailpatels (approved):
Thanks for catching this issue and for your contribution!

PR #8480, opened 2022-10-05
[Feature:TAGrading] TA grading image opening w/rubric pages

DefinitiveAbove:
Currently, when opening a component with a page assignment in the rubric panel in TA grading, it will always attempt to only open "upload.pdf" and scroll to the page.

### What is the new behavior?
When opening a component with a page assignment in the rubric panel in TA grading, it will do the following:

If a file is open in the viewer - if it is an image in the bulk upload image name format ("._upload_page_<num>.<image_extension>"), it will attempt to open the image with the specified page number from the component; otherwise (or if it fails), it will try to open the PDF at the specified page number.

If there is no file open in the viewer - it will check the last file that the user opened in a full view - if it was an image in the bulk upload image name format ("._upload_page_<num>.<image_extension>"), it will attempt to open the image with the specified page number from the component; otherwise (or if it fails), it will try to open the PDF at the specified page number.

### Other information?
Depends on #7804 due to file name changes with bulk upload images; previous bulk upload images will not be automatically opened with this PR.
Updated scroll offset for PDF scrolling to be more accurate.

### DIFF
--- a/site/app/controllers/grading/ElectronicGraderController.php
+++ b/site/app/controllers/grading/ElectronicGraderController.php
@@ -4,6 +4,8 @@ namespace app\controllers\grading;
 
 use app\libraries\DateUtils;
 use app\libraries\DiffViewer;
+use app\libraries\GradeableType;
+use app\libraries\response\RedirectResponse;
 use app\libraries\routers\AccessControl;
 use app\models\gradeable\Component;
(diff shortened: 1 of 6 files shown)


Eceptonsu (approved):
Works as expected

PR #8489, opened 2022-10-07
Bugfix:Submission added leading zero at left of exam image files

### LINKED ISSUE #8430: add leading zeros on bulk upload page .pngs
When we bulk upload a exam with 12 pages, the images of each page are numbered without leading zeros.  And the default sort-by-string is out of order and can be confusing.
Screen Shot 2022-09-23 at 7 15 19 PM

Let's add leading zeros on these filenames to fix this problem.
NOTE: The goal is not just to fix the display in TA grading, but also to fix anywhere we might want to view & access these files (e.g., linux 'ls').

Assuming that the test will be < 100 pages is fine.

ibonkonesa:
This  commit adds a leading zero at left of exam images and Fixes #8430

### What is the new behavior?
Exam images from 1 to 9 will be name 01,02,03...despites 1,2,3

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -117,4 +117,5 @@ Matt Raneri
 Chris Reed  
 Zachary Wimer  
 Fu Chai
+Asher Gottlieb
 For their bug reports of specific security vulnerabilities :)  
--- a/site/app/controllers/grading/ElectronicGraderController.php
+++ b/site/app/controllers/grading/ElectronicGraderController.php
@@ -1124,6 +1124,11 @@ class ElectronicGraderController extends AbstractController {
             return;
         }
 
+        if (strpos($_POST['attachment'], "..") !== false) {
+            $this->core->getOutput()->renderJsonFail('Invalid path.');
+            return;
+        }
+
         $attachment_path = FileUtils::joinPaths($this->core->getConfig()->getCoursePath(), 'attachments', $gradeable->getId(), $submitter_id, $grader->getId(), $_POST["attachment"]);
         if (is_file($attachment_path)) {
             if (@unlink($attachment_path)) {


shailpatels (changes requested):
Thanks for the contribution!

I think we want to make sure the filenames are decided in one place so that different parts of the site are not changing it incase we need to make updates in the future.

I think a better location for this fix is to investigate the submitty python jobs that are responsible for creating the split images. Currently there is code that attempts to prefix it, but maybe this broke recently?

See here 
https://github.com/Submitty/Submitty/blob/main/sbin/submitty_daemon_jobs/submitty_jobs/bulk_upload_split.py#L66-L70

and 

https://github.com/Submitty/Submitty/blob/main/sbin/submitty_daemon_jobs/submitty_jobs/bulk_qr_split.py#L128-L130

MasterOdin:
@shailpatels What are you suggesting to be changed in the python code? The PHP code in `SubmissionController::ajaxUploadSplitItem` on main completely ignores the filename scheme that the python code generates, so regardless if the python code is correct/incorrect, we'll still have the "wrong numbering" without a change to the PHP code.

However, per this comment:
https://github.com/Submitty/Submitty/blob/cf22c6bf72d681abbea2fa57f246ebfd602fd07a/site/app/controllers/student/SubmissionController.php#L635

and for

> I think we want to make sure the filenames are decided in one place so that different parts of the site are not changing it incase we need to make updates in the future.

the change I think is that for each `$image` we pull out of `$image_files`, we want to grab the page number suffix that's already there and re-use that, as opposed to having PHP incorporate its own counter, doing something like:

```php
        foreach ($image_files as $image) {
            preg_match("/\d*$/", basename($image), $matches);
            if (!@copy($image, FileUtils::joinPaths($version_path, "upload_page_" . $matches[0] . "." . $image_extension))) {
                return $this->uploadResult("Failed to copy uploaded image {$image} to current submission.", false);
            }
            // other code
        }
```

This is a similar pattern as we use a handful of lines above on merging the previous version:

https://github.com/Submitty/Submitty/blob/cf22c6bf72d681abbea2fa57f246ebfd602fd07a/site/app/controllers/student/SubmissionController.php#L662-L664

shailpatels:
@MasterOdin  Ah my bad, you're correct. I didn't read the existing copy file function in the PHP side closely enough. I'd have to take a look at the output of what the python code is producing but I think the images are prefixed with what split pdf they belong to and then what page number they represent so I think the PHP code isn't using the original filename to only include the page info in the name

PR #8519, opened 2022-10-18
[Feature:SubminiPolls] Create Tomorrow's Polls Section

### LINKED ISSUE #8500: Separate Instructor "Future Polls" category into "Tomorrow's Polls" and "Future Polls"
An instructor who is re-using polls and has lots of "future" polls, would like to confirm that the polls for tomorrow's lecture are prepared and have the right day.

Screen Shot 2022-10-10 at 11 41 33 PM

Just show the polls that are one day away in a separate section.
Separate Instructor "Future Polls" category into "Tomorrow's Polls" and "Future Polls"

The order of the categories should be:

Today's Polls
Old Polls
Tomorrow's Polls
Future Polls

Yes, they are not chronological.  Only Today's Polls and Old Polls should be visible to non-instructors.

jssunray34:
### What is the new behavior?
"Future Polls" have been separated into "Tomorrow's Polls" and "Future Polls."

Closes #8500


williamjallen (changes requested):
Code generally looks good.  It would probably be good to add some tests to `site/cypress/e2e/polls.spec.js` for this case.

site/cypress/e2e/polls.spec.js
```diff
@@ -59,6 +59,14 @@ describe('Test cases revolving around polls functionality', () => {
         cy.contains('Poll 5');
 
         cy.get('#tomorrow-table').contains('Poll 5');
+
+        // delete the poll
+        cy.get('#tomorrow-table-dropdown').click();
+        cy.contains('Poll 5').siblings(':nth-child(2)').click();
+        cy.wait(1000);
```
williamjallen, line 66: Adding `cy.wait()` calls can result in flaky tests in the CI and should generally be avoided.  Is there another way to do this?

williamjallen (changes requested):

jssunray34:
@williamjallen I removed the `wait()` call, and the test still passes. I had initially added it because a similar test does the same (for "Poll Cypress Test"). Do you have any suggestions for how to guarantee the test passes with respect to timing?

Also, something weird was happening where I was inputting what I thought would be tomorrow's date into the date field within my cypress test. However, it appears to be two days ahead (and thus the test fails), and I'm trying to figure out what is going on since it worked previously. Could you see if you're getting the same behavior when you run the test suite?

jssunray34:
Also, one of the lint tests is failing, but I'm not sure why (the error message is not very clear to me).

williamjallen:
I believe that the failing python linting test will be resolved by updating from main.  

As for your failing e2e test, it looks like it is trying to navigate to `/courses/f22/sample/polls`, and is receiving an error page at that point.  Does the main polls page currently work for both students and instructors?

williamjallen:
I also just remembered why there was a `cy.wait()` command in there originally.  When you click the dropdown, a rather trivial AJAX request is made to update a cookie, which was timing out in the CI occasionally.  I have removed that request in my changes to #6708, so I think any potential flakiness in the tests should be resolved by that PR.

jssunray34:
@williamjallen I looked into the failing e2e tests and am unsure where the error is coming from. I also made sure that both the student and instructor poll pages do not error out when a new poll (to be released "tomorrow") is added. The test suite does not fail locally, so it's a bit difficult to debug. Would you be able to take a look if you have a chance?

williamjallen:
I saw some similar issues with the tests while working on #6708, so for now I think it is best to wait for #6708 to be merged and then I can fix this PR to work with the new system.

jssunray34:
What is the status of this PR?

williamjallen:
#6708 will (hopefully) be finished and merged this weekend.  Since this PR conflicts with #6708 significantly, I'll take over updating this PR to use Doctrine after that's merged.

EdenBendheim:
I saw #6708 was merged, if you would not like to review this further I would be happy to test this PR @williamjallen :).

williamjallen:
> I saw #6708 was merged, if you would not like to review this further I would be happy to test this PR @williamjallen :).

 It needs an (almost) complete rewrite to use the new Doctrine backend.  I promised that I'd do it and I haven't gotten around to it yet.  I was actually planning to work on it this week, but if you're interested you're welcome to take it on instead.

PR #8520, opened 2022-10-18
[Bugfix:TAGrading] PHP error when viewing pdf 

Eceptonsu:
There were PHP errors when viewing a pdf in the grading interface.
There were PHP errors when trying to download the pdf in the grading interface for individual and team gradeables that prevents the files from being downloaded.

### What is the new behavior?
Removed the PHP errors when viewing pdfs in the grading interface, also removes the PHP error when downloading the pdf.

### Other information?
Remove the `getAnonPath()` function since it is no longer used.


site/app/controllers/pdf/PDFController.php
```diff
@@ -14,6 +14,20 @@ public function __construct(Core $core) {
         parent::__construct($core);
     }
 
+    public function getRealPath(string $file_path, string $id): string {
```
williamjallen, line 17: Should this function throw an exception if the file doesn't exist, or if you don't have access to the file?  At a minimum, it should probably check for `..` in strings.  This isn't an active security issue but it might be good to future-proof things a bit so we don't accidentally add a security bug in the future.

cjreed121 (changes requested):
We should check for `..` and possibly even do a canI check on the path when doing getRealPath().

Eceptonsu:
> We should check for `..` and possibly even do a canI check on the path when doing getRealPath().

The original file path is actually invalid because it contains the anonymous id (thats why the php error is there), I have added checks for the file after the conversion by using `file_exist`. The other checks in the function checks for the access so I do not think `canI` is needed.

tkoz0 (approved):
Tested and works as expected.

site/app/controllers/pdf/PDFController.php
```diff
@@ -18,6 +18,10 @@ public function getRealPath(string $file_path, string $id): string {
         $real_path = "";
         $file_path_parts = explode("/", $file_path);
         for ($index = 1; $index < count($file_path_parts); $index++) {
+            if ($file_path_parts[$index] === "..") {
```
williamjallen, line 21: Would it potentially be more effective to just add a check for ".." at the top of this function?  E.g., instead of checking every part, just use `str_contains()` to check for ".." in the file path.

cjreed121: No we can't do this. `file..file.pdf` is a valid file name that would not raise security issues.

@Eceptonsu can you change this back

williamjallen: Perhaps it's outside the scope of this PR, but my thought here was that we should prohibit files with `..` in them altogether to reduce the attack surface in case someone makes a bad change somewhere down the line.  If you feel strongly that we should allow files with `..` in the file name, I'm fine with that.  Perhaps we should document this information in a comment here though.

cjreed121: I mean `file..pdf.pdf` is a normal file and wouldn't want it to not be renderable because of its name. I still believe the loop `..` check is necessary as that strictly checks for `..`. Maybe `/../` could be checked for but I think doing the split on `/` and checking for `..` is already pretty efficient. And `/../` might not cover all cases.

We could just leave this conversation unresolved and it'll stay visible.

PR #8527, opened 2022-10-19
[Bugfix:TAGrading] Check gradeable type on more endpoints

DefinitiveAbove:
Many endpoints from ElectronicGraderController.php/SimpleGraderController.php do not check gradeable type and try to use methods that are unavailable as a result (usually causing a frog robot).

### What is the new behavior?
Endpoints in ElectronicGraderController.php/SimpleGraderController.php that use a gradeable check the gradeable type and redirect out on fail.

### Other information?
Extends off of #8513


shailpatels (approved):
Changes look good! Tried URL hacking the endpoints between the simple and electronic gradeables and saw the newly added error messages

PR #8528, opened 2022-10-19
[Refactor:TAGrading] Enable collapsing in non-mobile view

DefinitiveAbove:
In the mobile view of the grading index for TA grading, collapsing grading sections is available. This is not available in the non-mobile view of the grading index.

### What is the new behavior?
In the non-mobile view of the grading index, collapsing/expanding grading sections is now available.

Expanded/collapsed grading section info is saved on a course-by-course basis in localStorage.


cjreed121 (changes requested):
It functions well. I think it would be nice to have some sort of horizontal line or spacing between sections once they are minimized as it feels like all sections are just merged together once they are closed right now. Also it would be nice if when hovering over the arrow to toggle, your mouse cursor turned to a pointer so you could tell it does something.

cjreed121 (changes requested):
Looks great but there is an issue with the auto collapse/expand. The cookie doesn't update for me so I never see it working. It seems cookies forbid special characters so I think the newline join is breaking it. Could probably just store a simple JSON object here. Also shouldn't this be local storage instead of a cookie? I think cookies should only be used if the server also needs to see and control its value.

williamjallen:
> Also shouldn't this be local storage instead of a cookie? I think cookies should only be used if the server also needs to see and control its value.

Cookies allow the server to give the page an initial state (open or closed).  If local storage was used, the user would see a brief flash between the page initially rendering and the JS updating the relevant value from local storage and then re-rendering.

cjreed121:
> > Also shouldn't this be local storage instead of a cookie? I think cookies should only be used if the server also needs to see and control its value.
> 
> Cookies allow the server to give the page an initial state (open or closed). If local storage was used, the user would see a brief flash between the page initially rendering and the JS updating the relevant value from local storage and then re-rendering.

Yeah I agree that could be beneficial in that case but we aren't using the cookies server side. So I think cookies should either be used server side or this should actually use local storage.

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -441,6 +441,8 @@ function sortTable(sort_element_index){
      * @return string
      */
     public function detailsPage(Gradeable $gradeable, $graded_gradeables, $teamless_users, $graders, $empty_teams, $show_all_sections_button, $show_import_teams_button, $show_export_teams_button, $show_edit_teams, $past_grade_start_date, $view_all, $sort, $direction, $anon_mode, $overrides, $anon_ids) {
+        $collapsed_sections = isset($_COOKIE['collapsed_sections']) ? explode("\n", $_COOKIE['collapsed_sections']) : [];
```
williamjallen, line 444: Aren't the cookies supposed to be being used right here on the server side?  The value is then used in `Details.twig`.  I agree that the formatting will need to be changed to remove the newline for it to actually work.  Perhaps I misunderstood what you meant?

cjreed121: Oh whoops, I thought this was all new code functionality. That all makes sense now for it to be cookies.

cjreed121: Actually it is new functionality but I just skipped that line in my review before it seems.

cjreed121 (approved):
Works well!

PR #8529, opened 2022-10-20
[Refactor:TAGrading] Speedup Details Page

cjreed121:
Currently when loading the details page for an electronic gradeable there is 2 n-length queries. When a course has a lot of users this slows down the page a lot.

### What is the new behavior?
Both n-length queries have been reduced down to single-length queries instead.

### Other information?
I tested on a mock production server (around 800 users) where the page load time was initially 2.7 seconds and is now 0.9 seconds.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -4465,6 +4465,11 @@ AND gc_id IN (
         return $this->course_db->rows();
     }
 
+    public function getAllTeamAnonIdsByGradeable($g_id) {
+        $this->course_db->query("SELECT team_id, anon_id FROM gradeable_teams WHERE g_id=?", [$g_id]);
+        return $this->course_db->rows();
+    }
+
     /**
      * Get gradeable-specific user anon_id
      *
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -3724,6 +3724,20 @@ ORDER BY gt.{$section_key}",
      * @return SimpleGradeOverriddenUser[]
      */
     public function getUsersWithOverriddenGrades($gradeable_id) {
+        $return = [];
+        foreach ($this->getRawUsersWIthOverriddenGrades($gradeable_id) as $row) {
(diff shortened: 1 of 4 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -3723,7 +3723,21 @@ public function getUsersWithExtensions($gradeable_id) {
...
+        return $return;
+    }
+
+    /**
+     * Return an array of users with overridden Grades
+     *
+     * @param  string $gradeable_id
+     * @return array
+     */
+    public function getRawUsersWithOverriddenGrades(string $gradeable_id) {
```
williamjallen, line 3740:
```suggestion
    public function getRawUsersWithOverriddenGrades(string $gradeable_id): array {
```

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -440,7 +440,7 @@ function sortTable(sort_element_index){
      * @param bool $show_edit_teams
      * @return string
      */
-    public function detailsPage(Gradeable $gradeable, $graded_gradeables, $teamless_users, $graders, $empty_teams, $show_all_sections_button, $show_import_teams_button, $show_export_teams_button, $show_edit_teams, $past_grade_start_date, $view_all, $sort, $direction, $anon_mode) {
+    public function detailsPage(Gradeable $gradeable, $graded_gradeables, $teamless_users, $graders, $empty_teams, $show_all_sections_button, $show_import_teams_button, $show_export_teams_button, $show_edit_teams, $past_grade_start_date, $view_all, $sort, $direction, $anon_mode, $overrides, $anon_ids) {
```
williamjallen, line 443: Should type hints be added to all of these parameters as well?

williamjallen (changes requested):

site/app/views/grading/ElectronicGraderView.php
cjreed121: I didn't here cause there was just so many arguments. Not sure how good the type checking is here anyways cause it gets dynamically passed these.

williamjallen (approved):
Changes look good.  I tested this as thoroughly as possible and saw no issues for any level of user with any sort of gradeable.

PR #8531, opened 2022-10-22
[Refactor:TAGrading] Fetch Verified Grader Eagerly

cjreed121:
Every time a GradedComponent is created, if a verifier_id exists it fetches the database for the verifier User. In a gradeable with a lot of verified marks this can get very expensive.

### What is the new behavior?
The verifier user was added to our GradedGradeable query so that info can get picked up there if it exists.

### Other information?
I tested this and I could still verify graders and see that marks were verified.


bmcutler (approved):
Yay!  Amazing speed improvement for large course with verified test grades!

PR #8532, opened 2022-10-22
[Refactor:CourseMaterials] Course Materials Access Eager Load

cjreed121:
Right now the course materials access and sections do not get eager loaded. This makes a lot of database queries which could be simplified.

### What is the new behavior?
I switched from DQL to findBy so they can be eager loaded and added fetch eager to course materials accesses. Now significantly less queries are used.


jenaybarela (approved):
Looks good!

PR #8534, opened 2022-10-23
[Refactor:SubminiPolls] Simplify repetitive queries

williamjallen:
The main polls page currently makes 4n queries for n polls in a given course.  In addition, each of the responses is loaded each time.  For a moderately-sized class which uses polls regularly such as Data Structures at RPI, this corresponds to roughly 50,000 (yes, fifty thousand) individual queries being executed in the span of a few minutes, which puts a significant strain on the database server. 

### What is the new behavior?
The polls model has been refactored to lazily load data for improved performance if those data are not needed.  Data is also queried in aggregate instead of creating a large number of small queries.  In a local test with 10 polls, the average load time dropped from 0.33 seconds to 0.24 seconds.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -5315,22 +5315,6 @@ AND gc_id IN (
     }
 
     /**
-     * Gets user or team submitters by id
-     *
-     * @param  string[] $ids User or team ids
-     * @return Submitter[]
-     */
-    public function getSubmittersById(array $ids) {
-        //Get Submitter for each id in ids
-        return array_map(
-            function ($id) {
-                return $this->getSubmitterById($id);
-            },
-            $ids
-        );
-    }
-
-    /**
      * Gets a single GradedGradeable associated with the provided gradeable and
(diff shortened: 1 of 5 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7785,68 +7785,61 @@ public function openPoll($poll_id) {
...
         $this->course_db->query("SELECT * from polls order by poll_id ASC");
         $polls_rows = $this->course_db->rows();
-        $user = $this->core->getUser()->getId();
 
         foreach ($polls_rows as $row) {
-            $polls[] = $this->getPoll($row["poll_id"]);
+            $polls[] = new PollModel($this->core, $row["poll_id"], $row["name"], $row["question"], $row["question_type"], $row["status"], $row["release_date"], $row["image_path"], $row["release_histogram"]);
         }
 
         return $polls;
```
cjreed121, line 7797: Let's add some optional joins here with the responses, answers, and user_responses so that the full PollModel can be populated initially if we know we're going to need it later. This could either probably be in this function with a boolean or a separate function.

cjreed121 (changes requested):

PR #8539, opened 2022-10-24
[Feature:Forum] Display name changes

### LINKED ISSUE #8433: Add last name to instructors & full access in the forum
**Describe the bug**
Similar to #8431 & #8432

**Expected behavior**
Let's update the authors of forum post.
If the person in an instructor or a full access TA, lets make the author "Preferredfirstname Preferredlastname"
(unless they have the anonymous flag checked, then of course, leave it anonymous)
 
Leave the limited access graders and all students as "First name last initial"

jssunray34:
### What is the old behavior?

The display name on the discussion forum was previously "First name last initial" for all user groups. We would like to change this to "Preferredfirstname Preferredlastname" for instructors and full access graders.

### What is the new behavior?

For instructors and full access graders, the display name is now "Preferredfirstname Preferredlastname" but remains "First name last initial" for limited access graders.

Closes #8433


cjreed121 (changes requested):
This looks like it will cause a couple n-length queries where the query will occur in a for loop and create potentially a lot of database queries when it can be condensed. Could you change it so it only ever performs one query here.

jssunray34:
I looked back at the code to see how I can avoid making repeated calls to `isInstructorOrFullAccess` and am not sure how to avoid calling it in the for loop since it relies on the specific post/thread. Can you suggest how I may be able to condense the query calls?

cjreed121:
@jssunray34 Sure. A strategy you can use here is to make your query perform a `WHERE IN ()` rather than individually searching one at a time. DatabaseQueries has a helpful function called createParamaterList which takes a length and will produce `(?, ?, ..)` of specified length. Before you even enter these loops, you could quickly collect all user ids you know you will see, make the query storing result in an array, and in the loop access this array. When I say array, it should probably be a map for efficiency purposes.

jssunray34:
@cjreed121 I tried consolidating the database queries using your suggestion. Can you let me know if this aligns with what you were thinking? I was a little unsure how createParameterList would help me, so I didn't end up using it.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -818,6 +824,17 @@ public function isStaffPost($author_id) {
         return intval($this->course_db->rows()[0]['user_group']) <= 3;
     }
 
+    public function isInstructorOrFullAccess($author_id) {
+        $this->course_db->query("SELECT user_group FROM users WHERE user_id=?", [$author_id]);
+        return intval($this->course_db->rows()[0]['user_group']) <= 2;
```
williamjallen, line 829:
```suggestion
        return intval($this->course_db->row()['user_group']) <= 2;
```

williamjallen: It would also be good to use a named constant here instead of an arbitrary integer constant which is not as readable.

site/app/views/forum/ForumThreadView.php
```diff
@@ -841,10 +858,15 @@ public function displayThreadList($threads, $filtering, &$activeThreadAnnounceme
                 $email = trim($user_info['user_email']);
                 $first_name = trim($user_info["first_name"]);
                 $last_name = trim($user_info["last_name"]);
+                $visible_username = $first_name . " " . substr($last_name, 0, 1) . ".";
+
+                if ($this->core->getQueries()->isInstructorOrFullAccess($first_post["author_user_id"])) {
+                    $visible_username = $first_name . " " . $last_name;
+                }
```
williamjallen, line 865: This query is nested inside a `foreach` loop, which will result in a lot of repetitive queries.  It would be better to make one query which gets all of the necessary data and caches it instead.

site/app/views/forum/ForumThreadView.php
```diff
@@ -949,6 +971,10 @@ public function createPost($thread_id, $post, $unviewed_posts, $first, $reply_le
             $reply_level = 1;
         }
 
+        if ($this->core->getQueries()->isInstructorOrFullAccess($post["author_user_id"])) {
+            $visible_username = $first_name . " " . $last_name;
+        }
```
williamjallen, line 976: Same comment about repeated queries here

williamjallen (changes requested):

jssunray34:
What is the status of this PR?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -818,6 +824,17 @@ public function isStaffPost($author_id) {
         return intval($this->course_db->rows()[0]['user_group']) <= 3;
     }
 
+    public function isInstructorOrFullAccess($author_id) {
+        $this->course_db->query("SELECT user_group FROM users WHERE user_id=?", [$author_id]);
+        return intval($this->course_db->row()['user_group']) <= User::GROUP_FULL_ACCESS_GRADER;
+    }
```
cjreed121, line 830: Unused so should be deleted

cjreed121 (changes requested):
PR works well and adds minimal amount of database queries. Just one more comment and if you could update the PR description to contain old behavior that would be good. After that, I think it's good to go and will approve.

cjreed121 (commented):
You unfortunately have merge conflicts now. Can you fix those? Then I'll approve.

site/app/views/forum/ForumThreadView.php
```diff
@@ -949,6 +997,10 @@ public function createPost($thread_id, $post, $unviewed_posts, $first, $reply_le
             $reply_level = 1;
         }
 
+        if ($post["author_user_group"] <= User::GROUP_FULL_ACCESS_GRADER) {
+            $visible_username = $first_name . " " . $last_name;
```
cjreed121, line 1001: Don't forget to change this to given and family as well

cjreed121 (changes requested):

hansongu123:
It seems like this PR would be failing Cl run on the forum selenium tests more often than on main

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -768,6 +768,10 @@ public function getPosts() {
     }
 
     public function getPostsInThreads($thread_ids) {
+        if (count($thread_ids) == 0) {
```
williamjallen, line 771:
```suggestion
        if (count($thread_ids) === 0) {
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -768,6 +768,10 @@ public function getPosts() {
...
+            $rows = [];
+            return $rows;
```
williamjallen:
```suggestion
            return [];
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -825,6 +829,10 @@ public function isStaffPost($author_id) {
     }
 
     public function getAuthorUserGroups($author_ids) {
+        if (count($author_ids) == 0) {
```
williamjallen, line 831:
```suggestion
        if (count($author_ids) === 0) {
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -825,6 +829,10 @@ public function isStaffPost($author_id) {
...
+            $rows = [];
+            return $rows;
```
williamjallen:
```suggestion
            return [];
```

williamjallen (changes requested):

PR #8549, opened 2022-10-26
[Bugfix:TAGrading] Resolution change for downloaded pdf

Eceptonsu:
When downloading the PDF in the grading interface, the resulted file will have a low resolution.


tkoz0 (approved):
Tested and works as expected.

PR #8550, opened 2022-10-26
[Refactor:Submission] Improve graded gradeable query

DefinitiveAbove:
Query for retrieving graded gradeables is slow; a query run on the "Gradeables" view in "sample" course took 1.218 seconds

### What is the new behavior?
Query time for retrieving graded gradeables has been improved - changed the join for gradeable_component_data to be a lateral join to allow for filtering before joins with gradeable/gradeable_data - when tested, query performance on the same view improved to 19.1 ms.

Added index on gradeable_component_data for gd_id - when tested, query performance on the same view improved to 9.01 ms.


williamjallen (approved):
I tested this for all levels of users and for all gradeable types.

PR #8555, opened 2022-10-28
[Bugfix:System] sync registration type, main->course db

tkoz0:

Edits to registration type used to be made in the course database and were not updated automatically in the main database. This left different data in both the main and course databases if the registration type was changed for any user.

### What is the new behavior?
This adds a migration that fires the trigger to copy the data from the main database into the course databases so that the data is consistent in both.

### Other information?
Tested by manually editing users' registration type in course databases and checking that the data in the main database gets copied to replace the changed values when running the migration.


cjreed121 (approved):
Worked for me locally.

PR #8558, opened 2022-10-29
[Testing:Developer] Python linter fix

williamjallen:
### What is the new behavior?
A faulty assumption was made while creating PR #8541.  Even though the `cleanup_job()` method is in an abstract class, several subclasses do not implement it which causes errors during the bulk upload process amongst other things.  This PR implements the `cleanup_job()` function in all of the relevant locations such that the abstract base class is fully abstract.

### Other information?
This mistake could have been easily prevented by having a handful of tests for the bulk upload process.  Since problems with the bulk upload process seem to be a recurring issue, perhaps a few e2e tests would be enough to catch basic stuff like this.


sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
```diff
@@ -52,8 +52,7 @@ def validate_job_details(self):
     def run_job(self):
         pass
 
-    @abstractmethod
-    def cleanup_job(self):
+    def cleanup_job(self):  # noqa: B027
         pass
```
MasterOdin, line 56: Does changing this to a bare `return` also fix the `B027` error? Perhaps not as "pythonic" for an empty function, but :shrug:

williamjallen: There were only a limited number of locations where this function needed to be implemented so I have just gone ahead and implemented it in those locations.

shailpatels (approved):

PR #8562, opened 2022-10-30
[Bugfix:Submission] Unnecessary Bulk UI Elements

Eceptonsu:
Bulk related UI elements appears on the non-bulk related gradeables.

### What is the new behavior?
Removed the unnecessary bulk UI elements.
Cleaned the variable list for `SubmitBox.twig `
Updated the existing test cases to match the current behavior, in addition, a new sample course, Bulk Upload Scanned Exam, is created for testing purposes.


shailpatels:
Just so I'm caught up with the context, 

Before the behavior was that any gradeable could technically be a bulk upload gradeable if an instructor selected the radio button for bulk upload and uploaded something

But now an instructor must select the specific options in a gradeable config to make it a bulk upload in order to actually use it, is that correct?

There's a test that would check for all 3 elements (normal, submit for student, and bulk upload) so I think we want to separate that into 2 tests, one to make sure only the first two option exists on any gradeable and another to check that the bulk upload stuff only shows up on an actual bulk upload

Eceptonsu:
> @Eceptonsu Just so I'm caught up with the context,
> 
> Before the behavior was that any gradeable could technically be a bulk upload gradeable if an instructor selected the radio button for bulk upload and uploaded something
> 
> But now an instructor must select the specific options in a gradeable config to make it a bulk upload in order to actually use it, is that correct?
> 
> There's a test that would check for all 3 elements (normal, submit for student, and bulk upload) so I think we want to separate that into 2 tests, one to make sure only the first two option exists on any gradeable and another to check that the bulk upload stuff only shows up on an actual bulk upload

Yes, there is currently a specific gradeable for bulk submission. 
And yes, I will change the corresponding test cases.

PR #8575, opened 2022-11-01
[Testing:TAGrading] TA Rubric Attachments Access

ashergottlieb:

No change in behavior

### Other information?
Adds a e2e selenium test for the attachment endpoint.
Essentially a regression test for 0e894156e9c4521fa199f6511b34abce91538817, with an additional test for uploading and deleting an attachment through normal means.
Currently unclear if the test will work in CI given the different environment than the vagrant development VM.
However, this test does run successfully in the vagrant VM. Additionally, it uses the requests API in addition to selenium, as one request needs to be performed manually. The requests API appears to be installed on the vagrant VM, but not explicitly: it is installed by apt as a dependency of the `cloud-init` package.

As noted in the code, some tests fail with chromium + chromium-chromedriver when chromium is installed from the snap store (implicitly via `apt install chromium-browser`), and I speculate that this is due to sandboxing. I worked around this by using google chrome, although this issue isn't explicitly addressed in the docs.


tests/e2e/test_access.py
```diff
@@ -0,0 +1,97 @@
...
+        # This is kept as a single test since information from the first test
+        # is needed for the second (specifically, the anon_id)
+
+        cookies = {k: self.driver.get_cookie(k)['value']
+                   for k in ['submitty_session', 'PHPSESSID', 'submitty_token']}
+        csrf = self.driver.find_element(By.CSS_SELECTOR, 'body').get_attribute('data-csrf-token')
+
+        # 1. Go to grading details page for "Grading Homework" gradeable
+        # obtain valid anon_id of peer gradeable assignment
+        self.get('/courses/f22/sample/gradeable/grading_homework/grading/details')
```
DefinitiveAbove, line 36: "f22" should be replaced with the current semester from "self.semester" here

tests/e2e/test_access.py
```diff
@@ -0,0 +1,97 @@
...
+        del_btn = '#attachments-list-ta > div:nth-child(1) > div:nth-child(1) > a:nth-child(4)'
+        presence = EC.presence_of_element_located((By.CSS_SELECTOR, del_btn))
+        WebDriverWait(self.driver, 10).until(presence)
+        self.driver.find_element(By.CSS_SELECTOR, del_btn).click()
+        # confirm that we do want to delete the attachment
+        Alert(self.driver).accept()
+
+        # attempt delete another file
+        file_rel_path = 'submissions/closed_homework/aphacker/1/part1_buggy2.py'
+        assert os.path.exists('/var/local/submitty/courses/f22/sample/' + file_rel_path)
```
DefinitiveAbove, line 86: It might be better to change this to test if the file exists using the `display_file`/`download` endpoints from MiscController since some developers run selenium tests on their local machine against the VM for various reasons (i.e. testing in non-headless mode to debug tests).

DefinitiveAbove (changes requested):
Test works on my machine, requested some minor changes.

tests/e2e/test_access.py
```diff
@@ -0,0 +1,126 @@
...
+
+        cookies, csrf = self.get_credentials()
+
+        # 1. Go to grading details page for "Grading Homework" gradeable
+        # obtain valid anon_id of peer gradeable assignment
+        self.get(f'/courses/{self.semester}/sample/gradeable/grading_homework/grading/details')
+        # accept responsibilities as grader modal popup
+        self.driver.find_element(By.ID, 'agree-button').click()
+        # click view all sections
+        self.driver.find_element(By.CSS_SELECTOR, 'a.btn:nth-child(1)').click()
```
DefinitiveAbove, line 70: A more specific selector should probably be used here if possible (when testing on main the test is failing here for me)

DefinitiveAbove (changes requested):

tests/e2e/test_access.py
```diff
@@ -0,0 +1,143 @@
...
+from selenium.webdriver.common.alert import Alert
+from selenium.common.exceptions import TimeoutException, ElementNotInteractableException
+import os
+import unittest
+from urllib.parse import urlparse, parse_qs
+
+# requests is not explicitly installed system wide in our requirements.txt
+# (.setup/pip/system_requirements.txt), but exists currently installed
+# in the vagrant vm as required by other packages
+import requests
```
williamjallen, line 14: Why not explicitly list it in `.setup/pip/system_requirements.txt`?

williamjallen (changes requested):
I'm a little worried about adding more tests which aren't run in the CI.  Is there a particular reason why these are disabled in the CI?  In general tests which aren't run regularly are almost meaningless--dozens of such selenium tests are currently failing, but aren't run in the CI so we don't see it.

williamjallen:
We definitely have the `sample` course in the CI--that's what almost all of our e2e tests run against.  We don't have the other courses (beyond the blank course), and we don't have any submissions in the sample course as far as I am aware though.

IDzyre:
See #9598 for more info, closing because Selenium has been removed, maybe re-write in Cypress at a later date.

PR #8576, opened 2022-11-02
[Refactor:InstructorUI] New Autograding Config Selector

Viyerelu23333:

old

### What is the new behavior?

new

### Other information?

This also fixes the error message when overriding time limit for non-existed student


site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -68,6 +68,9 @@ public function newPage($template_id = null) {
         $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('flatpickr', 'plugins', 'shortcutButtons', 'shortcut-buttons-flatpickr.min.js'));
         $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('flatpickr', 'flatpickr.min.css'));
         $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('flatpickr', 'plugins', 'shortcutButtons', 'themes', 'light.min.css'));
+        $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('select2', 'js', 'select2.full.min.js'));
```
DefinitiveAbove, line 71: Just to check - is there a specific reason that the "full" version of select2 is being used over the "standard" version here? (see https://select2.org/getting-started/builds-and-modules for differences)

Viyerelu23333: I think you are correct -- just tested, and it is safe to use the standard version here.  Thanks for your review

DefinitiveAbove (approved):

bmcutler (approved):
much better than before!
(minor request: wish we could edit the current text but I can't edit previous typing or even copy paste)

PR #8585, opened 2022-11-05
[Refactor:HelpQueue] Remove repetitive queries

williamjallen:
The office hours queue page makes a large number of repetitive queries, which slows Submitty down on days when the queues are busy and leads to longer load times.

### What is the new behavior?
The repetitive queries have been simplified.  On my developer environment, with only a small number of students in the queue, the average page load time was improved by 50% on the instructor/TA side, and improved by 25% on the student side.  These gains are likely larger in production when there are dozens of students in several queues at one time.  The office hours queue page on the student side still makes a relatively large number of unique queries, but simplifying these will likely require a more significant refactor and is thus outside the scope of this PR.  All repetitive queries have been fixed, which is the most important part.


Eceptonsu (approved):
Code looks good, and there is a significant performance improvement when testing around 15-20 students on my end.

PR #8597, opened 2022-11-09
[Bugfix:CourseMaterials] Unwanted deletion

akshatbatra:

Deleting a folder (suppose A) also deletes the files present inside the same folder as folder A (but outside folder A) that begin with the same character(s) (e.g. ABC.pdf) from the database.

#### To reproduce on the main branch:

1. Navigate to course materials page of 'Sample' course.
2. Delete the folder named 'words'.
3. Observe that the file 'words_1463.pdf' (in the outermost directory) also gets deleted.

### What is the new behavior?
When a folder is deleted, only the database records of that folder and the files inside it are deleted.


Eceptonsu (approved):
Works as expected

PR #8598, opened 2022-11-09
[Feature:CourseMaterials] Number of files/links to be deleted

akshatbatra:

When deleting a folder, no information regarding the number of files/links that'll be deleted is given.

### What is the new behavior?

Popup box mentions the number of files/links to be deleted:

number_of_files_and_links


site/public/js/server.js
```diff
@@ -171,7 +171,20 @@ function newDeleteCourseMaterialForm(id, file_name) {
...
+        let num_of_links = files_or_links.filter(function () {
+            return (($(this).siblings('.file-viewer').children('a[data-is-link="1"]').length) === 1);
+        }).length;
+        let num_of_files = files_or_links.length - num_of_links;
+        let file_s = (num_of_files > 1) ? 's' : '';
+        let link_s = (num_of_links > 1) ? 's' : '';
+        let num_links_txt = (num_of_links === 0) ? '</em>)' : ` and <b>${num_of_links}</b> link${link_s}</em>)`
+        cm_message += ` (<em>contains <b>${num_of_files}</b> file${file_s}` + num_links_txt;
+    }
+    $('.delete-course-material-message', form).text(cm_message);
```
williamjallen, line 187: Many of these variables can be declared using `const` instead of `let`.

site/public/js/server.js
```diff
@@ -182,7 +182,20 @@ function newDeleteCourseMaterialForm(id, file_name) {
...
+        const num_of_links = files_or_links.filter(function () {
+            return (($(this).siblings('.file-viewer').children('a[data-is-link="1"]').length) === 1);
+        }).length;
+        const num_of_files = files_or_links.length - num_of_links;
+        const file_s = (num_of_files > 1) ? 's' : '';
+        const link_s = (num_of_links > 1) ? 's' : '';
+        const num_links_txt = (num_of_links === 0) ? '</em>)' : ` and <b>${num_of_links}</b> link${link_s}</em>)`
+        cm_message += ` (<em>contains <b>${num_of_files}</b> file${file_s}` + num_links_txt;
+    }
+    $('.delete-course-material-message', form).html(cm_message);
```
williamjallen, line 198: Doing this introduces the possibility for XSS attacks.  The likelihood of one instructor attacking another via the filename is extraordinarily low of course, but it's generally bad practice to use `.html()`.  It would be better to use `document.createElement()` or similar instead.

williamjallen (changes requested):

site/public/js/server.js
akshatbatra: jQuery also provides a neat way to create an element along with allowing to specify inner text, see [this commit](https://github.com/Submitty/Submitty/pull/8615/commits/94e8b7beb2b772adb98864647ca5adc2aaf67118) from one of my PRs for example. Using [setHTML()](https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTML) is probably the simplest but isn't recommended as it is still experimental and support across browsers isn't satisfactory.

site/public/js/server.js
```diff
@@ -182,8 +182,12 @@ function newDeleteCourseMaterialForm(id, file_name, str_id = null) {
 
     $('.popup-form').css('display', 'none');
     var form = $("#delete-course-material-form");
-    $('.delete-course-material-message', form).html('');
-    let cm_message = '<b>'+file_name+'</b>';
+    var deleteMessageElement = form.find('.delete-course-material-message')[0];
```
williamjallen: It's generally better to use `let` and `const` instead of `var` in JS due to `var`'s loose scoping rules.

```suggestion
    const deleteMessageElement = form.find('.delete-course-material-message')[0];
```

site/public/js/server.js
```diff
@@ -182,8 +182,12 @@ function newDeleteCourseMaterialForm(id, file_name, str_id = null) {
...
+    deleteMessageElement.innerHTML = '';
+
+    var cm_message = document.createElement('b');
```
williamjallen:
```suggestion
    const cm_message = document.createElement('b');
```

williamjallen (changes requested):

Youssef-Chip (approved):
I changed the variable types from var to const, it is good now to get merged.

PR #8605, opened 2022-11-13
[Testing:Developer] Add phpstan-doctrine plugin

williamjallen:
We do not currently use any PHPStan plugins for Doctrine, which means that read-only properties of Doctrine entities generate false positives in PHPStan.

### What is the new behavior?
The [`phpstan-doctrine`](https://github.com/phpstan/phpstan-doctrine) plugin has been added.  Among other things, this extension also performs some amount of DQL verification, and provides better type inference on the result of DQL queries.


williamjallen:
@cjreed121 I have made some edits to the forum and course materials entities to resolve the warnings generated by PHPStan Doctrine.  Since I'm not particularly familiar with this part of the code base, you might take a close look at the null/not null stuff.  Some of the fields were somewhat vague and it wasn't entirely clear whether some could have null values or not.

site/app/libraries/routers/Enabled.php
```diff
@@ -17,7 +17,7 @@
  * Example Usage:
  *
  * ```php
- * @Enabled("forum")
+ * @ Enabled("forum")
```
cjreed121, line 20: We can actually get rid of the space here. Doctrine annotations should be okay with it here since it has a parameter for the constructor. Since this is technically a usage example I think it should be as accurate as possible. And of course the same for FeatureFlag.php.

site/phpstan-baseline.neon
```diff
@@ -305,6 +300,16 @@ parameters:
 			count: 1
 			path: app/controllers/superuser/SuperuserEmailController.php
 
+		-
+			message: "#^Property app\\\\entities\\\\VcsAuthToken\\:\\:\\$expiration type mapping mismatch\\: property can contain DateTime\\|null but database expects DateTimeInterface\\.$#"
+			count: 1
+			path: app/entities/VcsAuthToken.php
```
cjreed121, line 306: Can we add that nullable flag here?

site/phpstan-baseline.neon
```diff
@@ -305,6 +300,16 @@ parameters:
...
+
+		-
+			message: "#^Property app\\\\entities\\\\course\\\\CourseMaterialAccess\\:\\:\\$course_material type mapping mismatch\\: database can contain app\\\\entities\\\\course\\\\CourseMaterial\\|null but property expects app\\\\entities\\\\course\\\\CourseMaterial\\.$#"
+			count: 1
+			path: app/entities/course/CourseMaterialAccess.php
```
cjreed121, line 311: Can nullable be used here? Or something else?

site/app/entities/forum/Post.php
```diff
@@ -29,14 +29,14 @@ class Post {
     /**
      * @ORM\ManyToOne(targetEntity="\app\entities\forum\Thread", inversedBy="posts")
      * @ORM\JoinColumn(name="thread_id", referencedColumnName="id")
-     * @var Thread
+     * @var Thread|null
      */
     protected $thread;
```
cjreed121, line 34: A post must have a thread so the null case is impossible.

cjreed121 (changes requested):
Thanks for updating the other entities.

site/phpstan-baseline.neon
williamjallen: The nullable flag can't be used on the `@ORM\ManyToOne` annotation.  It needs a join column to be specified, which potentially changes the behavior.  The same goes for the other ignored error.

bmcutler:
@williamjallen please resolve conflicts
@cjreed121 please re-review

cjreed121 (approved):

PR #8615, opened 2022-11-18
[Feature:CourseMaterials] Overwriting the materials

akshatbatra:

If a material is uploaded and a file already exists with the same name then a JS alert is shown to inform the user about the clash.

### What is the new behavior?
If file(s) already exist then a popup is shown allowing for single-click overwrite.

Overwrite_Course_Materials


EggBagel:
Does this affect just instructor/grader views, or does it also update the upload process for students? I was about to start work on adding a similar feature for student assignment uploads, and will move on to something else if this covers that.

akshatbatra:
> Does this affect just instructor/grader views, or does it also update the upload process for students? I was about to start work on adding a similar feature for student assignment uploads, and will move on to something else if this covers that.

This PR only deals with course materials. Course materials can only be uploaded by instructors, so, it's not related to student assignment submission. If you think there's a possibility to add a similar feature there, then you can continue your work.

site/public/js/server.js
```diff
@@ -178,6 +178,42 @@ function newDeleteCourseMaterialForm(id, file_name) {
...
+                html = html.replaceAll(elem[0], elem[1]);
+            });
+            return html;
+        });
+    }
+    else {
+        const singular = ['Material', 'file', 'name', 'one'];
+        form.html((i, html) => {
+            singular.forEach((elem) => {
+                html = html.replaceAll(RegExp(`( a){0,1}( ${elem})([ .:])`, 'g'), '$2s$3');
```
cjreed121, line 200: Can you put a comment what this reg exp is doing

site/public/js/server.js
```diff
@@ -178,6 +178,42 @@ function newDeleteCourseMaterialForm(id, file_name) {
...
+            });
+            return html;
+        });
+    }
+    var clash_list = $('#existing-names');
+    clash_list.html("");
+    clashing_names.forEach((elem) => {
+        clash_list.append(`<li>${elem}</li>`);
+    });
```
cjreed121, line 209: Having any sort of html element like this is a bit worrying. Could you make it dynamically add html elements and set text content.

cjreed121 (changes requested):

akshatbatra:
This PR should only be merged after #8684.

site/public/js/server.js
akshatbatra: Please see if the change is sufficient or something else would be needed.

hansongu123 (approved):
I can approve this PR functionality wise and since it's passing the site tests. @cjreed121 Chris please re-review to see if the requested changes are addressed by now. I can take over fixing it if not.

site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -788,4 +803,31 @@ private function addDirs(string $requested_path, string $upload_path, array &$di
             $j--;
         }
     }
+
+    private function resolveClashingMaterials(string $upload_path, array $file_names, bool $overwrite_all) {
```
williamjallen, line 807: This function should have a PHP type annotation for the return type

site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -142,7 +142,9 @@
                     const cmPath = document.getElementById('cm_path').innerHTML;
                     const requestedPath = document.getElementById("input-provide-full-path").value;
                     const sections_lock = document.getElementById('all_Sections_Showing_yes').checked === true;
-                    makeSubmission(expandZip, hideFromStudents, cmPath, requestedPath, cmTime, sort, sections, sections_lock);
+                    let overwrite_input = document.body.querySelector('input[name=overwrite-all]');
```
williamjallen, line 145:
```suggestion
                    const overwrite_input = document.body.querySelector('input[name=overwrite-all]');
```

site/cypress/e2e/courseMaterials.spec.js
```diff
@@ -782,4 +820,5 @@ describe('Test cases revolving around course material uploading and access contr
             });
         }
     });
+
```
williamjallen, line 823: this line isn't related to the rest of this PR and should be reverted to prevent potential merge conflicts with other PRs.

site/public/js/server.js
```diff
@@ -178,6 +178,45 @@ function newDeleteCourseMaterialForm(id, file_name) {
     form.find('.form-body').scrollTop(0);
 }
 
+function newOverwriteCourseMaterialForm(clashing_names, is_link, is_edit_form) {
+    const form = $('#overwrite-course-material-form');
+    form.css('display', 'block');
+    if (clashing_names.length === 1) {
+        let to_replace = [[' All', '']];
```
williamjallen, line 185:
```suggestion
        const to_replace = [[' All', '']];
```

site/public/js/server.js
```diff
@@ -178,6 +178,45 @@ function newDeleteCourseMaterialForm(id, file_name) {
...
+        const singular = ['Material', 'file', 'name', 'one'];
+        form.html((i, html) => {
+            singular.forEach((elem) => {
+                // replace singular words with plural by appending 's', taking care of the occurrences ending with . and :
+                html = html.replaceAll(RegExp(`( a){0,1}( ${elem})([ .:])`, 'g'), '$2s$3');
+            });
+            return html;
+        });
+    }
+    var clash_list = $('#existing-names');
```
williamjallen, line 206:
```suggestion
    const clash_list = $('#existing-names');
```

williamjallen (changes requested):
A handful of minor code style things

williamjallen (approved):
Changes seem to work as expected

PR #8624, opened 2022-11-22
[Feature:HelpQueue] Added counter for times in queue

jenaybarela:

Currently the only way to see how many time a student has been in the queue is to search for their name and that only provides information for today.

### What is the new behavior?
With this new feature it is easy to see how many time a student has been in a queue today as well as how many times they have been in any queue this week. 
image


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -6474,12 +6474,13 @@ public function getUsersNotFullyGraded(Gradeable $gradeable, $component_id = "-1
...
         return $this->course_db->rows();
     }
 
     public function getPastQueue() {
-        $this->course_db->query("SELECT ROW_NUMBER() OVER(order by time_out DESC, time_in DESC),* FROM queue where time_in > ? AND current_state IN ('done') order by ROW_NUMBER", [$this->core->getDateTimeNow()->format('Y-m-d 00:00:00O')]);
+        $current_date = $this->core->getDateTimeNow()->format('Y-m-d 00:00:00O');
+        $this->course_db->query("SELECT ROW_NUMBER() OVER(order by time_out DESC, time_in DESC),* FROM queue q LEFT JOIN (SELECT user_id as uid, COUNT(user_id) AS times_helped FROM queue WHERE extract(WEEK from time_in) = extract(WEEK from ?::DATE) AND (removal_type = 'helped' OR removal_type = 'self_helped') GROUP BY user_id) as h ON q.user_id = h.uid LEFT JOIN (SELECT user_id as uid, queue_code as qc, COUNT(queue_code) AS helped_today FROM queue WHERE time_in > ? AND (removal_type = 'helped' OR removal_type = 'self_helped') GROUP BY user_id, queue_code) as h1 ON q.user_id = h1.uid AND q.queue_code = h1.qc WHERE time_in > ? AND current_state IN ('done') order by ROW_NUMBER", [$current_date, $current_date, $current_date]);
```
williamjallen, line 6483: These two queries should be broken into multiple lines for readability.  Additionally, do you need the full left join here, or do you only need the count?

williamjallen (changes requested):

site/app/libraries/database/DatabaseQueries.php
jenaybarela: I'll make it more readable. The left joins are necessary because we need to pull all the data even for the students that have yet to be helped. If we do it without left join it doesn't get students that haven't been helped today.

williamjallen (changes requested):
This UI is somewhat confusing in my opinion, since it isn't obvious that "times helped today" is per-queue, while "times helped this week" is across all queues.  I don't have any particular suggestions which would be less confusing though.  My best idea is to simply make the "times helped this week" reflect the number of times helped in this queue to match the number for today.

image


I am also a little bit concerned about the performance hit resulting from the left joins in these two queries.  Postgres (using the `explain` tool) indicates that the changes make each of the queries substantially slower.  This is of expected given that a left join is being added, but I wonder if they could be made more efficient somehow.

sbelsk (approved):
Besides William's comments, I tested and all looks good!

williamjallen (approved):

PR #8625, opened 2022-11-22
[UI/UX:System] Changing SAML Management page colors

jenaybarela:

Currently there is an ugly brown color on the SAML Management. There is also some misalignment with the activate and deactivate button.

### What is the new behavior?
image


williamjallen:
Minor UI question: Is there a reason why we need the "Active?" column if there's already a button which clearly indicates whether the user is active or not?

jenaybarela:
I got rid of the active column and justified things to the left
Screenshot 2023-02-07 100359

williamjallen:
Would it be possible to include a screenshot of the "before" picture to compare to the "after" picture included in the PR body?

site/app/templates/superuser/SamlManager.twig
```diff
@@ -21,36 +21,34 @@
...
-                        <button class="btn btn-warning">Deactivate</button>
+                        <button class="btn btn-default">Deactivate</button>
                     </form>
                 {% else %}
                     <form method="POST" action="{{ change_active_saml_url }}" style="display: inline;">
                         <input type="hidden" name="csrf_token" value="{{ csrf_token }}">
                         <input type="hidden" name="activate" value="1">
                         <input type="hidden" name="id" value="{{ row['id'] }}">
-                        <button class="btn btn-warning">Activate</button>
+                        <button class="btn btn-primary" style="padding: 6px 21px">Activate</button>
```
williamjallen, line 51: This padding shouldn't be set manually like this.  I'm not certain what the intent is, but doing this is prone to conflicts in the future, given that the button already has padding set.  If we were to change the styling of the `btn`  class, this particular button would be left behind.

williamjallen (changes requested):

PR #8639, opened 2022-12-05
[Feature:Developer] Query performance warning

williamjallen:
Many Submitty pages query the database excessively.  A common cause of this problem is repetitive queries which can be reduced to a single larger query.  Some pages also just make a lot of different queries unnecessarily.

### What is the new behavior?
This PR adds a warning banner to the top of any page which executes duplicate queries or executes more than 20 different queries. This banner only appears in the development environment.

As a somewhat rudimentary guard against regression in functions which have been optimized, this PR also logs any route which violates the performance constraints to the site error logs (development environment only).  This will cause the CI to fail if any warnings are generated.  Many routes are not currently optimized, with some creating hundreds of queries.  These routes have been added to a new "ignore" list and should be removed if optimizations are completed to bring them into alignment with the query limits introduced in this PR.

New behavior:
image


site/app/libraries/Core.php
```diff
@@ -265,6 +265,56 @@ public function getCourseQueries(): array {
...
+        if ($this->submitty_debug_stack !== null) {
+            foreach ($this->submitty_debug_stack->queries as $query) {
+                $queries[] = $query['sql'];
+            }
+        }
+
+        return count($queries) !== count(array_unique($queries));
+    }
+
+    protected function logPerformanceWarning(): void {
```
cjreed121, line 292:
```suggestion
    private function logPerformanceWarning(): void {
```
Can be private I'm assuming cause there is no child of Core

site/app/templates/GlobalFooter.twig
```diff
@@ -43,6 +43,14 @@
         {% endif %}
     {% endif %}
     {% if is_debug %}
+        {% if performance_warning %}
+            <script>
+                $(document).ready(function() {
+                    $('#performance-warning-system-message').text('Developer Warning: Excessive or duplicate database queries observed: ' +
+                        '{{ submitty_queries|length + course_queries|length }} queries executed. Click "Show Page Details" at the bottom of the page for more info.').show();
+                });
```
cjreed121, line 51: Would be nice if this can be closed too

cjreed121 (changes requested):

cjreed121 (approved):
Looks good!

PR #8641, opened 2022-12-06
[Bugfix:Autograding] Edit error message in autograding 

dpshelio:
…ing files in the queue

In a very weird scenario, autograding logs a "non existence error" like:

```
== RuntimeError: ('ERROR: the file does not exist', '/var/local/submitty/in_progress_grading/primary_1/<semester>__<course>__<gradeable>__<userid>__1')
```

However, the file does exist at the path shown. For some inexplicable¹ reason the file does exist, but the runner of the autograding script doesn't have permissions to access it.

This results in the error above which gives the wrong error message but correctly throws an error indicating that something is wrong.

### What is the new behavior?

The autograding script now distinguishes between a non-existent file error and a permission error. Correctly reporting the cause of the error.

### Other information?

- Is this a breaking change?
  No

- How did you test
  Function doesn't have tests yet. The change doesn't introduce any alternative behaviours.

¹ - We are still investigating an error on our Submitty instance that has happened without any clear reason. All the jobs added to the queue were throwing non-existence file error messages, yet the files were present. The submissions were produced by a custom bulk submission script. That script has been tested and produces the correct behaviour.

@willGraham01


prestoncarman:
Are you able to create a test to demonstrate the error messages are working as expected?

bmcutler (approved):

PR #8647, opened 2022-12-08
[Feature:Submission] Time-Zone Check Against Local Timezone 

### LINKED ISSUE #8574: Add a timezone check against the user's local timezone.
Currently if your local timezone is changed, submitty doesn't update your timezone automatically or give you a warning. This can lead to some unfortunate circumstances. I suggest a check and display a warning if a timezone mismatch is detected.

marymx95:

Currently if your local timezone is changed, submitty doesn't update your timezone automatically or give you a warning.
Fixes #8574

### What is the new behavior?

When the user is on the home page, there is a time-zone check that alerts the user if there is a time-zone mismatch with their set time-zone on their profile and the users local time-zone.

Screen Shot 2022-12-08 at 10 36 58 AM
Screen Shot 2022-12-08 at 10 37 11 AM
Screen Shot 2022-12-08 at 10 37 26 AM

### Other information?

I tested this by updating the time-zone on the user profile to a different UTC than my computer's, then went back to the homepage. This resulted in seeing the warning message.

To make this better, we can keep track of if the warning was displayed already and stop it from displaying the next time the user goes to the homepage.


williamjallen:
Thanks for this contribution!  One question: should this banner be shown on other pages as well?

Vagrantfile
```diff
@@ -99,7 +99,7 @@ Vagrant.configure(2) do |config|
     ubuntu.vm.network 'forwarded_port', guest: 1511, host: 1511   # site
     ubuntu.vm.network 'forwarded_port', guest: 8443, host: 8443   # Websockets
     ubuntu.vm.network 'forwarded_port', guest: 5432, host: 16442  # database
-    ubuntu.vm.network 'forwarded_port', guest: 7000, host: 7000   # saml
+    ubuntu.vm.network 'forwarded_port', guest: 7000, host: 7001   # saml
     ubuntu.vm.network 'forwarded_port', guest: 22, host: 2222, id: 'ssh'
```
williamjallen, line 103: Is this change intended?

site/app/controllers/HomePageController.php
```diff
@@ -101,25 +102,64 @@ public function getGroups($user_id = null): MultiResponse {
      * @return MultiResponse
      */
     public function showHomepage() {
+        $user = $this->core->getUser();
         $courses = $this->getCourses()->json_response->json;
 
+
+
```
williamjallen, line 109: This whitespace is unnecessary.

site/app/controllers/HomePageController.php
```diff
@@ -101,25 +102,64 @@ public function getGroups($user_id = null): MultiResponse {
...
         return new MultiResponse(
             null,
             new WebResponse(
                 ['HomePage'],
                 'showHomePage',
                 $this->core->getUser(),
+                $this->core->getCsrfToken(),
```
williamjallen, line 116: Is this change necessary?  I believe that every page has access to the CSRF token since it's included in the global header.  (`GlobalHeader.twig`)

site/app/controllers/HomePageController.php
```diff
@@ -101,25 +102,64 @@ public function getGroups($user_id = null): MultiResponse {
...
 
+    /**
+     * @Route("/home/get_user_time_zone", methods={"GET"})
+     * @return JsonResponse
+     *
+     * Handle ajax request to update the currently logged in user's time zone data.
+     *
+     * Will return a json success or failure response depending on the result of the operation.
+     */
+    public function getUserTimeZone() {
```
williamjallen, line 131: Is there a reason why this is an AJAX request?  The system timezone could be included in the Twig logic for this page, essentially just setting the value of a JS constant.

site/app/controllers/HomePageController.php
```diff
@@ -101,25 +102,64 @@ public function getGroups($user_id = null): MultiResponse {
...
+
+
+
     /**
      * @Route("/home/courses/new", methods={"POST"})
      * @Route("/api/courses", methods={"POST"})
      */
     public function createCourse() {
+
+
```
williamjallen, line 162: This whitespace change is unnecessary.  (the same comment applies elsewhere in this file)

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -136,6 +136,10 @@ public function updateSubmittyUserTimeZone(User $user, string $time_zone) {
         $this->submitty_db->query("update users set time_zone = ? where user_id = ?", [$time_zone, $user->getId()]);
         return $this->submitty_db->getRowCount();
     }
+    public function getSubmittyUserTimeZone(User $user) {
+        $this->submitty_db->query("SELECT time_zone FROM users WHERE user_id = ?", [$user->getId()]);
+        return ($this->submitty_db->getRowCount() > 0) ? $this->submitty_db->row()['time_zone'] : null;
+    }
```
williamjallen, line 142: Instead of creating a separate query, is it possible to access this information through an existing `User` model?  Database queries are notoriously slow, so we want to minimize the total number of queries we make.

site/app/views/HomePageView.php
```diff
@@ -2,16 +2,20 @@
 
 namespace app\views;
 
+//check if needed
+//use app\libraries\Access;
```
williamjallen, line 6: The changes in this file should be reviewed.  I think a lot of them are probably unnecessary/unintended.  Also: see prior comment about not needing the CSRF token since it's already on every page.

site/public/js/homepage.js
```diff
@@ -0,0 +1,58 @@
...
+    const sign = (date.getTimezoneOffset() > 0) ? '-' : '+';
+    const offset = Math.abs(date.getTimezoneOffset());
+    let hours = Math.floor(offset / 60);
+    hours = (hours < 10 ? `0${hours}` : hours);
+    return `${sign + hours}:00`;
+}
+
+// once the homepage loads, does a timezone check against the user's local timezone and
+//if the user set a timezone,  warning will pop up if there is a timezone mismatch
+window.addEventListener('load', function () {
```
williamjallen, line 18:
```suggestion
$(document).ready(() => {
```

site/public/js/homepage.js
```diff
@@ -0,0 +1,58 @@
...
+    console.log("home page loaded");
+
+    $.getJSON({
```
williamjallen, line 21: See prior comment about AJAX request being unnecessary.  This code could just check the value of a JS constant set via Twig.

site/public/js/server.js
```diff
@@ -1,31 +1,32 @@
 ////////////Begin: Removed redundant link in breadcrumbs////////////////////////
 //See this pr for why we might want to remove this code at some point
+
```
williamjallen, line 3: Most of the whitespace changes in this file are distracting, and unrelated to this PR.  It would be good to lint this file eventually, but such changes should happen in a different PR.

williamjallen (changes requested):

site/app/controllers/HomePageController.php
marymx95: The AJAX request is used to retrieve the time zone that that user sets on the profile which is saved in the database.

marymx95:
> Thanks for this contribution! One question: should this banner be shown on other pages as well?
I believe that this banner should be show only on the homepage once the user logs in but I am unsure how to keep track of that. Any suggestions would be appreciated!

site/app/controllers/HomePageController.php
marymx95: Not necessary, was removed!

charvimehradu:
It seems to me that this pull request has been abandoned. Is there any way I can help or contribute to this?

bmcutler:
thanks for the contribution!
this work has been refactored and this PR is now replaced by #9309

PR #8648, opened 2022-12-09
[Bugfix:TAGrading] fixed peer grading stats 

### LINKED ISSUE #8275: Peer grading stats incorrect
**Describe the bug**
When looking at the grading stats of a peer gradeable, many of them are not sensible or just completely wrong.

Here is a screenshot of some averages:
image

Also the stats for the manually graded components were much higher than they actually were. Can't attach screenshot of this though.

**Expected behavior**
The stats should be accurate.

**To Reproduce**
Unsure of exact steps to reproduce since this was noticed on production. This assignment only had peer components.

**Configuration**
 - OS: Mac
 - Browser: Brave (Chromium)

cmkolb8:
### **What is the current behavior?**

Fixes #8275. 
The denominator of the stats "Current percentage of Peer Grading Done" currently only accounts for the number of students who were assigned one or more peer students to grade. However, since a single student can be assigned multiple peers, and numerator counts how many peer grades were graded in total, the numerator could be higher than the denominator, leading to percentages over 100%. 

When displaying individual peer grading stats, including the average and stdev of all assignments a student has graded of a certain component, the grading avg and stdev  currently takes the sum of all total points for a peer assignment. This results in displaying the total number of graded points for the avg and 0 for the stdev. 

### **What is the new behavior?**

I, thus, changed the denominator to add up the count of the number of peer assignments each student was assigned. (Ie, if there were only two students assigned peer assignments, and one was assigned two peers and the other four peers, the denominator would now say 6). Thus, when each student grades all of their assignments, it will then be 100%. 

The grading avg now takes the average of all assignments of that component that a grader has graded as well as the stdev. It sums over all assignments of that component by that grader, and divides by the number and performs a similar calculation for the stdev. 

### **Other Information?**
Previous behavior: 
Screen Shot 2022-12-09 at 1 51 26 PM 1
Current behavior:
Screen Shot 2022-12-09 at 1 50 58 PM
Previous behavior: 
https://user-images.githubusercontent.com/55092742/207722659-ced18d33-ff7f-4c15-87cc-90eb031dfd5f.png
Current Behavior: 
Screen Shot 2022-12-15 at 2 09 56 PM

Tested through creating different peer matrixes with students assigned to multiple peers to grade, to ensure percentage was correct and never went over 100%.


LunaFang1016 (commented):
LGTM!

LunaFang1016 (approved):
LGTM!

cjreed121 (changes requested):
Thanks for the contribution! Definitely seems to fix the percentage being over 100%. There is another part of this issue though. If you look at the right side with statistics of manually graded components, they aren't correct. Here's an example: 
image
Yet, as submitty-admin, my average is clearly not 1 as shown below.
image

If you can fix this too that would be great! Otherwise we can't close #8275 just yet.

Eceptonsu (approved):
Work for me.

PR #8650, opened 2022-12-12
[Bugfix:TAGrading] Fixed download zip button with dropdown

### LINKED ISSUE #8423: Download all submissions (& checkout & results) for all students & active only option
Currently you can download all students submissions folder for all versions.  
For a specific gradeable, on the  ta grading / 'grading stats' page there is a button at the bottom left.
The download zip includes the user_assignment_settings.json file indicates which version is the 'active' version (by default the highest numbered)

But an instructor might want to only download the active version (less to download & does not require consulting the user_assignment_settings.json file.
The download button has a drop down with checkboxes to select 'active only' and to request also downloading the results folder.  

BUG 1: But these options do not appear to work at all.  (OR maybe I'm confused about the UI...  but if so, this interface is not intuitive)

Furthermore, if the assignment is a submit by git, the download of the submissions folder is not the main assignment, we MUST ALSO include the parallel checkout folder, which contains the actual files in the repo at the time of submissions. 

BUG 2:   We are missing the .submit.timestamp file (and maybe other dotfiles) from the submissions folder.

BUG 3: download all for team submissions seems very broken....  Instead of having a layer of directories for each team name, there are just lots of directories that appear to be the submission versions for all students jumbled together.  NOTE:  I tested this with a gradeable that was both git & team, but all combinations should be tested & debugged.

LunaFang1016:
Fixes #8423 
For bug 1 mentioned in the issue, the "Download Zip of All / Assgined Students / Teams" button at the bottom left of  TA / instructor grading -> grading stats page has a drop-down menu. Instructors or TAs can choose to only download the current active version if they check the corresponding option. However, the checkbox does not work properly that all versions from all students' assignment are downloaded anyway even if the "only current active version" box is checked.

For the follow-up of bug 1, bug 2, and bug 3 described in the issue, I don't think they exist anymore, since when I tried to reproduce these errors, it turned out that the mentioned .timestamp file exists in each student's submission folder, and each team has one folder by their team id.

### What is the new behavior?
I located the bug to be in the `site\app\templates\grading\electronic\ta_status\StatusBase.twig` folder. Originally, the `editZipDownload` failed to assign the url to the correct button by button id. So I changed `#zip_link` to `#zip_link_limited` and `#zip_link_full` respectively to match the variable names before.

### Other information?
**Previous behavior:**
image
The url can't be configured according to checkbox selection. 
error_reproduce_2
The downloaded zip file thus still includes all versions even with the "Only Current Active Version" checked.
**Current behavior:**
fixed_download_url_active
`?type=active` when checking the second dropdown box.
fixed_download_url_both
`?type=both` when checking both dropdown boxes.
fixed_download_url_results
`?type=results` when checking the first dropdown box.
The url is configured properly.
fixed_download_file_active
And the downloaded file only contains one active version (if there is any) in every student's folder.
Tested by carrying out the download procedure through limited grader access, teamwork assignment, and individual assignment. Verified the validity of change by observing the url and downloaded file configuration.
Unit testing and documentation update not applicable.


alice-nie:
LGTM!

alice-nie (commented):
LGTM :)

cjreed121 (approved):
Looks good to me and fixes the problem. Thanks for the contribution!

PR #8660, opened 2022-12-12
[Refactor:Submission] Reducing splitted jpg sizes

Eceptonsu:
The splitted images for the uploaded pdfs in bulk submission currently take too much space.

### What is the new behavior?
The total size reduces to around 20% of the original total size.


bmcutler (approved):
Thanks for this improvement!

PR #8663, opened 2022-12-13
[Testing:InstructorUI] Cypress for DockerUI

Viyerelu23333:

No tests for DockerUI.  Docker hub API might get deprecated without any checks.

### What is the new behavior?

Added cypress tests for DockerUI.


.github/workflows/submitty_ci.yml
```diff
@@ -407,6 +408,10 @@ jobs:
             run: |
               cd $SUBMITTY_REPOSITORY
               sudo -E env "PATH=$PATH" bash .setup/testing/setup.sh
+              sudo jq '. |= . + {"verified_submitty_admin_user": "submitty-admin"}' /usr/local/submitty/config/submitty_users.json \
+                | sudo sponge /usr/local/submitty/config/submitty_users.json
```
shailpatels, line 412: Is there a reason the `sponge` command is being used here? I haven't run the full command locally but could you instead use something like `sudo tee --append <file>` instead ?

That way we could avoid installing `moreutils` in CI

shailpatels (changes requested):

.github/workflows/submitty_ci.yml
Viyerelu23333: As `jq` reads all the bytes in `submitty_users.json`, if the script contains something like 

```bash
jq 'PATTERN' submitty_users.json > submitty_users.json
```

the file will be cleared first, before `jq`.

The `sponge` here merely enables `jq` can process the content first — after receiving all data from pipe, it then opens the file and writes the data.

If this really matters, I could use something similar:

```bash
tmp=$(mktemp)
sudo jq 'PATTERN' submitty_users.json > ${tmp}
sudo mv -fv ${tmp} submitty_users.json
```

site/cypress/e2e/docker_ui.spec.js
```diff
@@ -0,0 +1,154 @@
...
+    cy.reload();
+    cy.then(() => {
+        return condition().then((result) => {
+            if (result || timeout <= 0) {
+                return result;
+            }
+            return waitAndReloadUntil(condition, timeout - wait, wait);
+        });
+    });
+};
```
williamjallen, line 14: This function might be useful elsewhere as well.  Perhaps it could be moved to the libraries instead of residing in this file.

.github/workflows/submitty_ci.yml
shailpatels: Oh gotcha, didn't realize that order was required - that makes sense to me how you did it then

shailpatels (approved):

site/cypress/e2e/docker_ui.spec.js
Viyerelu23333: Sure, how does registering a command in `/site/cypress/support/commands.js` sound?

williamjallen: That seems like a good idea.  I know of at least a few other use cases for this function.  In particular, I was struggling with this exact issue some time ago when I was working on Lichen UI tests.

williamjallen:
@Viyerelu23333 All of the tests seem to have passed.  Has the issue you brought up in the meeting yesterday been resolved at this point?  Given that the test ran and passed, it seems fine to me unless I'm missing the point.

Viyerelu23333:
> All of the tests seem to have passed. Has the issue you brought up in the meeting yesterday been resolved at this point? Given that the test ran and passed, it seems fine to me unless I'm missing the point.

Just tested few times again, and it is *surprisingly* working...
Perhaps I forgot to change my local Cypress files before the meeting... so it is probably not a recursion-related problem.

williamjallen (approved):
Approved if tests pass

Issue #8446, opened 2022-09-27
Preferred Name Order: Given Family vs Family Given


bmcutler:
**What problem are you trying to solve with Submitty**
When Submitty combines given & family names, we currently assumes "Given Family" is the correct order.  But the preferred order of names varies in different areas of the world -- users may prefer the combined order "Family Given". 

**Describe the way you'd like to solve this problem**
Allow students to specify in the profile which order their names should appear, when we display a combined name.

**Describe any potential alternatives you'd tried to solve the problem**
We are not currently storing the middle name.  This may impact the flexibility in constructing the correct combined name.  We may not be able to just assume that attaching the middle name after the given name is appropriate.  E.g.,  "Given Middle Family" vs "Family Given Middle" vs. "Family Middle Given".
Adding a middle name is probably not necessary for teaching & grading.  And we probably don't have a source of data for this (our direct registrar registration input does not include middle name).

Smeyer025:
Hi! I'm new to open source, and I would like to try to solve this issue. Do I need to be assigned the issue, or can I just get started?

LeoJung1:
Hi, is this issue open for change? If so, we would like to work on this issue as a team along with issue #8436!

abhishekyakhmi:
I would like to work on this issue  , if it's still open.

pranavsambyal:
Hey, I would like to propose a solution , that is to use ask the user to choose a display  name while signup which an be used. This avoids implementing a toggle switch for all  users preferred orders.

VaibhavKambar7:
Has this issue been solved?

sanyamk30:
Hey @bmcutler , is the issue still open? I would like to work on it

ziesski:
I reviewed PR #10259 today, and while golfing I was thinking about this PR and the design decision, and personally think this is NOT the way we should solve the problem.

It would be confusing if we allow users to decide which order individually. Usually it is obvious enough to tell which is Given and which is Family, but it is also common for people to have interchangeable Given and Family name. This would be hard to tell for users (and even worse for international students where English is not their first language). 

I rather would propose it to be set in submitty.json during initial configuration process, and if Submitty is used in different cultural location, admin can decide which order to use. 

If current implementation is applied to forum, it would possible make giant mess, and I think it is over empowering users. We should stick to the convention, and keep it consistent throughout the course. 

@bmcutler @micpap25

micpap25:
[...]

PR #8667, opened 2022-12-13
[Feature:API] User preferred order of their first and last names

LeoJung1:

* [ x ] Tests for the changes have been added/updated (if possible)
* [ x ] Documentation has been updated/added if relevant

This PR resolves issue #8446

The current Submitty name order only supports first name, last name order. People in regions where last name comes first would have to stick with this functionality.

### What is the new behavior?
The new feature adds the dropdown where the user would be able to choose the name order. The chosen name order will be saved in the database and be shown in the future documents that will be shown.

### DIFF
--- a/migration/migrator/migrations/master/20221206153408_add_user_preferred_name_order.py
+++ b/migration/migrator/migrations/master/20221206153408_add_user_preferred_name_order.py
@@ -10,7 +10,7 @@ def up(config, database):
     :param database: Object for interacting with given database for environment
     :type database: migrator.db.Database
     """
-    sql = "ALTER TABLE ONLY users ADD COLUMN IF NOT EXISTS user_preferred_name_order SMALLINT NOT NULL DEFAULT 1;"
+    sql = "ALTER TABLE ONLY users ADD COLUMN IF NOT EXISTS user_preferred_name_order SMALLINT NOT NULL DEFAULT 0;"
     database.execute(sql)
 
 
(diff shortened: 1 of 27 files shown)


migration/migrator/migrations/master/20221206153408_add_user_preferred_name_order.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Run down migration (rollback).
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    sql = "ALTER TABLE ONLY users DROP COLUMN IF EXISTS user_preferred_name_order RESTRICT;"
+    database.execute(sql)
```
cjreed121, line 27: In Submitty, we usually prefer that our down migrations don't delete data unless absolutely necessary for functionality. This this case we'd probably want the down migration to just be `pass`.

site/app/models/User.php
```diff
@@ -436,8 +448,17 @@ private function setDisplayedLastName() {
...
+        $first_name = $this->getDisplayedFirstName();
+        $last_name = $this->getDisplayedLastName();
+        switch ($this->preferred_name_order) {
+            case self::FIRST_LAST:
+                return $first_name . ' ' . $last_name;
+            case self::LAST_FIRST:
+                return $last_name . ' ' . $first_name;
+            default:
+                return $first_name . ' ' . $last_name;
+        }
```
cjreed121, line 461: We could simplify this a bit:
```suggestion
        switch ($this->preferred_name_order) {
            case self::LAST_FIRST:
                return $last_name . ' ' . $first_name;
            case self::FIRST_LAST:
            default:
                return $first_name . ' ' . $last_name;
        }
```

cjreed121 (changes requested):

cjreed121:
If you'd like me to continue reviewing, please mark it as ready for review.

PR #8684, opened 2022-12-15
[Bugfix:System] Handle array while formatting query

akshatbatra:

`formatQuery` function in `site/app/libraries/database/DatabaseUtils.php` throws an error if `$param` is an array.

### What is the new behavior?
`formatQuery` function can now handle an array (`$param`) inside `$params`.

### Other information?
This issue was discovered while working on #8615 where the following query is generated:

`SELECT t0.id AS id_1, t0.path AS path_2, t0.type AS type_3, t0.release_date AS release_date_4, t0.hidden_from_students AS hidden_from_students_5, t0.priority AS priority_6, t0.url AS url_7, t0.url_title AS url_title_8, t9.id AS id_10, t9.user_id AS user_id_11, t9.timestamp AS timestamp_12, t9.course_material_id AS course_material_id_13, t14.section_id AS section_id_15, t14.course_material_id AS course_material_id_16 FROM course_materials t0 LEFT JOIN course_materials_access t9 ON t9.course_material_id = t0.id LEFT JOIN course_materials_sections t14 ON t14.course_material_id = t0.id WHERE t0.path IN (?)`

Here, the relevant portion is at the last i.e. `WHERE t0.path IN (?)`. This is possibly generated by Doctrine's `findBy` call [here](https://github.com/Submitty/Submitty/blob/8fa2ce8e37d1a270bc3a151a0f65ab5af8e1a7e7/site/app/controllers/course/CourseMaterialsController.php#L800).

The error was encountered while trying to upload a material, see screenshot:
StackTrace

As can be observed in the above screenshot, this issue probably stayed hidden until recently a PR related to DB performance warning was merged.

### DIFF
--- a/site/app/libraries/database/DatabaseUtils.php
+++ b/site/app/libraries/database/DatabaseUtils.php
@@ -18,6 +18,29 @@ class DatabaseUtils {
             }
             if (is_array($param)) {
                 $str_param = json_encode($param);
+
+                /*
+                // in case queries with json function calls are not formatted properly
+                $json_functions = ["json_array_length", "json_array_elements", "json_array_elements_text"];
+                $function_call = false;
+                $qmark_info = [];
+                preg_match('/\?/', $sql, $qmark_info, PREG_OFFSET_CAPTURE);
+                foreach ($json_functions as $json_func) {
+                    $func_name_offset = $qmark_info[0][1]-strlen($json_func)-1;
+                    if ($func_name_offset >= 0 && substr($sql, $func_name_offset, strlen($json_func)) === $json_func) {
+                        $function_call = true;
+                        break;
+                    }
+                }
+                if ($function_call) {
+                    $param = "'{$str_param}'";
+                }
+                else {
+                    $param = substr($str_param, 1, strlen($str_param) - 2);
+                }
+                */
+
+                // remove square brackets
                 $param = substr($str_param, 1, strlen($str_param) - 2);
             }
             elseif (!is_numeric($param)) {
--- a/site/app/libraries/database/DatabaseUtils.php
+++ b/site/app/libraries/database/DatabaseUtils.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace app\libraries\database;
 
 use app\libraries\DateUtils;
-use app\libraries\Utils;
 
 class DatabaseUtils {
     public static function formatQuery(string $sql, ?array $params): string {
@@ -18,7 +17,7 @@ class DatabaseUtils {
                 $param = DateUtils::dateTimeToString($param);
             }
             if (gettype($param) === "array") {
-                $param = Utils::arrayToCommaSepString($param);
+                $param = json_encode($param);
             }
             elseif (!is_numeric($param)) {
                 $param = "'{$param}'";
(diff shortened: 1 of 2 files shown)


williamjallen (commented):
Thanks for catching this!  I noticed that the parameters weren't always getting put in the right spot a few days ago and didn't have time to look into why that was the case.

One comment: would it not be easier to use `json_encode` to convert the arrays to strings?

akshatbatra:
> Thanks for catching this! I noticed that the parameters weren't always getting put in the right spot a few days ago and didn't have time to look into why that was the case.
> 
> One comment: would it not be easier to use `json_encode` to convert the arrays to strings?

`json_encode` is definitely better, thanks for the suggestion.

site/app/libraries/database/DatabaseUtils.php
```diff
@@ -16,7 +16,13 @@ public static function formatQuery(string $sql, ?array $params): string {
             if ($param instanceof \DateTime) {
                 $param = DateUtils::dateTimeToString($param);
             }
-            $sql = preg_replace('/\?/', is_numeric($param) ? $param : "'{$param}'", $sql, 1);
+            if (gettype($param) === "array") {
```
cjreed121, line 19: We could probably do `is_array()` to follow the style of `is_numeric()`

site/app/libraries/database/DatabaseUtils.php
```diff
@@ -16,7 +16,13 @@ public static function formatQuery(string $sql, ?array $params): string {
...
+                $param = json_encode($param);
```
cjreed121, line 20: json_encode seems to be a decent option but when it gets substituted in it'll produce `([1, 2, ...])`. Then `[` and `]` are invalid SQL. So if we use json_encode we should make sure those braces get removed so it's valid SQL.

cjreed121 (changes requested):

site/app/libraries/database/DatabaseUtils.php
akshatbatra: I realized that but then thought about [PostgreSQL's JSON functions](https://www.postgresql.org/docs/9.5/functions-json.html). I was not sure if Doctrine accepts an array in some method and then uses it as an argument in a call to a JSON function in the resulting query, in which case it may be correct to keep square brackets and also single quotes e.g. `json_array_elements_text('["foo", "bar"]')`. Could this be a possible scenario?

cjreed121: Yeah I think using `json_array_elements_text` is a good option. Just make sure it works for both strings and integers (in the sense that it's valid SQL and collects the right data). Then we won't have to worry about some json edge case.

akshatbatra: `json_array_elements_text` creates issues concerning valid syntax. As of now, any issue related to some JSON edge case hasn't been discovered, handling it may result in unneeded additional complexity.
Still, I have added some code that deals with some JSON functions, but as it may be redundant, I've commented it. It can be used later if the need may arise...

cjreed121 (approved):
Sorry for not getting to this sooner. Looks good to me!

PR #8726, opened 2023-01-06
[Feature:Autograding] Config option to keep `.git`

Gerzer:

* [X] Documentation has been updated/added if relevant _(will be submitted as a separate pull request to the documentation repository)_

Submitty automatically deletes the `.git` directory while copying the current state of a student’s VCS submission to the temporary auto-grading workspace when a student clicks the “Grade My Repository” button.

### What is the new behavior?
Submitty automatically deletes the `.git` directory while copying the current state of a student’s VCS submission to the temporary auto-grading workspace when a student clicks the “Grade My Repository” button unless the auto-grading configuration’s `config.json` file has the `"keep_git_directory"` field set to the Boolean `true` at the root of the JSON object.

Any value other than `true`—including the string `"true"`—is implicitly coerced to `false`. Furthermore, in the absence of the JSON field, the current behavior is maintained. In other words, the default behavior remains to delete the `.git` directory, but there’s now an option to override that behavior for a specific auto-grading configuration.

### Other information?
This is not a breaking change because it relies on an entirely new, optional auto-grading configuration JSON field to change the current behavior. No existing auto-grading configuration should include this field because until now, it was invalid. Auto-grading configurations that continue not to include this field will continue to observe the same behavior as before.

I and @cjreed121 tested the change with a local development Submitty installation.

This change supports the needs of CSCI-2961 Rensselaer Center for Open Source in first few weeks of the Spring ’23 semester. We’re working on an auto-grading configuration with a custom validation script to grade the structure and commit history of students’ Git repositories, which needs to interact with the `.git` directory directly. This is part of an initiative to ensure that all RCOS students have baseline Git skills at the start of the semester. As such, we would greatly appreciate a timely deployment to submitty.cs.rpi.edu!

### DIFF
--- a/bin/json_schemas/complete_config_schema.json
+++ b/bin/json_schemas/complete_config_schema.json
@@ -515,6 +515,10 @@
             }
           }
         }
+      },
+      "keep_git_directory": {
+        "type": "boolean",
+        "description": "Specifies whether the .git directory should be copied into the temporary workspace for a custom validation script"
       }
   },
   "definitions" : {
--- a/bin/json_schemas/complete_config_schema.json
+++ b/bin/json_schemas/complete_config_schema.json
@@ -518,7 +518,7 @@
       },
       "keep_git_directory": {
         "type": "boolean",
-        "description": "Specifies whether the .git directory should be copied into the temporary workspace for a custom validation script"
+        "description": "Specifies whether the .git directory should be kept on clone into the student's checkout submission. Defaults to false."
       }
   },
   "definitions" : {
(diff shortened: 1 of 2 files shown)


bin/json_schemas/complete_config_schema.json
```diff
@@ -515,6 +515,10 @@
             }
           }
         }
+      },
+      "keep_git_directory": {
+        "type": "boolean",
+        "description": "Specifies whether the .git directory should be copied into the temporary workspace for a custom validation script"
```
cjreed121, line 521:
```suggestion
        "description": "Specifies whether the .git directory should be kept on clone into the student's checkout submission. Defaults to false."
```

cjreed121 (changes requested):

bmcutler:
Discussed offline.  Alternate solution found that does not involve storing a potentially large/complete history .git subdirectory.

PR #8733, opened 2023-01-11
[Bugfix:Submission] Generate team anon id on creation

DefinitiveAbove:
When creating a team for a gradeable, an anon_id is not automatically generated for it (generated when getAnonId).

### What is the new behavior?
When creating a team for a gradeable, an anon_id is automatically generated as part of the creation process.

A migration has been added to fix cases in the database where an anon_id was not generated for a team (when creating teams using the bulk-upload interface).


bmcutler (approved):
thanks for fixing this!

PR #8753, opened 2023-01-22
[Feature:HelpQueue] Generate Help Queue Data

shinyclovers:

Currently, there is no way for developers to test features related to the help queue without manually logging onto multiple accounts and manually joining the queue. This would be troublesome if a particular feature would require many students to be in the queue at once

### What is the new behavior?
A script has been created in order to automate the process of students joining the Help Queue without the need of the developer to open multiple windows in order to simulate it.

### Other information?

The script is still in its early stages where it makes a lot of assumptions (for example, which course and what queue it would be populating, the amount of students to add to the queue, whether the data should also include students who are helped). I have plans to turn those assumptions into default values of command line arguments that the developer would pass in. It would be appreciated if developers can provide feedback on what they would like to see in the script as the script should benefit developers trying to develop features with Help Queue.


.setup/bin/populate_office_hours_queue.py
```diff
@@ -0,0 +1,195 @@
...
+from datetime import datetime, timedelta
+
+DB_HOST = "localhost"
+DB_PORT = 5432
+DB_USER = "submitty_dbuser"
+DB_PASS = "submitty_dbuser"
+
+def main():
+    # get today's date to determine date and time
+    today = datetime.today()
```
shailpatels, line 16: There's a similar function in the python submitty utils lib that gets either the summer or fall term.

```python
from submitty_utils import dateutils

_ = dateutils.get_current_semester()
```

The function is under `python_submitty_utils/submitty_utils/dateutils.py`, it doesn't handle `u` for summer terms but I think that's ok since this is just sample data

.setup/bin/populate_office_hours_queue.py
```diff
@@ -0,0 +1,195 @@
+#!/usr/bin/env python3
+
+import argparse
+from sqlalchemy import create_engine, Table, MetaData, select, update, insert
+import random
+import os
+from datetime import datetime, timedelta
+
+DB_HOST = "localhost"
```
shailpatels, line 9: Is this intended to be run from inside or outside the VM?
A lot of the scripts in .setup are expected to be executing from inside so they can access the configuration data that gets generated with the first VM up

so from inside the VM you could do something like
`python3 GIT_CHECKOUT/Submitty/.setup/bin/<YOUR_SCRIPT.py>` and in python access the `$SUBMITTY_INSTALL_DIR/.setup/submitty_conf.json` file

which contains whatever the user configured their DB's host/port/username/pswd to be

shailpatels (changes requested):

.setup/bin/populate_office_hours_queue.py
```diff
@@ -0,0 +1,191 @@
...
+    if len(all_queues) == 0:
+        return
+
+    # Find all students in class
+    res = conn.execute(select(users_table))
+    all_users = res.fetchall()
+    res.close()
+
+    # Find all ids of students and graders
+    all_student_ids = [x['user_id'] for x in all_users if x['user_group'] == 4 and x['registration_section'] != None]
```
williamjallen, line 47: It's generally preferable to use `is`/`is not` when comparing something to `None` in Python.

```suggestion
    all_student_ids = [x['user_id'] for x in all_users if x['user_group'] == 4 and x['registration_section'] is not None]
```

.setup/bin/populate_office_hours_queue.py
```diff
@@ -0,0 +1,191 @@
...
+DB_PASS = "submitty_dbuser"
+
+def main():
+    # get current semester
+    term_id = dateutils.get_current_semester()
+
+    # set up the database connection
+    database = "submitty_" + term_id + "_sample"
+    engine = create_engine("postgresql:///{}?host={}&port={}&user={}&password={}"
+                           .format(database, DB_HOST, DB_PORT, DB_USER, DB_PASS))
```
williamjallen, line 22: Might be better to use an f-string here.

.setup/bin/populate_office_hours_queue.py
```diff
@@ -0,0 +1,191 @@
...
+    # Find all ids of students and graders
+    all_student_ids = [x['user_id'] for x in all_users if x['user_group'] == 4 and x['registration_section'] != None]
+    all_grader_ids = [x['user_id'] for x in all_users if x['user_group'] < 4]
+
+    # create a lookup table from user id to name
+    name_lookup = dict()
+    for x in all_users:
+        user_name = ""
+
+        if x["user_preferred_givenname"] != None:
```
williamjallen, line 55:
```suggestion
        if x["user_preferred_givenname"] is not None:
```

.setup/bin/populate_office_hours_queue.py
```diff
@@ -0,0 +1,191 @@
...
+            user_name += x["user_preferred_givenname"]
+        else:
+            user_name += x["user_givenname"]
+        
+        user_name += " "
+
+        if x["user_preferred_familyname"] != None:
```
williamjallen, line 62:
```suggestion
        if x["user_preferred_familyname"] is not None:
```

.setup/bin/populate_office_hours_queue.py
```diff
@@ -0,0 +1,191 @@
...
+            queue_entry["removed_by"] = random.choice(all_grader_ids)
+
+        # contact information
+        if queues_lookup[queue_entry["queue_code"]]["contact_information"]:
+            queue_entry["contact_info"] = queue_entry["user_id"] + "@sample.com"
+        else:
+            queue_entry["contact_info"] = None
+
+        res = conn.execute("SELECT max(time_in) FROM queue WHERE user_id = '{}' AND UPPER(TRIM(queue_code)) = UPPER(TRIM('{}')) AND (removal_type IN ('helped', 'self_helped') OR help_started_by IS NOT NULL)".format(queue_entry["user_id"], queue_entry["queue_code"]))
+        queue_entry["last_time_in_queue"] = res.fetchall()[0][0]
```
williamjallen, line 161: It would be good to format this into a more readable state.

.setup/bin/populate_office_hours_queue.py
```diff
@@ -0,0 +1,191 @@
...
+        # people usually won't be pausing
+        if random.random() < 0.8:
+            queue_entry["paused"] = False
+            queue_entry["time_paused"] = 0
+            queue_entry["time_paused_start"] = None
+        # this user is pausing or has paused
+        else:
+            # currently pausing
+            queue_entry["paused"] = random.choice([True, False])
+            queue_entry["time_paused"] = random.choice([0, random.randint(0, time_in)])
```
williamjallen, line 179: Why do we need the `random.choice()` here?  Do we want there to be a 50% chance that the time paused is 0?  Just to be clear, I'm just wondering about the logic here, not necessarily requesting changes.

williamjallen (changes requested):

.setup/bin/populate_office_hours_queue.py
```diff
@@ -0,0 +1,191 @@
...
+            # student is being emptied/removed out of the queue
+            queue_entry["removed_by"] = random.choice(all_grader_ids)
+
+        # contact information
+        if queues_lookup[queue_entry["queue_code"]]["contact_information"]:
+            queue_entry["contact_info"] = queue_entry["user_id"] + "@sample.com"
+        else:
+            queue_entry["contact_info"] = None
+
+        res = conn.execute("SELECT max(time_in) FROM queue WHERE user_id = '{}' AND UPPER(TRIM(queue_code)) = UPPER(TRIM('{}')) AND (removal_type IN ('helped', 'self_helped') OR help_started_by IS NOT NULL)".format(queue_entry["user_id"], queue_entry["queue_code"]))
```
tkoz0, line 160: Maybe use an f-string here

tkoz0 (changes requested):
I left a comment about using an f-string. As stated in the PR description, I would suggest adding a few options such as how many rows to add to the table and which course to add to. I'm also wondering if seeding the random number generator should be considered such as for the sample courses where having the same data every time is desirable for testing.

williamjallen:
> I left a comment about using an f-string. As stated in the PR description, I would suggest adding a few options such as how many rows to add to the table and which course to add to. I'm also wondering if seeding the random number generator should be considered such as for the sample courses where having the same data every time is desirable for testing.

This should definitely use a seeded random number generator.  That way, results are reproducible between runs.  Additionally, we could run this script as part of the setup sample courses script and run tests against the output if the results are reproducible.

.setup/bin/populate_office_hours_queue.py
shinyclovers: I believe I added the `random.choice()` here because I think it's more likely that a user haven't paused before and having it as `random.randint(0, time_in)` would give less "realistic" data. (However, based on testing I think the pausing time is not working as I intend so code around this area might be changed)

.setup/bin/populate_office_hours_queue.py
shinyclovers: This script is intended to be run inside the VM. I just hardcoded the variables from setup_sample_course as I was trying to have a semi-functional prototype while I coded it up, I will add a note at the top of the script saying its usage. I will also grab the information from the file you suggested.

bmcutler:
@williamjallen 
This is not intended to be used in setup sample courses because we generally want queue data to be very recent (within the last hour) rather than when we last rebuilt the machine.

jenaybarela (approved):
Works as expected.

PR #8755, opened 2023-01-22
[Bugfix:Developer] increase vagrant boot timeout

bmcutler:
increase vagrant boot timeout to 10 minutes

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -79,12 +79,13 @@ end
 Vagrant.configure(2) do |config|
   mount_options = []
 
+  # The time in seconds that Vagrant will wait for the machine to boot and be accessible. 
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -84,6 +84,8 @@ Vagrant.configure(2) do |config|
   # that one) as well as making sure all non-primary ones have "autostart: false" set
   # so that when we do "vagrant up", it doesn't spin up those machines.
 
+  config.vm.boot_timeout = 600
```
shailpatels, line 87: I think it would be useful to leave a comment saying this is in seconds and its the time to wait till the VM is up and reachable

shailpatels (approved):

PR #8760, opened 2023-01-24
[Feature:Developer] CodeMirror6 Package Install

Eceptonsu:
Submitty does not support CodeMirror6.

### What is the new behavior?
Add CodeMirror6 modules that are compatible with current CodeMirror5 into Submitty.


shailpatels (commented):
I think this looks good overall, is the plan to use codemirror6 in a different PR? Also is copying the files over enough - do we need any additional build steps like bundling them together to be able to be used in a different file?

Eceptonsu:
> I think this looks good overall, is the plan to use codemirror6 in a different PR? Also is copying the files over enough - do we need any additional build steps like bundling them together to be able to be used in a different file?

We dont need any additional build steps. I think CodeMirror6 split everything into different modules, and we can just copy over the required files :)

shailpatels (approved):

PR #8780, opened 2023-01-28
[Bugfix:System] Disabled audio, usb, uart on vagrant vb

Yujjio:

Nothing inside the `vagrant.configure`

Windows users may timeout at the `SSH auth method: private key` part when installing Submitty VM.

Related links:
* https://github.com/hashicorp/vagrant/issues/11777
* https://bugs.launchpad.net/cloud-images/+bug/1829625
* https://bugs.launchpad.net/cloud-images/+bug/1874453

### What is the new behavior?
Disabled unnecessary VM functions, including audio, USB, and UART1-4.

### Other information?
Tested in 2 Windows laptops and no timeout happened


bmcutler:
@Yujjio Please read
https://submitty.org/developer/getting_started/how_to_contribute

I believe this PR would be better titled [Bugfix:System]
(DevDependency is for 3rd party library updates)

Yujjio:
> @Yujjio Please read https://submitty.org/developer/getting_started/how_to_contribute
> 
> I believe this PR would be better titled [Bugfix:System] (DevDependency is for 3rd party library updates)

Got it! Changed the PR title

Vagrantfile
```diff
@@ -77,6 +77,17 @@ def mount_folders(config, mount_options)
 end
 
 Vagrant.configure(2) do |config|
+  config.vm.boot_timeout=600
+  config.vm.provider :virtualbox do |v|
+    v.customize ["modifyvm", :id, "--uart1", "0x3F8", "4"]
+    v.customize ["modifyvm", :id, "--uartmode1", "file", File::NULL]
+    v.customize ["modifyvm", :id, "--audio", "none"]
```
d0ngeun, line 84: From my tests, these modifications are not really necessary. The first two lines with uart1 and uartmode1 seem to fix the freeze on their own.

d0ngeun (commented):
The first 2 modifications with uart1 and uartmode1 are the only necessary ones needed to fix the freeze bug. I suggest removing the other vm customizations.

Vagrantfile
```diff
@@ -77,6 +77,17 @@ def mount_folders(config, mount_options)
 end
 
 Vagrant.configure(2) do |config|
+  config.vm.boot_timeout=600
```
MasterOdin, line 80: This is already done on line 94 and can be removed.

Vagrantfile
```diff
@@ -77,6 +77,17 @@ def mount_folders(config, mount_options)
...
+  config.vm.provider :virtualbox do |v|
```
MasterOdin, line 81: We already have a `config.vm.provider` block and I'd suggest merging the two to have just one block.

Yujjio: Yes I just saw that and I will merge it. Thanks

Yujjio:
> The first 2 modifications with uart1 and uartmode1 are the only necessary ones needed to fix the freeze bug. I suggest removing the other vm customizations.

I tried to only add the first 2 lines at first but it still will timeout from time to time, by adding the other modification, I never meet timeout again so I think these lines are also helpful

Vagrantfile
Yujjio: which version's line 94? I didn't see that in the main branch or in my branch

MasterOdin: Sorry, meant line 84:

https://github.com/Submitty/Submitty/blob/4a6309822740858cab5d1052768fd33815ea0aec/Vagrantfile#L84

Yujjio: OK, I see. Will modify it soon. Thanks

d0ngeun (approved):

MasterOdin (approved):

MasterOdin:
Thanks for the PR to improve the situation on vagrant!

PR #8790, opened 2023-01-29
[Bugfix:System] Fix vagrant SSH freeze

d0ngeun:

There was a bug where if you were trying to install Submitty it would freeze at 'SSH auth method: private key' and time out. This was commonly seen on Windows devices but also could have stemmed from a problem with the Ubuntu box. 
### What is the new behavior?
After modifying the vagrant file and adding some instructions regarding the serial ports, the installation was able to continue. 
### Other information?

Further bug and solution descriptions can be found here:
https://bugs.launchpad.net/cloud-images/+bug/1829625
https://github.com/hashicorp/vagrant/issues/11890

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -82,6 +82,11 @@ Vagrant.configure(2) do |config|
   # The time in seconds that Vagrant will wait for the machine to boot and be accessible. 
   config.vm.boot_timeout = 600
 
+  config.vm.provider :virtualbox do |v|
+    v.customize ["modifyvm", :id, "--uart1", "0x3F8", "4"]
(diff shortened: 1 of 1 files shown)


bmcutler:
I believe this PR is similar to PR #8780

d0ngeun:
yeah I did see afterwards that Ray solved the same problem earlier, but I was unsure what to do during conflicting pr scenarios. Is there a way to merge them? Will make sure to check github before next commit. Also, the difference would be that my pr solves the freeze without disabling audio or USB features, those extra lines weren't necessary. @bmcutler

bmcutler:
@d0ngeun
Then please make those comments on the other PR as a review and close your PR.  Thanks.

PR #8800, opened 2023-01-31
[Refactor:Submission] Gradeables query improvement

DefinitiveAbove:
### What is the new behavior?
An index on gradeable_component_mark_data added for gd_id, avoiding a sequential scan of the table and reducing the average query cost directly related to that table from ~3 ms to ~0.12 ms (getGradedGradeablesUserOrTeam).

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -2428,6 +2428,13 @@ CREATE INDEX gradeable_component_mark_data_gcm_id_idx ON public.gradeable_compon
 
 
 --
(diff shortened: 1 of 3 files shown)


williamjallen (approved):
Looks good to me!  This looks like it might have a significant impact on performance.  On my developer VM, Postgres shows estimated values of `cost=306.81..2577.00` before and `cost=207.68..1387.44` after the index is added.

I didn't see a major improvement in the real-world running time for the gradeables page as a whole, but this likely has a greater effect on production where there is more data present.

PR #8811, opened 2023-02-02
[Bugfix:Forum] Add required attribute to textareas

### LINKED ISSUE #8696: Discussion forum makes request when body of post is empty
**Describe the bug**
If the new post text box is empty, a request is still made.

**Expected behavior**
Client-side validation should tell me it was empty without making a full network request.

**To Reproduce**
1. Go to forum post
2. Hit Submit Reply to All

**Configuration**
 - OS: [e.g. Windows, Linux, ...] MacOS
 - Browser [e.g. chrome, safari] Chrome

**Screenshots**

**Additional context**
It doesn't let us post but it's a waste of a network request.

PythonCoderAS:

Currently, whenever I try to post an empty post, Submitty makes a network request to the server which rejects the post due to the lack of text.

Fixes #8696

### What is the new behavior?

Textareas are required so some text has to be in them. This means that checking the post is not empty is done client side and saves a full network request.

### DIFF
--- a/site/public/templates/misc/MarkdownArea.twig
+++ b/site/public/templates/misc/MarkdownArea.twig
@@ -80,6 +80,9 @@
             rows="10" cols="30"
             {# 524288 characters is the default max length for text input fields #}
             maxlength="{{ (no_maxlength|default(null) ? '524288' : (textarea_maxlength|default(null) ? textarea_maxlength :'5000'))}}"
+            {% if required %}
+                required
+            {% endif %}
             {% if textarea_onchange is defined %}
                 onchange="{{ textarea_onchange }}"
             {% endif %}
--- a/site/public/templates/misc/MarkdownArea.twig
+++ b/site/public/templates/misc/MarkdownArea.twig
@@ -80,7 +80,7 @@
             rows="10" cols="30"
             {# 524288 characters is the default max length for text input fields #}
             maxlength="{{ (no_maxlength|default(null) ? '524288' : (textarea_maxlength|default(null) ? textarea_maxlength :'5000'))}}"
-            {% if required %}
+            {% if required is defined and required %}
                 required
             {% endif %}
             {% if textarea_onchange is defined %}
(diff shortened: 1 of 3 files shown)


williamjallen:
Please fill out the PR description a little bit more, even if it's a brief description of the problem and how this PR fixes it.

PythonCoderAS:
@williamjallen I have updated the description

williamjallen:
The e2e tests are failing because an error page was generated.  I suspect that the `required` parameter is not being passed to the `MarkdownArea.twig` template in some cases.

PythonCoderAS:
I am not knowledgeable about how Twig operates. How do I make something akin to a boolean parameter with a default of `False`?

williamjallen (approved):

site/public/templates/misc/MarkdownArea.twig
```diff
@@ -80,6 +80,9 @@
             rows="10" cols="30"
             {# 524288 characters is the default max length for text input fields #}
             maxlength="{{ (no_maxlength|default(null) ? '524288' : (textarea_maxlength|default(null) ? textarea_maxlength :'5000'))}}"
+            {% if required %}
```
MasterOdin, line 83: Could we have:

```
            {% if required is defined and required %}
```

here as well? I get that twig.js doesn't have the same requirement, but would be good to keep them the same as much as possible.

MasterOdin (changes requested):

site/public/templates/misc/MarkdownArea.twig
PythonCoderAS: I forgot that the two templates have to be kept in sync. I still don't understand why a symlink couldn't be utilized though.

MasterOdin: Windows has had issues with symlinks for a long time. I think there's been some progress made in the last few years that have really improved things, but looking it up, it does require activating a developer setting in Windows, as well as that some clients may still not work quite right, though I'm not sure.

Easier to avoid the very few symlinks we'd want than field potential issues on getting new developers setup in my opinion.

MasterOdin (approved):

site/public/templates/misc/MarkdownArea.twig
PythonCoderAS: I think a decent alternative would be a script that automatically copies the newer of the two to the same location, and maybe a GitHub Action that ensures the two items have the same hash.

PR #8828, opened 2023-02-03
[Documentation:System] Updated README_utm_arm.txt

EdenBendheim:

Outdated instructions for installing Submitty on ARM based MacBooks.

### What is the new behavior?
Just updates the instructions to what worked for spring 2023.

### Other information?
Still needs to be tested for M2 MacBook, and if the instructions are different, we should probably split this up into two files.


bmcutler (approved):

LittleSilver33 (commented):
Good changes to the documentation. I ran through the install and didn't run into any issues that weren't talked about, which is definitely a good thing. It is important to note that I tested this on my M2 so it should work for the M1 or M2, and hopefully future apple silicon chips. I didn't find any bugs.

PR #8830, opened 2023-02-04
[Feature:Submission] Added a Warning text for bulk upload

### LINKED ISSUE #8820: Warning text for bulk upload gradeable that files will not be released to students
**Describe the bug**
When the instructor makes a bulk upload / scanned gradeable, these are the default settings on the submissions/autograding tab:

Screen Shot 2023-02-02 at 10 14 42 PM

I would like to keep these settings, but add a warning message next to the "view and download submitted files" option.
If "No" is selected, let's display warning text in bold red font that says "WARNING: Students will see the grading score and TA feedback but will not be able to download the scanned and uploaded file."  (or similar) text.

related to #8437

Suryansh5545:

Fixes #8820
Screenshot 2023-02-04 211301
 currently no warning message is shown when "View and download submitted files? "NO" " is selected

### What is the new behavior?
A warning message is shown the "NO" checkbox is selected.


shailpatels:
Hi @Suryansh5545 Could you add an image to the UI changes you made for reviewers and link to an issue this PR is supposed to fix

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -21,7 +21,7 @@
 
     <div id="student_download_view">
         <div>
-            <fieldset><legend>View and download submitted files?</legend>
+            <fieldset><legend>View and download submitted files? <span class="warning-text" id="warning"> WARNING: Students will see the grading score and TA feedback but will not be able to download the scanned and uploaded file.</span></legend>
```
shailpatels, line 24: Let's change the class here to use `red-message` which is already defined in `site/public/css/server.css`, this would make the message look like how the other warning/error messages look like - it would also be needed to meet our accessibility standards for the amount of contrast required.

image

shailpatels, line 24: I think the id `warning` should be changed to something more descriptive, like maybe `view-files-warning`

site/public/css/admin-gradeable.css
```diff
@@ -136,6 +136,10 @@ input[type="checkbox"] {
     padding-left: 20px;
 }
 
+.warning-text {
```
shailpatels, line 139: Once you switch to the `red-message` class, this can be removed

shailpatels (changes requested):

shailpatels (changes requested):
A couple of general comments

It looks like this warning is shown for all types of gradeables, based on the issue description I think you want to only show it for bulk upload type, since normal electronic gradeable's don't have a concept of scanned/uploaded files

Also there appears to be a bug if the "yes" radio button is already checked, then if you select "no" the warning message still appears, then when you check "yes" afterwards, the message disappears which is the wrong state. This should show the message when "yes" is selected either by default on page load or when the user selects it

Suryansh5545:
sorry for this issue, radio button bug is fixed and warning tag replaced by the suggested one.

shailpatels (changes requested):
I beleive the issue of the non bulk gradeables showing the warning is fixed, but it looks like the issue of an existing bulk upload gradeable has the warning flipped still if the user selected yes originally

Suryansh5545:
@shailpatels i am confused a bit, so the warning text should show when the "Yes" Radio button is selected? the #8820 issue states to show it when the "NO" checkbox is selected?

shailpatels:
@Suryansh5545 ah sorry about that, I didn't see it in the issue and misread the negative. You are correct, disregard my last comment and I'll do a final review later today but other then that this should be close to done!

shailpatels (approved):
Changes looks good and implement the feature as described in the issue

Thanks for the contribution!

PR #8832, opened 2023-02-04
[Feature:Submission] Fixed Thread Field not being mandatory

### LINKED ISSUE #8816: Omitting thread id errors & clears form
**Describe the bug**
When an instructor fillis out the "New Gradeable" form, if they say "Yes" to "Is there a discussion forum component... " but they do not fill in the forum thread id, submission of the form will fail, and the form boxes are cleared -- losing the users work/effort.

Note: the forum must be enabled for this course.

Screen Shot 2023-02-02 at 4 21 34 PM

Screen Shot 2023-02-02 at 4 21 41 PM

**Expected behavior**
Instead, let's properly treat this form box as a required field and don't allow submission, don't clear & lose the data.

NOTE:  Thoroughly test this form and make sure all other fields are correctly checked for mandatory values, etc. to prevent data loss in attempting to submit with any incomplete form state.

Suryansh5545:

Fixes #8816 The form to create new gradeable with Forum Discussion on was submit able without a thread id.

### What is the new behavior?
The form to create new gradeable with Forum Discussion on is no longer submit able without a tread id and the thread id field is hidden if the requirements are not fulfilled. 
Screenshot 2023-02-04 211036

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -125,6 +125,20 @@ Vagrant.configure(2) do |config|
     # vb.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
     vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
 
+    # Sometimes Vagrant will error on trying to SSH into the machine when it starts, due to a bug in how
+    # Ubuntu 20.04 and later setup the virtual serial port, where the machine takes minutes to start, plus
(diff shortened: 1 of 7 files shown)


shailpatels (approved):
Changes look good and fixes the issue, thanks for the contribution!

Issue #8388, opened 2022-09-12
Notebook - instructor cannot submit as another student


bmcutler:
**Describe the bug**

The instructor view of a notebook gradeable includes the radio button option to submit as student:

<img width="415" alt="Screen Shot 2022-09-12 at 3 54 46 PM" src="https://user-images.githubusercontent.com/5871417/189743854-b7f79d6c-1531-4d0c-9319-d0ae0d43fd00.png">

But this option does not work.  (It was only implemented to work for regular file upload submissions.
It doesn't even have a box to type in a student name!

**Expected behavior**

First, let's remove this option from the webpage for all notebook gradeables, it's not implemented and it's confusing.

Second, let's actually implement this functionality.  It would be useful for an instructor giving an in-class electronic quiz.  Most of the students successfully submit using their laptop, but if one student doesn't have a working laptop, they can submit on paper and the instructor or TA can type up and submit their answers afterwards.

NOTE:  Let's make sure it's not easy to accidentally submit the form as a student.  For both notebook and regular file uploads, let's have a popup "Are you sure you want to submit as student smithj?" after clicking the "Submit" button.  Maybe also there should be big colored box with notices at the top & bottom of the page when the 'submit as student' radio button is selected.

kevinochoa8266:
This is my first open-source contribution. Can I participate?

williamjallen:
Hi @kevinochoa8266!  Submitty welcomes contributions from anyone.  If you'd like to tackle this issue, go right ahead!  Once you make a pull request, it will undergo a review, and hopefully be merged once any requested changes have been addressed.  If you need help with anything, feel free to ask in our [Slack](https://submitty.org/index/contact).

ratika-12:
Can I work on this ? @bmcutler

PR #8834, opened 2023-02-04
[UI/UX:InstructorUI] Added confirm box to submit for student

jason490:
The instructor would click the submit button for a student. This might cause issues to the instructor submitting as a wrong student.
Linking issue #8388 as this solves one of the issues stated in there.

### What is the new behavior?
Added a confirm box that would show up after the submit button is clicked for only submitting as a student. Clicking cancel will prevent the submission from going through.
comfirm_box_UI


shailpatels (changes requested):
I think this looks pretty good, just 1 small tweak

After the instructor uploads a file and hits cancel on the confirmation box, the green 'submit' box is still disabled. I think to fully reset the state to before we should re-enable that button to indicate the page is ready to submit again 

- it would also prevent awkwardly removing the file(s) and re-adding them to upload again if an instructor accidentally hits cancel

jason490:
I added the changes you requested for.
Is there any other minor changes you want me to make?

shailpatels (approved):
Looks good, changes fix the issue, and work as described.

Thanks for the first PR!

PR #8837, opened 2023-02-05
[UI/UX:InstructorUI] Adds exclude section in rotating section

### LINKED ISSUE #8429: Updating existing rotating sections, exclude specific registration sections
When rotating sections are creating initially and students are assigned to rotating sections, instructor has the option to exclude section (e.g., they can include a 'staff' registration section).

But if students add after that, we only have the option to all all non-NULL registration section students to rotating section.  We cannot exclude the staff section.

jason490:

Closes #8429 
The radio button that puts all the NULL rotating sections users to rotating sections and all the NULL registration sections to NULL rotating sections 

### What is the new behavior?
The radio button now has a pop down with the option to exclude certain sections from placing NULL rotating sections users to rotating sections
fewest_exclude_sections


site/app/templates/admin/users/RotatingSectionsForm.twig
```diff
@@ -9,12 +9,18 @@
             else {
 				$('#redo-data').show();
             }
+            if (val != "fewest"){
```
shailpatels, line 12: You would want a `!==` here for strict inequality
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality

Normally the JS linter would catch this but it can't work on JS embedded inside html/twig pages

site/public/css/rotatingsectionsform.css
```diff
@@ -7,6 +7,17 @@ fieldset > * {
     margin-bottom: 0.5em;
 }
 
+#fewest-exclude {
+    margin-left: 13px;
+    border-left: solid #cccccc 1px;
```
shailpatels, line 12: Instead of a specific color value here I think you'd want to use the CSS `var` functionality and use one of the already defined colors in `site/public/css/colors.css `

shailpatels (changes requested):

jason490:
I added the changes you requested for. Do you see any other issues?

hansongu123 (approved):
Routine looks good! Although doing the second radio button after doing third radio button crashes the site. (existing bug in the main build)

jason490:
I found the bug you were talking about.
The issue is when there is no one in the rotating sections the curr_section_sizes would contain NULL elements, which would break the array_map, since its subtracting using a NULL element

hansongu123:
so, you are able to reproduce this on your end as well. Would you want to make a new issue for this and solve it? Maybe either by allowing it to function or putting up a warning before this happens.

jason490:
I was thinking of either disabling the button if no one is in the rotating section or filling the curr_section_sizes with the sections

hansongu123:
Disabling the button could also work. One thing is, when you do a refresh or back up on the crashed site after this, you'd see the button actually did the job(?) - people in the NULL rotating sections are put into the sections again with the options toggled.

shailpatels (commented):
It looks like the list of normal registration sections is being used for the selection to exclude? Is this intentional, I'm a bit confused on what happens if I select a selection that's not one of the 3 listed, it doesn't look like anything would happen

image

shailpatels:
@jason490 I wasn't sure if you created an issue or not mentioned in the discussion above but I made one here 
https://github.com/Submitty/Submitty/issues/8900

Feel free to close if its a duplicate though or update if you have additional details

hansongu123:
@shailpatels The list of sections on the drop-down window showing on the right are the "registration sections", the ones showing on the left are "rotating sections". Students are added to the registration sections when they are first added to the course - this can be set manually by an instructor, whereas the rotating sections are random, rotating teams... and can be set to a different number than that of the registration sections.

shailpatels:
@hansongu123 Ah ok that makes sense, I was thinking the operation would avoiding specific rotating section but it re-reading the issue I understand its operating from the pool of existing normal sections

shailpatels (approved):

PR #8845, opened 2023-02-06
[Feature:InstructorUI] Add ability to delete docker imgs in UI

### LINKED ISSUE #8524: Docker GUI -- remove/cleanup old docker image
The UI for instructors to add docker images to specific worker machines works great.
But over time, the worker machines will clutter with docker images that were incomplete (through testing instructor realized something was missing and they made & installed a new image) and from old assignments / prior terms

Give instructor ability to remove & uninstall docker images that are no longer necessary.  
instructor should only be able to remove images that they added (to prevent mistakes uninstalling other instructors images).
super users can remove any docker image.

Suryansh5545:
Fixes #8524. Currently only adding a Docker container is possible through the GUI. If the instructor needs to delete an image, it is possible only through the command line.

### What is the new behavior?
Users can now use the "Remove" button to delete an unnecessary Docker container. The Docker UI Cypress test has been expanded to test this functionality.

image


site/public/js/docker_interface.js
```diff
@@ -64,6 +67,53 @@ function addFieldOnChange() {
     }
 }
 
+// eslint-disable-next-line no-unused-vars
```
williamjallen, line 70: It would be better to put `exported removedTicked` at the top of the file instead of putting inline ignores everywhere as per [this](https://submitty.org/developer/software_and_system_design/coding_style_guide/javascript#style-rules) documentation page.  Apply this throughout the rest of these changes.

williamjallen (changes requested):

site/phpstan-baseline.neon
```diff
@@ -33,7 +33,7 @@ parameters:
 
 		-
 			message: "#^Call to function is_null\\(\\) with app\\\\models\\\\User will always evaluate to false\\.$#"
-			count: 3
+			count: 4
```
williamjallen, line 36: We shouldn't be adding more ignored errors unless strictly necessary.  The new error should be fixed rather than ignored.  I suspect the issue is coming from line 162 of [DockerInterfaceController.php](https://github.com/Submitty/Submitty/pull/8845/files#diff-bfd0eb036127c1252522af67074ab2cfa90757bd093a6bf5947cdca7235e6060).

williamjallen (changes requested):

Suryansh5545:
I will get it fixed. Also i looked into the ways of fulfilling the permission requirement for this delete action and the only way to do it seems to be to add the instructor or user info inside the autograding_containers.json file.  
With this kind of structure

{
    "instructor": {
        "cpp": [
            "submitty/docker-test-saml-idp:latest",
            "submitty/java:11"
        ]
    }
}

This will require changes in DockerView and Docker.twig file which are working now. But after this implementation the update function stops working. i suspect it is due to the update_and_install_workers.py not being able to get the image tag due to change in the structure of the json. so that will also be changed.

So is this implementation good or shall i think about another way of doing it?

site/phpstan-baseline.neon
Suryansh5545: Reviewing this it seems that there is no need of is_null check for the removeImage function since it is being already done by the showDockerInterface function.

site/public/js/docker_interface.js
Suryansh5545: Fixed

Suryansh5545:
@williamjallen I have successfully implemented the feature to remove Docker images when the 'Update Docker and Machine' button is pressed. Additionally, I have incorporated a check for the user ID so that only the instructor who added the image can remove it. However, I am currently encountering some difficulties with the superuser portion. If the implementation looks good, I will proceed to create the test cases.

shailpatels (changes requested):
Thanks for the work so far!

I think the front end is a pretty good idea of what needs to be done but can be simplified . Instead of having a radio button on each table row, you could add a column that contains a button to delete the container, or an icon to represent deleting them. I think we should also add a confirmation to make sure the user knows what deleting a container means.

This would save the user a click and make the implementation more simple, it would also prevent issues were the instructor selected the wrong container and weren't aware of it 

Here's an example in the office hours queue table that has a column filled with buttons to remove students from it

image

And here's the course landing page with the list of gradeables, as an instructor they have a trash can symbol on the left to delete each one 

image



As for the additional information I'd have to look into more on seeing where that file is also used, an alternative would be to create a separate file or maybe we can create a database entry to track users to images they've uploaded

Suryansh5545:
@shailpatels "Thanks for the suggestion. the database integration is far more easy and better compared to the previous  JSON implementation. I've made the necessary changes to the user interface as well. Now, whenever a user or instructor adds an image, the JSON is updated and a corresponding entry is made in the database. If the image is removed, the corresponding database entry is also deleted. Currently, I'm considering the best approach for preloaded images, such as whether they should be assigned to a user or left unassigned. This would determine whether they can be deleted or not. Additionally, I'd like to find a solution to the mismatch in the migration process that's causing problems with the db_check. Can you help me on this issue?
DockerGIF

shailpatels:
@Suryansh5545 yeah thats a good point to consider, I'm thinking we can treat preloaded images as special, for example a system administrator with server privileges would have permission to install them, so they should be the only person that would be able to uninstall, so a user through the web UI should not be able to delete those.

@bmcutler Before work continues could we get some buy-in on storing which user uploaded which image on the database, I think this would streamline permissions handling but just wanted to make sure you didn't want that information to be also in the JSON file. 

As far as I know the worker machines are the only other thing that need the specific format of the containers json file? But still would be nice to avoid changing its structure

Suryansh5545:
@shailpatels I have made some updates to the code in response to issue #8524. Superusers can now remove any Docker image, even those that are not listed in the database. Additionally, I have fixed the DB migration.

I believe that all of the required changes have been made, and the pull request is now ready to be reviewed. Please let me know if there is anything else that needs to be addressed.

site/app/templates/admin/Docker.twig
```diff
@@ -133,10 +134,16 @@
                         <span class="badge badge-{{ capability_to_color[capability] }}">{{capability}}</span>
                     {% endfor %}
                 </td>
+                {% for capability in value %}
+                <td>
+                <button class="send-button btn btn-primary" id="{{ key }}" onclick=confirmationdialog("{{admin_url}}/remove_image",id,'{{capability}}')>Remove</button>
```
shailpatels, line 139: If you remove the class `send-button` here, the button will then be centered 

image

Also I think the `btn-primary`  should be `btn-danger`  or `btn-default` (personally I think `btn-default` is better because its a button most users won't use so better to keep it out of attention) 

Here's some rational behind the different colors
https://submitty.org/developer/software_and_system_design/interface_design_style_guide

site/public/js/docker_interface.js
```diff
@@ -64,6 +66,40 @@ function addFieldOnChange() {
...
 }
 
+function confirmationdialog(url,id,capability) {
+    if (confirm(`Are you sure you want to remove ${id} image?`)) {
+        removeImage(url,id,capability);
+    }
+}
+
+function removeImage(url,id,capability) {
+    $.ajax({
```
shailpatels, line 76: You could the `fetch` api which is a little bit easier to use especially since we're not using any ajax/xhr specific features 

```js
const response = await fetch(url, {
   'method' : 'POST',
   'body' : JSON.stringify({
        'image' : id,
        'capability' : capability,
        'csrf_token' : csrfToken
    }),
    headers: {
        'Content-Type': 'application/json'
    },
});

response.json()
.then((json) => { //handle response })
.catch((err) => { //handle fetch fail });
 
```

Or something like that, see the docs here

site/app/controllers/DockerInterfaceController.php
```diff
@@ -153,6 +155,39 @@ public function addImage() {
...
+        $user_id = $this->core->getUser()->getId();
+        $jsonFilePath = FileUtils::joinPaths(
+            $this->core->getConfig()->getSubmittyInstallPath(),
+            "config",
+            "autograding_containers.json"
+        );
+        $json = json_decode(file_get_contents($jsonFilePath), true);
+        $key = array_search($_POST['image'], $json[$_POST['capability']]);
+        $Verify = $this->core->getQueries()->removeDockerImageOwner($_POST['image'], $user_id);
+        if ($key !== false && $Verify != false) {
```
shailpatels, line 173: I think we can check if the key is false before we attempt to perform the query, if the container is not listed in the json file then I don't the query would succeed either so we can avoid trying it

site/app/controllers/DockerInterfaceController.php
```diff
@@ -153,6 +155,39 @@ public function addImage() {
...
+    public function removeImage() {
+        $user = $this->core->getUser();
+        $user_id = $this->core->getUser()->getId();
+        $jsonFilePath = FileUtils::joinPaths(
+            $this->core->getConfig()->getSubmittyInstallPath(),
+            "config",
+            "autograding_containers.json"
+        );
+        $json = json_decode(file_get_contents($jsonFilePath), true);
+        $key = array_search($_POST['image'], $json[$_POST['capability']]);
```
shailpatels, line 171: The `$_POST['capability']` and `$_POST['capability']` should be tested if they're actually set and is a string. Otherwise a user could use a client like postman and spoof the POST body with null / bad data.

We could also do some validation by checking if the image is a valid name, see some restrictions of what a valid name is here https://docs.docker.com/engine/reference/commandline/tag/#description

but this isn't super necessary, checking its some sort of string I think is good for this first implementation

shailpatels (changes requested):

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -57,6 +57,20 @@ def update_docker_images(user, host, worker, autograding_workers, autograding_co
...
+            # List all images
+            image_current_list = client.images.list()
+            container_list = images_str.split(', ')
+            container_list = set(container_list)
+            image_set = {image.attrs['RepoTags'][0] for image in image_current_list}
+            image_to_remove = set.difference(image_set, container_list)
+
+            # Remove images
+            for imagesRemoved in image_to_remove:
+                client.images.remove(imagesRemoved)
```
shailpatels, line 70: Yeah I think this makes sense just cleaning up whats not listed on the file the next time we do an update, as long as the user doesn't see them listed

shailpatels (dismissed):
Looks like you accidentally checked in a top-level package-lock.json that should be deleted

shailpatels (commented):
Thanks for the work so far! I think this is almost there to approve - definitely a challenging issue as an external contributor to tackle!

site/public/js/docker_interface.js
Suryansh5545: @shailpatels I'm having trouble with implementing the fetch feature as I'm receiving a 302 response. I've attempted to modify the CORS mode and the JSONResponse backend, but I haven't been able to resolve the issue. Can you please offer any suggestions or guidance on how to resolve this problem?

Suryansh5545:
@shailpatels I have added a cypress test to test the functionality and i think the PR is ready to be reviewed/merged now. if there is anything more required please let me know

Suryansh5545:
@shailpatels the PR is ready except the fetch api, i tried implementing it but it doesn't return anything no errors eithers.

DarthNyan (approved):

site/public/js/docker_interface.js
```diff
@@ -64,6 +64,40 @@ function addFieldOnChange() {
...
+function confirmationdialog(url, id) {
+    if (confirm(`Are you sure you want to remove ${id} image?`)) {
+        removeImage(url, id);
+    }
+}
+
+function removeImage(url, id) {
+    $.ajax({
+        url: url,
+        type: 'POST',
```
williamjallen, line 76: You could use `$.post()` instead.

site/app/templates/admin/Docker.twig
```diff
@@ -142,10 +145,16 @@
                         <span class="badge badge-{{ capability_to_color[capability] % 8 + 1 }}">{{capability}}</span>
                     {% endfor %}
                 </td>
+                <td>
+                    {% if (docker_image_owners[key] is defined and user_id == docker_image_owners[key]) or superuser %}
+                    <button class="send-button btn btn-default" id="{{ key }}" data-image-id = "{{ key }}" onclick=confirmationdialog("{{admin_url}}/remove_image",id)>Remove</button>
```
williamjallen, line 150: Why is the admin_url variable required?

williamjallen (changes requested):

migration/migrator/migrations/master/20240219141515_docker_image.py
```diff
@@ -0,0 +1,44 @@
...
+def down(config, database):
+    """
+    Run down migration (rollback).
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    pass
```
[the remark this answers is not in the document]
DarthNyan: I don't think so, because if a new image is added between running up and down migrations, then dropping the table would lose who added the image.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -9368,4 +9368,56 @@ public function checkNonMappedUsers() {
...
+     */
+    public function setDockerImageOwner(string $image, string $user_id): void {
+        $current_owner = $this->getDockerImageOwner($image);
+        if ($current_owner === false) {
+            $this->submitty_db->query("INSERT INTO docker_image (image_name, user_id) values (?, ?);", [$image,$user_id]);
+        // If an instructor wants to add an image they didn't upload to a capability, the image will have no owner.
+        // Only sysadmin will be able to remove it.
+        }
+        elseif ($current_owner !== $user_id) {
+            $this->submitty_db->query("UPDATE docker_image SET user_id = '' WHERE image_name = ?", [$image]);
```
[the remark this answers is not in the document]
DarthNyan: If one instructor adds an image to a capability, and then a different instructor adds that image to a different capability, neither instructor should be able to remove that image from the system, only the sysadmin should.

site/app/templates/admin/Docker.twig
DarthNyan: I don't know if it's required per se, but this is how adding and updating images are handled in this file so I assume it was copied from there.

site/public/js/docker_interface.js
DarthNyan: I haven't seen $.post used in any of the code that I've worked on, and I ran a search and it seems to only be used twice in the entire codebase. Do you think this is a pressing change or can it wait for a later refactor that replaces a bunch of $.ajax requests with $.post?

williamjallen (approved):
All of my requests have been addressed, and the code looks reasonable to me.  I didn't test it and I'm not familiar with these parts of the codebase, so it would be good for someone else to take a look as well.

site/app/controllers/DockerInterfaceController.php
```diff
@@ -153,6 +159,48 @@ public function addImage() {
...
+                        $key = array_search($_POST['image'], $capability, true);
+                        if ($key !== false) {
+                            unset($json[$capability_key][$key]);
+                        }
+                    }
+                    file_put_contents($jsonFilePath, json_encode($json, JSON_PRETTY_PRINT));
+                    return JsonResponse::getSuccessResponse($_POST['image'] . ' removed from docker images!');
+                }
+                else {
+                    return JsonResponse::getFailResponse('This image is owned/managed by another instructur/superuser.');
```
cjreed121, line 191:
```suggestion
                    return JsonResponse::getFailResponse('This image is owned/managed by another instructor/superuser.');
```

site/app/libraries/Access.php
```diff
@@ -212,6 +212,10 @@ public function __construct(Core $core) {
         $this->permissions["forum.merge_thread"] = self::ALLOW_MIN_LIMITED_ACCESS_GRADER | self::CHECK_CSRF;
 
         $this->permissions["admin.wrapper"] = self::ALLOW_MIN_INSTRUCTOR;
+
+        //Docker UI permissions
+        $this->permissions["dockerui.add_image"] = self::CHECK_CSRF | self::ALLOW_MIN_INSTRUCTOR;
+        $this->permissions["dockerui.remove_image"] = self::CHECK_CSRF | self::ALLOW_MIN_INSTRUCTOR;
```
cjreed121, line 218: No need for this

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -9368,4 +9368,59 @@ public function checkNonMappedUsers() {
...
+        }
+        elseif ($current_owner !== $user_id) {
+            $this->submitty_db->query("UPDATE docker_images SET user_id = NULL WHERE image_name = ?", [$image]);
+        }
+    }
+
+    /**
+     * @param string $image the full name of the image to remove
+     * @param User $user the user who is removing the image
+     * @return bool true iff image was deleted
```
cjreed121, line 9415:
```suggestion
     * @return bool true if image was deleted
```

site/public/js/docker_interface.js
```diff
@@ -64,6 +64,40 @@ function addFieldOnChange() {
     }
 }
 
+function confirmationdialog(url, id) {
```
cjreed121, line 67:
```suggestion
function confirmationDialog(url, id) {
```
Do this for better readability. Also make sure to fix where it's called.

migration/migrator/migrations/master/20240219141515_docker_image.py
```diff
@@ -0,0 +1,45 @@
...
+    images = set()
+    for capability in container_data:
+        for image in container_data[capability]:
+            images.add(image)
+    existing_images = set()
+    for row in database.execute("SELECT image_name FROM docker_images").all():
+        existing_images.add(row[0])
+    for image in images:
+        if image not in existing_images:
+            database.session.execute("INSERT INTO docker_images (image_name, user_id) VALUES (:name, NULL);", {"name": image})
```
cjreed121, line 33: Thoughts on setting user_id to the submitty admin account rather than just null? This way we can enforce a non-null constraint.

cjreed121 (changes requested):

migration/migrator/migrations/master/20240219141515_docker_image.py
DarthNyan: Is there an admin user that is common across production and development environments? If so I can add that but I don't want to assign a foreign key column a default value that might not be present.

bmcutler (commented):
I tried to test this PR.  I'm on an M2 Mac so I think that impacts my testing.

I like the UI design, I think that's good.

It appears that the user can only delete the images they uploaded, which is good.
(but I haven't tested it extensively).

I ran into problems confirming that the images were updated correctly on the server.
It appears to have deleted all of the default/previously installed docker images.  When I run "docker image ls" it only shows the new images I added.  I thought maybe this was because most of our docker images do not have an ARM version that is required for M2.

There were errors similar to this displayed on the UI:
Screenshot 2024-05-23 at 1 14 42 PM

However, the new images I selected for install do not have an ARM version either so I'm just confused.

I don't have a non-ARM machine to test this on right now, so I can't approve.

williamjallen (commented):
@bmcutler 

> Do we need the docker containers for lichen and saml test? Should we add them to the autograding containers file so they get preserved? Or should we special case them?

The Lichen image appears to be preserved based on my local testing.  See line 65 [here](https://github.com/Submitty/Submitty/pull/8845/files#diff-75e24156073c99922058c5a4e1476249455b847af1f35395103800709c9a69c3) for the "special case" Tate implemented for Lichen at the last hackathon.  You make a good point about the other examples though.  I think it would be a good idea to add a config file somewhere which lists "system" images instead of embedding the images in the code itself.

> I get errors when I try to "Update dockers and machines"

I'm also unable to reproduce this.  

@bmcutler did you forget to pull before testing this PR perhaps?  I believe @DarthNyan recently fixed some of the issues you observed.

dorian451 (approved):
Seems to work

williamjallen (changes requested):
This still needs the list of images to preserve to be formalized somewhat as per the conversation above.

site/app/controllers/DockerInterfaceController.php
```diff
@@ -153,11 +152,45 @@ public function addImage() {
...
     }
 
-    /**
-     * @return JsonResponse
-     */
+    #[Route("/admin/remove_image", methods: ["POST"])]
+    public function removeImage(): JsonResponse {
+        $pattern = '/^[a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+\/[a-z0-9]+[a-z0-9._(__)-]*[a-z0-9]+:[a-zA-Z0-9][a-zA-Z0-9._-]{0,127}$/';
+        $user = $this->core->getUser();
+        $user_id = $user->getId();
```
ziesski, line 159:
```suggestion
```

Unused variable

powe97: Good catch. Surprised that didn't get caught by the static analysis.

ziesski (approved):
I tested after my change, and it works. The rest of the code looks good to me.

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -52,6 +52,34 @@ def update_docker_images(user, host, worker, autograding_workers, autograding_co
...
+            image_set = {
+                image_name
+                for image in client.images.list()
+                for image_name in image.attrs["RepoTags"]
+            }
+
+            images_to_remove = set.difference(image_set, images_to_update)
+
+            # Prevent removal of system docker containers
+            with open("/usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/data/system_docker_containers.json") as json_file:
```
williamjallen, line 66: We should not be referencing files in the `GIT_CHECKOUT` directory directly in source code for things like this.

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -52,6 +52,34 @@ def update_docker_images(user, host, worker, autograding_workers, autograding_co
...
+
+            images_to_remove = set.difference(images_to_remove, set(system_docker_containers))
+
+            # Remove images
+            for imageRemoved in images_to_remove:
+                try:
+                    image_id = client.images.get(imageRemoved).id
+                    print("Removed image " + imageRemoved)
+                except docker.errors.ImageNotFound as e:
+                    print("Couldn't find image ", e)
```
williamjallen, line 77: This message could easily be lost in the logs without someone realizing that an error occurred.  It would be better to explicitly add the text "ERROR:" to this string, like the other error messages in this file.  Ideally we should also print errors to stderr, but that's a broader issue outside the scope of this PR.

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -52,6 +52,34 @@ def update_docker_images(user, host, worker, autograding_workers, autograding_co
...
+                    continue
+                client.images.remove(image_id, True)
+        except Exception as e:
+            # Handle the exception
+            print("An error occurred:", e)
```
williamjallen, line 82: Same here.

williamjallen (changes requested):

sbin/shipper_utils/update_and_install_workers.py
```diff
@@ -52,14 +53,48 @@ def update_docker_images(user, host, worker, autograding_workers, autograding_co
...
         for image in images_to_update:
             print(f"locally pulling the image '{image}'")
             try:
                 repo, tag = image.split(':')
                 client.images.pull(repository=repo, tag=tag)
             except Exception as e:
-              print(f"ERROR: Could not pull {image}")
-              traceback.print_exc()
+                print(f"ERROR: Could not pull {image}: {e}", file=sys.stderr)
+                traceback.print_exc(file=sys.stderr)
```
ziesski, line 97:
```suggestion
              print(f"ERROR: Could not pull {image}: {e}", file=sys.stderr)
              traceback.print_exc(file=sys.stderr)
```

ziesski:
FIX tested works, good to go.

site/app/controllers/DockerInterfaceController.php
```diff
@@ -57,19 +63,12 @@ public function showDockerInterface(): MultiResponse {
...
-    /**
-     * @return JsonResponse | MultiResponse
-     */
+
     #[Route("/admin/add_image", methods: ["POST"])]
     #[Route("/api/admin/add_image", methods: ["GET"])]
-    public function addImage() {
+    public function addImage(): JsonResponse {
         $user = $this->core->getUser();
-        if (is_null($user) || !$user->accessFaculty()) {
```
DarthNyan, line 67: Is there another place where user access is checked, or do we want this page to be accessible to anyone?

ziesski: Although we do have several access control before add image, we should still have one. I have just added jsonreponse check, PR #7905 by Chris also did same check for `updateDockerCall` function so followed same structure.

shailpatels (commented):
Having to images with different `name:tags` but point to the same image ID appears to break the autograding_containers.json file

Using images `ubuntu/squid:edge` and `ubuntu/squid:latest` both point `87507c4542d0` (as of 12/11). 

Before I had the autograding_containers.json of
```
{
    "default": [
        "grycap/cowsay:latest"
    ],
    "cpp": [
        "grycap/cowsay:latest"
    ],
    "et-cetera": [
        "grycap/cowsay:latest"
    ],
    "python": [
        "ubuntu/squid:latest",
        "ubuntu/squid:edge"
    ]
}
``` 

Then removing one of the names that point to the same image causes it to look like

```
{
    "default": [
        "grycap\/cowsay:latest"
    ],
    "cpp": [
        "grycap\/cowsay:latest"
    ],
    "et-cetera": [
        "grycap\/cowsay:latest"
    ],
    "python": [
        "ubuntu\/squid:latest"
    ]
}
```

and then doing the process of adding and removing again totally breaks it with some errors in the docker logs

```
{
    "default": [
        "grycap\/cowsay:latest"
    ],
    "cpp": [
        "grycap\/cowsay:latest",
        "ubuntu\/squid:edge"
    ],
    "et-cetera": [
        "grycap\/cowsay:latest"
    ],
    "python": {
        "1": "ubuntu\/squid:edge"
    }

```

```
[Last ran on: 2024-12-11 23:24:05]
ERROR: Could not pull 1: not enough values to unpack (expected 2, got 1)
  File "/usr/local/submitty/sbin/shipper_utils/update_and_install_workers.py", line 94, in update_docker_images
    repo, tag = image.split(':')
ValueError: not enough values to unpack (expected 2, got 1)
Mode Set: only updating docker images.
-------------------------------------------------------
UPDATE MACHINE: primary
download/update docker images on localhost
localhost needs ubuntu/squid:edge, grycap/cowsay:latest, 1
Description:	Ubuntu 22.04.4 LTS

Removed image ubuntu/squid:latest
locally pulling the image 'ubuntu/squid:edge'
locally pulling the image 'grycap/cowsay:latest'
locally pulling the image '1'
Docker Version: 27.3.1
Tag: ubuntu/squid:edge
	-id: sha256:87507c4542d0
	-created: 2024-09-11T04:44:30.763585694Z
	-size: 242382063
Tag: lichen:latest, submitty/lichen:latest
	-id: sha256:53f99dde23f7
	-created: 2023-09-11T02:05:15.054839161Z
	-size: 2133006237
Tag: grycap/cowsay:latest
	-id: sha256:80cd02790675
	-created: 2017-06-01T13:34:53.717797922Z
	-size: 166214560
ERROR: Failed to pull one or more required docker images on primary
FAILURE TO UPDATE MACHINE primary
ERROR: FAILURE TO UPDATE ONE OR MORE MACHINES

[Last ran on: 2024-12-11 23:24:30]
```

shailpatels:
Since this Pr has been open for a while maybe it makes sense to install it (feature flag it?) and then work the extra issues mentioned above separately

PR #8853, opened 2023-02-07
[Bugfix:InstructorUI] Add/remove sections

### LINKED ISSUE #8471: Incorrect sections in auto-created customization.json
**Describe the bug**

Here are the registration sections for my course:  
A "STAFF" section + sections 1-10.  Note:  The staff section was created first.

Screen Shot 2022-10-02 at 10 15 59 PM

Then I used the form to create the customization.json for Rainbow Grades:
Grade Reports -> Web-Based Rainbow Grades Customization

The "Section Labels" portion of the UI is incorrect.  There is no section 11.  And we are missing Section 1.

Screen Shot 2022-10-02 at 10 17 40 PM

This error propagates to the auto-generated customization.json:

Screen Shot 2022-10-02 at 10 19 01 PM

ANS1514:
Prints the format properly now, as well as corrects the json file. Corrected the way to add and remove sections, as well as the edge case where a section is removed and added at the same time

Looks to fix #8471, currently counting wrong when creating the customization.json file. Also after the intial creation of sections, unable to add and delete sections properly. Also found that the database array was correct but the File array wasn't updating properly

### What is the new behavior?

Now prints out the correct format in the customization.json, and able to add/remove sections in proper order. Took the database array and converted into a temp array in the format of the file array then compared what was missing between the two then added/deleted sections accordingly.

### Other information?

Signed in as an instructor on localhost, added a new course, went to Manage Sections and Added Registration Sections "STAFF" then sections 1-5. Went to Grade Reports tab, pressed Generate Grade Summaries, then went to Web-Based Rainbow Grades Generation, pressed Save Changes, refreshed the page then checked the ordering of the Section Lables (if you didn't var_dump the arrays to check the order).

Also use
root@vagrant:/var/local/submitty/courses/s23/*YOUR_COURSE_NAME*/rainbow_grades# rm customization.json
if you want to clear your sections in the file array. You have to manually delete every section in Manage Sections to clear the database


site/app/models/RainbowCustomization.php
```diff
@@ -423,17 +423,32 @@ public function getSectionsAndLabels() {
...
-            // in the database, add new fields into sections from file with defaults
-            $sectionsFromFileCount = count($sectionsFromFile);
+            // gets the number of sections in the database
             $sectionsCount = count($sections);
 
-            if ($sectionsFromFileCount != $sectionsCount) {
-                for ($i = $sectionsFromFileCount + 1; $i <= $sectionsCount; $i++) {
-                    $sectionsFromFile[$i] = (string) $i;
+            // creates a copy of the database sections to compare against the file sections
+            $temp = [];
```
cjreed121, line 430: Could we replace temp with some more descriptive variable name?

site/app/models/RainbowCustomization.php
```diff
@@ -423,17 +423,32 @@ public function getSectionsAndLabels() {
...
+            for ($i = 0; $i < $sectionsCount; $i++) {
+                $temp[$db_sections[$i]['sections_registration_id']] = (string) $db_sections[$i]['sections_registration_id'];
+            }
+
+            // checks whether or not a sections has been added and upodates the file sections
+            $result = array_diff_key($temp, $sectionsFromFile);
+            if ($result) {
```
cjreed121, line 437: We know the return type of array_diff_key to be an array so the if statement would check if it's empty or not. For someone looking at this code for the first time, it would probably be better to do something like this instead:
```suggestion
            if (count($result) > 0) {
```

Do the same for line 445.

cjreed121 (changes requested):
Works well. Couple requested changes for future clarity but then I think it's good.

site/app/models/RainbowCustomization.php
```diff
@@ -423,17 +423,32 @@ public function getSectionsAndLabels() {
...
-            if ($sectionsFromFileCount != $sectionsCount) {
-                for ($i = $sectionsFromFileCount + 1; $i <= $sectionsCount; $i++) {
-                    $sectionsFromFile[$i] = (string) $i;
+            // creates a copy of the database sections to compare against the file sections
+            $temp = [];
+            for ($i = 0; $i < $sectionsCount; $i++) {
+                $temp[$db_sections[$i]['sections_registration_id']] = (string) $db_sections[$i]['sections_registration_id'];
+            }
+
+            // checks whether or not a sections has been added and upodates the file sections
```
williamjallen, line 435:
```suggestion
            // checks whether or not a sections has been added and updates the file sections
```

williamjallen (changes requested):

cjreed121 (approved):
Looks good to me!

ANS1514 (commented):
Made the changes from reviews

PR #8857, opened 2023-02-08
[Testing:Developer] Increase PHPStan level to 5

williamjallen:
PHPStan is currently set to level 4.  No errors are generated by PHPStan if the specified type does not match the type passed to a function.  This results in many cases of dubious type inference, where the code happens to work as expected, but not necessarily in a readable way.  As we start preparing to switch to PHP 8.x, it is possible that many of these cases will become problematic due to changes in type conversion.  In particular, I am worried about the changes in comparisons between strings and integers outlined [here](https://www.php.net/manual/en/migration80.incompatible.php).

For reference, the PHPStan rule levels can be found here: https://phpstan.org/user-guide/rule-levels

### What is the new behavior?
I have increased PHPStan's rule level to level 5.  This will ensure that types match when specified.  An attempt was made to cut down the number of errors, but only relatively safe changes were made.  The rest of the errors have been added to the baseline.


mahi-pas:
Error on "seek team/partner" button in teams
image
image
if I click the button, the screen becomes this error

mahi-pas:
It appears the buttons (documentation and run query) in SQL toolbox aren't working, they are pressed but nothing happens
image

mahi-pas:
Not sure if this is just because I'm running on vagrant but in office hours queue, today's history is blank, even after I clicked finish helping on Alyssa P Hacker. It says the websocket isn't connected
image

williamjallen:
@mahi-pas Thanks for finding these issues!  I will look into them.

Just a comment about reviewing: it is generally better to request changes instead of commenting if you find an issue like this.  You can do that by going to the "Files Changed" tab at the top of this page and clicking "Review changes"

williamjallen:
@mahi-pas 

1. "seek team/partner" error -- this seems to exist on `main` as well, so it is a bug which should be fixed, but is unrelated to my changes in this PR.  From a brief glance at the offending code, this issue seems to be unrelated to types altogether.  This function will throw an exception if no rows are returned, which is the source of the error you found.
https://github.com/Submitty/Submitty/blob/4d6120e98b3ef3be3a4f48573a55c72081bc5781/site/app/libraries/database/DatabaseQueries.php#L3446-L3456
2. I am unable to identify any issues with the SQL toolbox.  Writing a SQL query, and then clicking "Run Query" generates the expected results.  In your screenshot, I don't see a `FROM` clause in your SQL query, which might be causing issues.  Try something like `SELECT * FROM polls`.
3. OH Queue -- I am unable to find any issues on the office hours queue.  It is normal to not see anyone in the "today's history" section, because you probably created your VM more than a day ago.  If you add a new student to the queue and then finish helping them, the new student should show up in the history.  As for the websockets issue, that is probably a problem with your local environment.

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -155,7 +155,15 @@ private function editPage(Gradeable $gradeable, $semester, $course, $nav_tab = 0
...
-            return $a[0] > $b[0];
+            if ($a[0] > $b[0]) {
+                return 1;
+            }
+            elseif ($a[0] < $b[0]) {
+                return -1;
+            }
+            else {
+                return 0;
+            }
```
cjreed121, line 166: Could we just use `strcmp`?

site/app/libraries/plagiarism/PlagiarismUtils.php
```diff
@@ -68,7 +68,15 @@ public static function constructIntervalsForUserPair(string $filename, string $u
...
-            return $a->getStart() > $b->getStart();
+            if ($a->getStart() > $b->getStart()) {
+                return 1;
+            }
+            elseif ($a->getStart() < $b->getStart()) {
+                return -1;
+            }
+            else {
+                return 0;
+            }
```
cjreed121, line 79: Could we just return the difference between a's start and b's start?

cjreed121 (commented):
Changes all look good. Couple comments to possibly make things shorter/simpler.

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -155,7 +155,7 @@ private function editPage(Gradeable $gradeable, $semester, $course, $nav_tab = 0
             $repo_id_number++;
         }
         usort($all_repository_config_paths, function ($a, $b) {
-            return $a[0] > $b[0];
+            return strcmp($a, $b);
```
cjreed121, line 158: Haven't looked too much at this but wouldn't we want this instead?
```suggestion
            return strcmp($a[0], $b[0]);
```

cjreed121 (changes requested):

Issue #6014, opened 2020-11-11
Forum doesn't scroll to bottom of page when using filter with lots of wide categories.


bmcutler:
If you enable filtering in a course with lots of wide category names, so they wrap onto a second or more lines of categories, and you view a lengthy thread, you cannot scroll to the bottom of the posts.  It stops early.  And may not be able to access the buttons at the bottom of the page.  Might depend on zoom level and width of window.

williamjallen:
After some investigation, it seems that the issue is that the sidebar and content of the forum is being set to a fixed height of 81vh, a fixed number relative to the page height.  The parent element does not have a fixed height set which I suspect is why the original author used this CSS hack.  To fix this bug, some Javascript which changes the height based on the height of the element containing the filters could be added but this would be ugly and unreliable.  The better option here would be to almost completely rewrite the forum CSS but it would be at least a week's work so I don't know if it would be worth it for one minor bug.  If there are other bugs which could be fixed by redoing the CSS or if the UI needs to be overhauled anyway, perhaps it would be reasonable.  Any changes here require changes all the way up to the parent element for the entire page which then cascade back down and require changes to the CSS for other parts of the page.

SatendraDiwakar:
Hello there,
I have provided a fix for this issue by changing some css, please review the [PR](https://github.com/Submitty/Submitty/pull/8861) and let me know if anything needs to be changed as this is my first PR here.
Thanks in advance :)

PR #8861, opened 2023-02-08
[Bugfix:Forum] Scroll issue in sidebar of forum

SatendraDiwakar:
- When filter options are open then the last thread in sidebar gets hidden because of fixed height in child elements.
- **Fixes** #6014 

### What is the new behavior?
- **Reason**: The child elements **threads_list** and **posts_list** is having fix height of 81vh but the parent **thread_container** depends on its own parent due to which the child's scroll is not increasing.
- So I added fix height of **81vh** to **thread_container** and updated child height to **100%** so that they can increase there scroll based on content.

### To reproduce the issue
- Make sure your screen size is having height less than **800px**.
You can change your resolution from developer tools. My screen size is **1500x740**.
- Go to discussion forum and open any thread.
- Check the sidebar which has thread list, before filter options opened up.
- Check the sidebar which has thread list, after filter options opened up. You'll find that the last thread got hidden from the bottom.

### Screenshots

- **_Before fix_**

forumScrollIssue

- **_After fix_**

forumFixed


Eceptonsu (commented):
It seems like on my computer, the PR does not make a difference as describled (Windows 10, Chrome)? Here are the two screenshots. The first one is from main.
image
image

In addition, this PR creates the following behavior. The forums are not in a "container" anymore.
image
image

SatendraDiwakar:
Hi @Eceptonsu ,
Thankyou so much for the review,

> **In addition, this PR creates the following behavior. The forums are not in a "container" anymore.**

I have fixed this please review it again.

And I wanna specify that the issue was produce on screen height less than 800px as per my findings.
And as specified in description, for the issue to be reproduced the filter options has to be opened up.

I have also added reproducing steps in description of the PR.

Here is screenshot of tested resolution

image

SatendraDiwakar:
Hi @Eceptonsu , cc: @williamjallen ,
Just a reminder, I think this PR was missed in the traffic, please review the new changes and also check my comment above, whenever you get chance.
Thanks in advance

jason490 (approved):
Looks good.
The functionality works as stated.

PR #8863, opened 2023-02-09
[Bugfix:Developer] Fix invalid YAML in dependabot file

williamjallen:
https://github.com/Submitty/Submitty/pull/8788 mistakenly added invalid YAML to the `dependabot.yml` file, which has been causing intermittent errors on various PRs since https://github.com/Submitty/Submitty/pull/8788 was merged.  It is unclear why the tests on the PR did not pick up the failure.

### What is the new behavior?
`dependabot.yml` now passes [this](https://jsonformatter.org/yaml-validator) online validator, which is suggested by GitHub.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,8 +17,8 @@ updates:
     directory: "/site"
     schedule:
       interval: "weekly"
-        day: "monday"
-        time: "10:00"
+      day: "monday"
+      time: "10:00"
     open-pull-requests-limit: 15
     labels:
       - dependencies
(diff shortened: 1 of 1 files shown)


MasterOdin (approved):

MasterOdin:
I don't think we have any explicit tests that the `dependabot.yml` file is valid yml, though I think it's edited so infrequently that not sure it's worth it?

Could perhaps though add a step that validates that, as well as that the GH actions are also valid yml?

williamjallen:
> I don't think we have any explicit tests that the `dependabot.yml` file is valid yml. Could perhaps though add a step that validates that, as well as that the GH actions are also valid yml?

I had the exact same thought.  I'm actually about to open an issue about adding [yamllint](https://yamllint.readthedocs.io/en/stable) for exactly that reason.

The  Dependabot usually adds a failing test if there are issues with the `dependabot.yml` file, but it didn't on #8788 for whatever reason.  You can see an example of the failing test [here](https://github.com/Submitty/Submitty/pull/8251/checks?check_run_id=11154492359).

Issue #7885, opened 2022-05-13
Remaining Installation Issues on ARM64 AArch64 / Apple M1


bmcutler:
PR #6815 adds instructions for creating a development Virtual Machine on ARM64 AArch64 / Apple M1.

This issue records unresolved/incomplete tasks:

MasterOdin:
> [ ] with UTM, the ssh keys are not copied (.setup/install_system.sh)

With how the if statement works, this breaks usage of M1 with vagrant-parallels, since we're just checking for `aarch64`. I'd like to propose to add a new flag to `install_system.sh` (`--utm`) that would be used in place of `--vagrant`, and that would activate a new `${UTM}` variable within the installer. The existing `${UTM_ARM}` and its guards would then be split to either `${UTM}` or `${ARM}`, where most of the existing usages would remain using `${ARM}`, though some (like the above) would use `${UTM}`. Setting `--utm` would implicitly activate `--vagrant` for now, though we could work to make the underlying variables a bit more generically named (e.g. have an overarching `${DEV_VM}` for common actions, and then `${VAGRANT}` for specific vagrant things), but that's certainly not a priority.

PR #8890, opened 2023-02-11
[Refactor:System] Separate UTM variable from ARM

skara9:
#### Current Behavior
The VM installation scripts currently check for an `aarch64` architecture and assign the shell variable `UTM_ARM` to track it.
Also, the install script skips installing haskell stack for ARM machines.

#### Updated Behavior
* A separate `--utm` flag has been added to indicate running on UTM, differentiating between ARM configurations not running on UTM and configurations of any architecture running on UTM.
* A `DEV_VM` variable has been created to indicate a script running on either Vagrant or UTM.
* Furthermore, the install script has been updated to install haskell stack for ARM given that stack ships an ARM binary since [v2.9.1](https://github.com/commercialhaskell/stack/releases/tag/v2.9.1).

Partly addresses #7885


bmcutler:
@MasterOdin Can you help review this please? thanks

.setup/install_system.sh
```diff
@@ -59,6 +59,7 @@ source ${CURRENT_DIR}/bin/versions.sh
 # PROVISION SETUP
 #################
 
+export UTM=0
```
MasterOdin, line 62: add `export DEV_VM=0` up here as well.

MasterOdin (changes requested):
Changes overall seem reasonable given the concept that we are potentially supporting multiple different VM platforms. Some minor comments.

tkoz0:
I tested `vagrant up` on this branch and it works as expected. My host OS is Ubuntu 22.04 on an Intel x86_64 CPU and my hypervisor for the VM is VirtualBox 6.1.

bmcutler:
@MasterOdin could you please re-review?  Thanks!

EdenBendheim (commented):
I ran into an installation error:
image

Not too sure of how to fix it. Looked around in Install_system and the files for installing it were there.

MasterOdin:
@EdenBendheim You're hitting that error when going through UTM? I think we stopped trying to install that package from Ubuntu 20.04 and onwards though so I'm not sure why you're hitting that assuming you're installing Ubuntu 20.04. If you're trying to install an older Ubuntu, that's unsupported.

PR #8891, opened 2023-02-11
[Feature:System] Vagrant libvirt + remodel box selection

skara9:
Currently, the Vagrant file is set to override boxes based on virtual machine. This is redundant because both virtual machines used for Submitty are configured to use the same box.

Also, a global variable is used to indicate running on an M1 Mac, which appends the string '-arm64' to the box name. This is not a universal naming convention and only works with the current `bento/ubuntu-20.04` box which has a `bento/ubuntu-20.04-arm64` counterpart that serves as an alternative box for parallels desktop on arm64.

There is also no existing way to override the default box specified in the Vagrant file without manually updating the file itself. This makes for a poor development experience as changes made to the box name in the Vagrant file cannot be committed without affecting other users as the Vagrant file is tracked by git.

---
**Changes**
* Added configuration for the libvirt provider to the Vagrantfile
* Remodeled box selection in the Vagrantfile to be more robust and allow for encapsulated provider-specific overrides, which are much more useful than the current vm-specific overrides
* Added support for the environment variable `VAGRANT_BOX` to override the default box used in the Vagrant file
* Added support for the `vagrant-env` plugin which allows reading from a `.env` file (also added to gitignore)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -149,7 +149,8 @@ Vagrant.configure(2) do |config|
   config.vm.provider "parallels" do |prl, override|
     # Default box for parallels on arm64
     unless ENV.has_key?('VAGRANT_BOX')
-      if (`uname -m`.chomp == 'arm64' || `uname -m`.chomp == 'aarch64')
+      arch = `uname -m`.chomp
+      if (arch == 'arm64' || arch == 'aarch64')
         override.vm.box = "bento/ubuntu-20.04-arm64"
       end
     end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,6 +11,13 @@
 #   WORKER_PAIR=1 vagrant up
 #
 #
+# If you want to override the default image used for the virtual machines, you can set the
+# environment variable VAGRANT_BOX. See https://vagrantup.com/boxes/search for a list of
+# distributed boxes. For example:
+#
+# VAGRANT_BOX=ubuntu/focal64 vagrant up
+#
+#
 # If you want to install extra packages (such as rpi or matlab), you need to have the environment
 # variable EXTRA set. The easiest way to do this is doing:
 #
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -52,12 +52,6 @@ VERSION=$(lsb_release -sr | tr '[:upper:]' '[:lower:]')
 bash ${GIT_PATH}/.setup/install_worker.sh #{extra_command} 2>&1 | tee ${GIT_PATH}/.vagrant/install_worker.log
 SCRIPT
 
-box_extra = if Vagrant::Util::Platform.darwin? && (`uname -m`.chomp == "arm64" || (`sysctl -n machdep.cpu.brand_string`.chomp.include? 'M1'))
-  '-arm64'
-else
-  ''
-end
-
 def mount_folders(config, mount_options)
   # ideally we would use submitty_daemon or something as the owner/group, but since that user doesn't exist
   # till post-provision (and this is mounted before provisioning), we want the group to be 'vagrant'
@@ -77,6 +71,14 @@ def mount_folders(config, mount_options)
 end
 
 Vagrant.configure(2) do |config|
+  # Default box
+  config.vm.box = "bento/ubuntu-20.04"
+
+  # Custom box
+  if ENV.has_key?('VAGRANT_BOX')
+    config.vm.box = ENV['VAGRANT_BOX']
+  end
+
   mount_options = []
 
   # The time in seconds that Vagrant will wait for the machine to boot and be accessible. 
@@ -88,7 +90,6 @@ Vagrant.configure(2) do |config|
   # so that when we do "vagrant up", it doesn't spin up those machines.
 
   config.vm.define 'submitty-worker', autostart: autostart_worker do |ubuntu|
-    ubuntu.vm.box = "bento/ubuntu-20.04#{box_extra}"
     # If this IP address changes, it must be changed in install_system.sh and
     # CONFIGURE_SUBMITTY.py to allow the ssh connection
     ubuntu.vm.network "private_network", ip: "172.18.2.8"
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -160,6 +173,20 @@ Vagrant.configure(2) do |config|
     mount_folders(override, [])
   end
 
+  config.vm.provider "libvirt" do |libvirt, override|
+      # Default box with libvirt support
+      unless ENV.has_key?('VAGRANT_BOX')
+        override.vm.box = "generic/ubuntu2004"
```
MasterOdin, line 179: Should we move the regular base image over to `generic/ubuntu2004` instead of using `bento/ubuntu-20.04`? The generic project doesn't look like it'll support arm64 anytime soon so parallels with arm64 will always need an alternative, but that doesn't stop non-arm64 people.

If we want to keep using bento as our base image, we should also look to create an issue on their tracker seeing if they'd be open to adding `libvrt` boxes as that would then allow us to drop the usage of `generic/ubuntu2004` here.

Vagrantfile
```diff
@@ -147,6 +152,14 @@ Vagrant.configure(2) do |config|
   end
 
   config.vm.provider "parallels" do |prl, override|
+    # Default box for parallels on arm64
+    unless ENV.has_key?('VAGRANT_BOX')
+      arch = `uname -m`.chomp
+      if (arch == 'arm64' || arch == 'aarch64')
+        override.vm.box = "bento/ubuntu-20.04-arm64"
```
MasterOdin, line 159: From the other comment, with this PR, we would define separate boxes in 3 different places without any close proximity to each other, which would certainly lead to a potential problem someone updates only the base image to use ubuntu-22.04, but leaves parallels and libvrt alone, and then something accidentally breaks.

I would much prefer to see that we define at the top of this file something like:

```
# All images should point to the same version of ubuntu
base_image = "bento/ubuntu-20.04"
base_parallels_image = if (arch == 'arm64' || arch == 'aarch64')
  "#{base_image}-arm64"
else
  base_image
end
base_libvrt_image = "generic/ubuntu2004"
```

This creates a much stronger spatial link between these box names, and should make it harder to have them differ in ubuntu verison.

Within the code, we'd use `config.vm.box = base_image`, `override.vm.box = base_parallels_image`, etc.

MasterOdin (changes requested):

Vagrantfile
```diff
@@ -77,6 +78,19 @@ def mount_folders(config, mount_options)
...
+    config.env.enable
+  end
+
+  # Default box
+  config.vm.box = base_boxes.default
+
+  # Custom box
+  if ENV.has_key?('VAGRANT_BOX')
+    config.vm.box = ENV['VAGRANT_BOX']
+  end
```
MasterOdin, line 92: The more idiomatic way to write this would be:

```suggestion
  config.vm.box = ENV.fetch('VAGRANT_BOX', base_boxes.default)
```

I would also drop the `# Default box` and `# Custom box` comments. They just give a "what" the line does, and the line itself is simple enough to not warrant the comment.

Vagrantfile
```diff
@@ -52,11 +52,12 @@ VERSION=$(lsb_release -sr | tr '[:upper:]' '[:lower:]')
...
 SCRIPT
 
-box_extra = if Vagrant::Util::Platform.darwin? && (`uname -m`.chomp == "arm64" || (`sysctl -n machdep.cpu.brand_string`.chomp.include? 'M1'))
-  '-arm64'
-else
-  ''
-end
+base_boxes = Hash[]
+
+# Should all be base Ubuntu 20.04 boxes
```
MasterOdin, line 57: I would reword this to drop the "20.04" part of this, so that when we upgrade to 22.04 (and beyond), it's one less line that needs to change.

```suggestion
# Should all be base Ubuntu boxes that use the same version
```

Vagrantfile
```diff
@@ -52,11 +52,12 @@ VERSION=$(lsb_release -sr | tr '[:upper:]' '[:lower:]')
...
+base_boxes.default         = "bento/ubuntu-20.04"
+base_boxes[:arm_parallels] = "bento/ubuntu-20.04-arm64"
+base_boxes[:libvirt]       = "generic/2004"
```
MasterOdin, line 60: I think there's a typo on this line?
```suggestion
base_boxes[:libvirt]       = "generic/ubuntu2004"
```

Vagrantfile
```diff
@@ -147,6 +159,14 @@ Vagrant.configure(2) do |config|
   end
 
   config.vm.provider "parallels" do |prl, override|
+    # Default box for parallels on arm64
```
MasterOdin, line 162: Same for this comment, it's just the "what" of this block, and it's simple enough already to not warrant it.

Vagrantfile
```diff
@@ -160,6 +180,20 @@ Vagrant.configure(2) do |config|
     mount_folders(override, [])
   end
 
+  config.vm.provider "libvirt" do |libvirt, override|
+      # Default box with libvirt support
```
MasterOdin, line 184: Same for this comment, it's just the "what" of this block, and it's simple enough already to not warrant it.

Vagrantfile
```diff
@@ -77,6 +78,19 @@ def mount_folders(config, mount_options)
 end
 
 Vagrant.configure(2) do |config|
+  # Optional plugin for reading from env file
```
MasterOdin, line 81: Same here, it's not a meaningful comment.

Vagrantfile
```diff
@@ -77,6 +78,19 @@ def mount_folders(config, mount_options)
...
+  if Vagrant.has_plugin?('vagrant-env')
+    config.env.enable
+  end
+
+  # Default box
+  config.vm.box = base_boxes.default
+
+  # Custom box
+  if ENV.has_key?('VAGRANT_BOX')
```
MasterOdin, line 90: Please also update the comment at the top of this page to mention `VAGRANT_BOX` and how one could use it.

MasterOdin (changes requested):

MasterOdin (approved):

PR #8893, opened 2023-02-11
[Bugfix:InstructorUI] Button click -> calendar gone

### LINKED ISSUE #4712: Course materials date calendar can go off screen
On small screens:
Go into course materials, click upload course materials.

In the modal, click on the input box under "Choose a time to release the file being uploaded".

The "now" button to finalize the date can go off screen.

saumyaborwankar:

The calendar did not go away initially when you clicked any buttons or anywhere else on the popup menu. You needed to click on another input to make it go away. Also, the calendar did not move relative to the popup menu, it used to move relative to the background page.

Closes #4712

### What is the new behavior?
When you press a button between "Beginning of Time", "Now", "Tomorrow" or "End of time", the date calendar goes away.

https://user-images.githubusercontent.com/46644513/221048033-5520ee51-9af7-44d3-8f27-32e15af9de95.mp4

### Other information?
 Not a breaking change.
 Tested it locally.


bmcutler:
Thank you for your contribution.

Please read the instructions for making a pull request here:
https://submitty.org/developer/getting_started/make_a_pull_request

Please expand the body of the PR message and consider adding a screenshot.

saumyaborwankar:
@bmcutler Can you review this PR and give me some info about the error I'm getting from the WAVE extension
image


And this is from UploadCourseMaterialsForm.twig
image

I did not change the input tag so I'm not sure why it wants me to put a label for it. Any info on this will be helpful.

hansongu123:
Thank you for your contribution! I think it would be a good idea to separate the new "set time" button from rest of the buttons by adding more space between it and "End of time" and maybe apply a different color/style to the button or text. Maybe also change the text to "Done" and have the other buttons: "Beginning of Time", "Now", "Tomorrow", "End of time" not close the calendar on click - since they would be equivalent as selecting a time slot manually.

hansongu123:
Another suggestion will be to glue the date picker calendar to the "choose a time..." window to allow it to be movable by scroll in the pop-up but not by scrolling on the other part of the page - as mentioned in the original issue.

hansongu123:
If you get stuck or encounter an issue, feel free to join the Submitty Slack and to discuss with your fellow developers!

saumyaborwankar:
> Another suggestion will be to glue the date picker calendar to the "choose a time..." window to allow it to be movable by scroll in the pop-up but not by scrolling on the other part of the page - as mentioned in the original issue.

Hey @hansongu123 thanks for your inputs, I actually made these changes (relative scroll of the date picker with the popup)in my last commit but it was failing the e2e workflow because flatpickr adds an additional <div> tag inside the <label> tag so it fails the test as <div>tags cant be enclosed inside <label> tags. So I had to roll back my changes. Do you have any idea how i could approach this?

According to the flatpickr documentation "The flatpickr doc suggests using the static option if the flatpickr is enclosed within a scrollable page, The only problem we have is the <label> tag enclosing the flatpickr. "

saumyaborwankar:
> Thank you for your contribution! I think it would be a good idea to separate the new "set time" button from rest of the buttons by adding more space between it and "End of time" and maybe apply a different color/style to the button or text. Maybe also change the text to "Done" and have the other buttons: "Beginning of Time", "Now", "Tomorrow", "End of time" not close the calendar on click - since they would be equivalent as selecting a time slot manually.

Ill work on this and update the PR. Thanks.

hansongu123:
> Hey @hansongu123 thanks for your inputs, I actually made these changes (relative scroll of the date picker with the popup)in my last commit but it was failing the e2e workflow because flatpickr adds an additional

>tag inside the tag so it fails the test as
tags cant be enclosed inside tags. So I had to roll back my changes. Do you have any idea how i could approach this?

>According to the flatpickr documentation "The flatpickr doc suggests using the static option if the flatpickr is enclosed within a scrollable page, The only problem we have is the tag enclosing the flatpickr. "


Maybe [this](https://github.com/flatpickr/flatpickr/issues/1160) closed Flatpickr GitHub issue can answer your question about how to make the `date_picker` object sticking to a position in a page. (Also do note that we don't want the picker to be scrollable outside of the pop-up.)

hansongu123:
Do note that Flatpickr is also an open source project on GitHub. You can always search within their GitHub closed issues for one that is similar to your question.  Additionally, [this](https://stackoverflow.com/questions/66335759/how-to-fix-missing-form-label-error-in-material-ui-select-using-chrome-wave-to) Stack Overflow post maybe can help you with the error you encountered from WAVE when you set `static: true` in that Flatpickr object.

hansongu123 (commented):
I just looked into this again. An immediate solution will be to ignore the on-scroll error appearing at the moment, since it is only affecting the date-pickers appearing on the modals (modal in a modal) - and trying to solve the problems mentioned in the linked issue by: 1. making the original "set time" button closes the date-picker and saves input; 2. clicking outside of the date-picker closes the date-picker but doesn't save the input; 3. clicking on all other buttons in the bottom of the date-picker closes the date-picker and saves the input. And then most importantly, making those buttons immediately usable by a user by positioning the date-picker better so they are not hidden. 

image

hansongu123:
I'll approve it as long as the changes are reasonable and they satisfy for a solution for problems intended by the linked issue and within its comments.

saumyaborwankar:
Hi @hansongu123 I think there is a problem with requirement 2 as per your last comment. Clicking outside does not register as the datepickr considered the entire popup menu as a body that doesn't allow the datepickr to close. Ideally the code without my changes should have been sufficient to close the datepickr if someone clicked outside. I read about this from an issue on flatpickr https://github.com/flatpickr/flatpickr/issues/964. I dont know if i have the correct understanding of this yet.

A solution that comes to my mind is using the onChange() function which weirdly registers the clicking outside the datepickr also this function will close the flatpickr when you change the day/time, which will eventually solve our problem and we don't need a "set time" button in that case. This function will also solve the problem of closing the datepickr when you click outside it. So which option should we go for?
(I still don't understand why flatpickr never gave an onDayChange() and onTimeChange() function )

Let me know your thoughts.

https://user-images.githubusercontent.com/46644513/223816860-2980dd3f-e54b-4fff-8f7a-67bd92a9ec09.mp4

hansongu123:
Logic-wise, the "Beginning of time", "Now", "End of Time" buttons all don't require a manually written "time of the day" input. I assume the "Tomorrow" button acts the same way although it isn't that intuitive to the user of what time it is picking. They are okay to be left close-on-click as long as they update the correct information. I like the idea that clicking on outside of the calendar closes it - the only problem you're having is that it doesn't discard the input or registers to empty, but on a second thought this is rather a stylistic choice, it is okay to ignore it for now.  The scrolling is actually a big deal, but this can't be solved rather quickly. I think you can ignore it for now too since this is regarding a mechanism within `flatpickr` itself. I think I'll make a separate issue for it to either re-implement the date-pickers or find a way around it, and you can make a PR if you're interested.  In the video it shows that manually clicking on one of the day slots in the calendar closes it, this would mean that an user must input the "time of the day" beforehand if they want it to be taken into account - which can be sometimes troublesome. Let's remove the close-on-click for clicking on a day in the calendar and make the save mechanism for it to be only clicking away.
To summarize, make: 
1.  Buttons on the bottom of the pop-up calendar close-on-click
2.  Clicking outside of calendar registers the input instead of discards it 
3.  Remove the "set time" button since it now overlaps in functionality with 2

The purpose of this PR is to make the date-picker as conventional and easy-to-use as possible. Normally the calendar would appear either on-top or below of the window that calls it and it is totally depending on the size of browser window it is opened in of whether is the case. Now, is there a way to make sure an user doesn't need to scroll to see the entire calendar and the buttons under it? (refer to the comments in the linked issue about scrolling) I'd say to fix it in a certain position that allows the user to see everything - but that is the lazy way to do it. Lemme know if you come up with something like-wise or a better solution, and push your changes up so I can look at the diff. Thanks!

PR #8894, opened 2023-02-11
[Feature:InstructorUI] Course Materials upload folder selector

### LINKED ISSUE #8263: Course Materials Upload folder/subfolder selector
When an instructor uploads a new course material, they have an option of specifying a subfolder.
The webbrowser does autofill previous data you've typed in that input box on the webform -- but it's not necessarily a folder/subfolder that exists in the current course (it could be for another course).

If the instructor makes a small unintentional typo in the name, they might create a new folder e.g., 'Homeworks'  or "homework' instead of 'Homework'.  

It would be nice to change this form to have a selection dropdown with all of the existing folders & subfolders in the current courses course materials uploads.  
And then a 'new folder/subfolder' button if you did want to specify something new.
[ note: this design suggestion is incomplete, please propose an intuitive and efficient UI. ]

Note: Instructors use course materials organization in many different ways.  Let's make sure we support all:
Some instructors put all files in the top level, and don't use any folders.
Some instructors have 1 level of folders.
Some instructors have multiple levels of folders & subfolders.
Some folders contain a mix of files & subfolders.
Some folders only have subfolders.

@holzbh comments?

jason490:

Closes #8263 
When creating a new course material, there is an option to put the material in a folder or make a new folder if the folder doesn't exist, but it doesn't tell the user what folders exist nor choose the directory that already exist

### What is the new behavior?
Added a autocomplete feature which would show options of existing folders
Screenshot from 2023-02-11 18-08-43
Screenshot from 2023-02-11 18-06-56


site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -207,6 +207,22 @@
         dropzone.addEventListener("dragleave", draghandle, false);
         dropzone.addEventListener("drop", dropWithMultipleZips, false);
     </script>
+    <script>
+        $(function() {
+            var paths = Object.keys($.parseJSON('{{ directory_priorities|json_encode|raw }}'));
```
williamjallen, line 212: It is not good practice to use `var` in JS code.  You should use `let` or `const` instead, as appropriate.  The differences are discussed in a very good stack overflow post here: https://stackoverflow.com/a/11444416/6768191

```suggestion
            const paths = Object.keys($.parseJSON('{{ directory_priorities|json_encode|raw }}'));
```

site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -207,6 +207,22 @@
...
+            for (let x = 0; x < paths.length; x++){
+                paths[x] = paths[x].slice(69);
```
williamjallen, line 214: Why was 69 chosen for the maximum path length here?  The logic behind this decision should be documented in a comment for developers who may stumble upon this code in the future.  As it is now, this number seems rather arbitrary.

site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -207,6 +207,22 @@
         dropzone.addEventListener("dragleave", draghandle, false);
         dropzone.addEventListener("drop", dropWithMultipleZips, false);
     </script>
+    <script>
+        $(function() {
+            var paths = Object.keys($.parseJSON('{{ directory_priorities|json_encode|raw }}'));
+            for (let x = 0; x < paths.length; x++){
```
williamjallen, line 213: It might be better to use `forEach()` here instead of iterating over the array manually.

site/app/templates/course/UploadCourseMaterialsForm.twig
```diff
@@ -207,6 +207,22 @@
         dropzone.addEventListener("dragleave", draghandle, false);
         dropzone.addEventListener("drop", dropWithMultipleZips, false);
     </script>
+    <script>
```
williamjallen, line 210: Why are there three different sets of `<script>` tags back to back here?  All of this should be combined into one single `<script>` tag.

Ideally some of this JS code should be moved to a `.js` file if possible (everything except twig-rendered data), but splitting that into a separate file isn't necessarily related to this change, and should be done in a separate PR.

williamjallen (approved):

site/app/templates/course/UploadCourseMaterialsForm.twig
williamjallen: If that is the case, a new approach needs to be taken here.  Course names don't necessarily have a consistent length.  A hacky way to do this would be to split on slashes, and then remove the first 8 components, but there is probably a more correct way to do this.

williamjallen (changes requested):
I clicked the wrong button when reviewing earlier.

jason490:
> From my testing it looks like clicking on an option in the autocomplete dropdown hides the dropdown without selecting the value. I'm only able to select options using the keyboard. Tested on all major browsers.

After testing on certain browsers, the issue still doesn't fix the textbox not filling when clicking on one on the options.
 I tried moving the autocomplete UI as a child to different elements and it still doesn't fill in the textbox after clicking the option. Based on my testing, it only works when its defaulted to the body of the website page.

site/app/templates/course/UploadCourseMaterialsForm.twig
jason490: Yeah, I hardcoded to 69, assuming the path will remain the same. So maybe when it sees course_materials, then cut out the rest of the path names, since course_materials should stay the same name?

PR #8896, opened 2023-02-13
[Bugfix:System] Secondary Email Course Sync

### LINKED ISSUE #8448: Adding user to class removes secondary email
**Describe the bug**
Adding an user to a new class removes the secondary email of that user

**Expected behavior**
The secondary email of the user should remain the same after being added to a class

**To Reproduce**
1. Set secondary email in an user's account
2. Add user to a new class
3. The user will see that their secondary email is gone

cjreed121:

When adding a student to a course or modifying certain information about a student, their secondary email address will be removed. The notify both email addresses field also gets set to false even if it was true.
Closes #8448

### What is the new behavior?
Adding or modifying students keeps the previous secondary email address and notify setting.

### Other information?
Tested locally and worked as expected.


site/app/controllers/admin/UsersController.php
```diff
@@ -355,7 +355,7 @@ public function updateUser($type = 'users') {
             $this->core->addSuccessMessage("User '{$user->getId()}' updated");
         }
         else {
-            if ($this->core->getQueries()->getSubmittyUser($_POST['user_id']) === null) {
+            if (($u = $this->core->getQueries()->getSubmittyUser($_POST['user_id'])) === null) {
```
williamjallen, line 358: This syntax is somewhat confusing, especially given that `$u` is used again 368.  It might be more clear to do the variable assignment on the line above, and then only perform the check on this line.

site/app/controllers/admin/UsersController.php
```diff
@@ -355,7 +355,8 @@ public function updateUser($type = 'users') {
             $this->core->addSuccessMessage("User '{$user->getId()}' updated");
         }
         else {
-            if ($this->core->getQueries()->getSubmittyUser($_POST['user_id']) === null) {
+            $submitty_user = $this->core->getQueries()->getSubmittyUser($_POST['user_id']);
+            if ($submitty_user  === null) {
```
williamjallen, line 359:
```suggestion
            if ($submitty_user === null) {
```

Another small nitpick comment here.  It's interesting that PHPCS didn't pick this up--as far as I am aware it used to complain about stuff like this.  I'll look into it.

tkoz0 (approved):
Tested, works. I just want someone else to confirm having the down migration is necessary. It makes sense to me since I would think the secondary email was added to the database at some point and these changes to the trigger function would not work if rolling back that far.

Issue #8854, opened 2023-02-07
Panels over extending in TA grading 


jenaybarela:
**Describe the bug**
When grading in two panel mode if one of the panels is long it over extends all the other panels leading to a lot of blank screen to scroll through. This is only a problem not in full screen mode.

**Expected behavior**
One panel should not impact the size of other panels. It would be better if the panel scrolled like it does in full screen mode.

**To Reproduce**
1.  As a grader in the sample course from the gradable page grade Grading Homework.
2. Pick a student to grade
3. Open two panel mode with four panels 
4. Fill the panels with one of them being the file panel
5. expand all the files
6. See error

**Configuration**
 - OS: Windows
 - Browser: Chrome

**Screenshots**
![Screenshot 2023-02-07 165631](https://user-images.githubusercontent.com/25368899/217376925-4a2ba3b5-3786-49ad-93fd-fca3a09b5fc0.png)

DefinitiveAbove:
Issue fixed in #8902

PR #8902, opened 2023-02-14
[Bugfix:TAGrading] Fix scrolling issue on TA grading page

DefinitiveAbove:
The global and electronic CSS force the minimum height of the page wrapper to be the max height of the content (to disable sidebar scrolling). This has a side-effect of making the TA grading page be taller than the browser viewport height and allowing the panels to extend very far height-wise.

### What is the new behavior?
Added CSS to undo the min-height requirement for only the TA grading page, making the TA grading page height fit the viewport again.


bmcutler:
does this close #8854?
if so ask @jenaybarela to review

Eceptonsu (approved):
Works as expected, and does close #8854.

PR #8904, opened 2023-02-15
[Feature:Forum] Unread and Deleted Posts a Gray-Blue 

### LINKED ISSUE #7027: Unread & deleted thread/post
Only instructors (& other teaching staff?) can view deleted thread/posts.  

[in light mode] When a post is new/unread, it is light blue.  When a post is deleted it is grey.  However, when a post is both deleted and unread, it is blue.  It is rare, but it can be confusing.  Let's it a different darker grey blue in this case -- how about '#aabbcc'.

...  also a pick a color that works in dark mode.

04vmatsibekker:

Fixes #7027. Threads that are both deleted and unread show up as a light blue which is confusing when compared to just an unread thread.

### What is the new behavior?
The threads that are both deleted and unread are a gray-blue. (This also takes care of different colors in dark mode)
In dark mode I also changed the deleted thread's text be black rather than white so that there is a visible contrast for users.

https://user-images.githubusercontent.com/117249183/223539255-cdabae75-be95-4c17-a4db-a477c3803f00.mp4

### Other information?
Note that when an Instructor clicks on the thread and views it, the color will only change to the deleted color, Gray, when the site is refreshed.


bmcutler:
Please read through
https://submitty.org/developer/getting_started/make_a_pull_request

So you can pass the PR title check and also include a screenshot.

jenaybarela (changes requested):
Looks good in light mode but its a bit hard to read in dark mode. A different color would probably be a good idea to fix this.

jssunray34 (changes requested):
Nice job! However, I agree with Jenay that it's a little hard to read in dark mode. Also, when you click on a deleted unread post, the sidebar thread remains gray-blue until you refresh. We probably want the color to change in both places at the same time. Here is an image showing what I mean:
Screenshot 2023-02-28 at 11 15 10 AM

site/app/views/forum/ForumThreadView.php
```diff
@@ -747,6 +747,10 @@ public function displayThreadList($threads, $filtering, &$activeThreadAnnounceme
                 $class .= " new_thread";
             }
             if ($thread["deleted"]) {
+                $array_of_class_IDs = explode(" ", $class);
+                if (in_array("new_thread", $array_of_class_IDs)) {
```
cjreed121, line 751: It might be a bit better to use the actual variables that determines whether it's a new thread or not rather than the class list. So on line 746 you could just store that result and substitute that new variable in here and on 746.

site/app/views/forum/ForumThreadView.php
```diff
@@ -969,6 +973,9 @@ public function createPost($thread_id, $post, $unviewed_posts, $first, $reply_le
         }
         if ($post["deleted"]) {
             $classes[] = "deleted";
+            if (in_array("new_post", $classes)) {
```
cjreed121, line 976: Same for this as my previous comment

cjreed121 (changes requested):
Tested locally and it works well. Just a few changes for future maintainability. Also make sure to include a screenshot in your PR description and update the other information as that is now fixed.

site/app/views/forum/ForumThreadView.php
04vmatsibekker: Alright, I will get to work on it!

cjreed121 (approved):
Thanks for making the changes. Works as described and looks good.

I noticed you were failing the linters a couple times in a row. To save some time, you can always run these locally by following https://submitty.org/developer/testing/index.

PR #8910, opened 2023-02-16
[Feature:SubminiPolls] Show Correct Poll Answers

hansongu123:

Currently, there's no way to display answers of polls to students after they're closed. In the poll editing tool, there are two dropdown windows responsible for its behavior - respectively when to release a poll, and when to release a response histogram to students. First window takes in a date in the calendar for its input, and second takes the options "Never", "After the Poll Ended", and "Always".

image

### What is the new behavior?
Add a third dropdown window responsible for disclosing answers to students, that has the options similar to the second dropdown window.

closes https://github.com/Submitty/Submitty/issues/7352

Handles scenarios:
1. "While the poll is open for response, the instructor may wish to allow students to see the current histogram of answers... but not yet view which answer(s) are correct."
2. "Once the poll is closed, the instructor may not wish to display which answers are correct -- because that information may be misleading. Sometimes a poll has one perfect answer that will receive full credit but it also has a partially correct answer that the instructor also decides to give full credit. It would be misleading to show both of them as equally worth full credit. In this case the instructor may want to communicate this in other ways."

image


williamjallen (changes requested):
It seems that you have checked in some irrelevant changes here which should be reverted.

hansongu123:
@williamjallen Yes. That currently has no effect on my code as I'm still experimenting with making changes. I'm trying to track all of the changes made by doctrine migration, which I know you worked on, and at the same time replicate Miles' PR on this feature as much as possible. Is there anything you care to share or details to follow?

hansongu123 (commented):
also revert Vagrantfile to default and deleted _exclude_o

site/app/entities/poll/Poll.php
```diff
@@ -167,6 +173,26 @@ public function isHistogramAvailable(): bool {
         return ($this->release_histogram === "always" && !$this->isClosed()) || ($this->release_histogram === "when_ended" && $this->isEnded());
     }
 
+    /**
+     * Note: This function should only be used if the actual string is desired.  (exporting poll data for example)
+     *       isShowAnswer() is preferred if at all possible.
+     */
+
+    public function setReleaseAnswer(string $status): void {
```
williamjallen, line 181:
```suggestion
    /**
     * Note: This function should only be used if the actual string is desired.  (exporting poll data for example)
     *       isReleaseAnswer() is preferred if at all possible.
     */

    public function setReleaseAnswer(string $status): void {
```

site/app/libraries/PollUtils.php
```diff
@@ -54,4 +55,5 @@ public static function isSingleResponse(string $poll_type): bool {
                 || ($poll_type == "single-response-multiple-correct")
                 || ($poll_type == "single-response-survey"));
     }
+
```
williamjallen, line 58:
```suggestion
```

migration/migrator/data/course_tables.sql
```diff
@@ -1389,7 +1389,8 @@ CREATE TABLE public.polls (
     release_date date NOT NULL,
     image_path text,
     question_type character varying(35) DEFAULT 'single-response-multiple-correct'::character varying,
-    release_histogram character varying(10) DEFAULT 'never'::character varying
+    release_histogram character varying(10) DEFAULT 'never'::character varying,
+    release_answer character varying(10) DEFAULT 'never'::character varying
```
williamjallen, line 1393: There doesn't seem to be a migration associated with this change.  You should make a migration, run the migration, and then dump the changes (which will make edits to this file).  You can read more about it here: https://submitty.org/developer/development_instructions/migrations

williamjallen (changes requested):
These changes seem to be almost complete.  There isn't a migration checked in so I can't test this PR locally very easily, but once the migration is complete, I'll review the functionality.    I also noted a couple other trivial code style issues.

hansongu123 (commented):
The previous few pushes are attempts to revert back some editor made auto-spacing or new lines, which affect status of DB checking but not the actual content of changed files.

hansongu123:
Cypress tests added. And also created test cases for release-histogram selector. Added ids for two toggleable buttons in the student poll interface.

site/app/templates/polls/PollForm.twig
```diff
@@ -110,6 +110,16 @@
...
             <option value="always" {{ poll is not null and poll.getReleaseHistogram() == 'always'? "selected" : "" }}>Always</option>
         </select>
+        <br><br>
+        <label for="student-answer-release-setting" class="option-title">
+            Show correct response to students:
+        </label>
+        <select id="student-answer-release-setting" name="release_answer">
+            <option value="never" {{ poll is null or poll.getReleaseAnswer() == 'never'? "selected" : "" }}>Never</option>
+            <option value="when_ended" {{ poll is not null and poll.getReleaseAnswer() == 'when_ended'? "selected" : "" }}>After Poll Ended</option>
+            <option value="always" {{ poll is not null and poll.getReleaseAnswer() == 'always'? "selected" : "" }}>Always</option>
```
williamjallen, line 120:
```suggestion
            <option value="never" {{ poll is null or poll.getReleaseAnswer() == 'never' ? "selected" : "" }}>Never</option>
            <option value="when_ended" {{ poll is not null and poll.getReleaseAnswer() == 'when_ended' ? "selected" : "" }}>After Poll Ended</option>
            <option value="always" {{ poll is not null and poll.getReleaseAnswer() == 'always' ? "selected" : "" }}>Always</option>
```

site/cypress/e2e/polls.spec.js
```diff
@@ -149,6 +149,8 @@ describe('Test cases revolving around polls functionality', () => {
         cy.get('#poll-date').clear({force: true});
         cy.get('#poll-date').type('1970-01-01', {force: true});
         cy.get('h1').click(); // get rid of the date picker
+        cy.get('#student-histogram-release-setting').select('never');
+        cy.get('#student-answer-release-setting').select('never');
```
williamjallen, line 153: If you're creating a new poll and these are default values, there's no need to set them.  Furthermore, this would be a good chance to make sure the default values are set correctly.

I can't comment on unchanged lines so I'll leave my comment here:
It might be good to add some checks for the release histogram and answer release setting a little bit after line 186.  Adding checks around line 212 might be good for the same reason.

site/cypress/e2e/polls.spec.js
```diff
@@ -279,6 +283,62 @@ describe('Test cases revolving around polls functionality', () => {
...
+        cy.login();
+        cy.visit(['sample', 'polls']);
+        cy.contains('Poll Cypress Test').siblings(':nth-child(6)').children().click();
+        cy.contains('Poll Cypress Test').siblings(':nth-child(1)').children().click();
+        cy.url().should('include', 'sample/polls/editPoll');
+        cy.get('#breadcrumbs > :nth-child(7) > span').should('have.text', 'Edit Poll');
+        cy.get('#student-histogram-release-setting').select('when_ended');
+        cy.get('#student-answer-release-setting').select('when_ended');
+        cy.get('button[type=submit]').click();
+        cy.wait(1000);
```
williamjallen, line 297: This wait is probably unnecessary, and slows down the test suite quite a bit.  It also has the potential to result in a flaky test if the test runner is running slowly sometime.  The other waits should be eliminated as well.

site/cypress/e2e/polls.spec.js
```diff
@@ -279,6 +283,62 @@ describe('Test cases revolving around polls functionality', () => {
...
+
+        // log into student to verify histogram and answer are released
+        cy.logout();
+        cy.login('student');
+        cy.visit(['sample', 'polls']);
+        cy.contains('Poll Cypress Test').siblings(':nth-child(3)').contains('View Poll').click();
+        cy.get('#toggle-histogram-button').click();
+        cy.wait(1000);
```
williamjallen, line 305: As per the comment:
> log into student to verify histogram and answer are released

I think some additional logic needs to happen here to actually check that the histogram and answer are properly released.  Perhaps you can target something on the histogram, such as checking the color of each bar.

The same comment applies throughout the rest of these changes.

site/tests/app/entities/poll/PollTester.php
```diff
@@ -7,9 +7,8 @@
 use app\entities\poll\Response;
 use DateTime;
 use ReflectionProperty;
-use tests\BaseUnitTest;
 
-class PollTester extends BaseUnitTest {
+class PollTester extends \PHPUnit\Framework\TestCase {
```
williamjallen, line 11: What necessitated this change?  It seems like we have a number of tests extending `BaseUnitTest` and a number of tests extending `TestCase`, so it's a plausible change.  I'm just wondering about the logic behind the decision.

site/tests/app/entities/poll/PollTester.php
```diff
@@ -268,4 +270,15 @@ public function testHistogramRelease(): void {
         $this->expectException(\RuntimeException::class);
         $this->my_polls[0]->setReleaseHistogram("aaaaaaaaa");
     }
+    public function testAnswerRelease(): void {
```
williamjallen, line 273: It would be good to add a newline between these functions

williamjallen (changes requested):

site/app/templates/polls/ViewPoll.twig
```diff
@@ -63,10 +63,13 @@
                                 "content" : option.getResponse(),
                                 "style": "display: inline"
                             } only %}
+                            {% if option.isCorrect() and poll.isReleaseAnswer()%}
+                                <p id='correctTag' style="color: #0eed0e; margin-left: 0rem ">correct!</p>
```
williamjallen, line 67: I personally think it looks better when it's capitalized personally, but I'm open to other opinions.  This is a suggestion, not a requested change.

```suggestion
                                <p id='correctTag' style="color: #0eed0e; margin-left: 0rem ">Correct!</p>
```

williamjallen, line 67: As per the [style guide](https://submitty.org/developer/software_and_system_design/coding_style_guide/css), this ID should be dash-separated.  e.g., `correct-tag` instead of `correctTag`.

site/cypress/e2e/polls.spec.js
```diff
@@ -279,6 +279,68 @@ describe('Test cases revolving around polls functionality', () => {
...
+
+        // log into student and verify we can see both histogram and answer
+        cy.logout();
+        cy.login('student');
+        cy.visit(['sample', 'polls']);
+        cy.contains('Poll Cypress Test').siblings(':nth-child(3)').contains('View Poll').click();
+        cy.get('#toggle-info-button').should('be.visible');
+        cy.get('#toggle-histogram-button').should('be.visible').click();
+        cy.get('#poll-histogram').should('be.visible');
+        cy.get('#correctTag').should('be.visible');
```
williamjallen, line 324: It would be good to verify both that the tag is visible, but also that it is next to the correct answer.  I think it would probably be fine to do this by searching for the text of the answer in the siblings of this element.

williamjallen (changes requested):
A couple minor changes, and then I think it should be good to go.

site/app/templates/polls/ViewPoll.twig
hansongu123: I think this tag would fit more uniformly if it's either all capped or not capitalized at all - I'm leaving it as it is for now

hansongu123: Done

williamjallen: What about capitalizing the entire word and putting it in a green pill like the course materials page has?

As it is now, the text has poor contrast in light mode.  The pills on the course materials page were designed to fix this exact contrast issue.

image

site/cypress/e2e/polls.spec.js
hansongu123: The final way is reserved there but reduced to 500 to ensure the test runs smoothly.

site/app/templates/polls/ViewPoll.twig
```diff
@@ -64,7 +64,7 @@
                                 "style": "display: inline"
                             } only %}
                             {% if option.isCorrect() and poll.isReleaseAnswer()%}
-                                <p id='correct-tag' style="color: #0eed0e; margin-left: 0rem ">correct!</p>
+                                <span id='correct-tag'>&#10004;</span>
```
williamjallen, line 67: This should be replaced with an icon from [Font Awesome](https://fontawesome.com/icons/check?s=solid&f=classic) since this checkmark character isn't guaranteed to look the same on all platforms.  You can find numerous examples of FA icons across the codebase to reference for this.

site/public/css/colors.css
```diff
@@ -14,6 +14,7 @@
     --focus-warning-brown: #ad7c37;
     --hover-warning-brown: #9c5b00;
     --active-warning-brown: #b58a4e;
+    --default-green: #018201;
```
williamjallen, line 17: This change seems reasonable to me but I'd like to get @jenaybarela's approval before adding a new color.  This is related to the corresponding change on line 322.

site/public/css/polls.css
```diff
@@ -101,3 +101,10 @@ form {
     flex-grow: 100;
     margin-left: 0.5em;
 }
+
+#correct-tag {
+    font-family: Copperplate;
+    font-weight: 900;
```
williamjallen, line 106: These changes will be unnecessary after switching to a FA icon.

williamjallen (changes requested):

site/app/templates/polls/ViewPoll.twig
```diff
@@ -64,7 +64,7 @@
                                 "style": "display: inline"
                             } only %}
                             {% if option.isCorrect() and poll.isReleaseAnswer()%}
-                                <span id='correct-tag'>&#10004;</span>
+                                <i class="fa-solid fa-square-check" id="correct-tag"></i>
```
williamjallen, line 67: Oops, I didn't see this earlier.  In HTML, the `id` attribute should be unique.  You should use a `class` instead, since there's a possibility that there might be multiple correct answers.  Many browsers will render things properly if there are duplicate IDs, but that is technically undefined behavior as far as I am aware.

This might be why you weren't able to count them in Cypress.

williamjallen (changes requested):

PR #8912, opened 2023-02-16
[Testing:Developer] Remove dependabot title fix workflow

MasterOdin:

We have a workflow that is triggered whenever dependabot opens a PR that is intended to fix the title of PRs prefixed with `[DevDependency]:` with `[DevDependency] ` (note the space). This was to get around an issue in Dependabot where it only allowed prefixes to be 15 characters long (see https://github.com/dependabot/dependabot-core/issues/2934#issue-777358410), and so we could not use the `[DevDependency] ` prefix ourselves since it's 16 characters long.

### What is the new behavior?

Per https://github.com/dependabot/dependabot-core/issues/2934#issuecomment-1416715959, the prefix length has been raised. Additionally, #8788 updated the prefix to leverage the increased length already, and as can be seen in recent PRs that use the prefix (e.g. #8880), the PR is created with the expected prefix we want, and this workflow did nothing.


williamjallen (approved):
Looks good to me.  After this PR, we can probably archive the [repo](https://github.com/Submitty/action-dependabot-title-fixer) for this action.

PR #8914, opened 2023-02-16
[Bugfix:Submission] PHP Error when seeking team

### LINKED ISSUE #8911: Exception thrown when seeking a team
**Describe the bug**
As described in https://github.com/Submitty/Submitty/pull/8857#issuecomment-1430153644, clicking the "Seek Team/Partner" button results in an error.  The direct cause of the error is this function, which does not handle cases where no rows are returned gracefully.  https://github.com/Submitty/Submitty/blob/4d6120e98b3ef3be3a4f48573a55c72081bc5781/site/app/libraries/database/DatabaseQueries.php#L3446-L3456

**Expected behavior**
No error should appear.

**To Reproduce**
1. Go to a gradeable with teams enabled (for example: http://localhost:1511/courses/s23/sample/gradeable/open_team_homework/team)
2. Click "Seek Team/Partner" button
3. See error

skara9:

PHP error thrown when seeking team and there is no seek message in the database.
Resolves #8911 

### What is the new behavior?
Seek message query method returns null as expected when no message found in the database.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -3452,7 +3452,9 @@ VALUES(?, ?, ?, ?, 0, 0, 0, 0, ?)",
             [$g_id, $user_id]
         );
 
-        return $this->course_db->rows()[0]['message'];
+        $rows = $this->course_db->rows();
+        if (!$rows) return null;
+        return $rows[0]['message'];
     }
 
     /**
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -3452,11 +3452,11 @@ VALUES(?, ?, ?, ?, 0, 0, 0, 0, ?)",
             [$g_id, $user_id]
         );
 
-        $rows = $this->course_db->rows();
-        if (count($rows) === 0) {
+        $row = $this->course_db->row();
+        if (count($row) === 0) {
             return null;
         }
-        return $rows[0]['message'];
+        return $row['message'];
     }
 
     /**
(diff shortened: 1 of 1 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -3452,7 +3452,11 @@ public function getSeekMessageByUserId($g_id, $user_id) {
             [$g_id, $user_id]
         );
 
-        return $this->course_db->rows()[0]['message'];
+        $rows = $this->course_db->rows();
+        if (empty($rows)) {
```
williamjallen, line 3456:
```suggestion
        if (count($rows) === 0) {
```

In general, it is preferable to explicitly count the number of rows instead of using `empty()` since `empty()` can come with some unintended results.  See [this](https://www.php.net/manual/en/function.empty.php) documentation page for further details.  It probably wouldn't make a huge difference here, but it's just one more place we'll have to clean up whenever I finish some of my linting PRs.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -3452,7 +3452,11 @@ public function getSeekMessageByUserId($g_id, $user_id) {
...
+            return null;
+        }
+        return $rows[0]['message'];
```
williamjallen, line 3459:
```suggestion
        $row = $this->course_db->row();
        if (empty($rows)) {
            return null;
        }
        return $row['message'];
```

If we expect there to be one row all the time, we can use `row()` instead of `rows()[0]`.

williamjallen (changes requested):

williamjallen (approved):
Works as expected.

Issue #8818, opened 2023-02-03
ta grading assignment with discussion forum thread id - auto-open image attachments


bmcutler:
**What problem are you trying to solve with Submitty**
When grading an assignment that has a discussion forum thread, we can use the "Discussion" panel to display all posts within the specified thread(s).  If the student has included an image attachment, it requires the grader to click to load and open and display that image.

**Describe the way you'd like to solve this problem**
Let's auto open all image attachments if the discussion panel is displayed.

<img width="500" alt="Screen Shot 2023-02-02 at 8 35 10 PM" src="https://user-images.githubusercontent.com/5871417/216490711-a288b175-942f-4662-a585-37b5036f7e6a.png">

sonustar:
One possible solution is to modify the existing code for the "Discussion" panel to automatically load and display image attachments when the panel is displayed. It could be done by adding some JavaScript or JQuery code to the page that detects when the "Discussion" panel is opened, and then automatically triggers the loading and display any image attachments within that panel.

Do you allow me to contribute for this problem .

sonustar:
Is there any way that I can talk to the mentor ? if any then please let me. I need to ask a few things

saumyaborwankar:
Hey, @bmcutler I wasn't able to find any linked discussion forum on the assignment grading page. Am I missing something?
Is there a certain homework I'm supposed to be looking at?
I'm logged in as "ta2"

![image](https://user-images.githubusercontent.com/46644513/218383642-ae3565f2-0621-4bc7-9f49-c2acb6b9967b.png)

Ahmed89shaaban:
Want to share all information about mechanical engineering to collect data and books and courses and how to be excellent

PR #8917, opened 2023-02-17
Adds code to auto open attachments

ANS1514:

Looks to fix issue #8818, currently attachments don't open automatically in the TA discussion forum panel.

### What is the new behavior?
The attachments now open automatically in the TA discussion forum panel.

### Other information?

I logged in as an instructor, made a thread on the discussion forum and created a homework. I linked the homework to the discussion forum thread and then went into TA grading to check whether the attachments were opening or not.

### DIFF
--- a/site/app/templates/forum/CreatePost.twig
+++ b/site/app/templates/forum/CreatePost.twig
@@ -109,6 +109,16 @@
 
         <div id="{{ post_attachment.params.well_id }}" class="attachment-well well well-lg"></div>
 
+        {# Checks whether the page is on TA grading or the Discussion forum and opens attachments based off that#}
+        <script>
+            function openAttachments() {
+                if ( window.location.pathname == '/courses/s23/hashtag8471/gradeable/hw01/grading/grade') {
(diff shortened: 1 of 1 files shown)


site/app/templates/forum/CreatePost.twig
```diff
@@ -109,6 +109,16 @@
 
         <div id="{{ post_attachment.params.well_id }}" class="attachment-well well well-lg"></div>
 
+        {# Checks whether the page is on TA grading or the Discussion forum and opens attachments based off that#}
+        <script>
+            function openAttachments() {
+                if ( window.location.pathname == '/courses/s23/hashtag8471/gradeable/hw01/grading/grade') {
```
cjreed121, line 115: This is hard-coded to only your course on your development environment. It would probably also be nicer to put this logic in the TA grading JavaScript code instead.

cjreed121 (changes requested):
It looks like you included Tutorial as a submodule. We definitely don't want that so please remove it.

PR #8922, opened 2023-02-17
[Documentation:System] Install system success message

EdenBendheim:

* Updated the Submitty.org website to clarify this new message should be shown if the installation has succeeded. 

The current behavior is that there are many "Done" messages in the install_system.sh. It isn't very clear for new students installing submitty.

### What is the new behavior?
Changed the final message so it is more clear when the installation has finished.

### Other information?
I tested it when I re-made my VM to test #8874.

### DIFF
--- a/.setup/install_system.sh
+++ b/.setup/install_system.sh
@@ -877,5 +877,23 @@ if [ ${WORKER} == 0 ]; then
 fi
 
 
-echo "Done."
+echo "
+#####################################################################
+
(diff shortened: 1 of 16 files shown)


williamjallen (changes requested):
It looks like you checked in some extra files accidentally here.  Otherwise, changes look good.

EdenBendheim:
I did indeed! I accidentally had files here that were different, I guess I maybe didn't branch this off of main or something 😬.

PR #8923, opened 2023-02-17
[Documentation:System] Added a Installation Success message

EdenBendheim:

Updated documentation on Submitty.org in order to include this final install message.

There were many 'Done exit 0' throughout the installation. It was confusing whether it had really ended or not.

### What is the new behavior?
Added a large success message so users installing submitty for the first time would be sure that the installation had completed.

### Other information?
Tested while remaking my VM in order to test Mac using QEMU for #8874.

### DIFF
--- a/.setup/install_system.sh
+++ b/.setup/install_system.sh
@@ -847,5 +847,23 @@ if [ ${WORKER} == 0 ]; then
 fi
 
 
-echo "Done."
+echo "
+#####################################################################
+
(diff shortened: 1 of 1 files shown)


shailpatels (approved):
I think this is a good idea, its hard to tell when the VM is done especially with the amount of output

If you could link to the PR that updates the documentation for this PR that would be helpful

@bmcutler I think this is good to go if you're happy with the message

PR #8927, opened 2023-02-18
[Feature:TAGrading] Option to show all version on TA page

### LINKED ISSUE #8899: More options for TA Grading regrade drop down menu

From the TA grading page, the grader can queue the students submission for regrade.
They do this from the "Autograding" panel.
Currently, they have 2 choices, the currently marked active version and all versions.

Screen Shot 2023-02-13 at 7 48 52 PM

NOTE:  The TA can view & modify which version is "active".
They do this from the "Student Information Panel":

Screen Shot 2023-02-13 at 7 49 07 PM

Sometimes it is useful to regrade one of the non-active versions (if there was an autograding misconfiguration), but important to not regrade all versions (if there are alot of versions and/or the autograding for the gradeable is computationally intensive).

Let's modify the regrade dropdown (first image above) to list all versions of the students submission
with the active one clearly labeled (similar to the dropdown in the student information panel.

NOTE: selecting to regrade all versions should still be available as a simple one click option.

NOTE: currently the TA can switch the version then regrade active, then switch it back.
But this is error prone, potentially confusing if the student happens to visits the page in the middle of this process.
Also, there is a permanent record of the version change in the version change history log.

Suryansh5545:

Fixes #8899 Currently the TA grading page show only two options of regrading the version "All" and "Active".

### What is the new behavior?
All versions are now shown on the TA grading page and the active versions is shown with the text "GRADE THIS VERSION"
TAGradingPage


williamjallen:
I generally like this UI a lot.  My only small nitpick is that I think the text of the button should change to reflect the item selected.  I think it should read "Regrade: <version info here>", with the default option being the default version.

Just my thoughts here--I'm definitely open to different opinions if someone else disagrees.

BairongZhang1 (approved):

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1172,7 +1172,7 @@ public function renderGradingPanelHeader(bool $isPeerPanel, bool $isStudentInfoP
 
     /**
      * Render the Autograding Testcases panel
-     * @param AutoGradedVersion $version_instance
+     * @param AutoGradedVersion|null $version_instance
```
DefinitiveAbove, line 1175: It may be good to add the typing in the function definition also if it's being updated here (for a nullable AutoGradedVersion it would be `AutoGradedVersion? $version_instance`)

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1192,6 +1192,24 @@ public function renderAutogradingPanel($version_instance, bool $show_hidden_case
             $id = $graded_gradeable->getSubmitter()->getId();
         }
 
+        // TODO: this is duplicated in Homework View
+        $version_data = array_map(function (AutoGradedVersion $version) use ($gradeable) {
+            return [
+                'points' => $version->getNonHiddenPoints(),
+                'days_late' => $gradeable->isStudentSubmit() && $gradeable->hasDueDate() ? $version->getDaysLate() : 0
```
DefinitiveAbove, line 1199: `days_late` isn't being used anywhere so it can be removed here

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -1200,6 +1218,10 @@ public function renderAutogradingPanel($version_instance, bool $show_hidden_case
             "is_vcs" => $gradeable->isVcs(),
             "gradeable_id" => $gradeable->getId(),
             "user_id" => $id,
+            "display_version" => $display_version,
```
DefinitiveAbove, line 1221: `display_version` isn't used so this & the logic above should be removed

site/ts/twig/grading/electronic/AutogradingPanel.ts
```diff
@@ -53,5 +60,11 @@ $(() => {
     $('#autograding-results-regrade-all').on('click', () => {
         regrade(0, HIGHEST_VERSION, GRADEABLE_ID, USER_ID);
     });
+    $('.dropdown-item').on('click', () => {
```
DefinitiveAbove, line 63: It may be better to change this to a different class with a class named uniquely to the autograding panel regrade dropdown menu buttons (since the dropdown-item class is used for styling on other pages)

site/app/templates/grading/electronic/AutogradingPanel.twig
```diff
@@ -12,6 +12,9 @@
                 </button>
                 <div class="dropdown-menu">
                     <button id="autograding-results-regrade-all" class="dropdown-item">Regrade All Versions</button>
+                    {% for version in versions %}
+                    <button  id={{loop.index}} class="dropdown-item">Version #{{ loop.index }} Score: {{ version.points }} / {{ total_points }} {%- if loop.index == active_version -%} &nbsp;&nbsp;&nbsp;GRADE THIS VERSION {%- endif -%}</button>
```
DefinitiveAbove, line 16: The button here (and all buttons in this file) should have the `type="button"` attribute on them.

The id attribute of these buttons should be changed to something more specific (i.e. `autograding-results-regrade-{id}`).

The dropdown menu also currently sticks on the panel when you have it open and scroll past the `Regrade Active Version` button - while this isn't an issue itself, it becomes a problem when trying to regrade assignments that students have submitted many times (15 on my computer, possibly less on mobile) - the menu becomes taller than the panel and anything outside of the panel height cannot be clicked on, even if you scroll down since the dropdown menu is static on the panel. One way to solve this could be to make it so the dropdown menu is no longer sticky on the panel.

DefinitiveAbove (changes requested):
Looks great and functionality works as expected - some changes requested

Suryansh5545:
@DefinitiveAbove Sorry for the unnecessary line of code in my pull request. I have now made the necessary changes and updated the pull request accordingly. Additionally, I tried making the dropdown menu non-sticky, but for some reason, it wasn't working as expected. Therefore, I opted to make it scrollable instead.
Capture

Eceptonsu (approved):
Code looks good and works, there is only a small portion of duplication and can be addressed later.

PR #8929, opened 2023-02-19
[Bugfix:Autograding] Wrapping Description Text Properly

### LINKED ISSUE #7978: Autograding summary text doesn't wrap properly
**Describe the bug**
When autograding test case descriptions are very long and consist of just letter characters, the test doesn't wrap properly, as in test#5 in the screenshot. It should wrap like in test#4, where the string also has dash characters.
image

04vmatsibekker:

Fixes #7978 . Originally test case descriptions wouldn't wrap correctly.

### What is the new behavior?
Now the description wraps correctly, and the sentence will move to the next line.

### DIFF
--- a/site/public/css/server.css
+++ b/site/public/css/server.css
@@ -691,10 +691,6 @@ td > a.active-sort {
     margin-bottom: 10px;
 }
 
-code {
-    word-wrap: break-word;
-}
-
 .box-block > code,
 .box-block > pre {
     padding: 10px;
--- a/site/public/css/server.css
+++ b/site/public/css/server.css
@@ -691,6 +691,10 @@ td > a.active-sort {
     margin-bottom: 10px;
 }
 
+code {
+    word-wrap: break-word;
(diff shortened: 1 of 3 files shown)


04vmatsibekker:
image

site/public/css/server.css
```diff
@@ -691,6 +691,10 @@ td > a.active-sort {
     margin-bottom: 10px;
 }
 
+code {
+    word-wrap: break-word;
+}
```
williamjallen, line 696: This is not necessarily in the right spot.  As it is now, you're affecting all `<code>` elements everywhere on the system, which may be unintended.  You should find a more specific selector to target a specific element on this page.

williamjallen (changes requested):

site/public/css/server.css
04vmatsibekker: Alright, Thank You!

Issue #8084, opened 2022-06-22
Allow for Instructors to Hide Course Materials Folders


MBon1:
**What problem are you trying to solve with Submitty**
- When an instructor edits a Course Materials folder to be hidden from the students or all files in the folder are hidden, the files are hidden from the students but the folder is still visible to the students.  

**Describe the way you'd like to solve this problem**
- If a folder is not released, marked as hidden, or if all contents within folder are not visible nor released, the folder should not be visible to the students.  

**Additional context**
 - Instructor View
    ![image](https://user-images.githubusercontent.com/20442222/175143917-75d982f4-f99c-437a-8913-15054b252ecd.png)
 - Current Student View
    ![image](https://user-images.githubusercontent.com/20442222/175143533-dc69df8b-12d5-4a9f-915b-c53ce90c08ae.png)
 - Desired Student View
    ![image](https://user-images.githubusercontent.com/20442222/175144091-2fea2d36-32a4-4149-91ad-6601c6a9518d.png)

PR #8956, opened 2023-02-24
[Bugfix:CourseMaterials] Hide course materials folders

BairongZhang1:
Fix issue #8084 

When an instructor edits a Course Materials folder to be hidden from the students or all files in the folder are hidden, the files are hidden from the students but the folder is still visible to the students.

### What is the new behavior?
Student won't be able to see the folder if all the course materials in that folder is hidden.

### Other information?
1. Login as instructor and set open course material panel, set all files in one folder hidden.
2. Login as student and check if the folder is hidden.


shailpatels (changes requested):
When I log in as a student (aphacker) I see the following error on this branch

```

app\exceptions\OutputException (Code: 0) thrown in /usr/local/submitty/site/app/libraries/Output.php (Line 371) by:


Message:
Key "isHiddenFromStudents" for array with keys "b" does not exist. in /usr/local/submitty/site/app/templates/course/CourseMaterials.twig:159

Stack Trace:
#0 /usr/local/submitty/site/app/views/course/CourseMaterialsView.php(152): app\libraries\Output->renderTwigTemplate()
#1 unknown file(unknown line): app\views\course\CourseMaterialsView->listCourseMaterials()
#2 /usr/local/submitty/site/app/libraries/Output.php(233): call_user_func_array()
#3 /usr/local/submitty/site/app/libraries/Output.php(202): app\libraries\Output->renderTemplate()
#4 /usr/local/submitty/site/app/libraries/response/WebResponse.php(43): app\libraries\Output->renderOutput()
#5 /usr/local/submitty/site/public/index.php(142): app\libraries\response\WebResponse->render()
```

My course materials from the instructor look like this:

image

williamjallen (changes requested):

BairongZhang1:
The current version will check five layers of folder of course materials to see if all files in it set hidden, or the folder will be shown. IF there are more than five layers in this folder, the folder will be shown even all files in it set hidden.

site/app/templates/course/CourseMaterials.twig
```diff
@@ -155,8 +155,47 @@
 
 {% macro display_folder(name, course_materials, links, user_group, user_section, id, indent, display_file_url, folder_path, directory_priorities, material_list, date_format, folder_ids, seen) %}
     {% set showFolder = false %}
-    {% import _self as macros %}
-    {{ macros.hideFolderCheck(course_materials,indent,user_group,user_section)}}
+    {% for course_material in course_materials %}
```
williamjallen, line 158: The indentation here is way too deep.  Some of this logic should either be moved to a macro (like it was originally), or simplified to have less indentation.  In addition, the logic is repetitive and doesn't handle cases beyond 5 levels of nestedness.

williamjallen (changes requested):

EdenBendheim:
The code seems to work fine from my test, but the readability of the code is a bit of an issue. I don't know too much about twig, but is there no way to do this in its own loop? Or recursively somehow?

EdenBendheim (commented):
Review in the comment above ^

site/app/templates/course/CourseMaterials.twig
BairongZhang1: I didn't find a way to create a recursive macro, since macro can't return anything and the result of checking hidden material will lost after every loop. I also not sure how to simplify this.

BairongZhang1:
> EdenBendheim
Since macro can't return anything, the result of checking hidden material will lose after every loop. So I can't apply a recursive structure on it. I'm also new to twig and I am not sure I can create a functionable recursive macro.

Issue #8770, opened 2023-01-25
Add "course_id" to the database


bmcutler:
**What problem are you trying to solve with Submitty**
Each course section has an ID number with the registrar.  At RPI it is a 5 digit number that is unique for the semester and we call it a CRN.  We would like to store this information in the database and have it available in Rainbow Grades.

**Describe the way you'd like to solve this problem**

Step 1:

In the main Submitty database, let's add a column named 'course_id' to the courses_registration_sections table.
And in the course database, let's add a column named 'course_id' to the sections_registration table.

This is an optional piece of information, and different schools might have different types (numbers vs strings) or other requirements (uniqueness, etc.). So at this time, let's not attempt to put these requirements on the database.  Let's set the type of this field as 'character varying' (string).  It can have any length (# of characters) and is allowed to be empty string.

This field should be automatically synced between the Submitty database & each course database via database triggers.

Step 2:

When student data is exported for Rainbow Grades, in addition to the field, e.g., "registration_section" : "5", let's add "course_id" : "12345".

Step 3:

Let's display the section ID on the Manage Sections page for the course in the top section under "Manage Registration Sections", add a column to the "Current Registration Section" table.

Step 4:

It would be nice to add a feature to set the course id for a section from the Manage Sections page.  This information should update both the main database and the course database via triggers from step 1.  I expect we at RPI will mostly be inputting the section data via sysadmin scripts with data from the registrar.

bmcutler:
@JManion32  (or other develoeprs)

If you would like to work on Step 4 above, that is the only part of this issue that is still unfinished.

(The rainbow grades work is a different issue/future work)

PR #8957, opened 2023-02-24
[Feature:InstructorUI] section_id exported in student data

AdamTuhacek:

course_section_id is not included in student grade reports

### What is the new behavior?

Adds course_section_id to the student json files created through generating grade summaries. 

### Other information?

Step 2 of #8770 

============================================================
Updated - aphacker json file now contains course_section_id
Screen Shot 2023-06-13 at 12 32 53 PM


williamjallen:
The PR number should be removed from the PR title for consistency with our PR title formatting rules

AdamTuhacek:
> The PR number should be removed from the PR title for consistency with our PR title formatting rules

Oops, yes, you're right

site/app/models/User.php
```diff
@@ -451,6 +455,10 @@ public function setRegistrationSection($section) {
         $this->registration_section = ($section !== null) ? $section : null;
     }
 
+    public function CourseSectionId($course_section_id) {
+        $this->course_section_id = ($course_section_id !== null) ? $course_section_id : null;
+    }
+
```
tkoz0, line 461: This should probably be called `setCourseSectionId`. The null check also is not necessary, the same thing happens without it. Please ask one of the people who is more familiar with PHP about the setter and getter functions not needing to be explicitly defined, I remember seeing them come from some comment at the top of the file.

site/app/models/gradeable/Submitter.php
```diff
@@ -109,6 +109,14 @@ public function getRegistrationSection() {
...
     }
 
+    /**
+     * Get the course_section_id for the submitter
+     * @return string
+     */
+    public function getCourseSectionId() {
+        return $this->team_or_user->getCourseSectionId();
+    }
+
```
tkoz0, line 119: Again, please check on whether the function needs to be explicitly defined.

tkoz0 (changes requested):
I left comments about the setter and getter functions in the PHP code. Please discuss them with someone who is more familiar with PHP.

AdamTuhacek:
> I left comments about the setter and getter functions in the PHP code. Please discuss them with someone who is more familiar with PHP.

Ok, thanks for the feedback. I'll do some research and maybe discuss with William later.

site/app/models/User.php
AdamTuhacek: Yeah, I renamed it and added it as a "magic" function, and it works the same.

site/app/models/gradeable/Submitter.php
AdamTuhacek: This seems like it could be done without being explicitly defined, but it looks like none of the functions in submitter.php are this way. I think for consistency it's maybe better to leave it here.

site/app/models/gradeable/Submitter.php
```diff
@@ -109,6 +109,14 @@ public function getRegistrationSection() {
         return $this->team_or_user->getRegistrationSection();
     }
 
+    /**
+     * Get the course_section_id for the submitter
+     * @return string
+     */
+    public function getCourseSectionId() {
```
williamjallen, line 116: It's better to use native PHP type declarations when possible because PHP will enforce them in the development environment.

```suggestion
    /**
     * Get the course_section_id for the submitter
     */
    public function getCourseSectionId(): string {
```

williamjallen (changes requested):
One minor tweak, and then I think this should be ready to go.

site/app/models/User.php
```diff
@@ -35,6 +35,8 @@
  * @method void setGroup(integer $group)
  * @method void setRegistrationType(string $type)
  * @method string getRegistrationSection()
+ * @method string setCourseSectionId()
+ * @method void setCourseSectionId(string $Id)
```
williamjallen, line 39: Is the first one supposed to be `getCourseSectionId()`?

AdamTuhacek: Yes, you're right

bmcutler:
@AdamTuhacek @tkoz0 @williamjallen 
What is the status of this PR?  Why is a draft?
I thought we had already merged this...  it should be a priority to finish

AdamTuhacek:
> @AdamTuhacek @tkoz0 @williamjallen What is the status of this PR? Why is a draft? I thought we had already merged this... it should be a priority to finish

This was almost merged but I found out it wasn't working as intended, I'm getting the database errors now. I think I have a better idea of why, so I can try and take another look and get it (hopefully) fixed in the next couple of days.

bmcutler:
@ziesski please take over this PR

Youssef-Chip (approved):
The code looks clean and the functionality works as expected.

bmcutler (changes requested):

bmcutler (approved):

PR #8958, opened 2023-02-25
[Testing:TAGrading] Added Test case for simple-grading

### LINKED ISSUE #8895: Write Cypress test cases for Numeric/Text gradeable
See comment on PR #8867 

"Given that this is a relatively low-traffic page which seems to get broken on a regular basis, it might be good to write some basic Cypress tests for this page. In particular, this action of navigating both the UI and the text within each of the cells seems prone to breaking in the future. It should be relatively straightforward to write a basic test for this."

Suryansh5545:

Fixes #8895 Currently there is no test case for the simple grading page.

### What is the new behavior?
Added test case simple_grading.spec.js
Capture
 to test grading lab and grading test page.


williamjallen (approved):
Thanks for this contribution!

These test cases might be a little fragile if future edits are made to the page, but I don't see a good alternative.

In the future, it would be good to also test the other functionality on this page, but that can be done in a separate PR.

PR #8982, opened 2023-02-27
[Documentation:System] Install system success message

EdenBendheim:
Please check if the PR fulfills these requirements:

Updated the Submitty.org website to clarify this new message should be shown if the installation has succeeded.

What is the current behavior?

The current behavior is that there are many "Done" messages in the install_system.sh. It isn't very clear for new students installing submitty.

What is the new behavior?

Changed the final message so it is more clear when the installation has finished.

Other information?

I tested it when I re-made my VM to test 
https://github.com/Submitty/Submitty/pull/8874.

This is a re-make of a different PR that accidentally included more files

### DIFF
--- a/.setup/install_system.sh
+++ b/.setup/install_system.sh
@@ -847,5 +847,23 @@ if [ ${WORKER} == 0 ]; then
 fi
 
 
-echo "Done."
+echo "
+ #####################################################################
+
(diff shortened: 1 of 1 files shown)


williamjallen:
Is this change not a duplicate of https://github.com/Submitty/Submitty/pull/8923, which was already merged?

PR #8984, opened 2023-02-28
[Feature:TAGrading] Auto open Images

ANS1514:

Images don't auto open in the TA grading panel

### What is the new behavior?
Images will auto open in the TA grading panel and will stay open even if you refresh the page.

### Other information?

I made a course on my local machine and created a discussion forum thread and made a post with an attachment. Then I made a homework linked to that thread, made a submission then went to grade it and use the discussion forum panel in the TA grader to see if the image opened.


cjreed121 (changes requested):
Do you think you could instead put this logic into [DiscussionForumPanel.twig](https://github.com/Submitty/Submitty/blob/main/site/app/templates/grading/electronic/DiscussionForumPanel.twig)? That is much more closely tied to where we want this to happen. This would add a lot of extra computation on the forum since it would never apply.

site/app/templates/forum/CreatePost.twig
```diff
@@ -108,6 +108,7 @@
 
 
         <div id="{{ post_attachment.params.well_id }}" class="attachment-well well well-lg"></div>
+
```
williamjallen, line 111: This change seems to be unrelated and should be reverted

site/app/templates/grading/electronic/DiscussionForumPanel.twig
```diff
@@ -9,11 +9,15 @@
...
-		});
+$('#posts_list').css('min-width', '-webkit-fill-available');
+$(function() {
+	generateCodeMirrorBlocks(document);
+});
+
+$(document).ready(function() {
+    loadAllInlineImages(true);
+
+});
```
williamjallen, line 21: All of this code should probably be moved to the relevant `.js` file.  Perhaps `ta-grading.js`.

williamjallen (changes requested):

site/public/js/forum.js
```diff
@@ -2501,3 +2501,4 @@ function pinAnnouncement(thread_id, type, csrf_token) {
         });
     }
 }
+
```
williamjallen, line 2504: This line is unrelated to the rest of the changes and should be reverted.

williamjallen (changes requested):
One quick change to prevent potential merge conflicts with other PRs and I'll approve.

williamjallen (approved):

PR #8987, opened 2023-03-01
Links (anchor tag) replaced by buttons (button tag) for non-navigation buttons 

charvimehradu:
Pull request for Issue#7819

A number of places throughout Submitty use anchor tags and are stylized as buttons but do not perform navigation and instead rely on an onclick event. This can confuse screen readers which might interpret the element as a link.

### What is the new behavior?
The button tag is now used for elements that do an action without changing the page. The button tag also automatically fires onclick when a key is pressed on the keyboard, eliminating the need for the key to click class to be used everywhere.

### Other Information
-The original issue has some articles listed for the support of the above statements.
-The PR#7863 had no progress since Oct 30, 2022 and had some problems unfixed.
-Thank you @immortalcodes for mentioning the specific lines of code that contained the bugs.

### DIFF
--- a/site/app/templates/admin/users/StudentList.twig
+++ b/site/app/templates/admin/users/StudentList.twig
@@ -56,7 +56,7 @@
                                 <form method="post" action="{{ view_grades_url }}">
                                     <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
                                     <input type="hidden" name="student_id" value="{{ student.getId() }}"/>
-                                    <a onclick="$(this).parent().submit();"><i class="fas fa-chart-line"></i></a>
+                                    <button onclick="$(this).parent().submit();"><i class="fas fa-chart-line"></i></button>
                                 </form>
                             </td>
                             <td>{{ student.getRegistrationType() }}</td>
--- a/site/public/js/server.js
+++ b/site/public/js/server.js
@@ -1165,7 +1165,7 @@ function refreshOnResponseOverriddenGrades(json) {
     }
     else {
         json['data']['users'].forEach(function(elem){
-            let delete_button = "<a onclick=\"deleteOverriddenGrades('" + elem['user_id'] + "', '" + json['data']['gradeable_id'] + "');\"><i class='fas fa-trash'></i></a>"
+            let delete_button = "<button onclick=\"deleteOverriddenGrades('" + elem['user_id'] + "', '" + json['data']['gradeable_id'] + "');\"><i class='fas fa-trash'></i></button>"
             let bits = ['<tr><td class="align-left">' + elem['user_id'], elem['user_givenname'], elem['user_familyname'], elem['marks'], elem['comment'], delete_button + '</td></tr>'];
             $('#grade-override-table').append(bits.join('</td><td class="align-left">'));
         });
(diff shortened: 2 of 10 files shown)


MasterOdin:
Thanks for the PR @charvimehradu! As part of getting this ready to merge, would you be able to rework the title so that it matches our format? See (4) on https://submitty.org/developer/getting_started/make_a_pull_request for more details.

charvimehradu:
thankyou for guiding me @MasterOdin I hope now it looks better!

AdamTuhacek (approved):
Looks good, code changes make sense and no problems were encountered in my testing.

site/app/templates/admin/users/StudentList.twig
mahi-pas, line 1: This overrides the theme of the button, should be fixed (I am currently trying to find one)

mahi-pas (changes requested):

mahi-pas:
For anyone in the future, if buttons overwrite the theme of the enclosed item, add
`style="background: none;border:none;all:unset;"`

example:
`<button style="background: none;border:none;all:unset;"></button>`

site/app/templates/admin/users/StudentList.twig
mahi-pas, line 1: Edit and delete button should also remove <a>

mahi-pas (changes requested):
Won't merge due to this issue:
The previous behavior for links was that it changed the cursor type to a hand and also had a hover effect that changed the color (below image)
image

With the change, it doesn't change the cursor or have a hover functionality.
image

site/public/js/server.js
```diff
@@ -1217,7 +1217,7 @@ function refreshOnResponseOverriddenGrades(json) {
     }
     else {
         json['data']['users'].forEach(function(elem){
-            let delete_button = "<a onclick=\"deleteOverriddenGrades('" + elem['user_id'] + "', '" + json['data']['gradeable_id'] + "');\"><i class='fas fa-trash'></i></a>"
+            let delete_button = "<button onclick=\"deleteOverriddenGrades('" + elem['user_id'] + "', '" + json['data']['gradeable_id'] + "');\" style='background: none;border:none;all:unset;'><i class='fas fa-trash'></i></button>"
```
prestoncarman, line 1220: Use "no-style-invisible-button" instead of the style tag.

site/public/css/global.css
```diff
@@ -189,3 +189,9 @@ footer .footer-separator {
 #loading-bar-percentage {
     margin-left: 10px;
 }
+
+#no-style-invisible-button {
```
prestoncarman, line 193: ID's should only be used once in a HTML file. They represent a unique item. I think this should be a class. See <https://css-tricks.com/the-difference-between-id-and-class/> for more details.

site/app/templates/admin/users/StudentList.twig
```diff
@@ -58,7 +58,7 @@
                                 <form method="post" action="{{ view_grades_url }}">
                                     <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
                                     <input type="hidden" name="student_id" value="{{ student.getId() }}"/>
-                                    <a onclick="$(this).parent().submit();"><i class="fas fa-chart-line"></i></a>
+                                    <button onclick="$(this).parent().submit();" id="no-style-invisible-button"><i class="fas fa-chart-line"></i></button>
```
prestoncarman, line 61: Switch from id to class. See comment in the CSS file.

prestoncarman (changes requested):

PR #8991, opened 2023-03-02
[Bugfix:CourseMaterials] Fixes Folders don't close properly

### LINKED ISSUE #8986: Folders don't close properly when clicking Open/Close All folders button
**Describe the bug**
In course materials, I noticed that the the folders are displayed as open or closed based on a cookie saved when a folder is open or closed. However, while the Open/Close all folders button temporarily opens and closes the folders, the changes are not saved and refreshing the page undos the action.

**Expected behavior**
Folders stay closed/open when the button is clicked and the page is refreshed

**To Reproduce**
1.  Go to course materials
2. Open and close random folders, and folders withing folders
3.  Click the open/close all folders button
4. Refresh the page, and the folders should revert to how you had them before
5. See error

**Configuration**
 - OS: Windows 11
 - Browser: Chrome

**Screenshots**
Initial config:
image
After open/close all folders button:
image
After refresh:
image

**Additional context**
I would start with checking if the button modifies the cookie, as that is probably the core issue.

Suryansh5545:

Fixes #8986 Currently the folder structure doesn't remain open due to the issue with subdir cookies and folderOpen cookies.

### What is the new behavior?
Added a loop to be called instead so that both subdir and folderOpen cookies are changed. This will make sure both cookies remain or change to same value when the open/close all button is pressed.


site/app/templates/course/CourseMaterials.twig
```diff
@@ -181,6 +181,37 @@
 {% endmacro %}
 
 <script>
+    function folderopencookie(){
+        var fileNames = [];
```
williamjallen, line 185:
```suggestion
        const fileNames = [];
```

site/app/templates/course/CourseMaterials.twig
```diff
@@ -181,6 +181,37 @@
...
+            var c = ca[i];
+            while (c.charAt(0) == ' ') {
+                c = c.substring(1);
+            }
+            if (c.indexOf(name) == 0) {
+                return c.substring(name.length, c.length);
+            }
+        }
+        return "";
+    }
```
williamjallen, line 213: This logic is no longer necessary given that #9314 has been merged.

site/cypress/e2e/courseMaterials.spec.js
```diff
@@ -185,7 +185,7 @@ describe('Test cases revolving around course material uploading and access contr
             cy.get('#submit-materials').click();
         });
 
-        cy.get('#cm-toggle-folders-btn').click();
+        cy.get('[onclick=\'folderopencookie();\']').click();
```
williamjallen, line 188: Why is this change necessary?  It is better to target things by ID or by class, rather than by a bit of nasty HTML?  The same comment applies to the rest of the changes in this file.

williamjallen (changes requested):

PR #8994, opened 2023-03-03
[Bugfix:Submission] Blocked bulk upload when no grade access

### LINKED ISSUE #8802: Bulk upload allowed even when grading is not
**Describe the bug**
A grader can upload files even when they are not assigned to grade the assignment.

**Expected behavior**
When a grader is not assigned to grade an assignment they should not be able to upload 

**To Reproduce**
1. Create a new bulk upload gradeable 
2. Make sure in grader assignment the lowest privileged user is instructor 
3. Open the gradable to ta
4. From the ta account they will now have access to upload even though they are not able to grade

jason490:

Fixes #8802 
When the instructor sets the highest grade access to instructor only, the tas are able to upload bulk gradeables.

### What is the new behavior?
When the highest grading access is set to instructor only, the tas are not able to upload bulk gradeables


site/app/views/NavigationView.php
```diff
@@ -515,7 +515,7 @@ public static function getSubmitButton(Core $core, Gradeable $gradeable, $graded
             if (!$gradeable->hasDueDate()) {
                 $date_text = "";
             }
-            if (!$gradeable->isStudentSubmit() && $core->getUser()->accessFullGrading()) {
+            if (!$gradeable->isStudentSubmit() && $core->getUser()->accessFullGrading() && $core->getUser()->getGroup() <= $gradeable->getMinGradingGroup()) {
```
shailpatels, line 518: The second part of the condition could also be checked with 
```
$this->core->getAccess()->canI("grading.electronic.grade", ["gradeable" => $gradeable]);
```

which is kind of the centralized way of making sure the same check is applied everywhere and runs the same getMinGradingGroup.

shailpatels (approved):
Changes correctly fix the issue as described, can make the change I commented above, but if this needs to be merged earlier I think it should be good

site/app/views/NavigationView.php
jason490: I tried implementing that code by replacing the condition with the condition you showed me, and it seems like it created a php exception.

shailpatels: Oh it was probably due to the function being static and using the `$this` keyword

I think if you remove that and just start with $core like how the other functions do it, it should work?

Otherwise could you post the exception message

jason490: Yep, that solved the issue

site/app/views/submission/HomeworkView.php
```diff
@@ -476,6 +476,7 @@ private function renderSubmitBox(Gradeable $gradeable, $graded_gradeable, $versi
             'show_no_late_submission_warning' => !$gradeable->isLateSubmissionAllowed() && ($gradeable->isSubmissionClosed() && !$gradeable->isTaGradeReleased()),
             // This is only used as a placeholder, so the who loads this page is the 'user' unless the
             //  client overrides the user
+            'bulk_upload_access' => $this->core->getUser()->getGroup() <= $gradeable->getMinGradingGroup(),
```
shailpatels, line 479: Should this check also be updated to `$this->core->getAccess()->canI("grading.electronic.grade", ["gradeable" => $gradeable]);`

shailpatels (changes requested):

shailpatels (approved):
Retested after changes and works correctly

PR #8995, opened 2023-03-03
[Bugfix:Autograding] Wrapping Description Text Properly #8929

### LINKED ISSUE #7978: Autograding summary text doesn't wrap properly

04vmatsibekker:

Fixes #7978 . Originally test case descriptions wouldn't wrap correctly.

Additional Fix: Originally there was a weird indent for the description text. 

### What is the new behavior?
Now the description wraps correctly, and the sentence will move to the next line.
Have the indent continue for the rest of the description.

### DIFF
--- a/site/public/css/reset.css
+++ b/site/public/css/reset.css
@@ -25,9 +25,6 @@ time, mark, audio, video {
     background:transparent;
 }
 
-code{
-    word-wrap: break-word;
-}
 
 article,aside,details,figcaption,figure,
 footer,header,hgroup,menu,nav,section {
--- a/site/public/css/reset.css
+++ b/site/public/css/reset.css
@@ -25,6 +25,10 @@ time, mark, audio, video {
     background:transparent;
 }
 
+code{
+    word-wrap: break-word;
+}
+
 article,aside,details,figcaption,figure,
 footer,header,hgroup,menu,nav,section {
     display:block;
--- a/more_autograding_examples/cpp_cats/config/config.json
+++ b/more_autograding_examples/cpp_cats/config/config.json
@@ -44,7 +44,7 @@
         },
         {
             "title" : "Test of by word comparison",
-            "details" : "./byWord.exe input.txt output.txt",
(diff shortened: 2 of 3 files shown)


williamjallen:
Can you please add a before/after screenshots to your PR description?

hansongu123:
@williamjallen It will be looking like this:
image
image
Although, I only tested out the functionality in the general cases (in the original issue, Shelley only mentioned about a long string put in "details" in a test case will break out from the text box), it worked pretty well.  Good work!

site/public/css/reset.css
```diff
@@ -25,6 +25,7 @@ time, mark, audio, video {
     background:transparent;
 }
 
+
```
bmcutler, line 28: please revert this edit (just inserting a blank line)

hansongu123:
The additional indentation fix here looked very nice too. 
I see that the same edit is being applied to many files related to the integration tests: is this the only way you can test this? I don't think they are testing the fix you added since that's only about the display issue, and they seem like they were actually designed for different styles of inputs already (I can see that e2e is failed for a similar reason). Maybe it would be helpful to make a new test case about the changes and remove it after the PR is considered stable enough.

hansongu123 (changes requested):

williamjallen (commented):
I'm a little bit confused about why the tests were modified here.  This is a simple UI change and these test changes do nothing to test the correctness of the UI.

The code changes themselves seem reasonable.

hansongu123 (commented):
We had a simple exchange on this and decided maybe it is a good idea to remove the edits you added in those tests in order to avoid confusion since they practically have no function. William also mentioned that this kind of UI change is best tested using a Cypress-Snapshot plugin, which is not set up by Submitty at the moment. I don't think the indentation and where the text breaks necessarily are going to break things - but sometimes it could be acting a little funky. I'd say revert the edits you applied on the tests and resolve the new merge conflict, then I'll check on it again!

bmcutler:
I requested the changes to the test :)

hansongu123 (approved):

PR #9020, opened 2023-03-10
[Bugfix:Submission] Notebook Copy In Multiple Choice

### LINKED ISSUE #8345: Markdown multiple choice duplicates
**Describe the bug**
If a notebook/markdown gradeable has duplicate values, the gradeable will build fine and produce unexpected behavior to the student submitting. If 2 options had the same value, the website may pick the incorrect value to set when loading the page.

**Expected behavior**
The gradeable should just not build successfully because not all values were unique.

**To Reproduce**
1. Create a notebook/markdown gradeable config with a multiple choice but have duplicate values.
2. Build the gradeable on Submitty
3. Pick the higher order option and submit
4. See that after the refresh, the option changed to the lower one

See comments for a full example

04vmatsibekker:

Fixes #8345. If there is a multiple choice where two of the values are the same, the notebook gradeable will be built.

### What is the new behavior?
The Notebook will not be built and will display an error message.

### DIFF
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -1804,7 +1804,7 @@ nlohmann::json ValidateANotebook(const nlohmann::json& notebook, const nlohmann:
           for (auto it = choices.begin(); it != choices.end(); ++it){
               auto it_next = it;
               for (++it_next; it_next != choices.end(); ++it_next){
-                assert (it->value() != it_next->value()); 
+                assert (it.value() != it_next.value()); 
               }
               // Reassign the value of this iteration to choice
               nlohmann::json choice = it.value();
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -1805,7 +1805,7 @@ nlohmann::json ValidateANotebook(const nlohmann::json& notebook, const nlohmann:
               auto it_next = it;
               for (++it_next; it_next != choices.end(); ++it_next){
                 if (it.value() == it_next.value()) 
-                  std::cout  << "ERROR: " << it.value() " == " << it_next.value() << std::endl;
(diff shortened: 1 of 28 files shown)


grading/load_config_json.cpp
```diff
@@ -1813,6 +1813,16 @@ nlohmann::json ValidateANotebook(const nlohmann::json& notebook, const nlohmann:
               assert(value != "");
               assert(description != "");
 
+              auto it_next = it;
+              for (++it_next; it_next != choices.end(); ++it_next){
```
williamjallen, line 1814: I believe this variable declaration can be moved into the for-loop declaration

grading/load_config_json.cpp
```diff
@@ -1813,6 +1813,16 @@ nlohmann::json ValidateANotebook(const nlohmann::json& notebook, const nlohmann:
...
+
+                nlohmann::json choice_next = it_next.value();
+                std::string value_next = choice_next.value("value", "");
+                if (value == value_next)
+                  std::cout  << "ERROR: " << it.value() << " == " << it_next.value() << std::endl;
```
williamjallen, line 1819:
```suggestion
                if (value == value_next) {
                  std::cout  << "ERROR: " << it.value() << " == " << it_next.value() << std::endl;
                }
```

williamjallen, line 1819: This error message isn't very descriptive.  It would be better to indicate both the problem and a suggested resolution.

williamjallen (changes requested):

04vmatsibekker:
@williamjallen
Does the new text that is outputted as the error message look better?

williamjallen:
The code changes seem reasonable to me.  I'll leave the feature/functionality review to someone more knowledgeable about this part of the codebase.

Eceptonsu (approved):
Tested on my machine, works well.

PR #9022, opened 2023-03-13
[Refactor:System] Switch to highlight.js

cjreed121:

We currently use Commonmark Highlighter to allow for code highlighting in our markdown. While this makes it super easy to handle all in the backend, there is an unseen flaw. Behind the scenes the library maintains 100+ language json files which get included every time twig is loaded (pretty much every request). This takes a toll on the system to constantly read in all these files. There is also no caching within this library and they aren't php files so OPCache can't help as far as I know. The main advantage to use this library is that there is less work put on the front end.

### What is the new behavior?
Commonmark Highlighter has been removed and replaced with the original highlight.js library. It is just as functional but it's possible you may notice the code highlighting lazy load in rather than be immediate. Most code editors have a lazy load feel to their syntax highlighting anyways. Performance benefits below.

### Other information?
**Backend Performance**
Before:
Running a profile on the home page showed Output::loadTwig() taking about 25% of the overall request time.

After:
Same profile only showed about 5% of request time.

I can do performance tests on the front end if we want but I don't imagine this will add too much.

I also did consider having the backend only read certain JSON files and support caching but it would require extending/modifying quite a few library classes. That'd make it tough for future maintainability.


ANS1514 (approved):

bmcutler:
@cjreed121 @ANS1514 
Can one of you please resolve merge conflicts?  Thanks

bmcutler:
@cjreed121 @ANS1514
Can one of you please resolve merge conflicts? Thanks

williamjallen (approved):
Changes seem reasonable and work as expected.  The merge conflicts need to be addressed and then I think this can be merged.

bmcutler:
@cjreed121 conflicts again

PR #9033, opened 2023-03-13
[Bugfix:TAGrading] Creating New Annotations

Eceptonsu:
While attempting to add new text and strokes, the current PDF annotation does not function properly. Due to changes in #8549, the newly created annotations will not be stored or displayed.

### What is the new behavior?
The PR fixes the issue mentioned above. The newly created annotations will be appropriately stored and displayed when viewing the modified PDFs.


BairongZhang1 (approved):
Work as expected, annotations are able to be saved.

site/app/controllers/pdf/PDFController.php
```diff
@@ -14,22 +14,36 @@ public function __construct(Core $core) {
...
+     */
+    public function getPath(string $file_path, string $id, bool $real): string {
+        $path = "";
+        $file_path_parts = explode("/", $file_path);
+
+        if ($real) {
+            if (str_contains($file_path, "..")) {
+                return "INVALID FILE PATH";
+            }
         }
```
williamjallen, line 29: It would be better to combine these into a single `if`-statement

site/app/controllers/pdf/PDFController.php
```diff
@@ -14,22 +14,36 @@ public function __construct(Core $core) {
...
-        $file_path_parts = explode("/", $file_path);
         for ($index = 1; $index < count($file_path_parts); $index++) {
             if ($index === 9) {
-                $real_path .= "/" . $id;
+                if ($real) {
+                    $path .= "/" . $id;
+                }
+                else {
+                    $user_id = $file_path_parts[$index];
+                    $anon_ids = $this->core->getQueries()->getSubmitterIdFromAnonId($user_id, $id);
```
williamjallen, line 38: This is going to result in repetitive queries which slow down the page load time.  It would be better to move this outside the loop.

williamjallen (changes requested):

site/app/controllers/pdf/PDFController.php
Eceptonsu: I think this will only be checked once since its in the condition for index === 9

williamjallen (approved):
Code looks good and functions as expected.

PR #9035, opened 2023-03-14
[Bugfix:TAGrading] PDF Download Correct File

Eceptonsu:
If a submission contains multiple PDFs, only the first opened PDF will be downloaded when downloading a PDF in the TA grading interface. That is, if one switches to another PDF and tries to download the file, the downloaded file will be the previously opened one.

### Other information?
Changes in [PDFAnnotateEmbedded.js](https://github.com/Submitty/Submitty/compare/main...muti_pdf_download#diff-5decbeb402a10fd07dbe4e3abfc8ac4931564dfa942400a4d6b2641dff137b07) remove a console error and do not affect the functionality of PDF download.


BairongZhang1 (approved):
Work properly, currently able to download correct pdf.

PR #9038, opened 2023-03-14
[Feature:SubminiPolls] Create Tomorrow's Polls Section in polls

### LINKED ISSUE #8500: Separate Instructor "Future Polls" category into "Tomorrow's Polls" and "Future Polls"

yanliw123:
add a new tomorrow's section in polls
tomorrow's poll now separate from future polls

closes #8500 
test by create a new poll and set the release date to tomorrow.


hansongu123 (commented):
Although editing an existing poll's release date can now move it into the "Tomorrow's Polls" column, I'm no longer able to create a new poll in any combinations because it throws an exception every time after clicking on "Add Poll" button in the instructor's UI.

hansongu123:
Clicking on the "View Poll" button as an instructor would take me out to the course selection page instead of showing the poll page, which persists after I repeat the action.

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -142,6 +142,69 @@
...
+                    <th scope="col"></th>
+                    <th scope="col">Date Released</th>
+                    <th scope="col">Responses</th>
+                    <th scope="col"></th>
+                </tr>
+            </thead>
+
+            <tbody>
+                {% for poll in future_polls %}
+                    {% if poll.getReleaseDate() | date("Y-m-d") == ('now'|date_modify("+1 day")|date('Y-m-d')) %}
```
hansongu123, line 174: I would suggest instead of picking out the polls that are dated tomorrow from the future polls, it is more consistent creating an extra array in the poll controller to organize just tomorrow's polls. This would help the situation when we just want to manage those polls alone although they still appear in the future polls' section.

hansongu123 (commented):
I can't comment on the unedited part of the code, but around line 64 where it says toggling the "older polls" and "future polls" dropdown, you need to also test functionality of the "tomorrow's polls" drop down since it is a newly added component too.

site/cypress/e2e/polls.spec.js
```diff
@@ -116,8 +116,34 @@ describe('Test cases revolving around polls functionality', () => {
...
         cy.get('#chartContainer').contains('1 * 3 = 4');
-    });
+        cy.go('back');
 
+        //tomorrow's poll test, name poll 5
+        cy.contains('New Poll').click();
+        cy.get('#poll-name').type('Poll 5');
+        cy.get('#poll-question').type('What is your favorite class?');
+        cy.get('#poll-date').clear({force: true});
+        const today = new Date(Date.now());
```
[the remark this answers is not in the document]
williamjallen: I don't have any problems with this choice personally.  The goal here is overall readability, and this seems to be clear and concise.  It is indeed inconsistent with other places in these tests as you mention, but that doesn't necessarily mean that these changes are bad.

site/cypress/e2e/polls.spec.js
```diff
@@ -116,8 +116,34 @@ describe('Test cases revolving around polls functionality', () => {
...
+        cy.get('h1').click();
+
+        // submit and verify on main polls page, poll should be in tomorrow table
+        cy.get('#poll-form-submit').click();
+        cy.url().should('include', 'sample/polls');
+        cy.contains('Poll 5');
+        cy.get('#tomorrow-table').contains('Poll 5');
+
+        // delete the poll
+        cy.contains('Poll 5').siblings(':nth-child(2)').click();
```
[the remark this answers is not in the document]
yanliw123: should I move all tomorrow's poll test cases to the bottom section where it says "should verify making, editing, and deleting poll works as expected"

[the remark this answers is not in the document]
yanliw123: > In general, each test case should clean up after itself. Each test case should delete any polls it creates.
> 
> That said, I think the broader issue here is that these changes are placed in the wrong spot. This test case is specifically testing the sample polls which are generated by the sample course generation script. I believe that all of these changes can be moved to a completely new test case since they seem to be completely independent of everything else.

should I move all tomorrow's poll test cases to the bottom section where it says "should verify making, editing, and deleting poll works as expected"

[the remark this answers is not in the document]
williamjallen: All of the test code you wrote here is completely independent of everything else in the test case you added it to.  Given that it's entirely independent (creating, modifying, and deleting a poll for testing purposes), this code should be moved to a new test case.

site/cypress/e2e/polls.spec.js
```diff
@@ -116,8 +116,8 @@ describe('Test cases revolving around polls functionality', () => {
         cy.get('#chartContainer').contains('2 * 3 = 6');
         cy.get('#chartContainer').contains('8 / 2 = 3');
         cy.get('#chartContainer').contains('1 * 3 = 4');
-    });
 
+    });
```
hansongu123, line 120:
```suggestion
    });
    
```

site/cypress/e2e/polls.spec.js
```diff
@@ -360,7 +360,49 @@ describe('Test cases revolving around polls functionality', () => {
...
+        // log in from instructor account
+        cy.visit(['sample', 'polls']);
+        cy.login();
+
+        //tomorrow's poll test, name poll 5
+        cy.contains('New Poll').click();
+        cy.get('#poll-name').type('Poll 5');
+        cy.get('#poll-question').type('What is your favorite class?');
+        cy.get('#poll-date').clear({force: true});
+        const tomorrow = new Date(Date.now());
```
hansongu123, line 375:
```suggestion
        const tomorrow = new Date();
```

site/cypress/e2e/polls.spec.js
```diff
@@ -360,7 +360,49 @@ describe('Test cases revolving around polls functionality', () => {
...
+        cy.login('student');
+        cy.visit(['sample', 'polls']);
+        cy.get('poll 5').should('not.exist');
+
+        //log into instructor, and delete poll
+        cy.logout();
+        cy.login();
+        cy.visit(['sample', 'polls']);
+
+        // delete the poll 5
```
hansongu123, line 400:
```suggestion
```

hansongu123 (changes requested):
Around line 15, add a test for clicking on the new dropdown window and its visibility like what was done for the polls in the other sections, especially in the "Future's Polls" section.
Potentially, you can add some test to see what would happen if you move a polls in the "Tomorrow's Polls" section out to other sections, and vice versa.

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -145,6 +145,69 @@
...
+                        </td>
+                        <td>
+                                <a class="fas fa-trash black-btn" tabindex="0" href='javascript:newDeletePollForm("{{ poll.getId() }}", "{{ poll.getName() }}", "{{ base_url }}");' title="delete poll {{ poll.getName() }}" aria-label="delete poll {{ poll.getId() }}"></a>
+                        </td>
+                        <td style="text-align: left;">
+                            {{ poll.getName() }}
+                        </td>
+                        <td> {{ poll.getReleaseDate() | date("Y-m-d") }} </td>
+
+                        <td id="poll_{{ poll.getId() }}_responses"> {{ poll.getUserResponses()|length }} </td>
```
williamjallen, line 189: This line is the source of the failing tests.  As the loop containing this code iterates over each poll, this line will attempt to load all of the user responses to each poll, and then count them.  This is a bad approach, because it creates a separate query each time the loop iterates, and gathers unnecessary data from the database anyway.  A better option would be to mimic the same line in the "today's polls" section by only querying the number of answers to each poll.

williamjallen: To reproduce the duplicate query warning, you just have to put multiple tests in the "tomorrow's polls" section.

hansongu123, line 189:
```suggestion
                        <td id="poll_{{ poll.getId() }}_responses"> {{ poll.getId() not in response_counts|keys ? 0 : response_counts[poll.getId()] }} </td>
```

hansongu123 (changes requested):

hansongu123 (approved):

williamjallen (changes requested):
Changes mostly look good, except for the "tomorrow's polls" dropdown arrow which doesn't work as expected.  When you load the page and open the dropdown, reloading the page doesn't preserve its state.  I also think that the dropdown should just start in the open state like the rest of the dropdowns.

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -145,6 +145,69 @@
     <br/>
     <hr>
 
+    <div class="dropdown-wrapper">
+        <div id="tomorrow-table-dropdown" onclick="updateDropdownStates($(this).find('.dropdown-arrow').hasClass('down'), 'tomorrow_polls_dropdown', '{{ base_url }}')" class="dropdown-bar">
+            <h2> Tomorrow's Polls </h2>
+            <div class="dropdown-btn">
+                <i class="dropdown-arrow{{dropdown_states['future'] ? ' down' : ' right'}} fa fa-chevron-right"></i>
```
williamjallen, line 152: I believe this line is the source of the problem.  The "tomorrow's polls" dropdown matches the state of the "future polls" dropdown.

williamjallen (changes requested):

site/app/templates/polls/AllPollsPageInstructor.twig
hansongu123: Hard to say if it is an intended feature by the old PR or not. Anyways, I added one more dropdown status so now the "tomorrow's" dropdown doesn't listen to the status of other dropdowns, and made it open by default just like "Today's Polls". I brought up the order of "Tomorrow's Polls" to be directly under "Today's" to make it more accessible to the user. 

Now the order of the dropdowns goes like this: 
Today's => Tomorrow's => Old => Future
I'd think it makes sense to swap the order of "Old Polls" and "Future Polls" dropdowns as well.

williamjallen (changes requested):
It might be a good idea to add a Cypress test case which tries opening and closing all of the dropdowns, ensuring that they maintain their state after reloading the page.

hansongu123:
> It might be a good idea to add a Cypress test case which tries opening and closing all of the dropdowns, ensuring that they maintain their state after reloading the page.

I think that is already tested by checking the polls in the expanded dropdowns. The polls inside them wouldn't be considered "visible" if the dropdowns were closed. And I think creating, editing or deleting polls, alongside of logging out and logging out, and then logging back in would be considered as refreshing the page as well. At least these are true for all the dropdowns that are not open by default. That's the same reason I ignored the testing of "Old Polls" dropdown table in the new test case - because the test case above it also tested this functionality.

I could add a few lines in the same test case about checking the opening and closing of the "Today's Polls" and "Tomorrow's Polls" dropdown bars though and a few checks about refreshing if you want... Over all this shouldn't be too much of a worry since it is blended in in other test cases. The issue appeared in your last comment only occurred because "Tomorrow's" dropdown bar's open status binds to that of "Future's" dropdown bar, and it should be just intended by the original PR. The change I made was straightforward.

williamjallen (approved):

PR #9043, opened 2023-03-15
[Bugfix:Submission] Fixes gradeable relocking

### LINKED ISSUE #9037: Relocking the gradeable doesn't work
**Describe the bug**
Relocking the unlocked gradable doesn't work

**Expected behavior**
Relocking the unlocked gradable is supposed to lock the gradable

**To Reproduce**
1. Create a new gradable
2. Under the submissions/auto-grading, Lock the gradable by choosing one of the dropdowns and set the auto grader points to be a reasonable value  
3. Unlock the gradable
4. Try relocking the gradable
5. See error

**Configuration**
 - OS: Linux
 - Browser safari

**Screenshots**
image

Suryansh5545:

Fixes #9037 The gradeable was not being re locked. it was though to be only limited to that but since the page change only sended depend_on or depend_on_points at a time the locking function wasn't working. 

### What is the new behavior?
Both the data values are send at the same time and the depend_on_point also get checked for null value in the controller.
Below is a gif of the implementation working

RelockingFixed

### DIFF
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -1133,7 +1133,7 @@ class AdminGradeableController extends AbstractController {
                             $gradeable->setDependsOnPoints(0);
                         }
                         else {
-                            if ($depends_on_points < 0 || ($depends_on_points > $temp_gradeable->getDependsOnPoints() && $temp_gradeable->getDependsOnPoints() != null)) {
+                            if ($depends_on_points < 0 || ($depends_on_points > $temp_gradeable->getDependsOnPoints() && $temp_gradeable->getDependsOnPoints() !== null)) {
                                 $errors['depends_on_points'] = "Invalid depends on points!";
                             }
                         }
--- a/site/app/controllers/admin/AdminGradeableController.php
+++ b/site/app/controllers/admin/AdminGradeableController.php
@@ -1133,7 +1133,7 @@ class AdminGradeableController extends AbstractController {
                             $gradeable->setDependsOnPoints(0);
                         }
                         else {
-                            if ($depends_on_points < 0 || $depends_on_points > $temp_gradeable->getDependsOnPoints()) {
+                            if ($depends_on_points < 0 || ($depends_on_points > $temp_gradeable->getDependsOnPoints() && $temp_gradeable->getDependsOnPoints() != null)) {
                                 $errors['depends_on_points'] = "Invalid depends on points!";
                             }
                         }
--- a/site/public/js/admin-gradeable-updates.js
+++ b/site/public/js/admin-gradeable-updates.js
@@ -223,7 +223,7 @@ $(document).ready(function () {
             }
             data[val.name] = $(val).val();
         };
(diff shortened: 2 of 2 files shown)


site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1133,7 +1133,7 @@ private function updateGradeable(Gradeable $gradeable, $details) {
                             $gradeable->setDependsOnPoints(0);
                         }
                         else {
-                            if ($depends_on_points < 0 || $depends_on_points > $temp_gradeable->getDependsOnPoints()) {
+                            if ($depends_on_points < 0 || ($depends_on_points > $temp_gradeable->getDependsOnPoints() && $temp_gradeable->getDependsOnPoints() != null)) {
```
williamjallen, line 1136: It is better to use the strict comparison operators in PHP.

```suggestion
                            if ($depends_on_points < 0 || ($depends_on_points > $temp_gradeable->getDependsOnPoints() && $temp_gradeable->getDependsOnPoints() !== null)) {
```

site/public/js/admin-gradeable-updates.js
```diff
@@ -223,6 +223,12 @@ $(document).ready(function () {
             }
             data[val.name] = $(val).val();
         };
+        if (data['depends_on'] != null) {
+            data['depends_on_points'] = points;
+        }
+        else if (data['depends_on_points'] != null) {
+            data['depends_on'] = gradeable;
+        }
```
williamjallen, line 231: same here.
```suggestion
        if (data['depends_on'] !== null) {
            data['depends_on_points'] == points;
        }
        else if (data['depends_on_points'] != null) {
            data['depends_on'] = gradeable;
        }
```

williamjallen (changes requested):

ch3nnyj (approved):
I've tested your branch and verified that the issue no longer appears when re-locking a gradeable. I see no issues in your code as well. Maybe William might find something but overall it looks like you fixed the issue.

site/public/js/admin-gradeable-updates.js
```diff
@@ -223,6 +223,12 @@ $(document).ready(function () {
             }
             data[val.name] = $(val).val();
         };
+        if (data['depends_on'] !== null) {
+            data['depends_on_points'] == points;
```
williamjallen, line 227: Is the `==` operator here a typo?

williamjallen (changes requested):

site/public/js/admin-gradeable-updates.js
Suryansh5545: It was in the suggested change in last conversation and the functionality still worked so I changed both. I will revert it.

Suryansh5545: @williamjallen it was in the last suggested change and didn't break the functionality. So I changed both from != and = to !== and == . I have changed it back now.

ch3nnyj (approved):
Functionality didn't change so I must have missed that type on my end when reviewing. I should be all set now to merge.

Issue #9045, opened 2023-03-16
"Set Folder Release Date" Buttons for Course Material Folders Overlap Visually 


hansongu123:
**What problem are you trying to solve with Submitty**
In the Course Material page within the instructor's UI, "Set Folder Release Date" buttons for course material folders overlap visually and create a blue lump when there are multiple.
![image](https://user-images.githubusercontent.com/123511202/225755230-7fa69c5a-2b5e-436c-8dbc-ded89bc0bb0b.png)

**Describe the way you'd like to solve this problem**
Find a way to separate out the buttons from each other.

**Describe any potential alternatives you'd tried to solve the problem**
By adding reasonable amount of space in between those folder items (slight) or reducing the size of background for those buttons just in this page.

**Additional context**
If adding more space distort the current structure of showing folders or make too big of a difference visually when there are many items, abort and jump on to other suggestions.

SwagatikaBehera:
Here buttons looks a bit bigger we can reduce the size and adjust border-radius or we can add margin-bottom around the listed items div or using gap property of display:flex to listed div we could fix it. I would love to fix this problem @hansongu123.

PR #9046, opened 2023-03-17
[UI/UX:CourseMaterials] Fix buttons overlap visually

BairongZhang1:
Fix issue #9045

In the Course Material page within the instructor's UI, "Set Folder Release Date" buttons for course material folders overlap visually and create a blue lump when there are multiple.
bef

### What is the new behavior?
aft

### Other information?
None


site/app/templates/course/CourseMaterials.twig
```diff
@@ -167,7 +167,7 @@
                 {% if user_group == 1 %}
                     <a class="key_to_click" tabindex="0" onclick='newEditCourseMaterialsFolderForm(this)' data-id="{{ folder_ids[folder_path] }}" data-priority="{{ directory_priorities[folder_path] }}"> <i class="fas fa-pencil-alt black-btn" aria-hidden="true" style="font-size: 16px; margin: 5px;"></i></a>
                     <a class="key_to_click" tabindex="0" onclick='newDeleteCourseMaterialForm("{{ folder_ids[folder_path]  }}", "{{ name }}");'> <i class="fas fa-trash" aria-hidden="true" style="font-size: 16px; margin: 5px;"></i></a>
-                    <a onclick='setFolderRelease("{{ folder_path }}","","{{ id }}","{{ folder_ids[folder_path] }}")' class="btn btn-primary key_to_click" tabindex="0">Set Folder Release Date</a>
+                    <a onclick='setFolderRelease("{{ folder_path }}","","{{ id }}","{{ folder_ids[folder_path] }}")' class="btn btn-primary key_to_click" tabindex="0" style="line-height: 15px; height: 30px; margin: 1px">Set Folder Release Date</a>
```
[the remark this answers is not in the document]
BairongZhang1: I moved the change of style to bootstrap.css.

site/public/css/bootstrap.css
williamjallen, line 1: We generally don't want to make changes to this file since it is originally from a 3rd-party [library](https://getbootstrap.com).  `course-materials.css` seems like a more appropriate place for these changes.

site/public/css/bootstrap.css
```diff
@@ -201,6 +201,12 @@ fieldset[disabled] .btn-primary.focus {
     background-color: var(--btn-default-white);
 }
 
+.btn-course-material-folder-set-time {
+    line-height: 15px;
+    height: 30px;
+    margin: 1px;
+}
```
williamjallen, line 208: I believe that just adding a little bit of margin on the bottom of each row would be more appropriate than adjusting the button itself.

williamjallen (changes requested):

yanliw123 (changes requested):
As William said we should make changes in the course-materials.css file since this is mainly a UI/display issue. I looked at the bottons' class name, it is btn.btn-primary.key_to_click. So the change should be something looks like
.btn.btn-primary.key_to_click{
    margin: 1.5px;
}
you can try to add this in course-materials.css file, and will probably solve the issue.

yanliw123 (approved):

site/public/css/course-materials.css
```diff
@@ -44,3 +44,7 @@ a.popout-item > .fas:hover {
 .delete-course-material-message {
     font-weight: bold;
 }
+
+a.btn-primary {
+    margin: 1.5px;
+}
```
williamjallen, line 50: This is going to target *all* buttons inside links.  That's probably not a good thing if we have a button somewhere else.  It would be better to give these buttons a class and target that class with CSS.

williamjallen (changes requested):

site/public/css/course-materials.css
```diff
@@ -44,3 +44,9 @@ a.popout-item > .fas:hover {
 .delete-course-material-message {
     font-weight: bold;
 }
+
+.btn-course-material-folder-set-time {
+    line-height: 15px;
+    height: 30px;
```
williamjallen, line 50: Are these two lines necessary?  Why are we setting the height of the button?

site/app/templates/course/CourseMaterials.twig
```diff
@@ -167,7 +167,7 @@
                 {% if user_group == 1 %}
                     <a class="key_to_click" tabindex="0" onclick='newEditCourseMaterialsFolderForm(this)' data-id="{{ folder_ids[folder_path] }}" data-priority="{{ directory_priorities[folder_path] }}"> <i class="fas fa-pencil-alt black-btn" aria-hidden="true" style="font-size: 16px; margin: 5px;"></i></a>
                     <a class="key_to_click" tabindex="0" onclick='newDeleteCourseMaterialForm("{{ folder_ids[folder_path]  }}", "{{ name }}");'> <i class="fas fa-trash" aria-hidden="true" style="font-size: 16px; margin: 5px;"></i></a>
-                    <a onclick='setFolderRelease("{{ folder_path }}","","{{ id }}","{{ folder_ids[folder_path] }}")' class="btn btn-primary key_to_click" tabindex="0">Set Folder Release Date</a>
+                    <a onclick='setFolderRelease("{{ folder_path }}","","{{ id }}","{{ folder_ids[folder_path] }}")' class="btn btn-primary btn btn-primary btn-course-material-folder-set-time" tabindex="0">Set Folder Release Date</a>
```
williamjallen, line 170: That's a pretty long class name...

williamjallen (changes requested):
These changes are a little better but I still think it's the wrong approach.  As it is now, you're changing the padding on the button, not the row as a whole.  The padding on the buttons should remain the same to ensure all buttons across the system look the same, and you should add padding to the entire row to separate the rows a bit.

image

williamjallen (approved):
These changes look good to me.  My only comment is that `0.8px` of padding is an almost trivial amount.  I'll leave the issue of the proper amount of padding up for further discussion.

As of now, the UI looks like this:
image

PR #9047, opened 2023-03-17
[Feature:System] Recreate submissions script

tkoz0:

There is no automated way to copy a gradeable to another place for testing.

### What is the new behavior?
This PR adds a script that can upload student submissions to a gradeable automatically, simplifying the process of copying submissions to a new gradeable for testing.

### Other information?
In order to test this, see the instructions at the top of the file `.setup/bin/recreate_submissions.py`. The steps are
1. Download submissions for a gradeable by going to the grading index
2. Create a new gradeable (if it is in a new course, then use the CSV upload to add the students from the original course to the new course)
3. Configure the gradeable and make 1 submission to it, probably as instructor
4. Run the command as described in the instructions


.setup/bin/recreate_submissions.py
```diff
@@ -0,0 +1,294 @@
...
+        func_add_submission: typing.Callable[[str,str,int,str],None],
+        func_update_active: typing.Callable[[str,str,int],None]):
+    '''
+    Create a new submission as a student
+    semester,course,gradeable = specify gradeable to make submission to
+    user = user to submit as
+    data = path to directory containing the files for submission
+    '''
+    gradeable_path = os.path.join(ARG_DATA_DIR,'courses',semester,course,
+        'submissions',gradeable)
```
williamjallen, line 110: It would be better to use [pathlib](https://docs.python.org/3/library/pathlib.html) instead of `os.path` throughout this script.

.setup/bin/recreate_submissions.py
```diff
@@ -0,0 +1,294 @@
...
+    if os.path.exists(os.path.join(user_path,UASF_NAME)):
+        with open(os.path.join(user_path,UASF_NAME),'r') as uasf:
+            UAS = json.load(uasf)
+    else: # initialize empty
+        UAS = {
+            'active_version': 0,
+            'history': []
+        }
+    highest_version = 0 if len(UAS['history']) == 0 else max(obj['version']
+        for obj in UAS['history'])
```
williamjallen, line 126: Given that this statement has to be broken into two lines anyway, it would be better to avoid an inline if-statement here.

.setup/bin/recreate_submissions.py
```diff
@@ -0,0 +1,294 @@
...
+    dbconn_course.close()
+    # initiate batch regrade
+    print(f'ADDING SUBMISSIONS TO GRADING QUEUE')
+    regrade_py = os.path.join(ARG_INSTALL_DIR,'bin','regrade.py')
+    gradeable_path = os.path.join(ARG_DATA_DIR,'courses',
+        ARG_SEMESTER,ARG_COURSE,'submissions',ARG_GRADEABLE)
+    os.system(f'{regrade_py} {gradeable_path}')
+
+if __name__ == '__main__':
+    if getpass.getuser() != 'root':
```
williamjallen, line 291: You could possibly use `os.geteuid() != 0` here instead to prevent an import just for this one command.

.setup/bin/recreate_submissions.py
```diff
@@ -0,0 +1,294 @@
...
+        print('Please run this as root, exiting...')
+        quit()
```
williamjallen, line 293: Why not restructure this to only call `main()` if the user has root access instead of exiting if the user doesn't?

williamjallen (changes requested):
All of the files in `.setup/bin` are currently ignored by our Python linter.  Since this is a new file, it should meet our Python code formatting requirements.  It would be good to explicitly ignore all of the necessary individual paths in `.flake8` instead of the whole directory so that this gets linted.

.setup/bin/recreate_submissions.py
tkoz0: What makes pathlib better?

.setup/bin/recreate_submissions.py
```diff
@@ -288,7 +289,8 @@ def update_egver(g_id: str, user_id: str, active_version: int):
     os.system(f'{regrade_py} {gradeable_path}')
 
 if __name__ == '__main__':
-    if getpass.getuser() != 'root':
+    if os.getuid() == 0:
+        main()
+    else:
         print('Please run this as root, exiting...')
         quit()
```
williamjallen, line 296: According to the [docs](https://docs.python.org/3/library/constants.html#constants-added-by-the-site-module), `quit()` should not be used outside the interpreter.  Better discussion [here](https://stackoverflow.com/questions/19747371/python-exit-commands-why-so-many-and-when-should-each-be-used).

Why do you need an exit command at all?  Why not just let the program end?

williamjallen (changes requested):

.setup/bin/recreate_submissions.py
tkoz0: You're right, it is unnecessary.

Eceptonsu (approved):
The script looks good in my opnion, and the recreate submission functionality works on my machine.

tkoz0:
@williamjallen Could you please review this again when you have a chance?

williamjallen (approved):

Issue #8434, opened 2022-09-24
Last name v Last initial for Grade Inquiries


bmcutler:
**Describe the bug**
Similar to #8431 #8432 and #8433...

**Expected behavior**
Grade inquiry post should be

Preferredfirstname Preferredlastname for instructors & full access graders

Preferredfirstname Last initial for limited access graders

NOTE:  limited access graders are anonymous for grading, but if they are interactive with grade inquiries, then we will intentionally show their names.

kevinochoa8266:
Where can I find this bug?

snehalchaudhari98:
Hi, if this issue is still open, I would love to work on this.

PR #8608, opened 2022-11-15
[Feature:TAGrading] Grade inquiries display name

jssunray34:
### What is the new behavior?

For grade inquiries, the display name is "Preferredfirstname Preferredlastname" for instructors and full access graders and is "Preferredfirstname Last initial" for limited access graders.

Closes #8434

PR #9048, opened 2023-03-17
[Feature:TAGrading] Grade inquiries display name - Additions

04vmatsibekker:
What is the new behavior?
For grade inquiries, the display name is "Preferredfirstname Preferredlastname" for instructors and full access graders and is "Preferredfirstname Last initial" for limited access graders.

Closes https://github.com/Submitty/Submitty/issues/8434


williamjallen (changes requested):

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -813,9 +813,25 @@ public function removeNotificationsPost($post_id) {
...
+    public function isInstructorOrFullAccessGraderPost(string $author_id): bool {
+        $this->course_db->query("SELECT user_group FROM users WHERE user_id=?", [$author_id]);
+        return intval($this->course_db->row()['user_group']) <= User::GROUP_FULL_ACCESS_GRADER;
+    }
+
+    public function getAuthorUserGroups(array $author_ids): array {
+        $placeholders = $this->createParamaterList(count($author_ids));
+        $this->course_db->query("SELECT user_id, user_group FROM users WHERE user_id IN {$placeholders}", $author_ids);
+        return $this->course_db->rows();
     }
```
williamjallen, line 835: I believe that all of these functions are unnecessary.  It would be better to use the relevant functions in `site/app/models/User.php`.

williamjallen (changes requested):

ch3nnyj (approved):
Functionality it looks good to me, once you fix what Will suggested I think it should be good to go.

williamjallen:
Is this a duplicate of https://github.com/Submitty/Submitty/pull/8608?

bmcutler:
@williamjallen 
Yes, the plan was to close and replace #8608 since there were many conflicts
But I see you've successfully updated that PR.
Which one should we close?

williamjallen:
> @williamjallen
> Yes, the plan was to close and replace #8608 since there were many conflicts
> But I see you've successfully updated that PR.
> Which one should we close?

Given how similar the two PRs are, I think it would be best to give preference to the older PR and continue working on #8608.  I'm not certain about the current status of either of these, but I left comments on #8608 which are still relevant, and also relevant to this PR.  It's close, but still needs a bit of work I think.

04vmatsibekker:
The PR looks great and the code is very clean. The only problem is that the name for the instructor and the grader only appear after the site is reloaded.

04vmatsibekker:
I approve the PR

Youssef-Chip (approved):

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -836,6 +836,11 @@ public function getAuthorUserGroups($author_ids) {
         return $this->course_db->rows();
     }
 
+    public function getAuthorUserGroup($author_id) {
+        $this->course_db->query("SELECT user_id, user_group FROM users WHERE user_id = ?", [$author_id]);
+        return $this->course_db->row();
+    }
+
```
williamjallen, line 843: This function is essentially a duplicate of the function above it and should be deleted.

williamjallen (changes requested):

site/app/libraries/database/DatabaseQueries.php
Youssef-Chip: The function above takes as an argument multiple author IDs and returns an array of rows, and I only have one post with one author ID.

site/app/views/submission/HomeworkView.php
```diff
@@ -1186,14 +1187,41 @@ public function showGradeInquiryDiscussion(GradedGradeable $graded_gradeable, bo
...
+
                 // format posts
                 $posts = [];
                 foreach ($grade_inquiry_posts[$grade_inquiry->getId()] as $post) {
                     if (empty($post)) {
                         break;
                     }
-                    $is_staff = $this->core->getQueries()->isStaffPost($post['user_id']);
-                    $name = $this->core->getQueries()->getUserById($post['user_id'])->getDisplayedGivenName();
+                    $is_staff = $queries->isStaffPost($post['user_id']);
```
williamjallen, line 1211: This is going to result in the same performance issues I commented on above.  This one seems relatively simple to convert to a single larger query.

williamjallen (commented):
I noted a few performance issues below.  The grade inquiries code has rather poor performance already, and is a low-traffic page, so optimizing these queries is not necessarily a merge blocker for this PR.  If a decision is made to merge this PR without making the requested changes, an issue detailing the performance issues on the grade inquiries page should be created so we don't forget to come back and refactor.

PR #9050, opened 2023-03-17
[Feature:Submission] Notebook print pdf

### LINKED ISSUE #5646: "Print to PDF" a Notebook gradeable
The instructor has formatted a quiz/test as a Submitty Notebook.  
Can the student download a PDF of the quiz/test?  
The student may wish to print it on paper, fill it out, then scan & upload (depending on what the student wants to be uploaded).

What does each standard browser "print page" to PDF to with a notebook gradeable?  
Can we do better?
Remove the topbar/sidebar/footer, have page breaks between widget items as appropriate... ???

04vmatsibekker:

Fixes #5646.

### What is the new behavior?
 Now when we press print to pdf. We download a pdf file of the quiz.
image

image


mahi-pas:
Print button looks very different in style to the other submitty buttons. Try to make it flat like the others.
image
image

shailpatels (changes requested):
So I think overall this fixes the issue, but there might be additional elements of the page that should be blocked/removed before printing

Chrome and firefox usually add something to the top like the page name, you could look into removing this

image
(https://stackoverflow.com/questions/585254/how-to-remove-the-url-from-the-printing-page)

The time due is shown which I think is ok, but the time remaining would probably want to be removed
image


The site footer and assignment submit buttons are shown
image

Sometimes the pagination is broken
image

That one is a little tougher but there is a CSS element that controls it but you'd probably have to edit the logic to generate the notebook HTML to place them in a container that wraps each question

https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside

shailpatels:
@04vmatsibekker incase you didn't know there is a development option in firefox to simulate print mode, its not 100% close to what the final result looks like which differs between browsers but is pretty close and might be faster then seeing the print preview during development incase you'd prefer to use that during development 

Screenshot from 2023-03-27 23-40-45

shailpatels (approved):
Looks good, the overall result on firefox and chrome match what a student would see on the web interface.

Sometimes a question gets cut in the middle (e.g first set of questions on page1 , the rest on the next) but I don't think we can control this much more and the user will need to mess with the print settings and page size to fix that.

Maybe a solution for that can be made in the future but for now getting the basic functionality should be good

attached is an example PDF generated

[Notebook Basic - Development.pdf](https://github.com/Submitty/Submitty/files/11302809/Notebook.Basic.-.Development.pdf)

shailpatels (commented):
The clear and reset buttons are still shown, not a big deal but if you want to quickly hide them that would be good. But if this PR needs to get merged ok to make an issue to track that separately 

image

image

Eceptonsu (approved):
Code change looks good and works on my machine

PR #7424, opened 2021-12-08
[Feature:TAGrading] Add late submission stats

jportorreal00:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?

-Fixes #4588
-Fixes the grader index "total" column for late submissions.

### What is the new behavior?
- When a user submits an assignment that is too late, the TA should have a red button that says "Too Many Late Days" instead of a blue "Grade" button. 

The "late-box" around grading total of students also works correctly now.

There should also an option to remove all late submissions from the "grading to be done" stat page.

This PR is a recreation of #5631. It now uses the late day cache system created. It should fix #6005 and #4588

### Other information?

Using different late submissions

Screen Shot 2023-07-12 at 10 34 22 AM

PR #9056, opened 2023-03-18
[Refactor:Developer] Delete dead DatabaseQueries.php code

williamjallen:
`DatabaseQueries.php` currently has a number of functions which are no longer used.

### What is the new behavior?
The dead code has been eliminated.


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1486,48 +1470,6 @@ public function getLateDayCacheForUser(string $user_id): array {
...
-
-        // If cache still doesn't exist, the gradeable is not associated with
-        // LateDays OR there has been a computation error
-        if (empty($row)) {
-            return null;
-        }
-
-        return $row;
-    }
-
```
cjreed121, line 1530: This isn't currently used but it looks like it was planned on being used in https://github.com/Submitty/Submitty/pull/7424.

cjreed121 (changes requested):
Most of these should be fine. I found one instance where we might want to keep. Not sure if any other open PRs rely on this code.

PR #9077, opened 2023-03-21
[UI/UX:SubminiPolls] Text Indentation on Poll Options

hansongu123:

image

### What is the new behavior?

1. Increased checkmark icon size.
2. Fix indentation for text surrounding a radio button.
 
image

Also fixed indentation and unnecessary spacing for text following a checkbox.


Yujjio (changes requested):

site/public/css/polls.css
```diff
@@ -43,6 +43,11 @@
     vertical-align: top;
 }
 
+.radio {
```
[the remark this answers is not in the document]
hansongu123: No, that's because the "no response" one doesn't use text indent here... I'm still trying figure out the best way to make it to and solve the problem with formatting the square checkbox buttons - which was more messed up originally when there are multiple lines.

site/public/css/polls.css
```diff
@@ -43,6 +43,11 @@
     vertical-align: top;
 }
 
+.response {
+    margin-left: 24.2px;
+    text-indent: -24.2px;
+}
```
williamjallen, line 49: In general, manually specifying widths is a bad idea.  A better option would be to use flexbox to format the columns of inputs and text

site/public/css/polls.css
```diff
@@ -64,6 +69,11 @@ input[type="radio"][disabled=""]:checked::before {
     background: var(--standard-medium-dark-gray);
 }
 
+input[type="checkbox"] {
+    margin: 0;
+    transform: scale(0.8);
+}
```
williamjallen, line 75: It would probably be better to target a specific class given to the relevant inputs instead of targeting all checkboxes.  If we decide to add another checkbox to the UI somewhere, this will cause confusion.

williamjallen (changes requested):

site/public/css/polls.css
hansongu123: Done. Now it only targets checkboxes whose name = "answer[ ]", which was originally defined in the file.

site/public/css/polls.css
hansongu123: Buttons and text columns are now organized into table to fix the problem of indenting. Thanks for the suggestion and your template!

hansongu123:
Note: new styles are being applied to "no response" option in student's interface. Cypress e2e is failing because now it has to invoke the elements in the table cells instead of directly at elements themselves.

LittleSilver33 (approved):
Looks good to me. I read through the code and tested it several times with different length responses and didn't find any issues.

site/app/templates/polls/ViewPoll.twig
```diff
@@ -22,34 +22,35 @@
                 <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
                 <input type="hidden" name="poll_id" value="{{ poll.getId() }}"/>
                 <fieldset>
+                    <table>
```
Yujjio, line 25: Should the elements inside `<table>` and `</table>` be indented?

Yujjio:
Looks good to me but this branch is currently behind the main branch. Can you update it?

site/app/templates/polls/ViewPoll.twig
hansongu123: They automatically get indented when placed as td in a tr as now the text has its own cell.

site/app/templates/polls/ViewPoll.twig
```diff
@@ -58,21 +59,21 @@
...
-                        </div>
+                                <i class="fa-solid fa-square-check fa-lg correct-tag"></i>
+                            {% endif %}</td>
+                        </tr>
                         {% set index = index + 1 %}
                     {% endfor %}
+                    </table>
                 </fieldset>
                 {% if not user_admin and poll.isOpen() %}
+                    <br/>
```
williamjallen, line 76: In general, relying upon `<br>` for spacing is a bad idea.  It would be better to do this with CSS instead.

site/app/templates/polls/ViewPoll.twig
```diff
@@ -22,34 +22,35 @@
...
-                            >
-                            No response
-                        </label>
-                        <br/>
-                        <br/>
+                            /></td>
+                            <td class="no-response option-text">
+                            no response
+                            </td>
+                        </tr>
```
williamjallen, line 46: I think this should still read "No response".  It's common in English to capitalize the first letter of a phrase or sentence such as in this case.

site/public/css/polls.css
```diff
@@ -64,6 +75,11 @@ input[type="radio"][disabled=""]:checked::before {
     background: var(--standard-medium-dark-gray);
 }
 
+input[type="checkbox"][name="answers[]"] {
+    transform: scale(0.75);
+    margin: 0 1px 0 0;
+}
```
williamjallen, line 81: Is there a better way to target this?  In general, it's better to target things with classes or IDs.

williamjallen (changes requested):

site/app/templates/polls/ViewPoll.twig
hansongu123: right, this should be some leftover code and I haven't deleted yet

site/app/templates/polls/ViewPoll.twig
hansongu123: More of an opinion thing, this definitely is not a complete sentence.
But, yes, chief.

site/public/css/polls.css
hansongu123: This should be good enough. I could add classes for the radio buttons and checkboxes that specifically in the answer field but I thought that wasn't necessary because there's already a name defined.

site/app/templates/polls/ViewPoll.twig
hansongu123: I'm going to abandoned the styling on "No response" if the first letter is capitalized - since now its distracting.

site/public/css/polls.css
hansongu123: Yeah, I eventually did some class manipulation to make the targeting in the new table easier.

site/app/templates/polls/ViewPoll.twig
hansongu123: Targeted the submit button this is for and added padding instead.

site/cypress/e2e/polls.spec.js
```diff
@@ -188,15 +188,17 @@ describe('Test cases revolving around polls functionality', () => {
...
+        cy.get('table').contains('td','No response');
+        cy.get('table > tbody > tr:nth-child(1) > td:nth-child(2)').contains('No response');
+        cy.get('table > tbody > tr:nth-child(1) > td:nth-child(1) > input').should('be.disabled');
+        cy.get('table > tbody > tr:nth-child(1) > td:nth-child(1) > input').should('be.checked');
+        cy.get('table > tbody > tr:nth-child(2) > td:nth-child(2)').contains('Answer 1');
+        cy.get('table > tbody > tr:nth-child(2) > td:nth-child(1) > input').should('be.disabled');
+        cy.get('table > tbody > tr:nth-child(3) > td:nth-child(2)').contains('Answer 2');
+        cy.get('table > tbody > tr:nth-child(3) > td:nth-child(1) > input').should('be.disabled');
+        cy.get('table > tbody > tr:nth-child(4) > td:nth-child(2)').contains('Answer 3');
+        cy.get('table > tbody > tr:nth-child(4) > td:nth-child(1) > input').should('be.disabled');
```
williamjallen, line 201: It might be better to target the table by its class instead.  As it is now, adding another table to this page could potentially break this test.

williamjallen: Apply throughout.

site/public/css/polls.css
```diff
@@ -39,10 +39,28 @@
...
+    vertical-align: middle;
+}
+
+.option-btn input[type="radio"]{
     vertical-align: top;
 }
 
+.option-btn input[type="checkbox"]{
+    transform: scale(0.8);
+}
```
williamjallen, line 52: Is this still needed now that everything is in a table?

site/public/css/polls.css
```diff
@@ -39,10 +39,28 @@
...
+
+.student-submit[type="submit"] {
+    margin-top: 2em;
+}
+
+.option-text {
+    padding-top: 0.55em;
+    padding-bottom: 0.55em;
+    padding-left: 0.25em;
+}
```
williamjallen, line 62: same here

site/public/css/polls.css
hansongu123: Yes, a checkbox in the page is way too large comparing to a radio button. I tried to retain as much of the original size but still keep it inline with the size of a row of text and all that. This should over make it to have a nicer look but does not break things since it is scaled down not up. The functionality should be tested throughout though, using different browsers.

site/public/css/polls.css
hansongu123: Yes, I tried to keep as little CSS as possible and also replicate its original look. This does make the text cell look nice with alignment, spacing, and stuff.

site/cypress/e2e/polls.spec.js
hansongu123: Sure. I think I've already set a class for the table, I think I can swap it on really quick.

site/public/css/polls.css
williamjallen: Yes, the checkboxes are larger than most radio buttons, but I don't see that as a problem.  Different systems are going to show them differently, and rescaling things here just makes this page different from the rest of Submitty.  I'm personally of the opinion that the checkbox size should just be left alone, but it's a rather subjective argument.

site/public/css/polls.css
williamjallen: That seems reasonable.  Hundredths of `em`s are very small values, so you might just stick with `0.5em` for the top and bottom, but that's just my personal opinion.  I'll approve it either way.

hansongu123: Makes sense.

site/public/css/polls.css
hansongu123: In the functionality test, the other guys also tested the scaled down version of the checkbox in this page. It should be good to apply here - because they're really strangely big. In main it is also the source of breaking lines of text of an answer field because its size is adding extra spacing between the first row and second row of text. 

But that problem is solved by putting it in a table cell and centering it vertically. I think it is good either way too. I'm removing the change.

site/cypress/e2e/polls.spec.js
williamjallen: Was this test missed while making other changes?  The other tests target by `.poll-content`.

site/public/css/polls.css
```diff
@@ -39,10 +39,24 @@
...
-.response-btn {
+.option-btn {
+    vertical-align: middle;
+}
+
+.option-btn input[type="radio"]{
     vertical-align: top;
 }
 
+.student-submit[type="submit"] {
```
williamjallen, line 50: If this button is unique, there's no reason why we need to specify that it is of type submit.
```suggestion
.student-submit {
```

site/cypress/e2e/polls.spec.js
hansongu123: They were missed. Oops.

williamjallen (approved):

hansongu123:
I made small edits to the Cypress file to clean up the code after merging "Tomorrow's Polls" PR, whch also made changes to the same file.

PR #9078, opened 2023-03-21
[Feature:Autograding] Added auto pulling for docker image

### LINKED ISSUE #8171: Better error message for autograding issues (e.g., missing docker image)
Currently when autograding requires a docker container that the worker does not have, autograding fails and the error message shows up in the global Submitty autograding log (not the most obvious place to check).

This could also apply to for other problems where autograding fails. Better error messages would help with debugging.

Autograding should give a more obvious error message to make debugging easier.

The message is available in `/var/local/submitty/logs/autograding`, but the immediate information (on the autograding results page) is a generic error message that indicates nothing about how to fix the problem.

**Additional context**

Suryansh5545:

Fixes #8171 **Partially** When a image is not present in the worker, a error is logged but not shown on the submission page. But displaying this error will require editing grading code and have it pull the autograding stack and replace the message with it. another approach mentioned in issue is to pull the container

### What is the new behavior?
Container is pulled when not available on the worker image

### Other information?
I am currently working on implementing the error log message feature on the submission page as requested in the issue. However, I would appreciate some help in understanding the code and figuring out the best way to implement it. It is important to have a more informative error message in case the image name is completely wrong(In that case we will need a better message than the current generic one), and I believe that implementing this feature will greatly benefit our users.


prestoncarman:
Can you create a test for this area of the code?

saumyaborwankar (approved):
Looks good to me and seems to work. But we still need to make a new PR for better debugging of error message for docker related issues.

bmcutler:
Thanks for the contribution, and sorry for the much delayed review.

Unfortunately I'm not going to merge this PR because I'm worried about using unreviewed docker images.
(There is a potential for a malicious docker container.)

Currently only users with 'faculty' designation can add docker images to the worker machines through the webUI.
This PR will allow any user who uploaded & built an autograding configuration (instructor access in a single course) can specify a docker image.

I will update the Issue.

PR #9079, opened 2023-03-21
[UI/UX:System] localization/language translation

### LINKED ISSUE #9052: Lang support
There is currently no support for languages other than English on the site.

Support for other languages should be added and presented as an option for users on the site.

skara9:

Resolves #9052
Lack of support for other languages on the site.

### What is the new behavior?
Adds support for other languages through configuration files.

---

Site admin can choose the default language in initial setup, and all users of the site will see text in said default language, falling back to `en_US` for all gaps.

In the Twig template files, plain text will need to be updated from:
```twig
<p>My Text</p>
```
to
```twig
<p>{{ localize('phrases.myText', "My Text") }}</p>
```

where the language files will each have:
```json
{
  "phrases": {
    "myText": "[translation of 'My Text']",
    ...
  },
  ...
}
```

This new method for displaying website text serves to make Submitty more friendly and accessible for academic institutions outside of the US where the primary language is not English, allowing the site to benefit as many Computer Science students as possible in their intellectual pursuits.

---

#### For Testing:

* Check out the PR locally and run `submitty_install`
* Edit one of the Twig files in the way shown above *(the "phrases.myText" and "My Text" can be replaced with anything)*
* Enter the VM and navigate to `/usr/local/submitty/GIT_CHECKOUT/Localization/lang`
* Choose a locale name such as `en_US` and create a JSON file with the name (ending in `.json`)
* Add the key you wrote in the Twig to the JSON object as shown in the example above
* Open `/usr/local/submitty/config/submitty.json`, and add the key `"default_locale": "en_US"`, where `en_US` is your chosen locale name
* Run `submitty_install_site` and see the changes on the website. *(Note: this will need to be run every time the language json is changed, but not necessary for changing the default locale config key)*


cjreed121:
I have some performance comments.

1. Core gets injected into Twig which already causes some slowdowns (this should ideally be fixed but we use it around a lot so it's not easy to remove) so I don't think we should rely on it more. Instead maybe some sort of custom twig function can be made which can handle this. This could also make the syntax a little shorter and nicer. Regardless some profiling should be done to see how this affects performance. Make sure to test this with a decently sized JSON file. It's also possible some of this got better with Twig 3 since I think I last looked at this with Twig 2.
2. While JSONs are decent options to store all this data, a PHP file might be better. This file is going to have to be parsed on every request. The advantage of PHP files is that OPCache will kick in and preload the file so that no parsing needs to be done.

Youssef-Chip (approved):
Code looks clean and the functionality works as described.

.setup/distro_setup/ubuntu/20.04/setup_distro.sh
```diff
@@ -49,7 +49,7 @@ apt-get install -qqy libpam-passwdqc
 apt-get install -qqy ssh sshpass unzip
 apt-get install -qqy postgresql-12
 apt-get install -qqy apache2 apache2-suexec-custom libapache2-mod-authnz-external libapache2-mod-authz-unixgroup libapache2-mod-wsgi-py3
-apt-get install -qqy php php-cli php-fpm php-curl php-pgsql php-zip php-mbstring php-xml php-ds php-imagick
+apt-get install -qqy php php-cli php-fpm php-curl php-pgsql php-zip php-mbstring php-xml php-ds php-imagick php-intl
```
bmcutler, line 52: This edit should be in the 22.04 file instead/also

bmcutler (changes requested):

PR #9081, opened 2023-03-22
[Bugfix:InstructorUI] Cookies store sorting information.

### LINKED ISSUE #9080: Sorting broken on gradeable grading page
**Describe the bug**
Sorting is not working on the gradeable grading page for all except user id.

**Expected behavior**
When the column is clicked it should sort by that column. 

**To Reproduce**
1. As a TA go to a gradable grading page 
2. Try to sort by first name

**Configuration**
 - OS: Windows and Mac
 - Browser: Chrome and Firefox

**Additional context**
Sorting is working on other grading pages such as the lab page.

SaadAhmedGit:

Fixes #9080
Previously, sorting by `first` or `last` name would still sort by `id`.

### What is the new behavior?
Rows are sorted correctly according to the intended attribute.
The information about how to sort columns is now stored in the browser cookies instead of being sent as `GET` request.
### Other information?
I tested by successfully attempting to sort according to `first` and `last` names which were previously not working.


cjreed121 (changes requested):
Thanks for the contribution. We probably want to keep it using $_COOKIE rather than $_GET. If we look at the git blame on this file we can trace it back to https://github.com/Submitty/Submitty/pull/7799. By setting this back to $_GET we'd cause that other issue again. Can you change it so the sort buttons edit the cookie value rather than set a GET parameter.

SaadAhmedGit:
> Thanks for the contribution. We probably want to keep it using $_COOKIE rather than $_GET. If we look at the git blame on this file we can trace it back to https://github.com/Submitty/Submitty/pull/7799. By setting this back to $_GET we'd cause that other issue again. Can you change it so the sort buttons edit the cookie value rather than set a GET parameter.

Ah, I see why `$_COOKIE` was being checked for the `sort` key in the first place. Let me make the requested changes.

SaadAhmedGit:
> If we use cookies, php is not reading sort params. any suggestions, how to go around this? 
> to reproduce, change sort value in URL and update the page
> image
> 

Wasn't the point of using cookies that the sort settings are saved in local storage and not sent with the get request in order to maintain previous state? The changes I made no longer require changes in the URL and these sort params are now stored in the cookies.

SaadAhmedGit:
Hi @cjreed121 , I hope you're doing well. I've made the changes you suggested, and I was wondering if you had a chance to review them yet? Thank you for your help!

cjreed121 (commented):
Thanks for making the changes. Seems to work well. Can you improve your PR description a bit now that this has some changes to cookies? After that and assuming the tests pass, I'll approve this.

SaadAhmedGit:
@cjreed121 I have updated the PR description as suggested.

Issue #4508, opened 2019-09-18
Add ability to order threads in different ways


BenKelly99:
**Describe the bug**
In the forum, the "reverse-time" search method (Chronological \\/) does not change the order in which the threads are displayed

**Expected behavior**
It should sort the posts from oldest to newest, perhaps with recent announcements at the top.

**To Reproduce**
1. Go a forum with some posts
2. Rank them by "reverse-time"/Chronological \\/
3. See error

**Configuration**
 - OS: Windows
 - Browser: Chrome

elihschiff:
It looks like the different ways of sorting only apply to comments within a thread not the list of threads on the left. I tested and the modes seem to be working fine for the comments. 

However I do think it would be helpful to add a second sort button that is for the list of threads on the left as they are always sorted by post date as far as I can tell.

AngadSethi:
Hello!

I am working on this issue. I'm adding options to sort the threads by _alphabetical order of thread titles_, _chronological ascending_, _chronological descending_ and by _thread IDs_. Is there any other option that I should incorporate?

Also, is it fine if I modify the function `getSortedThreads()` at `site/app/controllers/forum/ForumController2.php ` to get the job done?

A side note : I think the bug label should be removed since it really isn't a bug. The sorting function works perfectly in the posts section

Regards
Angad

AngadSethi:
Hello!

Linked a pull request. Please review.

Regards
Angad Sethi

bmcutler:
Closing this issue request because it is low priority and unlikely to be highly used.  

The UI already contains other ways to sort that are rarely used (but used by power users w/ teaching staff access for specific grading needs).

The issue stems from a confusion about what the current chrono sort feature does (it only sorts posts within a thread, never the threads).  We will be tweaking the visual/text to make the existing functionality more clear.

See also #9375

PR #9082, opened 2023-03-22
seems to work

niyozbek:
-sort by row_number is overridden by id sorting
-cookie seems to be useless to removed while passing option from javascript -sorting button appears inside "more" dropdown, related button hover

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -563,7 +563,7 @@ SQL;
      * @param  int      $thread_id          If blockNumber is not known, find it using thread_id
      * @return array    Ordered filtered threads - array('block_number' => int, 'threads' => array(threads))
      */
-    public function loadThreadBlock($categories_ids, $thread_status, $unread_threads, $show_deleted, $show_merged_thread, $current_user, $blockNumber, $thread_id) {
+    public function loadThreadBlock($categories_ids, $thread_status, $unread_threads, $show_deleted, $show_merged_thread, $current_user, $blockNumber, $thread_id, $sortOption = null) {
         $blockSize = 30;
         $loadLastPage = false;
 
@@ -600,6 +600,11 @@ SQL;
         }
         $query_offset = ($blockNumber - 1) * $blockSize;
         $this->buildLoadThreadQuery($categories_ids, $thread_status, $unread_threads, $show_deleted, $show_merged_thread, $current_user, $query_select, $query_join, $query_where, $query_order, $query_parameters, true, true);
+        if ($sortOption == "time") {
+            $query_order = "t.id ASC";
+        } else if ($sortOption == "reverse-time") {
+            $query_order = "t.id DESC";
+        }
         $query = "SELECT {$query_select} FROM threads t {$query_join} WHERE {$query_where} ORDER BY {$query_order} LIMIT ? OFFSET ?";
         $query_parameters[] = $blockSize;
         $query_parameters[] = $query_offset;
--- a/site/app/controllers/forum/ForumController.php
+++ b/site/app/controllers/forum/ForumController.php
@@ -865,14 +865,14 @@ class ForumController extends AbstractController {
         return null;
     }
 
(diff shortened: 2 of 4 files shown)


site/app/libraries/database/DatabaseQueries.php
```diff
@@ -600,7 +600,14 @@ public function loadThreadBlock($categories_ids, $thread_status, $unread_threads
...
         $this->buildLoadThreadQuery($categories_ids, $thread_status, $unread_threads, $show_deleted, $show_merged_thread, $current_user, $query_select, $query_join, $query_where, $query_order, $query_parameters, true, true);
-        $query = "SELECT {$query_select} FROM threads t {$query_join} WHERE {$query_where} ORDER BY {$query_order} LIMIT ? OFFSET ?";
+        if ($sortOption == "time") {
+            $query_order = "t.id ASC";
+        }
+        elseif ($sortOption == "reverse-time") {
+            $query_order = "t.id DESC";
+        }
+        $query = "SELECT {$query_s
+        elect} FROM threads t {$query_join} WHERE {$query_where} ORDER BY {$query_order} LIMIT ? OFFSET ?";
```
williamjallen, line 610: Was this change intentional?

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -600,7 +600,14 @@ public function loadThreadBlock($categories_ids, $thread_status, $unread_threads
         }
         $query_offset = ($blockNumber - 1) * $blockSize;
         $this->buildLoadThreadQuery($categories_ids, $thread_status, $unread_threads, $show_deleted, $show_merged_thread, $current_user, $query_select, $query_join, $query_where, $query_order, $query_parameters, true, true);
-        $query = "SELECT {$query_select} FROM threads t {$query_join} WHERE {$query_where} ORDER BY {$query_order} LIMIT ? OFFSET ?";
+        if ($sortOption == "time") {
```
williamjallen, line 603: It's generally better to use strict equality in PHP
```suggestion
        if ($sortOption === "time") {
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -600,7 +600,14 @@ public function loadThreadBlock($categories_ids, $thread_status, $unread_threads
...
+            $query_order = "t.id ASC";
+        }
+        elseif ($sortOption == "reverse-time") {
```
williamjallen, line 606: Same here

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -600,7 +600,14 @@ public function loadThreadBlock($categories_ids, $thread_status, $unread_threads
...
+            $query_order = "t.id DESC";
+        }
```
williamjallen, line 608: Do we need an "else" case here?

williamjallen (changes requested):

site/app/libraries/database/DatabaseQueries.php
niyozbek: So, by default, it was not set here. When I checked value of $query_order was  set by buildLoadThreadQuery as "row_number". So I did not write else condition not to delete the behavior that works without sorting. What would be a better approach?

site/app/libraries/database/DatabaseQueries.php
niyozbek: no, it is not. i fixed it

hansongu123 (changes requested):
Despite the small changes, clicking on the two new options in the forum dropdown doesn't seem to be changing the ordering of the threads or posts.
image
image
Which unfortunately doesn't close #4508

(But the feature seems to be working in the Instructor's UI)

04vmatsibekker (commented):
For me, when testing it out the changes seem to work fine. When on the discussion forum you could change the order of the threads, and once you have entered a thread you could only change the order of the posts. 


I tried the changes as a student and as an instructor, and I saw that both worked

bmcutler:
Thanks for the contribution, but we've decided that the original issue was due to a confusion about the existing functionality rather than a need for new functionality.  And the new functionality will make the interface more cluttered and potentially confusing.  So we will be closing this issue and this PR.

Note:  related new issue #9375

PR #9084, opened 2023-03-23
[Bugfix:System] Turn on HTTP only for cookies

cjreed121:
The HTTP only flag is not set for cookies such as `submitty_session`.

### What is the new behavior?
All cookies that use the `Utils::setCookie` function will now default to being HTTP only. This flag will tell browsers to not allow the cookie to be readable by JavaScript. All other cookies not set by this function should not be effected.

### Other information?
A quick local testing worked fine but I want to see how the tests perform. All current cookies will not be set to HTTP Only automatically, but within a week all `submitty_session` cookies will get set again with this new flag.

### DIFF
--- a/site/app/libraries/Utils.php
+++ b/site/app/libraries/Utils.php
@@ -105,16 +105,17 @@ class Utils {
      * @param string        $name name of the cookie
      * @param string|array  $data data of the cookie, if array, will json_encode it
      * @param int           $expire when should the cookie expire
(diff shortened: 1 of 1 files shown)


williamjallen (approved):
Changes seem reasonable.  I wasn't able to find any instances where this might break existing functionality.

MasterOdin (approved):

PR #9089, opened 2023-03-24
[Bugfix:Forum] Websocket order fix

ANS1514:

Say that two people have the discussion forum open. If student1 were to create a post it would show up at the top of the page, above pinned threads, for student2. Ideally this is not the case and it shows up below the pinned threads.

### What is the new behavior?
The behavior I saw in my tests show that the post now shows up below the pinned threads.

### Other information?

I tested this behavior by making pinned posts and regular posts in the discussion forum on a Submitty site on microsoft edge and I saw how it changed a seperate submitty site I had open on google chrome.


ch3nnyj (approved):
I have reviewed and tested your changes and can assure you it has fixed the issue. Your code looks functional and I don't see any changes that need to be made. You have ensured that the insertion is below every pinned discussion. Nice work!

site/public/js/forum.js
```diff
@@ -377,7 +377,16 @@ function socketNewOrEditThreadHandler(thread_id, edit=false) {
                         const last = last_bookmarked.length === 0 ? last_announcement : last_bookmarked;
 
                         if (last.length === 0) {
-                            $(new_thread).insertBefore($('.thread_box_link').first()).hide().fadeIn('slow');
+                            if ($(new_thread).find('.thread-announcement-expiring').length === 1) {
+                                $(new_thread).insertBefore($('.thread_box_link').first()).hide().fadeIn('slow');
+                            }
+                            else {
+                                let spot_after_announcements = $('.thread_box_link').first();
```
williamjallen, line 384: This variable doesn't appear to be reassigned after the initial declaration, so it should be defined as a constant.

```suggestion
                                const spot_after_announcements = $('.thread_box_link').first();
```

williamjallen (changes requested):
This new code is quite a hack, but not particularly out of character for the discussion forum code unfortunately, so I'll approve it as long as the functionality is correct.

The original issue is fixed, but now you have the opposite problem: both regular posts and new announcements appear under the previous announcements when updating via websockets.  When you reload the page, the new announcements move to the top of the list.  The intent here is for new posts to show up in the same order they are displayed when the page is first loaded.

williamjallen (approved):

PR #9096, opened 2023-03-26
[Bugfix:InstructorUI] Set confetti element display to `none` after finishing.

### LINKED ISSUE #9091: confetti cannot automatically stop
**Describe the bug**
Currently, if a user clicks his full-credit auto-grading block, the `<canvas>` element's display attribute will change from **none** to **block**. During the animation, if the user clicks the block again, it will change back to **none**. 
However, after the animation, the `canvas` element will not automatically change from **block** to **none**. In that case, the user has to click the block twice to intrigue the animation again(first click to change **block** to **none** and second click to change **none** to **block**)

**Expected behavior**
Set a time interval. When the animation is finished, set the `canvas` element from **block** to **none**

**To Reproduce**
1. click full-credit auto-grading block
2. wait for the animation to finish
3. click the block again
4. See error(you will see that there is no response after you clicked it. You need to click again to intrigue the animation again)

**Screenshots**
confetti

SaadAhmedGit:

The `display` property of the Confetti animation remained set to `block` after completion, forcing users to click twice - once to reset it to `none` and again to register the click. This fixes  #9091.

### What is the new behavior?
The `canvas` element which displays the confetti animation is now set to `none` after 9 seconds which is the time it takes for the animation to end.
### Other information?
I tested by opening the browser developer tools and seeing the property change back to `none` as soon as the animation ended and, my first click after animation ended would register correctly as well.

### DIFF
--- a/site/public/js/confetti.js
+++ b/site/public/js/confetti.js
@@ -163,23 +163,39 @@ function addConfetti() {
             return;
         }
 
-        update();
-        ctx.clearRect(0, 0, canvas.width, canvas.height);
+        // Run each step of the animation for 9 seconds then hide the confetti.
+        let start = null;
+        function confettiStepAnimation(timestamp) {
+            if (!start) {
+                start = timestamp;
+            }
+            const progress = timestamp - start;
+            if (progress < 9000) {
+                update();
+                ctx.clearRect(0, 0, canvas.width, canvas.height);
 
-        pieces.forEach((p) => {
-            ctx.save();
+                pieces.forEach((p) => {
+                    ctx.save();
 
-            ctx.fillStyle = p.color;
+                    ctx.fillStyle = p.color;
 
-            ctx.translate(p.x + p.size / 25, p.y + p.size / 2);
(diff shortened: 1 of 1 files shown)


shailpatels (changes requested):
Hi @SaadAhmedGit, thanks for the PR.

There is already a section of code that has the logic for when the confetti should end here, but it looks like its not doing anything with the div,
https://github.com/Submitty/Submitty/blob/main/site/public/js/confetti.js#L141

If you can update that section instead, the change will be simpler then adding a secondary timer

SaadAhmedGit:
Hi @shailpatels , could you review this again? I've made the requested changes. Sorry for the weird revert commit; I should have done that before making these changes. I'm still new to open source, so pardon me. Thank you!

Yujjio:
Hello @SaadAhmedGit, thanks for the PR. 

The bug cannot be fixed by changing the time since it varies with the monitor's resolution(a strange feature in JS). Instead, we need to check the last segment's height; hide the animation If it is out of the screen. Currently, I have modified that part.

PR #9105, opened 2023-03-27
[Testing:Developer] Split E2E Tests using Composite Action

### LINKED ISSUE #8611: Split up the e2e testing GitHub action to speed up receiving test results
The Submitty tests running on GitHub take over an hour to run. Primarily, due to one test that incorporates many different types of tests into a single threaded process.

Instead of using one action on distinct tests, the action could be split up to run in parallel and improve overall test suite running time. The total test time will not change, but results will be produced quicker. The e2e test could be split into these testing areas:

* Selenium tests
* Cypress test part 1
* Cypress test part 2
* Integration tests

**Additional context**
Here is a slack comment from @williamjallen:

> For the ~50-minute run I linked above, ~7 minutes were spent running Selenium tests, ~15 minutes were spent running Cypress tests, and ~10 minutes were spent running integration tests.  Those three categories combined to make up about 70% of the total running time.

And a follow up from @shailpatels:

> ... I can definitely see making an e2e job for selenium and just for cypress to get them in parallel. We could also split cypress into a subset of tests between 2 jobs as well

You can find the full thread on slack <https://submitty.slack.com/archives/C9A0XMXEX/p1666843464571729>. These comments come towards the end of the thread.

04vmatsibekker:

Fixes #8611. 

### What is the new behavior?
Now the e2e tests are split into three: selenium, cypress, and integrations tests to speed up the process


.github/actions/set-up-composite/action.yml
```diff
@@ -0,0 +1,224 @@
+name: 'copyFiles'
+descriptions: 'Uses the composite action to copy and set up files'
+
+# inputs:
+#   name:
+#     description: 'A name to include in the greeting'
```
williamjallen, line 6: This should probably be removed

.github/actions/set-up-composite/action.yml
```diff
@@ -0,0 +1,224 @@
...
+  PHP_VER: 7.4
+  NODE_VERSION: 16
+  PYTHON_VERSION: 3.8
+
+# runs:
+#   using: "composite"
+#   steps:
+#     - name: Greet someone
+#       run: echo "Hello"
+#       shell: bash
```
williamjallen, line 26: Same here

.github/actions/set-up-composite/action.yml
```diff
@@ -0,0 +1,224 @@
+name: 'copyFiles'
+descriptions: 'Uses the composite action to copy and set up files'
```
williamjallen, line 2: Should these be a bit more descriptive?

.github/actions/set-up-composite/action.yml
williamjallen, line 1: Should the filename be a bit more descriptive?

.github/actions/set-up-composite/action.yml
```diff
@@ -0,0 +1,224 @@
...
+        sudo -H pip3 install .
+        sudo pip3 show submitty_utils
+      shell: bash
+
+    - name: Install accessibility checker
+      run: |
+        wget https://github.com/validator/validator/releases/download/20.3.16/vnu.jar_20.3.16.zip
+        unzip vnu.jar_20.3.16.zip
+        sudo mv dist/vnu.jar /usr/bin/
+      shell: bash
```
williamjallen, line 115: I believe this only needs to be installed for the selenium tests (for the tests in `tests/e2e/test_accessibility.py` specifically)

.github/actions/set-up-composite/action.yml
```diff
@@ -0,0 +1,224 @@
...
+
+    - uses: nanasess/setup-chromedriver@master
+    - name: Setup chromedriver
+      run: |
+        sudo systemctl restart submitty_autograding_worker
+        sudo systemctl restart submitty_autograding_shipper
+        export DISPLAY=:99
+        chromedriver --url-base=/wd/hub &
+      shell: bash
```
williamjallen, line 224: This is only necessary for the selenium tests

.github/workflows/submitty_ci.yml
```diff
@@ -273,8 +274,6 @@ jobs:
           run: sudo apt-get install -y shellcheck
         - name: Run ShellCheck
           run: python3 run_shellcheck.py  # Uses the default Python installed with Ubuntu
-
-
```
williamjallen, line 277: It might be good to restore the space between jobs to aid readability

.github/workflows/submitty_ci.yml
```diff
@@ -300,195 +299,48 @@ jobs:
...
+    e2eCypressOne:
+        needs:
+          - js-lint
+          - js-unit
+          - php-lint
+          - php-unit
+          - python-lint
+          - python-unit
+          - css-lint
+          - shellcheck
```
williamjallen, line 322: We can shave off another 3-4 minutes from the overall running time by not adding these as dependencies.

.github/workflows/submitty_ci.yml
```diff
@@ -600,6 +520,76 @@ jobs:
           - name: Check for site errors
             working-directory: ${{ env.SUBMITTY_REPOSITORY }}
             run: sudo bash tests/test_site_error_log.sh
```
williamjallen, line 522: This needs to be run after every job.  Perhaps there is a way to do this automatically by adding something like [this](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runspost) to the setup action you created.

.github/workflows/submitty_ci.yml
```diff
@@ -547,7 +399,76 @@ jobs:
           - name: Restart autograding worker
             run: |
               sudo systemctl restart submitty_autograding_worker
+    e2eCypressTwo:
```
williamjallen, line 402: I don't think there's a need to have two separate Cypress jobs here.  All of the stuff in this job takes less than 5 minutes to complete, based on the most recent CI runs.  The setup will likely take longer than the actual testing here.

.github/workflows/submitty_ci.yml
```diff
@@ -600,6 +520,76 @@ jobs:
...
+              POSTGRES_PASSWORD: submitty_dbuser
+              POSTGRES_USER: postgres
+              TZ: America/New_York
+            options: >-
+              --health-cmd pg_isready
+              --health-interval 10s
+              --health-timeout 5s
+              --health-retries 5
+            ports:
+              - 5432:5432
```
williamjallen, line 547: I don't think we need Postgres for the integration tests

.github/workflows/submitty_ci.yml
```diff
@@ -600,6 +520,76 @@ jobs:
...
+
+        steps:
```
williamjallen, line 549: I think most of the stuff in this section can be eliminated.  The integration tests are mostly unit tests, and don't need the sample courses or website to be working as far as I am aware.

williamjallen (changes requested):

williamjallen:
Given the number of commits here, you might also consider squashing your current changes into a single commit.

.github/actions/set-up-composite/action.yml
04vmatsibekker: I am pretty sure the filename has to be called action.yml for it to work. That is at least what I have seen when looking it up.

williamjallen: Sure, but can `set-up-composite` be changed?

04vmatsibekker: yup and I will change it

.github/workflows/submitty_ci.yml
williamjallen: Something like [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) might work.  If I'm understanding the docs correctly, it should kill all the other running jobs whenever a job with the fail fast option fails.

.github/workflows/submitty_ci.yml
```diff
@@ -105,7 +106,7 @@ jobs:
                 working-directory: site
         steps:
             - uses: actions/checkout@v3
-            - uses: shivammathur/setup-php@2.25.2
+            - uses: shivammathur/setup-php@2.25.1
```
williamjallen, line 109: It looks like this dependency has been updated since you created this PR.  Please bump the version to 2.25.2 like we have on main.

.github/workflows/submitty_ci.yml
```diff
@@ -140,7 +141,7 @@ jobs:
               working-directory: site
       steps:
           - uses: actions/checkout@v3
-          - uses: shivammathur/setup-php@2.25.2
+          - uses: shivammathur/setup-php@2.25.1
```
williamjallen, line 144: Same here

.github/workflows/submitty_ci.yml
```diff
@@ -273,333 +274,197 @@ jobs:
           run: sudo apt-get install -y shellcheck
         - name: Run ShellCheck
           run: python3 run_shellcheck.py  # Uses the default Python installed with Ubuntu
+    e2eSelenium:
```
williamjallen, line 277: It might be good to add some whitespace here to make it clear that this is a new job

.setup/pip/system_requirements.txt
```diff
@@ -40,7 +40,7 @@ clang==14.0
 # Python libraries for QR bulk upload
 pyzbar==0.1.9
 pdf2image==1.16.3
-numpy==1.24.3
+numpy==1.24.2
```
williamjallen, line 43: This change should be reverted.

.github/workflows/submitty_ci.yml
```diff
@@ -273,333 +274,197 @@ jobs:
...
-              path: ${{env.SUBMITTY_REPOSITORY}}/site/cypress/videos
 
-          - name: Check for site errors
-            working-directory: ${{ env.SUBMITTY_REPOSITORY }}
-            run: sudo bash tests/test_site_error_log.sh
+    e2eIntegrationTests:
+      runs-on: ubuntu-20.04
+      services:
+        postgres:
+          image: postgres
```
williamjallen, line 452: It would be good to add a comment here stating why postgres is listed here, as well a TODO to remove it eventually.

williamjallen (changes requested):
I still need to do a more thorough review, I'll leave these comments here in the meantime:

williamjallen (changes requested):
In general, I think things are looking pretty good here and we should be able to get this merged pretty soon.  We can probably make some more improvements eventually, but this should serve as a decent starting place at least.

Something to note: It looks like the regular e2e job is marked as "waiting".  I don't see an e2e job listed in the yaml file, so I assume that the waiting job is related to the `workflow_dispatch` trigger and will go away once this is merged.

Unrelated to the PR itself, but general development advice:  Good commit messages help you stay organized, and help reviewers understand why each commit was made.  Currently I see this:
image
These commit messages tell me *what* was changed, but not *why*.  I can see what was changed by looking at the diff, so the commit message is rather useless.  A better commit message would be something like "Add comment about x, y, and z" or "Re-add database service to integration tests", for example.

williamjallen (approved):
Changes seem reasonable to me.  Any further changes can be made as separate PRs.

PR #8985, opened 2023-02-28
[Feature:InstructorUI] Teaching Staff Locked Gradeables Access

ch3nnyj:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?

Currently locked gradeables were previously inaccessible for teaching staff and instructors. 
This would prevent students from getting help with certain gradeables if teaching staff hasn't completed the prerequisite.
When teaching staff attempts to access locked graded the below alert pops up.
Screenshot 2023-02-28 at 2 25 36 PM
Closes #8355 
### What is the new behavior?

Now locked gradeables can be bypassed with a confirmation screen allowing teaching staff to access the submission page.
After clicking ok teaching staff will be redirected to the submission page.
Screenshot 2023-02-28 at 3 23 19 PM

### Other information?
To Test:
1.) Create a gradeable that requires another gradable as a prerequisite.
2.) Attempt to access the as a student and verify that there is no access to that gradeable.
3.) Verify all teaching staff (Mentors/TAs/Instructors) can access the gradeable after click ok on the conformation prompt.

PR #9115, opened 2023-03-31
[Bugfix:InstructorUI] locked gradeable override message

bmcutler:
fixing message from PR #8985


ch3nnyj (approved):
Works for me Barb, finally we have a prompt!!!

PR #9116, opened 2023-03-31
[Bugfix:Autograding] stop confetti automatically

Yujjio:

Confetti animation will stop based on cumulative animation time, which, however, is cumulated based on the screen's resolution. In that case, checking the time to determine whether to stop is inaccurate. 

More than that, the canvas's height currently will be initialized as a whole web page's size instead of the visible part's size. 

### What is the new behavior?

Change the canvas's height to the visible area's height.

Stop animation based on the piece list. When a piece is out of the visible area, it will be removed from the list. In that case, when every piece is out of the screen and no more visible piece, stop the animation. 

Modified the number of pieces based on the screen size so that it will not be too crowded

### DIFF
--- a/site/public/js/confetti.js
+++ b/site/public/js/confetti.js
@@ -33,14 +33,13 @@ function addConfetti() {
     canvas.width  = window.innerWidth;
     const body = document.body;
     const html = document.documentElement;
-    canvas.height = Math.max( body.scrollHeight, body.offsetHeight,
-        html.clientHeight, html.scrollHeight, html.offsetHeight );
+    canvas.height = document.body.clientHeight;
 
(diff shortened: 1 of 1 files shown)


ch3nnyj (changes requested):
When testing confetti, I noticed when clicking on a perfect score to activate it then switching to a different tab, then back to current submission tab. The confetti instead of continuing like on main it bundles up on top of the page. 

Update - Issue only exist on Safari
Issue is the same as main

ch3nnyj (approved):
All set and looks, tested it multiple times. Code wise I think its all good.

ANS1514 (approved):
Looks good

Issue #9114, opened 2023-03-31
Add Solid Background to the Checkmark Icon for Polls in Dark mode


hansongu123:
**What problem are you trying to solve with Submitty + How to solve it**

When an option in a poll is considered correct, and the instructor allowed answer releasing, a checkmark would appear next to the correct option's text in student's poll page like this:
![image](https://user-images.githubusercontent.com/123511202/229147727-fda9d625-0ace-4667-b757-ab95ecae7ea3.png)
However, this solid **FA icon** uses whatever the background color it is in to color the inside (while the outside is colored by a defined color in Submitty)

Change the inside color of the checkmark to a defined white color in Submitty, regardless of what mode it is in (Light, Dark, Black), so it is less uglier when in a darker background.

**Additional context**
_Please note that the corners of the square check are round, it would take a little bit of extra engineering._

StephCurry07:
Did you try overwriting it with CSS?
Something like:
`.fa-check:before {
  font-weight: bold;
  content: "\f00c";
  color: white !important; /* Set the color to white */
  -webkit-text-fill-color: white !important; /* For Safari */
}
`

PR #9118, opened 2023-03-31
[UI/UX:SubminiPolls] Fix Checkmark In Dark Mode

BairongZhang1:

Fix issue #9114

### What is the new behavior?
image
### Other information?
None

### DIFF
--- a/site/public/css/polls.css
+++ b/site/public/css/polls.css
@@ -103,6 +103,11 @@ form {
 }
 
 .correct-tag {
+    background-image: linear-gradient(to bottom, 
+    transparent 0%, transparent 25%,
+    white 25%, white 75%,
+    transparent 75%, transparent 100%
+    );
     color: var(--standard-dark-green);
     margin-left: 0.5em;
 }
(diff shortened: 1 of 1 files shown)


williamjallen:
Attempting to add borders/backgrounds to FA icons generally doesn't result in a clean-looking UI, since you're applying CSS styling to a SVG image.  Unless FA releases bordered icons, I think any attempt to add an outline around the checkmark will look worse than the current state.  Just my thoughts--I'm open to other opinions here.

hansongu123:
What would happen if I want to change into a different shape like a circle check or just simply adjusting the size? I doubt these changes in CSS would still work.

hansongu123:
FA icons do support features like changing background colors or background shapes. There should be enough templates on the internet just by googling it. What would others do if they want to adjust a FA icons background color in the dark mode? 

@williamjallen I agree it is generally better not to add custom styling for a preset image especially if it's imported like this, but maybe some minimal shape and depth manipulation is good. It would be really cool if we can design our own check mark icon though, I can see it to be useful in many interactive interfaces within Submitty.

hansongu123:
You can try to change it size and shape, and add paddings to see whether or not the changes would still work. Also, different browsers and different OS sometimes have different rendering for the icons.

hansongu123:
@BairongZhang1 I'm going to close this PR now. Lets move to the other PR on the same issue, I've requested a review from you and please feel free to commit.

PR #9132, opened 2023-04-03
[Testing:Submission] Adds cypress tests for gradeables

jason490:

There are no cypress tests for gradeables, which can cause small errors to be overlooked

### What is the new behavior?
Adds a general test to regular gradeables that makes sure the submission has no errors and makes sure all of the files that are submitted, can be downloaded and verifies the content of the file.

### Other information?
The downloaded file gets cleaned up after every test in a headless run, but it doesn't get cleaned up in a cypress open


site/cypress/e2e/gradeables/gradeable.spec.js
```diff
@@ -0,0 +1,41 @@
+describe('Tests cases revolving around gradeable access and submition', () => {
+    ['student','instructor'].forEach((user) => {
```
williamjallen, line 2: Why not also check for other access levels as well?

williamjallen (commented):
As per your comment in the PR description, the downloaded file doesn't get cleaned up when running tests locally.  Is there any particular reason why you can't clean that up after each test?

jason490:
> As per your comment in the PR description, the downloaded file doesn't get cleaned up when running tests locally. Is there any particular reason why you can't clean that up after each test?

I was searching around for a way to clean up the downloaded files when running it locally, but there isn't a official way to delete all the downloaded files. And the easy way to do it is to install [this](https://www.npmjs.com/package/cypress-delete-downloads-folder) plugin.

site/cypress/e2e/gradeables/gradeable.spec.js
jason490: I was thinking that student and instructor is good enough regular tests because the ta access level has almost the same access level as a instructor and if the instructor doesn't give access to the ta, then the gradeable wouldn't even show, but I realized that we might add more restrictions in the future, so I will add a ta too.

williamjallen: if you're going to add a full access grader, might as well add a limited access grader (users `ta` and `grader` for example)

williamjallen (commented):
Another thing you could possibly test is whether the submit button is disabled or not.

williamjallen:
It also looks like you're failing the performance warning test.  To resolve that, you should just ignore the route by adding `/courses/<semester>/<course>/download` to `.performance_warning_ignore.json`.

hansongu123:
More tests are always good, quote William. Don't worry about testing the components that aren't that easy to test. We would usually go around them if they require too much extra steps or could disturb the pattern of the test cases in the same file. There's a reason why this hasn't been done before for the gradeables...
 
I think for now it would be fine just testing the components in the main gradeables page, the visibility and the functionalities of the items that would be there for the default setting, then a test case for opening a most general gradeables item (Open Homework in this case) and the components in that page. For the main content (inside of a gradeable), you can just test the file names, icon properties, version, and autograding information. 

Writing the Cypress tests for gradeables would be a very big task, so I would suggest breaking it up into smaller PR's, so we can have more thorough tests across the three UI's since the three of them would have different looking pages, and interactions with each other.

williamjallen (approved):
Changes look good.  This seems like a decent test of the basic functionality of the page.  As Hanson commented above, it's always good to test more things, but we can get to that later as needed.

PR #9133, opened 2023-04-03
[Testing:CourseMaterials] Changed attachFile to selectFile

jason490:

The previous cypress test used functions that are no longer supported in cypress

### What is the new behavior?
Updated the cypress test to use functions that are supported

### Other information?
Changed attachFile to selecFile and added the path to the file.
Link to where `attachFile` was deprecated:
https://docs.cypress.io/guides/references/migration-guide#Migrating-from-cypress-file-upload-to-selectFile


williamjallen (commented):
Do we still need the `cypress-file-upload` plugin?  If not, it should be removed from our `package.json`.

jason490:
> Do we still need the `cypress-file-upload` plugin? If not, it should be removed from our `package.json`.

It seems like it is still required

hansongu123:
The newest Cypress dependabot PR from two days ago seems hasn't been failing tests.

williamjallen:
> The newest Cypress dependabot PR from two days ago seems hasn't been failing tests.

`attachFile` (and the `cypress-file-upload` plugin as a whole) has been deprecated, but has not been removed.  That means that using `attachFile` is no longer recommended, but existing usages won't break anything (yet).

williamjallen (approved):

PR #9134, opened 2023-04-03
[UI/UX:SubminiPolls] Rework Correct Icon for Polls

hansongu123:
Rework the FA icon that'd appear on the "View Polls" page, now the checkmark in the center should have its own controllable solid color instead of being transparent - by using a stacked icon.

Also made change to the background shape of the checkmark so it doesn't replicate an html checkbox.

image
image

For issue #9114


hrose6967 (approved):
Tested and new icon changes work.

hansongu123:
@BairongZhang1 Can you review the functionality and see how it is different from the PR you made on  the same issue? You are free to add more commits using this as template to make it look nicer and I'll let Hannah review it again if you do make changes.

BairongZhang1 (approved):
Useful change on the button.

PR #9146, opened 2023-04-04
[Feature:System] Vagrant Port Overrides

skara9:
Vagrant port forwards are hardcoded into the configuration, which poses an issue in cases where a developer has another application running on their machine that occupies one or more of the ports.

### What is the new behavior?
Host ports can be overriden through environment variables.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -118,11 +118,11 @@ Vagrant.configure(2) do |config|
 
   # Our primary development target, RPI uses it as of Fall 2021
   config.vm.define 'ubuntu-20.04', primary: true do |ubuntu|
-    ubuntu.vm.network 'forwarded_port', guest: 1511, host: 1511   # site
-    ubuntu.vm.network 'forwarded_port', guest: 8443, host: 8443   # Websockets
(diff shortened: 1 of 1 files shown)


shailpatels (approved):
Looks good, thanks for this I think this will be a convenient update 

Adding some docs to the submitty.org page about the VM installation , for example you could add a blurb saying if you get the following error:

```
Vagrant cannot forward the specified ports on this VM, since they
would collide with some other application that is already listening
on these ports. The forwarded port to #### is already in use
on the host machine.

To fix this, modify your current project's Vagrantfile to use another
port. Example, where '1234' would be replaced by a unique host port:

  config.vm.network :forwarded_port, guest: 1511, host: 1234

Sometimes, Vagrant will attempt to auto-correct this for you. In this
case, Vagrant was unable to. This is usually because the guest machine
is in a state which doesn't allow modifying port forwarding. You could
try 'vagrant reload' (equivalent of running a halt followed by an up)
so vagrant can attempt to auto-correct this upon booting. Be warned
that any unsaved work might be lost.
```

do a 

```
<PORT_ENV_VAR>=#### vagrant up
```

where PORT_ENV_VAR is one of the 5 options you can override that's already in use

Issue #9112, opened 2023-03-30
Export status of open grade inquiry to Rainbow Grades


bmcutler:
**What problem are you trying to solve with Submitty**
If a student has submitted a grade inquiry for a specific assignment, and it has not yet been resolved (it is open), it would be helpful to note that on the Rainbow Grades spreadsheet.  

**Describe the way you'd like to solve this problem**
First, the state of the grade inquiry should be included in the .json file exported for each students in Generate Grade Summaries.  I suggest 3 states.   "None" (never started a grade inquiry), "Open", and "Resolved". 

Ideally we visualize all 3 states in the RainbowGrades output, with "Resolved" being a minor asterisk, and "Open" being a larger more visible tag <-- the instructor should not assign final grades if there is an unresolved grade inquiry!

**Additional context**
related to #5260 

I don't have a final proposal for the visual design -- it should be prototyped and those prototypes presented and discussed by a group. 

Perhaps we outline in red (or some other color) the cells for each student & gradeable that has an active grade inquiry.  The color needs to be noticeable, but should be different than color that is used for other visuals in #5260.

Once the grade inquiry is resolved, perhaps the cell has a thick black border?  Or maybe a symbol (e.g., '!') after the score "83!" to indicate it was resolved.   The symbol needs discussion, we are already using a * for use of a late day and @ for academic integrity.  Those symbols could be redesigned.  

We should have a key at the bottom of the page summarizing the visuals.

ziesski:
![Screen Shot 2023-04-06 at 01 13 17](https://user-images.githubusercontent.com/123261952/230277911-920550b0-895e-4f9e-803b-dff59503c817.png)

I just modified some html from instructor gradebook, and applied possible cases that @bmcutler mentioned. Is this similar to your rough blueprint? 

--- Stuff Barb mentioned in #5260 ---
red outline with a 0 score in the cell = submitted, but too late or not enough late days => automatic zero
yellow outline with a non-zero score = grade override (often used for makeup exams, or other special cases)
empty box, no score = no submission
orange outline with a '?' score in the cell = submitted, but not yet graded
green(?) outline with a non-zero score = grade inquiry in progress
red outline with a @ score in the cell = academic integrity violation on this assignment

ziesski:
[...]

PR #9153, opened 2023-04-07
[Feature:RainbowGrades] Export inquiry status in grade summaries

AdamTuhacek:

Inquiry status is not included in the json files created by generating grade summaries for rainbow grades

### What is the new behavior?

New field for the inquiry status, which is set to "None", "Open", or "Resolved" based on the current status.

This is step 1 of #9112


hansongu123 (commented):
Is the status of an inquiry a new variable? It doesn't seem to be included in the list right above the changes in the same function. 

We should also initialize `inquiry_status` to open and use if else statements to modify its content accordingly.

AdamTuhacek:
> This change should work for the first step in the original issue. I think it would be a good idea to add "regrade" to the variable name so it is more obvious of what it's representing.
> 
> Is there a way we can initialize the status to "none" and then modify it using the if else statement?

Yeah, we can discuss that, I'm not sure it matters much since this will only be parsed by rainbow grades. 

The field could be set to None by default and then updated accordingly, not sure that would make much of a difference either, but I can look into it.

Yujjio:
LGTM

Yujjio (approved):

PR #9155, opened 2023-04-07
[Bugfix:TAGrading] Correct PDF Annotation Rotation

Eceptonsu:
The PDF annotations in TA grading does not rotate with the PDF when the rotation tool is used.

### What is the new behavior?
Correctly fixes the abnormal behavior related to rotation and now the annotations move with the PDF. See the image below.
image


williamjallen (changes requested):
Functionality review comments (discussed at hackathon):

PDF can be rotated right, but not left.  Evan will try to reproduce.

ziesski:
The problem also happens with right rotation if you zoom out to 70-80% and then try rotating it.

ziesski:
After minor edit in rotation function, annotation rotation now functions as expected in both direction.

However, while testing I discovered new error - when the PDF is rotated, the text box does not work, pressing save button won't save the text and dismisses. I will open a new issue.

04vmatsibekker (commented):
Works well, but once rotated to the side, you can not successfully annotate.

PR #9156, opened 2023-04-07
[Bugfix:TAGrading] Annotated PDF Download Resolution

Eceptonsu:
The downloaded annotated PDF in TA grading interface currently has a OK resolution but not the best. In addition, the annotation does not appear in the correct spot in the PDF.

### What is the new behavior?
The downloaded PDF will have correct resolution that match the original file, no matter if the pdf is annotated or not. If the pdf has annotations, the corresponding elements will be adjusted as well.


site/public/js/pdf/PDFAnnotateEmbedded.js
```diff
@@ -115,7 +115,8 @@ function renderPageForDownload(pdf, doc, num, targetNum, file_name) {
             doc.addPage();
         }
         pdf.getPage(num).then((page) => {
-            const viewport = page.getViewport({scale:1.5});
+            const scale = 3; // Increase the scale factor for higher resolution
+            const viewport = page.getViewport({scale: scale});
```
ch3nnyj, line 119: If all you did was change 1.5 to 3 here we don't need to introduce a new variable. You can do it on one line.

ch3nnyj (changes requested):
Everything else looks good to me, if you change the variable thing then everything else is good!

williamjallen (approved):
Seems to work as expected.  It's worth noting that the downloaded file size for a PDF in the sample course is significantly larger after these changes (17.5MB vs 2.5MB for an average PDF in the sample course).

It's also worth noting that the lines on the annotations are not smooth at the corners.
image

PR #9159, opened 2023-04-07
[Bugfix:System] Fixed Minor Issues with Pronouns UI

### LINKED ISSUE #9147: Minor issues with My Profile pronouns UI
**Describe the bug**

1. On Mac Chrome the width of the selector is too narrow for the default text:
(I actually cannot scroll to see what the rest of the text says)
Screen Shot 2023-04-05 at 12 38 08 AM

2.  Before I've made a selection, there are multiple options in the dropdown:
  Screen Shot 2023-04-05 at 12 38 46 AM
  But once I've made a selection, there is only that one option:
  (It would be nice to still see all of the options.  Yes, if I clear out my selection, then I can see all the options again, but I don't think that should be necessary.)
  Screen Shot 2023-04-05 at 12 39 03 AM

3.  If I click on the pencil to consider editing my selection, but don't change anything pressing "Submit" does not close the   window.  I just get an error message and the modal stays open.  I didn't notice the error, so it was annoying, I kept pressing "Submit" and nothing happened.  Go ahead and close it when I hit submit, even if I didn't change my pronouns.
Screen Shot 2023-04-05 at 12 39 19 AM

mahi-pas:

Closes #9147 

- The width of the pronouns selector is too narrow
- Datalist filtering needs to be removed in pronouns selector
- Pressing submit doesn't close the Change Pronouns popup when no text is changed

### What is the new behavior?

- Width of pronouns selector is now 30 characters (max pronouns size is 30 characters)
- Pressing submit closes the Change Pronouns popup when no text is changed
- Trash button that removes typed pronouns in form (compromise for selector filtering- see other information)

### Other information?
I've looked into how to use the data list without filtering and couldn't find anything that works.
The closest I found was this: https://stackoverflow.com/questions/41678677/avoid-filtering-of-datalist-items-in-an-input-element
but it didn't work correctly (for me the dropdown didn't show, and when trying the solution in an online interpreter it only showed the dropdown after typing at least 1 character). Filtering is required by data list, and I would need to implement a custom dropdown to bypass it.


mahi-pas:
Known bug: hitting the trash button and then the close button, and then opening the pronouns form again gives empty pronouns. It should load the set pronouns.

hansongu123:
Despite Barb's scolding, I would still suggest re-implementing the drop down and follow as a template for how the pronouns field is done by many popular sites, _such as on _GitHub's profile page.__ You can then add some JavaScript to allow typing before specifying input type. You don't really need to change other things around since this would still have the same listening method. This would at least resolve bulleted point 2 in the linked issue.

mahi-pas:
this PR is now complete and ready for review

site/public/js/user-profile.js
```diff
@@ -1,10 +1,10 @@
 /* exported updateUserPronouns, showUpdatePrefNameForm, showUpdatePronounsForm,
 showUpdatePasswordForm, showUpdateProfilePhotoForm, showUpdateSecondaryEmailForm,
 updateUserPreferredNames, updateUserProfilePhoto, updateUserSecondaryEmail,
-changeSecondaryEmail
+changeSecondaryEmail, clearPronounsBox
  */
 
-var pronounsLastVal = null;
+let pronounsLastVal = null;
```
williamjallen, line 7: Is there an easy way to avoid having a global variable here?

williamjallen (commented):
Other than the one small nitpick mentioned below, the code for this PR looks decent

site/app/templates/EditPronounsForm.twig
```diff
@@ -10,22 +10,24 @@
...
+                   size="30">
                 <datalist id="default_pronouns_list">
                     <option value="He/Him"></option>
                     <option value="She/Her"></option>
                     <option value="They/Them"></option>
-                    <option value="Custom"></option>
+                    <option value="(Type Custom Here)"></option>
                 </datalist>
             </input>
+            <a href="javascript:clearPronounsBox();"><i class="fas fa-trash" style="padding:0.5rem;"></i></a>
```
shailpatels, line 30: Typically we should not be using `a` tags for non navigation purposes since they can mess with screen readers and act a little weird on certain browsers 

This should be a button wrapping the icon and you can remove the styling for it so only the icon shows up
```html
<button aria-label="Clear pronoun input" type="button" style="background: none;border:none;" onclick="clearPronounsBox();"><i> icon... </i></button
```

The aria-label will prevent the accessibility checker from complaining about a button with no text and the style removes any bordering. The type prevents the button from submitting the form 

You could also optionally add a title field to the icon, this will show some help text when a user hover's over it, something like 

```html
<i title="Clear pronouns field" ></i>
```

Here's some additional context on the subject
https://css-tricks.com/buttons-vs-links/
https://fontawesome.com/docs/web/dig-deeper/accessibility
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#accessibility

shailpatels (changes requested):

shailpatels (commented):
If you want to use the jquery autocomplete, it looks like the dropdown is hidden on the modal because its not being attached to the right element.

You can fix this by adding the class `ui-front` to the file https://github.com/Submitty/Submitty/blob/main/site/app/templates/generic/Popup.twig#L21

so that it looks like
```html
<div class="form-body ui-front">
```

This would be good so that if anyone else wants to use the autocomplete jquery functionality in a popup it'll work for them too

Then you can add the JS logic to setup autocomplete, note you'll need to delete the `datalist` stuff since it'll conflict. This logic should go in its own script tag in a file like the  `site/app/templates/UserProfile.twig` file since you need access to twig variables

```js
var opts = [
            "{{ user_pronouns }}",
            "He/Him",
            "She/Her",
            "They/Them"
 ];

        
 $("#user-pronouns-change").autocomplete({
        source: (request, response) => {
            response(opts); //always return the full list no matter what the user types
        },
        minLength: 0
    }).focus(function () {
        $(this).autocomplete('search', '');
    });
```

The `.focus` function basically shows the dropdown as soon as you click on the input instead of when you start typing, and the override for source will return the entire list of pronouns instead of filtering

site/app/templates/EditPronounsForm.twig
mahi-pas: Done!

mahi-pas:
> If you want to use the jquery autocomplete, it looks like the dropdown is hidden on the modal because its not being attached to the right element.
> 
> You can fix this by adding the class `ui-front` to the file https://github.com/Submitty/Submitty/blob/main/site/app/templates/generic/Popup.twig#L21
> 
> so that it looks like
> 
> ```
> <div class="form-body ui-front">
> ```
> 
> This would be good so that if anyone else wants to use the autocomplete jquery functionality in a popup it'll work for them too
> 
> Then you can add the JS logic to setup autocomplete, note you'll need to delete the `datalist` stuff since it'll conflict. This logic should go in its own script tag in a file like the `site/app/templates/UserProfile.twig` file since you need access to twig variables
> 
> ```js
> window.addEventListener('load', () => {
> 
>     var opts = [
>             "{{ user_pronouns }}",
>             "He/Him",
>             "She/Her",
>             "They/Them"
>     ];
> 
>         
>      $("#user-pronouns-change").autocomplete({
>         source: (request, response) => {
>             response(opts); //always return the full list no matter what the user types
>         },
>         minLength: 0
>      }).focus(function () {
>         $(this).autocomplete('search', '');
>      });
> });
> ```
> 
> The `.focus` function basically shows the dropdown as soon as you click on the input instead of when you start typing, and the override for source will return the entire list of pronouns instead of filtering
> 
> Here are more details on how it works https://api.jqueryui.com/autocomplete/#option-source

I wasn't able to get a version of this that ran nicely. It would never fill in the data, and sometimes it would randomly close the window

site/public/js/user-profile.js
mahi-pas: We discussed

Yujjio (approved):
LGTM

bmcutler (approved):

Issue #9177, opened 2023-04-13
Undo button overlaps when scrolling in Grading lab


ziesski:
**Describe the bug**

Search for student, undo, redo button overlaps with Color info of check points, when scrolling down.

**Expected behavior**

Expect to stay in its original position, rather than following scroll.
I checked the original PR #3994, but there is no clear description of overlapping as intended behavior. 

**To Reproduce**
1. Go to Lab grading of gradable. 
2. Scroll down
3. See error

**Configuration**
 - OS: [e.g. Windows, Linux, ...]

Mac / chrome

**Screenshots**
![Screen Shot 2023-04-12 at 22 37 38](https://user-images.githubusercontent.com/123261952/231633637-44bf58eb-a89e-4895-aea3-5711a9f3f920.png)

![Screen Shot 2023-04-12 at 22 38 14](https://user-images.githubusercontent.com/123261952/231633650-092877eb-638c-42d4-8992-3653e07e80d4.png)

**Additional context**

Checked mobile view, and it stays in its original position when scrolling. (works as expected)
iOS / Safari

Non critical, yet bugged me for the whole semester.

yanliw123:
i would like to work on this

PR #9180, opened 2023-04-14
[Bugfix:InstructorUI] student search function overlap

yanliw123:
i think the original purpose of the function for student search is to have it follow when scroll. So the user may use the search function whenever instead of scrolls to top every time. But I removed all related codes since the purpose of issue #9177 is to have the student search function fixed at its position.

Search for student, undo, redo button overlaps with Color info of check points, when scrolling down.

### What is the new behavior?
Search for student function now fixed at its position. 
### Other information?

To Reproduce

Go to Lab grading of gradable.
Scroll down


ANS1514 (approved):
The change works as intended when I tested locally. Could be worth asking some TA's if they'd like the sticky search bar but that's a conversation for the future. Overall the fix works for the issue it looks to solve.

yanliw123:
> The change works as intended when I tested locally. Could be worth asking some TA's if they'd like the sticky search bar but that's a conversation for the future. Overall the fix works for the issue it looks to solve.

I discussed this with Jaeseok, and we believe that ctrl + f can do the same thing. However, I can try to make it sticky if overall more people believe that's a better way of solving this.

ziesski (approved):
Works as intended, I checked to see if there is any code dangling, and it seems good to me.

bmcutler:
@yanliw123 please resolve merge conflicts on this PR

Issue #9207, opened 2023-04-18
Course Materials Folder Open/Close All Folders


LittleSilver33:
The Open/Close all folder button does not accurately adjust the image of the folder that is present. For example, if a button is opened and then the close all folders button is clicked then the folder is closed but the image still shows an open folder image. 

When the open/close folder button is pressed then every folder image should revert to the intended state. For example, if the folder was open and then is closed by the button, then the folder should then be closed. Additionally, when every folder is tried to be opened using the button then only the first folder in the list is adjusted to be shown open. No sub-folder is adjusted, nor is any other folder. 

**To Reproduce**
1. Open a folder
2. Press close all folders button
3. Image does not show closed folder
And
1. Press open all folders button
2. Only the first outer folder shows as open

<img width="234" alt="Screenshot 2023-04-18 at 5 02 12 PM" src="https://user-images.githubusercontent.com/84456159/232904482-0e1c2605-e3e6-4a41-bc6a-d81b16c754c9.png">

Jack-Kiefer:
I would like to work on this!

LittleSilver33:
> I would like to work on this!

As long as you make a PR, that is fine. Please follow the conventions laid out on Submitty.org. Feel free to assign me as a reviewer and ping me on GitHub when you need a review when it is finished.

Suryansh5545:
Is this issue to fix the image display only? or functionality too? I have a pr open for fixing the folder open and closing but unable to resolve the cypress check fail as it is happening on random on my machine. #8991

Jack-Kiefer:
Awesome we'll do that!

PR #9212, opened 2023-04-23
[Bugfix:CourseMaterials] Change image when Open/Close all folders

Yujjio:

According to [#issue9207](https://github.com/Submitty/Submitty/issues/9207), if a folder is opened and the close all folders button is clicked then the folder is closed but the image still shows an open folder image. 

### What is the new behavior?
A wrong jQuery selector causes the bug. By adding a for loop, every folder image would change after the user click the Open/Close All Folders button.


BairongZhang1 (approved):
Work properly.

PR #9226, opened 2023-04-25
[Feature:InstructorUI] Abbreviated Name Format Option

### LINKED ISSUE #8436: Add user Profile Option for Family/Last Name Initial/Abbreviation
We currently assume that the users last/family/sur name should be abbreviated by taking only the first initial.  

* This is correct for users with a single last name.
'John' 'Smith' =>  "John S."

* However, this is incorrect for users with multiple last/family/sur names separated by a space:
'John' 'Smith Jones'  => "John S.J." or "John SJ." or "John SJ"

* Similarly, users with hyphenated last/family/sur names may want both initials:
'John' 'Smith-Jones'  => "John S-J."

* And let's add other standard name abbreviation conventions to this list...

Let's add an option to the user profile page.
By default, we'll chose the first option, to grab the first letter of their family/last/sur name string.
But add radio buttons to let them select one of several other options.
NOTE:  Let's not let them type anything for their last name string.  
It must be based on their preferred last/family/sur name and be 
one of our set of choices above (expanded to include all conventions).

Also, let's have an option to not display anything at all about the last name.
(Make this the last choice in the list.  Displaying some abbreviation about the last name 
is helpful in large classes where multiple users might have the same given/first name)

skara9:

Last initials are always obtained by taking the first letter of the user's family name.
Resolves #8436

### What is the new behavior?
There is an option to change how last initials are formatted, allowing users with more complex last names to abbreviate their name in the way they prefer.

image


site/app/models/User.php
```diff
@@ -449,8 +473,50 @@ public function getDisplayFullName() {
...
+        if ($last_initial_format === self::LAST_INITIAL_MULTI) {
+            $spaced = preg_split('/\s+/', $family_name, -1, PREG_SPLIT_NO_EMPTY);
+            foreach ($spaced as $part) {
+                $last_initial .= $part[0] . '.';
+            }
+        }
+        elseif ($last_initial_format === self::LAST_INITIAL_HYPHEN) {
+            $spaced = preg_split('/\s+/', $family_name, -1, PREG_SPLIT_NO_EMPTY);
+            foreach ($spaced as $part) {
+                $dashed = preg_split('/-/', $part, -1, PREG_SPLIT_NO_EMPTY);
```
williamjallen, line 491: No need to use `prep_split` for a single character here.  You can use `explode` for this.

williamjallen (changes requested):

site/app/controllers/UserProfileController.php
```diff
@@ -158,6 +158,33 @@ public function changeUserName() {
...
+                    $user->setUserUpdated(true);
+                    $this->core->getQueries()->updateUser($user);
+                    return JsonResponse::getSuccessResponse([
+                        'message' => "Last initial format successfully updated!",
+                        'format' => $user->getLastInitialFormat(),
+                        'display_format' => $user->getDisplayLastInitialFormat(),
+                        'new_abbreviated_name' => $user->getDisplayAbbreviatedName()
+                    ]);
+                } finally {
+                }
```
williamjallen, line 182: Why do we need an empty `finally` block here?

williamjallen (changes requested):

site/app/controllers/UserProfileController.php
williamjallen: I'm a little confused about what you're trying to do here.  Would it not be exactly equivalent to remove the `try-finally` stuff entirely?  If you don't catch an exception, it's going to fail either way.

williamjallen: I think the best way to do that would be to catch the specific exception you're expecting and return an error if so.  As it is now, the try-finally isn't doing anything to catch exceptions, so it will fail if an exception occurs.

williamjallen (approved):

PR #8921, opened 2023-02-17
[Feature:System] Pronouns info in user profile

Yujjio:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
No pronouns column on the user's profile page

### What is the new behavior?
A column is added to the user database that allows this information to be stored and a series of AJAX functions are created to make sure that user can submit successfully

### Other information?
haven't modified course database yet and the submit cannot work currently

PR #8803, opened 2023-02-01
[Testing:Autograding] Add unit tests for Autograding Shipper

IDzyre:
### What is the current behavior?
Currently there are only a few tests in the 'test_submitty_autograding_shipper.py' file, and none of them for the checkout_vcs_repo function. 

### What is the new behavior?
This PR adds tests for the checkout_vcs_repo function in the submitty_autograding_shipper file. It first attempts to checkout the repository, and succeed, having no 'failed' messages/files, while also checking that the paths and files checked out are correct. The second test, checks the output when the function fails to clone the repository. The third test checks the output when the function fails to construct a valid repository directory path.

### Other information?
This was tested using pytest.

PR #9227, opened 2023-04-25
[Testing:Developer] Remove Unused Config Parameter

IDzyre:
This changes one of the test configs brought in by #8803 to remove a parameter made obsolete by #8921 

@shailpatels wrote this as a comment for whichever was merged first, to change this afterwards. 
https://github.com/Submitty/Submitty/pull/8803/files/e67fdfc56e927f713f0993c206c37c2f4351c8fe#r1155356109

This change doesn't break anything, and the tests don't have to be edited because it is removing a parameter that wasn't being used in the tests anyway, but it will keep the test configs consistent with the actual configs used. 

I am also unsure if there is a better type/module for this PR's title.

### DIFF
--- a/autograder/tests/data/submitty_config.json
+++ b/autograder/tests/data/submitty_config.json
@@ -12,7 +12,6 @@
         "cgi_url": "https://submitty.test.com/cgi-bin",
         "websocket_port": 8443,
         "institution_name": "Test University",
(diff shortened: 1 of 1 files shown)


williamjallen:
Our current PR title convention doesn't support cases like this very well, but my suggestion would be to use `[Testing:Developer]` for the PR title.

PR #9229, opened 2023-04-27
[Bugfix:CourseMaterials] Fix Datepicker Scrolling

### LINKED ISSUE #9106: Date-Picker does not attach to the Pop-up they are opened in
**Describe the bug**
See the comment section of Issue [#4712](https://github.com/Submitty/Submitty/issues/4712).

When a date-picker is opened in a pop-up form, it is not controlled by the scrolling inside of the pop-up but instead gets attached to the main page. PR [#8893](https://github.com/Submitty/Submitty/pull/8893) tried to resolve this issue by setting it to static but experienced a error in WAVE test's error detection. (see the comments within that PR).

This appears to be the nature within all pop-up forms in Submitty when a modal or a detachable object is opened within them. 

**Expected behavior**
A date-picker calendar object or any modal/pop-up object should attach to the subpage or the pop-up component that calls it, and not allowed to be scrolled or invoked outside.

**To Reproduce**

An example:
1.  Go to Course Material tab.
2.  Click the Upload button or Edit Course Materials.
3.  Click on a date/time-selection window, a calendar would pop-up, then scroll the page.
4.  See error.

**Configuration**
- All

**Screenshots**
**Taken from the comment section of the above mentioned issue.*
image

**Additional context**
The issue maybe related to the way the "pop-up forms" are set up in Submitty. Note that in some part of the website, many of those and other modals are hacked for functionality.  A general fix or even a re-implementation of the pop-up should be presented in the future.

Chris' PR on implementing close-on-click feature to Pop-ups might be causing the issue with it not closing properly (PR [#7711](https://github.com/Submitty/Submitty/pull/7711)) (fixed)

hansongu123:

Although the problem of flatpickr elements not being closed properly and some related issues are resolved internally by the newest release, they would still only be scrollable through scrolling on the page directly under the pop-ups they're called in.

### What is the new behavior?
Flatpickr instances opened in modals or popups in the Course Materials page on Instructor's UI now will append to the parent elements or html input windows calling them.

Closes #9106 and #4712

### Other information?
This fix utilizes a combination of flatpickr's own configuration and some JavaScript for a temporary solution. In the long run, a replacement of the date-picker dependency or improvement on the pop-up forms across Submitty are expected.

Because flatpickr instances close and save the input only when clicked outside, I limited the trigger for pointer event from the whole flex-column to only on the input windows that call them in the Upload and 2 Edit forms, so it is easier to use.

### DIFF
--- a/site/app/templates/course/EditCourseMaterialsFolderForm.twig
+++ b/site/app/templates/course/EditCourseMaterialsFolderForm.twig
@@ -46,7 +46,7 @@
             <span>
                 <strong>Choose a time to release the files being uploaded:</strong>
             </span>
-            <input id="edit-folder-picker" class="edit-folder-picker type="text" size="26" onchange="enableFullUpdate()" style="pointer-events:auto;" ><br>
+            <input id="edit-folder-picker" class="edit-folder-picker" type="text" size="26" onchange="enableFullUpdate()" style="pointer-events:auto;" ><br>
             <span><em>Note: Everyone with grading access will be able to view all the materials irrespective of the release date/time.</em></span>
         </label>
         <label>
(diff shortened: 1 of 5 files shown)


williamjallen (changes requested):
It looks like the "Set Folder Release Date" button also needs this fix.  I also noted a few minor code changes.

PR #9230, opened 2023-04-27
[SECURITY][Bugfix:Calendar] Change HTML strings to JS elements

cjreed121:
HTML strings are used to create new elements for the calendar. This raises XSS security issues.

### What is the new behavior?
Instead of HTML strings, JS elements are created to properly escape any concerning values.

### Other information?
Tested locally and calendar still worked as expected.


williamjallen (approved):
Changes look good.  Building up the HTML like that looks rather painful and should probably be refactored in the future, but it's good enough for now.

PR #9231, opened 2023-04-27
[Feature:Autograding] Autograding example for Verilog

tkoz0:

There is no autograding example using Verilog.

### What is the new behavior?
An example has been added for Verilog autograding.

### Other information?
The docker container be pushed to a Submitty Docker Hub account before approving this PR.


AdamTuhacek (approved):
Looks good, I was able to set up a test gradable using the sample files and the autograder works as expected.

grading/execute.cpp
```diff
@@ -104,6 +104,8 @@ bool system_program(const std::string &program, std::string &full_path_executabl
     { "spim",                    "/usr/bin/spim" },
     { "clang",                   "/usr/bin/clang" },
     { "gdb",                     "/usr/bin/gdb" },
+    { "iverilog",                "/usr/bin/iverilog" },
```
prestoncarman, line 107: My question is more about the whole docker process then this PR.
* Where is the Submitty docker image defined. Does the image contain the Submitty code base?
* I am wondering why you need to add verlog commands here? My initial thought was they would only be accessible in the docker.

prestoncarman: Thanks for clarifying.

more_autograding_examples/verilog_hello_world/config/config.json
```diff
@@ -0,0 +1,44 @@
...
+    "autograding_method": "docker",
+    "autograding": {
+        // compile .v files
+        "submission_to_compilation": ["**/*.v"],
+        // run a.out
+        "compilation_to_runner": ["**/a.out"]
+    },
+    // use the verilog container for grading
+    "container_options": {
+        "container_image": "tkoz0/submitty-verilog"
```
saumyaborwankar, line 18: I think this needs to be changed to "tkoz0/submitty-verilog:latest" as there is a need to add this custom docker to the "Docker UI" section which will only accept a docker with the correct tag <Image name>:<tag>. And currently there is no way of manipulating the alias of the docker added.

saumyaborwankar (changes requested):
I think there is just one change needed in the config.json file. Along with rebasing with main. I had to bring your changes to my branch and test them due to an error while executing the `submitty_install` inside the VM. 


"ERROR: docker.errors.DockerException: Error while fetching server API version: request() got an unexpected argument 'chunked'"

prestoncarman:
Would it make sense to have a Submitty supported **verilog** docker image? Or is this not a common use case?

bmcutler:
@saumyaborwankar please update now that https://github.com/Submitty/DockerImages/pull/17 was merged

cjreed121 (approved):
Tested locally and works as expected

PR #9232, opened 2023-04-27
[Feature:Calendar] Global Calendar

LittleSilver33:
What is the current behavior?
Currently, the course calendar and global calendar are two separate pages which are treated independently. The intended behavior of this PR is to consolidate the course calendar and global calendar into just a single global calendar. 

What is the new behavior?
Allow instructors to add items, edit items, and delete items to the calendar from the global calendar from each of their different courses.


AdamTuhacek (commented):
I did some testing, mostly looks pretty good. The only things I noted:
- I wasn't able to create Announcements through the New Calendar Item button, I was getting the "This is not a valid calendar item" error.
- When editing a item, it seems like the calendar always moves to the current month after submitting. For example, when editing a item in February it always moves me back to April after submitting. Maybe it would be nice to stay in the current month, not sure if this is intended or not.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -78,52 +51,75 @@ public function createMessage(): RedirectResponse {
...
-            $this->core->addErrorMessage($e->getMessage());
-            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+            $this->core->addErrorMessage("Text exceeds 255 characters which is not allowed");
+            return new RedirectResponse($this->core->buildUrl(['calendar']));
+        }
+
+        if (isset($_POST['course'])) {
+            $set_course = $_POST['course'];
+        }
+        else {
```
EdenBendheim, line 93: Is it possible to instead of redirecting them back to the calendar page, to instead just show an error message and let them continue? I see this happen a lot and it could probably get annoying. Again if its too much work a future PR could try to get around this.

EdenBendheim (changes requested):
I also ran into an issue where when trying to make a new announcement, I got a "That is not a valid calendar type" error.
image
I think just making a new announcement is broken.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -78,52 +51,75 @@ public function createMessage(): RedirectResponse {
...
+            return new RedirectResponse($this->core->buildUrl(['calendar']));
         }
         $calendar_item->setDate(new \DateTime($date));
         try {
             $calendar_item->setText($text);
         }
         catch (\InvalidArgumentException $e) {
-            $this->core->addErrorMessage($e->getMessage());
-            return new RedirectResponse($this->core->buildCourseUrl(['calendar']));
+            $this->core->addErrorMessage("Text exceeds 255 characters which is not allowed");
```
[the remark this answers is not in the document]
LittleSilver33: That is absolutely a good change. I agree that this should probably be a future PR.

site/app/controllers/calendar/CalendarController.php
LittleSilver33: This is definitely a good suggestion. I think that for now, this is a good start. If this is a problem that wants to be addressed in the future, it could become its own PR.

LittleSilver33:
> I did some testing, mostly looks pretty good. The only things I noted:
> 
> * I wasn't able to create Announcements through the New Calendar Item button, I was getting the "This is not a valid calendar item" error.
> * When editing a item, it seems like the calendar always moves to the current month after submitting. For example, when editing a item in February it always moves me back to April after submitting. Maybe it would be nice to stay in the current month, not sure if this is intended or not.

I just fixed the first issue. Committing shortly. As for the second issue, I think that would be a good addition. As for the specifics of it, I also mentioned something similar to Eden about changing the view and how that also defaulted back to today. I think that we are going to make this in a future PR although I do think it's a great idea. Thanks for the feedback!

LittleSilver33:
> I also ran into an issue where when trying to make a new announcement, I got a "That is not a valid calendar type" error. image I think just making a new announcement is broken.

I believe I have fixed the issue. Requesting a rereview shortly.

AdamTuhacek (commented):
Nice, the announcements seem to be working now. I did notice that when I logged in as student I was able to open the edit menu, I wasn't able to actually make edits, but that probably shouldn't happen. 

Also, I was able to crash the site by entering a really long message text. Probably not a big deal but maybe better to check the input length somewhere.

LittleSilver33:
> Nice, the announcements seem to be working now. I did notice that when I logged in as student I was able to open the edit menu, I wasn't able to actually make edits, but that probably shouldn't happen.
> 
> Also, I was able to crash the site by entering a really long message text. Probably not a big deal but maybe better to check the input length somewhere.

That is something I will look into. It definitely should not happen, you are correct. After finals I'll get it sorted.

LittleSilver33:
> Nice, the announcements seem to be working now. I did notice that when I logged in as student I was able to open the edit menu, I wasn't able to actually make edits, but that probably shouldn't happen.
> 
> Also, I was able to crash the site by entering a really long message text. Probably not a big deal but maybe better to check the input length somewhere.
I have fixed the first issue. As for the second, I check for the input length now and return an error if it is too long. As for visuals, if you get anywhere above 50 characters (which is still over 200 below the character limit) it appears to be too long for the box. I believe @EdenBendheim has a PR on some CSS changes for the calendar in the works so I will likely leave that up to him.

williamjallen (changes requested):
I think this should be good to go after a final review once these changes are addressed.

LittleSilver33:
> Maybe it would also make sense to error out a calendar item with no message text? They look kind of weird, and I don't think there's a way to delete an empty note since there's no border/text to click on.
> 
> image

This is a great catch. I'll add that tomorrow if I have time.

AdamTuhacek (approved):
Functionality looks good now, I can't find any more errors. Might be useful to have someone else to look at the code as well before it's merged.

04vmatsibekker (commented):
I think this looks really great. When I was testing it, after the instructor made some announcements and notes for different courses, it was hard to distinguish where announcements and notes were for which courses. This may be a feature for a future PR though. But awesome job!

bmcutler (changes requested):
@LittleSilver33 (or @04vmatsibekker or @Youssef-Chip)
Sorry, I tried to resolve the merge conflict but the page is now broken:

Screen Shot 2023-06-06 at 11 18 28 PM

site/app/controllers/calendar/CalendarController.php
```diff
@@ -103,27 +76,50 @@ public function createMessage(): RedirectResponse {
...
+        }
+        else {
+            $this->core->addErrorMessage("Invalid course given.");
+            return new RedirectResponse($this->core->buildUrl(['calendar']));
+        }
+
+        $instructor_courses = $this->core->getQueries()->getInstructorLevelUnarchivedCourses($this->core->getUser()->getId());
+        $exists = false;
+        foreach ($instructor_courses as $course) {
+            if ($set_course === ($course['semester'] . $course['course'])) {
```
cjreed121, line 101: It would be better to check semester and course name each on their own. Here's an example where this could cause issues as is. Let's say there is are two semesters, one with code 's2' and another with code 's22'. Then there is also courses '2' and '22' in each semester. Instructors in semester 's2' course '22' could now create items on semester 's22' course '2'.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -132,76 +128,130 @@ public function editMessage(): RedirectResponse {
...
+                    return new RedirectResponse($this->core->buildUrl(['calendar']));
+                }
+                $calendar_item->setDate(new \DateTime($date));
+                try {
+                    $calendar_item->setStringType($type);
+                }
+                catch (\InvalidArgumentException  $e) {
+                    $this->core->addErrorMessage($e->getMessage());
+                    return new RedirectResponse($this->core->buildUrl(['calendar']));
+                }
```
cjreed121, line 199: You could consolidate the two try catches by putting the statements inside both try's inside one try catch instead.

site/app/views/calendar/CalendarView.php
```diff
@@ -44,7 +43,6 @@ public function showCalendar(CalendarInfo $info, bool $in_course = false): strin
         $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('flatpickr', 'plugins', 'shortcutButtons', 'themes', 'light.min.css'));
         $this->core->getOutput()->addInternalCss('table.css');
         $this->core->getOutput()->enableMobileViewport();
-        $this->core->getOutput()->addBreadcrumb($in_course ? "Course Calendar" : "Calendar");
```
cjreed121, line 47: Should we still have at least a calendar breadcrumb?

site/app/views/calendar/CalendarView.php
```diff
@@ -6,7 +6,6 @@
 
 use app\libraries\FileUtils;
 use app\models\CalendarInfo;
-use app\models\User;
```
cjreed121, line 9: This needs to stay to fix the issue Barb pointed out.

cjreed121 (changes requested):

site/app/controllers/calendar/CalendarController.php
04vmatsibekker: To fix this issue, I had the $set_course incorporate a space between the semester and the course, I think this will improve readability when Instructors are specifying their course (In the problem you mention before if  professor was teaching "s2 22" and "s22 2", both would appear to read as "s222".) But this will also fix the issue you mention in your comment.

Youssef-Chip (approved):

site/app/controllers/calendar/CalendarController.php
04vmatsibekker: It fixes issue because now the $set_course for these two would be different. One would be "s2 22" the other would be "s22 2"

04vmatsibekker (approved):

PR #9239, opened 2023-04-27
[UI/UX:Forum] Websockets post ordering

### LINKED ISSUE #5683: Order new posts fetched with websockets
The current behavior of adding new posts fetched with websockets is working only for _"Hierarchical"_ ("tree") option while there exists other options a user can sort viewed posts with including, but not limited to, _Chronological_, _Alphabetical_ and _Alpha by Registration_.

- Check for the cookie **"forum_display_option"** 
- Switch cases for the value of the cookie and add the new fetched post in its correct position, see https://github.com/Submitty/Submitty/blob/master/site/public/js/forum.js#L238

**Additional context**
To enable websockets please check https://submitty.org/sysadmin/version_notes/v20.06.00

ANS1514:

Posts on the forum would come in with websockets out of order in most ordering selections except for hierarchical. 

### What is the new behavior?
closes #5683
Looks to have posts come in with websockets in the correct ordering based off the forum display option

### Other information?

Tested by adding in posts to a thread with every forum display option and seeing where the websocket would load in the post.


AdamTuhacek (commented):
Works fine in all cases other than chronological descending. It looks like the first post goes to the top and the successive posts go below.

04vmatsibekker (commented):
Sátvik is correct. When a new post in made for someone in hierarchical mode, the change appears in hierarchical mode for all other users in real-time even if their forum post order are not in hierarchical mode.

image

04vmatsibekker:
Also if create a reply for the first post, it works fine, but if you try to reply to a later post, the problem occurs

04vmatsibekker:
Now this PR should work again without having to revert prior work

Youssef-Chip (approved):

PR #9254, opened 2023-04-28
[Feature:Calendar] Multi week view

EdenBendheim:

Only monthly view is available.

### What is the new behavior?
Creates a dropdown to view either the month, two weeks, or one week. Also adds cookies to keep the current view.

### Other information?
Testing the PR should be pretty simple. As long as the arrows in the other views move you forward and back one week for every view that is not month view, along with the page rendering correctly.


site/app/templates/calendar/Calendar.twig
```diff
@@ -8,10 +8,25 @@
...
+        if(x.value === "month"){
+            document.cookie = "view=month";
+        }
+        else if(x.value === "two_week"){
+            document.cookie = "view=two_week";
+        }
+        else if(x.value === "one_week"){
+            document.cookie = "view=one_week";
+        }
+        console.log(x.value);
```
cjreed121, line 27: Remove console log

Vagrantfile
```diff
@@ -219,4 +219,4 @@ Vagrant.configure(2) do |config|
     config.ssh.password = 'vagrant'
     config.ssh.insert_key = 'true'
   end
-end
```
cjreed121, line 222: Un-needed change

site/app/templates/calendar/Calendar.twig
```diff
@@ -25,9 +40,15 @@
                     <i class="fas fa-plus"></i> New Item
                 </button>
             {% endif %}
+            
+        <select id="calendar-item-type-edit" name="type" onchange="changeView(this);">
+                    <option value="month" {% if (cookies["view"] is defined) and cookies["view"] == "month" %} selected {% endif %}>Month</option>
+                    <option value="two_week" {% if (cookies["view"] is defined)  and cookies["view"] == "two_week" %} selected {% endif %}>Two Week</option>
+                    <option value="one_week" {% if (cookies["view"] is defined)  and cookies["view"] == "one_week" %} selected {% endif %}>One Week</option>
+                </select>
```
cjreed121, line 48: Can we format this a bit nicer?

site/app/views/calendar/CalendarView.php
```diff
@@ -59,7 +59,8 @@ public function showCalendar(CalendarInfo $info, bool $in_course = false): strin
             "empty_message" => $info->getEmptyMessage(),
             "in_course" => $in_course,
             "is_instructor" => $this->core->getUser()->getGroup() === User::GROUP_INSTRUCTOR,
-            "colors" => $info->getColors()
+            "colors" => $info->getColors(),
+            "cookies" => $_COOKIE
```
cjreed121, line 63: Probably better to just pass what's relevant for cookies rather than every cookie. So just $_COOKIE['view'] assuming you don't need anything else.

cjreed121 (changes requested):

Vagrantfile
EdenBendheim: Hey, I'm not sure how to remove this change, I don't want to change the vagrant file, and I didn't even realize I was changing it (my bad). Also is it a problem that this PR has all of the previous commits?

site/app/views/calendar/CalendarView.php
EdenBendheim: Thank you! This makes it a lot cleaner in the twig too 👍

site/app/views/calendar/CalendarView.php
LittleSilver33, line 1: If you view the page for the first time then $_COOKIE['view'] is not set. This gives an undefined index problem. You may want to assign a default value of 'month' if $_COOKIE['view'] is undefined.

site/app/templates/calendar/Calendar.twig
LittleSilver33, line 1: If I am currently looking at the month view of a given month in the calendar, and then I switch the view to something else, the calendar will default back to today. Is there any way that you can pass a view_day value into the rendering of the Twig template in CalendarView.php to stay in roughly the same area? Going from a more specific view to a less specific view should be easy, simply set the first day of the week being shown as the first day of the first week being shown. However, changing to a more specific day might be more complicated.

LittleSilver33 (changes requested):

site/public/js/calendar.js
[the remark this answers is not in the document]
EdenBendheim: No, currently it doesn't do anything. I'm planning on making it so that when it does come through, you can just view your semester. Maybe the arrows can lead you back or forward a semester? It's gonna require changing the way the header load function works so ill just take the code out and save it for a later PR.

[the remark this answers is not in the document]
EdenBendheim: The reason type is passed in is that the function that array calls needs it to know whether it will view a one-week or a two-week (since I use the prev and next Week funcitons for both weeks). These get passed directly into another function, so I can't really use push.

site/app/views/calendar/CalendarView.php
EdenBendheim: Thank you!

site/app/templates/calendar/Calendar.twig
EdenBendheim: This is a good feature, but I think ill make it a separate PR.

site/app/templates/calendar/Calendar.twig
```diff
@@ -8,10 +8,24 @@
     const isGlobal = {% if not in_course %}true{% else %}false{% endif %};
     const isInstructor = {% if is_instructor %}true{% else %}false{% endif %};
 
-    window.addEventListener("load", function () {
-        loadCalendar({{ view_month }}, {{ view_year }});
+    $( document ).ready(function() {
+        // Get value of the type dropdown
+        var type = $("#calendar-item-type-edit").val();
```
williamjallen, line 13:
```suggestion
        const type = $("#calendar-item-type-edit").val();
```

site/app/templates/calendar/Calendar.twig
```diff
@@ -8,10 +8,24 @@
...
+    window.changeView = function(x){
+        if(x.value === "month"){
+            document.cookie = "view=month";
+        }
+        else if(x.value === "two_week"){
+            document.cookie = "view=two_week";
+        }
+        else if(x.value === "one_week"){
+            document.cookie = "view=one_week";
+        }
```
williamjallen, line 26: We should try to maintain the eslint formatting, even though eslint can't lint twig files.

```suggestion
        if (x.value === "month") {
            document.cookie = "view=month";
        }
        else if (x.value === "two_week") {
            document.cookie = "view=two_week";
        }
        else if (x.value === "one_week") {
            document.cookie = "view=one_week";
        }
```

site/app/templates/calendar/Calendar.twig
```diff
@@ -8,10 +8,24 @@
...
 
-    window.addEventListener("load", function () {
-        loadCalendar({{ view_month }}, {{ view_year }});
+    $( document ).ready(function() {
+        // Get value of the type dropdown
+        var type = $("#calendar-item-type-edit").val();
+        loadCalendar(curr_month, curr_year, curr_day, type);
     });
 
+    window.changeView = function(x){
```
williamjallen, line 17: Is there any reason why you can't just use `function changeView(x) {` here?

site/app/views/calendar/CalendarView.php
```diff
@@ -59,7 +59,8 @@ public function showCalendar(CalendarInfo $info, bool $in_course = false): strin
             "empty_message" => $info->getEmptyMessage(),
             "in_course" => $in_course,
             "is_instructor" => $this->core->getUser()->getGroup() === User::GROUP_INSTRUCTOR,
-            "colors" => $info->getColors()
+            "colors" => $info->getColors(),
+            "view_cookie" => isset($_COOKIE['view']) ? $_COOKIE['view'] : "view=month"
```
williamjallen, line 63: Why is the string we're passing around here `view=month`?  Would it not be more appropriate to just pass `month`?

site/public/js/calendar.js
```diff
@@ -12,13 +12,13 @@ const monthNamesShort = ['Dec', 'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul',
  * @param year : int the current year
  * @returns {number[]} : array<int> {previous_month, year_of_previous_month}
```
williamjallen, line 13: Is this return type still correct?  Should a new parameter be specified here?

site/public/js/calendar.js
```diff
@@ -28,15 +28,51 @@ function prevMonth(month, year) {
  * @param year : int the current year
  * @returns {number[]} : array<int> {next_month, year_of_next_month}
  */
-function nextMonth(month, year) {
+function nextMonth(month, year, day) {
```
williamjallen, line 31: Same comment here

site/public/js/calendar.js
```diff
@@ -28,15 +28,51 @@ function prevMonth(month, year) {
...
-    return [month, year];
+    return [month, year, day, 'month'];
 }
 
+/**
+ * Gets the previous week of a given month
+ * @param month : int the current month (1 as January and 12 as December)
+ * @param year : int the current year
+ * @param day : int the current day
+ * @returns {number[]} : array<int> {previous_month, year_of_previous_month}
```
williamjallen, line 45: The specified return type doesn't seem to match what you're actually returning here.  You're returning an array of three things.

site/public/js/calendar.js
```diff
@@ -28,15 +28,51 @@ function prevMonth(month, year) {
...
+}
+
+/**
+ * Gets the next week of a given week
+ * @param month : int the current month (1 as January and 12 as December)
+ * @param year : int the current year
+ * @param day : int the current day
+ * @returns {number[]} : array<int> {next_month, year_of_next_month}
+ */
+function nextWeek(month, year, day) {
```
williamjallen, line 66: Same here

site/public/js/calendar.js
```diff
@@ -346,9 +404,9 @@ function buildSemesterHeader(semester_name) {
  * @param view_month : int month that the calendar is viewing (1 as January and 12 as December)
  * @returns {HTMLElement} the HTML Element with the entire calendar
  */
-function generateCalendarOfMonth(view_year, view_month) {
+function generateCalendarOfMonth(view_year, view_month, view_day) {
```
williamjallen, line 407: Documentation for this new parameter should be added

site/public/js/calendar.js
```diff
@@ -396,6 +454,132 @@ function generateCalendarOfMonth(view_year, view_month) {
     return table;
 }
 
+/**
+ * This function creates a table that shows the calendar for one week.
+ *
+ * @param view_year : int year that the calendar is viewing
+ * @param view_month : int month that the calendar is viewing (1 as January and 12 as December)
+ * @param view_day : int day that the calendar is viewing
+ * @returns {string} the HTML string contains the entire calendar table displaying view_month/view_year
```
williamjallen, line 463: Doesn't this function actually return an `HTMLElement`, similar to the function above it?

site/public/js/calendar.js
```diff
@@ -396,6 +454,132 @@ function generateCalendarOfMonth(view_year, view_month) {
...
+    return table;
+}
+
+/**
+ * This function creates a table that shows the calendar for two weeks.
+ *
+ * @param view_year : int year that the calendar is viewing
+ * @param view_month : int month that the calendar is viewing (1 as January and 12 as December)
+ * @param view_day : int day that the calendar is viewing
+ * @returns {string} the HTML string contains the entire calendar table displaying view_month/view_year
```
williamjallen, line 520: same here

site/public/js/calendar.js
```diff
@@ -408,12 +592,9 @@ function generateFullCalendar(start, end, semester_name) {
     // Header area: two buttons to move, and month
     const table = generateCalendarHeader(buildSemesterHeader(semester_name));
     const tableBody = document.createElement('tbody');
-
```
williamjallen, line 411: All of these whitespace changes are unnecessary and should be reverted due to the risk of merge conflicts, given that multiple PRs are touching similar code.  Apply throughout.

site/public/js/calendar.js
```diff
@@ -451,20 +628,28 @@ function generateFullCalendar(start, end, semester_name) {
...
 
+
 /**
  * Changes the calendar div to the required month and year.
  *
  * @param month_ : int month that the calendar will show (1 as January and 12 as December)
  * @param year_ : int year that the calendar will show
  */
-function loadCalendar(month_, year_) {
+function loadCalendar(month_, year_, day_, type) {
```
williamjallen, line 641: The documentation for this function should be updated as well

williamjallen (changes requested):

site/public/js/calendar.js
EdenBendheim: Oh, yea those were accidental, I appreciate you catching them. Is there any way to just easily remove them?

site/app/templates/calendar/Calendar.twig
EdenBendheim: Thannk you!

EdenBendheim: They're all set with view= before the view type, so in the case we don't have a cookie, I'm pretending there is a view one.

site/app/views/calendar/CalendarView.php
EdenBendheim: I passed it as "view=month" since if the cookie doesn't exist, then I pass in a string instead of a cookie.

EdenBendheim: it would!

williamjallen (approved):
Works well!

04vmatsibekker (approved):

PR #9309, opened 2023-05-15
[Feature:Submission] Timezone mismatch warning

### LINKED ISSUE #8574: Add a timezone check against the user's local timezone.

skara9:

No warnings for mismatch between user timezone and local.
Resolves #8574 

Note: this is a revision of PR #8647 by @marymx95

### What is the new behavior?
A warning message is displayed on page load in the case where the local timezone does not match the one stored in the user's profile. Once displayed, the warning will only appear again if an hour has passed since the last warning and the timezones are still different.

image


site/public/js/server.js
```diff
@@ -1784,3 +1784,33 @@ function addMarkdownCode(type){
...
+
+    const user_off = (h + m / 60) * (user_offstr[0] === '-' ? -1 : 1);
+
+    const local_off = new Date().getTimezoneOffset() / -60;
+    if (user_off === local_off) {
+        return;
+    }
+
+    const THRESHOLD = 60*60*1000; // will wait one hour after each warning
+    const last = Number(document.cookie.replace(/(?:(?:^|.*;\s*)last_tz_warn_time\s*\=\s*([^;]*).*$)|^.*$/, "$1"));
```
williamjallen, line 1809: It is generally good to add a comment stating what a complicated regex like this is doing.

williamjallen (approved):
Code changes seem reasonable to me.  I haven't tested it thoroughly, so it might be good for someone else to take a look as well.

PR #9334, opened 2023-05-23
[SYSADMIN ACTION][Refactor:System] PHP8 / Ubuntu 22 Upgrade

cjreed121:

We are currently using PHP 7.4 which has reached its end of life. We are also running Ubuntu 20.04 which uses PHP 7.4 as it's primary PHP version.

### What is the new behavior?
Submitty will now use PHP 8.1. Ubuntu 22.04 is now used for Vagrant as its PHP default version is 8.1

### Other information?
Most pages seem to load fine. Much more testing is still needed.

This PR upgrades both PHP and Ubuntu at the same time to avoid the PPA but I can split the two version bumps if needed.


cjreed121:
> Initial installation fails at:
> 
> 
> 
> ```
> 
> ubuntu-22.04: # Need to change this otherwise it will hang the script in interactive mode
> 
> ubuntu-22.04: sed -i "s/#\$nrconf{restart} = 'i';/\$nrconf{restart} = 'a';/" needrestart.conf
> 
> ubuntu-22.04: sed: can't read needrestart.conf: No such file or directory
> 
> ubuntu-22.04: err_message
> 
> ```

Thanks for testing. I know the issue and will edit it.

NiaHeermance (commented):
The installation failed for me as well. I've attached my [install_ubuntu_22.04.log file](https://github.com/Submitty/Submitty/files/11579860/install_ubuntu_22.04.log).

site/app/libraries/GradingQueue.php
```diff
@@ -361,6 +361,9 @@ private function updateDetailedQueueCount(
         array $open_autograding_workers_json,
         array &$course_info
     ): array {
+        if ($queue_or_grading_file === "/var/local/submitty/to_be_graded_queue/s23__sample__future_no_tas_homework__instructor__2") {
+            $thing = 2;
+        }
```
williamjallen, line 366: Should this be removed?

site/app/views/GlobalView.php
```diff
@@ -17,17 +17,16 @@ public function header($breadcrumbs, $wrapper_urls, $sidebar_buttons, $notificat
...
-        }
-
         $page_title = "Submitty";
         if ($this->core->getUser() === null) {
             $page_title = "Login";
         }
         elseif ($this->core->getConfig()->isCourseLoaded()) {
+            $course_name = ucwords(strtolower($this->core->getFullCourseName()));
+            // We assume that if there is no page breadcrumb (only course), we are on gradeables
+            if ($course_name == ucwords(strtolower($page_name))) {
```
williamjallen, line 27:
```suggestion
            if ($course_name === ucwords(strtolower($page_name))) {
```

williamjallen (commented):
When we moved from 18.04 to 20.04, we retained the ability to create both kinds of VMs at the same time, and adjusted all the ports by one.  Should we do that again?  It's probably unlikely that anyone will want to use 20.04 once 22.04 is available, but it might also be nice to have the ability to support it in case we need to debug a 20.04-specific issue on production for whatever reason.

bmcutler:
@cjreed121 
Can you please resolve conflicts
And remove it from draft stage if you think it's finished (if you have no remaining tasks)

cjreed121:
> @cjreed121 Can you please resolve conflicts And remove it from draft stage if you think it's finished (if you have no remaining tasks)

I can resolve the merge conflicts although every time we merge a PHP dependency PR we'll need to resolve merge conflicts. So I'd prefer not to until this is about to be merged or we really want to test a new feature. This PR (https://github.com/Submitty/Submitty/pull/9361) needs to get merged before this.  I don't know much about it's usage. There seems to be an issue with one of the integration tests but it passes locally. After those I think it's good to come out of draft stage (unless you want me to pull it out of draft now).

If anyone else testing this PR wants to try running that failing test locally to see if it passes or fails, that would be much appreciated.

cjreed121:
> image
> 
> I get this image when I try to switch to this branch and run submitty_install.

Switching to this branch requires a new VM since it bumps the entire OS version. So you can do a vagrant destroy and up or you could vagrant up in a new dir if you don't want to destroy existing.

cjreed121:
> > > image
> 
> > > I get this image when I try to switch to this branch and run submitty_install.
> 
> > 
> 
> > Switching to this branch requires a new VM since it bumps the entire OS version. So you can do a vagrant destroy and up or you could vagrant up in a new dir if you don't want to destroy existing.
> 
> 
> 
> So basically I would vagrant destroy on main branch, then switch to this PR, and then do vagrant up?

That would work, yes.

ziesski:
I did see a green duck with installation success, but I cannot login with any id (instructor, student, aphacker, bitdiddle etc)
Although line "  ubuntu-22.04: Creating user Ben Bitdiddle (bitdiddle)... " was on the terminal installation process. 
Screen Shot 2023-06-17 at 8 50 52 PM

Screen Shot 2023-06-17 at 8 51 12 PM

bmcutler:
@cjreed121 
Please resolve conflicts ASAP
thanks

bmcutler:
@cjreed121 
Failing php linting

.setup/install_system.sh
```diff
@@ -536,7 +536,7 @@ EOF
     # being that we do not disable phpinfo() on the vagrant machine as it's not a function that could be used for
     # development of some feature, but it is useful for seeing information that could help debug something going wrong
     # with our version of PHP.
-    DISABLED_FUNCTIONS="popen,pclose,proc_open,chmod,php_real_logo_guid,php_egg_logo_guid,php_ini_scanned_files,"
```
bmcutler, line 535: chmod was intentionally on our list of disabled functions:
https://submitty.org/sysadmin/installation/system_customization#disable-php-functions

If we can find an alternate workaround that would be preferred.

bmcutler (changes requested):

.setup/install_system.sh
cjreed121: Twig calls it and causes an issue if we don't let it. There isn't a way around that unless we modify their source code. I'm of course not changing the version of Twig so the code isn't changing. It seems that PHP 7 was happy ignoring errors of a disabled function being called. PHP 8 is more strict and now errors in that case. While I don't like removing functions from this list, I don't think removing chmod is of high issue/concern.

bmcutler: ok, we'll have to update the submitty.org docs too.

site/app/libraries/database/AbstractDatabase.php
```diff
@@ -172,11 +172,6 @@ public function query($query, $parameters = []) {
             elseif ($identity === QueryIdentifier::SELECT) {
                 $columns = $this->getColumnData($statement);
                 $this->results = $statement->fetchAll(\PDO::FETCH_ASSOC);
-                // Under normal circumstances, we don't really need to worry about $this->results being false.
-                // @codeCoverageIgnoreStart
-                if ($this->results === false) {
-                    return false;
-                }
                 // @codeCoverageIgnoreEnd
```
bmcutler, line 175: should this end comment also be removed

bmcutler (changes requested):

site/app/libraries/database/AbstractDatabase.php
cjreed121: Yeah I think it should. Missed that before. Thanks for catching it.

bmcutler (approved):

Issue #9342, opened 2023-05-24
PDF   ✦   (0/4)


skara9:

im-ajaymeena:
Can this issue be assigned to me?

PR #9355, opened 2023-05-26
[Testing:Developer] Cypress PDFs test

im-ajaymeena:
There are PDF access tests that are written in selenium, but not run in CI. 

### What is the new behavior?
These tests have been re-written in Cypress and _can_ now run in CI

### Other information?
These tests include testing that,
* instructors can access all types of gradeable's PDFs,
* ta's can access all types of gradeables PDF's,
* graders can access their assigned gradeable PDF's, and not other sections
* students can access peer assigned PDF's, and not other sections

Resolves https://github.com/Submitty/Submitty/issues/9342


im-ajaymeena:
Hi @skara9 , I had few queries 

- In e2e python test script there are tests for grader's pdfs access, when running locally and logged in as grader I couldn't find any submissions or any pdfs. I get this

image

If this is expected should I remove those tests (they are commented in currently test file)

im-ajaymeena:
@skara9 
Could you help me know why are the test failing, I think the actual tests are passing but the step involving 
`sudo bash tests/test_site_error_log.sh` fails. Is there anything missing in cypress test file?

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,105 @@
+import {getCurrentSemester} from '../support/utils.js';
+
+function switch_settings(setting, pdf_type) {
+    cy.visit(`/courses/${getCurrentSemester()}/sample`);
```
williamjallen, line 4: The `/courses/<semester>` is unnecessary here.  You can just use `cy.visit(['sample'])`, like the code [here](https://github.com/Submitty/Submitty/blob/36c58b64c779c4141bf4be995f2bf79dae2e4b03/site/cypress/e2e/late_days_allowed.spec.js#L4).  Apply throughout the rest of this PR.

This particular line is also completely unnecessary because you navigate to a page, and then promptly navigate away to log in.

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,105 @@
...
+    cy.login('instructor');
+    cy.get(`a[href*="/sample/gradeable/${pdf_type}/update"]`).click();
```
williamjallen, line 6: You don't need to click on links like this.  Just navigate to the page you want to cut down on the overall running time.  Apply throughout.

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,105 @@
...
+        cy.get(`a[href*='/sample/gradeable/${gradeable_id}/grading/status']`).click();
+        cy.contains('Grading Index').should('be.visible').click();
+    }
+    cy.get('#details-table').should('be.visible');
+    if (user_id !== 'student') {
+        cy.get('.markers-container .btn-default').each(($el) => {
+            if ($el.text().trim() === 'View All') {
+                cy.wrap($el).click();
+            }
+        });
```
williamjallen, line 43: Why do you need an `if` in a `.each()`?  Can't you just target the things you want and then click it?

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+    cy.get(`#details-table > tbody.details-content.panel-content-active > tr:nth-child(${tr_number}) > td:nth-child(${td_number}) > a`).eq(0).click();
```
williamjallen, line 45: Is there a more specific way to target this via an identifier of some sort?

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,105 @@
...
+describe('Test cases for PDFs access', () => {
+    it('users should have access to basic pdfs', () => {
+        const pdf_type = 'grading_homework_pdf';
+        switch_settings('Limited Access Grader', pdf_type);
+
+        pdf_access('instructor', '3', '8' ,pdf_type, 'words_1463.pdf');
+        pdf_access('ta', '3', '8', pdf_type, 'words_1463.pdf');
+        pdf_access('grader', '4', '8', pdf_type, 'words_881.pdf');
+
+        revert_settings(pdf_type);
```
williamjallen, line 69: `revert_settings()` should be run in a `.afterEach()` function to make sure all of the changes get reverted regardless of whether the test passes or fails.  Apply throughout.

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,105 @@
...
+    cy.get('#div_viewer_sd1').find(`a[file-url*='${pdf_name}']`).click();
+    cy.get('#pageContainer1').should('be.visible');
+    cy.get('a[onclick*="collapseFile"]').click();
+    cy.logout();
+}
+
+describe('Test cases for PDFs access', () => {
+    it('users should have access to basic pdfs', () => {
+        const pdf_type = 'grading_homework_pdf';
+        switch_settings('Limited Access Grader', pdf_type);
```
williamjallen, line 63: `switch_settings()` appears to be used at the start of every test case here.  It would be better to move it to a `.beforeEach()` function to reduce the amount of code duplication.  Apply throughout.

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,105 @@
...
+
+        pdf_access('instructor', '3', '8' ,pdf_type, 'words_1463.pdf');
+        pdf_access('ta', '3', '8', pdf_type, 'words_1463.pdf');
+        pdf_access('grader', '4', '8', pdf_type, 'words_881.pdf');
```
williamjallen, line 67: Should you also check to make sure students *can't* access the PDF?

williamjallen (changes requested):

williamjallen:
Please also fill out the PR body a bit with information about what exactly is being tested.

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,102 @@
...
+    cy.visit(['sample', 'gradeable', gradeable_id, 'grading', 'details']);
+    cy.get('[data-testid="banner-message"]').should('be.visible').and('contain.text', 'You do not have permission');
+}
+
+const gradeable_type = ['grading_homework_team_pdf', 'grading_homework_pdf', 'grading_pdf_peer_homework', 'grading_pdf_peer_team_homework'];
+
+describe('Test cases for PDFs access', () => {
+    before(() => {
+        cy.visit('/');
+        cy.login('instructor');
```
williamjallen, line 52: We don't need to visit `/` if we're going to log in right afterwards, right?

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,102 @@
...
+        // Change the settings that aren't already what we need
+        switch_settings('grading_homework_team_pdf');
+        switch_settings('grading_pdf_peer_homework');
+        switch_settings('grading_pdf_peer_team_homework');
+        cy.logout();
+    });
+
+    beforeEach(() => {
+        cy.visit('/');
+    });
```
williamjallen, line 62: Is this necessary?

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,102 @@
...
+    cy.get('[data-testid="details-table"]').should('be.visible');
+    select_gradeable();
+}
+
+function no_pdf_access(gradeable_id) {
+    cy.visit(['sample', 'gradeable', gradeable_id, 'grading', 'details']);
+    cy.get('[data-testid="banner-message"]').should('be.visible').and('contain.text', 'You do not have permission');
+}
+
+const gradeable_type = ['grading_homework_team_pdf', 'grading_homework_pdf', 'grading_pdf_peer_homework', 'grading_pdf_peer_team_homework'];
```
williamjallen, line 47: Just a comment: it is sometimes cleaner to format arrays with each element on a separate row so it's easier to read a diff if someone changes one single element in the future.

site/cypress/e2e/pdfs.spec.js
```diff
@@ -0,0 +1,102 @@
...
+    });
+    select_gradeable();
+}
+
+function minimum_pdf_access(gradeable_id) {
+    cy.visit(['sample', 'gradeable', gradeable_id, 'grading', 'details']);
+    cy.get('#agree-button').click({ force: true });
+    cy.get('[data-testid="details-table"]').should('be.visible');
+    select_gradeable();
+}
```
williamjallen, line 40: It would be good to add some sort of check to make sure we *don't* see stuff we're not supposed to see.  Otherwise, what makes this different from `check_pdf_access`?  I believe that this test case would pass any time `check_pdf_access` passes.

williamjallen (changes requested):

site/cypress/e2e/pdfs.spec.js
IDzyre: Cypress needs to have a location to login to, if you just do cy.login, since there is no webpage loaded, it doesn't find the elements.

site/cypress/e2e/pdfs.spec.js
IDzyre: The 'grader' and 'student' users don't have the ability to view different sections, like the TA and Instructor do, and since the TA is not assigned to all of the gradeables, you sometimes have to click the 'view all' button in order to view the table. This function _may_ succeed when check_pdf_access succeeds, however it is for the student and grader users, because the 'view all' or 'view sections' button isn't visible to them.

site/cypress/e2e/pdfs.spec.js
IDzyre: see comment above

site/cypress/e2e/pdfs.spec.js
IDzyre: This has been added.

04vmatsibekker (approved):
Tested it out locally. Looks good!

site/cypress/e2e/pdfs.spec.js
IDzyre: I have added a check to make sure the grader/student _don't_ see the 'view all' button.

williamjallen (approved):
Changes look good.  I left an inline comment with my suggested array formatting, but it's not a blocker for this getting merged.

PR #9356, opened 2023-05-26
[Feature:Submission] Remove Course-wide Grade Inquiry Setting

NiaHeermance:
Before, Submitty courses had a global grade inquiry setting. This setting had to be toggled on before individual gradeables could have grade inquiries activated in their gradeable-specific settings.

### What is the new behavior?
Now, the course-wide setting is removed, and grade inquiries are available to use always. When a new gradeable is created, grade inquiries are off by default, but they can easily be turned on during the gradeable's creation or through the gradeable's settings.


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -315,7 +315,7 @@ public function statusPage(
             "bulk_stats_url" => $this->core->buildCourseUrl(['gradeable', $gradeable->getId(), 'bulk_stats']),
             "details_url" => $details_url,
             "grade_url" => $this->core->buildCourseUrl(['gradeable', $gradeable->getId(), 'grading', 'grade']),
-            "regrade_allowed" => $this->core->getConfig()->isRegradeEnabled(),
+            "regrade_allowed" => false,
```
williamjallen, line 318: Is there a reason why this parameter can't be removed entirely?

NiaHeermance: Regrade_allowed is the course-specific grade inquiry setting, not the global setting. The global setting was removed. Unless you're specifically referring to the file ElectronicGraderView? To be honest I don't know much about this file; I think it can be removed as the issue never really came up in my testing until the unit tests online were done.

williamjallen: After looking into the issue further, I think leaving this as-is is probably fine for now.  I was primarily concerned about the fact that we were just passing a constant to a Twig file, but it might be a challenge to simplify the logic so I think it's fine.

bmcutler (approved):

PR #9381, opened 2023-06-01
[Testing:Developer] Selenium to Cypress grading_stats.spec

bobpaw:
Maintains current test functionality.
Continues to not run in CI.
Addresses #9346.

* [x] ~~Documentation has been updated/added if relevant (not relevant)~~

Grading stats info e2e test is checked in Selenium.

### What is the new behavior?
Grading stats info e2e test is checked in Cypress.

### Other information?

Needs to be reviewed locally.


site/cypress/e2e/grading_stats.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    describe('Grading stats as an instructor', () => {
+        beforeEach(() => {
+            cy.login('instructor');
+            cy.visit(['sample']);
+        });
+        afterEach(() => {
+            cy.logout();
+        });
+        it('should be accurate for teams.', () => {
+            cy.get('a[href~="/sample/gradeable/grading_team_homework/grading/status"]').click();
```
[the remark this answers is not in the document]
bobpaw: Actually these tests shouldn't be running in e2e at all. Line 1 is supposed to disable all tests in a CI environment. Latest commit should address this.

site/cypress/e2e/grading_stats.spec.js
```diff
@@ -0,0 +1,111 @@
+import { skipOn } from '@cypress/skip-test';
+
+skipOn(Cypress.env('run_area') === 'CI', () => {
```
williamjallen, line 3: Is there a reason why this can't run in the CI?  Tests which aren't run regularly as part of our CI are almost useless because they quickly become outdated.  Many of the Selenium tests aren't run as part of the CI for historical reasons, but our CI is now more complete and should be able to support almost all tests.  If we can't support the entire test, can some subset of the test be enabled at least?

bobpaw:
The sample course seems to have changed. The old spec checks for `Teams who have submitted: 32 / 36 (88.9%)`, but the new sample site has `Teams who have submitted: 30 / 36 (83.3%)`. Also, not sure if this is out because i'm running tests out of order, but I get a "Your responsibility as a grader" popup as a TA.

site/cypress/e2e/grading_stats.spec.js
```diff
@@ -0,0 +1,111 @@
...
+    describe('Grading stats as an instructor', () => {
+        beforeEach(() => {
+            cy.visit(['sample']);
+            cy.login('instructor');
+        });
+        afterEach(() => {
+            cy.logout();
+        });
+        it('should be accurate for teams.', () => {
+            cy.get('a[href*="/sample/gradeable/grading_team_homework/grading/status"]').click();
```
mahi-pas, line 13: This line fails when testing on my machine (0 matched elements)
`cy.get('a[href*="/sample/gradeable/grading_team_homework/grading/status"]').click();`

site/cypress/e2e/grading_stats.spec.js
```diff
@@ -0,0 +1,111 @@
...
+            text.contains('Students who have submitted: 73 / 101 (72.3%)');
+            text.contains('Current percentage of TA grading done: 73 / 73 (100.0%)');
+            text.contains('Section 1: 12 / 12 (100.0%)');
+            text.contains('Number of students who have viewed their grade: 50 / 73 (68.5%)');
+        });
+    });
+
+    describe('Grading stats as a ta', () => {
+        beforeEach(() => {
+            cy.login('ta');
```
mahi-pas, line 42: Not sure why, but "grading stats as a ta" fails for me while testing as before each. (get #saml-login 0 matches found) Likely because the instructor failed. For similar reasons, grader doesn't even run

mahi-pas (changes requested):
The test doesn't work on my machine, I have sent the issues I am encountering

site/cypress/e2e/grading_stats.spec.js
bobpaw: Sorry that was my bad. I corrected the order of cy.login() and cy.visit() in the beforeEach clauses.

bobpaw:
@williamjallen Should I change the expected output in the test? I'm not sure if the sample course has been updated since the Python test was created or if there is a different cause for the aforementioned difference. The other errors I'm getting are because of the grader responsibility popup and an error in how the cy.logout() command checks for a window reload. As I understand, Cypress is supposed to wait for clicks to load by default but in this case it isn't since the page url isn't changing for logout. A better option might be using [cy.intercept() and cy.as()](https://docs.cypress.io/faq/questions/using-cypress-faq#How-do-I-wait-for-multiple-requests-to-the-same-url) in cy.logout() but that constitutes a separate issue. Grader responsibility popup:
image
I can add a line to click close and it will popup again every time or I can click accept once.

bobpaw:
Added code to click the close popup button and removed the call to logout (which should be unnecessary because each test case starts in a new window). Also enabled the test in CI and switched to using `.should()` assertions instead of chaining `.contains()`. Now the checks are failing because the actual text on the sample is different. Also, one TA page has multiple close buttons, some of which are hidden?

NiaHeermance:
Sorry, this is a silly question, but how do I review a PR like this where the functionality is Github CI related?

mahi-pas:
> Sorry, this is a silly question, but how do I review a PR like this where the functionality is Github CI related?

Run cypress on your machine and make sure it works well. Make sure the tests are well written and don't have any flaws
https://submitty.org/developer/testing/cypress

NiaHeermance (commented):
grading_stats indeed passed locally!

NiaHeermance (approved):
Meant to click approve.

PR #9393, opened 2023-06-05
[Testing:Forum] Part replace of test_forum_merge_threads

IDzyre:
Selenium tests are flaky, so we are porting them to cypress. 

### What is the new behavior?
This is a basic forum test, which adds three different forum threads, comments to them, merges them into one, then deletes the resulting thread.


mahi-pas:
Should this be a draft?

site/cypress/e2e/test_forum.spec.js
```diff
@@ -0,0 +1,125 @@
...
+const reply3 = 'Cypress Reply 3 Cypress';
+const merged1 = 'Merged Thread Title: '.concat(title3, '\n\n', content3);
+const merged2 = 'Merged Thread Title: '.concat(title2, '\n\n', content2);
+
+describe('Test cases revolving around initializating, modifying, and merging discussion forum threads', () => {
+
+    ['instructor'].forEach((user) => {
+        beforeEach(() => {
+            cy.visit('/');
+        });
```
williamjallen, line 18: We should put the login code here so it doesn't clutter every single test case.

williamjallen (changes requested):

site/cypress/e2e/test_forum.spec.js
IDzyre: Done

IDzyre:
I am working to refactor these tests to be more like the tests in profile.spec.js (instead of duplicating code, making functions to call instead)

site/cypress/e2e/test_forum.spec.js
```diff
@@ -0,0 +1,84 @@
...
+
+const removeThread = (title) => {
+    cy.get('.thread-left-cont > .thread-list-item').contains(title).click();
+    cy.get('.first_post > .post-action-container > .delete-post-button').click();
+    cy.get('.thread-left-cont > .thread-list-item').contains(title).should('not.exist');
+};
+
+describe('Test cases revolving around creating, replying to, merging, and removing discussion forum threads', () => {
+
+    ['instructor'].forEach((user) => {
```
williamjallen, line 50: Why do we need to iterate over an array with one element?

IDzyre: I had left this in case I needed to test with student or ta, but with the current testing, I removed the forEach part.

site/cypress/e2e/test_forum.spec.js
```diff
@@ -0,0 +1,82 @@
...
+};
+
+const mergeThreads = (fromThread, toThread, mergedContent) => {
+    // Add more to tests for uploading attachments
+    cy.get('.thread-left-cont > .thread-list-item').contains(fromThread).click({ force: true });
+    cy.get('[title="Merge Thread Into Another Thread"]').click();
+    cy.get('.chosen-single > span').click();
+    cy.wait(500);
+    cy.get('.active-result').contains(toThread).click({ force: true });
+    cy.get('#merge-threads > form > .popup-box > .popup-window > .form-body > .form-buttons > .form-button-container > .btn-primary').click({ force: true });
```
williamjallen, line 38: Is there any better way to target this?  It just seems a bit fragile if someone changes the underlying HTML in the future.

site/cypress/e2e/test_forum.spec.js
```diff
@@ -0,0 +1,82 @@
...
+        cy.get('#nav-sidebar-collapse-sidebar').click();
+    });
+
+    it('Create, reply to, merge, and delete threads', () => {
+        // Comment
+        createThread(title1, content1, 'Comment ');
+        // Question
+        createThread(title2, content2, 'Question ');
+        // Tutorials
+        createThread(title3, content3, 'Tutorials ');
```
williamjallen, line 64: Is the space necessary after the category name?

williamjallen (commented):
This test seems reasonable to me, but I need to see the CI pass before I can approve.  Please ping me or re-request a review when the tests are passing.

site/cypress/e2e/test_forum.spec.js
IDzyre: It is not necessary, I was just copying what the twig file had, but it is now removed to be cleaner.

williamjallen (approved):

PR #9395, opened 2023-06-05
[Feature:System] Add display name order

04vmatsibekker:

User cannot store the format in which they want their name to be displayed (i.e. given family, family given)

### What is the new behavior?
A column is added to the user database that allows this information to be stored.

Closes issue #8446


migration/migrator/migrations/master/20230605230211_add_display_name_order_column.py
```diff
@@ -0,0 +1,178 @@
...
+def up(config, database):
+    """
+    Run up migration.
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    # add column
+    database.execute("ALTER TABLE users ADD COLUMN IF NOT EXISTS display_name_order character varying(255) NOT NULL DEFAULT 'GIVEN_F';")
```
williamjallen, line 13: It would be good to execute this entire migration in a single transaction via a single `database.execute()` call.  That way, any SQL error will result in the entire transaction getting rolled back if the migration fails.

williamjallen (changes requested):

williamjallen (approved):

bmcutler (changes requested):
The trigger function is wrong.  It deletes the recently added pronouns field -- perhaps other things are also wrong.  Please do a line by line comparison of the edits to the trigger function.

Issue #9380, opened 2023-06-01
VCS subdirectory improvements


bmcutler:
2 requests following after PR #9317 VCS Subdirectory Support

* Let's allow the instructor to switch an existing VCS gradeable with no subdirectory specified to having a subdirectory specified & vice versa.  (Currently the instructor can set the subdirectory to '.' which does the same thing as not having a subdirectory, but this is an imperfect workaround.)

* Let's make it an 'error' (or 'warning') if the repository is completely empty (with or without a subdirectory specified), or if the specified subdirectory exists but is empty.   rationale: this is probably student error (not understanding how, or mistakenly not commiting & pushing files to the repo before requesting grading.)

PR #9399, opened 2023-06-07
[Feature:Submission] VCS submission error messages

IDzyre:
closes part of #9380 

Currently if the repository for a VCS gradeable is empty, either no files get checked out, or a failure text file is put in the checkout path. The name of this text file is the only error message that occurs when something goes wrong with submitting files.

### What is the new behavior?
If no files are submitted (with a repository that allows empty commits) it shows 
image
If a 'failure' text file is found., it displays the 'tips' found in the failure text file.
image
image


site/public/css/submitbox.css
```diff
@@ -13,6 +13,15 @@
     margin-left: auto;
 }
 
+#error-message {
+    text-align: center;
+    margin-left: auto;
+}
+
+#error-message > h1 {
+    color: red;
```
williamjallen, line 22: It would be better to use one of our existing colors from `colors.css`.  A good choice might be the same color used for the red warning text shown above the submit box in your screenshot.

site/app/views/submission/HomeworkView.php
```diff
@@ -1030,6 +1030,16 @@ private function renderVersionBox(GradedGradeable $graded_gradeable, $version_in
             }
         }
 
+        $failed_file = '';
+        // See if the grade has succeeded or failed
+        foreach ($param['files'] as $file) {
+            if (str_contains($file['relative_name'], 'failed')) {
```
ziesski, line 1036: Single file containing "failed" will generate amount of error message, but we discussed on the meeting and should not worry too much about it.

ziesski (approved):
Looks good to me! Ready to merge after minor tweak.

PR #9400, opened 2023-06-08
[Bugfix:Autograding] backwards compatible vcs subdirectory

bmcutler:
A VCS gradeable created and built before installing v23.06.00 will not have the partial_path field in the form.json file.  This will cause an error upon upgrading unless the gradeable is rebuilt.

### What is the new behavior?
Backwards compatible logic allows grading of gradeables built before v23.06.00.
New gradeables, and old gradeables built after installation work as well.


IDzyre (approved):
Assuming the failing tests are due to the flakyness of Selenium, this is good.

-- An oversight on my part...

Issue #9402, opened 2023-06-08
Hiding Folder on Same Level as Another Folder with Same Name Start Hides Both


NiaHeermance:
**Describe the bug**
In Course Materials, suppose two folders are on the same level, and the second folder's name is a longer version of the first
folder's name. (e.g. `sub` and `sub2` folders are both on the root level or are both in a folder called, say, `main`. 'sub2' is an
extension of the name of 'sub'.) Then clicking the "Hide from Students" button for one folder also effectively clicks the
"Hide from Students" button of the other folder.

**Expected behavior**
The "Hide from Students" button for a folder should only hide the files inside that folder, not hide the files of a folder on the same level.

**To Reproduce**
1.  Go to Course Materials as an instructor.
2. Add two folders on the same level. with the second folder's name being an extension of the first. (e.g. Add any file with path sub and add another file in path sub2.)
3. Ensure that all files are currently set to _**not**_ "Hide from Students."
4. Click the pencil for one of the folders. Select "Hide from Students."
5. Go to Course Materials as a student.
6. See error

**Configuration**
 - OS: Windows 11.
 - Browser: Firefox.

PR #9403, opened 2023-06-08
[Bugfix:CourseMaterials] Hide button only affects same folder

NiaHeermance:

Currently, if two folders are in the same subfolder (or if these two folders are both on the top level), and the second folder's
name is an extension of the first folder's name (e.g. `apple` and `apple_pear`), the recursive hide button ("Hide from Students")
for the shorter name folder will actually affect the other folder as well.

### What is the new behavior?
Now the "Hide from Students" button will always affect only the current folder.

See issue #9402 for more information.


site/app/controllers/course/CourseMaterialsController.php
```diff
@@ -328,11 +328,17 @@ public function modifyCourseMaterialsFileTimeStamp($newdatatime): JsonResponse {
...
         foreach ($course_materials as $course_material) {
-            if (
-                str_starts_with(pathinfo($course_material->getPath(), PATHINFO_DIRNAME), $main_course_material->getPath())
-                && $course_material->getPath() != $main_course_material->getPath()
-            ) {
+            $course_material_path = $course_material->getPath();
+            $course_material_dir = pathinfo($course_material->getPath(), PATHINFO_DIRNAME);
+
+            $same_start = str_starts_with($course_material_dir, $main_path);
+            $not_same_file = $course_material_path != $main_path;
```
williamjallen, line 338: It's better to use the strict equality operators (`===` and `!==`) in PHP to prevent unexpected type conversions from occurring.

See here for more info: https://www.php.net/manual/en/types.comparisons.php

```suggestion
            $not_same_file = $course_material_path !== $main_path;
```

williamjallen (changes requested):

NiaHeermance:
Changes made! I'm new to PHP so I forgot about that.

Youssef-Chip (changes requested):
Wouldn't be a bit better to switch the order of the if statements at the end: First one :             
if (!$same_start || !$not_same_file || $course_material_path[strlen($main_path)] !== '/') {
            continue;}                                                                                                                                    The other two if/else stay the same after iy

NiaHeermance:
> Wouldn't be a bit better to switch the order of the if statements at the end: First one : if (!$same_start || !$not_same_file || $course_material_path[strlen($main_path)] !== '/') { continue;} The other two if/else stay the same after iy

I would agree if the indenting was many levels deep. However, there's only one indent inside, so I don't think this simplification is necessary.

Issue #9397, opened 2023-06-06
Cookie names should be hardcoded


skara9:
**What problem are you trying to solve with Submitty**
Currently, the Submitty website uses a multitude of cookies with names that store variable data.
For example, in the course materials view, every time a folder is opened/closed a cookie is set with its name as the path to the folder, and its value as whether the folder is open or closed.
(e.g. `%2Fvar%2Flocal%2Fsubmitty%2Fcourses%2Fsemester%2Fcourse%2Fuploads%2Fcourse_materials%2Ffolder=opensd1`)

This is bad practice and leads to poor organization, as the names of cookies are meant to be used as identifiers, not values. Cookies with values stored as names can clutter the namespace and potentially lead to collisions. Furthermore, retrieving data from such cookies becomes unsafe and unidiomatic.

**Describe the way you'd like to solve this problem**
Usage of cookies throughout the site should follow this guideline: cookie names should never be generated, always hardcoded.
This way, the site codebase will be more organized and also safer.

**Additional context**
This is a large issue proposing changes that, if done incorrectly, can break large portions of the site. Hence, it would be beneficial to tackle it with several PRs, each modifying a smaller portion of the site frontend.

PR #9404, opened 2023-06-09
[Refactor:InstructorUI] Course Material Cookies Refactor

skara9:
Partially addresses #9397 

Course material folder open/closed cookies have been refactored to be more organized and use hardcoded cookie names so as to not pollute the namespace.


site/app/templates/course/CourseMaterials.twig
```diff
@@ -181,10 +181,14 @@
...
+        Cookies.set('cm_folder_data', JSON.stringify(folderData));
+    }
+
     function toggleFoldersOpen(){
-        {% for file in material_list %}
-        Cookies.remove("{{ file.getPath }}".substr(0,'{{ file.getPath }}'.lastIndexOf("/")));
-        {% endfor %}
+        Cookies.remove('cm_folder_data');
         Cookies.set("foldersOpen", toggleCMFolders());
     }
```
williamjallen, line 193: It would be a good idea to move these functions to a `.js` file so they can be linted.

site/app/templates/course/CourseMaterials.twig
```diff
@@ -221,46 +225,14 @@
 
     window.onload = function () {
```
williamjallen, line 226: Same comment for the body of this function as well.

williamjallen: Serving "dead" code to lower-access users for the sake of maintainability is fine in my opinion.  Static JS files are often cached by browsers, and are negligible in size compared to media, so there should be no performance hit associated with the extra bytes being transmitted.

site/app/templates/course/CourseMaterials.twig
mahi-pas: I second this

mahi-pas (commented):
The code works correctly in my testing. It would be nice if the js in CourseMaterials.twig were in a different file.

site/app/templates/course/CourseMaterials.twig
```diff
@@ -221,46 +225,14 @@
...
-                        //if the folder should be closed, close it
-                        closeCMFolder(subFolderCookie.substr(6));
-                        //there's no way to open it in that function
-                    }
-                }
-            }{% endfor %}
+        const foldersOpen = Boolean(JSON.parse(Cookies.get('foldersOpen') || 'false'));
+        const folderData = JSON.parse(Cookies.get('cm_folder_data') || '{}');
+        //open/close folders on screen as directed
+        toggleCMFolders(foldersOpen);
```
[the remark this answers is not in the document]
mahi-pas: I see, thank you

mahi-pas (approved):
LGTM

site/app/templates/course/CourseMaterials.twig
williamjallen: That seems reasonable to me.  Please make an issue so we don't forget to do it.

PR #9416, opened 2023-06-13
[Refactor:InstructorUI] Separate course material script

skara9:

Course materials page has a large amount of JavaScript within the HTML document.

### What is the new behavior?
This code has been moved to a separate script file and refactored to run without Twig and pass eslint.


site/public/js/course-materials.js
```diff
@@ -0,0 +1,170 @@
+/* global IS_INSTRUCTOR, toggleCMFolders, toggleCMFolder */
+
+// eslint-disable-next-line no-unused-vars
+function setFolderCookie(folderPath, id, open) {
```
williamjallen, line 4: It would be better to mark these as `exported` at the top of the file, similar to the `global` comment.

site/public/js/course-materials.js
```diff
@@ -0,0 +1,170 @@
...
+    const folderData = JSON.parse(Cookies.get('cm_folder_data') || '{}');
+    folderData[folderPath] = { id, open };
+    Cookies.set('cm_folder_data', JSON.stringify(folderData));
+}
+
+// eslint-disable-next-line no-unused-vars
+function toggleFoldersOpen() {
```
williamjallen, line 11: Same here.

site/public/js/course-materials.js
```diff
@@ -0,0 +1,170 @@
...
+    Cookies.remove('cm_folder_data');
+    Cookies.set('foldersOpen', toggleCMFolders());
+}
+
+function handleHideMaterialsCheckboxChange(clicked) {
+    const warningIds = {'hide-materials-checkbox': 'upload-form', 'hide-materials-checkbox-edit': 'edit-form', 'hide-folder-materials-checkbox-edit': 'edit-folder-form'};
```
williamjallen, line 17: This line is pretty long.  It would be better to break it into multiple lines.

Something like this perhaps:
```suggestion
    const warningIds = {
        'hide-materials-checkbox': 'upload-form',
        'hide-materials-checkbox-edit': 'edit-form',
        'hide-folder-materials-checkbox-edit': 'edit-folder-form'
    };
```

site/public/js/course-materials.js
williamjallen: I'm not sure what you mean.  The `no-unused-vars` error being ignored here is caused by the linter not seeing any usages of this function in this file.  This function is being used in a Twig file and there's no way for eslint to know that.  The solution is to tell eslint that it is used outside this file by marking it as "exported".

mahi-pas (changes requested):
On my device the cookies get stored, but the folders don't open/close as they should

mahi-pas (approved):
Tested functionality for instructor, ta, student, and grader. It all works properly.
Code looks good as well

williamjallen:
This looks good to me, but I think we should wait for the tests to pass before merging since we have a decent number of course materials tests which could potentially catch issues with these changes.

Issue #9154, opened 2023-04-07
Display Pronouns In Discussion Forum


mahi-pas:
**What problem are you trying to solve with Submitty**
Pronouns are only shown to instructors in Manage Students

**Describe the way you'd like to solve this problem**
Make an optional opt-in feature that displays your pronouns beside your name in the discussion forum

User profiles must also store a boolean value that is true when they would like to show their names in discussion forum

sahil-ansari01:
1. Add a Pronoun Field to User Profiles:
To add a field for pronouns to user profiles in Submitty, you will need to modify the code that generates the user profile form. You can add a new input field for pronouns and store this information in the database.

2. Display Pronouns in Discussion Forum Posts:
To display a user's pronouns in their forum posts, you will need to modify the code that generates the user information block that appears next to each post. You can add a new field for pronouns and retrieve this information from the database.

PR #9420, opened 2023-06-14
[Feature:Forum] Display pronouns in forum option

yanliw123:
display pronouns option in profile page and function to display pronoun in forum works good, read for review

### Please check if the PR fulfills #9154

display pronouns option in profile page 
My Profile - Google Chrome 2023_6_14 14_14_36

### What is the new behavior?
once checked the box for display pronouns, you should able to see your pronouns next to your name in discussion forum.


NiaHeermance (changes requested):
I changed Alyssa P. Hacker's pronoun's to She/Her, clicked the button to display on the forum. When I logged in as instructor and viewed [this thread](http://localhost:1511/courses/s23/sample/forum/threads/7), the pronouns are indeed there. Nice!

However, when I went to [the same thread](http://localhost:1511/courses/s23/sample/forum/threads/7) as aphacker (or student), I received the following error:

```

app\exceptions\OutputException (Code: 0) thrown in /usr/local/submitty/site/app/libraries/Output.php (Line 372) by:


Message:
Key "pronouns" does not exist as the array is empty. in /usr/local/submitty/site/app/templates/forum/CreatePost.twig:97

Stack Trace:
#0 /usr/local/submitty/site/app/views/forum/ForumThreadView.php(290): app\libraries\Output->renderTwigTemplate()
#1 unknown file(unknown line): app\views\forum\ForumThreadView->showForumThreads()
#2 /usr/local/submitty/site/app/libraries/Output.php(234): call_user_func_array()
#3 /usr/local/submitty/site/app/libraries/Output.php(203): app\libraries\Output->renderTemplate()
#4 /usr/local/submitty/site/app/controllers/forum/ForumController.php(1059): app\libraries\Output->renderOutput()
#5 unknown file(unknown line): app\controllers\forum\ForumController->showThreads()
#6 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(262): call_user_func_array()
#7 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(240): app\libraries\routers\WebRouter->run()
#8 /usr/local/submitty/site/public/index.php(131): app\libraries\routers\WebRouter::getWebResponse()
```

Could this error be addressed?

NiaHeermance (approved):
Works fully now! I approve.

NiaHeermance (changes requested):
Actually, I just tested rolling back the migration and I get a crash. Re-migrating it does not fix it either. Here is the error:

```

app\exceptions\ConfigException (Code: 0) thrown in /usr/local/submitty/site/app/models/Config.php (Line 562) by:
                throw new ConfigException("Missing config setting '{$section}.{$key}' in configuration json file");


Message:
Missing config setting 'course_details.grade_inquiry_message' in configuration json file

Stack Trace:
#0 /usr/local/submitty/site/app/models/Config.php(507): app\models\Config->setConfigValues()
#1 /usr/local/submitty/site/app/libraries/Core.php(142): app\models\Config->loadCourseJson()
#2 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(278): app\libraries\Core->loadCourseConfig()
#3 /usr/local/submitty/site/app/libraries/routers/WebRouter.php(183): app\libraries\routers\WebRouter->loadCourse()
#4 /usr/local/submitty/site/public/index.php(131): app\libraries\routers\WebRouter::getWebResponse()
```

Could this be fixed?

mahi-pas (changes requested):
image
In my profile there should also be a field "Display Pronouns in Discussion Forum" and editing it would go to the pronouns modal, similar to first and last name opening the same modal

mahi-pas (changes requested):
image
Is there a reason why there is a space between the parenthesis and the pronouns?

mahi-pas (commented):
Functionality and code seem to be good. I don't have the expertise to verify the migrations.

migration/migrator/migrations/course/20230609110652_display_pronouns.py
```diff
@@ -0,0 +1,38 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    sql = """ALTER TABLE users DROP COLUMN IF EXISTS display_pronouns;"""
+    database.execute(sql)
```
williamjallen, line 37: We generally don't want to delete any data in the down migration in case something gets rolled back temporarily and then rolled forward again shortly thereafter.

migration/migrator/migrations/master/20230609110630_display_pronouns.py
```diff
@@ -0,0 +1,177 @@
...
+def up(config, database):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    pass
```
williamjallen, line 13: this line isn't necessary

migration/migrator/migrations/course/20230609110652_display_pronouns.py
```diff
@@ -0,0 +1,38 @@
...
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    sql = """ALTER TABLE users ADD COLUMN IF NOT EXISTS display_pronouns
+      boolean DEFAULT false;"""
+    database.execute(sql)
+    pass
```
williamjallen, line 20: You don't need `pass` if you already have text in a given block of code in Python.

migration/migrator/migrations/course/20230609110652_display_pronouns.py
```diff
@@ -0,0 +1,38 @@
...
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    sql = """ALTER TABLE users DROP COLUMN IF EXISTS display_pronouns;"""
+    database.execute(sql)
+    pass
```
williamjallen, line 38: Same comment here

migration/migrator/migrations/master/20230609110630_display_pronouns.py
```diff
@@ -0,0 +1,177 @@
...
+                        RAISE EXCEPTION 'query_string error in trigger function sync_courses_user() when doing UPDATE';
+                    END IF;
+                    PERFORM dblink_exec(db_conn, query_string);
+                END IF;
+                -- All done.
+                RETURN NULL;
+            END;
+            $$;"""
+    database.execute(sql)
+    pass
```
williamjallen, line 91: same comment here

migration/migrator/migrations/master/20230609110630_display_pronouns.py
```diff
@@ -0,0 +1,177 @@
...
+                    END IF;
+                    PERFORM dblink_exec(db_conn, query_string);
+                END IF;
+                -- All done.
+                RETURN NULL;
+            END;
+            $$;"""
+
+    database.execute(sql)
+    pass
```
williamjallen, line 177: same comment here

williamjallen (changes requested):

NiaHeermance (changes requested):
Migration rollback works now! There's still two things I notice:

- On the My Profile page, when you set the Display Pronouns in Discussion Forum option, the popup does not close automatically. This may be an intentional choice. However, some popups on that page do close automatically, so I want to make sure this was intentional.

- After changing whether you want your pronouns to be displayed, the text says "true" instead of "True" and "false" instead of "False." On a refresh, the capitalization is fixed, but the capitalization should probably be right to begin with.

site/app/templates/EditPronounsForm.twig
```diff
@@ -29,6 +29,14 @@
             </input>
             <button aria-label="Clear pronoun input" type="button" style="background: none;border:none;" onclick="clearPronounsBox();"><i class="fas fa-trash" style="padding:0.5rem;" title="Clear pronouns field"></i></button>
         </div>
+        <br>
```
mahi-pas, line 32: I'm pretty sure we shouldn't use `<br>` but rather use CSS. Either way, the gap is too large so it looks a bit odd
image

The gap should only be a little bit. But honestly, if you think it's unnecessary then we can just do without the gap and leave it to how it was before

mahi-pas (changes requested):
Overall pretty good, I think I'll approve after this change

NiaHeermance:
I now see this on the login page: 
image

NiaHeermance:
I'm experiencing this now when I go to this page: http://localhost:1511/courses/f23/sample.

image

yanliw123:
> I'm experiencing this now when I go to this page: http://localhost:1511/courses/f23/sample.
> 
> image

hmm, that's weird, I just tested the instructor and student account and both works fine without any errors when go to f23 sample course page

NiaHeermance (approved):
Everything works perfect. I approve!

NiaHeermance:
I realized that I should be using submity_install when I first load a branch. I believe that was what was causing the echo on the side. Apologies for not realizing that. I have approved the PR.

site/app/controllers/UserProfileController.php
```diff
@@ -107,14 +107,18 @@ public function changePronouns() {
...
             if ($user->validateUserData('user_pronouns', $newPronouns) === true) {
                 $user->setPronouns($newPronouns);
+                $user->setDisplayPronouns($newDisplayPronouns);
                 $user->setUserUpdated(true);
                 $this->core->getQueries()->updateUser($user);
                 return JsonResponse::getSuccessResponse([
                     'message' => "Pronouns updated successfully",
-                    'pronouns' => $newPronouns
+                    'pronouns' => $newPronouns,
+                    'before' => $newDisplayPronouns,
```
bmcutler, line 120: what is the purpose of this 'before' field vs the 'display_pronouns' field?

site/app/controllers/admin/UsersController.php
```diff
@@ -350,6 +354,8 @@ public function updateUser($type = 'users') {
 
         $user->setPronouns(trim($_POST['user_pronouns']));
 
+        $user->setDisplayPronouns(trim($_POST['display_pronouns']));
```
bmcutler, line 357: I don't think we should be 'trim'ming a bool type value

site/app/views/forum/ForumThreadView.php
```diff
@@ -103,6 +103,8 @@ public function searchResult($threads) {
                 $given_name = trim($user_info["given_name"]);
                 $family_name = trim($user_info["family_name"]);
                 $visible_username = $given_name . " " . substr($family_name, 0, 1) . ".";
+                $pronouns = trim($user_info["pronouns"]);
+                $display_pronouns = trim($user_info["display_pronouns"]);
```
bmcutler, line 107: why trim this?

site/app/views/forum/ForumThreadView.php
```diff
@@ -887,10 +889,12 @@ public function displayThreadList($threads, $filtering, &$activeThreadAnnounceme
 
             if ($is_full_page) {
                 $user_info = $this->core->getQueries()->getDisplayUserInfoFromUserId($first_post["author_user_id"]);
-                $email = trim($user_info['user_email']);
+                $email = trim($user_info["user_email"]);
                 $given_name = trim($user_info["given_name"]);
                 $family_name = trim($user_info["family_name"]);
                 $visible_username = $given_name . " " . substr($family_name, 0, 1) . ".";
+                $pronouns = trim($user_info["pronouns"]);
+                $display_pronouns = trim($user_info["display_pronouns"]);
```
bmcutler, line 897: why trim this?

site/app/views/forum/ForumThreadView.php
```diff
@@ -997,6 +1003,8 @@ public function createPost($thread_id, $post, $unviewed_posts, $first, $reply_le
         $given_name = trim($user_info["given_name"]);
         $family_name = trim($user_info["family_name"]);
         $visible_username = $given_name . " " . substr($family_name, 0, 1) . ".";
+        $pronouns = trim($user_info["pronouns"]);
+        $display_pronouns = trim($user_info["display_pronouns"]);
```
bmcutler, line 1007: why trim this

site/app/views/forum/ForumThreadView.php
```diff
@@ -1114,6 +1126,24 @@ public function createPost($thread_id, $post, $unviewed_posts, $first, $reply_le
             }
         }
 
+        if ($this->core->getUser()->getGroup() == 4) {
+            $info_name = $given_name . " " . $family_name . " (" . $post['author_user_id'] . ")";
+            $visible_user_json = json_encode($visible_username);
+            $pronouns = trim($user_info["pronouns"]);
+            $display_pronouns = trim($user_info["display_pronouns"]);
```
bmcutler, line 1133: why trim this?

bmcutler (changes requested):
asking questions, not sure if they are required changes

site/app/controllers/UserProfileController.php
yanliw123: that was for error-checking purpose, I will remove the before field and only leave the 'pronouns' and 'display_pronouns’ fields for future use.

bmcutler: please remove the before field ASAP.  thanks

bmcutler (changes requested):
Unfortunately we have a bug with anonymous posts in the forum.
The pronouns are still displayed:

Screenshot 2023-07-26 at 3 02 20 PM

mahi-pas (changes requested):
If you set pronouns in discussion forum to true
image
image
And then reload the page, it reverts to false.
image

 It appears the value works because it is shown in the discussion forum, but the display is wrong

mahi-pas (changes requested):
image
When i make a thread on instructor and look at the student page to see if it's synced, the name and pronouns are not visible. if I refresh the page they are

NiaHeermance (commented):
Why is display_pronouns added both to the course users table and the master user table? Shouldn't it only be added to the master user table?

bmcutler:
The original design of the databases is that the course database is sufficient to render any page within the course.
One advantage of this is that an instructor can request a dump of the complete course database for their course and in theory import it to their own local instance of Submitty for archiving, etc.

The master database was created so that we knew which courses a specific student had access to.   (So we didn't have to loop over all courses on the system to check).

Much of the data about each user is therefore duplicated between the master and course databases.  That is a known redundancy/duplication (a slight but accepted disadvantage).  The trigger functions are supposed to ensure the duplication is in sync across all courses.

NiaHeermance:
I see! Thanks for explaining that to me.

bmcutler:
Also, in the original design, in theory, we should be able to destroy and then recreate the master database just with the info in each course database.  (Not sure if this is still true, and I've never attempted to do this.)

bmcutler (approved):
thanks for updating this PR :)

PR #9423, opened 2023-06-15
[Feature:InstructorUI] Add/Remove Subdirectory after created

IDzyre:
Subdirectories were added in #9317, but were defaulted to not being able to be added/removed after the gradeable has been created. 

Fixes part 1 of #9380

### What is the new behavior?
Instructors can now specify if they want to use/not use a subdirectory _after_ a gradeable has been created. 

image


ziesski (approved):
All functionality seems to be working as expected.

PR #9424, opened 2023-06-15
[Refactor:API] Change config & course semester to term

NiaHeermance:
Currently, Submitty uses a mix of `term` and `semester` throughout the code. However, we want to change `semester` to `term` everywhere to be consistent as well as be more general for schools that setup Submitty and do not have a semester system. (Some schools have trimesters or quarters.) 

### What is the new behavior?
Three changes are made towards this goal of removing the word `semester`:

- `.performance_warning_ignore.json` instead of `<semester>` now uses `<term>`.

- `config->getSemester()` is now called `config->getTerm()`.

- `course->getSemester()` and `course->getLongSemester()` are now `course->getTerm()` and `course->getLongTerm()` respectively.

### Note:
EmailRepository's `getSemester()` will be changed in a future PR as it requires a migration.


04vmatsibekker (commented):
image

NiaHeermance:
> image
> 
> These are three places in the code where getSemester() is still being used

Actually, these are usuing EmailEntity's getSemester(), which I have not changed yet. Changing that one requires a migration, so it's coming in a separate PR.

04vmatsibekker (commented):
I tested out adding comments in the calendar [semester and course are important for making announcements], I switched from course to course and used some of the features. I even made a new course and played with the course settings. Nothing broke.

04vmatsibekker (approved):

pbailie (commented):
I do not anticipate conflicts with the sysadmin Autofeed or Accounts scripts with this PR.

bmcutler (approved):
code looks ok.

Issue #9426, opened 2023-06-15
Mark as resolved does not work unless on thread


NiaHeermance:
**Describe the bug**
When a Discussion Forum thread is created, you can set it as `Unresolved`. An instructor then needs to click `Mark as resolved` to give it a green check mark rather than a question mark. You can click `Mark as resolved` both on the [forum main page itself](http://localhost:1511/courses/s23/sample/forum) and a thread's specific page. However, only the `Mark as Resolved` button on the thread itself currently is working and actually changes the question mark to a check mark.

**Expected behavior**
The green check mark will appear when you click `Mark as resolved` and the page will refresh.

**To Reproduce**
1. Login as any user and create a Discussion Forum thread. Make sure it is set to `Unresolved`.
2. As instructor, click that course and on the sidebar click "Discussion Forum."
3. Locate the thread in the vertical list, and click `Mark as resolved`.
4. Refresh and notice the thread is still unresolved.

**Configuration**
 - OS: Windows 11
 - Browser: Firefox

**Screenshots**

![Mark as Resolved Example - Bad Button](https://github.com/Submitty/Submitty/assets/51209504/981ec6fc-71c3-433e-b602-bb4345474545)

![Mark as Resolved Example - Good Button](https://github.com/Submitty/Submitty/assets/51209504/ea0e6fba-353c-4cd3-ad6f-4bff36a754a2)

PR #9428, opened 2023-06-15
[Bugfix:Forum] Remove threadlist resolve button

NiaHeermance:
A broken `Mark as resolved` button appears on unresolved threads in the discussion forum's main page (Thread list).

### What is the new behavior?
The broken button is gone, rather than fixed as decided by myself and @bmcutler.

See issue #9426.


04vmatsibekker (approved):
Simple change and fixes the issue. Great job!

PR #9433, opened 2023-06-16
[Testing:Developer] Adding pronouns tests to cypress

### LINKED ISSUE #9418: Tests for Pronouns
In Spring 2023, we added pronoun data to user profiles. This data still does not have automated tests

Create tests using cypress/ unit tests to test the functionality of pronouns and ensure that it is all working properly

Example Tests:
* Testing changing pronouns in user profile and reloading the page to ensure it changed
* Testing a student changing their pronouns and then logging into TA and making sure it updated in student photos (then changing it back)
* Testing a student changing their pronouns and then logging into instructor and making sure it updated in manage students (then changing it back)
* etc.

Come up with your own as well if you think more are needed! (or merge these into one)

*Note: if pronouns in discussion forum is in main, then also make tests for this* (it is currently PR #9420, if it is merged then this is necessary)

mahi-pas:

Partially Fixes #9418
There were no cypress tests for pronouns

### What is the new behavior?
Adds cypress tests to pronouns

New test does the following:

1. Logs in as student and changes pronouns
2. Logs in as instructor and sees if the change is reflected in manage students
3. Logs in as instructor and sees if the change is reflected in student photos
4. Logs in as student and reverts pronouns to original value

### Other information?
Next PR regarding pronouns testing will be about PHP unit tests


site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+        //save old pronouns
+        e.then(($pronounsInput) => {
+            oldPronouns = $pronounsInput.val();
+        });
+
+        //type in new pronouns
+        cy.get('button[aria-label="Clear pronoun input"]').click(); //clear input using trash can
+        e.type('They/Them');
+        cy.get('#edit-pronouns-form .form-button-container > .btn-primary').first().click();
```
williamjallen, line 22: Can we target this button in a more specific way?  It's very possible that someone will add another primary button to this form and thus break the test.  It would be better to give the button an ID, or otherwise target it in a more unique way.

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+    });
+
+    it('Verifies changed pronouns as instructor in Manage Students', () => {
+        cy.visit(['sample','users']);
+        cy.login('instructor');
+
+        //ensure pronouns column is on
+        cy.get('a[href="javascript:toggleColumnsForm()"]').click(); //open toggle columns form
```
williamjallen, line 31: It would be better to target this by something other than the name of a JS function.  We don't want to have to change this test if someone refactors the JS and changes the name of the function.

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+        //Select text from photo area and parse to get pronoun
+        e.invoke('text')
+            .then(text => text.split('\n'))                 // split by newline
+            .then(texts => texts.map(text => text.trim()))  // trim whitespace
+            .then(texts => texts.filter(text => text))      // remove empty
+            .then(texts => {
+                //texts stores split text, should be [Joe Student, student, They/Them]
+                expect(texts[2]).to.equal('They/Them');
+            });
```
williamjallen, line 56: Would it not be roughly equivalent to just use `e.contains('They/Them')`?

site/cypress/e2e/pronouns_change.spec.js
mahi-pas: done

site/app/templates/EditPronounsForm.twig
```diff
@@ -38,5 +38,5 @@
 {% endblock %}
 {% block buttons %}
     {{ block('close_button') }}
-    <input class="btn btn-primary" type="submit" value="Submit" />
+    <input class="btn btn-primary" type="submit" value="Submit" id="submit" />
```
williamjallen, line 41: An ID of "submit" is overly general and is likely to conflict with other fields elsewhere on the page.  Please use a more specific identifier.

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,68 @@
+
+describe('Tests cases abut changing user pronouns', () => {
+
+    let oldPronouns = '';
```
williamjallen, line 4: Why is this variable defined here?  If you're trying to handle data unexpected existing before the test suite runs, you might want to take a look at `beforeEach()` and `afterEach()` to handle setting the expected data and then restoring the old data when the tests finish.

site/app/templates/admin/users/ToggleManageStudentsColumns.twig
```diff
@@ -64,6 +64,6 @@
 
 {% block buttons %}
     {{ block('close_button') }}
-    <input name="submit" class="btn btn-primary key_to_click" type="submit" value="Submit" onclick="updateManageStudentsColumns()"/>
+    <input name="submit" class="btn btn-primary key_to_click" type="submit" value="Submit" id="submit" onclick="updateManageStudentsColumns()"/>
```
williamjallen, line 67: Same comment here.  This identifier needs to be more specific.

williamjallen (changes requested):

site/app/templates/EditPronounsForm.twig
04vmatsibekker: Yeah I agree, especially since even in the cypress tests you made you are using two different submit buttons. Perhaps you could rename this one submit-pronoun-changes instead or something and the other could be submit-manage-student-columns.

mahi-pas: done

site/app/templates/admin/users/ToggleManageStudentsColumns.twig
mahi-pas: done

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+        //save old pronouns
+        e.then(($pronounsInput) => {
+            oldPronouns = $pronounsInput.val();
+        });
+
+        //type in new pronouns
+        cy.get('button[aria-label="Clear pronoun input"]').click(); //clear input using trash can
+        e.type('They/Them');
+        cy.get('#submit').click();
```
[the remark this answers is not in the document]
mahi-pas: I tried a few things but it seems infeasible. Cypress doesn't have a pre-written function for datalists. Also, datalists should be working as it's implemented by the browser so there's no need to test

site/cypress/e2e/pronouns_change.spec.js
mahi-pas: 1) This isn't to handle unexpected data, it's to change the data back
2) These are both changes and tests, if I wrapped it in a before/after call it wouldn't display as a test in cypress. My understanding is before each is usually just for setting the stage for each test, but this requires logging into student and getting the data on a page.
3) If I made it beforeEach/afterEach it would run twice which would slow down the test

williamjallen: Would using `before()` and `after()` address point 3?

04vmatsibekker (approved):

site/cypress/e2e/pronouns_change.spec.js
mahi-pas: If I did it as `before()` or `after()`, it wouldn't show up as a test, which would be bad as it does test if it changed on the profile page

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+    let oldPronouns = '';
+
+    it('Tests changing pronouns for student', () => {
+        cy.visit('/user_profile');
+        cy.login('student');
+
+        //open pronouns form
+        cy.get('#pronouns_val').click();
+        const e = cy.get('#user-pronouns-change');
```
IDzyre, line 12:
```suggestion
        cy.get('#user-pronouns-change').as('e');
```
[Assigning Return Values](https://docs.cypress.io/guides/references/best-practices#Assigning-Return-Values)

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+        //save old pronouns
+        e.then(($pronounsInput) => {
```
IDzyre, line 15:
```suggestion
        cy.get('@e').then(($pronounsInput) => {
```

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,76 @@
...
+            oldPronouns = $pronounsInput.val();
+        });
+
+        //type in new pronouns
+        cy.get('button[aria-label="Clear pronoun input"]').click(); //clear input using trash can
+        e.type('They/Them');
```
IDzyre, line 21:
```suggestion
        cy.get('@e').type('They/Them');
```

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+    });
+
+    it('Verifies changed pronouns as instructor in Student Photos', () => {
+
+        cy.visit(['sample','student_photos']);
+        cy.login('instructor');
+
+        //select Joe Student's photo
+        const e = cy.get('.student-image-container > .name').first();
```
IDzyre, line 49: Why does this line need to be split with the .contains?

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    it('Changes pronouns back', () => {
+        cy.visit('/user_profile');
+        cy.login('student');
+
+        //change back to old pronouns
+        cy.get('#pronouns_val').click();
+        const e = cy.get('#user-pronouns-change');
+        cy.get('button[aria-label="Clear pronoun input"]').click(); //clear input using trash can
+        if (oldPronouns !== '') {
+            e.type(oldPronouns);
```
IDzyre, line 65: Same question, about splitting get and .type

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,76 @@
+
+describe('Tests cases abut changing user pronouns', () => {
+
+    let oldPronouns = '';
+
+    it('Tests changing pronouns for student', () => {
```
IDzyre, line 6: Even though it seems good to have multiple points being tested, since these tests are using the same values, and the tests 'rely' on each other passing, it would be better to combine them into one bigger test. This would be as easy as just removing the separate 'it' statements, and putting the description as a comment, and explicitly logging out of each user before logging in as a different one.

IDzyre (changes requested):

site/cypress/e2e/pronouns_change.spec.js
mahi-pas: Is cy.get('@e') preferrable in any way? otherwise it may make the code less readable

IDzyre:
Content of the tests is good, just a couple of small(ish) changes requested.

site/cypress/e2e/pronouns_change.spec.js
IDzyre: [Assigning Return Values](https://docs.cypress.io/guides/references/best-practices#Assigning-Return-Values)

It is just the best practice for cypress. In theory (and best practice) const, let, and var shouldn't really ever be used.

mahi-pas: I feel like it's bad since it's hard to read for someone unfamiliar with cypress, but since it's declared as the way to do it I will

site/cypress/e2e/pronouns_change.spec.js
mahi-pas: It logs in for each test already. `cy.login()` automatically logs out and then logs in again (or at least, it does in these test cases). 

Combining the test would have the issue of making it harder to know what failed during the test

site/cypress/e2e/pronouns_change.spec.js
mahi-pas: Fixed according to these suggestions

bmcutler:
@mahi-pas 
This PR is failing js-lint, please update :)

site/cypress/e2e/pronouns_change.spec.js
mahi-pas: Should be addressed in latest commit. Let me know if you have any questions/concerns

williamjallen (approved):
These tests are relatively weak and it's easy to find cases where these tests might not catch bugs, but I think they're good enough for now given how simple this feature is.

IDzyre (approved):

mahi-pas:
> These tests are relatively weak and it's easy to find cases where these tests might not catch bugs, but I think they're good enough for now given how simple this feature is.

Which cases?

site/cypress/e2e/pronouns_change.spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+    it('Verifies changed pronouns as instructor in Student Photos', () => {
+
+        cy.visit(['sample','student_photos']);
+        cy.login('instructor');
+
+        //Select text from photo area and parse to get pronoun
+        cy.get('.student-image-container > .name').first().contains('They/Them');
+
+    });
```
williamjallen, line 73: You're not actually verifying that anything changed here--just that the text reads "They/Them".  If the text is "They/Them" at the start of the test and the change functionality is broken, this test case won't catch anything.  The same general idea applies to the other test case as well.  To verify that something actually changed, you need to set it to two separate values.

williamjallen (commented):
> > These tests are relatively weak and it's easy to find cases where these tests might not catch bugs, but I think they're good enough for now given how simple this feature is.
> 
> Which cases?

PR #9444, opened 2023-06-19
[Feature:CourseMaterials] move files, different display_name views

04vmatsibekker:

Instructors can upload file, but once they do so, there is no way to change file name/file location
### What is the new behavior?
Now instructors can change the file name/location, and display a different file name (link_title, for now) to non-owners of file.

New UI for both file/URL changes
Edit

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -47,3 +47,14 @@ updates:
       - dependencies
     commit-message:
       prefix: "[DevDependency] "
+  - package-ecosystem: "github-actions"
+    directory: "/actions/e2e-Setup-Composite"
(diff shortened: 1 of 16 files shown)


NiaHeermance (changes requested):
I uploaded a file (let's call it `test.txt`) with path `cool_place_1/`. Then I tried to move it to path `cool_path_2/`. When I tried to redownload the file, I got this message:

image

I repeated this experiment multiple times and got the same result.

Also, if you try to change a file's path to a location that doesn't exist yet, an error occurs rather than the path being created (which I think the latter is what one would want).

PR #9445, opened 2023-06-19
[Refactor:Autograding] Dockerize existing autograding examples

saumyaborwankar:
### What is the new behavior?

Many/most autograding examples in more_autograding_examples have been updated to run inside of autograding docker containers.

Most of these updates have been tested both on a modern ARM-based Mac and an older-mode intel x86 Mac.

More manual testing is needed, and some of these autograding examples need more edits to fix them.
We also need to create cypress end-to-end tests of autograding -- the existing integration tests of autograding do not test the resource limits and docker containers.


micpap25:
@bmcutler drawing attention to this; should we continue doing this? and should I confirm this is done for my Qiskit examples?

bmcutler:
@micpap25 
Yes, this PR is still a priority to complete (the names of some of the docker images may need to be updated with the recent/ongoing overhaul of the dockerimages repo)

micpap25:
I believe there is some issue in `container_network.py` (evidenced by errors in the Autograding Status page) that are causing the non-`submitty/autograding-default.py` containers to fail. Or maybe something is wrong with the `latest` containers; I will test the latter momentarily.

micpap25:
Re: the above comment; there is an issue with the `clang:10` (latest) and `python:3.7` (latest) containers. I believe they likely run some command that is not allowed somehow based on the error it is giving, though I can't confirm. This issue will need to be on hold until the images are fixed.

micpap25:
@cjreed121 getting you in-the-loop on this as discussed with William yesterday

bmcutler (approved):
Updating the examples still needs more work, but I'm going to merge this PR now.

PR #9448, opened 2023-06-19
[Bugfix:Submission] Fix error message on do not grade

IDzyre:
#9399 was merged, and if the user selects 'do not grade this assignment', a site error would show with 'count must be used on an array'
image

### What is the new behavior?
This is now fixed


Youssef-Chip (approved):
The new changes fix the bug correctly. It should be good to merge

PR #9451, opened 2023-06-20
[UI/UX:TAGrading] Update sort_type for "Total" field

musaabimran:

What problem are you trying to solve with Submitty
Currently in the grading details page you can sort by user id, first name, or last name. However, when grading certain things it would be useful to grade by how many point the person got on the assignment. Currently to do this you have to go through the list and look for the scores manually.

### What is the new behavior?
Now the sorting can be done based on how many points the person got on the assignment.

### DIFF
--- a/site/app/views/grading/ElectronicGraderView.php
+++ b/site/app/views/grading/ElectronicGraderView.php
@@ -493,7 +493,7 @@ HTML;
                 $columns[]     = ["width" => "10%", "title" => "Grading",          "function" => "grading_blind"];
             }
             if ($peer === false) {
-                $columns[]     = ["width" => "15%", "title" => "Total",            "function" => "total"];
+                $columns[]     = ["width" => "15%", "title" => "Total",            "function" => "total",  "sort_type" => "first"];
             }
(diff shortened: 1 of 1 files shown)


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -493,7 +493,7 @@ public function detailsPage(Gradeable $gradeable, $graded_gradeables, $teamless_
                 $columns[]     = ["width" => "10%", "title" => "Grading",          "function" => "grading_blind"];
             }
             if ($peer === false) {
-                $columns[]     = ["width" => "15%", "title" => "Total",            "function" => "total"];
+                $columns[]     = ["width" => "15%", "title" => "Total",            "function" => "total",  "sort_type" => "first"];
```
jenaybarela, line 496: Adding sort type of first will only sort by the first name column. This will not sort by auto grading scores.

jenaybarela (changes requested):
More is needed to get the column to actually sort by score.

jenaybarela (changes requested):
A new sort type needs to be added. The current sort types only support first, last, and user. Adding one of those sort types will not sort by score.

PR #9452, opened 2023-06-20
[Feature:System] Migrate trigger functions separately

skara9:

Database migrations that update trigger functions currently need to copy the entire function into the sql command and update what is needed. This makes development challenging, especially for dealing with merge conflicts between pull requests that update the same trigger functions.

### What is the new behavior?
The database trigger functions have been moved into separate files so they can be more easily updated and tracked by git.


cjreed121 (commented):
This is great! It appears to only do the triggers in the master database. There is also triggers in the course database. Could we also extract them out to sql files?

migration/migrator/main.py
```diff
@@ -470,3 +473,45 @@ def dump(args):
         semester = f"{'s' if today.month < 7 else 'f'}{str(today.year)[-2:]}"
         dump_database(f'submitty_{semester}_sample', data_dir / 'course_tables.sql')
         print('DONE')
+
+
+def load_triggers(args, silent=False):
+    print('Loading trigger functions to master...', end='')
+
+    if 'master' not in args.environments:
```
MasterOdin, line 481: What about triggers for the course DBs? Will they be added later?

MasterOdin (commented):
How often are triggers being updated?

The expectation is that someone would run `migrate` and `load_triggers` always to install Submitty?

williamjallen:
> How often are triggers being updated?

We currently have multiple PRs open which modify trigger functions including #9420 and #9395.

Multiple PRs also modified trigger functions in the spring.  See the blame around [this line](https://github.com/Submitty/Submitty/blame/e8c8f4fa010e06064a490da24363fc190bc2e5da/migration/migrator/data/submitty_db.sql#L102) for reference.

migration/migrator/main.py
```diff
@@ -281,6 +281,9 @@ def handle_migration(args):
         if missing_migration.exists():
             missing_migration.unlink()
 
+    if 'master' in args.environments:
```
MasterOdin, line 284: Given that `load_triggers` has its own check on `args.environments`, I think you can get rid of this if statement altogether.

migration/migrator/main.py
```diff
@@ -470,3 +473,66 @@ def dump(args):
...
+def load_triggers(args, silent=False):
+    for environment in args.environments:
+        if environment != 'master' and environment != 'course':
+            continue
+
+        trigger_dir = get_triggers_path() / environment
+        if not trigger_dir.is_dir():
+            if silent:
+                continue
+            raise SystemExit('Error: Could not find triggers directory for ' + environment)
```
MasterOdin, line 487: This error condition should only happen if someone were to delete the `course` or `master` directory within the repo. I don't think this is an error case that would happen in production unless someone was purposely trying to break their install and if it happens in development, fine to have a noisy error raised from when do `trigger_dir.iterdir()`.

As such, I'd remove this as it adds unnecessary complexity.

migration/migrator/main.py
```diff
@@ -470,3 +473,66 @@ def dump(args):
         semester = f"{'s' if today.month < 7 else 'f'}{str(today.year)[-2:]}"
         dump_database(f'submitty_{semester}_sample', data_dir / 'course_tables.sql')
         print('DONE')
+
+
+def load_triggers(args, silent=False):
+    for environment in args.environments:
+        if environment != 'master' and environment != 'course':
```
MasterOdin, line 480:
```suggestion
        if environment not in ['master', 'course']:
```

MasterOdin (commented):
To remove a trigger, the expectation is that someone would need to:
1. delete the trigger `.sql` file
2. write a migration that drops that trigger

?

migration/migrator/triggers/master/sync_user.sql
```diff
@@ -0,0 +1,41 @@
...
+                db_conn := format('dbname=submitty_%s_%s', course_row.semester, course_row.course);
+                query_string := 'UPDATE users SET user_numeric_id=' || quote_nullable(NEW.user_numeric_id) || ', user_pronouns=' || quote_literal(NEW.user_pronouns) || ', user_givenname=' || quote_literal(NEW.user_givenname) || ', user_preferred_givenname=' || quote_nullable(NEW.user_preferred_givenname) || ', user_familyname=' || quote_literal(NEW.user_familyname) || ', user_preferred_familyname=' || quote_nullable(NEW.user_preferred_familyname) || ', user_last_initial_format=' || quote_literal(NEW.user_last_initial_format) || ', user_email=' || quote_literal(NEW.user_email) || ', user_email_secondary=' || quote_literal(NEW.user_email_secondary) || ',user_email_secondary_notify=' || quote_literal(NEW.user_email_secondary_notify) || ', time_zone=' || quote_literal(NEW.time_zone)  || ', display_image_state=' || quote_literal(NEW.display_image_state)  || ', user_updated=' || quote_literal(NEW.user_updated) || ', instructor_updated=' || quote_literal(NEW.instructor_updated) || ' WHERE user_id=' || quote_literal(NEW.user_id);
```
williamjallen, line 30: This line is pretty long, and the diff for any future changes will be difficult to read.  It would be better to break it into multiple lines for each individual column.

williamjallen (commented):
A couple quick code style changes, and then I think this is ready to be merged.

migration/migrator/main.py
```diff
@@ -470,3 +472,61 @@ def dump(args):
...
+    for environment in args.environments:
+        if environment not in ('master', 'course'):
+            continue
+
+        trigger_dir = get_triggers_path() / environment
+        sql = [(f, f.read_text()) for f in trigger_dir.iterdir()
+               if f.is_file() and f.suffix == '.sql']
+
+        if len(sql) == 0:
+            print('Loading trigger functions to {}...DONE'.format(environment))
```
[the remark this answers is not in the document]
bmcutler: can we reduce this print statement to be 1 line per course?
We have hundreds of  courses and usually these trigger updates will be zero change.


the initial migration output per course is compact when there are no migrations, but gets lengthy when one or more migrations run per course -- this is good! I notice that we are installing some migrations and I am reminded to look for potential issues that arise related to those migrations.

Screen Shot 2023-07-20 at 3 28 09 PM

but this new trigger output is bulky and uninteresting (it always prints even when nothing changed)
Screen Shot 2023-07-20 at 3 29 33 PM

cjreed121 (approved):
Works well. I tested locally and changed a trigger which I saw the new behavior occur. I did not do a Python code review.

migration/migrator/main.py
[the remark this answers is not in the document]
williamjallen: Why do we even need the info message at all?  Perhaps it should just print a message when an error occurs.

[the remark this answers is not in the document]
bmcutler: yeah, let's print nothing if there is no error on the trigger function install.

migration/migrator/triggers/master/sync_courses_user.sql
```diff
@@ -0,0 +1,79 @@
...
+                query_string text;
+            BEGIN
+                db_conn := format('dbname=submitty_%s_%s', NEW.semester, NEW.course);
+
+                IF (TG_OP = 'INSERT') THEN
+                    -- FULL data sync on INSERT of a new user record.
+                    SELECT * INTO user_row FROM users WHERE user_id=NEW.user_id;
+                    query_string := 'INSERT INTO users (
+                        user_id,
+                        user_pronouns,
```
bmcutler, line 20: for order consistency pronouns & numeric_id should be swapped
(they are swapped in the sync_user.sql trigger function)

migration/migrator/triggers/master/sync_courses_user.sql
```diff
@@ -0,0 +1,79 @@
...
+                        user_numeric_id,
+                        user_givenname,
+                        user_preferred_givenname,
+                        user_familyname,
+                        user_preferred_familyname,
+                        user_last_initial_format,
+                        user_email,
+                        user_email_secondary,
+                        user_email_secondary_notify,
+                        user_updated,
```
bmcutler, line 30: we are missing the time_zone and display_image_state fields -- they should be added
(they are in the sync_user.sql trigger file)

bmcutler (changes requested):

bmcutler (approved):

bmcutler (changes requested):
remove print statements

bmcutler (approved):

PR #9453, opened 2023-06-20
[Bugfix:System] validity constraint semester & course

04vmatsibekker:

Beforehand, the course and semester could have trailing and leading spaces, but that should not be allowed. 

### What is the new behavior?

Migration adds a constraint for spaces around course and semester names. Also when an instructor creates a new course, the semester and course names automatically get trimmed.


migration/migrator/migrations/master/202306205342134_add_course_and_semester_constraint.py
```diff
@@ -0,0 +1,8 @@
+def up(config, database):
+    database.execute("ALTER TABLE courses ADD CONSTRAINT no_surrounding_spaces_course CHECK (course ~ '^[^ ].*[^ ]$')")
+    database.execute("ALTER TABLE terms ADD CONSTRAINT no_surrounding_spaces_semester CHECK (term_id ~ '^[^ ].*[^ ]$')")
+
+def down(config, database):
+    database.execute("ALTER TABLE courses DROP CONSTRAINT no_surrounding_spaces_course")
+    database.execute("ALTER TABLE terms DROP CONSTRAINT no_surrounding_spaces_semester")
```
williamjallen, line 7: `up()` and `down()` should each contain a single `database.execute()` call to make sure the entire migration gets rolled back as a single transaction if something goes wrong.

williamjallen (changes requested):
The PR body needs to be filled out so it is clear why this PR is necessary and what changes were made.

migration/migrator/migrations/master/202306205342134_add_course_and_semester_constraint.py
```diff
@@ -0,0 +1,7 @@
+def up(config, database):
+    sql = "ALTER TABLE courses ADD CONSTRAINT no_surrounding_spaces_course CHECK (course ~ '^[^ ].*[^ ]$'); ALTER TABLE terms ADD CONSTRAINT no_surrounding_spaces_semester CHECK (term_id ~ '^[^ ].*[^ ]$');"
+    database.execute(sql)
+
+def down(config, database):
+    sql = "ALTER TABLE courses DROP CONSTRAINT no_surrounding_spaces_course; ALTER TABLE terms DROP CONSTRAINT no_surrounding_spaces_semester;"
+    database.execute(sql)
```
williamjallen, line 7: These lines are pretty long, so it would probably be a good idea to break them into multiple lines (although still as part of a single string).

bmcutler:
If we're adding a constraint, let's make it a more useful constraint, not limited only to spaces on the calendar page.
(I think this constraint is not even strong enough for that use -- spaces in the middle could also break things).

Semester & should be restricted to letters, numbers, underscore & dash like the owner & group.
Also, please name the constraint similarly to the owner & group constraints.

bmcutler (changes requested):
(see comment above)

bmcutler (approved):

PR #9456, opened 2023-06-21
[Bugfix:System] update migration for legacy courses

bmcutler:

Unfortunately the migration in PR #9398 assumed that a constraint exists, however, this constraint was added early in Fall 2019 without an appropriate migration, so courses created during and before Fall 2019 are missing this constraint.

### What is the new behavior?

The migration has been modified to add the constraint before running the rename operation.

### DIFF
--- a/migration/migrator/migrations/course/20230601123230_change_regrade_to_grade_inquiry.py
+++ b/migration/migrator/migrations/course/20230601123230_change_regrade_to_grade_inquiry.py
@@ -28,6 +28,18 @@ def up(config, database, semester, course):
     :param course: Code of course being migrated
     :type course: str
     """
+
+    # First we need to patch missing constraints from existing courses prior to Fall 2019
+    database.execute("""
+
+        alter table electronic_gradeable drop constraint if exists eg_regrade_allowed_true;
+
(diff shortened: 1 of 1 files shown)


migration/migrator/migrations/course/20230601123230_change_regrade_to_grade_inquiry.py
```diff
@@ -28,6 +28,18 @@ def up(config, database, semester, course):
...
+
+        alter table electronic_gradeable drop constraint if exists eg_regrade_allowed_true;
+
+        alter table electronic_gradeable add constraint eg_regrade_allowed_true CHECK
+            (((eg_regrade_allowed IS TRUE) OR (eg_regrade_allowed IS FALSE)));
+
+    """)
+
+
     database.execute("""
```
williamjallen, line 43: It would be better to combine everything into a single `database.execute()` call so all of the commands in this migration get run as a single transaction, and get rolled back together if something goes wrong in any of the steps.

bmcutler: Yes, I agree in general.  
As this is a patch to inconsistent/legacy courses due to migration flaws in the past, I'm just going to leave this as is.

PR #9457, opened 2023-06-21
[Bugfix:Forum] allow user to only write 0 and post

yanliw123:
change empty() to strlen() ===0, so that the user can only write 0 and post

User can not only write 0 and post, error will pop up
### What is the new behavior?
User can only write 0 and post.

### DIFF
--- a/site/app/controllers/forum/ForumController.php
+++ b/site/app/controllers/forum/ForumController.php
@@ -345,7 +345,7 @@ class ForumController extends AbstractController {
         foreach ($_POST["cat"] as $category_id) {
             $categories_ids[] = (int) $category_id;
         }
-        if (empty($thread_title) || empty($thread_post_content)) {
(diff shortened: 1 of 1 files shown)


site/app/controllers/forum/ForumController.php
```diff
@@ -345,7 +345,7 @@ public function publishThread() {
         foreach ($_POST["cat"] as $category_id) {
             $categories_ids[] = (int) $category_id;
         }
-        if (empty($thread_title) || empty($thread_post_content)) {
+        if (strlen($thread_title) === 0 || strlen($thread_post_content) === 0) {
```
04vmatsibekker, line 348: This change only fixes the 0 issue when someone is making a new thread. If someone is replying to a thread by making a post, then the 0 issue still remains 

image

PR #9468, opened 2023-06-23
[Bugfix:TAGrading] Fix Grade Inquiry Components

### LINKED ISSUE #9467: Crash with multiple grade inquiries on different components from same student
**Describe the bug**

* For an electronic/rubric gradeable with manual TA grading and multiple components
* Enable grade inquiries, enable grade inquiries per component.
* Have a student submit 2 grade inquiries from 2 different components of the rubric -- you can do this either from the student UI, or post to the grade inquiry system from the instructor/TA interface.

After opening/posting from the second component, we see the following error:
Screenshot 2023-06-22 at 8 11 58 PM

NOTE:  I see a different line number depending on what exact version of the code I have installed (both recent versions), but I have reproduced the error on 2 machines.

cjreed121:
If a student makes multiple grade inquiries by component, the grading and submission page will crash.
Fixes #9467

### What is the new behavior?
Both pages now load fine.

### Other information?
The issue was caused by `[1=>'USER_ID']` being passed into a query when it expected some element at index 0. To fix this a simple `array_values` was used to reset the keys.

### DIFF
--- a/site/app/views/submission/HomeworkView.php
+++ b/site/app/views/submission/HomeworkView.php
@@ -1229,7 +1229,7 @@ class HomeworkView extends AbstractView {
                 $author_user_ids = array_map(function ($post) {
                     return $post["user_id"];
                 }, $grade_inquiry_posts_for_id);
-                $author_user_groups = $queries->getAuthorUserGroups($author_user_ids);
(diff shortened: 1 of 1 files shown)


bmcutler (approved):
problem fixed!  thanks

PR #9472, opened 2023-06-25
[SYSADMIN ACTION][Feature:System] HTTP/2 Support

### LINKED ISSUE #3850: Setup http/2 on the server

04vmatsibekker:

* [x] Documentation has been updated/added if relevant (https://github.com/Submitty/submitty.github.io/pull/438/)

Closes #3850 

Currently, the Apache2 serves all content in HTTP/1.1.  And the server uses `prefork` as the multiprocessing module.  The module is introduced by `mod_php`, which is not used in production.  

### What is the new behavior?

Most browsers only support HTTP/2 when the TLS is enabled.  The script will also enable HTTPS feature in the development environment.  If a browser does not support HTTP/2, then the server will serve contents in HTTP/1.1.

Developers could use `dev-upgrade-h2.sh` to change the HTTP version:

- `dev-upgrade-h2.sh up`: Upgrade
  After upgrading, the server uses TLS and HTTP/2; or HTTP/1.1 + TLS if a browser does not support `h2`.

- `dev-upgrade-h2.sh down`: Downgrade
  The server uses HTTP/1.1, or HTTP/2 over TCP (`h2c`) if a browser supports `h2c`.

`mod_php` will be disabled and another multiprocessing module (`mpm_event`) will be used in Apache2.  This is effective before upgrading/downgrading.

### Other information?

- NGINX
  The NGINX server still serves WebSocket traffics with HTTP/1.1.

  **Note that** most browsers could not trust WebSocket traffics with self-signed certificates.  If you are dealing with WebSocket related features, there are some workarounds:
    - Copy the certs from VM and trust it on your host system; OR
    - Trust/Ignore the certificates on browsers; OR
    - Downgrade to HTTP/1.1 using `.setup/dev-upgrade-h2.sh down`.

  Trusting/ignoring certificates on browsers:
    - For Chrome, start Chrome with `--ignore-certificate-errors` option.
    - For Firefox...
      1. trust the certificate when the warning pops up;
      2. type `https://localhost:8443/ws` and select trust again.

  This will not impact the production as it is equipped with standard certificates from the trusted CA.

- Performance Tuning, Production
  See documentations above.

### DIFF
--- a/.github/actions/e2e-Setup-Composite/action.yml
+++ b/.github/actions/e2e-Setup-Composite/action.yml
@@ -117,6 +117,7 @@ runs:
         sudo apt-get install php${PHP_VER}-fpm
         sudo apt-get install poppler-utils
         sudo apt-get install valgrind
+        sudo apt-get install ca-certificates
         sudo apt-get install moreutils
       shell: bash
 
@@ -155,7 +156,7 @@ runs:
     - name: Set up apache
       run: |
         cd $SUBMITTY_REPOSITORY
-        sudo a2enmod include rewrite actions cgi alias headers suexec authnz_external headers proxy_fcgi proxy_http proxy_wstunnel
+        sudo a2enmod include rewrite actions cgi alias headers suexec authnz_external headers proxy_fcgi proxy_http proxy_wstunnel ssl
         sudo cp .setup/php-fpm/pool.d/submitty.conf /etc/php/$PHP_VER/fpm/php-fpm.conf
         sudo mkdir -p /run/php
         sudo chown www-data:www-data /run/php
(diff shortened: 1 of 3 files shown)


04vmatsibekker:
PR made by [Viyerelu23333](https://github.com/Viyerelu23333)

Issue #8295, opened 2022-08-15
Create Gradeable Submission Page Tests


MBon1:
**What problem are you trying to solve with Submitty**
Currently, there are no tests for the gradeable submission page.  

**Describe the way you'd like to solve this problem**
It would be nice to have Cypress e2e tests, testing features found on the gradeable submission page such as: 
 - Playing confetti animation
 - Showing details of an individual test case
 - Hiding details of a loaded test case
 - Canceling a loading test case (#7986)
 - Expand/Collapse All Test Cases (#8217)

PR #9487, opened 2023-06-27
[Testing:Submission] Cypress Gradeable Test Case Tests

mahi-pas:
Partially Fixes Issue #8295 

### What is the new behavior?
Tests: 
- Playing confetti animation
- Showing details of an individual test case
- Hiding details of a loaded test case
- Canceling a loading test case (https://github.com/Submitty/Submitty/pull/7986)
- Expand/Collapse All Test Cases (https://github.com/Submitty/Submitty/pull/8217)


site/cypress/e2e/gradeables/gradeable_results.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    beforeEach(() => {
+        cy.visit(['sample','gradeable','closed_team_homework']);
+        cy.login('student');
+
+    });
+
+    it('Should display confetti', () => {
+        cy.get('#confetti_canvas[style="display: block;"]').should('not.exist');
+        cy.get('div[onclick="addConfetti();"]').click();
+        cy.get('#confetti_canvas[style="display: block;"]').should('exist');
```
williamjallen, line 13: These need to be targeted in a more specific way.  Targeting based on a JS function name or an inline CSS style is fragile.  It would be more robust to target these elements based on an ID or class, or via a data tag similar to what Cameron did in #9503.

site/cypress/e2e/gradeables/gradeable_results.spec.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    it('Should show and hide details', () => {
+        //Open
+        cy.get('#testcase_1').should('have.css', 'display').and('match', /none/);
+        cy.get('.loading-tools-show').eq(1).click();
+        cy.get('#testcase_1').should('have.css', 'display').and('not.match', /none/);
+        //Close
+        cy.get('.loading-tools-hide').first().click();
+        cy.get('#testcase_1').should('have.css', 'display').and('match', /none/);
```
williamjallen, line 23: We should not be targeting things based on CSS properties for any reason.  It would be better to use `.should('exist')` or .should('be.visible')` here.  Apply throughout.

williamjallen (changes requested):

williamjallen (changes requested):
It looks like you've checked in a bunch of irrelevant changes here which need to be reverted.

04vmatsibekker (changes requested):
I believe you have a bad merge into main, and it is making the code look like a lot of files are changed. Please go through the code and fix the merge main mistake.

site/cypress/e2e/gradeables/gradeable_results.spec.js
mahi-pas: done

site/cypress/e2e/gradeables/gradeable_results.spec.js
mahi-pas: done

mahi-pas:
Fixed the merge stuff, now it's down from 1,500 lines to just 50 lines added lol

site/cypress/e2e/gradeables/gradeable_results.spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+describe('Testing functionality of Autograder Results', () => {
+    beforeEach(() => {
+        cy.visit(['sample','gradeable','closed_team_homework']);
+        cy.login('student');
+
+    });
+
+    it('Should display confetti', () => {
+        cy.get('#confetti_canvas').should('have.css', 'display').and('not.match',/block/);
```
IDzyre, line 11: I don't have a working VM to test this locally, is there a reason to test the CSS 'display' instead of using .should('be.visible') or .should('exist')  (Similar to William's previous comment)

williamjallen: Yep, this needs to be fixed before this PR can be merged.

williamjallen (changes requested):
@IDzyre's comment needs to be addressed before this can be merged.

site/cypress/e2e/gradeables/gradeable_results.spec.js
mahi-pas: The reason I didn't change it is because should('be.visible') doesn't work on confetti canvas because its ancestor element has position:fixed. The way it's displayed in js code is by changing the display property, so querying this specifically is better for testing the code.

mahi-pas:
It wasn't passing because github CI doesn't simulate autograding, thus the tests would never find any proper values. I talked with Barb and she said we should leave it as a local only test. As for the testing css display value for confetti, the way the js changes the visibilty of the confetti is by changing the css display value, so it's the best way to test. 

TLDR:
- This test wont work on github CI, it should be kept as a local test
- CSS display value for confetti is the only way I can think of to test confetti

PR #9500, opened 2023-07-02
[Refactor:System] Changes for Code Quality

musaabimran:

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -912,7 +912,8 @@ function showEditPostForm(post_id, thread_id, shouldEditThread, render_markdown,
                 const thread_title = json.title;
                 const thread_lock_date =  json.lock_thread_date;
                 const thread_status = json.thread_status;
-                let expiration = json.expiration.replace(/-/g, '/');
+                let expiration = json.expiration.replace('-', '/');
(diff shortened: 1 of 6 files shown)


site/public/js/forum.js
```diff
@@ -1680,7 +1679,7 @@ function refreshCategories() {
         $(this).trigger('eventChangeCatClass');
     });
 
-    $('.cat-buttons').bind('eventChangeCatClass', changeColorClass);
+    $('.cat-buttons').on('eventChangeCatClass', changeColorClass);
```
williamjallen, line 1682: `.bind()` and `.on()` are equivalent in the current version of jQuery.  I believe that the `.on()` is slightly preferred, but all current usages of `.bind()` should be converted to `.on()` and an eslint rule prohibiting `.bind()` should be added if you want to make this change.

williamjallen (changes requested):
The changes in this PR are completely independent of one another, and should be split into separate PRs.  That said, several of the changes in this PR are also trivial, and are probably not worth making a PR to fix by themselves.

Legitimate bugs should be fixed one at a time in separate PRs.  If you want to make code style changes, it would be better to apply a consistent set of changes to a broader section of the codebase.

Issue #9335, opened 2023-05-23
Port Selenium tests to Cypress


skara9:
Currently, we do e2e testing in both selenium and cypress, two different testing frameworks. This slows down our workflows and requires more effort to maintain. Most of our selenium tests are also flagged to be skipped on CI mostly due to the lack of necessary components. The codebase has changed significantly while all of these tests were being skipped, so it is likely that a majority of them need major refactor if they are to work at all with the current or future codebase.

Furthermore, the tests that do work in CI fail most of the time. Workflows often need to be re-run 3-4 times before they pass selenium. This makes ongoing development challenging and slows productivity.

Hence, we will be gradually porting over the selenium tests to cypress so we can run all of our e2e tests under the single framework, saving time and energy.

---

Here is the breakup of our current selenium tests that need to be rewritten in Cypress, along with their complexity (represented with ✦'s), and the number of tests that are currently being run in CI:
- [x] #9337
  * Relies on java dependency and config files
- [x] #9338
  * Currently fails the most often (on merge_threads)
- [x] #9339
- [x] #9340
- [x] #9341
   * Out of sync with latest sample queue data
- [x] #9342
- [x] #9343
- [x] #9344
- [x] #9345
- [x] #9346
- [x] #9347

PR #9503, opened 2023-07-03
[Testing:Developer] Change sidebar tests to Cypress

### LINKED ISSUE #9344: Sidebar   ✦✦   (7/10)

IDzyre:
Currently the sidebar tests are written in Selenium

### What is the new behavior?
This PR re-writes the sidebar tests to be written in Cypress. 

Related to #9335
Closes #9344


site/cypress/e2e/sidebar.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    cy.sidebarContains('My Courses', '/home');
+    cy.sidebarContains('My Profile', '/user_profile');
+    cy.sidebarContains('Authentication Tokens', '/authentication_tokens');
+    cy.sidebarContains('Calendar', '/calendar');
+    if (user.includes('instructor')) {
+        cy.sidebarContains('Docker', '/admin/docker');
+        cy.sidebarContains('New Course', '/home/courses/new');
+        cy.sidebarContains('Autograding Status', '/autograding_status');
+    }
+    cy.sidebarContains('Collapse Sidebar', 'javascript: toggleSidebar();');
```
williamjallen, line 34: Is there a better way to target this item?  Targeting based on a JS function name seems fragile...

site/cypress/e2e/sidebar.spec.js
```diff
@@ -0,0 +1,83 @@
...
+        else if (title.includes('Discussion Forum')) {
+            cy.get('#main > .content').should('contain', 'Create Thread');
+        }
+        else if (title.includes('Collapse Sidebar')) {
+            return;
+        }
+        else {
+            cy.get('#main > .content').should('contain', title);
+        }
+    }
```
williamjallen, line 21: Complicated logic like this is prone to failure.  What prevents something from changing and accidentally cutting off this branch, leaving your tests un-run?

I think it would be good to cut back on the branches here (and elsewhere in the custom commands you've created), but I can't think of any specific suggestions as for how to make it better.  If you aren't able to find something cleaner without a significant amount of effort, this is probably good enough.

site/cypress/e2e/sidebar.spec.js
```diff
@@ -0,0 +1,83 @@
...
+Cypress.Commands.add('baseSidebar', (user) => {
+    cy.sidebarContains('My Courses', '/home');
+    cy.sidebarContains('My Profile', '/user_profile');
+    cy.sidebarContains('Authentication Tokens', '/authentication_tokens');
+    cy.sidebarContains('Calendar', '/calendar');
+    if (user.includes('instructor')) {
+        cy.sidebarContains('Docker', '/admin/docker');
+        cy.sidebarContains('New Course', '/home/courses/new');
+        cy.sidebarContains('Autograding Status', '/autograding_status');
+    }
```
williamjallen, line 33: Why not have a separate command for the instructor sidebar items?  Doing that ensures that the tests get run every time, as opposed to the current solution which could become fragile if something cuts off this branch somehow.

site/cypress/e2e/sidebar.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    cy.sidebarContains('Office Hours Queue', `/courses/${currentSemester}/sample/office_hours_queue`);
+    cy.sidebarContains('Polls', `/courses/${currentSemester}/sample/polls`);
+    cy.sidebarContains('Course Materials', `/courses/${currentSemester}/sample/course_materials`);
+    cy.sidebarContains('Discussion Forum', `/courses/${currentSemester}/sample/forum`);
+    cy.sidebarContains('My Late Day', `/courses/${currentSemester}/sample/late_table`);
+
+    if (user !== 'student') {
+        cy.sidebarContains('Student Photos', `/courses/${currentSemester}/sample/student_photos`);
+        if (user !== 'ta') {
+            cy.sidebarContains('New Gradeable', `/courses/${currentSemester}/sample/gradeable`);
```
williamjallen, line 52: Don't you also want to assert that these sidebar items *don't* show up for students?

shailpatels (changes requested):

site/cypress/e2e/sidebar.spec.js
```diff
@@ -0,0 +1,102 @@
+import { getCurrentSemester } from '../support/utils.js';
+
+const currentSemester = `${getCurrentSemester()}`;
+
+Cypress.Commands.add('sidebarContains', (title, extension, header = title) => {
```
shailpatels, line 5: You don't have to change this but just as a heads for for future cases you can use simple js `function`s to group similar behavior since you'll still have access to the `cy` object. You can also return the result of a `cy` action to chain the result in the caller.

Typically commands are used for test suite wide logic since you can import and use them wherever

shailpatels (approved):
Looks good, thanks for the test updates!

williamjallen (approved):

Issue #9482, opened 2023-06-26
Refactor Cypress Tests to Use Cypress Variables


mahi-pas:
**What problem are you trying to solve with Submitty**
Cypress recommends using its own variable system when returning Cypress elements rather than javascript variables. In order to make our code reflect the best practices, we should refactor our tests to use Cypress's system.

For more info:
https://docs.cypress.io/guides/references/best-practices#Assigning-Return-Values
https://docs.cypress.io/guides/core-concepts/variables-and-aliases

**Describe the way you'd like to solve this problem**
Refactor all the tests in `site/cypress/e2e` to use Cypress variables when needed.

**Additional context**
For information on installing and using Cypress, consult the documentation:
https://submitty.org/developer/testing/cypress

To see a test with proper cypress variables, reference `site/cypress/e2e/Cypress-UI/pronoun_change.spec.js` or read the comments on the PR #9433 which added that test.

mahi-pas:
@tarunsamanta2k20 would you like to work on this?

Mehul-Agrawal410:
hey I am new to contributing and would like to try this out if you could assign it to me

IDzyre:
The way the Submitty organization works, is we cannot assign external contributors to issues. However if you create a pull request and link it to the issue, we, and other contributors will know you are working on it.

tarunsamanta2k20:
@IDzyre please run the cypress test once for my PR. And send  me the log what is the error.

PR #9504, opened 2023-07-05
[Refactor:Developer] Refactor forum tests functions

IDzyre:

#9393 added forum Cypress tests, however I created them with javascript functions. 

### What is the new behavior?
Related to #9482
This PR refactors the forum cypress tests to use Cypress commands functions instead of javascript functions.

### DIFF
--- a/site/cypress/e2e/test_forum.spec.js
+++ b/site/cypress/e2e/test_forum.spec.js
@@ -10,7 +10,7 @@ const reply3 = 'Cypress Reply 3 Cypress';
 const merged1 = 'Merged Thread Title: '.concat(title3, '\n\n', content3);
 const merged2 = 'Merged Thread Title: '.concat(title2, '\n\n', content2);
 
-const createThread = (title, content, category) => {
+Cypress.Commands.add('createThread', (title, content, category) => {
     // Add more to tests for uploading attachments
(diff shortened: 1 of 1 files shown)


williamjallen (approved):
Changes seem reasonable to me.  It's worth noting that a few other Cypress tests seem to use functions, so it would be a good idea to refactor those as well (in separate PRs).

shailpatels (changes requested):
I don't think this PR specifically addresses the main concerns in the issue, changing the function signatures to cypress commands doesn't seem to make a difference since they all just contain a bunch of `cy` functions in them which can be done 

The issue opened seems to be more concerned with trying to use the result of a cypress function in a js variable which is not occurring here

https://docs.cypress.io/api/cypress-api/custom-commands#Best-Practices

IDzyre:
> I don't think this PR specifically addresses the main concerns in the issue, changing the function signatures to cypress commands doesn't seem to make a difference since they all just contain a bunch of `cy` functions in them which can be done
> 
> The issue opened seems to be more concerned with trying to use the result of a cypress function in a js variable which is not occurring here
> 
> https://docs.cypress.io/api/cypress-api/custom-commands#Best-Practices

Apparently I did not read all of the cypress API docs, thanks for bringing this up.

PR #9505, opened 2023-07-10
[Testing:Developer] Change selenium profile tests to cypress

IDzyre:
Currently there are some profile tests that are run in selenium.

- Make sure the correct names are shown before editing
- Testing the time zone selector
- Uploading a profile photo
- Flagging a profile photo as inappropriate

### What is the new behavior?
These have been re-written in cypress


site/cypress/e2e/profile.spec.js
```diff
@@ -71,6 +71,61 @@ describe('Test cases revolving around user profile page', () => {
...
+        cy.get('[data-testid="givenname-row"]').should('contain.text', 'Quinn');
+        cy.get('[data-testid="familyname-row"]').should('contain.text', 'Instructor');
+        cy.get('[data-testid="pronouns-row"]').should('contain.text', 'She/Her');
+        cy.get('[data-testid="email-row"]').should('contain.text', 'instructor@example.com');
+    });
+
+    // Selenium test_time_zone_selection
+    it('Time zone selector should work', () => {
+        // 420 time zones
+        const random_number = Math.floor(Math.random() * 420);
```
williamjallen, line 86: Random numbers in tests are *BAD*.  What if this test fails in one single case out of the possible 420?  Rerunning the test would likely make it pass and look like a flaky test.  Instead of this, check fixed edge cases which can be deterministically checked.

williamjallen (changes requested):

site/cypress/e2e/profile.spec.js
IDzyre: Wouldn't it be good to catch that single case rather than testing just a few cases? The old selenium tests chose one of the time zones at random,   `# Now check if updating the time_zone is correctly working
        # Choose a random option and select it`

IDzyre: I changed this away from random selection.

site/cypress/e2e/profile.spec.js
```diff
@@ -71,6 +71,59 @@ describe('Test cases revolving around user profile page', () => {
...
+
+    // Selenium test_time_zone_selection
+    it('Time zone selector should work', () => {
+        cy.get('[data-testid="time-zone-dropdown"]').should('contain.text', 'NOT_SET/NOT_SET');
+        cy.get('[data-testid="time-zone-dropdown"]').select('(UTC-04:00) America/New_York');
+        // Since the login success message is still up, we get the next message.
+        cy.get('[data-testid="popup-message"]').next().should('contain.text', 'Time-zone updated successfully');
+        cy.get('[data-testid="time-zone-dropdown"]').select('NOT_SET/NOT_SET');
+    });
+
```
04vmatsibekker, line 91: So perhaps I a misunderstanding something, but in the Selenium tests I saw a test called "test_edit_preferred_names" that I expected to be in this cypress test at around this part of the code.

04vmatsibekker (changes requested):

site/cypress/e2e/profile.spec.js
IDzyre: This functionality is already tested with the `it('Should test the modifying of the values')` test, I believe

site/cypress/e2e/profile.spec.js
williamjallen: > Wouldn't it be good to catch that single case rather than testing just a few cases? The old selenium tests chose one of the time zones at random, `# Now check if updating the time_zone is correctly working # Choose a random option and select it`

Yes, if there are edge cases, you definitely want to have a specific test case for them.  You just don't want to risk running into an unforeseen edge case accidentally by making the test non-deterministic.

williamjallen (approved):
Changes look good to me, pending the resolution of @04vmatsibekker's comments.

04vmatsibekker (approved):

PR #9512, opened 2023-07-11
[Feature:InstructorUI] Number of unresolved inquiry per grader

### LINKED ISSUE #9256: Show list of outstanding grade inquiries
I teach a large course (200+) students and have 5 TAs doing grading. It would be very convenient to be able to see all the open grade inquiries and which grader they are assigned to, so I can ping them if they are left often long, or resolve them myself. Currently I have to click and scroll through each gradeable.

Add a button the gradeables page to view all open grade inquiries. The information shown should include which grader and TA were assigned to the section, how long it has been open, which student requested it, and link directly to each inquiry so it is convenient for us to resolve them. The full instructors should be able to see all open inquiries, while TAs and graders should see those assigned to them.

Currently I tell my TAs to resolve grade inquiries and then follow behind them by manually opening each gradeable and scrolling through each of them to see which are still open. It would save time and effort (and I think also make the TAs close inquiries faster) if there was simple feature that lists all the open inquiries that have yet to be resolved and who owns them at one click.

**Additional context**
None available.

ziesski:

Fixes #9256
Currently, Stats page does not have number of unresolved inquiries of each graders.
Which makes it harder for instructors to pester corresponding grader.

### What is the new behavior?
With new improvement, Stats page will include the number of unresolved inquiries that are assigned to each graders.
The behavior varies depending on gradeable component status. 
If it is by component, the count of exact grader who graded the component will be incremented.
If it is not by component, the count of all graders who was involved in the student's gradeable will be incremented.
Hence, the sum may not add up exactly in such case.

Screen Shot 2023-07-12 at 3 20 58 PM

One idea (or further improvement) may be to sort the gradeable page by inquiries.


site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1190,6 +1190,11 @@ private function updateGradeable(Gradeable $gradeable, $details) {
                 $this->core->getQueries()->convertInquiryComponentId($gradeable);
             }
 
+            if ($prop === 'grade_inquiry_per_component_allowed' && $post_val === true && !$gradeable->isGradeInquiryPerComponentAllowed()) {
+                $this->core->getQueries()->revertInquiryComponentId($gradeable);
+                echo "this";
```
Youssef-Chip: Code looks clean just remove this debug statement here.

Youssef-Chip (approved):
Functionality works as expected and the code is clean. Once it passes the test, should be ready to merge

site/app/controllers/admin/AdminGradeableController.php
ziesski: Thanks for pointing out, removed :)

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5158,14 +5158,51 @@ public function getNumberGradeInquiries($gradeable_id, $is_grade_inquiry_per_com
...
+    /**
+     * This is used to revert non-component inquiries back to by-component inquiries
+     * convertInquiryComponentId function modifies the lowest gc_id of each student of the gradeable to null.
+     * If instructor decides to switch to non-component from by-component then switch back to by-component,
+     * this revertInquiryComponentId function will fetch the gc_id from grade_inquiry_discussion and update accordingly.
+     * @param Gradeable $gradeable
+     * @return void
+     */
+    public function revertInquiryComponentId($gradeable): void {
+        $this->course_db->query("UPDATE grade_inquiries SET gc_id=(SELECT b.gc_id FROM grade_inquiry_discussion b WHERE grade_inquiries.id = b.grade_inquiry_id) WHERE grade_inquiries.gc_id IS NULL AND grade_inquiries.g_id=?;", [$gradeable->getId()]);
```
williamjallen, line 5203: This line is incredibly long and hard to read.  Please format it so it's more readable.  Perhaps something like this:
```suggestion
        $this->course_db->query("
            UPDATE grade_inquiries
            SET gc_id = (
                SELECT b.gc_id
                FROM grade_inquiry_discussion b
                WHERE grade_inquiries.id = b.grade_inquiry_id
            )
            WHERE
                grade_inquiries.gc_id IS NULL
                AND grade_inquiries.g_id = ?
        ", [$gradeable->getId()]);
```

williamjallen: Also, it might be good to choose a different alias for `grade_inquiry_discussion`.  `b` is seemingly completely unrelated to `grade_inquiry_discussion`.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5158,14 +5158,51 @@ public function getNumberGradeInquiries($gradeable_id, $is_grade_inquiry_per_com
...
     public function convertInquiryComponentId($gradeable) {
         $this->course_db->query("UPDATE grade_inquiries SET gc_id=NULL WHERE id IN (SELECT a.id FROM (SELECT DISTINCT ON (t.user_id) user_id, t.id FROM (SELECT * FROM grade_inquiries ORDER BY id) t WHERE t.g_id=?) a);", [$gradeable->getId()]);
     }
 
+    /**
+     * This is used to revert non-component inquiries back to by-component inquiries
+     * convertInquiryComponentId function modifies the lowest gc_id of each student of the gradeable to null.
+     * If instructor decides to switch to non-component from by-component then switch back to by-component,
+     * this revertInquiryComponentId function will fetch the gc_id from grade_inquiry_discussion and update accordingly.
+     * @param Gradeable $gradeable
```
williamjallen, line 5199: This should be switched to a native PHP type annotation since those can be enforced by the language itself.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5158,14 +5158,51 @@ public function getNumberGradeInquiries($gradeable_id, $is_grade_inquiry_per_com
...
+     * @return void
```
williamjallen, line 5200: You already have a native PHP type annotation so you don't need to have a PHPDoc annotation as well.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5158,14 +5158,51 @@ public function getNumberGradeInquiries($gradeable_id, $is_grade_inquiry_per_com
...
     }
 
+    /**
+     * get the grader and the count of the gradeable component for inquiry
+     * If the grade is not-by component, function will return all graders who were involved in grading the student
+     * @param string $gradeable_id
+     * @param  bool $is_grade_inquiry_per_component_allowed
+     * @return array<string, int> of gcd_grader_id and count of unresolved grade inquiries of the grader
+     */
+    public function getGraderofGradeInquiry($gradeable_id, $is_grade_inquiry_per_component_allowed = true) {
```
williamjallen, line 5168:
```suggestion
     * @return array<string, int> of gcd_grader_id and count of unresolved grade inquiries of the grader
     */
    public function getGraderofGradeInquiry(string $gradeable_id, bool $is_grade_inquiry_per_component_allowed = true): array {
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5158,14 +5158,51 @@ public function getNumberGradeInquiries($gradeable_id, $is_grade_inquiry_per_com
...
+        $return = [];
+        if ($is_grade_inquiry_per_component_allowed) {
+            $this->course_db->query("SELECT count(b.*), b.gcd_grader_id FROM grade_inquiries a JOIN gradeable_component_data b ON a.gc_id = b.gc_id JOIN gradeable_data c ON b.gd_id = c.gd_id AND a.user_id = c.gd_user_id WHERE a.status = '-1' AND a.g_id=? GROUP BY b.gcd_grader_id", [$gradeable_id]);
```
williamjallen, line 5171: This line is very long and is difficult to read.  Also, the aliases don't really make sense.  Please change them to a character, or series of characters which makes sense given the context.

```suggestion
            $this->course_db->query("
                SELECT
                    count(b.*),
                    b.gcd_grader_id
                FROM grade_inquiries a
                INNER JOIN gradeable_component_data b ON (
                    a.gc_id = b.gc_id
                )
                INNER JOIN gradeable_data c ON (
                    b.gd_id = c.gd_id
                    AND a.user_id = c.gd_user_id
                )
                WHERE
                    a.status = -1
                    AND a.g_id = ?
                GROUP BY b.gcd_grader_id
            ", [$gradeable_id]);
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5158,14 +5158,51 @@ public function getNumberGradeInquiries($gradeable_id, $is_grade_inquiry_per_com
...
+            foreach ($this->course_db->rows() as $row) {
+                $return[$row['gcd_grader_id']] = $row['count'];
+            }
+            return $return;
+        }
+        else {
+            $this->course_db->query("SELECT count(result.*), result.gcd_grader_id FROM (SELECT DISTINCT a.user_id, b.gcd_grader_id FROM grade_inquiries a JOIN gradeable_data c ON a.g_id = c.g_id AND a.user_id = c.gd_user_id JOIN gradeable_component_data b ON c.gd_id = b.gd_id WHERE a.status = '-1' AND a.g_id=?) AS result GROUP BY result.gcd_grader_id", [$gradeable_id]);
```
williamjallen, line 5178: Same here.

```suggestion
            $this->course_db->query("
                SELECT
                    count(result.*),
                    result.gcd_grader_id
                FROM (
                    SELECT
                        DISTINCT a.user_id,
                        b.gcd_grader_id
                    FROM grade_inquiries a
                    INNER JOIN gradeable_data c ON (
                        a.g_id = c.g_id
                        AND a.user_id = c.gd_user_id
                    )
                    INNER JOIN gradeable_component_data b ON (
                        c.gd_id = b.gd_id
                    )
                    WHERE 
                        a.status = -1 
                        AND a.g_id = ?
                ) AS result
                GROUP BY result.gcd_grader_id
            ", [$gradeable_id]);
```

williamjallen: Why is the subquery necessary here?

williamjallen (changes requested):
This is on the right track, but it needs to be cleaned up a bit.

site/app/libraries/database/DatabaseQueries.php
ziesski: I agree that it is hard to read, and I did initially wrote with necessary space and indentation, yet the code style I referenced were mostly condensed to one line, which I thought was intentional. I will make adequate changes.

site/app/libraries/database/DatabaseQueries.php
ziesski: thanks for your insight

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5158,14 +5158,95 @@ public function getNumberGradeInquiries($gradeable_id, $is_grade_inquiry_per_com
...
+            SET gc_id = (
+                SELECT gid.gc_id
+                FROM grade_inquiry_discussion gid
+                WHERE grade_inquiries.id = gid.grade_inquiry_id
+            )
+            WHERE
+                grade_inquiries.gc_id IS NULL
+                AND grade_inquiries.g_id = ?
+        ", [$gradeable->getId()]);
+    }
```
williamjallen, line 5248: It feels like there might be edge cases we're missing here, but I can't quite figure out why it feels wrong.

williamjallen (approved):
Code seems reasonable.  It might be worthwhile to have someone who is more familiar with the grade inquiries DB schema take a look at the query I noted before this gets merged.

site/app/libraries/database/DatabaseQueries.php
ziesski: After some thoughts, I agree that there are some corner cases that is not handled. One case is submissions made during non-component setting. In such case grade_inquiry_discussion gc_id is null as well, so there is no point of updating the grade_inquiries table (b/c both gc_id are null). Hence, it works only for submissions made during by_component setting. However, I tested locally and students submitted during non-component will continue to see All inquiry discussion, so it should not be a huge problem.

bmcutler: Discussed with @ziesski 
This is actually building on a bug/oversight in the original code.
When we switch from 'by component' to 'not by component' ALL gc_ids in the inquiry should be cleaned up and made NULL.  This function should not exist.  We should not be attempting to revive stale gc_id data in these super rare 'what if' scenario.  This function should be simplified/deleted when this bug is removed.  

@ziesski is making a separate issue and will like this PR and this function.

bmcutler (approved):
@ziesski is making an issue to report existing bug ...  when we switch from 'by component' to 'not by component' in some cases we get an 'internal server error' and the switch fails to save.

PR #9517, opened 2023-07-11
[Feature:Developer] Integrate Cypress Cloud 

IDzyre:
Currently we just use the testing component of cypress.

### What is the new behavior?
This will add the setup to have cypress link with cypress cloud, which will allow for easier parallelization, and it shows recordings of the tests in the cloud. 

### Other information?
~~The open source license allows for 100,000 runs (specific it('', ()=>{})  functions, not spec files) per month, so this is the testing period to see if this is enough, if not, it may be better to manually parallelize the tests and use cloud as a different feature, to see why cypress is failing in CI and not locally, but we will see. I have also communicated with some people at cypress, and they indicated that we may be able to get more runs, but I haven't communicated more about that.~~

This information is old. We now have an OSS cypress cloud 'organization', which allows for unlimited runs, not 100,000/month.


.yamllint
```diff
@@ -8,3 +8,4 @@ rules:
 ignore: |
   site/node_modules/*
   site/vendor/*
+  .github/workflows/cypress_ci_testing.yml
```
williamjallen, line 11: This probably shouldn't be checked in.

williamjallen (commented):
Cypress Cloud looks interesting, but I believe that 100,000/month test runs is not enough to meet our demand at peak times.  I need to take a deeper look later, but these changes seem to be heading down the right path.

.yamllint
IDzyre: I thought it was a file created by cypress, but then I realized it was a biproduct of my testing on a forked repo, this has been removed.

IDzyre:
This should be merged _after_ (if) Submitty gets approved for an open source license, because apparently its different if you try to record, and don't have a valid key, and if you try and record, and don't have any runs left. 

If you don't have any tests left, it doesn't run the cypress tests, a critical flaw that I am looking into.

PR #9521, opened 2023-07-12
[Bugfix:Submission] Notebook Zip Download No User Responses

### LINKED ISSUE #9481: Notebook Zip Download Shouldn't Include User Responses
Splits PR #9087 into 2 PRs.

In Sample > Notebook Filesubmission, PR #9087 hides the responses for problem 1 and 3 (which are multiple choice and text input)
image

However, if you download the zip, problems 1 and 3 are still there.
image

This is a security issue, in case instructors encode sensitive information in response files.

Make sure the download zip button does not include files generated by Notebooks.

Where the download zip function is in code:
It calls the function `downloadSubmissionZip()` in `server.js`, which accesses a route that runs
`downloadSubmissionZip()` in `MiscController.php`

Make sure the function excludes files that are generated by the notebook.

Hacky solution:
Perhaps it can use the gradeable_id to see if it's a notebook, and then see if the file name isn't in the notebook to make sure it's good to include.

In PR #9087, a similar system was implemented in `CurrentVersionResults.twig` to hide the notebook files from being individually downloaded.
image

Better Solution:
have the notebook files return as unable to read in the if statement in line 429 of `MiscController.php`
image

mahi-pas:

Closes #9481 
Currently, notebook generated submissions are still downloadable via the "download all" button.
image

### What is the new behavior?
Notebook-generated submissions won't be downloaded via download all.

In addition, certain submitty-generated files will be removed from the zip download of all gradeables (not just zip). The files are `.submit.timestamp`, `.submit.notebook`, and `.user_assignment_access.json`
image

### Other information?

Test for zip downloads for notebook and non-notebook gradeables to ensure nothing has been affected. Test at different access levels, such as instructor, grader, ta, and student


04vmatsibekker (approved):

NiaHeermance (changes requested):
As a student, indeed Notebook files do not download. However, as an instructor, they do, an unintended behavior.

NiaHeermance (changes requested):
Additionally, TA and Grader also have the Notebook files downloaded when they should not.

NiaHeermance (approved):
The notebook files do not download for all users now. Yay!

bmcutler (changes requested):
For the development course Notebook Filesubmission gradeable I uploaded these 2 files in the Problem 4 box:

Screenshot 2023-07-28 at 11 51 44 AM

However, since the name problem_3_animal.txt matches the instructor chosen name for another question, this file is incorrectly excluded from both the individual file download list and the zip of all files:

Screenshot 2023-07-28 at 11 52 53 AM

site/app/models/gradeable/Gradeable.php
```diff
@@ -930,6 +930,47 @@ public function canStudentDownload() {
         return $this->student_download;
     }
 
+    public function canStudentDownloadFile(int $version, string $file_path, string $root_path): bool {
```
NiaHeermance, line 933: You should add a function comment about the parameters and return type.

NiaHeermance (approved):
Works and looks good! I'm approving since my suggested change is small/doesn't effect functionality, but I would recommend adding a function comment to canStudentDownloadFile since our linter looks for that now.

PR #9522, opened 2023-07-13
[Testing:Developer] Change cypress login to post request

IDzyre:
Currently the cypress login function logs in manually using the UI at the /authentication/login page, and the logout function clicks the logout button on the sidebar. 

### What is the new behavior?
Cypress uses a POST request to manually login and logout without using the UI. 
see https://docs.cypress.io/guides/references/best-practices#Organizing-Tests-Logging-In-Controlling-State
and  https://www.youtube.com/watch?v=5XQOK0v_YRE

### Other information?
After this is merged, it would be good to remove all of the `cy.visit('/')` instances that happen just before a `cy.login()`, as they are now obsolete.


site/cypress/support/commands.js
```diff
@@ -24,50 +24,71 @@
...
+        })
+    })
+});
+
+/**
+* Log into Submitty using the UI, assumes no one is logged in already and at login page
+*
+* @param {String} [username=instructor] - username & password of who to log in as
+*/
+Cypress.Commands.add('loginViaUI', (username = 'instructor') => {
```
williamjallen, line 59: Why does this need to be a command?  Are there any cases where we might want to login by clicking around the UI instead of just making a HTTP request directly?  Shouldn't this just be a test in the login tests?

site/cypress/e2e/login.spec.js
williamjallen, line 1: It would be good to also test the logout button somewhere in here

williamjallen (changes requested):

williamjallen (approved):

PR #9553, opened 2023-07-21
[Feature:TAGrading] Default TA Anonymous Setting

yanliw123:
Anonymous mode for instructors and full access graders is disabled by default for every gradeable.

### What is the new behavior?
On a Gradeable's config page, one can now set the default setting for anonymous mode for instructors and full access graders. Then an individual instructor/full access grader can override the default on a Gradeable's Index/Details page if they desire.

### Images
Gradeable's Settings/Config page to set default anonymous mode for full access gradeers:
image

Gradeable's Details page to override default with "Enable/Disable Anonymous Mode" button:
image


NiaHeermance (commented):
I like the changes so far. However, this setting needs to actually do something, as currently it is hooked up to nothing. The goal is that when "Blinded" is on, Enable Anonymous Grading on the Gradeable's details page is true, and when "Unblinded" is on, Enable Anonymous Grading on the Gradeable's details page is false.

NiaHeermance (changes requested):
Forgot to select Request changes.

NiaHeermance:
Oh also: Enable Anonymous Mode is controlled by a cookie. It's a per browsing setting currently rather than a per gradeable setting, as the same string is used for the cookie for all gradeables. So you may want to modify the cookie so that it is specific to each gradeable. I can help with this if you need it.

Youssef-Chip (commented):
Functionnality works and the code looks clean. Although is the following setting really necessary : What should the default anonymization of students be for instructors and full access graders? 
Since instructors and full access graders can manually change it for themselves in the grading index, my concern is that we're adding too much text in this page ( especially that in peer gradeables, there would be 3 different anonymous settings ( full access graders - limited access graders - students )

NiaHeermance:
That's a fair concern, but adding this setting was the point of this PR, so if we decide we don't want this setting, we shouldn't merge this PR. Barb and I were talking about it, and we thought this setting was a good idea as instructors sometimes want to turn on anonymous mode for TAs, which is what this setting allows.

Youssef-Chip (approved):

NiaHeermance (approved):

PR #9560, opened 2023-07-25
[Feature:Calendar] Filter Calendar Info By Course

### LINKED ISSUE #9558: Filter Calendar By Courses
The current submitty Calendar shows all information at the same time

There should be a modal that allows you to select which classes you want to display

mahi-pas:

Fixes #9558

Can't filter calendar by courses

### What is the new behavior?
You can filter calendar by courses
image
image
image

Also added a cookie for where in the calendar the user left off. previously, if the real month is august and the user was looking at September and they reloaded the screen, the calendar would go to august. Now, it saves the month as a session cookie.

### Other information?

Test by checking that it filters properly, creating archived courses and seeing how it interacts. Test for week, 2 week, and month views.
Test calendar month cookies by making sure it saves properly. Test for week, 2 week, and month views.


04vmatsibekker (changes requested):
2 things.

I think you talked about it last time, but If I go to the next month, then switch to only one course, it goes back to August. Not quite ideal, but it's ok.

Also each course has a color when I am on all courses, I kinda would prefer it if when I went to a particular course, that course's color didn't change, just so that color doesn't throw a user off and make them question if they went to the right course for a second.

mahi-pas:
> 2 things.
> 
> I think you talked about it last time, but If I go to the next month, then switch to only one course, it goes back to August. Not quite ideal, but it's ok.
> 
> Also each course has a color when I am on all courses, I kinda would prefer it if when I went to a particular course, that course's color didn't change, just so that color doesn't throw a user off and make them question if they went to the right course for a second.

I'm currently working on the month issue. As for the colors, I'm going to overhaul how colors are decided when I implement legend, so I will definitely make that the case later.

04vmatsibekker:
Yay! Looks pretty good so far otherwise though.

04vmatsibekker (approved):
Looks pretty good, just fix the tests and I think it will be ready to merge

PR #9564, opened 2023-07-26
[Bugfix:System] Build submitty utils not in shared mount

skara9:

Installation errors on some machines/configurations due to filesystem errors from running setuptools build script within shared directory on developer virtual machine.

### What is the new behavior?
Build has been moved outside of the mounted directory.

### DIFF
--- a/.setup/INSTALL_SUBMITTY_HELPER.sh
+++ b/.setup/INSTALL_SUBMITTY_HELPER.sh
@@ -164,7 +164,9 @@ fi
 
 echo -e "Install python_submitty_utils"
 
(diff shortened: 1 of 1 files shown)


Youssef-Chip (approved):

bmcutler (changes requested):
instead of "cp", we want to do rsync, so it can be repeated (not just on the initial vagrant up, but on each time we re-run INSTALL_SUBMITTY.sh).
Like other places in the code, probably "rsync -ruz --delete" is best.

bmcutler (approved):

PR #9569, opened 2023-07-28
[Feature:InstructorUI] Course Materials Folder Suggestions

NiaHeermance:
### Previous PR Relation
This PR replaces PR #8894 and closes issue #8263.

When uploading a new course material, the user can choose what folder the material is in by specifying the path to the file. This action will create the folder if it does not already exist. However, the user can not see whether the folder already exists when typing the path.
 
### What is the new behavior?
There is now a dropdown list of all folders the user can put the file in, and this list updates as the user's path becomes more specific.

Additionally, when users move a file from one folder to another, folder suggestions appear there as well.

### Images

Dropdown Menu of Folder Paths

image


04vmatsibekker (approved):

bmcutler (approved):
looks good.  thanks :)

PR #9582, opened 2023-08-01
[Bugfix:System] Registration Section Input Validation

### LINKED ISSUE #9573: Registration Section Input Validation
**Describe the bug**
When a long string is entered in the registration section it breaks the Instructor and TA's panel.

**Expected behavior**
The input should be validated:
Limit to ~20 characters
Only allow alpha a-zA-Z, 0-9, -, _ 

**To Reproduce**
1. Login as instructor 
2. Go to a course 
3. Select "Manage Sections"
4. Add a string in "Add Registration Section"
5. See error

**Configuration**
 - OS: Windows, Linux
 - Browser: Chrome, Firefox, Brave

**Screenshots**
image

musaabimran:
When a long string is entered in the registration section it breaks the Instructor and TA's panel.
image

### What is the new behavior?
The input should be validated and the panel does not break and error message is prompted. Closes #9573


site/app/models/User.php
```diff
@@ -611,8 +611,9 @@ public static function validateUserData($field, $data): bool {
                 return preg_match("~^[1-4]{1}$~", $data) === 1;
             case 'registration_section':
                 //Registration section must contain only alpha (upper and lower permitted), numbers, underscores, hyphens.
+                // AND between 0 and 20 chars.
                 //"NULL" registration section should be validated as a datatype, not as a string.
-                return preg_match("~^(?!^null$)[a-z0-9_\-]+$~i", $data) === 1 || is_null($data);
+                return preg_match("~^(?!^null$)[a-zA-Z0-9_\-]{1,20}$~", $data) === 1 || is_null($data);
```
ziesski, line 616: I am not sure if the registration section should be case sensitive, which could potentially confuse users.

ziesski: Will discuss about it during the meeting and update the comment.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2561,7 +2561,7 @@ public function getUsersByRotatingSections($sections, $orderBy = "rotating_secti
      * @return array
      */
     public function getRegistrationSections() {
-        $this->course_db->query("SELECT * FROM sections_registration ORDER BY SUBSTRING(sections_registration_id, '^[^0-9]*'), COALESCE(SUBSTRING(sections_registration_id, '[0-9]+')::INT, -1), SUBSTRING(sections_registration_id, '[^0-9]*$') ");
+        $this->course_db->query("SELECT * FROM sections_registration ORDER BY SUBSTRING(sections_registration_id, '^[^0-9]*'), CAST(SUBSTRING(sections_registration_id, '[0-9]+') AS NUMERIC), SUBSTRING(sections_registration_id, '[^0-9]*$');");
```
ziesski, line 2564: There are two semi colons, do we need the semi colon inside the double quotation?

site/app/models/User.php
musaabimran: No, it doesn't let me update it.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -2561,7 +2561,7 @@ public function getUsersByRotatingSections($sections, $orderBy = "rotating_secti
      * @return array
      */
     public function getRegistrationSections() {
-        $this->course_db->query("SELECT * FROM sections_registration ORDER BY SUBSTRING(sections_registration_id, '^[^0-9]*'), COALESCE(SUBSTRING(sections_registration_id, '[0-9]+')::INT, -1), SUBSTRING(sections_registration_id, '[^0-9]*$') ");
+        $this->course_db->query("SELECT * FROM sections_registration ORDER BY SUBSTRING(sections_registration_id, '^[^0-9]*'), CAST(SUBSTRING(sections_registration_id, '[0-9]+') AS NUMERIC), SUBSTRING(sections_registration_id, '[^0-9]*$')");
```
ziesski, line 2564: What is the reason to change from COALESCE to CAST, I do see that the functionality is same, but there is also other functions that use identical code, would it be better to switch other queries as well?

site/app/models/User.php
ziesski: Update: Barb also thinks it should be case-insensitive, doesn't make sense to have Staff section and STAFF section.

site/app/libraries/database/DatabaseQueries.php
musaabimran: In this query, CAST is being used to convert a substring extracted from "sections_registration_id" into a numeric value because I had to deal with the case that if 20 digit long number in entered, INT or Bigint are not able to handle it.

ziesski: That is true, ::INT will convert to PHP int max, which is 2147483647, I tested it with our current query, and it does fail after 2147483647. However, with your query (using CAST) I am getting below error. 
Screen Shot 2023-08-05 at 10 45 02 AM

site/app/models/User.php
```diff
@@ -632,8 +632,9 @@ public static function validateUserData($field, $data): bool {
                 return preg_match("~^[1-4]{1}$~", $data) === 1;
             case 'registration_section':
                 //Registration section must contain only alpha (upper and lower permitted), numbers, underscores, hyphens.
+                // AND between 0 and 20 chars.
                 //"NULL" registration section should be validated as a datatype, not as a string.
-                return preg_match("~^(?!^null$)[a-z0-9_\-]+$~i", $data) === 1 || is_null($data);
+                return (preg_match("~^(?!^null$)[a-z0-9_\-]{1,20}$~i", $data) === 1 || is_null($data));
```
ziesski, line 637: is the extra parenthesis necessary?

ziesski (changes requested):

ziesski (approved):
INT -> NUMERIC change seems reasonable, which can handle very large digits. 
and, regex limits the length to 1-20. 

Below are some keys I tested, it still fails with 2147483648 (i dont know what is wrong with it)
Screen Shot 2023-08-07 at 1 03 03 PM

Rest of them are fine 
Screen Shot 2023-08-07 at 1 02 24 PM

musaabimran:
> INT -> NUMERIC change seems reasonable, which can handle very large digits. and, regex limits the length to 1-20.
> 
> Below are some keys I tested, it still fails with 2147483648 (i dont know what is wrong with it, may be b/c it is one pass the php int max? ) Screen Shot 2023-08-07 at 1 03 03 PM
> 
> Rest of them are fine Screen Shot 2023-08-07 at 1 02 24 PM

I tried this and it works.
image

PR #7428, opened 2021-12-09
[Feature:TAGrading] NULL section filter

jportorreal00:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

### What is the current behavior?
Closes #7427 

### What is the new behavior?
The new filter allows the stats page to contain NULL sections

PR #9586, opened 2023-08-01
[Feature:InstructorUI] Overridden gradeable is fully graded

ziesski:
Helps #7428 

Currently the grade stat page does not count overridden gradeable as fully graded.
Screen Shot 2023-08-01 at 1 36 15 PM

With this update, the stat page will correctly reflect overridden gradeable.


williamschen23 (approved):
Functionality looks good, didn't look into the code that much

williamjallen (approved):
I'm not convinced that I understand the code fully, but I don't see anything that looks harmful about it.

PR #9588, opened 2023-08-01
[Bugfix:Developer] Remove chrome driver setup

IDzyre:
The Cypress and Integration tests occasionally fail on the setup phase of the github action, with the error 
image

### What is the new behavior?
Since we are no longer using Selenium for CI, and Cypress and Integration tests do not use the chrome driver, the setup has been removed, which should also remove the occasional errors in the setup phase.


04vmatsibekker (approved):
If there were any issues with removing chrome driver, it would have showed up in the ci tests. Thanks for the fix!

MasterOdin:
Selenium was the only thing within CI that required chromedriver, and since selenium was removed, chromedriver can also be removed. Cypress (and most of the node based E2E test frameworks) will download their own drivers / browser instances to help provide more stability for test environments.

PR #9594, opened 2023-08-02
[Bugfix:System] Redundancy of NULL section 

### LINKED ISSUE #9593: Redundancy of NULL Section
**Describe the bug**
Null section can be added by the user causing it to appear twice.

**Expected behavior**
NULL section shouldn't appear twice.

**To Reproduce**
1. Goto Manage Sections
2. Add NULL section
3. See error

**Configuration**
 - OS: Windows, Linux
 - Browser Chrome, Firefox

**Screenshots**
image

musaabimran:
Null section can be added by the user causing it to appear twice.
image

### What is the new behavior?
NULL section does not appear twice. Close #9593

### DIFF
--- a/site/app/models/User.php
+++ b/site/app/models/User.php
@@ -611,8 +611,9 @@ class User extends AbstractModel {
                 return preg_match("~^[1-4]{1}$~", $data) === 1;
             case 'registration_section':
                 //Registration section must contain only alpha (upper and lower permitted), numbers, underscores, hyphens.
+                // AND between 0 and 20 chars.
                 //"NULL" registration section should be validated as a datatype, not as a string.
-                return preg_match("~^(?!^null$)[a-z0-9_\-]+$~i", $data) === 1 || is_null($data);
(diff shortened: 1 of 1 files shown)


ziesski:
I was not able to reproduce the issue, and below are the names I tested. 
Screen Shot 2023-08-04 at 12 24 47 AM

Perhaps a small hint of debugging would be that there are two different types of error messages. 

If I try to make "Section Hi" again, it will return 
Screen Shot 2023-08-04 at 12 26 32 AM

While if I try NULL, null, Null, "NULL" etc it will return below. 
Screen Shot 2023-08-04 at 12 27 03 AM

Please give us more detailed instruction to replicate the issue.

ziesski (changes requested):
I was not able to reproduce the issue, and below are the names I tested. 
Screen Shot 2023-08-04 at 12 24 47 AM

Perhaps a small hint of debugging would be that there are two different types of error messages. 

If I try to make "Section Hi" again, it will return 
Screen Shot 2023-08-04 at 12 26 32 AM

While if I try NULL, null, Null, "NULL" etc it will return below. 
Screen Shot 2023-08-04 at 12 27 03 AM

Please give us more detailed instruction to replicate the issue.

PR #9613, opened 2023-08-04
[Bugfix:Submission] Notebook Generated Files Filter Fix

mahi-pas:

If a student submits a file with the same name as a notebook generated file, it won't be shown in downloads

### What is the new behavior?
Now it will be shown

### Other information?
Make sure this is the case for all user levels, and that it doesn't accidentally show anything it shouldn't (ie, notebook generated files)


NiaHeermance (approved):
Works! Great that the code is shorter now, too.

PR #9528, opened 2023-07-13
Make poster

04vmatsibekker:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant

PR #9617, opened 2023-08-04
[Feature:Developer] Update DockerUI selectors

### LINKED ISSUE #9614: Refactor Docker UI test to use better selectors
Currently the DockerUI cypress tests use fragile selectors, `cy.get('.machine-table > tbody:nth-child(1) > tr:nth-child(3) > td:nth-child(4)')`

Refactor some of the selectors to use test-id's, so UI changes later won't break the tests 

see [Selecting Cypress Elements](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements)

IDzyre:

The DockerUI tests were written with a couple of fragile selectors, and a recent PR #9528 found that any UI change would break the tests.
Closes #9614
### What is the new behavior?
These fragile selectors were changed to use data- tags, so Cypress can find them, even if the amount of elements on the page changes.


04vmatsibekker (approved):

04vmatsibekker (commented):
When running it locally I got this error 
image

Issue #9615, opened 2023-08-04
Unable to access Email Status Page --Server Error


yanliw123:
**Describe the bug**
Impossible to access a key "error" on an object of class "app\entities\email\EmailEntity" that does not implement ArrayAccess interface. in /usr/local/submitty/site/app/templates/EmailStatus.twig:33

**To Reproduce**
1. Login in as an Instructor account 
2. go to Email Status Page.

**Configuration**
 - OS: [e.g. Windows, Linux, ...] Windows 11
 - Browser [e.g. chrome, safari] Chrome

**Screenshots**
![image](https://github.com/Submitty/Submitty/assets/114595115/446fb47d-3e54-4784-9106-d2a90ab8edc7)

RayReyRam:
@yanliw123 the directions to replicate this bug are pretty straight forward, and I'm unable to reproduce this bug. Do you think this bug has been fixed with the related merged PR?

PR #9620, opened 2023-08-04
[Feature:Developer] Check for server error in sidebar tests

IDzyre:

The sidebar Cypress tests only check that the title header is accurate for the page, and since not every page has it's own Cypress test yet, so some errors fall through the cracks. 
Related to #9615
### What is the new behavior?
A line has been added to the sidebar tests to catch the server errors that still allow the page to be viewed.


ziesski (approved):
Please address the merge conflict, otherwise it looks good :)

bmcutler:
@IDzyre please resolve conflicts

PR #9621, opened 2023-08-04
[Bugfix:TAGrading] Add Team anon id column in grading index

yanliw123:
seems to be the right fix. not sure if this will break other tests. See if this passes the cypress tests

Solves the bug when login as student and visit grade for Released Peer Team PDF Homework
image

### What is the new behavior?
add new team anon id in grading index page.
Hide student names in anon mode when grading team homework (bug)
image


NiaHeermance (changes requested):
The fix is great. As an instructor, when I see the Grading Status page of a Team Gradeable as an instrucotor in Anonymous Mode, there is now a team anonymous ID. And if I login as a limited access grader, if the instructor has set for limited access graders to be anonymized, then I see the team anonymous ID too.

I notice something strange though, which I believe existed before your branch. If an instructor is in non-anonymous mode, there is a team ID. However, limited access graders don't see the non-anonymous team ID when they are not blinded. Instead, they just see no team ID at all:

image


Maybe you could fix this in this PR?

PR #9622, opened 2023-08-06
[Feature:RainbowGrades] UI for Academic Integrity

### LINKED ISSUE #9320: add plagiarism text to Rainbow Grades web generator for customization.json
The rainbow grades web form for customization is missing input for academic integrity violations. 

Here is an example of the syntax needed in the json file:

```
    "plagiarism" : [
        { "user" : "XXXXX",     "gradeable" : "XXXXX",      "penalty" :  1.0 },
        { "user" : "XXXXX",     "gradeable" : "XXXXX",      "penalty" :  0.0 },
        { "user" : "XXXXX",     "gradeable" : "XXXXX",      "penalty" :  0.5 }
    ]
```

Let's add a "add academic integrity violation" button to the web UI.  And then the instructor types the user id, gradeable id, and penalty.  We should do some basic validation on the user id string and gradeable id string.   And the penalty should be 0.0-1.0 in value.

ziesski:
closes #9320  
Cannot insert plagiarism data from web-based customization page GUI.
### What is the new behavior?
web-based customization page has space for plagiarism data. 
Screen Shot 2023-08-16 at 3 16 17 PM

Screen Shot 2023-08-16 at 3 19 11 PM

Checks for duplicate entry, penalty value between 0-1


ziesski:
Resolved merge conflict from display option

site/app/templates/admin/RainbowCustomization.twig
```diff
@@ -208,6 +267,84 @@ To remove a curve simply blank all boxes."></i>
         </div>
     </div>
 
+    <script>
+
+        function addToTable() {
```
Youssef-Chip, line 272: Would it be possible to write this function in a js file then call it in the twig file instead ?

Youssef-Chip (changes requested):

site/app/templates/admin/RainbowCustomization.twig
ziesski: Thanks for pointing it out.

site/app/models/RainbowCustomizationJSON.php
```diff
@@ -69,6 +73,16 @@ public function getGradeables() {
         return $this->gradeables;
     }
 
+    /**
+     * Get array of plagiarism
+     *
+     * @return array<mixed>
+     */
+    public function getPlagiarism(): array {
+        return $this->plagiarism;
```
williamjallen, line 82: You have marked the return type of this function as `array<mixed>`, but `$this->plagiarism` is marked as an array of objects.

site/app/models/RainbowCustomizationJSON.php
```diff
@@ -276,6 +294,23 @@ public function addMessage(string $message) {
         $this->messages[] = $message;
     }
 
+
+    /**
+     * Add plagiarism entry to existing array
+     *
+     * @param object $plagiarismEntry
+     */
+    public function addPlagiarismEntry(object $plagiarismEntry): void {
```
williamjallen, line 303: Does this object have a more specific type?

site/public/js/rainbow-customization.js
```diff
@@ -271,6 +271,100 @@ function getGradeableBuckets() {
...
+            user: user,
+            gradeable: gradeable,
+            penalty: penalty,
+        });
+    }
+
+    return plagiarismData;
+}
+
+// eslint-disable-next-line no-unused-vars
```
williamjallen, line 296: It would be better to put the name of this function in an `/** exported */` block at the top of the file as discussed [here](https://submitty.org/developer/software_and_system_design/coding_style_guide/javascript#style-rules).

site/public/js/rainbow-customization.js
```diff
@@ -271,6 +271,100 @@ function getGradeableBuckets() {
...
+function addToTable() {
+    const USERID = document.getElementById('user_id').value.trim();
+    const gradeable = document.getElementById('g_id').value.trim();
+    const penalty = document.getElementById('marks').value.trim();
+
+    // Check for empty fields
+    if (USERID === '' || gradeable === '' || penalty === '') {
+        alert('Please fill in all the fields fyi.');
```
williamjallen, line 304: Keep it concise and professional please.

```suggestion
        alert('Please fill in all the fields.');
```

site/public/js/rainbow-customization.js
```diff
@@ -271,6 +271,100 @@ function getGradeableBuckets() {
...
+    const cellDelete = newRow.insertCell();
+    cellDelete.innerHTML = '<a onclick="deleteRow(this)"><i class="fas fa-trash"></i></a>';
+
+    // Clear the form fields
+    document.getElementById('user_id').value = '';
+    document.getElementById('g_id').value = '';
+    document.getElementById('marks').value = '';
+}
+
+// eslint-disable-next-line no-unused-vars
```
williamjallen, line 362: Same comment here about moving this to an exported comment block at the top of the file.

site/public/js/rainbow-customization.js
```diff
@@ -326,13 +420,16 @@ function buildJSON() {
...
     };
 
     ret = JSON.stringify(ret);
-
     return ret;
 }
 
+
+
+
```
williamjallen, line 432: These whitespace changes are seemingly irrelevant and should be reverted.  The same comment applies to the other extraneous whitespace changes throughout this file.

site/public/js/rainbow-customization.js
```diff
@@ -413,9 +510,9 @@ function ajaxUpdateJSON(successCallback, errorCallback) {
     catch (err) {
         $('#save_status').html(err);
     }
-
 }
 
+
 function displayChangeDetectedMessage() {
     $('#save_status').html('Changes detected, press "Save Changes" to save them.');
```
williamjallen, line 517: It would be better to use jQuery's `.text()` function here.

williamjallen (changes requested):

site/public/js/rainbow-customization.js
```diff
@@ -271,6 +271,100 @@ function getGradeableBuckets() {
...
+    cellUSERID.innerHTML = USERID;
+
+    const cellGradeable = newRow.insertCell();
+    cellGradeable.innerHTML = gradeable;
+
+    const cellPenalty = newRow.insertCell();
+    cellPenalty.innerHTML = penalty;
+
+    const cellDelete = newRow.insertCell();
+    cellDelete.innerHTML = '<a onclick="deleteRow(this)"><i class="fas fa-trash"></i></a>';
```
williamjallen, line 354: It's also not a good idea to be using `.innerHTML` here because it open up the possibility of XSS attacks in the future.

williamjallen (changes requested):

site/app/models/RainbowCustomizationJSON.php
ziesski: class PlagiarismEntry {
    public string $user;
    public string $gradeable;
    public int $penalty;

    public function __construct(string $user, string $gradeable, int $penalty) {
        $this->user = $user;
        $this->gradeable = $gradeable;
        $this->penalty = $penalty;
    }
}

Should I define new class for more specific type?

ziesski: public function __construct(string $user, string $gradeable, int $penalty) {
        $this->user = $user;
        $this->gradeable = $gradeable;
        $this->penalty = $penalty;
    }
Do you think if we should define new class to specify the type?

williamjallen: I don't think there's a need for a whole data class here.  You can just specify the object shape with PHPDoc annotations and I'll be satisfied: https://phpstan.org/writing-php-code/phpdoc-types#object-shapes

bmcutler (commented):
I'll approve, but its possible to enter an invalid user id
@ziesski will fix that quick??

bmcutler (approved):

PR #9629, opened 2023-08-08
[Bugfix:TAGrading] LimAccess blindness affecting FullAccess

NiaHeermance:
On a Gradeable's configuration page, there's a tab called "Grader Assignment" where instructors can set permissions on those who can grade the Gradeable and how much can they see. One setting controls whether Limited Access Graders can see/are blind to who they are grading. However, this setting was also unintentionally affecting whether instructors and Full Access Graders can see who they are grading.

Another issue is that the Limited Access Blindness settings does not blind Team members names on a Gradeable's Status page for limited access graders. Notice that the first issue mentioned above is about limited access settings affecting full access, while this bug is limited access settings not working fully as intended.

### What is the new behavior?
The setting now only affects Limited Access Graders.

### Screenshots

The setting in question:
image

The setting was accidentally hiding the student information in the top left for instructors and Full Access Graders.

image

A blinded Gradeable Status page for a team Gradeable.
image


Youssef-Chip (commented):
There is another issue unrelated to this specific situation. For **team** gradeables, when blinded grading is chosen for limited access graders, it would correctly not display the team members after clicking on grade, but unfortunately on the grading index page, the names would still appear. If you can fix that in this PR it would be great.

Youssef-Chip (approved):

PR #9635, opened 2023-08-09
[Bugfix:Developer] Fix usage of @prop

skara9:

Class properties in the PHP codebase are currently annotated with docstring comments beginning with `@prop @var`. This is not valid phpdoc syntax and results in the types being ignored by static analysis.
Additionally, the phrase `@property` was used interchangeably with `@prop`. This is problematic as `@property` is a recognized keyword used for defining virtual properties on classes.

### What is the new behavior?
All usage of `@prop` and its derivatives have been separated from `@var` so as to not hinder static analysis. Incorrect uses of `@property` have been replaced with `@prop`.

### Other information?

The `@prop` keyword is a custom keyword used in PHP typedoc comments throughout the Submitty codebase. Our code relies on it for certain tasks such as the hydration of models. We also have derivatives such as `@prop-read` and `@prop-write`, which serve as makeshift access controls for properties.

It is worth discussing if there is a better approach to solving this issue, and in which case whether or not this PR should be merged in as a temporary solution while a more permanent refactor is under the process of being worked on.


williamjallen (approved):
Good catch here!  This is definitely a problem we should address.

I think this is a decent solution for now.  Let's go ahead and merge this, and then figure out a long-term solution afterwards.

It's worth noting that since we now use PHP 8, we can use native PHP type declarations on member variables rather than relying on PHPDoc type hints.  In fact, using native type declarations should be preferred, because they can be enforced by PHP itself.  Unfortunately, some of our PHPDoc type hints are not correct, so it might not be a trivial swap.

MasterOdin:
Moving more fully to using doctrine would remove almost all of the models, which could render some of the conversation around `@prop` moot.

PR #9659, opened 2023-08-14
[Feature:System] Create Git Safe DIR during installation

ziesski:
Related to https://github.com/Submitty/RainbowGrades/pull/56

Screen Shot 2023-07-31 at 12 10 48 PM
Currently, Git has security feature called safe directory. Checking version on web-based customization page is ran by submitty_daemon, which has permission problem. (Running it locally, the user is most likely a root user, won't encounter this issue) 

### What is the new behavior?
Screen Shot 2023-08-14 at 11 40 40 AM
During install_system process, the script will now create .gitconfig file in /home/submitty_daemon with owner/group being submitty_daemon. In .gitconfig file contains the safe directory: usr/local/submitty/GIT_CHECKOUT/RainbowGrades

### Other information?
/etc/gitconfig file has hierarchy over this - causing permission error, so be careful when running git config --global --edit command or --system. Although local config can override system/global config, in our case it is checking etc/gitconfig file first and gives error (fatal: unable to access '/etc/gitconfig': permission denied)


04vmatsibekker (approved):

migration/migrator/migrations/system/20230815125306_add_git_safe_directory.py
```diff
@@ -0,0 +1,23 @@
...
+
+    gitconfig_path = f"/home/{DAEMON_USER}/.gitconfig"
+
+    gitconfig_content = """[safe]
+    directory = {}""".format(RAINBOWGRADES_REPOSITORY)
+
+    try:
+        with open(gitconfig_path, "w") as gitconfig_file:
+            gitconfig_file.write(gitconfig_content)
+        subprocess.run(["sudo", "chown", f"{DAEMON_USER}:{DAEMON_USER}", gitconfig_path], check=True)
```
bmcutler, line 16: should you also have a similar chown in install_system.sh as well?

bmcutler (changes requested):

migration/migrator/migrations/system/20230815125306_add_git_safe_directory.py
ziesski: Thanks for the insight, it is always better practice to be explicit than relying on previous code. Just made the change

williamjallen (commented):
Is there any particular reason why we shouldn't just mark each of these directories as "safe" on a system level?  You can run `git config --system --add safe.directory <directory>` to do this if desired.

ziesski:
> Is there any particular reason why we shouldn't just mark each of these directories as "safe" on a system level? You can run `git config --system --add safe.directory <directory>` to do this if desired.

yes, please refer to the original post, I mentioned 

"/etc/gitconfig file has hierarchy over this - causing permission error, so be careful when running git config --global --edit command or --system. Although local config can override system/global config, in our case it is checking etc/gitconfig file first and gives error (fatal: unable to access '/etc/gitconfig': permission denied)"  

So, if we were to create on on the system level, we need to edit /etc/gitconfig, which Barb thought was not a good idea for submitty_daemon to have access to /etc directory.

cjreed121 (changes requested):

migration/migrator/migrations/system/20230815125306_add_git_safe_directory.py
```diff
@@ -0,0 +1,23 @@
...
+
+    except Exception as e:
+        print("Error:", e)
+        print("Failed to write .gitconfig entry or change ownership.")
+
+def down(config):
+    pass
```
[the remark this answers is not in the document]
ziesski: Thanks for your suggestion.

PR #9665, opened 2023-08-15
[Refactor:System] Iclicker removal 

ziesski:
Removing iClicker from install script and bin.


bmcutler:
please also remove the iclicker gradeable example in more_autograding_examples and where that gradeable is created in development course

bmcutler (approved):

PR #9667, opened 2023-08-16
[Bugfix:Calendar] Fixed Calendar for 0-1 Courses

mahi-pas:

image
This bug occurs when a user has 0 courses (i.e. superuser), has the `calendar_show_all` cookie set to 0, and doesn't have a `calendar_course` cookie set. The code does two things wrong:
1. It will by default choose a course from your courses, and use its data for the `calendar_course` cookie. This will give an error if you don't have any courses.
2. It uses the index of 1 rather than 0 to find a course, so if you have only 1 course it will also give an issue as you will access a course that doesn't exist.

### What is the new behavior?
Now that bug should be patched!


musaabimran (approved):
Works Perfectly!

PR #9675, opened 2023-08-16
[Feature:Developer] Script for lint testing

ziesski:
Currently, we have to type long commands to run php static analysis, php linting, javascript linting, css linting locally. 
Some developers copy/paste the command from website to run them. 

### What is the new behavior?
With this new update, devs can use submitty_test <option>
Similar to how we use submitty_install_stie, submitty_install.
submitty_test php_stan will run php static analysis
It also takes third argument, for instance, 
submitty_test php_stan --memory-limit=4G
submitty_test php_stan --generate-baseline
and all other options will work the same as if running php vendor/bin/phpstan analyze app public/index.php socket/index.php

submitty_test php_lint will run php linting.
There are options like php_lint_stan which will run both, or all (javascript, css as well)
and help, which displays the commands. 

May further work on including python linting.


bmcutler:
@ziesski 
Please also make a PR to documentation to add instructions for these shortcuts

.setup/SUBMITTY_TEST.sh
```diff
@@ -0,0 +1,51 @@
...
+        php vendor/bin/phpstan analyze app public/index.php socket/index.php "${@:2}"
+    elif [ "$1" == "js_lint" ]; then
+        npm run eslint
+    elif [ "$1" == "css_lint" ]; then
+        npm run css-stylelint
+    elif [ "$1" == "all" ]; then
+        php vendor/bin/phpstan analyze app public/index.php socket/index.php
+        php vendor/bin/phpcs --extensions=php ./app
+        npm run eslint
+        npm run css-stylelint
```
williamjallen, line 29: Why are we duplicating code here?

.setup/SUBMITTY_TEST.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+
+current_dir=$(pwd)
+
+cd /usr/local/submitty/GIT_CHECKOUT/Submitty/site || {
```
williamjallen, line 6: Consider using `pushd`/`popd` instead.  See [here](https://github.com/Submitty/Lichen/blob/7e7d22975cb9b4c62a26fa8255002969974bafb4/install_lichen.sh#L57) for an example.

.setup/SUBMITTY_TEST.sh
```diff
@@ -0,0 +1,51 @@
...
+    echo "Failed to change to the Submitty/site directory.
+    Please check if you have /usr/local/submitty/GIT_CHECKOUT/Submitty/site and valid permission"
+    exit 1
+}
+
+
+submitty_test() {
```
williamjallen, line 13: Why are we creating a function and then calling it immediately afterwards?

williamjallen (changes requested):
I think this is a good feature in general, but it needs some more refinement.

I think we should do some more thinking about the arguments to `submitty_test` in particular.  What about using these instead?
- `submitty_test php` (runs PHPStan)
- `submitty_test php-style` (runs PHPCS)
- `submitty_test js` (runs eslint)
- `submitty_test css` (runs stylelint)
- `submitty_test bash` (runs shellcheck)

In all cases, I think we should pass any additional arguments through to the tool being used.

Why should we stop at linters though?  Should we add `php-unit`, `js-unit`, and `python-unit` commands as well?  (Cypress can't be easily run in this way, so users will have to run it on their host machine)

.setup/SUBMITTY_TEST.sh
ziesski: Great point, Thanks

ziesski:
> I think this is a good feature in general, but it needs some more refinement.
> 
> I think we should do some more thinking about the arguments to `submitty_test` in particular. What about using these instead?
> 
> * `submitty_test php` (runs PHPStan)
> * `submitty_test php-style` (runs PHPCS)
> * `submitty_test js` (runs eslint)
> * `submitty_test css` (runs stylelint)
> * `submitty_test bash` (runs shellcheck)
> 
> In all cases, I think we should pass any additional arguments through to the tool being used.
> 
> Why should we stop at linters though? Should we add `php-unit`, `js-unit`, and `python-unit` commands as well? (Cypress can't be easily run in this way, so users will have to run it on their host machine)

I agree we should also implementing php-unit, js-unit etc.. Wouldn't the argument php-lint, php-stan naming makes more sense as it draws clear distinction.

ziesski:
Additionally, lets add every check possible, and allow user to run one single command in vagrant (ie. submitty_test all)

ziesski:
Since there are some issues with installing eslint jslint in vagrant, lets first finish up php part, which works fine, as we will have incoming devs, and to complete things in small pieces.

ziesski:
@williamjallen 
One additional reason for naming the tests php-lint, css-lint, js-lint would be how our CI is currently labeled.

Screen Shot 2023-08-31 at 5 30 12 PM

As devs will be familiar with the Github CI naming, I still think we should coincide the submitty_test with it.
Hence, I suggest that submitty_test php-lint to test both php stan, php cs, just like php-lint in Github CI runs both.
We can have submitty_test php to test only php stan, and php-style to test php cs.

williamjallen:
@ziesski I think that sounds reasonable.  If the command options share a name with a a CI job, that makes things easier to remember.

.setup/SUBMITTY_TEST.sh
```diff
@@ -0,0 +1,41 @@
...
+run_php_stan() {
+    php vendor/bin/phpstan analyze app public/index.php socket/index.php "${@:2}"
+}
+
+run_php_cs() {
+    php vendor/bin/phpcs --extensions=php ./app
+}
+
+submitty_test() {
+    if [ "$1" == "php-stan" ]; then
```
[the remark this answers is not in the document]
ziesski: Done :)

ziesski (commented):
https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md

Used this as reference for installing composer

ziesski:
For whomever testing, please note that alias only works after fresh vagrant up with this PR branch. 
If you do not wish to vagrant up, try '/usr/local/submitty/.setup/SUBMITTY_TEST.sh php-lint' 
If you do have alias you may use 'submitty_test php-lint'

site/composer.json
```diff
@@ -52,7 +52,7 @@
   },
   "scripts": {
     "test": "phpunit",
-    "lint": "phpcs --extensions=php ./app",
-    "static-analysis": "phpstan analyze app public/index.php socket/index.php"
+    "lint": "php vendor/bin/phpcs --extensions=php ./app",
+    "static-analysis": "php vendor/bin/phpstan analyze app public/index.php socket/index.php"
```
williamjallen, line 56: Should these changes be reverted in favor of the changes in https://github.com/Submitty/Submitty/pull/9902?

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -653,6 +653,12 @@ chmod 700       "${SUBMITTY_INSTALL_DIR}/.setup/bin/reupload"*
 chown root:root "${SUBMITTY_INSTALL_DIR}/.setup/bin/track_git_version.py"
 chmod 700       "${SUBMITTY_INSTALL_DIR}/.setup/bin/track_git_version.py"
 
+###############################################
+# submitty_test script
+cp  "${SUBMITTY_REPOSITORY}/.setup/SUBMITTY_TEST.sh"        "${SUBMITTY_INSTALL_DIR}/.setup/"
+chown root:root "${SUBMITTY_INSTALL_DIR}/.setup/SUBMITTY_TEST.sh"
+chmod 700       "${SUBMITTY_INSTALL_DIR}/.setup/SUBMITTY_TEST.sh"
+
```
williamjallen, line 661: Is this necessary?  I'm not familiar enough with the setup scripts and permissions structure to comment with any certainty, but I think it would be good for @bmcutler to take a look before this gets merged.

.setup/SUBMITTY_TEST.sh
williamjallen, line 1: Are you also planning to add our Python and JS linters to this script?  Both should be relatively straightforward I think.

.setup/SUBMITTY_TEST.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+
+pushd /usr/local/submitty/GIT_CHECKOUT/Submitty/site  > /dev/null || {
+    echo "Failed to change to the Submitty/site directory.
+    Please check if you have /usr/local/submitty/GIT_CHECKOUT/Submitty/site and valid permission"
```
williamjallen, line 6: What about this instead?
```suggestion
    echo "Failed to change to /usr/local/submitty/GIT_CHECKOUT/Submitty/site."
```

.setup/SUBMITTY_TEST.sh
```diff
@@ -0,0 +1,46 @@
...
+        echo "Unknown test type: $1
+            use phpstan, phpcs, php-lint
+            or help for detail"
+    fi
+}
+
+submitty_test "$@"
+
+popd > /dev/null || {
+    echo "Failed to return to the previous directory. Check if you have valid permission."
```
williamjallen, line 44: This is a suggestion.  Feel free to take it or leave it as you see fit.
```suggestion
    echo "Failed to return to the previous directory."
```

.setup/SUBMITTY_TEST.sh
williamjallen: I'm still a bit confused about this.  Can't you just move all of this logic outside the function?

.setup/INSTALL_SUBMITTY_HELPER.sh
ziesski: I would say it is a good practice, to make sure script has executable bit. Without chmod, it won't. 
Although github now tracks the executable bit, it doesn't work for our case.

site/composer.json
ziesski: As I mentioned in #9902, this is, in fact, the correct way. I admit I got somewhat unorganized. I will close that PR.

.setup/SUBMITTY_TEST.sh
ziesski: Yea, I was just trying to make it neat, because initially I got confused by checking extra flags/arguments with "$@"
If you don't like it I can make the change.

.setup/SUBMITTY_TEST.sh
ziesski: I remember myself adding js and css, then faced some issue with npm, I will try to work on it this week. If it is simple enough I will try to add to this PR, if not new PR. (so that I have at least sth to say on friday)

.setup/INSTALL_SUBMITTY_HELPER.sh
williamjallen: Yes, but should we be installing this script on all systems?

.setup/SUBMITTY_TEST.sh
williamjallen: Seems reasonable to me.

.setup/SUBMITTY_TEST.sh
williamjallen: Can't you still do that without using the function?

site/composer.json
williamjallen: Sounds good to me.  You can still remove the files to be analyzed since https://github.com/Submitty/Submitty/pull/9900 has been merged though.

.setup/SUBMITTY_TEST.sh
ziesski: Yup, made the changes :)

.setup/INSTALL_SUBMITTY_HELPER.sh
ziesski: I guess we should have it installed, and it is up to devs to use it or not.

ziesski: I guess the concern is in the production server right?

site/composer.json
ziesski: done!

.setup/INSTALL_SUBMITTY_HELPER.sh
williamjallen: Yep.

williamjallen (approved):
Looks good to me.  We can add additional features in future PRs.

PR #9689, opened 2023-08-20
Grading Stats

musaabimran:
"Number of unresolved grade inquiries per grader" should be displayed only when there are grade inquiries greater than 0
image


NiaHeermance (approved):
Works!

PR #9696, opened 2023-08-20
[UI/UX:Autograding] Autograding Table Display

musaabimran:
The auto grading table had display issue as the lines were not clear.
image

### What is the new behavior?
Now there is no display issue.
image


NiaHeermance (approved):
Works!

bmcutler:
I prefer the original design.

PR #9699, opened 2023-08-21
[UI/UX:InstructorUI] Hide Header for Component Inquiry

NiaHeermance:
### What is the new behavior?
Extending PR #9689, now the "Number of unresolved grade inquiries" per component headers hide when that specific component has no active grade inquiries.

### Images
image

image


musaabimran (approved):
Work fine!

PR #9707, opened 2023-08-22
[Testing:Forum] Markdown Example in Forum

mahi-pas:

No forum post that shows all features of markdown. An example post will help with testing and also show devs what features are available

### What is the new behavior?
Added a forum post with markdown examples

### Other information?

Make sure that every type of markdown command is shown in this markdown post


NiaHeermance (approved):
Good addition! I do notice that the markdown tags are of course going to be printed as raw text on the side bar, but that seems okay.

image

mahi-pas:
> Good addition! I do notice that the markdown tags are of course going to be printed as raw text on the side bar, but that seems okay.
> 
> image

I might as well do the quick fix to this in this PR

PR #9729, opened 2023-08-28
[Dependency] Add Vite

NiaHeermance:
We currently do not have any way to create client-side rendered templates.

### What is the new behavior?
Vite is now installed for rendering Vue templates.


.setup/install_submitty/install_site.sh
```diff
@@ -38,11 +38,11 @@ set_permissions () {
     esac
 }
 
-set_mjs_permission () {
```
shailpatels, line 41: hey just a heads up the `mjs` was for javascript modules which are distinct from normal JS (the stuff you see in script tags usually embedded inside of twig templates) 

Modules live in the `site/ts` folder and I believe they get transpiled to the a directory called `mjs` if they're typescript

PR #9737, opened 2023-08-29
[Testing:Developer] Manual parallelizing in Github Actions

IDzyre:

Currently, Cypress Cloud parallelizes the runs, however this only works for Submitty members, not external contributors. It also has a limit of 100,000 different test runs (Specific `It('should test...', ()=>{});` runs, not spec files or describe instances). With increasing development, and the creation of new spec files that add to the tests, this will most likely not be enought. 

### What is the new behavior?
The runs are now split up into _five_ different categories manually, and Cypress Cloud has been removed from CI:

- Admin -- Testing admin features, or website options
- Feature -- Testing user features, polls, forums, late day caching, etc.
- Gradeable -- Testing creation, editing, submission of gradeables
- System -- Testing things that need other system functions to run, and accessibility testing. 
- UI -- Testing misc user interfaces


jason490 (changes requested):
You can try using the github matrix to get rid of doing similar operations
ex:
```
Cypress-Test:
    ...
   strategy:
      matrix:
           containers: ["gradeable","admin","UI",etc...]
   ...
   ...
         spec: cypress/e2e/${{matrix.containers}}
```

jason490 (approved):

.github/workflows/submitty_ci.yml
```diff
@@ -468,11 +474,9 @@ jobs:
         with:
           config: baseUrl=http://localhost
           working-directory: ${{env.SUBMITTY_REPOSITORY}}/site
-          spec: cypress/e2e/cypress-feature/*
+          spec: cypress/e2e/Cypress-${{matrix.containers}}/*
           browser: chrome
           env: run_area=CI
-          record: true
-          parallel: true  # Runs test in parallel using settings above
```
williamjallen, line 475: (Leaving this here since I can't comment on lines you didn't edit)

Are the env variables under this necessary now?

williamjallen (commented):
Changes seem reasonable to me.  I do worry about needing to rebalance the sets of tests occasionally, but that's certainly manageable, and better than the alternative of just running all of the tests sequentially.

Is there any parallelization you can do in one or more of the test sets?  We'd have to make sure none of the tests potentially interfere with one another and the benefits of parallelization would be hampered by the limited resources available on the CI runners, but it may reduce the time slightly at least.

.github/workflows/submitty_ci.yml
IDzyre: No they are not, I will remove them tonight.

IDzyre:
> Changes seem reasonable to me. I do worry about needing to rebalance the sets of tests occasionally, but that's certainly manageable, and better than the alternative of just running all of the tests sequentially.
> 
> Is there any parallelization you can do in one or more of the test sets? We'd have to make sure none of the tests potentially interfere with one another and the benefits of parallelization would be hampered by the limited resources available on the CI runners, but it may reduce the time slightly at least.

I think its probably possible, but to make the CI run more smoothly as it is, I think that can be a future PR.

williamjallen (approved):
Good work on this!

jason490 (approved):

PR #9739, opened 2023-08-30
[Bugfix:HelpQueue] Fix Queue Creation

NiaHeermance:
When someone attempts to create a queue without a password, a server error is produced.

### What is the new behavior?
The server error is gone! You can now create queues without passwords as usual.

### DIFF
--- a/site/app/controllers/OfficeHoursQueueController.php
+++ b/site/app/controllers/OfficeHoursQueueController.php
@@ -59,7 +59,10 @@ class OfficeHoursQueueController extends AbstractController {
         if ($token !== null) {
             preg_match_all($re, $token, $matches_token, PREG_SET_ORDER, 0);
         }
-        if (count($matches_code) !== 1 || count($matches_token) !== 1) {
+        if (
+            is_array($matches_code) && count($matches_code) !== 1
+            ||  is_array($matches_token) && count($matches_token) !== 1
+        ) {
             $this->core->addErrorMessage('Queue name and secret code must only contain letters, numbers, spaces, "_", and "-"');
             return MultiResponse::RedirectOnlyResponse(
                 new RedirectResponse($this->core->buildCourseUrl(['office_hours_queue']))


site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -59,7 +59,10 @@ public function openQueue() {
         if ($token !== null) {
             preg_match_all($re, $token, $matches_token, PREG_SET_ORDER, 0);
         }
-        if (count($matches_code) !== 1 || count($matches_token) !== 1) {
+        if (
+            is_array($matches_code) && count($matches_code) !== 1
+            || is_array($matches_token) && count($matches_token) !== 1
+        ) {
```
cjreed121, line 65: We could probably just initialize these variables to empty arrays (before the preg_match_all call) and then we can always check the count without issue.

cjreed121 (changes requested):

site/app/controllers/OfficeHoursQueueController.php
NiaHeermance: Good idea. Fixed!

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -55,11 +55,11 @@ public function openQueue() {
 
         $re = '/^[\sa-zA-Z0-9_\-]+$/m';
         preg_match_all($re, $queue_code, $matches_code, PREG_SET_ORDER, 0);
-        $matches_token = 1;
+        $matches_token = [];
```
cjreed121, line 58: Do this for matches_code too

site/app/controllers/OfficeHoursQueueController.php
```diff
@@ -55,11 +55,11 @@ public function openQueue() {
...
         if ($token !== null) {
             preg_match_all($re, $token, $matches_token, PREG_SET_ORDER, 0);
         }
-        if (count($matches_code) !== 1 || count($matches_token) !== 1) {
+        if (count($matches_code) !== 1 && count($matches_token) !== 1) {
```
cjreed121, line 62: I don't think this should have changed?

cjreed121 (changes requested):

site/app/controllers/OfficeHoursQueueController.php
NiaHeermance: Sure.  Matches code wasn't the one causing the crash, but better safe than sorry.

site/app/controllers/OfficeHoursQueueController.php
NiaHeermance: Yeah, that was a typo and I spot it just before you commented on it, ha ha.

NiaHeermance:
Hmm, with the refactor the queue creation, while no sever error, still isn't successful. Let me fix it.

NiaHeermance:
Fixed!

PR #9752, opened 2023-09-02
[Bugfix:InstructorUI] Fixed Improper use of Date strings in JS

mahinigam:

#### Currently in our JavaScript code, there are numerous calls to the Date constructor which pass in a date string.

Ex.) new Date("02/01/1970 05:00") => new Date("1970-02-01T05:00:00")


site/public/js/latedays.js
```diff
@@ -37,7 +37,7 @@ $(document).ready(() => {
                             date = new Date();
                             break;
                         case 1:
-                            date = new Date('9998-01-01');
+                            date = new Date("9998-01-01");
```
williamjallen, line 40: This change isn't necessary and should be reverted.

williamjallen (changes requested):

site/public/js/latedays.js
mahinigam-28: done

ziesski (commented):
Please resolve the javascript lint error (double quotation->single quotation)

VVVITH (approved):
After I fix the branch conflict, the code looks good now. So I approve it, but I'm not sure is this change a good change and make real improve? More reviews needed.

cjreed121 (approved):

PR #9753, opened 2023-09-04
[Refactor:System] Organizing setup_sample_courses.py

jason490:

There is approximately 2250 lines in setup_sample_courses.py, which is inconvenient to attach more/new functionality too and harder to troubleshoot 
### What is the new behavior?
The file has been split up in python packages and files that is more manageable


williamjallen (commented):
Now that you have split `setup_sample_courses.py` into smaller pieces, is it possible to enable flake8 linting for some of the files with minimal edits?

jason490:
It seems like this one test keeps on failing on GitHub actions, but I just tried running this pr with cypress running locally and sometimes it would work and sometimes not. This also happens in main, when simple_grading.js will pass the test and sometimes not.

.setup/bin/sample_courses/__init__.py
```diff
@@ -26,20 +26,21 @@ def parse_args():
     parser.add_argument("--courses_path", default=os.path.join(SETUP_DATA_PATH, "courses"),
                         help="Path to the folder that contains .yml files to use for course creation. Defaults to "
                              "../data/courses")
-    parser.add_argument("--install_dir", type=str, default=SUBMITTY_INSTALL_DIR, help="install path of submitty")
-    parser.add_argument("--data_dir", type=str, default=SUBMITTY_DATA_DIR, help="data path of submitty")
```
jason490: Changed the default to what the default value of the variable was because this function was moved to __init__.py, to make it easier to see how each variable is being set

.setup/bin/sample_courses/main.py
```diff
@@ -51,61 +51,20 @@
...
-
-    args = parse_args()
-    DB_ONLY = args.db_only
-    NO_SUBMISSIONS = args.no_submissions
-    NO_GRADING = args.no_grading
-    SUBMITTY_INSTALL_DIR = args.install_dir
-    SUBMITTY_DATA_DIR = args.data_dir
-    SUBMITTY_REPOSITORY = os.path.join(SUBMITTY_INSTALL_DIR, "GIT_CHECKOUT/Submitty")
-    MORE_EXAMPLES_DIR = os.path.join(SUBMITTY_INSTALL_DIR, "more_autograding_examples")
-    TUTORIAL_DIR = os.path.join(SUBMITTY_INSTALL_DIR, "GIT_CHECKOUT/Tutorial", "examples")
```
jason490: Condensed the redeclaring of variables and placed it into __init__
diff:
[](https://github.com/Submitty/Submitty/pull/9753/commits/2745aa6e3a28c7eaba685e88f444b145c361344f)

.setup/bin/sample_courses/__init__.py
```diff
@@ -26,20 +26,21 @@ def parse_args():
...
     return parser.parse_args()
 
 # Start of global variables
-args = parse_args()
+
 yaml = YAML(typ='safe')
 
 CURRENT_PATH = os.path.dirname(os.path.realpath(__file__))
-SETUP_DATA_PATH = os.path.join(CURRENT_PATH, "..", "data")
+SETUP_DATA_PATH = os.path.join(CURRENT_PATH, "..", "..", "data")
```
jason490: I had to add extra /../ to get of the current directory

.setup/bin/sample_courses/__init__.py
```diff
@@ -42,17 +55,34 @@ def parse_args():
...
+        raise SystemError("The following directory does not exist: " + os.path.join(
+            SUBMITTY_DATA_DIR, directory))
+    
+with open(os.path.join(SUBMITTY_INSTALL_DIR, "config", "database.json")) as database_config:
+    database_config_json = json.load(database_config)
+    
+    DB_USER = database_config_json["database_user"]
+    DB_HOST = database_config_json["database_host"]
+    DB_PORT = database_config_json["database_port"]
+    DB_PASS = database_config_json["database_password"]
```
jason490, line 86: I had to move this from main.,py to __init__.py because this was changing the ports value and if you use from filename import variable, the variables sort of become a constant variable that can't be changed
Diff:
https://github.com/Submitty/Submitty/pull/9753/commits/2635f7e0a27d93154b00d74f8b1fdfd448821447

.setup/bin/sample_courses/models/course/__init__.py
jason490, line 1: Had to split the function create because the function itself was around 700 lines
Diff:
https://github.com/Submitty/Submitty/pull/9753/commits/405ad3ac5f8737d4ec23ca4d3efb3bdb082890c4

.setup/bin/sample_courses/models/course/course_generate_utils.py
```diff
@@ -0,0 +1,269 @@
...
+
+from sqlalchemy import Table, select, func
+
+from sample_courses import *
+from sample_courses.utils import get_current_semester
+from sample_courses.utils.create_or_generate import generate_random_user_id
+from sample_courses.models.course.course_utils import Course_utils
+from sample_courses.models.course.course_data import Course_data
+
+class Course_generate_utils(Course_utils,Course_data):
```
jason490, line 17: For some reason I don't see the diff, but compare it with the lower comment

.setup/bin/sample_courses/models/course/course_create_gradeables.py
```diff
@@ -0,0 +1,400 @@
...
+    generate_random_ta_comment,
+    generate_versions_to_submit,
+    generate_random_user_id,
+    create_gradeable_submission,
+    create_pdf_annotations
+    )
+from sample_courses.models.course.global_var import Table_var
+
+
+class Course_create_gradeables:
```
jason490, line 30: I removed the inheritance from the class Course_create_gradeables because I'm already inheriting all the classes in the course class and if I inherit classes that are already inherited, I believe it loads double the classes. Also, I don't currently know what functions that are being called through the course class.

.setup/bin/sample_courses/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+                        help="course code to build. If no courses are passed in, then it'll use "
+                             "all courses in courses.json")
+    return parser.parse_args()
+
+# Start of global variables
+
+yaml = YAML(typ='safe')
+
+CURRENT_PATH = os.path.dirname(os.path.realpath(__file__))
+SETUP_DATA_PATH = os.path.join(CURRENT_PATH, "..", "..", "data")
```
IDzyre, line 56: I don't know if this would do anything, but would it be more consistant with the rest of the file to declare this _after_ the args are parsed, and then use and `SETUP_DATA_PATH = os.path.join(SUBMITTY_REPOSITORY, '.setup/data')`

jason490: The funny thing is we can't because the `parse_args` function relies on the SETUP_DATA_PATH variable for the default path.
Also, don't really think we need to really use `os.path.dirname(os.path.realpath(__file__))`, but its doing its job well for now because `SETUP_DATA_PATH = "/usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/data"` is the absolute directory in the VM.
In this PR, my goal is to just split up this file into readable parts, then create another PR to then clean the code up, and lastly make another PR to make it more efficient

.setup/bin/sample_courses/models/course/__init__.py
jason490, line 1: I felt the class course should be in __init__.py because all it does currently is initialize all the global vars inside the classes.
I can't make the global vars into a file without having a bigger change because the course data is being passed into the class

.setup/bin/sample_courses/models/course/course_create_gradeables.py
```diff
@@ -0,0 +1,415 @@
...
+    def __init__(self):
+        # Anything that needs to be initialized goes here
+        pass
+
+    def add_gradeables(self):
+        anon_ids = {}
+        for gradeable in self.gradeables:
+            #create gradeable specific anonymous ids for users
+            prev_state = random.getstate()
+            for user in self.users:
```
jason490, line 56: In a future PR, I will split this function into parts and move them into more functions and eventually into classes to become more maintainable.

.setup/bin/sample_courses/models/course/course_generate_utils.py
```diff
@@ -0,0 +1,269 @@
+"""
+Contains Create_Course class that inheits Course_utils and Course_data class
```
IDzyre, line 2:
```suggestion
Contains Create_Course class that inherits Course_utils and Course_data class
```

.setup/bin/sample_courses/__init__.py
```diff
@@ -0,0 +1,101 @@
+"""
+Contains all the global constant variables that are used
+throughout the sample courses.
+None of the variables should be changed after they are set here.
+If you ever want to change the global variables in another file,
+you have to use import sample_courses and not from sample_courses import *,
+but its highly unrecommended.
+
+To prevent cicurlar imports, don't import any files in this package and if
```
IDzyre, line 9:
```suggestion
To prevent circular imports, don't import any files in this package and if
```

.setup/bin/sample_courses/models/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        print("Creating user {}...".format(self.id))
+        os.system(
+            "useradd -m -c 'First Last,RoomNumber,WorkPhone,HomePhone' {}".format(
+                self.id
+            )
+        )
+        self.set_password()
+
+    def create_non_ssh(self):
+        print("Creating user {}...".format(self.id))
```
IDzyre, line 162: Would using f-strings like in line 170 be better than .format(variable) in making the files consistent? 

This could also be in a future PR...

jason490: Yep, I will change it in the next PR, hopefully after this PR gets merged

.flake8
```diff
@@ -9,6 +9,7 @@ exclude=
     .setup/bin/reupload_generate_csv.py,
     .setup/bin/reupload_old_assignments.py,
     .setup/bin/setup_sample_courses.py,
+    .setup/bin/sample_courses/models,
```
williamjallen, line 12: Let's ignore all of the individual files rather than an entire directory.  That way, if someone adds a new file, it will get linted.

.flake8
```diff
@@ -9,6 +9,7 @@ exclude=
     .setup/bin/reupload_generate_csv.py,
     .setup/bin/reupload_old_assignments.py,
     .setup/bin/setup_sample_courses.py,
```
williamjallen, line 11: This file has been deleted, so it can be removed...

williamjallen (changes requested):
A couple quick linting changes, and then I'll click approve.

IDzyre:
I think that this is a good start, it breaks it up into readable and editable pieces.  Once William approves, it is good. There are some things that could change, but the for the scope of this PR, I think this is sufficient

williamjallen (approved):

PR #9760, opened 2023-09-07
[UI/UX:Calendar] Calendar uses course display name

yanliw123:
1. go to course setting, add/change a name for current course.
image

2. go to calendar. Calendar should show display name(if there is one) instead of course name/id. 

image
current behavior is calendar will only show course name/id in format of [course_title course _term]
### What is the new behavior?
now the calendar will show course display name if there is one in format of [course_displayname course_term] and if the course does not have a display name it will show the course in old format [course_title course _term]


site/app/views/calendar/CalendarView.php
```diff
@@ -86,7 +100,7 @@ public function showCalendar(CalendarInfo $info, array $courses, bool $in_course
             "colors" => $info->getColors(),
             "instructor_courses" => $this->core->getQueries()->getInstructorLevelUnarchivedCourses($this->core->getUser()->getId()),
             "view_cookie" => isset($_COOKIE['view']) ? $_COOKIE['view'] : "month",
-            "course_names" => $formatted_courses,
+            "course_names" => $combinedArray,
```
ziesski, line 103: Can we have better variable name choice, more descriptive.

site/app/views/calendar/CalendarView.php
```diff
@@ -40,11 +40,25 @@ public function showCalendar(CalendarInfo $info, array $courses, bool $in_course
...
+            }
+            else {
+                $name_string = sprintf("%s %s", $course->getTitle(), $course->getTerm());
+            }
+            array_push($display_names, $name_string);
+        }
+        $combinedArray = [];
+        foreach (array_map(null, $formatted_courses, $display_names) as [$course, $name]) {
+            $combinedArray[] = ['course' => $course, 'name' => $name];
+        }
```
ziesski, line 61:
```suggestion
$combinedArray = [];

foreach ($courses as $course) {
    $displayName = $course->getDisplayName();
    $title = $course->getTitle();
    $term = $course->getTerm();
    
    $name_string = ($displayName ? $displayName : $title) . ' ' . $term;
    
    $combinedArray[] = ['course' => $course, 'name' => $name_string];
}

```

ziesski (changes requested):

NiaHeermance (changes requested):
I agree that ``combinedArray`` should be called something else, perhaps ``courseWithName``. Otherwise looks good!

site/app/views/calendar/CalendarView.php
yanliw123: I refactored this part of the code. In the Combinearray( renamed to courseWithName) we need to use the formatted_courses instead of course.

site/app/views/calendar/CalendarView.php
```diff
@@ -40,11 +40,20 @@ public function showCalendar(CalendarInfo $info, array $courses, bool $in_course
...
         //$course_colors = [];
         foreach ($courses as $course) {
             $course_string = sprintf("%s %s", $course->getTitle(), $course->getTerm());
             array_push($formatted_courses, $course_string);
         }
+        $courseWithName = [];
+        foreach ($courses as $index => $course) {
+            $nameString = $course->getDisplayName() ?
+                        sprintf("%s %s", $course->getDisplayName(), $course->getTerm()) :
+                        sprintf("%s %s", $course->getTitle(), $course->getTerm());
```
williamjallen, line 53: Why is `sprintf` being used here?  Can't you just do this with regular string concatenation?

site/app/views/calendar/CalendarView.php
```diff
@@ -40,11 +40,20 @@ public function showCalendar(CalendarInfo $info, array $courses, bool $in_course
...
+
+            $courseWithName[] = ['course' => $formatted_courses[$index], 'name' => $nameString];
```
williamjallen, line 55: It would be better to break this into multiple lines so future diffs are easier to read.

```suggestion
            $courseWithName[] = [
                'course' => $formatted_courses[$index],
                'name' => $nameString,
            ];
```

williamjallen (changes requested):

site/app/views/calendar/CalendarView.php
yanliw123: I used sprintf here because above code used sprintf. So i tried to keep the code in same format. But I changed them all to regular concatenation

yanliw123: > Why is `sprintf` being used here? Can't you just do this with regular string concatenation?
I used sprintf here because above code used sprintf. So i tried to keep the code in same format. But I changed them all to regular concatenation

site/app/views/calendar/CalendarView.php
```diff
@@ -40,11 +40,24 @@ public function showCalendar(CalendarInfo $info, array $courses, bool $in_course
...
         foreach ($courses as $course) {
-            $course_string = sprintf("%s %s", $course->getTitle(), $course->getTerm());
+            $course_string = $course->getTitle() . " " . $course->getTerm();
             array_push($formatted_courses, $course_string);
         }
+        $courseWithName = [];
+        foreach ($courses as $index => $course) {
+            $nameString = $course->getDisplayName() ?
+                        ($course->getDisplayName() . " " . $course->getTerm()) :
+                        ($course->getTitle() . " " . $course->getTerm());
```
williamjallen, line 53: An even cleaner solution you could consider:
```suggestion
            $displayName = $course->getDisplayName()
            $nameString = ($displayName !== '' ? $displayName : $course->getTitle()) . $course->getTerm();
```

site/app/templates/calendar/CalendarOptions.twig
```diff
@@ -7,17 +7,17 @@
         {% if course_names|length > 1 %}
         <option value="show all">All Courses</option>
         {% endif %}
-        {% for course in course_names %}
-        <option value="{{course}}">{{course}}</option>
+        {% for item in course_names %}
+        <option value="{{item.course}}">{{item.name}}</option>
```
williamjallen, line 11: `item` isn't a very descriptive identifier.  The original `course` was much more descriptive.  Can you please choose a better identifier?

site/app/templates/calendar/CalendarOptions.twig
```diff
@@ -7,17 +7,17 @@
...
         {% endfor %}
     </select>
     <hr class="line-break">
     <h2 class="cal-options-section-head">Colors</h2>
     <div class="options-colors">
-        {% for course in course_names %}
+        {% for item in course_names %}
```
williamjallen, line 17: Same here.

williamjallen (changes requested):

NiaHeermance (commented):
Why does your screenshot not have the f23 in it?

image

yanliw123:
> Why does your screenshot not have the f23 in it?
> 
> image

that was the old screenshot, I will update it with a new one

NiaHeermance (approved):
Looks great!

taorif25 (approved):

PR #9773, opened 2023-09-11
[Bugfix:Submission] Remove redundant late day cache call

Youssef-Chip:

There is an unnecessary call to generate the late day cache in LateDays.php. This is causing duplicate keys in the late day cache.
### What is the new behavior?
Remove unnecessary and redundant call

### DIFF
--- a/site/app/models/gradeable/LateDays.php
+++ b/site/app/models/gradeable/LateDays.php
@@ -43,24 +43,10 @@ class LateDays extends AbstractModel {
         $graded_gradeables = array_filter($graded_gradeables, function (GradedGradeable $gg) {
             return $gg->getGradeable()->getType() === GradeableType::ELECTRONIC_FILE;
         });
-        $ggs = [];
-        foreach ($graded_gradeables as $gg) {
-            $ggs[$gg->getGradeableId()] = $gg;
-        }
+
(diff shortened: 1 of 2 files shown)


site/app/models/gradeable/LateDays.php
```diff
@@ -43,24 +43,10 @@ public function __construct(Core $core, User $user, array $graded_gradeables, $l
...
-        $ggs = [];
-        foreach ($graded_gradeables as $gg) {
-            $ggs[$gg->getGradeableId()] = $gg;
-        }
+
         // Get the late day updates that the instructor will enter
         $this->late_days_updates = $late_day_updates ?? $this->core->getQueries()->getLateDayUpdates($user->getId());
-        $this->core->getQueries()->generateLateDayCacheForUser($user->getId());
         $late_day_cache = $this->core->getQueries()->getLateDayCacheForUser($user->getId());
-        // Construct late days info for each gradeable
```
ziesski, line 54: Don't we need to use getLateDayCacheForUserGradeable function instead to replace line 55-63, which calls generateLateDayCacheForUser function inside.

ziesski (approved):
Looks good, although cypress test needs a refactor to not check latedaycache, as we discussed. (should be a seperate PR)

PR #9781, opened 2023-09-15
[Testing:Developer] Add slack message on vagrant up failure

### LINKED ISSUE #9596: Add broader notification for `vagrant_up.yml` failure
Whenever the [`vagrant_up.yml` scheduled workflow](https://github.com/Submitty/Submitty/blob/main/.github/workflows/vagrant_up.yaml) fails, it sends a notification to the person who  last updated the schedule, which happens to be me. If someone else updates the schedule, then they'll be the only person who gets a notification this workflow is failing. This isn't super useful as it creates a bus factor of 1 on knowing the status of this workflow, and is easy to miss/ignore.

Instead, should send out a broader notification if the workflow fails (use the `if: ${{ failure() }}` syntax for a step to have it run only on failure) so that more of the team knows this is happening and be encouraged to fix it. Two options for type of notification would be to create a GH issue or to send a slack message. I would lean towards the latter as it's easier to gain visibility vs a GH issue.

IDzyre:

Currently the vagrant up github action is a cron job, so it is run on a schedule. Due to this, it only sends notifications to the last person to edit the workflow file. If this person is away for any significant period, the failing job can go unnoticed by other developers. This is a problem, as new developers won't be able to reliably run 'vagrant up', which will slow development. 

### What is the new behavior?
This adds a notification VIA a slack app that notifies all who are in the slack channel of the vagrant up failure. 
image

The first two messages were debugging, and the last one was a mock failure to show it works. 

Closes #9596

### DIFF
--- a/.github/workflows/vagrant_up.yaml
+++ b/.github/workflows/vagrant_up.yaml
@@ -23,6 +23,7 @@ jobs:
       - name: Validate image
         run: curl --show-error --fail --include http://localhost:1511
       - name: Acquire Job ID
+        if: failure()
         id: get-job-id
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/vagrant_up.yaml
+++ b/.github/workflows/vagrant_up.yaml
@@ -34,4 +34,4 @@ jobs:
       - name: Send slack message on failure
         if: failure()
         run: |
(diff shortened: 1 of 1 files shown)


cjreed121 (commented):
This looks good to me. Could we however put the exact run log link instead? For example something like https://github.com/Submitty/Submitty/actions/runs/6216708666/job/16870883135. I think we should be able to fetch the job and run ID through variables.

IDzyre:
> This looks good to me. Could we however put the exact run log link instead? For example something like https://github.com/Submitty/Submitty/actions/runs/6216708666/job/16870883135. I think we should be able to fetch the job and run ID through variables.

It is possible, at the moment Github doesn't have a job-id environment variable that I could find easily, but it is possible to get it through the API though, so it has been added.

cjreed121 (approved):
Looks good! Also, your PR description seems to have gotten cut off in your main paragraph.

Issue #9738, opened 2023-08-30
Temporary override when student incorrectly loses registration access


bmcutler:
**What problem are you trying to solve with Submitty**
Approximately once per term we have corrupted datafeed.  Students are incorrectly omitted from the course enrollment, and an automated script removes the student from the course.  The student loses Submitty access and cannot obtain files, access the discussion forum, upload homework.  Students send email, and must wait for the instructor or sysadmin to manually add them back into the course -- but precious time is lost.

**Describe the way you'd like to solve this problem**

Proposal:

* If a student has accessed a course *recently* but they are currently in the NULL registration section of an active course, they can temporarily re-add themselves to the course.

* "recently accessed" = has database access information within the last 24 hours (or maybe 48 or 72 hours?)

* database access information = most recent date in all the categories on the "Student Activity Dashboard"

* Their list of Submitty courses on the main "My Courses" Submitty page lists those courses as "Recently Dropped".

* If they click on that course (or otherwise directly access a course URL), instead of getting this message:

    <img width="419" alt="Screenshot 2023-08-29 at 10 45 59 PM" src="https://github.com/Submitty/Submitty/assets/5871417/6acfbe1a-4e15-49c0-82b9-a01d6f33eda3">

    They will instead get a message that says something like "You were recently removed from this course.  If you believe this is an error, click here to re-add yourself to this course.  An email will be automatically sent to both your instructor(s) and the system administrators so they are alerted about this potential error."
    
* These instructor & system administrator emails will be sent through the normal Submitty email system.  I the student was indeed correctly removed from the course, then the instructor & system administrator can resolve the situation to permanently remove access if needed.

* If the daily registration scripts incorrectly remove the same students repeatedly, the student can keep re-enrolling themselves (and these emails will be repeatedly sent) until the situation is appropriately resolved.

* When we re-add a student to the course, we will need to know what that students "last valid registration section" was.  We should add an additional column to the database and fill it in appropriately.   Maybe this can just be stored in the course database?  (I don't think it's necessary to store it in the master database and clutter up the trigger function.)

* If this last value registration section data is not available, we should still allow the student to be re-added, just put them in the first registration section.    The email to the instructor and sysadmin should include the information about which section the student has been re-enrolled to (in case this is incorrect and the student needs to be moved).

Tagging @williamjallen @cjreed121 @NiaHeermance @pbailie for discussion.  

Also linking this to the feature request for "self-registration" courses #3950

bmcutler:
[...]

PR #9783, opened 2023-09-17
[Feature:System] Rejoin Previous Section When Removed

NiaHeermance:
PR #9809 follows this one. Continues addressing issue #9738.

When a student rejoins a course, they are readded to the course's first section.

### What is the new behavior?
Now they are added to the last section they were in.


cjreed121 (changes requested):

migration/migrator/migrations/master/20230916171141_update_last_nonnull_section_trigger.py
```diff
@@ -0,0 +1,48 @@
...
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    database.execute("""
+        ALTER TABLE courses_users
+        ADD COLUMN last_nonnull_registration_section VARCHAR;
```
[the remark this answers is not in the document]
NiaHeermance: Oh true. I'll fix that.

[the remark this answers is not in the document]
NiaHeermance: If a student is currently in the null section, then I'll let section 1 be the last nonnull section they were a part of even if that's not necessarily true as I have no other information to go off of. This only matters for the migration and not the future.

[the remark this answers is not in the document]
NiaHeermance: Done!

williamjallen:
Given that this PR is waiting for another one to be merged, should this be marked as a draft?

cjreed121 (changes requested):
This breaks adding users manually to a course when there is no registration section. Try running `adduser_course.py` on a course where there is no sections.

migration/migrator/migrations/course/20230921190909_update_last_nonnull_rotating_section_trigger.py
```diff
@@ -0,0 +1,76 @@
...
+            RETURN NEW;
+        END;
+        $$;
+
+        -- Attatch trigger function
+        CREATE TRIGGER before_update_users_update_last_nonnull_rotating_section
+        BEFORE UPDATE ON public.users
+        FOR EACH ROW EXECUTE PROCEDURE update_last_nonnull_rotating_section();
+
+        -- Set existing users' last nonnull rotating section.
```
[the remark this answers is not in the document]
NiaHeermance: I updated adduser_course.py and am now able to add users manually. I'll check out the sysadmin scripts to see if any changes need to be made there.

pbailie (commented):
Most importantly — so far this appears to be working OK with the current autofeed script.

One observation:  When a student is first inserted at the beginning of the term, their `previous_registration_section` is null, and remains null until their `registration_section` changes (either to a new section or is set to null presumably because they dropped their enrollment).

My understanding is that this is per spec.  But I am bringing this up to consider should `previous_registration_section` being null is desirable or not.  Or at least consider — can (should?) a student "rejoin" a null section?  (and this can be disregarded if this concern has already been resolved).

NiaHeermance:
> Most importantly — so far this appears to be working OK with the current autofeed script.
> 
> One observation: When a student is first inserted at the beginning of the term, their `previous_registration_section` is null, and remains null until their `registration_section` changes (either to a new section or is set to null presumably because they dropped their enrollment).
> 
> My understanding is that this is per spec. But I am bringing this up to consider should `previous_registration_section` be null is desirable or not. Or at least consider — can (should?) a student "rejoin" a null section? (and this can be disregarded if this concern has already been resolved).

Originally, that column was going to be called `previous_nonnull_registration_section` for that very concern. However, this name was confusing when adding students to a course before a nonnull section was added to the course. The good news is that `previous_registration_section` works just fine, as when a student is in the null section, their previous section was nonnull. And once a student is in a course, they are not allowed to "rejoin" a course they are already in, so they can't accidentally rejoin the null section. So your concern can be disregarded as you said.

migration/migrator/data/submitty_db.sql
```diff
@@ -416,6 +435,7 @@ CREATE TABLE public.courses_users (
     registration_section character varying(255),
     registration_type character varying(255) DEFAULT 'graded'::character varying,
     manual_registration boolean DEFAULT false,
+    previous_registration_section character varying,
```
cjreed121, line 438: Should this be character varying(255) as well if the source where it would come from is also 255?

migration/migrator/migrations/course/20230921190909_add_previous_rotating_section_trigger.py
```diff
@@ -0,0 +1,58 @@
...
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("""
+        ALTER TABLE users
+        DROP COLUMN previous_rotating_section;
```
cjreed121, line 53: Wouldn't we want to keep this?

migration/migrator/migrations/course/20230921190909_add_previous_rotating_section_trigger.py
```diff
@@ -0,0 +1,58 @@
...
+            LANGUAGE plpgsql
+            AS $$
+        BEGIN
+            RETURN NEW;
+        END;
+        $$;
+
+        -- Attatch trigger function
+        CREATE TRIGGER before_update_users_update_previous_rotating_section
+        BEFORE UPDATE ON public.users
```
cjreed121, line 33: This will only need to get run if the rotating section changes, right? If so, can we change it to `BEFORE UPDATE OF rotating_section ON public.users`?

migration/migrator/migrations/master/20230916171141_add_previous_registration_section_trigger.py
```diff
@@ -0,0 +1,50 @@
...
+            LANGUAGE plpgsql
+            AS $$
+        BEGIN
+            RETURN NEW;
+        END;
+        $$;
+
+        -- Attatch trigger function
+        CREATE TRIGGER before_update_courses_update_previous_registration_section
+        BEFORE UPDATE ON public.courses_users
```
cjreed121, line 29: Same as above

migration/migrator/migrations/master/20230916171141_add_previous_registration_section_trigger.py
```diff
@@ -0,0 +1,50 @@
...
+    Run down migration (rollback).
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    database.execute("""
+        ALTER TABLE courses_users
+        DROP COLUMN previous_registration_section;
```
cjreed121, line 45: Same as above

cjreed121 (changes requested):
Code looks good. Just some small requested changes.

migration/migrator/data/submitty_db.sql
NiaHeermance: Is 255 the character limit? Gotcha!

migration/migrator/migrations/course/20230921190909_add_previous_rotating_section_trigger.py
NiaHeermance: Good point.

cjreed121 (approved):
Code looks good to me

PR #9790, opened 2023-09-21
[Feature:Developer] Composer script

ziesski:
Our new plan is to make composer script for PHP stuff then bash script (#9675) will call this composer script.


ziesski:
Still trying to figure out for composer script to take additional argument

williamjallen (approved):
Looks good to me and matches the docs [here](https://submitty.org/developer/testing/linting_static_analysis).

PR #9793, opened 2023-09-21
[Dependency] update DrMemory 2.6.0

bmcutler:
https://drmemory.org/page_download.html
This update is necessary to use Ubuntu22 / g++11


NiaHeermance:
Where I can see the changes for the new DrMemory version? :)

bmcutler:
I don't know if there are formal release notes
https://drmemory.org/page_download.html

This update is necessary to use Ubuntu22 / g++11

PR #9797, opened 2023-09-22
[Testing:Forum] cypress test- display pronouns in forum

yanliw123:
some more cypress test for display pronouns in forum. 
1.create thread
2.verify pronouns is shown in overall forum page
3.comment on the thread, verify pronouns is shown
4.remove thread

also add code to change display pronouns option back.


NiaHeermance (changes requested):
Looks good! I would just also include logging in as another user and seeing Joe's pronouns are there.

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -36,16 +44,34 @@ describe('Tests cases abut changing user pronouns', () => {
...
+
         cy.get('button[aria-label="Clear pronoun input"]').click(); //clear input using trash can
         if (oldPronouns !== '') {
             cy.get('#user-pronouns-change').type(oldPronouns);
         }
-        cy.get('#edit-pronouns-submit').first().click();
 
-        //ensure pronouns changed on page
+        //change display option back
+        cy.get('#pronouns-forum-display').then(() => {
```
IDzyre, line 54: if this is only running with the student profile, would it be more concise to hard-code the resetting of the variables, because you should be able to see what the original values are ?

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -72,4 +97,67 @@ describe('Tests cases abut changing user pronouns', () => {
...
+
+        //comment on the thread, verify pronouns is shown
+        cy.get('.create-post-head').should('contain', 'Test pronouns display');
+        cy.get('#reply_box_2').type('my pronouns is They/Them{ctrl}{enter}');
+        cy.contains('Submit Reply to All').click();
+        cy.get('.post_box').should('contain', 'my pronouns is They/Them');
+        cy.get('.post_user_pronouns').should('contain', 'They/Them');
+
+        //remove thread
+        cy.get('.first_post > .post-action-container > .delete-post-button').click();
```
IDzyre, line 127: Is there a better selector for this?

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -72,4 +97,67 @@ describe('Tests cases abut changing user pronouns', () => {
...
+        cy.get('.create-post-head').should('contain', 'Test Anonymous thread, should not show pronouns');
+        cy.get('.thread-anon-checkbox').filter(':visible').click();
+        cy.get('#reply_box_2').type('I can not see your pronouns{ctrl}{enter}');
+        cy.contains('Submit Reply to All').click();
+        cy.get('.post_box').should('contain', 'I can not see your pronouns');
+        cy.get('.post_user_id').should('contain', 'Anonymous');
+        cy.get('.post_user_pronouns').should('not.exist');
+
+        //remove thread
+        cy.get('.first_post > .post-action-container > .delete-post-button').click();
```
IDzyre, line 158: same with better selector

IDzyre (changes requested):

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
yanliw123: yea, but the variables might be different or changed by different users. So In the future, if people run the cypress test locally. Their student account's variable will be changed to the original values I set, instead of their original value.

IDzyre: Will they be changing the student account pronoun for a reason on a dev VM? (besides to test something)

williamjallen: I think it makes sense to try to accept a system in any state, and then restore the system to the initial state when the test is done whenever possible.  That said, we do have a fair number of tests which rely upon the system being in a fresh state, so it's not terrible to do it that way either...

site/package.json
```diff
@@ -71,7 +71,8 @@
     "regenerator-runtime": "^0.14.0",
     "stylelint": "^15.10.3",
     "stylelint-config-standard": "^33.0.0",
-    "typescript": "^5.0.4",
+    "ts-node": "^10.9.1",
+    "typescript": "^5.2.2",
```
williamjallen, line 75: These changes are unrelated and should be reverted.

site/package-lock.json
williamjallen, line 1: These changes are also unrelated.

williamjallen (dismissed):

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -36,16 +30,18 @@ describe('Tests cases abut changing user pronouns', () => {
...
-        }
+
+        //set to false
+        cy.get('#pronouns-forum-display').uncheck();
+
+        cy.get('#edit-pronouns-submit').click();
+
+        //ensure pronouns and display option changed on page
+        cy.get('#pronouns_val').should('have.text', ' ');
+        cy.get('#display_pronouns_val').contains('False');
```
IDzyre, line 44: if you are asserting that an element contains a value, the most clear way is to do cy.get().should('contain', 'text_to_contain').

cy.get().contains('text_to_contain') will have the same output (it will fail if the element doesn't have the text), but it is for requesting the element, to then chain commands after it (like on line 78), instead of asserting that the element exists.

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -72,4 +67,79 @@ describe('Tests cases abut changing user pronouns', () => {
...
+        //verify pronouns exist and remove thread
+        cy.contains('Test pronouns display').find('.post_user_pronouns').should('contain','They/Them');
+        cy.contains('Test pronouns display').find('.post_user_pronouns').click();
+        cy.get('.post_user_pronouns').should('contain', 'They/Them');
+        cy.get('.first_post').find('.fa-trash').click();
+        cy.contains('Test pronouns display').should('not.exist');
+
+        //verify pronouns do not exist (post thread Anonymously) and remove thread
+        cy.contains('Test Anonymous thread, should not show pronouns').find('.post_user_id').should('contain','Anonymous');
+        cy.contains('Test Anonymous thread, should not show pronouns').find('.post_user_pronouns').should('not.be.exist');
```
IDzyre, line 137:
```suggestion
        cy.contains('Test Anonymous thread, should not show pronouns').find('.post_user_pronouns').should('not.exist');
```

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -72,4 +67,79 @@ describe('Tests cases abut changing user pronouns', () => {
...
+
+        //verify pronouns is shown in overall fourm page
+        cy.get('#nav-sidebar-forum').click();
+        cy.get('.thread-list-item').should('contain', 'Test pronouns display');
+        cy.contains('Test pronouns display').find('.post_user_pronouns').should('contain','They/Them');
+        cy.contains('Test pronouns display').find('.post_user_pronouns').click();
+
+        //comment on the thread, verify pronouns is shown
+        cy.get('.create-post-head').should('contain', 'Test pronouns display');
+        cy.get('#reply_box_2').type('my pronouns is They/Them{ctrl}{enter}');
```
IDzyre, line 91:
```suggestion
        cy.get('#reply_box_2').type('my pronouns are They/Them{ctrl}{enter}');
```

IDzyre (changes requested):
I like how everything is being tested in one 'it' statement. This may be a 'little harder to read' for newer developers than splitting the test up, however its best practice to have all tests be self-sufficient, and not require others tests before to setup.

NiaHeermance (approved):
Test now checks if pronouns are these as another user. Looks good!

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -12,18 +9,15 @@ describe('Tests cases abut changing user pronouns', () => {
...
-        //type in new pronouns
+        //type in new pronouns and check display in forum option
         cy.get('button[aria-label="Clear pronoun input"]').click(); //clear input using trash can
         cy.get('@e').type('They/Them');
+        cy.get('#pronouns-forum-display').check();
         cy.get('#edit-pronouns-submit').click();
 
-        //ensure pronouns changed on page
+        //ensure pronouns and display option changed on page
         cy.get('#pronouns_val').contains('They/Them');
```
IDzyre, line 19: Same comment about the should('contain', 'text') for this line and the one after.

IDzyre (changes requested):
One more comment about cypress assertions, then it looks good.

IDzyre (approved):

bmcutler:
@yanliw123 @IDzyre 
Can one of you please resolve the merge conflicts on this PR?
Thanks!

PR #9782, opened 2023-09-16
[Feature:System] Can Email Anyone

NiaHeermance:
### What is the current behavior?
Currently, Submitty's email API only allows emails to be sent to people with a Submitty user id.

### What is the new behavior?
Now, developers can send emails to anyone. This PR also adds emailing the sysadmin when a student readds themselves to a course, see PR #9768.

### How to Test?
As an instructor, move a student to the Null section and make sure you untick "Manual." Then, as that student, go to the course's page. On the No Access page you get, click the join button. Make sure that before you dropped the student, the student had done some sort of activity in the course within the past three days. Now, as an instructor, go to the email status page and see that one of the emails is to "Sysadmin."

### Images
image

This PR should only be merged after PR #9768 is merged.

PR #9768, opened 2023-09-09
[Feature:System] Student Override Back Into Course

NiaHeermance:
### What is the current behavior?
When student's are dropped from a course by automated scripts, they have no way of rejoining the course without contacting the instructor. This can cause them to get behind on work, and the instructor has to manage readding students.

### What is the new behavior?
The No Access page for a course now has an option for students to rejoin a course, provided they meet certain conditions:

- The student was not manually removed from the course by the instructor
- The student was in the course to begin with
- It is within two weeks of the start of the term OR the student has accessed a gradeable, poll, forum post, course material, etc within the last 72 hours.

### Relationship to Other PRs and Issues
This PR address issue #9738 but does not close it, as future PRs will address other parts of that issue. This PR is followed by PR #9799 and PR #9783, which are independent of each other and can be merged in any order.  This PR, PR #9768, must be merged first.

### How to test?
As an instructor, move a student to the Null section and make sure you untick "Manual." Then, as that student, go to the course's page. On the No Access page you get, click the join button. Make sure that before you dropped the student, the student had done some sort of activity in the course within the past three days.

### Images
image
and then,
image

PR #9799, opened 2023-09-22
[Feature:System] Can Email Anyone

NiaHeermance:
(This PR is a recreation of PR #9782 as I accidentally deleted its branch when testing something.)
This PR addresses more of issue #9738.

Currently, Submitty's email API only allows emails to be sent to people with a Submitty user id.

### What is the new behavior?
Now, developers can send emails to anyone. This PR also adds emailing the sysadmin when a student readds themselves to a course, see PR #9768.

### How to Test?
As an instructor, move a student to the Null section and make sure you untick "Manual." Then, as that student, go to the course's page. On the No Access page you get, click the join button. Make sure that before you dropped the student, the student had done some sort of activity in the course within the past three days. Now, as an instructor, go to the email status page and see that one of the emails is to "Sysadmin."

### Images
image


site/app/controllers/SelfRejoinController.php
```diff
@@ -129,6 +129,14 @@ private function sendRejoinedStudentEmail($joined_section): void {
             $email = new Email($this->core, $details);
             array_push($emails, $email);
         }
+        $sysadamin_email = $this->core->getConfig()->getSysAdminEmail();
+        if (!empty($sysadamin_email)) {
```
williamjallen, line 133: `empty()` performs a very loose comparison.  Given that `getSysAdminEmail()` returns a string, you should compare it to the empty string.

```suggestion
        if ($sysadamin_email !== '') {
```

site/app/controllers/SelfRejoinController.php
```diff
@@ -129,6 +129,14 @@ private function sendRejoinedStudentEmail($joined_section): void {
...
+            unset($details["to_user_id"]);
+            $details["email_address"] = $sysadamin_email;
+            $details["to_name"] = "Sysadmin";
+            $email = new Email($this->core, $details);
+            array_push($emails, $email);
```
williamjallen, line 138: The array push shorthand is generally cleaner:
```suggestion
            $emails[] = $email;
```

site/app/entities/email/EmailEntity.php
```diff
@@ -53,6 +56,13 @@ public function getUserId(): string {
         return $this->user_id;
     }
 
+    /**
+     * @return string
+     */
+    public function getToName(): string {
```
williamjallen, line 62: There's no need to put a redundant PHPDoc comment given that you're not providing any additional information

site/app/libraries/NotificationFactory.php
```diff
@@ -234,6 +234,16 @@ public function sendEmails(array $emails, bool $force_secondary = false, bool $s
         $current_user = $this->core->getUser();
         $flattened_emails = [];
         foreach ($emails as $email) {
+            if (!empty($email->getEmailAddress())) {
```
williamjallen, line 237: Again, `empty()` provides a very loose comparison, and `getEmailAddress()` returns a string, so you can use string comparison instead.

```suggestion
            if ($email->getEmailAddress() !== '') {
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5267,11 +5267,11 @@ public function insertNotifications(array $flattened_notifications, int $notific
      */
     public function insertEmails(array $flattened_params, int $email_count) {
         // PDO Placeholders
-        $row_string = "(?, ?, current_timestamp, ?, ?, ?, ?)";
+        $row_string = "(?, ?, current_timestamp, ?, ?, ?, ?, ?)";
```
williamjallen, line 5270: (leaving a comment here since I can't comment on lines you didn't change)

Can you add a PHPDoc comment specifying the [array shape](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes) for `$flattened_params`?  As it is now, it's difficult to understand what this array contains if you were to only see a small portion of the relevant code.

site/app/models/Email.php
```diff
@@ -14,34 +14,52 @@
...
+     */
     public function __construct(Core $core, array $details = []) {
         parent::__construct($core);
         if (count($details) == 0) {
             return;
         }
-        $this->setUserId($details["to_user_id"]);
-        $this->setSubject($this->formatSubject($details["subject"]));
-
+        if (array_key_exists("to_user_id", $details)) {
```
williamjallen, line 56: You could alternatively use `isset()` here.  This isn't a requested change, just a suggestion.

```suggestion
        if (isset($details["to_user_id"])) {
```

site/app/templates/EmailStatus.twig
```diff
@@ -36,14 +36,18 @@
                     {% endfor %}
                     {% for row in pending[subject] %}
                         <li class="status status-warning">
-                            <div class="status-message">Recipient: {{row.getUserId()}}</div>
+                            <div class="status-message">Recipient: {{
+                                (row.isToSubmittyUser()) ? row.getUserId() : row.getToName()
```
williamjallen, line 40: The extra parentheses aren't necessary here:
```suggestion
                                row.isToSubmittyUser() ? row.getUserId() : row.getToName()
```

site/app/templates/EmailStatus.twig
```diff
@@ -36,14 +36,18 @@
...
+                            }}</div>
                             <div class="status-message">Time Sent: Not Sent</div>
                             <div class="status-message">Email Address: {{row.getEmailAddress()}} </div>
                         </li>
                     {% endfor %}
                     {% for row in successes[subject] %}
                         <li class="status status-success"> 
-                            <div class="status-message">Recipient: {{row.getUserId()}}</div> 
+                            <div class="status-message">Recipient: {{
+                                (row.isToSubmittyUser()) ? row.getUserId() : row.getToName()
```
williamjallen, line 49: Same here:
```suggestion
                                row.isToSubmittyUser() ? row.getUserId() : row.getToName()
```

williamjallen (dismissed):

site/app/controllers/SelfRejoinController.php
NiaHeermance: I wanted to also check if it's nonnull (which is why I was using null), so I'll make the condition include that too.

site/app/controllers/SelfRejoinController.php
NiaHeermance: I prefer the other syntax, but I understand the latter is more PHP-esque, so I'll use what you recommend.

site/app/libraries/NotificationFactory.php
NiaHeermance: Would this be better than?
`$email->getEmailAddress() !== null && $email->getEmailAddress() !== ''`

site/app/models/Email.php
```diff
@@ -14,34 +14,52 @@
...
-        $this->setUserId($details["to_user_id"]);
-        $this->setSubject($this->formatSubject($details["subject"]));
-
+        if (isset($details["to_user_id"])) {
+            $this->setUserId($details["to_user_id"]);
+        }
+        else {
+            $this->setEmailAddress($details["email_address"]);
+            $this->setToName($details["to_name"]);
+        }        $this->setSubject($this->formatSubject($details["subject"]));
```
cjreed121, line 62: Split to a second line

site/app/libraries/NotificationFactory.php
```diff
@@ -234,6 +234,16 @@ public function sendEmails(array $emails, bool $force_secondary = false, bool $s
         $current_user = $this->core->getUser();
         $flattened_emails = [];
         foreach ($emails as $email) {
+            if ($email->getEmailAddress() !== null && $email->getEmailAddress() !== '') {
```
cjreed121, line 237: Could we just use !empty() here?

cjreed121 (changes requested):
Looks good. Just some minor changes.

site/app/libraries/NotificationFactory.php
NiaHeermance: I had that, but Will wanted me to change it to this to be more explicit. I think I will change it back though, as I don't think Will understood that I was also considering the null case in my logic (he recommended `$email->getEmailAddress() !== ''` originally).

NiaHeermance: @williamjallen

site/app/libraries/NotificationFactory.php
williamjallen: `getEmailAddress()` returns a string (as per its PHP type annotation), so the null check should not be necessary.

site/app/libraries/NotificationFactory.php
williamjallen: `getEmailAddress()` returns a string, so we can be more explicit by checking against an empty string.  `empty()` provides a pretty loose comparison, so it should be avoided if possible.

site/app/controllers/SelfRejoinController.php
williamjallen: I see that it's theoretically possible for it to be null, so it can either be left with `empty()`, or a more explicit check for null or the empty string.

site/app/libraries/database/DatabaseQueries.php
NiaHeermance: Done!

migration/migrator/migrations/master/20230914092120_add_to_name_to_email_database.py
```diff
@@ -0,0 +1,34 @@
...
+        ADD COLUMN to_name VARCHAR;
+        --- Now user_id null if to_name used instead.
+        ALTER TABLE emails
+        ALTER COLUMN user_id DROP NOT NULL;
+        ALTER TABLE emails
+        ADD CONSTRAINT name_or_email CHECK (
+            (user_id is NOT NULL)
+            <>
+            (to_name is NOT NULL)
+        );
```
cjreed121, line 23: This migration is not runnable twice. You will need to add some things like `IF NOT EXISTS` or add some stuff to your down migration.

cjreed121 (changes requested):

pbailie (commented):
So far — my testing is finding no conflicts with the autofeed script.

migration/migrator/migrations/master/20230914092120_add_to_name_to_email_database.py
NiaHeermance: Fixed!

cjreed121 (approved):

PR #9804, opened 2023-09-24
[Bugfix:TAGrading] Checkpoint Grading Empty Section

cjreed121:

If you try to load the checkpoint grading page with an empty section, a frog robot error will occur. This is due to the null coalesce being null rather than an empty array.

### What is the new behavior?
The page loads normally as expected. The null coalesce is now just an empty array as it should be.

### Other information?
Tested locally and it does fix the issue.


ziesski (approved):
Appropriate null handling.

PR #9809, opened 2023-09-26
[Feature:System] Display Recently Dropped Courses

NiaHeermance:
This PR should only be merged after PR #9783 is merged. PR #9818 follows this one. This PR continues to address issue #9738.

Students cannot see their recently dropped courses that they can rejoin on the course list.

### What is the new behavior?
These dropped courses now appear.

### Images
image

image


site/phpstan-baseline.neon
```diff
@@ -10532,12 +10537,12 @@ parameters:
 			path: app/views/AutoGradingView.php
 
 		-
-			message: "#^Method app\\\\views\\\\AutoGradingView\\:\\:convertToAnonPath\\(\\) has parameter \\$gradeable with no type specified\\.$#"
```
04vmatsibekker, line 10535: You might want to check with William, but I think he has commented for one of my PRs, and one of Anson's PRs that you might be able to fix some of these errors rather than ignoring them, but you should ask him for more info to double check.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -4919,24 +4920,29 @@ public function getPeerGradingAssignmentsForGrader($grader_id) {
...
+            $force_nonnull = "NOT";
         }
 
         $query = <<<SQL
 SELECT t.name AS term_name, u.term, u.course, u.user_group, u.registration_section
 FROM courses_users u
 INNER JOIN courses c ON u.course=c.course AND u.term=c.term
 INNER JOIN terms t ON u.term=t.term_id
-WHERE u.user_id=? ${extra} AND (u.registration_section IS NOT NULL OR u.user_group<>4)
+WHERE u.user_id=? ${include_archived} AND ${force_nonnull} (u.registration_section IS NOT NULL OR u.user_group<>4)
```
micpap25, line 4945: There are two places where registration section is changed; in the "submitty" database and the specific "submitty_semester_course" database. This PR (as well as all previous PRs) only cause users to see their dropped course as being "dropped" (eg. not in "My Courses") if the former is changed; if the latter is changed, the user is still denied access but does not see it until they attempt to access the course.

micpap25 (commented):
PR seems good to me and works as intended for the most part. I bugtested a bit in the database and found an issue with this chain of PRs if the submitty database cluster goes out-of-sync with itself in terms of registration section; not sure how major this is. Would appreciate some discussion over if this is necessary to fix. Otherwise LGTM.

site/app/libraries/database/DatabaseQueries.php
micpap25: Highlighting this line because this is where the dependence on the overall submitty database comes in (the registration status on the course database is not considered here).

micpap25 (approved):
Upon further thinking I may consider the issue I had to be out-of-scope for this PR since it affects all courses, even in Submitty right now if there was a database issue. So I'll approve this to get it moved along because it accomplishes what it wants to do.

PR #9813, opened 2023-09-26
[Bugfix:Developer] Add cypress videos back

IDzyre:
With Cypress 13.0, the video recording option defaults to _false_ instead of _true_. 
see [Cypress Changelog](https://docs.cypress.io/guides/references/changelog#13-0-0)
### What is the new behavior?
To fix this, another config option was added, video: true,

This sets cypress to record the videos to cypress/videos

These should now be visible in the artifacts after a failing test.


NiaHeermance:
Where do I access the config for cypress when running it locally?

IDzyre:
I believe that it is the same file, is it not ?

NiaHeermance:
Okay! I don't know much about cypress videos so I'll have to experiment when reviewing this.

NiaHeermance (commented):
I got this error, but a video was recorded. Is this fine?
image

NiaHeermance (approved):
Videos appear when a test fails. Looks good!

PR #9818, opened 2023-09-27
[Feature:System] Recently Dropped Icon

NiaHeermance:
This PR should only be merged after PR #9809 is merged.

There is no way for instructors to see the recently dropped students. Recently dropped students can rejoin a course, so it can be useful for an instructor to know who can readd themselves and who needs a manual readd.

### What is the new behavior?
There is now an icon for recently dropped students.

### Images
image


site/ts/activity-dashboard.js
```diff
@@ -134,7 +134,6 @@ export function applySettings() {
         let s_num_polls = data[i].num_poll_responses;
         const s_off_hours = data[i].office_hours_queue;
         const s_course_mat = data[i].course_materials_access;
-
```
micpap25, line 137: Can this be undone? It clutters up the PR.

site/app/controllers/admin/UsersController.php
```diff
@@ -83,6 +84,24 @@ public function getStudents() {
             }
         }
 
+        $can_rejoin_in_null = [];
```
micpap25, line 87: Is "can rejoin in null" a good name? Would "can rejoin" suffice?

site/app/controllers/SelfRejoinController.php
```diff
@@ -76,7 +127,7 @@ private function canRejoinCourse() {
             }
         }
 
-        $term_start_date = $this->core->getQueries()->getCurrentTermStartDate();
+        $term_start_date = $this->core->getQueries()->getTermStartDate($term, $user);
         // If never accessed course but today is within first two weeks of term, can re-add self.
```
micpap25, line 131: How is "never accessed course" enforced here? If it isn't, this line should be changed to avoid confusion

micpap25 (changes requested):
Works as intended on the user end, just a few code nitpicks

williamjallen:
Is this ready for review?  It is currently a draft and there are outstanding review comments to resolve, so I have moved this back to "Work in Process".

micpap25:
PR should be merged per @bmcutler, the CI issue is a performance issue from excessive requests since the `canRejoinCourse` function wasn't designed with calling on lots of users at once. Opening a new issue for that.

micpap25 (approved):

ziesski (approved):
Didn't test the functionality, but the code looks good. As described in new issue, it does have excessive query, which should be optimized, we could add new column or re-write the query into one giant query.

PR #9810, opened 2023-09-26
[Refactor:InstructorUI] Consolidate color definitions

DarthNyan:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant
* [ ] Screenshots are attached to Github PR if visual/UI changes were made

### What is the current behavior?
Addresses issue #9679. CSS color definitions are in various files and often hardcoded.

### What is the new behavior?
Moved all color definitions in various css files to colors.css. Started with server.css, then worked alphabetically. Some colors altered slightly due to being similar to existing colors in the color palette, but changes should barely be noticeable if at all.

### Other information?
WIP. Will test with WAVE before final pull request.

PR #9837, opened 2023-10-01
[Feature:Calendar] Clickable buttons darken when hovering.

### LINKED ISSUE #9705: Idea: Show Clickable Calendar Entries with two tones of color
There is no way to show entries are clickable.

Make a calendar entry have a lighter/darker shade to show it is clickable or not clickable

**Additional context**

DarthNyan:

Closes #9705. Calendar entries with links embedded appear no different from those without. It is difficult to tell whether a button has a link embedded without clicking it.

### What is the new behavior?
When hovering the mouse over a button with a link, the button becomes slightly darker. Buttons without links do not change color. The screenshot doesn't show the cursor, but it's hovering over the "Docker Network Python" button in the top screenshot, but not in the bottom.
clickcaldiff
clickcallbef

### Other information?
I built this on top of my previous branch, PR https://github.com/Submitty/Submitty/pull/9810 . All the changes to CSS files are related to that. This project only required editing calendar.js

Tested by accessing calendar page and ensuring proper behavior of entries with and without links.


williamjallen:
Given that this PR depends upon https://github.com/Submitty/Submitty/pull/9810, it should be marked as a draft until https://github.com/Submitty/Submitty/pull/9810 is merged.

site/public/js/calendar.js
```diff
@@ -97,6 +97,31 @@ function dateToStr(year, month, day) {
     return `${year.toString()}-${month.toString().padStart(2, '0')}-${day.toString().padStart(2, '0')}`;
 }
 
+/**
+ * This function returns a slightly darker color than the color variable name passed.
+ *
+ * @param colorstr : string the color to darken in the form "var(--color-name)"
+ * @returns {string} a hex code for a slightly darker shade
+ */
+function darken(colorstr) {
```
04vmatsibekker, line 106: What happens if the color passed into this function is like the darkest black. What will be returned by this function? Will it still work?

And if not how can we try to fix this? Also, when I switch to dark mode, I feel like the change in color is just not noticeable enough, but that might just be me.

04vmatsibekker (commented):
Wh

site/public/js/calendar.js
DarthNyan: I see what you're saying about the color change being very subtle on dark mode. I just doubled the amount the function darkens colors by. Let me know if that's better.

The function should still work for black and very dark grays. It doesn't alter values that are '0' or '1', so it would just leave black as-is. I can make it return a slightly lighter gray, but there aren't any calendar buttons that use black or very dark gray backgrounds so I don't know if it's necessary.

benjaminlee755 (approved):
Everything looks good to me. The buttons with links darken when hovered.

site/public/js/calendar.js
04vmatsibekker: It could be good to have the function account for a very dark color even if we don't have to deal with those scenarios for now, just in case we run into that situation on future PRs.

taorif25 (approved):
Code changes work as expected on the site. Darkening the icons make the UI more intuitive and it looks good. Changes are isolated to the issue and don't seem like they could cause any problems with the site. Good job!

PR #9838, opened 2023-10-02
[Bugfix:Developer] Fix vagrant up slack message

IDzyre:
After I was testing the slack message on failure, I accidentally deleted the line to get the JOB ID on failure. This causes the vagrant up notification to not have the JOB ID, so the link is bad.

### What is the new behavior?
The line to run this on failure has been re-added, this will make the link accurate.


williamjallen (approved):
Changes seem reasonable to me.

PR #9844, opened 2023-10-03
[Testing:Developer] Refactor Cypress Tests

### LINKED ISSUE #9482: Refactor Cypress Tests to Use Cypress Variables

liny18:

Partially Fixes #9482; changes only implemented for test files under site/cypress/e2e/Cypress-UI.
Some places are not using Cypress variables when possible.
Some places are not using aliases when it can be used for better efficiency/cleaner code.
Some strings are hardcoded multiple times when a variable can be used.

### What is the new behavior?
Cypress variables are used whenever possible.
Aliases are used whenever they result in benefit.
Slightly reduced the use of repetitive code.

### Other information?

docker.spec.js is failing, but I made no changes to the file.

### DIFF
--- a/site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
+++ b/site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
@@ -31,7 +31,8 @@ describe('Tests cases abut changing user pronouns', () => {
         cy.login('student');
 
         //change back to old pronouns
-        cy.get('@pronounsVal').click();
+        cy.get('#pronouns_val').as('pronounsVal').click();
+        cy.get('#user-pronouns-change').as('pronounsInput');
         cy.get('button[aria-label="Clear pronoun input"]').click(); //clear input using trash can
         if (oldPronouns !== '') {
             cy.get('@pronounsInput').type(oldPronouns);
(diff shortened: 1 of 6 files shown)


site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -58,7 +55,7 @@ describe('Tests cases abut changing user pronouns', () => {
         cy.get('#toggle-student-col-submit').first().click();
 
         //Ensure correctness in table
-        cy.get('.td-pronouns:eq( 12 )').should('have.text', 'They/Them');
+        cy.get('.td-pronouns:eq( 12 )').should('have.text', newPronouns);
```
IDzyre, line 58: Ideally this would use a different selector, one that wouldn't break if the order of the students changes.

liny18: Yea, I definitely agree. But I would argue that isn't the concern for the issue that this PR is trying to address.

IDzyre: I agree, hence not a requested change, just a comment.

IDzyre (approved):

PR #9850, opened 2023-10-05
[Bugfix:System] Stricter Course Rejoin Requirements

NiaHeermance:
Students can "rejoin" courses they are already in if they specifically go to the null access page. They can also rejoin archived courses.

### What is the new behavior?
Students can only rejoin in the null section, and the course must be not archived.

### DIFF
--- a/site/app/controllers/SelfRejoinController.php
+++ b/site/app/controllers/SelfRejoinController.php
@@ -36,14 +36,35 @@ class SelfRejoinController extends AbstractController {
         $course = $this->core->getConfig()->getCourse();
         $term = $this->core->getConfig()->getTerm();
 
-        // If manually removed from course, this was probably intentional removal.
-        if (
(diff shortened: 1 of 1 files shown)


taorif25 (approved):
Minimal changes, everything seems to be working as intended. Good comments!

bmcutler:
@NiaHeermance 
Please check back on this PR, it failed a Cypress test after I merged in @IDzyre s fix for Cypress testing.  
So it might be a real failure?
Trying another merge master & re-run.

I approve the new logic you've added, but I wonder if its possible that the access information might be flawed (and never not be true).  If a student attempts to access a page (and fails because they've been dropped) do we update the database with the attempt and use that new access data to check for 'recent activity'?  Basically I'm asking if we've tested if its possible to actually fail the access time check?  It can wait for another PR if there is a problem.  I will merge this when it passes the tests.

PR #9852, opened 2023-10-06
[Bugfix:Autograding] Fix Autograding to use term or semester

### LINKED ISSUE #9839: Autograding in sample courses never finishes
**Describe the bug**
Autograding in sample courses never finishes

**Expected behavior**
Autograding in sample courses is supposed to finish

**To Reproduce**
1. Have a submitty vm of the lastest build and do `vagrant ssh`
2. Run `sudo systemctl stop submitty_autograding_shipper`
3. Run `sudo systemctl stop submitty_autograding_worker`
4. Then run `sudo su -c '/usr/local/submitty/autograder/submitty_autograding_shipper.py' submitty_daemon`
5. See error

**Configuration**

**Screenshots**
This is what I see as my error output
Test1_latest_ver

**Additional context**
It seems like the error is being thrown in submitty_auto_grading.py line 408

jason490:

Closes #9839
Autograding scripts would run into an error because it was reading 'semester' when it was changed to 'term' in some areas of autograding config files in sample courses

### What is the new behavior?
Fixes autograding scripts. Added a backwards compatibility with "semester" and autograding should function


autograder/submitty_autograding_shipper.py
jason490, line 1: For some reason on line 482, doesn't need to be changed to semester for the autograding to not run into errors

jason490:
Currently the autograding works, but it cannot reach some files as some folders are not created/exist

jason490:
Looking at this again, it seems like not all of the autograding config files have been changed from 'semester' to 'term'. I will write something that will fallback to 'semester' when accessing 'term' fails

jason490:
When looking at the autograding status it seems like there is a stack trace of not being able to read a file because it seems like the file doesn't exist
But, otherwise it doesn't seem like there is any major errors

NiaHeermance (approved):
This is a good fix. You could refactor the .setup folder to use "term" instead of "sample" when creating autograding jobs, but if you don't feel comfortable doing that, then this works for the time being until I get around to doing that refactor myself.

jason490:
When fixing the autograder shipper to view the 'term' or 'semester', how do you update the shipper or worker after doing `submitty_install`?
I tried restarting both the shipper and the worker and that didn't work, but it started throwing a different error?
I need someone who has time to test this PR, Fix autograding to use term or semester out, with using submitty install with a currently running version of main

I had to do a vagrant destroy and vagrant up, and on PR#9852 (Fix autograding to use term or semester). Which fixed the autograding issue?

Also, it seems like you can't re-create a grading file (clicking regrade) while its in the process of regrading (probably the file exists already)? 
The issue with that is what happens when the autograding is broken and there is a fix for it, but the clicking regrade all doesn't replace the autograding files because it already exists

Otherwise,  regrade all versions work only when all the autograding finishes / most
batch regrade works

bmcutler:
I've pushed a few changes to update the queue files for regrading

Scanned PDF Exam / Bulk upload is buggy

I don't know if it's related to this PR.  When I go to view the TA grading page (listing all students) for a bulk upload gradeable,
I get a pile of error messages:

Screenshot 2023-10-22 at 2 24 54 PM

jason490:
> I've pushed a few changes to update the queue files for regrading
> 
> Scanned PDF Exam / Bulk upload is buggy
> 
> I don't know if it's related to this PR. When I go to view the TA grading page (listing all students) for a bulk upload gradeable, I get a pile of error messages:
> 
> Screenshot 2023-10-22 at 2 24 54 PM

I made an PR for that bug. PR #9923

bmcutler (approved):
Ok, I've tested this, I think we are ok to install this.
Still lots of "semester"s to replace, but I think regrading & submission works ok.
[ I did not test sample course - vagrant destroy/up ]

shailpatels (commented):
Sorry for the late review but did a quick checkout + test of bulk upload and looks good!
The PHP error for the invalid id was unrelated to bulk upload and fixed with #9923 as well.

And yeah updating all of bulk upload to use 'term' instead will touch a couple of other things, mostly the `jobs.py` under `sbin/submitty_daemon_jobs/submitty_jobs` is expected "semester" to be a present key in the job file dropped and many of the other jobs in that file depend on that behavior as well

PR #9853, opened 2023-10-06
[Refactor:TAGrading] Late day cache improvement

cjreed121:

When the late day cache is fetched, it fetches the cache for each user on its own query. This becomes quite expensive as the class size grows.

### What is the new behavior?
Up front, a cache gets created with all user's late day info. Then per call it will fetch from that cache instead.

### Other information?
Went from 154 queries down to 16 queries on one of the TA grading pages.


ziesski (approved):
Looks good

Youssef-Chip (approved):

PR #9857, opened 2023-10-06
[Feature:System] Add Reason For an Excused Extension

### LINKED ISSUE #9819: Add new 'notes' column to excused absense
Instructors have requested the ability to note and track the reason for an excused absence.
(on the "Excused Absense Extension" form).
Standard reasons/explanations/notes might include "illness", "interview", "travel-conference", "travel-personal", "schedule conflict", ... anything else?

* add a 'notes' column (type = string) to the late_day_exceptions database table to be used for the reason/explanation for grading the excused absense.

* default this new column to empty string.  

* Write a migration to insert the column and empty string to all existing courses.

* on the UI, add a dropdown box for input of this data.  The dropdown should default to "unspecified" which is empty string.  The dropdown should include the standard reasons above but also have an "other" option that allows the instructor to type an arbitrary string.

* make sure the CVS upload option works with this new option.  ideally it is backwards compatible (accepts the old format of csv with 3 columns) or the new format with 4 columns)

* make sure to add the reason column to the table with all current data that is displayed at the bottom when a specific gradeable is selected.

taorif25:
Closes #9819 

Currently, instructors don't have the ability to give a reason for an excused late day extension.
image

### What is the new behavior?
The option to add a reason has been added to the Excused Extension page. The database has been updated with a new column in the late_day_exceptions section. CSV file upload has been updated to include this new parameter, but is still backwards compatible with the old 3 parameter format. Also fixes issue where CSV upload results in a JSON parsing error. Users can input a custom reason or select an option from a drop down menu.
image

### Other information?

Note: CSV upload will still result in an error parsing data. See PR #9931 for fix on this.


site/app/templates/admin/Extensions.twig
```diff
@@ -43,12 +43,24 @@
...
+                    </select>
+                </div>
+            </label>
             <a class="btn btn-primary" href="javascript:updateHomeworkExtension()">Submit</a>
         </fieldset>
         <fieldset>
             <legend class="option-title" id="csv-upload-title">Multiple Student Entry Via CSV Upload</legend>
             <label for="csv-upload" id="csv-upload-format">
-                Do not use column headers. CSV must be of the following form: student_id, gradeable_id, days_of_extension
+                Do not use column headers. CSV must be of the following form: student_id, gradeable_id, days_of_extension, reason_for_extension<br>
```
NiaHeermance, line 62: Add a period:
`Do not use column headers. CSV must be of the following form: student_id, gradeable_id, days_of_extension, reason_for_extension.<br>`

NiaHeermance (changes requested):
Looks good! I tested adding a reason both through the page and through a CSV file. The backwards compatibility works as intended.

Maybe you could updated the gradeable submit page to tell a student the reason. Example:

`You have a 2 day deadline extension for this assignment for this reason: illness.`

I also wonder if the selector for the reason could change for dark mode option. It's totally okay if it can't, as HTML elements like that can be unflexible. Still, look into it.

A couple of bugs:
- I can set a custom reason when using a csv file, but I can't via the webpage.
- You get an error message when submitting, even when it works.
- The delete reason button (the trash can) does not work. 

Fix these bugs and address my other comments and then I can approve. :D

site/app/controllers/admin/LateController.php
```diff
@@ -226,6 +226,13 @@ public function updateExtension() {
                     JsonResponse::getFailResponse($error)
                 );
             }
+            $reason_for_extension = null;
+            if (isset($_POST['reason_for_exception'])) {
+                $reason_for_exception = $_POST['reason_for_exception'];
+            }
+            else {
+                $reason_for_exception = 'unspecified';
+            }
```
williamjallen, line 235: This is a great chance to use PHP's [null coalescing](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce) operator!

```suggestion
            $reason_for_exception = $_POST['reason_for_exception'] ?? 'unspecified';
```

I assume the "exception" vs "extension" naming thing here is a typo?

site/app/controllers/admin/LateController.php
```diff
@@ -359,8 +366,8 @@ private function parseAndValidateCsv(string $csv_file, array &$data, string $typ
                 return trim($k);
             }, $fields);
 
-            //Each row has three fields
-            if (count($fields) !== 3) {
+            //All types have 3 fields except for exceptions, which can have 3 or 4 rows.
+            if (count($fields) !== 3 && !($type == 'extension' && count($fields) == 4)) {
```
williamjallen, line 370: It's almost always a good idea to use PHP's strict comparison operators whenever possible.

```suggestion
            if (count($fields) !== 3 && !($type === 'extension' && count($fields) === 4)) {
```

site/app/controllers/admin/LateController.php
```diff
@@ -400,6 +407,10 @@ private function parseAndValidateCsv(string $csv_file, array &$data, string $typ
                     "error" => "Third column must be an integer greater or equal to zero, got '{$fields[2]}' on row {$row_number}",
                 ];
             }
+            //$fields[3] added if not present to extension type. Allows for backwards compatibility.
+            if ($type == "extension" && count($fields) == 3) {
```
williamjallen, line 411: Same comment here:
```suggestion
            if ($type === "extension" && count($fields) === 3) {
```

site/app/views/admin/ExtensionsView.php
```diff
@@ -31,17 +34,20 @@ public function displayExtensions($gradeables) {
...
                                           'user_givenname' => $user->getDisplayedGivenName(),
                                           'user_familyname' => $user->getDisplayedFamilyName(),
-                                          'late_day_exceptions' => $user->getLateDayExceptions()];
+                                          'late_day_exceptions' => $user->getLateDayExceptions(),
+                                          'reason_for_exception' => $user->getReasonForException()];
         }
         if (empty($current_exceptions)) {
             $current_exceptions = null;
         }
+        $reasons = ["illness", "interview", "travel", "personal issue"];
```
williamjallen, line 43: Why can't you just inline this variable?

williamjallen (changes requested):

site/app/views/admin/ExtensionsView.php
taorif25: Could you explain what that means please?

williamjallen: `$reasons` is only used in one location after being defined here.  You could eliminate the temporary variable by removing this line and changing the block of code below this to something like the following:

```
return $this->core->getOutput()->renderTwigTemplate("admin/Extensions.twig", [
             "gradeables" => $gradeables,
             "student_full" => $student_full,
             "current_gradeable" => $current_gradeable,
             "current_exceptions" => $current_exceptions,
             "reasons" => [
                 "illness",
                 "interview",
                 "travel",
                 "personal issue",
             ],
         ]);
```

This is just a suggestion though, so feel free to disregard if you disagree though.

site/app/views/admin/ExtensionsView.php
```diff
@@ -8,13 +8,20 @@
...
         $this->core->getOutput()->addInternalCss('exceptionforms.css');
         $this->core->getOutput()->addInternalCss('table.css');
         $this->core->getOutput()->addInternalJs('extensions.js');
         $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('flatpickr', 'flatpickr.min.js'));
         $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('flatpickr', 'flatpickr.min.css'));
         $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('flatpickr', 'plugins', 'shortcutButtons', 'shortcut-buttons-flatpickr.min.js'));
         $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('flatpickr', 'plugins', 'shortcutButtons', 'themes', 'light.min.css'));
+        $this->core->getOutput()->addVendorJs(FileUtils::joinPaths('select2', 'js', 'select2.min.js'));
+        $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('select2', 'css', 'select2.min.css'));
+        $this->core->getOutput()->addVendorCss(FileUtils::joinPaths('select2', 'bootstrap5-theme', 'select2-bootstrap-5-theme.min.css'));
```
cjreed121, line 24: These are duplicated

cjreed121 (changes requested):

site/app/views/admin/ExtensionsView.php
taorif25: Just fixed this, thanks for letting me know!

cjreed121 (approved):
Looks good. Tested and works as expected.

NiaHeermance (approved):
All my comments have been addressed. Works great!

NiaHeermance:
One thing: In the future, don't put apostrophes in your branch name as it makes it harder to switch to your branch on the command line. (I had to put quotation marks on your branch name, so not a big deal, but something to keep in mind for the future.)

PR #9870, opened 2023-10-10
[UI/UX:TAGrading] grade inquiry only warning

yanliw123:
### Previous behavior:
image
the grading page looks like this before, when viewing grade inquiry only
### What is the new behavior?
click on the grade inquiry only button will show a warning that reminds the grader is viewing grade inquiry only 
image

### How to test?
go into a gradeable, click on the grade inquiry only button, see warning


williamjallen:
What was the previous behavior?

yanliw123:
> What was the previous behavior?

image
It looks like this previously, if the grader is viewing grade inquiry, and if there isn't any active grade inquiry. It might be confusing for the graders sometimes.

williamjallen:
Thanks!  Can you edit that into your PR description so it's easier for reviewers to tell what the before/after comparison should look like?

bmcutler:
@williamjallen @NiaHeermance Can one/both of you please review? thanks

site/app/templates/grading/electronic/Details.twig
```diff
@@ -64,14 +64,23 @@
...
-            } else {
-                reverse_inquiry_only();
-                button.innerHTML = "Grade Inquiry Only: Off";
+            } 
+            else {
+                $('[data-testid=grade-button]').each(function() {
+                        $(this).closest('[data-testid="grade-table"]').show();  // show all gradeable items
+                    });
+                    button.innerHTML = "Grade Inquiry Only: Off";
             }
```
williamjallen, line 80: The indentation is weird here.

Also: we try to avoid using `.innerHTML()` because it can inadvertently lead to XSS vulnerabilities down the line if future developers are not cautious.  Use `.text()` or similar instead.

site/app/templates/grading/electronic/Details.twig
```diff
@@ -64,14 +64,23 @@
             var button = document.getElementById('inquiryButton');
             status = Cookies.get('inquiry_status')
             if (status === 'on') {
-                grade_inquiry_only();
+                $('[data-testid=grade-button]').each(function() {
+                    const hasGradeInquiry = typeof $(this).attr('data-grade-inquiry') !== 'undefined';
+                    if (!hasGradeInquiry) {
```
williamjallen, line 69: It might be simpler to inline `hasGradeInquiry` here and get rid of the temporary variable.

site/app/views/grading/ElectronicGraderView.php
```diff
@@ -832,6 +832,10 @@ public function detailsPage(Gradeable $gradeable, $graded_gradeables, $teamless_
         elseif (count($grader_registration_sections) === 0) {
             $message = 'Notice: You are not assigned to grade any students for this gradeable.';
         }
+        if ($inquiry_status) {
+            $message .= ($message === "") ? "Notice: You are viewing students with active grade inquiry." : "\nNotice: You are viewing students with active grade inquiry.";
```
williamjallen, line 836: I'm confused about what's going on here.  If you are trying to conditionally add a newline, can't you just do that rather than duplicating the same string?

Also, this is not grammatically correct.

williamjallen (changes requested):

site/app/templates/grading/electronic/Details.twig
```diff
@@ -64,14 +64,22 @@
...
+                        $(this).closest('[data-testid="grade-table"]').hide();  // hide gradeable items without active inquiries
+                    }
+                });
+                button.textContent = "Grade Inquiry Only: On";
+            } 
+            else {
+                $('[data-testid=grade-button]').each(function() {
+                        $(this).closest('[data-testid="grade-table"]').show();  // show all gradeable items
+                    });
+                    button.textContent = "Grade Inquiry Only: Off";
```
williamjallen, line 78: The indentation is still weird here

site/app/templates/grading/electronic/Details.twig
```diff
@@ -64,14 +64,22 @@
...
-                grade_inquiry_only();
-                button.innerHTML = "Grade Inquiry Only: On";
-            } else {
-                reverse_inquiry_only();
-                button.innerHTML = "Grade Inquiry Only: Off";
+                $('[data-testid=grade-button]').each(function() {
+                    if (typeof $(this).attr('data-grade-inquiry') === 'undefined') {
+                        $(this).closest('[data-testid="grade-table"]').hide();  // hide gradeable items without active inquiries
+                    }
+                });
```
williamjallen, line 71: Is there a better way to query these elements?

williamjallen (changes requested):

site/app/templates/grading/electronic/Details.twig
yanliw123: I don't think there is a better way to query those. 
image
since I'm trying to show/hide the entire row. I think I have to query '[data-testid="grade-table"]'

site/app/templates/grading/electronic/Details.twig
```diff
@@ -62,16 +62,25 @@
     <script>
         window.onload = function() {
```
williamjallen, line 63: Can you move this code to a `.js` file so it can be linted by our JS linters?

site/app/templates/grading/electronic/Details.twig
```diff
@@ -62,16 +62,25 @@
...
             var button = document.getElementById('inquiryButton');
-            status = Cookies.get('inquiry_status')
+            var status = Cookies.get('inquiry_status');
```
04vmatsibekker, line 65: maybe use const or let instead of var

site/app/templates/grading/electronic/Details.twig
```diff
@@ -94,7 +103,7 @@
     <button type="button" class="btn btn-primary" onclick="expandAllSections()">Expand All Sections</button>
     {% if message is not empty %}
     <div class="details-warning-box{% if message_warning %} details-warning-highlight{% else %} details-warning-neutral{% endif %}">
-        {{ message }}
+        {{ message|nl2br }}
```
04vmatsibekker, line 106: maybe add the br in an {%if %} {%else%} condition, because I don't see this syntax a lot in the code, but I don't know it's probably ok

williamjallen (approved):
Code seems reasonable.  Needs an additional feature review.

VVVITH (approved):
Tested the features, works as expected.

PR #9872, opened 2023-10-10
[Bugfix:Developer] Remove unused cached files

IDzyre:
The CI tests that required the VM to make a mock instance of Submitty has been flaky/failing due to disk space issues.

### What is the new behavior?
Lines have been added to submitty_ci.yml to remove unnecessary files that are from the image used by Github Actions runners. 

### Other information?
[Github Issue comment](https://github.com/actions/runner-images/issues/2840#issuecomment-790492173)


.github/workflows/submitty_ci.yml
```diff
@@ -305,6 +305,15 @@ jobs:
...
     steps:
+      - name: Delete huge unnecessary folders, and clean apt cache
+        run: |
+          rm -rf /opt/hostedtoolcache
+          rm -rf /usr/share/dotnet
+          rm -rf /opt/ghc
+          rm -rf "/usr/local/share/boost"
+          rm -rf "$AGENT_TOOLSDIRECTORY"
+          sudo apt-get clean
+
```
williamjallen, line 316: Instead of duplicating this code, is there any reason why you can't move this to the setup composite action?

.github/workflows/submitty_ci.yml
```diff
@@ -305,6 +305,15 @@ jobs:
           - 5432:5432
 
     steps:
+      - name: Delete huge unnecessary folders, and clean apt cache
+        run: |
+          rm -rf /opt/hostedtoolcache
+          rm -rf /usr/share/dotnet
+          rm -rf /opt/ghc
+          rm -rf "/usr/local/share/boost"
```
williamjallen, line 313: Does this interfere with our usage of Boost?

.github/workflows/submitty_ci.yml
IDzyre: There is not, it has been moved to composite.

.github/workflows/submitty_ci.yml
IDzyre: im not sure. It works without this line, so I have removed it.

williamjallen (approved):

PR #9880, opened 2023-10-12
[Bugfix:Calendar] fixes edit/delete item in calendar

yanliw123:
1. When the instructor makes a note/announcement in the calendar, the notes/ announcement is not editable or deletable
2. Hover the mouse over the note/announcement does not change the mouse to a pointer (indicating the item is clickable)

### What is the new behavior?
1. Instructor clicks on a note/announcement item to edit/delete an item
2. Hover over the item will change the mouse to a pointer


RitaLei123 (approved):
The three step review process:
1. Is the proposed change a good idea?
Yes, editing comments is a great idea because typos are common.

2. Did the author take the right approach?
It appears that the author fixed the java script so that it would check for length instead of just checking if it exists.

3. Are all the implementation details correct?
Yes, the implementation details are correct. Additionally, maybe create a new issue to add the feature to allow professors to edit the course as well because sometimes, they could post it without checking the course.

Manual testing:
1. Reproduce on main: Yes, those features are not available and a fix was needed.
2. Check topic branch
3. Reproduce on main again: Those features are added.

PR #9899, opened 2023-10-18
[Documentation:Developer] Wrap placeholder text in HTML comments

williamjallen:
Our issue templates currently include placeholder text.  If the placeholder text is not removed manually prior to opening an issue, the placeholder text will remain visible once the issue is opened.

### What is the new behavior?
To reduce the amount of spam in our new issues, I have wrapped the placeholder text in HTML comments.  The placeholder text will now be hidden if it is not removed prior to opening an issue.


ziesski (approved):
Looks good to me

MasterOdin (approved):

PR #9900, opened 2023-10-18
[Testing:Developer] Add files to PHPStan config

williamjallen:

We currently require developers to list the files to check when running PHPStan.  PHPStan supports the ability to list these files in the configuration file.

### What is the new behavior?
It is no longer necessary to specify the files to check when running PHPStan.


ziesski (approved):
This is awesome, if I had known this sooner, I would have been much less motivated to do the script. (I still see some value in script, as long as we add most of our testing)

PR #9901, opened 2023-10-18
[Feature:Developer] Enable PHPStan strict rules plugin

williamjallen:
### What is the new behavior?
The [phpstan-strict-rules](https://github.com/phpstan/phpstan-strict-rules) plugin has been enabled.  Amongst other things, PHPStan will now require strict type comparisons and disallow functions which perform loose type comparisons.  This change will automate a large portion of my manual code review work.


ziesski (approved):
I see that it disallow `empty()` but we use isEmpty() instead, so I guess it is fine.

PR #9902, opened 2023-10-18
[Testing:Developer] update phpstan command

ziesski:
PHPStan supports the ability to list files as a parameter in the configuration file. <-Check #9900 for more detail.
Update composer script for static-analysis accordingly.

### DIFF
--- a/site/composer.json
+++ b/site/composer.json
@@ -53,6 +53,6 @@
   "scripts": {
     "test": "phpunit",
     "lint": "phpcs --extensions=php ./app",
-    "static-analysis": "phpstan analyze app public/index.php socket/index.php"
+    "static-analysis": "php vendor/bin/phpstan analyze"
   }
 }


site/composer.json
```diff
@@ -53,6 +53,6 @@
   "scripts": {
     "test": "phpunit",
     "lint": "phpcs --extensions=php ./app",
-    "static-analysis": "phpstan analyze app public/index.php socket/index.php"
+    "static-analysis": "php vendor/bin/phpstan analyze"
```
williamjallen, line 56: Any reason why you can't just use `phpstan analyze` here?

ziesski: Actually, this is related to my lint PR as well. 
Currently the code will run into permission error even as root, to execute it without php interpreter.
For more detail look at
https://stackoverflow.com/questions/38521329/how-to-fix-permission-denied-while-running-vendor-bin-phpunit-in-a-laravel-proje

ziesski:
Replaced by #9675

PR #9890, opened 2023-10-17
[Dependency] Add Luxon date library

oskurtt:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant
* [ ] Screenshots are attached to Github PR if visual/UI changes were made

### What is the current behavior?
We are using JS' default Date Library

### What is the new behavior?
Instead of using the builtin JS date library, we will be using Luxon. This PR only installs the dependency. Other than that, no files are being affected.

PR #9909, opened 2023-10-20
[Refactor:Dates] 9731 new date

### LINKED ISSUE #9731: Improper use of Date strings in JS
Currently in our JavaScript code, there are numerous calls to the Date constructor which pass in a date string.

Ex.) `new Date("02/01/1970 05:00")` => `new Date("1970-02-01T05:00:00")`

oskurtt:

We use the builtin JS date library in a number of places. While this mostly works, we keep getting bit by browser implementation specific stuff and pitfalls because it's not a very developer friendly interface. This is especially a bit hairy as we then do bits of math with these dates, which is equally error prone.
Fixes #9731, however, do NOT close that issue because there are more files that needs to be refactored. 

### What is the new behavior?
Instead, we are using Luxon which abstracts this stuff (as well as takes advantage of the Intl library for better international support). This would also provide the same interface across all browsers so that we're confident we're not going to hit a bug where the date displays properly on Chrome, but not on Safari or FireFox.


williamjallen:
Can you please give this PR a more descriptive title?

Also, it is good to link to related work.  PR https://github.com/Submitty/Submitty/pull/9890 added Luxon, but didn't refactor any of our existing code to use it.  This PR appears to be a follow-up PR which refactors some files to use Luxon.

Youssef-Chip (approved):
I tested the functionality of latedays, extensions and confetti display and they appear to work as expected with the Luxon Library. The code also looks clean

williamjallen:
Can you improve the PR title?  Out of context, the title does not provide much information about what this PR does.

PR #9912, opened 2023-10-20
[Bugfix:Calendar] Fix pronouns cypress test

yanliw123:
fix pronouns cypress test

### DIFF
--- a/site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
+++ b/site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
@@ -5,8 +5,9 @@ describe('Tests cases abut changing user pronouns', () => {
         cy.visit('/user_profile');
         cy.login('student');
 
-        cy.get('#pronouns_val').as('pronounsVal').click(); // Alias for pronouns value
-        cy.get('#user-pronouns-change').as('pronounsInput'); // Alias for pronouns input
+        //open pronouns form
+        cy.get('#pronouns_val').click();
+        cy.get('#user-pronouns-change').as('e');
 
         //type in new pronouns and check display in forum option
         cy.get('button[aria-label="Clear pronoun input"]').click(); //clear input using trash can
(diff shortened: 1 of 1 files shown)


site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -5,8 +5,9 @@ describe('Tests cases abut changing user pronouns', () => {
         cy.visit('/user_profile');
         cy.login('student');
 
-        cy.get('#pronouns_val').as('pronounsVal').click(); // Alias for pronouns value
-        cy.get('#user-pronouns-change').as('pronounsInput'); // Alias for pronouns input
+        //open pronouns form
+        cy.get('#pronouns_val').click();
+        cy.get('#user-pronouns-change').as('e');
```
IDzyre, line 10: I think that the 'refactor cypress tests' PR that was merged before this one changed the variable that this was saved as, and I think that it would be better to keep the pronounsInput variable name, and change the place later in this file to be @pronounsInput instead of @e

IDzyre (changes requested):

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
yanliw123: fixed! I changed it back to @pronounsInput

yanliw123: > I think that the 'refactor cypress tests' PR that was merged before this one changed the variable that this was saved as, and I think that it would be better to keep the pronounsInput variable name, and change the place later in this file to be @pronounsInput instead of @e

fixed! i changed it back to @pronounsInput

IDzyre (approved):

williamjallen:
Can you please update the PR description to include a little bit more information about what was wrong and how you fixed it?  Your PR description is the permanent documentation of this change, and if we continue to have issues with this test it will be useful to be able to look back at this change and know what was wrong and how the issue was fixed without having to look at the diff.

PR #9916, opened 2023-10-21
[Refactor:CourseMaterials] Use one variable for beginning of time

### LINKED ISSUE #9762: Code refactor for Setting the Begining of time
We recently changed the beginning of time feature in the datetime picker in Course Materials to 1900.  
To use the code related to the beginning of time, we have to write out :  ```date = new Date("1900-01-01 00:00:00");```. in five different files. 
We should call a single variable in all of the places where we use the beginning of time code instead of manually writing 1900-01-01 00:00:00 each time.

Youssef-Chip:

The beginning of time feature is used in all date time picker in course materials. We recently changed the date from 1000 to 1900. 
### What is the new behavior?
In case we decide to change it again later, this refactor would allow the developer to change one variable in one file, instead of 5 different ones in 5 different files.
### Other information?
Closes #9762

### DIFF
--- a/site/app/views/course/CourseMaterialsView.php
+++ b/site/app/views/course/CourseMaterialsView.php
@@ -5,6 +5,7 @@ namespace app\views\course;
 use app\entities\course\CourseMaterial;
 use app\libraries\FileUtils;
 use app\views\AbstractView;
+use app\libraries\DateUtils;
 
 class CourseMaterialsView extends AbstractView {
     public function listCourseMaterials(array $course_materials_db) {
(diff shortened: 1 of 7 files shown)


site/app/views/course/CourseMaterialsView.php
```diff
@@ -26,7 +26,7 @@ public function listCourseMaterials(array $course_materials_db) {
         $folder_ids = [];
         $links = [];
         $base_view_url = $this->core->buildCourseUrl(['course_material']);
-
+        $begining_of_time_date = "1900-01-01 00:00:00";
```
williamjallen, line 29: Instead of hardcoding it here, you could add it to our [date utilities class](https://github.com/Submitty/Submitty/blob/main/site/app/libraries/DateUtils.php).

taorif25 (approved):
Code works as intended and will make development easier in the future.

site/app/views/course/CourseMaterialsView.php
```diff
@@ -26,7 +27,7 @@ public function listCourseMaterials(array $course_materials_db) {
         $folder_ids = [];
         $links = [];
         $base_view_url = $this->core->buildCourseUrl(['course_material']);
-
+        $begining_of_time_date = DateUtils::BEGINING_OF_TIME;
```
williamjallen, line 30: Looks good to me.  One minor nitpick you don't necessarily have to change: since you're just using this variable in one place, it might be better to just inline it below.

williamjallen (approved):

PR #9917, opened 2023-10-21
[Feature:Developer] Add notifications for good first issue

IDzyre:

Related to #9903 
Currently when good first issues are posted to the Github, there is no notification to developers that an issue has been posted. If it is not in your workflow to check issues regularly, they can be lost in the hundreds of issues we currently have. 

### What is the new behavior?
A workflow has been added to notify the Zulip channel on the labelling of an issue as a 'good first issue'.  The wording of the content isn't concrete, so if there is something that should be added/removed it is easily changeable. 

image


williamjallen (commented):
Looks reasonable to me.  Should we also provide a link to a filtered list of all of the other good first issues as well?  https://github.com/Submitty/Submitty/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

williamjallen:
After discussing the issue with Barb, it looks like we should wait until the Submitty long-term planning meeting on Tuesday before doing anything further with this PR.  We need to figure out how to better handle good first issues, but it's possible that there are better ways to go about doing it rather than having a Zulip channel for them.

bmcutler:
Yes, let's try it out for a while.  We can modify or remove if it's more annoying than useful.

PR #9922, opened 2023-10-22
[Bugfix:InstructorUI] Fixed incorrect url results

### LINKED ISSUE #9915: Writing incorrect url results in a server error
**Describe the bug**
When the following url is typed:  ```/courses/<term>/<course>/gradeable/<gradeable>/grading/grade```
(Example: http://localhost:1511/courses/f23/sample/gradeable/closed_homework/grading/grade), a server error occurs.

**Expected behavior**
When an incorrect url is usually typed, it redirects to the home page.
Also the url above is only correct when you add a query string at the end, to grade a specific student: 
http://localhost:1511/courses/f23/sample/gradeable/closed_homework/grading/grade?who_id=aphacker

**To Reproduce**
1. Type the url above
2. See error

**Configuration**

**Screenshots**
image
**Additional context**

teazhi:

When the following URL is typed:  ```/courses/<term>/<course>/gradeable/<gradeable>/grading/grade```
(Example: http://localhost:1511/courses/f23/sample/gradeable/closed_homework/grading/grade), a server error occurs.

When an incorrect or invalid URL is typed, it should redirect to the home page.

**Screenshot**
image

### What is the new behavior?

If you do not add a query string at the end to grade a specific student, the page will get redirected to the home page of Submitty.

Correct URL Link Example: http://localhost:1511/courses/f23/sample/gradeable/closed_homework/grading/grade?who_id=aphacker 

### Other information?

closes #9915


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -1676,6 +1676,10 @@ public function showGrading(
             // Get the graded gradeable for the $from user
             $id_from_anon = $this->core->getQueries()->getSubmitterIdFromAnonId($from, $gradeable_id);
             if ($blind_grading !== "unblind" || $anon_mode) {
+                if ($id_from_anon === null) {
+                    header('Location: /home');
+                    exit;
```
williamjallen, line 1681: We do not want to abruptly exit, or set headers like this.  It would be better to use `return new RedirectResponse('/home');` instead.  That said, is this query parameter mandatory?  If so, should it simply be added to the route?

williamjallen (dismissed):

site/app/controllers/grading/ElectronicGraderController.php
Youssef-Chip: "The submitter id " query is mandatory for the specific route to be able to grade that specific student. I think it should return to the home page, because there isn't a default submitter it should go to if no "submitter id" is provided.

Youssef-Chip: Yes, we should update the route then.

williamjallen:
@teazhi It looks like you're still failing the title check action.  Please edit your title to meet our PR title [guidelines](https://submitty.org/developer/getting_started/make_a_pull_request).

ArnavPawar (changes requested):
It looks good and works well with no bugs. 
However, I think you should take it back to the grades page instead of the homepage though, because if they are already on the grades page and reach an error it should only move them back one step in the site.

DarthNyan (changes requested):
Looks good to me.

Ultimately whether this redirects home or to the gradeables page is a design decision, and I think either option is fine. If we later want to change the redirect it'll be an easy fix.

DarthNyan (approved):
Oops I meant to hit approve

VVVITH (approved):
As the suggestion from reviewers, when the wrong url is about missing student id, a redirection to grading details may be better than redirect to home page. Talk to Barb, we agree with the grading detail page choice. Thus I make the changes and tested alright.
Need more reviews and re-reviews, since I make this change so my review should not count.

cjreed121 (approved):

PR #9923, opened 2023-10-22
[Bugfix:Submission] User not found in late day cache

jason490:

Error in late day cache. The cache for the user was trying to be accessed without being created.

### What is the new behavior?
If the user doesn't exist in the cache, then the user gets added to the late day cache, otherwise it continues getting the late day cache.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -1690,10 +1690,9 @@ WHERE term=? AND course=? AND user_id=?",
             }
         }
 
-        $row = $cache[$user_id];
-
+        $row = NULL;
         // If cache doesn't exist, generate it and query again
-        if (empty($row)) {
+        if (!array_key_exists($user_id, $cache) || empty($cache[$user_id])) {
             $params = [$user_id, $g_id];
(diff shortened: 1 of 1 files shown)


bmcutler (approved):
thanks
seems to fix the problem!

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1690,10 +1690,9 @@ public function getLateDayCacheForUserGradeable(string $user_id, string $g_id):
             }
         }
 
-        $row = $cache[$user_id];
-
+        $row = null;
         // If cache doesn't exist, generate it and query again
-        if (empty($row)) {
+        if (!array_key_exists($user_id, $cache) || empty($cache[$user_id])) {
```
williamjallen, line 1695: There's no need to change this now that it has been merged, but just FYI `empty()` does an `isset()` check for you.  The issue is that by setting `$row`, `$cache[$user_id]` is evaluated and an index not found error is thrown.

You could do this more concisely by using `if (empty($cache[$user_id])) {` alone.

ziesski (commented):
Sorry for reviewing late, the change is reasonable.

PR #9925, opened 2023-10-24
[Bugfix:System] Fix session cookie not updating

cjreed121:

When the session gets closer to expiration it will update on the database but not the cookie. Thus the cookie will still expire and log the user out after the session expiration time.  The issue was introduced in https://github.com/Submitty/Submitty/pull/8620.

### What is the new behavior?
The cookie will now update when the session updates. The user will only get logged out after 2 weeks of inactivity.


williamjallen (approved):
Changes look good to me (once you get the tests passing).  Thanks for fixing this!  I encountered this problem on RPI's Submitty instance a few days ago.

PR #9927, opened 2023-10-24
[Bugfix:Forum] Fix syntax error for error handling

shailpatels:

* [] Screenshots are attached to Github PR if visual/UI changes were made (N/A)

If the forum backend sends an invalid JSON response the client-side code to handle it throws an exception due to invalid syntax this prevents the cypress test for forum pronoun changes to fail since it will abort on an exception thrown from our software

### What is the new behavior?
Removes the error allowing cypress to pass that part of the test and prevents the forum from silently putting exceptions in the console if this happens to a real user.

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -1766,8 +1766,8 @@ function deletePostToggle(isDeletion, thread_id, post_id, author, time, csrf_tok
                 }
                 catch (err) {
                     // eslint-disable-next-line no-undef
-                    displayErrorMessage('Error parsing data. Please try again').
-                        return;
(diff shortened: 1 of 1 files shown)


ziesski (approved):
valid syntax change, looks good / but not sure if it fixes pronoun issue

MasterOdin (approved):
Fixed two other syntax errors in this file. I did a grep over the rest of the codebase to see if there were anymore instances of this particular error (`\)\.\n[\s]+return;`).

I've also reworded the title since while this bug was surfaced by Cypress, they are things that could happen in production, so  it's a bugfix, vs just affecting testing efforts.

PR #9928, opened 2023-10-24
[Bugfix:Notifications] Fix flattened email count

cjreed121:

If you make an announcement, a site error will occur due to the flattened email count being divided by the wrong number.

### What is the new behavior?
Making an announcement no longer creates an error and that count is now divided by the right number.

### Other information?
Tested locally and works as expected now.


bmcutler (approved):
thanks! fixed it

Issue #9848, opened 2023-10-04
Manage Student - additional toggle columns options


bmcutler:
**What problem are you trying to solve with Submitty**
From the Manage Students page the instructor can see many columns of data for the student.

The "Toggle Columns" option allows them to remove some columns from display (it can be overly crowded with columns that aren't necessary useful to some instructors/courses).

Additional data about the user is available by clicking "edit student" (user numeric id, legal first name, legal last name, email, secondary email)

It may be helpful to view that additional data for all students at once (rather than click edit student for each student).  E.g. if a student writes their user numeric id on their exam instead of their username, we could search the manage students page for the matching student row.

**Describe the way you'd like to solve this problem**

- [ ] Add 5 more options to the "toggle columns" selector: (user numeric id, legal first name, legal last name, email, secondary email).  These columns should be off/not visible by default.

- [ ] make the "Download Users" option download the data for all currently visible/selected columns.  Currently it is a fixed format and you can't add additional columns (e.g., pronouns are not exported).

ADDED A NEW ITEM 12/1/23

- [ ] let's allow the order of the columns to be rearranged by the instructor.  A draggable interface to order the options similar to the TA grading rubric marks interface and the rainbow grades syllabus categories and gradeable interface would be nice.

sagnik3788:
can i work on this issue @bmcutler

Akshun-01:
Hey, is someone working on this issue? I can try it if no one's currently working on it.

bmcutler:
Hi @sagnik3788 

Thank you for your interest in the Submitty open source project.
We welcome contributions from new developers!
However we do not use the Github issue 'assign' feature for first time prospective contributors.

Please read our documentation on how to to get started with Submitty:
https://submitty.org/developer/getting_started/index
Specifically in setting up your development environment:
https://submitty.org/developer/getting_started/vm_install_using_vagrant
And making a pull request:
https://submitty.org/developer/getting_started/make_a_pull_request
You do not need to be assigned to an issue to create a pull request that will be reviewed by our team and then merged if it appropriately resolves the issue.
Also join our Slack server to discuss technical questions:
https://submitty.org/index/contact

PR #9929, opened 2023-10-24
[Feature:InstructorUI] New Toggle Column Options

hayeonohh:

The columns in the Manage Students page is missing columns for additional data we have on students. The only way to view this information is to press the "Edit Student" button for each individual student.

image

### What is the new behavior?
There are five new column options in the Manage Students page when the Toggle Columns button is pressed. When checked off there will be a new column for the additional data showing the information for all students instead of one. The new options are turned off by default.

image
(Image is cut off, you can scroll left/right to see all the columns on the site.)
image

### Other information?
First task in #9848

### DIFF
--- a/site/app/controllers/admin/UsersController.php
+++ b/site/app/controllers/admin/UsersController.php
@@ -78,8 +78,8 @@ class UsersController extends AbstractController {
         }
         else {
             //Expires 10 years from today (functionally indefinite)
-            if (setcookie('active_student_columns', implode('-', array_fill(0, 17, true)), time() + (10 * 365 * 24 * 60 * 60))) {
-                $active_student_columns = implode('-', array_fill(0, 17, true));
+            if (setcookie('active_student_columns', implode('-', array_merge(array_fill(0, 12, true), array_fill(0, 5, false))), time() + (10 * 365 * 24 * 60 * 60))) {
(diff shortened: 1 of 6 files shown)


VVVITH (changes requested):
I test the columns and the buttons, I works as expected. But as I tested, I find the new added columns are not defaultly checked off(turn on). Which is confilct to your decription. Is the decription wrong or your implement wrong? Please fix it.

VVVITH (changes requested):
The description is now same as you implement. But I find a new problem that the new columns are not center alignment to the title as the columns we have before. I don't think it's very aesthetically pleasing. Could you fix it?
截屏2023-10-31 09 45 50

VVVITH (approved):
Looks good now!

IDzyre:
The accessibility test is failing because the size of the table is smaller (13) than the amount of columns associated with it (17/18)(according to the test)

taorif25 (approved):
Code works as expected and successfully closes the first part of the issue. Columns are toggled off by default and correctly display the information with proper formatting. Approved.

DarthNyan (commented):
Feature works for me but I have one question. Would it be possible to move the new columns so that the legal names are near the preferred names, the numeric id near the user id, and keeping the edit/delete columns on the far right? The current layout of columns feels a little disorganized to me.

bmcutler:
Updated the original issue to add the feature request suggested by @DarthNyan  to allow the columns to be reordered.

PR #9930, opened 2023-10-24
[Bugfix:Forum] Proper editing of top post in a thread

ANS1514:
Before this PR an edit to the top post in a thread would only register in the thread list of the discussion forum when viewing a thread.

Now it will properly edit when outside of the thread and inside of the thread.

I did find another bug within this that existed before where it also deletes the next thread underneath the thread that is being edited.


oskurtt:
- For title, change "BugFix" to "Bugfix" to pass PR Title Check.
- Changes in ForumThreadView.php look good

DarthNyan (approved):
Looks good to me. I was able to reproduce the bug and this did fix it.

ArnavPawar (approved):
The bug does not appear anymore for me. Also, the code does seem to only add fixes for the specific bug only and doesn't affect anything else. Looks good to be pulled to main.

taorif25 (approved):

taorif25 (changes requested):
I've been testing this on my machine using two users, and I'm still getting the error message for the user that isn't editing the post. I'm not sure why that is as it seems like the changes are working for other people. Here's a screenshot:
image

ANS1514:
> I've been testing this on my machine using two users, and I'm still getting the error message for the user that isn't editing the post. I'm not sure why that is as it seems like the changes are working for other people. Here's a screenshot: image

I tested this again with some updates to main, I think that some of the code from PR#9968 (https://github.com/Submitty/Submitty/pull/9968/files) seems to make it work 100% of the time for me. If you want to try it again and see.

taorif25:
Retested and the changes are working as expected now. It looks like some cypress tests are failing now though after you merged main.

taorif25 (approved):
PR passes all tests now, approved!

PR #9931, opened 2023-10-24
[Bugfix:System] Fixes CSV Upload Error

### LINKED ISSUE #9871: Error message when uploading CSV files to the Excused Absence Extension page
**Describe the bug**
When uploading a CSV file on the Excused Absence Extension page, an error message is given. However, the CSV input still modifies the extensions in the expected way.

**Expected behavior**
The CSV should work properly without producing an error message.

**To Reproduce**
1. Go to the Excused Absence Extension Page as an instructor.
2. Upload a CSV file. (see specific csv used below)

**Configuration**
 - OS: windows
 - Browser: edge

**Screenshots**
image

**Additional context**
CSV file used: [CSVnoReason.csv](https://github.com/Submitty/Submitty/files/12861909/CSVnoReason.csv)
Error seems to be produced in line 31 of extensions.js

taorif25:
Closes #9871
Currently, when uploading a CSV file on the excused absence extension page, an error is given due to invalid json being parsed.
This is due to a deprecated code warning in lateController.php for ini_set("auto_detect_line_endings", true);.
image

### What is the new behavior?
Deprecated code is replaced by manual newline parsing, removing the error.

### DIFF
--- a/site/app/controllers/admin/LateController.php
+++ b/site/app/controllers/admin/LateController.php
@@ -342,7 +342,10 @@ class LateController extends AbstractController {
             ];
         }
 
-        $rows = file($csv_file, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
+        $file_content = file_get_contents($csv_file);
+        $file_content = preg_replace("/\r\n|\r/", "\n", $file_content);
+        $rows = explode("\n", $file_content);
+        $rows = array_filter($rows, 'strlen');
         if ($rows === false) {
             $data = null;
             return [
--- a/site/app/controllers/admin/LateController.php
+++ b/site/app/controllers/admin/LateController.php
@@ -343,16 +343,16 @@ class LateController extends AbstractController {
         }
 
         $file_content = file_get_contents($csv_file);
-        $file_content = preg_replace("/\r\n|\r/", "\n", $file_content);
-        $rows = explode("\n", $file_content);
-        $rows = array_filter($rows, 'strlen');
-        if ($rows === false) {
+        if ($file_content === false) {
             $data = null;
             return [
                 "success" => false,
                 "error" => "Could not open CSV file for reading",
             ];
         }
+        $file_content = preg_replace("/\r\n|\r/", "\n", $file_content);
+        $rows = explode("\n", $file_content);
+        $rows = array_filter($rows, 'strlen');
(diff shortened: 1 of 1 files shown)


williamjallen (commented):
Would PHP's [`fgetcsv`](https://www.php.net/manual/en/function.fgetcsv.php) function be cleaner here?

taorif25:
> Would PHP's [`fgetcsv`](https://www.php.net/manual/en/function.fgetcsv.php) function be cleaner here?



> Would PHP's [`fgetcsv`](https://www.php.net/manual/en/function.fgetcsv.php) function be cleaner here?

I looked into it and it seems that fgetcsv would still have the problem of needing auto_detect_line_endings enabled in order to detect \r line endings. This manual parsing solution allows the detection of all line endings without using deprecated auto_detect_line_endings.

taorif25:
> Could you improve the documentation of Submitty by adding a sample format of the csv file (in https://submitty.org/student/submission/late_days), because I made a file and I am still getting an error

Yeah sure. The csv formatted should also be listed on the excused absence extensions page. It should be in the format of student_id, gradeable_id, days_of_extension. Could you link the file that you used?

williamjallen:
> I looked into it and it seems that fgetcsv would still have the problem of needing auto_detect_line_endings enabled in order to detect \r line endings. This manual parsing solution allows the detection of all line endings without using deprecated auto_detect_line_endings. https://www.php.net/manual/en/function.fgetcsv.php

Seems like reasonable logic to me.  I think the current approach is a good alternative.

taorif25:
> > I looked into it and it seems that fgetcsv would still have the problem of needing auto_detect_line_endings enabled in order to detect \r line endings. This manual parsing solution allows the detection of all line endings without using deprecated auto_detect_line_endings. https://www.php.net/manual/en/function.fgetcsv.php
> 
> Seems like reasonable logic to me. I think the current approach is a good alternative.

Is there anything else I should change or do things look good?

Youssef-Chip (approved):
I can't manually test this on my machine because the csv upload feature doesn't work on macbook M1 ( I have made #9942 specific to macbook arm chips). However, I think this PR should get merged.

NiaHeermance (commented):
Since no one tested functionality, I just went ahead and did so. Error is gone and uploading of csv works!

Issue #9934, opened 2023-10-26
Auto-label all new issues with "Awaiting Triage"


bmcutler:
**What problem are you trying to solve with Submitty**

New issues, especially those created by junior developers or external non-developer users may:
* not be tagged with all relevant github labels,
* lack details necessary to reproduce the bug,
* not have a clear and fully-developed design/implementation plan, or
* may not be relevant to the overall mission and future goals of Submitty.

Before a developer undertakes solving a new issue, a senior member of the Submitty development and project management team should review the issue and make any necessary recommendations.  Issues that have not yet been reviewed should be clearly marked.

**Describe the way you'd like to solve this problem**

* Let's create a bot/action to automatically add the "Awaiting Triage" github label to all new issues.

* A senior member of the Submitty development and project management team will aim to promptly review all issues with the "Awaiting Triage" label and after making any necessary adjustments, will remove this label.

* Developers should not start implementation of or make a pull request for any issue which still has the "Awaiting Triage" label.  Developers are welcome to communicate with Submitty development and project management team 
(https://submitty.org/index/contact) if an issue has this label and they are interested in starting work immediately.

PR #9898, opened 2023-10-18
[Documentation:Developer] Automatically tag feature request issues

williamjallen:
### What is the current behavior?
Issues created using the feature request template currently are not currently tagged properly.

### What is the new behavior?
Feature request issues are now tagged appropriately.

PR #9937, opened 2023-10-26
[Feature:Developer] Label new issues as Awaiting Triage

IDzyre:

Implements #9934
Submitty has over 400 issues currently, and some of them are from external contributors, so we are unsure if they are still issues, or even reproducible. 

### What is the new behavior?
A Github workflow has been added, so on opening a new issue, the label 'Awaiting Triage' will be added automatically, and a Submitty Org developer will go through and make sure enough information is in the issue, and review the problems.


williamjallen (commented):
Is there a reason why having an action as opposed to just adding the tag to the default tags in our issue templates is beneficial?

IDzyre:
> Is there a reason why having an action as opposed to just adding the tag to the default tags in our issue templates is beneficial?

I did not realize that there was default tags, so that would be the easier and more efficient option probably...

bmcutler:
If there are default tags, the tag can be removed

williamjallen:
@bmcutler External contributors cannot edit tags.  The ability to modify tags is granted by the "Triage" access level and above.  External contributors also cannot make issues which do not use one of the existing issue templates.  To see what it looks like for people without the Triage access level, take a look at the [PHPStan repository](https://github.com/phpstan/phpstan/issues/new/choose).

bmcutler:
External contributors cannot even choose which of existing set of labels should be applied?  Well that's an unfortunate decision by Github.  Even more reason we need to have someone with access review & correctly label new issues from junior devs and devs outside the organization.

williamjallen:
Contributors who do not have at least the Triage access level cannot add individual tags to their issues, but they can choose from the set of issue templates, which can have default sets of tags.  It doesn't really make sense for Submitty due to our module tagging system, but for something like PHPStan it makes a lot of sense.  PHPStan only has a few categories of issues, and I can't think of many situations where a new contributor would need a set of tags that isn't provided by default.

In addition to being a good motivation for a manual triage step, this is exactly why we should carefully choose our issue templates to appropriately cover the maximum number of cases new contributors might run into.  That's also why I created [this](https://github.com/Submitty/Submitty/pull/9898) PR a few weeks ago.

Should this PR be closed again?

PR #9938, opened 2023-10-26
[Bugfix:InstructorUI] Fix Filter Activity by Poll Count

NiaHeermance:
Filtering by poll count highlights all rows as green.

### What is the new behavior?
Now only rows where the poll count is greater than or equal the specified value are highlighted as green.

### Images
image


site/ts/activity-dashboard.js
```diff
@@ -130,10 +130,13 @@ export function applySettings() {
         const s_grad_sub = data[i].gradeable_submission;
         const s_forum_view = data[i].forum_view;
         const s_forum_post = data[i].forum_post;
-        const s_num_polls = data[i].num_poll_responses;
+        let s_num_polls = data[i].num_poll_responses;
         const s_off_hours = data[i].office_hours_queue;
-        let flag = false;
+        if (s_num_polls === null) {
+            s_num_polls = 0;
+        }
```
williamjallen, line 137: You could use the [null(ish)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) coalescing operator for this.

Nicc041 (approved):
Other than what William said, the code works as expected; the rows are filtered by the numbers of polls answered and turn red or green.

ANS1514 (approved):

bmcutler:
@NiaHeermance 
please update this PR with William's feedback

site/ts/activity-dashboard.js
NiaHeermance: Unfortunately, as we figured out in the meeting yesterday, our linter doesn't recognize that operator. We need to update the linter.

PR #9939, opened 2023-10-26
[Testing:Developer] Notify Zulip on main failing CI

IDzyre:
If we aren't paying close attention, if Main fails CI we may not know it immediately, which can cause other PR's to fail the CI and wonder what is causing it. 

### What is the new behavior?
A workflow has been added, that runs when the Submitty CI workflow is run on the main branch, to notify Zulip if the CI has failed on main.


.github/workflows/submitty_ci.yml
```diff
@@ -520,3 +520,14 @@ jobs:
       - uses: ./.github/actions/e2e-Setup-Composite
       - name: Run integration tests
         run: sudo -E env "PATH=$PATH" python3 /usr/local/submitty/test_suite/integrationTests/run.py
+
+  Notify-Failure:
+    runs-on: ubuntu-latest
+    if: ${{ always() && contains(join(needs.*.result, ','), 'failure') && github.event.pull_request.title }}
+    needs: [css-lint, js-lint, js-unit, php-lint, php-unit, python-lint, python-unit, yaml-lint, ShellCheck, Integration, Cypress-System, Cypress]
```
williamjallen, line 527: Is there a way to say `needs: [*]` or similar so you don't have to specify all of the different jobs?  I didn't see anything like that in my brief look through the docs, but it would be a good change if it is possible to do.

.github/workflows/submitty_ci.yml
```diff
@@ -520,3 +520,14 @@ jobs:
       - uses: ./.github/actions/e2e-Setup-Composite
       - name: Run integration tests
         run: sudo -E env "PATH=$PATH" python3 /usr/local/submitty/test_suite/integrationTests/run.py
+
+  Notify-Failure:
+    runs-on: ubuntu-latest
+    if: ${{ always() && contains(join(needs.*.result, ','), 'failure') && github.event.pull_request.title }}
```
williamjallen, line 526: Do you need something about only the main branch here?  It might be useful to put a few comments describing the logic here in case someone stumbles into this in the future.

IDzyre: Yes, that got lost with my testing earlier... I will add comments too.

.github/workflows/submitty_ci.yml
IDzyre: I don't see one, if I try needs[*] the github actions syntax doesn't allow it.

ziesski (approved):
Looks good, well structured.

williamjallen:
The way this is currently written, the Notify-Failure job shows up as "skipped" in the checks for all PRs.  Is there a different approach you can take which is only triggered on pushes to `main`?  Can workflows depend upon jobs in other workflows?  Is there a way for Zulip to accept web hooks from GitHub directly perhaps?

I'm not certain about what the best approach is.  This approach works, but I wonder if there's a better approach out there somewhere.

ziesski:
> The way this is currently written, the Notify-Failure job shows up as "skipped" in the checks for all PRs. Is there a different approach you can take which is only triggered on pushes to `main`? Can workflows depend upon jobs in other workflows? Is there a way for Zulip to accept web hooks from GitHub directly perhaps?
> 
> I'm not certain about what the best approach is. This approach works, but I wonder if there's a better approach out there somewhere.


Can we use something like if github.event_name = 'push' with github.ref == main? to trigger push only?

williamjallen:
As far as I am aware, that will do essentially the same thing the current setup does.  If any logic in your job runs for any reason, the checks will list the job.

IDzyre:
> The way this is currently written, the Notify-Failure job shows up as "skipped" in the checks for all PRs. Is there a different approach you can take which is only triggered on pushes to `main`? Can workflows depend upon jobs in other workflows? Is there a way for Zulip to accept web hooks from GitHub directly perhaps?
> 
> I'm not certain about what the best approach is. This approach works, but I wonder if there's a better approach out there somewhere.

I can make a workflow that is called when the CI workflow is run on main, would that be a better approach?

williamjallen:
Possibly, but only as long as you can reliably access the status of jobs in other workflows.

williamjallen (commented):
This looks like a reasonable solution to me.  Can you test this by targeting this branch and causing a test failure and then reverting the change before we merge this?

ziesski (approved):
Looks good to me, William's comment should be checked tho :)

IDzyre:
> This looks like a reasonable solution to me. Can you test this by targeting this branch and causing a test failure and then reverting the change before we merge this?

I don't know if these workflows can be run before they are on the main branch. I tested on my fork, and it works correctly.

williamjallen:
> I don't know if these workflows can be run before they are on the main branch. I tested on my fork, and it works correctly.

Ok, I'll trust that you've tested this thoroughly, and we can make any necessary repairs via a separate PR if required.

.github/workflows/notify_main_fail.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  Notify-Failure:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send notification to Zulip on failure
+        if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+        run: |
+            curl -X POST https://submitty.zulipchat.com/api/v1/messages -u ${{ secrets.ZULIP_AUTHENTICATION }} --data-urlencode 'type=stream' --data-urlencode 'to=Submitty Developer Studio' \
+            --data-urlencode 'topic=Main CI Failures' --data-urlencode \
+            'content=The Github Actions CI has failed on the Main branch, View here: [Main CI](https://github.com/Submitty/Submitty/actions/runs/${{ github.event.workflow_run.run_id }})'
```
williamjallen, line 18: Would it be useful to specify which job failed specifically?

PR #9945, opened 2023-10-30
[Refactor:InstructorUI] Move Select2 Widget Code

NiaHeermance:
Throughout Submitty, we repeat the code for creating and styling Select2 widgets.

### What is the new behavior?
Select2 widgets still look the same, but their code and styling has been to a few central files.

### How to test?
Go to a gradeable's edit page, select Submissions/Autograding, and see if you can still choose an autograding configuration successfully.

Go to Excused Absences Extensions on the sidebar, and see if you can still add an excused extension with a reason.

### Images
image

image

image


site/app/templates/admin/Extensions.twig
```diff
@@ -52,6 +52,7 @@
                             <option value="{{ reason }}">{{ reason }}</option>
                         {% endfor %}
                     </select>
+                    <script> registerSelect2Widget("reason-for-exception"); </script>
```
jason490, line 55: I think we should place all the scripts in one area in the twig file instead of adding separate script tags in different places.

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -93,6 +93,7 @@
                         <option value={{path.1}}>{{path.0}}</option>
                     {% endfor %}
                 </select>
+                <script> registerSelect2Widget("autograding_config_selector"); </script>
```
jason490, line 96: Same here

jason490 (changes requested):

jason490 (commented):
Everything works great and has been tested. I will approve after what I said above is addressed.

jason490:
image
This might not be relevant to this pr, but I found this error in the console when I changed the page to mobile view

RitaLei123 (approved):
1. Is the proposed change a good idea?
Yes, it simplifies the code, while keeping the functionality. This makes it easier to read the code as well.

2. Did the author take the right approach?
Yes, Nia replaced the relevant code with Select2 widgets and removed code that might interfere with this. I'm not familiar with the coding languages used but it looks correct (Nia removed the old code and replaced it with Select2 Widget code).

3. Are all the implementation details correct?
Yes, the implementation details are correct.

Manual testing:
1. Reproduce on main: Those features were working before the changes.
2. Check topic branch
3. Reproduce on main again : Those features still work after the changes.

site/app/templates/admin/Extensions.twig
NiaHeermance: That's fair. I was thinking I would put it there because it would be more clear to new developers wanting to implement select2 widgets that you need that script tag for it to work, but you're right that putting it with the other scripts is probably more organized.

NiaHeermance:
> Everything works great and has been tested. I will approve after what I said above is addressed.

Fixed!

jason490 (approved):
Everything looks good

site/app/libraries/SelectWidgets.php
```diff
@@ -0,0 +1,32 @@
...
+class SelectWidgets {
+    /**
+     * Adds the necessary CSS and JS to use a Select2 widget.
+     * Select2 widgets allow you to preselect from options as well as
+     * type your own option.
+     *
+     * @param Core $core The core of the Submitty application we will add CSS and Js to.
+     * @return void
+     */
+    public static function addSelect2WidgetCSSAndJs(Core $core): void {
```
MasterOdin, line 21: Instead of having a static function that requires passing `Core` to it, how about just putting this function into the `Output` class, given that it's highly coupled to how that functions.

bmcutler:
@NiaHeermance 
Please update this PR with MasterOdin's feedback

PR #9951, opened 2023-10-31
[Bugfix:Submission] Warning message for late submissions

RitaLei123:

When a instructor gives only one team member a late day extension, and the other team members don't have an extension, all the other team members will get a zero when the team member with the extension submits.

### What is the new behavior?
When the team member with a late day extension submits, they will be notified with a alert box saying like this:
image

### Other information?

1. Go to the instructor and create a late gradable.
2. Remove all the late days and extensions that exists for the student.
3. Give one student in a team a late day extension and not the rest of them (this can be done by clicking no when asked should all the team members be given a extension).
4. Go to the student with a late day extension and submit something for that assignment.
5. The alert box should pop up upon submission.


Youssef-Chip (approved):
The new changes do effectively handle that edge case. The code also looks clean and the logic makes sense.

taorif25 (changes requested):
I wasn't able to reproduce the bug exactly, but this PR still seems to solve some issues for team extensions. On main, when only one student had an extension, the warning was already showing up for me. However, even when all members of the team had the exception, the warning would still show up. 
Here's a screenshot with both team members having an extension:
image

On this branch, when only one student had an extension, the warning would show up. But when all team members had the extension, the warning would not show up, which is good.
One suggestion would be to use this new logic to get rid of this red warning box if all team member have an extension:
image

taorif25:
> > I wasn't able to reproduce the bug exactly, but this PR still seems to solve some issues for team extensions. On main, when only one student had an extension, the warning was already showing up for me. However, even when all members of the team had the exception, the warning would still show up. Here's a screenshot with both team members having an extension: image
> > On this branch, when only one student had an extension, the warning would show up. But when all team members had the extension, the warning would not show up, which is good. One suggestion would be to use this new logic to get rid of this red warning box if all team member have an extension: image
> 
> This red warning box should appear because in that picture you submitted, the user had already a bad submission. The red box should disappear after you make a valid (late) submission, but it should stay to warn the user that the current version they have is a bad submission.

Ok it turns out there were 2 assignments called closed peer team homework in sample. Red box went away when submitting on the correct assignment!

taorif25 (approved):

PR #9968, opened 2023-11-03
[Bugfix:Forum] Fix edit post from main page

### LINKED ISSUE #9966: Error when editing a post without clicking into it on the discussion forum
**Describe the bug**
On the discussion forum, when attempting to  edit a post from the main discussion forum page, an error message is given and the message isn't updated.

**Expected behavior**
The post should be able to be edited with no errors.

**To Reproduce**
1.  Create a discussion forum post
2.  From the main discussion forum page, attempt to edit the post
3. See error

**Configuration**
 - OS: Windows
 - Browser: Edge

**Screenshots**

image

**Additional context**

Let me know if you're able to reproduce this bug!

DarthNyan:

Closes #9966 . Currently, there is an error when trying to edit a post from the main forum page.

### What is the new behavior?
This fixes the error and allows users to edit posts from the main forum page.

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -2210,19 +2210,20 @@ function updateThread(e) {
     e.preventDefault();
     const cat = [];
     $('input[name="cat[]"]:checked').each(item => cat.push($('input[name="cat[]"]:checked')[item].value));
+    const post_box_id = $('#edit-user-post').find('.thread-post-form').data('post_box_id');
 
     const data =  {
         edit_thread_id: $('#edit_thread_id').val(),
         edit_post_id: $('#edit_post_id').val(),
         csrf_token: $('input[name="csrf_token"]').val(),
         title: $('input#title').val(),
-        thread_post_content: $('textarea#reply_box_0').val(),
+        thread_post_content: $(`textarea#reply_box_${post_box_id}`).val(),
         thread_status: $('#thread_status').val(),
         Anon: $('input#thread_post_anon_edit').is(':checked') ? $('input#thread_post_anon_edit').val() : 0,
         lock_thread_date: $('input#lock_thread_date').text(),
         expirationDate: $('input#expirationDate').val(),
         cat,
-        markdown_status: parseInt($('input#markdown_input_0').val()),
+        markdown_status: parseInt($(`input#markdown_input_${post_box_id}`).val()),
     };
 
     $.ajax({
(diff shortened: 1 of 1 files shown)


taorif25 (approved):
Works as intended, bug is gone on this branch. Code changes are minimal and are isolated to the bug.

site/public/js/forum.js
```diff
@@ -2216,13 +2216,13 @@ function updateThread(e) {
         edit_post_id: $('#edit_post_id').val(),
         csrf_token: $('input[name="csrf_token"]').val(),
         title: $('input#title').val(),
-        thread_post_content: $('textarea#reply_box_').val(),
+        thread_post_content: $('textarea#reply_box_0').val(),
```
ziesski, line 2219: The `.val()` is used to get the current value of selected element, which will return the corresponding value of `reply_box_` so manually inserting a digit should not be the solution.

site/public/js/forum.js
```diff
@@ -2216,13 +2216,13 @@ function updateThread(e) {
...
         thread_status: $('#thread_status').val(),
         Anon: $('input#thread_post_anon_edit').is(':checked') ? $('input#thread_post_anon_edit').val() : 0,
         lock_thread_date: $('input#lock_thread_date').text(),
         expirationDate: $('input#expirationDate').val(),
         cat,
-        markdown_status: parseInt($('input#markdown_input_').val()),
+        markdown_status: parseInt($('input#markdown_input_0').val()),
```
ziesski, line 2225: same thing here

site/public/js/forum.js
DarthNyan: The bug was that '#reply_box' didn't select anything because each reply box has an int appended to their id to distinguish them. The box in the edit post form is always zero (EditPostForm.twig line 13), so the hardcoded number would have worked. I did just make the change to select the post box id and use that though.

DarthNyan:
This shouldn't affect the forum cypress test. At present, it only tests creating, replying, merging, and deleting but not editing or splitting threads as far as I can tell.

IDzyre:
Correct, the forum tests currently only test creating, commenting, merging, and deleting forum posts. This may be a thing you can create an issue on, more tests for the forum so we catch this type of bug before it happens.

ziesski (approved):
Looks good to me :)

PR #9972, opened 2023-11-06
[Bugfix:Developer] Cypress tests use non-changing timezone

DarthNyan:

Currently, profile.spec.js fails because it can't find "(UTC-04:00) America/New_York", due to Daylight Savings Time ending.

### What is the new behavior?
The test now selects for  "(UTC-07:00) America/Phoenix". Arizona doesn't observe Daylight Savings Time, so this should continue to work past March.


jason490 (approved):
Everything looks good and accounted for future tests by using a timezone that doesn't have daylight savings.

PR #9974, opened 2023-11-07
[Feature:Calendar]  global calendar item for all courses

yanliw123:
### What is the new behavior?
Allow Superuser to post Calendar item for all courses
image
### Other information?

1. Login as Superuser
2. post calendar item, should be able to edit/delete the item
3. login as instructor/ta/student, should able to see the item but not able to edit/delete


migration/migrator/migrations/master/20231106203826_superuser_announcement.py
```diff
@@ -0,0 +1,55 @@
...
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    # Create the sequence for superuser_announcements id
+    database.execute(
+    """
```
04vmatsibekker, line 15: It would be preferable to have the database.execute be only called once, which would mean saving the sequel code in a string and then calling the execute function

site/public/js/calendar.js
```diff
@@ -270,6 +306,7 @@ function generateDayCell(year, month, day, curr_view_month, view_semester=false)
     itemList.classList.add('cal-cell-items-panel');
     for (const i in gradeables_by_date[cell_date_str]) {
         itemList.appendChild(generateCalendarItem(gradeables_by_date[cell_date_str][i]));
+        console.log((gradeables_by_date[cell_date_str][i]));
```
04vmatsibekker, line 309: when your PR is finished I don't think console.log should be needed

VVVITH (approved):
Fixed the php-lint check, then all should be good. Feature work as expected, ready to be reviewed and merged if all cypress passed.

site/app/controllers/calendar/CalendarController.php
```diff
@@ -298,4 +301,159 @@ public function deleteMessage(): ResponseInterface {
...
+        if ($announcement === null) {
+            $this->core->addErrorMessage("Announcement not found");
+            return new RedirectResponse($this->core->buildUrl(['calendar']));
+        }
+
+        // Edit the announcement
+        try {
+            $announcement->setText($text);
+            $announcement->setDate(new \DateTime($date));
+            //$calendar_item->setStringType($type);
```
cjreed121, line 412: Can this comment go?

site/app/controllers/calendar/CalendarController.php
```diff
@@ -298,4 +301,159 @@ public function deleteMessage(): ResponseInterface {
...
+        catch (\InvalidArgumentException $e) {
+            $this->core->addErrorMessage($e->getMessage());
+            return new RedirectResponse($this->core->buildUrl(['calendar']));
+        }
+
+        // Persist the changes and flush to save them
+        $this->core->getSubmittyEntityManager()->flush();
+
+        // Disconnect from the master database
+        $this->core->getSubmittyDB()->disconnect();
```
cjreed121, line 423: Don't disconnect

site/app/controllers/calendar/CalendarController.php
```diff
@@ -298,4 +301,159 @@ public function deleteMessage(): ResponseInterface {
...
+    public function deleteGlobalAnnouncement(): ResponseInterface {
+        if (isset($_POST['id'])) {
+            $id = $_POST['id'];
+        }
+        else {
+            $this->core->addErrorMessage("Error: No id specified");
+            return new RedirectResponse($this->core->buildUrl(['calendar']));
+        }
+
+        $this->core->loadMasterDatabase();
```
cjreed121, line 441: Don't load

site/app/controllers/calendar/CalendarController.php
```diff
@@ -298,4 +301,159 @@ public function deleteMessage(): ResponseInterface {
...
+            ->findOneBy(['id' => $id]);
+
+        if ($item === null) {
+            $this->core->addErrorMessage("Announcement not found");
+            return new RedirectResponse($this->core->buildUrl(['calendar']));
+        }
+
+        $this->core->getSubmittyEntityManager()->remove($item);
+        $this->core->getSubmittyEntityManager()->flush();
+        $this->core->getSubmittyDB()->disconnect();
```
cjreed121, line 455: Don't disconnect

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -4873,6 +4873,21 @@ public function getInstructorLevelUnarchivedCourses(string $user_id): array {
...
+     */
+    public function getGlobalCalendarItems(): array {
+        $query = "
+            SELECT id, type, text, date
+            FROM global_calendar_items
+            ORDER BY date DESC
+        ";
+        $this->submitty_db->query($query);
+        return $this->submitty_db->rows();
+    }
```
cjreed121, line 4889: You don't need to make new queries when using Doctrine. Instead fetch them with Doctrine utilities. See other controllers that use doctrine entities for examples.

site/app/models/GlobalCalendarInfo.php
```diff
@@ -0,0 +1,83 @@
...
+                'status_note' => '',
+                'color' => '',
+                'type' => 'item',
+                'date' => $date,
+                'id' => $item->getId()
+            ];
+            $info->items_by_date[$date][] = $curItem;
+        }
+
+        $info->empty_message = "There are currently no assignments posted.  Please check back later.";
```
cjreed121, line 68:
```suggestion
        $info->empty_message = "There are currently no assignments posted. Please check back later.";
```

site/app/templates/calendar/Calendar.twig
```diff
@@ -6,6 +6,8 @@
     const curr_month = {{ curr_month }};
     const curr_day = {{ curr_day }};
     const instructor_courses = {{ instructor_courses|json_encode|raw }};
+    const global_items_by_date = {{global_items_by_date|json_encode|raw }};
+    const is_superuser = {{ is_superuser|json_encode|raw }};
```
cjreed121, line 10: Why is this json encoded and raw?

cjreed121 (changes requested):

JWei17 (approved):
This looks good, but can you allowed superuser to add color to the event? 
imageApproved.

JoJo-ESC (changes requested):
functionality checked. I find it necessary to color the  super user when they use the global event feature. Without some sort of coloring the messages get lost.

dorian451:
Based on our in-person discussion, we should not color the global announcements since they should be not as prominent as course announcements.

dorian451 (approved):
Seems to work

bmcutler (changes requested):
Tested - looks good - except its not clear what is difference about note vs announcement.   
  * apparently notes are italics.  ok...  Asked @dorian451 to add a little documentation about that to the UI when a global calendar item is added.

Reviewed code - sql changes are good -- missed one line comment from @cjreed121 
  * @dorian451 will made handle the final comment.

site/app/templates/calendar/Calendar.twig
dorian451: Fixed with 12cbcd8b445807334be6359623f883723fbe7db6

bmcutler (approved):

bmcutler:
@dorian451 
failing a CI test, please fix.

dorian451:
Should be fixed now

site/app/templates/calendar/Calendar.twig
```diff
@@ -6,6 +6,8 @@
     const curr_month = {{ curr_month }};
     const curr_day = {{ curr_day }};
     const instructor_courses = {{ instructor_courses|json_encode|raw }};
+    const global_items_by_date = {{global_items_by_date|json_encode|raw }};
+    const is_superuser = {{ is_superuser ? "true" : "false" }};
```
IDzyre, line 10: This is inside of a script, so having the const is_superuser doesn't affect the twig later in this file. In a later PR,  would it be better to move this JS script into the calendar.js file?

dorian451: That `const is_superuser` is referenced here: https://github.com/Submitty/Submitty/blob/7f397a530ff73531df6ecae0a68a45808f2e4488/site/public/js/calendar.js#L246

So it is nessesary to have a `is_superuser` variable, at least the way this code is currently set up. We could theoretically move the script to a js file but then we would need a different way to allow the php backend to let `calendar.js` know if the currently logged in user is a superuser or not.

PR #9988, opened 2023-11-14
[Refactor:Forum] Replace iterated database queries

### LINKED ISSUE #9981: Replace iterated queries in discussion forum
**Describe the bug**

Currently, the discussion forum makes a lot of queries from within loops. The result is hundreds of repetitive queries being needed to render the pages.

**Expected behavior**

These should be replaced with broader queries whose results can be iterated over.

**To Reproduce**
1. 
2. 
3. 

**Configuration**

**Screenshots**

**Additional context**

DarthNyan:

Closes #9981 Currently, loading the forum pages require executing loops and querying the database from within the loops. This resulted in hundreds of repetitive queries, slowing the page down.

### What is the new behavior?
Now the forum makes a few queries outside of loops, then iterates over the results. This will make the forum much faster.

### Other information?

The main functions that render the forum are displayThreadList (for the main forum page, lists all the threads) and generatePostList (displays the posts in the current thread). Queries inside the large for loop in displayThreadList have been moved outside, and the main forum page now requires 15 queries to load. generatePostList calls createPost in a loop, so I queried the data needed in createPost outside the loop in generatePostList. The thread pages are now down to 31-32 queries.

Moving these queries also caused some old query functions to never be called, so I deleted those. I also made the change to showThreads so I could delete getResolveState (the data provided is already queried by getThread).


williamjallen (approved):
Seems reasonable to me.  I'm not sure if I tested all of the possible cases given that the changes are pretty complicated and I'm unfamiliar with the forum code, so it would be worth having a couple more people test it out as well.

site/app/views/forum/ForumThreadView.php
```diff
@@ -441,9 +441,13 @@ public function generatePostList($currentThread, $posts, $unviewed_posts, $curre
         $expiring = $activeThread['pinned_expiration'] <= date("Y-m-d H:i:s", strtotime("+7 day"));
 
         $thread_id = $activeThread['id'];
+        if ($thread_id == -1) {
```
cjreed121, line 444:
```suggestion
        if ($thread_id === -1) {
```

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -5016,6 +5055,28 @@ public function getDisplayUserInfoFromUserId($user_id) {
...
+        $this->course_db->query("SELECT * FROM users WHERE user_id IN {$placeholders};", $unique_users);
+        $return = [];
+        foreach ($this->course_db->rows() as $row) {
+            $return[$row['user_id']] = [
+                "given_name" => (isset($row["user_preferred_givenname"]) && strlen($row["user_preferred_givenname"]) > 0) ? $row["user_preferred_givenname"] : $row["user_givenname"],
+                "family_name" => " " . ((isset($row["user_preferred_familyname"]) && strlen($row["user_preferred_familyname"]) > 0) ? $row["user_preferred_familyname"] : $row["user_familyname"]),
+                "user_email" => $row["user_email"],
+                "pronouns" => $row["user_pronouns"],
+                "display_pronouns" => $row["display_pronouns"],
+                "is_staff" => intval($row["user_group"]) <= 3,
```
cjreed121, line 5074: Rather than 3 could we use one of the constants defined in User.php

cjreed121 (commented):
Great work on this. Page definitely feels a lot faster and didn't see any issues. Code looks good. Just some small changes.

PR #9989, opened 2023-11-14
[Refactor:Calendar] Luxon Date Objects

### LINKED ISSUE #9731: Improper use of Date strings in JS

oskurtt:

Fixes #9731
We use the builtin JS date library in a number of places. While this mostly works, we keep getting bit by browser implementation specific stuff and pitfalls because it's not a very developer friendly interface. This is especially a bit hairy as we then do bits of math with these dates, which is equally error prone. Many of the strings passed in do not follow the ISO8601 format and rely on non-standard browser behavior to parse the date.
### What is the new behavior?
Instead of using the builtin JS date library, we will be using Luxon. I have only made this change in latedays.js so the team can review and approve of integrating Luxon for the rest of Submitty.


taorif25 (approved):
Code changes are straightforward and successfully convert the files to luxon dates. I tested calendar features impacted by this branch and everything seems to be working as expected. Looks good!

DarthNyan (approved):
Looks good to me.

The only thing I would change, and this is very minor, but in showEditPostForm you could append the ":00" to validIsoString where it's declared so that it is a valid ISO string on its own. That way you wouldn't need to append stuff to it in line 879. But that's just my personal preference.

williamjallen:
Can you please update the PR title to reflect the fact that you are refactoring a few specific files

bmcutler:
@oskurtt 
please resolve merge conflicts

site/public/js/drag-and-drop.js
DarthNyan, line 1: Changes in this file should probably be in a separate PR. They don't appear to have much to do with Luxon and I don't know what they're supposed to do.

DarthNyan (changes requested):
Calendar and Forum still work, but the changes to drag-and-drop.js don't seem to be related to Luxon.

williamjallen:
@DarthNyan Would you be interested in taking over this PR?  Most of the content seems to be good, but the changes to the file you commented on contain an XSS vulnerability, and are seemingly irrelevant.  You could probably just revert anything that that is broken/unnecessary, and we could merge whatever is left.

DarthNyan (approved):

williamjallen (approved):

cjreed121 (approved):

PR #9990, opened 2023-11-14
[Feature:Developer] auto label abandoned pull requests

ANS1514:
Implemented a new GitHub workflow that runs once a day, it finds PR's that have had no comments/commits for the past 12 days and add a comment under it warning them it will be labeled abandoned without any work within 48 hours. Then if no work is completed, it is labeled abandoned 2 days after.


.github/workflows/label_stale_prs
```diff
@@ -0,0 +1,20 @@
+name: Label Stale PRs
```
cjreed121, line 1: Make sure this file gets a `.yml` file extension.

.github/workflows/label_stale_prs
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  schedule:
+  - cron:  '0 0 * * *'
```
cjreed121, line 5: Add a comment here to say what that cron job means in English

.github/bin/label_abandoned_prs
```diff
@@ -0,0 +1,43 @@
...
+d = datetime.timedelta(days = 14)
+a = today - d
+a = str(a)
+year = int(a[0:4])
+month = int(a[5:7])
+day = int(a[8:10])
+twoWeeks = date(year, month, day)
+
+for x in output:
+    already_warned = 0
```
cjreed121, line 21: Why not use a boolean?

.github/bin/label_abandoned_prs
```diff
@@ -0,0 +1,43 @@
...
+    number = str(x['number'])
+    delta =  prDate - twoWeeks
+    numDays = int(str(delta)[0:2])
+    already_stale = 0
+    already_abandoned = 0
+    for labels in x['labels']:
+        if(labels['name'] == 'Abandoned PR - Needs New Owner'):
+            already_abandoned = 1
+    if((numDays == 2 and already_stale == 0 and already_warned == 0)):
+        subprocess.run(['gh', 'pr', 'comment', number, '--body', 'This PR has been inactive (no commits and no review comments) for 12 days. If there is no new activity in the next 48 hours, this PR will be labeled as Abandoned PR - Needs New Owner and either another developer will finish the PR or it will be closed.'])
```
cjreed121, line 41: What user will this comment as?

ANS1514: github actions bot

.github/bin/label_abandoned_prs
ANS1514: I'll change it to a boolean

.github/bin/label_abandoned_prs.py
```diff
@@ -0,0 +1,42 @@
+import json
+import subprocess
+import datetime
+from datetime import timedelta
+
+pr_json = "gh pr list --json updatedAt,labels,number,comments"
+string = subprocess.check_output(pr_json, shell=True, text=True)
+output = json.loads(string)
+
+for x in output:
```
cjreed121, line 10: Is there a more descriptive variable name than x?

.github/bin/label_abandoned_prs.py
```diff
@@ -0,0 +1,42 @@
...
+            already_warned = True
+    json_time = datetime.datetime.fromisoformat(string.replace('Z', '+00:00'))
+    eastern = datetime.timezone(datetime.timedelta(hours=-5))
+    et_time_update = json_time.astimezone(eastern)
+    today = datetime.datetime.now()
+    two_weeks = timedelta(weeks=2)
+    twelve_days = timedelta(days=12)
+    two_days = timedelta(days=2)
+    et_today = today.astimezone(eastern)
+    tdiff = et_today - et_time_update
```
cjreed121, line 28: Do we need to use a timezone here since all we care about is difference?

cjreed121 (changes requested):

.github/bin/label_abandoned_prs.py
ANS1514: if you don't set both times into the same timezone it causes an error, so I figured it would be easiest to convert both into eastern time

.github/bin/label_abandoned_prs.py
```diff
@@ -4,16 +4,16 @@
...
+terminal_output = subprocess.check_output(pr_json, shell=True, text=True)
+json_output = json.loads(terminal_output)
 
-inactive_comment = 'This\
-                         PR has been inactive (no commits and no review comments)\
-                         for 12 days. If there is no new activity in the next 48 hours,\
-                         this PR will be labeled as Abandoned PR - Needs New Owner and\
-                         either another developer will finish the PR or it will be closed.'
+inactive_comment = "This"\
+" PR has been inactive (no commits and no review comments)"\
```
williamjallen, line 11: You could combine these two lines

cjreed121 (approved):

PR #9993, opened 2023-11-17
[Bugfix:Developer] Fix flakyness of Cypress tests

Youssef-Chip:

There are 3 current bugs with cypress tests: 
Flakyness in edit-gradeable.spec.js
Removing screenrecorded videos when no tests where being run
Flakyness in Docker_ui.spec.js
### What is the new behavior?
This PR fixes the first 2 bugs.

### DIFF
--- a/site/cypress/e2e/Cypress-Gradeable/edit_gradeable.spec.js
+++ b/site/cypress/e2e/Cypress-Gradeable/edit_gradeable.spec.js
@@ -151,7 +151,9 @@ describe('Tests cases revolving around modifying gradeables', () => {
         cy.get('#add-peer-grader').click();
         cy.get('.form-button-container > a').click({ force: true, multiple: true });
 
-        cy.get('#page_5_nav').click();
+        //clicking on page 5 directly, loads the page too fast, which causes flakyness
+        //so we better visit the url
+        cy.visit(['sample', 'gradeable', 'open_peer_homework', 'update?nav_tab=5']);
 
         notBeVisible(['#has_due_date_no'], ['#date_due']);
 
--- a/site/cypress.config.ts
+++ b/site/cypress.config.ts
@@ -1,5 +1,6 @@
 import { defineConfig } from 'cypress'
 import * as fs from 'fs'
+const path = require('path');
 
 export default defineConfig({
   video: true,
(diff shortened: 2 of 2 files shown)


williamjallen:
If you have fixes for two separate issues, the fixes should be made as two separate PRs.

site/cypress/e2e/Cypress-Gradeable/edit_gradeable.spec.js
```diff
@@ -151,7 +151,9 @@ describe('Tests cases revolving around modifying gradeables', () => {
         cy.get('#add-peer-grader').click();
         cy.get('.form-button-container > a').click({ force: true, multiple: true });
 
-        cy.get('#page_5_nav').click();
+        //clicking on page 5 directly, loads the page too fast, which causes flakyness
+        //so we better visit the url
+        cy.visit(['sample', 'gradeable', 'open_peer_homework', 'update?nav_tab=5']);
```
williamjallen, line 156: I believe you could still test the link by first asserting that the link is visible, and then clicking on it.

Youssef-Chip:
> If you have fixes for two separate issues, the fixes should be made as two separate PRs.

The problem is The tests would still fail, so I want a PR that passes all the test

site/cypress/e2e/Cypress-Gradeable/edit_gradeable.spec.js
Youssef-Chip: doesn't cy.visit already do that by default?

IDzyre: Cy.visit() doesn't check the links that are on the page, it just tries to visit a URL.

williamjallen:
> The problem is The tests would still fail, so I want a PR that passes all the test

The flaky tests being addressed here are in two separate test suites though, so it should be clear whether each fix works independently.  We just don't want to block one fix if the other fix runs into a snag during the review process (something like the above comment, for example).

site/cypress/e2e/Cypress-Gradeable/edit_gradeable.spec.js
Youssef-Chip: I tried to put ```cy.get('#page_5_nav').should('be.visible')```  followed by ```cy.get('#page_5_nav').click()``` or         ```cy.get('#page_5_nav').as('page5');  cy.get('@page5').click();``` but it is still flaky

PR #10000, opened 2023-11-18
[UI/UX:System] party duck 

ziesski:
This is to celebrate (10th) birthday of Submitty (more meaningful to have it as #10000 th issue/PR number!) 
party-duck (1)


ArnavPawar (approved):
Just a svg file can be mergerd

VVVITH (approved):
Happy Birthday!

RitaLei123 (approved):
Is the proposed change a good idea?
Yes, it’s a nice anniversary duck.

Did the author take the right approach?
Yes, the author added a party hat and some confetti. 

Are all the implementation details correct?
Yes, the implementation details are correct.

PR #10003, opened 2023-11-20
[Feature:Developer] Specify allowed Mime Types

musaabimran:
Multiple Checks where in place for mime type which could have caused security problem in case the a different mime type is used.

### What is the new behavior?
Created a white list of mime types which can only be used so that different malicious or vulnerable content type is not used.


mimetype.patch
```diff
@@ -0,0 +1,111 @@
+diff --git a/site/app/controllers/MiscController.php b/site/app/controllers/MiscController.php
```
cjreed121, line 1: Don't commit the patch

site/app/controllers/MiscController.php
```diff
@@ -166,15 +167,15 @@ public function displayFile($dir = null, $path = null, $gradeable_id = null, $us
         $corrected_name = pathinfo($path, PATHINFO_DIRNAME) . "/" .  $file_name;
         $mime_type = mime_content_type($corrected_name);
         $file_type = FileUtils::getContentType($file_name);
-        if ($mime_type === "application/pdf" || (str_starts_with($mime_type, "image/") && $mime_type !== "image/svg+xml")) {
+        if (in_array($mime_type, FileUtils::Allowed_Inline_Types)) {
```
cjreed121, line 170: Add `true` as a third argument to specify strict comparison. By default this is `false`.

site/app/controllers/MiscController.php
```diff
@@ -166,15 +167,15 @@ public function displayFile($dir = null, $path = null, $gradeable_id = null, $us
...
             $this->core->getOutput()->useHeader(false);
             $this->core->getOutput()->useFooter(false);
             header("Content-type: " . $mime_type);
             header('Content-Disposition: inline; filename="' . $file_name . '"');
             readfile($corrected_name);
             $this->core->getOutput()->renderString($path);
         }
-        else {
+        elseif (in_array($mime_type, FileUtils::Allowed_Text_Types)) {
```
cjreed121, line 178: Same here

site/app/libraries/FileUtils.php
```diff
@@ -13,6 +13,13 @@ class FileUtils {
     const IGNORE_FOLDERS = [".svn", ".git", ".idea", "__macosx"];
     const IGNORE_FILES = ['.ds_store'];
     const ALLOWED_IMAGE_TYPES = ['jpg', 'jpeg', 'png', 'gif'];
+    const Allowed_Text_Types = [
```
cjreed121, line 16: const vars are usually preferred to be all caps

cjreed121 (changes requested):

micpap25:
I did some testing; all these filetypes work but I've found two separate issues that should be addressed (and probably need some discussion before addressing):
1. the course materials display was updated by this PR but the pop-out for viewing files in current gradable version was not, and that functions incorrectly when an octet stream or any other filetype that shouldn't be viewed inline is popped out in this way. Leading into:
2. Both course materials and current version views use another explicitly defined list of filetypes to choose which files may be popped out and viewed inline. I think these need to be removed and perhaps synched up with the allowed MIME types?
3. 
In other news, the MIME types for most files (even python, java, erlang, etc.) are just "text/html". I'm not certain why this is happening and it's not an issue on our end, it's how the files are interpreted.

micpap25:
Putting this on draft until we fix the existing issues with the implementation of preventing pop-outs based on file extension.

bmcutler:
closing this PR as it needs significant work.
Related to issue #10618

PR #10004, opened 2023-11-20
[Dependency] Bump league/commonmark from 1.6.7 to 2.4.2 in /site

jason490:

We couldn't bump commonmark to 2.4.0 because a no longer maintained a package that wrapped league/commonmark into a twig extension, making it easy to render the input using a filter.

### What is the new behavior?
League/commonmark is bumped to 2.4.0

### How to test?
Go to the discussion forum and make sure the markdown is rendering, so if its wrap in code blocks then the element should be rendered as a code block.
Also, make sure it is also being rendered in the notebook gradeables


benjaminlee755 (commented):
This looks good. Tested the commonmark with bolding, italicizing, titles, and code. Just waiting for the Submitty markdown package to have its first release.

site/app/libraries/Output.php
```diff
@@ -4,19 +4,19 @@
 
 use app\controllers\GlobalController;
 use app\exceptions\OutputException;
-use app\libraries\FileUtils;
+//use app\libraries\FileUtils;
```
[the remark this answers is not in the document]
VVVITH: Talk to Chirs, this line may should be delete since FileUtils and Output.php are in same directory that don't need to 'use'.

VVVITH (approved):
Test feature only, all the code, bold, Italic, url works well with markdown area.
Need more reviews about the code.

ANS1514 (approved):

site/app/libraries/CustomCodeInlineRenderer.php
```diff
@@ -2,24 +2,24 @@
...
         $this->baseRenderer = new CodeRenderer();
     }
 
-    public function render(AbstractInline $inline, ElementRendererInterface $htmlRenderer) {
-        $element = $this->baseRenderer->render($inline, $htmlRenderer);
+    public function render(Node $node, ChildNodeRendererInterface $childRenderer) {
+        $element = $this->baseRenderer->render($node, $childRenderer);
         $attrs = [
-            "class" => "inline-code"
+            "class" => "inline-code hljs"
```
cjreed121, line 22: Why put `hljs` here?

cjreed121 (changes requested):
Code looks good. Seems to work fine. One class might be removable though before merge.

site/app/libraries/CustomCodeInlineRenderer.php
jason490: It adds the surrounding inline code to be darkened. Should we place this in another PR because it was there before I took over this PR.

powe97:
Resolved complicated merge conflicts on this, hopefully correctly.

powe97:
The php-unit fail

> PHP Fatal error:  Class PHPStan\Parser\PhpParserDecorator contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (PhpParser\Parser::getTokens) in phar:///home/runner/work/Submitty/Submitty/site/vendor/phpstan/phpstan/phpstan.phar/src/Parser/PhpParserDecorator.php on line 11

appears to be due to PHPStan shipping with PHP-Parser v4 instead of PHP-Parser v5, which we can't do much about. See also phpstan/phpstan#10620 and https://github.com/phpstan/phpstan-src/commit/9dac90dfd5fc809dd84fdac0532ecd51c6ae66bf.

powe97:
Tried to fix the CI fail but didn't work. This PR appears to be blocked by #10674.

bmcutler:
superseded by #11661

PR #10036, opened 2023-12-01
[Bugfix:RainbowGrades] Remove invalid display options

ziesski:
Currently there are invalid checkboxes in web based customization page, which creates errors (OOPS).
Indeed, this is a bad implement that leads a lot of confusion, because we are currently misusing/handling "display options" and "token" 

Display options includes following
` ["instructor_notes", "exam_seating", "academic_sanction_details" ("moss_details"),   "final_grade", "grade_summary", "grade_details", "display_rank_to_individual"(no longer used)]`

Tokens, includes following
`[messages, bonus_latedays, gradeables, final_cutoff, manual_grade, independentstudy, audit, earned_late_days]` etc

I have no idea why I tried to handle two separate entities at the same time. 

I think the correct way to handle this is

Create entry section for all tokens (just like we did for plagiarism)
Optionally we can keep the idea of checkbox to hide/appear those entry sections.

Anyhow, since we have instructors facing trouble, this would be initial fix

### DIFF
--- a/site/app/models/RainbowCustomizationJSON.php
+++ b/site/app/models/RainbowCustomizationJSON.php
@@ -28,20 +28,14 @@ class RainbowCustomizationJSON extends AbstractModel {
     private $gradeables = [];
 
     // The order of allowed_display and allowed_display_description has to match
-    const allowed_display = ['grade_summary', 'grade_details', 'benchmark_percent',
-        'exam_seating', 'section', 'messages', 'warning', 'final_grade', 'manual_grade', 'final_cutoff', 'instructor_notes'];
+    const allowed_display = ['grade_summary', 'grade_details',
+        'exam_seating', 'final_grade', 'instructor_notes'];
 
     const allowed_display_description = [
         "Display a column(row) for each gradeable bucket on the syllabus.", //grade_summary
         "Display a column(row) for each gradeable within each gradeable bucket on the syllabus.", //grade_details
-        "not used", //benchmark_percent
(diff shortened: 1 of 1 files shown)


site/app/models/RainbowCustomizationJSON.php
```diff
@@ -28,20 +28,14 @@ class RainbowCustomizationJSON extends AbstractModel {
     private $gradeables = [];
 
     // The order of allowed_display and allowed_display_description has to match
-    const allowed_display = ['grade_summary', 'grade_details', 'benchmark_percent',
-        'exam_seating', 'section', 'messages', 'warning', 'final_grade', 'manual_grade', 'final_cutoff', 'instructor_notes'];
+    const allowed_display = ['grade_summary', 'grade_details',
+        'exam_seating', 'final_grade', 'instructor_notes'];
```
williamjallen, line 32: If you format arrays such that every field is on a separate line, the diff is much easier to read when changes are made in the future.  Not a requested change--just a general suggestion.
```suggestion
    const allowed_display = [
        'grade_summary',
        'grade_details',
        'exam_seating',
        'final_grade',
        'instructor_notes',
    ];
```

taorif25 (commented):
Code looks good and it successfully removes the not implemented / invalid options for rainbow grades customization. Just curious, are these features eventually going to be added back to the site or are they not being implemented? Also, why were the invalid options there to begin with?

ziesski:
After conversation with Barb, this PR is replaced by /Submitty/RainbowGrades/66

PR #10037, opened 2023-12-02
[UI/UX:System] menorah duck

ziesski:
We strive to be inclusive and welcoming to all communities. 
Here, in effort to embrace and celebrate the diversity of religious traditions, we are adding Hanukkah, the festival of lights, a Jewish holiday. 

We are trying something new, menorah duck consists of 8 images, and the image will change adding lighting additional candle each day. 

menorah-new


ziesski:
Note that this file does not have rudolph duck, but I have added "random" option, so this is a dependent PR to #9999 so merge 9999 first before merging this.

bmcutler (approved):
A future PR will update the calculations for start date of Hanukkah to work for every year.

PR #10046, opened 2023-12-07
[Feature:InstructorUI] Course Material Tracking

### LINKED ISSUE #10005: Record user_id & timestamp of course material upload & last course material edit
We have no history of the date and author of course material uploads & edits

Let's add 4 new columns to the course database course_materials table a few extra columns....

uploaded_by - character varying - foreign key to users table user_id
upload_date - timestamp with timezone

last_edit_by - character varying - foreign key to users table user_id
last_edit_date - timestamp with timezone

The migration to add this information should set all of these columns to be null for existing courses.  

When a new course material is uploaded, the last_edit columns are NULL.
When a course material is edited, we overwrite any data that might be in the last_edit_by columns.  (We don't store the full history of edits to the course material, only the last edit.)

sahilsuman933:
fixes #10005 

=> Currently we have no history of the date and author of course material uploads & edits.

### What is the new behavior?
=> Added uploaded_by, upload_date, last_edit_by and last_edit_date to the table course_material.
 

This PR is lacking detailed information. 

스크린샷 2024-06-18 오후 1 39 40

With the updated changes, the columns are well defined, and time is set accordingly. 

In above screenshot, 
The `-04` at the end of the timestamp `2024-06-18 00:57:09-04` indicates the timezone offset from Coordinated Universal Time (UTC). Basically, timezone that is 4 hours behind UTC. 

### Other information?
=> None.


xixvtt:
hi @sahilsuman933  I'm new to this project, I've been trying this issue but didn't really know how to use migration to make changes to db, is it possible to ask you a few questions on this? many thanks

cjreed121:
@sahilsuman933 if you still need help, join our Zulip and ask there so we can better assist you. Go [here](https://submitty.org/index/contact) for instructions to join.

migration/migrator/migrations/course/20231206173634_course_materials.py
```diff
@@ -0,0 +1,42 @@
...
+    :type course: str
+    """
+    
+    # Add columns to the course_materials table.
+    database.execute("ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS uploaded_by character varying(255) REFERENCES users(user_id) DEFAULT NULL;")
+    database.execute("ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS upload_date TIMESTAMP WITH TIME ZONE DEFAULT NULL;")
+    database.execute("ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS last_edit_by character varying(255) REFERENCES users(user_id) DEFAULT NULL;")
+    database.execute("ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS last_edit_date TIMESTAMP WITH TIME ZONE DEFAULT NULL;")
+
+    database.execute("UPDATE course_materials SET last_edit_by = uploaded_by, last_edit_date = upload_date;")
```
williamjallen, line 24: What if this up migration was already run once?  Wouldn't this overwrite any values?

williamjallen (commented):
It looks like you've added the requested columns, but they aren't hooked up to anything yet.  Do you intend to implement that part in this PR?  If you're currently working on this PR, you can convert it to a "draft" so other developers know that it isn't ready for review yet.

migration/migrator/migrations/course/20231206173634_course_materials.py
sahilsuman933: It will overwrite the values and to prevent this I converted this into a single transaction. So, if the first column already exists then the update command will not run. 

I have done this in the latest commit.

sahilsuman933:
> It looks like you've added the requested columns, but they aren't hooked up to anything yet. Do you intend to implement that part in this PR? If you're currently working on this PR, you can convert it to a "draft" so other developers know that it isn't ready for review yet.

I misread the issue, I thought the only thing that I had to do was to create new columns. I will update it in my next commit meanwhile I will draft this PR. Sorry for the inconvenience.  

@williamjallen

migration/migrator/migrations/course/20231206173634_course_materials.py
williamjallen: I believe the code does the exact same thing still.  The `IF EXISTS` has no effect on anything other than the `ALTER TABLE` statement.  Everything in a single migration is run in a single transaction, but that doesn't' really matter here.  A better solution is to set the value to the `upload_date` value if the column is null.

sahilsuman933: Okay. I will update it when I will push the changes related to it when I will complete this PR.

sahilsuman933:
@williamjallen, could you please review this updated commit? I have incorporated the necessary data and removed the unnecessary Update query, as I have already set the default value to NULL.

migration/migrator/migrations/course/20231206173634_course_materials.py
```diff
@@ -0,0 +1,40 @@
...
+    :type course: str
+    """
+    
+    # Add columns to the course_materials table.
+    database.execute("""
+                     ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS uploaded_by character varying(255) REFERENCES users(user_id) DEFAULT NULL;
+                     ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS upload_date TIMESTAMP WITH TIME ZONE DEFAULT NULL;
+                     ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS last_edit_by character varying(255) REFERENCES users(user_id) DEFAULT NULL;
+                     ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS last_edit_date TIMESTAMP WITH TIME ZONE DEFAULT NULL;
+                     UPDATE course_materials SET last_edit_by = uploaded_by, last_edit_date = upload_date;
```
williamjallen, line 24: This `UPDATE` command will still overwrite any existing data if the table already exists.  You need to add a `WHERE` clause to only update if the field is null.  Also, should the new columns be non-null, given that you're essentially setting them to default values here?

williamjallen (changes requested):

migration/migrator/migrations/course/20231206173634_course_materials.py
sahilsuman933: In previous commits, I mistakenly redid the migration function because the `UPDATE` query was still present. However, we do not need that update query.

The implementation we are aiming for is that during the migration of the database, the default value for new columns, for which some data already exists, will be set to `NULL`. 

When new course material is added, the upload-related column will be populated with the data. Additionally, when the course is edited, the last_edit related column will be populated. This is the desired behavior from the website.

In the latest commit, all of these tasks are now performed.

thegame615 (commented):
I can confirm that this PR does add the columns as promised, and I am pretty sure it conforms to all of williamjallen's requested changes. It is worth mentioning however, that the original issue [https://github.com/Submitty/Submitty/issues/10005](#10005) has not been fully addressed yet. As far as I can see, none of the course materials history data is being stored yet, but the database is setup to allow that now. Although the changes good and are a movement in the right direction, I am unsure whether I should approve this yet, as I don't see any reason for it to be merged until it implements the feature specified by the issue.

sahilsuman933:
Thank you for reviewing my PR, @thegame615. I will also implement the feature to have a course material history section in few days.

sahilsuman933:
Now that I think about it, I have already implemented the course history feature mentioned in the original issue. I forgot about it because it was a long time ago when I made those changes. Now that I am reviewing my changes, it has already been implemented and everything seems to be working fine.

I have attached screenshots for your reference:

No course file is uploaded:
image
Uploaded a basic course file:
image
Changed the name of the course file:
image


// @thegame615

site/app/entities/course/CourseMaterial.php
```diff
@@ -72,7 +72,19 @@ class CourseMaterial {
...
+    #[ORM\Column(type: Types::DATETIMETZ_MUTABLE, nullable: true)]
+    protected ?DateTime $upload_date;
+
+    #[ORM\Column(type: Types::STRING, nullable: true)]
+    protected ?string $last_edit_by;
+
+    #[ORM\Column(type: Types::DATETIMETZ_MUTABLE, nullable: true)]
+    protected ?DateTime $last_edit_date;
+
+    public function __construct(int $type, string $path, \DateTime $release_date, bool $hidden_from_students, float $priority, ?string $url, ?string $title, ?string $uploaded_by, ?\DateTime $upload_date, ?string $last_edit_by, ?\DateTime $last_edit_date) {
```
NiaHeermance, line 87: Since you're editing this function's header, it would be good  to add a function level comment, as we're slowly trying to comment old functions.

NiaHeermance (commented):
Looks good! One thing I noticed is that if a course has 0 course materials, then when you add the very first course material (and you put that material in a subfolder, though the issue may be reproduceable if the course material goes to the base folder), the course material has no edit date. For every initial upload of a material after the first one, though, the edit date is nonnull as desired.

site/app/entities/course/CourseMaterial.php
sahilsuman933: I will update the function header as soon as possible. Currently, I'm occupied with some other work.

migration/migrator/migrations/course/20231206173634_course_materials.py
```diff
@@ -0,0 +1,37 @@
...
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    
+    # Add columns to the course_materials table.
+    database.execute("ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS uploaded_by character varying(255) REFERENCES users(user_id) DEFAULT NULL;")
+    database.execute("ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS upload_date TIMESTAMP WITH TIME ZONE DEFAULT NULL;")
+    database.execute("ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS last_edit_by character varying(255) REFERENCES users(user_id) DEFAULT NULL;")
+    database.execute("ALTER TABLE course_materials ADD COLUMN IF NOT EXISTS last_edit_date TIMESTAMP WITH TIME ZONE DEFAULT NULL;")
```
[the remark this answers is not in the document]
sahilsuman933: What kind of constraint should I apply to ensure this column is not empty? Currently, it seems acceptable for this column to be empty, particularly when we first upload a file. This is because no one has edited this information and the default value is null. An alternative approach could be to update these columns during file upload, forcing this field to be non-empty.

site/app/entities/course/CourseMaterial.php
sahilsuman933: Added the comment for the function header @NiaHeermance.

site/app/entities/course/CourseMaterial.php
```diff
@@ -118,6 +148,22 @@ public function getReleaseDate(): \DateTime {
...
+        return $this->upload_date;
+    }
+
+    public function getLastEditBy(): string {
+        return $this->last_edit_by;
+    }
+
+    public function getLastEditDate(): \DateTime {
+        return $this->last_edit_date;
+    }
```
williamjallen, line 165: All of these could return null, right?

site/app/entities/course/CourseMaterial.php
```diff
@@ -81,8 +108,11 @@ public function __construct(int $type, string $path, \DateTime $release_date, bo
         $this->sections = new ArrayCollection();
         $this->url = $url;
         $this->title = $title;
+        $this->uploaded_by = $uploaded_by;
+        $this->upload_date = $upload_date;
+        $this->last_edit_by = $last_edit_by;
+        $this->last_edit_date = $last_edit_date;
     }
-
```
williamjallen, line 85: This whitespace change should be reverted.

site/tests/app/entities/course/CourseMaterialTester.php
williamjallen, line 1: Can you add some tests for the methods you added?

williamjallen (changes requested):
Apologies for the very delayed review!  Things appear to be in good shape for the most part, but I left comments on a few things which need minor changes.

site/app/entities/course/CourseMaterial.php
sahilsuman933: Yeah.

site/tests/app/entities/course/CourseMaterialTester.php
sahilsuman933: I have covered all the required cases. Can you be more specific about the what exactly do I need to add here?

site/app/entities/course/CourseMaterial.php
sahilsuman933: Sure.

site/app/entities/course/CourseMaterial.php
williamjallen: Null should be added to the return type then.

site/tests/app/entities/course/CourseMaterialTester.php
williamjallen: You've added the necessary parameters to the constructor, but it would be good to add specific test cases for the getters and setters you've created, such as `setUploadedBy()`.  Given that the getters and setters are simple, you just need to create a basic test which gets and sets each new value.

site/app/entities/course/CourseMaterial.php
sahilsuman933: It can be either of two types: DateTime or Null. Should I add both?

site/tests/app/entities/course/CourseMaterialTester.php
sahilsuman933: Got it.

site/app/entities/course/CourseMaterial.php
sahilsuman933: @williamjallen any update on it?

williamjallen: You can mark the return type as a nullable DateTime with `?DateTime`.  Alternatively, you could use an equivalent union type like `DateTime|null`.

migration/migrator/data/course_tables.sql
```diff
@@ -714,7 +714,11 @@ CREATE TABLE public.course_materials (
     hidden_from_students boolean,
     priority double precision NOT NULL,
     url text,
-    title character varying(255)
+    title character varying(255),
+    uploaded_by character varying(255) DEFAULT NULL::character varying,
+    upload_date timestamp with time zone,
```
ziesski, line 719: 스크린샷 2024-06-11 오전 10 22 15

I am getting frog robot when I visit http://localhost:1511/courses/s24/sample/course_materials
I did try submitty_install & submitty_install clean
Could you let me know if you can reproduce the issue?

ziesski (changes requested):

migration/migrator/data/course_tables.sql
sahilsuman933: Let me reinstall it and test it out.

sahilsuman933: Fixed.

ziesski:
This PR is lacking detailed information. 

스크린샷 2024-06-18 오후 1 39 40


With the updated changes, the columns are well defined, and time is set accordingly. 

In above screenshot, 
The `-04` at the end of the timestamp `2024-06-18 00:57:09-04` indicates the timezone offset from Coordinated Universal Time (UTC). Basically, timezone that is 4 hours behind UTC.

site/tests/app/entities/course/CourseMaterialTester.php
```diff
@@ -64,4 +72,40 @@ public function testLinkNoSections() {
...
+        $details = [
+            'type' => CourseMaterial::LINK,
+            'path' => '/tmp/file.txt',
+            'release_date' => new \DateTime('9998-01-01 05:00:00'),
+            'hidden_from_students' => true,
+            'priority' => 2.4,
+            'uploaded_by' => "Dummy Uploader",
+            'uploaded_date' => new \DateTime('9998-01-01 05:00:00'),
+            'last_edit_by' => "Dummy User",
+            'last_edit_date' => new \DateTime('9998-01-01 05:00:00'),
```
ziesski, line 86: it looks like edit_date is not updated when it is first uploaded. ie. edit_date is null when we first upload file. I discussed with Barb about it, and it is fine. 
But uploaded_date and last_edit_date cannot be same, with the current implementation. So can you just fix the last_edit_date to some time after uploaded_date?

ziesski (approved):
Looks good to me, trivial fix needed, but tested. I wonder if we would display it to course materials in next PR.

williamjallen (commented):
Adding on to @ziesski's comments, and after discussion in person, it seems like adding a database constraint as well as a PHP constraint would help prevent potential bugs in the future.

PR #10048, opened 2023-12-07
[Bugfix:InstructorUI] Upload Validation Error

### LINKED ISSUE #9986: Bulk Upload w/ wrong number of pages fails with no error message
**Describe the bug**
If the instructor bulk uploads a scanned exam and makes a mistake with the page count (or the scanned exam packet has missing or extra pages causing the page count to be wrong), the upload process fails without any error or indication that there was a problem.

For example, upload a pdf with 64 pages (8 students x 8 pages per student) and say that the per exam count is 6 pages.  

We can see the error in the daemon jobs logs, but this is not visible on the website:

```
sudo systemctl status submitty_daemon_jobs_handler

Nov 14 13:30:30 submitty-srv2 submitty_daemon_jobs.py[1454094]: quiz1a_20231018144519124.pdf not divisible by 6
```

**Expected behavior**
This should give an error since the pdf page count is not a multiple.  Instead there is no error on the website.

sahilsuman933:

Fixes #9986 

If the instructor bulk uploads a scanned exam and makes a mistake with the page count (or the scanned exam packet has missing or extra pages causing the page count to be wrong), the upload process fails without any error or indication that there was a problem.

For example, upload a pdf with 64 pages (8 students x 8 pages per student) and say that the per exam count is 6 pages.

We can see the error in the daemon jobs logs, but this is not visible on the website:

```
sudo systemctl status submitty_daemon_jobs_handler

Nov 14 13:30:30 submitty-srv2 submitty_daemon_jobs.py[1454094]: quiz1a_20231018144519124.pdf not divisible by 6
 
```
### What is the new behavior?

It shows an error if we provide an invalid Split by Page count. 

image

### Other information?
=> I was trying to fix this issue by executing a shell command (exec or related commands), but since this functionality is disabled in PHP due to security reason, I used a Python script via CGI to validate whether the file is divisible by num_page.


sahilsuman933:
@williamjallen, could you please tell me what the optimal solution is for this issue? While I was working on it, I found only one way to fix it, which was to either use an external library or utilize the CGI method. Since PyPDF2 was already being used in the Python code, I chose to go with the CGI method. This allowed me to avoid adding any additional dependencies, which is always preferable to minimize dependencies as much as possible.

site/app/controllers/student/SubmissionController.php
```diff
@@ -486,6 +486,22 @@ public function ajaxBulkUpload($gradeable_id) {
 
                 $bulk_upload_job  = "/var/local/submitty/daemon_job_queue/bulk_upload_" . $uploaded_file["name"][$i] . ".json";
 
+                // exec() and similar functions are disabled by security policy,
+                // so we are using a python script via CGI to validate whether file is divisible by num_page or not.
+                $pdf_full_path = $pdf_path . "/" . $job_data["timestamp"] . "/" . $job_data["filename"];
```
04vmatsibekker, line 491: use FileUtils::joinPaths instead I think

04vmatsibekker (changes requested):

site/app/controllers/student/SubmissionController.php
sahilsuman933: Okay. I will make this change and push the changes as soon as possible.

sahilsuman933: Done. @04vmatsibekker

site/app/controllers/student/SubmissionController.php
```diff
@@ -486,7 +486,23 @@ public function ajaxBulkUpload($gradeable_id) {
...
-                $bulk_upload_job  = "/var/local/submitty/daemon_job_queue/bulk_upload_" . $uploaded_file["name"][$i] . ".json";
+                $bulk_upload_job  = FileUtils::joinPaths("/var/local/submitty/daemon_job_queue/bulk_upload_", $uploaded_file["name"][$i], ".json");
+
+                // exec() and similar functions are disabled by security policy,
+                // so we are using a python script via CGI to validate whether file is divisible by num_page or not.
+                $pdf_full_path = $pdf_path . "/" . $job_data["timestamp"] . "/" . $job_data["filename"];
+                $ch = curl_init();
+                curl_setopt($ch, CURLOPT_URL, $this->core->getConfig()->getCgiUrl() . "pdf_page_check.cgi?pdf_path={$pdf_full_path}&num_page={$num_pages}&file_name={$job_data['filename']}");
+                curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+                $output = curl_exec($ch);
```
ziesski, line 497:
```suggestion
                $output = curl_exec($ch);
                curl_close($ch);
```

ziesski (changes requested):

site/app/controllers/student/SubmissionController.php
```diff
@@ -486,7 +486,24 @@ public function ajaxBulkUpload($gradeable_id) {
...
                 ];
 
-                $bulk_upload_job  = "/var/local/submitty/daemon_job_queue/bulk_upload_" . $uploaded_file["name"][$i] . ".json";
+                $bulk_upload_job  = FileUtils::joinPaths("/var/local/submitty/daemon_job_queue/bulk_upload_", $uploaded_file["name"][$i], ".json");
+
+                // exec() and similar functions are disabled by security policy,
+                // so we are using a python script via CGI to validate whether file is divisible by num_page or not.
+                $pdf_full_path = FileUtils::joinPaths($pdf_path, $job_data["timestamp"], $job_data["filename"]);
+                $ch = curl_init();
+                curl_setopt($ch, CURLOPT_URL, $this->core->getConfig()->getCgiUrl() . "pdf_page_check.cgi?pdf_path={$pdf_full_path}&num_page={$num_pages}&file_name={$job_data['filename']}");
```
ziesski, line 495: could we use `rawurlencode` for filename and pdf_full_path? maybe even for num_pages, URL encoding is good practice to handle unsafe characters.

ziesski (changes requested):

site/app/controllers/student/SubmissionController.php
ziesski: I think there are some example use of `rawurlencode` in our codebase. Feel free to use them as references.

sahilsuman933: Sure.

sahilsuman933: Done.

ziesski (approved):
I made minor change on displaying the error message. 
Instead of using alert, now using home-made `displayErrorMessage` function. 

스크린샷 2024-06-18 오전 11 36 32

I think once php-lint is fixed it is good to merge. I tested it, and works.

site/app/controllers/student/SubmissionController.php
```diff
@@ -360,7 +360,7 @@ public function ajaxBulkUpload($gradeable_id) {
             return $this->uploadResult("Invalid gradeable id '{$gradeable_id}'", false);
         }
 
-        $num_pages = $_POST['num_pages'];
+        $num_pages = rawurlencode($_POST['num_pages']);
```
williamjallen, line 363: Why is this necessary?

site/app/controllers/student/SubmissionController.php
```diff
@@ -486,7 +486,24 @@ public function ajaxBulkUpload($gradeable_id) {
                     "is_qr"     => false
                 ];
 
-                $bulk_upload_job  = "/var/local/submitty/daemon_job_queue/bulk_upload_" . $uploaded_file["name"][$i] . ".json";
+                $bulk_upload_job  = rawurlencode("/var/local/submitty/daemon_job_queue/bulk_upload_". $uploaded_file["name"][$i]. ".json");
```
williamjallen, line 489: Same here?

site/cgi-bin/pdf_page_check.cgi
williamjallen, line 1: Is there a reason why this can't be done in PHP?

williamjallen (changes requested):

site/app/controllers/student/SubmissionController.php
sahilsuman933: I added the rawurlencode as per the suggestion of @ziesski. The reason for adding it was that URL encoding is a good practice to handle unsafe characters.﻿

site/cgi-bin/pdf_page_check.cgi
sahilsuman933: It can be done, and it was done directly in PHP by installing another package.﻿ 

https://github.com/Submitty/Submitty/pull/10053

sahilsuman933: I just didn't do it because I didn't want to introduce a new dependency just for this functionality.

Issue #10051, opened 2023-12-08
Rainbow grades does not match "My Late Days" page for bulk upload gradeables


williamjallen:
**Describe the bug**
Submissions were made to a gradeable via bulk upload three days after the due date.  Rainbow grades displayed three stars next to the gradeable, but the "My Late Days" page did not display a row for the bulk upload gradeable.  As a result, there was an inconsistency between what the "My Late Days" page showed and my true number of late days.

**Expected behavior**
This issue needs further discussion.  One possible solution is to only display bulk upload gradeables on the "My Late Days" only if a nonzero number of late days were used.

RuchitTripathi:
[Update-"Late-Days"-page-to-show-late-days-used-for-bulk-upload-gradeables-#10051](https://github.com/RuchitTripathi/Submitty/tree/Update-%22Late-Days%22-page-to-show-late-days-used-for-bulk-upload-gradeables-%2310051) branch on [RuchitTripathi/Submitty](https://github.com/RuchitTripathi/Submitty).                                                                                                
 Hey checkout this branch. I hope it resolves the issue and let me know if there are more specific updates.

manish2022-creator:
I want to solve this issue. Is it is still open so that I can start my work?

bmcutler:
Hi @manish2022-creator 

Thank you for your interest in the Submitty open source project.
We welcome contributions from new developers!
However we do not use the Github issue 'assign' feature for first time prospective contributors.

Please read our documentation on how to to get started with Submitty:
https://submitty.org/developer/getting_started/index
Specifically in setting up your development environment:
https://submitty.org/developer/getting_started/vm_install_using_vagrant
And making a pull request:
https://submitty.org/developer/getting_started/make_a_pull_request
You do not need to be assigned to an issue to create a pull request that will be reviewed by our team and then merged if it appropriately resolves the issue.
We also encourage you to join our Zulip server to discuss technical questions:
https://submitty.org/index/contact

ziesski:
During today's discussion there was opinion that bulk upload should not have due date, which is reasonable. Bulk upload is used to upload paper exams/worksheets to students uploaded by instructor. (make sure to keep release date)

PR #10052, opened 2023-12-08
Update "My Late Days" page to show late days used for bulk upload gradeables. #10051

RuchitTripathi:
In the calculateLateDays function, I added a check to see if the selected gradeable is a bulk upload and whether late days have been used. If it's a bulk upload with zero late days used, the late days value won't be updated.

In the updateLateDays function, I included information about bulk uploads by appending a is_bulk_upload field to the form data.

Additionally, I added an event listener to the gradeable selection to recalculate late days when the user changes the selected gradeable.

### DIFF
--- a/site/public/js/latedays.js
+++ b/site/public/js/latedays.js
@@ -1,6 +1,7 @@
 /* global luxon */
 const DateTime = luxon.DateTime;
 
+// Update the calculateLateDays function to consider bulk upload gradeables with nonzero late days
 function calculateLateDays(inputDate) {
     const select_menu = document.getElementById('g_id');
     if (select_menu.selectedIndex === 0) {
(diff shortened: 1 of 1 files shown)


site/public/js/latedays.js
```diff
@@ -54,29 +99,20 @@ $(document).ready(() => {
...
-        data: fd,
-        processData: false,
-        contentType: false,
-        success: function() {
-            window.location.reload();
-        },
-        error: function() {
-            window.alert('Something went wrong. Please try again.');
-        },
+    // ... (rest of your existing code)
```
ziesski, line 103: Seems like using ChatGPT without understanding the code.

ziesski (changes requested):

ziesski:
This is not the fix we were looking for. This PR does not show correct fix, nor valid understanding of the issue. Using ChatGPT to fix the issue without proper understanding/reviewing. Unfortunately, this PR will be closed. 
Please feel free to contribute with new fixes.

PR #10053, opened 2023-12-09
[Bugfix:Submission] Bulk upload validation error

### LINKED ISSUE #9986: Bulk Upload w/ wrong number of pages fails with no error message

s026134:
image

Fixes #9986 - When there is a bulk upload with an incorrect number of split pages, there is no error that appears on the front end other than what is in the daemon log.

### What is the new behavior?
The new behavior shows an alert that appears saying "The PDF page count (# split pages) is not divisible by the expected count per exam (# pages in pdf)" and once the user hits "ok" it returns back to the submission page for the user to reload. The daemon log as described in the issue, no longer appears because the check for divisibility occurs before the job starts. 

### Other information?
It was tested manually by uploading multiple-page PDFs and inputting different incorrect split page counts. 
This code involves a library called FPDI which is a free pdf document importer for PHP. 
This is the link to FPDI documentation: [](https://github.com/Setasign/FPDI)

### DIFF
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -22,7 +22,6 @@ use app\models\GradingOrder;
 use Symfony\Component\Routing\Annotation\Route;
 use app\models\notebook\SubmissionCodeBox;
 use app\models\notebook\SubmissionMultipleChoice;
-
 // Adds Fpdi library for getting page count of pdfs
 use setasign\Fpdi\Fpdi;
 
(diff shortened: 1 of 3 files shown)


williamjallen:
This appears to be a duplicate of https://github.com/Submitty/Submitty/pull/10048.  These PRs take different approaches, but I think neither one is optimal.  Since https://github.com/Submitty/Submitty/pull/10048 was created first, I'm going to close this PR.  Please feel free to discuss the best solution to this issue on https://github.com/Submitty/Submitty/issues/9986, or on Zulip.

PR #10055, opened 2023-12-11
[Bugfix:Forum] Add more empty list checks

DarthNyan:

There's an issue in the forum that's been affecting FOCS that occurred when the professor went to the forum and filtered by unread. It happens when SQL tries to query an empty list of parameters, and I was able to reproduce this by filtering by unread when I had no unread threads.

### What is the new behavior?
I've added checks for empty lists before all the queries used to display the thread list. I'm able to filter by unread even with no unread threads now.


cjreed121 (approved):

ArnavPawar (approved):
Works, looks like you just added more checks to return empty if the thread list is empty. And code implements this correctly in the correct spots it seems. All tests pass.

PR #10059, opened 2023-12-13
[Refactor:System] Menorah duck dates

ziesski:
Current code has hard coded dates for hanukkah just for 2023.

Now the code correctly automates the date for every year.
In addition, to reflect the culture accurately, the it will only be available from sunfall of first day to sunfall of last day. 
(Although we hard coded sunfall = 4PM, we agree that it is good enough) 

Has been tested functionally, quick side note that it is conversion from hebrew -> julian -> gregorian.


williamjallen (approved):
Seems reasonable to me

PR #10069, opened 2023-12-14
[Bugfix:InstructorUI] Fix grading button

IDzyre:
There are two class parameters in the HTML element for the grade button's on the TA grading page. The element only takes the first one, and omits the next. 
image

### What is the new behavior?
These have now been combined into one class parameter, and everything works as intended.
image


bmcutler (approved):
thanks @IDzyre !
has fixed the rendering bug :)

Issue #10078, opened 2023-12-21
Download output.csv from rainbow grades from webpage


bmcutler:
**What problem are you trying to solve with Submitty**
Instructors want to download the .csv of the grades from rainbow grades.
This file is located in /var/local/submitty/courses/SEMESTER/COURSE/rainbow_grades/output.csv
Currently they must use scp to download that file (and its not documented)

**Describe the way you'd like to solve this problem**
Let's add a blue "Download CSV of grades" button to get this file from the top of the "Gradebook" page.

We probably want to add a note on this page about when the grades were last updated (grab the timestamp that "generate grade summaries" was run.)

TusharPuri10:
I can make adjustments to meet your preferences. Let me know if there are any specific changes you have in mind.

rutuja-khaire:
I want to work on this issue please assign me this issue

Issue #9805, opened 2023-09-25
Using rainbow grades customization UI before "generate grade summaries" causes error


bmcutler:
**Describe the bug**
In a new course (with no prior rainbow grades work), if you first go to the "web based rainbow grades generation" and create and save settings BEFORE doing "generate grade summaries", there are errors in the save and generate rainbow grades html step.  The errors are related to rsync.  (Sorry I didn't screenshot the problem).  

We couldn't fix the problem from the UI alone.  Generating rainbow grades and then returning to the web based customization didn't fix things.  I did fix the problem by deleting everything in the unix rainbow_grades folder for the course and starting over.  but I think this required sudo access, which instructors don't have.

**Expected behavior**
I believe we should have "generate grade summaries" happen automatically on each "save and generate" from the web-based customization page -- especially if it hasn't ever been run.

TusharPuri10:
Can I work on this issue ?

bmcutler:
@TusharPuri10 

Thank you for your interest in the Submitty open source project.
We welcome contributions from new developers!
However we do not use the Github issue 'assign' feature for first time prospective contributors.

Please read our documentation on how to to get started with Submitty:
https://submitty.org/developer/getting_started/index
Specifically in setting up your development environment:
https://submitty.org/developer/getting_started/vm_install_using_vagrant
And making a pull request:
https://submitty.org/developer/getting_started/make_a_pull_request
You do not need to be assigned to an issue to create a pull request that will be reviewed by our team and then merged if it appropriately resolves the issue.
We also encourage you to join our Zulip server to discuss technical questions:
https://submitty.org/index/contact

PR #10079, opened 2023-12-21
[Feature:InstructorUI] Download Button for Rainbow Grades

TusharPuri10:
fixes https://github.com/Submitty/Submitty/issues/10078
closes #10213 

=> Currently the .csv of the grades from rainbow grades is located in /var/local/submitty/courses/SEMESTER/COURSE/rainbow_grades/output.csv

### What is the new behavior?
=> Added a blue "Download CSV of grades" button to get this file from the top of the "Gradebook" page.
image

### Other information?
There was PR with an minor update of RainbowGrades.twig made from PR #10213, which has been cherry picked to this PR, as well as the video demo. The work should be credited accordingly to @MUFFANUJ 

https://github.com/Submitty/Submitty/assets/151390316/06a29734-d22b-4074-9ffc-853bfa60783f

Some useful feedbacks from #10213 was not carried over, please consider reviewing those comments as well when taking over this PR.


site/app/controllers/admin/ReportController.php
```diff
@@ -789,4 +789,49 @@ public function displayGradebook() {
...
+        $csvFilePath = '/var/local/submitty/courses/f23/development/rainbow_grades/output.csv';
+
+        // Check if the file exists
+        if (file_exists($csvFilePath)) {
+            // Set appropriate headers for file download with a custom filename
+            header('Content-Description: File Transfer');
+            header('Content-Type: application/csv');
+            header('Content-Disposition: attachment; filename="' . $this->generateCustomFilename() . '"');
+            header('Expires: 0');
+            header('Cache-Control: must-revalidate');
```
cjreed121, line 825: This route should not be necessary. See the `downloadCourseFile` function in `site/app/controllers/MiscController.php`. Our `Access.php` already has rainbow_grades as a valid directory so you should be able to create a URL to the `/courses/{semester}/{course}/download` path.

site/app/templates/admin/Report.twig
```diff
@@ -10,6 +10,23 @@
...
+            <div class="card-btn-cont">
+                <button onclick="location.href='{{ rainbow_grades_csv_url }}'"
+                class="btn btn-primary key_to_click"
+                tabindex="0"
+                >
+                    Download CSV of grades
+                </button>
+                <div class="card-btn-info">
+                    Last ran: <span id="grade-summaries-last-run">{{ grade_summaries_last_run }}</span>
+                </div>
```
cjreed121, line 26: You might even be able to take advantage of the `downloadFile` function from server.js instead which would let you download the file even easier.

cjreed121 (changes requested):
Thanks for the contribution! Your changes seem that it would address the issue but there is a more preferred way to address the problem which I noted below.

TusharPuri10:
Hi chris,

Thanks for your input. I'll update the code as suggested by using the
existing `downloadCourseFile` function. I'll also explore integrating the
`downloadFile` function from `server.js` for a smoother download process.

I'll keep you posted on the progress.

Regards

site/app/templates/admin/Report.twig
```diff
@@ -3,13 +3,37 @@
         $(document).ready(function() {
             // Bind click listener to grade summaries button
             $('#grade-summaries-button').click(function() {
-                $('#grade-summaries-last-run').html('Running...');
+                $('.grade-summaries-last-run').html('Running...');
+            });
+            // Bind click listener to download buttons
+            $('.download-btn').click(function() {
+                let path = $(this).data('path');
+                let dir = $(this).data('dir');
```
cjreed121, line 11: `const` is preferred over `let`

site/app/templates/admin/Report.twig
```diff
@@ -23,7 +47,7 @@
                     Generate Grade Summaries
                 </button>
                 <div class="card-btn-info">
-                    Last ran: <span id="grade-summaries-last-run">{{ grade_summaries_last_run }}</span>
+                    Last ran: <span class="grade-summaries-last-run">{{ grade_summaries_last_run }}</span>
```
cjreed121, line 50: Why is this changing?

site/app/templates/admin/Report.twig
```diff
@@ -3,13 +3,37 @@
...
     </script>
 
     <div id="grade-report">
         <h1>Grade Reports</h1>
+        <div id="rainbow-grade-report-csv" class="card">
+            <div class="title-cont">
+                Pushing this button will generate a CSV file (for download) of Rainbow Grades.
+            </div>
+            <div class="card-btn-cont">
+                <button class="btn btn-primary key_to_click download-btn"
```
cjreed121, line 24: An ID would be a better choice for the download button since there will only be one

site/app/templates/admin/Report.twig
```diff
@@ -3,13 +3,37 @@
...
+                        data-path="/var/local/submitty/courses/f23/development/rainbow_grades/output.csv"
+                        data-dir="rainbow_grades"
```
cjreed121, line 26: These are hardcoded values. path should be dynamically generated for the specific semester/course context we are in. Also dir is completely static so that could be moved entirely to the JS.

cjreed121 (changes requested):

site/app/templates/admin/Report.twig
TusharPuri10: I apologize for the oversight

site/app/templates/admin/Report.twig
TusharPuri10: Got it

TusharPuri10:
@cjreed121 I solved the issue but the user first has to **Generate the Rainbow Grades** to actually create the path to output.csv else it will throw
`$this->core->getOutput()->showError(self::GENERIC_NO_ACCESS_MSG);`
image

because the file does not exist

**path:  \Submitty\site\app\libraries\Access.php**
```php
public function canUser($user, string $action, $args) {
...
            //This is not a valid directory
            //checks for top level dirs like annotations, checkout, course_materials, submissions ... )
            if (!array_key_exists($dir, $this->directories)) {
                return false;
            }
            elseif (!file_exists($path)) {
                //checks for the existence of path which is asked for
                return false;
            }
...
}
```
How do you think this can be solved? Should I add message "Please generate rainbow grades first" or add default columns to table to generate it on click itself.

TusharPuri10:
https://github.com/Submitty/Submitty/issues/9805 is part of this issue. I think solving it and then adding the similar solution to the "Download CSV of grades" button will solve the issue I stated in above comment.

bmcutler:
@TusharPuri10 

In the original issue I requested this be placed on the "Gradebook" page, not on the "Grade Reports" page.
The gradebook page link appears on the left sidebar when "Display Rainbow Grades" is selected on "Course settings"

Thanks.

site/app/libraries/Access.php
```diff
@@ -914,8 +914,17 @@ public function resolveDirPath(string $dir, string $path) {
...
-        if (!str_starts_with($path, $info["base"])) {
+        // Diffferent base path for downloading output.csv of rainbow grades
+        if ($dir === "rainbow_grades" && substr($path, -11) === "/output.csv") {
+            // Removing "/upload" from path
+            $base = substr($info["base"],0,-23) . "/rainbow_grades";
+            //checking the path
+            if(!str_starts_with($path, $base))
+            {
+                $path = $base . $path;
+            }
```
cjreed121, line 925: Is this change just to remove the `/upload` part from the path? You should be able to just add `/upload` in the JS call.

cjreed121 (changes requested):
Making the change that @bmcutler noted should address the issue you are having.

site/app/libraries/Access.php
TusharPuri10: yes, it is to remove the `/upload` part from path. Otherwise I had to change the base path of `rainbow_grades` directory
``` php
private function loadDirectories() {
...
        $this->directories["rainbow_grades"] = [
            "base" => $this->core->getConfig()->getCoursePath() . "/uploads/rainbow_grades",
            "subparts" => [],
            "permissions" => [
                "path.read" => "path.read.rainbow_grades",
                "path.write" => "path.write.rainbow_grades",
            ]
        ];
...
}
```
to bypass the checks of `Access.php`. This will effect other URLs as well that's why I thought this will be a faster solution. If you insist I can make those changes by changing the base path.
> @TusharPuri10
> 
> In the original issue I requested this be placed on the "Gradebook" page, not on the "Grade Reports" page. The gradebook page link appears on the left sidebar when "Display Rainbow Grades" is selected on "Course settings"
> 
> Thanks.

 I will also address this issue.

TusharPuri10:
Added a blue "Download CSV of grades" button to get this file from the top of the "Gradebook" page. I apologize for the inconvenience as I have misread the issue first time. @cjreed121 please review.
image

TusharPuri10:
@cjreed121 Can you please tell me about Cypress (UI) check. I was unable to solve it.

TusharPuri10 (commented):
Can you please tell me about Cypress (UI) check. @cjreed121

TusharPuri10 (commented):
I apologize for that. It won't happen again.

site/app/libraries/Access.php
```diff
@@ -914,8 +914,16 @@ public function resolveDirPath(string $dir, string $path) {
...
+        // Diffferent base path for downloading output.csv of rainbow grades
+        if ($dir === "rainbow_grades" && substr($path, -11) === "/output.csv") {
+            // Removing "/upload" from path
+            $base = substr($info["base"], 0, -23) . "/rainbow_grades";
+            //checking the path
+            if (!str_starts_with($path, $base)) {
+                $path = $base . $path;
+            }
+        }//Make sure it starts with the dir base
+        elseif (!str_starts_with($path, $info["base"])) {
```
cjreed121, line 926: Sorry, I'm actually going to go back on what I said. I thought the Access.php rainbow_grades directory pointed to this one but didn't realize it had the upload on it. I don't think we should add something like this to Access.php. If you could just grab your code from before in the git log that had the new route which downloaded the file, that would be great.

cjreed121 (changes requested):

ziesski:
There was PR with an minor update of RainbowGrades.twig made from PR #10213, which has been cherry picked to this PR. The work should be credited accordingly to @MUFFANUJ 

Some useful feedbacks from #10213 was not carried over, please consider reviewing those comments as well when taking over this PR.

site/app/libraries/Access.php
dorian451: Fixed in https://github.com/Submitty/Submitty/pull/10079/commits/7fcf166721524d006e8da11a8feb43b52365c2fb

zacharymnp:
the new button produces a different .csv than the old button - although it is still a Rainbow Grades .csv

zacharymnp (approved):

site/app/controllers/admin/ReportController.php
```diff
@@ -135,13 +135,10 @@ public function getGradeSummariesLastRun() {
 
     /**
      * Generates and offers download of CSV grade report
+     * @AccessControl(role="INSTRUCTOR")
```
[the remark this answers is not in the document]
dorian451: Fixed in 964525f73b3d7ccc480cfed597b0003e316abca4

site/app/controllers/admin/ReportController.php
```diff
@@ -776,8 +772,56 @@ public function displayGradebook() {
...
             )
         );
     }
+
+    /**
+     * Generate a custom filename for the downloaded CSV file
+     *
+     * @return string
+     */
+    private function generateCustomFilename() {
```
williamjallen, line 786: It would be better to use native PHP type annotations instead of PHPDoc comments, because native type annotations can be enforced by the language itself.
```suggestion
    /**
     * Generate a custom filename for the downloaded CSV file
     */
    private function generateCustomFilename(): string {
```

site/app/controllers/admin/ReportController.php
```diff
@@ -776,8 +772,56 @@ public function displayGradebook() {
...
+        $course = $this->core->getConfig()->getCourse();
+        $timestamp = date("ymdHis");
```
williamjallen, line 788: You could use `DateUtils::getFileNameTimeStamp()` for this instead.

site/app/controllers/admin/ReportController.php
```diff
@@ -776,8 +772,56 @@ public function displayGradebook() {
...
+        return "{$course}_rainbow_grades_{$timestamp}.csv";
+    }
+
+
+    /**
+     * Download CSV file for Rainbow Grades
+     * @return void
+     */
+    #[Route("/courses/{_semester}/{_course}/reports/rainbow_grades_csv")]
+    public function downloadRainbowGradesCSVFile() {
```
williamjallen, line 798: Same comment here:
```suggestion
    /**
     * Download CSV file for Rainbow Grades
     */
    #[Route("/courses/{_semester}/{_course}/reports/rainbow_grades_csv")]
    public function downloadRainbowGradesCSVFile(): void {
```

site/app/controllers/admin/ReportController.php
```diff
@@ -776,8 +772,56 @@ public function displayGradebook() {
...
+        // Check if the file exists
+        if (file_exists($csvFilePath)) {
+            // Set appropriate headers for file download with a custom filename
+            header('Content-Description: File Transfer');
+            header('Content-Type: application/csv');
+            header('Content-Disposition: attachment; filename="' . $this->generateCustomFilename() . '"');
+            header('Expires: 0');
+            header('Cache-Control: must-revalidate');
+            header('Pragma: public');
+            header('Content-Length: ' . filesize($csvFilePath));
```
williamjallen, line 816: [this](https://github.com/Submitty/Submitty/pull/10079#discussion_r1438422010) review comment from Chris still needs to be addressed.

site/app/controllers/admin/ReportController.php
```diff
@@ -776,8 +772,56 @@ public function displayGradebook() {
...
+
+            // Output the file content
+            readfile($csvFilePath);
+            exit;
+        }
+        else {
+            // Handle the case where the file does not exist
+            $this->core->getOutput()->showError($csvFilePath . " was not found or was not readable.\nMaybe you have not <a\thref='https://submitty.org/instructor/course_settings/rainbow_grades/automatic_setup'>generated the rainbow grades</a> yet?");
```
williamjallen, line 824: Putting a link in an error message banner seems like a bad idea...

site/app/templates/submission/RainbowGrades.twig
```diff
@@ -1,11 +1,32 @@
...
+                        onclick='window.open("{{ rainbow_grades_csv_url }}", "_blank").focus()'
+                        tabindex="0"
+                    >
+                        Download CSV of grades
+                    </button>
+                </div>
+                <p>
+                    Grades last updated on: 
+                    <b>
+                    <span>{{ grade_summaries_last_run }}</span>
```
williamjallen, line 24: Is this `<span>` necessary?

site/public/css/rainbow-grades.css
```diff
@@ -1,3 +1,11 @@
...
     color: black;
 }
+
+#rainbow-grades-header > .btn-wrapper {
+    margin: auto;
+}
+
+#rainbow-grades-header > p {
+    float: right;
+}
```
williamjallen, line 11: Targeting elements by raw selectors like `p` is discouraged.  Please use a class or ID instead.  You can also probably find an existing CSS selector to do this.

site/app/templates/submission/RainbowGrades.twig
```diff
@@ -1,11 +1,32 @@
...
+    <header>
+        <h1 class="label">
+            {% if extra_label is defined %}
+                Rainbow Grades {{extra_label}}
+            {% else %}
+                Rainbow Grades
+            {% endif %}
+        </h1>
+        <div id="rainbow-grades-header">
+            {% if show_summary and extra_label is defined and extra_label == "For All Students" %}
```
williamjallen, line 11: This check against a string to be shown in the UI feels like a hack.  Why not create a variable specifically for the "all students" case?

site/app/controllers/admin/ReportController.php
```diff
@@ -776,8 +772,56 @@ public function displayGradebook() {
...
+    /**
+     * Download CSV file for Rainbow Grades
+     * @return void
+     */
+    #[Route("/courses/{_semester}/{_course}/reports/rainbow_grades_csv")]
+    public function downloadRainbowGradesCSVFile() {
+        // Check if the user has the necessary permissions (e.g., admin access)
+        if (!$this->core->getUser()->accessAdmin()) {
+            $this->core->getOutput()->showError("This account cannot access admin pages");
+        }
```
williamjallen, line 802: I believe this isn't necessary since the entire class is annotated with `@AccessControl(role="INSTRUCTOR")`.

site/app/controllers/admin/ReportController.php
```diff
@@ -138,10 +138,6 @@ public function getGradeSummariesLastRun() {
      */
     #[Route("/courses/{_semester}/{_course}/reports/csv")]
     public function generateCSVReport() {
-        if (!$this->core->getUser()->accessAdmin()) {
-            $this->core->getOutput()->showError("This account cannot access admin pages");
-        }
-
```
williamjallen, line 144: While this is a good change to make, it doesn't seem to be related to the rest of this PR, and should be done separately.

williamjallen (changes requested):

site/app/controllers/admin/ReportController.php
dorian451: https://github.com/Submitty/Submitty/pull/10079#discussion_r1530954806

site/app/controllers/admin/ReportController.php
dorian451: Why?

site/app/controllers/admin/ReportController.php
```diff
@@ -776,8 +775,49 @@ public function displayGradebook() {
...
+
+
+    /**
+     * Download CSV file for Rainbow Grades
+     */
+    #[Route("/courses/{_semester}/{_course}/reports/rainbow_grades_csv")]
+    public function downloadRainbowGradesCSVFile(): void {
+
+        // Path to the CSV file for Rainbow Grades
+        $csvFilePath = '/var/local/submitty/courses/' . $this->core->getConfig()->getTerm() . '/' . $this->core->getConfig()->getCourse() . '/rainbow_grades/output.csv';
```
ziesski, line 801:
```suggestion
        $csvFilePath = FileUtils::joinPaths(
    '/var/local/submitty/courses',
    $this->core->getConfig()->getTerm(),
    $this->core->getConfig()->getCourse(),
    'rainbow_grades',
    'output.csv'
);

```

ziesski (changes requested):

site/app/controllers/admin/ReportController.php
ziesski: Using `FileUtils::joinPaths` ensures that the correct path separators are used, regardless of the operating system.

dorian451: I forgot, my bad

site/app/controllers/admin/ReportController.php
williamjallen: If the banner message disappears after a period of time, so too will the documentation the users wants to click on.  Couldn't you display this else in the UI somewhere instead?  At a minimum, the HTML should be injected via Twig.

site/app/controllers/admin/ReportController.php
williamjallen: Don't we already have some existing code for file downloads?

site/app/controllers/admin/ReportController.php
```diff
@@ -776,8 +775,55 @@ public function displayGradebook() {
...
+        // Check if the file exists
+        if (file_exists($csvFilePath)) {
+            // Set appropriate headers for file download with a custom filename
+            header('Content-Description: File Transfer');
+            header('Content-Type: application/csv');
+            header('Content-Disposition: attachment; filename="' . $this->generateCustomFilename() . '"');
+            header('Expires: 0');
+            header('Cache-Control: must-revalidate');
+            header('Pragma: public');
+            header('Content-Length: ' . filesize($csvFilePath));
```
IDzyre, line 818: Could this use the DownloadResponse from 'app\libraries\response\DownloadResponse' instead of setting headers manually?

IDzyre (changes requested):

site/app/controllers/admin/ReportController.php
IDzyre: If you to file_get_contents($csvFilePath) instead of readfile() does it work?

site/app/controllers/admin/ReportController.php
dorian451: The error message will never disappear, as it is being rendered by ErrorPage.twig, which never hides the error message. I believe that the only way to move the link out of file would be to either to change how error pages are created in Submitty or make a brand-new error page specifically for this one error. I could also just remove the link, but I figured having it there would make it easier for users to debug their problem.

site/app/controllers/admin/ReportController.php
dorian451: Unfortunately no, it just returns an empty page.

IDzyre: You have to return the download response for it to work.

dorian451: If I do that PHP gets mad and says `Fatal error: A void function must not return a value in /usr/local/submitty/site/app/controllers/admin/ReportController.php on line 826`

dorian451: Fixed in 23c6f701bcd9d1650355e79cdd1006c539af06bc

ziesski (approved):
Code looks good, we just need to decide when to clean up the old generate csv report button from report page. I made the issue and linked to this PR #10591 
Sudden change may be confuse existing users (in this case I don't think there are many users using this feature) so might be nice if we can have notification & redirect link next to the old button saying "this button will be moved to [here](http://localhost:1511/courses/s24/sample/gradebook)"

PR #10080, opened 2023-12-28
[UI/UX:System] Sticky Sidebar

### LINKED ISSUE #5902: Floating Sidebar
**What problem are you trying to solve with Submitty?**
Some gradeable assignments can get very long (like exams). If you want to click on the sidebar while deep within the gradeable, you have to scroll all the way up to the top to see it.

A floating sidebar stays in the same place on your screen, even when scrolling. Thus, you are always able to click on the sidebar pages no matter your location in the gradeable.

Patch4747:

If on a long page in Submitty and currently scrolled far down into it, the sidebar is gone and the user must scroll all the way to the top to use it.

### What is the new behavior?
 This PR uses sticky positioning so the sidebar instead remains at its position on the page. Screenshot is an example of sidebar appearance when scrolled down on page. Closes #5902.

image

### DIFF
--- a/site/public/css/sidebar.css
+++ b/site/public/css/sidebar.css
@@ -99,10 +99,13 @@ aside {
         position: absolute;
         margin: 12px 0 0 32px;
     }
+    
+    aside ul {
+        position: sticky;
+        top: 10px;
+    }
 
     [data-theme="dark"] aside ul {
         background-color: var(--background-blue);
-        position: sticky;
-        top: 10px;
     }
 }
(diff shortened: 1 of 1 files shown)


cjreed121 (changes requested):
Thanks for the contribution! This definitely addresses the issue from the student side but does create another issue from the instructor side. Instructors have a very long list of pages on the sidebar that has probably gotten too long but it is what it is for now. This change will make it so instructors can't see the bottom of the list until they scroll all the way down. Here is a screenshot of me midway down a course with this change applied where I can't get the bottom of the sidebar list (there is more below grade reports):
image

I'm honestly not sure what the best solution to this would be. Maybe a scroll bar for the side bar but I'm not sure if that's the best UI choice? If you have any ideas on this, please let us know.

PR #10095, opened 2024-01-01
[Bugfix:InstructorUI] Change of constraints in team name

### LINKED ISSUE #10074: Team Gradeables - team name length
**Describe the bug**
Students can optionally specify a team name for their gradeable.
Currently we don't seem to have a limit on the length of the team name or on the characters that can be used for a team name.
If the team name is excessively lengthy, this impacts the rendering of the TA grading index page and makes it difficult for graders to do their work.  Also, if there are weird or un-renderable characters in the team name, the display is confusing.

NOTE: instructors can edit a team name from the grading index page.  So lengthy team names can be manually replaced one-at-at-time.  But this is temporary, students can re-edit the team name.

**How to fix it**

NOTE: At this time, let's not write a migration to change any existing team names.  

* Let's truncate long team names already in the database before displaying the TA grading index page.  About 30 characters should be sufficient.  If the name is longer than 30 characters let's add "..." (ellipses) to make clear its been truncated.

* On the page where the students specify their team name or edit their team name, let's add some text:
"Team names are limited to 30 characters"
And if a student tries to make a longer team name, they should get an error message and the new name should not save.
Let's make sure this applies to new teams, new team names, and editing names of existing teams.

* On the page where instructors can edit the name of a team, let's also cap the length of edits to team names to 30 characters.

* Let's also decide on a reasonable set of characters that can be used in team names:  letters, numbers, spaces, dash, underscore for sure.  Would be nice to include characters with accents also.  But is it necessary to have emojis?  This is unnecessary and outside the original purpose of team names - to help instructors in large classes where it is helpful to label big groups and facilitate efficient course management.

GO-viper7:

* [✅] Screenshots are attached to Github PR if visual/UI changes were made

Fixes #10074

Presently, there is no restriction on the length or allowed characters for team names. However, extensive team names can disrupt the layout of the TA grading index page, posing challenges for graders in carrying out their tasks. Additionally, the presence of unusual or unrenderable characters in team names can lead to a confusing display.

### What is the new behavior?

Truncating the Team name with greater than 30 characters and adding "..." (ellipses)
Text to indicate team name cannot have more than 30 characters

Displaying error response for the below after user clicks on submit: 
1. If team name contains emojis
2. If team name contains special characters (['-', '_' ] are omitted and the array in code can be expandable upon further requests)
3. If team name has more than 30 characters.

TeamElipses
TeamLimit

### DIFF
--- a/migration/migrator/data/submitty_db.sql
+++ b/migration/migrator/data/submitty_db.sql
@@ -427,7 +427,7 @@ CREATE TABLE public.courses_users (
 
 CREATE TABLE public.emails (
     id bigint NOT NULL,
-    user_id character varying NOT NULL,
+    user_id character varying,
     subject text NOT NULL,
     body text NOT NULL,
     created timestamp without time zone NOT NULL,
@@ -435,7 +435,9 @@ CREATE TABLE public.emails (
     error character varying DEFAULT ''::character varying NOT NULL,
     email_address character varying(255) DEFAULT ''::character varying NOT NULL,
     term character varying,
-    course character varying
+    course character varying,
+    to_name character varying,
+    CONSTRAINT name_or_email CHECK (((user_id IS NOT NULL) <> (to_name IS NOT NULL)))
 );
 
 
--- a/migration/migrator/migrations/master/20230914092120_add_to_name_to_email_database.py
+++ b/migration/migrator/migrations/master/20230914092120_add_to_name_to_email_database.py
@@ -1,0 +1,37 @@
+"""Migration for the Submitty master database."""
+
+
+def up(config, database):
+    """
+    Run up migration.
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    database.execute("""
+        ALTER TABLE emails
+        ADD COLUMN IF NOT EXISTS to_name VARCHAR;
+        --- Now user_id null if to_name used instead.
+        ALTER TABLE emails
+        ALTER COLUMN user_id DROP NOT NULL;
+
+        ALTER TABLE emails
+        DROP CONSTRAINT IF EXISTS name_or_email;
+        ALTER TABLE emails
+        ADD CONSTRAINT name_or_email CHECK (
+            (user_id is NOT NULL)
+            <>
+            (to_name is NOT NULL)
+        );
+    """)
+
+def down(config, database):
+    """
+    Run down migration (rollback).
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    """
+    pass
--- a/sbin/send_email.py
+++ b/sbin/send_email.py
@@ -134,7 +134,7 @@ def construct_mail_client():
 
 def get_email_queue(db):
     """Get an active queue of internal emails waiting to be sent."""
-    query = """SELECT id, user_id, email_address, subject, body FROM emails
+    query = """SELECT id, user_id, to_name, email_address, subject, body FROM emails
     WHERE email_address SIMILAR TO :format AND sent is NULL AND
     error = '' ORDER BY id LIMIT 100;"""
     domain_format = '%@(%.' + EMAIL_INTERNAL_DOMAIN + '|' + EMAIL_INTERNAL_DOMAIN + ')'
@@ -144,9 +144,10 @@ def get_email_queue(db):
         queued_emails.append({
             'id': row[0],
             'user_id': row[1],
-            'send_to': row[2],
-            'subject': row[3],
-            'body': row[4]
+            'to_name': row[2],
+            'send_to': row[3],
+            'subject': row[4],
+            'body': row[5]
             })
 
     return queued_emails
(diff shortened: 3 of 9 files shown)


site/app/controllers/student/TeamController.php
```diff
@@ -497,6 +497,28 @@ public function setTeamName($gradeable_id): RedirectResponse {
             $_POST['team_name'] = null;
         }
 
+        if(strlen($_POST['team_name']) > 30) {
```
shailpatels, line 500: I'd suggest turning this into a constant so its located in one place

The [team model](https://github.com/Submitty/Submitty/blob/main/site/app/models/Team.php) would be a good place, you can add a new constant in the class, something named `MAX_TEAM_NAME_LENGTH = 30` then you can you import the file (`use app\models\Team;`) and access this via `Team::MAX_TEAM_NAME_LENGTH` where needed

site/app/controllers/student/TeamController.php
```diff
@@ -497,6 +497,28 @@ public function setTeamName($gradeable_id): RedirectResponse {
...
+            $this->core->addErrorMessage("Number of characters in team name must be less than 30");
+            return new RedirectResponse($return_url);
+        }
+        
+        if(preg_match('/[\x{1F600}-\x{1F64F}\x{1F300}-\x{1F5FF}\x{1F680}-\x{1F6FF}\x{1F700}-\x{1F77F}\x{1F780}-\x{1F7FF}\x{1F800}-\x{1F8FF}\x{1F900}-\x{1F9FF}\x{1FA00}-\x{1FA6F}\x{1FA70}-\x{1FAFF}\x{2600}-\x{26FF}\x{2700}-\x{27BF}\x{2300}-\x{23FF}]/u',$_POST['team_name'])) {
+            $this->core->addErrorMessage("Team name should not contain emojis");
+            return new RedirectResponse($return_url);
+        } 
+
+        function hasSpecialCharacters($str) {
```
shailpatels, line 510: It would be good to move this function outside `setTeamName`, I think just leaving it in the same controllerFile and making it a `static` function would be good so you can call it from anywhere and it'll be easier to write a unit test for it

site/app/templates/submission/EditTeamName.twig
```diff
@@ -4,7 +4,10 @@
 {% block body %}
     <p class="std-margin">
         Please fill in an appropriate name for your team if you would like.
+        <br>
+        Team names are limited to 30 characters
     </p>
+    
     <div class="flex-col" id="team-name-modal">
         <input id="team-name-csrf" type="hidden" name="csrf_token" value="{{ csrf_token }}" />
         <div class="edit_user_row">
```
shailpatels, line 13: You can put the attribute `maxlength` https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength to perform a client side check that the input length is correct. You could also pass the MAX_LENGTH as a variable so that the front end backend will always be synced to the same constraint. 

This twig file is rendered from this view `site/app/views/submission/TeamView.php` and gets imported via `site/app/templates/submission/Team.twig` 

The diff isn't letting me click it but its line 12, the input where the name is "team_name"

site/app/templates/submission/EditTeamName.twig
```diff
@@ -4,7 +4,10 @@
 {% block body %}
     <p class="std-margin">
         Please fill in an appropriate name for your team if you would like.
+        <br>
+        Team names are limited to 30 characters
```
shailpatels, line 8: See comment below about passing the "30" from the backend so its always the same everywhere

site/app/controllers/student/TeamController.php
```diff
@@ -497,6 +497,28 @@ public function setTeamName($gradeable_id): RedirectResponse {
             $_POST['team_name'] = null;
```
shailpatels, line 497: you may want to put your logic in an else statement after this check, if the user does not input a team name then no point in doing validation on it

shailpatels (changes requested):
Thanks for the PR! Looks like the main functionality is there just a couple comments related to code quality

GO-viper7:
@shailpatels 
Hey, thanks for the comments!
hope, the updated files met those requirements which were stated
Please go through them and mention if any further changes are needed
Thank you.

GO-viper7:
hey @shailpatels 
So, there's an unsuccessful check for php-lint because of the following : 
1. starting an else statement right after ending the if statement instead of starting it in next line
2. writing parentheses without a space after an if statement
and etc.

Do you want me to make these changes and commit again or can it be merged irrespective of these changes with your authorization.

williamjallen:
@GO-viper7 All of the checks must pass before this can be merged.

shailpatels:
@GO-viper7  I think I accidentally broke your Pull Request trying to force push some fixes after your force push. (I removed the last instance of the hardcoded '30' and PHP lint errors).

I can no longer push to your branch. When you forked did you check the box to allow contributors to push to your branch? Not sure if thats still an option but could you try and do that, otherwise you could recreate the branch from the one I have here and remake your PR. 

https://github.com/Submitty/Submitty/tree/team-constraint
and here is the diff
https://github.com/Submitty/Submitty/compare/main...team-constraint

I think if you make a new branch and pull that one into yours, that would be the easiest way to rebuild the PR, I think the issue is you forked to main instead of a branch and may of had some branch protections? Not fully sure what happened but sorry again - let me know if you need some help remaking the PR!

PR #10102, opened 2024-01-13
[Refactor:Developer] Lint remaining JS files

williamjallen:
Eslint is currently run on most of our files, but the target JS version is too low to be able to effectively parse a handful of remaining files.

### What is the new behavior?
This PR changes our JS target version to `es2020`, and finishes linting the rest of the JS files.  All of our JS files are now linted on every CI run.


arougey (approved):

cjreed121 (commented):
Changes look good. I think on ignore you added could be solved pretty simply which I noted below.

cjreed121 (approved):

IDzyre (approved):

PR #10103, opened 2024-01-14
[Bugfix:Submission] Fix duplicate team anon IDs

cjreed121:

https://github.com/Submitty/Submitty/pull/8733 had a bug in the migration where any gradeable missing anonymous IDs, would receive duplicate anonymous IDs. This causes some old courses to get duplicate anonymous IDs and then prevent loading the grading page.

### What is the new behavior?
This PR wipes any duplicate anonymous IDs and then regenerates them with the same code but with a bugfix. This also adds a unique constraint so that it is not an issue going forward.

### Other information?
I tested locally and confirmed that any duplicates do get fixed.

### DIFF
--- a/migration/migrator/data/course_tables.sql
+++ b/migration/migrator/data/course_tables.sql
@@ -2021,6 +2021,14 @@ ALTER TABLE ONLY public.threads ALTER COLUMN id SET DEFAULT nextval('public.thre
 
 
 --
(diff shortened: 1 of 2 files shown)


bmcutler:
Thanks @cjreed121 this fixed the bug from the S20 course!

bmcutler (approved):

Issue #10106, opened 2024-01-15
Bulk Upload UI requires toggle to display exam page count form


bmcutler:
**Describe the bug**

* Instructor creates a new gradeable "TA/Instructor will (bulk) upload scanned .pdf for online manual grading" (a bulk pdf / scanned exam gradeable)

* When instructor first enters to "BULK UPLOAD" from the gradeables page, they see this on the submission/upload page:

  <img width="509" alt="Screenshot 2024-01-15 at 5 20 20 PM" 
   src="https://github.com/Submitty/Submitty/assets/5871417/dddc7483-d220-468d-8c51-854e52c5bd36">

* This is buggy.  There is nowhere to specify either split by QR code or page count per student pdf.

* The instructor can hack to fix it by toggling "Make submission for student" and then back to "Bulk PDF Upload".  Then they correctly see:

<img width="550" alt="Screenshot 2024-01-15 at 5 20 29 PM" src="https://github.com/Submitty/Submitty/assets/5871417/17144e6c-a523-4296-bfb8-0562310bc4e5">

**Expected behavior**

Toggling should not be necessary when you first load the page.

ProgrammingPirates:
Hell Sir @bmcutler  i want to work on this issue, can you assign me please

bmcutler:
HI @ProgrammingPirates 

Thank you for your interest in the Submitty open source project.
We welcome contributions from new developers!
However we do not use the Github issue 'assign' feature for first time prospective contributors.

Please read our documentation on how to to get started with Submitty:
https://submitty.org/developer/getting_started/index
Specifically in setting up your development environment:
https://submitty.org/developer/getting_started/vm_install_using_vagrant
And making a pull request:
https://submitty.org/developer/getting_started/make_a_pull_request
You do not need to be assigned to an issue to create a pull request that will be reviewed by our team and then merged if it appropriately resolves the issue.
We also encourage you to join our Zulip server to discuss technical questions:
https://submitty.org/index/contact

PR #10108, opened 2024-01-16
[Bugfix:InstructorUI] Fix bulk upload not showing options

TusharPuri10:
fixes https://github.com/Submitty/Submitty/issues/10106

A bug exists in the initial view of the "BULK UPLOAD" feature on the gradeables page, preventing users from specifying QR code or page count for student PDFs. Instructors can temporarily fix it by toggling "Make submission for student" before switching back to "Bulk PDF Upload," revealing the correct options. 

image

## What is the new behavior?
This new logic checks for a previous setting in session-Storage, and if found, it sets the appropriate radio button accordingly. If there is no previous setting, it defaults to clicking the "Bulk Upload" radio button. **( The logic is,  it must be clicked if it exists at first place )** .This modification ensures that the correct view is displayed, resolving the problem identified in the initial rendering of the "BULK UPLOAD" feature.

image


powe97:
Tested, looks good.

powe97 (approved):

ziesski (approved):
Didn't test the functionality, but the code seems minimal change, fixing the introduced issue.

IDzyre (changes requested):
The tests need to pass before this is good to merge.

bmcutler:
I don't understand why this code would fix the problem.  Did you test both bulk upload gradeables and regular submission gradeables?  I think this change will probably break regular submission gradeables.

bmcutler (changes requested):
Code may fix the problem but breaks regular gradeables, this can be seen in the Cypress tests.

williamjallen:
Superseded by https://github.com/Submitty/Submitty/pull/10108, as per the PR description.

PR #10109, opened 2024-01-16
[Feature:Forum] Add attachments when editing posts

### LINKED ISSUE #9795: Add ability to add attachments when editing discussion forum post
While struggling to add an attachment to my post on the discussion forum (#9794), I discovered that it is impossible to add an attachment to an existing post.  In my opinion, users should be able to do anything via the edit menu that they can do when making a post normally.

Users should be able to add attachments when editing existing posts.

DarthNyan:

Currently, users can't add attachments when they edit posts on the discussion forum.

### What is the new behavior?
Closes #9795. Users can now add attachments while editing their posts and threads.

### Other information?

The increments to post_box_id are due to the existing file upload code storing files in an array indexed by post_box_id - 1, and each reply/edit form has a unique id. The edit form's hardcoded id of 0 was trying to access index -1 and throwing an error, so I incremented every post_box_id by 1.


jeffreyC4:
The implementation looks good. Works as intended when I tested the add attachments feature when editing posts on the forum.

jeffreyC4 (commented):
The implementation looks good. Works as intended when I tested the add attachments feature when editing posts on the forum.

ArnavPawar (approved):
Code looks good.
Tested with and without your changes and it works as expected. I am able to add an attachment to an already existing post and tested with a post and a reply to a post etc. 
Was not able to break and it code makes sense. 
Looks good to merge.

jeffreyC4 (approved):

williamlin6803 (approved):
Code looks good and was able to make sense of it. 
I tested it on my pc and it worked fine. Can attach things when editing a post.

RayReyRam (commented):
When testing on my machine I was able to edit and then attach a file. Though on some occasions it would prompt me to attach a file again after clicking a file to attach(Check video below). Other than that the function of the code works

https://github.com/Submitty/Submitty/assets/148298134/dedb7aed-cae0-4c90-bcd0-3f75a7348fb0

site/app/views/forum/ForumThreadView.php
```diff
@@ -628,7 +628,7 @@ public function generatePostList($currentThread, $posts, $unviewed_posts, $curre
                 "merge_thread_content" => $merge_thread_content,
                 "csrf_token" => $csrf_token,
                 "activeThreadTitle" => $activeThreadTitle,
-                "post_box_id" => $post_box_id,
+                "post_box_id" => $post_box_id + 1,
```
IDzyre, line 631: Is there a reason to add one to all of the objects?

DarthNyan: Yes. The file upload functions in drag-and-drop.js use an array that's indexed by post_box_id-1. The edit form's id had been hardcoded to 0, so this was trying to access index -1 and would throw an error. I figured the simplest fix would be to shift all the ids by 1 rather than trying to give the edit form an id based on the number of posts or rewrite the drag-and-drop code.

I can take a look and clean up how the ids are adjusted or try an alternative approach.

williamjallen (approved):
Changes seem reasonable to me.  It would be good to add some Cypress tests for this eventually, either in this PR or in a separate one.  Small features like this, particularly in complicated/messy parts of the codebase like the forum are prone to being broken inadvertently.

PR #10115, opened 2024-01-19
[Testing:Developer] Improve flakiness of vagrant up workflow

MasterOdin:

The vagrant up workflow often fails right now, where VirtualBox is stuck in a "gurumeditation" state. There is https://github.com/actions/runner-images/issues/8730 which indicates others are running into the same problem, and while GH has closed the issue with it's fixed and sometimes the workflow does succeed, it's still often failing. My only guess is that per https://github.com/actions/runner-images/issues/8730#issuecomment-1834391700, there are different machines that run macOS-12 builds and certain ones just always fail :shrug:.

### What is the new behavior?

PR updates the vagrant up workflow to use macOS 13 which while in beta seems to provide a much more stable interface for vagrant/virtualbox, and that I was able to run the workflow 4 times in a row (with one disconnect failure, which I think is due to machine running out of resources?), something we haven't seen in a long time (https://github.com/Submitty/Submitty/actions/workflows/vagrant_up.yaml).

### DIFF
--- a/.github/workflows/vagrant_up.yaml
+++ b/.github/workflows/vagrant_up.yaml
@@ -14,6 +14,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - run: brew install virtualbox vagrant
       # Installing this is skipped in CI as using it was causing some corruption in
       # attempting to mount the shared folder. The guest and host guest additions
       # versions are close enough that we are fine without it.
--- a/.github/workflows/vagrant_up.yaml
+++ b/.github/workflows/vagrant_up.yaml
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   vagrant-up:
-    runs-on: macos-latest
+    runs-on: macos-13
 
     steps:
       - uses: actions/checkout@v4
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -32,6 +32,8 @@ $stderr.sync = true
 
 require 'json'
 
+ON_CI = !ENV.fetch('CI', '').empty?
+
 def gen_script(machine_name, worker: false)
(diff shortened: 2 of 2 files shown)


cjreed121:
That's great to see it passing more often now. I will say though that the old action was completing in about 1hr 5min (when it would pass) while now it's sitting at about 1hr 50min. I saw you were getting the disconnect issue with the higher resources which is strange. Is it worth experimenting to see if we can keep at least some of the higher resources (i.e. 1 cpu & 2024 MB or 2 cpu & 1024 MB)?

williamjallen:
Did you intend to include all of the vagrant cloud commits in your PR?

IDzyre:
> Did you intend to include all of the vagrant cloud commits in your PR?

No, apparently when merging this PR into my fork to see how it works with the cloud, it merged my changes into this PR. It should be fixed now.

MasterOdin:
I've merged this as it's a definite right step in terms of improving flakiness, though the resource stuff is annoying.

> Is it worth experimenting to see if we can keep at least some of the higher resources (i.e. 1 cpu & 2024 MB or 2 cpu & 1024 MB)?

I'm not sure. I wouldn't expect more RAM to help as I don't think we're doing anything that's too heavy there. More CPU could help, though I'm not sure what the sweet spot would be. I think that using `macos-13` in general is going to be slower due to a change in the underlying architecture that GH uses, and from what I've read in the linked issue above (and issues linked off that) is that the newer macos machines don't handle virtualization as well, and if we want better performance the best bet would be to move to a linux based runner.

Right now, given that this runs once a day pretty async, it's probably kind of fine for it to be a bit slower, and that I don't think anyone will really notice or care, vs it always failing and always posting in zulip (and emailing me).

Issue #10070, opened 2023-12-14
Markdown posts have padding that non-markdown posts do not


PythonCoderAS:
**Describe the bug**

When making a post with Markdown enabled, even if no markdown is used, the resulting post has extra padding because by default the Markdown renderer will add in `<p>` tags for every paragraph. In contrast, when not using Markdown, the text is inserted as is, without `<p>` tags. This leads to an inconsistency in post padding.

Additionally, normal posts have spaces prepended to them at the beginning, but Markdown posts do not.

**Expected behavior**
 Both modes would make the post look the same

**To Reproduce**
1. Make a post with normal text
2. Edit post to use Markdown

**Configuration**
 - OS: [e.g. Windows, Linux, ...] MacOS
 - Browser [e.g. Chrome, Safari] Chrome

**Screenshots**

**Additional context**

Regular post:

```html
<span class="post_content">    Post text here 
</span>
```

Markdown:

```html
<span class="post_content">    <div class="markdown " style=""><p>Post text here</p>
</div>
</span>
```

pankaj-2503:
Assign this issue to me i can solve this issue.

bmcutler:
Hi @pankaj-2503 

Thank you for your interest in the Submitty open source project.
We welcome contributions from new developers!
However we do not use the Github issue 'assign' feature for first time prospective contributors.

Please read our documentation on how to to get started with Submitty:
https://submitty.org/developer/getting_started/index
Specifically in setting up your development environment:
https://submitty.org/developer/getting_started/vm_install_using_vagrant
And making a pull request:
https://submitty.org/developer/getting_started/make_a_pull_request
You do not need to be assigned to an issue to create a pull request that will be reviewed by our team and then merged if it appropriately resolves the issue.
We also encourage you to join our Zulip server to discuss technical questions:
https://submitty.org/index/contact

PR #10118, opened 2024-01-19
[Bugfix:Forum] Indentation w/ Markdown Disabled

hansongu123:
First part of #10070 

Before:
image
After:
image

Barb on Zulip:
"...the indentation on the forum is looking a little squirrelly to me on some threads/some posts within the thread. See how each post in this screenshot has a little extra space at the start of the first line? But It doesn't happen on every post in every thread." 

When a forum post is created (including replies) without markdown option toggled, the text is indented automatically. The text with markdown appears to be normal. 

### What is the new behavior?

Forum post text is indented correctly regardless.

### DIFF
--- a/site/app/templates/forum/RenderPost.twig
+++ b/site/app/templates/forum/RenderPost.twig
@@ -1,7 +1,7 @@
 {% if render_markdown %}
     {% include "misc/Markdown.twig" with { content: post_content } only %}
 {% else %}
-    {{ post_content }}
+    <p>{{ post_content }}</p>>
 {% endif %}
 {% if post_attachment is defined and post_attachment.exist == true %}
     <table>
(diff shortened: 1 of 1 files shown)


site/app/templates/forum/RenderPost.twig
```diff
@@ -1,7 +1,7 @@
 {% if render_markdown %}
     {% include "misc/Markdown.twig" with { content: post_content } only %}
 {% else %}
-    {{ post_content }}
+    <p>{{ post_content }}</p>>
```
jason490, line 4:
```suggestion
    <p>{{ post_content }}</p>
```
Otherwise this PR fixes the issue of an extra space

jason490 (changes requested):

jason490 (approved):
Looks good and fixes the added spaces

PR #10125, opened 2024-01-24
[UI/UX:Calendar] Fix calendar announcement cursor

### LINKED ISSUE #9248: Cursor changing when hovering announcement

sadam-04:

Fixes #9248 
When moving cursor onto an announcement in the calendar, the cursor changes to a pointer as if the announcement is clickable for a split second, before changing back.

### What is the new behavior?
The cursor does not change at all when dragged onto the announcement.

### Other information?
This is not a breaking change.
I tested locally from a student login as well as an instructor login.


cjreed121 (changes requested):
For students, the announcements shouldn't appear clickable. With your change it is a pointer when they hover which when they click it does nothing. It should remain the default mouse cursor for the entire time they hover.

jeffreyC4 (approved):
The changes look good. I verified for both instructor and student accounts. Removing the cursor assignment in developer tools re-introduced the problem, so I believe this properly addresses the issue.

cjreed121 (approved):

Issue #10126, opened 2024-01-24
Forum reply button should be disabled when text area is empty


williamjallen:
**Describe the bug**
The "submit reply to all" button is enabled and clickable when no content has been entered in the forum text box.

**Expected behavior**
The submit button should be disabled when no content has been entered.

**Screenshots**
![image](https://github.com/Submitty/Submitty/assets/16820599/226958de-937f-490c-b6e1-e52eec71b0ca)

NewtonFernandis:
Hey @williamjallen, I will take this up. Could you please assign this to me?

jeffreyC4:
I think I can handle this issue by adding event listener to text-area.

byteWizard07:
Hi @williamjallen, I would like to work on this issue. please assign it to me.

Abhishek3777:
Hi @williamjallen , I can work on this issue.

shivanshusawan66:
Hi @williamjallen  I would like to work on this issue.

PR #10129, opened 2024-01-25
[Bugfix:Forum] Disable Forum Reply Button

jeffrey-asm:

Fixes [#10126](https://github.com/Submitty/Submitty/issues/10126)
When a user is currently on a forum thread, any reply button is always enabled, regardless of the empty text area above. This can be misleading to users and it would be better if reply buttons were always disabled on empty text area input.

### What is the new behavior?
The reply buttons are now always set to disabled whenever the text area is empty. This is checked on page load, thread change, and any change in text area input. This also applies to replying to any thread block or even editing your post on the forum.

### Other information?
This is a small change. Should be working as intended as it was tested on student and instructor accounts. Changes were tested for posts created by current users and posts created by others.

image

image


site/public/js/forum.js
```diff
@@ -2417,6 +2417,37 @@ function clearReplyBoxAutosave(replyBox) {
...
 
+function setupDisableReplyThreadForm() {
+    const threadPostForms = document.querySelectorAll('.thread-post-form');
+
+    threadPostForms.forEach((form) => {
+        //For all thread forms either reply's or posts, ensure that when text area is empty, the submit button appears to be disabled
+        const textArea = form.querySelector('textarea');
+        const submitButton = form.querySelector('input[type="submit"]');
+
+        if (textArea.id === 'reply_box_0' || textArea.id === 'reply_box_1') {
```
DarthNyan, line 2428: #10109 is going to shift these from 0 and 1 to 1 and 2. If that gets merged first be sure to update this. Otherwise I'll adjust that when this gets merged.

DarthNyan (commented):
Overall looks good. Feature works and code looks good. 

I've started rerunning the cypress tests that failed as they can be flaky sometimes. If they keep failing, we can figure out why and either change this or change the tests to fit the new expected behavior.

sadam-04 (commented):
Tested the new behavior, looks good.
I noticed that the setupDisableReplyThreadForm() metod could possilby be called twice. I'm not sure if this could lead to problems with multiple identical event listeners being created.

NewtonFernandis (approved):

jeffreyC4:
The one failing Cypress tests works locally on my VM with current implemented changes.

site/public/js/forum.js
```diff
@@ -2417,6 +2417,37 @@ function clearReplyBoxAutosave(replyBox) {
...
 }
 
+function setupDisableReplyThreadForm() {
+    const threadPostForms = document.querySelectorAll('.thread-post-form');
+
+    threadPostForms.forEach((form) => {
+        //For all thread forms either reply's or posts, ensure that when text area is empty, the submit button appears to be disabled
+        const textArea = form.querySelector('textarea');
+        const submitButton = form.querySelector('input[type="submit"]');
+
```
DarthNyan, line 2427: I had a thought:
```suggestion
pbid = $('#edit-user-post').find('.thread-post-form').data('post_box_id');
```
will get the post_box_id for the edit post form (currently 0). Then below if you look for reply boxes pbid and pbid+1, that should resolve the conflicts with my PR and this should be ready to merge.

AavanVadiwala:
A trait that I found was that when replying if the user wants to upload a document without any text, it wouldn't be possible since the reply button isn't highlighted, preventing the user from replying. I do not know if this was intended behavior or not but it might be good to bring to attention.

AavanVadiwala:
> A trait that I found was that when replying if the user wants to upload a document without any text, it wouldn't be possible since the reply button isn't highlighted, preventing the user from replying. I do not know if this was intended behavior or not but it might be good to bring to attention.

barb says it's good that it requires text, so I think everything looks good

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -79,15 +80,17 @@ describe('Tests cases abut changing user pronouns', () => {
...
         cy.get('.thread-list-item').should('contain', 'Test pronouns display');
         cy.contains('Test pronouns display').find('.post_user_pronouns').should('contain', 'They/Them');
         cy.contains('Test pronouns display').find('.post_user_pronouns').click();
 
         //comment on the thread, verify pronouns is shown
         cy.get('.create-post-head').should('contain', 'Test pronouns display');
-        cy.get('#reply_box_2').type('my pronouns are They/Them{ctrl}{enter}');
+        // For future cypress tests involving elements with event listeners, you must
+        // manually trigger change event to call that specific event listener within the browser
+        cy.get('#reply_box_2').type('my pronouns are They/Them{ctrl}{enter}').trigger('change');
```
IDzyre, line 93: Does the {ctrl}{enter} need to be in this line, and 118?

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -27,20 +27,21 @@ describe('Tests cases abut changing user pronouns', () => {
...
+        cy.get('button[aria-label="Clear pronoun input"]').click({force: true}); //clear input using trash can
 
         //set to false
-        cy.get('#pronouns-forum-display').uncheck();
+        cy.get('#pronouns-forum-display').uncheck({force: true});
+        cy.get('#edit-pronouns-submit').click({force: true});
 
-        cy.get('#edit-pronouns-submit').click();
+        //Wait for HTML content to update
+        cy.wait(10000);
```
IDzyre, line 40: Is there a reason for such a long delay? If you are waiting for the page to update, would cy.waitPageChange() be a better option (defined in cypress/support/commands.js) If not, could you try and find something on the page that should be different and assert the presence/absence of the item with a longer timeout instead of a long wait?

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -126,19 +129,22 @@ describe('Tests cases abut changing user pronouns', () => {
...
         //verify pronouns exist and remove thread
         cy.contains('Test pronouns display').find('.post_user_pronouns').should('contain', 'They/Them');
-        cy.contains('Test pronouns display').find('.post_user_pronouns').click();
+        cy.contains('Test pronouns display').find('.post_user_pronouns').click({force: true});
         cy.get('.post_user_pronouns').should('contain', 'They/Them');
-        cy.get('.first_post').find('.fa-trash').click();
-        cy.contains('Test pronouns display').should('not.exist');
+        cy.get('.first_post').find('.fa-trash').click({force: true});
+        // Wait for element to be removed and search by selector
+        cy.wait(10000);
```
IDzyre, line 136: Same comment about the wait, and on line 146

IDzyre (changes requested):

Vishesh17:
Hey, I'd like to work on it, can i get started?

hansongu123:
Just a nitpick on the disabling logic, I think it is better if the button is set to be disabled at default and enabled only if user types something into the field to reflect user action. It is also better if the user is allowed to post an attachment without a caption to avoid confusion.

ArnavPawar (approved):
It looks good. Just make sure that you adjust your code to work with what Tate says and after you change that it should be good to be merged.

iamdanielyun (approved):

JWei17 (approved):
I have test it out text area - Bottom disabled when text is empty & enable when is none empty.
Insert a attachment - Bottom disable when text is empty & not attach a file, enable when text empty & attach a file, and Bottom disable when text is empty & delete the attach file.
All those features looks fine!

IDzyre (commented):
Cypress tests look better, I have not tested the code at all.

williamjallen (changes requested):
Can you please add some Cypress tests for this?  Small features like this (particularly in the forum) are prone to being broken inadvertently, and such breakages often go unnoticed.

site/cypress/e2e/Cypress-Feature/test_forum.spec.js
```diff
@@ -45,6 +49,26 @@ const removeThread = (title) => {
...
+    // Enter text to submit and clear text to disable reply all
+    cy.get('[value="Submit Reply to All"]').should('be.disabled');
+    cy.get('#reply_box_2').type(reply).trigger('change');
+    cy.get('[value="Submit Reply to All"]').should('not.be.disabled');
+    cy.get('#reply_box_2').clear().trigger('change');
+    cy.get('[value="Submit Reply to All"]').should('be.disabled');
+
+    // Enter attachment to submit and clear attachment to disable reply all
+    cy.get(attachment).attachFile(attachment);
+    cy.wait(500);
```
IDzyre, line 65: These waits shouldn't be necessary

site/cypress/e2e/Cypress-Feature/test_forum.spec.js
```diff
@@ -45,6 +49,26 @@ const removeThread = (title) => {
...
+
+    // Enter text to submit and clear text to disable reply all
+    cy.get('[value="Submit Reply to All"]').should('be.disabled');
+    cy.get('#reply_box_2').type(reply).trigger('change');
+    cy.get('[value="Submit Reply to All"]').should('not.be.disabled');
+    cy.get('#reply_box_2').clear().trigger('change');
+    cy.get('[value="Submit Reply to All"]').should('be.disabled');
+
+    // Enter attachment to submit and clear attachment to disable reply all
+    cy.get(attachment).attachFile(attachment);
```
IDzyre, line 64: An example of how to attach file
```
cy.get('[data-testid="upload"]').selectFile('file/to/select', {action: 'drag-drop'});
```

jeffrey-asm:
Once [#10109](https://github.com/Submitty/Submitty/pull/10109) is merged into main, this will be updated to increment all occurrences of reply_box_x by 1.

site/cypress/e2e/Cypress-Feature/test_forum.spec.js
```diff
@@ -45,6 +46,20 @@ const removeThread = (title) => {
...
+    cy.get('.thread-left-cont > .thread-list-item').contains(title).click();
+    // Reply button should be disabled by default with no text
+    cy.get('[value="Submit Reply to All"]').should('be.disabled');
+
+    // Ensure reply button is not disabled when attachments are added
+    cy.get('#input-file2').attachFile(attachment);
+    cy.get('[value="Submit Reply to All"]').should('not.be.disabled').click();
+
+    // Wait for submission and ensure attachment with no text is visible
+    cy.get('.attachment-btn').should('exist').click();
```
IDzyre, line 59: You don't have to have the .should('exist') part because that is implied when you get the attachment button. You should only need to assert existence if you are only checking if it exists, or if it exists, and has a few other parameters (like CSS or class, or id, etc)

IDzyre (changes requested):
One small change then looks good after. I will approve once it is changed.

williamjallen (approved):
Code looks reasonable to me.

PR #10133, opened 2024-01-27
added a way to dynamically set the user's group when adding to course

octagonprogramming:

Currently when adding a user to a course they are first added as an instructor using the script

### What is the new behavior?
The new behavior allows the user's group to be set when running the script. It will default to instructor to keep original functionality

### DIFF
--- a/sbin/adduser_course.py
+++ b/sbin/adduser_course.py
@@ -30,6 +30,8 @@ def parse_args():
     parser.add_argument('course', help='title of the course')
     parser.add_argument('registration_section', nargs='?', default=None,
                         help='registration section that the user is added into')
+    parser.add_argument('user_group', default=1,
+                        help='group the user belongs to 1:Instructor 2:Full Access Grader 3:Limited Access Grader 4:Student 5:None')
 
     return parser.parse_args()
 
@@ -39,6 +41,7 @@ def main():
     user_id = args.user_id
     semester = args.semester
     course = args.course
+    user_group = args.user_group
     registration_section = args.registration_section
 
     conn_str = db_utils.generate_connect_string(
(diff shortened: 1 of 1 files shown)


prestoncarman:
LGTM

IDzyre:
I think the title would be better suited with something about adding the user group, since the script is coming later.

sbin/adduser_course.py
```diff
@@ -30,6 +30,8 @@ def parse_args():
     parser.add_argument('course', help='title of the course')
     parser.add_argument('registration_section', nargs='?', default=None,
                         help='registration section that the user is added into')
+    parser.add_argument('user_group', default=1,
+                        help='group the user belongs to 1:Instructor 2:Full Access Grader 3:Limited Access Grader 4:Student 5:None')
```
cjreed121, line 34:
```suggestion
                        help='group the user belongs to 1:Instructor 2:Full Access Grader 3:Limited Access Grader 4:Student')
```

We don't support a value of 5 so it's better to leave off. The DB constraint would fail if it was attempted.

cjreed121 (changes requested):
Thanks for the change! Just have one thing to update in the code and also please update the comment at the top of the file as the behavior is now slightly different. After those changes I will approve.

cjreed121:
Can you also update the comment at the top of the file. It says the user will always be added as an instructor when now it's changeable.

sbin/adduser_course.py
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 
 """
-Use this script to add a user to courses. Any user added to a course
-will be an instructor.
+Use this script to add a user to courses. You can choose the group
+that the user belongs to.
```
cjreed121, line 5:
```suggestion
Use this script to add a user to courses. By default, the user
will be an instructor but you can specify a different group
with an argument.
```

cjreed121 (changes requested):

cjreed121 (approved):

IDzyre (approved):

PR #10139, opened 2024-01-30
[Bugfix:TAGrading] Fix Peer Grading Status on Button

### LINKED ISSUE #10127: Grading Status Changed From One Peer Affecting Another
**Describe the bug**
On the Status page for a Gradeable with Peer Grading, if one Peer Component out of multiple for of a Gradeable is graded by User A for User C, then when a User B logs in to peer grade, they will see User C has "Grading Incomplete" even though User B has not started grading User C. Additionally, after User A finishes grading User C, User B will see User C as "0/{Total Points}" even though User B has not started grading User C.

**Expected behavior**
User A peer grading User C should not affect the status page of User B.

**To Reproduce**
1. As the instructor, create a gradeable with peer grading components.
2. Login as at least three students and submit to this gradeable.
3. As the instructor, in the gradeable's configuration, generate peer grading assignments in the peer grading tab.
4. As one of the students who submitted, grade one component of another student.
5. Login as the third student, navigate to the gradeable's status page, and see how the status page is partially filled in when it should not be.

**Configuration**
 - OS: Windows.
 - Browser: Firefox, Edge.

**Screenshots**
When User A grades one component of User C:
User A:
image

User B:
image
image

After User A finishes grading User C:
User A:
image

User B:
image

Xiayucheng1212:
Fixes #10127 

Currently one peer component's grade of User C will be affected by other peer graders. Let's say a User C has been assigned to User A and User B for peer grading. As long as User A start grading User C, the grading status shown for User B will also be affected, even User B has not started grading yet. 

### What is the new behavior?
User B(hilpef) will not be affected by User A(student), in any situation below:
1. User A "Grading incomplete":
截圖 2024-01-30 上午12 14 56
User B shows:
截圖 2024-01-30 上午12 14 09

3. User A Finish Grading:
截圖 2024-01-30 上午11 13 22
User B shows:
截圖 2024-01-30 上午12 14 09


site/app/templates/grading/electronic/Details.twig
```diff
@@ -470,19 +470,19 @@
             {% set contents = "Grade" %}
             {% set btn_class = "btn-primary" %}
 
-        {% if graded_gradeable.getOrCreateTaGradedGradeable().anyGrades() %}
+        {% if graded_gradeable.getOrCreateTaGradedGradeable().anyGrades(core.getUser()) %}
```
williamjallen, line 473: Using `core` in Twig files is discouraged.  Instead, you can pass the current user into the template.

williamjallen (changes requested):

site/app/templates/grading/electronic/Details.twig
micpap25: @williamjallen Perhaps an issue for cleaning out usage of core in Twig should be made? There are many such cases in the code, I could see how an external contributor would think it is good practice

micpap25:
I think we should just merge this PR and make a new issue for cleaning up core usage in Twig @williamjallen. There is a lot of Twig to clean up, and I'm more than happy to handle that myself, but I think leaving the code in a state where two different approaches to querying database info are used in the same file, let alone the same line, would lead to confusion.

micpap25 (changes requested):
@Xiayucheng1212 There is a major issue which is causing the checks to fail; this change also prevents TA graders from seeing the grade that TAs / instructors have assigned for non-peer grading. Is there any way you could change this code such that it does not affect non-peer grading? Thank you.

image

micpap25:
Marked as a draft; feel free to re-open once it is ready for review again!

zacharymnp:
@micpap25 Please correct me if I am wrong, but I think the last failing cypress test is in error. On main, the test `student should have access to some pdfs` passes because the buttons are labeled incorrectly; the PR fixes the buttons, causing it to fail.

The cypress test for the PR fails because no gradeable button label contains `/ 12`, but the `/ 12` labels are correctly changed to `Grade` in the PR.
image
^ main

micpap25:
> @micpap25 Please correct me if I am wrong, but I think the last failing cypress test is in error. On main, the test `student should have access to some pdfs` passes because the buttons are labeled incorrectly; the PR fixes the buttons, causing it to fail.
> 
> The cypress test for the PR fails because no gradeable button label contains `/ 12`, but the `/ 12` labels are correctly changed to `Grade` in the PR.

Yes, sounds correct to me, we can make that change.

micpap25 (changes requested):
Spoke to Zach in-person but needs a quick change (probably a flip of an if statement) to work

zacharymnp (approved):

site/app/models/gradeable/GradedGradeable.php
```diff
@@ -133,11 +133,12 @@ public function isTaGradingComplete() {
     /**
      * Gets whether a peer grader has graded all of the peer components for this submitter/gradeable
      * Later this will take in a userId and determine if that user graded all components
+     * @param User|null $grader Peer grader to check if all peer components associated with this grader has been graded.
      * @return bool
      */
-    public function isPeerGradingComplete() {
+    public function isPeerGradingComplete(User $grader = null) {
```
ziesski, line 139: Let add return type annotation while editing this function, and apply throughout

ziesski (changes requested):

ziesski (commented):
LGTM

PR #10140, opened 2024-01-30
[Feature:Submission] Add color description for rubric grading

ZheyuDeng:
Rows in rubric grading for peer review are highlighted without explanation for students.
image

### What is the new behavior?
There is now a question symbol next to it. If you put your cursor onto it "mark will be displayed to ALL students" is going to appear and you can click and jump to the explanatory website.
image


zhehua-zhang:
reviewed with code runned.
looks good.
I feel like the link only need to show once, instead of what is currently showes for every colored rubric.

williamjallen (commented):
In the screenshot provided in the PR description, the icon does not appear to be centered vertically, and is squished up against the text.

image

I also question whether this is the best possible UI choice.  In addition to linking to submitty.org documentation, perhaps it would be beneficial to add a tooltip as well.  As for the placement and choice of icon, it feels like it could be cleaner, but I'm struggling to come up with a better design.  The potential duplication of help icons is particularly concerning to me.  Perhaps it would make sense to move the help icon and some help text outside the mark itself.

ZheyuDeng:
@williamjallen 
We discussed in the meeting today and decided to go with this ui.
image

bmcutler (approved):
Looks great!  Thanks!
I tested this on Mac Chrome and Mac Firefox.
(I had to force reload the page in Chrome and clear my cache in Firefox to see the change)

PR #10169, opened 2024-02-08
[UI/UX:Developer] Black history month duck

ziesski:
Screen Shot 2024-02-08 at 1 52 51 AM

To embrace and welcome all community, we are adding black history month duck. 
The design is the three common theme color of  black history month. 
It will be visible randomly during entire february.


RayReyRam (approved):
I've tested this code with every log in, and the ducks appearance is correct through out. I've also went into different categories and the ducks appearance is also updated. The code changed looks good and clean , as well as commented with the type of duck style it is. These code is ready to merge

JoJo-ESC (approved):
Duck looks good ready to merge!

PR #10170, opened 2024-02-08
[Bugfix:InstructorUI] Cant use Notebook Builder in Gradeables

chenjohnny-cs:

Notebook builder is completely unable to be accessed whether no matter what upload configuration you choose. It should not be that way, and you should not be able to modify PROVIDED folder.

### What is the new behavior?
Notebook Builder cannot access the PROVIDED folder through edit, but is able to create a new notebook builder anywhere into the UPLOADED folder.

### Other information?
Linked to https://github.com/Submitty/Submitty/pull/5945

### DIFF
--- a/site/public/js/notebook_builder/widgets/widget.js
+++ b/site/public/js/notebook_builder/widgets/widget.js
@@ -1,5 +1,5 @@
 /* exported NUM_MARKDOWN, Widget */
-var NUM_MARKDOWN = 0;
+let NUM_MARKDOWN = 0;
 
 class Widget {
     /**
--- a/site/public/js/notebook_builder/widgets/widget.js
+++ b/site/public/js/notebook_builder/widgets/widget.js
@@ -1,5 +1,4 @@
 /* exported NUM_MARKDOWN, Widget */
-
 var NUM_MARKDOWN = 0;
 
 class Widget {
--- a/site/public/js/notebook_builder/widgets/widget.js
+++ b/site/public/js/notebook_builder/widgets/widget.js
@@ -1,6 +1,6 @@
 /* exported NUM_MARKDOWN, Widget */
 
-const NUM_MARKDOWN = 0;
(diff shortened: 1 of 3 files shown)


site/public/js/notebook_builder/widgets/widget.js
```diff
@@ -1,6 +1,6 @@
 /* exported NUM_MARKDOWN, Widget */
 
-const NUM_MARKDOWN = 0;
+var NUM_MARKDOWN = 0;
```
ziesski, line 3: This fails our js-lint check, could you explain why we need to change it from const to var?

ziesski (changes requested):

site/public/js/notebook_builder/widgets/widget.js
chenjohnny-cs: The markdown-widget.js is changing the value on NUM_MARKDOWN each time a markdown is created, with a const variable I believe that you can't change that value

I don't know if the export at the top changes something with the variable

chenjohnny-cs: I updated the variable from var to let to pass the js-lint check.

RayReyRam (approved):
Previously we talked about being unable to get into the notebook builder, this fixes it, you are able to enter the notebook builder without any issues. The code looks good, not much changes, cant seem to break anything and it passes all the lint test.

JoJo-ESC (commented):
Code looks good changes are seen and works as expected.

JoJo-ESC (approved):

ziesski (approved):
Code looks good to me, haven't tested the functionality.

JoJo-ESC (approved):
Changes look good tested functionality related to new builder as well as provision edit permissions and expected behavior was shown.

williamjallen (approved):
Code looks reasonable.  Did not test manually.

Issue #10171, opened 2024-02-08
Cursor chaging when hovering over Note in calendar


sadam-04:
**Describe the bug**
This bug is almost identical to the one described in issue [#9248](https://github.com/Submitty/Submitty/issues/9248), except in this one the issue is with Note entries in the calendar instead of Announcements.

When the cursor is dragged over the edge of a Note entry in the calendar from a student, grader, or ta login, the cursor changes to a pointer for a split second (as if the border is clickable) before switching back to the arrow cursor.

**Expected behavior**
The cursor should not change when moved over the edge of the Note.

**To Reproduce**
1. Open calendar view from student, grader, or ta login.
2. Find a Note object
3. Drag cursor over the edge of the Note

**Configuration**
 - OS: Windows 11
 - Browser: Chrome

**Screenshots**
Cursor changes too quickly to easily capture in a screenshot

PR #10172, opened 2024-02-08
[UI/UX:Calendar] Fix calendar Note cursor

sadam-04:

Fixes issue [#10171](https://github.com/Submitty/Submitty/issues/10171) 
This PR fixes an almost identical bug as was fixed in PR [#10125](https://github.com/Submitty/Submitty/pull/10125), except this fixes it for notes, instead of announcements in the calendar.

As the cursor is moved over the edge of a Note entry in the calendar view, the cursor changes to a pointer for a split second, before switching back to the default arrow cursor.

### What is the new behavior?
The cursor stays as an arrow, and doesn't change when dragged over the edge of Notes.

### Other information?
This is not a breaking change.
Tested from student, grader, and ta logins.

### DIFF
--- a/site/public/css/calendar.css
+++ b/site/public/css/calendar.css
@@ -202,12 +202,14 @@ a.cal-gradeable-status-text {
     padding: 0;
     border: none;
     box-shadow: none;
+    cursor: default;
 }
 
 a.cal-gradeable-status-text:hover {
     color: var(--text-black);
     padding: 0;
     border: none;
+    cursor: default;
 }
 
 a.cal-gradeable-status-text:focus {
--- a/site/public/css/calendar.css
+++ b/site/public/css/calendar.css
@@ -225,12 +225,13 @@ a.cal-gradeable-status-ann {
     border-style: solid;
     border-width: 3px;
     box-shadow: none;
+    cursor: default;
 }
 
 a.cal-gradeable-status-ann:hover {
+    cursor: default;
     padding: 0;
     font-style: italic;
-    cursor: default;
     border-style: solid;
     border-width: 3px!important;
 }
--- a/site/public/css/calendar.css
+++ b/site/public/css/calendar.css
@@ -202,14 +202,12 @@ a.cal-gradeable-status-text {
     padding: 0;
     border: none;
     box-shadow: none;
-    cursor: default;
 }
 
 a.cal-gradeable-status-text:hover {
     color: var(--text-black);
     padding: 0;
     border: none;
-    cursor: default;
 }
 
 a.cal-gradeable-status-text:focus {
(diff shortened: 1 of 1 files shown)


michaelkrar (approved):
Had difficulty replicating the issue at first. Looks like in a firefox browser(on ubuntu LTS) the bug is either too fast to see or doesn't occur, while in chrome it did visibly occur. (not very relevant but worth pointing out)

When testing the fix, the behavior in firefox appears to be identical to that without the fix, but when I tested it in chrome the behavior was fixed. 

Tested with student, ta, grader on chrome and firefox and the behavior was correct for all.
Also tested instructor specific functionalities in the calendar view (making new notes, announcements) and all worked as intended.
The fix also does not appear to be a breaking change.

arougey (approved):
Tested it in Firefox and could not reproduce the error.  Tested on student, instructor and ta logins. Based on the changed files, it looks like an effective fix for other browsers

site/public/css/calendar.css
```diff
@@ -202,13 +202,14 @@ a.cal-gradeable-status-text {
...
     box-shadow: none;
+    cursor: default;
 }
 
 a.cal-gradeable-status-text:hover {
     color: var(--text-black);
     padding: 0;
-    cursor: default;
     border: none;
+    cursor: default;
```
cjreed121, line 212: Revert this change

site/public/css/calendar.css
```diff
@@ -228,11 +229,11 @@ a.cal-gradeable-status-ann {
 }
 
 a.cal-gradeable-status-ann:hover {
-    cursor: default;
     padding: 0;
     font-style: italic;
     border-style: solid;
     border-width: 3px!important;
+    cursor: default;
```
cjreed121, line 236: Revert this change

cjreed121 (changes requested):
This does fix the issue and is good functionality wise but please address the following requested changes.

williamlin6803:
Tested it on my machine and it worked as normal.

ryvaru (approved):
Tested this code on my machine and everything works as intended, the cursor no longer changes when dragged over the edge of a note. Code looks logical and fixes the issue. Approved.

AavanVadiwala:
Test as an instructor, student, and ta. Approved. When the cursor is dragged along the edge of the border of a note, the cursor remains an arrow unless dragged into the note, which then the cursor is switched to a pointer.

AavanVadiwala (approved):
Test as an instructor, student, and ta. Approved. When the cursor is dragged along the edge of the border of a note, the cursor remains an arrow unless dragged into the note, which then the cursor is switched to a pointer.

cjreed121 (approved):

PR #10173, opened 2024-02-09
[Bugfix:Autograding] Copy all named executables to runner

### LINKED ISSUE #10121: autograding executable name must have a dot in the name
**Describe the bug**
When creating an autograding for a compiled language, the target executable name cannot simply be named e.g., "program", but it must be named "program.exe" or "program.out".  If it is named without a dot, compilation will succeed, and points awarded.  But all testing will fail.

I suspect it's a bug with file copying to the runner and/or permissions.

DarthNyan:

Fixes #10121 . Currently, the autograder only copies .out and .class files from the compiler to the runner. This means if the instructor creates a compilation testcase that compiles into anything other than .out or .class, the file will not be copied to the runner and other testcases will fail.

### What is the new behavior?
Now, in addition to copying all .out and .class files to the runner, any file that is listed in the "executable_name" field in the gradeable's config.json will also be copied to the runner. 

### Other information?

This needs more testing. I've been able to verify that it works when a single executable is produced, but I haven't been able to test whether it correctly handles an array of executable names. I'm currently working on a java example for this, and I'll upload it as a separate PR when it's done.

### DIFF
--- a/grading/load_config_json.cpp
+++ b/grading/load_config_json.cpp
@@ -96,9 +96,7 @@ void AddAutogradingConfiguration(nlohmann::json &whole_config) {
         if (exe_name->is_array()) {
           for (typename nlohmann::json::iterator itr = exe_name->begin(); itr != exe_name->end(); itr++) {
             if (itr->is_string()) {
-              std::string name = to_string(*itr);
-              // to_string function inserts "" characters around the executable name. Need to strip.
-              whole_config["autograding"]["compilation_to_runner"].push_back("**/" + name.substr(1, name.size() - 2));
+              whole_config["autograding"]["compilation_to_runner"].push_back("**/" + itr->template get<std::string>());
             }
             else {
               throw std::invalid_argument("Unable to parse provided executable_name");
@@ -106,9 +104,7 @@ void AddAutogradingConfiguration(nlohmann::json &whole_config) {
           }
         }
         else if (exe_name->is_string()) {
-          std::string name = to_string(*exe_name);
-          // to_string function inserts "" characters around the executable name. Need to strip.
-          whole_config["autograding"]["compilation_to_runner"].push_back("**/" + name.substr(1, name.size() - 2));
+          whole_config["autograding"]["compilation_to_runner"].push_back("**/" + exe_name->template get<std::string>());
         }
       }
     }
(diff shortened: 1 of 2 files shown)


williamlin6803 (commented):
Code looks good to me. I think this is a fantastic enhancement to the autograding system, and it directly addresses the limitations highlighted in issue #10121.

grading/load_config_json.cpp
```diff
@@ -84,6 +84,26 @@ void AddAutogradingConfiguration(nlohmann::json &whole_config) {
...
+          continue;
+        }
+        // Handle single file or array of files.
+        if (exe_name->is_array()) {
+          for (typename nlohmann::json::iterator itr = exe_name->begin(); itr != exe_name->end(); itr++) {
+            whole_config["autograding"]["compilation_to_runner"].push_back(*itr);
+          }
+        }
+        else {
+          whole_config["autograding"]["compilation_to_runner"].push_back(*exe_name);
```
cjreed121, line 102: Should `**/` go in front of the executable name? I don't think we can be certain where they compiled the file.

grading/load_config_json.cpp
```diff
@@ -95,11 +95,20 @@ void AddAutogradingConfiguration(nlohmann::json &whole_config) {
         // Handle single file or array of files.
         if (exe_name->is_array()) {
           for (typename nlohmann::json::iterator itr = exe_name->begin(); itr != exe_name->end(); itr++) {
-            whole_config["autograding"]["compilation_to_runner"].push_back(*itr);
+            if (itr->is_string()) {
+              std::string name = to_string(*itr);
```
cjreed121, line 99: `to_string` is not the best choice here. According to the [docs](https://json.nlohmann.me/api/basic_json/to_string/), `to_string` is used for getting the serialized result which is why you have to strip the quotes. I think you should just be able to get the value from the iterator. Or maybe std's to_string could help. There is definitely a way to accomplish this without stripping quotes.

cjreed121 (changes requested):

cjreed121 (approved):
Code looks good

grading/load_config_json.cpp
```diff
@@ -84,6 +84,31 @@ void AddAutogradingConfiguration(nlohmann::json &whole_config) {
...
+            }
+            else {
+              throw std::invalid_argument("Unable to parse provided executable_name");
+            }
+          }
+        }
+        else if (exe_name->is_string()) {
+          whole_config["autograding"]["compilation_to_runner"].push_back("**/" + exe_name->template get<std::string>());
+        }
+      }
```
bmcutler, line 109: I think we should add an else case here to error out if its not an array or a string.
(similar to the error when the array contains a non-string).
More error messages to the instructor is better :)

bmcutler (changes requested):
Looks good overall, just asking for 1 more error message.

bmcutler (approved):
thanks for the refactor

PR #10176, opened 2024-02-10
[Bugfix:Forum] Fix Forum Post Spacing

### LINKED ISSUE #10070: Markdown posts have padding that non-markdown posts do not

hansongu123:
closes #10070 

#10118 fixes the first part of #10070 by containing the post text in a block, but this way it adds extra spacing on top and bottom of a post to match that of a markdown enabled post, causing there to be more wasted space.

### What is the new behavior?

This PR adds whitespace control to the non-markdown enabled text and removes the extra spacing for the posts w/ markdown.
A 5px top-margin is added to improve the look and feel.

Before:
image

Now:
image

### DIFF
--- a/site/app/templates/forum/CreatePost.twig
+++ b/site/app/templates/forum/CreatePost.twig
@@ -34,7 +34,7 @@
 
     {% endif %}
 
-    <div class='pre-forum'><span class="post_content">{% include "forum/RenderPost.twig" with { "post_content": post_content, "render_markdown": render_markdown } only %}</span></div>
+    <div class='pre-forum'><div class='post_content'>{% include "forum/RenderPost.twig" with { "post_content": post_content, "render_markdown": render_markdown } only %}</div></div>
 
     <hr style="margin-bottom: 3px;">
 
--- a/site/app/templates/forum/RenderPost.twig
+++ b/site/app/templates/forum/RenderPost.twig
@@ -1,7 +1,7 @@
 {% if render_markdown %}
     {% include "misc/Markdown.twig" with { content: post_content } only %}
 {% else %}
-    <p>{{ post_content }}</p>
+    {{- post_content -}}
 {% endif %}
 {% if post_attachment is defined and post_attachment.exist == true %}
     <table>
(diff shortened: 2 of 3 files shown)


Kush-Raval (approved):
I did a functionality test and there didn't seem to be any glitches. I tested with different size screens and nothing broke the general function of the page. 
The code changes make sense as to how you get the desired result.
Looks good to merge

jeffreyC4 (changes requested):
The fix looks good on my end. Works as intended for single lines and multiple paragraphs. I would change the selector to "post-content" to fix the CSS Lint issues. I would also look into removing ".post_content p:first-child/last-child" because there is no paragraph tag within any of the "post_content" containers.

hansongu123:
@jeffreyC4 there's <p> block created for every paragraph if the post is markdown enabled, and each paragraph get's the same top and bottom margining hence we want to remove them for the first and last paragraphs.  With disabling css lint for these two lines there should be no more failing check.

@Kush-Raval @jeffreyC4 guys can you re-review and make sure the white space control (- -) doesn't affect other things?

Kush-Raval (approved):
Ran this on my local machine and saw the desired effects. Messed with different screen sizes and formats and the change still worked. Drastic improvement on space inside posts.
Tested as instructor and student.
The code makes sense and I can see why it will result in the desired effect.
Ready to merge.

PR #10178, opened 2024-02-12
[Bugfix:InstructorUI] Unassign graders to grade team members

### LINKED ISSUE #10131: Team Peer Grading Generation Assigns Teams to Grade Themselves
**Describe the bug**
When on the Peer Matrix tab as the instructor, when you click "Generate" for the peer grading matrix, it will assign students to grade their own team's work.

**Expected behavior**
Students should only be assigned to grade other teams' work.

**To Reproduce**
1. As the instructor, create a gradeable with peer grading components and teams.
2. Login as at least three students. Have two of the students form a team together. Have the two teams (a team of 2 and a team of 1 or more) submit.
3. As the instructor, in the gradeable's configuration, generate peer grading assignments in the peer grading tab.
4. Notice how students in the same team are assigned to grade themselves.

**Configuration**
 - OS: Windows.
 - Browser: Firefox, Edge.

**Screenshots**
image
Notice how aphacker is assigned to grade millsm even though aphacker and millsm are on a team together.

GO-viper7:
Closes #10131

-  [✅] Screenshots are attached to Github PR if visual/UI changes were made

On the Peer Matrix tab as an instructor, "Generate" button generates graders and students they grade for the peer grading matrix, it is even assigning students to grade their own team's work.

### What is the new behavior?
Upon clicking "Generate", graders and students are generated and none of the students are assigned to grade their own team members.
buckrz, fadelo and abernl are team members. Thus, they are not present in assigned students list for abernl.

student-teams-buckrz
student-teams-fadelo

### Other information?
Along with the above issue, there was a major functionality bug which was unnoticed while adding new students for a grader

Upon clicking on edit students, it lets us to add new students for a grader who can even be a grader's team member

### What is the new behavior?
In case of adding a team member to a grader, it notifies that "unable to add a team member". Also disables the submit button.

student-teams


site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -132,14 +132,40 @@ public function generateHistogramData($overall_scores) {
...
     }
-
+    /**
+     * function to filter out students in the same team as Grader
+     * @param Gradeable $gradeable
+     * @param string $grader
+     * @param Array $students
+     * @return Array $filtered_grading_info
+     */
+    private function filterTeamStudents(Gradeable $gradeable, string $grader, array $students) {
```
Xiayucheng1212, line 142: This should be a resonable method but what if the following condition happens?
A team with 2 members are assigned to grad for each other and the `number_to_grade` equals to 1. After implementing the fileterTeamStudents() function, this team will not be graded by anyone. 
So how about adding a notice for those team assignments which are not properly assigned?

GO-viper7: sure! @Xiayucheng1212 
In this particular case, I can inform user about the issue and suggest on using  "Edit Students" to add a new student who isn't a team member for that grader
i hope any of the maintainers to give me suggestions on where to notify the user as we have lot of ways (using alert, on the peer matrix, UI, pop-up and etc.)

GO-viper7:
Hello @jeffreyC4, can you please review this PR if you don't mind.

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -323,13 +331,16 @@
                 $(error_div_id).text("Please enter a valid new grader that is in the course and not already in the matrix.");
             } else if (self_grading) {
                 $(error_div_id).text("Please note that student is not able to grade themselves.");
+            } else if (team_grading) {
+                $(error_div_id).text("Please note that student is not able to grade their team members.");
```
04vmatsibekker, line 335: Isn't all javascript supposed to be in single quotes and not double quotes?

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
```diff
@@ -323,13 +331,16 @@
...
             } else if (valid_students.includes(false)) {
                 $(error_div_id).text("Please enter valid new students that are in the course and not already assigned to the grader.");
             } else if (new_grader && !valid_students.includes(true)) {
                 $(error_div_id).text("Please enter at least one student for the new grader.");
             }
             $(error_div_id).show();
             $(button_id).prop('disabled', true);
+            team_grading = false //resetting the status for team_grading
```
04vmatsibekker, line 343: add semi-colon. Make sure the java script code in here is up to code standards

04vmatsibekker (changes requested):

site/app/templates/admin/admin_gradeable/AdminGradeablePeers.twig
GO-viper7: Sorry for the delay.

and Yes, almost all the js files sticking with the single quotes
So, changed all of the double quotes to single.

GO-viper7:
Hello @04vmatsibekker 
As you can see php lint errors keep popping up even though I'm trying to clear the following : 
1. Loose checking the null is not allowed so I'm using !==,  though it does not resolve it and displays an additional error
2. the method `in_array` has some issues with the third argument even though I'm setting it to `true`

Can you please look into it. Let me know if you need anything

04vmatsibekker:
> Hello @04vmatsibekker As you can see php lint errors keep popping up even though I'm trying to clear the following :
> 
> 1. Loose checking the null is not allowed so I'm using `!==` instead of `!=`,  though it does not resolve it and displays an additional error
> 2. the method `in_array` has some issues with the third argument even though I'm setting it to `true`
> 
> Can you please look into it. Let me know if you need anything

I can try to take a look at the issue over the next couple of days.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -965,7 +995,7 @@ public function showDetails(string $gradeable_id) {
             foreach ($all_teams as $team) {
                 $student_ids = array_diff($student_ids, $team->getMembers());
                 $team_section = $gradeable->isGradeByRegistration() ? $team->getRegistrationSection() : $team->getRotatingSection();
-                if ($team->getSize() > 0 && (in_array($team_section, $sections) || $show_all)) {
+                if ($team->getSize() > 0 && (in_array($team_section, $sections, true) || $show_all)) {
```
04vmatsibekker, line 998: The cause for php-lint error I believe is that $team_section is of type int and $sections is of type array<string>, this is a type mismatch as either $sections should be type array<int> or $team_section should be type string.

Because of mismatching datatypes, this function is always returning false.

site/app/libraries/database/DatabaseQueries.php
04vmatsibekker, line 1: Line 927 "  if ($value != 'false') {"   --> should be !== and make $value is of type string not boolean at first when entering loop

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -4146,6 +4146,44 @@ public function removeFromSeekingTeam($g_id, $user_id) {
...
+                gt.g_id = ?
+            GROUP BY 
+                gt.team_id
+            ORDER BY
+                team_id",
+            [$gradeable->getId()]
+        );
+        $teams = [];
+        $rows = $this->course_db->rows();
+        if ($rows !== null) {
```
04vmatsibekker, line 4178: rows will always be of type array (never a null case), so the only thing it could be is an empty array. Therefore I don't think you need this check

04vmatsibekker (changes requested):
These are the changes that I think will perhaps fix your php lint issues, although I myself can't directly make these changes for you from the way I pulled your fork. So I leave making the changes up to you. Good luck! And let me know if you need help again.

site/app/controllers/grading/ElectronicGraderController.php
GO-viper7: So, I have used debugger and found out the following : 
team_section was actually giving string and
sections were giving array 

which are as intended, I still don't know the reason why I'm failing here

site/app/libraries/database/DatabaseQueries.php
GO-viper7: You are right 
it worked well when I checked the size of array instead of checking null

GO-viper7: Thanks for the suggestions @04vmatsibekker 
Most of them worked correctly

and right when I pushed the changes, new php static errors popped up
if its this way, I think there will be lot more lint/static errors 
May be I should let one of the maintainers know about this issue ?

GO-viper7:
Thanks for the suggestions @04vmatsibekker 
Most of them worked correctly

and right when I pushed the changes, new lint errors popped up
if its this way, I think there will be lot more lint/static errors
May be I should let one of the maintainers know about this issue ?

bmcutler:
@GO-viper7 
Can you please resolve the merge conflicts on this PR?  Thanks!

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -3753,7 +3753,7 @@ public function getSubmittersWhoGotMarkBySection($mark, $grader, $gradeable, $an
         // Switch the column based on gradeable team-ness
         $type = $mark->getComponent()->getGradeable()->isTeamAssignment() ? 'team' : 'user';
         // TODO: anon teams?
-        $user_type = ($type == 'user' && $anon != 'unblind') ? 'anon' : $type;
+        $user_type = ($type == 'user' && $anon !== 'unblind') ? 'anon' : $type;
```
ziesski, line 3756:
```suggestion
        $user_type = ($type === 'user' && $anon !== 'unblind') ? 'anon' : $type;
```

Should use more strict comparison

site/app/controllers/grading/ElectronicGraderController.php
ziesski: Lets use alert for now.

ziesski (changes requested):

GO-viper7:
@ziesski , could you please specify on using alerts ?
Are you suggesting to not to filter the students and instead make an alert if students are present in same team as grader's ?

ziesski:
> expected

Sorry for late response, and Yes, after generate button

ZheyuDeng (approved):
I fixed a couple merge conflict and php lint errors. Also verified this PR fixed the bug described in the issue. I had aphacker and student in a group and they weren't assigned to grade each other.
image

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -701,6 +701,12 @@ public function findThread($thread_id) {
         return $this->course_db->row();
     }
 
+    public function existsAnnouncementsId(int $thread_id): bool {
```
zacharymnp, line 704: Is this function doing anything? It appears to be unused, and I'm not sure what it has to do with peer grading.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -966,13 +972,44 @@ public function visitThread($current_user, $thread_id, string $last_viewed_post_
...
      * @param int $thread_id
-     * @return null
-     */
-    public function unreadThread(string $current_user, int $thread_id) {
-        $this->course_db->query("DELETE FROM viewed_responses where thread_id = ? and user_id = ?", [$thread_id, $current_user]);
-        return null;
+     * @param bool $newStatus
+     * @return bool|null
+     */
+    public function setDeletePostStatus(int $post_id, int $thread_id, bool $newStatus): bool|null {
```
zacharymnp, line 980: Is this function doing anything? It appears to be unused, and I'm not sure what it has to do with peer grading.

zacharymnp (changes requested):

site/public/js/admin-gradeable-updates.js
```diff
@@ -452,6 +452,14 @@ function setRandomGraders(gradeable_id, p_values, successCallback, errorCallback
             if (res.data === 'Clear Peer Matrix') {
                 $('#save_status').text('Peer Matrix Cleared').css('color', 'var(--text-black)');
             }
+            else {
+                for (let i = 0; i < res.data.length; i++) {
```
zacharymnp, line 456: Could this loop be modified to include in the alert the list of graders that have not been assigned to a given number of students? If **Generate** is used with a low number of students, a given grader may not have any students to grade, which results in them not showing up in the GUI. This might make it difficult for instructors to manually assign them students to grade.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -965,14 +965,11 @@ public function visitThread($current_user, $thread_id, string $last_viewed_post_
...
 
-    /**
-     * @param string $current_user
-     * @param int $thread_id
-     * @return null
-     */
-    public function unreadThread(string $current_user, int $thread_id) {
-        $this->course_db->query("DELETE FROM viewed_responses where thread_id = ? and user_id = ?", [$thread_id, $current_user]);
-        return null;
+    public function unreadThread(string $current_user, int $thread_id): void {
```
zacharymnp, line 968: Why is this function being changed? Is it related to the PR?

zacharymnp (changes requested):
Before changes, graders would be assigned to grade students on their team.
All Grade All without changes

After changes, graders are not assigned to grade students on their team.
All Grade All with changes

My main concern is that the 'Some of the graders haven't yet been assigned to a given number of students, please manually edit them' alert pops up every single time the **Generate** button is used. An issue should definitely be made to make it less likely, and the popup should convey more information to make it easier for the instructor to find which graders they need to manually edit.

bmcutler:
As I noted on the original issue, this desired solution to this problem is incompletely specified.  This PR results in an uneven number of reviewers for each team submission.  Some teams may end up with few or even no assigned reviewers.  And other teams may end up with many more reviewers than desired.

I've closed the original issue as underspecified and low-priority.

PR #10184, opened 2024-02-14
[Feature:SubminiPolls] Poll-Timer

Kush-Raval:

* [x] Poll stops accepting after the given Time Interval
* [x] Does not take away the ability to close and open polls manually
* [x] Does not crash or have any bugs

Instructors have to open manually and close polls

### What is the new behavior?
Give instructors an option to add a timer to the poll where as soon as the "accept answers" checkbox is ticked it will start a timer and will close after the given duration.

How To Test:
Make a new poll in instructor and make it last different amounts of time, try accessing the poll in student accounts as well.


site/app/controllers/PollController.php
```diff
@@ -429,9 +449,22 @@ public function openPoll(): RedirectResponse {
...
+        if ($duration->h > 0 || $duration->i > 0 || $duration->s > 0 || $duration->days > 0 || $duration->m > 0 || $duration->y > 0) {
+            // Duration evaluates to > 0
+            $end_date = $this->core->getDateTimeNow();
+            $end_date->add($duration);
+            $poll->setEndDate($end_date);
+        }
+        else {
+            // If duration is 0, set end date to a future date.
+            //If duration is 0, it means that the user wants to manually close it.
+            $end_date = new \DateTime("9999-01-01T00:00:00");
```
williamjallen, line 463: It would be better to use `DateUtils::MAX_TIME` here

site/app/controllers/PollController.php
```diff
@@ -467,9 +501,10 @@ public function closePoll(): RedirectResponse {
             $this->core->addErrorMessage("Invalid Poll ID");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
-        $poll->setClosed();
+        //Setting the time to the beginning of time indicates that it is closed.
+        //Should be based on status.
+        $poll->setEndDate(new DateTime('1900-01-01'));
```
williamjallen, line 506: Same here: use `DateUtils::BEGINING_OF_TIME` instead.

site/app/entities/poll/Poll.php
```diff
@@ -58,16 +66,17 @@ class Poll {
     #[ORM\JoinColumn(name: "poll_id", referencedColumnName: "poll_id")]
     protected Collection $responses;
 
-    public function __construct(string $name, string $question, string $question_type, \DateTime $release_date, string $release_histogram, string $release_answer, string $image_path = null) {
+    public function __construct(string $name, string $question, string $question_type, \DateInterval $duration, \DateTime $release_date, string $release_histogram, string $release_answer, string $image_path = null, ) {
         $this->setName($name);
         $this->setQuestion($question);
         $this->setQuestionType($question_type);
+        $this->setDuration($duration);
+        $this->setEndDate(new \DateTime("1900-01-02"));
```
williamjallen, line 74: Is there a reason for this default end date?

site/app/templates/polls/ViewPoll.twig
```diff
@@ -1,6 +1,22 @@
...
+        {% endif %}
+        {% if duration.h > 0 %}
+            {% set durationString = durationString ~ duration.h ~ ' hour(s) ' %}
+        {% endif %}
+        {% if duration.i > 0 %}
+            {% set durationString = durationString ~ duration.i ~ ' minute(s) ' %}
+        {% endif %}
+        {% if duration.s > 0 %}
+            {% set durationString = durationString ~ duration.s ~ ' second(s) ' %}
+        {% endif %}
```
williamjallen, line 17: You can make the UI ever so slightly by handling the singular and plural cases automatically, rather than adding `(s)` to the end of every string.

site/public/css/polls.css
```diff
@@ -116,6 +116,23 @@ form {
...
+#poll-hours{
+    width: 60px;
+    margin-right: 8px;
+    margin-bottom: 10px;
+}
+
+#poll-minutes{
+    width: 60px;
+    margin-bottom: 10px;    
+}
```
williamjallen, line 134: It is generally better to use text-relative units (`em`) instead of raw pixel values for things like this.  I believe you can do this with the existing CSS code for the page though, so this shouldn't be needed anyway.

migration/migrator/data/course_tables.sql
```diff
@@ -1533,7 +1533,9 @@ CREATE TABLE public.polls (
     image_path text,
     question_type character varying(35) DEFAULT 'single-response-multiple-correct'::character varying,
     release_histogram character varying(10) DEFAULT 'never'::character varying,
-    release_answer character varying(10) DEFAULT 'never'::character varying
+    release_answer character varying(10) DEFAULT 'never'::character varying,
+    duration character varying(255) DEFAULT 'P0D'::character varying,
```
williamjallen, line 1537: This should be converted to a proper interval type before this PR can be merged.  If that turns out to be challenging, you could store it as integer seconds if necessary.

williamjallen (changes requested):

site/app/entities/poll/Poll.php
Kush-Raval: It should 1900-01-01, I changed that.

site/app/templates/polls/ViewPoll.twig
Kush-Raval: This is just to test the duration and time. It should be the Poll Name and a JS timer displayed.

site/app/controllers/PollController.php
```diff
@@ -429,9 +449,22 @@ public function openPoll(): RedirectResponse {
             $this->core->addErrorMessage("Invalid Poll ID");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
+
+        $duration = $poll->getDuration();
+        if ($duration->h > 0 || $duration->i > 0 || $duration->s > 0 || $duration->days > 0 || $duration->m > 0 || $duration->y > 0) {
```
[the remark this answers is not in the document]
Kush-Raval: Isn't a very standard way of checking if the duration is empty. I made it if format() === "00000000" I think it is a bit more concise.

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
```diff
@@ -0,0 +1,36 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("ALTER TABLE polls ADD COLUMN IF NOT EXISTS duration VARCHAR(255) DEFAULT 'P0D'")
+    database.execute("ALTER TABLE polls ADD COLUMN IF NOT EXISTS end_date timestamptz NOT NULL DEFAULT '9999-01-01'")
```
04vmatsibekker, line 18: Could you combine into one execute statement?

site/app/controllers/PollController.php
```diff
@@ -190,8 +193,17 @@ public function addNewPoll(): RedirectResponse {
                     $this->core->addErrorMessage("Invalid poll answer release setting");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
+        //Handle the two inputs, both will be ints (one hours, one minutes). I have to combine both of them into a "DateInterval" and create the PollObject based on that.
```
04vmatsibekker, line 196: fix comments when your done

site/app/controllers/PollController.php
```diff
@@ -309,10 +321,24 @@ public function submitEdits(): RedirectResponse {
...
+        if ($poll->isOpen()) {
+            $endDate = $this->core->getDateTimeNow();
+            $endDate->add($newDuration);
+            $poll->setEndDate($endDate);
+        }
         if ($date === false) {
             $this->core->addErrorMessage("Invalid poll release date");
             return new RedirectResponse($returnUrl);
         }
+
```
04vmatsibekker, line 341: Remove whitespace

04vmatsibekker (changes requested):

site/app/controllers/PollController.php
Kush-Raval: Fixed

site/app/controllers/PollController.php
Kush-Raval: Fixed

site/public/css/polls.css
Kush-Raval: I removed the poll-hours and poll-minutes and concise it to just poll-timer. And I switched from pixels to em. 
I didn't find anything else in the poll.css with small boxes I could put into.

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
Kush-Raval: I will leave it at is, because I also added another thing and it might be complicated to see it all in one execute.

site/app/controllers/PollController.php
```diff
@@ -651,7 +686,7 @@ public function importPollsFromJSON(): RedirectResponse {
                 implemented don't have this data. At the time, there
                 only existed questions of type single response. */
             $question_type = array_key_exists("question_type", $poll) ? $poll['question_type'] : 'single-response-multiple-correct';
-            $poll_entity = new Poll($poll['name'], $poll['question'], $question_type, \DateTime::createFromFormat("Y-m-d", $poll['release_date']), $poll['release_histogram'], $poll['release_answer']);
+            $poll_entity = new Poll($poll['name'], $poll['question'], $question_type, new \DateInterval($poll['duration']), \DateTime::createFromFormat("Y-m-d", $poll['release_date']), $poll['release_histogram'], $poll['release_answer']);
```
williamjallen, line 689: Is `$poll['duration']` allowed to be empty here?

site/app/controllers/PollController.php
```diff
@@ -429,9 +452,21 @@ public function openPoll(): RedirectResponse {
             $this->core->addErrorMessage("Invalid Poll ID");
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
-        $poll->setOpen();
-        $em->flush();
 
+        $duration = $poll->getDuration();
+        if ($duration->format('%r%Y%m%d%H%i%s') !== '000000000000') {
```
williamjallen, line 457: Is there a cleaner way to do this, potentially by comparing against a 0 date interval?

site/app/entities/poll/Poll.php
```diff
@@ -49,6 +58,7 @@ class Poll {
     #[ORM\OneToMany(mappedBy: "poll", targetEntity: Option::class, orphanRemoval: true)]
     #[ORM\JoinColumn(name: "poll_id", referencedColumnName: "poll_id")]
     #[ORM\OrderBy(["order_id" => "ASC"])]
+
```
williamjallen, line 61: this whitespace should be reverted.

site/app/entities/poll/Poll.php
```diff
@@ -27,6 +28,14 @@ class Poll {
 
     #[ORM\Column(type: Types::TEXT)]
     protected $status;
+    /**
+     * @var string
+     */
+    #[ORM\Column(type: Types::STRING)]
+    protected $duration;
```
williamjallen, line 35: Is there a reason why this is a PHPDoc annotation?  It would be better to use a native PHP type annotation if possible.
```suggestion
    #[ORM\Column(type: Types::STRING)]
    protected string $duration;
```

site/app/entities/poll/Poll.php
```diff
@@ -91,39 +103,52 @@ public function getQuestion(): string {
...
     }
-
     public function setOpen(): void {
-        $this->status = "open";
+        $this->end_date = new \DateTime(DateUtils::MAX_TIME);
+    }
+    public function setEnded(): void {
+        $temp = DateUtils::getDateTimeNow();
+        $tempString = $temp->format('Y-m-d');
+        $this->end_date = new DateTime($tempString);
```
williamjallen, line 115: Is there a reason why all of this logic is necessary?  Why can't you just use `$this->end_date = DateUtils::getDateTimeNow();` directly?

site/app/entities/poll/Poll.php
```diff
@@ -91,39 +103,52 @@ public function getQuestion(): string {
...
+        $now = DateUtils::getDateTimeNow();
+        return $now < $this->end_date;
     }
 
-    public function setClosed(): void {
-        $this->status = "closed";
+    public function isEnded(): bool {
+        $now = DateUtils::getDateTimeNow();
+        $closeDate = '1900-02-01T00:00:00';
+        return $now > $this->end_date && $this->end_date->format('Y-m-d\TH:i:s') !== $closeDate;
```
williamjallen, line 125: It would be better to compare DateTime objects rather than their corresponding strings.

site/app/entities/poll/Poll.php
```diff
@@ -91,39 +103,52 @@ public function getQuestion(): string {
...
     }
 
     public function isClosed(): bool {
-        return $this->status === "closed";
+        $now = DateUtils::getDateTimeNow();
+        return $now > $this->end_date && $this->end_date->format('Y-m-d') === '1900-02-01';
```
williamjallen, line 130: Same comment here.  It would be better to compare against a DateTime object.

site/app/templates/polls/AllPollsPageInstructor.twig
```diff
@@ -262,4 +262,4 @@
             </tbody>
         </table>
     </div>
-</div>
+</div>
```
williamjallen, line 265: This whitespace-only change appears to be inadvertent and should be reverted.

site/tests/app/entities/poll/PollTester.php
```diff
@@ -35,13 +38,14 @@ public function setUp(): void {
...
                 "multiple-response-survey",
+                new DateInterval("PT1M30S"),
                 new DateTime('now'),
                 "when_ended",
                 "when_ended",
                 "/var/local/submitty/courses/s21/sample/uploads/polls/poll_image_3_colors.png"
             )
         ];
-
+        $this->my_polls[0]->setClosed();
```
williamjallen, line 48: Why is this change necessary?

site/tests/app/libraries/PollUtilsTester.php
```diff
@@ -47,12 +51,13 @@ public function testExportDataWithNonEmptyPolls() {
...
                 "Is this the fourth poll?",
                 "multiple-response",
+                new DateInterval("PT1M30S"),
                 new DateTime("2020-01-13"),
                 "when_ended",
                 "when_ended"
             ),
         ];
-
+        $polls[0]->setClosed();
```
williamjallen, line 60: Same here.  Why is this change necessary?

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
```diff
@@ -0,0 +1,40 @@
...
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("ALTER TABLE polls ADD COLUMN IF NOT EXISTS duration VARCHAR(255) DEFAULT 'P0D'")
+    database.execute("ALTER TABLE polls ADD COLUMN IF NOT EXISTS end_date timestamptz NOT NULL DEFAULT '9999-02-01'")
+    database.execute("ALTER TABLE polls ALTER COLUMN status SET NOT NULL")
+    database.execute("UPDATE polls SET end_date = CASE "
+                      "WHEN status = 'open' THEN '9999-02-01' "
+                      "WHEN status = 'closed' THEN '1900-02-01' "
+                      "ELSE now() END")
```
williamjallen, line 23: You could use a string denoted by triple-quoted string to avoid needing multiple strings.
```suggestion
    database.execute("""
        UPDATE polls SET end_date = CASE
            WHEN status = 'open' THEN '9999-02-01'
            WHEN status = 'closed' THEN '1900-02-01'
            ELSE now() END
    """)
```

williamjallen (changes requested):
Everything seems to be on track to be merged soon.  This is definitely now in the final stages of development.

A few things that still need to be addressed:
1. All of the test need to be passing.  It looks like you've changed some of the behavior when it comes to deciding the visibility of a poll, so those tests need to be figured out.
2. The DB column type still needs to be worked on some.  It would be best to use a native Interval type, but integer seconds would also be alright if you can't get that working.

site/app/entities/poll/Poll.php
Kush-Raval: Fixed

site/app/entities/poll/Poll.php
Kush-Raval: Fixed

site/tests/app/entities/poll/PollTester.php
Kush-Raval: When a poll is constructed the endDate is set to the beginning of time '1900-02-01' if I don't set it closed, it won't correctly test the functionality. It will just compare with '1900-02-01' for each poll.

site/tests/app/libraries/PollUtilsTester.php
Kush-Raval: Same reason as above poll endDate is initialized at beginning of Time.

site/tests/app/entities/poll/PollTester.php
Kush-Raval: Wait, you are right it shouldn't be necessary.

site/tests/app/libraries/PollUtilsTester.php
Kush-Raval: nevermind

site/app/templates/polls/AllPollsPageInstructor.twig
Kush-Raval: Fixed

site/app/entities/poll/Poll.php
Kush-Raval: This method is only used for the test. I need it to be in Y - m - d format because I am checking for specifically that and have set the "expected output" in the test to match this format. It wouldn't work with hours minutes and seconds because those could deviate.

site/app/controllers/PollController.php
Kush-Raval: I tried multiple ways and ended up reverting this and checking each duration member variable separately. That was the only way it seemed to work. This and comparing it against a 0-date interval didn't work.

site/app/entities/poll/Poll.php
Kush-Raval: Fixed this.

site/app/entities/poll/Poll.php
Kush-Raval: Fixed

ryvaru (dismissed):
When creating or editing a poll, if an instructor presses the down arrow on the default (or zero) value, it goes to a negative value for the hours, minutes, and seconds boxes. This should not be the case, it should instead jump to a high positive value.

Kush-Raval:
> When creating or editing a poll, if an instructor presses the down arrow on the default (or zero) value, it goes to a negative value for the hours, minutes, and seconds boxes. This should not be the case, it should instead jump to a high positive value.

site/app/controllers/PollController.php
Kush-Raval: By default, a duration will always have 0. It will never be empty.

site/app/entities/poll/Poll.php
Kush-Raval: Reverted, but made it look cleaner by checking with static variable BEGINNING_OF_TIME.

site/app/entities/poll/Poll.php
Kush-Raval: Reverted, but same as last comment.

site/tests/app/entities/poll/PollTester.php
Kush-Raval: Fixed

site/app/entities/poll/Poll.php
```diff
@@ -91,39 +98,74 @@ public function getQuestion(): string {
...
+
+        $days = floor($seconds / (24 * 60 * 60));
+        $seconds -= $days * (24 * 60 * 60);
+
+        $hours = floor($seconds / (60 * 60));
+        $seconds -= $hours * (60 * 60);
+
+        $minutes = floor($seconds / 60);
+        $seconds -= $minutes * 60;
+        return new \DateInterval("P{$years}Y{$months}M{$days}DT{$hours}H{$minutes}M{$seconds}S");
```
hansongu123, line 144: getDuration() should only be expected to return the duration of the poll, in seconds, rather than wrapped as a `DateInterval` object - we should do this manipulation elsewhere if needed.

site/app/entities/poll/Poll.php
```diff
@@ -91,39 +98,74 @@ public function getQuestion(): string {
...
+        $totalSeconds += $duration->d * 86400;
+        $totalSeconds += $duration->m * 2592000;
+        $totalSeconds += $duration->y * 31536000;
+        $this->duration = $totalSeconds;
+    }
+
+    public function setEndDate(\DateTime $end_date): void {
+        $this->end_date = $end_date;
+    }
+
```
hansongu123, line 168: Again, we should also simplify the setting of a poll's duration.

migration/migrator/data/course_tables.sql
```diff
@@ -1528,12 +1528,14 @@ CREATE TABLE public.polls (
...
-    status text NOT NULL,
+    status text,
     release_date date NOT NULL,
     image_path text,
     question_type character varying(35) DEFAULT 'single-response-multiple-correct'::character varying,
     release_histogram character varying(10) DEFAULT 'never'::character varying,
-    release_answer character varying(10) DEFAULT 'never'::character varying
+    release_answer character varying(10) DEFAULT 'never'::character varying,
+    duration integer DEFAULT 0,
+    end_date timestamp with time zone DEFAULT '1900-02-01 00:00:00-05'::timestamp with time zone NOT NULL
```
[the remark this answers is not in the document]
hansongu123: Also, what's the purpose of "duration" if a poll already has "end_date (time)", we can just calculate the difference between "now" and when the poll is ended... this way the poll's duration wouldn't reset if an instructor wants to edit a poll's other attributes like title or questions.

site/app/templates/polls/PollForm.twig
```diff
@@ -87,7 +87,15 @@
...
         <br/>
-
+        
+        <br/>
+        <label for="poll-hours" class="option-title">
+            Expected Duration of Poll:
+        </label>
+        <input type="number" name="poll-hours" id="poll-hours" class="poll-timer" style="width: 70px" min=0 placeholder="Hour" value="{{ poll is not null ? poll.getDuration().h }}" />
+        <input type="number" name="poll-minutes" id="poll-minutes" class="poll-timer" style="width: 70px" min=0 placeholder="Min" value="{{ poll is not null ? poll.getDuration().i }}" />
+        <input type="number" name="poll-seconds" id="poll-seconds" class="poll-timer" style="width: 70px" min=0 placeholder="Sec" value="{{ poll is not null ? poll.getDuration().s }}" />
```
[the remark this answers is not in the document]
hansongu123: There should also be an indication if a poll timer is set to infinity rather than just empty fields - maybe a button to toggle on/off the feature?

hansongu123 (changes requested):
The changes look good in general. I left some comments about the code. 

For this feature to provide better UX, there seems to be still a lot a work to do for the look and feel, such as a better timer/picker for instructor's UI and a countdown timer for student's UI. For now, I think we can just settle on some indication of when a poll is out of time, such as a toast or a message on the page.

migration/migrator/data/course_tables.sql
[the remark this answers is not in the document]
Kush-Raval: we get the end_Date based on duration by appending the duration given by input to DateTimeNow(). The reason we can't just do this with end_date is we need to hold the values the instructor gives for the duration and set the endDate based on when they click "accepting poll responses"

site/app/templates/polls/PollForm.twig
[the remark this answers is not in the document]
Kush-Raval: I don't quite follow, the instructors can see which boxes are hours minutes and seconds because of the placeholder. If the instructor doesn't do anything the poll-timer doesn't do anything, the functionality of the poll remains unchanged. Therefore, I don't think we need to set an on/off feature it is just an optional feature they can use if they want.

site/app/entities/poll/Poll.php
Kush-Raval: The way I have it, outside the PHP Poll Class, the duration acts as a date interval. Due to database storing issues inside the class, I was forced to convert it to an integer that holds the seconds. The getters and setters are coded in a way such that the duration member variable appears as an int within the class but a date interval outside.

migration/migrator/data/course_tables.sql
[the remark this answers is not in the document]
Kush-Raval: > Would be nice if the variable name for this can also capture the connotation of "time" instead of "date".

I will change the end_date to endTime to make it more intuitive.

site/app/entities/poll/Poll.php
Kush-Raval: The main reason for making it a DateInterval is the ease of access where you can add a DateInterval directly to a DateTime Object.

JWei17 (approved):
I will approve this but make a check box for unlimited time duration.

ryvaru (approved):
Tested the functionality of the code and everything works as far as I can tell. Edge cases, such as negative numbers, 000, or any amount of time higher than 24 hours are handled correctly. Frontend still needs to be implemented, but backend works well. Approved.

jeffrey-asm (changes requested):
I was not able to test the implementation over potential merge conflicts. I tried to set up the poll timer, but after saving the specific time frame, no indication of the timer was shown for the instructor or student. I would address these conflicts below for future reviews:


Screenshot 2024-03-21 at 8 45 18 AM

hansongu123 (approved):
Changes look good to me.

site/app/templates/polls/PollForm.twig
```diff
@@ -101,7 +100,28 @@
...
+            </div>
+            <div class="poll-input">
+                <label class="option"> Min</label><br>
+                <input name="poll-minutes" id="poll-minutes" class="poll-timer" value="{{ poll is not null ? poll.getDuration().i }}"/>
+            </div>
+            <div class="poll-input">
+                <label class="option"> Sec</label><br>
+                <input name="poll-seconds" id="poll-seconds" class="poll-timer" value="{{ poll is not null ? poll.getDuration().s }}"/>
+            </div>
+        </div>
```
jeffrey-asm, line 124: It would be nice to add a break tag, <br />, here for consistency I think.

site/public/css/polls.css
```diff
@@ -116,6 +116,25 @@ form {
...
 }
 
+#timer-inputs {
+    display: inline-block;
+}
+
+.poll-timer {
+    display: inline-block;
+    text-align: center;
+    margin-right: 0.5em;
```
jeffrey-asm, line 126: I would remove margin-right to align the text.

jeffrey-asm (approved):
The proposed changes look good. The poll timer worked well with a variety of intervals. I would also look into the comment about margins to align the text in poll form.

site/public/css/polls.css
Kush-Raval: Fixed, looks much better

site/app/controllers/PollController.php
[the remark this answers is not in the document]
jeffrey-asm: In Poll.php, you can return null if the duration is less than or equal to 0, which would be a cleaner check.

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
jeffrey-asm: Try updating the migration to a single transaction for up and down methods.

site/app/controllers/PollController.php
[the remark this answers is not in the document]
Kush-Raval: The problem with doing this is that I would be allowing duration to be null, but I am throughout my code guaranteeing that duration won't be null. I am representing an "empty duration" as one that is just 0.

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
Kush-Raval: I don't understand what you mean.
Do you mean making it into one execute statement for up and down, I talked to William previously and I think we decided it would make the one line very messy. The multiple lines are for readability.

site/app/templates/polls/PollForm.twig
```diff
@@ -202,6 +221,36 @@
         }
         submitErrorChecks();
     };
+    window.onload = function() {
+        toggleTimerInputs();
+    };
+    function toggleTimerInputs() {
+        var enableTimerCheckbox = document.getElementById("enable-timer");
+        var timerInputs = document.getElementById("timer-inputs").querySelectorAll(".poll-timer");
```
williamjallen, line 229: Using `const` or `let` (as appropriate) is better than `var` in JS.

```suggestion
        const enableTimerCheckbox = document.getElementById("enable-timer");
        const timerInputs = document.getElementById("timer-inputs").querySelectorAll(".poll-timer");
```

site/app/templates/polls/PollForm.twig
```diff
@@ -202,6 +221,36 @@
...
+                    input.setAttribute("placeholder", "Hour");
+                    input.setAttribute("max", "24");
+                } else if (input.classList.contains("poll-minutes")) {
+                    input.setAttribute("placeholder", "Min");
+                } else if (input.classList.contains("poll-seconds")) {
+                    input.setAttribute("placeholder", "Sec");
+                }
+            });
+        } else {
+            document.getElementById("timer-inputs").style.display = "none";
```
williamjallen, line 246: I believe you can use `.hide()` instead of manipulating the style directly.

site/app/templates/polls/PollForm.twig
```diff
@@ -202,6 +221,36 @@
...
     };
+    window.onload = function() {
+        toggleTimerInputs();
+    };
+    function toggleTimerInputs() {
+        var enableTimerCheckbox = document.getElementById("enable-timer");
+        var timerInputs = document.getElementById("timer-inputs").querySelectorAll(".poll-timer");
+
+        if (enableTimerCheckbox.checked) {
+            document.getElementById("timer-inputs").style.display = "block";
```
williamjallen, line 232: You can use `.show()` instead of manipulating the style directly.

site/app/templates/polls/PollForm.twig
```diff
@@ -202,6 +221,36 @@
         }
         submitErrorChecks();
     };
+    window.onload = function() {
+        toggleTimerInputs();
+    };
+    function toggleTimerInputs() {
+        var enableTimerCheckbox = document.getElementById("enable-timer");
```
williamjallen, line 228: It would be better to use jQuery instead of manually querying the DOM.  In this case, it would look something like this: `$('#enable-timer')` instead of `document.getElementById("enable-timer")`.

Apply the same throughout the following function.

site/public/css/polls.css
```diff
@@ -116,6 +116,25 @@ form {
...
 }
 
+#timer-inputs {
+    display: inline-block;
+}
+
+.poll-timer {
+    display: inline-block;
+    text-align: center;
+    width: 4em !important;
```
williamjallen, line 126: Why is the `!important` necessary here?

williamjallen (changes requested):
Almost there.  After this round of changes, I think it'll be pretty much good to go.

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
```diff
@@ -0,0 +1,43 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("ALTER TABLE polls DROP COLUMN IF EXISTS end_time")
+    database.execute("ALTER TABLE polls DROP COLUMN IF EXISTS duration")
```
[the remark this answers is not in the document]
Kush-Raval: Removed the lines

site/app/entities/poll/Poll.php
```diff
@@ -25,9 +26,17 @@ class Poll {
     #[ORM\Column(type: Types::TEXT)]
     protected $question;
 
+    //Duration should be stored as an interval
+    //Tried doctrine annotation to be DATEINTERVAL and interval type in database but there wasnt a proper conversion.
+    #[ORM\Column(type: Types::INTEGER)]
+    protected int $duration;
+
```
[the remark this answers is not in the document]
Kush-Raval: Added a comment to explain what the integer represents

site/app/templates/polls/PollForm.twig
Kush-Raval: Changed this

site/app/templates/polls/PollForm.twig
Kush-Raval: changed the function to use jquery

site/app/templates/polls/PollForm.twig
Kush-Raval: I changed the function to use jquery library.

site/app/templates/polls/PollForm.twig
Kush-Raval: Fixed, changed function to use jquery library.

site/public/css/polls.css
Kush-Raval: Removing it messes up the UI, 
Not sure why but I think removing it makes the CSS styling get overridden by something else.

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    database.execute("ALTER TABLE polls ADD COLUMN IF NOT EXISTS duration INTEGER DEFAULT 0")
+    database.execute("ALTER TABLE polls ADD COLUMN IF NOT EXISTS end_time timestamptz NOT NULL DEFAULT '1900-02-01'")
+    database.execute("ALTER TABLE polls ALTER COLUMN status DROP NOT NULL")
+    database.execute("""
+        UPDATE polls SET end_time = CASE
+            WHEN status = 'open' THEN '9999-02-01'
+            WHEN status = 'closed' THEN '1900-02-01'
+            ELSE now() END
+    """)
```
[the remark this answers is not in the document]
Kush-Raval: I think as it is now, we have endTime, duration, and status columns. It was my intention to set end_time values based on the values of status. 
After the PR is merged, I will open up an issue to remove the status column and remove the status completely.

site/app/controllers/PollController.php
```diff
@@ -191,7 +193,20 @@ public function addNewPoll(): RedirectResponse {
             return new RedirectResponse($this->core->buildCourseUrl(['polls']));
         }
 
-        $poll = new Poll($_POST['name'], $_POST['question'], $_POST['question_type'], $date, $_POST['release_histogram'], $_POST["release_answer"]);
+        //set to 0 if it is not found in $_POST.
+        $hours = intval($_POST['poll-hours'] ?? 0);
+        $minutes = intval($_POST['poll-minutes'] ?? 0);
+        $seconds = intval($_POST['poll-seconds'] ?? 0);
+        if (($hours * 3600) + (60 * $minutes) +  $seconds > 86400) {
```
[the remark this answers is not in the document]
Kush-Raval: Implemented new logic, that compares DateTimes. 
Comparing DateIntervals isn't supported by PHP, so I used a workaround with adding DateIntervals to DateTimes and comparing those.

site/app/templates/polls/PollForm.twig
```diff
@@ -225,33 +226,34 @@
...
+                $(this).attr("type", "number")
+                    .attr("min", "0");
+                if ($(this).hasClass("poll-hours")) {
+                    $(this).attr("placeholder", "Hour")
+                        .attr("max", "24");
+                } else if ($(this).hasClass("poll-minutes")) {
+                    $(this).attr("placeholder", "Min");
+                } else if ($(this).hasClass("poll-seconds")) {
+                    $(this).attr("placeholder", "Sec");
                 }
```
williamjallen, line 244: Why do you need to do this?  Can't you just add these in the HTML markup directly?

williamjallen (changes requested):

site/app/templates/polls/PollForm.twig
Kush-Raval: Fixed

site/app/templates/polls/PollForm.twig
```diff
@@ -202,7 +222,20 @@
...
+    function toggleTimerInputs() {
+        const enableTimerCheckbox = $("#enable-timer");
+        const timerInputs = $(".poll-timer");
 
+        if (enableTimerCheckbox.prop("checked")) {
+            $("#timer-inputs").show();
+        } else {
+            $("#timer-inputs").hide();
+        }
+    }
```
williamjallen, line 237: It would be good to move this to `polls.js` so the JS code gets linted.

site/app/templates/polls/PollForm.twig
```diff
@@ -202,7 +222,20 @@
         }
         submitErrorChecks();
     };
+    window.onload = function() {
+        toggleTimerInputs();
+    };
```
williamjallen, line 227: Is this needed now?  I would be better to render everything on the server side when the page initially loads rather than relying on client-side rendering to render the initial state of the page.

.setup/data/polls/polls_data.json
```diff
@@ -16,6 +16,7 @@
             2
         ],
         "release_date": "2021-07-19",
+        "end_time":"2021-08-19",
```
williamjallen, line 19: The same spacing issue appears throughout the changes in this file.  Please apply throughout.

```suggestion
        "end_time": "2021-08-19",
```

williamjallen (changes requested):

site/app/templates/polls/PollForm.twig
Kush-Raval: Fixed

.setup/data/polls/polls_data.json
Kush-Raval: Fixed

site/app/templates/polls/PollForm.twig
Kush-Raval: Changed to the server side.

williamjallen (approved):

site/app/templates/polls/PollForm.twig
```diff
@@ -202,7 +222,7 @@
         }
         submitErrorChecks();
     };
-
+
```
williamjallen, line 225: This whitespace change should be reverted.

site/app/templates/polls/PollForm.twig
```diff
@@ -193,7 +214,6 @@
             }
         }
     }
-
```
williamjallen, line 196: Same here.

site/cypress/e2e/Cypress-Feature/polls.spec.js
```diff
@@ -196,6 +196,17 @@ describe('Test cases revolving around polls functionality', () => {
         cy.get('#poll-date').clear({force: true});
         // manually setting the release date to some time in the past
         cy.get('#poll-date').type('1970-01-01', {force: true});
+        // Testing the Pol-Timer
```
williamjallen, line 199:
```suggestion
        // Testing the poll timer
```

site/cypress/e2e/Cypress-Feature/polls.spec.js
```diff
@@ -260,6 +271,24 @@ describe('Test cases revolving around polls functionality', () => {
...
+        //Editing the poll to remove timer
+        cy.contains('Poll Cypress Test').siblings(':nth-child(1)').children().click();
+        //Checking if user input for duration saved
+        cy.get('#enable-timer').should('be.checked');
+        cy.get('#timer-inputs').get('#poll-hours').invoke('val').should('eq', '0');
+        cy.get('#timer-inputs').get('#poll-minutes').invoke('val').should('eq', '0');
+        cy.get('#timer-inputs').get('#poll-seconds').invoke('val').should('eq', '5');
+        cy.get('#poll-seconds').clear();
+        cy.get('#enable-timer').get('#poll-hours').clear().type('3');
+        cy.get('button[type=submit]').click();
```
williamjallen, line 287: Is there a more specific selector to target this button with?  This will cause problems if another form with a submit button is added to the page in the future.

site/cypress/e2e/Cypress-Feature/polls.spec.js
```diff
@@ -364,7 +400,10 @@ describe('Test cases revolving around polls functionality', () => {
         cy.logout();
         cy.login();
         cy.visit(['sample', 'polls']);
-        cy.contains('Poll Cypress Test').siblings(':nth-child(6)').children().click();
+        cy.wait(8000);
```
williamjallen, line 403: Is this 8-second wait as tight a bound as it can be?

site/public/js/polls.js
```diff
@@ -109,3 +109,13 @@ function updateDropdownStates(curr_state, cookie_key) {
 function importPolls() {
     $('#import-polls-form').submit();
 }
+
+function toggleTimerInputs() {
+    const enableTimerCheckbox = $('#enable-timer');
+    if (enableTimerCheckbox.prop('checked')) {
```
williamjallen, line 115: This could be inlined since it's only used once.  Note: this is just a suggestion, so it is not a blocker.

site/tests/app/entities/poll/PollTester.php
```diff
@@ -226,32 +230,56 @@ public function testReleaseDate(): void {
...
-        $this->my_polls[2]->setEnded();
-        $this->assertFalse($this->my_polls[2]->isOpen());
-        $this->assertFalse($this->my_polls[2]->isClosed());
-        $this->assertTrue($this->my_polls[2]->isEnded());
+        $this->my_polls[1]->setEnded();
+        $this->assertFalse($this->my_polls[1]->isOpen());
+        $this->assertFalse($this->my_polls[1]->isClosed());
+        $this->assertTrue($this->my_polls[1]->isEnded());
+    }
+    public function testDuration(): void {
```
williamjallen, line 262:
```suggestion
    }

    public function testDuration(): void {
```

williamjallen (changes requested):
The new changes seem reasonable to me.  While looking through the code again, I caught a handful of small code quality things that would be good to clean up.

site/app/templates/polls/PollForm.twig
Kush-Raval: Fixed

site/cypress/e2e/Cypress-Feature/polls.spec.js
Kush-Raval: I changed it to 6000 seconds, it might cause it to be more flaky I'll run the cypress test multiple times to test

site/cypress/e2e/Cypress-Feature/polls.spec.js
Kush-Raval: This is the same way as the pre-existing logic for editing a poll and submitting changes.

site/public/js/polls.js
Kush-Raval: Makes sense, added suggestion

site/tests/app/entities/poll/PollTester.php
Kush-Raval: Applied change

site/cypress/e2e/Cypress-Feature/polls.spec.js
Kush-Raval: Tested three times times through github and locally tested as well. Seems to work

site/cypress/e2e/Cypress-Feature/polls.spec.js
```diff
@@ -364,7 +400,11 @@ describe('Test cases revolving around polls functionality', () => {
         cy.logout();
         cy.login();
         cy.visit(['sample', 'polls']);
-        cy.contains('Poll Cypress Test').siblings(':nth-child(6)').children().click();
+        //Wait 6 seconds to wait out the time remaining for poll to close
+        cy.wait(6000);
```
williamjallen, line 404: I'm still a little bit concerned about whether this will lead to flaky tests down the road, but perhaps it's good enough for now.

williamjallen (approved):
New changes seem reasonable to me.

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
```diff
@@ -0,0 +1,49 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("ALTER TABLE polls ADD COLUMN IF NOT EXISTS duration INTEGER DEFAULT 0")
+    database.execute("ALTER TABLE polls ADD COLUMN IF NOT EXISTS end_time timestamptz DEFAULT NULL")
```
williamjallen, line 18: I believe the `DEFAULT NULL` is redundant here.

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
```diff
@@ -0,0 +1,49 @@
...
+    database.execute("""
+        UPDATE polls SET 
+            end_time = CASE
+                WHEN status = 'open' THEN NULL
+                ELSE now() 
+            END,
+            is_visible = CASE
+                WHEN status IN ('open', 'ended') THEN TRUE
+                ELSE FALSE
+            END
```
williamjallen, line 30: This can be broken into two separate update statements, and the `end_time` changes can be simplified to something like this:

```
UPDATE polls
SET end_time = now()
WHERE status != 'open'
```

Would it also make sense to set the end time to a date in the past instead of `now()`?

site/app/entities/poll/Poll.php
```diff
@@ -163,10 +169,14 @@ public function setDuration(\DateInterval $duration): void {
         $this->duration = $totalSeconds;
     }
 
-    public function setEndTime(\DateTime $end_time): void {
+    public function setEndTime(?\DateTime $end_time): void {
         $this->end_time = $end_time;
     }
 
+    public function isTimerEnabled(): bool {
+        return !($this->end_time === null);
```
williamjallen, line 177:
```suggestion
        return $this->end_time !== null;
```

site/app/libraries/PollUtils.php
```diff
@@ -23,7 +23,7 @@ public static function getPollExportData(array $polls): array {
                 'responses' => [],
                 'correct_responses' => [],
                 'duration' => $poll->getDuration()->format('P%yY%mM%dDT%hH%iM%sS'),
-                'end_time' => $poll->getEndTime()->format('Y-m-d'),
+                'end_time' => $poll->getEndTime() === null ? $poll->getEndTime() : $poll->getEndTime()->format('Y-m-d'),
```
williamjallen, line 26: Isn't this rather silly?  If `$poll->getEndTime()` returns null, why call it again?

Perhaps something like this would be better:
```suggestion
                'end_time' => $poll->getEndTime()?->format('Y-m-d'),
```

williamjallen (changes requested):

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
```diff
@@ -0,0 +1,49 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    pass
```
williamjallen, line 47: Is there a safe down migration which can be applied here to ensure a reasonable status is set in case this gets rolled back for whatever reason?

site/app/entities/poll/Poll.php
```diff
@@ -91,39 +105,78 @@ public function getQuestion(): string {
...
+        if ($this->end_time === null && $this->is_visible) {
+            return false;
+        }
+        $now = DateUtils::getDateTimeNow();
+        return $now > $this->end_time && $this->is_visible;
     }
 
     public function isClosed(): bool {
-        return $this->status === "closed";
+        return (!$this->is_visible);
```
williamjallen, line 137:
```suggestion
        return !$this->is_visible;
```

site/app/templates/polls/PollForm.twig
```diff
@@ -101,7 +100,29 @@
...
+            <label class="option-title">
+                Expected Duration of Poll:
+            </label>
+            <div class="poll-input">
+                <label class="option"> Hours </label><br>
+                <input type="number" name="poll-hours" id="poll-hours" class="poll-timer" placeholder="hrs" max="24" min="0" value="{{ poll is not null ? poll.getDuration().h }}" />
+            </div>
+            <div class="poll-input">
+                <label class="option"> Minutes </label><br>
+                <input type="number" name="poll-minutes" id="poll-minutes" class="poll-timer" placeholder="min" min="0" value="{{ poll is not null ? poll.getDuration().i }}"/>
```
williamjallen, line 118: Should there be a maximum value of 60 for minutes and seconds, or is this a deliberate decision to allow instructors to set "90 seconds", for example?

williamjallen (changes requested):
Another few small comments:

site/app/templates/polls/PollForm.twig
Kush-Raval: This was a design choice. They should be able to choose 90 seconds. The 24 hour maximum for hours is because we are limiting duration to 24 hours.

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
Kush-Raval: Oh okay makes sense. I changed

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
Kush-Raval: if it is closed, the is_visible would be false so the end_time wouldn't matter. If the poll was set to open, the end_time should be null and is_Visible should be true therefore signaling that it is open. If the poll was ended, is_visible would be true but there would be a endtime that is less than (now) so we could change it to beginning of time if that is more suitable

Kush-Raval: Actually, it does make sense for it to be the beginning of time. I will change it to 1900-02-01

site/app/libraries/PollUtils.php
Kush-Raval: ahh, didn't know you could do that applied this change

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
Kush-Raval: status columns wouldn't ever change. So rolling it back should be fine.

Kush-Raval: idk if there is a need to have a down migration

site/app/entities/poll/Poll.php
Kush-Raval: fixed

migration/migrator/migrations/course/20240205154116_add_poll_timer.py
Kush-Raval: Never mind, I made the down migration reset the status constraint back to the original.

williamjallen: I believe the down migration leaves the system in a broken state if new polls are created with these changes, and then the changes are rolled back.  The new polls will not have a valid status.  Now that you've added the constraint back, I believe the down migration itself will fail due to the null values in new polls.

This migration needs to be 100% reversible and repeatable in both directions.

Kush-Raval: I think I understand

williamjallen (approved):

PR #10100, opened 2024-01-12
[Feature:InstructorUI] Ability to Upload JSON to make gradeable

IDzyre:
### Please check if the PR fulfills these requirements:

* [x] Tests for the changes have been added/updated (if possible)
* [x] Documentation has been updated/added if relevant
* [x] Screenshots are attached to Github PR if visual/UI changes were made

### What is the current behavior?
The Submitty UI only has the ability to manually create gradeables VIA the UI

### What is the new behavior?
image
image

There is now a button to upload a JSON file containing all of the relevant data. 

# Template
```
{
  "title": "Testing Json",
  "instructions_url": "",
  "id": "hw-1-1-1222",
  "type": "Electronic File"/"Checkpoints"/"Numeric",
  "vcs": {
    "repository_type": "submitty-hosted",
    "vcs_path": "path/to/vcs",
    "vcs_subdirectory": "subdirectory"
  },
  "bulk_upload": true/false,
  "team_gradeable": {
    "team_size_max": 3,
    "inherit_from": "gradeable_id",
    "gradeable_teams_read": false,
  },
  "grading_inquiry": {
    "grade_inquiry_per_component_allowed": true/false,
  },
  "ta_grading": true/false,
  "discussion_thread_id": "thread_id",
  "syllabus_bucket": "Homework"
}
```

Dates and other features will be in future PRs
### Other information?
* [x] Add tests
* [x] Review Security
* [x] Add API route
* [x] Write Documentation

PR #10188, opened 2024-02-15
[Feature:InstructorUI] Download JSON of Gradeable

IDzyre:

image

### What is the new behavior?
As an addition to #10100, this allows for the download of a JSON from an existing gradeable, which will allow for recreation in a new course/semester. 

Variables downloaded:
Title
Gradeable Type
Instructions URL
Gradeable ID
VCS:
  VCS Path
  VCS Repository Type
  VCS Subdirectory
Team assignments
  Team max size
  Inherit teams from ID
TA grading
Bulk Upload
Grade Inquiries
Autograding Config Path
Dates:
  TA view date
  Grade start date
  Grade due date
  Grade released date
  Team lock date
  Submission open date
  Submission due date
  Grade inquiry start date
  Grade inquiry due date
  Has due date
  Has release date
  Late submissions allowed
  Late days (number)


site/public/js/gradeable.js
```diff
@@ -373,6 +373,41 @@ function ajaxUploadGradeable() {
     return true;
 }
 
+// Uploads a gradeable via JSON POST request
+function ajaxDownloadGradeable($gradeable_id) {
```
shailpatels, line 377: Could do something like this

```js
//downloads a gradeable's json config
function downloadGradeableJsonConfig
    fetch(buildCourseUrl([gradeable_id, 'download']))
        .then( response => response.json() )
        .then( json => {
            if (json['status'] === 'success') {
                const temporaryElement = document.createElement('a');
                const file = new Blob([JSON.stringify(json['data'], null, 4)]);
                temporaryElement.href = URL.createObjectURL(file);
                temporaryElement.download = `${json['data']['id']}.json`;
                temporaryElement.click();
            }
            else{
                displayErrorMessage(json['message']);
            }
        })
        .catch(err => {
            displayErrorMessage(err);
        });
}
```

Fetch is a newer API thats a little bit less terse then ajax calls, there's also the ability to do `await` and `async` to mark when a function is used async vs not so thats a bit of older nomenclature the codebase used before it was standardized where we put `ajax` in front of the function to indicate it would do something async or make a request. Now you probably don't need and its more important to give a good descriptive name, especially one in a js file because it gets included in the global scope and will reduce the chance of a name collision 

The `displayErrorMessage` is a function defined globally in `server.js` and would show the nice standard red error banners we have around the site.

createObjectURL will also handle the encoding for the tmp link so you don't have to deal with uri encoding yourself and setting the preamble for the url

based on: https://stackoverflow.com/questions/34156282/how-do-i-save-json-to-local-text-file

shailpatels (changes requested):

shailpatels:
image

screenshot of the new button on the right of the edit gradeable page on sample course

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -51,11 +51,82 @@ public function editGradeableRequest($gradeable_id, $nav_tab = 0) {
...
+                $team_properties = [
+                    'team_max_size' => $config->getTeamSizeMax(),
+                    'inherit_from' => ''
+                ];
+                $return_json['team_gradeable'] = $team_properties;
+            }
+            if ($config->isTaGrading()) {
+                $return_json['ta_grading'] = true;
+                if ($config->isGradeInquiryAllowed()) {
+                    $return_json['grade_inquiries'] = true;
```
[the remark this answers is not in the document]
IDzyre: Yes, I changed the code for the upload slightly, and forgot to change the docs. I will create a PR for the docs change soon.

IDzyre:
> Approving after date changes, download functionality works the same from initial review.
> 
> I tried uploading the resulting download for a new gradeable and noticed the dates aren't read in from the downloaded json, is that an unimplemented feature or a bug?
> 
> Also not sure if this is being considered but a pretty cool e2e test would be to download the json, delete the original gradable and see if it can be recreated from the upload

Currently dates are not in main for upload, they are added in #10252 . Once both of these PRs are merged, i will create more e2e tests that involve uploading a previously downloaded file, and ensuring dates are accurate for download.

site/public/js/gradeable.js
```diff
@@ -373,6 +373,30 @@ function ajaxUploadGradeable() {
...
+            temporaryElement.download = `${json['data']['id']}.json`;
+            temporaryElement.click();
+        }
+        else {
+            displayErrorMessage(json['message']);
+            return false;
+        }
+    });
+    return true;
+}
```
williamjallen, line 398: Why is this necessary?  Can't you just send it with the appropriate headers set on the server side such that it gets downloaded automatically?

williamjallen (changes requested):

site/public/js/gradeable.js
IDzyre: I am not quite sure what you mean by that, could you explain a bit more?

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -51,11 +51,102 @@ public function editGradeableRequest($gradeable_id, $nav_tab = 0) {
...
+            $dates['has_due_date'] = $config->hasDueDate();
+            $dates['has_release_date'] = $config->hasReleaseDate();
+            $dates['late_submission_allowed'] = $config->isLateSubmissionAllowed();
+            $dates['late_days'] = $config->getLateDays();
+            $return_json['dates'] = $dates;
+        }
+
+        $this->core->getOutput()->useHeader(false);
+        $this->core->getOutput()->useFooter(false);
+        echo json_encode($return_json, JSON_PRETTY_PRINT);
```
williamjallen, line 141: The `render()` function in `JsonResponse.php` essentially does the same thing, and would probably be a better option here.  The link I sent you to the example in `PlagiarismController.php` was probably not the best example in hindsight.

williamjallen (changes requested):

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -50,14 +50,104 @@ public function editGradeableRequest($gradeable_id, $nav_tab = 0) {
         'electronic_bulk' => 'TA/Instructor will (bulk) upload scanned .pdf for online manual grading'
     ];
 
+    /**
+     * @Route("/api/{_semester}/{_course}/{gradeable_id}/download", methods={"GET"})
+     * @Route("/courses/{_semester}/{_course}/{gradeable_id}/download", methods={"GET"})
+     */
+    public function downloadJson(string $gradeable_id) {
```
williamjallen, line 57: This should fix the PHPStan error:
```suggestion
    public function downloadJson(string $gradeable_id): void {
```

williamjallen (dismissed):
Code looks good.  One minor change and then I think it's good to go.

site/app/controllers/admin/AdminGradeableController.php
williamjallen, line 57: One side effect of this decision is that there is now an API route which does not return data in the standard success/failure format.  I feel like that kind of inconsistency could be problematic in the future.  Do you think it would be a good idea to make two routes: one regular download route which keeps the logic as-is, and an API route which packages the data up in a success response?

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -50,14 +50,112 @@ public function editGradeableRequest($gradeable_id, $nav_tab = 0) {
...
+        $json_response = JsonResponse::getSuccessResponse($returned_json);
+        // Make the JSON only the data, not the data and the success status.
+        $json_response->json = $json_response->json['data'];
+        $json_response->render($this->core);
+    }
+
+    /**
+     * @return array<mixed>
+     */
+    public function getGradeableJson(Gradeable $gradeable): array {
```
williamjallen, line 72: This would be a great place to specify the [array shape](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes), so that it is clear exactly what fields are being returned, and what their expected types are.

cjreed121 (approved):

PR #10193, opened 2024-02-19
[Feature:TAGrading] Grading panel visibility to peer graders

### LINKED ISSUE #10114: specify which ta grading panels are available to peer graders
When a peer grader logs in to grade an assignment, they get the TA grading interface.
This interface is designed for TAs who do lots of grading, but it can be confusing for first time graders.

Screenshot 2024-01-18 at 1 28 45 PM

Why are the panels in this order:  autograding, rubric, files, solution notes, discussion?

The most common use case for peer grading is to look at the files tab and fill out the rubric tab.
The may never need the autograding, solution notes, and discussion tab (discussion tab only appears if the assignment has a form thread associated).

If they are just giving feedback on an in class presentation they may only need the rubric tab.

Let's allow the instructor to specify which tabs should be displayed to peer graders.

On the edit gradeable / grader assignment tab let's add a question a set of checkboxes underneath this question:

Screenshot 2024-01-18 at 1 33 43 PM

e.g., let's add:
"Which tabs should be visible to peer graders?" 
- [ ] autograding
- [ ] rubric
- [ ] files
- [ ] solution/notes
(and if relevant to this gradeable - [ ] discussion)
All should be checked by default but can be unselected

GO-viper7:
Closes #10114

* [✅] Screenshots are attached to Github PR if visual/UI changes were made

When a peer grader logs in and opens TA grading interface , all the available panels may confuse the grader

### What is the new behavior?
Instructor can configure the availability of grading panels to peer graders 
This can be done using checkboxes in Grader Assignment tab of Edit Gradeable Configuration

[peerGradingPanel.webm](https://github.com/Submitty/Submitty/assets/81346327/3b5e3e49-e063-4bde-b1e0-ba6345f10fb8)

### Visual Changes in `Edit Gradeable`
image

### Instructor disabling access for Autograding and Solution/Notes, and changes getting reflected on grading interface for peer graders
image
image

### Other information?
Added migration to create table "peer_grading_panel" with a foreign key relation to "electronic_gradeable" table
Each column inside this new table corresponds to a panel which follows boolean data type 
True : displays respective panel to peer graders
False: doesn't display respective panel to peer graders


site/app/views/grading/ElectronicGraderView.php
```diff
@@ -958,6 +958,9 @@ public function hwGradingPage(Gradeable $gradeable, GradedGradeable $graded_grad
         $isDiscussionPanel = false;
         $isGradeInquiryPanel = false;
         $is_peer_grader = false;
+        // Converting array from '{"a", "b"}' to [a, b]
+        $isPeerAccessible = explode(",", trim($gradeable->getPeerPanel(), '{}'));
```
jeffreyC4, line 962: Why is $gradeable->getPeerPanel() the first parameter to trim() given it returns an array of integers?

jeffreyC4 (commented):
I'm not too comfortable with the peer grading aspect of Submitty, but the intended changes worked for the specific account shown in the video above. I could not get it to work for the limited-access grader account (grader), but I'm not sure if this is a peer grader. Most of the implementation looks solid, but review the comment I left because it could lead to unintended behavior.

site/app/views/grading/ElectronicGraderView.php
GO-viper7: Thanks for the review,
Here's why `$gradeable->getPeerPanel()` is being trimmed : 
PostgreSQL supports array data types and lets us to store arrays in this format : "{"A", "B"}"
This is why im trimming this down to [A, B] 
Let me know if you want me to elaborate the comment in line 961

Also, I don't think its intended for limited-access grader accounts as per the issue # 10114

GO-viper7:
Hey, @jeffreyC4 
Kindly help me with the following  : 
As you can see, there's an unsuccessful check for Cypress (Gradeable) tests

Here's the reason for it 
image

But the thing is, my code changes are not concerned with the above 
So, I'm not knowing the exact reason for these test fails
Could you please go through and let me know where the issue is
Thank you.

jeffreyC4:
> Hey, @jeffreyC4 Kindly help me with the following : As you can see, there's an unsuccessful check for Cypress (Gradeable) tests
> 
> Here's the reason for it image
> 
> But the thing is, my code changes are not concerned with the above So, I'm not knowing the exact reason for these test fails Could you please go through and let me know where the issue is Thank you.

I tried running the same failing tests locally by changing to the site directory and running `npx cypress open`, but they are still failing. The main issue is that the new migration is leading to problems when creating a new gradable: 
image

I would try to focus on fixing the creation issue as the input type is not being allowed, which keeps leading to the error below during the Cypress tests. I'm not too comfortable with working with arrays stored in a column in SQL. It's generally good practice not to store an array of data as a column type as it can be harder to maintain and can lead to errors in formatting. 

image

GO-viper7:
> > Hey, @jeffreyC4 Kindly help me with the following : As you can see, there's an unsuccessful check for Cypress (Gradeable) tests
> > Here's the reason for it image
> > But the thing is, my code changes are not concerned with the above So, I'm not knowing the exact reason for these test fails Could you please go through and let me know where the issue is Thank you.
> 
> I tried running the same failing tests locally by changing to the site directory and running `npx cypress open`, but they are still failing. The main issue is that the new migration is leading to problems when creating a new gradable: image
> 
> I would try to focus on fixing the creation issue as the input type is not being allowed, which keeps leading to the error below during the Cypress tests. I'm not too comfortable with working with arrays stored in a column in SQL. It's generally good practice not to store an array of data as a column type as it can be harder to maintain and can lead to errors in formatting.
> 
> image

I see, you are right
But it will be redundant to maintain 5 variables for each of the below and circulate it throughout those files: 
- Autograding panel
- Rubric panel
- Files panel
- Solutions/Notes panel
- Discussion panel

Let me research a bit on cypress tests and storing arrays in Postgres
will inform you, if i found a solution to this issue 
or may be I can ask suggestions from mentors about this
Thank you

jeffreyC4:
> > > Hey, @jeffreyC4 Kindly help me with the following : As you can see, there's an unsuccessful check for Cypress (Gradeable) tests
> > > Here's the reason for it image
> > > But the thing is, my code changes are not concerned with the above So, I'm not knowing the exact reason for these test fails Could you please go through and let me know where the issue is Thank you.
> > 
> > 
> > I tried running the same failing tests locally by changing to the site directory and running `npx cypress open`, but they are still failing. The main issue is that the new migration is leading to problems when creating a new gradable: image
> > I would try to focus on fixing the creation issue as the input type is not being allowed, which keeps leading to the error below during the Cypress tests. I'm not too comfortable with working with arrays stored in a column in SQL. It's generally good practice not to store an array of data as a column type as it can be harder to maintain and can lead to errors in formatting.
> > image
> 
> I see, you are right But it will be redundant to maintain 5 variables for each of the below and circulate it throughout those files:
> 
> * Autograding panel
> * Rubric panel
> * Files panel
> * Solutions/Notes panel
> * Discussion panel
> 
> Let me research a bit on cypress tests and storing arrays in Postgres will inform you, if i found a solution to this issue or may be I can ask suggestions from mentors about this Thank you

Maybe try storing the state as a binary string. Like `00000` could imply no visibility and `11111` can imply all. You can trim this string and attach the index to a bucket in an array that represents which panel is which.

GO-viper7:
Hello @jeffreyC4 
So, the new migration cleared all the unwanted issues from the testing

and I just fixed those js and php lint errors 
You may run the tests again
Thanks.

migration/migrator/migrations/course/20240219030249_peer_panel_graders.py
```diff
@@ -0,0 +1,34 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("ALTER TABLE electronic_gradeable DROP COLUMN IF EXISTS eg_peer_panel;")
```
jeffreyC4, line 17: You can remove the first database execute as column doesn't exist on main.

GO-viper7: So, only the following line will stay ?
`database.execute("ALTER TABLE electronic_gradeable ADD COLUMN eg_peer_panel VARCHAR(5) DEFAULT '11111';")`

But, It also needs to include CREATE TABLE electronic_gradeable  query right?

GO-viper7:
> Most of the implemented changes look good. I left a comment to update the migration file. The one issue I found was that the SolutionTaNotesPanel.twig file is always being rendered, so if none of the checkboxes are toggled, the peer graders will always see the TA Solution/Notes without the button at the top. I would fix this.

Fixed both the issues, thanks for pointing it out!
You may review.

jeffreyC4 (approved):
This implementation looks solid and makes sense. I could not find any bugs when addressing all possible panel options for peer graders. All higher grading privilege accounts had no effect due to the proposed changes.

GO-viper7:
@jeffreyC4 , In the main(first) comment of this PR
I edited the representation of an entry of column from array to binary string, so that no one gets confused
You may run test for title-check, thanks.

wuj18 (approved):
I tested it and it works. The panel change is made to the peer grader account and does not affect the limited/full access grader. The code change also makes sense.

williamjallen:
The video in the PR description seems to be broken.  Screenshots would probably be good enough in this case.  Can you please add some before/after screenshots instead of the video?

migration/migrator/migrations/course/20240219030249_peer_panel_graders.py
```diff
@@ -0,0 +1,33 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("ALTER TABLE electronic_gradeable ADD COLUMN eg_peer_panel VARCHAR(5) DEFAULT '11111';")
```
williamjallen, line 17: Essentially storing data as a bit string seems rather prone to error, and will be challenging to maintain in the future if additional panels get added.  Creating a new table with a foreign-key relationship to the `electronic_gradeable` column is probably the best way to go about doing this.

williamjallen (changes requested):

GO-viper7:
> The video in the PR description seems to be broken. Screenshots would probably be good enough in this case. Can you please add some before/after screenshots instead of the video?

Sure, thanks for pointing it out
Added screenshots of newly added feature

migration/migrator/migrations/course/20240219030249_peer_panel_graders.py
GO-viper7: yes, you are right 
it will be tedious to append a default character to the string for each of the gradeable out there, if a new panel is added.
I'll soon implement the foreign-key relation which you have mentioned
And I plan to use Boolean datatype in new table for each of the column instead of binary
is it ok ?

williamjallen: Using a boolean would be great.

bmcutler (changes requested):
I tested it on an existing peer gradeable in sample course, and while I could uncheck some panels for visibility to peers, it did not change the peer view (peer could still see everything).

I made a new gradeable and it worked correctly.

Other reviewers and/or @GO-viper7 please test the existing gradeables again.

Also.  I don't know if it's related, but I had access problems when trying to enter the peer grading page for specific students.  (Some of the user was assigned to grade worked fine, some failed to load.)  When I got the error, this was the error message:

Screenshot 2024-06-06 at 1 06 37 PM

dorian451 (changes requested):
I also experience the same problem that @bmcutler  does, where checking or unchecking panels in an existing peer grading course does not change a peer grader's view.  I also get the access denied error occasionally when trying to peer grade, but I also get that on `main` so maybe that's a different problem

migration/migrator/migrations/course/20240508080451_add_peer_grading_panels_table.py
```diff
@@ -0,0 +1,44 @@
...
+            files BOOLEAN NOT NULL DEFAULT TRUE,
+            solution_notes BOOLEAN NOT NULL DEFAULT TRUE,
+            discussion BOOLEAN NOT NULL DEFAULT TRUE,
+            FOREIGN KEY (g_id) REFERENCES electronic_gradeable(g_id)
+        );
+    """)
+    pass
+
+
+def down(config, database, semester, course):
```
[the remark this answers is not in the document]
dorian451: Ok after reading our documentation I turns out this is the right way to do it so please disregard this comment.

williamschen23 (changes requested):
I did not experience the same issues that Dorian and Barb did, it worked well on my end. Some comments.

It does not make sense for someone grading to not see the rubric and/or the submission. I would keep at least one of these boxes greyed out, meaning that it would automatically be selected because it is important to peer grade.

migration/migrator/migrations/course/20240508080451_add_peer_grading_panels_table.py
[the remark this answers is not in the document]
GO-viper7: Alright

GO-viper7:
> I did not experience the same issues that Dorian and Barb did, it worked well on my end. Some comments.
> 
> It does not make sense for someone grading to not see the rubric and/or the submission. I would keep at least one of these boxes greyed out, meaning that it would automatically be selected because it is important to peer grade.

yes, you are right and we need a little discussion on which panels to be displayed here in default 
@bmcutler can you please specify this for us. 
Thanks.

williamschen23:
> > I did not experience the same issues that Dorian and Barb did, it worked well on my end. Some comments.
> > It does not make sense for someone grading to not see the rubric and/or the submission. I would keep at least one of these boxes greyed out, meaning that it would automatically be selected because it is important to peer grade.
> 
> yes, you are right and we need a little discussion on which panels to be displayed here in default @bmcutler can you please specify this for us. Thanks.

From the meeting, we all decided that the best course of action was to not change the current implementation because of different usages for peer grading.

williamschen23 (approved):

williamschen23:
To the people having trouble, I also noticed that the sample has two of the same courses with the same title. Make sure that you are selecting the right one.
image

ziesski (changes requested):

site/app/templates/grading/electronic/GradingPanelHeader.twig
```diff
@@ -1,16 +1,20 @@
 {% import _self as self %}
 <script>
+    let autograding_results = {% if not isPeerGrader or (isPeerGrader and isPeerAutograding) %} true {% else %} false {% endif %};
+    let grading_rubric = {% if not isPeerGrader or (isPeerGrader and isPeerRubric) %} true {% else %} false {% endif %};
+    let submission_browser = {% if not isPeerGrader or (isPeerGrader and isPeerFiles) %} true {% else %} false {% endif %};
+    let solution_ta_notes = {% if not isPeerGrader or (isPeerGrader and isPeerSolutions) %} true {% else %} false {% endif %};
     let student_info = {% if isStudentInfoPanel %} true {% else %} false {% endif %};
     let grade_inquiry_info = {% if isGradeInquiryPanel %} true {% else %} false {% endif %};
-    let discussion_browser = {% if isDiscussionPanel %} true {% else %} false {% endif %};
+    let discussion_browser = {% if isDiscussionPanel and (not isPeerGrader or (isPeerGrader and isPeerDiscussion)) %} true {% else %} false {% endif %};
```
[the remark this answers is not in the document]
GO-viper7: Sure, updated them.

site/app/templates/grading/electronic/GradingPanelHeader.twig
```diff
@@ -1,16 +1,20 @@
...
 
     updateThePanelsElements({
-        autograding_results: true,
-        grading_rubric: true,
-        submission_browser: true,
-        solution_ta_notes: true,
+        autograding_results: autograding_results,
+        grading_rubric: grading_rubric,
+        submission_browser: submission_browser,
+        solution_ta_notes: solution_ta_notes,
```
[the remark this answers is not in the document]
GO-viper7: you are right, I could directly pass them as we are using same vars throughout the files.

GO-viper7:
Thanks for the review @ziesski 
Made changes as required.

migration/migrator/data/course_tables.sql
```diff
@@ -2404,6 +2418,14 @@ ALTER TABLE ONLY public.peer_feedback
     ADD CONSTRAINT peer_feedback_pkey PRIMARY KEY (g_id, grader_id, user_id);
 
 
+--
+-- Name: peer_grading_panel peer_grading_panel_pkey; Type: CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.peer_grading_panel
```
ziesski, line 2425: Did you manually edit `course_tables.sql`file to add constraint? I am reviewing the code, and it looks like your migration `20240508080451_add_peer_grading_panels_table.py` does not add any constraints but .sql file has them. You should not edit the .sql file directly, but rather use migrator dump to edit it. Here is the [documentation.](https://submitty.org/developer/development_instructions/migrations)

ziesski (changes requested):

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7303,7 +7358,6 @@ private function createGradedComponent(GradedComponent $graded_component) {
             VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)";
         $this->course_db->query($query, $param);
     }
-
```
bmcutler, line 7306: revert whitespace change

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -6240,7 +6240,6 @@ public function getGradeableConfigs($ids, $sort_keys = ['id']) {
 
         // Generate the ORDER BY clause
         $order = self::generateOrderByClause($sort_keys, []);
-
```
bmcutler, line 6243: revert whitespace change

site/public/js/admin-gradeable-updates.js
```diff
@@ -234,12 +238,13 @@ $(document).ready(() => {
         else if (data['depends_on_points'] !== null) {
             data['depends_on'] = gradeable;
         }
-
```
bmcutler, line 237: revert whitespace edit

bmcutler (changes requested):
I tested it again on an existing gradeable in the sample course and it worked great.

You'll need to fix the problem @ziesski pointed out above.  Your .sql edits should match the changes in the migration (you're missing the constraints in the migration) -- this flaw should cause the CI db test...  but maybe the test is broken.

migration/migrator/data/course_tables.sql
ziesski: @GO-viper7 Please let us know :)

migration/migrator/migrations/course/20240508080451_add_peer_grading_panels_table.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    database.execute("""
+        CREATE TABLE IF NOT EXISTS peer_grading_panel (
+            g_id VARCHAR(255) PRIMARY KEY,
+            autograding BOOLEAN NOT NULL DEFAULT TRUE,
+            rubric BOOLEAN NOT NULL DEFAULT TRUE,
+            files BOOLEAN NOT NULL DEFAULT TRUE,
+            solution_notes BOOLEAN NOT NULL DEFAULT TRUE,
+            discussion BOOLEAN NOT NULL DEFAULT TRUE,
+            FOREIGN KEY (g_id) REFERENCES electronic_gradeable(g_id)
```
ziesski, line 25: After careful review, I noticed that we actually have `FOREIGN KEY` here, which is stipulating that the g_id column in the peer_grading_panel table is referencing the g_id column in the electronic_gradeable table. Hence indirectly adds constraint for primary key and foreign key.
I was initially so focused on the migration matching .sql file and had tunnel vision. And of course DB-check was not failing because it was correct migration.

GO-viper7: Sorry for the delay.
So, @ziesski 
Are you saying that no changes from my side are required on .sql file ?
Here's the history actually : 
There was a merge conflict in .sql file due to some course_materials changes I believe, I eventually found out and resolved it manually (Hope this is not an issue here)

Let me know if you still need anything. Thanks.

site/app/models/gradeable/Gradeable.php
```diff
@@ -353,6 +378,26 @@ public function __construct(Core $core, array $details) {
...
+            $this->setPeerFiles($details['files'] !== null ? $details['files'] : true);
+        }
+
+        if (array_key_exists('solution_notes', $details)) {
+            $this->setPeerSolutions($details['solution_notes'] !== null ? $details['solution_notes'] : true);
+        }
+
+        if (array_key_exists('discussion', $details)) {
+            $this->setPeerDiscussion($details['discussion'] !== null ? $details['discussion'] : true);
+        }
```
ziesski, line 399:
```suggestion
$mapping = [
    'autograding' => 'setPeerAutograding',
    'rubric' => 'setPeerRubric',
    'files' => 'setPeerFiles',
    'solution_notes' => 'setPeerSolutions',
    'discussion' => 'setPeerDiscussion'
];

foreach ($mapping as $key => $method) {
    if (array_key_exists($key, $details)) {
        $this->$method($details[$key] ?? true);
    }
}

```

ziesski: Can we do something like this to make it less repetitive? I think we can even apply this to surrounding code.  Maybe in new PR for refactor.

site/app/templates/grading/electronic/GradingPanelHeader.twig
```diff
@@ -1,16 +1,20 @@
...
-    let is_notebook = {% if is_notebook %} true {% else %} false {% endif %};
+    const autograding_results = {% if not isPeerGrader or (isPeerGrader and isPeerAutograding) %} true {% else %} false {% endif %};
+    const grading_rubric = {% if not isPeerGrader or (isPeerGrader and isPeerRubric) %} true {% else %} false {% endif %};
+    const submission_browser = {% if not isPeerGrader or (isPeerGrader and isPeerFiles) %} true {% else %} false {% endif %};
+    const solution_ta_notes = {% if not isPeerGrader or (isPeerGrader and isPeerSolutions) %} true {% else %} false {% endif %};
+    const student_info = {% if isStudentInfoPanel %} true {% else %} false {% endif %};
+    const grade_inquiry_info = {% if isGradeInquiryPanel %} true {% else %} false {% endif %};
+    const discussion_browser = {% if isDiscussionPanel and (not isPeerGrader or (isPeerGrader and isPeerDiscussion)) %} true {% else %} false {% endif %};
+    const peer_info = {% if isPeerPanel %} true {% else %} false {% endif %};
+    const is_notebook = {% if is_notebook %} true {% else %} false {% endif %};
```
ziesski, line 11:
```suggestion
    const autograding_results = {{ not isPeerGrader or (isPeerGrader and isPeerAutograding) ? true : false }};
    const grading_rubric = {{ not isPeerGrader or (isPeerGrader and isPeerRubric) ? true : false }};
    const submission_browser = {{ not isPeerGrader or (isPeerGrader and isPeerFiles) ? true : false }};
    const solution_ta_notes = {{ not isPeerGrader or (isPeerGrader and isPeerSolutions) ? true : false }};
    const student_info = {{ isStudentInfoPanel ? true : false }};
    const grade_inquiry_info = {{ isGradeInquiryPanel ? true : false }};
    const discussion_browser = {{ isDiscussionPanel and (not isPeerGrader or (isPeerGrader and isPeerDiscussion)) ? true : false }};
    const peer_info = {{ isPeerPanel ? true : false }};
    const is_notebook = {{ is_notebook ? true : false }};
```

ziesski: Lets use ternary operator to make it concise.

ziesski (changes requested):

site/app/models/gradeable/Gradeable.php
GO-viper7: sure @ziesski , it removes redundancy
let me commit suggestion right here.

site/app/templates/grading/electronic/GradingPanelHeader.twig
GO-viper7: Alright, looks good.

GO-viper7:
@ziesski 
could you please look at the static error which popped up
It says method is variable.

site/app/models/gradeable/Gradeable.php
```diff
@@ -353,6 +378,20 @@ public function __construct(Core $core, array $details) {
...
+            'files' => 'setPeerFiles',
+            'solution_notes' => 'setPeerSolutions',
+            'discussion' => 'setPeerDiscussion'
+        ];
+
+        foreach ($mapping as $key => $method) {
+            if (array_key_exists($key, $details)) {
+                $this->$method($details[$key] ?? true);
+            }
+        }
```
ziesski, line 393:
```suggestion
        foreach ($mapping as $key => $method) {
            if (array_key_exists($key, $details) && method_exists($this, $method)) {
                call_user_func([$this, $method], $details[$key] ?? true);
            }
        }
```

ziesski (changes requested):

site/app/models/gradeable/Gradeable.php
ziesski: This fixes the lint

site/app/models/gradeable/Gradeable.php
```diff
@@ -96,6 +96,16 @@
  * @method int getLimitedAccessBlind()
  * @method void setPeerBlind($peer_blind)
  * @method int getPeerBlind()
+ * @method void setPeerAutograding($peer_autograding)
+ * @method boolean getPeerAutograding()
```
IDzyre, line 100:
```suggestion
 * @method bool getPeerAutograding()
```

Apply for others

IDzyre (changes requested):

site/app/models/gradeable/Gradeable.php
GO-viper7: Alright, updated (but it was passing all checks before, this might not be the reason for failing tests)

GO-viper7:
@IDzyre could you please look at the failing cypress tests, they were passing till yesterday and failing tests aren't relevant.

site/app/templates/grading/electronic/GradingPanelHeader.twig
```diff
@@ -1,16 +1,20 @@
 {% import _self as self %}
 <script>
-    let student_info = {% if isStudentInfoPanel %} true {% else %} false {% endif %};
-    let grade_inquiry_info = {% if isGradeInquiryPanel %} true {% else %} false {% endif %};
-    let discussion_browser = {% if isDiscussionPanel %} true {% else %} false {% endif %};
-    let peer_info = {% if isPeerPanel %} true {% else %} false {% endif %};
-    let is_notebook = {% if is_notebook %} true {% else %} false {% endif %};
+    const autograding_results = {{ not isPeerGrader or (isPeerGrader and isPeerAutograding) ? true : false }};
```
williamschen23, line 3: Use 0 and 1's here as discussed in the meeting

williamschen23: Also be careful about comparing (===) 0 to false

williamschen23: let student_info = {% if isStudentInfoPanel %} true {% else %} false {% endif %};
    These still work, so I think we should just use these statement because its much clearer than why we are using 0 and 1's

GO-viper7: You are right, we should just revert back to if else because of 
1. Strong comparison we are doing across the site.
2. True or false is much clearer.

williamschen23:
:partying_face: :tada:

GO-viper7:
> 🥳 🎉

Thanks Williams!

ziesski (approved):
Looks good to me

PR #10198, opened 2024-02-21
[Bugfix:InstructorUI] Different repo_name in generate_repos.py

### LINKED ISSUE #10137: incorrect errors for generate_repos script when specified gradeable does not exist
**Describe the bug**
We have bugs in the generate_repos script if improper arguments are provided.

Documentation is here:
https://submitty.org/instructor/assignment_preparation/managing_git

```
sudo /usr/local/submitty/bin/generate_repos.py <semester> <course_code> <project_name/gradeable_id>
```

Firstly, if you specify a gradeable_id that does exist but is not a vcs gradeable, it goes ahead and makes repositories.
The script should warn the instructor that this gradeable is not a vcs gradeable.
And then the script should stop and not make the repositories.

Secondly, if the specified `gradeable_id` doesn't exist, we get a warning, and the instructor is asked if they want to continue.
If they say yes, but the script continues and then fails with this error:
AttributeError: 'NoneType' object has no attribute 'eg_vcs_subdirectory'
which is confusing to the instructor. 

The warning needs to be more firm (it was too easy to say yes).
It should say "please make the gradeable before attempting to run this script.

However, we do have an option to make a gradeable that has repos that do not match the gradeable id.
This is intended for individual (non team) assignments that will share a repo over multiple gradeables.
In this case the instructor specifies the repo name that will be used in common for multiple gradeables:

Screenshot 2024-01-29 at 10 05 03 AM

So if the gradeable_id doesn't exist, the script should check if any gradeable has the specified repo name specified as the the shared repository name (for one or more gradeables).  The script should only continue if that is true.  And there may be a bug here...  this error should not happen:
AttributeError: 'NoneType' object has no attribute 'eg_vcs_subdirectory'

Xiayucheng1212:
Fixes #10137 

### What is the new behavior?
First create 2 different gradeable to test the following three conditions.

Gradeable 1: Not a VCS gradeable
Gradeable 2: A VCS gradeable with vcs_type = " choose repository name (can be used for multiple gradeables)"
截圖 2024-02-21 下午11 55 16

## Test result
Case 1: It will exit directly and show warning messages. (use Gradeable 1)
case1_result
Case 2: It will show different warning message, asking the user to create a gradeable in advance. (Random string to mock "gradeable id")
case2_result
Case 3: It will show that the "repo_name" was found in at least one gradeable, and then ask the user whether to continue creating/update. (use Gradeable 2)
vcs-multiople-result

### Notes
I am not sure whether this work flow is acceptable by users, please give me some suggestions! Thank you!

### DIFF
--- a/bin/generate_repos.py
+++ b/bin/generate_repos.py
@@ -178,7 +178,7 @@ is_team = False
 course_conn_string = db_utils.generate_connect_string(
     DATABASE_HOST,
     DATABASE_PORT,
-    f"submitty_{args.semester}_{args.course}", # database name
+    f"submitty_{args.semester}_{args.course}",
     DATABASE_USER,
     DATABASE_PASS,
 )
@@ -208,7 +208,7 @@ elif not args.non_interactive:
     for gradeable in gradeables:
         # the eg_vcs_partial_path has pattern like `gradeable_id/user_id`, so we need to use regex to match the gradeable_id
         if gradeable.eg_vcs_host_type == 1 and re.match(f'^{args.repo_name}/', gradeable.eg_vcs_partial_path):
-            print("Find matching gradeable_id '{}' in the course.".format(gradeable.g_id))
+            print("Found matching gradeable_id '{}' in the course that uses the requested repo name.".format(gradeable.g_id))
             is_repo_name_in_gradeables = True
             response = input ("Should we continue and make individual repositories named '"+args.repo_name+"' for each student/team? (y/n) ")
             if not response.lower() == 'y':
@@ -216,7 +216,7 @@ elif not args.non_interactive:
                 sys.exit()
 
     if is_repo_name_in_gradeables == False:
-        print ("Warning: Please make the gradeable before attempting to run this script!")
+        print ("ERROR: Please make the gradeable before attempting to run this script!")
         print ("exiting")
         sys.exit()
 


williamjallen:
Hi @Xiayucheng1212, thanks for this contribution!  I just approved a CI run for you.  I'm not particularly knowledgeable about this script, but I'll try to track down someone who is able to review this appropriately for you.

Xiayucheng1212:
> Hi @Xiayucheng1212, thanks for this contribution! I just approved a CI run for you. I'm not particularly knowledgeable about this script, but I'll try to track down someone who is able to review this appropriately for you.

Sure! Thank you for your help!

bin/generate_repos.py
```diff
@@ -191,19 +193,38 @@ def create_or_update_repo(folder, subdirectory, which_branch):
...
+    if eg.eg_vcs_host_type == -1:
+        print((WARNING + "Warning: Semester '{}' and Course '{}' contains gradeable_id '{}' but it is not a VCS gradeable." + DEFAULT).format(args.semester, args.course, args.repo_name))
+        print("exiting")
+        sys.exit()
+# gradeable does not exist in the course
 elif not args.non_interactive:
-    print ("Warning: Semester '{}' and Course '{}' does not contain gradeable_id '{}'.".format(args.semester, args.course, args.repo_name))
-    response = input ("Should we continue and make individual repositories named '"+args.repo_name+"' for each student? (y/n) ")
-    if not response.lower() == 'y':
+    print ((WARNING + "Warning: Semester '{}' and Course '{}' does not contain gradeable_id '{}'." + DEFAULT).format(args.semester, args.course, args.repo_name))
```
cjreed121, line 203: Can you remove the colors on the print. This output may go to a log so it doesn't make sense to put the color there.

bin/generate_repos.py
```diff
@@ -191,19 +193,38 @@ def create_or_update_repo(folder, subdirectory, which_branch):
...
+    is_repo_name_in_gradeables = False
+    for gradeable in gradeables:
+        # the eg_vcs_partial_path has pattern like `gradeable_id/user_id`, so we need to use regex to match the gradeable_id
+        if gradeable.eg_vcs_host_type == 1 and re.match(f'^{args.repo_name}/', gradeable.eg_vcs_partial_path):
+            print("Find matching gradeable_id '{}' in the course.".format(gradeable.g_id))
+            is_repo_name_in_gradeables = True
+            response = input ("Should we continue and make individual repositories named '"+args.repo_name+"' for each student/team? (y/n) ")
+            if not response.lower() == 'y':
+                print ("exiting")
+                sys.exit()
```
cjreed121, line 215: If they push 'y' then it should fetch the necessary variables for this gradeable and then continue generating repos, not exit.

cjreed121 (changes requested):

bin/generate_repos.py
Xiayucheng1212: Sure, done.

bin/generate_repos.py
Xiayucheng1212: Yes, it will continue generating repos. It will exit only when the reponse is not 'y'.

ZheyuDeng (approved):
Was able to reproduce all three cases. When I was testing the third case I got fatal: detected dubious ownership in repository at '/var/local/submitty...' at first and was able to run the script after bypassing it.

7e6d3ac0c2f65cfd876f7c8935bc847

Issue #10202, opened 2024-02-23
fix the expended issue for the text bar 


JWei17:
**Describe the bug**
The text bar in the edit post page should be expanded initially. But currently, the text bar won't get expanded until the user make change to the text in the text bar.  

**Expected behavior**
When the instructor or grader entering the edit post page, the text bar in that page should be automatically expanded.  

**To Reproduce**
1. Login as an instructor
2. Go to ether post under Discussion forum, and click edit the post.
3. The text bar won't get expanded until you make change in the text.

**Configuration**
 - OS: Windows
 - Browser Chrome

**Screenshots**
![Screenshot 2024-02-23 171738](https://github.com/Submitty/Submitty/assets/127140351/64bd353a-fc4e-4d3d-a937-6500d6458f09)

**Additional context**

PR #10203, opened 2024-02-23
[Bugfix:Forum] text box sizing for edit post

ZheyuDeng:
Fixes [10202](https://github.com/Submitty/Submitty/issues/10202)
Also Fixes [10258](https://github.com/Submitty/Submitty/issues/10258)
When editing longer threads, instructores can not see the full thread without typing things in.

### What is the new behavior?
Now each thread expanded by default to display entire thread

image


jeffreyC4 (changes requested):
This is a great idea, but when I tested it, the reply to all textarea in the thread was significantly larger than usual on page load, and inputting the first piece of text while editing a post led to a large increase in text area height. I would recommend resizing all text areas at build time. I'm also not sure why you're using overflowY on the textarea if you're increasing the height on each input. I'm attaching a picture of what I mean below. 

image

site/app/templates/misc/MarkdownArea.twig
```diff
@@ -109,14 +109,37 @@
     </div>
 </div>
 <script>
-    $('#{{markdown_area_id}}').ready(function() {
-        enableTabsInTextArea("#{{markdown_area_id}}");
-        const markdown_area = $('#{{markdown_area_id}}').closest('.markdown-area')
-        markdown_area.find('[data-initialize-preview="1"]').trigger('click');
+$(document).ready(function() {
+    function resizeTextarea(textarea) {
```
jeffrey-asm, line 113: I would make this an event listener to apply to all textarea's on the forum thread for future updates.

jeffrey-asm (changes requested):
This works great with posts that are long, but it does not increase the size of smaller posts when the size of the text increases beyond the container height. I would add an event listener to listen on input of the all textareas on the page to manually resize.

jeffrey-asm (approved):
The implementation looks solid. All of the textarea's in the forum, including both posts and replies, are properly resized to grow or size down as the input changes accordingly.

site/public/templates/misc/MarkdownArea.twig
```diff
@@ -104,14 +104,55 @@
     </div>
 </div>
 <script>
-    $('#{{markdown_area_id}}').ready(function() {
-        enableTabsInTextArea("#{{markdown_area_id}}");
-        const markdown_area = $('#{{markdown_area_id}}').closest('.markdown-area')
-        markdown_area.find('[data-initialize-preview="1"]').trigger('click');
+<script>
```
cjreed121, line 107: Remove this line. This is likely causing your Cypress tests to fail.

site/app/templates/misc/MarkdownArea.twig
```diff
@@ -109,14 +109,54 @@
...
+    // Resize textareas when the content is dynamically added
+    var observer = new MutationObserver(function(mutations) {
+        mutations.forEach(function(mutation) {
+            mutation.addedNodes.forEach(function(node) {
+                if (node.nodeType === 1 && node.matches('textarea')) {
+                    resizeTextarea(node);
+                }
+            });
+        });
+    });
```
cjreed121, line 149: This will look for any textarea and update that. We probably only want to focus on the markdown area that this is responsible for. Otherwise it could impact other areas which we don't expect. If we desire this behavior everywhere, it should be in a more global spot. If you plan to just keep this here then add an extra check to ensure the element is the correct ID.

site/public/templates/misc/MarkdownArea.twig
```diff
@@ -104,14 +104,55 @@
...
+
+    // Resize textareas when the content is dynamically added
+    var observer = new MutationObserver(function(mutations) {
+        mutations.forEach(function(mutation) {
+            mutation.addedNodes.forEach(function(node) {
+                if (node.nodeType === 1 && node.matches('textarea')) {
+                    resizeTextarea(node);
+                }
+            });
+        });
```
cjreed121, line 144: Same here

cjreed121 (changes requested):

williamjallen (changes requested):
Many of the changes seem to be duplicated in `site/app/templates/misc/MarkdownArea.twig` and `site/public/templates/misc/MarkdownArea.twig`.

Additionally, there are a significant number of JS lint errors which are not being enforced because this code is embedded in Twig files.  Would it be reasonable to move the JS to a dedicated file so it can be linted?

ZheyuDeng:
> Many of the changes seem to be duplicated in `site/app/templates/misc/MarkdownArea.twig` and `site/public/templates/misc/MarkdownArea.twig`.
> 
> Additionally, there are a significant number of JS lint errors which are not being enforced because this code is embedded in Twig files. Would it be reasonable to move the JS to a dedicated file so it can be linted?

At the top of this two files it says I will need to update both file consistently though

site/app/templates/misc/MarkdownArea.twig
```diff
@@ -108,15 +108,4 @@
...
-        markdown_area.find('[data-initialize-preview="1"]').trigger('click');
-    });
-    $('.markdown-mode-tab').on('click', function(){
-        $(this).addClass('active');
-        const markdown_area = $(this).closest('.markdown-area');
-        markdown_area.find('.markdown-mode-tab').not(this).removeClass('active');
-    });
-</script>
\ No newline at end of file
+<script src="/js/markdown.js"></script>
```
williamjallen, line 111: I think it would make sense to add this file to every request via `setInternalResources()` in `Output.php`.

williamjallen (changes requested):
This does not appear to fix the problem.  When I open a forum post, the box is fixed at the small initial size, and cannot be resized.  It also doesn't get resized when the box is clicked.

ZheyuDeng:
> This does not appear to fix the problem. When I open a forum post, the box is fixed at the small initial size, and cannot be resized. It also doesn't get resized when the box is clicked.

Hum this is strange I tested on my machine and my friends there wasn't any issue though.
image

ZheyuDeng:
> This does not appear to fix the problem. When I open a forum post, the box is fixed at the small initial size, and cannot be resized. It also doesn't get resized when the box is clicked.



> This does not appear to fix the problem. When I open a forum post, the box is fixed at the small initial size, and cannot be resized. It also doesn't get resized when the box is clicked.

It was caused by the merge of main make the markdown area id from 0 to 1. I have updated it.

bmcutler (approved):
thanks!  tested this and it looks good :)

PR #10204, opened 2024-02-24
[Bugfix:Forum] Find search terms within punctuation

### LINKED ISSUE #10141: Submitty Forum search feature fails to search on certain search terms
**Describe the bug**
Submitty Forum search feature seems to be failing to produce any results for certain search terms. For instance, here is a Forum post that clearly contains term ``module-info``:

image

However when I search for ``module-info`` the resultset is just empty:

image

image

**Expected behavior**
Results that match the search term should be shown.

**To Reproduce**
1. Navigate to the Forum interface, e.g., https://submitty.cs.rpi.edu/courses/s23/csci2600/forum
2. Enter the search term, like ``module-info`` in the ``Search here...`` box
3. Click ``Search`` button
4. Observe the ``No threads match your search criteria.`` message

**Configuration**
 - OS: Windows 11 Home Version 23H2 OS build 22631.3007 Experience Windows Feature Experience Pack 1000.22681.1000.0
 - Browser: Mozilla Firefox 121.0.1 (64-bit)

**Screenshots**
Provided above

**Additional context**
If I search for a slightly different term ,like ``module`` (i.e., omitting the ``-info`` part, I get some results from the search:

image

iamdanielyun:

Fixes  #10141 

### What is the new behavior?
Able to do a more robust sql query to search for dashes as well in submitty forum serach field

*Code looks mostly the same except for reformatting, but the key detail is plainto_tsquery(:q) => phraseto_tsquery(:q)


DarthNyan (commented):
This manages to search for dashes, but I remember doing a bit of research into this and I found the root cause was that the query was searching for a whole word, maybe with punctuation, surrounded by whitespace, and that the reason the search term failed to find was because of the ".java". I think a better fix would be to search for parts of words, not necessarily full words.

I really appreciate you splitting the query into multiple lines, but I still think you can break it into more lines to improve readability.

RayReyRam (dismissed):
When testing in my machine, the search function was still not working, Logged in as instructor, then created a thread having the same message as the initial bug and it still says that there is no thread that match my criteria. Heres some screenshots
image1
image2

micpap25:
I would appreciate a review on this greatly. My observation is that periods are a "special character" in the context of submitty due to their frequent use in file names. Therefore, making it so users can easily search both the name of the file without the filetype, or the filetype without the name of the file, is important. This push resolves the linked issue but may require further testing to make sure it isn't too loose.

zacharymnp (approved):
Query results for 'file.extension' and 'file extension' will be identical, but the query itself is more robust

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -1180,7 +1180,21 @@ private function findChildren($post_id, $thread_id, &$children, $get_deleted = f
...
-        $this->course_db->query("SELECT post_content, p_id, p_author, thread_id, thread_title, author, pin, anonymous, timestamp_post FROM (SELECT t.id as thread_id, t.title as thread_title, p.id as p_id, t.created_by as author, t.pinned_expiration as pin, p.timestamp as timestamp_post, p.content as post_content, p.anonymous, p.author_user_id as p_author, to_tsvector(p.content) || to_tsvector(t.title) as document from posts p, threads t JOIN (SELECT thread_id, timestamp from posts where parent_id = -1) p2 ON p2.thread_id = t.id where t.id = p.thread_id and p.deleted=false and t.deleted=false) p_doc where p_doc.document @@ plainto_tsquery(:q) ORDER BY timestamp_post DESC", [':q' => $searchQuery]);
+        $this->course_db->query(
+            "SELECT post_content, p_id, p_author, thread_id, thread_title, author, pin, anonymous, timestamp_post
+            FROM (SELECT t.id as thread_id, t.title as thread_title, p.id as p_id,
+                t.created_by as author, t.pinned_expiration as pin, p.timestamp as timestamp_post,
+                p.content as post_content, p.anonymous, p.author_user_id as p_author,
+                to_tsvector('english', replace(p.content, '.', ' '))
```
ziesski, line 1188: Notice that the format has improved, which is more organized, but would be better to state what you have changed to address the issue, especially when the github diff checker is not parsing the diff precisely. 

In this case I noticed that language parameter in to_tsvector is specified to english, and used replace function to remove period(`.`) from the content and title field. 

I think the changes are overall good improvements.

ziesski (changes requested):
스크린샷 2024-06-03 오전 11 31 05

I found a corner case while testing this PR. 
The problem is query finds 9876 and 4444, but it will not find 5432

To reproduce, create post with `int-int=int`
search for the middle int, and see the issue.

ziesski (approved):
Tested, the query now handles the above mentioned corner case as expected. Looks good to me.

PR #10206, opened 2024-02-24
[UI/UX:InstructorUI] Fix CodeMirror size for "Notebook.twig"

### LINKED ISSUE #10182: Scroll Down Feature on some Text boxs are still broken
**Describe the bug**

When you trying to resize the text box in a gradeable, the text box resize feature is arbitrary.
截屏2024-02-13 02 28 39

**Expected behavior**

The resize should be limited. We have fixed similar resize problem before, but gradeable text box seems been overlooked.
You may refer to PR[#10016](https://github.com/Submitty/Submitty/pull/10016) to see the similar case we have been work on and you may also find a possible similar solution.

**To Reproduce**
1. Get into a gradeable with a direct text area
2. Try to resize the text box
3. See error

**Configuration**

Xiayucheng1212:
Fixes #10182 

The text box for short answer question in a notebook can be manually resized arbitrarily. It sometimes conflicts with other UI components near by. 

### What is the new behavior?

https://github.com/Submitty/Submitty/assets/34077605/6363ca8a-3766-46d4-babe-fdd79912eff0


ArnavPawar (changes requested):
Reviewed this same PR earlier still not sure if it matches the issue that you is linked. If you can confirm with other people that it matches then It will be good to merge.

Xiayucheng1212:
> Reviewed this same PR earlier still not sure if it matches the issue that you is linked. If you can confirm with other people that it matches then It will be good to merge.

@ArnavPawar  Thank you for review. I think this PR fulfills the issue #10182, but I would like to check with @VVVITH again. 
@VVVITH Hi, could you please check my PR to see it matches the issue number? Thank you!

justin-sadakhom (approved):
@Xiayucheng1212 Confirming that this PR matches #10182. Looks good from my end.

Xiayucheng1212:
@ArnavPawar The change has been confirmed. I think it's good to go now.

wuj18 (commented):
I tested the changes and everything works as expected. The textbox's width and minimum height are fixed, and resizing is only allowed vertically. However, I'm noticing another issue: when inputting a long string, it does not wrap to the next line upon reaching the maximum width. Pressing the Enter button doesn't move to a new line either. I'm not sure if it falls within the scope of this issue. If not I can create a new issue to address it.

https://github.com/Submitty/Submitty/assets/75406429/95acfdc3-6dbc-4da0-a61c-15faf6292042

Xiayucheng1212:
@wuj18 Yeah, I think it should be included in this PR. Do you have any suggestion to solve it?

micpap25:
Just a couple thoughts, looping in @ArnavPawar @wuj18 due to relevance to the issue.

Is it necessary to be able to resize the box for a "short answer" type of question where the answer should be constrained to a single line anyways? What value does increasing the height of the box provide? Perhaps we could allow users to increase and decrease the size of the box horizontally up to a limit? This only applies for short answer boxes, of course. 

For the other kinds of boxes, I think this change looks good to me. But I I would just like to change the functionality for short answer boxes.

site/public/css/gradeable-notebook.css
```diff
@@ -110,7 +110,9 @@ img {
     border: 1px solid var(--standard-light-gray) !important;
     border-radius: 4px !important;
     vertical-align: middle !important;
-    resize: both;
+    width: 100% !important;
+    resize: vertical;
+    min-height: 300px;
```
micpap25, line 115: For a short answer, this is much too large.

site/public/css/gradeable-notebook.css
```diff
@@ -110,7 +110,9 @@ img {
     border: 1px solid var(--standard-light-gray) !important;
     border-radius: 4px !important;
     vertical-align: middle !important;
-    resize: both;
+    width: 100% !important;
+    resize: vertical;
```
micpap25, line 114: For a short answer, it may be useful to be able to extend horizontally, and it is definitely not useful to be able to extend vertically.

micpap25 (changes requested):

micpap25:
Would really appreciate a review on this! How does it look graphically? Also, maybe a new name since it's also making notebook CSS a bit more flexible?

zacharymnp (approved):
https://github.com/Submitty/Submitty/assets/56311867/ea6fe3cd-4424-41b1-966d-dd50313d6f43

When short answer height is not default, only the vertical size can be changed. When short answer height is default, only the horizontal size can be changed.

micpap25:
Updated main PR post to reflect changes

Issue #10207, opened 2024-02-25
Warning on CalendarView.php 105 line leading to non-responsive calender


shreyashrpawar:
**Describe the bug**
I have just installed the submitty using VM and logged in as a superuser. But while navigating to the calender menu, it is showing a warning as "Warning : Undefined array key 0 in /usr/local/submitty/site/app/views/calendar/CalendarView.php on line 105". Because of the warning the Calender is leaving it's location and aligning to the left of the screen which is not giving a good experience. I have added the respective screenshot below.

**Expected behavior**
There should not be any warning showing on the page. Also, the calender should align in the center of the page.

**To Reproduce**
1. Login as superuser
2. Go to calender menu

I have installed Submitty first time. So, i don't know if the existing users are getting this error or no.

**Configuration**
 - OS: Windows
 - Browser : Chrome

**Screenshots**
![submitty bug](https://github.com/Submitty/Submitty/assets/87687490/505e55f0-b03c-45b3-bcb8-fd9ef0c593ef)

PR #10209, opened 2024-02-25
[Bugfix:Calendar] Check if array null in CalendarView.php

shreyashrpawar:
Solved issue  #10207

The issue is generating because we are taking a value of first index of an array which is empty. So, I have added nested ternary block which will take that first element if the length of an array is greater than 0.

Now, because of the modification it is working fine

### DIFF
--- a/site/app/views/calendar/CalendarView.php
+++ b/site/app/views/calendar/CalendarView.php
@@ -102,7 +102,7 @@ class CalendarView extends AbstractView {
             "show_legend" => $show_legend,
             "color_options" => $course_colors,
             "show_all_cookie" => isset($_COOKIE['calendar_show_all']) ? $_COOKIE['calendar_show_all'] : 1,
-            "calendar_course_cookie" => isset($_COOKIE['calendar_course']) ? $_COOKIE['calendar_course'] : (count($formatted_courses)>0 ? $formatted_courses[0] : NULL),
+            "calendar_course_cookie" => $_COOKIE['calendar_course'] ?? ($formatted_courses[0] ?? NULL),
         ]);
     }
 }
--- a/site/app/views/calendar/CalendarView.php
+++ b/site/app/views/calendar/CalendarView.php
@@ -102,7 +102,7 @@ class CalendarView extends AbstractView {
             "show_legend" => $show_legend,
             "color_options" => $course_colors,
             "show_all_cookie" => isset($_COOKIE['calendar_show_all']) ? $_COOKIE['calendar_show_all'] : 1,
-            "calendar_course_cookie" => isset($_COOKIE['calendar_course']) ? $_COOKIE['calendar_course'] : $formatted_courses[0],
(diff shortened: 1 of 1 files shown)


michaelkrar (changes requested):
Not an issue in firefox. Was able to replicate the issue on chrome. The changes completely fix the issue and are non-breaking. Changes are a reasonable solution and the code is formatted well / logic is simple to follow. Also changes appear to not break anything in Firefox as well as student/ta logins.
PR Title shouuld be changed to be more specific as to what bug is being fixed, and also include the prefix: [Bugfix:Calendar]

GO-viper7 (changes requested):
PR title should be bit more concise, as it is limited to 40 characters excluding [Bugfix:Calendar] and gives title-check fail if it is not followed.

shreyashrpawar (commented):
Simplified the code as instructed by [arougey](https://github.com/arougey).
Also changed the title as suggested by [GO-viper7](https://github.com/GO-viper7).

IDzyre:
All of these warnings cause misalligned UI, it would be better to state what the warning is for, and/or where it is found, like `Fix Superuser Calendar View` or `Check if calendar array is empty`

shreyashrpawar:
> All of these warnings cause misalligned UI, it would be better to state what the warning is for, and/or where it is found, like `Fix Superuser Calendar View` or `Check if calendar array is empty`

Thanks for the suggestion. I have changed the title as more precise. Kindly approve. [IDzyre](https://github.com/IDzyre)   [GO-viper7](https://github.com/GO-viper7)   [arougey](https://github.com/arougey)   [michaelkrar](https://github.com/michaelkrar)

iamdanielyun (approved):

JoJo-ESC (approved):
Code looks good and changes can be seen on my machine. If I had any common it would be to pls change the title to calendar_fix.

site/app/views/calendar/CalendarView.php
```diff
@@ -102,7 +102,7 @@ public function showCalendar(CalendarInfo $info, array $courses, bool $in_course
             "show_legend" => $show_legend,
             "color_options" => $course_colors,
             "show_all_cookie" => isset($_COOKIE['calendar_show_all']) ? $_COOKIE['calendar_show_all'] : 1,
-            "calendar_course_cookie" => isset($_COOKIE['calendar_course']) ? $_COOKIE['calendar_course'] : $formatted_courses[0],
+            "calendar_course_cookie" => $_COOKIE['calendar_course'] ?? ($formatted_courses[0] ?? NULL),
```
IDzyre, line 105:
```suggestion
            "calendar_course_cookie" => $_COOKIE['calendar_course'] ?? ($formatted_courses[0] ?? null),
```

for php lint

IDzyre (changes requested):

IDzyre (approved):

cjreed121 (approved):
Looks good to me

PR #10214, opened 2024-03-01
[Feature:Forum] Live Chat Room

### LINKED ISSUE #7207: Live Lecture Chat
Live text chat/discussion between instructor, (possible TAs), and students was useful during pandemic 100% remote live video lectures.  This feature was integrated with Webex, Zoom, Google Meetings, etc.

When we return to the fully in-person classroom, having this same digital chat can be beneficial:
* In a large classroom, it can be hard to hear the question coming from a person in the back of the room.  Having the question typed out can improve accuracy of communication.
* Shy students may wish to stay (mostly) anonymous, but still need to ask a question or want to answer the instructor's question to the class.
* A student may want a specific detail of lecture or an announcement repeated or clarified, and another student in the course can do that.

* Let's add a "Live Chat" option to Submitty.  (Design note: We can discuss/revise the name for this feature.)
* It's a new page on Submitty accessed from the left sidebar.  
* Like the office hours queue, a member of the teaching staff can enable or disable Live Chat.  It's probably only enabled during lecture or other class meeeting times.
* When it is enabled, students & instructor can see a scrolling, dynamically updated list of messages.
* By default the author displayed for each message is the same as forum (preferred) first name plus first letter of (preferred) last name.  Like the forum instructor/grader messages should be highlighted.
* The display should be kept very simple, not threaded.  We should discuss if there are any educational-use specific features that are important...  are you asking a question - who is it to?  any of your classmates?  or you need an answer specifically from the instructor?  are you replying to a specific person.  We can have some simple labeling of messages to help instructor who is watching the feed simultaneous to lecturing.
* Formatting of messages is also (probably) very simple, probably no markdown, perhaps a total character count limit, etc.

* All messages to Live Chat and the author username are permanently saved in the database, and this information is available to the instructor (and possible entire teaching staff??).  This should sufficiently discourage inappropriate usage of the Live Chat.  
* Students do not have access to full history of Live Chat.

**Describe potential technical challenges**
This is a novel technical challenge, unlike the forum and the polls.  Performance is crucial.  We'll need to debug & tune our websockets code.

let's discuss more...

hansongu123:
### What is the new feature and why is it important?

Live text chat/discussion between instructor, (possible TAs), and students was useful during pandemic 100% remote live video lectures. This feature was integrated with Webex, Zoom, Google Meetings, etc.

When we return to the fully in-person classroom, having this same digital chat can be beneficial:

In a large classroom, it can be hard to hear the question coming from a person in the back of the room. Having the question typed out can improve accuracy of communication.
Shy students may wish to stay (mostly) anonymous, but still need to ask a question or want to answer the instructor's question to the class.
A student may want a specific detail of lecture or an announcement repeated or clarified, and another student in the course can do that.

Closes #7207.
Future PRs will refine and improve this feature.

### What is the new behavior?

Steps to enable the feature:

run submitty_install
In instructor's UI:

go to course settings

scroll down to toggle on live lecture chat:

Now a "Live Lecture Chat" option should be available on the left side menu.

Toggle start session to activate a chat room, then the chat room will be available for the students to join.
image

------------------------------------------------------------------------------------------------------------------------------------
Screenshots
image

image

454499560-3a65158d-8a4a-4c6c-bacb-7904e8bd797c

Feature includes:
Anonymous join - enable/disable in instructor's UI upon creating an editing a chatroom

Notes:
The dynamic display of messages is currently through the use of WebSockets.


migration/migrator/migrations/course/20240209021641_lecture_chat.py
```diff
@@ -0,0 +1,48 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("DROP TABLE IF EXISTS chatrooms")
+    database.execute("DROP TABLE IF EXISTS chatroom_messages")
```
cjreed121, line 46: We shouldn't drop the tables so we can prevent any data loss.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * @Route("/courses/{_semester}/{_course}/chat/{chatroom_id}/edit", name="edit_chatroom", methods={"POST"})
+ * @AccessControl(role="INSTRUCTOR")
+ */
+public function editChatroom(int $chatroom_id): RedirectResponse {
+    $em = $this->core->getCourseEntityManager();
+    $chatroom = $em->getRepository(Chatroom::class)->find($chatroom_id);
+
+    if (!$chatroom) {
```
cjreed121, line 124: Direct comparison on null would be better

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,187 @@
...
+            $em->remove($message);
+        }
+        $em->remove($chatroom);
+        $em->flush();
+
+        return JsonResponse::getSuccessResponse();
+    }
+
+/**
+ * @Route("/courses/{_semester}/{_course}/chat/{chatroom_id}/edit", name="edit_chatroom", methods={"POST"})
```
cjreed121, line 117: We don't ever use Route names so you could remove those from the docblock. Not necessary to remove them but I'm just pointing out that they don't do anything for us. So I'll leave it up to you.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    if (isset($_POST['title'])) {
+        $chatroom->setTitle($_POST['title']);
+    }
+    if (isset($_POST['description'])) {
+        $chatroom->setDescription($_POST['description']);
+    }
+
+    $em->persist($chatroom);
```
cjreed121, line 136: You don't need to persist again. You only need to persist when first creating an object.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,187 @@
...
+        /** @var Chatroom|null */
+        $chatroom = $repo->findByChatroomId($chatroom_id);
+
+        if ($chatroom === null) {
+            return JsonResponse::getFailResponse('Invalid Chatroom ID');
+        }
+
+        foreach ($chatroom->getMessages() as $message) {
+            $em->remove($message);
+        }
```
cjreed121, line 109: Will this function work without this because of the cascade you have in the entity?

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,187 @@
...
+                'content' => $message->getContent(),
+                'timestamp' => $message->getTimestamp()->format('Y-m-d H:i:s'),
+                'user_id' => $message->getUserId(),
+                'display_name' => $message->getDisplayName(),
+                'role' => $message->getRole()
+            ];
+        }, $messages);
+
+        return JsonResponse::getSuccessResponse($formattedMessages);
+    }
```
cjreed121, line 162: Will there be a restriction for only messages from that day?

site/app/entities/chat/Message.php
```diff
@@ -0,0 +1,95 @@
...
+#[ORM\Entity]
+#[ORM\Table(name: "chatroom_messages")]
+class Message {
+
+    #[ORM\Id]
+    #[ORM\Column(name: "id", type: Types::INTEGER)]
+    #[ORM\GeneratedValue]
+    private int $id;
+
+    #[ORM\ManyToOne(targetEntity: "Chatroom")]
```
cjreed121, line 21: For target entity you can use the ::class syntax to specifically refer to that class. It is better for linting purposes and would be helpful if we ever changed the directory structure.

site/app/views/ChatroomView.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public function showChatPageStudent(array $chatrooms) {
+        return $this->core->getOutput()->renderTwigTemplate("chat/ChatPageStu.twig", [
+            'csrf_token' => $this->core->getCsrfToken(),
+            'base_url' => $this->core->buildCourseUrl() . '/chat',
+            'semester' => $this->core->getConfig()->getTerm(),
+            'course' => $this->core->getConfig()->getCourse(),
+            'chatrooms' => $chatrooms
+        ]);
```
cjreed121, line 39: These are very similar (and the contents in the twig files are very similar too). Could we combine them and just do some conditional rendering in the twig file?

site/public/js/chatroom.js
```diff
@@ -0,0 +1,176 @@
...
+    }
+    message.innerHTML = `
+        <div class="message-header">
+            <span class="sender-name">${display_name}</span>
+            <span class="timestamp">${timestamp}</span>
+        </div>
+        <div class="message-content">
+            ${content}
+        </div>
+    `;
```
cjreed121, line 73: This is dangerous to do because you are using .innerHTML. This could be subject to an XSS attack. Any variable that contains data coming from the user should be sanitized. Easiest way to do this is to create any needed elements and then insert user data with .innerText or .text() if using jQuery.

cjreed121 (changes requested):
Code looks pretty good overall but I do have some requested changes.

There is also another security concern where any user can send messages over the websocket server to all other users without it being logged. They can also pretend to be other users too. In order to address this, the websocket messages must be sent from the PHP side and we must not allow clients to send these messages. The latter part might need to be done in another PR. You should be able to move the websocket send message into PHP easily by referring to either office hour or forum code.

migration/migrator/migrations/course/20240209021641_lecture_chat.py
hansongu123: I'm currently still making adjustments to the tables, will remove these two lines after the PR is converted to open

site/app/controllers/ChatroomController.php
hansongu123: I'll remove either this or cascade

site/app/views/ChatroomView.php
hansongu123: I'll try to reduce redundancy. I wanted more clarity at this point because student's and instructor's UI could be more different.

site/app/controllers/ChatroomController.php
hansongu123: Good question. Currently there's no restriction to a student's access to the history of messages. We could either set this restriction to either day-based, session-based, or decide something else.

site/app/views/ChatroomView.php
williamjallen: Although it may be tempting to have separate student and instructor UIs, we learned from the polls page that that is a bad idea.  Even though it adds additional complexity, it is important to have all of the information for a given page in one place.

migration/migrator/migrations/course/20240209021641_lecture_chat.py
```diff
@@ -0,0 +1,48 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("CREATE TABLE IF NOT EXISTS chatrooms(id SERIAL PRIMARY KEY, host_id character varying NOT NULL, host_name character varying, title text NOT NULL, description text, is_active BOOLEAN DEFAULT false NOT NULL, allow_anon BOOLEAN DEFAULT true NOT NULL)")
+    database.execute("CREATE TABLE IF NOT EXISTS chatroom_messages(id SERIAL PRIMARY KEY, chatroom_id integer NOT NULL, userId character varying NOT NULL, display_name character varying, role character varying, content text NOT NULL, timestamp timestamp(0) with time zone NOT NULL)")
```
williamjallen, line 19: Can you wrap these lines to make them more readable on most displays?

migration/migrator/migrations/course/20240209021641_lecture_chat.py
```diff
@@ -0,0 +1,48 @@
...
+    course_dir = Path(config.submitty['submitty_data_dir'], 'courses', semester, course)
+    # add boolean to course config
+    config_file = Path(course_dir, 'config', 'config.json')
+    if config_file.is_file():
+        with open(config_file, 'r') as in_file:
+            j = json.load(in_file)
+            j['course_details']['chat_enabled'] = False
+
+        with open(config_file, 'w') as out_file:
+            json.dump(j, out_file, indent=4)
```
williamjallen, line 30: Instead of editing all exiting configs, why can't the default value be false instead?  The field could then be optional.

migration/migrator/migrations/course/20240209021641_lecture_chat.py
williamjallen, line 19: These need foreign keys to be set up wherever possible.  Chatroom messages should have foreign-key relationships with chatrooms and users at a minimum.

As an aside: what is the purpose of the `host_id` and `host_name` columns?

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+use app\libraries\routers\Enabled;
+use app\views\ChatroomView;
+use Symfony\Component\Routing\Annotation\Route;
+use Doctrine\ORM\EntityManagerInterface;
+
+class ChatroomController extends AbstractController {
+
+    public function __construct(Core $core) {
+        parent::__construct($core);
+    }
```
williamjallen, line 21: This isn't necessary.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+            return new WebResponse(
+                'Chatroom',
+                'showChatPageStudent',
+                $active_chatrooms
+            );
+        }
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/chat/newChatroom", methods={"POST"})
```
williamjallen, line 49: `.../chat/newChatroom` seems a bit redundant.  What about `.../chat/new`?

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+     * @AccessControl(role="INSTRUCTOR")
+     */
+    public function addChatroom(): RedirectResponse {
+        $em = $this->core->getCourseEntityManager();
+
+        $chatroom = new Chatroom();
+        $chatroom->setTitle($_POST['title']);
+        $chatroom->setDescription($_POST['description']);
+        $chatroom->setHostId($this->core->getUser()->getId());
+        $chatroom->setHostName($this->core->getUser()->getDisplayFullName());
```
williamjallen, line 59: Would this be a good opportunity to use a constructor?  See the `Poll` type for an example.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+     * @return RedirectResponse|WebResponse
+     */
+     public function getChatroom(string $chatroom_id): WebResponse|RedirectResponse {
+         if (!is_numeric($chatroom_id)) {
+            $this->core->addErrorMessage("Invalid Chatroom ID");
+            return new RedirectResponse($this->core->buildCourseUrl(['chat']));
+        }
+
+        $repo = $this->core->getCourseEntityManager()->getRepository(Chatroom::class);
+        $chatroom = $repo->find($chatroom_id);
```
williamjallen, line 80: What if no chat room was found?

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+            $chatroom,
+        );
+     }
+
+     /**
+     * @Route("/courses/{_semester}/{_course}/chat/{chatroom_id}/anonymous", methods={"GET"})
+     * @param string $chatroom_id
+     * @return RedirectResponse|WebResponse
+     */
+     public function getChatroomAnon(string $chatroom_id): WebResponse|RedirectResponse {
```
williamjallen, line 94: What's the purpose of this route?  It appears to be essentially identical to the route without the `/anonymous`

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+
+        return new WebResponse(
+            'Chatroom',
+            'showChatroom',
+            $chatroom, true,
+        );
+     }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/chat/deleteChatroom", methods={"POST"})
```
williamjallen, line 111: What about `.../chat/delete`?

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+     * @AccessControl(role="INSTRUCTOR")
+     */
+    public function deleteChatroom(): JsonResponse {
```
williamjallen, line 114: Should we have the ability to delete chatrooms?  Perhaps we should keep them around but add a boolean "deleted" column in case there's ever a desire to dig up the chatroom in the future.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+        $em->flush();
+
+        return JsonResponse::getSuccessResponse();
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/chat/{chatroom_id}/edit", methods={"POST"})
+     * @AccessControl(role="INSTRUCTOR")
+     */
+    public function editChatroom(int $chatroom_id): RedirectResponse {
```
williamjallen, line 141: Will the `int` type hint work here?  You used `string` earlier in this file, and I seem to recall that the type is always `string` for query parameters in Submitty.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+
+        $this->core->addSuccessMessage("Chatroom successfully updated");
+        return new RedirectResponse($this->core->buildCourseUrl(['chat']));
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/chat/{chatroom_id}/toggleOnOff", methods={"POST"})
+     * @AccessControl(role="INSTRUCTOR")
+     */
+     public function toggleChatroomOnOf(int $chatroom_id): RedirectResponse {
```
williamjallen, line 173:
```suggestion
     public function toggleChatroomOnOff(int $chatroom_id): RedirectResponse {
```

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+        $formattedMessages = array_map(function ($message) {
+            return [
+                'id' => $message->getId(),
+                'content' => $message->getContent(),
+                'timestamp' => $message->getTimestamp()->format('Y-m-d H:i:s'),
+                'user_id' => $message->getUserId(),
+                'display_name' => $message->getDisplayName(),
+                'role' => $message->getRole()
+            ];
+        }, $messages);
```
williamjallen, line 206: I believe using `toArray()` on the Doctrine collection will have the same effect.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,234 @@
...
+
+        return JsonResponse::getSuccessResponse($formattedMessages);
+    }
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/chat/{chatroom_id}/send", methods={"POST"})
+     */
+    public function addMessage(int $chatroom_id): JsonResponse {
+        $em = $this->core->getCourseEntityManager();
+        $content = $_POST['content'] ?? '';
```
williamjallen, line 216: Should we allow people to send blank messages?

williamjallen: It would probably be good to add constraints like this to the database table as well.

site/app/models/Chatroom.php
williamjallen, line 1: Is there a reason why there's an empty file here?

site/app/repositories/chat/ChatroomRepository.php
```diff
@@ -0,0 +1,35 @@
...
+                    ->getResult();
+    }
+
+    public function findAllInactiveChatrooms() {
+        return $this->createQueryBuilder('c')
+                    ->where('c.isActive = :isActive')
+                    ->setParameter('isActive', false)
+                    ->getQuery()
+                    ->getResult();
+    }
```
williamjallen, line 34: I believe it's recommended to use DQL instead of the query builder for performance.

williamjallen (changes requested):
I haven't finished reviewing the entire PR yet, but here's my first round of comments.  In general, things look very good, but we need to make sure everything is robust since this will be exposed to hundreds of students at a time.  This also needs to have some tests added.  For now, some basic Cypress tests would be good at least, if only to ensure that there are no repeated queries occurring behind the scenes.  Any sort of repeated queries would be detrimental to the success of this feature.

site/app/models/Chatroom.php
hansongu123: nope

migration/migrator/migrations/course/20240209021641_lecture_chat.py
hansongu123: These columns currently don't have other purposes except for showing the name of the host on students' UI. I intended. to include them to prepare for if we plan to have different hosts hosting different chat rooms in the future.

site/app/controllers/ChatroomController.php
hansongu123: I thought of this before. I added a column in the database table to allow a chat room to be "session" based. Toggle on a chat room will allow it to be active and available to the students, and stopping the session will take away students' ability to type into the chat and/or entering the chat room. So instead of a "deleted" column, a chat room would just be "inactive", and the host and activate it again easily if they wish to dig it up or reuse it.

On the other hand if a chat room is deleted it is deleted, so there's no confusion.

RayReyRam (changes requested):
I switched to this branch and ran a sumbitty_install as instructor went to sample and have this error preventing me to enter 
error

hansongu123:
> I switched to this branch and ran a sumbitty_install as instructor went to sample and have this error preventing me to enter 

@RayReyRam I updated the migration and the dump file, can you pull from this branch and try again? If that doesn't work, try (in vagrant ssh) calling `migrator -e course rollback` and` migrator -e course migrate` to manually apply the migration. If none of this works, try re-creating the sample courses: 
`sudo bash /usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/bin/recreate_sample_courses.sh --no_submissions` in vagrant ssh.

@williamjallen Is this caused by the removal of the script in the migration file that edits all exiting configs?

RayReyRam (changes requested):
Manually migrated worked.
I believe that live chat feature should be on by default, as I had trouble finding it and stumbled upon it luckily in settings. 

I also think there should be a reply feature, maybe a student knows the answer to another student's question, or a ta directly wants to answer and could reply to it. 

Just a thought but maybe have some sort of tag feature, like if it's a question a tag could be placed making easier for the instructor/ta to see. 

I found an error that when the instructor ends the session of the live chat, the student is still able to message in it as long they stay on it. Or just go back a page. They should be kicked out the live chat, unless that is intended.

Also don't remember if this was talked about but allowing a student to view a previously closed session. This could be a seeing toggled by the instructor, but i feel like it would be useful if a student can go back into the chat especially if questions were answered in the session.

Other than that the live chat seems to work I'm able to message through 2 accounts without any problem.

hansongu123:
@RayReyRam Thanks for the review, I find it really helpful! I added a step in the PR description for how to enable it as a sidebar item. I haven't added the mechanism to disable students to entering messages upon terminating a session, or limiting students' access to the entire chat history just yet. I would like to hear back from the reviewers and decide how to design these mechanisms.

JoJo-ESC (commented):
Overall I love this idea and I think it was executed exceptionally well. There are a couple things that I would like to see changed before merged.

If live chat won't be enabled automatically, please, if possible, move the live lecture on option to be right below the forum option. I think grouping things that deal with similar ideas imporves clarity. Additionally Im not sure if this idea was discussed already but I really think it might be worth it to consider implementing something to stop students from blatant spam. If there's anyway to track how many messages X user has sent we can say that if they exceed a certain message threshold they get paused for a minute. Another thing is that I think messages really get lost when they start to stack up. There are 2 possible solutions either take the discord approach of doing different colors (probably won't work due to sheer number of users per live chat) or the second option is to implement 2 changes 1 is to highlight in some shade of yellow, and allow instructors to  pin/highlight comments of the instructors choice and give them some special shade as well.

This is an excellent addition to submitty and haas the potential to positively impact many users experience.

cjreed121:
This is on hold of being merged till https://github.com/Submitty/Submitty/issues/10296 is closed

williamschen23 (changes requested):
all routes should be refactored to use php8 syntax
Example: `#[Route("/home/courses", methods: ["GET"])]`

zacharymnp (changes requested):
There is some unusual behavior that should be corrected.

A chatroom can be toggled between anonymous and not anonymous, but the checkbox is always unchecked. An anonymous chatroom needs to be checked on, then back off to turn off anonymous join.
https://github.com/user-attachments/assets/3d4512cd-40f5-48ea-a9f6-ea707b2866de

Leaving and then re-entering a chat changes any message from the user from USERNAME to 'me'.
https://github.com/user-attachments/assets/16ec65c6-4031-4988-bc19-d50f1b423cd2

Leaving and re-entering a chatroom as an anon will give a new ID every time. This may be intentional, but I think it should be consistent if it is possible.
https://github.com/user-attachments/assets/5488872b-57c6-42aa-b165-6592bae1cae4

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,267 @@
...
+        $em = $this->core->getCourseEntityManager();
+
+        $repo = $em->getRepository(Chatroom::class);
+
+        $chatroom = $repo->find($chatroom_id);
+        if ($chatroom === null) {
+            return JsonResponse::getFailResponse('Invalid Chatroom ID');
+        }
+        foreach ($chatroom->getMessages() as $message) {
+            $em->remove($message);
```
jeffrey-cordero: Wouldn't removing the chatroom itself here be more efficient in the case of hundreds of messages for a given instance? `chatroom_id integer NOT NULL REFERENCES chatrooms(id) ON DELETE CASCADE` should handle deleting the messages via cascade.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        if (isset($_POST['title'])) {
+            $chatroom->setTitle($_POST['title']);
+        }
+        if (isset($_POST['description'])) {
+            $chatroom->setDescription($_POST['description']);
+        }
+        if (isset($_POST['allow-anon'])) {
+            $chatroom->setAllowAnon(true);
```
jeffrey-cordero: We could simplify this into one line - `$chatroom->setAllowAnon(isset($_POST['allow-anon']));`

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,267 @@
...
+                $display_name = $user->getDisplayFullName();
+            }
+            else {
+                $display_name = $user->getDisplayedGivenName() . " " . substr($user->getDisplayedFamilyName(), 0, 1) . ".";
+            }
+        }
+
+        $role = $user->accessAdmin() ? 'instructor' : 'student';
+        $msg_json = [];
+        $msg_json['content'] = $_POST['content'];
```
jeffrey-cordero: Verifying non-empty content would be nice here to not send it to others over the WebSocket.

site/app/templates/chat/AllChatroomsPage.twig
```diff
@@ -0,0 +1,119 @@
...
+                    <tr>
+                        <th scope="col">Name</th>
+                        <th scope="col">Host</th>
+                        <th scope="col">Description</th>
+                        <th scope="col" style="text-align: left"></th>
+                    </tr>
+                </thead>
+                <tbody>
+                    {% for chatroom in chatrooms %}
+                        {% if chatroom.isActive() %} <!--students only have access to active chat rooms-->
```
jeffrey-cordero, line 88: Would additional functionality to allow instructors to disable new messages, but still allow students to read prior messages, be a good addition to the current work?

site/app/views/ChatroomView.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public function showChatroom(Chatroom $chatroom, bool $anonymous = false): string {
+        $this->core->getOutput()->addBreadcrumb("Chatroom");
+        $user = $this->core->getUser();
+        $display_name = $user->getDisplayFullName();
+        $roomId = $chatroom->getId();
+        $sessKey = "anon_name_chatroom_{$roomId}";
+        $boolKey = "anon_name_chatroom_{$roomId}_bool";
```
jeffrey-cordero: I don't see `boolKey` being invoked within this method. Probably a good idea to remove it.

site/public/js/chatroom.js
```diff
@@ -0,0 +1,237 @@
+/* global csrfToken */
+
+// eslint-disable-next-line no-unused-vars
+function fetchMessages(chatroomId, my_id, when = new Date(0)) {
+    $.ajax({
+        // eslint-disable-next-line no-undef
```
jeffrey-cordero: You can remove this ESLint line if you add `buildCourseUrl` to the global line above. Same goes for  `displayErrorMessage` - `/* global csrfToken, buildCourseUrl, displayErrorMessage */`

site/public/js/chatroom.js
```diff
@@ -0,0 +1,237 @@
+/* global csrfToken */
+
+// eslint-disable-next-line no-unused-vars
+function fetchMessages(chatroomId, my_id, when = new Date(0)) {
```
jeffrey-cordero: None of the function params after `chatroomId` are required here.

site/public/js/chatroom.js
```diff
@@ -0,0 +1,237 @@
+/* global csrfToken */
+
+// eslint-disable-next-line no-unused-vars
```
jeffrey-cordero: Many of these `no-unused-vars` rules don't apply to the file, as shown in the following output below. Please remove as many as you can.

```bash
$ npm run eslint

> eslint
> eslint .


/.../site/public/js/chatroom.js
  116:1  warning  Unused eslint-disable directive (no problems were reported from 'no-unused-vars')
  123:1  warning  Unused eslint-disable directive (no problems were reported from 'no-unused-vars')
  133:1  warning  Unused eslint-disable directive (no problems were reported from 'no-unused-vars')
  171:1  warning  Unused eslint-disable directive (no problems were reported from 'no-unused-vars')
  198:9  warning  Unused eslint-disable directive (no problems were reported from 'no-unused-vars')
```

site/public/js/chatroom.js
```diff
@@ -0,0 +1,237 @@
...
+            csrf_token: csrfToken,
+            user_id: userId,
+            display_name: displayName,
+            role: role,
+            content: content,
+        },
+        success: function (response) {
+            try {
+                // eslint-disable-next-line no-unused-vars
+                const json = JSON.parse(response);
```
jeffrey-cordero: For simplicity, I'd recommend simply having this line as `JSON.parse(response);` as we don't need to use the value here, but just verify we can parse the server response.

site/public/js/chatroom.js
```diff
@@ -0,0 +1,237 @@
...
+        error: function () {
+            window.alert('Something went wrong with storing message');
+        },
+    });
+}
+
+function appendMessage(displayName, role, ts, content) {
+    let timestamp = ts;
+    if (!timestamp) {
+        timestamp = new Date(Date.now()).toLocaleString('en-us', { year: 'numeric', month: 'short', day: 'numeric', hour: 'numeric', minute: 'numeric' });
```
jeffrey-cordero: This block could be simplified with input of `ts || Date.now()` for the `Date` constructor. I don't see when `ts` would be `undefined` as it's based on well-formatted `GET` request responses or WebSocket messages from the server.

site/public/js/chatroom.js
```diff
@@ -0,0 +1,237 @@
...
+    appendMessage(msg.display_name, msg.role, msg.timestamp, msg.content);
+}
+
+function initChatroomSocketClient(chatroomId) {
+    // eslint-disable-next-line no-undef
+    window.socketClient = new WebSocketClient();
+    window.socketClient.onmessage = (msg) => {
+        switch (msg.type) {
+            case 'chat_message':
+                socketChatMessageHandler(msg);
```
jeffrey-cordero, line 101: I'd add another `default` block to either log the invalid WebSocket message to via `console.error` or add an error banner. The former seems more reasonable for many concurrent messages.

site/public/js/chatroom.js
jeffrey-cordero: Wouldn't checking the `status` property from the server response be more reasonable here? It'd also align with the other methods in the file.

site/public/js/chatroom.js
```diff
@@ -0,0 +1,237 @@
...
+                window.alert('Something went wrong. Please try again.');
+            },
+        });
+    }
+}
+
+// eslint-disable-next-line no-unused-vars
+function toggle_chatroom(chatroomId, active) {
+    const form = document.getElementById(`chatroom_toggle_form_${chatroomId}`);
+    if (active) {
```
jeffrey-cordero: I'd recommend to update the boolean condition to `active && confirm('This will terminate this chatroom session. Are you sure?')` to not follow this questionable nested `if` statement approach.

site/app/templates/chat/AllChatroomsPage.twig
```diff
@@ -0,0 +1,119 @@
...
+                                        <i class="fas fa-pause"></i>
+                                        end session
+                                    </button>
+                                {% endif %}
+                            </td>
+                            <td>
+                                <a href="{{ base_url }}/{{ chatroom.getId() }}" class="btn btn-primary">join</a>
+                                {% if chatroom.isAllowAnon() %}
+                                    <i> or </i>
+                                    <a href="{{ base_url }}/{{ chatroom.getId() }}/anonymous" class="btn btn-default">join as anon.</a>
```
jeffrey-cordero: It'd be more consistent with the site to have these buttons capitalized. Similarly, for the `leave` button as well, within an active chatroom.

image

site/app/entities/chat/Chatroom.php
```diff
@@ -0,0 +1,114 @@
...
+
+    public function getDescription(): string {
+        return $this->description;
+    }
+
+    public function setDescription(string $description): void {
+        $this->description = $description;
+    }
+
+    public function toggle_on_off(): void {
```
jeffrey-cordero: `toggleActiveStatus` would make this a bit more clearer for future maintainers. Similarly for `toggleChatroomOnOff` to `toggleChatroomActiveStatus`.

migration/migrator/migrations/course/20250325021641_lecture_chat.py
```diff
@@ -0,0 +1,68 @@
...
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    database.execute("DROP TABLE IF EXISTS chatrooms cascade")
```
jeffrey-cordero: Would manual dropping of the tables be more appropriate for keeping an existing log of these chatroom messages?

site/public/js/chatroom.js
```diff
@@ -0,0 +1,237 @@
...
+}
+
+document.addEventListener('DOMContentLoaded', () => {
+    const pageDataElement = document.getElementById('page-data');
+    if (pageDataElement) {
+        const pageData = JSON.parse(pageDataElement.textContent);
+        // eslint-disable-next-line no-unused-vars
+        const { chatroomId, userId, displayName, user_admin, isAnonymous } = pageData;
+
+        showJoinMessage(`You have successfully joined as ${displayName}.`);
```
jeffrey-cordero, line 268: Displaying this on every page load may be bothersome for most users. 

image

I believe it'd be cleaner to have it above the input container. Additionally, I believe it'd be more consistent with the site to leverage the existing notification and button components, but this is more of a UI/UX decision for everyone to agree on.

image

site/app/templates/chat/Chatroom.twig
```diff
@@ -0,0 +1,26 @@
...
+    <div class="chatroom-container">
+        <div class="chatroom-header">
+            <span class="chatroom-title" title="{{ chatroom.getTitle() }}"> {{ chatroom.getTitle() | length > 40 ? chatroom.getTitle() | slice(0, 40) ~ '...' : chatroom.getTitle() }} </span>
+            <a href="{{ base_url }}" class="leave-room btn btn-danger">leave</a>
+        </div>
+        <div class="chatroom-content">
+            <div class="messages-area"></div>
+            <div class="input-container">
+                    <textarea name="content" class="message-input" placeholder="enter your message here..." maxlength=1500></textarea>
+                <button class="send-message-btn btn btn-primary">Send</button>
```
jeffrey-cordero, line 12: I'd recommend to make this more mobile-friendly as many students use Submitty on their phones.

image

site/public/js/chatroom.js
```diff
@@ -0,0 +1,237 @@
...
+    });
+}
+
+function appendMessage(displayName, role, ts, content) {
+    let timestamp = ts;
+    if (!timestamp) {
+        timestamp = new Date(Date.now()).toLocaleString('en-us', { year: 'numeric', month: 'short', day: 'numeric', hour: 'numeric', minute: 'numeric' });
+    }
+    else {
+        timestamp = new Date(ts).toLocaleString('en-us', { year: 'numeric', month: 'short', day: 'numeric', hour: 'numeric', minute: 'numeric' });
```
jeffrey-cordero: I'd also recommend to add the seconds to the timestamp to handle multiple messages within the same minute.

site/app/templates/chat/Chatroom.twig
```diff
@@ -0,0 +1,26 @@
...
+<div class="content chatroom-page dim-overlay">
+    <div class="chatroom-container">
+        <div class="chatroom-header">
+            <span class="chatroom-title" title="{{ chatroom.getTitle() }}"> {{ chatroom.getTitle() | length > 40 ? chatroom.getTitle() | slice(0, 40) ~ '...' : chatroom.getTitle() }} </span>
+            <a href="{{ base_url }}" class="leave-room btn btn-danger">leave</a>
+        </div>
+        <div class="chatroom-content">
+            <div class="messages-area"></div>
+            <div class="input-container">
+                    <textarea name="content" class="message-input" placeholder="enter your message here..." maxlength=1500></textarea>
```
jeffrey-cordero: `Enter your message here` would be a cleaner message

site/public/css/chatroom.css
```diff
@@ -0,0 +1,108 @@
+.chatroom-container {
```
jeffrey-cordero, line 1: I believe it would be more user-friendly to make the container expand the entire width. A better divider between the messages and chatroom header would also be nice, if possible.

image

jeffrey-cordero (changes requested):
This is a solid implementation so far for adding live lecture chatrooms into Submitty. I've left many comments related to improvements that could be made, mainly related to the styling of the components or code optimizations that should be looked into.

site/app/templates/chat/AllChatroomsPage.twig
martig7: It's definitely a good idea, probably something to take note of for a future PR to be honest.

site/public/js/chatroom.js
martig7: I moved the badge location, not sure about using the notification component and button though.

migration/migrator/migrations/course/20250325021641_lecture_chat.py
martig7: That would make more sense yeah.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,264 @@
...
+        $msg_json['user_id'] = $user_id;
+        $msg_json['display_name'] = $display_name;
+        $msg_json['role'] = $role;
+        $msg_json['type'] = 'chat_message';
+        $msg_json['timestamp'] = date("Y-m-d H:i:s");
+        $msg_json['page'] = $this->core->getConfig()->getTerm() . '-' . $this->core->getConfig()->getCourse() . "-chatroom_$chatroom_id";
+        $message = new Message($msg_json['user_id'], $msg_json['display_name'], $msg_json['role'], $msg_json['content'], $chatroom);
+
+        try {
+            $client = new Client($this->core);
```
jeffrey-cordero: Forgot to mention this, but I would recommend having this in a separate function to handle setting up and sending WS messages with method signature `private function sendSocketMessage(array $msg_array): void`. I'd look into `OfficHoursQueueController.php` to see how to implement this to ensure consistency across the codebase.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,264 @@
...
+    /**
+     * @AccessControl(role="INSTRUCTOR")
+     */
+    #[Route("/courses/{_semester}/{_course}/chat/new", methods:["POST"])]
+    public function addChatroom(): RedirectResponse {
+        $em = $this->core->getCourseEntityManager();
+
+        $hostId = $this->core->getUser()->getId();
+        $hostName = $this->core->getUser()->getDisplayFullName();
+        $chatroom = new Chatroom($hostId, $hostName, $_POST['title'], $_POST['description']);
```
jeffrey-cordero: Please ensure to return an error if `$_POST['title']` is an empty string after trimming it.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,264 @@
...
+    public function editChatroom(string $chatroom_id): RedirectResponse {
+        $em = $this->core->getCourseEntityManager();
+        $chatroom = $em->getRepository(Chatroom::class)->find($chatroom_id);
+
+        if (!$chatroom) {
+            $this->core->addErrorMessage("Chatroom not found");
+            return new RedirectResponse($this->core->buildCourseUrl(['chat']));
+        }
+
+        if (isset($_POST['title'])) {
```
jeffrey-cordero: Similarly, please ensure the title is non-empty.

site/app/templates/chat/Chatroom.twig
```diff
@@ -0,0 +1,26 @@
+<div class="chatroom-toast"></div>
+<div class="content chatroom-page dim-overlay">
+    <div class="chatroom-container">
+        <div class="chatroom-header">
+            <span class="chatroom-title" title="{{ chatroom.getTitle() }}"> {{ chatroom.getTitle() | length > 40 ? chatroom.getTitle() | slice(0, 40) ~ '...' : chatroom.getTitle() }} </span>
+            <a href="{{ base_url }}" class="leave-room btn btn-danger">leave</a>
+        </div>
+        <div class="chatroom-content">
+            <div class="messages-area"></div>
+            <div class="input-container">
```
jeffrey-cordero, line 10: You can fix the responsiveness to appear as shown below. You'd override the `input-container` or add an additional class to have `min-width: none; flex: 1;`.

image

site/app/templates/chat/Chatroom.twig
```diff
@@ -0,0 +1,26 @@
+<div class="chatroom-toast"></div>
+<div class="content chatroom-page dim-overlay">
+    <div class="chatroom-container">
+        <div class="chatroom-header">
+            <span class="chatroom-title" title="{{ chatroom.getTitle() }}"> {{ chatroom.getTitle() | length > 40 ? chatroom.getTitle() | slice(0, 40) ~ '...' : chatroom.getTitle() }} </span>
+            <a href="{{ base_url }}" class="leave-room btn btn-danger">leave</a>
```
jeffrey-cordero: This should be capitalized.

site/app/templates/chat/Chatroom.twig
```diff
@@ -0,0 +1,26 @@
...
+        </div>
+        <div class="chatroom-content">
+            <div class="messages-area"></div>
+            <div class="input-container">
+                    <textarea name="content" class="message-input" placeholder="Enter your message here" maxlength=1500></textarea>
```
jeffrey-cordero, line 11: This should be an `<input type="text">` or you'd remove the enter key functionality to allow users to enter long text.

site/public/css/chatroom.css
```diff
@@ -0,0 +1,118 @@
...
+    font-size: 1rem;
+    min-width: 85%;
+    margin-left: 1rem;
+}
+
+.message-input:focus {
+    background-color: var(--standard-hover-light-gray);
+}
+
+.send-message-btn {
```
jeffrey-cordero, line 95: The consistency of the buttons still feels off. 

image

Can we stick to the regular "Submitty" buttons?

image

site/public/css/chatroom.css
```diff
@@ -0,0 +1,118 @@
+.chatroom-container {
+    box-shadow: 0 0 2px 0 var(--default-black);
```
jeffrey-cordero, line 2: I feel that the border feels out of sync within this page. The following removes the `box-shadow` and adds a `margin: -5px`. I'd recommend this approach.

image

site/public/css/chatroom.css
```diff
@@ -0,0 +1,118 @@
...
+    padding-top: 0.5rem;
+    margin-right: 1rem;
+}
+
+.message-container:hover {
+    background-color: var(--standard-hover-light-gray);
+}
+
+.message-header {
+    margin-bottom: 3px;
```
jeffrey-cordero: The header box shadow only seems to apply in dark mode and disappears when hovering over it. I'd recommend removing it entirely, as the divider is not necessarily needed, given vertical overflow scrolling exists.

image

site/public/css/chatroom.css
```diff
@@ -0,0 +1,118 @@
...
+}
+
+.message-input {
+    border-radius: 5px;
+    outline: none;
+    resize: none;
+    overflow: hidden;
+    font-size: 1rem;
+    min-width: 85%;
+    margin-left: 1rem;
```
jeffrey-cordero: A `margin-right: 1rem` would ensure consistency with the input margin as well.

image

site/public/css/chatroom.css
```diff
@@ -0,0 +1,118 @@
...
+
+.send-message-btn {
+    margin-left: 1rem;
+    border: none;
+    border-radius: 25px;
+    cursor: pointer;
+    outline: none;
+}
+
+.chatroom-toast {
```
jeffrey-cordero, line 99: The toast looks pretty clean. I'd recommend adding this to the PR description for future reviewers.

image

site/app/templates/chat/AllChatroomsPage.twig
```diff
@@ -0,0 +1,119 @@
...
+                                    <input type="hidden" name="csrf_token" value="{{ csrf_token }}"/>
+                                </form>
+                                {% if not chatroom.isActive() %}
+                                    <button class="btn btn-primary" onclick="toggle_chatroom({{ chatroom.getId() }}, false)">
+                                        Start Session
+                                    </button>
+                                {% else %}
+                                    <button class="btn btn-danger" onclick="toggle_chatroom({{ chatroom.getId() }}, true)">
+                                        <i class="fas fa-pause"></i>
+                                        End Session
```
jeffrey-cordero, line 59: I'd recommend sending out a WS message when a session ends or is deleted, which is similar to what `PollController.php` does. You'd disable inputs when the session ends or redirect users when a chatroom is deleted with a notice message.

site/app/templates/chat/Chatroom.twig
martig7: Do we want to remove the ability for users to enter long text? I guess there's a concern about spam but allowing new lines would allow for code to be typed in the chat which I think would be a common use case.

jeffrey-cordero: Definitely. I'd recommend having a similar input as the discussion forum, but I'm not too certain about Markdown rendering for this page specifically. I'd be nice to demo this during the meeting to get some more input.

site/public/css/chatroom.css
martig7: I like the border, but I could go either way

site/public/css/chatroom.css
martig7: I changed the header shadow to present in both light mode and dark mode. I like the header shadow, but I agree that it shouldn't disappear on hover

site/public/css/chatroom.css
martig7: I put the send in line with the leave button

site/app/templates/chat/AllChatroomsPage.twig
martig7: After much suffering, it has been done.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,288 @@
...
+        $display_name = '';
+        if ($chatroom->isAllowAnon() && $_SESSION[$boolKey] === true) {
+            $display_name = $_SESSION[$sessKey];
+        }
+        else {
+            if ($user->accessAdmin()) {
+                $display_name = $user->getDisplayFullName();
+            }
+            else {
+                $display_name = $user->getDisplayedGivenName() . " " . substr($user->getDisplayedFamilyName(), 0, 1) . ".";
```
lavalleeale:
```suggestion
                $display_name = $user->getDisplayAbbreviatedName();
```

williamjallen:
@martig7 Are the screenshots in the PR description still up-to-date?

martig7:
> @martig7 Are the screenshots in the PR description still up-to-date?

All but the chat screenshot, I'll update that now.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,306 @@
...
+            $client = new Client($this->core);
+            $client->json_send($msg_array);
+        }
+        catch (WebSocket\ConnectionException $e) {
+            $this->core->addNoticeMessage("WebSocket Server is down, page won't load dynamically.");
+        }
+    }
+
+    #[Route("/courses/{_semester}/{_course}/chat", methods: ["GET"])]
+    public function showChatroomssPage(): WebResponse {
```
cjreed121:
```suggestion
    public function showChatroomsPage(): WebResponse {
```

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,306 @@
...
+        $em->persist($chatroom);
+        $em->flush();
+
+        $this->core->addSuccessMessage("Chatroom successfully added");
+        return new RedirectResponse($this->core->buildCourseUrl(['chat']));
+    }
+
+    /**
+     * @return RedirectResponse|WebResponse
+     */
```
cjreed121: Can remove. Already documented by PHP return type

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,306 @@
...
+        return new WebResponse(
+            'Chatroom',
+            'showChatroom',
+            $chatroom,
+        );
+    }
+
+    /**
+     * @return RedirectResponse|WebResponse
+     */
```
cjreed121: Can also remove

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,306 @@
...
+    #[Route("/courses/{_semester}/{_course}/chat/{chatroom_id}/anonymous", methods: ["GET"], requirements: ["chatroom_id" => "\d+"])]
+    public function getChatroomAnon(string $chatroom_id): WebResponse|RedirectResponse {
```
cjreed121: This is almost identical to the prior function. Please refactor

site/app/entities/chat/Chatroom.php
```diff
@@ -0,0 +1,114 @@
...
+    #[ORM\Id]
+    #[ORM\Column(name: "id", type: Types::INTEGER)]
+    #[ORM\GeneratedValue]
+    private int $id;
+
+    #[ORM\Column(type: Types::STRING)]
+    private string $host_id;
+
+    #[ORM\Column(type: Types::STRING)]
+    private string $host_name;
```
cjreed121: Probably not the best idea to store this since it's duplicate (non-ID) information. It can also change if the user updates their information while it won't be reflected here. Instead it should be fetched off the user dynamically. Perhaps take advantage of the UserEntity class here. Might have to add some logic to that

site/app/repositories/chat/ChatroomRepository.php
```diff
@@ -0,0 +1,43 @@
...
+class ChatroomRepository extends EntityRepository {
+    /**
+     * @return array<Chatroom>
+     */
+    public function findAllChatroomsByHostId(string $hostId): array {
+        return $this->createQueryBuilder('c')
+                    ->where('c.host_id = :hostId')
+                    ->setParameter('hostId', $hostId)
+                    ->getQuery()
+                    ->getResult();
```
cjreed121: You aren't doing anything fancy enough to require the use of query builder. Tbh you don't even need this repository at all. You can just do a `findBy(['host_id' => $hostId])` or the like for each other method

site/app/views/ChatroomView.php
```diff
@@ -0,0 +1,91 @@
...
+                $nouns      = ["Duck","Goose","Swan","Eagle","Parrot","Owl","Sparrow","Robin","Pigeon","Falcon","Hawk","Flamingo","Pelican","Seagull","Cardinal","Canary","Finch","Hummingbird"];
+                $adj  = $adjectives[array_rand($adjectives)];
+                $noun = $nouns[array_rand($nouns)];
+                $_SESSION[$sessKey] = "Anonymous {$adj} {$noun}";
+            }
+            $display_name = $_SESSION[$sessKey];
+        }
+        else {
+            if (!$user->accessAdmin()) {
+                $display_name = $user->getDisplayedGivenName() . " " . substr($user->getDisplayedFamilyName(), 0, 1) . ".";
```
cjreed121: I think `getDisplayAbbreviatedName` in User.php might be what you are looking for

site/public/js/chatroom.js
```diff
@@ -0,0 +1,284 @@
...
+        <td>
+            <span class="display-short" title="${msg.description}">
+                ${msg.description.length > 45 ? `${msg.description.slice(0, 45)}...` : msg.description}
+            </span>
+        </td>
+        <td>
+            <a href="${msg.base_url}/${msg.id}" class="btn btn-primary">Join</a>
+            ${msg.allow_anon ? `<i> or </i><a href="${msg.base_url}/${msg.id}/anonymous" class="btn btn-default">Join As Anon.</a>` : ''}
+        </td>
+    `;
```
cjreed121: I placed the do not merge label because of this comment. Please do not self-resolve this comment and verify with me that it is properly addressed.

There is an XSS security vuln here. Need to do this functionality without using `.innerHTML`. Some good options are to manually create the elements in JS and use that to inject in (can be a bit of a pain) or to use something like Vue.js. See calendar.js as an example for the first one.

cjreed121 (changes requested):

site/public/js/chatroom.js
martig7: I just made a change to address this, lmk if this method of manually creating the elements is safe/unsafe.

site/app/repositories/chat/ChatroomRepository.php
martig7: I've removed the ChatroomRepository file.

site/app/entities/chat/Chatroom.php
martig7: I now pass in a UserEntity to the Chatroom object, and removed the host_name column from the chatroom table and its migration. I'm somewhat new to sql and doctrine so let me know if my implementation is bad.

site/app/controllers/ChatroomController.php
martig7: Done.

site/public/js/chatroom.js
```diff
@@ -0,0 +1,303 @@
...
+            }
+        },
+        error: function () {
+            window.alert('Something went wrong with storing message');
+        },
+    });
+}
+
+function appendMessage(displayName, role, ts, content) {
+    const timestamp = ts || new Date(Date.now()).toLocaleString('en-us', { year: 'numeric', month: 'short', day: 'numeric', hour: 'numeric', minute: 'numeric', second: 'numeric' });
```
cjreed121: Way too long of a line

cjreed121 (changes requested):

site/public/js/chatroom.js
cjreed121: Yes this should be safe. Thanks for addressing.

site/app/entities/chat/Chatroom.php
cjreed121: It appears to be fine. The one thing you might want to consider is eager fetching since this looks like it will make a second query to fetch the user info

site/app/entities/UserEntity.php
```diff
@@ -135,6 +135,7 @@ public function getDisplayInfo(): array {
         $out = [];
         $out["given_name"] = $this->user_preferred_givenname ?? $this->user_givenname;
         $out["family_name"] = $this->user_preferred_familyname ?? $this->user_familyname;
+        $out["full_name"] = $out["given_name"] . ' ' . $out["family_name"];
```
cjreed121: I would just recreate `getDisplayFullName()` from User.php and put it in this class.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,284 @@
...
+        $sessKey = "anon_name_chatroom_{$chatroom_id}";
+        $boolKey = "anon_name_chatroom_{$chatroom_id}_bool";
+        $msg_array = [];
+        $msg_array['content'] = $_POST['content'];
+        $msg_array['type'] = 'chat_message';
+        $msg_array['user_id'] = $user->getId();
+        $display_name = '';
+        if ($chatroom->isAllowAnon() && isset($_SESSION[$boolKey]) && $_SESSION[$boolKey] === true) {
+            $display_name = $_SESSION[$sessKey];
+        }
```
cjreed121: I would suggest looking for an alternative besides `$_SESSION`. One issue you are going to hit here is that these will conflict across different courses. That now could make someone more identifiable potentially.

imo `$_SESSION` should be used sparingly since it is also going to create a lot more local file IO and that data could be lost at any point.

martig7: I just made a lot of changes to address this, I think my approach is good but you should definitely check.

site/app/entities/chat/Chatroom.php
martig7: Alright, I've switched to eager fetching to prevent that second query.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,282 @@
...
+        if (!$chatroom->isActive() && !$user->accessAdmin()) {
+            return JsonResponse::getFailResponse("This chatroom is not enabled");
+        }
+        if (strcmp($_POST['content'], "") === 0) {
+            return JsonResponse::getFailResponse("Can't send blank message");
+        }
+        $msg_array = [];
+        $msg_array['content'] = $_POST['content'];
+        $msg_array['type'] = 'chat_message';
+        $msg_array['user_id'] = $user->getId();
```
cjreed121: If you are going to send `user_id`, that removes all form of anonymity as a user could just open up chrome dev tools and see the user id.

site/app/entities/chat/Chatroom.php
```diff
@@ -0,0 +1,132 @@
...
+
+    public function calcAnonName(string $user_id): string {
+        $adjectives = ["Quick","Lazy","Cheerful","Pensive","Mysterious","Bright","Sly","Brave","Calm","Eager","Fierce","Gentle","Jolly","Kind","Lively","Nice","Proud","Quiet","Rapid","Swift"];
+        $nouns      = ["Duck","Goose","Swan","Eagle","Parrot","Owl","Sparrow","Robin","Pigeon","Falcon","Hawk","Flamingo","Pelican","Seagull","Cardinal","Canary","Finch","Hummingbird"];
+        $session_started_at = $this->getSessionStartedAt() !== null ? $this->getSessionStartedAt()->format('Y-m-d H:i:s') : 'unknown';
+        $seed_string = $user_id . '-' . $this->getId() . '-' . $this->getHostId() . '-' . $session_started_at;
+        $adj_hash = crc32($seed_string);
+        $noun_hash = crc32(strrev($seed_string));
+        $adj_index = abs($adj_hash) % count($adjectives);
+        $noun_index = abs($noun_hash) % count($nouns);
```
cjreed121, line 119: This is very easy to reverse engineer. All components to this are identifiable or very easily guessable. Meaning given a list of user_id's, an external user (knowing this code), could easily calculate everyone's anonymous name.

cjreed121 (changes requested):

bmcutler (approved):
I tested this and I'm happy with the level of functionality.  We can add/modify features in later PRs.

In regards to the reverse engineering of the anonymous ids, we can improve that later.  It does have the start time in the hash so it isn't trivial to reverse engineer.  Related to that, when I stop and re-start the same chat room, because the timestamp changes, the anonymous ids will all change -- and that's weird.  So we probably want to modify it so at least for a single chat room the id stays the same.

bmcutler:
@martig7 

Actually, one more note.  Let's discuss the name for this feature on the sidebar.  Yes, I requested it for use in lecture in my course.  But I don't think the word "lecture" should be in the feature name.  Probably just "Live Chat".   We should update the database table names in addition to the terminology used on the page.

site/app/entities/UserEntity.php
```diff
@@ -120,6 +120,20 @@ public function getId(): string {
     #[ORM\ManyToMany(mappedBy: "upduckers", targetEntity: Post::class)]
     protected Collection $upducks;
 
+    public function getDisplayedGivenName(): string {
+        return $this->user_preferred_givenname ?? $this->user_givenname;
+    }
+    public function getDisplayedFamilyName(): string {
```
jeffrey-cordero, line 127: A space between the methods here would be nice.

site/app/controllers/ChatroomController.php
```diff
@@ -0,0 +1,282 @@
...
+            $msg_array['host_name'] = $chatroom->getHostName();
+            $msg_array['base_url'] = $this->core->buildCourseUrl(['chat']);
+            $msg_array['socket'] = "chatrooms";
+        }
+        else {
+            $msg_array = [];
+            $msg_array['type'] = 'chat_close';
+            $msg_array['id'] = $chatroom->getId();
+            $msg_array['socket'] = "chatrooms";
+            $indiv_msg_array = [];
```
jeffrey-cordero, line 209: Add a comment here for clarity in socket pages.

site/public/css/chatroom.css
```diff
@@ -0,0 +1,113 @@
...
+
+.message-input {
+    border-radius: 5px;
+    outline: none;
+    resize: none;
+    overflow: hidden;
+    font-size: 1rem;
+    flex: 1;
+    margin-left: 1rem;
+    margin-right: 1rem;
```
jeffrey-cordero: I would update the `margin-right` to `0.5rem` and remove the `margin-left` from the `.send-message-btn` to reduce the space between them.

image

site/public/css/chatroom.css
```diff
@@ -0,0 +1,113 @@
...
+    min-width: 0;
```
jeffrey-cordero, line 88: I would also add a `min-height:34px; padding-top:7px;` to ensure the heights are the same and the inner text is properly aligned.

image

site/public/js/chatroom.js
```diff
@@ -0,0 +1,305 @@
...
+            contentType: false,
+            success: function (data) {
+                try {
+                    const msg = JSON.parse(data);
+                    if (msg.status !== 'success') {
+                        console.error(msg);
+                        window.alert('Something went wrong. Please try again.');
+                    }
+                    else {
+                        window.location.reload();
```
jeffrey-cordero, line 172: This doesn't need to hold up the PR, but it would be a nice optimization to append the chatroom ID to each `<tr>` and delete the DOM element instead of reloading the page here.

site/public/css/chatroom.css
```diff
@@ -0,0 +1,114 @@
...
+    box-shadow: 0 -2px 10px 0 rgb(0 0 0 / 5%);
+    padding: 0.5rem;
+    flex-shrink: 0;
+    width: 100%;
+}
+
+.message-input {
+    border-radius: 5px;
+    outline: none;
+    resize: none;
```
jeffrey-cordero, line 81: I would remove the resize for now until Markdown is implemented in a future PR to handle large inputs.

site/public/js/chatroom.js
```diff
@@ -0,0 +1,306 @@
...
+            }
+        },
+        error: function () {
+            window.alert('Something went wrong with storing message');
+        },
+    });
+}
+
+function appendMessage(displayName, role, ts, content) {
+    const timestamp = ts || new Date(Date.now()).toLocaleString('en-us',
```
jeffrey-cordero: The output of the right-hand side timestamp differs from the `ts` input. Over the network, it's in the form `2025-06-23 08:16:04 (ts)`, but when you call it with the `Date` constructor, it becomes a localized string, where for me it's in the form `'Jun 23, 2025, 8:17:07 AM'`. I'd suggest removing the right-hand side here because `ts` should never be undefined, as it comes from the server.

site/public/css/chatroom.css
```diff
@@ -0,0 +1,114 @@
...
+    padding-top: 7px;
+    min-width: 0;
+}
+
+.message-input:focus {
+    background-color: var(--standard-hover-light-gray);
+}
+
+.send-message-btn {
+    margin-left: 1rem;
```
jeffrey-cordero: I'd suggest removing the left margin to remove the unnecessary gap.

image

image

jeffrey-cordero (approved):
The live lecture chat feature looks solid at this stage. While there may always be room for minor styling refinements, the current implementation is clean, functional, and complete. I'm confident it provides a strong foundation, and I'm looking forward to future pull requests that add end-to-end tests and enhance the styling and functionality of the feature.

PR #10235, opened 2024-03-07
[Testing:Developer] Cypress grade inquiries test.

### LINKED ISSUE #9339: Grade Inquiries   ✦✦   (0/4)

rajputanuj31:
Moving the selenium grade inquiries test to cypress.
Grade inquiries (1/4)
Closes #9339

# Summary
* extending the grade inquiry date so we can raise the inquiry
* Testing the submit grade inquiry button as it was firstly disable and then we type something in the markdown area then it will be enabled (covered that part also)
* tested same things for the ta and grader


IDzyre (changes requested):
See https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

We should ideally not be using any class or id attributes to select elements, as they may change if the UI is modified at all.

IDzyre:
> > See https://docs.cypress.io/guides/references/best-practices#Selecting-Elements
> > We should ideally not be using any class or id attributes to select elements, as they may change if the UI is modified at all.
> 
> To not use `id` or `classname` I need to add `data-*` attribut to html elements. Is it OK to add `data-*` to HTML elements?

Yes, you can and will have to add a new attribute to the html elements. They will only be used in testing so they shouldn't have any affect other than being able to access them in tests.

manzil-infinity180:
> See https://docs.cypress.io/guides/references/best-practices#Selecting-Elements
> 
> We should ideally not be using any class or id attributes to select elements, as they may change if the UI is modified at all.

@IDzyre sir using the data attribute is the best practice can we use this ? or we have to follow what the current project is using

site/cypress/e2e/Cypress-Gradeable/grade_inquiries.spec.js
```diff
@@ -0,0 +1,40 @@
...
+    };
+
+    it('should test normal submission grade inquiry panel', () => {
+        const gradeableId = 'grades_released_homework';
+        const gradeInquiryDeadlineDate = '9998-01-01 00:00:00';
+        setGradeInquiriesForGradeable(gradeableId, gradeInquiryDeadlineDate, true);
+
+        cy.get(`#${gradeableId} .btn-nav-grade`).click();
+        cy.get('[data-testid="view-sections"]').click();
+        cy.get(':nth-child(4) > :nth-child(1) > :nth-child(8) > [data-testid="grade-button"]').click();
```
IDzyre, line 35: Is there a way to select this without chaining selectors?

IDzyre (changes requested):

site/cypress/e2e/Cypress-Gradeable/grade_inquiries.spec.js
manzil-infinity180: > Is there a way to select this without chaining selectors?

this could be possible case if we are clicking on first grade-buttton and i tried it locally it works fine !

```js
cy.get('[data-testid="grade-button"]').first().click();
```

Screenshot 2024-04-04 at 8 34 57 PM

Screenshot 2024-04-04 at 8 34 46 PM

rajputanuj31:
Hi @IDzyre Sorry for the late response, actually I fell ill.
There are two ways to select this: `cy.get(':nth-child(4) > :nth-child(1) > :nth-child(8) > [data-testid='grade-button']') . click();` 
First as suggested by @manzil-infinity180 and second is selecting a specific student as is done in Selenium tests

`cy.get('[data-testid='grade-table']').contains('td', 'Farrell').nextAll('td').find('[data-testid='grade-button "]').Click()
`
Can you please suggest which one should I use?

IDzyre:
> Hi @IDzyre Sorry for the late response, actually I fell ill. There are two ways to select this: `cy.get(':nth-child(4) > :nth-child(1) > :nth-child(8) > [data-testid='grade-button']') . click();` First as suggested by @manzil-infinity180 and second is selecting a specific student as is done in Selenium tests
> 
> `cy.get('[data-testid='grade-table']').contains('td', 'Farrell').nextAll('td').find('[data-testid='grade-button "]').Click() ` Can you please suggest which one should I use?

The first one chains off of many different elements, the second one is better.

site/cypress/e2e/Cypress-Gradeable/grade_inquiries.spec.js
```diff
@@ -0,0 +1,40 @@
...
+        cy.get(`#${gradeableId} .fa-pencil-alt`).click();
+
+        if (allowed) {
+            cy.get('[data-testid="yes-grade-inquiry-allowed"]').click();
+        }
+        else {
+            cy.get('[data-testid="no_grade_inquiry_allowed"]').click();
+        }
+
+        if (date) {
```
IDzyre, line 16: Is there a reason for the branching if this function is only being called once?

IDzyre (changes requested):

site/cypress/e2e/Cypress-Gradeable/grade_inquiries.spec.js
rajputanuj31: Actually I am taking reference from Selenium Tests, and since it uses branching, I have used it here. If it is unnecessary I will change it.

site/cypress/e2e/Cypress-Gradeable/grade_inquiries.spec.js
IDzyre, line 1: This may be better to put in Cypress-Feature

IDzyre (changes requested):
Quick comment, I will test later today.

site/app/templates/admin/admin_gradeable/AdminGradeableDates.twig
```diff
@@ -76,7 +76,7 @@
 
             <br />
 
-            <input data-testid="grade_inquiry_due_date" name="grade_inquiry_due_date" id="date_grade_inquiry_due" class="date_picker grade_inquiry_due_date" type="text" value="{{ gradeable.getGradeInquiryDueDate()|date(date_format, timezone_string) }}">
+            <input data-testid="date-grade-inquiry-due" name="grade_inquiry_due_date" id="date_grade_inquiry_due" class="date_picker grade_inquiry_due_date" type="text" value="{{ gradeable.getGradeInquiryDueDate()|date(date_format, timezone_string) }}">
```
IDzyre, line 79:
```suggestion
            <input data-testid="grade-inquiry-due-date" name="grade_inquiry_due_date" id="date_grade_inquiry_due" class="date_picker grade_inquiry_due_date" type="text" value="{{ gradeable.getGradeInquiryDueDate()|date(date_format, timezone_string) }}">
``` 
This is more of a consistency suggestion between the other date testids, but I can be persuaded to use different testids if they are all consistent.

site/cypress/e2e/Cypress-Feature/grade_inquiries.spec.js
```diff
@@ -0,0 +1,47 @@
...
+            cy.get('[data-testid="close-button"]').should('exist');
+            cy.get('[data-testid="close-hidden-button"]').should('exist');
+            cy.get('[data-testid="agree-popup-btn"]').click();
+            if ( user === 'ta') {
+                cy.get('[data-testid="view-sections"]').click();
+            }
+            cy.get('[data-testid="grade-button"]').eq(3).click();
+            cy.get('[data-testid="grade-inquiry-info-btn"]').click();
+            cy.get('[data-testid="grading-label"]').should('contain', 'Grade Inquiry');
+            cy.get('[data-testid="grade-inquiry-actions"]').contains('Submit Grade Inquiry').and('be.disabled');
```
IDzyre, line 41:
```suggestion
            cy.get('[data-testid="grade-inquiry-actions"]').should('contain', 'Submit Grade Inquiry').and('be.disabled');
```

Apply for other assertions of contains

IDzyre (changes requested):

site/cypress/e2e/Cypress-Feature/grade_inquiries.spec.js
manzil-infinity180: here i am searching for it , previously i have done in that manner with some other data-testid but it failing for that so i change (what i think that element is disable so select just one top level then use `.contains `to search and then verify it ) but its failing again 

should i change to that one ??

IDzyre: If you are getting the action that contains 'submit grade inquiry', that is okay, but then change the assertion to ```contains(xxx).should('be.disabled')```

IDzyre:
The tests fail on CI because the gradeable does not have grade inquiries available at the time of the test, they become available at the end of time. The gradeable needs to be updated to change the dates.

manzil-infinity180:
Reason behind cypress failing 
- Like when we type the date and directly goes to some other page it does not save that date So the submit grade inquiry button does not exist there 
- Added to press enter for saving the date

manzil-infinity180:
> The tests fail on CI because the gradeable does not have grade inquiries available at the time of the test, they become available at the end of time. The gradeable needs to be updated to change the dates.
Fixed ✅
Reason behind cypress failing 
- Like when we type the date and directly goes to some other page it does not save that date So the submit grade inquiry button does not exist there 
- Added to press enter for saving the date

IDzyre (approved):

PR #10236, opened 2024-03-07
[Refactor:TAGrading] Use of data- selector instead of class's

### LINKED ISSUE #10015: Refactor simple grading cypress test to use better selectors (that aren't tied to classes/UI elements)
The simple_grading.spec.js file has a few lines that are extremely long, using selectors that are tied to classes. 

These lines should be rewritten to use a `data-` selector instead of long class selectors.

GO-viper7:
Closes #10015 

* [✅] Screenshots are attached to Github PR

In the file `simple_grading.spec.js`, extremely long selectors are tied using multiple classes.

### What is the new behavior?
Usage of `data-` selector instead of class selectors.
This in turn improves testing stability.

Upon running `simple_grading.spec.js` in Cypress UI : 

refactor-simple-grading

### DIFF
--- a/site/app/templates/generic/Popup.twig
+++ b/site/app/templates/generic/Popup.twig
@@ -10,7 +10,7 @@
                         {% block title_tag %}
                             <h1>{% block title %}Untitled Form{% endblock %}</h1>
                             <button
-                                data-close="close-button"
+                                data-testid="close-button"
                                 onclick="{{ block('close_click_action') }}"
                                 class="btn btn-default close-button key_to_click"
                                 tabindex="-1" type="button"
--- a/site/app/templates/generic/Popup.twig
+++ b/site/app/templates/generic/Popup.twig
@@ -10,6 +10,7 @@
                         {% block title_tag %}
                             <h1>{% block title %}Untitled Form{% endblock %}</h1>
                             <button
+                                data-close="close-button"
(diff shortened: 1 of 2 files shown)


GO-viper7:
I wonder what error this is
@IDzyre can you please go through it
image

site/app/templates/generic/Popup.twig
```diff
@@ -10,6 +10,7 @@
                         {% block title_tag %}
                             <h1>{% block title %}Untitled Form{% endblock %}</h1>
                             <button
+                                data-close="close-button"
```
IDzyre, line 13: See https://docs.cypress.io/guides/references/best-practices#Selecting-Elements
For best use with Cypress, we should use `data-cy, data-test, or data-testid`. I am partial to `data-testid` because it is more descriptive to what it is.

GO-viper7: alright, will replace `data-close` with `data-testid`

iamdanielyun (approved):

GO-viper7:
Hey @IDzyre 
Can you please look at the failed check or run the tests again
I don't understand what's causing it.
Thank you.

ArnavPawar (approved):
Fixes the complexity issue. Looks good and works just some simple js class changes.

PR #10245, opened 2024-03-13
[Refactor:System] Switch PDO to DBAL

cjreed121:

We currently wrap our AbstractDatabase around PDO for both Postgres and SQLite. We should also be upgrading DBAL to 3 (and eventually 4 since that was just released) which has been an open PR for a while. Doctrine ORM requires DBAL because that is what it uses as it's database layer. We pass in our existing PDO instance to ORM/DBAL to avoid creating a new database connection.

We are using DBAL 2.13.9.

### What is the new behavior?
This PR changes AbstractDatabase to now wrap DBAL instead of PDO. DBAL wraps PDO so we essentially just have an extra layer in between AbstractDatabase and PDO. If we were to keep AbstractDatabase using PDO, we would need to double our database connections to support DBAL 3 and higher. This is because DBAL removed the ability to pass in a user provided PDO instance. So we either would need to double the database connections or switch our connection. The idea is to maintain just 1 DBAL connection for each the master and course DBs and to just pass that into ORM now.

We are now using DBAL 3.8.2.

### Other information?
As little has been changed to replace PDO to DBAL in AbstractDatabase.

### DIFF
--- a/site/.performance_warning_ignore.json
+++ b/site/.performance_warning_ignore.json
@@ -44,5 +44,6 @@
   "/courses/<term>/<course>/polls/newPoll",
   "/courses/<term>/<course>/polls/editPoll/submitEdits",
   "/courses/<term>/<course>/polls/deletePoll",
-  "/courses/<term>/<course>/upload"
+  "/courses/<term>/<course>/upload",
+  "/courses/<term>/<course>/email_status_page"
 ]
--- a/site/app/libraries/Core.php
+++ b/site/app/libraries/Core.php
@@ -8,13 +8,10 @@ use app\exceptions\CurlException;
 use app\libraries\database\DatabaseFactory;
 use app\libraries\database\AbstractDatabase;
 use app\libraries\database\DatabaseQueries;
-use app\libraries\database\DatabaseUtils;
 use app\models\Config;
 use app\models\User;
 use app\entities\Session;
 use app\repositories\SessionRepository;
-use Doctrine\DBAL\DriverManager;
-use Doctrine\DBAL\Logging\DebugStack;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\ORMSetup;
 use Symfony\Component\Cache\Adapter\FilesystemAdapter;
@@ -45,14 +42,8 @@ class Core {
     /** @var EntityManager */
     private $submitty_entity_manager;
 
-    /** @var DebugStack */
-    private $submitty_debug_stack;
-
     /** @var EntityManager */
     private $course_entity_manager;
-
-    /** @var DebugStack */
-    private $course_debug_stack;
 
     /** @var AbstractAuthentication */
     private $authentication;
@@ -182,7 +173,7 @@ class Core {
         $this->session_manager = $manager;
     }
 
-    private function createEntityManager(AbstractDatabase $database, ?DebugStack $debug_stack): EntityManager {
+    private function createEntityManager(AbstractDatabase $database): EntityManager {
         $cache_path = FileUtils::joinPaths(dirname(__DIR__, 2), 'cache', 'doctrine');
         $cache = new PhpFilesAdapter("", 0, $cache_path);
         $config = ORMSetup::createAttributeMetadataConfiguration(
@@ -192,17 +183,7 @@ class Core {
             $cache
         );
 
-        if ($debug_stack) {
-            $config->setSQLLogger($debug_stack);
-        }
-
-        $conn = DriverManager::getConnection(
-            [
-                'driver' => 'pdo_pgsql',
-                'pdo' => $database->getConnection()
-            ]
-        );
-        return new EntityManager($conn, $config);
+        return new EntityManager($database->getConnection(), $config);
     }
 
     /**
@@ -224,8 +205,7 @@ class Core {
         $this->submitty_db->connect();
 
         $this->setQueries($this->database_factory->getQueries($this));
-        $this->submitty_debug_stack = $this->config->isDebug() ? new DebugStack() : null;
-        $this->submitty_entity_manager = $this->createEntityManager($this->submitty_db, $this->submitty_debug_stack);
+        $this->submitty_entity_manager = $this->createEntityManager($this->submitty_db);
     }
 
     public function setMasterDatabase(AbstractDatabase $database): void {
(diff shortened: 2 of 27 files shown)


site/app/libraries/Core.php
```diff
@@ -291,19 +262,7 @@ public function hasDBPerformanceWarning(): bool {
...
-                $queries[] = $query['sql'];
-            }
-        }
-        if ($this->submitty_debug_stack !== null) {
-            foreach ($this->submitty_debug_stack->queries as $query) {
-                $queries[] = $query['sql'];
-            }
-        }
-
-        return count($queries) !== count(array_unique($queries));
```
MasterOdin, line 306: We don't need this check anymore as the `hasDuplicateCheck()` that's run against the `$this->course_db` and `$this->submitty_db` cover what this was checking anyway, right?

site/app/libraries/database/QueryLogger.php
```diff
@@ -0,0 +1,40 @@
...
+     * @param $level
+     * @param \Stringable|string $message
+     * @param array<mixed> $context
+     * @return void
+     */
+    public function log($level, \Stringable|string $message, array $context = []): void {
+        if ($level === "debug") { // SQL queries are in the debug level
+            if (!isset($context['sql'])) {
+                return; // Ignore if SQL is not set
+            }
```
MasterOdin, line 26:
```suggestion
        // SQL queries are in the debug level, but we want to ignore any other debug log
        if ($level === "debug" && isset($context['sql'])) {
```

site/tests/app/controllers/AuthenticationControllerTester.php
```diff
@@ -13,6 +13,7 @@
 
 /**
  * @runTestsInSeparateProcesses
+ * @preserveGlobalState disabled
```
MasterOdin, line 16: What about these changes is causing something within global state to not be serializable? Is there anyway to avoid that?

MasterOdin:
Seems like there's a number of pages that now throw an exception from within cypress?

site/app/libraries/Core.php
cjreed121: Yeah these were strictly for the ORM queries that were issued. So the `hasDuplicateCheck()` will now cover both.

site/tests/app/controllers/AuthenticationControllerTester.php
cjreed121: William and I both spent a couple hours trying to figure out what was going on here. Bumping DBAL is what is causing this issue. They added a closure as a member variable of their [Configuration object](https://github.com/doctrine/dbal/blob/cbc2e57cf9806072c022f4e537b0d9c26dc0a254/src/Configuration.php#L39-L44). Somehow this gets saved into some global state we store and when moving to a separate process, the serialization fails since PHP can't serialize closures. The only 2 feasible options I have found is removing the testing in separate processes (wasn't exactly sure why we are doing this besides memory/isolation reasons?) or by adding this preserve global state option. If you have any other ideas here, please let me know and I'll try them.

cjreed121:
> Seems like there's a number of pages that now throw an exception from within cypress?

Yeah I still need to fix those. Just looking for any initial feedback on the changes since there is some design decisions here. Thanks for the comments so far.

site/app/libraries/database/QueryLogger.php
```diff
@@ -0,0 +1,38 @@
...
+     */
+    protected array $queries = [];
+
+    /**
+     * @param $level
+     * @param \Stringable|string $message
+     * @param array<mixed> $context
+     * @return void
+     */
+    public function log($level, \Stringable|string $message, array $context = []): void {
```
williamjallen, line 22: Is `$level` guaranteed to be a string here?

williamjallen (changes requested):

site/.performance_warning_ignore.json
```diff
@@ -44,5 +44,6 @@
   "/courses/<term>/<course>/polls/newPoll",
   "/courses/<term>/<course>/polls/editPoll/submitEdits",
   "/courses/<term>/<course>/polls/deletePoll",
-  "/courses/<term>/<course>/upload"
+  "/courses/<term>/<course>/upload",
+  "/courses/<term>/<course>/email_status_page"
```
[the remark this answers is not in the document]
cjreed121: See my change in EmailRepository.php. It wasn't ever logging before.

site/app/libraries/database/AbstractDatabase.php
```diff
@@ -62,13 +53,16 @@ public function __construct($connection_params) {
         }
     }
 
-    abstract public function getDSN();
+    /**
+     * @psalm-return Params
+     */
+    abstract public function getConnectionDetails(): array;
```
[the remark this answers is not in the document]
williamjallen: Resolved offline.

site/app/libraries/database/AbstractDatabase.php
```diff
@@ -158,31 +144,26 @@ public function query($query, $parameters = []) {
...
             }
         }
-        catch (\PDOException $pdoException) {
-            throw new DatabaseException($pdoException->getMessage(), $query, $parameters);
+        catch (DBALException $dbalException) {
+            throw new DatabaseException($dbalException->getMessage(), $query, $parameters);
         }
 
-        return $result;
+        return true;
```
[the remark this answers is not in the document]
cjreed121: Based on the code before, it appears like it would never return false due to it having the ERRMODE_EXCEPTION mode on. This means the PDO execute won't ever return false and will just throw an exception instead. I have only seen 2 places in the codebase that use this return value and they use it as bools. Maybe we should remove the return value entirely since if it fails, it will throw an exception?

site/app/libraries/database/AbstractDatabase.php
```diff
@@ -87,58 +81,51 @@ abstract public function fromPHPToDatabaseArray($array);
...
+     * This will additionally commit any open transactions before disconnecting.
      */
     public function disconnect() {
         if ($this->transaction) {
             $this->rollback();
         }
-        $this->link = null;
+        $this->conn->close();
+        $this->conn = null;
     }
```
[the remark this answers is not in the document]
cjreed121: I guess this comment was just wrong before. I didn't look at that close enough. I'm honestly not sure what the correct functionality should be here. We have been sticking with rollback on disconnects so there is a chance we change the behavior of something if we switch the rollback. I'm thinking of just updating the comment on this one. Then if any application code needs to commit before closing then they are responsible for that. Thoughts?

site/app/libraries/database/AbstractDatabase.php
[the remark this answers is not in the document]
williamjallen: Discussed in person.  Changes still necessary.

williamjallen (approved):

PR #10252, opened 2024-03-14
[Feature:InstructorUI] Add dates to upload

IDzyre:

#10100 added the ability to upload a JSON file to create a gradeable. This was added without the ability to use preset dates, and required you to edit the dates manually in the UI after the gradeable was created. 
### What is the new behavior?
Now you can add dates, and date-related fields, (has_due_date, has_release_date, late_days_allowed, late_days(amount)) to the JSON file and they will be added.


cjreed121:
@IDzyre can you please add details to this PR for what it is addressing

jeffrey-asm (approved):
The added changes and tests look great. Instructors can now add dates, including late days, to the upload gradable JSON file, and the date verification can be done as usual using the create gradable method for verification, catching any invalid date setup.  This PR has nothing to do with negative late days, but it would be nice to eventually fix the negative late-day input on the gradable form.

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -143,6 +160,9 @@ public function uploadGradeable() {
             return JsonResponse::getSuccessResponse($_POST['id']);
         }
         catch (\Exception $e) {
+            if ($e->getMessage() === 'Date validation failed') {
+                return JsonResponse::getErrorResponse('Date validation has failed, ensure the dates are in the correct sequential order');
```
cjreed121, line 164: You could catch this as a ValidationException and then return the actual Date error(s) which exists to provide more info of which Date is wrong

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1084,15 +1103,15 @@ private function createGradeable($gradeable_id, $details) {
...
+            'team_lock_date' => isset($details['team_lock_date']) ? new \DateTime($details['team_lock_date'], $this->core->getConfig()->getTimezone()) : (clone $tonight)->add(new \DateInterval('P7D')),
+            'submission_open_date' => isset($details['submission_open_date']) ? new \DateTime($details['submission_open_date'], $this->core->getConfig()->getTimezone()) : (clone $tonight),
+            'submission_due_date' => isset($details['submission_due_date']) ? new \DateTime($details['submission_due_date'], $this->core->getConfig()->getTimezone()) : (clone $tonight)->add(new \DateInterval('P7D')),
+            'grade_inquiry_start_date' => isset($details['grade_inquiry_start_date']) ? new \DateTime($details['grade_inquiry_start_date'], $this->core->getConfig()->getTimezone()) : (clone $tonight)->add(new \DateInterval('P15D')),
+            'grade_inquiry_due_date' =>  isset($details['grade_inquiry_due_date']) ? new \DateTime($details['grade_inquiry_due_date'], $this->core->getConfig()->getTimezone()) : (clone $tonight)->add(new \DateInterval('P21D'))
```
cjreed121, line 1114: Could you break up the long lines

cjreed121 (changes requested):
Looks good overall. Just some small cleanup and a suggestion (if you want to do that here or could be done in a future PR).

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -128,11 +128,28 @@ public function uploadGradeable() {
         if (array_key_exists('ta_grading', $_POST)) {
             $values['ta_grading'] = $_POST['ta_grading'];
             if (array_key_exists('grade_inquiries', $_POST)) {
-                $values['grade_inquiry_allowed'] = $_POST['grade_inquiries'];
+                $values['grade_inquiry_allowed'] = $_POST['grade_inquiries'] ?? 'false';
```
williamjallen, line 131: Is there a reason why the boolean value here (and elsewhere in this file) is a string?

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1083,16 +1102,71 @@ private function createGradeable($gradeable_id, $details) {
         if ($tonight->diff($this->core->getDateTimeNow())->h < 12) {
             $tonight->add(new \DateInterval('P1D'));
         }
+
+        $ta_view_start_date = (
+            isset($details['ta_view_start_date'])
+                ? new \DateTime($details['ta_view_start_date'], $this->core->getConfig()->getTimezone())
+                : (clone $tonight)
+        );
```
williamjallen, line 1110: Very similar logic is being repeated here.  Would it be better to simplify it, either by creating a function to do it, or by setting array key-value pairs of an array in a loop?  Regarding the actual construction of new DateTime objects, I wonder if it would make sense to create a method in the Core library class to do this, similar to the `getDateTimeNow()` method?

https://github.com/Submitty/Submitty/blob/b289ce22f17b429fc8f8e32761e97b60981404ec/site/app/libraries/Core.php#L769-L771

site/app/controllers/admin/AdminGradeableController.php
IDzyre: I believe this is an instance of getting values passed from html elements to php code, and never being converted to boolean. 

at first I tried using booleans, but it would fail later on in the process.

williamjallen (approved):
Code looks good to me.

cjreed121 (approved):
Code looks good

PR #10253, opened 2024-03-14
[Feature:SubminiPolls] Add WebSocket to Polls

jeffrey-cordero:

Currently, the histogram of the poll page on the instructor side does not update dynamically. Instructors must refresh the page to get the most up-to-date poll information to be reflected on the histogram. Also, any updates on the current open poll will not be reflected on the student interface unless stated directly by the instructor.

### What is the new behavior?
Instructors can now see updates reflected on the histogram as students submit their poll responses. On student submission, a WebSocket message will be sent only containing updates for the most recently changed option response counts. Students will also receive WebSocket messages when a poll has closed, re-opened, or has made significant changes.
 
### Other information?
This is a small change to the polls feature of Submitty. I tested by submitting various changes in polls side by side as instructor and student. As a student submits a change in a given poll, the histogram would update on the instructor side in constant time, reflecting current response counts. I also made various changes to the poll while a student was connected as shown in the short demonstration video below. This feature works toward completing [#5913](https://github.com/Submitty/Submitty/issues/5913).

https://github.com/Submitty/Submitty/assets/94331560/b154b6e7-592d-4301-8079-4404283acbdc


ZheyuDeng (commented):
I tested the case where a student edit his answer and submit again. The histogram will include both of his submission unless when the page is refreshed it will removes the first one. 
After second submission where student removes b and submit again, the histogram shows both submission as you can see the response goes to 2.
71666907f718c2a25571f2a77e17bdd

ZheyuDeng (approved):
Tested and worked for me!

cjreed121:
This is on hold of being merged until https://github.com/Submitty/Submitty/issues/10296 is closed

williamjallen:
@martig7 This would be a great PR to review if you're interested in some of the web socket stuff.

martig7 (approved):
This works well, I was able to test two people submitting a poll at the same time and it worked as expected. We should think about how we could test websockets using cypress.

site/app/controllers/PollController.php
```diff
@@ -482,7 +484,8 @@ public function submitEdits(): RedirectResponse {
         }
 
         $em->flush();
-
+        $web_socket_message = ['type' => 'poll_updated', 'poll_id' => $poll_id, 'socket' => 'student', 'message' => 'Poll updated'];
```
williamjallen, line 487: It's generally more readable to wrap large associative arrays.  An additional benefit of doing so is that the diff will be cleaner if someone edits the array in the future--the same reason why a trailing comma is useful.  The same applies below as well.

```suggestion
        $web_socket_message = [
            'type' => 'poll_updated',
            'poll_id' => $poll_id,
            'socket' => 'student',
            'message' => 'Poll updated',
        ];
```

site/app/templates/polls/ViewPoll.twig
```diff
@@ -107,6 +107,48 @@
                 </fieldset>
                 {% if not user_admin and poll.isOpen() %}
                     <button type="submit" data-testid="submit-answer" class="btn btn-primary student-submit">Submit</button>
+                    <script>
```
williamjallen, line 110: This logic should be moved to `site/public/js/polls.js` so it gets linted.

site/app/templates/polls/ViewPoll.twig
```diff
@@ -107,6 +107,48 @@
...
+                        function initializeStudentSocketClient() {
+                            window.socketClient = new WebSocketClient();
+
+                            window.socketClient.onmessage = (msg) => {
+                                const submit_button = document.querySelector('.student-submit');
+                                const custom_response_submit = document.querySelector('.custom-response-submit');
```
williamjallen, line 116: I'm torn regarding whether using jQuery for this would be cleaner.  We usually use jQuery whenever possible, but this code seems generally ok to me.  Thoughts?

site/app/templates/polls/ViewPoll.twig
```diff
@@ -165,10 +207,15 @@
...
+
+            // Store the position of each response for incoming websocket updates (-1, 0, 1)
+            const responseIndexMap = {};
+            let position = 0;
             const [correctColor, defaultColor] =
             [window.getComputedStyle(document.body).getPropertyValue('--histogram-correct'),
                 window.getComputedStyle(document.body).getPropertyValue('--histogram-default')];
             {% for option in poll.getOptions() %}
+                responseIndexMap["{{ option.getResponse()|replace({"\n": " ", "\r": " ", "\t": " "})|e('js') }}"] = position++;
                 data[0].x.push("{{ option.getResponse()|replace({"\n": " ", "\r": " ", "\t": " "})|e('js') }}");
```
williamjallen, line 219: Let's not duplicate code here and instead assign the value of the twig expression to a constant.

site/app/templates/polls/ViewPoll.twig
```diff
@@ -196,6 +243,39 @@
...
+                            updateHistogram(msg.message);
+                            break;
+                        default:
+                            console.error("Unknown web socket message received:", msg);
+                            break;
+                    }
+                };
+
+                window.socketClient.open('polls-{{ poll.getId() }}-instructor');
+            }
```
williamjallen, line 275: Let's move these functions to `site/public/js/polls.js` so they get linted.  The rest of the code can probably stay for now, although it all needs a bit of a refactor eventually.

williamjallen (changes requested):
Looks pretty reasonable overall.  A few code comments to address and then this should be good to go.

site/app/templates/polls/ViewPoll.twig
jeffrey-cordero: The `polls.js` almost entirely uses JQuery methods compared to my initial implementation in Twig. We should focus more on standard JS methods on new work, but file consistency is more important here.

martig7 (approved):
I tested it in the same way again and it still worked, I can't comment on the code form changes since I lack experience in that aspect.

williamjallen (approved):
Code looks reasonable to me.

PR #10259, opened 2024-03-15
[Feature:Submission] Select Preferred Name Order on Profile Page

04vmatsibekker:

In the database, we have a display_name_order column, but couldn't update the value in the database.

### What is the new behavior?

Can now select/specify the value of the preferred name order variable on the profile page
image

### Other information?
NOTE: This selection is not actually used on any pages in Submitty.  That will be done in future PRs following requests in issue #8446


site/app/templates/EditDisplayNameOrderForm.twig
```diff
@@ -0,0 +1,32 @@
...
+{% block title %}Edit Display Name Order{% endblock %}
+{% block body %}
+    <p class="std-margin">{{ "You can edit the order in which your name is displayed here!" }}</p>
+    <div class="flex-col" id="user_settings_modal">
+        <input id="display-name-order-csrf" type="hidden" name="csrf_token" value="{{ csrf_token }}" />
+        <h2>My Pronouns</h2>
+        <div class="edit_user_row ui-front">
+    <select name="display-name-order-change" id="display-name-order-change">
+        {% if user.getDisplayNameOrder() != 'REVERSED_F' %}
+        <option value="GIVEN_F">GIVEN_F</option>
```
jeffrey-asm, line 12: The option text should be more readable for the general audience like "Reversed" and "Given" instead of "REVERSED_F" and "GIVEN_F"

site/public/js/user-profile.js
```diff
@@ -150,6 +161,58 @@ function updateUserPronouns(e) {
...
+    // eslint-disable-next-line no-undef
+    const url = buildUrl(['user_profile', 'change_display_name_order']);
+    $.ajax({
+        url,
+        type: 'POST',
+        data,
+        processData: false,
+        contentType: false,
+        success: function(res) {
+            console.log(res);
```
jeffrey-asm, line 186: It would be a good idea to remove the console logs for the finalized method

site/app/controllers/UserProfileController.php
```diff
@@ -145,6 +145,29 @@ public function changePronouns() {
...
     }
 
+    /**
+     * @Route("/user_profile/change_display_name_order", methods={"POST"})
+     * @return JsonResponse
+     */
+    public function changeDisplayNameOrder() {
+        $user = $this->core->getUser();
+        if (isset($_POST['display-name-order'])) {
+            $newPronouns = trim($_POST['display-name-order']);
```
jeffrey-asm, line 155: newPronouns being set to display-name-order is a bit confusing given the form is described as "You can edit the order in which your name is displayed here!"

site/app/controllers/UserProfileController.php
```diff
@@ -145,6 +145,29 @@ public function changePronouns() {
...
+            //validPronouns() checks for valid option
```
jeffrey-asm, line 156: The comments here can also go unless they are essential for a valid display order, which should be validated here in the backend.

jeffrey-asm (changes requested):
Overall most of the implementation looks good as the display order of the user is updated properly. I'm still a bit confused as to what the display order is used for as I saw no differences in the forum or viewing student name orders as instructors. I would look into some of the comments I left behind for further clarity on the fix.

micpap25:
Wrapping up this PR today; we are going to roll out actually implementing this change in various locations (office hours queue, forum, etc.) as well as having it interact well with abbreviations separately.

micpap25:
Linking #8446

site/app/templates/EditDisplayNameOrderForm.twig
```diff
@@ -0,0 +1,32 @@
...
+    <select name="display-name-order-change" id="display-name-order-change">
+        {% if user.getDisplayNameOrder() != 'REVERSED_F' %}
+        <option value="GIVEN_F">Given Name First</option>
+        <option value="REVERSED_F">Family Name First</option>
+        {% endif %}
+        {% if user.getDisplayNameOrder() != 'GIVEN_F' %}
+        <option value="REVERSED_F">Family Name First</option>
+        <option value="GIVEN_F">Given Name First</option>
+        {% endif %}
+    </select>
```
IDzyre, line 19: Option values have a 'selected' parameter that you can add, like `<option value="GIVEN_F" {user.getDisplayNameOrder() == 'GIVEN_F' ? selected : ''>Family Name First</option>`, so the order can be consistent instead of switching the order depending on what is selected. 

Also, these variable names aren't intuitive, since from the displayed values, I think that `GIVEN_FIRST` and `FAMILY_FIRST` or something similar make more sense (if eventually we want to add reversing, so given name last, `REVERSED` would be confusing, as it isn't reversed.

IDzyre (changes requested):
Some syntax changes for PHP8, use native php annotations instead of PHPdoc annotations, confusing variable names, and use selected for optin

site/app/controllers/UserProfileController.php
```diff
@@ -144,6 +144,28 @@ public function changePronouns() {
     }
 
     /**
+     * @Route("/user_profile/change_display_name_order", methods={"POST"})
```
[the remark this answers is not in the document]
williamschen23: There's also another annotation for change_preferred_names

site/app/templates/EditDisplayNameOrderForm.twig
micpap25: Can't change `GIVEN_F` to anything else for database reasons, so changed `REVERSED_F` to `FAMILY_F`.

site/app/controllers/UserProfileController.php
```diff
@@ -144,6 +144,28 @@ public function changePronouns() {
...
+     * @return JsonResponse
+     */
+    public function changeDisplayNameOrder() {
```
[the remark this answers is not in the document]
micpap25: Made this change to all functions in the file.

site/app/templates/EditDisplayNameOrderForm.twig
```diff
@@ -0,0 +1,26 @@
+{% extends 'generic/Popup.twig' %}
+{% block popup_id %}edit-display-name-order-form{% endblock %}
+{% block title %}Edit Display Name Order{% endblock %}
+{% block body %}
+    <p class="std-margin">{{ "You can edit the order in which your name is displayed here!" }}</p>
+    <div class="flex-col" id="user_settings_modal">
+        <input id="display-name-order-csrf" type="hidden" name="csrf_token" value="{{ csrf_token }}" />
+        <h2>My Pronouns</h2>
```
dorian451, line 8: I'm assuming you forgot to change this

site/app/controllers/UserProfileController.php
```diff
@@ -143,11 +143,32 @@ public function changePronouns() {
         }
     }
 
+    /**
+     * @return JsonResponse
+     */
```
dorian451, line 148: If you're putting the return type in the function declaration I don't think you need it in the phpdoc as well
```suggestion
```

dorian451 (changes requested):
Looks good, other than a few small things. I've also noticed that changing the preferred name order doesn't seem to do anything, I'm assuming that will be in a separate PR?

site/app/controllers/UserProfileController.php
micpap25: @IDzyre thoughts?

IDzyre: I believe if you use the `function(): JsonResponse {}` notation, you don't need to have the phpdoc declaration also, (unless it is an array and you want to specify the variables inside the array) but @williamjallen would know more.

site/public/js/user-profile.js
```diff
@@ -149,6 +160,51 @@ function updateUserPronouns(e) {
...
+
+    const data = new FormData();
+    // eslint-disable-next-line no-undef
+    data.append('csrf_token', csrfToken);
+    data.append('display-name-order', displayNameOrder.val());
+    const url = buildUrl(['user_profile', 'change_display_name_order']);
+    $.ajax({
+        url,
+        type: 'POST',
+        data,
```
ziesski, line 178: This is very minor request, style-wise lets keep it consistent with rest of the codebase
```suggestion
        url: url,
        type: 'POST',
        data: data,
```

ziesski (approved):
Overall looks good, I tested and the popup is available on the profile page. I agree with one of the above review that this PR doesn't really change anything in anywhere (discussion forum etc) but it is a good starting point.

site/app/models/User.php
```diff
@@ -235,6 +240,13 @@ public function __construct(Core $core, $details = []) {
 
         $this->setPronouns($details['user_pronouns']);
 
+        if (isset($details['display_name_order'])) {
+            $this->setDisplayNameOrder($details['display_name_order']);
+        }
+        else {
+            $this->setDisplayNameOrder("GIVEN_F");
+        }
```
IDzyre, line 248: Is this needed, given that it is already set to GIVEN_F by default?

micpap25:
@bmcutler Should this be merged in accordance with our discussion on #8446? I would then like to have a discussion and make an issue regarding removing "first" and "last" name from Submitty; in particular how we should approach sorting by first name or by last name. (should this become sorting by family / given name? etc.)

bmcutler:
@micpap25 
"first" should be replaced with "given"
and "last" should be replaced with "family"
everywhere on the UI & within the code & DB.
That's just a cosmetic change (shouldn't impact columns or sorting).  
... but because it likely changes lots of lines of code and lots of files and could have merge conflicts, we may indeed want to do that in stages over multiple PRs.


Actually using the preferred name order where appropriate, requires thought and design.
And some revision to sorting / display of sorted names.  etc.

ziesski (approved):
Tested it works

dorian451 (commented):
seems to work

PR #10260, opened 2024-03-15
[Testing:Forum] Implement Image Diff For Markdown

04vmatsibekker:

Can't test markdown.

### What is the new behavior?

Implemented Image Diff so we can take a screenshot of the forum page and then test it looks correct


manzil-infinity180:
Removed the visual testing for autograding section (markdown_test.spec.js) 
I will make test that in separate pr

.github/workflows/submitty_ci.yml
```diff
@@ -560,7 +560,7 @@ jobs:
         if: failure()
         with:
           name: cypress-screenshots
-          path: ${{env.SUBMITTY_REPOSITORY}}/site/cypress/screenshots
+          path: ${{env.SUBMITTY_REPOSITORY}}/site/cypress-image-diff-screenshots
```
IDzyre, line 563: This should be both screenshots and image diff screenshots, if they exist

site/public/js/notebook_builder/widgets/widget.js
```diff
@@ -1,6 +1,7 @@
 /* exported NUM_MARKDOWN, Widget */
 // eslint-disable-next-line prefer-const
 let NUM_MARKDOWN = 0;
+
```
IDzyre, line 4: this should be reverted

IDzyre (changes requested):
Looks close, some small changes.

.github/workflows/submitty_ci.yml
manzil-infinity180: done [944617e](https://github.com/Submitty/Submitty/pull/10260/commits/944617ef1f778debf671e2eb1aaf66c648ee9234)

IDzyre: This should be BOTH cypress-screenshots AND the image diff screenshots, as the normal screenshots are for debugging failed tests

IDzyre:
Once the tests pass in both CI and locally, I'll do a final review, but the progress looks good.

manzil-infinity180:
> Once the tests pass in both CI and locally, I'll do a final review, but the progress looks good.

what you think about the threshold value , i used 0.02 

> Once the tests pass in both CI and locally, I'll do a final review, but the progress looks good.

is that threshold value is good like i used 0.02 in this test?

IDzyre:
Ideally it would be 0 thresh hold, but we can discuss it more in the meeting.

IDzyre:
Seeing the comments at the bottom of the JS package, it would be difficult to get this working locally and in CI without using docker, which we don't plan to use/support. If we want to get it to work in just CI we could, but it would probably be hard to maintain. 

> Please notice
Be aware that despite forcing a screenshot resolution to a particular height and width for a test, if this test is run on different computers (i.e a 13" Mac vs PC attached to a 30" monitor), the results will be different. So it's extremely important that you standardize where the tests will run, both locally and CI.
One way to handle this is by running it with docker container. This project tests use a docker container to run the tests so it could be used as an example.

bmcutler:
Closing this PR because unfortunately, due to differences in screen resolution (and perhaps also browser and native machine OS) the idea of performing automatic image comparisons will be challenging.  It will be difficult for developers to debug problems and maintain these tests long term.

Closing this PR for now.  
Perhaps this testing will be better supported and easier to do in the future, and we can re-implement/re-open this PR.

PR #10261, opened 2024-03-15
[Bugfix:InstructorUI] Images Widget in Notebook returns null

chenjohnny-cs:

The image widget currently throws a null error whenever there is no image uploaded. When this button was clicked, the save button & the show json button broke because of this.

### What is the new behavior?
This PR adds an error check to the images widget and requires the images to be uploaded to the widget so no more errors are thrown to the console.

### Other information?
This PR is linked to #10170


site/app/controllers/admin/NotebookBuilderController.php
```diff
@@ -43,12 +43,11 @@ public function builder(string $g_id, string $mode) {
         }
 
         $failure_url = $this->core->buildCourseUrl(['gradeable', $gradeable->getId(), 'update']) . '?nav_tab=1';
-
-        if (!$gradeable->isUsingUploadedConfig()) {
-            $this->core->addErrorMessage("Notebook builder may only edit uploaded configurations for the current course and semester.");
-            return new RedirectResponse($failure_url);
-        }
-
+        // Related to PR #5945
```
DarthNyan, line 46: If we're deciding we want to not check whether the config is custom, it's probably best to remove the related lines to declutter the code. IDK how often professors use each other's config files, but if they do it often it might be worth moving this check below the $mode === 'new' block so that one professor can't edit another's config.

DarthNyan (changes requested):
Feature works, but I don't like committing commented out blocks of code to main. I'm also a bit uncertain about removing the isUsingUploadedConfig check altogether, but I think checking that or whether it's using a Notebook config before allowing someone to edit should work.

site/app/controllers/admin/NotebookBuilderController.php
chenjohnny-cs: I completely removed this commented code out in a more updated PR #10285 , but I'm confused as to moving the isUsingUploadedConfig() check further down into $mode === "new".

Could you also expand on what you mean by professors can't edit another professors config?
From what I heard from Barb, the Notebook Builder should be accessible to create quizzes and should not be only accessible when a configuration is uploaded.

DarthNyan: The configs are stored in a directory on the server. My concern is that if two or more gradeables are using the same config in the same directory and one gets edited, that shouldn't affect the other gradeable. An example would be the provided configs returned by getDefaultConfigPaths(), which should not be editable via notebook builder. Imagine if a professor has a gradeable using a provided "upload a PDF" config, and they decide they want to alter the config for whatever reason so they click the button to edit an existing config with notebook builder. That should not enable them to edit the provided config.

I think I worded it poorly, but the code executed if $mode==='new' handles creating a new configuration, and below that (line 85) starts the code for loading the notebook editor. I think this check should be moved there, so that instructors can always create a new notebook config but cannot edit the existing config with notebook if it's something like a provided config.

bmcutler: It is actually common to re-use an autograding config, sometimes for multiple gradeables in the same course.  Or for the same course in a different semester.  Uploaded configs or configs built with the notebook editor or this new config editor should be stored in a directory per course.  Sometimes we have multiple instructors per course, it is possible they will work together to configure autograding for an assignment.

site/app/controllers/admin/NotebookBuilderController.php
```diff
@@ -43,12 +43,11 @@ public function builder(string $g_id, string $mode) {
...
+        // if (!$gradeable->isUsingUploadedConfig()) {
+        //     $this->core->addErrorMessage("Notebook builder may only edit uploaded configurations for the current course and semester.");
+        //     return new RedirectResponse($failure_url);
+        // }
```
04vmatsibekker, line 50: remove comments

04vmatsibekker (changes requested):

site/app/controllers/admin/NotebookBuilderController.php
chenjohnny-cs: Removed this check, moved it further down to line 85

site/app/controllers/admin/NotebookBuilderController.php
chenjohnny-cs: Check was moved down to line 85 as requested, and the functionality seems to work as intended to where you shouldn't be able to modify PROVIDED configs, but you can modify UPLOADED configs and create a new config anytime.

DarthNyan (approved):
Looks good to me.

williamjallen (approved):
Code looks reasonable.  Did not test.

PR #10263, opened 2024-03-18
[Bugfix:System] Move Doctrine Proxy Install

cjreed121:

Doctrine proxies are installed before composer gets updated. This causes issues when we update Doctrine related code and could potentially be using old proxies or modify code that relies on the newer code.

### What is the new behavior?
Doctrine proxies are generated after composer install.

### Other information?
This should be merged before #10245

### DIFF
--- a/.setup/install_submitty/install_site.sh
+++ b/.setup/install_submitty/install_site.sh
@@ -151,15 +151,19 @@ fi
 # create access control cache directory
 mkdir -p ${SUBMITTY_INSTALL_DIR}/site/cache/access_control
 
+# clear old doctrine proxy classes
+if [ -d "${SUBMITTY_INSTALL_DIR}/site/cache/doctrine-proxy" ]; then
+    rm -rf "${SUBMITTY_INSTALL_DIR}/site/cache/doctrine-proxy"
+fi
+# create doctrine proxy classes directory
+mkdir -p ${SUBMITTY_INSTALL_DIR}/site/cache/doctrine-proxy
+
 # clear old lang cache
 if [ -d "${SUBMITTY_INSTALL_DIR}/site/cache/lang" ]; then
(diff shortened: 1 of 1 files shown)


williamjallen (commented):
Do you need to move these lines as well?  https://github.com/Submitty/Submitty/blob/bb040a431a750cb881d2ec7627726f9952f3a94d/.setup/install_submitty/install_site.sh#L176-L178

williamjallen (approved):

.setup/install_submitty/install_site.sh
```diff
@@ -230,6 +216,20 @@ fi
 find ${SUBMITTY_INSTALL_DIR}/site/vendor -type d -exec chmod 551 {} \;
 find ${SUBMITTY_INSTALL_DIR}/site/vendor -type f -exec chmod 440 {} \;
 
+# clear old doctrine proxy classes
+if [ -d "${SUBMITTY_INSTALL_DIR}/site/cache/doctrine-proxy" ]; then
+    rm -rf "${SUBMITTY_INSTALL_DIR}/site/cache/doctrine-proxy"
+fi
+# create doctrine proxy classes directory
+mkdir -p ${SUBMITTY_INSTALL_DIR}/site/cache/doctrine-proxy
```
MasterOdin, line 224: I'd consider leaving this block where it was just so it's in the same block of deleting and recreating cache folders, where we'd benefit from DRY that up by using an array and a loop, vs all the repetition.

MasterOdin (changes requested):
Reviewing the script, looks like these lines should also be moved down after we run `composer install` since it loads composer autoload:

https://github.com/Submitty/Submitty/blob/c4a2aa52653ba3799c00c6a422f2db276f8612a8/.setup/install_submitty/install_site.sh#L161-L162

MasterOdin (approved):

Issue #10265, opened 2024-03-18
Calendar button that sets date to current day


arougey:
Now that there is/is going to be a way to quickly change the calendar dates, there should also be a way to quickly change the calendar date to the current day. 

In my mind, this would be a simple button somewhere on the upper navigation bar that simply has text that says "Today". 

An alternative could be an option within the dropdown menu, but I think this alternative is worse.

PGPradhan:
To address this issue I would like to give me some suggestion.
1.   Replace the existing date handling code with a modern date library like **Luxon or date-fns.** These libraries provide robust date manipulation capabilities and can handle date formatting and parsing more effectively.
2.  Submitty is used in different regions. Therefore, it may face an issue of **Localization**. To encounter this issue we can implement libraries like Moment.js or Globalize.js that can help us format dates according to different locales.

arougey:
Closed by #10267

PR #10249, opened 2024-03-14
[Feature:Calendar] Month and year dropdown menu

arougey:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [x] Documentation has been updated/added if relevant
* [x] Screenshots are attached to Github PR if visual/UI changes were made

### What is the current behavior?
Closes #10242, closes #10246. Currently, users have to repeatedly click the arrow buttons to move the calendar by a month. When going back more than a few months, this is tedious.

### What is the new behavior?
This PR implements a dropdown menu (4 years in the past, 1 year in the future) so that users can more easily go back and forwards in time on the calendar. It uses cookies to keep the calendar on the same state when reloading the page as well.

image

### Testing
This was tested on the VM as instructor with many different months and years selected.

PR #10267, opened 2024-03-19
[Feature:Calendar] Button to Set Calendar Date to Today

arougey:
Screenshot from 2024-04-06 13-09-31

The current behavior is that, to change the date to the current day, the user must press the arrow keys repeatedly.

The new behavior is that, there is now a blue button on the upper left-hand side of the calendar page that
sets the calendar date to the current day/month/year and also updates all cookies.

This should not be a breaking change and was tested on both student and instructor accounts. 

However, this pr could interfere with my other calendar pr #10249 which creates month/year dropdowns which may or may not update when the button is pressed.

### DIFF
--- a/site/app/templates/calendar/Calendar.twig
+++ b/site/app/templates/calendar/Calendar.twig
@@ -51,13 +51,9 @@
     }
 </script>
 
-
 {# Begin Calendar #}
 <div class="content">
     <h1>Calendar</h1>
-    <button type="button" onclick="setDateToToday()" class="btn btn-primary">
-        <i class="fas fa-clock"></i> Today
-    </button>
     <div class="cal-content">
         <div class="cal-toolbar">
             {% if instructor_courses|length > 0 %}
@@ -65,7 +61,6 @@
                     <i class="fas fa-plus"></i> New Item
                 </button>
             {% endif %}
-
                 <button type="button" onclick="openOptionsModal()" class="btn btn-primary">
                     <i class="fas fa-sliders"></i> Options
                 </button>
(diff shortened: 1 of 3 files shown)


arougey:
Forgot to mention that is closes #10265

sadam-04:
Looks good, i tested on student and instructor logins

sadam-04 (approved):
Looks good, i tested on student and instructor logins.

AavanVadiwala (commented):
Works. Went couple months ahead, and successfully leads user back to current date. Went couple months back, and successfully leads user back to current date. Checked on Student, Instructor, and ta. The comments that Github Action seems to make on the code in the file site/public/js/calendar.js seem small and don't affect the action, but need to be addressed to be merged.

ryvaru (approved):
Code looks good. Tested it on every login and it successfully brought me back to the current month going backwards or forwards a couple of days or years. I suggest having a bit of empty space between "Today" and "New Item" (Instructor) or "Today" and "Options" (Student, TA) for it to look better visually, but the current one tackles the issue well.

ArnavPawar (changes requested):
Looks good. As far as the UI though I think it should be aligned with the other blue buttons.

Kush-Raval (approved):
The button does what is intended.
Code change makes sense, code is ready to merge after JS lint is fixed.

arougey:
Button has been moved to the right

williamjallen:
> Button has been moved to the right

Can you update the screenshot in the PR description to reflect this?

williamjallen (approved):
Code seems reasonable.

PR #10277, opened 2024-03-24
[Feature:Submission] Add Search Functionality to Timezone

### LINKED ISSUE #10186: Search functionality for timezone in user profile
 in user profile , we have many timezones to select for, so it would be great if we have search functionality. 

simply,by adding an Event Listener to the dropdown,we can perform search of relavent inputs... 

Time-Zone-Fix

Aditya-Dhodapkar:
This PR introduces a searchable dropdown for the timezone selection in the user profile, improving the user interface and making it easier for users to find and select their correct timezone. 

The timezone drop-down element does not support a search feature that can help users search for their required timezone. The user has to manually scroll until they find the timezone they want.

### What is the new behavior?
The timezone drop-down element is now searchable. The user can now search for the timezone they want and find it faster.

###Changes:
- Implemented Select2 for the timezone dropdown.
- Adjusted the timezone dropdown's CSS to ensure consistency with the Submitty's UI theme.

###Testing:
 - The changes have been tested across different themes (light, dark, black) to ensure that the search functionality works as intended and that the UI is consistent across various use cases.
 -  New cypress test cases have been added that test the select2 implementation on the timezone element. Multiple variations of tests have been added.
 -  The changes have been tested across different browsers to ensure that the functionality works as intended. 

###Challenges Faced:
One of the challenges I faced when integrating select2 was the lack of support for themes in the control. With the bootstrap5 option, the select2 controls drop-down arrow UX does not match the default select drop-down arrow UI. To fix this issue I extended the select2 control and changed its SVG-rendered chevrom dropdown to look like the default select control. While doing this I also leverage the existing logic in the application that allowed me to use conditional styling base on data-theme.

###Feedback Request:
I've added the style overrides to user-profile.css that enable select2. I wanted feedback on the correct CSS file to add these overrides.

###Screenshots:

LightTheme_time_zone_drop_down

DarkTheme_time_zone_drop_down

Fixes #10186


Aditya-Dhodapkar:
Added a second commit to fix errors (lining and styling errors) that were shown in the tests above

JoJo-ESC (approved):
Functionality check. The search functionality was implemented well and looks good. As for the issue of where the override changes should go that seems to be an ok place to put it .

ZheyuDeng (approved):
tested functionality, didn't find any bug.

williamlin6803 (approved):
The code looks good to me and I tested it on my laptop and it works

site/cypress/e2e/Cypress-UI/profile.spec.js
```diff
@@ -85,11 +85,52 @@ describe('Test cases revolving around user profile page', () => {
...
     it('Time zone selector should work', () => {
-        cy.get('[data-testid="time-zone-dropdown"]').should('contain.text', 'NOT_SET/NOT_SET');
-        cy.get('[data-testid="time-zone-dropdown"]').select('(UTC-07:00) America/Phoenix');
-        // Since the login success message is still up, we get the next message.
-        cy.get('[data-testid="popup-message"]').next().should('contain.text', 'Time-zone updated successfully');
-        cy.get('[data-testid="time-zone-dropdown"]').select('NOT_SET/NOT_SET');
+
+        //Check that the default value is NOT_SET/NOT_SET
+        it('should have the correct default value', () => {
+            cy.get('[data-testid="time-zone-dropdown"]').should('contain.text', 'NOT_SET/NOT_SET');
```
IDzyre, line 91: One line tests in cypress are inefficient, and not good practice. In reality, all of these separate 'it' statements can be combined into one.

IDzyre (changes requested):

site/cypress/e2e/Cypress-UI/profile.spec.js
```diff
@@ -85,11 +85,52 @@ describe('Test cases revolving around user profile page', () => {
 
     // Selenium test_time_zone_selection
     it('Time zone selector should work', () => {
```
IDzyre, line 87: None of the tests are actually being run because they are inside of another 'it' statement, my suggestion would be to remove all of the inside 'it' statements, and have one big check for the time zone selector, and you can separate parts by comments so code readability is easier.

IDzyre (changes requested):
Cypress fixes needed.

site/cypress/e2e/Cypress-UI/profile.spec.js
Aditya-Dhodapkar: I'll fix this by Tomorrow, thanks for letting me know.

ZheyuDeng:
There is a problem when there isn't space to expand the menu below it will extand to the top but looks a bit weird.
image

IDzyre (changes requested):
Cypress tests are close, just a small comment and a CSS question.

PR #10280, opened 2024-03-25
[Bugfix:System] Fix main notifications

IDzyre:

Any branch named "main" will notify Zulip if the tests fail

### What is the new behavior?
It will only notify if the head_repository (the requesting repository, not the repository where the workflow file is located) is Submitty/Submitty

### DIFF
--- a/.github/workflows/notify_main_fail.yml
+++ b/.github/workflows/notify_main_fail.yml
@@ -8,7 +8,7 @@ on:
       - 'main'
 jobs:
   Notify-Failure:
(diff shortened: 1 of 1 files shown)


cjreed121 (commented):
Our original syntax should be fine according to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository

williamjallen:
Doing this adds a "skipped" job to the list of actions in every PR.  Is there a way to avoid that?

cjreed121 (approved):

PR #10284, opened 2024-03-26
[Refactor:Forum] Longer Markdown Cookies

### LINKED ISSUE #10251: Forum markdown button status should be preserved indefinitely
**Describe the bug**
If Markdown is enabled on the forum, the cookie will expire after one month.  This means that I have to repeatedly enable Markdown throughout the semester.

**Expected behavior**
It would be better to set the expiration time to the end of time so the Markdown status is preserved indefinitely.

RayReyRam:

Closes #10251 :Previously if markdown is enabled on the forum, the cookie will expire after one month. 

### What is the new behavior?
The markdown cookie will be enabled for a year, meaning you wont have to re-enable the box every month

### Other information?
To test enter the discussion form, turn on the markdown box then send a message, then inspect element and navigate to the cookies section(different for each browser) then see the markdown_enabled and see its expiration date.


arougey (approved):
Looks like an ordinary non-breaking change but seems a little redundant considering that the value is already huge. I'll still approve.

chenjohnny-cs (changes requested):
The PR definitely does work, but only when you comment on a forum post. I think it should be enabled when you click on the markdown button. In which case, I believe looking in forum.js, where the toggleMarkdown() button is at and experimenting with the Cookies.set({}) when toggling.

You can probably refer to this: 
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/set
and use the expires: (expiry date) tag

williamjallen (approved):
Code looks good.  I also tested this manually and verified that the cookie now expires after one year.

PR #10292, opened 2024-03-29
[Bugfix:System] Add flag for provisioning

### LINKED ISSUE #10286: Vagrant requests unnecessary provisioning
This problem occurs when developers use vagrant up to successfully start their Submitty server, specifically after the first time it is performed (the installation process). Even though the command was successful, vagrant will suggest the user run the 

```==> ubuntu-22.04: Machine already provisioned. Run `vagrant provision` or use the `--provision`
==> ubuntu-22.04: flag to force provisioning. Provisioners marked to run always will still `run.```

If this command is run by the user then they must use the vagrant destroy command and restart the installation process. This command breaks the installation that is otherwise working.

I think this problem should be solved by preventing vagrant from writing to the terminal after vagrant up runs successfully. If not possible, the documentation should be updated to warn against this.

smarizvi110:

Fixes #10286

### What is the new behavior?
Creates a flag after provisioning to prevent future provisioning of the same kind. Does this by passing an if/else statement to the shell for checking if the file exists or not. If it doesn't, provisioning of the kind in question goes ahead. If it does, provisioning of the kind in question is skipped.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -60,8 +60,17 @@ def gen_script(machine_name, worker: false, base: false)
     GIT_PATH=/usr/local/submitty/GIT_CHECKOUT/Submitty
     DISTRO=$(lsb_release -si | tr '[:upper:]' '[:lower:]')
     VERSION=$(lsb_release -sr | tr '[:upper:]' '[:lower:]')
-    mkdir -p ${GIT_PATH}/.vagrant/logs
-    #{setup_cmd}
+    
+    # Conditional provisioning check at the very start
+    if [ ! -f "/vagrant/#{machine_name}_provisioned" ]; then
+      echo "Provisioning #{machine_name} for the first time..."
+      mkdir -p ${GIT_PATH}/.vagrant/logs
+      #{setup_cmd}
+      # Mark this provisioning step as done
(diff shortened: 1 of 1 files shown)


IDzyre:
I don't think that this is a 'fixable issue' as the 'request for provisioning' is a message from Vagrant stating that the machine is already provisioned, and letting you know if you accidentally stopped re-provisioning.

thegame615:
> I don't think that this is a 'fixable issue' as the 'request for provisioning' is a message from Vagrant stating that the machine is already provisioned, and letting you know if you accidentally stopped re-provisioning.

I think that the changes that were made involving changing vanguard to use a prepackaged VM largely solves the issue of things being broken. Which means that to my understanding the issue may only persist on mac (last I heard there was no prebuilt image for mac). 

I also do think that it is possible to notify the user when they mistakenly try to re provision the machine, I just haven't found the time to work on it. To clarify, I am not saying to remove the message that vagrant writes, I am only suggesting that if a dev runs vagrant provision (which will never be necessary to my knowledge) it will stop them and tell them not to do it.

williamjallen:
After discussion, we've decided that the approach taken in https://github.com/Submitty/Submitty/pull/10470 is the best approach.

PR #10299, opened 2024-03-31
[Bugfix:InstructorUI] Doc link, display errors on upload

shailpatels:

If an upload gradeable via JSON configuration fails, the page redirects which prevents the user from seeing any error messages. The documentation link for the upload gradable model is broken

### What is the new behavior?
Page no longer redirects on error, allowing the alert to be shown. In the future this alert can be replaced with the usual red error message or with a modal that can display the errors better if the validation message is more complex

### Other information?

No

Easiest way to test the error message is to upload a gradable config for one that exists already

### DIFF
--- a/site/app/templates/admin/admin_gradeable/AdminGradeableUploadForm.twig
+++ b/site/app/templates/admin/admin_gradeable/AdminGradeableUploadForm.twig
@@ -7,7 +7,7 @@
     </p>
     <p>
         For more formatting information, see
-        <a target="_blank" href="https://submitty.org/instructor/assignment-preparation/upload_gradeable">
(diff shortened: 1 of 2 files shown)


IDzyre (approved):
When I run the upload, I can see the error message, but I think that this is an accurate change, if we are using the red error message in the future.

IDzyre:
> Oh interesting you were able to see the error message before this change? I tested on firefox on linux not sure if other browsers behave differently
 
I use Chrome and OperaGX on windows, so i wonder if chrome(like) browsers behave differently.

williamlin6803:
The code looks good and I tested it on my laptop and it works well!

williamlin6803 (approved):
The code looks good and I tested it on my laptop and it works well!

PR #10314, opened 2024-04-02
[Bugfix:Developer] Initialize "os" key in Docker UI

williamjallen:
https://github.com/Submitty/Submitty/pull/10271 missed the "os" key, which is believed to be the cause of [this](https://github.com/Submitty/Submitty/actions/runs/8515492568/job/23323056920) CI failure.

### What is the new behavior?
The "os" key is initialized to null.

The original source of the Docker UI test flakiness is PR https://github.com/Submitty/Submitty/pull/7932.  Notably, the last comment on that PR discusses the flakiness of the tests, and the subsequent [commit](https://github.com/Submitty/Submitty/pull/7932/commits/5acb72fb030ac02cbd54195e786a62dbea96f386) comments out a test in an attempt to fix the flakiness.  In the future, it would be worthwhile to investigate the commented-out test to see if it is now reliable.

### DIFF
--- a/site/app/views/admin/DockerView.php
+++ b/site/app/views/admin/DockerView.php
@@ -135,7 +135,7 @@ class DockerView extends AbstractView {
                 // Parse the OS description
                 $is_match = preg_match("/Description:\t(.+)/", $buffer, $matches);
                 if ($is_match) {
-                    $machine_system_details[$current_machine]["os"] = $matches[1];
+                    $machine_system_details[$current_machine]["os"] = $matches[1] ?? null;
                 }
 
                 // Parse the docker version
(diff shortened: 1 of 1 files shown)


DarthNyan (changes requested):
"os" is initialized earlier on line 138 of this file. This will overwrite it with null.

DarthNyan (changes requested):
This is running in a while loop over a string buffer. If a later buffer string doesn't match this pattern, it'll overwrite "os".

DarthNyan (approved):

cjreed121 (approved):

PR #10281, opened 2024-03-25
[Feature:Calendar] Changed Calendar Cells Based on View

arougey:
Changed the calendar cells to be larger based on the viewing mode(month, two-week, week)

Screenshot from 2024-03-25 15-29-22

Screenshot from 2024-03-25 15-29-09

The current behavior is that if you change viewing modes from 'Month' to 'Two Week' or 'Week', it simply hides the extra weeks. The purpose of the other view modes is to make cells larger to fit more information and be less crammed for users who have many items on their calendar.

The new behavior is that the cells will expand based on the viewing mode as well as the number of items in each cell. 

This is not a breaking change and was tested on instructor and student accounts.

PR #10318, opened 2024-04-02
[Bugfix:System] Check DB connected before closing

IDzyre:

Currently in some cases, when deconstructing the database object, an error is thrown trying to disconnect from a null object, due to the database not being connected. 

image

### What is the new behavior?
image
Now it checks if the database is connected before attempting to close the connection. 
See https://github.com/Submitty/Submitty/pull/10281 for discovery

### DIFF
--- a/site/app/libraries/Core.php
+++ b/site/app/libraries/Core.php
@@ -336,10 +336,10 @@ class Core {
             $this->logPerformanceWarning();
         }
 
-        if ($this->course_db !== null && $this->course_db->isConnected()) {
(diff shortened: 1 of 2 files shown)


cjreed121 (changes requested):
Seems to be related to the recent DBAL change. I think the better fix would be to do the check inside AbstractDatabase instead. Should just be a simple null check.

cjreed121:
@IDzyre can you fix the lint error then I'll approve

cjreed121 (approved):

PR #10324, opened 2024-04-03
[Bugfix:Forum] Highlight new posts in Forum

### LINKED ISSUE #10319: New forum posts added by sockets should be highlighted as unread
**Describe the bug**
When you are viewing a forum thread for the first time, all the posts within it are light blue (light blue = unread -- in light mode).  If a new post is added while you are on that thread, the websocket will add it as white post (white post = read -- in light mode).  This doesn't make sense to have the newest posts be read while the older posts are blue.

**Expected behavior**
All new posts should be unread.

**Related Note**
I've noted elsewhere (maybe not formally typed up as an issue) that the visualization of read vs unread posts in dark mode is unclear (or even buggy) since most users in dark mode aren't even aware that we have read vs. unread status in the forum.  This should be fixed.  The colors used for read vs unread mode should be adjusted so the difference is more obvious.

CCWizma99:
When a user got a new post while he is in the forum page, user's browser will render it as a read message. But, it should be rendered as an unread message.

### What is the new behavior?
Now, if a user got a post while user is in the forum page, browser will render it as an unread message.

### Other information?
(Changed according to the request from @DarthNyan.)
Changes : Added a code line to add 'new_post' class when displaying a new post.
I tested this on my VM.
It works on dark mode as well (a white color transparent overlay). 

However, I was unable to find the logical error why the system treat new messages like this. Let's try to find a more sustainable method soon. Working on it :)

closes #10319 

Thank you.

### DIFF
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -2607,7 +2607,6 @@ function newPostWhileUserOnline(){
         }
     });
     captureChange.observe(post_list, { childList: true});
-    alert('loaded');
 }
 setTimeout(function(){
     newPostWhileUserOnline();
--- a/site/public/js/forum.js
+++ b/site/public/js/forum.js
@@ -339,6 +339,7 @@ function socketNewOrEditPostHandler(post_id, reply_level, post_box_id=null, edit
                     original_post.remove();
                 }
 
+                $(`#${post_id}`).addClass('new_post');
                 $(`#${post_id}-reply`).css('display', 'none');
                 $(`#${post_id}-reply`).submit(publishPost);
                 // eslint-disable-next-line no-undef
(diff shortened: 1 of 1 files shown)


DarthNyan (changes requested):
The function `socketNewOrEditPostHandler` in this file in this file handles inserting a new post into an open page. A better fix will be to set a new post as unread there rather than listening for new posts.

CCWizma99:
> The function `socketNewOrEditPostHandler` in this file in this file handles inserting a new post into an open page. A better fix will be to set a new post as unread there rather than listening for new posts.

@DarthNyan, Thank you for your guiding. Now it works fine. Just added a code line `socketNewOrEditPostHandler`. 

Can you please review it again.

DarthNyan (approved):
Looks good to me.

DarthNyan:
I just ran the CI. Code looks good, but the PR title needs to be shortened a little and the "F" in "BugFix" should be lowercase. A title like, "[Bugfix:Forum] Highlight new posts on open threads" should work.

ArnavPawar (approved):
SImple change for next time get title changed as it gives you an error in github. Looks good besides that.

PR #10325, opened 2024-04-03
[Feature:InstructorUI] Text Editor for Gradeable Configs

DarthNyan:
The only ways to change the configuration of a gradeable are to switch the config directory to a different directory on the server, upload a new config file, or use Notebook Builder. There is no way to manually edit the gradeable's config.json or supplemental files such as sample output or custom validation code.

## What is the new behavior?
This PR creates a basic text editor that can edit any file in the config directory, provided the config was uploaded or created via Notebook Builder. It does not currently allow instructors to add files or subdirectories, just edit the existing files. It is intended to allow instructors to easily fix minor mistakes in their uploaded configurations without needing to upload a separate config.

## Other information?
**Config Editor Closed:**

config editor when it is closed

**Click "Manually edit config" to browse files in the config directory, then click on a file to edit and save your changes.**

manual configuration when it is opened

**When a provided configuration is selected:**

provided configuration is selected no editing allowed


jeffrey-cordero (dismissed):
Overall, the implementation looks good and the feature works well in terms of updating currently uploaded configurations given valid authentication. In terms of UI, I believe it would look better to make a new section like the Notebook Builder to have a button to manually edit the config and a short description. It would also be nice to check if a file type is in the correct format before saving, like invalid JSON format for example, but this may be infeasible as various file extensions are allowed. Deleting files could also be a possible addition to this feature as well.

DarthNyan:
I'll get a button added shortly.

As for checking JSON, I have it set to try to rebuild the gradeable once you save your edits, so if the JSON can't be parsed the build will fail and you'll be notified. 

As for adding and deleting files, I've thought about that but decided it's beyond the scope of this PR and would be best added in a follow-up.

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1875,4 +1877,74 @@ public function maxPoints($gradeable_id) {
         }
         $this->core->getOutput()->renderJsonError("Unknown gradeable");
     }
+
+    /**
+     * Loads config info for a gradeable to allow editing
+     * @Route("/courses/{_semester}/{_course}/gradeable/edit/load", methods={"POST"})
```
williamschen23, line 1883: change to PHP Route 8 annotations

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1875,4 +1877,74 @@ public function maxPoints($gradeable_id) {
...
+
+        $config_content = file_get_contents(FileUtils::joinPaths($file_path));
+        $output = [];
+        $output["config_content"] = $config_content;
+        $this->core->getOutput()->renderJsonSuccess($output);
+    }
+
+    /**
+     * Saves config info from a gradeable edit
+     * @Route("/courses/{_semester}/{_course}/gradeable/edit/save", methods={"POST"})
```
williamschen23, line 1912: same here

williamschen23 (changes requested):

site/app/controllers/admin/AdminGradeableController.php
zacharymnp: [23a1867](https://github.com/Submitty/Submitty/pull/10325/commits/23a1867a8acc2ed4fcafc810e0998c57d96a4eb3)

site/app/controllers/admin/AdminGradeableController.php
zacharymnp: [23a1867](https://github.com/Submitty/Submitty/pull/10325/commits/23a1867a8acc2ed4fcafc810e0998c57d96a4eb3)

GO-viper7 (dismissed):
Tested the editor in my browser, It works as intended
Code quality is good, everything makes sense

Few styling issues which need to addressed : 
1. In dark mode and dark mode(black), cancel button cannot be seen while hovering
2. When the editor gets focused, border of editor is overlapping with that of save button (little bit of margin would be good)
3. Browser based spell check for the editor needs to be removed in order to make those red lines disappear

Also, It would be good to see "tab" working inside the editor.

site/public/js/admin-gradeable-updates.js
```diff
@@ -878,3 +879,117 @@ function hideBuildLog() {
...
+
+let originalConfigContent = null;
+
+// When the text editor opens, the user shouldn't have to manually scroll to see the contents
+function scrollToBottom() {
+    window.scrollTo({ top: 800, left: 0, behavior: 'smooth' });
+}
+
+// When you load the editor
+function loadGradeableEditor(g_id, file_path) {
```
jeffrey-cordero, line 912: This sends a request on each click. Could we optimize this to only send the request if the element with ID of `gradeable-config-edit-bar` has a `display` of `none`? A step further would be to save the content in an object tied to the given path.

image

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
```diff
@@ -110,6 +111,21 @@
...
 
+    <div class="raw-config-info">
+        <div class="option-title raw-config-info-header">Manual Configuration:</div>
+        <a class="btn btn-primary" tabindex="0" id="open-config-editor" onclick="toggleGradeableConfigEdit()">Open Editor</a>
+        <div id="gradeable-config-structure">
+            <ul class="config-folder-view">
+                {{ _self.display_config_files(config_files, 1, gradeable.getId()) }}
+            </ul>
+            <div id="gradeable-config-edit-bar">
+                <textarea id="gradeable-config-edit" data-edited="false" aria-label="Edit Gradeable Configuration" spellcheck="false"></textarea>
```
jeffrey-cordero, line 123: Functionality to allow native tabs would be nice if it would add little overhead to the implementation. Otherwise, I think it would be a good future issue to note.

image

site/public/css/admin-gradeable.css
```diff
@@ -377,6 +377,38 @@ body::-webkit-scrollbar-track {
...
+    min-height: 500px;
+    margin-top: 0.5em;
+    margin-bottom: 0.5em;
+    overflow: auto;
+    resize: vertical;
+}
+
+#gradeable-config-edit-bar {
+    display: none;
+    overflow: hidden;
```
jeffrey-cordero: I would remove the `overflow: hidden` as it cuts off the save / cancel edit buttons.

image

site/app/templates/admin/admin_gradeable/AdminGradeableAuto.twig
JManion32: Implementing the native tab isn't difficult, however, I am having trouble getting it to work with the browser's undo/redo functionality. I think I'll leave this feature for a future PR so we can ensure it is properly implemented.

site/public/js/admin-gradeable-updates.js
```diff
@@ -878,3 +879,131 @@ function hideBuildLog() {
...
+function scrollToBottom() {
+    window.scrollTo({ top: 800, left: 0, behavior: 'smooth' });
+}
+
+let current_g_id = null;
+let current_file_path = null;
+
+function updateGradeableEditor(g_id, file_path) {
+    // If no file has been selected yet or it is not the currently selected one
+    if ((current_g_id === null && current_file_path === null) || (current_g_id !== g_id || current_file_path !== file_path)) {
```
jeffrey-cordero, line 904: There's just a small bug now when the instructor cancels changes, they must close then re-open the editor via the button to display the content again. Once this is fixed, all looks good for approval.

image

bmcutler (commented):
The UI is clunky and non-intuitive.   Further improvements are definitely needed.  
It is weird that it allows you to open the editor for files you don't have permission to edit, and type into those files, but gives the error only when the 'save' button is pressed.
But it seems to work, so I'm ok merging this now.

site/app/controllers/admin/AdminGradeableController.php
```diff
@@ -1983,4 +1986,74 @@ public function maxPoints($gradeable_id) {
         }
         $this->core->getOutput()->renderJsonError("Unknown gradeable");
     }
+
+    /**
+     * Loads config info for a gradeable to allow editing
+     */
+    #[Route("/courses/{_semester}/{_course}/gradeable/edit/load", methods: ["POST"])]
```
williamschen23, line 1993: this would be a GET request?

bmcutler: the functionality breaks with that change.
I'm going to merge this PR now, but please open an issue or PR if something further needs to be addressed.

bmcutler (approved):

PR #9971, opened 2023-11-06
[Testing:Developer] Add pylint to github actions

jason490:
### Please check if the PR fulfills these requirements:

* [ ] Tests for the changes have been added/updated (if possible)
* [ ] Documentation has been updated/added if relevant
* [ ] Screenshots are attached to Github PR if visual/UI changes were made

### What is the current behavior?
We did not have a static code analyzer for python. So we are using pylint.

### What is the new behavior?
Running pylint alongside flake8 with github actions to diversify python testing.

PR #10326, opened 2024-04-03
[Testing:Developer] Resolve pylint warnings

justin-sadakhom:

- Python linting step in the GitHub Actions workflow fails due to pylint warnings
- Please see the [base branch](https://github.com/Submitty/Submitty/pull/9971) for details

### What is the new behavior?

The command `python3 -m pylint --recursive=y .` no longer outputs any warnings.

### Other information?

- ~~There are some points I am unsure about~~
- Changes ignore files that fail linting instead of linting


williamjallen:
Thanks for taking over this effort!

In response to your comments in the PR description:
- We ignored any file which was failing the linter since this PR is just adding pylint itself.  Linting the files can be done in batches via separate PRs.
- Rather than adding the dependencies to the dev requirements.txt, I think it would be better to install both the dev and system requirements.txts.

justin-sadakhom:
Okay following up on your response:
- Ignoring an `__init__.py` file directly doesn't work, so I've adjusted it so the entire directory is ignored
- This has the side effect of condensing some of the file ignores into a single directory ignore

justin-sadakhom:
It seems one of the Cypress tests is failing. I suspect this is because the base branch doesn't have the latest changes from main. I think the best approach may be to merge this into the base branch and then merge main in.

williamjallen (approved):
It's unfortunate that `__init__.py` can't be explicitly ignored, but it does explain the errors I wasn't able to figure out.  Although it's not ideal to have entire directories ignored (new files won't be linted now), I suppose that it's probably the best that can be done for the moment.

Issue #8631, opened 2022-11-28
Student access to API


ryanting01:
I am trying to create a Visual Studio Code extension for students. Features would be to view their gradeables, submit from VS code to Submitty, see their results, etc.
I am having a problem as the API is not available for student as I when making a call it says i is only available for staff. I am just wondering if it is possible to implement these things through the API, and why is the API currently not available for students even though it says it is available in the documentation.

Hopefully this can be done and thank you for the help!

IDzyre:
Student API access has been added in #10330

Issue #6245, opened 2021-02-25
Visual Studio Code Extension


prestoncarman:
**What problem are you trying to solve with Submitty**
Integrate grading UI into the IDE which students are writing their code. For new students, it would be nice to cut down on the many places they have to go to complete an assignment.

**Describe the way you'd like to solve this problem**
Build an extension for Visual Studio Code that enables a student to:
* Trigger auto-grading on their assignment
* See auto-grading results
The process would be tightly integrated into the IDE, similar to the run button for executing their code.

**Thing to consider**
* These features could be added with a capable API from Submitty

**Additional context**
The goal would be to get students used to using the IDE without having to navigate to a web browser to see their test results. Update the Submitty grading process to look more closely with how typical unit testing works.

shailpatels:
@prestoncarman Do you have an initial design in mind for what you want the API endpoint to look like? 

It sounds like from your use case you allow students to submit as many times as they want so the submission endpoint probably won't have any way to notify students of late submissions/late day usage/ etc. This could probably be built into its own endpoint if needed though.

nilesh05apr:
Hello @shailpatels @prestoncarman is this task completed? if not I would like to work on this task, please share further details.

shailpatels:
> Hello @shailpatels @prestoncarman is this task completed? if not I would like to work on this task, please share further details. 

This task is still open, we've done a lot of work related to the API but nothing touching auto grading. This is probably still a large task but probably the first place is to look at the existing API functionality

ryanting01:
Hello! 
I am currently creating a Visual Studio Code extension and I am expanding the API functionality. I am trying to create an endpoint for students to submit a submission to a gradeable. I am trying to use this function https://github.com/Submitty/Submitty/blob/ed4fbf3f6485927f32ca8b1397af34d67084a304/site/app/controllers/student/SubmissionController.php#L1018 in an api endpoint like this:
`   
 /**
     * Submit a submission to a gradeable
     * @Route("/api/submit", methods={"POST"})
     *
     * @param string|null $user_id
     * @return MultiResponse
     */
    public function submitSubmission($user_id = null) {

        $this->addUploadFile("test1.txt", "", 1);

        $controller = new SubmissionController($this->core);
        $return = $controller->ajaxUploadSubmission('c_failure_messages');

    }
`
All dependancies should be working and it seems that I am getting an error whose origin doesn't seem to be related:

Oh no! Something irrecoverable has happened...

[...]

PR #10330, opened 2024-04-05
[Feature:API] Add student API support

IDzyre:

Currently, the API is only available to faculty (usually instructors). 

### What is the new behavior?
The API is being expanded to be able to be used by students, while keeping the restriction on most of the existing endpoints to instructors. 

To start, the endpoints will be retrieving autograding information for a graded assignment, and requesting a VCS gradeable be submitted. 

### Other information?

Related to #8631, #6245


site/app/controllers/student/SubmissionController.php
```diff
@@ -1020,6 +1020,63 @@ public function ajaxRegrade($gradeable_id): array {
...
+     * @param string $gradeable_id
+     */
+    #[Route("/api/{_semester}/{_course}/gradeable/{gradeable_id}/score", methods: ["GET"])]
+    public function ajaxGetGradeableScore(string $gradeable_id): JsonResponse {
+        // Instructor
+        if ($this->core->getUser()->getAccessLevel() !== 1) {
+            if (($_GET['user_id'] ?? '') !== $this->core->getUser()->getId()) {
+                return JsonResponse::getFailResponse('API key and specified user_id are not for the same user.');
+            }
+        }
```
williamjallen, line 1034:
```suggestion
        if ($this->core->getUser()->getAccessLevel() !== 1 && ($_GET['user_id'] ?? '') !== $this->core->getUser()->getId()) {
            return JsonResponse::getFailResponse('API key and specified user_id are not for the same user.');
        }
```

site/app/controllers/student/SubmissionController.php
```diff
@@ -1020,6 +1020,63 @@ public function ajaxRegrade($gradeable_id): array {
...
+        $gradeable_version_instance = $this->core->getQueries()->getGradedGradeable(
+            $gradeable,
+            $_GET['user_id'],
+            $gradeable->isTeamAssignment()
+        )->getAutoGradedGradeable()->getActiveVersionInstance();
+
+        if ($gradeable_version_instance === null || $gradeable_version_instance->isQueued() || $gradeable_version_instance->isGrading()) {
+            // Return null value for not graded, or not submitted gradeables.
+            return JsonResponse::getSuccessResponse(null);
+        }
```
williamjallen, line 1050: Would it make sense to return different values for each of these states, or should this endpoint be strictly for getting the score?  I'm fine with it either way, but I wanted to bring it up in case you hadn't thought about it.

Furthermore, why do we need an endpoint to get the score in the first place?  Would it not make sense to just have an endpoint which gets an entire gradeable, including the score and other metadata such as the grading status, grading time, etc?

site/app/controllers/student/SubmissionController.php
```diff
@@ -1020,6 +1020,63 @@ public function ajaxRegrade($gradeable_id): array {
...
+        $vcs_checkout = array_key_exists('vcs_checkout', $_POST) ? $_POST['vcs_checkout'] === "true" : false;
+        if (!$vcs_checkout) {
+            return JsonResponse::getFailResponse("API only supports requesting for VCS gradeables to be graded.");
+        }
+        if (!array_key_exists('git_repo_id', $_POST)) {
+            return JsonResponse::getFailResponse("API requires git_repo_id variable to be set.");
+        }
+
+        return $this->ajaxUploadSubmission($gradeable_id);
```
williamjallen, line 1077: Why doesn't this return a regular json success response?

williamjallen (changes requested):

site/app/controllers/student/SubmissionController.php
IDzyre: I will change it to get an array of the values instead of just score.

site/app/controllers/student/SubmissionController.php
IDzyre: This returns the response from the ajaxUploadSubmission function, which is in the long run a JsonResponse success or fail response.

williamjallen:
@IDzyre It looks like the tests are failing due to a flaky test which has since been fixed on main.  Updating this from main should revolve the issue.

site/app/controllers/student/SubmissionController.php
williamjallen: Why does the return type have array an option?

IDzyre: This function calls `uploadResult` which then calls `renderResultMessage`, which returns an array formatted as a success or fail JSON response, but as an array.

site/app/controllers/student/SubmissionController.php
```diff
@@ -1020,6 +1020,75 @@ public function ajaxRegrade($gradeable_id): array {
         return $this->core->getOutput()->renderResultMessage($msg, true);
     }
 
+    /**
+     * @return JsonResponse
+     * @param string $gradeable_id
+     */
+    #[Route('/api/{_semester}/{_course}/gradeable/{gradeable_id}/values', methods: ['GET'])]
+    public function ajaxGetGradeableValues(string $gradeable_id): JsonResponse {
+        if ($this->core->getQueries()->getUserById($_GET['user_id']) === null) {
```
williamjallen, line 1029: You're using `$_GET['user_id']` in multiple places, but not always checking that it is not null.  You could consider making a temporary variable instead.

site/app/controllers/student/SubmissionController.php
```diff
@@ -1020,6 +1020,75 @@ public function ajaxRegrade($gradeable_id): array {
...
+    /**
+     * @return JsonResponse|array{
+     *     status: string,
+     *     data: mixed
+     * }
+     */
+    #[Route('/api/{_semester}/{_course}/gradeable/{gradeable_id}/grade', methods: ['POST'])]
+    public function ajaxRequestGrade(string $gradeable_id): JsonResponse|array {
+        // Faculty = 2, Superuser = 1
+        if ($this->core->getUser()->getAccessLevel() > 2 && ($_POST['user_id'] ?? '') !== $this->core->getUser()->getId()) {
```
williamjallen, line 1078: Instead of magic numbers, it would be better to reference static constants if possible.

site/app/controllers/student/SubmissionController.php
williamjallen, line 1029: Doesn't this potentially reveal the students in the course?  Shouldn't you instead check to see if the requested user is the current authenticated user first, or otherwise a higher level role which can see the enrollment list?

williamjallen (changes requested):

site/app/controllers/student/SubmissionController.php
IDzyre: Yes, I didn't consider that, so now the order of the checks has been changed, to account for this.

williamjallen (approved):

site/app/controllers/student/SubmissionController.php
```diff
@@ -1020,6 +1020,80 @@ public function ajaxRegrade($gradeable_id): array {
...
+     * @param string $gradeable_id
+     */
+    #[Route('/api/{_semester}/{_course}/gradeable/{gradeable_id}/values', methods: ['GET'])]
+    public function ajaxGetGradeableValues(string $gradeable_id): JsonResponse {
+        $user_id = $_GET['user_id'] ?? '';
+        // Faculty and Superusers can get values for other users, if it is a student, require the $_GET user id to be the same as the
+        // API authenticated user.
+        if ($this->core->getUser()->getAccessLevel() === User::LEVEL_USER && ($user_id !== $this->core->getUser()->getId())) {
+            return JsonResponse::getFailResponse('API key and specified user_id are not for the same user.');
+        }
```
cjreed121, line 1034: This permission should be based on the user's course level (instructor, full access grader, limited access grader, or user) rather than system level access.

cjreed121 (changes requested):
On the authentication token page, we currently only allow for faculty access and up to retrieve API tokens. If we are allowing student access to the API, we should expand access for fetching tokens too.

site/app/controllers/student/SubmissionController.php
IDzyre: This has been changed to course level, I believe.

cjreed121 (approved):
Code looks good

PR #10332, opened 2024-04-06
[Feature:Forum] Add Unread Thread Button and Dropdown

### LINKED ISSUE #10158: Discussion Forum - "Mark unread from here"
When you click on a thread in Submitty, we make a record in the "viewed_responses" database table for that user, thread, and timestamp.
But if the user doesn't have time to view the entire (lengthy) thread of messages at that time, they might want to "mark the thread as unread, either entirely, or from a specific point, so they can come back later.

Let's add a button to the UI 
"Mark unread from here"
It should be subtle, possibly tucked-away feature, to avoid clutter.  
A white button or maybe in a three dot popout menu with this as an option.
Maybe we can combine a few other less-used forum features in a popout menu.

Xiayucheng1212:
Fixes #10158 

### What is the new behavior?
Add a new triple dot  user can click on and will have a menu including unread the whole thread function. 
image

After unreading it it will looks blue again.
image

### Other information?
In addition to adding the triple dots menu, we have decided to remove the legacy icons since they are included in the dropdown menu now.


CCWizma99:
@Xiayucheng1212 I was reviewing your request and I think, now it looks like this in your server as well...

image

Three dots and when you click it, you can see the option to unread the thread. I asked because in your example screenshots they look different. Your code worked on my device. I will review the code according to current codes.

CCWizma99 (changes requested):
Your code worked well @Xiayucheng1212.

But as @bmcutler explained in the issue #10158, maybe the user will need to **make thread unread from a specific point** in the forum instead of **entire thread**. So, you may add a button or dropdown here in this section...

image

... to make "Unread thread from here". So, the task will be completed. :)

You make take the **timestamp** of that specific point and update the "timestamp" in viewed_response table. 

(In dark mode, it has a contrast issue. But, it's not a problem in you code. I saw that in other dropdowns in site as well.) 

Good luck...

Xiayucheng1212:
@CCWizma99 Thank you for your suggestions. I am still working on the partial unread part. I am curious about what you mentioned in "You make take the timestamp of that specific point and update the "timestamp" in viewed_response table.", could you explain more about it?

CCWizma99:
> @CCWizma99 Thank you for your suggestions. I am still working on the partial unread part. I am curious about what you mentioned in "You make take the timestamp of that specific point and update the "timestamp" in viewed_response table.", could you explain more about it?

submitty_thread_unread

Here you can add a dropdown menu like that and add an option to unread the thread from that specific post we chose. So, we can get the timestamp in that post (red square in the image) and add the timestamp to the viewed_response table. Next time user go to the forum, system will show the thread as user was last online on that thread where user selected post. 

If you add a dropdown menu, you may remove the buttons for editing and deleting the post (green square) and add them into the dropdown as well.

I think you got the point 🙂.
While this is my point of view, you may need to discuss with an authorized maintainer...

Xiayucheng1212:
@CCWizma99  Sounds great! Will do that, thank you.

williamlin6803 (approved):
The un-marking for the entire thread does in fact as tested. Code looks fine

site/app/controllers/forum/ForumController.php
```diff
@@ -632,6 +632,33 @@ public function bookmarkThread(bool $type) {
...
+
+    /**
+     * @Route("/courses/{_semester}/{_course}/forum/posts/unread", methods={"POST"})
+     * @return JsonResponse
+     */
+    public function markPostUnread() {
+        $thread_id = $_POST["thread_id"];
+        $last_viewed_timestamp = $_POST["last_viewed_timestamp"];
+        // format the last viewed timestamp to be in the same format as the database
+        $last_viewed_timestamp = DateUtils::parseDateTime($last_viewed_timestamp, $this->core->getUser()->getUsableTimeZone())->format("Y-m-d H:i:sO");
```
CCWizma99, line 655: In format, is it valid to add 's0'? I think it should be 
format("Y-m-d H:i:s")

CCWizma99:
> @CCWizma99 Hi, all tasks have been solved, please help me check again, thank you.

Hello @Xiayucheng1212, it's a little sad news :(

Program didn't work correctly. Let's figure out a way. I will try to find where is the issue and will let you know.

First I checked the unread Thread functionality and It added new_post class to children of thread but, didn't DELETE the row in database. Did this functionality work on your server? 
Also, when I send a post, other user get an error as "Error parsing post. Refresh the page.". Check whether you also get these issues.

Don't worry. We will find a way...

04vmatsibekker (commented):
Hey it is really hard to tell that the three buttons are supposed to show the thread should be unread. Also to look at individual posts to mark as unread the dropdown is buggy: you can scroll when dropdown is toggled. 

Perhaps to mark a thread as unread you should put it under this more button: 
image

image

Xiayucheng1212:
@04vmatsibekker Yeah, I think you are right, let's move "unread thread" to the place you suggest

site/public/js/forum.js
IDzyre, line 1: I think it would be better to fix the JS lint errors instead of ignoring them

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
```diff
@@ -137,7 +137,7 @@ describe('Tests cases abut changing user pronouns', () => {
         cy.contains('Test Anonymous thread, should not show pronouns').find('.post_user_id').click();
         cy.get('.post_user_id').should('contain', 'Anonymous');
         cy.get('.post_user_pronouns').should('not.exist');
-        cy.get('.first_post').find('.fa-trash').click();
+        cy.get('.first_post > .post-action-container > .dropdown-menu ').find(':contains("Delete")').click({force: true});
```
IDzyre, line 140: Is there a reason for these changes?

IDzyre (changes requested):

site/cypress/e2e/Cypress-UI/pronouns_change.spec.js
Xiayucheng1212: Yes, since there is no 'fa-trash' icon, the delete button is wrapped in the dropdown menu.

site/public/js/forum.js
Xiayucheng1212: Sure, but some of them are necessary, since not all variables/functions are imported/invoked explicitly in this file.

site/public/js/forum.js
```diff
@@ -1895,6 +1895,52 @@ function bookmarkThread(thread_id, type) {
     });
 }
 
+// eslint-disable-next-line no-unused-vars
+function markThreadUnread(thread_id) {
+    // eslint-disable-next-line no-undef
+    const url = `${buildCourseUrl(['forum', 'threads', 'unread'])}`;
```
IDzyre, line 1901: I think you can add buildCourseUrl to the list of global functions at the top of the file and then you can remove the es lint disable for undefined

site/public/js/forum.js
```diff
@@ -1895,6 +1895,52 @@ function bookmarkThread(thread_id, type) {
...
+    $.ajax({
+        url: url,
+        type: 'POST',
+        data: {
+            thread_id: thread_id,
+            // eslint-disable-next-line no-undef
+            csrf_token: csrfToken,
```
IDzyre, line 1908: Same with csrfToken, add it to the globals

IDzyre (changes requested):

micpap25:
@IDzyre I cleaned up the relevant globals, I'll open an issue for the rest of them

micpap25 (changes requested):
This PR needs a bit more work; when you mark a post as unread (which is meant to make the whole thread past that point also unread), only that post changes color. You need to click off and click back on to see the thread has changed to unread as expected. 

image

Additionally, I think that the thee dots for marking as unread is too small, doesn't make sense as a dropdown with only one option in it, and the phrase "Unread From Post Time" is a bit clunky.

micpap25:
Marked as a draft; feel free to re-open once it is ready for review again!

micpap25 (changes requested):
Needs a change to fix how things are marked as unread which Zheyu and I discussed in-person

micpap25 (approved):
If you unread from a certain time in a thread, all other posts from the same time (even if they are above) will be marked as unread. This is a glitch but there isn't much we can do about it seeing as we only keep track of timestamps to the second.

Otherwise this is good.

site/app/templates/forum/CreatePost.twig
```diff
@@ -119,6 +120,22 @@
                 (<i>Last edit at {{ edit_date }}</i>)
             {% endif %}
         </span>
+
+        <a class="pinned-thread key_to_click" tabindex="0" data-testid="thread-dropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" title="Options">
```
zacharymnp, line 124: `key_to_click` was deprecated in PR #10740 since button elements have the same functionality by default. `key_to_click` could be removed if the `<a>` elements are replaced by `<button>` elements.
This should also be a button since only actual links should use the `<a>` tag.

I'm not sure if this can be done without changing the style of the buttons, and a lot of Submitty uses `<a>` anyway, so feel free to ignore this comment.

site/app/templates/forum/displayThreadList.twig
```diff
@@ -47,8 +47,8 @@
             </div>
             {% if is_full_page is defined and is_full_page %}
                 <div class="post-action-container">
-                    <span>
-                        {% if user_group <= 2 and thread.author_info.user_id != current_user %}
+                    <span>{# Removing the legacy icons as they are included in the dropdown now #}
+                        {#{% if user_group <= 2 and thread.author_info.user_id != current_user %}
```
zacharymnp, line 51: Is there any reason to leave this commented out code?

site/app/templates/forum/displayThreadList.twig
```diff
@@ -63,12 +63,12 @@
...
                                aria-label="Show full user information"
                             ><i class="fas fa-eye"></i></span>
-                        {% endif %}
-
+                        {% endif %}#}
                         {% if user_group <= 3 or thread.author_info.user_id == current_user %}
                             {% if not (thread.is_thread_locked != 1 or user_group <= 3) %}
                                 {#pass#}
                             {% else %}
+                                {#
```
zacharymnp, line 71: Is there any reason to leave this commented out code?

site/cypress/e2e/Cypress-Feature/test_forum.spec.js
```diff
@@ -0,0 +1,80 @@
...
+describe('Test cases revolving around creating, replying to, merging, and removing discussion forum threads', () => {
+    beforeEach(() => {
+        cy.login('instructor');
+        cy.visit(['sample']);
+        cy.get('#nav-sidebar-forum').click();
+        cy.get('#nav-sidebar-collapse-sidebar').click();
+    });
+
+    it('Create, reply to, merge, and delete threads', () => {
+        // Comment
```
zacharymnp, line 57: The comments here on lines 57, 59, 61, 77 don't add any information and should be removed.

zacharymnp (changes requested):

ziesski (changes requested):

site/app/templates/forum/CreatePost.twig
```diff
@@ -100,8 +100,9 @@
...
-                <a class="post_button edit-post-button key_to_click" data-testid="edit-post-button" tabindex = "0" onClick="showEditPostForm({{ post.id }}, {{ post.thread_id }}, {{ post_buttons.edit.shouldEditThread }}, {{ render_markdown?1:0 }}, '{{ csrf_token }}')" title="{{ post_buttons.edit.edit_button_title }}" aria-label="{{ post_buttons.edit.edit_button_title }}"><i class="fas fa-edit"></i></a>
+                {# Removing the delete and edit icons as they are included in the dropdown now #}
+                {# <a class="post_button delete-post-button key_to_click" tabindex = "0" onClick="deletePostToggle({{ post_buttons.delete.ud_toggle_status }}, {{ post.thread_id }}, {{ post.id }}, '{{ post.author_user_id }}', '{{ post_date }}', '{{ csrf_token }}' )" title="{{ post_buttons.delete.ud_button_title }}" aria-label="{{ post_buttons.delete.ud_button_title }}"><i class="fa {{ post_buttons.delete.ud_button_icon }}"></i></a>
```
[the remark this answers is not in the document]
ZheyuDeng: Ye Barb didn't seem certain about removing the button so I forgot to remove haha

site/app/templates/forum/CreatePost.twig
ZheyuDeng: Ye I tried the button looks a bit odd

ziesski (approved):
LGTM

PR #10349, opened 2024-04-10
[UI/UX:InstructorUI] Add margin to error text

### LINKED ISSUE #10344: Autograding error text needs padding
**Describe the bug**
Autograding error messages currently appear next to the test case name with no padding.  This unpolished look could be improved significantly.

image

**Expected behavior**
The error message should have a cleaner look.

CCWizma99:
There is no left-padding between h4 element and error-message (span.red-message). That makes the UI unclean.

### What is the new behavior?
image
There is a 5px of left-padding and the UI looks better.
closes #10344 

### Other information?
Changed electronic.css -> added a new style to elements with red-message class where parent element is [#autograding_results .box-block] which is where the auto grading error is displayed. So, other files will not be affected that use "red-message" class.


RayReyRam (approved):
I approve, this fixes the issue where the red error message has no spacing with the text. This code also does not affect another red message.

Kush-Raval (approved):
Code makes sense, it achieves the desired effect.
Since it is a UI change, I would put an image displaying the current behavior to highlight the change made.

williamjallen (commented):
While this is an improvement over the existing design, I question whether this is the best possible UI design for these error messages.  Is there a better way we can display these error messages?

CCWizma99:
> While this is an improvement over the existing design, I question whether this is the best possible UI design for these error messages. Is there a better way we can display these error messages?

@williamjallen I think there could be a better way. Check below design and please suggest me changes...
### New design
image
image

### Wave Extension check
image
image

site/public/css/electronic.css
```diff
@@ -590,6 +590,22 @@ progress::-webkit-progress-value {
     border: 5px solid var(--standard-vibrant-pink-red);
 }
 
+/* stylelint-disable-next-line selector-id-pattern */
+#autograding_results .box-block .red-message {
+    background-color: #940000;
```
ziesski, line 595: Please use color from color.css to maintain uniform color scheme across entire system

ziesski (changes requested):

site/public/css/electronic.css
CCWizma99: Oh okay. I will. Thank you for the guidance. Do you think "--alert-danger-red" will look good on this?

image

image

williamjallen:
I like the general direction the new design is going.  It feels like further refinements could be made, but I'm struggling to come up with a cleaner design.  Perhaps @bmcutler has thoughts on the UI?

bmcutler:
@CCWizma99 
Thanks for playing around with this design, but as I told @williamjallen yesterday, I am hesitant to make any significant changes to the visual design of these messages.  They are used in different ways for different assignments.  Sometimes they are length and forced to wrap on narrow screens or with different layouts for displaying text differences between student & instructor output.

I don't want to merge any significant changes.  Adding a little space is ok, but that's all for now.

CCWizma99:
> Adding a little space is ok, but that's all for now.

Oh okay. Let's just add a little padding then.

justin-sadakhom (approved):

bmcutler:
I just tried to test this and I'm not sure what changed.
(I did the full install & even force cleared my cache data and still didn't see anything change)
Here is a screenshot:

Screenshot 2024-05-22 at 1 50 03 PM

Maybe I'm not testing specifically which red message changed.  But I was expecting the error for p3.out missing to be shifted to the right.

bmcutler (changes requested):
I was unable to successfully test this fix.

CCWizma99:
> I just tried to test this and I'm not sure what changed. (I did the full install & even force cleared my cache data and still didn't see anything change)

I'm sorry to hear that. I deleted the whole Submitty repository in my computer and cloned it again from main branch. Then installed it on Virtual Box. And I only made the changes I did on this issue. I got the results. But, I'm a little confused that why isn't it work on your computer.

image

micpap25:
The reason Barb wasn't able to replicate the results is that CC's changes were only made to the TA Autograding panel and not results section itself. I've fixed this and added a comment for future developers.

ZheyuDeng:
> I just tried to test this and I'm not sure what changed. (I did the full install & even force cleared my cache data and still didn't see anything change) Here is a screenshot:
> 
> Screenshot 2024-05-22 at 1 50 03 PM
> Maybe I'm not testing specifically which red message changed. But I was expecting the error for p3.out missing to be shifted to the right.

There is a margin left 5px right now.
image

ZheyuDeng (approved):
Tested locally as the screenshot above shows it works.

bmcutler (approved):

PR #10350, opened 2024-04-10
[Bugfix:InstructorUI] Improved score pill design

### LINKED ISSUE #10343: Yellow background for score pill has poor contrast in dark mode
**Describe the bug**
The yellow background of the score pill has poor contrast when viewed in dark mode.

image

**Expected behavior**
The pill color should pass the WAVE accessibility checker.  One option is to do a border with a bright color like the one used on the course materials page for the "new" indicator.

CCWizma99:
Score pills have a low contrast specially in Yellow background.

### What is the new behavior?
Added a new design to improve the contrast and look of score pills. 
Light mode will not be affected

Look of new design...
#### Dark mode - Black
image

#### Dark mode
image

#### Wave extension didn't show an error about score pills.
Dark mode - Black
image

Dark mode
image

Closes #10343 

### Other information?
Edited this 1st comment on 17th April, 2024 with newest updated view images. So, read next comments carefully... :slightly_smiling_face: 

Changed 2 files -> server.css and server.js

Even though there is no use of "dark-blue-background" in this case, I thought it could be useful in any way as there was an existing class style as blue-background.

Please suggest for changes if you have any...


ArnavPawar (changes requested):
Code looks good but I don't think this is necessary. I think I liked the old color scheme better. 
Also, this is inconsistent ex)
image

I will check if other people like this and let you know.

CCWizma99:
> Code looks good but I don't think this is necessary. I think I liked the old color scheme better. 
> Also, this is inconsistent ex)
> image
> 
> I will check if other people like this and let you know.

Actually, the changes were consistent. Can you please check your badge.twig has been changed correctly for auto-score-yellow. Also, I made the pull request because of an issue listed by a member of Submitty saying that design needs changes to improve contrast and pass the wawe extension test. In that issue the member suggested that using bright color borders would be a better option. That's why I chose it.

Thank you :)

arougey (changes requested):
Although this does address the issue, I feel like it could be done in a more eye appealing way. The current full background color based on your grade makes more sense to me than a thin outline.

bmcutler:
I'm going to second @ArnavPawar 
I prefer the original design that had more pixels/larger area of color.  It was easier to see which tests passed/failed.  I would prefer a solution that only updates the colors in dark mode and doesn't change the style.

CCWizma99:
> Although this does address the issue, I feel like it could be done in a more eye appealing way. The current full background color based on your grade makes more sense to me than a thin outline.

Mmm okay. I will revert back to old way. Can you do a suggestion please, how I should address the issue like change the color or something else?

CCWizma99:
> I'm going to second @ArnavPawar 
> I prefer the original design that had more pixels/larger area of color.  It was easier to see which tests passed/failed.  I would prefer a solution that only updates the colors in dark mode and doesn't change the style.

Okay. I think I got it. The task would be to keep the colors in light mode as it is and change the colors in dark mode and no change in style. Am I correct and May I use JS instead of twig macros?

CCWizma99:
@ArnavPawar , @arougey and @bmcutler ...

How's these colors (Only change in dark and black themes, light theme will not be affected.)?

image
image

I changed server.js and server.css. In server.js I defined a function to change color of span.badge.(green or yellow or red bg), if html tag contain data-theme='dark' or data-black_mode='black' attributes.

What should be changed :handshake:

arougey (approved):
I like the color scheme a lot more, if it has a good wave score then I approve. Also make sure to change the pictures in the initial post.

ziesski:
I have to disagree/adjust the color here. The current selection may conform to wave, it is too vivarent for dark mode. Let me use Github color scheme as good example. The saturation/chroma needs to be reduced. 
Please refer to below screenshots. 

스크린샷 2024-04-18 오전 1 11 09
스크린샷 2024-04-18 오전 1 12 22
스크린샷 2024-04-18 오전 1 12 45

ziesski:
I have to disagree/adjust the color here. The current selection may conform to wave, it is too vivarent for dark mode. Let me use Github color scheme as good example. The saturation/chroma needs to be reduced. 
Please refer to below screenshots. 

스크린샷 2024-04-18 오전 1 11 09
스크린샷 2024-04-18 오전 1 12 22
스크린샷 2024-04-18 오전 1 12 45

CCWizma99:
@ziesski, mm okay. I can change it to a lower saturation. But, may I wait for a maintainer to change the colors as they prefer? If there should be a change in logics or methods, I will do it.

site/public/css/server.css
```diff
@@ -763,6 +763,26 @@ td > a.active-sort {
     background-color: var(--standard-medium-blue);
 }
 
+.dark-green-background{
+    background-color: #00ff11;
```
Kush-Raval, line 767: Use colors defined in colors.css to ensure that these changes match the pre-existing color scheme of submitty.

Kush-Raval (changes requested):
The code changes make sense, just apply the comment I made regarding defining new colors.

williamjallen:
Just to add my personal opinion to the mix, I personally like colored outlines in dark mode rather than bright colors.  We did something similar on the course materials page which might be a good idea here:
image
image

The colors of the grading scores are part of the Submitty brand, so getting it right is very important.

site/public/css/server.css
CCWizma99: Okay. I will change the colors to colors from colors.css.

ZheyuDeng (commented):
Thank you for your contribution!
I have switched the variable to color from color.css and fixed a small typo in your code. We also agreed to switch to less saturated colors during the meeting.
image

ziesski (approved):
I conform this change

zacharymnp (approved):
looks good

williamjallen:
@ZheyuDeng Can you update the screenshots in the PR description?

PR #10351, opened 2024-04-12
[Bugfix:TAGrading] Fix deleting all comments

cjreed121:

As a grader, if you clear an overall comment, it will clear for all other graded submissions.

### What is the new behavior?
If you clear an overall comment, it will only clear that submission's overall comment. A Cypress test was added to confirm this behavior.

### DIFF
--- a/site/app/libraries/database/DatabaseQueries.php
+++ b/site/app/libraries/database/DatabaseQueries.php
@@ -7222,6 +7222,12 @@ AND gc_id IN (
         $this->course_db->query($query, $params);
     }
 
+    /**
+     * @param string $gradeable_id
+     * @param string $grader_id
+     * @param string $submitter_id
+     * @return void
+     */
     public function deleteOverallComment($gradeable_id, $grader_id, $submitter_id) {
         $this->course_db->query(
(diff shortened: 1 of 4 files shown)


DarthNyan (approved):
Looks good to me. I tested it and it works.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -7222,8 +7222,17 @@ private function updateOverallComment(TaGradedGradeable $ta_graded_gradeable, $c
...
 
-    public function deleteOverallComment($gradeable_id, $grader_id) {
-        $this->course_db->query("DELETE FROM gradeable_data_overall_comment WHERE g_id=? AND goc_grader_id=?", [$gradeable_id, $grader_id]);
+    /**
+     * @param string $gradeable_id
+     * @param string $grader_id
+     * @param string $submitter_id
+     * @return void
+     */
+    public function deleteOverallComment($gradeable_id, $grader_id, $submitter_id) {
```
williamjallen, line 7231: It would be better to use native PHP type annotations if possible, since they are enforced by the interpreter itself, as opposed to PHPDoc comments which are only suggestions to the reader.

PR #10356, opened 2024-04-14
[Bugfix:Submission] Redirect out of submission if no team

### LINKED ISSUE #9566: disallow access to team gradeable page if you're not on a team
**Describe the bug**
As instructor, for a team gradeable, create a team, then go to the submission page for a team.

In another browser window, remove the instructor from the team (either have the instructor "leave team" or from the ta grading page leave the team).

Go back to the original browser window and hit reload.  The instructor incorrectly still has access to the submission page for that gradeable -- page rendering might be buggy / have errors and/or allow instructor to attempt to make a submission.

**Expected behavior**

Any user attempting to (re)load the submission page for a team gradeable when they aren't on a team should be booted out of the page and returned to the home page / main gradeables page for the course.

note: i think this doesn't happen for student users.  but make sure all users are treated the same and access is disallowed if not currently on a team.

related to #9494

GO-viper7:
Closes #9566 

Instructor still has access to submission page for a team-gradeable even after team has been abandoned

### What is the new behavior?
Instructor or any other user gets booted out of submission page on reload, if not present in a team

### To Test : 

- Open two browsers, login as same instructor in both
- In first browser, create a team and access a team-gradeable's submission page
- In second, go to manage teams and leave the team
- Come back to first and reload the page
- You can observe that the user gets redirected out of submission page with an alert saying "Must be on a team to access submission."

### DIFF
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -90,8 +90,9 @@ class SubmissionController extends AbstractController {
             $this->core->getOutput()->renderOutput('Error', 'noGradeable', $gradeable_id);
             return ['error' => true, 'message' => 'No gradeable with that id.'];
         }
-        elseif ($gradeable->isTeamAssignment() && $graded_gradeable === null && !$this->core->getUser()->accessAdmin()) {
-            $this->core->addErrorMessage('Must be on a team to access submission');
+        // Irrespective of the access level, user should be redirected out of page if not present in a team
+        elseif ($gradeable->isTeamAssignment() && $graded_gradeable === null) {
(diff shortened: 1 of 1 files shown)


RayReyRam (approved):
This works, as instructor when leaving the team and refreshing it kicks me out and alerts that message, it also works when trying to submit something either, code looks good, removed the part where if they are admin they would have access.

williamlin6803 (approved):
Code changes looks good and the functionality works on my machine.

williamjallen (approved):
Changes seem reasonable to me based on the details in the linked issue.  It would be good for @bmcutler to verify that this is the intended behavior change.

bmcutler (approved):

PR #10360, opened 2024-04-16
[Refactor:Autograding] Remove system program check for docker

cjreed121:
Only existing programs in the allowed list can be run in the jailed sandbox or in a container.

### What is the new behavior?
If we are grading inside docker, any command/program is allowed to be run. Docker gives us a more secure environment so we don't need to be as restrictive on what we can run.

### Other information?
Tested locally that this does allow other programs


thegame615 (approved):
Seems good to me. Just don't know if this is what Barb wants, but I'm assuming you checked with her.

grading/execute.cpp
```diff
@@ -70,23 +62,23 @@ bool system_program(const std::string &program, std::string &full_path_executabl
...
-        return true;
-      }
+      full_path_executable = program_obj["path"];
+      return true;
     }
   }
+  // if we are docker, allow any program. previous code is necessary
+  // to run before return for full path expansion.
+  if (running_in_docker) {
+    return true;
```
DarthNyan, line 80: It looks like full_path_executable isn't getting set here. When I tested, this was causing validate_program to return an empty string and the assertion on line 513 to fail in execution.

DarthNyan (changes requested):

micpap25 (approved):
Tested, looks all good to me!

Rerun the tests at some point, the failing PR is flaky.

PR #10363, opened 2024-04-16
Grading refactor @route php8

arougey:
Refactors all of the @Routes to the new PHP8 Syntax of #[Route()] for all grading files.


Kush-Raval (approved):
I looked at the code, and it successfully refactored the @route to the new PHP syntax.

williamjallen (approved):

PR #10383, opened 2024-04-22
[Refactor:TAGrading] speed up rubric grading page

wuj18:

On the TA grading rubric page, toggling components (switching to the next question for grading) takes a long time to load. This slowdown occurs because:
toggleComponent() consists of two steps: closeAllComponents() and openComponent()
- Closing all components: Attempts to call closeComponent() on all components. Each call to closeComponent() waits for the server to save and then calls updateTotals() to update the total score. updateTotals() rerenders the entire gradeable component, which is time-consuming.
- Opening a component: It waits for closeAllComponents() to finish, then waits again for the server to retrieve the component.

### What is the new behavior?
With the implemented changes:
- Component information is now stored in a global list, so open component can render using info from the list
- closeAllComponents() only calls closeComponent() on components that are actually open.
- Changes are saved in the background instead of waiting.
- updateTotals() has been replaced with refreshTotalScoreBox(), which only rerenders the total score box at the bottom.

### Other information?
Changes were made only to the normal grading and edit mode paths. I couldn't figure out what the instructor edit mode is so I rewrote the code in a way that the instructor edit path will still follow the old workflow. 

Potential Issue with the Change: If two people are grading the same student and both load the page simultaneously, there's a possibility of data being overwritten. For example, if Person 1 grades the first question and Person 2, without refreshing their page, grades the same question afterward, the later input will overwrite the earlier one. However, I don't consider this a significant downside because graders are usually assigned to different students or questions, making such overlaps unlikely. Additionally, similar overwrite issues could occur even before the change when two graders opened the same component simultaneously.

For me, it is a significant improvement. My computer is old and slow, so the response time has improved dramatically, from nearly 3 seconds to under 0.1 seconds.
before:
8d3e0670ce8c373d3bcc04ccb66942c
after:
7aa862038bb7961f899bddebc65927c

try to grade a student and see if anything breaks

### DIFF
--- a/site/public/js/ta-grading-rubric.js
+++ b/site/public/js/ta-grading-rubric.js
@@ -2319,7 +2319,7 @@ function reloadGradingRubric(gradeable_id, anon_id) {
         })
         .then((graded_gradeable) => {
             GRADED_GRADEABLE = graded_gradeable;
-            return loadComponentData(gradeable_tmp.components, gradeable_id, anon_id)
+            return loadComponentData(gradeable_tmp, graded_gradeable)
                 .then(() => {
                     return renderGradingGradeable(getGraderId(), gradeable_tmp, graded_gradeable,
                         isGradingDisabled(), canVerifyGraders(), getDisplayVersion());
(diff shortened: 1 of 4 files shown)


cjreed121 (commented):
Code looks fine and tested locally which worked. Couple requested changes. Also make sure to remove the time console.log's

site/public/js/ta-grading-rubric.js
```diff
@@ -18,6 +18,10 @@
  * Global variables.  Add these very sparingly
  */
 
+const GRADED_COMPONENTS_LIST = {};
+const COMPONENT_RUBRIC_LIST = {};
+let GRADED_GRADEABLE = Promise.resolve();
```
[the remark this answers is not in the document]
wuj18: This is a global variable that would store the gradable when reloadGradingRubric is called

site/public/js/ta-grading-rubric.js
```diff
@@ -1304,12 +1308,13 @@ function getGradedComponentFromDOM(component_id) {
 function getScoresFromDOM() {
     const dataDOMElement = $('#gradeable-scores-id');
     const scores = {
-        ta_grading_complete: getTaGradingComplete(),
+        user_group: GRADED_GRADEABLE.user_group,
```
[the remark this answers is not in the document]
wuj18: user_group needs to be passed into the TotalScoreBox.twig to get a correct total score rerender

site/public/templates/grading/Functions.twig
```diff
@@ -3,22 +3,17 @@ Function for getting the badge style for a given set of total and earned points
 #}
 {% macro getBadgeStyle(earned, total, complete) %}
```
cjreed121, line 4: Make sure to remove complete and also from the place it gets called

cjreed121 (changes requested):

site/public/templates/grading/Functions.twig
```diff
@@ -1,24 +1,19 @@
 {#
 Function for getting the badge style for a given set of total and earned points
 #}
-{% macro getBadgeStyle(earned, total, complete) %}
```
cjreed121, line 4: I didn't realize there was more usages of this. It might actually be better to keep this complete variable and then set it to null in the one case where it's called where we don't define `ta_grading_complete` anymore. And then in here check if it's null rather than undefined. Unless you have tested and verified this macro still works as expected without that last parameter?

cjreed121 (changes requested):

site/public/templates/grading/Functions.twig
williamjallen: That parameter doesn't appear to be used anywhere in the codebase.

cjreed121 (approved):
Code looks good

williamjallen (approved):
Code looks good, and I verified that the changes function for regular, peer, and team gradeables.

PR #10392, opened 2024-04-26
[Bugfix:Forum] Upduck counter wrap to next line bug

### LINKED ISSUE #10367: Upduck number wraps to next line on mobile if post is edited
**Describe the bug**

If the post was edited, it makes the line slightly too big, and forced the number of upducks to go onto the next line on a mobile device.

**Expected behavior**

They should all stay in one line.

**To Reproduce**
1. Find a post that was edited
2. Visit it

**Configuration**
 - OS: [e.g. Windows, Linux, ...] iOS
 - Browser [e.g. Chrome, Safari] Chrome

**Screenshots**

IMG_1375

**Additional context**

CCWizma99:
When a user edited a post in forum or user's display name is too long (for smaller screen sizes), Upduck counter will wrap to the next line. But, other elements (even the Upduck button doesn't.) don't go to next line.

image

### What is the new behavior?
Nested the Upduck button and counter inside a span and again nested it inside span.post-action-container. Now if the width exceed the space, whole span.post-action-container will go to the next line. Borders are optional. Added it to improve the UI.

#### Dark mode
image
#### Light mode
image

closes #10367 

### Other information?
I think there is no potential issue of nesting Upduck button and counter inside a span. Button and counter are working correctly.


site/public/css/forum.css
```diff
@@ -230,18 +230,35 @@
...
 }
 
+
 .post-email-toggle,
 .post-user-info {
     margin-right: 2px;
     display: inline-block;
 }
 
+.upduck-container{
```
micpap25, line 249: I'm not sure how I feel about the whole span being presented as a "button", especially since only the duck is clickable. I also feel that the design is a bit at odds with the rest of the post actions in terms of design.

micpap25 (commented):
Looks like a good start to me, just some thoughts on the design.

zacharymnp (approved):

zacharymnp:
image

Upduck counter/button remains on the first line while other text wraps around.

PR #10401, opened 2024-04-27
[Bugfix:System] Preserve submitty config keys

### LINKED ISSUE #8151: System Message clobbered by running CONFIGURE_SUBMITTY.sh
**Describe the bug**

If the "system_message" variable is stored in the /usr/local/submitty/config/submitty.json file, that value is lost when CONFIGURE_SUBMITTY.sh is run.

See also:
https://submitty.org/sysadmin/installation/system_customization#show-system-message-to-all-users

thegame615:
Currently, when the CONFIGURE_SUBMITTY.py script is run, the system_message field located in /usr/local/submitty/config/submitty.json is not preserved. This can be annoying for sysadmins who don't want to have to rewrite their system_message everytime after they run CONFIGURE_SUBMITTY.py.
Closes #8151 
### What is the new behavior?
If the submitty.json file has defined a system_message, field, it is saved into the config of the new file.

### Other information?
You can test this by running the CONFIGURE_SUBMITTY.py script located in /usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/CONFIGURE_SUBMITTY.py

Please try all sorts of edge cases. 
Examples:
Delete the submitty.json file and run
Change the field and run and make sure that it is preserved
leave the field empty
NOTE: we are assuming that the file is valid json, that is the user's responsibility

### DIFF
--- a/.setup/CONFIGURE_SUBMITTY.py
+++ b/.setup/CONFIGURE_SUBMITTY.py
@@ -494,6 +494,17 @@ if not args.worker:
 
 IGNORED_FILES_AND_DIRS = ['saml', 'login.md']
 
+system_message = None
+# this prevents clobbered system_message
+try:
+    with open(SUBMITTY_JSON,"r") as submitty_config:
+        load_submitty_json = json.load(submitty_config)
+        if "system_message" in load_submitty_json:
+            system_message = load_submitty_json["system_message"]
+except IOError:
+    pass
+
+
 if os.path.isdir(CONFIG_INSTALL_DIR):
     for file in os.scandir(CONFIG_INSTALL_DIR):
         if file.name not in IGNORED_FILES_AND_DIRS:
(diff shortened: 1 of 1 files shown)


justin-sadakhom (commented):
Overall looks good to me but a minor nitpick I have is to make sure you're consistent with your quotes. You mix single and double quotes; choose one and stick to it. I know the style in the file is all over the place but it's good to enforce consistency.

ziesski (approved):
made minor edit about consistent quotation from previous reviewer. Tested the functionality and reviewed the code.

.setup/CONFIGURE_SUBMITTY.py
```diff
@@ -494,6 +494,17 @@ def write(x=''):
 
 IGNORED_FILES_AND_DIRS = ['saml', 'login.md']
 
+system_message = None
+# this prevents clobbered system_message
+try:
+    with open(SUBMITTY_JSON,'r') as submitty_config:
+        load_submitty_json = json.load(submitty_config)
+        if 'system_message' in load_submitty_json:
```
skara9, line 502: Here we're giving special importance to the 'system_message' key.
What we can do instead is use the existing JSON as the base for the new data, which means any key that we don't change will not get deleted.

skara9 (changes requested):
Changes work as intended and code looks good, just one suggestion for improvement.

.setup/CONFIGURE_SUBMITTY.py
ziesski: I agree that in general we want to keep everything, but if we take a look at Config.php where these json tags are parsed and used, system_message is the only significant key that needs special importance.
I think that is what the initial issue expects to address. Hence, the request is beyond the scope of original issue. 
You could make it an additional issue and mention this PR instead.

skara9: (Discussed in meeting, @bmcutler requested for all keys to be preserved. PR scope was updated to meet this.)

ziesski (approved):
Works as expected, keeping all data, the code is good too.

PR #10426, opened 2024-05-06
[Bugfix:InstructorUI] Fix Rainbow Grades URL

oliiso:

This page has an invalid URL.
### What is the new behavior?
It has been updated to the correct URL.


IDzyre (approved):
Fixes the error.

PR #10446, opened 2024-05-16
[Feature:API] Add testcases to student api

IDzyre:

Currently, the student api only returns the total grade points, and doesn't include which of the tests the points came from. 

### What is the new behavior?
This adds the testcases the user can view to the return values, giving more information on what failed, and why. 
API tests will be converted to Jest in a future PR.

### DIFF
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -1052,6 +1052,30 @@ class SubmissionController extends AbstractController {
 
         $graded_gradeable = $graded_gradeable->getAutoGradedGradeable();
 
+        $testcase_array = [];
+        if ($graded_gradeable->hasActiveVersion()) {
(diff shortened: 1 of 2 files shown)


justin-sadakhom:
How did you test the API?

justin-sadakhom (commented):
Please also fix the PHP linting.

IDzyre:
Since this addition to the API is regarding the graded component of gradeables, I don't think it is currently able to be tested VIA the CI, unless we do some hacky thing like create a gradeable, submit to it, wait for it to be graded, then request it VIA the API, however I can create a test that runs only locally for now. I use Postman to call the API locally.

justin-sadakhom:
> Since this addition to the API is regarding the graded component of gradeables, I don't think it is currently able to be tested VIA the CI, unless we do some hacky thing like create a gradeable, submit to it, wait for it to be graded, then request it VIA the API, however I can create a test that runs only locally for now. I use Postman to call the API locally.

If it's possible to add a test case that can be run locally but is ignored by GitHub Actions please do that. Thank you!

Issue #10162, opened 2024-02-05
Label original author in forum replies


williamjallen:
**What problem are you trying to solve with Submitty**
When a student makes a post anonymously, it is currently impossible to determine if the same student is the author of a subsequent reply.  This can be confusing in some cases when follow-up questions are asked by multiple anonymous students.

**Describe the way you'd like to solve this problem**
This problem can be solved easily by adding a tag or other visual indicator which labels the student as the author of the original thread.

Xiayucheng1212:
How about change the signature from "Anonymous" to "Original Poster"?

bmcutler:
@MasterOdin 

We discussed this a bit in person and the Submitty and decided it is sufficient (for now) to only label the "OP" anonymous. It is unusual to have a lengthy anonymous back and forth with multiple anonymous users education forum where it is necessary to uniquely identify more than OP.

This is also related to "Live Lecture Chat" -- where we will need to worry about clearly labeling many anonymous users.
https://github.com/Submitty/Submitty/issues/7207

PR #10447, opened 2024-05-17
[Bugfix:Forum] Remove OP for anonymous author

### LINKED ISSUE #10381: Discussion forum anonymized OP
If someone posts anonymously on the discussion forum, but responds to the comment non-anonymously, the post and comment are inconsistently anonymous.

The `OP` marker should be disabled for non-anonymous replies when the initial post is made anonymously in order to keep the post anonymized.

**Additional context**
An example post where the OP is anonymous but is de-anonymized by commenting:
non-anonymous op

Xiayucheng1212:
fixes #10381

### What is the new behavior?
Only show the "OP" marker when the author didn't post anonymously. 

### Other information?
Add a new variable 'first_post_anonymous' to judge whether the marker is available.

The edit icon shown in this image demonstrates the first and third author are the same person and there is no 'OP' marked.

截圖 2024-05-17 下午12 10 08

### DIFF
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -540,3 +540,30 @@ jobs:
       - uses: ./.github/actions/e2e-Setup-Composite
       - name: Run integration tests
         run: sudo -E env "PATH=$PATH" python3 /usr/local/submitty/test_suite/integrationTests/run.py
+
+
+  ansible-install:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup SSH
+        shell: bash
+        run: |
+          ssh-keygen -t rsa -b 4096 -C "github_ci" -N '' -f /home/runner/.ssh/id_rsa
+          ssh-keyscan -t rsa localhost >> /home/runner/.ssh/known_hosts
+          cat /home/runner/.ssh/id_rsa.pub >> /home/runner/.ssh/authorized_keys
+          ssh -T localhost
+          sudo systemctl start postgresql
+
+      - name: Run ansible script
+        shell: bash
+        run: |
+          cd .setup/ansible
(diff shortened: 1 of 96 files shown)


cjreed121 (commented):
Don't forget this instance of createPost being called: https://github.com/Submitty/Submitty/blob/802ce7e74f3a686776be0fb521a949c41e065d61/site/app/controllers/forum/ForumController.php#L595-L614

justin-sadakhom (changes requested):
Please fix the [failing Cypress tests](https://github.com/Submitty/Submitty/actions/runs/9122816284/job/25084230121?pr=10447#step:7:258).

justin-sadakhom (approved):

williamschen23:
Should there be an option to prompt the original author if they want the OP tag? Sometimes it is useful to know that the person you are dealing with is the person who made the thread on the problem in the first place.

zacharymnp (changes requested):
image
image
image

Most anticipated behavior works correctly as pictured above. However, when OP makes an anonymous reply to a non-anonymous post, their reply is marked with OP, which is not ideal.

image

zacharymnp (approved):
Requested change was implemented:

image

bmcutler:
Thanks @zacharymnp for doing the 4 cases so clearly.

But, anonymous post + anonymous reply by same person SHOULD be labeled OP.  
That's the whole point of the original issue.  #10162 

The only time we don't want to label OP is when it deanonymizes a user because the user has inconsistent usage of the anonymous option.

bmcutler (changes requested):
see note above

ZheyuDeng:
Should be good now. Simplified condition.
is_OP = = ($post["author_user_id"] === $first_post_author_id) && ($first_post_anonymous === $post["anonymous"])
e234bbb64230e2c3814909915f5263e
134fa91b9c9425ce265f7013efd9246

zacharymnp (approved):
changes work

bmcutler (approved):

PR #10452, opened 2024-05-21
[Dependency] Pin requests version to 2.31.0

IDzyre:
Docker-py breaks with the current version of requests (2.32.0)

### What is the new behavior?
This PR pins requests version to 2.31.0, which will fix Docker until a hotfix is applied to docker-py.


IDzyre:
Would Dependabot update this since it is in the requirements file?

MasterOdin:
Yes, though you can just have it ignore the version by closing the PR or by modifying the dependabot configuration ([docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore)). I'd probably opt for the former as when that new version of `requests` is made, should be possible to unpin this dependency at that point.

PR #10465, opened 2024-05-25
[Feature:Submission] View submitted file button

skara9:

Currently, students can access course materials in two ways: download or view in browser.
However, for files submitted to a gradeable, students can only download the file if they want to read it.
This can make it annoying for students trying to verify that the file they submitted is the correct one,
or verifying that their answer to a problem was correct.
Also, downloading the file across different versions clutters the user's downloads folder with multiple
different versions of the same file.

### What is the new behavior?
A "view" button has been added to the gradeable submission view, allowing students to view a pop up window displaying a submitted file in their browser, parallel to the current grader view functionality.

image

### DIFF
--- a/site/app/templates/submission/homework/CurrentVersionResults.twig
+++ b/site/app/templates/submission/homework/CurrentVersionResults.twig
@@ -16,11 +16,18 @@
                         <span>
                             {{ file.relative_name }} ({{ (file.size / 1024) | number_format(2) | default(-1) }}kb)
                         </span>
-                    {# download icon if student can download files #}
-                    {% if student_download %}
-                        <button class = 'btn btn-primary key_to_click' onclick='downloadFile("{{ file.path|url_encode }}", "submissions")' aria-label="Download {{file.relative_name}}"> Download
-                            <i class="fas fa-download" title="Download the file"></i></button>
-                    {% endif %}
+                        {# view and download icons if student is permitted to access files #}
+                        {% if student_download %}
+                            <div>
+                                {% set extension = file.relative_name|split('.')|last|lower %}
+                                {% if '.' ~ extension in ['.pdf', '.jpg', '.jpeg', '.c', '.cpp', '.s', '.twig', '.py', '.java', '.png', '.txt', '.h', '.html', '.php', '.js', '.sql', '.sh', '.md', '.csv', '.salsa', '.erl', '.oz', '.pl', '.hs', '.gif'] %}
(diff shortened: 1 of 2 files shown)


justin-sadakhom (commented):
The project's pull request guidelines specify you should use light mode (default theme) for images. Please update the PR to reflect this.

williamjallen (commented):
Is a popup window the best user experience we can provide here?  Could it be better to simply use ` target="_blank"` to open the file in another browser tab/window?

Another thing to keep in mind is security.  Displaying an HTML file in the browser can be dangerous if not handled properly, for example.  cc: @cjreed121

skara9:
@williamjallen For course materials, the separate tab is preferable, because we want the student to be able to keep it open and read through the whole document as needed without having to go back to the course material list.

But for viewing a submitted file, a student does not need to keep it open or read the whole document, since it's something they uploaded themselves. They only need to check if they submitted the right file, and close the popup as soon as they are done. So for this, a popup is more user-friendly and limits clutter (especially if they are checking different versions of the same file).

And as far as security goes, this PR makes no changes to the backend, nor does it add any rendering code to the JavaScript. It only adds a button to the student view to call functionality that was already present in the grader view.
If that code has no security concerns, then this shouldn't either.

site/app/templates/submission/homework/CurrentVersionResults.twig
```diff
@@ -16,11 +16,18 @@
...
-                    {# download icon if student can download files #}
-                    {% if student_download %}
-                        <button class = 'btn btn-primary key_to_click' onclick='downloadFile("{{ file.path|url_encode }}", "submissions")' aria-label="Download {{file.relative_name}}"> Download
-                            <i class="fas fa-download" title="Download the file"></i></button>
-                    {% endif %}
+                        {# view and download icons if student is permitted to access files #}
+                        {% if student_download %}
+                            <div>
+                                {% set extension = file.relative_name|split('.')|last|lower %}
+                                {% if '.' ~ extension in ['.pdf', '.jpg', '.jpeg', '.c', '.cpp', '.s', '.twig', '.py', '.java', '.png', '.txt', '.h', '.html', '.php', '.js', '.sql', '.sh', '.md', '.csv', '.salsa', '.erl', '.oz', '.pl', '.hs', '.gif'] %}
```
zacharymnp, line 23: You shouldn't need to concatenate the '.' to extension if you remove from '.' each item in the list of acceptable extensions.

skara9: That code is copied over from `CourseMaterials.twig`; I didn't change anything to keep it consistent.
If we want to refactor it we should do it in both places.

zacharymnp (approved):

bmcutler (approved):
tested w/ directories as well, looks good :)

PR #10467, opened 2024-05-26
[Bugfix:InstructorUI] Bulk upload not showing options

### LINKED ISSUE #10106: Bulk Upload UI requires toggle to display exam page count form

ziesski:

fixes #10106
this PR replaces #10108 (wrong fix)
Initially instructor creates a new gradeable "TA/Instructor will (bulk) upload scanned .pdf for online manual grading" (a bulk pdf / scanned exam gradeable)

When instructor first enters to "BULK UPLOAD" from the gradeables page, they see this on the submission/upload page

### What is the new behavior?

Toggling is not necessary when you first load the page.

스크린샷 2024-05-27 오후 10 51 10

### DIFF
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -96,21 +96,21 @@
 		            <input type="text" id= "user_id" value ="" placeholder="{{ user_id }}"/>
                 </div>
             </div>
-            <div id="pdf-submit-button" class="sub">
+            <div id="pdf-submit-button" class="sub hide">
                 <div class="sub">
 
                     <div id="toggle-qr-split">
                         <label for="use-qr">Split by QR code?</label>
                         <input type="checkbox" id="use-qr">
                         {% if feature_flag_enabled('submitty_ocr') %}
-                        <span id="toggle-id-scan">
+                        <span id="toggle-id-scan" class="hide">
                             <label id ="use-ocr-id"for ="use-ocr">Detect Student ID's?</label>
                             <input type="checkbox" id="use-ocr">
                         </span>
                         {% endif %}
                     </div>
 
-                    <div id="qr-split-opts">
+                    <div id="qr-split-opts" class="hide">
                         <br>
 
                         <span id="prefix-prompt">QR code prefix/suffix:
@@ -540,39 +540,4 @@
         });
     });
     initializeTimer('{{ gradeable_id }}', {{ is_timed }});
-</script>
-
-
-<script>
-    window.onload = function() {
-        // capture radio elements
-        let radioBulk = document.getElementById('radio-bulk');
-        let radioNormal = document.getElementById('radio-normal');
-        let radioStudent = document.getElementById('radio-student');
-        let numericUploadOpts = document.getElementById('numeric-split-opts');
-        let qrUploadOpts = document.getElementById('qr-split-opts');
-
-        // capture the div elements
-        let pdfSubmitButton = document.getElementById('pdf-submit-button');
-
-        // add change event listeners to radio buttons
-        radioBulk.addEventListener('change', togglePdfSubmitButton);
-        radioNormal.addEventListener('change', togglePdfSubmitButton);
-        radioStudent.addEventListener('change', togglePdfSubmitButton);
-
-        function togglePdfSubmitButton() {
-            if (radioBulk.checked) {
-                // if Bulk Upload is checked, show the div
-                pdfSubmitButton.style.display = 'block';
-                qrUploadOpts.style.display = 'none';
-
-            } else {
-                // if other option is checked, hide the div
-                pdfSubmitButton.style.display = 'none';
-            }
-        }
-
-        // call the function on page load to set initial state
-        togglePdfSubmitButton();
-    };
 </script>
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -102,6 +102,7 @@
                     <div id="toggle-qr-split">
                         <label for="use-qr">Split by QR code?</label>
                         <input type="checkbox" id="use-qr">
+
                         {% if feature_flag_enabled('submitty_ocr') %}
                         <span id="toggle-id-scan" class="hide">
                             <label id ="use-ocr-id"for ="use-ocr">Detect Student ID's?</label>
@@ -541,3 +542,5 @@
     });
     initializeTimer('{{ gradeable_id }}', {{ is_timed }});
 </script>
+
+
--- a/site/app/templates/submission/homework/SubmitBox.twig
+++ b/site/app/templates/submission/homework/SubmitBox.twig
@@ -96,13 +96,12 @@
 		            <input type="text" id= "user_id" value ="" placeholder="{{ user_id }}"/>
                 </div>
             </div>
-            <div class="sub hide" id="pdf-submit-button">
+            <div id="pdf-submit-button" class="sub">
                 <div class="sub">
 
                     <div id="toggle-qr-split">
                         <label for="use-qr">Split by QR code?</label>
                         <input type="checkbox" id="use-qr">
-
                         {% if feature_flag_enabled('submitty_ocr') %}
                         <span id="toggle-id-scan">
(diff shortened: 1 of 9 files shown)


site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -544,3 +543,36 @@
 </script>
 
 
+<script>
```
williamjallen, line 546: This should be moved to an external JS file so it gets linted.

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -544,3 +543,36 @@
...
+                qrUploadOpts.style.display = 'none';
+
+            } else {
+                // if other option is checked, hide the div
+                pdfSubmitButton.style.display = 'none';
+            }
+        }
+
+        // call the function on page load to set initial state
+        togglePdfSubmitButton();
```
williamjallen, line 576: It would be better to set the initial state via Twig, to prevent a flash as the JS runs initially.

williamjallen (changes requested):

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -544,3 +543,36 @@
 </script>
 
 
+<script>
+    window.onload = function() {
```
shailpatels, line 547: You may want to move this logic into this file https://github.com/Submitty/Submitty/blob/main/site/ts/grader-submission.ts#L23 which handles a lot of UI controls for bulk upload already, and some of the logic added here might be duplicated there / attempted to change the style as well

shailpatels (changes requested):

site/app/templates/submission/homework/SubmitBox.twig
ziesski: yes, thanks for your advice, initially I was editing mentioned ts file, but then I got lazy with some trivial syntax, which I admit is my fault.

ziesski:
I did not put this is need review project board, but getting some feedback which is thankful (and insightful) but not needed at this moment, I was just exploring several different options, and am aware of mentioned critiques. For now I am switching it to draft.

site/ts/grader-submission.ts
```diff
@@ -20,7 +20,26 @@ declare global {
 
 const warning_banner = document.getElementById('submission-mode-warning');
 
+function initialsubmissionmode() {
+    const radioBulk = document.getElementById('radio-bulk') as HTMLInputElement;
+    const pdfSubmitButton = document.getElementById('pdf-submit-button');
+
+    if (radioBulk.checked) {
```
IDzyre, line 27: Not all gradeables have the bulk upload option right? 

Would checking if this is not null or something like that fix the cypress ?

ziesski: This was the problem. Good catch thanks :)

site/ts/grader-submission.ts
```diff
@@ -20,7 +20,50 @@ declare global {
...
+//         // If 'radio-bulk' is checked, show the submit button
+//         pdfSubmitButton!.style.display = 'block';
+//         sessionStorage.setItem(`${window.gradeable_id}-submission_mode`, 'bulk-upload');
+//         const message = 'Warning: Submitting files for bulk upload!';
+//         warning_banner!.firstChild!.textContent = message;
+//     }
+//     else {
+//         pdfSubmitButton!.style.display = 'none';
+//     }
+// }
```
williamjallen, line 37: Delete?

site/ts/grader-submission.ts
```diff
@@ -20,7 +20,50 @@ declare global {
...
+    if (radioBulk !== null) {
+        if (radioBulk.checked) {
+            if (pdfSubmitButton !== null) {
+                pdfSubmitButton.style.display = 'block';
+            }
+            sessionStorage.setItem(`${window.gradeable_id}-submission_mode`, 'bulk-upload');
+            const warning_banner = document.getElementById('warning-banner');  // ensure warning_banner is defined
+            if (warning_banner !== null && warning_banner.firstChild !== null) {
+                const message = 'Warning: Submitting files for bulk upload!';
+                warning_banner.firstChild.textContent = message;
```
williamjallen, line 52: It would be better to do this with jQuery.

site/ts/grader-submission.ts
```diff
@@ -20,7 +20,50 @@ declare global {
...
+            }
+        }
+        else {
+            if (pdfSubmitButton !== null) {
+                pdfSubmitButton.style.display = 'none';
```
williamjallen, line 57: You can use jQuery's `.hide()` method for this.

site/ts/grader-submission.ts
```diff
@@ -143,6 +186,7 @@ function changeSubmissionMode(event: Event) {
                 useQRCheckBox.click();
             }
             else {
+                qrUploadOpts!.style.display = 'none';
```
williamjallen, line 189: Same here.

williamjallen (changes requested):

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -96,22 +96,21 @@
 		            <input type="text" id= "user_id" value ="" placeholder="{{ user_id }}"/>
                 </div>
             </div>
-            <div class="sub hide" id="pdf-submit-button">
+            <div id="pdf-submit-button" class="sub hide">
```
williamjallen, line 99: This change isn't necessary and should be reverted.

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -542,5 +541,3 @@
     });
     initializeTimer('{{ gradeable_id }}', {{ is_timed }});
 </script>
-
-
```
williamjallen, line 546: This whitespace change should be reverted.

site/app/templates/submission/homework/SubmitBox.twig
```diff
@@ -96,22 +96,21 @@
...
                 </div>
             </div>
-            <div class="sub hide" id="pdf-submit-button">
+            <div id="pdf-submit-button" class="sub hide">
                 <div class="sub">
 
                     <div id="toggle-qr-split">
                         <label for="use-qr">Split by QR code?</label>
                         <input type="checkbox" id="use-qr">
-
```
williamjallen, line 105: same here

site/ts/grader-submission.ts
```diff
@@ -20,7 +20,37 @@ declare global {
...
+            if (warningBanner.length && warningBanner.children().length) {
+                const message = 'Warning: Submitting files for bulk upload!';
+                warningBanner.children().first().text(message);
+            }
+        }
+        else {
+            if (pdfSubmitButton.length) {
+                pdfSubmitButton.hide();
+            }
+        }
```
williamjallen, line 45: This can be combined into a single elseif:
```suggestion
        else if (pdfSubmitButton.length) {
            pdfSubmitButton.hide();
        }
```

williamjallen (changes requested):

site/app/templates/submission/homework/SubmitBox.twig
ziesski: This change was to have consistency with the rest of class= hide applied below.

williamjallen:
As discussed in my comment [here](https://github.com/Submitty/Submitty/pull/10476#issuecomment-2153228479), it looks like we have two very similar PRs attempting to fix the same issue.  I'm going to put this in "In Review" for now.  Please coordinate with @powe97 and put this back in "Awaiting Maintainer Review" once the duplicate PR issue is resolved.

williamjallen (approved):
Code looks good to me.

PR #10469, opened 2024-05-27
[Feature:Submission] Add warning for unaligned time setting

### LINKED ISSUE #10458: Add banner warning message for clock skew
If your native machine / web browser clock time is different than the server clock time -- not just that the timezone is different, but the actual time is difference -- this can cause rendering bugs, user confusion, and problematic errors.

Let's add a simple check as we are prepping to render each webpage.
What is the current server clock?  What is users machine clock?   
It's ok/normal if the server and the browser have different timezones but represent the same time when converted to UTC.

But are they different by more than 5 minutes?  
This is a bug, one of the machines is not correctly using ntp to sync to the worldwide agreed upon time.   If they are different by more than 5 minutes let's add a banner that says "WARNING: Your client / web browser clock does not match the server clock and this may cause rendering bugs or other errors."

What threshold should we use?  I think something between 1-5 minutes makes sense.

See also #10440 (a bug related to skew) and #9584 (mac users development machines regularly experience skew) 
Banner should be similar to warning message for socket server being down #7020

Xiayucheng1212:
fixes #10458 

No banner warn the unsynchronized time settings between server and browser.

### What is the new behavior?
If the difference between server time and browser time larger than 2 min, show the banner.

### DIFF
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -26,7 +26,7 @@
         var serverTime = {{ server_time }};
         var timeDifference = serverTime * 1000 - browserTime;
         if (Math.abs(timeDifference) > 120000) {
-            document.querySelector('#server-time-not-aligned').removeAttr('hidden');
+            document.querySelector('#server-time-not-aligned').removeAttribute('hidden');
         }
     </script>
 
--- a/site/app/templates/GlobalHeader.twig
+++ b/site/app/templates/GlobalHeader.twig
@@ -22,12 +22,14 @@
         }
     </script>
     <script>
-        var browserTime = new Date().getTime();
-        var serverTime = {{ server_time|e('js') }};
-        var timeDifference = serverTime * 1000 - browserTime;
-        if (Math.abs(timeDifference) > 120000) {
-            document.querySelector('#server-time-not-aligned').removeAttribute('hidden');
-        }
+        document.addEventListener("DOMContentLoaded", function() {
+            var browserTime = new Date().getTime();
+            var serverTime = {{ server_time|e('js') }};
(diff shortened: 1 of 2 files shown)


site/app/templates/GlobalHeader.twig
```diff
@@ -21,6 +21,14 @@
             })
         }
     </script>
+    <script>
```
skara9, line 24: Need to move all this code into a document load listener, or it won't be able to manipulate the element.

site/app/templates/GlobalHeader.twig
```diff
@@ -21,6 +21,14 @@
...
+        var browserTime = new Date().getTime();
+        var serverTime = {{ server_time }};
+        var timeDifference = serverTime * 1000 - browserTime;
+        if (Math.abs(timeDifference) > 120000) {
+            document.querySelector('#server-time-not-aligned').removeAttr('hidden');
```
skara9, line 29:
```suggestion
            document.querySelector('#server-time-not-aligned').removeAttribute('hidden');
```
'Attr' is jQuery

skara9 (changes requested):
Code looks mostly correct, but fails due to errors. Remember to test that your feature works as intended before marking your PR for review.
After applying the changes below, the message appears properly.

ziesski (approved):
After studying the code, I noticed that while Date.now() returns Unix epoch in user's local time zone, and server_time will provide Unix timestamp as well. So they are both derived from same epoch & UTC, and we can conclude that user's local time zone doesn't matter. 

I tried testing changing time on my mac, change from submitty profile page, change from my terminal (VM) and they behave as expected

skara9 (approved):

PR #10479, opened 2024-05-28
[Bugfix:System] Adduser group optional

cjreed121:

In https://github.com/Submitty/Submitty/pull/10133, a new group argument was added to the adduser script. However, it is not an optional argument since it is missing `--`. The create course script will error as is and not add the instructor to the course.

### What is the new behavior?
`--` has been added in front of the argument so this is now an optional argument. The create course script does not fail anymore.

### Other information?
Tested locally and it fixed it.


IDzyre (approved):
I noticed this after when testing a different change for another student

PR #10382, opened 2024-04-22
[Feature:Forum] Total Upducks in Forum Stats Page

DannyJiang1:
This PR is aimed to add the feature to display users' total upducks inside the forum stats page.

### Please check if the PR fulfills these requirements:

* [x] Tests for the changes have been added/updated (if possible)
* [x] Documentation has been updated/added if relevant
* [x] Screenshots are attached to Github PR if visual/UI changes were made

### What is the current behavior?
Closes #10378

### What is the new behavior?
Now the forum stats page will display the total upducks for each user.

### Other information?
To test, I created a couple of posts as both the instructor and the student, and I randomly upducked posts as both users. In the end, I compared the number in the stat page to the actual upducks within forum threads page.

Stats page:
Upducks

Forum page:
Upducks2

PR #10480, opened 2024-05-28
[Testing:Developer] Remove force click in test_forum

micpap25:

The force-clicks and wait in this test don't serve a purpose and are bad practice. This removes them.


micpap25:
So it turns out the force-clicks are used to account for what seems to be an issue with the merge-thread dropdown in that sometimes the thread you want to merge doesn't appear (potentially due to recency of creation?) Waiting wouldn't change this. Something underlying about this needs to be fixed.

IDzyre:
in this specific case, one of the force clicks are because the popup and dropdown for the merge thread goes over the edge of the popup, and scrolling down made even more flakiness, so just forcing it to click the desired dropdown option was the best route.

micpap25:
This should probably be on the backburner until underlying issues with merge threads are resolved

IDzyre:
The force clicks were the way to get around scrolling, since in Cypress, { force: true } clicks on the element no matter if it's visible or not. I agree that the wait should be removed. This was one of the first tests I worked on when I was first starting with Cypress, and so I didn't use the best selectors, and no testids. This one should be refactored to use testids.

williamschen23:
I think we should wait until #10382 is merged because it refactors some of the code to use better selectors, and I'm going to mark this as a draft.

williamjallen:
@williamschen23 and @micpap25, what's the status of this PR?  As per https://github.com/Submitty/Submitty/pull/10480#issuecomment-2158585287, this was waiting for https://github.com/Submitty/Submitty/pull/10382 which has now been merged.

williamschen23:
> @williamschen23 and @micpap25, what's the status of this PR? As per [#10480 (comment)](https://github.com/Submitty/Submitty/pull/10480#issuecomment-2158585287), this was waiting for #10382 which has now been merged.

I tried after #10382 got merged and did not get far. Since this was Michaels PR, I figured he would pick it back up. Since then, forums got another wait if I remember correctly.

micpap25:
Picking this back up, though I might just close it because the issue with it does stem back at merge threads being slow (which i can't address).

IDzyre (approved):

PR #10487, opened 2024-05-30
[Bugfix:System] Improve libvirt support

dorian451:

When using libvirt as the Vagrant backend, the GIT_CHECKOUT folder in the vm has the wrong permissions, and other small problems that prevent usage. 

### What is the new behavior? 
* Sets the permission of GIT_CHECKOUT to 0755 if the virtual machine is running in dev mode
* When running with libvirt, vagrant now uses libvirt's qemu/kvm user session instead of the system one 
* When running with libvirt, vagrant now explicitly uses rsync to keep the vm's files in sync with the host

### Other information?
* To sync files, the user must run `vagrant rsync-auto` on the host.

### Is this a breaking change?
No

### How did you test
`vagrant up --provider=libvirt`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -76,22 +76,27 @@ base_boxes[:arm_bento]     = "bento/ubuntu-22.04-arm64"
 base_boxes[:libvirt]       = "generic/ubuntu2204"
 base_boxes[:arm_mac_qemu]  = "perk/ubuntu-2204-arm64"
 
-def mount_folders(config, mount_options)
-  # ideally we would use submitty_daemon or something as the owner/group, but since that user doesn't exist
+
+def mount_folders_with_type(config, mount_options, type)
+ # ideally we would use submitty_daemon or something as the owner/group, but since that user doesn't exist
   # till post-provision (and this is mounted before provisioning), we want the group to be 'vagrant'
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -76,24 +76,29 @@ base_boxes[:arm_bento]     = "bento/ubuntu-22.04-arm64"
...
   optional_repos.each {|repo|
     repo_path = File.expand_path("../" + repo)
     if File.directory?(repo_path)
-      config.vm.synced_folder repo_path, "/usr/local/submitty/GIT_CHECKOUT/" + repo, owner: owner, group: group, mount_options: mount_options, smb_host: '10.0.2.2', smb_username: `whoami`.chomp
+      config.vm.synced_folder repo_path, "/usr/local/submitty/GIT_CHECKOUT/" + repo, owner: owner, group: group, mount_options: mount_options, smb_host: '10.0.2.2', smb_username: `whoami`.chomp, type:type
     end
   }
 end
 
+def mount_folders(config, mount_options)
```
IDzyre, line 98: Is creating a separate function necessary, can you just set the default value of the type parameter to be nil?

IDzyre (changes requested):
Question about creating a new function instead of default parameter value

Vagrantfile
dorian451: I wasn't aware ruby had default parameters, my bad

PR #10488, opened 2024-05-30
[Feature:System] Ansible Course Creation Scripts

ethanfordwork:

Currently, there is only one method for creating a course and creating a term, both following extensive manual process through command-line prompts.

### What is the new behavior?

With the Ansible Term and Course Creation Scripts, it will allow for using the Ansible automation tool to create a term and create a course, all in one playbook. Editing that playbook will allow for more customizable creation of the term and courses. This is a add on to the Ansible Install Script that we have, all in the .setup/ansible directory.


.setup/ansible/roles/submitty_course_add_instructor/handlers/main.yml
```diff
@@ -0,0 +1,5 @@
+---
```
shailpatels, line 1: You could define the handler in the playbook so its globally accessible to all roles as a suggestion to reduce duplication and let future roles re-use it as well.

.setup/ansible/roles/submitty_course_add_instructor/handlers/main.yml
```diff
@@ -0,0 +1,5 @@
...
+- name: Reload fpm
+  ansible.builtin.service:
+    name: php8.1-fpm
```
shailpatels, line 4: You could also move the version to a vars or inventory file so the php version can be changed in one place in the future
`name: "php{{ php_version }}-fpm"`

.setup/ansible/roles/submitty_course_add_instructor/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+submitty_course_add_instructor_username: instructor
```
shailpatels, line 2: Is the intention of this to be developer defaults? If not it seems like a bad idea to define these incase they accidentally become used in production. 

It might be a better idea to leave them undefined and if the user does not define or enter them have ansible crash out. Or explicitly give hints they are required and why https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_filters.html#forcing-variables-to-be-defined

.setup/ansible/playbooks/submitty_course_creation.yml
```diff
@@ -0,0 +1,58 @@
...
+
+  roles:
+    - role: submitty_term_creation
+      vars:
+        submitty_term_creation_key: term
+        submitty_term_creation_name: winter2024
+        submitty_term_creation_start_date: 01/01/2024
+        submitty_term_creation_end_date: 06/07/2024
+
+    - role: submitty_course_creation
```
shailpatels, line 14: I haven't been to close to the discussions related to ansible but is this playbook used for setting up a developer instance or is the idea that universities will maintain their own playbooks to manage their submitty instance?

.setup/ansible/roles/submitty_term_creation/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Run the create_term script to create a term.
+  ansible.builtin.command:
+    cmd: ./create_term.sh --amend "{{ submitty_term_creation_key }}"
```
shailpatels, line 4: Is there any plan / discussion of converting bash scripts like this to ansible? Might be beneficial to do, some of them could probably use the built in error handling and state manage ansible provides more than others

shailpatels (dismissed):
I haven't been too close to the conversation related to ansible recently so some of my comments may be ignored

But very cool to see progress being made in this area!

.setup/ansible/roles/submitty_course_add_instructor/handlers/main.yml
shailpatels: I think you can also access handlers accross roles if they're in the same play instance, but thats less explicit and not 100% sure about that and would of course depend on future playbooks using the role that defines the handler

.setup/ansible/playbooks/submitty_course_creation.yml
prestoncarman: I see these playbooks as examples of how a university could manage their infrastructure. Each university would set the parameters for their university in their own playbook. All universities would use the same roles with just different parameters. At least that is what I was thinking.

.setup/ansible/roles/submitty_term_creation/tasks/main.yml
prestoncarman: I think this is something we should talk about. Currently, we have bash and python scripts( and a partial API too).

.setup/ansible/roles/submitty_course_add_instructor/handlers/main.yml
ethanfordwork: We defined the handler as submitty_common_php_version and set that in the playbook for all roles. We could set the handler in the playbook all together, but we do not want to change the roles, but using the playbook to call the variables with set roles.

.setup/ansible/roles/submitty_course_add_instructor/handlers/main.yml
ethanfordwork: Thanks. Made this update consistent with all roles.

.setup/ansible/roles/submitty_course_add_instructor/defaults/main.yml
ethanfordwork: We removed our default values and instead, created a README that instructed the user to set the desired variables as they wish. If not, the script will produce an error prompting them to set the variables in the playbooks.

.setup/ansible/playbooks/submitty_course_creation.yml
ethanfordwork: The playbooks are used as an example, showing each University how they should set their parameters.

.setup/ansible/playbooks/submitty_course_creation.yml
```diff
@@ -0,0 +1,70 @@
...
+      vars:
+        submitty_course_add_instructor_term: term
+        submitty_course_add_instructor_course: course
+        submitty_course_add_instructor_username: instructor2
+        submitty_course_add_instructor_firstname: firstname
+        submitty_course_add_instructor_lastname: lastname
+        submitty_course_add_instructor_email: email
+        submitty_course_add_instructor_password: password
+    - name: Add Users
+      ansible.builtin.include_role:
```
IDzyre, line 43:
```suggestion
        submitty_course_add_instructor_password: password

    - name: Add Users
      ansible.builtin.include_role:
```

.setup/ansible/roles/submitty_course_creation/tasks/main.yml
```diff
@@ -0,0 +1,125 @@
...
+    cmd: ./adduser_course.py "{{ submitty_course_creation_username }}"
+      "{{ submitty_course_creation_term }}" "{{ submitty_course_creation_course }}"
+      "{{ submitty_course_creation_user_group }}"
+    chdir: /usr/local/submitty/sbin/
+  become: true
+  become_user: root
+  register: submitty_course_creation
+  changed_when: submitty_course_creation.rc != 0
+
+- name: Restart php8.1-fpm service.
```
prestoncarman, line 120: This should use the php variable.

prestoncarman (commented):
Do you need `ansible.cfg`? If not, remove the empty file.

prestoncarman:
@IDzyre I am wondering how to set up a test to verify these scripts are working using the existing Submitty CI testing tools.

.setup/ansible/roles/submitty_course_creation/tasks/main.yml
ethanfordwork: Updated name, inserted php variable name.

IDzyre (approved):
Changes  look good to me

PR #10509, opened 2024-06-04
[Bugfix:InstructorUI] Delete last rubric component break site

### LINKED ISSUE #10503: Deleting the last rubric item from a gradable breaks the site
**Describe the bug**
If the last rubric item is removed from a gradable, the site will become inaccessible in various places, at best preventing access to that gradable and at worst preventing access to the whole gradable page.

**Expected behavior**
Prevent the removal of the last rubric item.

**To Reproduce**
1. Create a new gradable with manual grading turned on.
2. Go to the Rubric tab.
3. Delete the one item that is there by default

**Configuration**
 - OS: Windows
 - Browser: Chrome

ZheyuDeng:
Fix #10503 

Deleting the last rubric component breaks the site. Had to run a vagrant destroy.

### What is the new behavior?
Now it will stop users from deleting the last component
image

To Reproduce

Create a new gradable with manual grading turned on.
Go to the Rubric tab.
Delete the one item that is there by default
See alert


micpap25 (approved):

site/public/js/ta-grading-rubric.js
```diff
@@ -1769,6 +1769,12 @@ function onRestoreMark(me) {
  * @param me DOM Element of the delete button
  */
 function onDeleteComponent(me) {
+    const componentCount = $('.component-container').length;
+    if (componentCount === 1) {
+        alert('Cannot delete the only component.');
```
IDzyre, line 1774: Would it be better to use displayErrorMessage() to have the red banner message instead of the JS alert?

ZheyuDeng: Ye I have switched to displayErrorMessage instead now
image

site/public/js/ta-grading-rubric.js
```diff
@@ -1769,9 +1769,16 @@ function onRestoreMark(me) {
  * @param me DOM Element of the delete button
  */
 function onDeleteComponent(me) {
+    const componentCount = $('.component-container').length;
+    if (componentCount === 1) {
+        displayErrorMessage('Can not delete the only component.');
```
ziesski, line 1774: Can we do Cannot instead of Can not, but everything else looks good :)

ziesski (approved):

micpap25:
@ziesski done

PR #10513, opened 2024-06-05
[Bugfix:InstructorUI] No dark mode for select2

ZheyuDeng:
Timezone not displaying in dark mode
15a49de88c9407256093ead47738460

### What is the new behavior?
017ec347b00850f8ae93b1a1aacf923


zacharymnp (approved):
image

Light mode and dark mode both work.

PR #10514, opened 2024-06-05
[Bugfix:InstructorUI] Fix manage grader save changes popup

ZheyuDeng:
In manage graders,
When cancel edit graders immediatly after opening it, the confirm(Changes to the form will be lost!) pops up and ask for your action which is quite annouying. You can test it here. http://localhost:1511/courses/s24/sample/graders). See video below:

https://github.com/Submitty/Submitty/assets/122962727/f49701f9-3092-4a01-ac75-8566cd7846ee

### What is the new behavior?
Fix the isForumEdited function so it functions correctly

### DIFF
--- a/site/public/js/userform.js
+++ b/site/public/js/userform.js
@@ -359,23 +359,30 @@ function isUserFormEdited() {
             current_grading_sections.push(+$(this).val());
         }
     });
-    // check if there is any changes made in the grading checkboxes.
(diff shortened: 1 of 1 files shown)


micpap25 (approved):
Tested, looks all good to me.

williamjallen:
Can you add screenshots and/or videos to the PR description?  You can add this PR back to the "Awaiting Maintainer Review" column directly after doing so.

PR #10526, opened 2024-06-07
[Refactor:System] Adding submitty_test to submitty_help

williamschen23:

1) The test shortcut for PHP linting, submitty_test, does not show up to submitty_help. 
2) Composer installs whenever the command is ran.

### What is the new behavior?
1) Added text to show that submitty_test exists.
2) Moved composer to only install when a PHP lint is called.

### DIFF
--- a/.setup/SUBMITTY_TEST.sh
+++ b/.setup/SUBMITTY_TEST.sh
@@ -16,7 +16,7 @@ run_php_cs() {
 }
 
 
-if [ "$1" == "help" ] || [ -z "$1" ]; then
+if [ -z "$1" ] || [ "$1" == "help" ]; then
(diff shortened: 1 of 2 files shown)


williamschen23:
Also, why don't we call this submitty_lint?

ziesski:
> Also, why don't we call this submitty_lint? It would make it so much easier to tell what it is doing when doing submitty_help
스크린샷 2024-06-10 오전 11 29 00

Because not all tests are linting, if you take a look at the screenshot, we also have unit testings, bash shell check etc. So we are using broader term.

ziesski (approved):
I haven't tested, but the changes seem reasonable.

PR #10527, opened 2024-06-07
[Bugfix:InstructorUI] Disable background scroll in Userprofile

ZheyuDeng:

In user profile scroll in background is still enable where there's a popup
See below:

https://github.com/Submitty/Submitty/assets/122962727/c254b8c2-bc43-442e-986c-871bd7526df1
### What is the new behavior?
Disabled scroll in background


ziesski (approved):
This is a good fix utilizing our existing function

PR #10530, opened 2024-06-10
[Feature:System]  Revamp worker setup and fix QEMU

skara9:

Worker configuration currently fails on QEMU due to lack of support for private networking.

### What is the new behavior?
The QEMU plugin has been modified to add support for socket networking.
The worker management configuration has been revamped to better accomodate the networking needs of the new QEMU plugin setup.

Testing instructions:
See Submitty/submitty.github.io#621


GO-viper7:
Please request a review after this PR gets out of draft.

skara9:
@ziesski When you install homebrew, it gives you a command to update your PATH. You may have missed this step.

ziesski:
It has been long while since I installed homebrew. 

Screen Shot 2024-06-21 at 11 33 17 AM

This is my current `HOMEBREW_PREFIX` What do you want me to update it to.

GO-viper7 (commented):
So, I was already on workers set up
Thus I removed only the worker-VM manually from GUI and ran the commands you mentioned above
You can see the result in the screenshot below  
It probably has an issue with the ssh connection or Do I need to remove even the primary machine to try this ?

image

ziesski:
@skara9 I am doing this time on my M3 MacOS 14, qemu, and getting below error. this is after I did `vagrant destroy`
Could you tell me what is wrong? 
스크린샷 2024-07-01 오전 9 50 42

GO-viper7 (approved):
I followed the testing instructions on Windows
All the commands show expected results, couldn't find any errors in CLI
and FYI, here's my `workers.json`
```json
{
    "version": 2,
    "provider": "virtualbox",
    "workers": {
        "worker-1": {
            "ip_addr": "192.168.56.2",
            "ssh_port": 2241
        }
    }
}
```

dorian451 (approved):
I think it works

ziesski:
Tested on intel Mac OS 12, works.
Will test on M3 once @skara9 provides the fix.

ziesski:
스크린샷-2024-07-15-오전-11 51 50
This is what I get after vagrant workers up, is this expected?
@skara9

bmcutler:
@skara

Notes on instructions above

* Please start with "vagrant destroy" (if you have an existing vagrant VM)
...  that is, assuming you must destroy the main VM first?  But maybe you don't?  I was getting errors, but I'm just starting to go through these steps.

* BASE=1 is the old version.  We switched to FROM_SCRATCH=1

* Where do I specify the # of workers?  These instructions are replacing the old instructions which had that as a clear option
https://submitty.org/developer/getting_started/worker_vm

skara9:
@bmcutler You don't need to destroy the main VM for setting up workers, but you will have to reprovision it at the end because the installation scripts on the main still send the SSH key to the workers.

For multiple workers, it is `vagrant workers generate -n X`, this is documented in the new instructions (not merged yet).

bmcutler:
@skara9 
Please resolve conflicts

GO-viper7:
@skara9 
I followed the testing instructions and here's what I got in windows 
Also, I made sure to set FROM_SCRATCH=1, as you have mentioned current prepackaged VM does not support workers.
It still didn't work as intended.

image

workers.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
+if ! which python3 >/dev/null; then
+  echo "Python 3 has not been configured. Make sure to install it and add it to your PATH."
+  exit 1
+fi
```
GO-viper7, line 6: It only checks for Linux / MacOS
May be this is the reason I'm failing on Windows.

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,96 @@
...
+    enabled += 1
+  total += 1
+  new_autograding_data[name] = worker_data
+
+with open(autograding_workers_path, 'w') as file:
+  json.dump(new_autograding_data, file, indent=4)
+print(f"Configuration saved with {enabled}/{total} machines enabled")
+print()
+
+os.system('systemctl restart submitty_autograding_shipper && systemctl restart submitty_autograding_worker')
```
bmcutler, line 95: Don't include autograding restart in this script

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,96 @@
...
+if 'version' in vagrant_workers_data and type(vagrant_workers_data['version']) == int:
+  provider = vagrant_workers_data['provider']
+  vagrant_workers_data = vagrant_workers_data['workers']
+else:
+  print("This script requires a worker configuration of v2 or greater. Please regenerate your configuration with 'vagrant workers generate'.")
+  exit(1)
+print("Done loading data")
+print()
+
+print("Generating SSH credentials...")
```
bmcutler, line 32: add a comment explaining what steps from the (non-vagrant) sysadmin instructions these is mimicing.
https://submitty.org/sysadmin/installation/worker_installation

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,96 @@
+import os
+import shutil
+import pwd
+import json
+import subprocess
+from collections import OrderedDict
+
+SUBMITTY_REPOSITORY = os.environ['SUBMITTY_REPOSITORY']
```
bmcutler, line 8: these environment variables don't exist on production, let's not use this here...

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,96 @@
+import os
```
bmcutler, line 1: this script should only be run on vagrant, let's add a check towards the top of the file to make sure this isn't accidentally run on production.  (this script could be damaging to a production install)

.setup/install_system.sh
```diff
@@ -182,6 +182,7 @@ alias migrator='python3 ${SUBMITTY_REPOSITORY}/migration/run_migrator.py -c ${SU
 alias vagrant_info='cat /etc/motd'
 alias ntp_sync='service ntp stop && ntpd -gq && service ntp start'
 alias recreate_sample_courses='sudo bash /usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/bin/recreate_sample_courses.sh'
+alias refresh_vagrant_workers='python3 /usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/bin/refresh_vagrant_workers.py'
```
bmcutler, line 185: I don't think this alias is necessary to add.
(I also believe recreate sample courses alias should only be on vagrant systems, not on production)

bmcutler (changes requested):
I tested it, and with tweaks to the instructions, I did get this to work with 1 worker machine, and successfully share autograding between primary & 1 worker on an M1 mac.

williamschen23 (approved):
Did not look at the code, works on m2 mac

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,95 @@
...
+
+print("Generating SSH credentials...")
+shutil.rmtree("/tmp/worker_keys", True)
+os.makedirs("/tmp/worker_keys", 0o500)
+daemon_stat = pwd.getpwnam(DAEMON_USER)
+os.chown("/tmp/worker_keys", daemon_stat.pw_uid, daemon_stat.pw_gid)
+
+DAEMON_HOME = os.path.realpath(subprocess.check_output(['su', DAEMON_USER, '-c', 'echo $HOME']).strip())
+if not os.path.exists(DAEMON_HOME):
+    print("Error: could not find home directory for daemon user")
```
jeffrey-cordero, line 40: Should we not also `exit(1)` at this point to prevent subsequent failures in the SSH key removals?

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,95 @@
...
+
+shutil.rmtree(os.path.join(DAEMON_HOME, b'.ssh'), True)
```
jeffrey-cordero, line 43: Removing the entire directory seems risky for existing machines. Should we not only delete the SSH files being modified in the following lines?

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,95 @@
...
+print()
+
+ssh_config = ''
+successful_machines = []
+for name, data in vagrant_workers_data.items():
+    print("Attempting to connect to " + name)
+    shutil.copyfile(f"{SUBMITTY_REPOSITORY}/.vagrant/machines/{name}/{provider}/private_key", f"/tmp/worker_keys/{name}")
+    os.chown(f"/tmp/worker_keys/{name}", daemon_stat.pw_uid, daemon_stat.pw_gid)
+    os.chmod(f"/tmp/worker_keys/{name}", 0o400)
+    w = subprocess.run(['su', DAEMON_USER, '-c', f"scp -i /tmp/worker_keys/{name} -o StrictHostKeyChecking=no ~/.ssh/id_rsa.pub root@{data['ip_addr']}:/tmp/workerkey"])
```
jeffrey-cordero, line 55: `StrictHostKeyChecking=no` makes the script vulnerable to MITM attacks as this will trust any host claiming a given worker machine.

jeffrey-cordero: We should also verify the process exit status code.

Vagrantfile
```diff
@@ -79,29 +79,28 @@ base_boxes[:arm_bento]     = "bento/ubuntu-22.04-arm64"
 base_boxes[:libvirt]       = "generic/ubuntu2204"
 base_boxes[:arm_mac_qemu]  = "perk/ubuntu-2204-arm64"
 
-
-def mount_folders(config, mount_options, type = nil)
+def mount_folders(config, mount_options, type = nil, host = '10.0.2.2')
```
jeffrey-cordero, line 82: This default host may not work for non-VirtualBox instances, but I'll test this with Qemu.

.setup/bin/refresh_vagrant_workers.py
skara9: Yeah, I think this flag was kept in from the really old single worker config (pre v1). The general idea was that our devs are going to be reprovisioning worker VMs back up with the same addresses, and host key checking means our script will fail on setting up the SSH connection between the main VM and workers.

Now that we're wiping known hosts, we're probably in the clear to remove this though.

.setup/bin/refresh_vagrant_workers.py
skara9: SSH dir for the daemon user should be managed entirely by this script; i.e. if there's anything else in there it would be a concern. Good call in general, but wiping it in this case actually saves us from potential trouble.

.setup/bin/refresh_vagrant_workers.py
skara9: Good catch, fixed.

.setup/bin/refresh_vagrant_workers.py
skara9: Could throw in a exit code check here, I just didn't think it was necessary. If the scp errors but the ssh works we don't really lose anything, right?

Chriun: Just tested and it seems that we do need `-o StrictHostKeyChecking=no` for scp at least, otherwise the host fails connection for the workers as a result of known hosts being empty. We couuld have a secure alternative for this by using ssh-keyscan on the worker and adding it directly to the known hosts. but I don't think MITM is that much of a concern considering this setup is purely local.

.setup/bin/refresh_vagrant_workers.py
jeffrey-cordero: My knowledge of this worker-to-main machine SSH connection workflow is lacking a bit. I just noted it because system administrators or developers could change the contents with root privileges, but I assume this isn't a common scenario.

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,96 @@
...
+    worker_data['address'] = data['ip_addr']
+    if name not in successful_machines:
+        worker_data['enabled'] = False
+    else:
+        enabled += 1
+    total += 1
+    new_autograding_data[name] = worker_data
+
+with open(autograding_workers_path, 'w') as file:
+    json.dump(new_autograding_data, file)
```
Chriun: Having indent=4 here would be really helpful here in terms of formatting the json. Currently, it dumps it all into a single line which is not pretty.

vagrant-workers/generate_workers.py
```diff
@@ -0,0 +1,139 @@
...
+
+        if ip is None:
+            raise IndexError("IP range insufficient for requested number of workers")
+
+        data = OrderedDict()
+        data['ip_addr'] = str(ip)
+        if provider == 'qemu':
+            data['mac_addr'] = args.mac_prefix + ":%02x:%02x:%02x" % tuple(random.randint(0, 255) for v in range(3))
+        else:
+            data['ssh_port'] = args.base_port + i
```
jeffrey-cordero, line 118: I would suggest handling conflicting ports by finding the next suitable port and potentially logging the conflict to the terminal.

vagrant-workers/generate_workers.py
```diff
@@ -0,0 +1,139 @@
...
+                    and str(ip).endswith('.1'))):
+            ip = next(ips, None)
+
+        if ip is None:
+            raise IndexError("IP range insufficient for requested number of workers")
+
+        data = OrderedDict()
+        data['ip_addr'] = str(ip)
+        if provider == 'qemu':
+            data['mac_addr'] = args.mac_prefix + ":%02x:%02x:%02x" % tuple(random.randint(0, 255) for v in range(3))
```
jeffrey-cordero, line 116: Though the risk is near zero, adding functionality to handle conflicting MAC addresses would add robustness. On macOS, you can run `ifconfig | grep ether | awk '{print $2}'` to list all existing addresses on a new line.

Chriun (commented):
Works on Windows 10. I did run into problems initially with the setup since I was building the pre-packaged vm instead of from scratch. I think this should be highlighted in the documentation as well.

vagrant-workers/workers.py
```diff
@@ -0,0 +1,145 @@
...
+    if args.command == 'up':
+        if platform.system() != 'Darwin' or config['provider'] != 'qemu':
+            up = subprocess.run(['vagrant', 'up', '--provider={}'.format(config['provider']), *sys.argv[2:]], env=worker_env)
+            exit(up.returncode)
+        PLUGIN_VERSION = '24.07.00'
+        plugin_info = subprocess.check_output(['vagrant', 'plugin', 'list', '--machine-readable'])
+        if 'vagrant-qemu,plugin-version,{}%'.format(PLUGIN_VERSION) not in str(plugin_info):
+            print('Updating QEMU plugin...')
+            with tempfile.TemporaryDirectory() as tmpdir:
+                with urllib.request.urlopen(f'https://github.com/Submitty/vagrant-qemu/releases/download/v{PLUGIN_VERSION}/vagrant-qemu-{PLUGIN_VERSION}.gem') as res:
```
jeffrey-cordero, line 122: This does not work on macOS for me.

```bash
➜  Submitty git:(qwork) ✗ vagrant workers up
Updating QEMU plugin...
  File "/Users/jeffreycordero/Developer/open-source/Submitty/vagrant-workers/workers.py", line 145, in <module>
    main()
    ~~~~^^
  File "/Users/jeffreycordero/Developer/open-source/Submitty/vagrant-workers/workers.py", line 122, in main
    with urllib.request.urlopen(f'https://github.com/Submitty/vagrant-qemu/releases/download/v{PLUGIN_VERSION}/vagrant-qemu-{PLUGIN_VERSION}.gem') as res:
         ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 189, in urlopen
    return opener.open(url, data, timeout)
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 495, in open
    response = meth(req, response)
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 604, in http_response
    response = self.parent.error(
        'http', request, response, code, msg, hdrs)
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 527, in error
    result = self._call_chain(*args)
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 466, in _call_chain
    result = func(*args)
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 720, in http_error_302
    return self.parent.open(new, timeout=req.timeout)
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 495, in open
    response = meth(req, response)
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 604, in http_response
    response = self.parent.error(
        'http', request, response, code, msg, hdrs)
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 533, in error
    return self._call_chain(*args)
           ~~~~~~~~~~~~~~~~^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 466, in _call_chain
    result = func(*args)
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/urllib/request.py", line 613, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 403: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
```

I was able to fix the authorization issue by adding the user agent to the headers of the request:

```python
with urllib.request.urlopen(urllib.request.Request(f'https://github.com/Submitty/vagrant-qemu/releases/download/v{PLUGIN_VERSION}/vagrant-qemu-{PLUGIN_VERSION}.gem', headers={'User-Agent': 'Submitty/1.0'})) as res:
```

.setup/bin/refresh_vagrant_workers.py
jeffrey-cordero: I originally noted this because it could be a vulnerability for production environments. This should be inspected by someone with more senior-level security experience, but I think the original approach is fine because it is similar to the existing connection command.

```bash
$su submitty_daemon -c "sshpass -p 'submitty' ssh-copy-id -i ~/.ssh/id_rsa.pub -o StrictHostKeyChecking=no submitty@${IP}"
```

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,96 @@
...
+shutil.rmtree(os.path.join(DAEMON_HOME, b'.ssh'), True)
+subprocess.run(['su', DAEMON_USER, '-c', "ssh-keygen -b 2048 -t rsa -f ~/.ssh/id_rsa -q -N ''"], check=True)
+print("Done generating")
+print()
+
+ssh_config = ''
+successful_machines = []
+for name, data in vagrant_workers_data.items():
+    print("Attempting to connect to " + name)
+    shutil.copyfile(f"{SUBMITTY_REPOSITORY}/.vagrant/machines/{name}/{provider}/private_key", f"/tmp/worker_keys/{name}")
```
jeffrey-cordero, line 52: This will fail if the file doesn't exist on the worker machine. I'd recommend wrapping this in a `try-catch`, where if the file doesn't exist, create it in the exception block, or update the instructions to ensure `sudo touch /usr/local/submitty/GIT_CHECKOUT/Submitty/.vagrant/machines/<worker>/<provider>/private_key` is run before running the script.

```bash
root@ubuntu:/usr/local/submitty# refresh_vagrant_workers
Loading existing data...
Done loading data

Generating SSH credentials...
Done generating

Attempting to connect to worker-1
  File "/usr/local/submitty/GIT_CHECKOUT/Submitty/.setup/bin/refresh_vagrant_workers.py", line 52, in <module>
    shutil.copyfile(f"{SUBMITTY_REPOSITORY}/.vagrant/machines/{name}/{provider}/private_key", f"/tmp/worker_keys/{name}")
  File "/usr/lib/python3.10/shutil.py", line 254, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/submitty/GIT_CHECKOUT/Submitty/.vagrant/machines/worker-1/qemu/private_key'`
```

Vagrantfile
```diff
@@ -144,22 +143,44 @@ Vagrant.configure(2) do |config|
...
+        ubuntu.vm.network 'private_network', ip: data[:ip_addr]
+        ubuntu.vm.network 'forwarded_port', guest: 22, host: data[:ssh_port], id: 'ssh' unless data[:ssh_port].nil?
+        ubuntu.vm.provision 'shell', inline: gen_script(worker_name, worker: true, base: true)
+
+        ubuntu.vm.provider "qemu" do |qe, override|
+          qe.ssh_host = data[:ip_addr]
+          qe.ssh_port = 22
+          qe.socket_fd = 3
+          qe.mac_address = data[:mac_addr]
+          mount_folders(override, [], 'smb', ENV.fetch('GATEWAY_IP', '10.0.2.2'))
```
jeffrey-cordero, line 165: This will not work at all if `GATEWAY_IP` is not defined. This may be in the instructions somewhere, but if not, please append it. This should be set to the local host IP on macOS because `10.0.2.2` is the default host gateway IP used by VirtualBox’s NAT networking.

.setup/bin/refresh_vagrant_workers.py
```diff
@@ -0,0 +1,96 @@
...
+    os.chown(f"/tmp/worker_keys/{name}", daemon_stat.pw_uid, daemon_stat.pw_gid)
+    os.chmod(f"/tmp/worker_keys/{name}", 0o400)
+    w = subprocess.run(['su', DAEMON_USER, '-c', f"scp -i /tmp/worker_keys/{name} ~/.ssh/id_rsa.pub root@{data['ip_addr']}:/tmp/workerkey"])
+    w = subprocess.run(['su', DAEMON_USER, '-c', f"ssh -i /tmp/worker_keys/{name} root@{data['ip_addr']} \"chown {SUPERVISOR_USER}:{SUPERVISOR_USER} /tmp/workerkey && su {SUPERVISOR_USER} -c \\\"mkdir -p ~/.ssh && mv /tmp/workerkey ~/.ssh/authorized_keys\\\"\""])
```
jeffrey-cordero: I'm infinitely blocked when using SSH from both local and main machines.

`ssh: connect to host <Worker IP> port 22: Connection refused`

This could be a networking issue with Qemu within the Vagrant file, but I'm a bit stuck on finding the right solution to this macOS-specific issue.

```bash
$Submitty git:(qwork) ✗ ping <Worker IP>             
PING <Worker IP>: 56 data bytes
Request timeout for icmp_seq 0
Request timeout for icmp_seq 1
Request timeout for icmp_seq 2
Request timeout for icmp_seq 3
ping: sendto: No route to host
```

Vagrantfile
```diff
@@ -144,22 +143,44 @@ Vagrant.configure(2) do |config|
...
-    ubuntu.vm.provision 'shell', inline: gen_script(vm_name, base: base_box)
+  if ENV.fetch('WORKER_MODE', '0') == '1'
+    get_workers.map do |worker_name, data|
+      config.vm.define worker_name do |ubuntu|
+        ubuntu.vm.network 'private_network', ip: data[:ip_addr]
+        ubuntu.vm.network 'forwarded_port', guest: 22, host: data[:ssh_port], id: 'ssh' unless data[:ssh_port].nil?
+        ubuntu.vm.provision 'shell', inline: gen_script(worker_name, worker: true, base: true)
+
+        ubuntu.vm.provider "qemu" do |qe, override|
+          qe.ssh_host = data[:ip_addr]
```
jeffrey-cordero, line 161: For some reason, on Qemu, the installation script for the workers only invokes for `"127.0.0.1"` as any other IP leads to connection refusal. Not sure if this works for anyone else on macOS.

jeffrey-cordero: This also only works, I see, when I remove the hardcoded file descriptor line. This approach still requires me to manually update `.vagrant/workers.json`, which is not optimal. I hope this works for someone else.

.setup/bin/refresh_vagrant_workers.py
skara9: Yes, idea is they shouldn't be changing it.

.setup/bin/refresh_vagrant_workers.py
skara9: I'm assuming that output is because you have a worker-1 configured but not provisioned, right?

Vagrantfile
skara9: This variable is set by script, user shouldn't be touching it. If it's not set, that's intentional, and it will use `10.0.2.2`, which is the host gateway from the qemu VM.

vagrant-workers/workers.py
skara9: It's strange that a user agent header would solve a 403. Have you retried without this change to make sure it wasn't just a temporary issue on GitHub's side?

Vagrantfile
skara9: If you're referring to ssh_host on this line, that can't be set to localhost. It should be the address of the worker VM.

skara9: What do you mean by file descriptor line? And what changes did you make to the workers.json?

.setup/bin/refresh_vagrant_workers.py
skara9: If the worker IP isn't up (and your worker VM is running), that's a networking issue.

.setup/bin/refresh_vagrant_workers.py
skara9: I'll just throw the flag back in for now then, we can deal with host keys later if we want to.

skara9:
@Chriun 
> Works on Windows 10 with 1 and 3 workers. I did run into problems initially with the setup since I was building the pre-packaged vm instead of from scratch. I think this should be highlighted in the documentation as well.

Ideally once this is merged we can package the worker setup as well.

.setup/bin/refresh_vagrant_workers.py
jeffrey-cordero: I assume that is the case. I followed your documentation steps, but I assume I need to follow the worker installation steps first - https://submitty.org/sysadmin/installation/worker_installation?

Vagrantfile
jeffrey-cordero: I mapped localhost to the worker IP and the hard-coded port within the worker.json, and it worked. I'll retry the entire process today. SSH should work on localhost as the worker VM was open on port 22, but this is not the correct approach for socket networking in this case.

vagrant-workers/workers.py
jeffrey-cordero: It must have been some server issue on GitHub's end that potentially corrupted my setup yesterday, leading to many issues. I had to run `vagrant plugin expunge --force` and no errors show up today.

.setup/bin/refresh_vagrant_workers.py
jeffrey-cordero: Probably. I'm following all your exact steps, and I'm still stuck in the SSH connection process now. I'll call it quits here and assume this works on your end.

```bash
➜  Submitty git:(qwork) vagrant workers up      
Bringing machine 'worker-1' up with 'qemu' provider...
==> worker-1: Checking if box 'perk/ubuntu-2204-arm64' version '20230712' is up to date...
==> worker-1: Importing a QEMU instance
    worker-1: Creating and registering the VM...
    worker-1: Successfully imported VM
==> worker-1: Preparing SMB shared folders...
    worker-1: You will be asked for the username and password to use for the SMB
    worker-1: folders shortly. Please use the proper username/password of your
    worker-1: account.
    worker-1:  
    worker-1: Username (jeffreycordero):
    worker-1: Password (will be hidden):
==> worker-1: Warning! The QEMU provider doesn't support any of the Vagrant
==> worker-1: high-level network configurations (`config.vm.network`). They
==> worker-1: will be silently ignored.
==> worker-1: Starting the instance...
==> worker-1: Waiting for machine to boot. This may take a few minutes...
    worker-1: SSH address: <IP>:22
    worker-1: SSH username: vagrant
    worker-1: SSH auth method: private key
    [BLOCKED INDEFINITELY]
```

skara9: Well if you're reviewing it, you should make sure to get it working on your end.
If it adds to our troubleshooting steps or highlights something to clarify in docs, that would prevent future devs from getting stuck wherever you did.

In case you need any help debugging, you can message me on Zulip.

.setup/bin/refresh_vagrant_workers.py
skara9: Nope, that is for system administrators not developers.

bmcutler:
NOTE:  This PR works on MacOS 14 and below.  We will open an issue describing the problems with MacOS 15.

PR #10551, opened 2024-06-13
[DevDependency] Update ESLint style rules

dorian451:
This PR  replaces the [deprecated ESLint style rules](https://eslint.org/blog/2023/10/deprecating-formatting-rules/) in our ESLint config to the newer, maintained ones by [ESLint Stylistic](eslint.style). Since the previous rules were broken and not working properly, this PR also updates  any non-conformant files to be conformant with our style rules.

This PR also changes our [arrow-parens](https://eslint.style/rules/js/arrow-parens) style rule to "always".

PR #10554, opened 2024-06-14
[DevDependency] Fix ESLint ignoring files it should not have

dorian451:
This PR allows ESLint to lint files we previously ignored, such as `.ts` files in `site/ts`. It also applies any fixes to make these now-linted files conformant to our linting rules.


site/package.json
```diff
@@ -52,34 +52,36 @@
...
     "stylelint": "^15.11.0",
     "stylelint-config-standard": "^33.0.0",
     "typescript": "^5.4.5",
+    "typescript-eslint": "^7.13.0",
     "vnu-jar": "^23.4.11"
   },
   "scripts": {
     "css-stylelint": "stylelint \"**/*.css\"",
     "css-stylelint:fix": "stylelint \"**/*.css\" --fix",
-    "eslint": "eslint \"public/js/**/*.js\" \"cypress/**/*.js\" \"ts/**/*.js\" \"ts/**/*.ts\" \"tests/ts/**/*.js\"",
```
IDzyre, line 81: Will this now try and lint the node_modules folder? (or other folders that aren't ours to fix)

IDzyre (changes requested):
Reviewed the changes to the Cypress files, the spacing changes look good, but is there a reason that some of the inline function declaration uses parentheses and some don't?

dorian451:
Same reason as https://github.com/Submitty/Submitty/pull/10551#discussion_r1647790828. The eslint style rules aren't strict with whether parentheses are used, because I tried to keep them as close to the original as I could. I can add a rule to enforce them being there or not if you want.

site/package.json
dorian451: It hasn't on my machine or the CI, so no. The only files that would be linted are the ones specified in `eslint.config.js`, and eslint seems to auto-exclude node_modules. Ill edit eslint.config.js to be more explicit though.

dorian451: Ok with dba82dcfa0ea1a8524e9acccb3e21a7d374856f7 I made it so that eslint lints all js/ts files by default and ignores ones we aren't supposed to lint. The config block with `ignores` by itself is applies to the entire config, so even with running eslint with `eslint .` it will still ignore things it's supposed to.

site/.build.js
```diff
@@ -6,20 +6,21 @@ function getTwigTSFiles(dir) {
...
         }
-        
+
         return acc;
     }, []);
 }
 
 function getAllFiles(dir) {
-    return getTwigTSFiles(path.join(dir, "twig")).concat(fs.readdirSync(dir, { withFileTypes: true }).reduce((acc, entry) => {
-        if (!entry.isDirectory() && entry.name.endsWith('.ts') || entry.name.endsWith('.js')) {
```
IDzyre, line 19: Just want to make sure this functionality is what is wanted, this is checking if the file is not a directory, and it either ends with `.js` or `.ts`, correct?

IDzyre (approved):
Other than the comment about parentheses, the JS files look good to me, I don't know much (anything) about typescript, so it would be good to have someone who understands it to review also.

site/.build.js
dorian451: Yes, that's correct, since it wouldn't make sense we accepted a directory that suffixed with '.js' when we want a js file.

PR #10562, opened 2024-06-17
[DevDependency] Lint Cypress/Jest tests

dorian451:

Cypress and Jest tests are unlinted.

### What is the new behavior?
This PR enables the recommended rules for Cypress and Jest tests, and applies them to the affected files.


site/cypress/e2e/Cypress-Feature/late_days_allowed.spec.js
```diff
@@ -55,12 +59,13 @@ describe('Test cases involving the late days allowed page', () => {
...
             // add some late days for bitdiddle
-            cy.get('#user_id').type('bitdiddle').blur();
+            cy.get('#user_id').type('bitdiddle');
+            cy.get('#user_id').blur();
             cy.get('#datestamp').type('2021-01-01');
             cy.get('#user_id').click(); // dismiss the calendar view
             cy.get('#late_days').type('3');
             cy.get('input[type=submit]').click();
-            cy.wait(1000); // make sure the late day registered
+            cy.reload(); // make sure the late day registered
```
IDzyre, line 68: Does this need to reload? I haven't tested it locally, but this (and the waits) seem a bit unnecessary

site/cypress/e2e/Cypress-Feature/polls.spec.js
```diff
@@ -349,7 +362,10 @@ describe('Test cases revolving around polls functionality', () => {
         cy.get('.correct-box').eq(1).should('not.be.checked');
         cy.get('.poll_response').should('contain', 'Answer 3');
         cy.get('.correct-box').eq(2).should('be.checked');
-        cy.get('textarea').contains('Answer 1').clear().type('Answer 0');
+        cy.get('textarea').contains('Answer 1').then(($el) => {
```
IDzyre, line 365: Is there a reason to do this one differently than the others?

site/cypress/e2e/Cypress-System/autograding_status_2.spec.js
```diff
@@ -7,7 +7,7 @@ skipOn(Cypress.env('run_area') === 'CI', () => {
     describe('Pre autograding test', () => {
         before(() => {
             cy.login();
-            cy.wait(500);
+            // cy.wait(500);
```
IDzyre, line 10: Should this be removed instead of commented out?

site/cypress/e2e/Cypress-UI/navigation_page.spec.js
```diff
@@ -12,7 +12,7 @@ function validate_navigation_page_sections(sections) {
 describe('navigation page', () => {
     beforeEach(() => {
         cy.visit('/');
-        cy.wait(500);
+        // cy.wait(500);
```
IDzyre, line 15: same about remove instead of comment

site/cypress/support/commands.js
```diff
@@ -118,7 +118,6 @@ Cypress.Commands.add('checkLogoutInAfterEach', () => {
  * @param {int} wait
  */
 Cypress.Commands.add('waitAndReloadUntil', (condition, timeout, wait = 100) => {
-    cy.wait(wait);
```
IDzyre, line 121: I think this one is good to leave in, as that is the main purpose of the function

site/public/js/rainbow-customization.js
```diff
@@ -16,6 +16,7 @@ function ExtractBuckets() {
     $('#custom_form').submit();
 }
 
+// Forces element's value to be non-negative
```
IDzyre, line 19: Is this a relative change?

site/public/js/rainbow-customization.js
IDzyre, line 1: IS this file meant to be changed?

IDzyre (changes requested):

site/cypress/e2e/Cypress-Feature/polls.spec.js
dorian451: The reason why I changed this to begin with is because cypress recommends against chaining additionally queries/actions after doing something that can modify the state of the page (like `clear()`). The selector `textarea` matches multiple things on the page, and the original code relied on using `contains()` to filter which text area to type `'Answer 0'` in. Since I removed the chaining, I thought this would be concise/readable way to contnue using that specific textarea. I could also just assign an alias to it if you would prefer:
```suggestion
        cy.get('textarea').contains('Answer 1').as('answer1-text');
        cy.get('@answer1-text').clear();
        cy.get('@answer1-text).type('Answer 0');
```

site/public/js/rainbow-customization.js
dorian451: Oops; reverted in 3fc613f26491a579df720aa48cdf42d3fa7a1eaa

site/cypress/e2e/Cypress-Feature/polls.spec.js
IDzyre: I don't really prefer either way, it was just different than the other changes and I was wondering. I think it's okay to leave

dorian451: :+1:

site/cypress/e2e/Cypress-System/autograding_status_2.spec.js
dorian451: Fixed in https://github.com/Submitty/Submitty/pull/10562/commits/8107197e747bb610b3ae3ba2ba25d659559ebf33

site/cypress/e2e/Cypress-UI/navigation_page.spec.js
dorian451: Fixed in https://github.com/Submitty/Submitty/pull/10562/commits/8107197e747bb610b3ae3ba2ba25d659559ebf33

site/cypress/support/commands.js
dorian451: Fixed in https://github.com/Submitty/Submitty/pull/10562/commits/96a3021a32be35b884d51bc2d2187d98513634e1

site/eslint.config.js
```diff
@@ -106,18 +106,24 @@ module.exports = tseslint.config(
...
         },
+        rules: {
+            'no-restricted-syntax': [
+                'error',
+                {
+                    selector: '[type=CallExpression][callee.object.name=cy][callee.property.name=waitAndReloadUntil]',
+                    message: 'Do not wait for arbitrary time periods',
+                },
+            ],
+        },
```
williamjallen, line 121: While this might be a desirable change eventually, I don't think it's necessarily related to the basic task of linting the tests.  As @IDzyre noted, we can't just blindly remove waits in some cases.  While some of the waits may be unnecessary, many others were specifically added to address flakiness issues and I worry that removing them may cause undesirable consequences.

williamjallen (changes requested):

site/eslint.config.js
dorian451: This rule serves the same purpose cypress's built-in [no-unnecessary-waiting](https://github.com/cypress-io/eslint-plugin-cypress/blob/master/docs/rules/no-unnecessary-waiting.md) rule, with the only difference being that it applies to our custom `cy.waitAndReloadUntil()` function rather than cypress's `cy.wait()` function. Since `cy.waitAndReloadUntil()` uses `cy.wait()`, I believe usage of `cy.waitAndReloadUntil()` should be beholden to the same standard as `cy.wait()`. The point of these rules is to discourage the usage of these functions; if a test needs them for any reason, a comment can be put explaining why, along with an `eslint-disable-next-line` comment to disable the relevant rule as a one time exception.

For examples of each, look at the following:
https://github.com/Submitty/Submitty/pull/10562/files#diff-3d193f8781cad654997675b84fc39f55904d8c13d9e929433a8ac82aabb3c2e3R84-R87
https://github.com/Submitty/Submitty/pull/10562/files#diff-caa8c69680af44607887452b399ee8414370cd89028b959f3c3036c3bebf4998R54-R58

Additionally, it seems that all of the `wait`s this PR does remove were actually not needed, since the tests they were removed from repeatedly pass CI checks (this can be seen in my commit history, there were only two tests failing that I had to fix).

site/cypress/e2e/Cypress-Feature/late_days_allowed.spec.js
dorian451: Seems like it doesn't; fixed in https://github.com/Submitty/Submitty/pull/10562/commits/42e999e19e31ae96fa4f481aea4877f409139cf6

williamjallen (approved):
This looks good to me overall, but I'd like @IDzyre to take a look before this gets merged.

IDzyre (approved):
Looks good to me, We should refactor the older tests to use testid selectors, so some of the .within() and .wrap(el) things might become redundant, but for now this is good, and adds linting which should hopefully make future tests better.

PR #10566, opened 2024-06-17
[Bugfix:Developer] add vagrant-timezone to Vagrantfile

bmcutler:
Vagrant VM (on Mac OSX hosts) can experience clock skew.

### What is the new behavior?
This should help fix the problem.

You'll also need to add the plugin to the host (documentation has also been added with 
https://github.com/Submitty/submitty.github.io/pull/612)
```
vagrant plugin install vagrant-timezone
```
and run `vagrant halt` and `vagrant up`


Vagrantfile
```diff
@@ -266,6 +266,15 @@ Vagrant.configure(2) do |config|
 
   config.vm.provision :shell, :inline => " sudo timedatectl set-timezone America/New_York", run: "once"
 
+
+  # to use the command below, first install the vagrant plugin:
+  #
+  #   vagrant plugin install vagrant-timezone
+  #
+  if Vagrant.has_plugin?("vagrant-timezone")
+    config.timezone.value = "America/New_York"
```
ziesski, line 275:
```suggestion
    config.timezone.value = :host
```

Do we want the timezone to be hard-coded NY or match host machine's timezone?

ziesski: `def ensure_plugin_installed(plugin)
  begin
    unless Vagrant.has_plugin?(plugin)
      system "vagrant plugin install #{plugin}"
      print "Waiting for #{plugin} plugin installation to complete."
      until Vagrant.has_plugin?(plugin)
        print "."
        sleep 1
      end
      puts "\n#{plugin} plugin installation completed."
    end
  rescue => e
    raise "Failed to install #{plugin} plugin: #{e.message}"
  end
end

ensure_plugin_installed("vagrant-timezone")`


I was thinking about installing script by default, make it seamless. Not having to mention to each new dev, oh you need to install this plugin

bmcutler: Yes...  but Line 267 also has america/new york hard-coded.
(this was already an issue with the vagrantfile before my addition)

bmcutler:
This didn't fix the problem.  
time skew still appeared after closing laptop for a few hours.
ntp is running...  but is not keeping the VM up to date.

PR #10568, opened 2024-06-17
[Bugfix:InstructorUI] Unload course after checking for rejoin

### LINKED ISSUE #10565: Losing access to a course messes up the homepage sidebar
**Describe the bug**

* A student is registered for class A - they are in a valid section.
 
* The student loads a page for class A, for example the discussion forum.  This is successful.

* The instructor removes the student from class A - they are moved to the NULL section, they will no longer have access to private course pages.  NOTE: I'm not talking about "delete" student (trashcan icon) - that does not work if student has any access or submission data in the course.  

* The student still has the webpage open to the discussion forum.  They click to view another private course page, e.g., the Gradeables page.  They get access denied, e.g.,:

    Screenshot 2024-06-17 at 1 38 04 PM

* Clicking on the "Submitty" item in the breadcrumb should take them to the homepage of Submitty, showing all the courses they DO have access to.  However, the sidebar still contains links to the course they have LOST ACCESS to.  E.g. this is shown:

    Screenshot 2024-06-17 at 1 39 19 PM

Attempts to "fix" that don't work:
* Clicking to course B that the student does have access to works.  But still when we return to the homepage, it is still contaminated with links for course A (that are broken when the user tries to click on them).
* Force refreshing the browser and logging in and out of the site did not fix the problem.
* deleteing cookies and cached data from the browser does not fix the problem.

What DOES fix the problem:

* adding the student BACK into course A.
* student clicks on private pages for course A - which successfully load.
* Student clicks on the "Submitty" in the breadcrumb and successfully loads the list of courses they have access to.
* (however, if I remove the student again, the homepage listing courses is 'broken' again)

**Expected behavior**

* Clicking on the "Submitty" item from the breadcrumb should take us to a page that is not associated with any specific course.  It should not have a "Course Home" or "Gradeables" link.  It should start with "My courses" and "My profile", e.g.:

    Screenshot 2024-06-17 at 1 41 13 PM

**Configuration**

Tested with Mac Chrome on both an ARM and non-ARM machine.

powe97:
Fixes #10565. When a student loses access to a course, their Submitty front page shows the sidebar for the course they no longer have access to. This is because Submitty checks if students can rejoin a course they're no longer part of, this rejoin check loads the course config, and the course config is never unloaded, so the sidebar is rendered.

### What is the new behavior?
The fix is to simply unload the course config after checking if the student is permitted to rejoin.

### Other information?
Tested by removing and re-adding students to the Sample course in the VM, and checking their front pages.


ziesski (changes requested):
i was able to reproduce the issue, ARM macOS 14 Chrome, however, this fix does not seem to fix the original issue. 
screenshot

powe97:
Ok, so I've partially fixed the PR. Now, the sidebar is fixed after a student is removed from a course, however this also causes the sidebar to disappear when an instructor opens the Manage Students menu. What appears to be happening is

#### When a student is removed from a course while having it loaded
1. The student no longer has access and is kicked to the main screen.
2. Submitty checks if they can rejoin the course.
3. The course is unloaded (thanks to this PR), and the sidebar does not render on the main screen. All good.

#### When an instructor opens the Manage Students page
1. Submitty, for whatever reason, checks if students can rejoin the course that is already loaded.
2. Since the course we are using and the course we are checking for reload are the same course, `$reload_previous_course` isn't triggered.
3. This causes the course to be unloaded, so the sidebar disappears!

I'm going to try figure out what's going on in more detail and mark this as a draft for now.

powe97:
A-ha: turns out that the rejoin check is also performed in `UsersController` on each `NULL` section student. This is what causes that sidebar to disappear. 
https://github.com/Submitty/Submitty/blob/924dd2c06c3888a16c3394a4df9f321641d31762/site/app/controllers/admin/UsersController.php#L91-L103
It looks like we do not actually need the bulk of `canRejoinCourse()` here. `canRejoinCourseHelper()` is what actually checks for rejoinability, and `canRejoinCourse()` is a wrapper around it that handles getting the user if it isn't provided, and unloading and reloading another course if one is loaded already for some reason. In this case, this functionality is unneeded.

powe97:
Pushed changes to use `canRejoinCourseHelper()` instead in `UsersController()`. I recognize that converting a private method to public is generally a bad idea, but I think it makes the most sense here.

dorian451 (approved):
Seems to work

ziesski (approved):
Tested and the fix works, code seems reasonable with minimal change.

zacharymnp (approved):

PR #10603, opened 2024-06-24
[UI/UX:TAGrading] Hide notebook control buttons

williamjallen:
### What is the current behavior?
The TA grading interface currently shows control buttons in the notebook panel.  These buttons serve no purpose in this context.

image

### What is the new behavior?
This PR removes the control buttons:
image

PR #10604, opened 2024-06-24
[UI/UX:TAGrading] Disable notebook fields when grading

williamjallen:
The answer fields in the notebook panel of the TA grading interface are only disabled for submissions other than the active submission.  These fields should always be disabled in the TA grading interface.

### What is the new behavior?
All notebook fields are disabled in the TA grading interface.  In

### Other information?
In the future, it would be good to put together a set of comprehensive tests for the notebook panel of the TA grading interface.


sahilsuman933 (approved):
Looks good to me.
image

This #10603 and #10604 fix this issue #10619

PR #10616, opened 2024-06-24
[Feature:Developer] Add FROM_SCRATCH and PREBUILT_VERSION

IDzyre:

If you want to do a full vagrant up without using the prebuilt vagrant box, you use `BASE_BOX=1 vagrant up`

Also, there is no way to specify the version of the prebuilt box without editing the vagrant file. 
### What is the new behavior?
`BASE_BOX=1` variable name is somewhat confusing, so `FROM_SCRATCH=1` is the new method, and `BASE_BOX=1` will be deprecated, and removed sometime in the future. 

`PREBUILT_VERSION={version}` has been added as a supported ENV variable.
 **NOTE** 
Use JUST the numbers and the decimals, NOT the leading V. 
An example :
 `PREBUILT_VERSION=24.04.01.2405050238 vagrant up` for mac and linux or 
`SET PREBUILT_VERSION=24.04.01.2405050238`
`vagrant up` for windows

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -130,9 +130,9 @@ Vagrant.configure(2) do |config|
   arch = `uname -m`.chomp
   arm = arch == 'arm64' || arch == 'aarch64'
   apple_silicon = Vagrant::Util::Platform.darwin? && (arm || (`sysctl -n machdep.cpu.brand_string`.chomp.start_with? 'Apple M'))
-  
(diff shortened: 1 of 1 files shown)


ziesski (approved):
Tested on intel MacOS 12, with VirtualBox, works as expected, code looks good too.

GO-viper7 (approved):
Tested on Windows, works as expected
I cannot find any errors in CLI and localhost's up and good after running commands.

PR #10624, opened 2024-06-26
[Bugfix:Forum] Auto-resizing for all reply_box

### LINKED ISSUE #10623: When editing Submitty Forum message, the text area is not resizeable and has no scrollbar.
**Describe the bug**
When writing/editing a message in the Submitty Forum, there is no way to resize the text area, and there are no scroll bar shown.

**Expected behavior**
There should be a way to grab the corner of the text area and drag it to increase the size. Also, if not all contents fit in the text area, scroll bar must be shown,

**To Reproduce**
1. Navigate to the Forum.
2. Create a new thread and write more lines of message than the default height of the text area.
3. See the absence of either scroll bar or the way to increase the vertical size of the area.

**Configuration**
 - OS: Windows 11 Home Version 23H2 OS build 22631.3593 Experience Windows Feature Experience Pack 1000.22700.1003.0
 - Browser: Mozilla Firefox 127.0 (64-bit)

**Screenshots**
firefox_2BQ7uQRjt4

**Additional context**
I believe that previously it was possible to resize the text area but this functionality seems to be gone (not sure whether it was intentional or not).

ZheyuDeng:
Fix #10623 
### What is the new behavior?
Auto resize the markdown for all reply box base on input.

https://github.com/Submitty/Submitty/assets/122962727/a93088ee-e4de-4643-97ba-7b300b13671f

As we discussed we will work toward combining dragling with auto resizing in the future.


zacharymnp (changes requested):
I've noticed that the markdown resizing has a lot of trouble when many words are added in a single line.
https://github.com/Submitty/Submitty/assets/56311867/c82cf3dd-28e6-467c-9674-bda19181eee4

GO-viper7 (changes requested):
It works in my local as shown, textarea adjusts height according to the content inside
and I'm able to elongate the area vertically
But this elongation is limitless, may be its better to limit this by `dynamic height of content + 200px`

zacharymnp (approved):
tested functionality

ziesski (approved):
Code seems reasonable.

PR #10628, opened 2024-06-26
[Bugfix:Developer] Re-add student history function

IDzyre:
#10551 had a bad merge conflict resolution, as it removed a function in the code instead of linting it. 

### What is the new behavior?
This function is re-added and linted properly.


ziesski (approved):

dorian451 (approved):
Sorry for accidentally getting rid of this

PR #10635, opened 2024-06-27
[Feature:TAGrading] Add a clear conflict button

### LINKED ISSUE #10424: Clearing a version conflict takes many clicks
**Describe the bug**
If a student shows a version conflict on the TA grading page, either the active submission version needs to be changed back to the version that the TA/instructor graded -- OR -- the new version needs to be confirmed for each component.

Confirming the grade for each component used to be just opening and closing each component (I believe this was the case).

But it appears now that more clicks are needed, that the marks from the original grader need to be unchecked, then the component saved, then opened and re-marked.  This also changes the grader.

**Expected behavior**
It should be easier to clear a version conflict, and it should not require changing the "who graded this" label on the component.

dorian451:

Closes #10424

### What is the new behavior?
Allows graders to simulaneonous resolve all version conflicts for a gradable for a given student, by pressing the "clear version conflicts" button.

### Other information?

No.

1.  Grade a version of a student's work.
2.  Change the active version for a student.
3.  Observe that the button now appears.
4.  Click the button.

preview


oliiso (approved):
Looks great. UI is good and clears the version conflicts well.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -2339,6 +2339,84 @@ public function saveGradedComponent(TaGradedGradeable $ta_graded_gradeable, Grad
...
+            "gradeable_id" => $gradeable_id,
+            "grader_id" => $this->core->getUser()->getId(),
+            "component_ids" => implode(",", $component_ids),
+            "action" => "CHANGE_GRADEABLE_GRADED_VERSION",
+            "submitter_id" => $submitter_id
+        ];
+        Logger::logTAGrading($logger_params);
+
+        try {
+            // do thing
```
williamschen23, line 2408: do thing

manzil-infinity180:
do it using data-testid rather than selecting element using id and class 
cypress best practices - https://docs.cypress.io/guides/references/best-practices

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -2339,6 +2339,84 @@ public function saveGradedComponent(TaGradedGradeable $ta_graded_gradeable, Grad
...
+     */
+    #[Route("/courses/{_semester}/{_course}/gradeable/{gradeable_id}/grading/graded_gradeable/change_grade_version", methods: ["POST"])]
+    public function ajaxUpdateGradedVersionForStudent(string $gradeable_id): void {
+        $anon_id = $_POST['anon_id'] ?? null;
+        $graded_version = intval($_POST['graded_version'] ?? null);
+        $component_ids = $_POST['component_ids'] ?? [];
+
+        if ($anon_id === null) {
+            $this->core->getOutput()->renderJsonFail('Missing anon_id parameter');
+            return;
```
williamjallen, line 2353: It would be better to return `JsonResponse::getFailResponse(...)` instead of calling renderJsonFail directly.  Apply throughout.

site/app/controllers/grading/ElectronicGraderController.php
williamjallen: this comment should either be removed or improved.

williamjallen (changes requested):

site/cypress/e2e/Cypress-TAGrading/clearing_version_conflicts.spec.js
```diff
@@ -0,0 +1,41 @@
...
+        cy.get('[data-testid="change-graded-version"]').should('not.exist');
+        cy.get('.version-warning').should('not.exist');
+    });
+
+    it('Button should appear if there are version conflicts', () => {
+        cy.login('instructor');
+        cy.visit(['sample', 'gradeable', 'grading_homework', 'grading', 'grade?who_id=K8jI3q4qpdCc1jw&sort=id&direction=ASC&gradeable_version=1']);
+        cy.get('#grading_rubric_btn').click();
+        cy.get('[data-testid="change-graded-version"]').should('exist');
+        cy.get('.version-warning').should('exist');
```
IDzyre, line 15: These should still be able to use testids, as they are editable in the Twig files

IDzyre (changes requested):

site/cypress/e2e/Cypress-TAGrading/clearing_version_conflicts.spec.js
dorian451: I figured I shouldn't since this PR doesn't touch those elements, should I add them anyway?

IDzyre: You can edit elements that need testids, because maybe later another test will use the same elements, and need the testids also. These are small changes, and easy to fix conflicts if any arise.

dorian451: Ok I added the test ids for the elements here: https://github.com/Submitty/Submitty/pull/10689. I'll update this PR's test once that's merged.

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -2365,6 +2365,78 @@ public function saveGradedComponent(TaGradedGradeable $ta_graded_gradeable, Grad
...
+
+        // Get the graded gradeable
+        $graded_gradeable = $this->tryGetGradedGradeable($gradeable, $submitter_id);
+        if ($graded_gradeable === false) {
+            return null;
+        }
+
+
+        // check if all the components provided are real and the user has permission
+        foreach ($component_ids as &$component_id) {
```
ziesski, line 2407:
```suggestion
        foreach ($component_ids as $component_id) {
```
Should we remove `&` from foreach ($component_ids as &$component_id) since there was no modification of `$component_id` in the loop?

ziesski (changes requested):

site/cypress/e2e/Cypress-TAGrading/clearing_version_conflicts.spec.js
dorian451: Fixed in https://github.com/Submitty/Submitty/pull/10635/commits/2e017659349bacf01807dca03e21011458b7b6b2

ziesski (approved):
Code looks good to me.

site/cypress/e2e/Cypress-TAGrading/clearing_version_conflicts.spec.js
```diff
@@ -0,0 +1,41 @@
...
+        cy.login('instructor');
+        cy.visit(['sample', 'gradeable', 'grading_homework', 'grading', 'grade?who_id=K8jI3q4qpdCc1jw&sort=id&direction=ASC&gradeable_version=1']);
+        cy.get('[data-testid="grading-rubric-btn"]').click();
+        cy.get('[data-testid="change-graded-version"]').should('exist');
+        cy.get('[data-testid="version-warning"]').should('exist');
+    });
+
+    it('Clicking the button should resolve those version conflicts', () => {
+        cy.login('instructor');
+        cy.visit(['sample', 'gradeable', 'grading_homework', 'grading', 'grade?who_id=K8jI3q4qpdCc1jw&sort=id&direction=ASC&gradeable_version=1']);
```
IDzyre, line 20: Would combining these tests into one make it slower at all? It seems like these could all be tested in the same test to reduce the amount of times the login is run, and since they are essentially three line tests, I think it would be good to combine all three into one slightly bigger test

IDzyre (changes requested):

site/cypress/e2e/Cypress-TAGrading/clearing_version_conflicts.spec.js
dorian451: Combing them probably wouldn't negatively impact performance. I thought it was cleaner / easier to debug if they were in separate tests but if it doesn't matter that much I'm ok with combining them.

dorian451: Fixed in https://github.com/Submitty/Submitty/pull/10635/commits/1eb30dd58dd98393e45e9b61ffc1cdde0aeac89a

site/app/controllers/grading/ElectronicGraderController.php
```diff
@@ -2365,6 +2365,78 @@ public function saveGradedComponent(TaGradedGradeable $ta_graded_gradeable, Grad
...
+        }
+
+        if (count($component_ids) < 1) {
+            return JsonResponse::getFailResponse('Missing component_ids parameter');
+        }
+
+        // Get the gradeable
+        $gradeable = $this->tryGetGradeable($gradeable_id);
+        if ($gradeable === false) {
+            return null;
```
IDzyre, line 2391: Should this return a fail response instead of null? Same with other null return values?

dorian451: Technically the tryGetGradables already prints the fail response so by returning a fail response I think we would be sending two fail responses. Or would returning a fail response overwrite that?

IDzyre: I believe that most of them have a 'render_json' bool variable, which you can pass to not render JSON from inside the function, and instead return a JsonResponse in this file. 

When looking through the codebase, it may be eventually good to refactor some of these functions to return a JsonResponse (Success or Fail), and then return that value if it is a fail response, as they have more descriptive error messages.

dorian451: Fixed b8fdb7e28ba93ae8e06cc0a8de73cae573df293d

IDzyre (approved):
Code looks good to me, tests look good, I haven't tested the functionality locally

PR #10638, opened 2024-06-27
[Feature:TAGrading] PDF Previews for DOCX files

### LINKED ISSUE #10454: Add viewer for .docx files in TA grading file browser
If a student submits a .docx file as part of their gradeable submission, this file cannot be viewed within Submitty.  Currently if you click on it, it displays as an unreadable wall of special characters.   The only thing we can do is download the file (or download the directory of all files) and use a .docx viewer on our native machine.

* Can we add a simple inline .docx viewer?  or

* We could run a .docx -> .pdf converter as part of the autograding process.  Downside: the instructor would have to configure this as part of the gradeable configuration.  Also, this can get complicated if the student is allowed to name the file whatever they want, or if there are muliple .docx files within the submission and/or there .docx files in subdirectories with arbitrary names, etc.  or

* We could run a .docx -> .pdf converter one file at a time upon request from the TA within the TA grading side.  That file could get added to the results_public folder (so the student could view it from the submission page as well).  This could get complicated as well (is it expensive, non-trivial time, do we put it in a queue, what if 2 TAs request the same file to be converted - do we have a race condition).

skara9:

Submitted word documents need to be downloaded by grader.
Resolves #10454 

### What is the new behavior?
Word documents can now be viewed as PDF's in the browser without having to be downloaded.

### DIFF
--- a/.setup/install_submitty/install_bin.sh
+++ b/.setup/install_submitty/install_bin.sh
@@ -65,6 +65,10 @@ rsync -rtz  ${SUBMITTY_REPOSITORY}/autograder/* ${SUBMITTY_INSTALL_DIR}/autograd
 find ${SUBMITTY_INSTALL_DIR}/sbin -type f -exec chown root:root {} \;
 find ${SUBMITTY_INSTALL_DIR}/sbin -type f -exec chmod 500 {} \;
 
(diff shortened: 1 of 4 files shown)


skara9:
Superceded by #10705

PR #10639, opened 2024-06-27
[Bugfix:Developer] Move NTP fix after CLI args check

powe97:
Currently, the NTP time skew fix is ran *before* the SUBMITTY_INSTALL_HELPER script checks for CLI arguments and that the user is root. This means that running something like `submitty_install --help` only prints the help message after waiting a few seconds for the NTP skew fix to run.

### What is the new behavior?
This PR moves the NTP time skew fix immediately after the argument parsing and root check.


williamjallen (approved):
LGTM.  I'll let Barb take a final look and merge.

PR #10640, opened 2024-06-28
[Feature:Developer] Add CI flag to setup scripts

IDzyre:
Vagrant up jobs are hanging at the setup of NTP because of a 'soliciting pool' infinite loop, which seemingly only happens in CI. 
Currently we don't really have a way to tell if the Vagrant Up job is happening in CI from within the setup scripts.

### What is the new behavior?
The `-c` optional parameter has been added to the setup_vagrant.sh file . If you have the `CI=1` environment variable set, and you run vagrant up, it will assume that you are running it in CI, and will pass the `-c` flag to setup_vagrant.sh. This creates a file in the root directory of Submitty, with the contents `This file is used to check if Submitty is being run in the Github Actions CI.`. Later in the setup scripts, you can test for the existence of this file, and therefor detect if the script is being run in CI, and skip some optional, or breaking steps to the setup process.

### Other information?
This can be tested locally by setting the environment variable CI=1, and running `vagrant up`. You should notice soon after the VM is setup, and all folders have been synced a message saying `Submitty is being run in the Github Actions CI.`
https://github.com/IDzyre/Submitty/actions/runs/9707159611/job/26791907078 -- old way
https://github.com/IDzyre/Submitty/actions/runs/9718510404 show this works


.setup/vagrant/setup_vagrant.sh
```diff
@@ -14,6 +14,16 @@ if [ -x "$(command -v pip3)" ]; then
     python3 ${SUBMITTY_REPOSITORY}/.setup/bin/reset_system.py
 fi
 
+OPTSTRING=":c"
+while getopts ${OPTSTRING} opt; do
+    case ${opt} in
+    c)
+    echo "This file is used to check if Submitty is being run in the Github Actions CI." > ${SUBMITTY_REPOSITORY}/.github_actions_ci_flag
```
williamjallen, line 21: Writing a file for one-time use like this feels like a hack.  Is there a way to use an environment variable instead?

.setup/vagrant/setup_vagrant.sh
```diff
@@ -14,6 +14,16 @@ if [ -x "$(command -v pip3)" ]; then
     python3 ${SUBMITTY_REPOSITORY}/.setup/bin/reset_system.py
 fi
 
+OPTSTRING=":c"
```
williamjallen, line 17: `-c` isn't very descriptive.  Would `--ci`, or the environment variable `CI=1` be a better option?

.setup/vagrant/setup_vagrant.sh
IDzyre: Talked to Barb in the meeting yesterday and she said that since vagrant and utm are doing this, she liked this way. We can discuss more in the meetings

.setup/vagrant/setup_vagrant.sh
IDzyre: This has been updated.

williamjallen (commented):
Looks reasonable to me.  @IDzyre, do you have an example CI run on this branch to show that it works?

IDzyre:
In the PR description there is a run with the old system, I just started a new one at https://github.com/IDzyre/Submitty/actions/runs/9718510404

williamjallen (approved):

micpap25:
Hi @williamjallen @IDzyre I believe this PR may have broken something in the install script? I am getting this issue whenever I run it. It still successfully completes, as far as i can tell.
image

PR #10645, opened 2024-06-28
[Refactor:System] Refactor dir setup & globals in install

powe97:
See issue #4011. Currently, `SUBMITTY_INSTALL_HELPER.sh`, the script called by `submitty_install`, is a giant script that does many things to set up the system. It is desirable to split it up into smaller scripts so that parts can be run separately.

### What is the new behavior?
This PR moves a bunch of code that sets global variables used by the install script to `.setup/install_submitty/get_globals.sh`, and moves the directory setup code to `.setup/install_submitty/setup_directories.sh`

In this PR, the code is directly cut-and-pasted to the new `setup_directories.sh` script, which is ignored by ShellCheck. Once this is merged, I plan to actually fix the ShellCheck issues in a new PR; they're pretty trivial but I was advised to keep it separate from a refactor.

### Testing
Testing was done using the Github CI, since for some reason its environment wasn't the same as my local machine.


powe97:
Marking as draft because clearly it does not work on the CI.

powe97:
Ah, the problem is environment variables not getting passed to the other script. I have no idea why the script works locally.

powe97:
```
Stopping all remote machine workers...
  File "/usr/local/submitty/sbin/shipper_utils/systemctl_wrapper.py", line 27, in <module>
    with open(AUTOGRADING_WORKERS_PATH, 'r') as open_file:
PermissionError: [Errno 13] Permission denied: '/usr/local/submitty/sbin/shipper_utils/../../config/autograding_workers.json'
Error: Process completed with exit code 1.
```
What the heck did I do?

dorian451 (approved):
Works for me

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -26,35 +26,10 @@ set -e
...
-if [ -f "${THIS_DIR}/../.github_actions_ci_flag" ]; then
-    CI=1
-fi
-
 SUBMITTY_REPOSITORY=$(jq -r '.submitty_repository' "${CONF_DIR}/submitty.json")
-SUBMITTY_INSTALL_DIR=$(jq -r '.submitty_install_dir' "${CONF_DIR}/submitty.json")
-WORKER=$([[ $(jq -r '.worker' "${CONF_DIR}/submitty.json") == "true" ]] && echo 1 || echo 0)
 
-source "${THIS_DIR}/bin/versions.sh"
+source "${SUBMITTY_REPOSITORY}/.setup/install_submitty/get_globals.sh"
```
bmcutler, line 31: why don't we do:

source "$( dirname "${BASH_SOURCE[0]}" )/get_globals.sh"

Like you have in line 4 of setup_directories.sh?

.setup/install_submitty/setup_directories.sh
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env bash
+set -ve
+
+source "$( dirname "${BASH_SOURCE[0]}" )/get_globals.sh"
+
```
bmcutler, line 5: should we add the check that this is run as root user at the top of this script (and all other helper install scripts)?

.setup/INSTALL_SUBMITTY_HELPER.sh
powe97: `setup_directories.sh` and `INSTALL_SUBMITTY_HELPER.sh` are in different directories, so this would not find the file correctly as `BASH_SOURCE` differs. I'm using `SUBMITTY_REPOSITORY` here to get an absolute path instead of a relative one.

.setup/install_submitty/setup_directories.sh
powe97: Sure. I left that in `SUBMITTY_INSTALL_HELPER.sh` without thinking much of it; we could put a check for `root` in `get_globals.sh` because it gets called by everything else.

bmcutler:
@powe97 
Sorry, I don't like the change to put the sudo check in get globals.
Just skimming quickly, these global variables are (semi)publicly readable.
The root user check was meant to be on scripts that defninitely should only be run as root.
Its ok to have that duplicated.  I like to see it explicitly.
(Perhaps not all scripts will require root to run them)

powe97:
OK, sounds good

.setup/install_submitty/get_globals.sh
```diff
@@ -0,0 +1,95 @@
...
+if [ "${WORKER}" == 0 ]; then
+    DAEMONPHP_GROUP=$(jq -r '.daemonphp_group' "${SUBMITTY_INSTALL_DIR}/config/submitty_users.json")
+else
+    DAEMONPHP_GROUP="${DAEMON_GROUP}"
+fi
+DAEMONCGI_GROUP=$(jq -r '.daemoncgi_group' "${SUBMITTY_INSTALL_DIR}/config/submitty_users.json")
+DAEMONPHPCGI_GROUP=$(jq -r '.daemonphpcgi_group' "${SUBMITTY_INSTALL_DIR}/config/submitty_users.json")
+SUPERVISOR_USER=$(jq -r '.supervisor_user' "${SUBMITTY_INSTALL_DIR}/config/submitty_users.json")
+
+export THIS_DIR
```
bmcutler, line 70: I'm worried about the THIS_DIR variable.
It isn't very descriptive.
And since we are refactoring & adding new scripts & moving toward allowing these scripts to be run in smaller chunks (one at a time), and those scripts might be in different directories, it's going to be easy to make mistakes.

It should not be exported to the other scripts with this name.
Instead I think we should use a relative path from SUBMITTY_REPOSITORY or SUBMITTY_INSTALL_DIR...

powe97: I agree it's poorly named; `THIS_DIR` was copied directly from the helper script. The problem with using a relative path from `SUBMITTY_REPOSITORY` or `SUBMITTY_INSTALL_DIR` is that those two variables are fetched from `${CONF_DIR}/submitty.json`, and to get `CONF_DIR` we need `THIS_DIR`. In the helper script, `THIS_DIR` is necessary to find `get_globals.sh`. With that said, we can remove the export since it is not used beyond this.

powe97: Also, note that the `BASH_SOURCE` variable used to get `THIS_DIR` is relative to `get_globals.sh`, not the script sourcing it, so it's fine as a local variable.

.setup/install_submitty/get_globals.sh
```diff
@@ -0,0 +1,95 @@
...
+export CONF_DIR
```
bmcutler, line 71: Let's also rename CONF_DIR to SUBMITTY_CONFIGURATION_DIR

.setup/install_submitty/get_globals.sh
```diff
@@ -0,0 +1,95 @@
...
+export VAGRANT
+export UTM
+export CI
+export SUBMITTY_REPOSITORY
+export SUBMITTY_INSTALL_DIR
+export WORKER
```
bmcutler, line 77: And rename these variables...
VAGRANT => IS_VAGRANT
UTM => IS_UTM
CI => IS_CI
WORKER => IS_WORKER

bmcutler (changes requested):
more requested edits...

.setup/install_submitty/get_globals.sh
bmcutler: The .setup directory in the REPOSITORY?  
Or the .setup directory in the INSTALLATION DIRECTORY?
(yeah, the fact that I'm asking is part of the problem)

No need to make a global variable for either one, just relative off of whichever

powe97: It's the `.setup` directory that the script is inside, which on the VM will always be the installation directory. But yes I see the problem.

powe97: Nevermind, it's actually the one in the repository— I thought the install dir was where the scripts were run from; this is not the case. This is confusing.

powe97: I've removed the export because the variable isn't actually used anywhere outside of `get_globals.sh`.

powe97:
@bmcutler Vagrant up works now!

.setup/install_submitty/get_globals.sh
```diff
@@ -0,0 +1,94 @@
...
+if [ "${IS_WORKER}" == 0 ]; then
+    DAEMONPHP_GROUP=$(jq -r '.daemonphp_group' "${SUBMITTY_INSTALL_DIR}/config/submitty_users.json")
+else
+    DAEMONPHP_GROUP="${DAEMON_GROUP}"
+fi
+DAEMONCGI_GROUP=$(jq -r '.daemoncgi_group' "${SUBMITTY_INSTALL_DIR}/config/submitty_users.json")
+DAEMONPHPCGI_GROUP=$(jq -r '.daemonphpcgi_group' "${SUBMITTY_INSTALL_DIR}/config/submitty_users.json")
+SUPERVISOR_USER=$(jq -r '.supervisor_user' "${SUBMITTY_INSTALL_DIR}/config/submitty_users.json")
+
+export SUBMITTY_CONFIGURATION_DIR
```
bmcutler, line 70: why do we need to export this?
I understand there is a bit of chicken & egg bootstrapping needed at the start
But ultimately, there are just 2 top level directory paths, the repositories location & the install location (/usr/local/submitty). [ And also the data director (/var/local/submitty) ].  
The config path is relative off of the install dir path.
The setup location is just a relative path off of the repo dir path.

bmcutler (changes requested):

.setup/install_submitty/get_globals.sh
powe97: `SUBMITTY_CONFIGURATION_DIR` is actually used in the rest of the helper script which is why I left it in:
https://github.com/Submitty/Submitty/blob/0652a11c09e7327b6ca98f63696d1bda1c01613f/.setup/INSTALL_SUBMITTY_HELPER.sh#L111
However it looks like it doesn't actually need to be exported. I can just remove it.

powe97:
Can I please get a review on this?

williamjallen:
I just kicked off a vagrant up job [here](https://github.com/Submitty/Submitty/actions/runs/10051721522).

powe97:
@williamjallen 

Good news!
image

.setup/install_submitty/get_globals.sh
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env bash
+set -ve
+
+# We assume a relative path from this repository to the installation
+# directory and configuration directory.
+SETUP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )/.."
```
bmcutler, line 6: Sorry to keep harping on this variable (I know you didn't add it), but I don't think we need a special variable name for this.  It is just SUBMITTY_INSTALLATION_DIR + ".setup".

Can't we just define here:
SUBMITTY_INSTALLATION_DIR = "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )/../.."
(adding an extra "/..")

And then eliminate SETUP_DIR everywhere.

.setup/install_submitty/get_globals.sh
```diff
@@ -0,0 +1,111 @@
...
+if [ -d "${SETUP_DIR}/../.utm" ]; then
+    IS_UTM=1
+fi
+
+IS_CI=0
+if [ -f "${SETUP_DIR}/../.github_actions_ci_flag" ]; then
+    IS_CI=1
+fi
+
+SUBMITTY_REPOSITORY=$(jq -r '.submitty_repository' "${SUBMITTY_CONFIGURATION_DIR}/submitty.json")
```
bmcutler, line 24: And then here, let's do an "assert" that the variable SUBMITTY_INSTALL_DIR we've defined above matches the value stored in this config file.  If somehow the value doesn't match, crash installation with an appropriate error.

bmcutler (changes requested):

.setup/install_submitty/get_globals.sh
powe97: Didn't think of that, yeah that works. I'll do that now.

.setup/install_submitty/get_globals.sh
powe97: Sounds good

powe97:
image
Uhoh. I'm not sure which is correct.

.setup/install_submitty/get_globals.sh
```diff
@@ -0,0 +1,124 @@
...
+set -ve
+
+# First, we determine SUBMITTY_REPOSITORY from BASH_SOURCE, as this
+# script is run from within GIT_CHECKOUT/Submitty. We use readlink to
+# get an absolute path from a relative path here.
+SUBMITTY_REPOSITORY="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )/../.."
+SUBMITTY_REPOSITORY="$(readlink -ne "$SUBMITTY_REPOSITORY")"
+echo "SUBMITTY_REPOSITORY=${SUBMITTY_REPOSITORY}" >&2
+
+SUBMITTY_INSTALL_DIR="$(readlink -ne "${SUBMITTY_REPOSITORY}/../..")"
```
bmcutler, line 11: this line doesn't make sense to me.  We should not assume this relationship between these 2 directories.

bmcutler (changes requested):

.setup/install_submitty/get_globals.sh
powe97: Hmmm... I see the problem but this is a little bit tricky to deal with. Right now, SUBMITTY_INSTALL_HELPER and company are run from within the repo instead of within the install dir. After looking through the rest of the code in this PR, this relationship is also assumed in this relative path in SUBMITTY_INSTALL_HELPER:
https://github.com/Submitty/Submitty/blob/787d35a2da9f669c6cbb6668aca9d557057857ba/.setup/INSTALL_SUBMITTY_HELPER.sh#L36
So we need a better way to deal with this. One thing we could do is copy scripts from the repo to the install dir before SUBMITTY_INSTALL_HELPER is run. Alternatively we can export a variable containing the location of the install dir before calling SUBMITTY_INSTALL_HELPER. I think copying scripts is probably a better way to do this, so I will work on that.

powe97:
Running `vagrant up` on Github actions right now to make sure installation isn't broken: https://github.com/Submitty/Submitty/actions/runs/10218252891.

powe97:
Vagrant up passed.

bmcutler (changes requested):
config directory and/or installdir = /usr/local/submitty/ should only be hardcoded in one place, not multiple.
pass in as command line argument if necessary.

.setup/INSTALL_SUBMITTY_HELPER.sh
```diff
@@ -133,8 +105,8 @@ fi
 ################################################################################################################
 # VALIDATE DATABASE SUPERUSERS
 
-if [ "${WORKER}" == 0 ]; then
-    DATABASE_FILE="$CONF_DIR/database.json"
+if [ "${IS_WORKER}" == 0 ]; then
+    DATABASE_FILE="$SUBMITTY_INSTALL_DIR/config/database.json"
```
IDzyre, line 109: Is this different from $SUBMITTY_CONFIG_DIR?

powe97: No, they are the same. In this script I avoided using `SUBMITTY_CONFIG_DIR` so that we can remove that variable in the future with another refactor.

.setup/bin/update_repos.sh
```diff
@@ -13,9 +13,15 @@ if [[ "$UID" -ne "0" ]] ; then
...
-MY_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
-SUBMITTY_REPOSITORY=${MY_PATH}/../..
+# Get arguments
+for cli_arg in "$@"
+do
+    if [[ $cli_arg =~ ^config=.* ]]; then
+        SUBMITTY_CONFIG_DIR="$(echo "$cli_arg" | cut -f2 -d=)"
+    fi
+done
+
```
bmcutler, line 23: this script should crash with an appropriate error message if SUBMITTY_CONFIG_DIR is not successfully set above

.setup/install_submitty/get_globals.sh
```diff
@@ -0,0 +1,105 @@
...
+set -ve
+
+# Get arguments
+for cli_arg in "$@"
+do
+    if [[ $cli_arg =~ ^config=.* ]]; then
+        SUBMITTY_CONFIG_DIR="$(readlink -f "$(echo "$cli_arg" | cut -f2 -d=)")"
+    fi
+done
+
```
bmcutler, line 11: same as above...
(this script should crash with an appropriate error message if SUBMITTY_CONFIG_DIR is not successfully set)

.setup/install_submitty/install_bin.sh
```diff
@@ -7,15 +7,22 @@
...
-CGI_USER=$(jq -r '.cgi_user' ${CONF_DIR}/submitty_users.json)
-DAEMON_USER=$(jq -r '.daemon_user' ${CONF_DIR}/submitty_users.json)
+# Get arguments
+for cli_arg in "$@"
+do
+    if [[ $cli_arg =~ ^config=.* ]]; then
+        SUBMITTY_CONFIG_DIR="$(readlink -f "$(echo "$cli_arg" | cut -f2 -d=)")"
+    fi
+done
+
```
bmcutler, line 17: same as above... crash with error message

.setup/install_submitty/install_site.sh
```diff
@@ -51,26 +51,20 @@ set_mjs_permission () {
...
+# Get arguments
+for cli_arg in "$@"
 do
-    if [ "$arg" == "browscap" ]; then
+    if [[ $cli_arg =~ ^config=.* ]]; then
+        SUBMITTY_CONFIG_DIR="$(readlink -f "$(echo "$cli_arg" | cut -f2 -d=)")"
+    elif [ "$cli_arg" == "browscap" ]; then
         BROWSCAP=true
     fi
 done
```
bmcutler, line 63: same...  add error & message

.setup/install_submitty/setup_directories.sh
```diff
@@ -0,0 +1,202 @@
...
+set -ve
+
+# Get arguments
+for cli_arg in "$@"
+do
+    if [[ $cli_arg =~ ^config=.* ]]; then
+        SUBMITTY_CONFIG_DIR="$(readlink -f "$(echo "$cli_arg" | cut -f2 -d=)")"
+    fi
+done
+
```
bmcutler, line 11: same

bmcutler (dismissed):
Now that you're adding command line arguments to install_site & install_bin, our development documentation needs to change... right?

But actually since we're running these scripts FROM the installed directory (/usr/local/submitty/.setup), we actually don't need the config directory, because we can assume its in ../config.

powe97:
Turns out there is even more hardcoding that I missed from before, which is why the Cypress tests are failing. The script `.setup/bin/update_repos.sh` is called during setup *without* a config argument, because it is called from `.setup/install_system.sh` instead of `.setup/INSTALL_SUBMITTY_HELPER.sh`. So, now I need to refactor `install_system.sh` to account for the new command line argument.

powe97:
The most recent commits here fixed an issue that went unnoticed by me for around 2 months. Specifically, this PR worked locally with `vagrant up` but it did not work in the CI. This is because there is a separate script run when the `e2e-Setup-Composite` action is called, which I never noticed and didn't update to have a config directory argument. Now, it does, and it appears that `e2e-Setup-Composite` works as intended.

PR #10649, opened 2024-06-30
[Feature:Submission] Hide Accessibility Text for Disabled

### LINKED ISSUE #10643: CodeMirror accessibility text should not be shown when area is disabled
**Describe the bug**
CodeMirror currently displays accessibility help text above all code boxes.  This text is confusing in places where the code box is disabled and it is impossible to navigate to it.
image

**Expected behavior**
The text should not be shown if a code box is disabled.

sahilsuman933:
Fix #10643

CodeMirror currently displays accessibility help text above all code boxes. This text is confusing in places where the code box is disabled and it is impossible to navigate to it.

### What is the new behavior?
This commit updates the `getLargeCodeMirror` function to conditionally display the accessibility help text only when the CodeMirror instance is enabled. This prevents confusion in scenarios where the code box is disabled and navigation is not possible.

image

### DIFF
--- a/site/app/templates/notebook/Notebook.twig
+++ b/site/app/templates/notebook/Notebook.twig
@@ -111,12 +111,13 @@
                     {% endif %}
 
                     if ({{ cell.rows }} > 0) {
-                        var editor_{{ num_codeboxes }} = getLargeCodeMirror(document.getElementById("codebox_{{ num_codeboxes }}"), config_{{ num_codeboxes }});
+                        var show_accessbility_message = {{ is_grader_view ? "false" : "true" }}
+                        var editor_{{ num_codeboxes }} = getLargeCodeMirror(document.getElementById("codebox_{{ num_codeboxes }}"), config_{{ num_codeboxes }}, show_accessbility_message);
                         editor_{{ num_codeboxes }}.setSize(null, rowsToPixels({{ cell.rows }}));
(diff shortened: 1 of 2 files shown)


williamjallen (commented):
Is it possible to detect whether the codemirror box can be navigated to on the client side alone?  It's true that the read-only boxes are currently correlated with the grading mode, but this is not necessarily guaranteed to be the case all the time.

ZheyuDeng (changes requested):
Code looks good not seeing the changes for some reasons. Have tried running a full submitty_install and clearing cache. 
image

sahilsuman933:
> Code looks good not seeing the changes for some reasons. Have tried running a full submitty_install and clearing cache. image

That's strange because I can see the change on my local environment. Can you please check on your end as well? @williamjallen﻿

sahilsuman933:
Yes, we can do it on the client-side itself, but it will be a tedious task whereas this solution is more feasible. We just need to pass a new parameter `show_accessibility_msg`, which, by default, is true. So, in case we don't want it, we can just set the value to false or leave it as it is.
﻿
﻿
// @williamjallen

ZheyuDeng (approved):
Tested and works!
image

williamjallen (approved):
The code looks reasonable to me, and I manually verified that the functionality works as expected.

Issue #9685, opened 2023-08-18
Rainbowgrade display options are not valid


ziesski:
Few options in customization page is no longer valid (i.e. display_benchmark) or missing (moss_detail for plagiarism)
<img width="315" alt="Screen Shot 2023-08-18 at 3 01 46 PM" src="https://github.com/Submitty/Submitty/assets/123261952/5e723a06-3a86-4fb0-a14b-8d4e8eaad9f8">

Need to fix both ways.

zacharymnp:
PR #10644 removed benchmark_percent option

zacharymnp:
PR #10653 removes manual_grade option

zacharymnp:
As new items are added to the page, keep the following order in mind:
1. display
2. messages
3. gradeables
4. benchmarks checkboxes
5. display %
6. section
7. final cutoffs
8. manual
9. notes
10. warning
11. plagiarism

PR #10653, opened 2024-07-01
[Feature:RainbowGrades] Add manual_grade to display

zacharymnp:

Login to Instructor, then navigate to Grade Reports > Web-Based Rainbow Grades Customization.

image
Display has a checkbox for manual_grade that has no functionality.

### What is the new behavior?
image
image
Can add manual grades for students using the Rainbow Grades customization GUI. The visibility of this customization item is linked to the final_grade checkbox under Display.

Example Gradebook with note "this is a note". This example is independent of the PR.
image

### Other information?

Related to #9685

Documentation for JSON configuration of some customization items, including manual_grade, is missing or incorrect. PR #[615](https://github.com/Submitty/submitty.github.io/pull/615) (Finish Customization Basics) will update the documentation.


site/app/templates/admin/RainbowCustomization.twig
```diff
@@ -155,6 +155,65 @@
             </div>
 
 
+            <div id="manual-grading" class="customization_item">
+                <h2>Manual Grading</h2>
+                <p class="rg_info_message">You may manually assign final grades to specific students. The note is only invisible to the instructor.</p>
```
micpap25, line 160: Typo, did you mean 'only visible'?

site/app/templates/admin/RainbowCustomization.twig
```diff
@@ -155,6 +155,65 @@
...
+                                <th>Delete</th>
+                            </tr>
+                        </thead>
+                        <tbody id="manual-grading-table-body" style="text-align:left">
+                            {% for entry in manual_grade %}
+                                <tr>
+                                    <td>{{ entry.user }}</td>
+                                    <td>{{ entry.grade }}</td>
+                                    <td>{{ entry.note }}</td>
+                                    <td><a onclick="deleteRow(this)"><i class="fas fa-trash"></i></a></td>
```
micpap25, line 208: deleting a row does not seem to prompt the user to save changes (it stills says "no changes detected")

site/app/templates/admin/RainbowCustomization.twig
```diff
@@ -155,6 +155,65 @@
...
+                            <option value="C+">C+</option>
+                            <option value="C">C</option>
+                            <option value="C-">C-</option>
+                            <option value="D+">D+</option>
+                            <option value="D">D</option>
+                            <option value="F">F</option>
+                        </select>
+                    </label>
+                    <label for="manual_grading_note">
+                        <span>Note:</span>
```
micpap25, line 186: note shouldn't appear if "instructor_notes" isn't selected.

micpap25 (changes requested):
pretty good, just a few nitpicks.

oliiso (changes requested):
Two notes:
- It could be nice to make the "Note" section optional instead of required for cases when the instructor does not want to leave a note.
- I was able to make multiple cases of different final manual grades for the same student. Maybe add a check for there to only be one manual grade instance allowed to each student, and the instructor has to delete the previous manual grade assigned before changing to another. 
Image 7-3-24 at 3 40 PM

site/app/templates/admin/RainbowCustomization.twig
zacharymnp: Good catch. Fixed in [964dbf8](https://github.com/Submitty/Submitty/pull/10653/commits/964dbf8e77ae3a5f370f09f9db93b38f726087eb)

site/app/templates/admin/RainbowCustomization.twig
zacharymnp: Fixed in [0d4e524](https://github.com/Submitty/Submitty/pull/10653/commits/0d4e52443ded2939cf2dcbf4021d557ef594b416)

zacharymnp:
> Two notes:
> 
> * It could be nice to make the "Note" section optional instead of required for cases when the instructor does not want to leave a note.
> * I was able to make multiple cases of different final manual grades for the same student. Maybe add a check for there to only be one manual grade instance allowed to each student, and the instructor has to delete the previous manual grade assigned before changing to another.
>   Image 7-3-24 at 3 40 PM

Fixed in [0e069ca](https://github.com/Submitty/Submitty/pull/10653/commits/0e069ca7734f291be76156ce89614a0a5cbb8d85) and [f30ef97](https://github.com/Submitty/Submitty/pull/10653/commits/f30ef97c994deaeb98d6d1f026b087914717bca1)

oliiso:
Image 7-5-24 at 3 48 PM
Could be nice to align the length of the final grade “select an option” box with the length of the above gradeable “select an option” box in the “penalties for academic dishonesty and plagiarism” to have consistency

zacharymnp:
I'm more inclined to do the opposite, with 'Note' extended a bit so that a longer message can be typed.
image
Do you have any thoughts on this?

oliiso:
> I'm more inclined to do the opposite, with 'Note' extended a bit so that a longer message can be typed. image Do you have any thoughts on this?

Agreed, that's a good idea. Supports the functionality of writing the note!

oliiso (approved):
Requested changes look good and functional.

ziesski:
I wonder if this note can also be displayed with the hover.

zacharymnp:
This is currently what the note looks like. The note column only shows up when instructor_notes is enabled.

image

site/app/templates/admin/RainbowCustomization.twig
zacharymnp: As discussed in meeting, instead of toggling visibility of the note, there will be a message letting the professor know that the note will only appear if instructor_notes is selected. This may be changed in the future. [d1fa086](https://github.com/Submitty/Submitty/pull/10653/commits/d1fa08633957b59ed11c5710a0d048ffe896eabe)

site/public/js/rainbow-customization.js
```diff
@@ -351,14 +351,105 @@ function addToTable() {
     cellDelete.appendChild(deleteLink);
 
     // Clear the form fields
-    document.getElementById('user_id').value = '';
+    document.getElementById('plagiarism_user_id').value = '';
     document.getElementById('g_id').value = '';
     document.getElementById('marks').value = '';
 }
 
+function getManualGrades() {
```
ziesski, line 359: This is almost duplication of plagiarism table code, instead of making a separate function, I think it is good practice to make it modular. Reduce redundancy, and  improve maintainability and integrity.

ziesski (changes requested):

bmcutler (changes requested):
@zacharymnp 

I don't think we need a checkbox for manual_grade at all.|
It should not appear in the "display" list either.
When the instructor enables final_grade, then manual grade is implied and the box to enter manual_grades for specific students should appear -- the instructor should not enable/disable this feature.  They are tied together.

site/public/js/rainbow-customization.js
zacharymnp: [2790559](https://github.com/Submitty/Submitty/pull/10653/commits/2790559ef68d82acfd5af1d6dd32d62c9bdc2b48)

site/app/templates/admin/RainbowCustomization.twig
```diff
@@ -155,6 +155,65 @@
...
+                            <option value="B">B</option>
+                            <option value="B-">B-</option>
+                            <option value="C+">C+</option>
+                            <option value="C">C</option>
+                            <option value="C-">C-</option>
+                            <option value="D+">D+</option>
+                            <option value="D">D</option>
+                            <option value="F">F</option>
+                        </select>
+                    </label>
```
ziesski, line 184:
```suggestion
<label for="manual_grading_grade">
    <span>Final Grade:</span>
    <select name="manual_grading_grade" id="manual_grading_grade">
        <option disabled selected value> -- select an option -- </option>
        {% set grades = ["A", "A-", "B+", "B", "B-", "C+", "C", "C-", "D+", "D", "F"] %}
        {% for grade in grades %}
            <option value="{{ grade }}">{{ grade }}</option>
        {% endfor %}
    </select>
</label>

```

site/app/templates/admin/RainbowCustomization.twig
```diff
@@ -307,7 +366,10 @@ To remove a curve simply blank all boxes."></i>
             $(`#override-${gradeable_id}`).html(override.val());
         });
 
-        $("#user_id").autocomplete({
+        $("#plagiarism_user_id").autocomplete({
+            source: {{ student_full|raw }}
+        });
+        $("#manual_grading_user_id").autocomplete({
```
ziesski, line 372: ["#plagiarism_user_id", "#manual_grading_user_id"].forEach(function(id) {
    $(id).autocomplete({
    });
});

make it concise

ziesski (changes requested):

site/app/templates/admin/RainbowCustomization.twig
zacharymnp: [f493878](https://github.com/Submitty/Submitty/pull/10653/commits/f4938789c799c3f17fa6f4ef4d1e514acd043624)

ziesski (approved):
스크린샷 2024-07-12 오후 4 13 30

If wrong value is injected from terminal, it reads it without causing any issue, so trivial issue.

site/app/models/RainbowCustomization.php
```diff
@@ -537,6 +537,14 @@ public function getPlagiarism(): array {
         return !is_null($this->RCJSON) ? $this->RCJSON->getPlagiarism() : [];
     }
 
+    /**
+     * Get manual grades from json file if there are any
+     *
+     * @return array<object>  array of manual grades JSON object
+     */
+    public function getManualGrades(): array {
+        return !is_null($this->RCJSON) ? $this->RCJSON->getManualGrades() : [];
```
williamjallen, line 546: This is a great opportunity to us PHP's null coalescing operator.  I believe the null coalescing operator eliminates the need to have a nullsafe operator.

```suggestion
        return $this->RCJSON->getManualGrades() ?? [];
```

site/app/models/RainbowCustomizationJSON.php
```diff
@@ -346,6 +361,31 @@ public function addPlagiarismEntry(object $plagiarismEntry): void {
...
+     * } $manualGradeEntry
+     */
+    public function addManualGradeEntry(object $manualGradeEntry): void {
+        $emptyArray = [
+            "user" => "",
+            "grade" => "",
+            "note" => ""
+        ];
+
+        $manualGradeArray = json_decode(json_encode($manualGradeEntry), true);
```
williamjallen, line 380: What's the goal here?  If you want to operate on an array, would it be better to accept an array instead?

site/app/templates/admin/RainbowCustomization.twig
```diff
@@ -155,6 +155,59 @@
...
+    <select name="manual_grading_grade" id="manual_grading_grade">
+        <option disabled selected value> -- select an option -- </option>
+        {% set grades = ["A", "A-", "B+", "B", "B-", "C+", "C", "C-", "D+", "D", "F"] %}
+        {% for grade in grades %}
+            <option value="{{ grade }}">{{ grade }}</option>
+        {% endfor %}
+    </select>
+</label>
+
+                    <label for="manual_grading_note">
```
williamjallen, line 179: The indentation is a bit off here.

site/app/templates/admin/RainbowCustomization.twig
```diff
@@ -155,6 +155,59 @@
...
+                    <label for="manual_grading_user_id">
+                        <span>USER ID:</span>
+                        <input class="option-input" type="text" name="manual_grading_user_id" id="manual_grading_user_id">
+                    </label>
+<label for="manual_grading_grade">
+    <span>Final Grade:</span>
+    <select name="manual_grading_grade" id="manual_grading_grade">
+        <option disabled selected value> -- select an option -- </option>
+        {% set grades = ["A", "A-", "B+", "B", "B-", "C+", "C", "C-", "D+", "D", "F"] %}
+        {% for grade in grades %}
```
williamjallen, line 173: Is the temporary variable needed?

site/app/templates/admin/RainbowCustomization.twig
```diff
@@ -155,6 +155,59 @@
...
+                    <table id="manual-grading-table" class="table table-striped table-bordered persist-area mobile-table">
+                        <thead>
+                            <tr style="text-align:left">
+                                <th>USER ID</th>
+                                <th>Final Grade</th>
+                                <th>Note</th>
+                                <th>Delete</th>
+                            </tr>
+                        </thead>
+                        <tbody id="manual-grading-table-body" style="text-align:left">
```
williamjallen, line 196: I believe the `text-left` class can replace the inline CSS here.  I didn't test it, but I see this Bootstrap 3 class being used elsewhere in the codebase.

site/public/css/rainbow-customization.css
```diff
@@ -223,6 +223,10 @@ hr {
     padding: 4px 6px;
 }
 
+.note-input {
+    width: 32.5em !important;
```
williamjallen, line 227: Is there a way to do this without `!important`?  Using `!important` can be challenging from a future maintenance perspective since it is difficult to override and should be avoided if at all possible.

site/public/js/rainbow-customization.js
```diff
@@ -267,78 +267,144 @@ function getGradeableBuckets() {
     return gradeables;
 }
 
-function getPlagiarism() {
-    const plagiarismData = [];
+/**
+ * Gets data to add to the JSON for the plagiarism and manual grade tables
+ * @param {int} table
+ *      0 -> plagiarism table
+ *      1 -> manual grading table
```
williamjallen, line 274: Is there a reason why int values are useful here?  It might be more readable to instead pass strings which have at least a little bit of descriptive value.

site/public/js/rainbow-customization.js
```diff
@@ -267,78 +267,144 @@ function getGradeableBuckets() {
...
+                penalty: parseFloat(thirdInput),
+            });
+        }
+        else if (table === 1) {
+            data.push({
+                user: firstInput,
+                grade: secondInput,
+                note: thirdInput,
+            });
+        }
```
williamjallen, line 309: These can be combined, with a ternary used to determine the value of the `note` field.

site/public/js/rainbow-customization.js
```diff
@@ -267,78 +267,144 @@ function getGradeableBuckets() {
...
+        case 1: { // manual grading table
+            if (firstInput === '' || secondInput === '') {
+                alert('Please fill in both user ID and final grade.');
+                return;
+            }
+            if (!studentFullDataValues.includes(firstInput)) {
+                alert('Invalid User ID. Please enter a valid one.');
+                return;
+            }
+            const allowed_grades = ['A', 'A-', 'B+', 'B', 'B-', 'C+', 'C', 'C-', 'D+', 'D', 'F'];
```
williamjallen, line 377: This set of grades is being reused in multiple places.  Is there an easy way to share code?

site/public/js/rainbow-customization.js
```diff
@@ -267,78 +267,144 @@ function getGradeableBuckets() {
...
 
-    // Check for duplicate entries
-    const rows = tableBody.getElementsByTagName('tr');
-    for (let i = 0; i < rows.length; i++) {
-        const row = rows[i];
-        const existingUSERID = row.cells[0].textContent.trim();
-        const existingGradeable = row.cells[1].textContent.trim();
+            // Check for duplicate entries
+            for (let i = 0; i < rows.length; i++) {
+                const row = rows[i];
```
williamjallen, line 385: A [`.foreach()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) could be helpful here.

site/public/js/rainbow-customization.js
```diff
@@ -596,6 +664,16 @@ function setCustomizationItemVisibility(elem) {
...
     }
+
+    // manual grading is dependent on final grade cutoffs
+    if (checkbox_name === 'final_grade') {
+        if (is_checked) {
+            $('#manual-grading').show();
+        }
+        else {
+            $('#manual-grading').hide();
+        }
```
williamjallen, line 675: This could be simplified with jQuery's [`.toggle(bool)`](https://api.jquery.com/toggle/#toggle-display) method.

williamjallen (changes requested):

site/public/js/rainbow-customization.js
ziesski: I was thinking about this too, but we will be adding more tables to it later on, so it should be ok.

site/public/js/rainbow-customization.js
zacharymnp: It needs the `.show()`/`.hide()` to load the page properly, since checkboxes are selected on page load based on the JSON.

site/app/templates/admin/RainbowCustomization.twig
zacharymnp: Unfortunately, `class="text-left"` doesn't work since it is overridden by alignment in `class="table"`.

site/public/js/rainbow-customization.js
zacharymnp: There's no way to return from a Javascript `.forEach()`, which makes things a bit needlessly complicated.

site/app/templates/admin/RainbowCustomization.twig
zacharymnp: [dbd3009](https://github.com/Submitty/Submitty/pull/10653/commits/dbd30092ef46a3ebf32207881d9ee672ab0c5f72)

site/app/models/RainbowCustomization.php
zacharymnp: [f0bbef1](https://github.com/Submitty/Submitty/pull/10653/commits/f0bbef17b04826d8fc110beea691e5546e418057)

site/app/templates/admin/RainbowCustomization.twig
zacharymnp: [a6f1920](https://github.com/Submitty/Submitty/pull/10653/commits/a6f19203f20d9f9cab94c59d71a4efabf705e9dd)

site/public/js/rainbow-customization.js
zacharymnp: [a6f1920](https://github.com/Submitty/Submitty/pull/10653/commits/a6f19203f20d9f9cab94c59d71a4efabf705e9dd)

site/app/templates/admin/RainbowCustomization.twig
zacharymnp: Used an id selector to override the class selector. [b912515](https://github.com/Submitty/Submitty/pull/10653/commits/b912515195b09f448b35cd35f1734f9958150dec)

site/app/models/RainbowCustomizationJSON.php
zacharymnp: This is a somewhat convoluted way of casting to an array that I reused from the plagiarism table. Looking into it, it may work better than a cast for nested arrays. Changed to a cast in [f943457](https://github.com/Submitty/Submitty/pull/10653/commits/f9434572aabbd0b6632281e2ae47f55213025f7b)

site/public/js/rainbow-customization.js
zacharymnp: Refactored both functions to use string parameters. [966545c](https://github.com/Submitty/Submitty/pull/10653/commits/966545cb738aaa5febb0cb4923c1dc5ee6ba8bda)

site/app/models/RainbowCustomization.php
zacharymnp: Need to use null-safe operator [59a1948](https://github.com/Submitty/Submitty/pull/10653/commits/59a19487b713b67a0fdbbfda2bcc8a04a6151bd2)

site/public/css/rainbow-customization.css
williamjallen: An ID should be unique on the page, while a class can appear many times.  In reality, most browsers can handle non-unique IDs, but their behavior in such cases is not defined.  The most common case where this comes up is when testing with Cypress.

Is there a reason why an ID works but a class doesn't?

site/public/js/rainbow-customization.js
williamjallen: You can pass a boolean to `.toggle()` to indicate the status it should toggle to.

site/app/models/RainbowCustomizationJSON.php
williamjallen: Can't you just make the function accept an array in the first place?

site/public/css/rainbow-customization.css
zacharymnp: More specific types of selectors override in CSS, and id counts as more specific than class.

site/public/js/rainbow-customization.js
zacharymnp: Changed in [ced8003](https://github.com/Submitty/Submitty/pull/10653/commits/ced80036145a5effcf9b52334bcf5d16f95aa506)

site/app/models/RainbowCustomizationJSON.php
zacharymnp: The value the function is using needs to be an object before and after the function, so there would need to be a cast at some point (unless there's a better way to check whether every value in an object is empty?).

williamjallen (approved):
Code looks reasonable to me.  In the future, it would be better to use jQuery selectors whenever possible, but the surrounding code uses JS to select HTML elements already so I don't see a reason to block this for something so minor.

PR #10671, opened 2024-07-03
[Feature:TAGrading] Allow user to remove hotkey

### LINKED ISSUE #10621: Toggle/Remove TA Hotkeys

TA Grading hotkeys can be rebound, but they cannot be turned off entirely. Some TAs have expressed that they would prefer to remove hotkeys from their interface.

When a hotkey is rebound, there should also be an option to unbind it from any key.

ZheyuDeng:
Fix #10621 
Users were only able to remove hotkey back to default key instead of completly remove it. It also gets pretty unusable.
image

### What is the new behavior?
User are now able to remove shrotcut. Also added buttons to restore all and remove all hotkeys.

https://github.com/Submitty/Submitty/assets/122962727/fba0846f-0726-4078-8801-3c7ae6fc3236

### Other information?

When testing you might need to clear browser cache.
Since some keys are locked from changing should user be able to remove them? If so how should they restore.


sahilsuman933 (changes requested):
It works properly, but it would be nice to have a button through which we can get the default settings.﻿

https://github.com/Submitty/Submitty/assets/34382211/a934f525-6dc4-49a6-ab3a-34bb95ad7222

sahilsuman933:
@ZheyuDeng js-lint is failing can you please fix it?

manzil-infinity180:
adding cypress test for **restore all** and **remove all** button could be better to test

ZheyuDeng:
I am currently writing cypress test for it a review for functionality/UI will be great!

zacharymnp (changes requested):
Functionality is good. Two small comments:

1. Every button was unset when I first opened the page. I had to click Restore All. The keybinds should probably be enabled by default.
2. I would recommend 'Unassigned' or 'Unbound' instead of 'Unset'.

zacharymnp (approved):

site/cypress/e2e/Cypress-Gradeable/hotkey.spec.js
```diff
@@ -47,4 +48,21 @@ describe('TA grading hotkey testing', () => {
...
         cy.get('[data-testid="posts-list"]').should('contain', 'Discussion Posts');
     });
+
+    it('allows user to remove hotkey', () => {
+        cy.login();
+        cy.visit(['sample', 'gradeable', 'grading_homework', 'grading', 'details']);
+        cy.get('[data-testid="view-sections"]').click();
+        cy.get('[data-testid="grade-button"]').eq(12).click();
+        cy.get('[data-testid="grading-setting-btn"]').click();
+        cy.get('#remap-unset-0').click();
```
IDzyre, line 58: Can you create testids for the remap elements here?

IDzyre (changes requested):

site/cypress/e2e/Cypress-Gradeable/hotkey.spec.js
```diff
@@ -47,4 +48,21 @@ describe('TA grading hotkey testing', () => {
...
+        cy.get('[data-testid="view-sections"]').click();
+        cy.get('[data-testid="grade-button"]').eq(12).click();
+        cy.get('[data-testid="grading-setting-btn"]').click();
+        cy.get('[data-testid="remap-unset-0"]').click();
+        cy.get('[data-testid="remap-0"]').should('contain', 'Unassigned');
+        cy.get('[data-testid="remove-all-hotkeys"]').click();
+        cy.get('[data-testid="remap-1"]').should('contain', 'Unassigned');
+        cy.get('[data-testid="remap-2"]').should('contain', 'Unassigned');
+        cy.get('[data-testid="restore-all-hotkeys"]').click();
+        cy.get('[data-testid="remap-0"]').should('not.contain', 'Unassigned');
```
IDzyre, line 64: Is there a value that you should expect to be here, not just `not unassigned`?

ZheyuDeng: Ye your right that makes a lot of sense

IDzyre (approved):
Tests look good, didn't test more than that.

manzil-infinity180 (approved):
Tested locally , looks good to me

sahilsuman933 (approved):
LGTM 🚀

PR #10672, opened 2024-07-03
[Bugfix:Developer] Fix CI Flag and Ignore Flag File

IDzyre:
The CI flag works on CI, but not locally because of a formatting issue. 

### What is the new behavior?
The CI flag is now fixed, and accurately skipping the NTPD setup on CI, but allowing it to run on local machines. 
Also, the flag file was added, so all users who didn't delete the file were running it as if they were in the CI, so this file has been added to the .gitignore file, so if someone is testing the feature, it won't be committed and sent to all users who pull.

### DIFF
--- a/.github_actions_ci_flag
+++ b/.github_actions_ci_flag
@@ -1,1 +1,0 @@
-This file is used to check if Submitty is being run in the Github Actions CI.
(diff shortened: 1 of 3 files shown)


williamjallen (approved):
Code changes seem reasonable to me.

micpap25 (approved):
Tested and works for me

PR #10677, opened 2024-07-08
[Feature:InstructorUI] Team Grade Override

### LINKED ISSUE #10588: Mark all members of a team as grade overridden
Right now in Submitty, only individual user's grades can be overridden at a time.

Copy the functionality of late day overrides for teams; if one student is marked as having their grade overridden, the user is prompted with the option to mark all students on the team as overridden with that grade.

micpap25:
## Why is this Change Important & Necessary?
fixes #10588 tested with #10697
When an instructor wants to override a team's grade, they have to do so for each individual student.

## What is the New Behavior?

Instructors now have the option to override the grades of all students on a team at once.

This is achieved by copying the functionality of extensions for teams; if one student is marked as having their grade overridden, the user is prompted with the option to mark all students on the team as overridden with that grade.

There is now Vue confirmation popups for confirming team grade overrides and team extensions.

team update modal for senger

## What steps should a reviewer take to reproduce or test the bug or new feature?
tested using Rainbow Grades. I found an assignment where `Joe Student` was on a team, and did a team override using his teammate's id. I then generated Rainbow Grades, and observed that his grade was updated along with his teammate's. There may be more intuitive ways to do thus, such as using the logs. View Rainbow Grades instructions [here](https://submitty.org/instructor/course_settings/rainbow_grades/manual_setup).

## Other information
This is not a breaking change. We can't test this in the same way as non-team gradeables because the grade buttons on the grading page do not display "Overridden".


micpap25:
@bmcutler currently the version in Extensions doesn't have any confirmation for deleting an extension; should that lack of confirmation exist in this too?

bmcutler:
@micpap25 
I think an "are you sure" would be good for both extensions and grade override.

IDzyre:
Ideally you'd test the functionality in this PR, but it is okay to split it into a different PR, as long as if anything is 'broken' in that PR, it gets fixed.

site/app/controllers/admin/GradeOverrideController.php
```diff
@@ -40,8 +40,42 @@ public function getOverriddenGrades($gradeable_id) {
...
+                        $this->core->getQueries()->getUserById($team_member_ids[$i])->getDisplayedFamilyName();
+                }
+                $popup_html = $this->core->getOutput()->renderTwigTemplate(
+                    "admin/users/MoreGradeOverrides.twig",
+                    ['member_list' => $team_members, 'is_delete' => true]
+                );
+                return $this->core->getOutput()->renderJsonSuccess([
+                    'is_team' => true,
+                    'popup' => $popup_html
+                ]);
```
williamjallen, line 72: Sending HTML as a field in a JSON response seems like a poor idea.  A better option would be to send the data via this response, and then deal with rendering HTML on the client side.

site/app/controllers/admin/GradeOverrideController.php
```diff
@@ -54,11 +88,45 @@ public function updateOverriddenGrades($gradeable_id) {
...
 
-        $this->core->getQueries()->updateGradeOverride($_POST['user_id'], $gradeable_id, $_POST['marks'], $_POST['comment']);
-        $this->getOverriddenGrades($gradeable_id);
+        $team = $this->core->getQueries()->getTeamByGradeableAndUser($gradeable_id, $_POST['user_id']);
+        //0 is for single submission, 1 is for team submission
+        $option = isset($_POST['option']) ? $_POST['option'] : -1;
+        if ($team != null && $team->getSize() > 1) {
+            if ($option == 0) {
+                $this->core->getQueries()->updateGradeOverride($_POST['user_id'], $gradeable_id, $_POST['marks'], $_POST['comment']);
+                return $this->getOverriddenGrades($gradeable_id);
```
williamjallen, line 101: This seems to be pretty similar to the block of code above.  Would it be productive to share code between these two code blocks?

site/app/templates/admin/GradeOverride.twig
```diff
@@ -79,3 +79,5 @@
         source: {{ student_full|raw }}
     });
 </script>
+
+{% include('admin/users/MoreGradeOverrides.twig') %}
```
williamjallen, line 83: "MoreGradeOverrides" doesn't really provide any information about what it's for.  Any suggestions for a better name?

site/public/js/server.js
```diff
@@ -1388,6 +1398,27 @@ function deleteOverriddenGrades(user_id, g_id) {
...
+    }
+    else {
+        $('input[name="option"]').val(option);
+        updateGradeOverride();
+        $('input[name="option"]').val(-1);
+    }
+}
+
+function overridePopup(json) {
+    $('.popup-form').css('display', 'none');
```
williamjallen, line 1415:
```suggestion
    $('.popup-form').hide();
```

site/public/js/server.js
```diff
@@ -1388,6 +1398,27 @@ function deleteOverriddenGrades(user_id, g_id) {
...
+    const form = $('#more_override_popup');
+    form[0].outerHTML = json['data']['popup'];
+    $('#more_override_popup').css('display', 'block');
```
williamjallen, line 1418:
```suggestion
    $('#more_override_popup').show();
```

site/public/js/server.js
```diff
@@ -1388,6 +1398,27 @@ function deleteOverriddenGrades(user_id, g_id) {
...
+        $('input[name="option"]').val(option);
+        updateGradeOverride();
+        $('input[name="option"]').val(-1);
+    }
+}
+
+function overridePopup(json) {
+    $('.popup-form').css('display', 'none');
+    const form = $('#more_override_popup');
+    form[0].outerHTML = json['data']['popup'];
```
williamjallen, line 1417: It's very easy for jQuery functions which insert raw HTML to result in XSS vulnerabilities.  Please do not directly insert HTML this way.

williamjallen (changes requested):

site/app/controllers/admin/GradeOverrideController.php
```diff
@@ -40,8 +40,42 @@ public function getOverriddenGrades($gradeable_id) {
...
+                }
+                return $this->getOverriddenGrades($gradeable_id);
+            }
+            else {
+                $team_member_ids = explode(", ", $team->getMemberList());
+                $team_members = [];
+                for ($i = 0; $i < count($team_member_ids); $i++) {
+                    $team_members[$team_member_ids[$i]] = $this->core->getQueries()->getUserById($team_member_ids[$i])->getDisplayedGivenName() . " " .
+                        $this->core->getQueries()->getUserById($team_member_ids[$i])->getDisplayedFamilyName();
+                }
```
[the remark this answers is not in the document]
micpap25: Partially fixed, I halved the number of queries but i could look into constant time. With that said teams generally shouldn't exceed 3-5 members so it shouldn't be a big deal, right?

site/app/controllers/admin/GradeOverrideController.php
micpap25: It definitely would be but I wouldn't know how to approach it in PHP, it would also be nice to share the code with the extensions page if we can.

micpap25:
@williamjallen Resolved various issues and commented on a few others. I'm going to do my best with HTML security now, that might take a little longer so I'll get back to you on that.

site/app/controllers/admin/GradeOverrideController.php
micpap25: Can JS render a twig template? I don't see an example of this in the code.

micpap25:
Assigned relevant people to this task so we can work together to render twig JS-side.

micpap25:
Additionally I will handle e2e tests in this PR since #10697 was merged.

micpap25:
@williamjallen should we use Twig.js or wait for Vue to get merged

site/app/controllers/admin/GradeOverrideController.php
williamjallen: One possible approach is to render a Twig template and then use JS to manipulate the data within it.  You'd render a Twig template with the general structure of the HTML, and then use jQuery to fill it with data.

williamjallen:
@micpap25 Apologies for the delay, I forgot to send my earlier comment after typing it.

We definitely don't want to use Twig.js, and I'm not sure we want to block this until Vue is ready for production use.  Instead, you can render the Twig template server side, and then fill it with data, the same way the forum, grade inquiries, TA grading, and other similar dynamic content is displayed.

micpap25:
@williamjallen could we discuss further on call or in person? I'm not sure I fully understand your point even after looking at the forum and grade inquiry code.

williamjallen:
After discussing in today's meeting, it looks like waiting for Vue (https://github.com/Submitty/Submitty/pull/10715) is probably the most effective option here.

site/app/controllers/admin/GradeOverrideController.php
```diff
@@ -46,19 +46,66 @@ public function deleteOverriddenGrades($gradeable_id) {
...
-        if (!isset($_POST['user_id']) || $_POST['user_id'] == "" || $isUserNotInCourse || $user->getId() !== $_POST['user_id']) {
-            $error = "Invalid Student ID";
-            return $this->core->getOutput()->renderJsonFail($error);
+        $user_id = $_POST['user_id']   ?? '';
+        $marks   = $_POST['marks']     ?? '';
+        $comment = $_POST['comment']   ?? '';
+        $option  = intval($_POST['option'] ?? -1);
+
+        $user = $this->core->getQueries()->getSubmittyUser($user_id);
+        if ($user === null || $user->getId() !== $user_id) {
```
jeffrey-cordero: I don't believe `$user->getId() !== $user_id` will ever be false at that point given the literal query is `SELECT * FROM users WHERE user_id=?"`

site/app/controllers/admin/GradeOverrideController.php
```diff
@@ -46,19 +46,66 @@ public function deleteOverriddenGrades($gradeable_id) {
...
-            $error = "Marks be a integer";
-            return $this->core->getOutput()->renderJsonFail($error);
+        if ($team !== null && $team->getSize() > 1) {
+            if ($option === 0) {
+                $this->core->getQueries()
+                    ->updateGradeOverride($user_id, $gradeable_id, $marks, $comment);
+                return $this->getOverriddenGrades($gradeable_id);
+            }
+            elseif ($option === 1) {
+                foreach ($this->getTeamMemberIds($team) as $member_id) {
```
jeffrey-cordero: Any chance we can optimize this into a single SQL query? I believe this was causing the performance warning, where if you convert the following "upsert" query into a batch update or insert query, it would probably address the warning.

```php
public function updateGradeOverride($user_id, $g_id, $marks, $comment) {
        $this->course_db->query(
            "
          UPDATE grade_override
          SET marks=?, comment=?
          WHERE user_id=?
            AND g_id=?;",
            [$marks, $comment, $user_id, $g_id]
        );
        if ($this->course_db->getRowCount() === 0) {
            $this->course_db->query(
                "
            INSERT INTO grade_override
            (user_id, g_id, marks, comment)
            VALUES(?,?,?,?)",
                [$user_id, $g_id, $marks, $comment]
            );
        }
    }
```

site/app/controllers/admin/GradeOverrideController.php
```diff
@@ -46,19 +46,66 @@ public function deleteOverriddenGrades($gradeable_id) {
 
     #[Route("/courses/{_semester}/{_course}/grade_override/{gradeable_id}/update", methods: ["POST"])]
     public function updateOverriddenGrades($gradeable_id) {
-        $user = $this->core->getQueries()->getSubmittyUser($_POST['user_id']);
-        $isUserNotInCourse = empty($this->core->getQueries()->getUsersById([$_POST['user_id']]));
-        if (!isset($_POST['user_id']) || $_POST['user_id'] == "" || $isUserNotInCourse || $user->getId() !== $_POST['user_id']) {
-            $error = "Invalid Student ID";
-            return $this->core->getOutput()->renderJsonFail($error);
+        $user_id = $_POST['user_id']   ?? '';
+        $marks   = $_POST['marks']     ?? '';
```
jeffrey-cordero: Although the spacing here looks pretty clean, it's best to follow the same spacing format in the method below to ensure consistency.

site/app/controllers/admin/GradeOverrideController.php
```diff
@@ -46,19 +46,66 @@ public function deleteOverriddenGrades($gradeable_id) {
...
+        $comment = $_POST['comment']   ?? '';
+        $option  = intval($_POST['option'] ?? -1);
+
+        $user = $this->core->getQueries()->getSubmittyUser($user_id);
+        if ($user === null || $user->getId() !== $user_id) {
+            return $this->core->getOutput()->renderJsonFail("Invalid Student ID");
+        }
+
+        if ($marks === '' || !ctype_digit($marks)) {
+            return $this->core->getOutput()->renderJsonFail("Marks must be an integer");
```
jeffrey-cordero: I'd suggest updating the error message to `Marks must be at least 0`.

site/public/js/server.js
```diff
@@ -1345,6 +1355,36 @@ function deleteOverriddenGrades(user_id, g_id) {
...
+        updateGradeOverride();
+        $('input[name="option"]').val(-1);
+    }
+}
+
+function overridePopup(json) {
+    $('.popup-form').hide();
+    $('#override_team_popup').remove();
+
+    const mount_id = `vue-${Math.floor(Math.random() * 1e9)}`;
```
jeffrey-cordero, line 1386: I'd suggest adding a simple comment here explaining that the unique mount ID is required to mount to a dynamic Vue app onto the DOM until we have a uniform Vue framework handling this.

site/app/controllers/admin/GradeOverrideController.php
```diff
@@ -46,19 +46,66 @@ public function deleteOverriddenGrades($gradeable_id) {
...
         }
+        $marks = (int) $marks;
+
+        $team = $this->core->getQueries()->getTeamByGradeableAndUser($gradeable_id, $user_id);
 
-        if (((!isset($_POST['marks'])) || $_POST['marks'] == "" || is_float($_POST['marks']))) {
-            $error = "Marks be a integer";
-            return $this->core->getOutput()->renderJsonFail($error);
+        if ($team !== null && $team->getSize() > 1) {
+            if ($option === 0) {
```
jeffrey-cordero: Any chance we can update `0` and `1` to be literal strings, such as `single` and `batch`? It'd make this controller method and the front-end code more readable and maintainable.

site/cypress/e2e/Cypress-Feature/grade_override.spec.js
```diff
@@ -12,10 +12,13 @@ describe('testing grade override', () => {
             cy.get('[data-testid="grade-override-score"]').click();
             cy.get('[data-testid="grade-override-score"]').type('10');
             cy.get('[data-testid="grade-override-submit"]').click();
+            if (!(gradeable === 'grading_homework')) {
```
jeffrey-cordero: Could you update this to be `!==` for conciseness and readability?

site/cypress/e2e/Cypress-Feature/grade_override.spec.js
```diff
@@ -12,10 +12,13 @@ describe('testing grade override', () => {
...
+                cy.get('[data-testid="confirm-team-override"]').click();
+            }
             cy.get('[data-testid="popup-message"]').should('contain', `Updated overridden Grades for ${gradeable}`);
             cy.get('[data-testid="load-overridden-grades"]').should('contain', 'student');
             cy.get('[data-testid="load-overridden-grades"]').should('contain', `Overridden Grades for ${gradeable}`);
-            // When the same functionaliy for teams added,then remove the if block with adding some additional test
+            // Verify that override was successful. TODO: Logic for verifying team override after PR to enhance feedback
```
jeffrey-cordero: We could verify that `Adela Wisozk (wisoza)` displays their overridden mark within `#grade-override-table`. You could add a `data-testid` for each row to identify the overridden grade for a student, and another to identify the mark component. Verifying both marks (student, wisoza) would add robustness to this feature.

site/public/js/server.js
```diff
@@ -1317,23 +1321,29 @@ function refreshOnResponseOverriddenGrades(json) {
     }
 }
 
-function deleteOverriddenGrades(user_id, g_id) {
-    const url = buildCourseUrl(['grade_override', g_id, 'delete']);
-    const confirm = window.confirm('Are you sure you would like to delete this entry?');
+function deleteOverriddenGrades(user_id, option) {
+    const url = buildCourseUrl(['grade_override', $('#g_id').val(), 'delete']);
```
jeffrey-cordero: We should keep the gradeable ID in the function parameters list because we already invoke the `user_id` in a similar way - `deleteOverriddenGrades($('#user_id').val(), option);`

site/app/controllers/admin/GradeOverrideController.php
```diff
@@ -46,19 +46,66 @@ public function deleteOverriddenGrades($gradeable_id) {
...
+            }
+            elseif ($option === 1) {
+                foreach ($this->getTeamMemberIds($team) as $member_id) {
+                    $this->core->getQueries()
+                        ->updateGradeOverride($member_id, $gradeable_id, $marks, $comment);
+                }
+                return $this->getOverriddenGrades($gradeable_id);
+            }
+            else {
+                $member_ids  = $this->getTeamMemberIds($team);
```
jeffrey-cordero, line 75: The spacing in this block is also a bit excessive.

jeffrey-cordero (changes requested):
Team grade override functionality looks solid. Grade overrides for non-team assignments and single vs. batch team grade overrides worked as expected. I did leave a few comments to optimize the performance and maintainability of this feature for the future.

site/app/controllers/admin/GradeOverrideController.php
jeffrey-cordero: After thinking about it, a single batch query may be difficult given that the row count may not match the update query result for missing rows. I'd suggest doing an insert query and appending `ON CONFLICT (user_id, g_id) DO UPDATE SET marks = EXCLUDED.marks, comment = EXCLUDED.comment;` at the end, or marking this as a future TODO comment for the future.

site/cypress/e2e/Cypress-Feature/grade_override.spec.js
JManion32: I made the Cypress test verify the headers of the table, then verify the top column to ensure Joe Student is entered correctly. I'm not sure if verifying all of the headers is necessary, but I figured it was better than the more general tests we had.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -4531,21 +4531,13 @@ public function updateExtensions($user_id, $g_id, $days, $reason) {
     public function updateGradeOverride($user_id, $g_id, $marks, $comment) {
         $this->course_db->query(
             "
-          UPDATE grade_override
-          SET marks=?, comment=?
-          WHERE user_id=?
-            AND g_id=?;",
-            [$marks, $comment, $user_id, $g_id]
+            INSERT INTO grade_override (user_id, g_id, marks, comment)
```
jeffrey-cordero, line 4534: I would suggest turning this into a transaction, potentially a new method, to ensure they all have the final mark, regardless of interleaving failures. This would also improve performance.

site/public/js/server.js
```diff
@@ -1345,6 +1355,39 @@ function deleteOverriddenGrades(user_id, g_id) {
...
+        updateGradeOverride();
+        $('input[name="option"]').val('');
+    }
+}
+
+function overridePopup(json) {
+    $('.popup-form').hide();
+    $('#override_team_popup').remove();
+
+    // Generate a unique mount ID for the dynamic Vue app.
```
jeffrey-cordero: I'd suggest merging the first and third comments into a single line for conciseness.

site/app/libraries/database/DatabaseQueries.php
jeffrey-cordero: This actually won't fix the performance warning, so I'd recommend a batch insert and update on conflict.

site/cypress/e2e/Cypress-Feature/grade_override.spec.js
```diff
@@ -12,10 +12,27 @@ describe('testing grade override', () => {
...
+            cy.get('@headers').eq(2).should('have.text', 'Family Name');
+            cy.get('@headers').eq(3).should('have.text', 'Marks');
+            cy.get('@headers').eq(4).should('have.text', 'Comments');
+            cy.get('@headers').eq(5).should('have.text', 'Delete');
+
+            cy.get('#grade-override-table tbody tr').first().within(() => {
+                cy.get('td').eq(0).should('contain', 'student');
+                cy.get('td').eq(1).should('contain', 'Joe');
+                cy.get('td').eq(2).should('contain', 'Student');
+                cy.get('td').eq(3).should('contain', '10');
```
jeffrey-cordero: Any chance we could also test a specific team gradeable to ensure another team member receives the same mark? It will help cover testing for this new feature.

site/app/libraries/database/DatabaseQueries.php
```diff
@@ -4540,6 +4540,31 @@ public function updateGradeOverride($user_id, $g_id, $marks, $comment) {
         );
     }
 
+    /**
+     * @param string[] $user_ids
```
jeffrey-cordero, line 4544: I'd suggest adding the other params here as well, potentially with a short description, but not required.

site/app/templates/admin/GradeOverride.twig
```diff
@@ -62,6 +62,7 @@
                         <th>Delete</th>
                     </tr>
                 </thead>
+                <tbody></tbody>
```
jeffrey-cordero: This should be reverted.

site/cypress/e2e/Cypress-Feature/grade_override.spec.js
```diff
@@ -12,10 +12,35 @@ describe('testing grade override', () => {
...
+            cy.get('#grade-override-table thead tr th').as('headers');
+            cy.get('@headers').eq(0).should('have.text', 'Student ID');
+            cy.get('@headers').eq(1).should('have.text', 'Given Name');
+            cy.get('@headers').eq(2).should('have.text', 'Family Name');
+            cy.get('@headers').eq(3).should('have.text', 'Marks');
+            cy.get('@headers').eq(4).should('have.text', 'Comments');
+            cy.get('@headers').eq(5).should('have.text', 'Delete');
+
+            cy.get('#grade-override-table tbody tr').first().within(() => {
+                cy.get('td').eq(0).should('contain', 'student');
```
jeffrey-cordero: I'd suggest moving these 4 Cypress assertions into a helper method to not duplicate code here.

site/cypress/e2e/Cypress-Feature/grade_override.spec.js
```diff
@@ -12,10 +12,35 @@ describe('testing grade override', () => {
...
+                cy.get('td').eq(1).should('contain', 'Joe');
```
jeffrey-cordero: Adding a `data-testid` to the `<td>` element would also be best practices here.

site/app/templates/admin/GradeOverride.twig
JManion32: I had to add that to test the table data in Cypress because there was no tbody for it to look at since the rows were being created in JS. Is there a better way to do it?

jeffrey-cordero: Ah, this makes sense now. I'd suggest adding a `data-testid` to be used in the Cypress test.

site/cypress/e2e/Cypress-Feature/grade_override.spec.js
jeffrey-cordero: This would also be updated to use `data-testid` in place of `tbody` and `tr`.

site/app/templates/admin/GradeOverride.twig
JManion32: Will do 👍

site/public/js/server.js
```diff
@@ -1308,16 +1312,16 @@ function refreshOnResponseOverriddenGrades(json) {
     }
     else {
         json['data']['users'].forEach((elem) => {
-            const delete_button = `<a onclick="deleteOverriddenGrades('${elem['user_id']}', '${json['data']['gradeable_id']}');" data-testid="grade-override-delete"><i class='fas fa-trash'></i></a>`;
+            const delete_button = `<a onclick="deleteOverriddenGrades('${elem['user_id']}', '${json['data']['gradeable_id']}', 'single');" data-testid="grade-override-delete"><i class='fas fa-trash'></i></a>`;
             const bits = [`<tr><td class="align-left">${elem['user_id']}`, elem['user_givenname'], elem['user_familyname'], elem['marks'], elem['comment'], `${delete_button}</td></tr>`];
-            $('#grade-override-table').append(bits.join('</td><td class="align-left">'));
+            $('#grade-override-table tbody').append(bits.join('</td><td class="align-left">'));
```
jeffrey-cordero: `data-testid` would also be used here, similar to `data-testid="grade-override-delete`.

site/cypress/e2e/Cypress-Feature/grade_override.spec.js
JManion32: Wouldn't that require creating a command in `commands.js`? I feel like that isn't necessary here as we can't use the helper method in any other file.

jeffrey-cordero: No, you can make a literal JS function and use `cy` methods.

```js
const verifyEmail = (subject, source) => {
    cy.get('[data-testid="email-subject"]')
        .invoke('text')
        .then((text) => expect(text.trim()).to.equal(
            `Email Subject: ${subject}`,
        ));
    cy.get('[data-testid="email-time-created"]')
        .invoke('text')
        .then((text) => expect(text.trim()).to.match(
            /Time Created: \d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$/,
        ));
    cy.get('[data-testid="email-source"]')
        .invoke('text')
        .then((text) => expect(text.trim()).to.equal(source));
};

```

JManion32: Oh okay I'll do that

jeffrey-cordero (approved):
The team grade override feature looks solid, and the end-to-end tests properly cover the functionality as intended.

PR #10680, opened 2024-07-09
[Bugfix:InstructorUI] Remove lowest null handling

ziesski:
Currently, when web-based GUI customization page is rendered, users may get below screenshot image warning (screenshot credit to @zacharymnp) 
스크린샷 2024-07-09 오전 11 07 16

This is occurring because when we are loading json, it is looking for remove_lowest even when there is none.

In order to reproduce/test this issue, go to reports page -> customization page -> drag few gradeable to assigned bucket. Hit save changes. Refresh page, and will see the screenshot warning. 

With this change, it checks whether remove_lowest is specified or not, and if it is not, it would set it to zero, and only parse when it is specified.

### DIFF
--- a/site/app/models/RainbowCustomization.php
+++ b/site/app/models/RainbowCustomization.php
@@ -122,10 +122,9 @@ class RainbowCustomization extends AbstractModel {
                     $this->bucket_counts[$bucket] = $json_bucket->count;
                 }
 
-                if (!empty($json_bucket->remove_lowest)) {
+                if (property_exists($json_bucket, 'remove_lowest')) {
                     $this->bucket_remove_lowest[$bucket] = $json_bucket->remove_lowest;
-                }
-                else {
+                } else {
                     $this->bucket_remove_lowest[$bucket] = 0;
                 }
             }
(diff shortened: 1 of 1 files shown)


site/app/models/RainbowCustomization.php
```diff
@@ -122,7 +122,12 @@ public function buildCustomization() {
                     $this->bucket_counts[$bucket] = $json_bucket->count;
                 }
 
-                $this->bucket_remove_lowest[$bucket] = $json_bucket->remove_lowest;
+                if (property_exists($json_bucket, 'remove_lowest')) {
+                    $this->bucket_remove_lowest[$bucket] = $json_bucket->remove_lowest;
+                }
+                else {
+                    $this->bucket_remove_lowest[$bucket] = 0;
+                }
```
williamjallen, line 130: PHP's null coalescing operator works on properties too, so this can be simplified as follows:
```suggestion
                $this->bucket_remove_lowest[$bucket] = $json_bucket->remove_lowest ?? 0;
```

williamjallen (changes requested):

PR #10688, opened 2024-07-10
[Bugfix:Autograding] Tolerance parse fix

micpap25:
Makes a fix on #8251 so it works with numbers that have non-period punctuation surrounding them will still be recognized as numbers.


dorian451 (approved):
Looks good to me

ziesski (approved):
Reviewed the code, and looks like it will recognize numbers numbers surrounded by non-period punctation, as described. Since it is cpp code, Barb should take another look before merging.

grading/clean.cpp
```diff
@@ -308,7 +313,23 @@ bool isNumber(const std::string &str) {
...
+  return atLeastOneDigit;
+}
+
+std::string isolateAlphanumAndNumberPunctuation(const std::string &str) {
+  std::string::const_iterator begin = str.begin();
+  while (!isalnum(*begin) && !(*begin == '.')) {
+    begin++;
+    if (begin == str.end())
+    {
+      return str;
```
bmcutler, line 325: so you return the whole string if there is no number at all?
(this is not intuitive)

grading/clean.cpp
```diff
@@ -308,7 +313,23 @@ bool isNumber(const std::string &str) {
       return false;
     }
   }
-  return true == atLeastOneDigit;
+  return atLeastOneDigit;
+}
+
+std::string isolateAlphanumAndNumberPunctuation(const std::string &str) {
```
bmcutler, line 319: this function is somewhat confusing in purpose and should at least have a comment at the start...

bmcutler (changes requested):

bmcutler (approved):
thanks!

PR #10700, opened 2024-07-11
[Feature:InstructorUI] Use cookie in grade override page

### LINKED ISSUE #10676: Have Grade Override use Cookies and remove Apply Changes Button
Right now, the Grade Override page does not remember what the last gradable you were accessing was. Instructors need to select the gradable and click "apply changes" to see anything. In comparison, the Excused Absence Extension page uses cookies to remember the last gradable the instructor was looking at.

The Grade Override page should respond automatically to a new selected gradable and show content. It should store the last gradable the user selected and automatically open that gradable when the page is accessed again.

ZheyuDeng:
Fix #10676 

While instructor grade override page is similar to excused absence extension, users need to hit apply changes after selecting a gradable to see the data where as excused absence extension page directly shows the data and remember it with cookie.

### What is the new behavior?
Now it shows the data right after selecting the gradable and have cookie enabled.

https://github.com/user-attachments/assets/9ef21722-f4b2-4d93-ba6d-4f2494487eea


site/app/templates/admin/GradeOverride.twig
```diff
@@ -80,3 +71,4 @@
         source: {{ student_full|raw }}
     });
 </script>
+<script src="/js/grade-override.js"></script>
```
zacharymnp, line 74: This should be added in `GradeOverrideView.php` using `$this->core->getOutput()->addInternalCss('grade-override.js');`

zacharymnp (changes requested):

zacharymnp (approved):

ziesski (approved):
After discussion, noticed cookie is used to remember last edited gradeable, so that when instructor comeback to the page, and have that gradeable page selected, this matches the excused absent page behavior. Code looks good.

PR #10705, opened 2024-07-12
[Feature:TAGrading] Word document to PDF conversion

### LINKED ISSUE #10454: Add viewer for .docx files in TA grading file browser

skara9:

Submitted word documents need to be downloaded by grader.
Resolves #10454 

### What is the new behavior?
Word documents can now be viewed as PDF's in the browser without having to be downloaded.

### DIFF
--- a/migration/migrator/migrations/system/20240712022346_libreoffice_py
+++ b/migration/migrator/migrations/system/20240712022346_libreoffice_py
@@ -1,0 +1,21 @@
+"""Migration for the Submitty system."""
+import os
+
+def up(config):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    os.system("apt-get install libreoffice-core")
+
+
+def down(config):
+    """
+    Run down migration (rollback).
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    pass
--- a/sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
+++ b/sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
@@ -11,7 +11,9 @@ import subprocess
 import stat
 import traceback
 import datetime
+import mimetypes
 from urllib.parse import unquote
+from tempfile import TemporaryDirectory
 from . import bulk_qr_split
 from . import bulk_upload_split
 from . import INSTALL_DIR, DATA_DIR
@@ -416,3 +418,46 @@ class UpdateSystemInfo(AbstractJob):
 
     def cleanup_job(self):
         pass
+
+
+class DocToPDF(AbstractJob):
+    def run_job(self):
+        log_dir = os.path.join(DATA_DIR, "logs", "doc_to_pdf")
+        today = datetime.datetime.now()
+        log_file = os.path.join(log_dir, "{:04d}{:02d}{:02d}.txt".format(today.year, today.month, today.day))
+
+        flag = os.O_EXCL | os.O_WRONLY
+        if not os.path.exists(log_file):
+            flag |= os.O_CREAT
+        log = os.open(log_file, flag)
+
+        term = self.job_details['term']
+        course = self.job_details['course']
+        gradeable = self.job_details['gradeable']
+        user = self.job_details['user']
+        version = self.job_details['version']
+
+        submission_path = os.path.join(DATA_DIR, 'courses', term, course, 'submissions', gradeable, user, version)
+
+        DOC_MIME_TYPES = ['application/vnd.openxmlformats-officedocument.wordprocessingml.document']
+
+        doc_files = []
+
+        for root, _, files in os.walk(submission_path):
+            for file in files:
+                file_path = os.path.join(root, file)
+                mimetype, _ = mimetypes.guess_type(file_path)
+                if mimetype in DOC_MIME_TYPES:
+                    doc_files.append(file_path)
+
+        for doc_file in doc_files:
+            with TemporaryDirectory() as tmpdir:
+                with os.fdopen(log, 'a') as output_file:
+                    result = subprocess.run(["libreoffice", "--headless", "--convert-to", "pdf", "--outdir", tmpdir, doc_file], stdout=output_file, stderr=output_file)
+                if result.returncode != 0:
+                    continue 
+                file = os.listdir(tmpdir)[0]
+                os.rename(file, os.path.join(os.path.dirname(doc_file), '.' + os.path.basename(doc_file) + '.pdf'))
+
+        log_msg = f"[Last ran on: {today.isoformat()}]\n"
+        logger.write_to_log(log_file, log_msg)
--- a/site/app/controllers/student/SubmissionController.php
+++ b/site/app/controllers/student/SubmissionController.php
@@ -1673,6 +1673,29 @@ class SubmissionController extends AbstractController {
         }
 
 
+        // Add to queue to convert Word documents to PDF files for viewing
+        $doc_queue_data = [
+            "job" => "DocToPDF",
+            "term" => $this->core->getConfig()->getTerm(),
+            "course" => $this->core->getConfig()->getCourse(),
+            "gradeable" => $gradeable->getId(),
+            "user" => $user_id,
+            "version" => $new_version
+        ];
+
+        $doc_queue_file_helper = implode("__", ["doc_to_pdf", $this->core->getConfig()->getTerm(), $this->core->getConfig()->getCourse(),
+            $gradeable->getId(), $who_id, $new_version]);
+        $doc_queue_file = FileUtils::joinPaths(
+            $this->core->getConfig()->getSubmittyPath(),
+            "daemon_job_queue",
+            $doc_queue_file_helper
+        );
+
+        if (@file_put_contents($doc_queue_file, FileUtils::encodeJson($doc_queue_data), LOCK_EX) === false) {
+            return $this->uploadResult("Failed to create file for pdf generation queue.", false);
+        }
+
+
         $queue_file_helper = [$this->core->getConfig()->getTerm(), $this->core->getConfig()->getCourse(),
             $gradeable->getId(), $who_id, $new_version];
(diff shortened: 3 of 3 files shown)


zacharymnp (changes requested):
This does not seem to work, although it's possible there are installation issues.
Unreadable text still appears when viewing a sample .docx
image
results_public is empty

skara9:
@zacharymnp This PR isn't ready for review; the generated PDF files are uploaded to the results_public folder, but the autograder overwrites them shortly after. See the meeting notes for more information.

migration/migrator/migrations/system/20240712022346_libreoffice_py
```diff
@@ -0,0 +1,21 @@
...
+import os
+
+def up(config):
+    """
+    Run up migration.
+
+    :param config: Object holding configuration details about Submitty
+    :type config: migrator.config.Config
+    """
+    os.system("apt-get install libreoffice")
```
cjreed121, line 11: Add a `-y`. Also libreoffice is a pretty large package due to all the dependencies. I'm assuming most of the functionality and dependencies we will never use. There looks to be a core or non-gui version which we can install which might be smaller. Do you remember how big the install of libreoffice was for you.

sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
```diff
@@ -416,3 +418,59 @@ def run_job(self):
...
+                if mimetype in DOC_MIME_TYPES:
+                    doc_files.append(file_path)
+
+        stat_parent = os.stat(processed_submission_path)
+        for doc_file in doc_files:
+            with TemporaryDirectory() as tmpdir:
+                with os.fdopen(log, 'a') as output_file:
+                    output_file.write('\n')
+                    output_file.flush()
+                    result = subprocess.run(["libreoffice", "--headless", "--convert-to", "pdf", "--outdir", tmpdir, doc_file], stdout=output_file, stderr=output_file)
```
cjreed121, line 467: I'm a bit concerned on security on this. libreoffice is a trustworthy program but it also supports a lot of different files types and functionality. The main possible risk would be some sort of arbitrary code execution or file read/write. I'm not sure if an exploit would be possible here and it would take a long time to craft some sort of payload to do something harmful (if it's even possible). However, the safer thing to do would be to run it as a less privileged user rather than running it as submitty_daemon. A container could also be another option but would be more expensive.

cjreed121 (dismissed):

migration/migrator/migrations/system/20240712022346_libreoffice_py
skara9: I initially tried the nogui and core options, but both failed on the convert command due to some missing package(s) so I switched to the full bundle. I think it was ~600mb for me. I'll test again (=once the rest of the issues on this PR are addressed and see if theres a smaller set of packages we can install.

sbin/submitty_daemon_jobs/submitty_jobs/jobs.py
skara9: Agreed. Which user do you think we should delegate this to? Or should we make a new one?

site/app/controllers/student/SubmissionController.php
```diff
@@ -1691,6 +1691,45 @@ public function ajaxUploadSubmission($gradeable_id, $merge = null, $clobber = nu
...
+
+        if ($has_docx) {
+            // Add to queue to convert Word documents to PDF files for viewing
+            $doc_queue_data = [
+            "job" => "DocToPDF",
+            "term" => $this->core->getConfig()->getTerm(),
+            "course" => $this->core->getConfig()->getCourse(),
+            "gradeable" => $gradeable->getId(),
+            "user" => $user_id,
+            "version" => $new_version
```
williamschen23: indent

site/app/templates/submission/homework/CurrentVersionResults.twig
```diff
@@ -31,7 +31,32 @@
...
+                                    {% endif %}
+                                    <button class = 'btn btn-primary' onclick='downloadFile("{{ file.path|url_encode }}", "submissions")' aria-label="Download {{file.relative_name}}"> Download
+                                        <i class="fas fa-download" title="Download the file"></i></button>
+                                </div>
+                            {% endif %}
+                        </div>
+                    {% endif %}
+                {% endfor %}
+            {% endif %}
+            {% if student_download and (files|length + processed_files|length) > 1 %}
```
williamschen23, line 59: Since we only generate pdf's, this code should only be pdfs

site/app/controllers/student/SubmissionController.php
```diff
@@ -1691,6 +1691,45 @@ public function ajaxUploadSubmission($gradeable_id, $merge = null, $clobber = nu
...
+                if (isset($uploaded_files[$i])) {
+                    for ($j = 0; $j < $count[$i]; $j++) {
+                        if (strtolower(pathinfo($uploaded_files[$i]["name"][$j], PATHINFO_EXTENSION)) === "docx") {
+                            $has_docx = true;
+                            break 2;
+                        }
+                    }
+                }
+            }
+        }
```
williamschen23: is this necessary? We loop through all the files in the job.

site/app/libraries/Access.php
```diff
@@ -208,6 +208,7 @@ public function __construct(Core $core) {
         $this->permissions["path.write.split_pdf"] = self::ALLOW_MIN_FULL_ACCESS_GRADER | self::CHECK_CSRF;
         $this->permissions["path.write.uploads"] = self::ALLOW_MIN_INSTRUCTOR | self::CHECK_CSRF;
         $this->permissions["path.write.site"] = self::ALLOW_MIN_INSTRUCTOR | self::CHECK_CSRF;
+        $this->permissions["path.write.submissions_processed"] = self::DENY_ALL | self::CHECK_CSRF;
```
williamschen23: deny all or check csrf? does this make sense

williamschen23 (changes requested):
Output from libreoffice container:
----------------------------------
Warning: failed to launch javaldx - java may not function correctly
convert /app/doc_files/0/ILE Research Template-SU25.docx as a Writer document -> /app/doc_files/0/out/ILE Research Template-SU25.pdf using filter : writer_pdf_Export

[Last ran on: 2025-07-08T01:57:33.614748]
not sure if this impacts anything but i have this in my log file.

image
Frog robot here after a pdf gets converted, this should get fixed

Manually setting anything to .docx gives a `Error: source file could not be loaded`. In addition, docx files that are marked as another extension are not processed (good behavior?)

a bit rambly at 2am but catch me tmr for better explaining

williamschen23 (approved):
Functionality looks good. Code looks reasonable

.setup/install_submitty/setup_directories.sh
```diff
@@ -108,6 +108,7 @@ mkdir -p "${SUBMITTY_DATA_DIR:?}/logs/autograding_stack_traces"
 if [ "${IS_WORKER:?}" == 0 ]; then
     mkdir -p "${SUBMITTY_DATA_DIR:?}/logs/access"
     mkdir -p "${SUBMITTY_DATA_DIR:?}/logs/bulk_uploads"
+    mkdir -p "${SUBMITTY_DATA_DIR:?}/logs/doc_to_pdf"
```
bmcutler: Should this be 'docx' instead of 'doc' everywhere?

bmcutler (changes requested):
the code is inconsistent, using 'doc' in most places, and using 'docx' in others.
I think it should be 'docx' everywhere.

bmcutler:
I tested it with a single file, and it seemed to work fine.

Screenshot 2025-07-22 at 12 41 39 PM

But when I tested it with a zip file with multiple docx at different places in the directory structure, it didn't seem to work.  No files appeared in the processed directory:

Screenshot 2025-07-22 at 12 40 25 PM

I think I waited long enough for them to appear, there are no queue files in the daemon_jobs folder.  It might be nice to have a message on the TA grading page that we are awaiting processing and files will be added to submissions processed.  Similar to how there is a message when this assignment is in the autograding queue.

NOTE:  Also, I think we should probably rename that 'submissions_processed/pdf' subdirectory.
We will have pdfs that are produced in other ways (e.g., bulk upload redactions).
I suggest 'submissions_processed/convert_docx_to_pdf' just to be super clear where they came from.

migration/migrator/migrations/course/20240722105042_submissions_processed_dir.py
```diff
@@ -0,0 +1,49 @@
...
+    :type config: migrator.config.Config
+    :param database: Object for interacting with given database for environment
+    :type database: migrator.db.Database
+    :param semester: Semester of the course being migrated
+    :type semester: str
+    :param course: Code of course being migrated
+    :type course: str
+    """
+    course_dir = Path(config.submitty['submitty_data_dir'], 'courses', semester, course)
+    submissions_processed_dir = Path(course_dir, 'submissions_processed')
```
bmcutler, line 21: This migration makes the submissions_processed directory for existing courses, but for new courses on existing systems or for brand new from scratch installations, the submissions_processed directory won't be created.  

You also need to edit:

emacs sbin/create_course.sh

bmcutler (approved):

PR #10678, opened 2024-07-08
[Feature:InstructorUI] add sloc in database

ziesski:
This is initial step of implementing source line of code(sloc) to submitty.
It is added to database with the course migration. 
Constraint is added for extra layer of protection

PR #10708, opened 2024-07-12
[Bugfix:System]  sloc drop constraint fix

ziesski:
Current migration #10678 has some issue, it does not check if the constraint already exists.
This is a problem if system goes over some migration up & rollback, and fails with error "sloc_non_negative" constraint already exist. To prevent this happening we need to check if the constraint exists. 
Hoewever, PostgreSQL does not support ADD CONSTRAINT IF NOT EXISTS 
So one of the many option is to have DROP IF EXISTS before ADD regardless.
Yes it is expensive operation but we are OK with it.
You can find some reference in other migrations that we did (ex: user_last_initial_format.py)


williamjallen:
Can you please update the PR description to include more information about what specifically is wrong, and link to the PR which caused this?

williamjallen (approved):
Looks reasonable to me.

PR #10715, opened 2024-07-15
[Feature:System] Add support for Vue pages

dorian451:

No Vue support

### What is the new behavior?
This PR makes a new Vue project `site/vue` and adds the necessary infrastructure to load pages made in that project onto the main site, with a `renderVue()` PHP function (same syntax as `renderTwig()`).

### Other information?
This isn't a breaking change, and documentation is still WIP.


site/app/libraries/Output.php
```diff
@@ -561,6 +561,24 @@ public function addInternalJs($file, $folder = 'js') {
...
+
+        return "<script>" .
+                "window.addEventListener('load', () => {" .
+                    "window.submitty.render('[data-vuemount]', '$page', " . json_encode($args) . ");" .
+                "})" .
+            "</script>" .
+            "<div class=\"content\" data-vuemount>" .
+                "<noscript>You need Javascript enabled to see this page.</noscript>" .
+            "</div>";
+    }
```
williamjallen, line 580: I think it would be better to render this via Twig.

site/package.json
```diff
@@ -45,13 +45,15 @@
...
     "shortcut-buttons-flatpickr": "0.4.0",
+    "vue": "^3.4.21",
     "twig": "1.17.1"
   },
   "devDependencies": {
     "@babel/core": "^7.24.3",
     "@babel/preset-env": "^7.24.3",
     "@babel/preset-typescript": "^7.22.11",
     "@cypress/skip-test": "^2.6.1",
+    "@tsconfig/node20": "^20.1.4",
```
williamjallen, line 56: Why is this necessary?

williamjallen (commented):
I think this is definitely on the right track overall.  In addition to the inline comments I left, I think all of the vue stuff should reside in `site/vue`.  `-frontend` is redundant, and embedding 3 in the directory name leaves a lot of technical debt to work through at a future point whenever we upgrade to the next major version of Vue.

site/package.json
dorian451: For vue's tsconfig

site/app/libraries/Output.php
dorian451: Implemented in https://github.com/Submitty/Submitty/pull/10715/commits/9a370b71dafbb0cf17c7779b9dd95d5a33ff9b0a

dorian451:
> I think this is definitely on the right track overall. In addition to the inline comments I left, I think all of the vue stuff should reside in `site/vue`. `-frontend` is redundant, and embedding 3 in the directory name leaves a lot of technical debt to work through at a future point whenever we upgrade to the next major version of Vue.

Implemented in https://github.com/Submitty/Submitty/pull/10715/commits/3275557e5355d3c52c642513f4464e970a7bafdc, https://github.com/Submitty/Submitty/pull/10715/commits/a08471d37d2bdd68bcfe1d10b9d537d170561039, and https://github.com/Submitty/Submitty/pull/10715/commits/fdd507a74691bedaa7600ad2cb93ebbb994438d0

site/app/templates/Vue.twig
```diff
@@ -0,0 +1,8 @@
+<script> 
+    window.addEventListener('load', () => {
+        window.submitty.render('[data-vuemount]', "{{ page }}",  {{ args | json_encode() | raw() }});
+    })
```
williamjallen, line 4: You could use jQuery to do this instead.

site/app/templates/Vue.twig
```diff
@@ -0,0 +1,8 @@
+<script> 
+    window.addEventListener('load', () => {
+        window.submitty.render('[data-vuemount]', "{{ page }}",  {{ args | json_encode() | raw() }});
```
williamjallen, line 3: It would be better to use the `js` filter instead of `raw` to prevent XSS attacks.

site/vue/src/main.ts
```diff
@@ -0,0 +1,29 @@
...
+import Unknown from './pages/Unknown.vue';
+
+const exports = {
+    async render(target: string, page: string, args: Record<string, unknown> = {}) {
+        const root_component = await (async () => {
+            try {
+                return (await import(`./pages/${page}.vue`) as { default: Parameters<typeof createApp>[0] }).default;
+            }
+            catch {
+                return Unknown;
```
williamjallen, line 11: Should we also log something to the console?

williamjallen (changes requested):
This seems to be on the right track.  I plan to approve once the minor inline changes are made.  It would be good to @bmcutler to take a look at the file permissions and directory structure stuff at least.  It would also be good for someone else with significant Vue (or other frontend framework) experience to take a look.

site/app/templates/Vue.twig
dorian451: I can't find this `js` filter, as Twig says it doesn't exist and it does not appear on the [documentation website](https://twig.symfony.com/doc/3.x/). If you meant to say to use the `escape` filter with the argument `'js'` (`e('js')`), that would defeat the entire point of why I'm using `raw`. In order for this to work, I need to pass a javascript object to `window.submitty.render`, and using the `e('js')` filter breaks that. I'm currently unaware of another way of doing this.
If it helps, I believe XSS attacks are impossible here anyway due to `args` being a trusted parameter that only the backend can provide.

site/app/templates/Vue.twig
dorian451: Addressed in https://github.com/Submitty/Submitty/pull/10715/commits/f08c60b4e8100eaf8d93c6ee67e8ad38bed300e9

site/vue/src/main.ts
dorian451: Addressed in https://github.com/Submitty/Submitty/pull/10715/commits/42b22c2d10a175de7e298df3442d5ec19cc8e4fc

site/app/templates/Vue.twig
williamjallen: I forgot that it's not a js filter, but rather an escape filter with js as an option.  Apologies for the confusion.  `... | json_encode | e('js')` should work here though, right?

While there are no *current* XSS issues here, we want to minimize the usage of the `raw` filter throughout the codebase if at all possible, because someone could come along later and accidentally open up something when they don't see the full picture about what's going on.  It's happened before...

dorian451: >I need to pass a javascript object to `window.submitty.render`, and using the `e('js')` filter breaks that.

Using `raw`:
```js
$('document').ready(() => {
    window.submitty.render('[data-vuemount]', "Example",  {"example":"value"});
})
```
Using `e('js')`:
```js
$('document').ready(() => {
    window.submitty.render('[data-vuemount]', "Example", \u007B\u0022example\u0022\u003A\u0022value\u0022\u007D);
})
```

I can attempt to find some other way to pass this data to Vue, or we could just add a comment explaining why it needs to be used here and how exactly it should be used. What would you prefer?

williamjallen: Alright, that makes sense.  I think it would be fine to stick with `raw` for now then.  Perhaps in the future we could create a `json` filter which handles this via some common code path.

williamjallen (approved):
Everything looks good to me.  Thanks for the great work on this!  Hopefully Vue will help us develop dynamic content for efficiently!

PR #10716, opened 2024-07-16
[Feature:Developer] Add colors.css check

powe97:
The `colors.css` file is prone to becoming bloated over time; there are many duplicate and unnecessary colors and the file has grown to be pointlessly large.

### What is the new behavior?
This PR adds a CI check to make sure that new PRs do not increase the size of `colors.css`.

### Other information?
In commit 570f9a1, the check fails because I added a couple lines to `colors.css`.
image
After reverting changes to `colors.css` in 4263c64, the check passes.

### DIFF
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -296,6 +296,28 @@ jobs:
         run: python3 run_shellcheck.py  # Uses the default Python installed with Ubuntu
 
 
+  colors-css:
+    name: colors.css check
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Fetch main branch
+        uses: actions/checkout@v4
+        with:
+          ref: main
+          path: MAIN
+      - name: Fetch PR branch
+        uses: actions/checkout@v4
+        with:
+          path: PR
+      - name: Check file size change
+        run: |
+          FILEPATH="site/public/css/colors.css"
+          LINES_DIFF="$(($(wc -l < PR/$FILEPATH) - $(wc -l < MAIN/$FILEPATH)))"
+          if test $LINES_DIFF -gt 0; then
+            exit $LINES_DIFF
+          fi
+
+
   Cypress-System:
     runs-on: ubuntu-22.04
     services:
--- a/.github/workflows/submitty_ci.yml
+++ b/.github/workflows/submitty_ci.yml
@@ -331,10 +331,11 @@ jobs:
           LINES_DIFF=$(($(wc -l < "$PR_FILE") - $(wc -l < "$MAIN_FILE")))
 
           if [ $LINES_DIFF -gt 0 ]; then
-            echo "Line difference: $LINES_DIFF"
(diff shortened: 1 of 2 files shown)


zacharymnp (approved):
Works as intended.

.github/workflows/submitty_ci.yml
```diff
@@ -296,6 +296,28 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          path: PR
+      - name: Check file size change
+        run: |
+          FILEPATH="site/public/css/colors.css"
+          LINES_DIFF="$(($(wc -l < PR/$FILEPATH) - $(wc -l < MAIN/$FILEPATH)))"
+          if test $LINES_DIFF -gt 0; then
+            exit $LINES_DIFF
+          fi
```
ziesski, line 318:
```suggestion
  colors-css:
    name: colors.css Check
    runs-on: ubuntu-22.04

    steps:
      - name: Checkout main branch
        uses: actions/checkout@v4
        with:
          ref: main
          path: MAIN

      - name: Checkout PR branch
        uses: actions/checkout@v4
        with:
          path: PR

      - name: Compare file size
        run: |
          FILEPATH="site/public/css/colors.css"
          MAIN_FILE="MAIN/$FILEPATH"
          PR_FILE="PR/$FILEPATH"

          if [ ! -f "$MAIN_FILE" ]; then
            echo "Main file $MAIN_FILE does not exist."
            exit 1
          fi

          if [ ! -f "$PR_FILE" ]; then
            echo "PR file $PR_FILE does not exist."
            exit 1
          fi

          LINES_DIFF=$(($(wc -l < "$PR_FILE") - $(wc -l < "$MAIN_FILE")))

          if [ $LINES_DIFF -gt 0 ]; then
            echo "Line difference: $LINES_DIFF"
            exit $LINES_DIFF
          else
            echo "No lines added."
          fi
```

ziesski: Few word changes, also should check if PR tries to remove colors.css entirely? (check will pass if we rm all).

ziesski (changes requested):

.github/workflows/submitty_ci.yml
powe97: Sounds good

ziesski (approved):
Havent tested functionality, but code looks good.

bmcutler:
@powe97 

I don't see a clear & descriptive error message in the code 
And when I look at the commit that errors for this action, I don't even see a message about line count.  (I clicked into the test, and clicked around a bit, but I'm only seeing an unrelated deprecation error)

If we're going to add this, it needs to clearly state the expectation (and maybe the WHY) and what the developer should do to correct the failure.
(And also have a message that this failure can be ignored/waived by the maintainer if the additions are necessary and appropriate relative to the PR purpose.)

Please take a screenshot of how the error will appear on a failed commit..

powe97:
@bmcutler I've added an error message and new test commit.

bmcutler (approved):
thanks.  i like the error text.  looks good

PR #10719, opened 2024-07-16
[Feature:HelpQueue] Add cookie for office hour sliders

### LINKED ISSUE #10590: Add cookie for enable notifications/notification Sounds in office hour quene

Right now `enable notification` and `notification sound` switches in office hour quene does not save which means everytime you refresh/leave the page they are back to default.
Cookie for them should be added after PR #10564(Change notification buttons to switch) is merged.

Add cookies to save user's choices.

ZheyuDeng:
Fix #10590 

There are currently no cookies for office hour notification sliders. So when TA leave/refresh the page or even remove/restore a student from the queue it will goes back to default.

### What is the new behavior?
Now choices will be remembered.
However the broswer wouldn't allow sound to be played without interaction.
So if TA makes any action sound will be played as expected. Will make a separate issue about this.

https://github.com/user-attachments/assets/34dd2915-a9c1-4789-859a-987acc82d733


site/app/templates/officeHoursQueue/QueueFooter.twig
```diff
@@ -146,6 +146,22 @@
...
+      var push_notifications_enabled = pushNotificationState === 'true' || localStorage.getItem('office_hour_queue_notifications') === "true";
+
+      if ("Notification" in window && Notification.permission === "granted") {
+          setNotificationState(push_notifications_enabled);
+      }
+      else {
+          setNotificationState(false);
+      }
+
+      document.getElementById('push-notification-switch').checked = push_notifications_enabled;
```
zacharymnp, line 162: You should probably replace the ~4 uses of `document.getElementById()` that you added in this PR with JQuery.

zacharymnp (changes requested):

zacharymnp (commented):
Adding the `autoplay` attribute to the audio element in GlobalHeader.twig (~line 46) might fix the issue with the page requiring user interaction before playing sounds. It seems to work, although the behavior seems a little odd.

ZheyuDeng:
> Adding the `autoplay` attribute to the audio element in GlobalHeader.twig (~line 46) might fix the issue with the page requiring user interaction before playing sounds. It seems to work, although the behavior seems a little odd.

I think the auto play is just letting the sound play whenever things load instead of manually control it. I still get the DOM when sound is played.

zacharymnp (approved):
In that case, this works as well as it can at the moment.

williamjallen (approved):
Code looks reasonable to me, given the style of the surrounding code.  I did not test this manually.

It would be great if we could move most of the JS logic in this file to an external file so it gets linted automatically.  That should be done in a separate PR though.

